PR #69919, opened 2026-07-10
feat(engineering-analytics): render stat tiles and cost trend with quill charts

pauldambra:
## Problem

The engineering analytics scenes hand-roll their stat tiles (`MetricTile`, `StatCard` with bespoke delta badges) and render the cost-per-merged-PR trend through the legacy Chart.js `Sparkline` wrapper. We're standardizing on the quill design system, so these surfaces should use `@posthog/quill-charts` like web analytics and mcp_analytics already do.

## Changes

- `MetricTile` and `StatCard` now render through quill's `MetricCard` (title/headline/change-pill/subtitle), keeping the product's card chrome, no-data `—` state, and delta semantics (`pp`/`h` units, good-when-down direction, ±0 suppression, >999% clamp) in a thin wrapper. `DeltaBadge` stays for table cells.
- The cost-per-merged-PR chart on the repo overview is now a quill `TimeSeriesLineChart`: the logic selector emits ISO bucket labels + interval so the chart owns date tick/tooltip formatting, with a currency y-axis and the existing compact-USD tooltip values.
- `MasterStatusTile` typography updated to match `MetricCard` so the stat strip stays visually uniform.
- `valueSuffix` folded into headline/caption at the few call sites that used it.

Not migrated (no quill equivalent yet): `RunActivityChart` (scatter + concurrency band + brush), `LifecycleStrip` (timeline), `FailureSparkline` (table-cell stacked bars), `ShareRow` (leaderboard rows).

## How did you test this code?

- `jest --testPathPattern='products/engineering_analytics'` — 7 suites, 116 tests pass.
- `tsgo --noEmit`: no new errors in the changed files (this environment lacks generated kea `*Type.ts` artifacts, so pre-existing implicit-any noise remains repo-wide).
- oxlint/oxfmt clean.
- Not manually verified in a running app — reviewers should eyeball the repo overview stat strip, PR/workflow detail tiles, PR-list/test-health filter cards, and the cost chart.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

None needed — internal UI component swap, no behavior/API change.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

- Built with Claude Code (PostHog Code cloud task). Assessed all engineering analytics visualizations against `@posthog/quill-charts`, then migrated the two 'directly replaceable today' surfaces: stat tiles → `MetricCard` and the cost trend → `TimeSeriesLineChart`.
[...]

PR #69516, opened 2026-07-08
feat(eng-analytics): redesign hub, prs, workflows + detail pages

rnegron:
## Problem

The engineering analytics read-only UI (the hub, the Pull requests page, and the author page) read as a stack of trimmed-down copies of the dedicated tables. Headline metrics were undifferentiated stat-tiles, scope and loading behavior differed page to page, and several tables carried filler subtitles and redundant controls.

## Changes

**Hub**
- Pipeline-verdict hero (master status + open-PR backlog pressure) above the date filter; a now/window two-zone split so the filter visibly governs only what it scopes.
- Headline metrics are stat + sentiment-sparkline trend cards: Pass rate, PR time to green, Median open→merge, Cost per merged PR.
- Per-commit master-health mini-bars (height = CI duration, colored by verdict) with click-to-commit; 5-row PR + workflow preview tables (Show more / View all).

**Pull requests page**
- The three big stat cards became a compact click-to-filter lens strip: Open, Failing CI, Stuck >7d, plus two lenses the control bar can't express in one click: CI thrash (re-run cycles) and Ready (open, non-draft, green CI).
- Dropped the repo and author filters (the source chip and search already cover them); a Created column ("27h ago") defaults to newest-first, replacing Open time; page size 25; rows reach GitHub only via the `#123` link.

**Author page**
- Date picker moved into the scope bar, consistent with every other page; it now scopes the PR list too, not just the cost tiles.
- CI cost reframed as cost per iteration loop (a loop is one push and its CI), not a running total.
- "Where their CI minutes go" rebuilt as a full-width, single-hue ranked bar chart (was a half-width rainbow); a clear skeleton loading state when the range changes.

**Per-push CI history (PR detail + PR tables)**
- The PR lifecycle strip is now a lifecycle × bar-chart hybrid: one bar per push — height is that push's wall-clock CI time on a shared scale, color its verdict — with duration sublabels.
- PR tables gained a push-history sparkline in the Pushes column (per-push duration, red/green/amber), backed by a new `push_history` series on the PR list endpoint (verdicts collapse latest-run-per-workflow, same rule as `ci_rollup`).
- PR header tiles: Latest push · CI wall time (new) · Pushes · CI cost, with skeletons while runs load.

**Workflows tab + detail pages**
- RESTful GitHub-shaped URLs (`/repos/:owner/:name/pull-requests/:n`, `/actions/runs/:id`, `/actions/workflows/:name`); drill-downs and back links carry the window scope.
[...]

PR #70555, opened 2026-07-13
fix(engineering-analytics): declare the @posthog/quill-charts dependency

sampennington:
## Problem

`products/engineering_analytics` imports `@posthog/quill-charts` in three frontend files (`RepoOverviewScene`, `MetricTile`, `StatCard` — added in #69919 / #69516) but never declared it in its `package.json`. With pnpm's isolated node_modules the import isn't resolvable from that package. It has worked by accident on existing checkouts, where leftover root `node_modules/@posthog/quill-*` symlinks from older installs satisfied the resolver.

On a clean `pnpm install` those stale links are pruned and local dev breaks hard: vite's dependency scan fails with

```
(!) Failed to run dependency scan. Skipping dependency pre-bundling.
  @posthog/quill-charts (imported by .../products/engineering_analytics/...)
```

With pre-bundling skipped, vite discovers dependencies at runtime instead — every page load triggers "optimized dependencies changed. reloading", the forced reload interrupts React mid-mount, and the app dies with `NotFoundError: Failed to execute 'removeChild' on 'Node'` on effectively every navigation.

## Changes

One line: add `"@posthog/quill-charts": "workspace:*"` to `products/engineering_analytics/package.json` dependencies (plus the lockfile entry), matching how `products/metrics` and `products/revenue_analytics` declare the same dependency.

- Reproduced the failure on a clean install: `require.resolve('@posthog/quill-charts', ...)` from the product directory fails, and the vite dev server logs the failed dependency scan followed by a reload loop.
- After the fix + `pnpm install`: resolution succeeds, the dependency scan completes ("scanning dependencies…" → "bundling dependencies…" with no error), and a headless browser loads the app with no console errors.
- No behavior change to any built artifact — production builds resolved this through the bundler workspace config all along, which is why CI never caught it; only dev serve on a clean install breaks.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed — dependency declaration only.

## 🤖 Agent context

- Authored by Claude in a PostHog Code session while debugging Sam's repeatedly-crashing local dev stack; the crash was chased through two red herrings (stale lockfile, stale quill dist builds) before a clean `pnpm install` made the failure deterministic and the failed vite dependency scan named this import.
- Also cross-checked every other `products/*` package for quill imports vs declarations — `engineering_analytics` is the only offender.

---

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3238,6 +3238,9 @@ importers:
       '@posthog/icons':
         specifier: 'catalog:'
         version: 0.38.0(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+      '@posthog/quill-charts':
+        specifier: workspace:*
+        version: link:../../packages/quill/packages/charts
       '@storybook/react':
         specifier: 'catalog:'
         version: 10.4.6(@types/react-dom@18.3.7(@types/react@18.3.27))(@types/react@18.3.27)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(storybook@10.4.6(@testing-library/dom@10.4.0)(@types/react@18.3.27)(prettier@3.8.2)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(typescript@6.0.3)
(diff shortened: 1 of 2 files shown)


rnegron (approved):
oops, thank you!

PR #70184, opened 2026-07-10
feat(capture): emit v2 ingestion warnings from capture validation drops

eli-r-ph:
## Problem

Capture silently drops events that fail v1 validation (missing name, oversized distinct_id, bad timestamp, duplicate UUID), and customers get no visibility into those drops.
This gives capture a database-free way to report them: it emits a `$$client_ingestion_warning` event onto the existing topic, and the Node.js consumer (#70557) resolves token to team and writes the row.

Stacked on #70557. Ships dark behind `CAPTURE_INGESTION_WARNINGS_ENABLED` (default off).

> [!NOTE]
> Correct typing depends on #70557; until it deploys the consumer still stores each warning, just under the generic `client_ingestion_warning` type. Capture defaults off, so there's no rollout hazard.

```mermaid
flowchart LR
    B1[capture v1 validation drop] --> B2{{common-ingestion-warnings: throttle + envelope}}
    B2 --> B3[client_ingestion_warning topic]
    B3 --> B4[nodejs consumer: resolves token to team]
    B4 --> B5[ingestion_warnings_v2]
    classDef phBlue fill:#1d4aff,stroke:#1d4aff,color:#fff;
    classDef phGray fill:#e5e7eb,stroke:#c7ccd1,color:#000;
    classDef phYellow fill:#f9bd2b,stroke:#f9bd2b,color:#000;
    class B2 phBlue;
    class B3 phGray;
    class B1,B5 phYellow;
```

## Changes

New service-agnostic crate `rust/common/ingestion_warnings` (depends only on `common-types` plus kafka/metrics libs, no service crate).

- `registry.rs`: `WarningType` is codegen'd from the Node registry at build time, and `from_tag` is a hand-written allowlist mapping capture error tags to types (unregistered tags emit nothing).
- `build.rs` + `capture_warning_types.generated.json`: `build.rs` generates the enum from a committed JSON mirror of the Node source of truth, so the Rust types can't drift.
- `serializer.rs`: builds the envelope with type/source/details in properties, where top-level `distinct_id` is always the token (never the offending id, since `distinct_id_too_large` is itself a warning type) and the real id rides in `details`.
- `throttle.rs`: per-`(token, type)` limiter (1/hour, burst 1, per pod) with a 100k-key cardinality cap that fails open, bounding worst-case memory under a spoofed-token flood.
[...]

PR #70557, opened 2026-07-13
feat(ingestion): preserve structured warning type/source from backend producers

eli-r-ph:
## Problem

Capture (Rust) can't resolve token to team, so it can't write the v2 warning row directly, but it can produce a `$$client_ingestion_warning` event that the Node.js consumer already turns into a row.
That consumer hard-codes every such warning as `client_ingestion_warning` / `info` / `plugin-server`, flattening any structured type a backend producer sends.

## Changes

Node.js foundation for backend-produced ingestion warnings; backward compatible and lands before the capture emitter (#70184).

- `handle-client-ingestion-warning-step.ts` now preserves a structured warning when the event carries `$$client_ingestion_warning_type` (plus optional `_details` / `_source`), else falls back to the legacy client-warning shape.
- Structured warnings skip `alwaysSend`, so the existing per-`(team, type)` limiter is the sole gatekeeper for backend volume.
- The producer's ids (the offending event) win over the synthetic event's ids, and `category` / `severity` always come from the registry, never the envelope.
- Registered the 12 capture validation-drop types in `INGESTION_WARNING_TYPES` and restricted capture-produced types to a backend allowlist.
- `IngestionWarningsView.tsx` describes the new capture validation-drop types in the UI.
- Threaded an optional `source` through the v2 serializer, defaulting to `plugin-server`.

Automated only (I'm Claude, agent-assisted); added Jest cases to `handle-client-ingestion-warning-step.test.ts`.

- Structured type/source/details preserved, producer ids win, and category/severity come from the registry.
- Unrecognized structured type falls back to `client_ingestion_warning` + `plugin-server`.
- Suite passes locally (8/8); the ClickHouse-backed integration path was not run.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change beyond internal warning-type additions.

## 🤖 Agent context

- Standalone consumer half of the capture ingestion-warnings stack; the capture emitter and the Rust/Node type-list codegen land in stacked #70184.
- We pivoted from a dedicated capture-to-ClickHouse write path to reusing this consumer (which already resolves token to team), dropping the planned ClickHouse schema and Django read-side changes.


mayteio (approved):
LGTM, please check bot comments before merging 🔥

jose-sequeira (approved):

PR #67249, opened 2026-07-01
feat(data-management): bulk edit tags on event definitions

jakesciotto:
## Problem

Event definitions could only be tagged one at a time — tagging a batch of events (say, marking a set `deprecated` or `pii`) meant opening each definition individually. Feature flags, insights, and dashboards already support bulk tag editing; this brings event definitions to parity.

## Changes

- **Frontend:** the event definitions table (`Data management → Events`) now has row selection. Selecting rows reveals the shared **Update tags** action (add / remove / replace) already used by flags, insights, and dashboards. On success the affected rows update in place (including the logic's URL cache) so tags reflect immediately without a refetch.
- **Backend:** `EventDefinitionViewSet` gains a `bulk_update_tags` action.

The interesting part is why this needed a viewset-specific override rather than reusing `TaggedItemViewSetMixin.bulk_update_tags` directly:

- **UUID PKs.** The shared action validates `ids` as integers; event definitions are keyed by UUID. A UUID-aware request/response serializer variant handles this.
- **Access model.** The shared action filters each object through object-level RBAC (`access_level_satisfied_for_resource`). `event_definition` isn't a registered access-controlled resource, so that filter would reject *every* event definition for *every* user. The override scopes by project (`team__project_id`) and relies on project membership — the same boundary the single-object `PATCH` already uses.
- Tags live on the base `EventDefinition` row (multi-table inheritance shares the PK with `EnterpriseEventDefinition`), so the override operates on `EventDefinition.objects` and works regardless of the enterprise extension.

The shared add/remove/set logic is extracted into `apply_bulk_tag_changes` so the mixin and the override don't diverge. `scope_object_write_actions` opts the action into personal-API-key `event_definition:write` access.

## How did you test this code?

I'm an agent, and this session ran in an environment without the full Python/JS toolchain, so I could not execute the suites locally — CI will. What I did run: **ruff** (lint + format) passes on all changed Python files.

Automated tests added — `posthog/api/test/test_event_definition.py`:
- `test_bulk_update_tags_with_uuid_ids` — guards the UUID-PK + access-model regressions: if the override is dropped and the inherited mixin runs, UUID ids 400 on the integer serializer or every object gets filtered out as inaccessible.
- `test_bulk_update_tags_ignores_event_definitions_in_other_project` — project-scoping / IDOR guard: a definition in another project is reported "Not found" and left untouched.

[...]

PR #70561, opened 2026-07-13
feat(feature-flags): bulk enable/disable flags

jakesciotto:
## Problem

Feature flags could be enabled or disabled one at a time only. The flags table already supports bulk selection (bulk delete, bulk tag editing), but flipping the status of a batch meant opening each flag individually. This brings status changes to parity with the other bulk actions, in the same vein as the recent event-definitions bulk tag editing work.

## Changes

- **Backend:** `FeatureFlagViewSet` gains a `bulk_update_status` action. It accepts either explicit `ids` or list-endpoint `filters` (mutually exclusive, mirroring `bulk_delete`), scopes by project, skips flags already in the target state, and refuses to enable archived flags.
- **Logging:** `queryset.update()` bypasses `ModelActivityMixin`, so the action writes one "updated" activity entry per changed flag via `bulk_log_activity`, recording the `active` before/after. The History tab now shows a bulk status change the same way it shows a single-flag toggle.
- **Frontend:** the flags table selection bar gets **Enable** and **Disable** buttons next to the existing tag/delete actions. Each opens a confirmation dialog (enabling rolls out immediately, disabling rolls back), then reports the result with a toast. "Select all matching" resolves to filters so we don't ship an unbounded id list.

Generated frontend/MCP types for the new endpoint are left for CI's `check-openapi-types` job to regenerate and auto-commit. The button calls `api.create()` directly and doesn't depend on the generated client.

> [!NOTE]
> No screenshot: this session had no dev stack to run the app. The buttons render in the same selection bar as the existing "Delete selected" button, only when rows are selected.

I'm an agent and this session had no local Postgres, no `node_modules`, and no `hogli`, so I could not run pytest, Jest, or the type/lint pipeline locally. CI will. What I did run: **ruff** (check + format) passes on both changed Python files, and the module imports cleanly under the project venv (blocked only at the DB connection, i.e. past all import/reference resolution). I also confirmed the activity-log encoder round-trips the boolean `before`/`after` so the logging assertions hold.

Automated tests added — `TestFeatureFlagBulkUpdateStatus` in `products/feature_flags/backend/api/test/test_feature_flag.py`. Each targets a status-specific regression not covered by the existing `bulk_delete` suite:
- `flips_active_and_skips_noops` (parameterized enable/disable) — guards the core flip plus the no-op skip (a flag already in the target state must not be re-written or reported).
- `logs_active_change_per_updated_flag` — the "logging that goes with it": if the manual `bulk_log_activity` call is dropped, the bulk path silently stops writing history. Asserts one `active` change entry per changed flag.
- `cross_project_isolation` — IDOR guard: a flag in another project is reported "Flag not found" and left untouched.
- `cannot_enable_archived_flag` — safety rule: a bulk enable must not silently roll out an archived flag.
- `by_filter_only_updates_matching_flags` — wiring guard that the filters branch (used by "select all matching") works for status, not just ids.
- `invalid_request_returns_400` (parameterized) — wiring guard for the manual validation (missing/non-boolean `active`, neither/both of ids+filters).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover per-flag status toggling that would need updating for the bulk equivalent.

## 🤖 Agent context

Built with Claude Code (Opus 4.8). Skills invoked while shaping this change: `/improving-drf-endpoints` (viewset/serializer + schema annotations), `/adopting-generated-api-types` (frontend API usage), and `/writing-tests` (the value gate — kept coverage to the six status-specific regressions above).

Key decisions:
- Modeled the action on the sibling `bulk_delete` rather than inventing a new shape — same ids/filters duality, same project scoping, same access-control filter, same batched-update + manual-cache-invalidation structure. The point of "same vein" is that a reviewer reads the two actions side by side.
- Matched the sibling `bulkDeleteFlags` in `flagSelectionLogic` (raw `api.create` + a small handwritten result type) instead of the generated client, since the generated function won't exist until CI regenerates types and nothing else imports it.
- No `report_user_action` call — `bulk_delete` doesn't emit one either, so "logging" here means the activity/audit log, which is what the flag History tab reads.

---

### DIFF
--- a/frontend/src/scenes/feature-flags/FeatureFlags.tsx
+++ b/frontend/src/scenes/feature-flags/FeatureFlags.tsx
@@ -395,8 +395,8 @@ export function OverviewTab({
 
     const { currentProjectId } = useValues(projectLogic)
     const { paramsFromFilters } = useValues(featureFlagsLogic({}))
-    const { bulkDeleteResponseLoading } = useValues(flagSelectionLogic)
-    const { bulkDeleteFlags } = useActions(flagSelectionLogic)
+    const { bulkDeleteResponseLoading, bulkUpdateStatusResponseLoading } = useValues(flagSelectionLogic)
+    const { bulkDeleteFlags, bulkUpdateFlagStatus } = useActions(flagSelectionLogic)
 
     const [matchingFlagIds, setMatchingFlagIds] = useState<readonly number[] | null>(null)
     const [matchingFlagIdsLoading, setMatchingFlagIdsLoading] = useState(false)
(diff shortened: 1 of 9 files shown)


jakesciotto:
Closing — this was opened against the wrong product. The request was for bulk **status** updates on event definitions (Data management → Events), in the same vein as #67249 (bulk tag editing), not feature flags. Opening a fresh PR for the correct change on the event-definitions table.

PR #70567, opened 2026-07-13
feat(data-warehouse): show managed warehouse data status in data ops

EDsCODE:
## Problem

Once a team has a managed warehouse, there is no way to tell whether their data has actually landed in it. Events and persons are backfilled month by month, imported source tables are copied in chunks and then kept fresh by applied batches, and none of that progress is visible in the product. Users are left guessing whether an empty query result means "no data" or "not copied yet".

There is also no signal when a backfill is stuck. A historical partition that fails repeatedly is not retried, and a source table's import queue can stall - both are invisible today.

## Changes

A new Overview tab in Data ops that reports warehouse readiness:

- **Events and persons** - readiness state, historical backfill progress (completed / total partitions), the partition currently running, and when the status last changed.
- **Imported source tables** - per-table warehouse state, backfill chunks copied, imported batches still pending, and the last upstream sync.
- A rolled-up **overall state** at the top, so "everything is fine" or "something needs attention" reads at a glance.

Readiness is expressed as one of `not_configured`, `waiting`, `backfilling`, `catching_up`, `up_to_date`, `needs_attention`, `unknown`, served by a new `managed-warehouse-data-status` endpoint on the data warehouse viewset and backed by a new `ManagedWarehouseBackfillPartition` model that the Dagster backfill job writes to. The tab polls every 15s while work is in flight and every 60s once things are stable.

Tab availability is now driven by whether a warehouse exists, not just by feature flags:

> [!NOTE]
> A project with no ready warehouse sees **no tab bar at all** - just the setup form. Overview and Settings only appear once a warehouse reaches `ready`. Previously an unprovisioned project landed on a tab whose only possible content was "Not configured", with the setup form hidden behind Settings.

The status is `null` both before it loads and when no warehouse exists, so resolution is tracked explicitly rather than inferred from the value, and the tab bar is held until it lands (otherwise a provisioned project flashes the setup form and then swaps in tabs). Clamping the requested tab moved out of `urlToAction`, which runs before the status arrives and would strand a `?tab=overview` request on Settings, into the `activeTab` selector, which re-derives once the tab set settles.

Automated, run by me (Claude):

- `dataWarehouseSceneLogic.test.ts` - 5 cases, all passing. Three parameterized over provisioning state (none / provisioning / ready) assert the exposed tabs. These catch the regression this PR fixes: an unprovisioned project must not be shown Overview. One asserts the tab set stays unresolved until the status lands, guarding the flash-then-swap. The last requests a tab before the status resolves and asserts it survives - that guards the `urlToAction` clamping race, which no existing test covered.
- `test_managed_warehouse_data_status.py` - readiness derivation for the events, persons, and source paths.
- Typecheck and lint clean on the changed frontend files.

Manual, in a local dev instance: I drove the running app with a headless browser and mocked the status endpoint to render each readiness state (backfilling, up to date, needs attention, not configured, loading, failed to load), plus both provisioning paths - unprovisioned renders the setup form with no tab bar, `ready` renders the tabs and lands on Overview. Screenshots were captured but not uploaded to the public PR assets repo, so this description has none. Happy to add them.

Not tested: a real end-to-end backfill writing `ManagedWarehouseBackfillPartition` rows from Dagster. The readiness states were exercised against mocked payloads and unit-level fixtures, not a live backfill.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed - the feature is behind `data-warehouse-scene` and not yet documented.

## 🤖 Agent context

The Overview tab, the status endpoint, the readiness derivation, and the partition model were written by Codex. I (Claude Code, Opus 4.8) reviewed the result against a running local instance and found the scene still exposed tabs purely off feature flags, so an unprovisioned project landed on an Overview that could only ever say "Not configured" while the setup form sat behind another tab. The second commit fixes that.

Two things I tried and rejected along the way. First, a dedicated feature flag for the Overview tab: it would let the tab roll out independently of the scene, but the real gate turned out to be warehouse existence, not a flag, so it was dropped. Second, clamping the requested tab inside `urlToAction` against `availableTabs` (the existing approach): it races the warehouse status load now that the tab set depends on it, so clamping moved into the `activeTab` selector where it re-derives.


EDsCODE:
Superseded by a stacked split: #70571 (migrations, model, readiness logic, status endpoint, generated types) and #70572 (Data ops Overview tab, stacked on top). Same final tree, reviewable in two halves. Closing this one.

PR #70580, opened 2026-07-13
feat(alerts): add email as a shared alert destination type (7/7)

MattPua:
## Problem

Alerting products want email notifications, and each does it differently today: insight alerts send direct SMTP via `EmailMessage` with their own Django templates and subscriber logic; logs alerts have no email at all. Same story as Slack/Teams/webhook before this stack: per-product delivery forks.

## Changes

Email becomes the fourth shared destination type in `common/alerting/destinations.py`:

- `build_email_destination_config` creates a `template-email` HogFunction wired to the same `alert_id` internal-event filter as the other three types, so dispatch, deletion, and ownership checks work identically with zero glue changes.
- Content renders from the same `EventKindSpec` vocabulary: `email_subject` / `email_html` / `email_text` lay out header, prose `body_lines`, detail rows, and action buttons. Products own what the email says, the platform owns the structure and flow.
- The stored input carries no `templating` key, which the CDP runtime resolves as hog templating, so the same `{project.url}`-style placeholder strings work across all four destination types (the template's `liquid` is only a UI default).

> [!NOTE]
> This replaces (on adoption) the direct `EmailMessage` path insight alerts use today rather than generalizing it. `template-email` requires the team's email integration to be configured, and the template is currently `status: hidden` / non-free, so product rollout gates on that.

Automated only. Extended `common/alerting/test/test_destinations.py`: the email builder's payload contract (template id, filter linkage, native_email fields, no templating key) and vocabulary rendering (prose lines and extra buttons appear in html and text). No existing test covered any email path. Logs destination tests pass unchanged (no behavior change for existing destinations).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

`adopting-shared-alerting` skill updated in the same PR.

## 🤖 Agent context

- Authored with PostHog Code (Claude); skills invoked: `/writing-tests`.
- Key decision: route email through the existing `template-email` CDP destination instead of generalizing the direct-SMTP path, so all four destination types share one dispatch, filter, and lifecycle contract. Verified the per-input templating override in the nodejs runtime (`hog-inputs.service.ts`) before relying on hog placeholders.

---

### DIFF
--- a/.agents/skills/adopting-shared-alerting/SKILL.md
+++ b/.agents/skills/adopting-shared-alerting/SKILL.md
@@ -53,8 +53,8 @@ The adapter owns the product-shaped translation (model → `AlertSnapshot`, and 
 Pure builders for alert notification destinations, delivered as HogFunctions.
 
 - `EventKindSpec` — describes an alert-event kind (labels, message text) so Slack/Teams/webhook bodies render consistently. Products customize within its declarative vocabulary, never by supplying their own rendering: metric-shaped products use `details` (label/value rows), prose-shaped products use `body_lines` (e.g. one breach description per series for insight alerts), and `extra_buttons` adds actions beyond the primary button. All optional fields default to the standard logs-style layout.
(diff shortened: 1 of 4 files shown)


willwearing:
Closing this PR because the alerting work is being continued elsewhere.

PR #70583, opened 2026-07-13
fix(data-warehouse): expose S3 file-changed errors on warehouse reads

thiagosalvatore:
## Problem

A customer reported getting a lot of "ClickHouse error while executing query." when querying a data warehouse table backed by a self-managed S3 source.
Digging into the underlying exceptions showed the real errors were:

- S3 `InvalidRange` ("The requested range is not satisfiable") on specific parquet files
- "Not a Parquet file (wrong magic bytes at the end of file)" on the same files
- `CANNOT_PARSE_BOOL` from a `!= 'null'` filter on a boolean warehouse column

The first two happen when the files backing a self-managed table are overwritten in place while ClickHouse is reading them: ClickHouse lists the files (caching sizes), then issues ranged GETs, and a replaced file fails the read.
All three error codes map to `InternalCHQueryError`, so users only ever see a generic HTTP 500 with no way to tell the problem is on their side. It looks like a PostHog outage when it isn't.

## Changes

- `posthog/errors.py`
  - New `CHQueryErrorS3FileChangedDuringRead(ExposedCHQueryError)`. S3 errors containing "The requested range is not satisfiable", and `INCORRECT_DATA` errors containing "Not a Parquet file" with a file uri, now surface as a 400 with an actionable message naming the failing file and suggesting not to overwrite files in place.
  - `CANNOT_PARSE_BOOL` (467) now passes through ClickHouse's own message (it's already categorized as a user error), so users can see the offending value instead of a generic 500.
  - The new class is added to `CH_TRANSIENT_ERRORS` so Celery/Temporal retry paths keep treating it as transient (including the hand-listed tuple in `posthog/tasks/calculate_cohort.py`).
- `posthog/hogql/query.py`
  - The interactive HogQL execution path now retries once (1s delay) on transient S3 errors. `CHQueryErrorS3Error` was already classified as transient but only background paths (Celery autoretry, Temporal) retried it; the synchronous `/api/query` path surfaced the first failure. One retry re-lists the S3 files, which recovers when the underlying files finished changing between attempts.

- Extended the existing `wrap_clickhouse_query_error` suite in `ee/clickhouse/test/test_error.py` with four cases: S3 InvalidRange → exposed with the file uri and no stack trace, "Not a Parquet file" → same, plain `INCORRECT_DATA` → still internal (guards against over-exposing), and `CANNOT_PARSE_BOOL` → exposed pass-through. No existing case covered conditional exposure within one error code.
- Added three retry tests to `posthog/hogql/test/test_query.py` (mocked `sync_execute`, mocked sleep): transient S3 error retried once and succeeds, retry exhaustion re-raises, and non-transient errors are not retried. These catch a regression where the retry silently disappears or starts retrying everything.
- Verified end to end locally with the query API: a mocked ClickHouse S3 InvalidRange failure on `POST /api/environments/:id/query` now returns a 400 with the actionable message (previously a 500 with "ClickHouse error while executing query.").
- Ran `ee/clickhouse/test/test_error.py`, `posthog/hogql/test/test_query.py`, and `products/exports/backend/tasks/test/test_failure_handler.py` locally, all green.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs changes needed, error message strings only.

## 🤖 Agent context

- Authored with PostHog Code (Claude). Investigated a customer report by pulling the session's exceptions from error tracking, traced the generic message to the `InternalCHQueryError` fallback in `posthog/api/query.py`, and confirmed the failing parquet files are written by the customer's own pipeline (not our warehouse sync, which writes immutable timestamped snapshot folders).
- Skills invoked: `/writing-tests`.
- Decisions: only the two storage-race signatures are exposed for S3/`INCORRECT_DATA` (other errors under those codes stay internal to avoid leaking infra details); retry lives in the HogQL executor rather than `sync_execute` so inserts and paths with their own retry policies are unaffected; the deeper fix of translating `'null'` string comparisons on non-string warehouse columns to `isNull`/`isNotNull` was deliberately left out as a follow-up.

---


thiagosalvatore:
* **#70583** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/70583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

georgemunyoro (approved):

PR #64997, opened 2026-06-20
feat(agent-platform): per-principal identity, credentials + approvals authority — act as the user across native tools & MCP

benjackwhite:
Gives the agent platform a real per-principal **identity + credential** model, so an agent can act *as the user talking to it* (a Slack / JWT / PostHog principal) instead of relying on post-hoc, session-owner credentials — and rebuilds the tool-**approval** gate around the same principal/agent authority split. Public-by-default; when a capability needs privileged access the agent hands back an OAuth **link**, and on callback we store a per-principal credential.

Two pluggable axes:
- **Principal source** (who's asking) — Slack / JWT / PostHog / … (ingress verifiers).
- **Credential provider** (`IdentityProvider`) — bring-your-own `oauth2` (GitHub/Linear/…) and the **managed PostHog** provider.

## The credential model
- **Killed the Slack→PostHog email-match bridge** — an unverified email isn't an authz signal (master independently removed it too).
- **Storage**: `agent_identity_credential` (Fernet-encrypted, one row per agent_user+provider, with a proven `subject`) + single-use PKCE `agent_identity_link_state`.
- **Generic `Oauth2AuthProvider`** (PKCE auth-code; serves GitHub/Linear/the `dogs` test IdP) + **managed `PostHogAuthProvider`** (`establishesIdentity`, subject from `/oauth/userinfo`). On **promote**, a `{kind:posthog}` provider auto-provisions a *normal, user-consented* `OAuthApplication` for the agent's org (not first-party → the user sees the consent screen) and injects its `client_id` into the live spec.
- **`binding: principal`** — per-asker, the only binding the runtime implements; the `agent` (one shared app credential) seam is now **rejected at the spec schema** until the app-scoped shape lands (was an enum value that validated + promoted then threw at runtime).
- **Subject refactor**: dropped `agent_user.posthog_user_id` for a generic `subject` on the credential; per-asker team-admin resolution reads it.
- Archiving an application now **revokes its linked credentials** (terminal archive → no decryptable bearers left on a retired agent).

## Unified resolution — PostHog is not a special case
Every credential flows through one surface — `ctx.identity.resolve(provider)` → `ok | link_required | unavailable`:
- **Native tools** declare `requires.provider {id, scopes}`; **MCP servers** declare `auth.provider`; **custom tools** declare `requires_identity`. The whole `@posthog/*` surface resolves this way — not just `@posthog/identity-fetch`.
[...]

PR #70584, opened 2026-07-13
fix(agent-platform): remove agent-builder slack trigger, seed prod from CI

dmarticus:
## Problem

Two related gaps in how the canonical Agent Builder deployment tracks this repo:

1. The agent-builder example bundle declares a `slack` trigger (added in #64997), but the canonical Agent Builder deployment has no dedicated Slack app — and per the platform's one-Slack-app-per-agent stance, none exists to point it at. Because the promote gate refuses when a slack trigger's required secrets (`SLACK_SIGNING_SECRET`, `SLACK_BOT_TOKEN`) are unset, the bundle as committed cannot be seeded to a live deployment: `seed.py` fails at the final promote step.
2. Nothing re-seeds the deployment when the repo changes — seeding is a manual `seed.py` run, so the live Agent Builder had silently drifted well behind master (pre-#64997 prompt, stale spec). The agent-builder is deliberately a special snowflake: changes to the platform (kernel skills, the bundle, the seeder) should flow into the live deployment, not wait for someone to remember.

## Changes

**Slack trigger removal:**
- `spec.json`: remove the `slack` trigger (chat + mcp remain; the `slack-manifest` MCP allowlist entry stays — the Agent Builder still helps users set up *their* Slack agents).
- `agent.md`: drop Slack as one of the Agent Builder's **own** surfaces; all guidance about **building Slack-triggered agents for users** is kept.
- `example-agent-builder.test.ts`: the trigger wiring assertion now pins `chat + mcp` and the **absence** of `slack`.

**CI seeding:**
- New `.github/workflows/cd-agent-builder-seed.yml`: runs `seed.py agent-builder` against US prod on master pushes touching the bundle, `seed.py`, or `backend/kernel_skills/`, plus a daily schedule. The schedule matters because kernel skills are injected at freeze by the **deployed** backend — a push-time seed can't pick up a kernel-skill change until the backend deploy carrying it rolls out, so the daily run converges them. Gated with `github.repository_owner == 'PostHog' && vars.CD_DEPLOY_ENABLED == 'true'`, serialized via a concurrency group, fails loudly if the PAT secret is missing.
- `seed.py`: post-freeze churn guard. Bundles with freeze-injected kernel skills never match the on-disk manifest, so the pre-flight no-op check can't fire for them and every run used to re-promote a new live revision even when nothing changed (the scheduled seed would have minted one per day). The guard skips promote only when the frozen `bundle_sha256` AND the frozen spec both match the live revision — the sha covers the materialized bundle files, while the spec (trigger auth, tool approvals) lives on the revision row, so both must be identical before archiving the duplicate draft.
- Bundle `README.md`: document the CI seeding flow, and fix the regression-test section's stale pre-#64997 claims (`spec.mcps` empty, inline native approval gating).

**Secret provisioning: ✅ done** — the `AGENT_BUILDER_SEED_PAT` org secret (a personal API key with `agents:write` scope for the primary-org project) has been created and granted to this repo, so the workflow is runnable on merge. If it is ever missing, the workflow fails loudly (by design) rather than letting prod drift silently.

If Slack for the Agent Builder is wanted later, revert the trigger removal and provision the Slack app + secrets as part of the same rollout.

- Validated `spec.json` parses as JSON. The updated wiring test catches the regression no other test does: re-adding a slack trigger without provisioning its required secrets, which breaks the seeded deployment at promote. Could not execute vitest in this environment (no installed JS deps) — the assertion is a direct swap against the new trigger set and runs in CI.
- The workflow reuses `seed.py` unchanged (stdlib-only, no dependency install), which was exercised against the live deployment during the investigation that produced this PR — including the promote-refusal path the secret-missing guard mirrors. `ci-actionlint` covers workflow syntax.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Bundle-local README updated in this PR (CI seeding flow + stale regression-test claims); no `docs/` pages reference the agent-builder's Slack surface or seeding.

## 🤖 Agent context

- Authored with PostHog Code (Claude). Skills invoked: `/writing-tests` (wiring-test assertion change), `/gating-production-deploys` (the seed workflow mutates prod, so it carries the `CD_DEPLOY_ENABLED` gate).
- Origin: seeding the canonical agent-builder bundle failed at promote on the slack trigger's missing required secrets; the owner decided the Agent Builder should not have a Slack app, and asked for CI to keep the deployment seeded going forward.
- Deliberate scope choices: only the Agent Builder's own Slack surface is removed — the agent.md guidance for authoring Slack-triggered agents and the `slack-manifest` allowlist entry are untouched; the `working-outside-the-console` kernel skill already treats Slack as not-yet-live. The workflow's path filter is targeted (bundle + seeder + kernel skills) rather than all of `products/agent_platform/**` because only freeze-time inputs affect the seeded artifact; the daily schedule covers everything that reaches the agent through the deployed backend instead.

---


.github/workflows/cd-agent-builder-seed.yml
```diff
@@ -0,0 +1,64 @@
+name: Seed agent-builder
+
+# Keeps the canonical Agent Builder deployment in sync with the repo.
+# The agent-builder is deliberately a special snowflake: its bundle source is
```
daniloc, line 4: indeed

daniloc (approved):
go for it

gantoine (approved):
just mendral's comment needs fixing, rest looks good

PR #70588, opened 2026-07-13
feat(insights): extend day-of-week filtering to stickiness, lifecycle, and funnel trends

sampennington:
## Problem

The `daysOfWeek` field on the shared `DateRange` schema filters insights down to events that occurred on selected weekdays (ISO 1=Monday…7=Sunday), evaluated in the project timezone. It replaced the old `hideWeekends` display toggle with a true query-level filter, but so far only trends applies it. Stickiness, lifecycle, and funnel trends ignore the field.

## Changes

Wires the existing `QueryDateRange.day_of_week_filter_expr` helper into three more runners, one commit each. Absent or empty `daysOfWeek` still means unfiltered (the helper returns `None`), so existing queries and HogQL snapshots are unchanged.

- **Stickiness** – filter ANDed into the event `where_clause`. "Days active" now counts activity on the selected days only. The actors query reuses the same events query, and each compare series resolves its own `QueryDateRange` subclass, so the persons list and compare-to-previous are filtered too.
- **Lifecycle** – filter ANDed into the shared `event_filter`, so new/returning/resurrecting/dormant statuses derive from selected-day activity only; the actors query inherits it.
- **Funnel trends** – filter ANDed into `FunnelEventQuery`'s WHERE assembly (events and data-warehouse branches), gated on `funnelVizType == trends`. Regular funnels and time-to-convert are deliberately untouched (dropping mid-sequence events there needs a product decision first). The compare path clones the query with a rebuilt `DateRange`, which silently dropped `daysOfWeek` – it now carries the filter into the previous-period series.

> [!NOTE]
> **Lifecycle keeps excluded-day buckets** (trends drops them for day interval). In lifecycle, "dormant" lands on the day *after* last activity, which can itself be an excluded day (active Monday → dormant Tuesday under a Mondays-only filter). Dropping those buckets would hide real dormant counts, so all buckets stay; new/returning/resurrecting are simply zero on excluded days. A test locks this decision in.

> [!NOTE]
> **Funnel trends filters every funnel event**, not just the entrance: a step-two event on a deselected day counts as a drop-off. This matches the "count only events on selected days" semantics and is asserted in a test.

Paths, retention, regular funnels, and the calendar heatmap remain out of scope, as does the frontend gate that exposes the day chips.

Ran the three touched test files locally against ClickHouse (`test_stickiness_query_runner.py`, `test_lifecycle_query_runner.py`, `test_funnel_trends.py`), plus `test_funnel.py` and `test_funnel_trends_actors.py` to guard the viz-type gate – all green, all HogQL snapshots unchanged. Data-warehouse-fixture tests couldn't run in the sandbox (no local object storage); they rely on CI here.

New parameterized tests per runner, each catching a regression nothing covered before:

- weekday filter changes counts (would pass unfiltered if the WHERE wiring is dropped)
- an event late Sunday UTC that is Monday in the project timezone is included by a Monday filter (catches a runner passing a non-timezone-converted timestamp to the helper)
- empty `daysOfWeek` = unfiltered
- persons list matches the chart (actors query wiring)
- stickiness + funnel trends: compare-to-previous series is filtered (funnels previously rebuilt `DateRange` without `daysOfWeek`)
- lifecycle: excluded-day buckets retained with dormant counts visible
- funnels: steps viz ignores `daysOfWeek` (guards the gate)

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Backend-only; the user-facing docs for day-of-week filtering apply once the frontend exposes it beyond trends.

## 🤖 Agent context

Authored by Claude (PostHog Code session) from a written brief. Skills invoked: `/writing-tests`. Key decisions: gate the funnel filter on `funnelVizType == trends` inside `FunnelEventQuery` rather than in `funnel_trends.py`, since the WHERE exprs are assembled in the shared event query; keep lifecycle's excluded-day buckets (dormant lands on excluded days); preserve `daysOfWeek` in the funnels previous-period query clone, which was found to drop it.

---

### DIFF
--- a/posthog/hogql_queries/insights/funnels/funnel_event_query.py
+++ b/posthog/hogql_queries/insights/funnels/funnel_event_query.py
@@ -263,7 +263,7 @@ class FunnelEventQuery(DataWarehouseSchemaMixin):
         select_from = ast.JoinExpr(table=ast.Field(chain=[table_entity.table_name]), alias=self.EVENT_TABLE_ALIAS)
 
         date_range = self._date_range()
-        where_exprs: list[ast.Expr] = [
+        where_exprs: list[ast.Expr | None] = [
             ast.CompareOperation(
                 op=ast.CompareOperationOp.GtEq,
                 left=ast.Field(chain=["timestamp"]),
@@ -274,10 +274,8 @@ class FunnelEventQuery(DataWarehouseSchemaMixin):
                 left=ast.Field(chain=["timestamp"]),
                 right=ast.Constant(value=date_range.date_to()),
             ),
+            self._day_of_week_filter_expr(ast.Field(chain=["timestamp"])),
         ]
-        day_of_week_filter = self._day_of_week_filter_expr(ast.Field(chain=["timestamp"]))
-        if day_of_week_filter is not None:
-            where_exprs.append(day_of_week_filter)
         where = ast.And(exprs=[expr for expr in where_exprs if expr is not None])
 
         if not skip_step_filter:
--- a/posthog/hogql_queries/insights/funnels/funnel_event_query.py
+++ b/posthog/hogql_queries/insights/funnels/funnel_event_query.py
@@ -13,6 +13,7 @@ from posthog.schema import (
     FunnelExclusionEventsNode,
     FunnelMathType,
     FunnelsDataWarehouseNode,
+    FunnelVizType,
     GroupNode,
     StepOrderValue,
 )
@@ -205,6 +206,7 @@ class FunnelEventQuery(DataWarehouseSchemaMixin):
 
         where_exprs = [
             self._date_range_expr(),
+            self._day_of_week_filter_expr(ast.Field(chain=[self.EVENT_TABLE_ALIAS, "timestamp"])),
             self._entity_expr(skip_entity_filter),
             *self._properties_expr(),
             self._aggregation_target_filter(),
@@ -273,6 +275,9 @@ class FunnelEventQuery(DataWarehouseSchemaMixin):
                 right=ast.Constant(value=date_range.date_to()),
             ),
         ]
+        day_of_week_filter = self._day_of_week_filter_expr(ast.Field(chain=["timestamp"]))
+        if day_of_week_filter is not None:
+            where_exprs.append(day_of_week_filter)
         where = ast.And(exprs=[expr for expr in where_exprs if expr is not None])
 
         if not skip_step_filter:
@@ -635,6 +640,14 @@ class FunnelEventQuery(DataWarehouseSchemaMixin):
                 ),
             ]
         )
+
+    def _day_of_week_filter_expr(self, timestamp_field: ast.Expr) -> ast.Expr | None:
+        # daysOfWeek only applies to the funnel trends visualization. Regular funnels and
+        # time-to-convert are untouched: dropping mid-sequence events has ambiguous semantics
+        # there, which needs a product decision first.
+        if self.context.funnelsFilter.funnelVizType != FunnelVizType.TRENDS:
+            return None
+        return self._date_range().day_of_week_filter_expr(timestamp_field)
 
     def _entity_expr(self, skip_entity_filter: bool) -> ast.Expr | None:
         query, funnelsFilter = self.context.query, self.context.funnelsFilter
--- a/posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
+++ b/posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
@@ -347,7 +347,6 @@ class LifecycleQueryRunner(AnalyticsQueryRunner[LifecycleQueryResponse]):
                     timings=self.timings,
                 )
             )
-            # Days of week: lifecycle statuses are derived from activity on the selected days only
             day_of_week_filter = self.query_date_range.day_of_week_filter_expr(self.timestamp_field)
             if day_of_week_filter is not None:
                 event_filters.append(day_of_week_filter)
--- a/posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
+++ b/posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
@@ -347,6 +347,10 @@ class LifecycleQueryRunner(AnalyticsQueryRunner[LifecycleQueryResponse]):
                     timings=self.timings,
                 )
             )
+            # Days of week: lifecycle statuses are derived from activity on the selected days only
+            day_of_week_filter = self.query_date_range.day_of_week_filter_expr(self.timestamp_field)
+            if day_of_week_filter is not None:
+                event_filters.append(day_of_week_filter)
         with self.timings.measure("properties"):
             if self.query.properties is not None and self.query.properties != []:
                 event_filters.append(property_to_expr(self.query.properties, self.team))
--- a/products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
+++ b/products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
@@ -414,6 +414,11 @@ class StickinessQueryRunner(AnalyticsQueryRunner[StickinessQueryResponse]):
             ]
         )
 
+        # Days of week: "days active" counts activity on the selected days only
+        day_of_week_filter = date_range.day_of_week_filter_expr(ast.Field(chain=["timestamp"]))
+        if day_of_week_filter is not None:
+            filters.append(day_of_week_filter)
+
         # Series
         if isinstance(series, EventsNode) and series.event is not None:
             filters.append(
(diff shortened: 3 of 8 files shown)


products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
```diff
@@ -414,6 +414,11 @@ def where_clause(self, series_with_extra: SeriesWithExtras) -> ast.Expr:
             ]
         )
 
+        # Days of week: "days active" counts activity on the selected days only
+        day_of_week_filter = date_range.day_of_week_filter_expr(ast.Field(chain=["timestamp"]))
```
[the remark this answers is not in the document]
sampennington: This resolves correctly for warehouse tables, including the decoy case. `where_clause` runs against the series' FROM table with the `DataWarehouseEventsModifier` applied, and the database build force-remaps the virtual `timestamp` field to the configured `timestamp_field` even when the source table has its own physical DateTime column named `timestamp` (see `posthog/hogql/database/database.py` lines ~1834-1866: `timestamp_field_is_remapped` wins over the physical column). The unqualified `timestamp` here resolves through exactly the same mapping as the date-range filters two lines up, which is what `test_stickiness_data_warehouse_uses_configured_timestamp_field` (the decoy-column regression test) exercises. A `daysOfWeek` filter on such a table therefore runs `toDayOfWeek` on the configured field (`event_time`), not the decoy column.

posthog/hogql_queries/insights/funnels/funnel_event_query.py
```diff
@@ -273,6 +275,9 @@ def _build_data_warehouse_table_query(
                 right=ast.Constant(value=date_range.date_to()),
             ),
         ]
+        day_of_week_filter = self._day_of_week_filter_expr(ast.Field(chain=["timestamp"]))
```
sampennington: 🤖 **sp-ship** · 🧪 `[test]` · **Blocker**

This data-warehouse trends branch applies the day-of-week filter through a second, independent code path gated the same way on `FunnelVizType.TRENDS`, but every new funnel test uses `EventsNode` and so only exercises `_build_events_table_query`. A regression here — wrong field chain, the `is not None` guard dropped, or the TRENDS gate diverging — would pass with all tests green. Add a funnel-trends test using a `FunnelsDataWarehouseNode` that asserts day-of-week filtering on the warehouse table (at minimum a Mondays-only vs unfiltered contrast).

posthog/hogql_queries/insights/funnels/funnel_event_query.py
```diff
@@ -273,6 +275,9 @@ def _build_data_warehouse_table_query(
...
+        if day_of_week_filter is not None:
+            where_exprs.append(day_of_week_filter)
```
sampennington: 🤖 **sp-ship** · ♻️✨📐 `[reuse+quality+xp]` · **Suggestion / follow-up**

The sibling events-table path just drops `_day_of_week_filter_expr(...)` straight into the `where_exprs` list and lets the downstream `[expr for expr in where_exprs if expr is not None]` discard a `None`. Here you hoist it into a local and add an explicit `if ... is not None:` guard — but the `ast.And(...)` on the next line already filters `None`, so the guard is redundant and the two adjacent methods now say the same thing two different ways. Consider appending `self._day_of_week_filter_expr(ast.Field(chain=["timestamp"]))` directly into `where_exprs` to match the events-table path and drop these three lines.

posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
```diff
@@ -347,6 +347,10 @@ def event_filter(self) -> ast.Expr:
                     timings=self.timings,
                 )
             )
+            # Days of week: lifecycle statuses are derived from activity on the selected days only
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

This comment mostly restates what the appended filter already expresses. Consider dropping it.

products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
```diff
@@ -414,6 +414,11 @@ def where_clause(self, series_with_extra: SeriesWithExtras) -> ast.Expr:
             ]
         )
 
+        # Days of week: "days active" counts activity on the selected days only
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

Same as the lifecycle runner: this comment restates the adjacent code. Consider dropping it.

sampennington (commented):
🤖 **sp-ship** · 🚫 Blockers

Solid, well-tested extension of day-of-week filtering into stickiness, lifecycle, and funnel trends — it correctly reuses the existing `QueryDateRange.day_of_week_filter_expr()` helper, handles the `None` (unfiltered) contract on every path, follows project timezone, and covers entrances/conversions, empty-means-unfiltered, timezone boundaries, actors-match-chart, and previous-period propagation.

One thing worth blocking on: the funnel data-warehouse branch (`_build_data_warehouse_table_query`) carries its own copy of the filter application but no test exercises it — every funnel test uses `EventsNode`, so a regression in that branch would ship green (inline comment). Smaller follow-ups that didn't need their own thread: lifecycle has no compare/previous-period test and the "all seven days = unfiltered" case is only covered for stickiness (cheap parity gaps to close); and for string-typed DW timestamp columns `toDayOfWeek` will bucket by UTC rather than project timezone — consistent with the pre-existing date-range filter on the same alias, so not a regression, but worth a follow-up since day-of-week bucketing is more boundary-sensitive.

products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage** acknowledged — non-actionable: the new day-of-week filter reuses the same unqualified `timestamp` field reference already used by the pre-existing date_from/date_to filters two lines above in this same query (which resolves against this query's own `select` alias). Not a new inconsistency introduced by this PR.

posthog/hogql_queries/insights/funnels/funnel_event_query.py
sampennington: 🤖 **sp-triage** deferring to a human — this needs a new test with a `FunnelsDataWarehouseNode` fixture asserting day-of-week filtering on the warehouse-table path, which is a non-trivial code addition (test data + assertions) rather than a safe mechanical fix. Leaving open for the author to address.

posthog/hogql_queries/insights/funnels/funnel_event_query.py
sampennington: 🤖 **sp-triage** acknowledged — follow-up / stylistic, not required for correctness. Leaving as-is for now.

posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
sampennington: 🤖 **sp-triage** acknowledged — minor comment-wording nit, not required for correctness. Leaving as-is for now.

products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
sampennington: 🤖 **sp-triage** acknowledged — minor comment-wording nit, not required for correctness. Leaving as-is for now.

posthog/hogql_queries/insights/funnels/funnel_event_query.py
sampennington: 🤖 **sp-triage**

Fixed in `38400b9` — added a parameterized trends test to `test_funnel_data_warehouse.py` (`test_data_warehouse_trends_days_of_week_filters_entrances`) that exercises the `_build_data_warehouse_table_query` day-of-week branch: Mondays-only keeps the single Monday row (1 entrance), empty `daysOfWeek` leaves all six days unfiltered (6 entrances). Note: the ClickHouse-backed warehouse suite could not be run in this sandbox (no local stack), so CI will be the first execution of it.

posthog/hogql_queries/insights/funnels/funnel_event_query.py
sampennington: Done in c891aad1 — the data-warehouse branch now drops the expression straight into `where_exprs` (typed `list[ast.Expr | None]`) and lets the existing comprehension filter the `None`, matching the events-table path.

posthog/hogql_queries/insights/lifecycle/lifecycle_query_runner.py
sampennington: Dropped in c891aad1.

products/product_analytics/backend/hogql_queries/stickiness/stickiness_query_runner.py
sampennington: Dropped in c891aad1.

PR #70603, opened 2026-07-14
feat(metrics): support threshold alerts on metrics insights

DanielVisca:
## Problem

Metrics just went to a flag-gated alpha ([#70541](https://github.com/PostHog/posthog/pull/70541)), but a metrics product without alerting is a chart you have to remember to look at. Saved metrics insights couldn't have alerts: `MetricsQuery` wasn't an alertable kind, so the whole alert pipeline (thresholds, schedules, notifications, the Slack/email fan-out) was unavailable for the one signal type people most want paged on. This is the first and highest-leverage item from the metrics strategy discussion: everything downstream (alert → auto-investigation → report) hangs off an alert existing.

## Changes

Adds `MetricsQuery` as a fourth alertable insight kind, mirroring the existing trends/funnels/SQL patterns 1:1:

**Backend**
- New `MetricsAlertConfig` schema type (authored in `schema-general.ts`, codegen'd) — minimal: `{type, check_ongoing_interval?}`.
- New `MetricsExtractor` (`products/alerts/backend/evaluation/metrics.py`) normalizes `MetricsQuerySeries` (`labels`/`points`/`metricName`/`clause`) into the shared `ComparableSeries` contract; registered in the `EXTRACTORS` dispatch registry.
- **Every series the query returns is evaluated** — group-by label-sets, multiple clauses, or the formula series — and the alert fires if any breaches. No series picker (want one clause? save the insight with one clause or a formula).
- Anchoring is positional on the observed-bucket grid (union of observed buckets, zero-filled): second-to-last bucket by default (skips the still-accumulating trailing bucket), last bucket with `check_ongoing_interval` — which requires an upper bound, same guard and reasoning as trends.
- Config-time validation registered in `_ALERT_CONFIG_VALIDATORS`; anomaly `detector_config` is rejected for metrics (no detector extractor yet).
- API gate: metrics alerts are enabled by the **`metrics` product flag itself** (org-scoped, same helper as the hogql/funnel gates) — anyone who can see the product can alert on it. No new flag.
- Empty query result evaluates as 0 via the shared zero-sentinel, so a lower-bound alert on a dead metric fires instead of silently skipping.

**Frontend**
- `areAlertsSupportedForInsight` accepts bare `MetricsQuery` nodes — metrics insights persist **unwrapped** (no `InsightVizNode`), unlike trends/funnels, which is the one asymmetry worth review attention.
- `MetricsAlertConfig` in the frontend `AlertConfig` union + guards + capability helpers; new-alert default config for metrics insights.
- Flag threaded through all four gating call sites (insight header, insight modals, dashboard card menu, `AlertsButton`).

**Isolation**
- The alerts extractor needs zero `products.metrics` imports (schema types live in `posthog.schema`; execution goes through `calculate_for_query_based_insight`). The alerts *tests* seed metrics fixtures via a new `products/metrics/backend/facade/testing.py` re-export, keeping `tach check` clean.

TDD (red→green, run locally against Postgres + ClickHouse):

- **New:** `posthog/tasks/alerts/test/test_metrics_alerts.py` — 10 tests, written first and observed failing at the "Alerts are not supported for this insight" gate, then passing after implementation. Each catches a distinct regression no existing test covers:
  - parameterized absolute lower/upper/within-bounds (extractor anchor value + dispatcher wiring; end-to-end through `run_alert_check` with real ClickHouse-seeded metrics)
  - empty result evaluates as 0 and fires a lower-bound alert (crash/skip regression)
  - group-by fires on *any* breaching label-set (first-series-only regression)
  - relative increase on the observed-bucket grid (anchor/previous index wiring)
  - flag-off create rejected; trends-config-on-metrics rejected; detector_config rejected; ongoing-interval-without-upper rejected (each guards one gate in the validation chain)
  - Tests use a per-test metric name because `metrics1` isn't truncated between tests — keeps them order-independent.
- **Extended:** `insightAlertsLogic.test.ts` (+3 cases, observed red first): bare-node support with flag, flag-off rejection, and the unsupported-reason label matrix including `metrics`.
- **Regression:** existing alert suites pass — `test_validate_alert_config.py`, `test_alert_evaluation.py`, `test_trends_absolute_alerts.py` (69 passed), full `products/alerts/frontend` jest suite (222 passed, 13 suites).
- `tach check --dependencies --interfaces` clean; `hogli ci:preflight` 0 failures.
- Typecheck: this worktree has globally stale kea typegen, so the full `typescript:check` fails pre-existing; I verified via stash-baseline diff that my changes introduce **zero new type errors** (error sets identical modulo line shifts). CI runs the real check.
- Not done: no browser-level walkthrough of the alert form on a metrics insight (no dev app running in this environment) — the form renders the generic threshold/condition sections via the capability helpers, but a visual pass before un-drafting would be wise.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Metrics docs (`/docs/metrics`) don't document alerts yet; worth a short section when this ships to alpha users. Not included since the feature is flag-gated.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Daniel directed the work and design decisions; assigned as DRI.

I (Claude) implemented this end to end from a strategy discussion that ranked metric alerting as the top "Now" item for the Metrics alpha. Skills invoked: `/writing-tests` (before authoring the test files) and `/improving-drf-endpoints` (before touching `AlertSerializer`). Design decisions made along the way: evaluate all returned series instead of adding a clause picker (avoids generalizing `series_index`, matches the comparator's any-breach semantics); positional anchoring on the observed grid rather than time-based bucket-completeness (the response doesn't carry the effective interval); gate on the existing `metrics` flag instead of minting a new one; expose the metrics test seeder through the facade rather than adding a tach exception.

---


products/alerts/backend/evaluation/dispatcher.py
```diff
@@ -19,6 +20,7 @@
     NodeKind.TRENDS_QUERY: TrendsExtractor(),
     NodeKind.HOG_QL_QUERY: HogQLExtractor(),
     NodeKind.FUNNELS_QUERY: FunnelsExtractor(),
+    NodeKind.METRICS_QUERY: MetricsExtractor(),
```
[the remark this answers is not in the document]
DanielVisca: Fixed in d00e5d384c17. Creating or repointing an alert to a metrics insight with a programmatic token (PAK/OAuth) now requires the metrics data scope in addition to `alert:write` — `_require_metrics_scope_for_programmatic_auth` in `products/alerts/backend/api/alert.py` runs from the object-level `validate()` so a PATCH that omits `insight` can't bypass it. Session auth (no token scopes) stays exempt, matching the house model where logged-in users are gated by team membership + insight viewer access. Coverage added in `test_metrics_alerts.py::test_api_key_metrics_scope_enforcement`: `["alert:write"]` → 403 with the missing-scope message; `["alert:write", "metrics:read"]` → 201. Worth noting the same oracle property exists for trends/SQL/funnel alerts today (SQL being the widest); this PR fixes the new metrics surface where no existing tokens can break, and I'd suggest a follow-up decision on the older kinds.

products/alerts/backend/api/alert.py
```diff
@@ -91,7 +92,10 @@ class AlertConfigUnion(RootModel):
     """Per-insight-kind alert config, discriminated by ``type`` — keeps the OpenAPI (and the
     generated frontend types and MCP tool schemas) in sync with every kind alerts support."""
 
-    root: Annotated[TrendsAlertConfig | HogQLAlertConfig | FunnelsAlertConfig, PydanticField(discriminator="type")]
+    root: Annotated[
+        TrendsAlertConfig | HogQLAlertConfig | FunnelsAlertConfig | MetricsAlertConfig,
+        PydanticField(discriminator="type"),
+    ]
```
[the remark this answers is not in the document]
DanielVisca: Fixed in d00e5d384c17. The feature-flag gate is now a shared module-level helper (`_enforce_alert_feature_flags`) called by both `AlertSerializer` and `AlertSimulateSerializer.validate_insight`, so simulate gives the same gated rejection as create/update — for metrics and also for the pre-existing hogql/funnel gaps this covered. With the flag on, a metrics insight still 400s cleanly at the detector registry ("Anomaly detection simulation isn't supported for MetricsQuery insights") since metrics has no detector extractor yet. Coverage: `test_simulate_rejected_when_metrics_flag_disabled` (gated rejection) and `test_simulate_rejects_metrics_insight_with_flag_on` (clean registry 400, not a 500).

products/alerts/backend/api/alert.py
```diff
@@ -87,11 +89,61 @@ class AlertConditionField(serializers.JSONField):
...
+        raise ValidationError("Metrics insight alerts are not enabled for your account.")
+
+
+def _require_metrics_scope_for_programmatic_auth(context: dict[str, Any], insight: Insight) -> None:
+    # An alert on a metrics insight executes the query as `created_by` and delivers the computed
+    # value and labels in notifications, so a programmatic token must also carry the metrics data
+    # scope — otherwise `alert:write` alone becomes a metrics-read oracle bypassing the explicit
+    # product-scope gate on the query endpoints. Session auth (no token scopes) is exempt: those
+    # users are gated by team membership and insight viewer access instead.
+    if insight.alertable_query_kind != NodeKind.METRICS_QUERY:
```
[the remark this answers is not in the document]
DanielVisca: Fixed in 15d1812f500f. The alerts queryset (which `get_object()` resolves detail routes from, so list/retrieve/update/delete + embedded checks are all covered) and the insight-nested thresholds endpoint now exclude alerts on metrics insights for programmatic tokens lacking `metrics:read`/`metrics:write` — the read-side twin of the write gate, via a shared `_token_lacks_metrics_scope` helper. Implementation note: the exclusion uses a null-safe filter-then-exclude-by-id shape because `exclude()` over a JSON path is three-valued and would silently drop alerts whose insight query lacks the key. Coverage: `test_api_key_read_paths_gate_metrics_alerts` — alert:read-only key sees the metrics alert in neither list, detail (404), nor thresholds; alert:read+metrics:read sees all three.

products/alerts/backend/api/alert.py
```diff
@@ -87,11 +89,61 @@ class AlertConditionField(serializers.JSONField):
...
+    if kind == NodeKind.HOG_QL_QUERY and not _insight_alert_flag_enabled(context, "hogql-insight-alerts"):
+        raise ValidationError("SQL insight alerts are not enabled for your account.")
+    if kind == NodeKind.FUNNELS_QUERY and not _insight_alert_flag_enabled(context, "funnel-insight-alerts"):
+        raise ValidationError("Funnel insight alerts are not enabled for your account.")
+    # Gated on the Metrics product flag itself: anyone who can see the product can alert on it.
+    if kind == NodeKind.METRICS_QUERY and not _insight_alert_flag_enabled(context, "metrics"):
+        raise ValidationError("Metrics insight alerts are not enabled for your account.")
+
+
+def _require_metrics_scope_for_programmatic_auth(context: dict[str, Any], insight: Insight) -> None:
```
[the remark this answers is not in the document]
DanielVisca: Fixed in 15d1812f500f. The alerts queryset (which `get_object()` resolves detail routes from, so list/retrieve/update/delete + embedded checks are all covered) and the insight-nested thresholds endpoint now exclude alerts on metrics insights for programmatic tokens lacking `metrics:read`/`metrics:write` — the read-side twin of the write gate, via a shared `_token_lacks_metrics_scope` helper. Implementation note: the exclusion uses a null-safe filter-then-exclude-by-id shape because `exclude()` over a JSON path is three-valued and would silently drop alerts whose insight query lacks the key. Coverage: `test_api_key_read_paths_gate_metrics_alerts` — alert:read-only key sees the metrics alert in neither list, detail (404), nor thresholds; alert:read+metrics:read sees all three.

products/alerts/frontend/views/EditAlertModal.tsx
```diff
@@ -92,11 +92,13 @@ export function EditAlertModal({
     const funnelStepLabels = (funnelSource?.series ?? []).map(
         (node, index) => getDisplayNameFromEntityNode(node) ?? `Step ${index + 1}`
     )
-    const insightAlertKind: 'hogql' | 'funnels' | 'trends' = containsHogQLQuery(query)
+    const insightAlertKind: 'hogql' | 'funnels' | 'trends' | 'metrics' = containsHogQLQuery(query)
```
MattPua: nit: as this point i would make this an actual type

products/alerts/frontend/views/EditAlertModal.tsx
```diff
@@ -92,11 +92,13 @@ export function EditAlertModal({
...
         ? 'hogql'
         : isFunnelInsight
           ? 'funnels'
-          : 'trends'
+          : isMetricsQuery(query)
+            ? 'metrics'
+            : 'trends'
```
MattPua: code smell: we should just make an actual function at this point, or switch case or something

products/alerts/frontend/logic/insightAlertsLogic.ts
```diff
@@ -33,7 +34,7 @@ export interface InsightAlertsLogicProps {
 
 export const areAlertsSupportedForInsight = (
     query?: Record<string, any> | null,
-    options: { hogqlAlertsEnabled?: boolean; funnelAlertsEnabled?: boolean } = {}
+    options: { hogqlAlertsEnabled?: boolean; funnelAlertsEnabled?: boolean; metricsAlertsEnabled?: boolean } = {}
```
MattPua: nit: we could also make this options a type as well

products/alerts/frontend/views/EditAlertModal.tsx
MattPua: actually you already have a type for this in alertFormLogic.ts

products/alerts/backend/evaluation/metrics.py
```diff
@@ -0,0 +1,123 @@
...
+        check_ongoing_interval = bool(config.check_ongoing_interval)
+        if (
+            check_ongoing_interval
+            and threshold.bounds.upper is None
+            and condition.type in (AlertConditionType.ABSOLUTE_VALUE, AlertConditionType.RELATIVE_INCREASE)
+        ):
+            raise ValueError(
+                f"check_ongoing_interval is only supported for alert condition {condition.type} "
+                "when upper threshold is specified"
+            )
```
MattPua: isnt this basically just the _validate_metrics_alert_config function?

MattPua (approved):

products/alerts/frontend/views/EditAlertModal.tsx
DanielVisca: Done in 1c5cfd654ae6 — good catch that `InsightAlertKind` already existed in `alertFormLogic.ts`; the resolver now returns it directly so the annotation is inferred.

products/alerts/frontend/views/EditAlertModal.tsx
DanielVisca: Done in 1c5cfd654ae6 — extracted `insightAlertKindForQuery(query)` into `alertFormLogic.ts` next to the `InsightAlertKind` type (early-return guards instead of the nested ternary). Also let me drop the now-redundant `isFunnelInsight` local in the modal.

products/alerts/frontend/logic/insightAlertsLogic.ts
DanielVisca: Done in 1c5cfd654ae6 — named it `AlertSupportFlagOptions` and reused it across `areAlertsSupportedForInsight`, `areAnomalyAlertsSupportedForInsight`, `alertableInsightTypesLabel`, and `alertsUnsupportedReason` (the last three were each repeating the inline shape).

products/alerts/backend/evaluation/metrics.py
DanielVisca: Yes — deliberately the same rule in both places, mirroring how trends does it: `_validate_trends_alert_config` rejects at save time and `TrendsExtractor.extract()` re-raises the identical guard at eval time (`trends.py`, the `check_ongoing_interval ... upper threshold` ValueErrors in the ABSOLUTE_VALUE/RELATIVE_INCREASE arms). The save-time check gives the clean 400; the extractor guard is the eval-time backstop for alerts that reach `extract()` without passing API validation — configs saved before a rule existed, or callers like `run_alert_check`/direct `check_alert_for_insight` that skip `prepare_alert`'s re-validation. Happy to remove both copies in a follow-up if you'd rather lean solely on `prepare_alert` re-validating on every check, but I kept metrics consistent with trends here.

MattPua: yeah seems like we hsould actually just remove the duplication

DanielVisca: Done in e6712363a599 — removed the extractor's copy; `_validate_metrics_alert_config` (save time + re-run by `prepare_alert` before every check) is now the single owner of that rule, with a comment in `extract()` pointing there. Left trends' equivalent untouched since it's pre-existing code outside this PR's scope. All 16 metrics alert tests still pass.

PR #70604, opened 2026-07-14
feat(desktop): Implement loops ➿

charlesvien:
## Problem

Loops put recurring PostHog work on autopilot. Instead of kicking off the same agent task by hand every time, you define a loop once and it runs itself on PostHog's managed cloud infrastructure: on a schedule, when a GitHub event fires, or when your own code calls its API. This PR adds the backend API to create and manage them. The desktop UI lives in the code repo [[companion PR]](https://github.com/PostHog/code/pull/3411).

## Changes

Loops are named, cloud-executed agent automations: a prompt + model + triggers, running on the existing tasks pipeline. This PR is the backend and MCP surface; the UI ships separately in PostHog/code. Spec and pressure-test docs are in `products/tasks/docs/`.

- New `Loop`, `LoopTrigger` and `LoopFire` (per-fire dedup) models, plus a nullable `loop` FK on `Task`. Triggers: cron or one-time schedules, GitHub webhook events (`pull_request`, `issues`, `issue_comment`, `push`) and an external API trigger
- Every trigger path funnels through one `fire_loop()` that applies dedup, daily rate caps (100 fires/loop, 500/team) and the loop's overlap policy (skip, allow or cancel previous) under a per-team advisory lock, then creates an internal Task + TaskRun on the standard tasks pipeline
- Endpoints under `loops/`: CRUD plus `run/` (manual fire), `trigger/` (external fire, callable with a project secret API key via the new `loop:write` scope, the first write-capable PSAK scope), `runs/` history and a dry-run `preview/`
- The GitHub webhook in `posthog/urls.py` becomes a per-event-type handler registry with per-handler Redis dedup (released if a handler fails so redeliveries retry); conversations, tasks and installation handlers keep their slots, loops register alongside
- Contexts: a loop can attach to a desktop context folder and file runs into its feed, maintain its `context.md` and/or rewrite a canvas dashboard; context-attached loops must be team-visible
- Guardrails: 100 loops/team, 25 triggers/loop and 1 repo/loop (structured 429), auto-pause after 5 consecutive failures, owner deactivation or GitHub App uninstall pauses affected loops and cancels their in-flight runs (cause recorded in `disabled_reason`), retention sweeps (200 tasks kept per loop, fire records pruned after 7 days)
- Team loops need an explicit ownership takeover to edit, re-validating repositories and connectors against the new owner's GitHub and MCP access; config changes, takeovers and manual fires are activity-logged, and personal loop activity stays hidden from other members
- Notifications per loop: in-app always, push/email/Slack opt-in with per-channel event filters, a failure cooldown and Slack auto-disable on permanent errors
- `loops-*` MCP tools plus a `loops-review` card (MCP UI app): an agent assembles a loop from natural language and the user confirms with a Create button instead of it being created blind

### Temporal changes

- One Temporal Schedule per schedule trigger (`loop-trigger-{id}`), cron or one-time. Always skip overlap with a 5 minute catchup window, so an outage never replays a missed window as a burst; the schedule's paused state mirrors the trigger and loop enabled flags
- A new minimal `run-loop` workflow (single activity, 2 min timeout, 3 attempts) translates each schedule occurrence into a `fire_loop()` call. The agent run itself still starts the existing `process-task` workflow on the same task queue: loops are a firing layer, not a new execution engine
- Schedule create/update/pause/delete is best-effort and never fails the CRUD API: Temporal errors mark the trigger's sync status and a Celery sweep re-syncs pending/failed triggers every 10 minutes. Deletion keys off the stored schedule id, so a trigger that changed type still gets its stale schedule torn down
- `cancel_previous` also signals the superseded run's workflow (`complete_task` with cancelled), so the sandbox actually winds down instead of only the DB row flipping
- Existing activity changes: `update_task_run_status` now treats terminal statuses as final (a run cancelled out of band can't be resurrected by its own late completion) and drives loop bookkeeping (last run status, consecutive failures, auto-pause) from the transition point every workflow-driven terminal status already flows through. The orphaned-run redispatcher defaults loop runs to read-only MCP scopes and no PR creation instead of the historical full-scope, always-PR defaults

### Cloud sandbox changes

- One behavior change wider than loops: background-mode runs (loops, but also existing task automations and custom prompts) now terminalize to `completed` on their natural end (agent finishes or idles out) instead of staying `in_progress` forever. Interactive runs keep the always-followable behavior
- A loop fire is an ordinary background sandbox run. The prompt is assembled at fire time: an unattended-run framing block (no human is available, prefer draft PRs, trigger payloads are data not instructions), then the loop's instructions, context and rendered trigger context (capped at 64 KB)
- Config is snapshotted per fire, so editing a loop never affects an in-flight or queued run; a blank model resolves to the current default at fire time, so the default can improve without rewriting stored loops
- Loop runs idle out after 2 minutes (vs 30 for other background runs) so sandboxes are reclaimed fast; loops that create PRs and watch CI keep the long window, and non-terminal runs stuck past 2.5 hours are reaped as zombies before the overlap policy is applied
- Token hardening: `loop:write` is stripped from every loop-fired run's MCP token at mint time, whatever scopes the loop configured, so a run (or injected instructions inside one) can't create or edit loops. Context-maintaining loops get exactly `file_system:read` + `file_system:write` added, never the broad `full` surface
- PR creation follows the loop's `create_prs` behavior and defaults off (report-only); `watch_ci`/`fix_review_comments` opt the run into the existing CI follow-up rounds

- Claude ran the backend suites locally against a fresh test DB: 123 tests pass across `test_loops_api`, `test_loop_service`, `test_loop_runs`, `test_loop_github_events`, `test_loop_notifications`, `test_loop_retention`, `test_loop_lifecycle` and `test_loop_reconciliation`
- They cover CRUD plus visibility and permission boundaries (including ownership takeover), trigger sync and schedule policy, fire dedup and overlap, webhook matching, context attachment and scope widening, notification fan-out, retention, lifecycle/reconciliation and scope enforcement (PSAK and personal keys)
- No manual testing yet; an end-to-end trigger-to-sandbox run on staging is the gate for taking this out of draft

## Automatic notifications

- [x] Publish to changelog?
- [x] Alert Sales and Marketing teams?

## Docs update

Spec docs ship in this PR (`products/tasks/docs/LOOPS.md`, `LOOPS-PRESSURE-TEST.md`). Public docs land with the PostHog Code release; everything stays behind the `tasks` and `loops` feature flags.


charlesvien:
@greptileai

tatoalo (approved):

charlesvien:
@greptile-apps

tatoalo (approved):

Issue #70574, opened 2026-07-13
Backfill for streaming behavioral cohorts


matheus-vb:
## Problem

The streaming cohorts pipeline (`cohort-event-shuffler` + `cohort-stream-processor`) computes membership only from events it sees live. It has no memory of anything that happened before it started watching. Five real situations depend on history it can never observe live:

| # | Trigger | What is missing |
|---|---------|-----------------|
| T1 | Team enabled (allowlist flip) | All windowed state for that team's cohorts |
| T2 | New cohort created | State for all history inside the cohort's window |
| T3 | Cohort edited (window, threshold, filter) | State under the new key; old members stay members forever (a confirmed cutover blocker) |
| T4 | Store loss or schema bump (e.g. the v3 wipe) | Everything since `now - max(window)` |
| T5 | Dormant persons | Person property state for people who never emit events |

None of this is built today. The flag serving gate `Cohort.is_flag_compatible` already requires `last_backfill_events_at` to be set for a behavioral cohort, but nothing in production writes that column, so the gate stays permanently closed.

## Core idea

We compute the historical state with the exact same evaluator the live processor uses, aggregate it into small mergeable tiles, and feed those tiles into the processor through a new topic. The processor merges them into its live state idempotently. Because batch and live can overlap, we make the two domains disjoint by construction and absorb any overlap with an idempotent merge.

```
Django (products/cohorts)                    ClickHouse events history
  detects edit / new cohort / enablement          |
  enqueues a pinned BackfillRun ---------> cohort-seeder (new Rust binary)
  compare-and-swap stamps readiness <--+     reuses the live Stage 1 crate + Rust HogVM
                                       |     scans team + day + event-name, evaluates each event
                                       |     aggregates matches into (person, condition, day) tiles
                                       |          |
                                       |     cohort_stream_seed_events (64p, same murmur2 key)
                                       |          |
                                       |     cohort-stream-processor (new 5th consumer)
                                       |          merges tiles into daily-bucket state (take the max)
                                       +---- reconcile: full membership snapshot when seeding finishes
                                                  |
                                             cohort_membership_changed(_shadow) -> harness / PG consumer
```

## How it works

1. **Seeder.** A new small Rust binary scans ClickHouse once per run, filtered to the union of event names the team's cohorts reference. It evaluates every event with the same Stage 1 crate and Rust HogVM the processor runs live, so matching is identical by construction. It resolves each event to its canonical person at scan time (avoiding the stale `person_id` trap on the events row).

[...]

PR #70613, opened 2026-07-14
feat(cohorts): implement backfill coordination foundation

matheus-vb:
## Problem

Behavioral realtime cohort edits can orphan persisted streaming state, while Django had no durable coordination records or safe readiness handoff for backfills.

First step of https://github.com/PostHog/posthog/issues/70574.

## Changes

Adds atomic leaf shape hashing and readiness invalidation, gated backfill scheduling, pinned run state, CAS readiness stamping, an operator command, migrations, and tests.

The combined backend suite passed 481 tests and 89 snapshots. Ruff, migration drift, and an isolated migration apply, rollback, and reapply check passed.

## Automatic notifications

* [ ] Publish to changelog?
* [ ] Alert Sales and Marketing teams?

## Docs update

No.


posthog/settings/cohorts.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+from posthog.settings.utils import get_from_env, str_to_bool
+
+# Off by default ("none"); enable per environment via the env override
+# A set-but-empty value survives as "", which the parser reads as "all teams".
+REALTIME_COHORT_TEAM_ALLOWLIST: str = os.getenv("REALTIME_COHORT_TEAM_ALLOWLIST", "none")
```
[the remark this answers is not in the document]
matheus-vb: intended

products/cohorts/backend/backfill/readiness.py
```diff
@@ -0,0 +1,93 @@
...
+    This update intentionally bypasses signals. B5 must explicitly invalidate feature-flag and
+    behavioral-cohort caches after a successful stamp.
+    """
+    participation = CohortBackfillRunCohort.objects.for_team(run.team_id).get(run_id=run.id, cohort_id=cohort_id)
+    updated = Cohort.objects.filter(
+        id=cohort_id,
+        team_id=run.team_id,
+        filters_shape_hash=participation.filters_shape_hash,
+        last_backfill_events_at__isnull=True,
+    ).update(last_backfill_events_at=Now())
```
gustavohstrassburger, line 61: `suggestion`: This CASes on the full `filters_shape_hash`, but edit-time invalidation only nulls `last_backfill_events_at` when the *behavioral* hash changes (see `_maintain_behavioral_shape` in `models/cohort.py`). A person-property or cohort-reference edit mid-backfill changes the full hash without nulling readiness, so when this function runs, both the CAS and the fallback SELECT miss against the stale pinned hash and the run gets marked SUPERSEDED, discarding a completed, still-valid events backfill.

`participation.filters_shape_hash` is pinned from `ensure_filters_shape_hash`, which returns the full hash (person + cohort + behavioral leaves). Since events readiness is about behavioral event state, this should CAS on `behavioral_filters_shape_hash` instead:

```suggestion
    updated = Cohort.objects.filter(
        id=cohort_id,
        team_id=run.team_id,
        behavioral_filters_shape_hash=participation.behavioral_filters_shape_hash,
        last_backfill_events_at__isnull=True,
    ).update(last_backfill_events_at=Now())
```

That needs `behavioral_filters_shape_hash` pinned on `CohortBackfillRunCohort` at run creation (it currently only stores the full hash), plus the same swap in the fallback SELECT a few lines down. `test_edit_between_completion_and_stamp_fails_cas` only covers a *behavioral* edit correctly superseding, worth adding a sibling for a person-only edit that asserts the run is NOT superseded.

products/cohorts/backend/models/leaf_shape.py
```diff
@@ -0,0 +1,78 @@
+import json
+import hashlib
+from collections.abc import Iterator
+
+_I32_MIN = -(2**31)
+_I32_MAX = 2**31 - 1
+
+
+def walk_filter_leaves(node: object) -> Iterator[dict]:
```
gustavohstrassburger, line 9: `suggestion`: This recurses into nested `AND`/`OR` groups, but every test builds a flat single-level `AND`. A behavioral leaf nested inside an inner `OR`, which real cohorts produce, isn't covered, so a regression in the recursion wouldn't be caught. Worth one test case with a nested group.

products/cohorts/backend/models/cohort.py
```diff
@@ -245,6 +250,81 @@ class Meta:
...
+        # or as a keyword. Intercept it so _maintain_behavioral_shape can extend the frozen set.
+        positional_update_fields = args[3] if len(args) > 3 else kwargs.get("update_fields")
+        maintained_update_fields = self._maintain_behavioral_shape(positional_update_fields)
+        if len(args) > 3:
+            args = (*args[:3], maintained_update_fields, *args[4:])
+        else:
+            kwargs["update_fields"] = maintained_update_fields
+        super().save(*args, **kwargs)
+
+    def _maintain_behavioral_shape(self, update_fields: Iterable[str] | None) -> set[str] | None:
```
gustavohstrassburger, line 264: `suggestion`: The try/except here swallows any hashing exception so `save()` never raises, but the existing "failure doesn't break save" test patches the signal-receiver's guard, not this function directly. Worth a test that makes `extract_leaf_shape_hash` raise and asserts `cohort.save()` still succeeds, that's the exact case this branch exists for.

gustavohstrassburger (approved):
Nice foundation! A few non-blocking suggestions.

PR #70652, opened 2026-07-14
feat(billing): surface billing activity in the activity log UI

yasen-posthog:
## Problem

Billing activity rows (spend limits, addons/products, plan switches, trials, credits) are being written to the activity log by the billing SQS consumer, but they were invisible/unhelpful in the UI: `"Billing"` was never registered as an activity scope, so it wasn't a valid filter value and rows fell back to the raw default renderer. This registers the scope and adds a describer so they render meaningfully.

## Changes

- Add `"Billing"` to the backend `ActivityScope` literal — this regenerates the `ActivityLogListScope` filter enum (the advanced-activity-logs `scope` filter is a `ChoiceField(choices=get_args(ActivityScope))`, so `"Billing"` is now a selectable, valid filter).
- Add `BILLING = 'Billing'` to the frontend `ActivityScope` enum.
- Add `billingActivityDescriber` (`scenes/billing/activityDescriptions.tsx`) and wire it into `describerFor`. Every `billing_activity` row is an `"updated"` activity; the describer maps `detail.name` to a short verb phrase ("updated the spend limits", "added products", "switched plan", …) and summarizes `detail.changes`, e.g. _"PostHog updated the spend limits in billing (product_analytics from 500 to 1000)"_. System-origin rows (Stripe/dunning) render as "PostHog" via the existing `is_system` handling.

Generated diff is Billing-only (8 lines across the two generated files).

## Remaining gate (not in this PR)

Billing logs are **org-scoped** (`team_id=NULL`). `apply_organization_scoped_filter` only includes org-scoped rows in a project's activity view when that team has `receive_org_level_activity_logs=True` — otherwise they're filtered out regardless of the describer. So making them appear for a given team still depends on that flag (or surfacing them via a dedicated org-level billing-activity view). Flagging as the follow-up product decision.

Confirmed via Grafana + prod DB that the pipeline already writes these rows end-to-end (billing producer + PostHog consumer both live); this PR is purely the render/registration layer. Ran `hogli build:openapi` (generated enum picks up `Billing`), targeted typecheck of the three changed source files (clean; unrelated pre-existing master tsgo errors elsewhere), ruff clean, and lint-staged `ty check` passed on commit. I did not exercise the describer in a running browser.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code. Diagnosis first: traced the "no activity logs" report through Grafana Loki (producer `send_billing_activity_to_sqs` succeeding; consumer `_process_billing_activity` running the deployed code with no write failures) and the code, concluding rows are written but hidden — `"Billing"` unregistered + org-scoping. This PR fixes the registration/render half. Modeled the describer on `organizationActivityDescriber`; kept copy in sentence case without em dashes per the repo copy guidelines.


posthog/models/activity_logging/activity_log.py
```diff
@@ -95,6 +95,7 @@
     "SignalScoutConfig",
     "StreamlitApp",
     "Metric",
+    "Billing",
```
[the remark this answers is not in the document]
yasen-posthog: Thanks — this is accurate, but it describes the intended behavior rather than a regression.

Billing activity is inherently org-scoped, so it's only surfaced to teams that have opted into `receive_org_level_activity_logs` (the advanced activity logs viewset already gates org-level rows on that flag). A team in the default state seeing an empty "Billing" filter is consistent with how every other org-level activity source already behaves — the flag is the boundary, not this scope.

Adding `Billing` to the scope enum is what lets the new describer render those rows for teams that *do* receive org-level activity, which is the goal of this PR. Gating the filter option itself on the flag would be a separate, cross-cutting change to the scope filter and isn't specific to billing. Resolving as working as intended.

frontend/src/scenes/billing/activityDescriptions.tsx
```diff
@@ -0,0 +1,54 @@
...
+    'Billing trial extended': 'extended a trial',
+    'Billing trial cancelled': 'cancelled a trial',
+    'Billing credits purchased': 'purchased credits',
+}
+
+const formatValue = (value: ActivityChange['before']): string =>
+    value === null || value === undefined ? 'none' : `${value}`
+
+const describeChanges = (changes: ActivityChange[]): string | null => {
+    if (changes.length === 1 && changes[0].action === 'changed' && changes[0].field) {
```
pawel-cebula: For `created` and `deleted` changes, we only show the field name. For example, a credit purchase renders as `(credits)` and leaves out the amount from `after`. Could we show `after` for created values and `before` for deleted values, then add tests for those cases?

posthog/models/activity_logging/activity_log.py
pawel-cebula, line 99: Could we add `Billing` to the list in `ReadActivityLog.scope`, or generate that list from `ActivityScope`? The tool accepts `Billing`, but its description says it isn't available. That makes the filter hard for HogAI to discover.

frontend/src/scenes/billing/activityDescriptions.tsx
```diff
@@ -0,0 +1,54 @@
...
+// The billing service sets detail.name; map it to a short verb phrase for the feed.
+const BILLING_ACTIONS: Record<string, string> = {
+    'Billing spend limits': 'updated the spend limits',
+    'Billing next-period limit reset': 'reset a next-period spend limit',
+    'Billing products activated': 'added products',
+    'Billing products deactivated': 'removed products',
+    'Billing plan switched': 'switched plan',
+    'Billing trial activated': 'started a trial',
+    'Billing trial extended': 'extended a trial',
+    'Billing trial cancelled': 'cancelled a trial',
```
pawel-cebula: nit from `AGENTS.md`: The event name also uses British spelling, not just the text shown to users. Could we rename it to `Billing trial canceled` in Billing and show `canceled a trial` here? We should keep `Billing trial cancelled` as an alias so existing activity rows still render correctly.

pawel-cebula (commented):
Left a few comments, approving in advance.

pawel-cebula (approved):

yasen-posthog:
### CI status

All CI relevant to this change is green — the deterministic gates (Validate OpenAPI types, Repo checks, Validate migrations, Discover product tests), Storybook/visual regression, every Django test shard, Frontend CI, and MCP CI all pass on the latest commit.

The only red checks are **`Product tests (warehouse-sources (4/4))`** and its aggregate gate **`Django Tests Pass`**. That shard is failing with a runner-infrastructure error, not a test failure:

> The self-hosted runner lost communication with the server.

This is a pre-existing, master-wide issue — the same shard fails with the identical error on every recent `master` run (5/5 checked) and is unrelated to this PR, which only touches the activity-log describer, the `ActivityScope` enum, and the regenerated OpenAPI types. I re-ran the job; it died the same way. `Django Tests Pass` is red only because it `needs:` that shard.

Nothing in this diff can fix a runner OOM/network issue on the warehouse-sources shard, so I'm leaving it for the infra owners rather than churning re-runs.

frontend/src/scenes/billing/activityDescriptions.tsx
yasen-posthog: Good catch. `describeChanges` now shows the value for single-field created/deleted changes too — created renders `field: <after>` and deleted renders `field: <before>` (multi-field changes still list just the names to stay compact). Added tests covering the created (credit amount), deleted, and changed cases in `activityDescriptions.test.tsx`.

posthog/models/activity_logging/activity_log.py
yasen-posthog: Added `Billing` to the `ReadActivityLog.scope` description so the tool advertises it. I kept the curated string list rather than generating it from `ActivityScope` — the list is already an intentional subset (it omits internal scopes like `SignalScoutConfig`/`StreamlitApp`), so auto-generating would surface scopes we don't want HogAI filtering on. Happy to revisit that as a separate change if we'd rather drive it from the enum.

frontend/src/scenes/billing/activityDescriptions.tsx
yasen-posthog: Done — the feed now shows `canceled a trial` and maps the American `Billing trial canceled` name, with `Billing trial cancelled` kept as an alias so existing rows still render. The rename on the billing service side (emitting `Billing trial canceled`) is a separate change in that repo; this alias means we can land them in either order. Both names are covered by a test.

yasen-posthog:
Updated status after addressing review feedback:

- All three review comments are addressed in the latest commits — created/deleted changes now render their value (not just the field name), the trial-cancel copy uses American spelling with the British name kept as an alias, and `Billing` is now advertised in the `ReadActivityLog` scope description. Added describer tests.
- Rebased onto `master` (kept both `TableCertification` and `Billing` in the scope list) — the branch is now up to date with no conflicts.
- Re-ran the two previously-red non-deterministic jobs: **Storybook / visual regression** and **API CLI release bundle** both pass now (the earlier failures were a flaky play-test and a one-off 15‑min build timeout respectively; the CLI bundle passes on every other recent branch).

The only remaining red is still **`Product tests (warehouse-sources (4/4))`** (and its aggregate **`Django Tests Pass`**), failing with the same runner-infrastructure error — "the self-hosted runner lost communication with the server". It's unrelated to this PR and reproduces on `master`, so it needs an infra owner rather than a change here.

frontend/src/lib/components/ActivityLog/describers.tsx
```diff
@@ -56,6 +57,8 @@ export const describerFor = (logItem?: ActivityLogItem): Describer | undefined =
             return batchExportActivityDescriber
         case ActivityScope.BATCH_IMPORT:
             return batchImportActivityDescriber
+        case ActivityScope.BILLING:
```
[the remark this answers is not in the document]
yasen-posthog: Thanks — this is the same visibility question raised earlier in this PR, and it describes existing, intended behavior rather than something this change introduces.

A few points:

- **The describer is pure frontend rendering.** It formats `logItem.detail.changes`, which is data the activity-log API has *already* returned to the client. It cannot expose anything the API didn't already send — hiding or changing the describer doesn't affect what the endpoint returns, as you note.
- **Org-level rows (Billing included) are already gated server-side** on `receive_org_level_activity_logs` in `posthog/api/advanced_activity_logs/viewset.py` (`_should_skip_parent_filtering`). That flag is itself owner/admin-only to toggle (`TeamSerializer.validate_receive_org_level_activity_logs`). So an org only surfaces org-level activity to project members after an owner/admin opts in — that is the deliberate boundary for *all* org-level activity, not just billing.
- **This PR doesn't change access control, storage, or that gating.** It adds a describer, adds `Billing` to the `ActivityScope` enum, and advertises `Billing` in HogAI's `ReadActivityLog.scope`. The HogAI path runs through the same gated viewset, so it doesn't bypass anything either.
- Adding a *billing-admin/owner-specific* restriction below the existing org-level flag would be a separate, cross-cutting change to the activity-log access model — it isn't specific to this describer, and it belongs server-side rather than in a rendering change.

This matches the earlier identical thread (resolved as working-as-intended) and the maintainer's approval, so resolving here as well.

PR #70661, opened 2026-07-14
feat(data-warehouse): attribute posthog code-created sources as created_via=self_driving

Gilbert09:
## Problem

#70242 made wizard-created warehouse sources distinguishable via `created_via=wizard`. PostHog Code is the next surface getting source onboarding in its UI (the self-driving area of the app), and those sources would currently land as plain `mcp` - indistinguishable from any other MCP client.

## Changes

Extends the same server-side derivation to PostHog Code:

- New `SELF_DRIVING = "self_driving"` value on `ExternalDataSource.CreatedVia` (+ choices-only migration, no DDL).
- The create path's single wizard upgrade becomes a transport map: a machine-injected `created_via=mcp` is upgraded to `wizard` for wizard transports and to `self_driving` when `get_event_source` resolves `EventSource.POSTHOG_CODE` (the existing `posthog/code` / `*.hog.dev` UA detection, which also matches the `; for posthog/code` suffix the MCP server appends when proxying).
- Same anti-spoof posture as wizard: `self_driving` is not accepted from callers (not in the create serializer's choices - a 400), and explicit `web`/`api` values are never rewritten. The MCP tool still injects `mcp` itself; agents can't set the field.
- Help texts updated on both serializers.

> [!NOTE]
> Generated OpenAPI types are not in this PR - local codegen isn't available in this environment, so the `check-openapi-types` CI job will regenerate and auto-commit them to this branch (same-repo PR path).

- Extended the existing parameterized transport test (`test_create_external_data_source_transport_user_agent_upgrades_mcp_created_via`) rather than adding a new one: new cases cover `posthog/code` UA, the MCP-server-proxied `; for posthog/code` UA shape, a plain MCP UA staying `mcp`, and explicit `web`/`api` staying untouched under a PostHog Code UA. The regression this catches: PostHog Code-driven creates silently reverting to `mcp`, or the proxied UA shape not matching.
- Added `self_driving` to the reject-invalid-created_via cases - a caller must not be able to self-label as PostHog Code-created.
- Ran the 18 `created_via` API tests locally (Python 3.13 venv): all pass.
- Not run locally: OpenAPI codegen (see note above) and the wider backend suite (CI covers both).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected - the field's API help text is the documentation surface and is updated in this PR.

## 🤖 Agent context

Claude Code implemented this by mirroring the merged wizard attribution PR (#70242). Checked prod first: no `posthog_code`-sourced warehouse creations exist yet, so this is preparatory for the app's onboarding flow and there is no backfill concern. Decisions: kept the upgrade rule scoped to machine-injected `mcp` values only (consistent with the wizard's anti-spoof rationale) rather than also upgrading defaulted `api` values, so a future direct-REST path in PostHog Code would need to opt in explicitly; converted the single wizard conditional into a transport map rather than chained ifs. Skills invoked: /django-migrations (choices-only AlterField, no DDL), /writing-tests (extended existing parameterized tests instead of new functions). The migration was hand-written mirroring the wizard's 0065 because the local default venv can't run master's `manage.py`; it is byte-equivalent to what `makemigrations` produces for a choices change.


estefaniarabadan (approved):

Gilbert09:
### CI status — one failing check is a master-wide infra flake, not this PR

The only red check is **`Product tests (warehouse-sources (4/4))`** (and the `Django Tests Pass` gate that aggregates it). Everything else on the latest commit is green (146 passing).

This shard is failing for an infrastructure reason, not a test in this diff:

- Every failure ends with the annotation *"The self-hosted runner lost communication with the server… Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error"* and uploads **no logs** (the runner dies before flushing them).
- It reproduces on **clean `master`** — shard `warehouse-sources (4/4)` has been red while shards 1–3 stay green across multiple recent master runs (e.g. `29329064920`, `29329033116`, `29328190837`, `29328058805`, `29327356460`). It was also red on this PR's pre-merge head, so merging master in did not introduce it.
- I re-ran the failed job twice; it died the same way each time. The consistency (always the heaviest shard, 4/4) points to the shard OOMing / the runner being reclaimed rather than a flaky test.

The change here is a small, self-contained addition (`self_driving` `created_via` enum value + migration + serializer/test updates); I verified the new test cases match the server-side derivation logic in `get_event_source`, and the MCP tool-schema snapshots have been regenerated so `Unit Tests` is green.

This needs a fleet-side fix (shard rebalancing or a larger runner for `warehouse-sources`) rather than a code change in this PR. Flagging for a maintainer — happy to re-run once the shard is healthy on master.

PR #70664, opened 2026-07-14
chore(agents): add repo skill for analyzing insights across teams

thmsobrmlr:
## TL;DR

When asked to "analyze insights across all teams", agents working in this repo stop at the current project's insights because the SQL guidance tells them other teams' data is inaccessible. In the dogfood project it *is* accessible — prod Postgres replicas (insights, dashboards, teams, US + EU) are synced into the data warehouse. This PR adds a repo-local skill telling agents exactly that, without changing the guidance every customer's agents see.

## Problem

The `querying-posthog-data` guidelines (inlined into the `execute-sql` tool description) state that another team's data cannot be accessed. That's correct as a general rule, but for agents working in this repo against the dogfood project it causes refused or incomplete cross-team insight analysis, because the synced warehouse replicas cover that data. An earlier revision of this PR relaxed the general guideline; per feedback this should be repo-scoped only, so that change is reverted here.

## Changes

- New repo-local skill `.agents/skills/analyzing-insights-across-teams/SKILL.md`: names the synced tables (US `postgres.posthog_*`, EU `eu_postgres_posthog_*`), the column-verification workflow, a validated example query, sync-lag and public-surface caveats.
- Revert of the earlier guideline edit in `products/posthog_ai/skills/querying-posthog-data/references/guidelines.md` (back to master's wording).

Docs/skill-only change. The example query and table names were validated live against the dogfood project's warehouse via `execute-sql` (catalog lookup through `system.information_schema.tables`, then the example aggregation). The skill's frontmatter was picked up by the agent harness skill list in-session.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — the change is itself agent-facing documentation, scoped to this repo.

## 🤖 Agent context

Produced with PostHog Code. Investigated why agents refuse cross-team insight analysis: traced the behavior to the team-scoping sentence in the `execute-sql` guidance, and confirmed via live `information_schema` queries that warehouse-synced replicas of the relevant prod tables exist and are queryable. First attempt edited the shared guidelines; reverted after feedback to keep the rule general and scope the exception to this repo as a `.agents/skills/` skill. Skills invoked: /writing-skills.

---

### DIFF
--- a/.agents/skills/analyzing-insights-across-teams/SKILL.md
+++ b/.agents/skills/analyzing-insights-across-teams/SKILL.md
@@ -20,6 +20,9 @@ production Postgres tables are replicated into the project's data warehouse,
 so cross-team entity metadata **is** queryable with `posthog:execute-sql`.
 Do not stop at `system.insights` when the question spans teams.
 
+This skill is deliberately repo-local (`.agents/skills/`): it documents PostHog's internal dogfood setup,
+applies only to agents working in this repo, and must not move into the packaged `products/*/skills/` bundle that ships to every team.
+
 ## Synced tables
 
 | Entity           | US (prod-us)                     | EU (prod-eu)                        |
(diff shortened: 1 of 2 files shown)


thmsobrmlr:
_Comment from PostHog Code on behalf of the PR author._

Addressed the two review-bot findings in b5934fe:

1. **"Skill Path Is Invisible"** — `.agents/skills/` placement is deliberate, not a packaging omission: this guidance is scoped to agents working in this repo checkout (the harness discovers `.agents/skills/` directly, same as the other ~50 skills there) and must not ship in the `products/*/skills/` bundle that goes to every team. The skill body now states this explicitly.

2. **"Cross-Team Metadata Lacks Gate"** — the access gate itself is project membership (the synced tables are only reachable by members of the internal dogfood project; a skill can't add authorization). What the skill *can* control is output handling, and that's now a required section rather than an advisory caveat: no customer team names or row-level metadata on public surfaces, `team_id`-level aggregates as the ceiling for public copy, and stop-don't-route-around on permission errors. The primary example no longer joins team names.

PR #70667, opened 2026-07-14
feat(ingestion): expose real-time debug API from rust ingestion consumer

jose-sequeira:
## Problem

There's no way to see what the ingestion consumer is actually doing in dev - batch lifecycle, deferrals, retries, worker health, commit order - short of tailing logs. The ingestion control plane UI needs live, structured data from the consumer to render this.

> [!NOTE]
> Stacked on #70660. Merge that first; this diff is only the debug API.

## Changes

Adds a real-time debug API to the consumer, behind `DEBUG_API_ENABLED` (default off), served from the existing health server and consumed by the ingestion control plane UI:

- `/debug/load` - cheap JSON snapshot (worker health + dispatcher in-flight/pins/stash), safe to poll fast.
- `/debug/state` - the same plus the retained event backlog.
- `/debug/events` - SSE stream: backlog replay, then live events (lagged subscribers skip dropped events instead of dying).

Under the hood a `DebugRecorder` (bounded in-memory buffer, 5000 events / 15 min, plus a broadcast channel) is injected into the consumer, dispatcher, transport, and worker registry, which record structured events at the same points that already emit metrics: batch dispatch/assignment/commit/failure, deferrals and flushes, send retries and exhaustion, worker state transitions and membership.

It's a pure observer: it never influences routing, and when disabled nothing is constructed or recorded.

All 172 Rust tests pass locally (lib incl. recorder unit tests, dispatcher/transport integration, e2e against real Kafka); fmt/clippy/shear clean.

Manual smoke test I (well, Claude) actually ran: booted the consumer with `DEBUG_API_ENABLED=true` against local Kafka with mock workers and a live producer - `/debug/load` returned the merged worker+dispatcher snapshot, and `/debug/events` streamed live SSE events (worker membership, state transitions, batch lifecycle) that the control plane UI consumes.

No new tests: the recorder's buffer/broadcast semantics were already unit-tested, and the wiring is exercised by the existing integration suites with the recorder injected as `None` (the production default when disabled).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Covered by the new README section in this PR; no `docs/` pages cover this consumer.

## 🤖 Agent context

Authored with Claude Code, directed by me. The recorder existed as a local prototype; Claude wired it in (module declaration, config flag, `Option<Arc<DebugRecorder>>` injection via pre-`Arc` setters to avoid constructor churn, SSE via `futures::stream::unfold` to avoid a `tokio-stream` dependency) and ran the tests and smoke test above. An earlier iteration also served a self-contained HTML debug page from the consumer; that moved into the ingestion control plane UI, so this PR ships only the JSON/SSE API.


rust/ingestion-consumer/src/config.rs
```diff
@@ -133,6 +133,23 @@ pub struct Config {
...
+    #[envconfig(from = "DEBUG_API_ENABLED", default = "false")]
+    pub debug_api_enabled: bool,
+
+    /// Shared secret callers must present as `X-Debug-Api-Secret` on every
+    /// `/debug/*` request. Dedicated to this one control-plane→consumer hop
+    /// (deliberately not `INTERNAL_API_SECRET` — see .agents/security.md).
+    /// Empty fails closed: the debug API is not mounted without it.
+    #[envconfig(from = "DEBUG_API_SECRET", default = "")]
+    pub debug_api_secret: String,
+
```
eli-r-ph, line 152: complete nit, feel free to ignore, more curious:

* Do these need a `from = ...`?
* Enable flag makes sense as raw bool 👍 but surprised this isn't an `Option<String>`

rust/ingestion-consumer/src/dispatcher.rs
```diff
@@ -613,6 +674,14 @@ impl Dispatcher {
         }
 
         gauge!("ingestion_consumer_dispatcher_pins_total").set(table.pins.len() as f64);
+        drop(table);
+
```
eli-r-ph, line 689: 🤔 why the new `drop` here?

eli-r-ph (approved):
❤️ fantastic change. These metrics endpoints will be really useful in control plane UI _and_ for monitoring, but I bet MCP will eventually be an even more powerful target

rust/ingestion-consumer/src/dispatcher.rs
jose-sequeira: To remove the lock, while doing the recording below

PR #70024, opened 2026-07-10
feat(tasks): rebind sandbox GitHub identity and git author to the actor

VojtechBartos:
## Problem

Stacked on #69618 (which rebinds the PostHog MCP identity per message). Commits and PRs from the agent sandbox are still authored by whoever created the task, even when a teammate is the one driving the turn — their work lands under someone else's name.

## Changes

Extends the per-message rebind to GitHub: when the acting user has a personal GitHub install covering the task's repository, the live sandbox's git remote token, `GITHUB_TOKEN`/`GH_TOKEN`, and `GIT_AUTHOR_*`/`GIT_COMMITTER_*` are swapped via the agentsh env file — which the exec wrapper re-sources per command, so the change reaches the agent's later `git`/`gh` calls (including `gh pr create`) with no sandbox restart. The facade's `rebind_sandbox_identity_for_user` gains the GitHub half; the Slack layer is untouched.

Guardrails, since two background systems re-apply GitHub credentials and would otherwise fight a swap:

- The credential refresh loop resolves the *swapped* integration instead of the task creator's, and token-rotation propagation skips sandboxes whose identity was swapped away — neither silently reverts a swap.
- A `github_user` identity mark lets same-actor messages skip integration resolution entirely (a DB query and, on a stale repo cache, a paginated GitHub API sync per message otherwise), and it survives swaps — including switching back to the creator.
- Sandbox liveness is checked before the token-rotation lock, so a dead sandbox never triggers a mint-and-revoke.
- Only user-authored, non-caller-token runs are ever swapped; actors without a covering install keep the previous identity (their message still goes through — they simply keep authoring as before).

## How did you test this code?

Automated only; manual multiplayer verification planned with the queue-workflow dev rollout.

- Six tests in `test_sandbox_credentials.py`, each pinning a distinct regression: the swap applies token + git author through the shared apply helper in one env write and records both identity marks; actors without a covering install, bot-authored runs, and same-identity messages are skipped; switching back to the creator re-applies their identity; and the creator fast path skips integration resolution on never-swapped runs.
- Full tasks process_task + slack_app forwarding suites green on this branch and on the base branch independently.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

N/A, internal credential mechanics behind the queue-workflow flag.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

[...]

PR #66212, opened 2026-06-25
feat(slack): add approval recovery flow

cvolzer3:
## Problem

Slack coworker approvals now only need user intervention for destructive actions after Phase 2. The next phase should build on that narrower contract instead of carrying the older generic broker flow forward.

Denied approvals also should not leave a run as a dead end. A denial is useful steering: the agent should receive the rejection context, re-plan around it when possible, and give the user a clear follow-up path when the run still ends terminally.

## Changes

- Routes all agent permission decisions (human approve/deny + broker auto-approvals) through the durable Temporal task workflow
- Agent uses denials as steering context that it can re-plan arounds
- Adds recovery hints & structure resume prompts for failed/cancelled runs
- Makes the acting Slack user (not the thread creator) the credential identity for every token minted during a run

## How did you test this code?

Ran the tests and thoroughly tested it locally

## Docs update

no

PR #70668, opened 2026-07-14
feat(tasks): rebind sandbox GitHub identity on actor transitions

VojtechBartos:
## Problem

Completes per-message identity for multiplayer Slack threads. #70454 makes the sandbox's MCP session follow the current speaker; this PR does the same for GitHub, and unifies both behind one gate so identity switching always happens in the same place.

Two GitHub-specific gaps today:

1. **No per-message switching.** GitHub credentials (token, git author) are injected at boot and re-applied by the background credential-refresh loop on token-TTL cadence. When a different user takes over a thread, commits, pushes, and PRs keep running as the previous actor.
2. **The refresh loop reads a stale actor.** `refresh_sandbox_credentials` receives the `TaskProcessingContext` captured at workflow start, whose `state` is a boot-time snapshot, so even its cadence refreshes resolve the boot-time actor forever. The actor-aware resolution added in #66212 never sees mid-run actor changes.

## Changes

One reconciliation gate, `ensure_sandbox_identity` (new `sandbox_identity.py` module), called from the follow-up delivery activity right before the turn runs. It compares the run's current actor against the identity last pushed to this sandbox, per surface:

```mermaid
flowchart TD
    M[Follow-up message arrives] --> A{{Resolve actor from run state}}
    A --> MCP{MCP: transition or freshness window expired?}
    A --> GH{GitHub: transition?}
    MCP -- yes --> RM[Mint OAuth token, push refresh_session]
    GH -- yes --> RG[Re-inject GitHub token + git author, notify authorship]
    RM --> K[(Sandbox-scoped identity + freshness marks)]
    RG --> K

    class A,RM,RG phBlue;
    class M phYellow;
    class K phGray;
```

- **MCP surface**: moved from `send_followup_to_sandbox.py` unchanged in behavior (transition bypasses the freshness window; same-actor repeats within the window skip).
- **GitHub surface** (new): transition-only, since token TTL between messages stays owned by the credential-refresh loop. On a speaker change it reuses `GitHubSandboxCredential.refresh` with the live run state (actor-aware token via #66212's resolution), rewrites the git author env (`get_git_identity_env_vars`), and notifies the agent-server with `refreshed_credentials=["github"]` and the effective PR authorship. A failed rebind leaves the mark untouched, so the next message retries.
- **Actor fallback policy**: a speaker without a usable GitHub connection falls through the existing resolution chain to the installation token, i.e. bot-authored work. Bot attribution beats wrong-person attribution.
- **Staleness fix**: `refresh_sandbox_credentials` re-reads the run state live before building credentials, so cadence refreshes also follow the current actor.
- The delivery activity input gains an optional `context` (the run's `TaskProcessingContext`, passed by both workflows); activities scheduled by pre-rollout histories deserialize without it and simply skip the GitHub surface.
- Boot records the GitHub session identity alongside the MCP one, keyed by sandbox id so replacement sandboxes start unmarked.

> [!NOTE]
> Stacked on #70454 (the diff base). The gate stays best-effort: a failed rebind never blocks follow-up delivery.

327 passed, 4 skipped across `process_task` + `execute_sandbox` suites locally. New coverage in `test_sandbox_identity.py`:

- `TestGithubIdentityTransitionGate` (new): a speaker change re-injects credentials with the *live* run state, rewrites the git author, notifies with authorship, and records the identity; same actor doesn't touch the sandbox; missing context / no GitHub credentials / missing sandbox id skip safely; a failed rebind leaves the mark unset so the next message retries; an unrefreshable credential (caller-token run) marks without notifying.
- MCP gate tests moved from `test_send_followup_to_sandbox.py` and retargeted to the shared gate, asserting identical behavior through `ensure_sandbox_identity`.

Not manually tested end to end; the log trail (`refresh_github_identity_transition`, `refresh_github_delivered`, `refresh_github_identity_failed`) mirrors the MCP one for rollout verification.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal behavior only, no docs affected.

## 🤖 Agent context

Claude Code (Claude Fable 5). Replaces closed #70024, rebuilt on master's run-state actor design. The gate extraction (rather than a second standalone refresh function) was an explicit ask: MCP and GitHub identity switching happen in the same place with the same transition semantics.

### DIFF
--- a/products/tasks/backend/temporal/execute_sandbox/workflow.py
+++ b/products/tasks/backend/temporal/execute_sandbox/workflow.py
@@ -1203,6 +1203,7 @@ class ExecuteSandboxWorkflow(PostHogWorkflow):
                 posthog_mcp_scopes=self._posthog_mcp_scopes,
                 artifact_ids=artifact_ids,
                 message_id=str(workflow.uuid4()),
+                context=self.context,
             ),
             start_to_close_timeout=timedelta(minutes=35),
             # See process_task: heartbeat detects worker restarts, message_id
--- a/products/tasks/backend/temporal/process_task/activities/get_task_processing_context.py
+++ b/products/tasks/backend/temporal/process_task/activities/get_task_processing_context.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import timedelta
 
 from django.conf import settings
@@ -93,6 +93,19 @@ class TaskProcessingContext:
     # The kill-switch flag wins over everything; otherwise a per-run state override
     # (the user's toggle) applies. Captured at workflow start so it's stable across retries.
     rtk_enabled: bool = True
+
+    def with_state(self, state: dict | None) -> "TaskProcessingContext":
+        """Copy of this context rebased onto the given (typically live) run state.
+
+        The context is captured at workflow start, so its ``state`` is a
+        boot-time snapshot while the run's actual state moves on (multiplayer
+        follow-ups update the actor mid-run). Anything resolving the *current*
+        actor must go through a rebased copy. The other fields intentionally
+        stay boot-time: they are either immutable for the run (ids, repository)
+        or last-resort fallbacks (``github_user_integration_id``,
+        ``distinct_id``) that actor-aware resolution overrides via state.
+        """
+        return replace(self, state=state)
 
     @property
     def mode(self) -> str:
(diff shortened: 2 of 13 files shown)


VojtechBartos:
Superseded by #71941, which does rebind **or** log out on failure (a superset of the rebind-only behavior here) and is built on the current, rebased PR3. Closing this stale branch.

PR #70671, opened 2026-07-14
feat(tasks): add team and user default AI run preferences

VojtechBartos:
## Problem

Every surface that starts a coding-agent task run picks the model independently. The Slack app has its own per-workspace/per-user AI preferences, the web tasks UI hardcodes its picker defaults, PostHog Code keeps a device-local "last used model", and the tasks backend applies no default at all. There is no way to set a project-wide default model, and no per-user preference that follows you across devices and surfaces.

This is the backend part of a three-PR series (backend → web frontend → PostHog Code client).

## Changes

Adds a central preference layer for the AI run triple (`runtime_adapter`, `model`, `reasoning_effort`), applied whenever a task run is created without an explicit runtime selection, from any surface.

**Models** (`products/tasks/backend/models.py` + migration)
- `TeamTasksConfig` - team extension model (one row per project) with an `ai_run_preferences` JSON triple.
- `UserTasksConfig` - per (user, project) override, fail-closed via `TeamScopedRootMixin`.
- Both use `db_constraint=False` FKs so the migration is pure `CREATE TABLE` with no locks on the hot team/user tables.

**Resolution** (`products/tasks/backend/logic/services/ai_run_defaults.py`, re-exported via `facade/ai_run_defaults.py`)
- Precedence: explicit per-run values > user preference > team default > none.
- Whole-triple swap, never field-by-field blending, mirroring the Slack app's `resolve_ai_preferences` semantics. A partial pin (adapter or model alone) counts as explicit and is left untouched.
- Lenient at resolve time: unknown model ids pass through (the agent server owns final fallback), a triple with a retired adapter falls through to the next level, an unsupported reasoning effort is dropped.

**Application points**
- `Task.create_run()` is the safety net: every run creation flows through it, so Slack, Max tools, the wizard, automations, and the REST API all inherit defaults with no caller changes. Internal (infra-agent) tasks are excluded. Injection also derives `provider` and the Codex `initial_permission_mode="auto"` side effects.
- The warm-run paths (`run_task`, `create_task`, `warm_task_sandbox`) resolve defaults before warm matching, so a warm run provisioned under a default triple still matches a submit that pinned nothing.
- Slack preferences arrive as explicit values, so the layering is: per-run pick > Slack user > Slack workspace > central user > central team.

**API**
- `GET/POST /api/projects/:id/tasks/config/` - project default (singleton, `task:read`/`task:write` scopes).
- `GET/POST /api/projects/:id/tasks/my_config/` - the requesting user's preference, plus a `resolved_ai_run_defaults` block that tells clients (web picker, PostHog Code) what a run will use if they send nothing. Session, personal API key, and OAuth auth, same stack as the other PostHog Code endpoints.
- OpenAPI regenerated; generated `tasksConfig*`/`tasksMyConfig*` client functions land in `products/tasks/frontend/generated/`.

Also fixes a pre-commit papercut hit while producing this PR: `hogli format:js` now passes `--no-error-on-unmatched-pattern` to oxlint, so commits whose only staged JS files are generated (oxlint-ignored) ones no longer fail lint-staged.

Ran locally:
- `hogli test products/tasks/backend/tests/test_ai_run_defaults.py` - 25 new tests. Regressions they catch: whole-triple swap turning into field blending (user's effort-less triple must not inherit the team's effort), partial pins getting "completed" from preferences, retired adapters/unsupported efforts leaking into run state, `create_run` losing the provider/permission-mode side effects, internal tasks inheriting preferences, acting-user threading falling back to the task creator, a default-carrying warm run no longer matching a pinless submit, and the two endpoints' precedence/validation wiring.
- `hogli test products/tasks/backend/tests/test_warm_facade.py test_models.py test_max_tools.py test_facade.py test_api.py` - 811 existing tests pass, confirming no behavior change for runs that pin their runtime explicitly.
- `hogli ci:preflight --fix` - clean.

Not manually tested in a running app; the settings UI arrives in the follow-up frontend PR.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code. Skills invoked: /django-migrations, /improving-drf-endpoints, /writing-tests. Design decisions worth knowing for review: preferences are stored as one JSON triple (not three columns) so they validate and swap as a unit, matching `SlackSettings.ai_preferences`; two models instead of one nullable-user table because the team default is a Team-extension singleton with different permissions than the self-serve user rows; `Task.create_run` was chosen as the injection choke point after verifying every run-creation path flows through it, with early resolution added only where warm-run matching compares triples. Model-id membership is deliberately not validated at write time because the gateway model list drifts. The `tasks/config` and `tasks/my_config` routes are registered before the `tasks` viewset so its `{pk}` detail pattern can't swallow them.


VojtechBartos:
* **#70683** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#70671** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/70671?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #70672, opened 2026-07-14
feat(insights): New date picker experiment

sampennington:
## Problem

The insight date filter is a long single-column list, and two shipped backend capabilities — `dateRange.daysOfWeek` (day-of-week exclusion) and `dateRange.excludeIncompletePeriods` — have no UI. We want to A/B test a compact, chip-based redesign against it, without modifying the design system for an experiment that might get scrapped.

This is the **app-only** incarnation of the design explored in #70591: identical UI, but every new piece lives in `frontend/src/`, and `packages/quill` is untouched. If the experiment wins, #70591 is the ready-made graduation path that folds the panel natively into `DateTimePicker`.

## Changes

Everything is behind `product-analytics-quill-date-filter`; with the flag off nothing changes.

**New app components** (`lib/components/DateFilter/`), all composed from existing quill primitives and the stock `DateTimePicker` (its documented `ranges={[]}` mode hides the built-in quick list):

- `DateRangePresetsPanel` — 3×3 rolling chips (`1h 24h 7d / 14d 30d 90d / 1w 1m 1y`), a "Last N units" stepper row, 2-column named periods (`Today … All time`), and a "Custom range…" expander row. Selection model (`rolling`/`fixed`/`custom`) plus range projection helpers, dayjs-based.
- `RelativeRangeInput` — the "N units" stepper (count + unit dropdown), built on quill `InputGroup`/`Select`.
- `QuillDateFilter` — the surface: panel beside a stock `DateTimePicker` that mounts when "Custom range…" expands (key-remount reseeds it when chips change the selection or the "Include time" toggle flips; staging semantics stay the picker's own).
- `DateFilterExclusionsControl` — the Exclude flyout (incomplete period + exclude-day chips).
- `dateRangeSelection.ts` — PostHog date-string ↔ selection mapping, with 20 round-trip tests.

**Insights integration:**

- `InsightQuillDateFilter` wires the surface to insight kea logics: rolling chips ↔ `-30d`-style strings, named periods through `dateMapping`, custom dates ↔ ISO (`explicitDate` when time is included). Exclusions write `daysOfWeek` (trends-only, hidden while smoothing) and `excludeIncompletePeriods` (hidden for retention); the exact-time switch writes `explicitDate`. These rows ride the panel's footer slot — other surfaces adopting the panel simply don't pass them.
- Flag-gated swap in `InsightDisplayConfig`; active exclusions surface in the results metadata line.

**`lemon-skin.scss`** (prototype): restyles quill primitives to Lemon's visual language behind a `data-lemon-skin` opt-in. Chips map to LemonTag metrics; `!important` Tailwind utilities are neutralized by re-scoping the token they read (`--muted-3000` for the footer band, `--text-xs` inside the stepper) rather than fighting the declaration.

Known semantic divergence, flagged deliberately: the `1w`/`1m` chips write rolling `-1w`/`-1m`, not the calendar-anchored `-1wStart/-1wEnd`. Production click data says `180d` (25k clicks/month) lost its chip in the 3×3 trim — those users fall back to the stepper; worth watching in the experiment's re-adjustment metric.

- 20 parameterized jest tests on the vocabulary mapping round-trip (`-7d`, `-1mStart/-1mEnd`, `all`, ISO customs, time preservation, garbage fallback) — the seam where a regression would silently rewrite saved queries. Ran locally, all pass.
- Frontend typecheck passes for all touched files.
- I (well, Claude) could not visually verify this exact branch in the running app — the local dev server needed a restart at handoff time. The identical UI composition was verified extensively in-app and in Storybook on #70591; the `Components/Quill Date Filter` story here renders both skinned and quill-native variants for review.
- With the flag off the legacy `InsightDateFilter` renders untouched.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None — dark-shipped experiment behind a flag; no user-facing docs change until rollout.

## 🤖 Agent context

- Authored by Claude in a PostHog Code session directed by Sam, as the app-only re-cut of #70591 (same design, zero design-system changes) after deciding quill shouldn't carry concept churn for an experiment.
- Skills invoked: /writing-tests (on the ancestor branch), /storyshot.
- Decisions: the panel/stepper/surface live in `lib/components/DateFilter/` composed from stock quill exports; the stock picker is remounted via `key` instead of the `followValue` sync that the quill-native variant added; chip vocabulary (3×3, `180d` dropped) and sizing came from Sam's visual iteration plus production autocapture click data on the legacy filter; `1w`/`1m` stay rolling for interaction consistency.


frontend/src/queries/nodes/InsightViz/InsightDisplayConfig.tsx
```diff
@@ -78,7 +82,11 @@ export function InsightDisplayConfig(): JSX.Element {
             <div className="flex items-center gap-x-2 flex-wrap gap-y-2">
                 {!isRetention && (
                     <ConfigFilter>
-                        <InsightDateFilter disabled={isFunnels && !!isEmptyFunnel} />
+                        {useQuillDateComposer ? (
+                            <InsightQuillDateFilter disabled={isFunnels && !!isEmptyFunnel} />
```
sampennington, line 86: lets not say composer

frontend/src/lib/constants.tsx
```diff
@@ -428,6 +428,7 @@ export const FEATURE_FLAGS = {
     PRODUCT_ANALYTICS_INSIGHTS_TOOLTIPS: 'product-analytics-insights-tooltips', // owner: #team-product-analytics, gates the unified quill DefaultTooltip for trends/retention/stickiness insight charts
     PRODUCT_ANALYTICS_PATHS_V2: 'paths-v2', // owner: @thmsobrmlr #team-product-analytics
     PRODUCT_ANALYTICS_QUARTER_YEAR_INTERVALS: 'product-analytics-quarter-year-intervals', // owner: @sampennington #team-product-analytics, gates quarter/year interval selection on insights
+    PRODUCT_ANALYTICS_QUILL_DATE_COMPOSER: 'product-analytics-quill-date-composer', // owner: @sampennington #team-product-analytics, swaps the insight date filter for the quill DateTimePicker presets panel (lemon-skinned)
```
sampennington: dont use composer

frontend/src/lib/components/DateFilter/QuillDateFilter.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    showTime={includeTime}
+                    weekStartsOn={weekStartsOn}
+                    value={valueForSelection(selection, now, weekStartsOn)}
+                    onApply={({ start, end }) => {
+                        onApplyCustom({ kind: 'custom', start, end, includesTime: includeTime })
+                        setCalendarOpen(false)
+                    }}
+                    onCancel={() => setCalendarOpen(false)}
+                    // flex-col + separator mt-auto pin the Cancel/Apply band to the surface bottom
+                    // when the panel column is the taller of the two
```
sampennington: we dont need comments here

frontend/src/lib/components/DateFilter/dateRangeSelection.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    ] as const)('resolves unrepresentable %s / %s to a custom range', (dateFrom, dateTo) => {
+        const selection = selectionForDateRange(dateFrom, dateTo)
+        expect(selection.kind).toBe('custom')
+        const custom = selection as Extract<typeof selection, { kind: 'custom' }>
+        expect(custom.start.getTime()).toBeLessThan(custom.end.getTime())
+    })
+
+    it('quarters resolve to roughly the right span, not a 7-day fabrication', () => {
+        const selection = selectionForDateRange('-2q', null)
+        if (selection.kind !== 'custom') {
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Blocker**

Branching logic inside a test body: `if (selection.kind !== 'custom') { throw ... }`. Use the same non-branching cast the earlier block uses — `const custom = selection as Extract<typeof selection, { kind: 'custom' }>` — after asserting `expect(selection.kind).toBe('custom')`, and drop the `if`.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    selection={selection}
+                    onSelectionChange={(next) => {
+                        setPendingSelection(next)
+                        // Rolling changes ride the stepper click-by-click, so let them debounce
+                        updateDateRange(dateRangeForSelection(next), next.kind !== 'rolling')
+                        if (next.kind !== 'rolling') {
+                            setOpen(false)
+                        }
+                    }}
+                    onApplyCustom={(next) => {
```
sampennington, line 136: 🤖 **sp-ship** · 💻📐 `[code+xp]` · **Blocker**

Two switches write the same query field and can silently fight: "Exact time range" (this footer) binds `dateRange.explicitDate` directly, while the calendar's "Include time" toggle flows through this `onApplyCustom` → `dateRangeForSelection`, which always emits `explicitDate: includesTime`. Turning on "Exact time range" and then applying a custom range with "Include time" off silently flips `explicitDate` back to false. Seed `includeTime` from the current `explicitDate` (or collapse the two toggles into one) so they can't disagree.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const { insightProps, editingDisabledReason } = useValues(insightLogic)
+    const { dateRange, trendsFilter, isTrends, isRetention } = useValues(insightVizDataLogic(insightProps))
+    const { updateDateRange, updateQuerySource } = useActions(insightVizDataLogic(insightProps))
+    const { weekStartDay } = useValues(teamLogic)
+    const [open, setOpen] = useState(false)
+
+    const storeSelection = selectionForDateRange(dateRange?.date_from ?? '-7d', dateRange?.date_to)
+    // The store updates only after the logic's 300ms debounce — reflect picks immediately
+    // so rapid stepper presses accumulate instead of re-deriving from a stale value.
+    const [pendingSelection, setPendingSelection] = useState<DateRangeSelection | null>(null)
```
sampennington, line 58: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion**

For an open-ended absolute range (`date_from='2026-07-01'`, `date_to=null`), `selectionForDateRange` defaults `end` to `dayjs()`, so `storeSelectionKey` changes every render and this reset effect fires spuriously (harmless today — `setPendingSelection(null)` on null bails — but it defeats the accumulation guard for that state). Consider a key that ignores the now-based `end`, or memoizing the selection off `date_from`/`date_to`.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const storeSelectionKey = selectionKeyOf(storeSelection)
+    useEffect(() => {
```
sampennington, line 60: 🤖 **sp-ship** · ⚛️ `[react]` · **Suggestion**

`rerender-derived-state-no-effect` — resetting `pendingSelection` in a `useEffect` keyed on `storeSelectionKey` costs an extra commit+render after the store catches up. The render-phase idiom (store the previous key in state, reset during render when it changes) drops that pass. Minor; the effect version is correct.

frontend/src/lib/components/DateFilter/dateRangeSelection.ts
```diff
@@ -0,0 +1,67 @@
...
+const ROLLING_DATE_FROM = /^-(\d+)([hdwmy])$/
+const UNIT_BY_LETTER: Record<string, RelativeRangeUnit> = {
+    h: 'hours',
+    d: 'days',
+    w: 'weeks',
+    m: 'months',
+    y: 'years',
+}
+const LETTER_BY_UNIT: Record<RelativeRangeUnit, string> = {
+    minutes: 'h', // not offered in the picker; mapped defensively
```
sampennington, line 19: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion**

`minutes: 'h'` maps minutes to the hours letter, so a `{unit:'minutes'}` selection would silently serialize as an hours range. Unreachable today (the stepper omits minutes), but the type can express something the serializer gets wrong — consider dropping `minutes` from the union instead of mapping it incorrectly.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+import { computeDaysOfWeekUpdate, getExcludedDaysOfWeek, type IsoDayOfWeek } from './daysOfWeekFilterUtils'
+
+// Chip labels double as dateMapping keys; filter at module load so a renamed chip can never
+// silently produce the wrong range.
+const NAMED_CHIPS = ['Today', 'Yesterday', 'This week', 'This month', 'Year to date', 'All time'].filter((name) =>
+    dateMapping.some(({ key }) => key === name)
+)
+
+function shortExclusionsLabel(parts: string[]): string {
```
sampennington, line 38: 🤖 **sp-ship** · ♻️✨📐 `[reuse+quality+xp]` · **Suggestion**

`shortExclusionsLabel` is duplicated verbatim here and in `QuillDateFilter.stories.tsx`, and the "Exact time range" label + info-tooltip JSX block is duplicated there too — user-facing copy that will drift. Export the label helper from `DateFilterExclusionsControl.tsx` (next to `dateFilterExclusionParts`) and consider lifting the exact-time row into a small shared component the story imports.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+        return selection.name
+    }
+    return `${dayjs(selection.start).format('MMM D')} – ${dayjs(selection.end).format('MMM D')}`
+}
+
+function rollingStart(count: number, unit: RelativeRangeUnit, now: Dayjs): Dayjs {
+    return now.subtract(count, unit)
+}
+
+function startOfWeek(date: Dayjs, weekStartsOn: 0 | 1): Dayjs {
```
sampennington: 🤖 **sp-ship** · ♻️ `[reuse]` · **Suggestion**

This `startOfWeek` is a line-for-line reimplementation of the private helper in `lib/utils/dateFilters.ts:9`. If the panel stays app-side, consider exporting and reusing that one so the week-start math is defined once.

frontend/src/lib/components/DateFilter/DateFilterExclusionsControl.tsx
```diff
@@ -0,0 +1,142 @@
...
+    days: string[]
+    incomplete: boolean
+}
+
+export function dateFilterExclusionParts({ days, incomplete }: DateFilterExclusions): string[] {
+    const parts: string[] = []
+    if (days.length > 0) {
+        const sorted = [...days].sort().join(',')
+        parts.push(
+            sorted === '6,7'
```
sampennington: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion**

Nested ternary (`sorted === '6,7' ? … : sorted === '1,2,3,4,5' ? … : …`) — disallowed in this codebase; rewrite as an if/else chain. Related: this is now the second days→label vocabulary (`daysOfWeekLabel` in daysOfWeekFilterUtils says "Sat, Sun" where this says "weekends"/"N days"), so the trigger and the results-metadata note describe the same state differently — worth one function owning days→label.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+            return { start, end: start.add(6, 'days').endOf('day') }
+        }
+        case 'This month':
+            return { start: now.startOf('month'), end: now }
+        case 'Last month':
+            return { start: now.subtract(1, 'month').startOf('month'), end: now.subtract(1, 'month').endOf('month') }
+        case 'Year to date':
+            return { start: now.startOf('year'), end: now }
+        default:
+            return { start: now.subtract(10, 'years'), end: now }
```
sampennington: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion**

The "All time" fallback `now.subtract(10, 'years')` is a magic span in the `default` branch. Name it (e.g. `ALL_TIME_SEED_YEARS`) or note that it only seeds the calendar view.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+    { label: '14d', selection: { kind: 'rolling', count: 14, unit: 'days' } },
+    { label: '30d', selection: { kind: 'rolling', count: 30, unit: 'days' } },
+    { label: '90d', selection: { kind: 'rolling', count: 90, unit: 'days' } },
+    { label: '1w', selection: { kind: 'rolling', count: 1, unit: 'weeks' } },
+    { label: '1m', selection: { kind: 'rolling', count: 1, unit: 'months' } },
+    { label: '1y', selection: { kind: 'rolling', count: 1, unit: 'years' } },
+]
+const DEFAULT_NAMED_CHIPS = ['Today', 'Yesterday', 'This week', 'This month', 'Year to date', 'All time']
+
+export function dateRangeSelectionLabel(selection: DateRangeSelection): string {
```
sampennington, line 50: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion**

`dateRangeSelectionLabel` is only used by the Storybook story (production uses `dateFilterToText`). If it isn't intended panel API, drop the export to avoid dead production code.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+        case 'Last month':
+            return { start: now.subtract(1, 'month').startOf('month'), end: now.subtract(1, 'month').endOf('month') }
+        case 'Year to date':
+            return { start: now.startOf('year'), end: now }
+        default:
+            return { start: now.subtract(10, 'years'), end: now }
+    }
+}
+
+/** Projects a selection to a concrete range for seeding the calendar. */
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion**

A few added doc comments restate what the name/signature already says and can be deleted: this one, the `onCalendarOpenChange`/`footer` prop docs below (lines 145/148), `presetsFooter` in QuillDateFilter.tsx:25, and `units` in RelativeRangeInput.tsx:25.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+export function valueForSelection(selection: DateRangeSelection, now: Dayjs, weekStartsOn: 0 | 1): DateTimeValue {
```
sampennington, line 91: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion**

`valueForSelection`, `fixedRange`/`startOfWeek` (week-boundary math parameterized on `weekStartsOn`), and `selectionKeyOf` ship untested. `now` is a parameter, so they test deterministically with no fake timers — and `selectionKeyOf` is load-bearing (reset dep + remount key in the host; a key collision between distinct selections would wedge the UI). Worth a small table test with a frozen now.

frontend/src/scenes/insights/filters/InsightDateFilter/daysOfWeekFilterUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export const ALL_DAY_NUMBERS: IsoDayOfWeek[] = [1, 2, 3, 4, 5, 6, 7]
+
+function sortDays(days: IsoDayOfWeek[]): IsoDayOfWeek[] {
+    return [...days].sort((a, b) => a - b)
+}
+
+/** [] means all days on both sides of the inversion. */
+export function invertDaysOfWeek(days: IsoDayOfWeek[]): IsoDayOfWeek[] {
```
sampennington, line 25: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion**

This module decides which days get filtered out of queries and what the "Excluding …" note reads, and ships untested. The inversion (incl. the `[] means all days` case) and the `0/7 → null` normalization are classic silent off-by-one territory — a table test over the excluded→update→stored→`getExcludedDaysOfWeek` round trip plus the label cases would be cheap and catch it.

sampennington (commented):

Two blockers: a branching `if`/throw inside a test body, and the "Exact time range" / "Include time" pair both writing `explicitDate` so a custom-range Apply can silently undo the other switch — worth resolving before experiment users see it. Beyond that the composition is clean and well-isolated behind the flag; the suggestions cluster around consolidating duplicated label/tooltip helpers and adding cheap deterministic tests for the new pure date/day-inversion logic.

frontend/src/lib/components/DateFilter/dateRangeSelection.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+        expect(dateRangeForSelection(selection)).toEqual({
+            date_from: '2026-07-01T09:30:00',
+            date_to: '2026-07-01T17:00:00',
+            explicitDate: true,
+        })
+    })
+
+    it('falls back to a concrete last-7-days custom range for unparseable input', () => {
+        const selection = selectionForDateRange('garbage', null)
```
sampennington, line 70: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion**

The `'garbage'` fallback and the "unrepresentable" cases assert only `kind === 'custom'` (+ `start < end`) — a regression producing the wrong span would pass. Consider asserting the fallback is ~7 days, mirroring the quarter-span test.

frontend/src/queries/nodes/InsightViz/InsightDisplayConfig.tsx
sampennington: 🤖 **sp-triage**

Human review comment — leaving open for @sampennington to action directly rather than making the wording call myself.

frontend/src/lib/constants.tsx
sampennington: 🤖 **sp-triage**

Human review comment — leaving open for @sampennington to action directly rather than making the wording call myself.

frontend/src/lib/components/DateFilter/QuillDateFilter.tsx
```diff
@@ -0,0 +1,102 @@
...
+export function QuillDateFilter({
+    selection,
+    onSelectionChange,
+    onApplyCustom,
+    shortChips,
+    namedChips,
+    weekStartsOn = 1,
+    presetsFooter,
+    portalProps,
+}: QuillDateFilterProps): JSX.Element {
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Same root cause as the sp-swarm Blocker on InsightQuillDateFilter.tsx:126 (the calendar's local `includeTime` state and the footer's `explicitDate` switch can independently write the same query field). Reconciling them is a UX call — which switch wins, whether to seed one from the other on open, or collapse them into one control — so deferring both together rather than picking an approach unilaterally.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const triggerLabel =
+        dateFilterToText(dateRange?.date_from ?? '-7d', dateRange?.date_to, 'Last 7 days', dateMapping, false) ??
+        'Last 7 days'
+
+    // The picker speaks excluded days; the query schema stores included days. The legacy
+    // display-only trendsFilter.hideWeekends is deliberately NOT folded in — different semantics.
+    const excludedDays = getExcludedDaysOfWeek(dateRange)
+    // The backend rejects daysOfWeek together with smoothing, so don't offer it
+    const smoothingActive = isTrends && (trendsFilter?.smoothingIntervals ?? 1) > 1
+    const showExcludedDays = isTrends && !smoothingActive
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: clearing `dateRange.daysOfWeek` whenever the exclusions control is hidden (smoothing turned on or the insight type leaves trends), so the query can't keep a stale filter the UI no longer exposes.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const showIncompletePeriod = !isRetention
+    const exclusions: DateFilterExclusions = {
+        days: isTrends ? excludedDays.map(String) : [],
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422 alongside the smoothing case: the same effect clears `dateRange.daysOfWeek` whenever `showExcludedDays` goes false, covering both the smoothing and non-trends transitions.

frontend/src/lib/components/DateFilter/dateRangeSelection.test.ts
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: replaced the branching `if/throw` with the non-branching cast used in the block above.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
sampennington: 🤖 **sp-triage**

Agreed this is a real bug, but the fix requires a UX call: whether "Include time" should seed from the current "Exact time range" state on open, which switch should win when they disagree, or whether the two controls should collapse into one. Deferring to @sampennington rather than picking an approach — leaving this open.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
sampennington: 🤖 **sp-triage**

Agreed on the mechanism, but as noted it's harmless today (`setPendingSelection(null)` bails on null) — resolving as a known, non-blocking edge case rather than adding key-derivation complexity for it now.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
sampennington: 🤖 **sp-triage**

Agreed the render-phase idiom would drop the extra commit, but per the comment itself the effect version is correct — resolving as a minor, non-blocking optimization we're not taking right now.

frontend/src/lib/components/DateFilter/dateRangeSelection.ts
sampennington: 🤖 **sp-triage**

Agreed `minutes` shouldn't map to `'h'`, but `RelativeRangeUnit` is a shared type used well beyond this stepper (e.g. DateRangePicker's rolling options), so narrowing it safely means auditing every call site, not a local edit. Deferring as follow-up rather than doing that here.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
sampennington: 🤖 **sp-triage**

Agreed on the duplication risk, but extracting a shared label helper and a shared exact-time-row component spans this file and `QuillDateFilter.stories.tsx` and touches copy/composition decisions — deferring as a follow-up rather than doing that consolidation in this pass.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: exported the existing `startOfWeek` helper from `lib/utils/dateFilters.ts` and switched the panel to import it instead of reimplementing the same math.

frontend/src/lib/components/DateFilter/DateFilterExclusionsControl.tsx
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: rewrote the nested ternary as an if/else chain. Left the days→label vocabulary duplication vs `daysOfWeekLabel` as a separate follow-up — consolidating those is a bigger call than this pass.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: named it `ALL_TIME_SEED_YEARS` with a comment on why it only seeds the calendar.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
sampennington: 🤖 **sp-triage**

Whether this is intended panel API or should be Storybook-only is a scope call for @sampennington — deferring rather than dropping the export myself.

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: removed the four doc comments that only restated the name/signature (this one, the `onCalendarOpenChange`/`footer` prop docs, `presetsFooter` in QuillDateFilter.tsx, and `units` in RelativeRangeInput.tsx).

frontend/src/lib/components/DateFilter/DateRangePresetsPanel.tsx
sampennington: 🤖 **sp-triage**

Agreed this coverage is worth adding, but writing it well (a frozen-`now` table test) is a real addition rather than a mechanical fix — tracking as a follow-up instead of adding it in this triage pass.

frontend/src/scenes/insights/filters/InsightDateFilter/daysOfWeekFilterUtils.ts
sampennington: 🤖 **sp-triage**

Agreed this module deserves a round-trip table test, but that's new test-writing rather than a mechanical fix here — tracking as a follow-up instead of adding it in this pass.

frontend/src/lib/components/DateFilter/dateRangeSelection.test.ts
sampennington: 🤖 **sp-triage**

Fixed in 16a20788abfbc631371d59ff0bdb15809f0a8422: the fallback test now asserts the span is ~7 days, mirroring the quarter-span test.

frontend/src/queries/nodes/InsightViz/InsightDisplayConfig.tsx
sampennington: 🤖 **sp-triage**

Fixed in `bbefb111d66` — flag renamed to `PRODUCT_ANALYTICS_QUILL_DATE_FILTER` / `product-analytics-quill-date-filter`, and the `useQuillDateComposer`/`composerEnabled` locals renamed to match. No "composer" left in the PR.

frontend/src/lib/constants.tsx
sampennington: 🤖 **sp-triage**

Fixed in `bbefb111d66` — flag renamed to `PRODUCT_ANALYTICS_QUILL_DATE_FILTER` / `product-analytics-quill-date-filter`, and the `useQuillDateComposer`/`composerEnabled` locals renamed to match. No "composer" left in the PR.

frontend/src/scenes/insights/filters/InsightDateFilter/InsightQuillDateFilter.tsx
sampennington: 🤖 **sp-triage**

Fixed in `b7b2e58dc67` — per the author's direction, this now maps exactly to the legacy filter's semantics: applying a range never writes `explicitDate` (only the "Exact time range" toggle does), and the calendar's "Include time" is purely serialization precision. The two switches can no longer overwrite each other.

frontend/src/lib/components/DateFilter/QuillDateFilter.tsx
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Fixed in `b7b2e58dc67` — per the author's direction, this now maps exactly to the legacy filter's semantics: applying a range never writes `explicitDate` (only the "Exact time range" toggle does), and the calendar's "Include time" is purely serialization precision. The two switches can no longer overwrite each other.

thmsobrmlr (approved):

PR #69741, opened 2026-07-09
feat(insights): add line style option to SQL insight display settings

sampennington:
## Problem

Follow-up to #69700, which added a per-insight Smooth | Straight line control to the insight-viz Options menu. SQL insights render their line charts through the same quill pipeline (so they get the same smooth default from `CHART_CONFIG_DEFAULTS`), but their display settings live in the SQL editor's Display tab and had no equivalent control.

## Changes

- `chartStyle?: ChartStyle` on `ChartSettings` (reusing the type from #69700), plus regenerated schema artifacts.
- "Line style" segmented control (Smooth | Straight) in the SQL insight Display tab's General panel, shown for line and area visualizations.
- The curve flows through `chartStyleCurve` into both `buildLineChartConfig` and `buildComboChartConfig` in `sqlLineGraphAdapter`, so combo charts' line series honor it too. An explicit value wins; unset falls through to the app default, which is smooth.
- `mergeChartSettings` deep-merges `chartStyle` like the other nested settings objects (`pie`, y-axis settings), so future `ChartStyle` fields won't clobber each other.

## Merge order

Land #70675 (removes the `quill-sql-charts` flag and the legacy chart.js path) first. The control is ungated, and the curve override only has an effect on the quill renderer — the chart.js path ignores it.

## How did you test this code?

I'm an agent, and I ran automated tests only:

- `sqlLineGraphAdapter.test.ts` and `DisplayTab.test.tsx` pass, plus `tsgo --noEmit` over the whole frontend.
- No new tests: the curve passthrough is the same one-field mechanism already covered by the `compareQuery` and transform tests in #69700.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

None needed. This mirrors the insight-viz control from #69700, and the SQL Display tab's individual options aren't enumerated in docs.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

- Authored with Claude Code, driven by @sampennington, as a follow-up to #69700 in the same session.
- Originally stacked on the #69700 branch, but that PR merged mid-flight, so this was rebased onto master (resolving one overlap with the just-merged SQL value-labels work in `DisplayTab`).
[...]

PR #70675, opened 2026-07-14
chore(insights): remove quill-sql-charts flag and legacy chart.js path

sampennington:
## Problem

SQL (DataVisualization) insight charts have been rendering through the new
[@posthog/quill-charts](https://github.com/PostHog/posthog) renderers behind the
`product-analytics-quill-sql-charts` flag. Now that the quill charts are the default,
the flag and the old chart.js code paths are dead weight.

**Why:** the rollout is complete — quill charts are the future for SQL insights — so this
removes the flag and cleans up the legacy renderers left behind it.

All the consumers that previously rode this module from outside SQL insights have since
been moved off it (#70834, #70837, #70835, #74136), so nothing outside SQL insights
depends on the paths being deleted here.

## Changes

- Remove `PRODUCT_ANALYTICS_QUILL_SQL_CHARTS` from `FEATURE_FLAGS`.
- Always route non-pie SQL charts to the quill `SqlLineGraph` / `SqlBarGraph` / `SqlComboGraph`
  renderers; delete the `LegacyLineGraph` chart.js implementation and its `lineGraphLogic`.
- Always render the quill `SqlPieGraph`; delete `LegacyPieChart` (unreachable once the always-true
  `canRenderSqlPieGraph` guard is dropped) and the now-unused `uniqueKey` prop.
- **Rename the SQL chart dispatcher `LineGraph` → `SqlChart`** (and `LineGraphProps` →
  `SqlChartProps`). After the legacy renderer was removed it's just a dispatcher for line/area/bar/
  stacked-bar/combo, so the old name was misleading and collided with the legacy trends `LineGraph`.
- Make the renderers' `SqlChartProps` imports type-only — `SqlChart` imports the three renderers and
  each imported the type back from it, an import cycle oxlint flagged on the new module.
- Update the affected tests, the SQL chart stories, and the `render-data-visualization` test
  harness so they no longer toggle the flag.

Net ~1000 lines of dead code removed.

The one remaining importer of this module from outside SQL insights is Customer analytics'
`AccountBillingChart`, which uses `sqlLineGraphAdapter`'s *pure builders* only (no React layer) —
a residual accepted in #74136's review. It touches none of the paths deleted here.

- Rebased on master; DataVisualization charts + Customer analytics Accounts Jest suites pass
  (21 suites, 402 tests).
- `tsgo --noEmit` reports no errors in any changed file. The remaining whole-app errors are
  pre-existing unbuilt-quill-package module resolution noise in a fresh checkout, all in unrelated
  directories; `@posthog/quill-charts` itself resolves, so the chart files are genuinely checked.
- `oxlint` and `oxfmt --check` clean on the changed files (oxlint is now warning-free here — the
  import cycle above was the last one).
- Storybook / Visual Review covers the render swap: the SQL chart stories no longer set the flag,
  so they exercise the quill path by default.

## Note for reviewers

#69741 (line style option for SQL insight display settings) also edits `sqlLineGraphAdapter.ts`.
It's mergeable against master today, but whichever of the two lands second will need a small
rename-only conflict fix (`LineGraphProps` → `SqlChartProps`).

## 🤖 Agent context

Authored with PostHog Code (Claude Code). Invoked the `/writing-tests` and
`/setting-feature-flags-in-storybook` skills before touching the test and story files.
Notable decisions: (1) the pie-chart legacy fallback looked live, but `canRenderSqlPieGraph`
only checks `visualizationType === ActionsPie` — which the `PieChart` wrapper always sets — so
`LegacyPieChart` was unreachable and safe to delete; (2) renamed the dispatcher to `SqlChart`,
which also surfaced and fixed a stale `uniqueKey` prop still passed to `PieChart` from the SQL
editor's OutputPane.

On the rebase: both conflicts (`EndpointsUsageTrendsNode.tsx`, `AccountBillingChart.tsx`) were in
files whose only change on this branch was the mechanical rename. #70835 and #74136 have since
rewritten both so they no longer import the renamed symbols at all, so both resolved by taking
master's version — the billing-chart rename commit dropped as empty. `BillingChartProps` is
structurally assignable to `SqlChartProps`, so the adapter calls still typecheck. Two stale doc
references to `LineGraph` (the billing chart's comment and the Accounts agent guide) were updated
to match.

---


lricoy (approved):
lgtm on the web/customer analytics changes

PR #70676, opened 2026-07-14
feat(workflows): always link email metrics to the invocations tab

dmarchuk:
## Problem

In a workflow's email metrics tab, the Bounced, Blocked, and Bounce prevented tiles deep-linked into the old Logs tab. They should point at the newer Invocations tab so people stay in one place when investigating email delivery problems — and, importantly, clicking "Bounced" should show *only* bounced runs, not everything.

## Changes

The email metric tiles now drill into the **Invocations tab** instead of the old Logs tab, filtered to just the runs behind that metric.

To make that possible, the Invocations tab's **main search box now searches log-message text in addition to exact IDs** — the same substring search the old Logs tab had. A typed term matches an exact invocation / event / distinct / person id **or** a run that logged a message containing it (`buildSearchClause` → a per-flow `log_entries` subquery, `message ILIKE concat('%', <term>, '%')`). It's driven by the `inv_search` URL param and applied to both the table and the sparkline. There's no separate filter widget — it's the one search box, so a term the user types and a term a tile passes in behave identically.

The tiles pass their own term, narrowed to the level that distinguishes the outcome (carried in the optional `inv_log_levels` param):

| Metric | Drill-down |
|---|---|
| Bounced | message contains `bounce` at WARN/ERROR |
| Blocked | message contains `Complaint` at WARN/ERROR |
| Bounce prevented | message contains `Skipping send` at INFO |

The level narrowing is what keeps "Bounced" from also matching the INFO `Email sent to bounce@…` line. It applies **only** to a tile drill-down; typing in the box manually clears it, so a hand-typed search is level-agnostic (matches any level). `email_failed` stays undrillable — its two SES events emit differently-worded messages with no shared substring.

On performance: the subquery is bound to a single flow's `log_entries` (never a full-table scan) and is the same query shape the Logs tab already runs in prod. Like `problem_only`, it's deliberately not date-scoped, so a bounce/complaint that lands after a run's scheduled window still matches — the outer `scheduled_at` filter bounds which invocations appear.

**Automated:**

- `hogInvocationsLogic.test.ts` (new) — `buildSearchClause` returns an empty clause when no search is set (guards the silent "show everything" regression), matches an exact id **or** a `message ILIKE` substring, and adds the `lower(level) IN (...)` narrowing only when levels are set.
- `workflowMetricsSummaryLogic.test.ts` — each drillable metric maps to the expected `inv_search` / `inv_log_levels` params.
- Both suites pass (18 tests). Lint/format clean; frontend typecheck clean on the changed files.

**Manual, in the browser (local dev, real data):**

- Triggered a **live email bounce** end to end — a workflow sending to AWS's `bounce@simulator.amazonses.com` through a verified SES integration; the bounce notification came back and recorded a WARN/ERROR "Permanent bounce" log. Clicking the **Bounced** tile landed on the Invocations tab at `inv_search=bounce&inv_log_levels=WARN,ERROR`, the search box pre-filled with `bounce`, and the list narrowed to exactly that bounced run.
- Same for **Blocked** (sending to `complaint@simulator.amazonses.com` → "Complaint" log, `inv_search=Complaint`) and **Bounce prevented** (a no-MX recipient → INFO "Skipping send" log, `inv_search=Skipping send`).
- **Manual search:** clearing the box and typing a message-only term (`Permanent`) found the run via its log message — confirming the box searches both IDs and messages, level-agnostic, like the old Logs tab.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change. Internal navigation/filtering tweak with no user-facing docs surface.

## 🤖 Agent context

Claude Code (Opus 4.8) made these changes end to end; I directed it. Skills invoked: `/writing-tests` (before adding/changing tests).

How it evolved (visible in the commit history):
1. Moved the drill-down from the Logs tab to the Invocations tab, initially behind the runs-tab flag with a Logs fallback, then unconditionally linking to Invocations.
2. Added a dedicated `log_search` filter + a removable chip so a tile could isolate its specific log line.
3. Reworked that into the current design: folded the log-message search into the **main search box** (`buildSearchClause`) and dropped the separate chip, so the Invocations search works like the old Logs tab (type to search messages or IDs) and the tile drill-downs just pre-fill it. The per-metric terms (`bounce` / `Complaint` / `Skipping send`) match what the old Logs-tab drill-down used.

This touches the shared `HogInvocations` component (also used by the standalone hog-function scene). The new search behavior and the optional `log_levels` field are additive, so that scene keeps working — worth a glance from whoever owns that tab.


products/workflows/frontend/Workflows/WorkflowMetrics.tsx
```diff
@@ -105,12 +105,19 @@ function WorkflowRunMetrics(props: WorkflowLogicProps): JSX.Element {
...
-        const searchParams = buildEmailMetricLogSearchParams(metricKey, dateFrom.toISOString(), dateTo.toISOString())
-        if (searchParams && props.id) {
-            router.actions.push(urls.workflow(props.id, 'logs'), searchParams)
+        const searchParams = buildEmailMetricInvocationSearchParams(
+            metricKey,
+            dateFrom.toISOString(),
+            dateTo.toISOString()
+        )
+        if (searchParams) {
+            router.actions.push(urls.workflow(props.id, 'invocations'), searchParams)
```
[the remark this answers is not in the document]
dmarchuk: `HOG_INVOCATION_RESULTS_RUNS_TAB` is going to be removed soon and is already enabled everywhere, so this isn't a problem

frontend/src/scenes/hog-functions/invocations/hogInvocationsLogic.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        expect(buildSearchClause(props, { date_from: '-24h' }).raw).toBe('')
+    })
+
+    it('matches an exact id OR a log-message substring', () => {
+        const clause = buildSearchClause(props, { date_from: '-24h', search: 'bounce' }).raw
+        // Exact-id arms — pasting a UUID still finds that one run.
+        expect(clause).toContain("invocation_id = 'bounce'")
+        expect(clause).toContain("event_uuid = 'bounce'")
+        expect(clause).toContain("distinct_id = 'bounce'")
+        expect(clause).toContain("person_id = 'bounce'")
```
meikelmosby: just from reading this test this confused me somehow. We look for 'bounce' and the invocationId, uuid, distinctId and personId is set to bounce? 😅

frontend/src/scenes/hog-functions/invocations/hogInvocationsLogic.tsx
```diff
@@ -89,6 +89,13 @@ export interface HogInvocationsFilters {
      * no cross-shard subquery against `persons`.
      */
     person_uuid?: string
+    /**
+     * Optional level narrowing for the main search's log-message match, set only by metric
+     * drill-downs (e.g. the "Bounced" tile carries WARN/ERROR). It keeps the drill-down from also
+     * matching the INFO "Email sent to bounce@…" log. Absent for manual searches, which match any
+     * level.
+     */
```
meikelmosby: i think we can drop this comment..

meikelmosby (approved):

frontend/src/scenes/hog-functions/invocations/hogInvocationsLogic.test.ts
dmarchuk: Good catch, that read badly. The search box takes one typed term and tries it against every id column (paste an invocation/event/distinct/person id) OR the log message (type words) — so the same term lands in all arms. Switched the test to an id-shaped value (`run-42`) and added a comment explaining it in 64e8dac9.

---

PR #65807, opened 2026-06-24
feat(workflows): add nodejs and frontend support for native push workflow action

dmarchuk:
## Problem

We want to send push notifications (FCM / APNS) from the hogflow workers as a native workflow action.

This revives #51554, which was auto-closed after going stale. The dependency PRs it was originally stacked on have since merged into master, so this is now a standalone change rebased onto current master.

Related: https://github.com/PostHog/posthog/issues/44444

## Changes

- New `template-native-push` hog function template (FCM + APNS), gated behind the `workflows-push-notifications` feature flag and currently `status: 'hidden'`.
- **Multi-channel delivery:** the step takes an `integration_multi` "channels" input (select one or more FCM/APNS integrations). The CDP worker fans out to every selected channel in a single invocation and isolates failures per channel — a recipient only has a token for the platform they registered, so other channels record `push_skipped`.
- `PushNotificationService` in the nodejs CDP worker: resolves the per-device token for a distinct ID and delivers via FCM (`fcm.googleapis.com`) or APNS (HTTP/2, ES256 JWT auth). APNS provider tokens are cached in Redis (per Apple team + key id, 45-min TTL) so the fleet reuses one token per key rather than minting one per send — avoiding Apple's `429 TooManyProviderTokenUpdates`.
- **Registration is platform-agnostic:** `/api/push_subscriptions` resolves the integration from the `app_id` (a Firebase `project_id` or an APNs `bundle_id`), not the device platform — so an iOS device can deliver via Firebase (FCM) or via APNs directly.
- **Dead-token pruning:** when a provider reports a token as no longer registered (FCM `UNREGISTERED`, APNs `410`), it's removed from the person and the channel is skipped rather than retried. Ambiguous errors (`INVALID_ARGUMENT`, `BadDeviceToken`) are left alone to avoid dropping a token that's actually valid.
- `sendPushNotification` async function wired into the hog executor queue handling.
- APNS integration kind end to end: serializer/model support, frontend integration metadata, and FCM/APNS channel setup modals in the Workflows scene.
- Push subscription token storage/lookup (`$device_push_subscription_<appIdentifier>` person property, encrypted at rest).
- Flat config form (like email): core message fields inline, Android/iOS platform overrides in collapsed sections; step inputs prefill their template defaults on add.

No frontend behavior changes are user-visible without the feature flag.

https://github.com/user-attachments/assets/2ff4ece5-d9c3-4440-b634-804faaa4e709

## How did you test this code?

[...]

PR #70678, opened 2026-07-14
feat(workflows): verify caller identity on push registration

dmarchuk:
## Problem

Follow-up to #65807, closing the one security finding we left open there: **"High: untrusted callers can rebind push subscriptions."**

A push device token (FCM registration token / APNs device token) is a delivery **address**, not a credential. FCM/APNs hand a token to any app instance, and an attacker owns their own token legitimately, so possession proves "deliver to this device" and never "this device belongs to user X". Today `POST /api/push_subscriptions` authenticates with the **public project token**, which is embedded in the app and world-readable. It identifies the project, not the person. So anyone with it can register their own device under a victim's `distinct_id` and receive that user's workflow push notifications (notification takeover).

This is a hard blocker before push goes GA. It doesn't affect the merged PR while push stays gated (no delivery yet), but it has to close before we turn delivery on.

> [!NOTE]
> This is a draft of what the fix should ideally look like. The backend half is here and testable. The mobile SDK half (attaching and refreshing the token) is a separate piece owned by the SDK team, and the SDK is still in development, so we can shape its contract now.

## Changes

I researched how the leading messaging platforms solve this. **Braze's "SDK Authentication"** is the reference design and the only one of the three I looked at that actually defends against the attack: the customer's backend (the only party that authenticated the end user) mints a short-lived signed token whose subject is the user id, and the platform verifies it. **Customer.io** and **Brevo/WonderPush** both "trust the client" - the client asserts the identity with no signature - which is exactly the hole we have now.

This PR adapts that pattern to PostHog's primitives:

- **Signed identity token** (`push_identity_tokens.py`). A short-lived JWT (HS256) signed with the project's **secret API key** (`Team.secret_api_token`), minted by the customer's backend, asserting `sub = distinct_id` and `app_id`. PostHog re-verifies it at registration. An attacker with only the public token can't forge it. Verification accepts the current **or** backup secret, so a key rotation doesn't reject in-flight tokens.
- **Graduated per-integration enforcement** on the endpoint (`push_identity_verification` in the Firebase/APNs integration `config`), mirroring Braze's staged rollout so it can be turned on without breaking existing traffic:

| Mode | Behavior |
| --- | --- |
| `disabled` (default) | token ignored - current behavior, nothing breaks |
| `optional` | token verified and outcome recorded (`push_subscription_identity_verification` metric), but registration still succeeds - lets a customer confirm their backend mints valid tokens before enforcing |
| `required` | a registration without a valid token is rejected (`401`) |

- A short integration doc (`push_identity_tokens.md`) describing the token, claims, flow, and rollout for the SDK/backend teams.

I chose symmetric HMAC over an asymmetric key pair (Braze uses RS256) because PostHog verifies its own ingestion - there's no third-party verifier that needs a public key - and the secret already ships with rotation. I also considered gating the endpoint behind a Project Secret API Key (PSAK), but PSAK auth is wired for team-scoped DRF viewsets, and this is a capture-style function view on `/api/push_subscriptions`. Signing keeps the SDK as the direct caller (good DX) while still requiring server-side proof.

**Before** - any holder of the public token can bind a victim's `distinct_id`:

```mermaid
flowchart LR
    App[App: public project token]:::phRed -->|"distinct_id + device_token"| EP["/api/push_subscriptions"]:::phYellow
    EP --> Store[(store token on person)]:::phGray
```

**After** (`required` mode) - only a backend-signed assertion can bind a `distinct_id`:

```mermaid
flowchart LR
    BE[Customer backend: authenticated user + secret key]:::phBlue -->|"mint signed identity_token"| App[App: public token]:::phRed
    App -->|"distinct_id + device_token + identity_token"| EP["/api/push_subscriptions"]:::phYellow
    EP -->|"verify sig, exp, sub==distinct_id"| Store[(store token on person)]:::phGray
    EP -->|"missing / invalid / wrong sub"| Reject[401]:::phRed
```

I (Claude) could not run the suite in this environment - no Django install available - so these tests have **not** run locally yet and I'm relying on CI to execute them. They are:

- `test_push_identity_tokens.py` (pure `SimpleTestCase`, no DB): the signature/claim matrix on `verify_push_identity_token` - valid current-secret and backup-secret tokens pass; wrong `sub`, wrong `app_id`, wrong secret, expired, malformed, and no-secret-configured all fail. Catches any regression that would let a token authorize a `(distinct_id, app_id)` it wasn't minted for (the core rebind guard) or wrongly reject a valid one.
- `test_push_subscriptions.py` (endpoint wiring guards): `required` mode accepts a valid token (200) and rejects both a missing token and **a token minted for a different `distinct_id`** (401, capture never called - the actual attack); `optional` mode stores even without a token (monitor mode must not block). These prove the mode wiring end to end, which the pure tests can't.

Invoked `/writing-tests` before writing them and pushed the claim matrix down to the pure `SimpleTestCase` per its guidance, keeping the endpoint tests as thin wiring guards.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Included an in-repo integration doc (`push_identity_tokens.md`). No user-facing posthog.com docs yet - the customer-facing setup doc should land with the SDK support and the Channels UI toggle.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) - directed by @dmarchuk, who should be the DRI/assignee (the create-PR tool couldn't set the assignee, so please assign yourself).

Authored by Claude ([session](https://claude.ai/code/session_01Bbav6vE7Tm24KkdRhZmvNM)). Skills invoked: `/adding-project-secret-api-key-auth` (to understand PostHog's secret-credential primitives), `/writing-tests`, and I followed the `/improving-drf-endpoints` rule for the endpoint. A research subagent gathered how Customer.io, Braze, and Brevo handle device-token-to-user binding (official docs), which is what pointed to the Braze signed-assertion pattern as the one that actually defends against this.

Decisions along the way: considered PSAK-gating the endpoint (rejected - it's a function-view capture route, not a team-scoped viewset) and an asymmetric RS256 JWT like Braze (rejected - no third-party verifier here, and HMAC off the existing secret reuses its rotation). Landed on HMAC-signed short-lived tokens with a per-integration `disabled -> optional -> required` rollout so it's safe to enable incrementally.

Known follow-ups, called out in the doc: SDK support for attaching/refreshing the token, a Channels UI control for the per-integration mode (today it's a `config` value), and clearing the binding on logout.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bbav6vE7Tm24KkdRhZmvNM)_

### DIFF
--- a/products/messaging/backend/api/push_identity_tokens.md
+++ b/products/messaging/backend/api/push_identity_tokens.md
@@ -1,0 +1,60 @@
+# Push subscription identity verification
+
+Securing the binding of a mobile push device token to a `distinct_id`, so an attacker who only holds
+the (public, app-embedded) project token cannot point another user's notifications at their own device.
+
+## Why this is needed
+
+A push device token (FCM registration token / APNs device token) is a delivery **address**, not a
+credential — FCM/APNs issue one to any app instance, and an attacker owns their own token legitimately.
+Possession proves "deliver here", never "this device belongs to user X". Today `POST /api/push_subscriptions`
+authenticates with the **public project token**, which identifies the project but not the person, so any
+holder of it can register a device under an arbitrary `distinct_id` (notification takeover / "rebind").
+
+This mirrors what leading platforms do: the fix must require a server-attested proof that the caller may
+act for that `distinct_id`. Braze's "SDK Authentication" is the reference design — the customer's backend
+mints a short-lived signed JWT whose subject is the user id, and the platform verifies it. (Customer.io and
+Brevo/WonderPush "trust the client" and do **not** defend against this — the state we're leaving behind.)
+
+## The token
+
+A short-lived **JWT (HS256)** signed with the project's **secret API key** (`Team.secret_api_token`).
+Minted by the **customer's backend** (the only party that authenticated the end user), never in the app.
+
+| Claim    | Value                                                    |
+| -------- | -------------------------------------------------------- |
+| `sub`    | the `distinct_id` being registered                       |
+| `app_id` | the FCM `project_id` / APNs `bundle_id` the token is for |
+| `aud`    | `posthog:push_identity`                                  |
+| `exp`    | short expiry (SDK refreshes on failure)                  |
+
+Verification accepts the current **or** backup secret, so a key rotation doesn't reject in-flight tokens.
+`sub` and `app_id` are both bound, so a token minted for one identity/app can't authorize another.
+
+Reference signer + verifier: `push_identity_tokens.py` (`sign_push_identity_token` shows exactly what the
+customer backend produces; PostHog's ingestion only ever calls `verify_push_identity_token`).
+
+## Flow
+
+1. Customer backend authenticates the user, then mints the token for `(distinct_id, app_id)` with the
+   project secret key.
+2. The app receives the token and calls `POST /api/push_subscriptions` with the usual fields plus
+   `identity_token`.
+3. The endpoint verifies the signature, `exp`, and that `sub`/`app_id` match the registration.
+
+## Rollout — per integration, three stages
+
+Set `push_identity_verification` in the Firebase/APNs integration's `config` (mirrors Braze's staged
+rollout so it can be turned on without breaking existing traffic):
+
+- **`disabled`** (default) — `identity_token` ignored; current behavior.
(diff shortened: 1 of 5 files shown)


dmarchuk:
Closing for now as per https://github.com/PostHog/posthog/pull/65807#discussion_r3580592044

PR #70683, opened 2026-07-14
feat(tasks): add team and user default models for task agent runs

VojtechBartos:
## Problem

There was no way to set a preferred model once and have it apply across all your tasks. Each run took whatever the launching surface defaulted to, so a team couldn't standardize the project on one model, and a person couldn't set their own preference and have every run they start — from the web composer, Slack, or desktop — just use it.

## Changes

- Adds project- and user-level default models for task agent runs (`TeamTasksConfig` / `UserTasksConfig`), resolved user-over-project.
- Settings UI to set both, sourcing its model list from the same catalogue the composer uses — so a model settable as a default is exactly one a run can use.
- The composer shows the resolved default with a `Default ·` prefix, and submitting without an explicit pick omits the model entirely so the server resolves it (which also lets a warm run match).
- Slack runs sit on the same chain rather than pinning their own model. A model named in the mention itself ("use fable for this one") wins, then the user's Slack-level preference from App Home; when neither applies, the run defers to the central user/team default above, and Slack's built-in floor is used only when no default is configured anywhere. The App Home card names whichever rung actually applies and links to the PostHog settings that own the central levels.

How a run's model resolves at creation time (`resolve_ai_run_selection`):

```text
run carries an explicit pin? ──yes──► use the pin as-is        source: explicit
        │ no
        ▼
user default (tasks/@me/config)
  usable triple? ─────────────yes──► use it wholesale          source: user
        │ no / unusable
        ▼
team default (tasks/config)
  usable triple? ─────────────yes──► use it wholesale          source: team
        │ no
        ▼
no default — the surface keeps its built-in model              source: none
```

A user triple replaces the team triple wholesale (no field blending), an effort the model doesn't support is dropped rather than sent, and environment teams read and write the project root's config row.

https://github.com/user-attachments/assets/c2ab4194-32f8-459a-8aa8-0b40d1132805

Screenshot 2026-08-18 at 14 35 25

Tested locally, both manually in the running app and via the automated suites: 855 frontend tests across 66 suites, and the run-defaults backend suite (resolution, run creation, warm-run matching, and the config API — auth, cross-organization access, scopes, validation, environment-team normalization). mypy and typecheck clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None.

## 🤖 Agent context

Claude Code (Opus 5) rebased this onto the catalogue stack and adapted it; the feature itself is the author's.

Worth knowing from the rebase: master independently dropped the Slack workspace-level AI preference while this sat open, so that commit's code is now redundant and master's version was taken — only the migration clearing stale rows remains. Master also consolidated `validate_ai_preferences` onto `validate_model_selection`, which is kept over this branch's equivalent.

Two things a reviewer should check, flagged by a cleanup pass and left alone as design calls: a Codex default may not apply (the composer treats it as "no default"), and the defaults logic doesn't refetch on project switch.


VojtechBartos:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/PostHog/posthog/70683?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#70683** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/70683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#70671** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70671?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

andrewm4894 (approved):

andrewm4894 (commented):
.

products/tasks/backend/models.py
```diff
@@ -502,12 +503,47 @@ def _assign_task_number(self) -> None:
...
+        from products.tasks.backend.logic.services.ai_run_defaults import (  # noqa: PLC0415 — breaks the circular import with ai_run_defaults, which imports this module
+            apply_ai_run_defaults,
+        )
+        from products.tasks.backend.temporal.process_task.utils import (  # noqa: PLC0415 — keeps temporalio off the import path (matches _build_task)
+            apply_runtime_adapter_run_state,
+        )
+
+        resolved = apply_ai_run_defaults(state, self.team_id, acting_user_id or self.created_by_id)
+        if resolved is None:
+            return
```
tatoalo, line 678: with this PR, we deleted the serializer guard that used to cover what we had as 
```
"This field requires runtime_adapter to be set."
```

so this might mean that withaa `codex` default, we would have a `POST /tasks/{id}/run/ {"initial_permission_mode": "acceptEdits"}` which persists a Claude-only mode against a Codex run 🤔 

using the effective adapter rather than only the resolved one would survive the early return?

products/tasks/backend/models.py
```diff
@@ -502,12 +503,47 @@ def _assign_task_number(self) -> None:
...
+        default preferences when the caller pinned none.
+
+        A partially pinned selection (either `runtime_adapter` or `model` present) is
+        treated as explicit and left untouched — overwriting half a pin would replace a
+        value the caller chose. Internal tasks (custom-prompt infra agents) keep their
+        pinned behavior and never inherit preferences. An explicitly set
+        `reasoning_effort` survives injection; the default triple's effort only fills a
+        gap.
+        """
+        if self.internal:
```
tatoalo: need to check for `self.runtime == Task.Runtime.PI` as well

posthog/temporal/ai/slack_app/activities/task_creation.py
```diff
@@ -1130,7 +1132,7 @@ def _resume_task_with_new_run(
     # reaches the sandbox with none and the agent server picks its own — the thread then
     # can't say what ran. Resolved again rather than carried over, like the keys above: a
     # follow-up honours whatever the picker says now.
-    run_prefs = resolve_run_preferences(integration, slack_user_id)
+    run_prefs = resolve_run_preferences(integration, slack_user_id, team_id=mapping.task.team_id, user_id=run_actor.id)
```
tatoalo: we are omitting `acting_user_id`, so we fall back to `created_by_id` 🤔

products/tasks/backend/logic/services/ai_run_defaults.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+    canonical_team_id = _canonical_team_id(team_id)
+
+    # Entitlement gating is lazy: the distinct_id lookup only happens when a level
+    # actually resolves to a flag-gated model, which keeps the per-run-creation
+    # hot path free of an extra query in the common case.
+    acting_distinct_id: str | None = None
+    acting_distinct_id_loaded = False
+
+    def _model_accessible(model: str | None) -> bool:
```
tatoalo, line 148: this puts a network call inside the warm-run row lock

tatoalo (approved):

products/tasks/backend/logic/services/ai_run_defaults.py
VojtechBartos: as we dicussed i am going to follow up and we are going to have single definition baked in as fallback

PR #70693, opened 2026-07-14
chore(ci): narrow and speed up product test sweeps

pauldambra:
## Problem

Several expensive product test buckets join broad backend sweeps when unrelated product code changes. Master product shards also rebuild the Django test schema from scratch, which adds substantial setup time across each matrix entry.

**Why:** Reduce unnecessary CI runs and shorten product test feedback without skipping changes to declared Python dependencies.

## Changes

- Scope the highest-cost product suites to their direct cross-product Python dependencies during product-only sweeps.
- Distinguish directly changed products from dependent tests affected by those changes, preserving product isolation behavior.
- Keep full product coverage for legacy backend, shared infrastructure, and CI workflow changes.
- Restore the previous master schema cache for product shards and apply migration deltas through pytest.

- `node --test .github/scripts/turbo-discover-staleness.test.js` verifies direct product-path detection and dependency-scoped sweep selection.
- `npx --yes turbo@2.9.14 run backend:test --dry-run=json` verifies Turbo resolves every scoped product's backend, shared, and cross-product inputs.
- `npx --yes oxfmt@latest --check products/*/turbo.json` for the changed product configs.
- `actionlint .github/workflows/ci-backend.yml`
- `git diff --check`

`bin/hogli lint:workflows` could not run because the task image has no Python executable; standalone actionlint passed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior or documented workflow changes.

## 🤖 Agent context

Implemented with PostHog Code using the `/authoring-ci-workflows` and `/writing-tests` skills. CI job data showed expensive product buckets were frequently pulled into sweeps by unrelated non-isolated product changes, while master shards repeatedly paid schema setup costs. The dependency maps are checked against current cross-product Python imports, and true legacy changes still run the complete product matrix.

---

### DIFF
--- a/.depot/workflows/ci-backend.yml
+++ b/.depot/workflows/ci-backend.yml
@@ -441,34 +441,39 @@ jobs:
               env:
                   BASE_REF: ${{ github.event.pull_request.base.ref }}
               run: git fetch --no-tags --depth=1000 --filter=blob:none origin "$BASE_REF:refs/remotes/origin/$BASE_REF"
-            - name: Compute schema cache key from merge-base
+            - name: Compute schema cache key
               id: schema-key
-              if: github.event_name == 'pull_request'
               env:
                   BASE_REF: ${{ github.event.pull_request.base.ref }}
               run: |
-                  # HEAD is the synthetic merge commit; HEAD^2 is the PR branch tip.
-                  # The fetch-depth:1000 checkout + base-ref fetch above ensure the
-                  # full ancestry needed to find the divergence point is available.
-                  MERGE_BASE=$(git merge-base HEAD^2 "origin/${BASE_REF}" 2>/dev/null || echo "")
-                  if [ -z "$MERGE_BASE" ]; then
+                  if [ "${{ github.event_name }}" = "pull_request" ]; then
+                      # HEAD is the synthetic merge commit; HEAD^2 is the PR branch tip.
+                      SCHEMA_BASE=$(git merge-base HEAD^2 "origin/${BASE_REF}" 2>/dev/null || echo "")
+                      MIGRATION_REF="$SCHEMA_BASE"
+                  else
+                      # Product jobs run in parallel with the migration job on master,
+                      # so restore the previous push's schema and let pytest apply any delta.
+                      SCHEMA_BASE=$(git rev-parse HEAD^ 2>/dev/null || echo "")
+                      MIGRATION_REF=HEAD
+                  fi
+                  if [ -z "$SCHEMA_BASE" ]; then
                       echo "key=" >> $GITHUB_OUTPUT
                       echo "migrations_key=" >> $GITHUB_OUTPUT
-                      echo "::notice::merge-base not found (branch too stale?) — schema cache will be skipped"
+                      echo "::notice::schema base not found — schema cache will be skipped"
                       exit 0
                   fi
-                  # Exact branch-point key, kept as the fallback restore key below.
-                  echo "key=posthog-schema-master-${MERGE_BASE}" >> $GITHUB_OUTPUT
-                  # Shared key: hash merge-base schema inputs, not PR head, so PR migrations still
-                  # apply as the pytest delta. uv.lock covers third-party migration changes; the
-                  # postgres image is folded in so a server-version bump invalidates the key.
+                  # Exact base key, kept as the fallback restore key below.
+                  echo "key=posthog-schema-master-${SCHEMA_BASE}" >> $GITHUB_OUTPUT
+                  # Shared key: hash the schema inputs. On PRs this is the merge base;
+                  # on master it is HEAD, with the previous SHA key as a safe fallback
+                  # when the current push introduces a migration.
                   # Non-Django migration frameworks are excluded from this Postgres schema dump.
                   # SCHEMA_CACHE_EPOCH (workflow env) prefixes the key as a manual cache-bust knob.
-                  MIG_FILES=$(git ls-tree -r --format='%(objectname) %(path)' "$MERGE_BASE" \
+                  MIG_FILES=$(git ls-tree -r --format='%(objectname) %(path)' "$MIGRATION_REF" \
                       | grep -E '/migrations/[^/]+\.py$|^[0-9a-f]+ uv\.lock$' \
                       | grep -vE '/(clickhouse|async_migrations)/migrations/' \
                       | LC_ALL=C sort) || true
-                  PG_IMAGE=$(git show "${MERGE_BASE}:docker-compose.base.yml" 2>/dev/null \
+                  PG_IMAGE=$(git show "${MIGRATION_REF}:docker-compose.base.yml" 2>/dev/null \
                       | grep -m1 -E '^[[:space:]]*image:[[:space:]]*postgres:' | tr -d '[:space:]')
                   if [ -z "$MIG_FILES" ]; then
                       echo "migrations_key=" >> $GITHUB_OUTPUT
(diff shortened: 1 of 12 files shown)


webjunkie (changes requested):
Requesting changes and closing this PR. I'm DRI for test selection in `turbo-discover.js`, and there is approved work in flight that addresses the same problem from the opposite direction (#70702, plus #70679 for the coupling that motivated it). The cost problem this PR targets is real, the effort is appreciated, and one piece of it is worth landing separately (see "Worth salvaging" at the end). But the core mechanism narrows test selection in a way that cannot be made sound, and I want to leave the full reasoning here so it's actionable rather than just a verdict. This review is written primarily for the agent that will process it, so it is explicit and evidence-heavy on purpose.

**What this PR does, decomposed**

Three independent changes travel together here:

1. Nine products get hand-written cross-product dependency globs as `backend:test` inputs in their `turbo.json` (`../../products/<dep>/**/*.py`).
2. `turbo-discover.js` gains `DEPENDENCY_SCOPED_PRODUCTS`: those nine products are dropped from full product sweeps unless Turbo marks them affected (`includeDependencyScopedProducts`). Additionally, the "non-isolated product changed → test everything" trigger is swapped from Turbo affectedness (`nonIsolatedAffectedProducts`) to raw changed file paths (`getDirectlyChangedProducts`).
3. The Postgres schema-dump cache restore, currently PR-only, is extended to master pushes.

Change 3 is good and fully separable. Changes 1 and 2 are the substance of the PR and are what I'm requesting changes on — and since they can't be fixed by iteration (argument below), I'm closing rather than leaving it open.

**The dependency globs are already incomplete, in exactly the way the PR's own comment predicts**

The in-code comment says: "Keep this list deliberately small: missing an input would skip a product test that should have run." Diffing the shipped globs against actual cross-product imports in the repo finds three missing edges. All three are function-level deferred imports — this repo deliberately defers cross-product imports to keep products off the startup path (the sanctioned `PLC0415` exception), so any top-of-file import scan systematically misses this entire class:

- `endpoints → signals`: `products/endpoints/backend/logic/execution.py:161` and `products/endpoints/backend/logic/strategies.py:260` both do `from products.signals.backend.facade.api import emit_signal`. The `products/endpoints/turbo.json` in this PR does not list `products/signals`.
- `tasks → posthog_ai`: `products/tasks/backend/logic/services/local_skills.py:154` does `from products.posthog_ai.scripts.build_skills import SkillBuilder`. The `products/tasks/turbo.json` in this PR does not list `products/posthog_ai`.
- `warehouse_sources → notifications`: `products/warehouse_sources/backend/temporal/data_imports/cdc/broken.py:99` imports from `products.notifications.backend.facade.api`. The `products/warehouse_sources/turbo.json` in this PR does not list `products/notifications`.

Concrete regression the first one enables: a PR changes `products/signals/backend/facade/api.py` in a way that breaks `emit_signal` callers. `signals` is non-isolated (its `turbo.json` added here declares only `backend:test`, no `backend:contract-check`), so the "test everything" branch fires — but under this PR, `endpoints` is dependency-scoped, its globs don't include `signals`, Turbo doesn't mark it affected, and `includeDependencyScopedProducts` drops it from the sweep. The breakage in `execution.py` ships to master untested. Under current master behavior, that branch runs all products including `endpoints`, and the break is caught pre-merge. This is the same failure class as the recent master break that motivated #70679/#70702 — a dependent product's tests not running when something it consumes changed — reintroduced through a new channel. The other two holes enable the analogous scenario for `tasks` and `warehouse-sources`.

**Why the list can't be kept correct, even with diligence**

These globs are a third, independently maintained copy of the cross-product dependency graph (alongside `tach.toml` and the contract-check inputs), and it's the only copy with no enforcement: `tach check` verifies import edges in CI, but nothing verifies these globs against reality. The moment anyone adds a new deferred import, the glob list is silently stale, and the failure mode is silently skipped tests — invisible until master breaks. A list that must be complete to be safe, has no completeness check, and shipped with three known-verifiable holes on day one is not a maintenance burden; it's a standing incident generator.

**Why even a complete import scan wouldn't be sufficient**

This is the deeper reason iteration can't fix the approach, and it's documented: `products/architecture.md`, section "What makes the skip sound". Skipping a product's tests is a claim that no change elsewhere can break them. Import analysis proves only part of that claim. Channels that carry real coupling with zero imports, all present in this repo today:

- Cross-cutting tests reach product endpoints by URL: the Django test client dispatches in-process into the view stack, so permissions/schema/activity-log style tests couple to a product's live behavior with no import edge.
- `apps.get_model("<app>", "<Model>")` string references: 113 sites across the repo. Example: `products/reminders/backend/constants.py` references the `error_tracking` models with no import of `products.error_tracking` anywhere in `reminders`.
- `importlib.import_module(...)` dynamic imports: 22 sites.

The doc's conclusion, verbatim: because this channel can't be enumerated, it is "closed by construction rather than inspection." This PR narrows by inspection — a hand-audited enumeration of who talks to whom. Six of the nine dependency-scoped products (`signals`, `logs`, `product-analytics`, `revenue-analytics`, `web-analytics`, `experiments`) are non-isolated, meaning they have not sealed the channels above, so no amount of glob completeness makes skipping them sound. The other three (`endpoints`, `tasks`, `warehouse-sources`) are isolated and already exit sweeps soundly when only their own internals change — what this PR adds for them is being dropped from other products' blast radius, and notably all three verified holes above are exactly consumer-side edges into these three.

**The fail-direction asymmetry**

The approved cascade work (#70702) only ever adds products to the matrix when a contract changes; if its dependency graph is missing or wrong, it widens to all products. An error costs CI minutes. This PR only ever removes products from the matrix; if its dependency list is wrong (it is, three times), the error silently skips tests and costs a master break. Same class of mechanism, opposite failure direction. Related detail: `getDirectlyChangedProducts` returns `[]` when `TURBO_SCM_BASE` is unset, which silently disables the "test everything" branch entirely — where the current Turbo-affectedness path falls back to Turbo's default base. That's a second independent silent-skip mode.

**Mechanical note**

This also collides textually with #70702 (approved), which rewrites the same regions of `turbo-discover.js` — the selection branch, `module.exports`, and the main block — so this branch would need a substantial rebase regardless.

**Worth salvaging: the master schema-cache restore**

The change extending the schema dump restore to master pushes (`.github/workflows/ci-backend.yml` + the `.depot` mirror, plus the key computation rename) is self-contained, orthogonal to test selection, and a real setup-time win for master shards — restore the previous push's schema, let pytest apply the migration delta. I'd welcome that as its own small PR; it needs none of the selection machinery here.

**For the agent processing this review — actionable summary**

1. Do not re-open this PR with expanded globs. Completeness is not achievable by scanning: deferred function-level imports, `apps.get_model` string references, `importlib`, and URL-coupled in-process tests are all invisible to import analysis, and the repo's isolation model (`products/architecture.md`, "What makes the skip sound") requires skips to be earned by construction, not enumeration.
2. The sanctioned route to the CI savings this PR wants is isolating the expensive products: seal the facade and presentation layer per `products/architecture.md`, then declare `backend:contract-check`. Each product isolated that way exits full sweeps soundly and permanently, and the approved dependent-cascade (#70702) keeps its consumers tested when its contract changes. `signals`, `logs`, `product-analytics`, `revenue-analytics`, `web-analytics`, and `experiments` are the high-value targets on this PR's own list.
3. If desired, resubmit the master schema-cache restore as a standalone PR — that part is good.

Thanks for digging into CI cost — the target is right, and the isolation route gets there without the skip risk.

webjunkie:
Closing per the review above: the dependency-glob narrowing can't be made sound (three verified holes, fail-dangerous direction), and it collides with the approved #70702. The master schema-cache restore piece is welcome as a standalone PR.

PR #70694, opened 2026-07-14
feat(ingestion): add ingestion control plane service

jose-sequeira:
## Problem

When ingestion partitions lag because of problematic messages, figuring out *what* is actually in the partition (which team, which distinct_id, what event types, how big) is a manual kcat-and-squint exercise. Separately, the ingestion-consumer is growing a routing debug API, but there was no way to view it across live pods.

## Changes

A new standalone Rust service, `rust/ingestion-control-plane/`: a purely internal web tool for the ingestion team, built as a single binary with an embedded UI and a sidebar of tools designed to grow over time. Two tools to start:

**Lagging partitions**

- `GET /api/lag/overview` scans every configured consumer group (committed offsets vs watermarks via a lightweight `BaseConsumer`, same pattern as personhog-leader's warming) and sorts by total outstanding messages, with per-partition drill-down.
- Analysis jobs read a slice of a partition and aggregate message headers only (payloads are dropped, only their size is recorded). Two modes: `head` starts at the group's committed offset (what the consumer is stuck on), `tail` samples the newest messages (what's arriving now).
- Results are broken down per token, resolved to `team_id` via a read replica, with events and distinct_ids nested under each token, plus a size distribution and header-flag counts (`historical_migration` etc.).
- Analyses run under a dedicated consumer group and never touch the real group's offsets. Memory and runtime are bounded by cardinality caps, a byte budget, and a deadline, with partial results marked `truncated_reason`.

icp-ui-lag

**Consumer debug**

- Discovers ingestion-consumer pods via the Kubernetes API (label selectors, `pods get/list`), with a static discovery mode for local testing, mirroring the consumer's own worker discovery modes.
- The consumer's routing debug UI now lives here, served per pod at `/pods/:name/` on top of a proxy that pipes the pod's debug API (JSON + SSE) through unchanged. The consumer itself only needs to expose the debug API, no HTML and no CORS handling.

Screenshot 2026-07-14 at 14 06 13

**Supporting changes**

- `common/k8s-awareness` gains a label-selector pod lister (`list_pods_by_selector`, `get_pod`, `DiscoveredPod`) so pod discovery stays in the shared module.
- `examples/seed_lag.rs` seeds a realistic scenario into local Kafka: one dominant team (70%, single hot distinct_id, large payloads) on a lagging partition, mixed healthy traffic elsewhere.
- Build matrix entry in `.github/rust-images.yml` (dedicated Depot project). Deploy matrix + charts app (RBAC, internal ingress, network policy) are follow-ups in the charts repo.

> [!NOTE]
> The `/debug/*` endpoints on the ingestion-consumer are a contract for now, they land in a separate branch. The proxy and embedded UI were verified against a locally running consumer that serves them.

> [!WARNING]
> This service is purely internal, it has no auth of its own and results contain cross-customer tokens/distinct_ids. It must only ever be deployed behind the internal ingress. The README states this prominently.

- 35 unit tests across the crate and `k8s-awareness`: lag math edge cases (null/clamped committed offsets), offset-range resolution for both analysis modes, pure header aggregation (per-token nesting, caps, flags, histogram), static pod parsing/resolution, selector matching, and the `Pod` → `DiscoveredPod` mapping.
- End to end against local Redpanda + the dev Postgres: seeded the demo scenario, verified the overview (1 lagging partition, 1,950 lag), ran head/tail analyses (dominant team resolved to a real `team_id`, correct truncation reasons), and cancelled/replaced jobs.
- Verified the debug proxy against a locally running ingestion-consumer serving the debug API: pod listing, proxied `/debug/state` and `/debug/load` JSON, and live SSE streaming through the proxy; unknown pod names are rejected (the proxy only dials pods resolved fresh from discovery, never client-supplied hosts).
- UI checked with headless Chromium (screenshots above), no console errors.
- What I could not test: kubernetes discovery mode against a real cluster (covered by unit-tested mapping + the static mode), and KIP-848 consumer groups on MSK (called out in the README follow-ups).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal service, documented in the crate README.

## 🤖 Agent context

Built with Claude Code, directed throughout by @jose-sequeira. Notable decisions: reused established workspace patterns (personhog-leader's `committed_offsets`/`fetch_watermarks` on `spawn_blocking`, kafka-assigner's `BaseConsumer` metadata client and kube pins, capture-logs' service bootstrap); analyses are background jobs polled by the UI rather than long requests to avoid proxy timeouts; the UI is embedded vanilla HTML/JS (`include_str!`) to keep the generic rust Dockerfile build-step-free, revisiting React if sections multiply. Mid-review course corrections by José: an all-groups lag overview instead of a group dropdown, one analysis at a time instead of a stacked list, per-token nesting of events/distinct_ids, a static pod-discovery mode for local testing, and moving the consumer's debug UI into this service entirely.


rust/ingestion-control-plane/src/api.rs
```diff
@@ -0,0 +1,169 @@
...
+async fn list_pods(State(state): State<AppState>) -> Result<Json<PodsResponse>, ApiError> {
+    let pods = state
+        .pods
+        .list_pods(&state.config)
+        .await
+        .map_err(|e| ApiError::unavailable(format!("pod discovery unavailable: {e:#}")))?;
+    Ok(Json(PodsResponse { pods }))
+}
+
+pub fn router(state: AppState) -> Router {
```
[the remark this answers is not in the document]
jose-sequeira: Authentication will be via cognito

rust/ingestion-control-plane/examples/seed_lag.rs
```diff
@@ -0,0 +1,222 @@
...
+//! offset is left near the start so the partition shows heavy lag. Partitions
+//! 1 and 2 get light, balanced traffic and are committed at their high
+//! watermark (healthy).
+//!
+//! Run against the dev stack:
+//!
+//! ```sh
+//! DATABASE_URL=postgres://posthog:posthog@db:5432/posthog \
+//!     cargo run -p ingestion-control-plane --example seed_lag
+//! ```
```
eli-r-ph, line 15: 🤔 only partway through the review, maybe you did this, but occurs to me maybe this should just run in local dev mprocs (hogli) rig all the time? 🤷 

_EDIT:_ I mean the service generally, not `seed_lag` 😆

rust/ingestion-control-plane/src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[envconfig(
+        from = "CONSUMER_TARGETS",
+        default = "events-ingestion-consumer=events_plugin_ingestion"
+    )]
+    pub consumer_targets: String,
+
+    /// Read-replica Postgres URL for token -> team_id resolution. When empty,
+    /// resolution is disabled and analyses report `team_id: null`.
+    #[envconfig(from = "DATABASE_URL", default = "")]
+    pub database_url: String,
```
eli-r-ph: same here (robot thing maybe?) not important, but lots of unneeded `from = ...` and empty-string defaults vs. `Option`s 🤷 feel free to ignore, nonblocking

eli-r-ph (approved):

rust/ingestion-control-plane/examples/seed_lag.rs
jose-sequeira: I was thorn between committing this or not, but found it could be useful for local testing. I didn't go super far in the local setup, want to see how it looks in prod

rust/ingestion-control-plane/src/config.rs
jose-sequeira: Ye, this could be Option to be fair 🤔 Let me review

PR #70697, opened 2026-07-14
fix(batch-exports): don't re-emit SLA breach warning on workflow replay

rossgray:
## Problem

A batch export run that was cancelled still fired "SLA breached" warnings roughly half an hour after it was done. That warning backs an alert, so it paged for a run that had already completed.

The only producer of that log line is `SLAWaiter.wait_for_sla` in `products/batch_exports/backend/temporal/metrics.py`. It runs in workflow context but had no replay guard. Temporal re-executes a workflow from the start on every replay (worker recovery, cache eviction, and crucially **queries** - the Temporal UI stack-trace tab and programmatic queries replay a finished run). On replay the SLA timer re-resolves instantly from history and re-logs the warning stamped with the replay's wall-clock time. Investigating the run in the UI can itself trigger this, which is why two duplicate warnings landed after completion.

## Changes

Guard the log emission with `workflow.unsafe.is_replaying()`, matching the existing pattern in `posthog/temporal/common/slo_interceptor.py` (which guards its emissions the same way, with the same rationale).

Two subtleties captured in a code comment:

- The guard sits **after** the timer resolves, not around task creation. The `asyncio.sleep` is a deterministic Temporal timer, so the waiter task must be created on every replay or history diverges (non-determinism error). Only the side effect - the log - may be conditional.
- `is_replaying()` is time-varying (True while catching up, False once live), so it's only meaningful at the instant the side effect fires. This also preserves the one legitimate live warning for a genuinely slow run that happened to be replayed before it breached.

The guard is prefixed with `workflow.in_workflow()` because `SLAWaiter` is also exercised directly in unit tests outside any workflow, where `is_replaying()` raises `_NotInWorkflowEventLoopError`. Same precedent as `products/tasks/backend/temporal/observability.py`.

> [!NOTE]
> Behavior change: a real SLA breach still logs exactly once (live). Only the replayed duplicates are suppressed.

I (Claude) ran the existing `products/batch_exports/backend/tests/temporal/test_metrics.py` suite locally against the dev ClickHouse - both `test_sla_waiter` cases pass. I also temporarily removed the guard and confirmed a breach re-emits, then restored it, to sanity-check the mechanism.

I did not add a new automated test. A mocked test would stub the exact `is_replaying()` signal the bug hinges on (circular), and a faithful test needs the Temporal `Replayer` harness against real workflow history - disproportionate machinery for a one-line guard that mirrors an already-reviewed pattern. Flagging this tradeoff explicitly rather than shipping a low-value test.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated and authored by Claude (PostHog Code), directed by @rossgray from a production log query. No repo skills shaped the code; I did invoke `/writing-tests`, which is what led to dropping the mocked test rather than shipping it.

Along the way I first proposed a "cancellation wind-down" theory (the cancelled workflow's cleanup outrunning the SLA timer). The reviewer pointed out the run had already finished at 09:09, which killed that theory - a completed run's tasks are evicted, so a live timer can't fire at 09:39. That reframed the cause as replay, confirmed by the logger having no replay awareness and `slo_interceptor.py` already guarding the identical footgun.

---

### DIFF
--- a/products/batch_exports/backend/temporal/metrics.py
+++ b/products/batch_exports/backend/temporal/metrics.py
@@ -552,6 +552,18 @@ class SLAWaiter:
         """Coroutine used to wait for SLA seconds."""
         await asyncio.sleep(self.sla.total_seconds())
 
+        # Temporal re-executes the workflow from the start on every replay (worker recovery,
+        # cache eviction, queries), which re-resolves the timer above from history and would
+        # re-emit a false, late "SLA breached" long after the run finished.
+        #
+        # The guard has to sit *here*, after the timer resolves, not around task creation:
+        # the `asyncio.sleep` above is a deterministic Temporal timer, so the task must be
+        # created on every replay or history diverges. And `is_replaying()` is time-varying
+        # (True while catching up, False once live), so it's only meaningful at the instant
+        # the side effect fires — suppress the log, never the timer.
+        if workflow.in_workflow() and workflow.unsafe.is_replaying():
+            return
+
         self._over_sla.set()
         LOGGER.warning(
             "SLA breached",
--- a/products/batch_exports/backend/temporal/metrics.py
+++ b/products/batch_exports/backend/temporal/metrics.py
@@ -555,8 +555,8 @@ class SLAWaiter:
         # Only log when the timer resolves live, not when it re-resolves from history during a
         # replay (worker recovery, cache eviction, queries).
         #
-        # This never drops a real breach. A breach that already happened was logged live the
-        # first time the timer fired; suppressing replays just avoids logging it again. A run
+        # A breach that already happened was logged closely to the first time the timer fired,
+        # so we don't expect a breach to be missed; suppressing replays just avoids logging it again. A run
         # that hasn't breached yet — including one recovering mid-flight — replays up to the
         # live edge and then fires its still-pending timer live, so it logs as normal.
         #
--- a/products/batch_exports/backend/temporal/metrics.py
+++ b/products/batch_exports/backend/temporal/metrics.py
@@ -552,15 +552,17 @@ class SLAWaiter:
         """Coroutine used to wait for SLA seconds."""
         await asyncio.sleep(self.sla.total_seconds())
 
-        # Temporal re-executes the workflow from the start on every replay (worker recovery,
-        # cache eviction, queries), which re-resolves the timer above from history and would
-        # re-emit a false, late "SLA breached" long after the run finished.
+        # Only log when the timer resolves live, not when it re-resolves from history during a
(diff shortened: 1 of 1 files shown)


products/batch_exports/backend/temporal/metrics.py
```diff
@@ -552,6 +552,18 @@ async def wait_for_sla(self) -> None:
         """Coroutine used to wait for SLA seconds."""
         await asyncio.sleep(self.sla.total_seconds())
 
+        # Temporal re-executes the workflow from the start on every replay (worker recovery,
+        # cache eviction, queries), which re-resolves the timer above from history and would
+        # re-emit a false, late "SLA breached" long after the run finished.
+        #
```
tomasfarias: How do you determine that the run finished and not that we are just recovering from a worker crash while the run is still going?

tomasfarias (commented):
I just have 1 question

products/batch_exports/backend/temporal/metrics.py
rossgray: updated the comment as per our discussion

products/batch_exports/backend/temporal/metrics.py
```diff
@@ -552,6 +552,20 @@ async def wait_for_sla(self) -> None:
...
+        # replay (worker recovery, cache eviction, queries).
+        #
+        # This never drops a real breach. A breach that already happened was logged live the
+        # first time the timer fired; suppressing replays just avoids logging it again. A run
+        # that hasn't breached yet — including one recovering mid-flight — replays up to the
+        # live edge and then fires its still-pending timer live, so it logs as normal.
+        #
+        # The guard sits here (after the timer resolves) rather than around task creation
+        # because the sleep is a deterministic Temporal timer: the task must be created on
+        # every replay or history diverges. So we suppress the log, never the timer.
```
tomasfarias:
```suggestion
        # A breach that already happened was logged closely to the first time the timer fired,
        # so we don't expect a breach to be missed; suppressing replays just avoids logging it again. A run
        # that hasn't breached yet — including one recovering mid-flight — replays up to the
        # live edge and then fires its still-pending timer live, so it logs as normal.
        #
        # The guard sits here (after the timer resolves) rather than around task creation
        # because the sleep is a deterministic Temporal timer: the task must be created on
        # every replay or history diverges. So we suppress the log, never the timer.
```

nit: This change introduces the possibility of missing breaches. So I don't think the comment should use words like "never".

tomasfarias (changes requested):

tomasfarias (approved):

PR #70700, opened 2026-07-14
feat(dashboards): nudge repeat dashboard viewers to subscribe

vdekrijger:
## Problem

Dashboards are viewed constantly, but subscriptions (scheduled email/Slack delivery) are barely discovered — most people who'd benefit from getting a dashboard delivered on a schedule never notice the feature exists. This is experiment candidate #1 from an uptake analysis of the subscriptions feature.

## Changes

Nudges repeat dashboard viewers toward subscriptions via the in-app notifications product plus a sticky toast (originally an inline banner; pivoted because the banner proved too easy to miss in hands-on testing):

- Triggered when a user has viewed the same dashboard 3+ times in the last 7 days (view recording deduped at 60s), has no existing subscription on it, can edit the dashboard, and is in the `test` variant of feature flag `dashboard-subscribe-nudge` (control/test — launch is manual).
- Delivery is two-layer: a new `POST /dashboards/:id/subscribe_nudge` endpoint creates a normal-priority in-app notification (new `subscription_nudge` type, dashboard-scoped access control, 30-day server-side dedupe per user+dashboard that is released if creation no-ops), and the client shows a sticky toast at the qualification moment (persists until acted on) with the same CTA.
- Both surfaces deep-link to the new-subscription form with `?prefill=nudge`, so a notification clicked days later in a fresh session still lands on a prefilled form; the params are consumed and stripped on arrival.
- The CTA skips the subscriptions list and opens the create form directly, prefilled with the viewer's email as recipient (frequency/day/time/destination already match the form's own weekly Monday-morning email defaults, so only the recipient needed prefilling).

Key pieces:
- `dashboardSubscribeNudgeLogic` (new) — records a view timestamp per dashboard in a persisted reducer, prunes anything older than 7 days, and exposes `isEligible` combining view count, dismissal, and dashboard-type. **Only reads the feature flag once `isEligible` is true** — reading a PostHog feature flag value reports its exposure event, so gating the read this way scopes experiment exposure to the population that could actually see the banner.
- Subscriptions are fetched only for dashboards already past the view threshold (not on every dashboard view), avoiding an extra API call for the vast majority of page views.
- `dashboardLogic` gained a small `subscriptionPrefill` action/reducer so the nudge can hand initial field values to the subscription form without going through URL query params.
- Three new analytics events: `dashboard subscribe nudge shown`, `dashboard subscribe nudge clicked` (carries `prefilled` and `via`: toast vs notification), and `dashboard subscribe nudge check failed` (with error detail) so a broken eligibility check is distinguishable from genuine ineligibility. There is deliberately no `dismissed` event — dismissal/read state belongs to the notifications product. `subscription created` additionally carries `ai_summary_prefilled`.

Automated only — no manual browser testing was done.

- `dashboardSubscribeNudgeLogic.test.ts` (new, 23 cases): view-count threshold (parameterized), 7-day pruning, dismissal (+ its captured event), and three cases proving the feature-flag read is gated on eligibility (unresolved subscription check, existing subscription, and the eligible/shown path with single-capture dedup). Guards the one behavior explicitly required here — that ineligible views never trigger flag exposure.
- Extended `subscriptionLogic.test.ts` (18 cases total): opening the "new" subscription route with a prefill lands with the prefilled values and marks the form changed (so "Create subscription" is enabled immediately); a plain "new" open (no prefill, existing test) now also asserts the form stays unmarked. Catches the prefill either not applying, or applying without enabling submit.
- `pnpm --filter=@posthog/frontend typescript:check` and `pnpm --filter=@posthog/frontend format` — clean on the touched files (pre-existing unrelated errors elsewhere on master, untouched by this change).

## Proof of work

CleanShot 2026-07-15 at 11 59 23

CleanShot 2026-07-15 at 11 21 49
CleanShot 2026-07-15 at 11 34 43
  

[AB test](https://us.posthog.com/project/2/experiments/383714)

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A — internal experiment banner, no documented workflow changed.

### Review hardening (second commit)

A multi-pass local review (simplify pass + six-reviewer swarm, four passes) hardened the initial implementation before marking ready: view-count dedupe so the 3-view bar means three real visits, subscription re-check on modal close so the banner cannot reappear right after subscribing, prefill lifecycle fixes (survives only on the 'new' form; insight auto-select no longer clobbers the dirty flag on dashboards with tiles), silent-failure conventions for the background check, and bounded localStorage in both dimensions. Both test suites grew accordingly (41 tests total).

## 🤖 Agent context

Built end-to-end by Claude (Claude Code) from a detailed brief plus a mid-session scope addition (prefilling the subscription form). Skills invoked: `/writing-kea-logics`, `/writing-tests`.

Decisions made along the way:
- Split "is this dashboard even a candidate" (view count, dismissal, dashboard type — all cheap, always available) from "is there already a subscription" (requires a fetch) into two tiers, so the subscriptions API is only called for the small slice of views that cross the threshold, not every dashboard view.
- The exposure-gating requirement (don't read the flag for ineligible users) is enforced by only touching `featureFlags[key]` inside a selector body guarded by `isEligible` — PostHog's feature flag proxy reports `$feature_flag_called` on property access, so this is the mechanism, not just a convention.
- Prefill values are threaded through kea props/actions (`dashboardSubscribeNudgeLogic.subscriptionPrefill` → `SubscriptionsModal` → `subscriptionLogic` `initialValues`), not URL query params, even though the codebase has an existing `?target_type=` query-param precedent for the same form. The prefill lives only while the modal route is the 'new' form, and the discard-changes guard is scoped to a captured prefill baseline so an untouched prefilled form never prompts.
- The prefill is applied via `setSubscriptionValues` (not folded into the loaded baseline) specifically so the form's dirty-tracking flips to true immediately — otherwise the create button stays disabled until the user edits something, defeating the "essentially one confirm click" goal.

Launch of the `dashboard-subscribe-nudge` flag is manual and out of scope for this PR.


frontend/src/scenes/dashboard/dashboardSubscribeNudgeStoreLogic.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+    return changed ? pruned : log
+}
+
+// Singleton store behind the dashboard subscribe nudge: one persisted map of recent views for
+// all dashboards (instead of a localStorage entry per dashboard ever viewed), plus the set of
+// dashboards permanently excluded from the nudge.
+export const dashboardSubscribeNudgeStoreLogic = kea<dashboardSubscribeNudgeStoreLogicType>([
+    path(['scenes', 'dashboard', 'dashboardSubscribeNudgeStoreLogic']),
```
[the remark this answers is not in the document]
vdekrijger: Addressed in ef9655261f8. `dashboardSubscribeNudgeStoreLogic` is now a keyed logic scoped by team + user (`dashboardNudgeScopeKey()` → `${teamId}:${userId}`), so kea derives a distinct localStorage key per account and view counts / suppression markers no longer cross accounts on a shared browser. Note we deliberately use the `OrNone` app-context accessors rather than `getCurrentTeamId()`, which throws when the team is unknown — this logic mounts on public/shared dashboards before the placement gate. Covered by a test asserting two scope keys keep isolated state.

products/dashboards/backend/api/dashboard.py
```diff
@@ -3255,6 +3285,57 @@ def create_unlisted_dashboard(self, request: Request, *args: Any, **kwargs: Any)
...
+        "dedupe window, so repeat calls return 200 with created=false.",
+    )
+    @action(methods=["POST"], detail=True, url_path="subscribe_nudge", required_scopes=["dashboard:write"])
+    def subscribe_nudge(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        dashboard = self.get_object()
+        user = cast(User, request.user)
+
+        dedupe_key = f"dashboard_subscribe_nudge:{user.pk}:{dashboard.pk}"
+        if not safe_cache_add(dedupe_key, "1", timeout=SUBSCRIBE_NUDGE_DEDUPE_TTL_SECONDS):
+            return Response({"created": False}, status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
vdekrijger: Addressed in 613af277527. Added the notifications facade's `has_been_dispatched(...)` as a durable dedupe check before `create_notification`, with the `safe_cache_add` sentinel kept in front purely as the cheap concurrency guard. Dedupe no longer depends on Redis health, and the contract is now once-ever per (user, dashboard) rather than a 30-day window — which also matches the client's permanent notified marker. The endpoint's OpenAPI description was updated accordingly, and a backend test wipes the cache entirely after a successful create and asserts the second call still dedupe-skips.

frontend/src/lib/constants.tsx
```diff
@@ -277,6 +277,9 @@ export const FEATURE_FLAGS = {
     DASHBOARD_INLINE_TILE_INSERTION: 'dashboard-inline-tile-insertion', // owner: @MattPua #team-analytics-platform
     DASHBOARD_LAYOUT_DISCARD_PROMPT: 'dashboard-layout-discard-prompt', // owner: @cory.s #team-analytics-platform
     DASHBOARD_POSTHOG_AI_BUTTON_LABEL: 'dashboard-posthog-ai-button-label', // owner: @pauldambra #team-product-analytics multivariate=control,control_b,test
+    DASHBOARD_QUICK_FILTERS_EXPERIMENT: 'dashboard-quick-filters-experiment', // owner: @vdekrijger #team-product-analytics multivariate=control,test
+    DASHBOARD_SUBSCRIBE_NUDGE: 'dashboard-subscribe-nudge', // owner: #team-analytics-platform multivariate=control,test, nudges repeat dashboard viewers with no subscription toward setting one up
+    DASHBOARD_SUBSCRIBE_PLACEMENT: 'dashboard-subscribe-placement', // owner: @MattPua #team-analytics-platform multivariate=control,button,menu
```
MattPua, line 282: i think thomas removed quick filters yesterday, and i removed subsribe placement, probbably need to rebase off master

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
```diff
@@ -0,0 +1,268 @@
...
+            },
+        ],
+        // Team-wide subscription count, fetched only for free-tier candidates: those orgs can
+        // create subscriptions until SubscriptionFreeTierLimit.COUNT, and nudging someone already
+        // at the limit would push them straight into the create-form paywall.
+        teamSubscriptionCount: [
+            null as number | null,
+            {
+                loadTeamSubscriptionCount: async (_?: unknown, breakpoint?: BreakPointFunction) => {
+                    // limit=1 keeps the payload tiny; `count` reflects the team's full total.
```
MattPua, line 90: nit: not a useful comment

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
```diff
@@ -0,0 +1,268 @@
...
+                    const response = await subscriptionsList(String(getCurrentTeamId()), { limit: 1 })
+                    breakpoint?.()
+                    return response.count ?? 0
+                },
+            },
+        ],
+        // Asks the backend to deliver the in-app nudge notification. The server dedupes per
+        // (user, dashboard) and reports whether a notification was actually created.
```
MattPua, line 98: nit: can probably remove this

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
```diff
@@ -0,0 +1,268 @@
...
+                        limit: 1,
+                    })
+                    breakpoint?.()
+                    return (response.count ?? 0) > 0
+                },
+            },
+        ],
+        // Team-wide subscription count, fetched only for free-tier candidates: those orgs can
+        // create subscriptions until SubscriptionFreeTierLimit.COUNT, and nudging someone already
+        // at the limit would push them straight into the create-form paywall.
```
MattPua, line 85: does this really matter, if our criteria of checking was whether or not they have subscriptions at all?

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
```diff
@@ -271,12 +280,47 @@ export const subscriptionLogic = kea<subscriptionLogicType>([
...
     })),
 
-    listeners(({ actions, values, props, selectors }) => ({
+    listeners(({ actions, values, props, selectors, cache }) => ({
         submitSubscriptionSuccess: ({ subscription }) => {
             if (subscription?.target_type === 'slack' && subscription.target_value && subscription.integration_id) {
                 recordRecentSlackChannel(subscription.integration_id, slackChannelId(subscription.target_value))
             }
         },
+        applyInsightSelectionDefaults: ({ selectedIds }) => {
```
MattPua, line 289: nit: i was confused for a while at what this was doing cuz of the naming, maybe should be something like

applyDefaultSelectedInsights

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
```diff
@@ -419,17 +463,55 @@ export const subscriptionLogic = kea<subscriptionLogicType>([
...
+                actions.setSubscriptionValues(prefill)
+                cache.prefillBaseline = { ...NEW_SUBSCRIPTION, ...prefill }
+                posthog.capture('dashboard subscribe nudge clicked', {
+                    dashboard_id: props.dashboardId,
+                    prefilled: !!values.user?.email,
+                    via:
+                        searchParams[SUBSCRIPTION_PREFILL_PARAMS.viaParam] === SUBSCRIPTION_PREFILL_PARAMS.viaToast
+                            ? SUBSCRIPTION_PREFILL_PARAMS.viaToast
+                            : SUBSCRIPTION_PREFILL_PARAMS.viaNotification,
+                })
```
MattPua, line 513: wouldnt it make more sense to add this capture event on the button in the nudge itself? or is it cuz the nudge can come from notifications or the toast

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
```diff
@@ -255,6 +262,8 @@ export const subscriptionLogic = kea<subscriptionLogicType>([
                         insight_short_id: props.insightShortId,
                         subscription_id: updatedSub.id,
                         target_type: updatedSub.target_type,
+                        // True when the nudge flow's deferred AI-summary default was applied to this form.
```
MattPua, line 265: nit: not a useful comment

services/mcp/src/api/generated.ts
```diff
@@ -64264,6 +64269,18 @@ export namespace Schemas {
       Xs: 'xs',
     } as const;
 
+    export type EnvironmentsDashboardsSubscribeNudgeCreateParams = {
```
MattPua, line 64272: this nudge is solely UI right? do we want MCP changes for this?

MattPua (approved):

frontend/src/lib/constants.tsx
vdekrijger: Fixed in follow-up draft PR https://github.com/PostHog/posthog/pull/71592. It removes the stale quick-filter and subscribe-placement constants that were reintroduced.

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
vdekrijger: Fixed in follow-up draft PR https://github.com/PostHog/posthog/pull/71592. I removed the low-value payload comment and clarified the loader through naming instead.

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
vdekrijger: Fixed in follow-up draft PR https://github.com/PostHog/posthog/pull/71592. I removed this comment.

frontend/src/scenes/dashboard/dashboardSubscribeNudgeLogic.ts
vdekrijger: The check is still needed because it is team-wide and enforces the free-tier creation cap, while the existing-subscription check is scoped to the current dashboard. Follow-up draft PR https://github.com/PostHog/posthog/pull/71592 renames it to `freeTierSubscriptionCount` and makes that distinction explicit.

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
vdekrijger: Fixed in follow-up draft PR https://github.com/PostHog/posthog/pull/71592 by renaming the action to `applyDefaultSelectedInsights`.

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
vdekrijger: It stays at route entry because both the toast and notification deep links enter the flow here. Follow-up draft PR https://github.com/PostHog/posthog/pull/71592 adds a concise comment explaining that choice.

products/subscriptions/frontend/components/Subscriptions/subscriptionLogic.ts
vdekrijger: Fixed in follow-up draft PR https://github.com/PostHog/posthog/pull/71592. I removed this comment.

services/mcp/src/api/generated.ts
vdekrijger: The endpoint is not exposed as an MCP tool: its entry in `products/dashboards/mcp/tools.yaml` is disabled. This generated API type is expected because MCP codegen consumes the shared OpenAPI schema. Verified while preparing follow-up draft PR https://github.com/PostHog/posthog/pull/71592.

PR #70701, opened 2026-07-14
feat(error-tracking): move autocapture opt-in to settings (phase 1)

ablaszkiewicz:
## Problem

The exception autocapture toggle is stored on `Team.autocapture_exceptions_opt_in`, but it belongs with the rest of the error tracking settings. We want it on the `ErrorTrackingSettings` model so the setting lives with its product and stops adding surface to the hot `Team` table.

This is the first of a few PRs. It only stages the data. Nothing reads the new location yet, so behavior is unchanged.

## Changes

Expand phase of a standard expand / backfill / contract move.

- New nullable `autocapture_exceptions_opt_in` column on `ErrorTrackingSettings`. Nullable on purpose: it matches the sibling fields, and `null` reads as disabled just like the old `bool(None)` coercion. It also sidesteps the cross-language `NOT NULL` hazard, since nodejs writes raw `INSERT`s into this table.
- A `post_save` signal on `Team` that dual-writes any opt-in change onto `ErrorTrackingSettings`. Enabling creates a row, disabling syncs the existing row, and teams that never opted in stay row-less.
- A `sync_autocapture_opt_in` helper as the single source of the "only opted-in teams get a row" rule, shared by the signal and the backfill.
- A `backfill_error_tracking_autocapture_opt_in` management command for teams that opted in before the signal shipped. Dry-run by default, idempotent, run manually on a pod.

Why a signal instead of hooking the serializers: there are three real writers of this field (the Team serializer, the Project serializer passthrough, and product enablement). One `post_save` receiver covers all of them plus admin and shell writes. It also keeps the dependency pointing product to core, so we avoid importing the isolated error tracking facade into the hot, early-loaded `team.py` and `project.py`. No serializer changes at all.

> [!NOTE]
> Reads still come from `Team`. Moving reads (remote config, serializers, the frontend toggle) and dropping the old column come in later PRs.

**Deploy order:** ship this first so the signal starts dual-writing, then run the backfill on a pod for the historical rows. Because the signal covers everything from deploy onward, there is no need to re-run the backfill before the read cutover.

Automated tests I (Claude) actually ran locally, all green:

- `products/error_tracking/backend/tests/test_autocapture_opt_in_sync.py` — the dual-write signal. Catches: enabling creates the row, disabling syncs the existing row without deleting it (the stale-row bug a snapshot-only backfill would leave), never-opted-in teams get no row, and a save that doesn't touch the field doesn't resync.
- `products/error_tracking/backend/management/commands/test/test_backfill_error_tracking_autocapture_opt_in.py` — the backfill only touches historical opted-in teams, creates or updates their row while preserving other settings, and writes nothing on a dry run.
- Regression sweep for the new global signal: `posthog/api/test/test_team.py` (218), `posthog/api/test/test_product_enablement.py` (11), and the error tracking settings API tests (10) all pass unchanged. No query-count assertions wrap Team saves, and the one unused-snapshot warning in `test_team.py` is pre-existing (confirmed by stashing this change).

I did not do manual UI testing.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (actually Claude, Opus 4.8) wrote this under my direction. We scoped it first: located every reader and writer of `autocapture_exceptions_opt_in`, and found that the companion `autocapture_exceptions_errors_to_ignore` field is effectively dead (not in remote config, no SDK consumer, no UI), so it is deliberately left out of this move.

The dual-write mechanism was the main design decision. A backfill-only Phase 1 leaves a hole: a team that disables autocapture between the backfill and the read cutover would keep capturing, because the backfill only writes `True`. So we added the signal to keep the new column continuously correct, and demoted the backfill to historical rows.

Signal vs serializer hooks was the other call. Explicit hooks would have meant three write sites and importing the isolated product facade into core hot paths. The signal is one chokepoint with a clean dependency direction, at the cost of a guarded no-op query on Team saves (rare settings actions, not per-request).

Skills invoked: `/django-migrations` (nullable column, hot-table and cross-language `NOT NULL` checks) and `/writing-tests` (gated each test to a named regression).


products/error_tracking/backend/management/commands/backfill_error_tracking_autocapture_opt_in.py
```diff
@@ -0,0 +1,154 @@
...
+        if not live_run:
+            logger.info("backfill_dry_run_complete", teams=len(candidate_ids))
+            return
+
+        start_time = time.monotonic()
+        synced = 0
+        for offset in range(0, len(candidate_ids), batch_size):
+            chunk = candidate_ids[offset : offset + batch_size]
+            synced += self._sync_batch(chunk)
+            logger.info("backfill_progress", processed=offset + len(chunk), synced=synced, last_team_id=chunk[-1])
```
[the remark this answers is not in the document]
ablaszkiewicz: fixed

products/error_tracking/backend/management/commands/backfill_error_tracking_autocapture_opt_in.py
```diff
@@ -0,0 +1,157 @@
...
+
+        elapsed = time.monotonic() - start_time
+        logger.info("backfill_complete", synced=synced, elapsed_s=round(elapsed, 2))
+
+    def _sync_batch(self, team_ids: list[int]) -> int:
+        # Re-read live state: only mirror teams that are STILL opted in, so a disable that
+        # landed since the candidate snapshot doesn't get clobbered back to True. A team no
+        # longer opted in and still row-less is already correct (row-less == disabled).
```
cat-ph: I think there's _still_ a race here? should it be a transaction? i.e. between still_opted_In and bulk_create, small enough but non-zero 😬

cat-ph (approved):
🚀

products/error_tracking/backend/management/commands/backfill_error_tracking_autocapture_opt_in.py
ablaszkiewicz: In order for this to happen, someone has to switch that setting during the exact time when we process a batch containing this team (I expect this window to be tens of milliseconds). This is extremely unlikely to happen.

I feel uncomfortable locking the team entity for such simple thing

cat-ph: yeahhhhh true, I was thinking about row locks on those updating rows, but still _extremely_ unlikely, I wouldn't worry about it either too tbh

ablaszkiewicz: eh

ablaszkiewicz: let's leave my comfort zone

ablaszkiewicz: just to be double sure

ablaszkiewicz: alright now it's bulletproof. Added a transaction

PR #70717, opened 2026-07-14
feat(data-warehouse): implement fireworks_ai import source

Gilbert09:
## Problem

Fireworks AI is a widely used inference and fine-tuning platform for open models, but its `fireworks_ai` warehouse source was only a scaffolded stub (`unreleasedSource=True`, no fields, no sync logic). Teams running training and inference on Fireworks can't join that platform data (training runs, dataset inventory, deployment fleet state) with their product data in PostHog.

**Why:** fill in the scaffolded Fireworks AI connector end-to-end so the source can sync real data, following the warehouse-source architecture contract.

## Changes

Implements the source per the `implementing-warehouse-sources` skill, split into the standard three files:

- **`settings.py`** – declarative endpoint catalog for 10 tables: `models`, `datasets`, `deployments`, `deployed_models`, `supervised_fine_tuning_jobs`, `reinforcement_fine_tuning_jobs`, `batch_inference_jobs`, `evaluation_jobs`, `evaluators`, `users`. Primary key is the AIP resource `name` (globally unique full path), partition key is the stable `createTime` (never `updateTime`).
- **`fireworks_ai.py`** – transport over the control-plane REST API (`https://api.fireworks.ai/v1/accounts/{account_id}/...`), Bearer auth via `make_tracked_session()`, `pageToken`/`nextPageToken` pagination (pageSize 200), tenacity retries for 429/5xx, and sanitized HTTP errors (no query strings or response bodies in stored error state). Account ID input is normalized so a pasted `accounts/my-account` prefix doesn't break the path.
- **`source.py`** – `ResumableSource` wiring: page-token resume state saved after each yielded page, `validate_credentials` (invalid-account-id short-circuit, 200/401/403/404 mapping, 403 accepted at source-create per skill guidance), `get_non_retryable_errors` for 401/403/404, static `get_schemas` with `lists_tables_without_credentials = True`.

Supporting pieces:

- `canonical_descriptions.py` for all 10 tables, sourced from the official API reference.
- Icon at `frontend/public/services/fireworks_ai.svg` (official brand SVG from fireworks.ai).
- `FireworksAISourceConfig` regenerated (`api_key`, `account_id`).
- `SOURCES.md`: moved `fireworks_ai` from Scaffolded to Implemented (HTTP / requests / tracked ✅).
- User-facing doc following the `documenting-warehouse-sources` template, kept at `products/.../fireworks_ai/docs/fireworks-ai.mdx` because no posthog.com checkout was available in this environment – it needs to move to `posthog.com/contents/docs/cdp/sources/fireworks-ai.mdx` (the `docsUrl` already points there).

Deliberate scope decisions:

> [!NOTE]
> - **Full refresh only.** The API documents an AIP-160 `filter` param but doesn't enumerate filterable fields, and server-side timestamp filtering couldn't be verified without live credentials, so no table advertises incremental sync (noted in code comments).
> - **`apiKeys` excluded.** The collection is nested per user and its schema carries key material – it must not land in a warehouse table.
> - Stays behind `unreleasedSource=True` with `releaseStatus=ALPHA` until it's validated against a live account.

No Django migration needed – the `FireworksAI` enum value and its migration already landed with the scaffold; `makemigrations --check` is clean.

- `pytest products/warehouse_sources/.../fireworks_ai/tests/` – 57 tests pass. `test_fireworks_ai.py` covers the regressions no existing test catches: paginator dropping or looping the page token, resume state saved before yield (data loss on crash), pasted `accounts/` prefix doubling the URL path, sanitized errors leaking pagination tokens, and unstable partition keys. `test_fireworks_ai_source.py` covers credential status-code mapping (incl. account-not-found), non-retryable error matching, and pipeline plumbing.
- `pytest .../sources/tests/` (1551 cross-source guard tests incl. source categories) – pass.
- `hogli ci:preflight --fix` – 0 failures.
- Endpoint paths, envelopes (`nextPageToken`, camelCase collection keys), and auth behavior verified against the Fireworks API reference plus unauthenticated probes of the live API (all list endpoints return 401 without a valid key). I could not run a live sync – no Fireworks credentials were available in this environment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc added in-repo at `products/warehouse_sources/backend/temporal/data_imports/sources/fireworks_ai/docs/fireworks-ai.mdx`; needs to land in the posthog.com repo (see Changes).

## 🤖 Agent context

Authored with Claude Code (PostHog Code cloud task). Skills invoked: `implementing-warehouse-sources`, `documenting-warehouse-sources`, `writing-tests`. Modeled on the `together_ai` and `anthropic` sources (closest analogs: Bearer-auth AI platform APIs; anthropic for the resumable page-token pattern). Chose a hand-rolled tracked-session transport over `rest_source.RESTClient` because the AIP page-token flow is a dozen lines and keeps resume checkpointing explicit. Ran `generate_source_configs` and kept only the FireworksAI hunk – the generator also emitted unrelated env-dependent drift (Stripe config, formatting) that was reverted to keep the diff minimal.

---


danielcarletti (approved):
LGTM

PR #70721, opened 2026-07-14
feat(data-warehouse): implement platform_sh import source

Gilbert09:
## Problem

The `platform_sh` warehouse source existed only as a scaffolded stub (registered with `unreleasedSource=True`, empty fields, no sync logic). Teams hosting on Platform.sh or Upsun couldn't sync their deployment data into the PostHog Data warehouse.

**Why:** Platform.sh's activity log is the audit trail of every deploy, push, cron, and backup - and the API prunes expired activities, so syncing it into the warehouse is what makes deployment history durable and queryable alongside product data.

## Changes

Implements the source end-to-end following the `implementing-warehouse-sources` architecture contract (`source.py` / `settings.py` / `platform_sh.py` split), as a `ResumableSource`:

| Table | Fan-out | Sync | Primary key |
| --- | --- | --- | --- |
| `organizations` | top-level | full refresh | `id` |
| `projects` | per org | full refresh | `id` |
| `environments` | per org → project | full refresh | `project_id`, `id` |
| `activities` | per org → project | incremental on `created_at` | `project_id`, `id` |
| `subscriptions` | per org | full refresh | `id` |
| `members` | per org | full refresh | `organization_id`, `id` |

Key behaviors:

- **Auth**: the user's API token is exchanged for a ~15-minute OAuth2 bearer token (`grant_type=api_token`, basic auth `platform-api-user:`). The client refreshes proactively before expiry and once reactively on a mid-run 401. Both brand host pairs are supported (Platform.sh and Upsun) via a select field, which is declared in `connection_host_fields` so retargeting re-requires the secret.
- **Pagination**: org-level lists follow the `{items, _links}` cursor envelope, with next-page URLs resolved relative to the request URL and pinned to the configured host (a tampered link can't exfiltrate the bearer token). The activities feed pages backwards via `count` + `starts_at` with id dedupe, matching the official PHP client / CLI behavior.
- **Incremental activities**: client-side watermark stop (the endpoint has no lower-bound filter), `sort_mode="desc"` so the watermark only finalizes after a fully successful sync, plus a 24h default lookback so activity state changes (in_progress → complete) are re-read instead of frozen.
- **Resumable**: checkpoints the organizations page URL after each page's children are fully processed; resume re-fans that page and merge dedupes.
- **Redaction**: activity `log` output is dropped and `basic_auth` credential blocks are stripped recursively, so account secrets never land in queryable tables. All HTTP rides `make_tracked_session()` with both the API token and the current bearer token masked.
- Canonical table/column descriptions from the official API docs, `lists_tables_without_credentials = True` for the public docs table catalog, official logo mark extracted from docs.platform.sh assets, SOURCES.md moved to Implemented.

The source stays behind `unreleasedSource=True` with `releaseStatus=alpha` for now, per the rollout plan for this batch of sources.

> [!NOTE]
> Verified live without credentials: the token-exchange endpoints on both brands (bogus token → the exact 401 body the client handles) and the endpoint/parameter surface against the current OpenAPI spec plus the official PHP client and CLI source. Not verifiable without a real account: whether `filter[updated_at]` on the org project list actually filters. I made the conservative call - `projects` ships as full refresh (it's a small list), with a comment in `settings.py` noting the documented filter as a future upgrade once verified against a live account.

Automated tests only (no live Platform.sh account was available):

- `tests/test_platform_sh.py` (transport): mid-run 401 triggers exactly one token re-exchange and retry (a long sync is guaranteed to cross the 15-minute token expiry - dropping the refresh breaks every big backfill); relative HAL `_links.next` resolution (breaking it silently syncs one page); off-host URL refusal (bearer-token exfiltration guard); activities backward paging with boundary dedupe and termination (without the nothing-new stop the walk loops forever); watermark stop + keeps-walking-without-watermark (without the stop every incremental sync re-walks full history); `basic_auth`/`log` redaction; resume-from-saved-state; per-endpoint `SourceResponse` config.
- `tests/test_platform_sh_source.py` (source class): schema flags (only `activities` incremental, with lookback), non-retryable error keys for both hosts (missing one means a bad credential retries forever), credential-validation and pipeline argument plumbing, resumable manager binding, `connection_host_fields`.

All 35 pass, plus the registry-wide suites (`test_source_config_generator.py`, `test_source_categories.py` - 1537 tests) and `hogli ci:preflight --fix` with no failures.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The user-facing doc for posthog.com is written per the `documenting-warehouse-sources` skill and passes `python manage.py audit_source_docs` against a posthog.com checkout, but it needs to land in the posthog.com repo separately (this PR can't carry it). `docsUrl` already points at `https://posthog.com/docs/cdp/sources/platform-sh`.

<details>
<summary>contents/docs/cdp/sources/platform-sh.md (to land in posthog.com)</summary>

```markdown
---
title: Linking Platform.sh as a source
sidebar: Docs
showTitle: true
availability:
  free: full
  selfServe: full
  enterprise: full
sourceId: PlatformSh
---

<AlphaRelease />

The Platform.sh connector syncs your Platform.sh (or Upsun) organizations, projects, environments, deploy activity, subscriptions, and organization members into PostHog, so you can analyze your deployments and infrastructure alongside your product data.

## Prerequisites

You need a Platform.sh or Upsun account with access to the organizations you want to sync, and the ability to create an API token for your user.

## Adding a data source

<SourceSetupIntro />

When linking Platform.sh, you'll need:

- **API token** – create one in the Console under **My profile** > **API tokens**. See the [Platform.sh docs](https://docs.platform.sh/administration/cli/api-tokens.html) or [Upsun docs](https://docs.upsun.com/administration/cli/api-tokens.html) for a walkthrough. The token has the same access as your user account, so no extra scopes are needed.
- **API endpoint** – pick the product you use: Platform.sh (`api.platform.sh`) or Upsun (`api.upsun.com`). A Platform.sh token doesn't work against the Upsun API, and vice versa.

## Sync modes

<SyncModes />

The activities table syncs incrementally on `created_at`. Platform.sh prunes expired activities from its API, so incremental sync is also what preserves your full deploy history in PostHog beyond the API's retention window. The remaining tables are small lists and sync as full refresh.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

Activity rows exclude the raw `log` output, and environment rows exclude HTTP basic-auth credentials, so secrets from your Platform.sh account don't land in queryable tables.

## Troubleshooting

- If you see an authentication error, check that the token is valid in the Console and that you picked the API endpoint matching where the token was created (Platform.sh vs. Upsun).
- If recent activity history looks shorter than expected, note that Platform.sh prunes expired activities from its API; PostHog keeps everything it has already synced.

<TroubleshootingLink />
```

</details>

## 🤖 Agent context

- Built with Claude Code following the repo skills `implementing-warehouse-sources` (workflow, architecture contract, tracked transport), `/writing-tests` (invoked before authoring the test modules), and `documenting-warehouse-sources` (doc template + audit).
- Decisions: chose `ResumableSource` without `WebhookSource` for this first cut - Platform.sh webhooks are per-project integrations (one webhook per project across every org), which doesn't fit the single-webhook management model and couldn't be verified without credentials; noted as a follow-up. Shipped `projects` as full refresh rather than trusting the documented-but-unverified `filter[updated_at]`. Made `activities` the one incremental table because API-side pruning makes full refresh lossy there. Verified the undocumented `count`/`starts_at` activity params against the official PHP client and CLI sources rather than the (silent) OpenAPI spec.
- The scaffolded enum/schema wiring (types.py, schema-general.ts, `_load_all.py`, migration 0066) already existed, so no migrations or schema rebuild were needed; only `generate:source-configs` was re-run (with an unrelated environment-dependent Stripe hunk reverted).

---


danielcarletti (approved):
LGTM

PR #70728, opened 2026-07-14
feat(conversations): create email-channel tickets for anonymous widget requesters

abigailbramble:
## Problem

Anonymous widget requesters who leave an email address get a widget-channel ticket. Team replies then only appear back in the widget, which someone who submitted from a page they won't return to (for example a logged-out auth page) may never reopen. Their support answer effectively goes nowhere.

## Changes

When an anonymous requester supplies a valid email trait and the team has email enabled with a verified channel, the ticket is created on the email channel, sending from the team's primary channel. Team replies are then emailed. The ticket keeps its widget session binding, so it also stays visible in the widget. Identity-verified users are unchanged: they keep widget-channel tickets and see replies in the side panel.

Backend only. Stacked on #70725 (primary email channel), which this uses to pick the send-from identity.

> [!WARNING]
> **Open decision — please weigh in before this leaves draft.** This applies to every conversations team's widget, not just PostHog's own support. An anonymous request can set `email_from` to any address it types; once a team (or a public AI) replies, that address receives mail from the team's verified domain. That's a potential backscatter / impersonation vector, bounded only by the widget throttles. The ticket is flagged as unverified in the agent view (an `IdentityBadge`), which warns the agent about the claimed identity but does not stop the outbound email. Options discussed: restrict email-channel promotion to identity-verified requesters, require address verification (double opt-in) before emailing, or gate behind an explicit per-team setting. Kept draft until we pick one.

Cheap hardening already included: the email is length-capped to the model's limit (avoids a 500 on insert), and the subject (taken from the first message line) is stripped of control characters before it becomes an email header.

Automated only:

- `test_widget.py`: an anonymous email trait creates an email-channel ticket with the right recipient, config, and subject while keeping the widget session binding; a parameterized set locks the guardrails (no email trait, email channel disabled, no verified channel, invalid email all stay widget-channel); identity-verified users keep widget-channel tickets even with an email trait; and the ticket uses the team's primary channel rather than the oldest. Full `test_widget.py` suite passes.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Abigail directed this; I (Claude, via PostHog Code) wrote it. Carved out of a larger switchover branch. Skills invoked: `/improving-drf-endpoints` (widget endpoint) and `/writing-tests`.

This started as an unconditional product-wide behavior. A multi-agent review flagged the outbound-email abuse surface, which is why it's a separate draft PR with the decision called out rather than folded into the routing change. The channel selection prefers the team's primary channel (from the base PR) and falls back to the oldest verified one.

---

### DIFF
--- a/products/conversations/backend/api/tests/test_widget.py
+++ b/products/conversations/backend/api/tests/test_widget.py
@@ -3,13 +3,14 @@ import uuid
 from posthog.test.base import BaseTest
 from unittest.mock import patch
 
+from parameterized import parameterized
 from rest_framework import status
 from rest_framework.test import APIClient
 
 from posthog.models.comment import Comment
 
-from products.conversations.backend.models import Ticket
-from products.conversations.backend.models.constants import ChannelDetail, Status
+from products.conversations.backend.models import EmailChannel, Ticket
+from products.conversations.backend.models.constants import Channel, ChannelDetail, Status
 from products.conversations.backend.services.identity import compute_identity_hash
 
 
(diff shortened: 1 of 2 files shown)


JakobStadlhuber:
we need this, we thinking about doing a posthog function if this will not be released soon. hope this is comming soon.

PR #70733, opened 2026-07-14
feat(experiments): add property-based exposure freeze for local evaluation

mp-hog:
## Problem

"Freeze exposure" narrows an experiment's flag to a static cohort of already-exposed persons. Static cohorts are excluded from the local-evaluation flag definitions.

## Changes

- New \`freeze_mode="property"\` on \`POST .../freeze_exposure\`: ANDs caller-supplied property conditions into every release condition instead of a snapshot cohort. Locally evaluable, instant (no ClickHouse scan, no cohort build, no exposed-user cap), and works for group-aggregated flags (group-property conditions matching the flag's aggregation). Unfreeze/reset/ship-variant strip it cleanly via a new \`exposure_frozen_properties\` group stamp.
- New \`GET .../freeze_exposure_check\`: recommends a freeze mode by measuring the share of the flag's recent \`$feature_flag_called\` events with \`locally_evaluated=true\`, plus the group-aggregation check.
- Frontend detect-and-default flow: clicking "Freeze exposure" runs the check first. Common case keeps today's one-click cohort confirm. When the cohort mode is unsafe, a modal explains why and asks the user to confirm editable property conditions (default: person \`created_at\` before now) – the property mode is a proxy that's only correct for enrollment gated on the property, so it is never applied silently. Group-aggregated experiments now get the freeze option too.
- The cohort mode is unchanged and stays the default.

> [!NOTE]
> Sticky bucketing remains the durable long-term answer; this is deliberately a narrow, local-evaluation-compatible stopgap on the existing freeze machinery.

Automated tests only (the sandbox has no Postgres/ClickHouse, so backend tests will run in CI):

- \`test_experiment_service.py\`: property-mode freeze stamps/strips round-trip byte-for-byte (including a user condition identical to the frozen one – strip removes only the appended occurrence), condition validation matrix (cohort/flag types, aggregation mismatch, missing key/operator/value), group-aggregated support, and \`check_freeze_exposure\` threshold logic including query-failure fallback. These catch regressions the cohort-mode tests can't: a freeze that silently builds a cohort anyway, or an unfreeze that can't reverse a property-mode freeze.
- \`test_presentation_api.py\`: wiring guards – property-mode request body reaches the service, missing conditions 400, check endpoint serializes.
- \`test_freeze_exposure_clickhouse.py\`: CH-backed proof the detection query counts \`locally_evaluated\` per flag within the window (the unit tests stub \`execute_hogql_query\`).
- Jest (\`experimentLogic.test.ts\`, ran locally, all green): check-recommends-property opens the modal and never the plain confirm dialog; cohort recommendation keeps the dialog; property freeze sends the request body.

Not done here: \`hogli build:openapi\` (needs a running DB; generated types for the new request/response schemas need a regen – preflight flags this as the only advisory).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (PostHog Code) from a handoff plan Marcel provided. Skills invoked: /improving-drf-endpoints, /writing-tests, /adopting-generated-api-types. Notable decisions: the freeze request field is \`freeze_mode\` (not \`mode\`, which is an enum-collision-prone name); conditions are a list (\`property_conditions\`) rather than a single condition since the UI's PropertyFilters naturally edits a list and ANDed person properties stay locally evaluable; the over-cap case is not pre-detected by the check endpoint (it would need the same expensive scan the freeze runs) – the cohort freeze's size rejection remains the signal there. The plan's open question about SDK behavior on unresolvable static cohorts was already answered by an existing design note in \`freeze_exposure\`, so no flags-team confirmation was needed.

---

### DIFF
--- a/frontend/src/scenes/experiments/ExperimentView/ExperimentModals.tsx
+++ b/frontend/src/scenes/experiments/ExperimentView/ExperimentModals.tsx
@@ -14,6 +14,8 @@ import {
     Link,
 } from '@posthog/lemon-ui'
 
+import { PropertyFilters } from 'lib/components/PropertyFilters/PropertyFilters'
+import { TaxonomicFilterGroupType } from 'lib/components/TaxonomicFilter/types'
 import { FEATURE_FLAGS } from 'lib/constants'
 import { featureFlagLogic } from 'lib/logic/featureFlagLogic'
 import { urls } from 'scenes/urls'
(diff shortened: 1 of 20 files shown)


mp-hog:
Abandoned this approach as it has two complexities:
- What property is available for local-eval
- What property is working for freezing (e.g. a created_at would only work for new users)

Likely too fragile and easy to misconfigure

PR #70752, opened 2026-07-14
feat(logs): serve group-by aggregation from the log_attributes rollup

jonmcwest:
## Problem

The logs group-by aggregation (`LOGS_GROUP_BY` flag) scans the raw `logs` table for every request under a hard 10 GB read budget with `read_overflow_mode='throw'`.
On wide windows for high-volume teams the scan blows the budget and the query fails outright, so grouping is unusable exactly where it is most useful.

The pre-aggregated `log_attributes` rollup now carries `severity_text` per row (gen 3, cut over in [#70496](https://github.com/PostHog/posthog/pull/70496), exposed to HogQL in [#70635](https://github.com/PostHog/posthog/pull/70635)), which was the missing dimension: group-by responses report per-group error counts and accept severity filters, so the rollup could not serve them before.

## Changes

**Group-by reads the rollup when it can, scans logs when it must** (`products/logs/backend/group_by_query_runner.py`):

```mermaid
flowchart LR
    Q[group-by request] --> E{filters expressible<br/>on the rollup?}
    E -->|date range, services,<br/>severity, resource filters| R[(log_attributes<br/>rollup)]
    E -->|body search, log-attribute or<br/>non-severity log filters,<br/>column grouping| L[(logs<br/>raw scan)]
    class Q phYellow;
    class E phBlue;
    class R,L phGray;
```

- Rollup path: `sum(attribute_count)` for log counts, `sumIf(..., severity IN error/fatal)` for error counts, `max(time_bucket)` for last-seen, resource-attribute filters via the same fingerprint subquery the facet counts use.
- The rollup path runs under a tighter budget (5 GB / 30 s vs the scan's 10 GB / 60 s): the rollup is orders of magnitude smaller than `logs`, so a read approaching the scan budget there means something pathological and should fail fast.
- Fallback path is the existing raw-logs scan, unchanged.
- The shared rollup filter fragments (`service_names_expr`, `severity_levels_expr`, `resource_fingerprint_expr`, `attributes_query_date_range`) move onto `LogsFilterBuilder` / the runner mixin, and the facet-values runner now reuses them instead of hand-rolling the same expressions (second commit).

> [!NOTE]
> On the rollup path `last_seen` has 10-minute bucket precision and window bounds land on bucket edges, the same trade-off resource-attribute facet counts already make (the `last_seen` API field's help_text documents this). Attribute keys/values over 256 chars are absent from the rollup by design (the MV drops them).

**Test-infra hardening** (`posthog/conftest.py`): kept/reused ClickHouse test databases now converge and stay clean.

- `create_clickhouse_tables` re-runs `CREATE OR REPLACE` schema statements even when the table exists, so a kept database picks up distributed-table cutovers (all five such statements in the canonical schema are idempotent repoints).
- The logs data tables join the session-teardown truncate list. The Postgres test database's team-id sequence restarts every run while kept ClickHouse data survives, so a previous run's rollup rows can share the next run's team ids — which made the new group-by tests report phantom groups locally. The group-by test class also truncates at `setUpClass` to cover interrupted runs.

Automated only (I'm an agent, no manual run against a live cluster):

- `products/logs/backend/test/`: group-by (25), facet values (21), and logs query runner suites all green on a fresh ClickHouse test database, twice (the second run is the real test for the cross-run contamination fix).
- New tests, each catching a regression nothing else covers:
  - routing matrix (9 cases): a query with an inexpressible filter served from the rollup returns silently wrong counts, and the reverse reintroduces read-cap failures.
  - severity filter on the rollup path: catches losing the severity predicate or the `severity_text` dimension end to end (MV → rollup → query).
  - resource-attribute filter on the rollup path: catches dropping the fingerprint-subquery wiring, which would silently ignore the filter.
  - window-bounds test parameterized over both paths: bucket-edge bounds on the rollup, per-row timestamp bounds on the scan.
- Adversarial review pass over the diff (routing completeness against every `LogsQuery` field the scan honors, AST-node reuse in the fingerprint subquery, facet-runner refactor equivalence) found no defects; the risky paths were verified by execution against ClickHouse.
- `hogli ci:preflight` clean (OpenAPI/MCP types regenerated).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Jon asked me (Claude, via PostHog Code) to finish the group-by-over-rollup work on top of the gen-3 rollup cutover ([#70496](https://github.com/PostHog/posthog/pull/70496)).
An earlier iteration of this branch carried the canonical-schema alignment and the HogQL `severity_text` column itself; both merged to master separately, and the branch was rebuilt on current master (hence the force-push).
Skills invoked: /writing-tests.
Decisions along the way: kept the group source `column` and any filter the rollup lacks a dimension for on the raw-scan fallback rather than approximating; only service/fingerprint context filters are pushed into the resource-fingerprint subquery, since the outer `attribute_key` constraint would contradict the subquery's own attribute matching and empty it out; the conftest hardening stays in this PR because the group-by tests are the first to read the rollup with per-test inserts, which is what surfaced the cross-run contamination.

---

### DIFF
--- a/posthog/conftest.py
+++ b/posthog/conftest.py
@@ -47,30 +47,35 @@ def create_clickhouse_tables():
     }
 
     def missing(queries):
-        return [q for q in queries if get_table_name(q) not in existing_tables]
+        # Returns the rendered SQL of the statements that still need to run. CREATE OR
+        # REPLACE statements are how the canonical schema repoints an existing table
+        # (e.g. a distributed table cut over to a new data table); they must run even
+        # when the table already exists so kept/reused databases converge.
+        built = map(build_query, queries)
+        return [sql for sql in built if get_table_name(sql) not in existing_tables or "CREATE OR REPLACE" in sql]
 
-    mergetree_queries = list(map(build_query, missing(CREATE_MERGETREE_TABLE_QUERIES)))
+    mergetree_queries = missing(CREATE_MERGETREE_TABLE_QUERIES)
     if mergetree_queries:
         run_clickhouse_statement_in_parallel(mergetree_queries)
 
-    distributed_queries = list(map(build_query, missing(CREATE_DISTRIBUTED_TABLE_QUERIES)))
+    distributed_queries = missing(CREATE_DISTRIBUTED_TABLE_QUERIES)
     if distributed_queries:
         run_clickhouse_statement_in_parallel(distributed_queries)
 
     if settings.IN_EVAL_TESTING:
-        kafka_table_queries = list(map(build_query, missing(CREATE_KAFKA_TABLE_QUERIES)))
+        kafka_table_queries = missing(CREATE_KAFKA_TABLE_QUERIES)
         if kafka_table_queries:
             run_clickhouse_statement_in_parallel(kafka_table_queries)
 
-    mv_queries = list(map(build_query, missing(CREATE_MV_TABLE_QUERIES)))
+    mv_queries = missing(CREATE_MV_TABLE_QUERIES)
     if mv_queries:
         run_clickhouse_statement_in_parallel(mv_queries)
 
-    view_queries = list(map(build_query, missing(CREATE_VIEW_QUERIES)))
+    view_queries = missing(CREATE_VIEW_QUERIES)
     if view_queries:
         run_clickhouse_statement_in_parallel(view_queries)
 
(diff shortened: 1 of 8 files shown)


jonmcwest:
Closing — rethinking the approach.

PR #70755, opened 2026-07-14
feat(flags): filter feature flags by payload presence

JamesPatrickGill:
**_Quick Preamble:_** 

First time contribution, hopefully deemed useful, maybe mergeable -> mainly done to play around in the PostHog repo.

---

## Problem

The feature flags list already filters by status, type, tags, creator, and runtime, but there's no way to narrow to flags that carry a JSON payload. If you're auditing which flags ship a payload (remote config, or an ordinary boolean/multivariant flag with a payload attached), you have to open each flag to check. This adds a payload filter to close that gap, on both the API and the flags list UI.

## Changes

- **API** (`FeatureFlagViewSet._apply_filters`): new `has_payloads` query param on `GET /api/projects/{id}/feature_flags/`. `true` returns flags whose `filters.payloads` is a non-empty object, `false` returns flags with no payload (missing key or `{}`). Declared with an `OpenApiParameter` on `list`, so it flows into the generated frontend types and the `feature-flag-get-all` MCP tool.
- **UI** (`FeatureFlagFiltersSection`): a **Payload** dropdown (Any / Has payload / No payload), matching the existing Runtime filter's shape. Wired into `FeatureFlags.tsx` and gated behind the existing `filtersConfig`.
- **List logic** (`featureFlagsLogic`): the flags list applies the filter server-side and re-applies it client-side over the current page, so I added `flagMatchesPayloads` to mirror the backend definition (non-empty `filters.payloads`). The URL param is normalized on read to `'true' | 'false' | undefined`, so an unexpected value (e.g. a hand-edited `?has_payloads=any`) doesn't silently apply a filter while the dropdown still reads "Any".
- Regenerated OpenAPI / MCP artifacts (`api.schemas.ts`, MCP `generated.ts` / `api.ts` / tool wiring).

> [!NOTE]
> "Has payload" is defined as a non-empty `filters.payloads` object, so remote config flags (which store their config there) count as having a payload. This is intentional and called out in the param description — payload presence is distinct from the `remote_config` type, since a plain boolean or multivariant flag can also carry a payload.

**Default (any):**

any_selected

**Has Payload:**

has_payload_selected

I clicked through the flags list UI manually — the Payload dropdown filters correctly across Any / Has payload / No payload, and the URL param round-trips. Alongside that, the automated coverage (the parts Claude ran):

- **Backend** `test_get_flags_with_has_payloads_filter` (passed): covers `has_payloads=true`/`false` across a boolean flag with a payload, a multivariant flag with a variant payload, a flag with an empty `payloads: {}`, and a flag with no `payloads` key. Catches the queryset predicate treating "empty dict" or "missing key" as "has payload".
- **Frontend** `featureFlagsLogic.test.ts` + `FeatureFlagFilters.test.tsx` (75 passed): new cases for the URL round-trip when the router coerces `has_payloads` to a boolean, for an unrecognized URL value being dropped rather than filtered, and for the Payload filter rendering both on its own and alongside the other filters.
- `hogli ci:preflight` reports 0 failures; generated types regenerated with no drift after rebasing on master.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs update needed. The API reference at posthog.com/docs/api is generated from the OpenAPI spec (the param description ships with it), and the sibling list filters (`has_evaluation_contexts`, `evaluation_runtime`) aren't hand-documented, so this follows the existing convention.

## 🤖 Agent context

Directed by @JamesPatrickGill; the payload filter itself was built beforehand and Claude picked it up from the working tree. In this session Claude ran `/code-review` over the diff, which surfaced an edge case: the URL param wasn't normalized, so a value the dropdown never emits (`any`, or anything non-`true`/`false`) would apply a real filter while the control still showed "Any". That's fixed at the parse boundary (strictly better than the `evaluation_runtime` sibling, which has the same latent quirk) with a regression test added. Claude also confirmed the backend/client payload definitions agree, checked docs in both this repo and posthog.com, and rebased onto master with the generated files regenerated cleanly.

### DIFF
--- a/frontend/src/scenes/feature-flags/FeatureFlagFilters.tsx
+++ b/frontend/src/scenes/feature-flags/FeatureFlagFilters.tsx
@@ -13,6 +13,7 @@ export interface FeatureFlagFiltersConfig {
     status?: boolean
     createdBy?: boolean
     runtime?: boolean
+    payloads?: boolean
     tags?: boolean
 }
 
@@ -37,10 +38,12 @@ export function FeatureFlagFiltersSection({
         status: false,
         createdBy: false,
         runtime: false,
+        payloads: false,
         tags: false,
         ...filtersConfig,
     }
-    const hasNonSearchFilters = config.type || config.status || config.createdBy || config.runtime || config.tags
+    const hasNonSearchFilters =
+        config.type || config.status || config.createdBy || config.runtime || config.payloads || config.tags
 
     return (
         <div className="flex justify-between gap-2 flex-wrap">
@@ -213,6 +216,35 @@ export function FeatureFlagFiltersSection({
                                 ]}
                                 value={filters.evaluation_runtime ?? 'any'}
                                 data-attr="feature-flag-select-runtime"
+                            />
+                        </>
+                    )}
+                    {config.payloads && (
+                        <>
+                            <span className="ml-1">
+                                <b>Payload</b>
+                            </span>
+                            <LemonSelect
+                                dropdownMatchSelectWidth={false}
+                                size="small"
+                                onChange={(hasPayloads) => {
+                                    const { has_payloads, ...restFilters } = filters || {}
+                                    setFeatureFlagsFilters(
+                                        {
+                                            ...restFilters,
+                                            ...(hasPayloads !== 'any' ? { has_payloads: hasPayloads } : {}),
+                                            page: 1,
+                                        },
+                                        true
+                                    )
+                                }}
+                                options={[
+                                    { label: 'Any', value: 'any', 'data-attr': 'feature-flag-select-payloads-any' },
+                                    { label: 'Has payload', value: 'true' },
+                                    { label: 'No payload', value: 'false' },
+                                ]}
+                                value={filters.has_payloads ?? 'any'}
+                                data-attr="feature-flag-select-payloads"
                             />
                         </>
                     )}
--- a/frontend/src/scenes/feature-flags/featureFlagsLogic.ts
+++ b/frontend/src/scenes/feature-flags/featureFlagsLogic.ts
@@ -90,11 +90,20 @@ export function flagMatchesType(flag: FeatureFlagType, type?: string): boolean {
     return true
 }
 
+export function flagMatchesPayloads(flag: FeatureFlagType, hasPayloads?: string): boolean {
+    if (!hasPayloads) {
+        return true
+    }
+    const hasAnyPayload = Object.keys(flag.filters.payloads ?? {}).length > 0
+    return hasPayloads === 'true' ? hasAnyPayload : !hasAnyPayload
+}
+
 export function flagMatchesFilters(flag: FeatureFlagType, filters: FeatureFlagsFilters): boolean {
     return (
         flagMatchesSearch(flag, filters.search) &&
         flagMatchesStatus(flag, filters.active) &&
         flagMatchesType(flag, filters.type) &&
+        flagMatchesPayloads(flag, filters.has_payloads) &&
         // Archived flags are hidden unless explicitly filtered for, mirroring the API default
         (filters.archived === 'true' ? !!flag.archived : !flag.archived) &&
         (!filters.created_by_id?.length ||
@@ -137,6 +146,7 @@ export interface FeatureFlagsFilters {
     order?: string
     page?: number
     evaluation_runtime?: string
+    has_payloads?: string
     tags?: string[]
     excluded_tags?: string[]
 }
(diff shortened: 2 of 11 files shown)


products/feature_flags/backend/api/feature_flag.py
```diff
@@ -3635,6 +3643,17 @@ def _apply_filters(self, filters: dict, queryset: QuerySet) -> QuerySet:
                     queryset = queryset.filter(eval_tag_count__gt=0)
                 else:
                     queryset = queryset.filter(eval_tag_count=0)
+            elif key == "has_payloads":
```
haacked, line 3646: `suggestion`: This works on the list endpoint but not on bulk_delete. `has_payloads` isn't declared on [`BulkDeleteFiltersSerializer`](https://github.com/posthog/posthog/blob/e2c819611b2b3d35dbda3f7acb6e859061414de7/products/feature_flags/backend/api/feature_flag.py#L2632), so passing it there gets rejected as an unknown filter key with a 400. `has_evaluation_contexts` is declared on that serializer, but `has_payloads` isn't.

This isn't a regression the PR introduces: the select-all-matching bulk delete flow already 400s on an unrelated pre-existing gap (`page` isn't in that serializer either), so adding just `has_payloads` won't make bulk-delete-by-payload work yet. Add it anyway so the two filter surfaces stay in sync once that other gap is fixed.

Add to `BulkDeleteFiltersSerializer` (feature_flag.py:2632), alongside `has_evaluation_contexts`:

```python
has_payloads = serializers.BooleanField(
    required=False,
    help_text="When true, only matches flags carrying at least one payload value.",
)
```

frontend/src/scenes/feature-flags/featureFlagsLogic.ts
```diff
@@ -90,11 +90,20 @@ export function flagMatchesType(flag: FeatureFlagType, type?: string): boolean {
     return true
 }
 
+export function flagMatchesPayloads(flag: FeatureFlagType, hasPayloads?: string): boolean {
```
haacked, line 93: `suggestion`: `flagMatchesPayloads` has no direct test. Only the URL round-trip tests cover `has_payloads`, so if the `hasPayloads === 'true'` branch flips or the `payloads ?? {}` guard breaks, all tests pass but the list shows the wrong flags.

Write a test block modeled on `flagMatchesType`, checking a few payload shapes:

```ts
it.each<[Record<string, unknown> | undefined, string | undefined, boolean]>([
    [{ true: '"blue"' }, 'true', true],
    [{ true: '"blue"' }, 'false', false],
    [{}, 'true', false],
    [{}, 'false', true],
    [undefined, 'true', false],
    [undefined, undefined, true],
])('payloads=%p with has_payloads=%p → %p', (payloads, hasPayloads, expected) => {
    const flag = { ...NEW_FLAG, filters: { groups: [], payloads } } as FeatureFlagType
    expect(flagMatchesPayloads(flag, hasPayloads)).toBe(expected)
})
```

frontend/src/scenes/feature-flags/featureFlagsLogic.ts
```diff
@@ -90,11 +90,20 @@ export function flagMatchesType(flag: FeatureFlagType, type?: string): boolean {
...
+    if (!hasPayloads) {
+        return true
+    }
+    const hasAnyPayload = Object.keys(flag.filters.payloads ?? {}).length > 0
```
haacked, line 97: `nit`: `Object.keys(flag.filters.payloads ?? {}).length > 0` allocates an array just to check for emptiness, but it's negligible at ~100 flags a page.

frontend/src/scenes/feature-flags/FeatureFlagFilters.tsx
```diff
@@ -216,6 +219,35 @@ export function FeatureFlagFiltersSection({
                             />
                         </>
                     )}
+                    {config.payloads && (
+                        <>
+                            <span className="ml-1">
+                                <b>Payload</b>
+                            </span>
+                            <LemonSelect
```
haacked, line 227: `nit`: This select lacks an `aria-label`, so screen readers can't identify which filter it is. The Type, Status, and Runtime selects right above it have the same gap; add `aria-label="Payload"` if you're editing this block.

haacked (approved):
Solid first contribution! A couple of suggestions and two small nits inline, nothing blocking.

haacked:
Thanks for the contribution! In general, we try not to have filters for fields that aren't visible to the user in the flags list.

JamesPatrickGill:
Makes sense! Was just about to address the suggestions so good timing. Thanks anyhow.

PR #70772, opened 2026-07-14
chore(funnels): use quill FunnelChart for the steps bar viz

sampennington:
## Problem

The product-analytics steps funnel hand-builds its funnel look on top of the generic `BarChart` (hatched drop-off track, rounded and shadowed bars, percent value axis, index-keyed bands, click-to-converted semantics). All of that now lives in the purpose-built `FunnelChart` wrapper that was just merged into `@posthog/quill-charts`, so the scene was duplicating the wrapper's job.

## Changes

- `FunnelStepsBarChart` renders `FunnelChart` instead of a `BarChart` wired up with `buildFunnelStepsBarConfig`.
- The per-step `StepLegend` row is now supplied via `FunnelChart`'s `stepFooter`, which pixel-aligns each cell under its step's bars. That removes the component's own band-width math and manual legend grid.
- Clicks go through `onStepClick` (converted vs drop-off track) rather than `onPointClick`.
- The tooltip-interaction helper became `funnelStepsBarTooltipConfig`, returning a `TooltipConfig` for `FunnelChartConfig` rather than rewriting a full `BarChartConfig`.

Behavior is meant to be equivalent: the persons-modal click routing, the compare `trackData` track ceilings, the custom tooltip, and the pinnable-tooltip flag path all carry over unchanged. The shared `buildFunnelStepsBarConfig` stays put because the MCP funnel visualizer still uses it.

I (Claude) could not run the frontend toolchain in this environment (no `node_modules`), so I did not run typecheck, lint, jest, or the Storybook visual snapshots locally. CI will cover those. The existing `funnelStepsBarTransforms` unit tests were updated for the renamed tooltip helper; the visual-regression stories in `FunnelStepsBarChart.stories.tsx` exercise the default, breakdown, compare, and full-scene render paths and will re-baseline on CI.

## 🤖 Agent context

Authored by Claude (PostHog Code). The task was to adopt the newly merged quill `FunnelChart` in the product-analytics funnels. I mapped the existing `FunnelStepsBarChart` config surface onto `FunnelChartConfig`, chose to use the wrapper's `stepFooter` for the `StepLegend` row (letting the chart own band alignment) rather than keeping the manual grid, and kept the horizontal-scroll wrapper and explicit chart-width math so wide/many-breakdown funnels still scroll. I left the config off the generic `useChartConfig` refresh-defaults path since `FunnelChart` owns the funnel styling and only consumes `showGrid`/`barCornerRadius` from that set (both already at the intended values). No repo skills were required for this change.

---


sampennington (commented):

Clean, well-scoped migration to the purpose-built `FunnelChart` — reviewers verified the click-data shape flows through `resolveFunnelStepClick` intact, the `series`-only destructure is safe, and the chart-styling defaults match the old `buildFunnelStepsBarConfig`, so no behavior regressions. No blockers. Two small non-blocking nits are inline (constant-valued abstractions that could be plain module consts). One out-of-diff follow-up worth noting: `buildFunnelStepsBarConfig` is now only used by the MCP `FunnelVisualizer` (still on the generic `BarChart`), so the two funnel-bar surfaces have diverged — a later cleanup is to migrate the MCP app onto `FunnelChart` too.

products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/funnelStepsBarTransforms.ts
```diff
@@ -114,15 +114,12 @@ function orderCompareSeriesPreviousFirst(
...
+/** Tooltip behaviour for the grouped funnel bars. A breakdown always puts one series per breakdown
  *  value at each step, so a pinnable tooltip here always covers multiple series —
  *  `resolveClickToNearestSeries` resolves the click to the nearest one and opens its persons modal
  *  directly instead of pinning first. */
-export function withFunnelStepsBarInteraction(baseConfig: BarChartConfig): BarChartConfig {
-    return {
-        ...baseConfig,
-        tooltip: { ...INSIGHT_TOOLTIP_CONFIG, resolveClickToNearestSeries: true },
-    }
+export function funnelStepsBarTooltipConfig(): TooltipConfig {
```
sampennington: 🤖 **sp-ship** · ♻️✨ `[reuse+quality]` · **Suggestion / follow-up**

Now that this takes no args and just returns a fixed object, the function wrapper no longer earns its keep — a plain module const (e.g. `FUNNEL_STEPS_BAR_TOOLTIP_CONFIG`, keeping the docstring) would be clearer and avoid re-allocating per call. Keeping it a function is only defensible for fresh-object-per-call mutation safety. Purely stylistic.

products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/FunnelStepsBarChart.tsx
```diff
@@ -66,7 +59,19 @@ export function FunnelStepsBarChart({
         [steps, getFunnelsColor]
     )
 
-    const config = useChartConfig(() => withFunnelStepsBarInteraction(chartConfig), [])
+    // Display step labels for the chart's bands. The built-in x-axis is hidden (`hideStepLabels`) —
+    // the StepLegend footer row renders the visible labels — but they still feed the tooltip header.
+    const stepLabels = useMemo(() => steps.map((step) => String(step.custom_name ?? step.name ?? '')), [steps])
+
+    const config = useMemo<FunnelChartConfig>(
```
sampennington: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion / follow-up**

This `useMemo` has empty deps and a fully constant body (only `DEFAULT_MARGINS.left` and the tooltip constant), so it adds no value over a module-level `const CHART_CONFIG: FunnelChartConfig = {...}` passed directly.

products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/funnelStepsBarTransforms.ts
sampennington: 🤖 **sp-triage**

Agreed this is stylistic only (Suggestion/follow-up) — resolving as a valid follow-up rather than blocking on it.

products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/FunnelStepsBarChart.tsx
sampennington: 🤖 **sp-triage**

Agreed this is stylistic only (Suggestion/follow-up) — resolving as a valid follow-up rather than blocking on it.

products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/FunnelStepsBarChart.tsx
```diff
@@ -117,13 +122,13 @@ export function FunnelStepsBarChart({
...
+                    <FunnelChart<FunnelStepsBarSeriesMeta>
+                        steps={stepLabels}
                         series={series}
-                        labels={labels}
                         theme={theme}
                         config={config}
                         tooltip={renderTooltip}
-                        onPointClick={showPersonsModal ? onPointClick : undefined}
+                        onStepClick={showPersonsModal ? onStepClick : undefined}
                         onError={handleChartError}
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed real: `FunnelChart` supports a `stepFooter` render-prop for pixel-aligned per-step content, but this component renders `StepLegend` in a separate row with its own width math (`barsWidth`, `stepBandWidthPx`) instead. Fixing this properly means restructuring the render to pass `StepLegend` through `stepFooter` and dropping the duplicated layout math — that is a design/scope decision beyond a tight localized fix, so deferring to a human rather than auto-fixing.

PR #70780, opened 2026-07-14
fix(insights): fill metric card in exported/shared insights

sampennington:
## Problem

Metric insights render weirdly in exported/shared insight images (e.g. subscription screenshots): lots of empty space, and the sparkline sits mid-card at a fixed height instead of filling the card and hugging the bottom the way it does in the web app. Raised in [this Slack thread](https://posthog.slack.com/archives/C0368RPHLQH/p1784043735142029?thread_ts=1784043735.142029&cid=C0368RPHLQH).

## Changes

The Metric display type only stretches its sparkline (`flex-1` + `sparklineFill`) when `inCardView` is true. In `TrendInsight` that flag was derived as `embedded && !inSharedMode`. The exporter renders insights with `inSharedMode`, so `inCardView` was `false`: the sparkline fell back to its fixed 120px height and the card sized to its content, floating at the top of the tall exporter container with empty space below.

- Derive `inCardView` from `embedded` alone, matching how `BoxPlot` and `InsightVizDisplay` already do it. Shared-mode interactivity is gated separately via `showPersonsModal`, and `BoldNumber` (the other consumer of these common props) ignores `inCardView`, so only the Metric layout changes.
- Size the exported metric card as a square (header + chart together, proportioned like a dashboard tile on a 2x2 grid) instead of stretching it to the full exporter height, drop the exporter's viewport-height floor for metric exports, and render metric image exports at a 600px viewport (instead of the default 800px) so the square card comes out at dashboard-tile size.
- Add an exporter Storybook snapshot for the Metric display type to lock in the filled layout.

Added an `Exporter/Trends` Storybook story (`TrendsMetricInsight`) that renders the Metric display type through the exporter, so visual regression captures the filled-card layout. Dependencies were not installed in the agent sandbox, so the frontend formatter, typecheck, and the snapshot run were not executed locally — please rely on CI for those.

## 🤖 Agent context

Investigated from a Slack thread report. Traced the divergence between the dashboard/web render path and the exporter path: both pass `embedded`, differing only on `inSharedMode`, which flowed into `inCardView` in `frontend/src/scenes/trends/Trends.tsx` and disabled the Metric card's fill behavior. Considered adding a separate export-specific flag but chose to align `inCardView` with the existing `embedded`-based convention used elsewhere, since the interactivity concern `inSharedMode` guards is already covered by `showPersonsModal`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0368RPHLQH/p1784043735142029?thread_ts=1784043735.142029&cid=C0368RPHLQH)*

### DIFF
--- a/frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
+++ b/frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
@@ -64,8 +64,6 @@ export function ExportedInsight({
     const trendsDisplay =
         isInsightVizNode(query) && isTrendsQuery(query.source) ? query.source.trendsFilter?.display : undefined
     const isBoxPlot = trendsDisplay === ChartDisplayType.BoxPlot
-    // A metric has no intrinsically tall content, so give it a square footprint (see ExportedInsight.scss)
-    // instead of letting the filling sparkline stretch to the full exporter height.
     const isMetric = trendsDisplay === ChartDisplayType.Metric
     const showLegend =
         legend &&
--- a/frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
+++ b/frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
@@ -64,6 +64,9 @@ export function ExportedInsight({
     const trendsDisplay =
         isInsightVizNode(query) && isTrendsQuery(query.source) ? query.source.trendsFilter?.display : undefined
     const isBoxPlot = trendsDisplay === ChartDisplayType.BoxPlot
+    // A metric has no intrinsically tall content, so give it a square footprint (see ExportedInsight.scss)
+    // instead of letting the filling sparkline stretch to the full exporter height.
+    const isMetric = trendsDisplay === ChartDisplayType.Metric
     const showLegend =
         legend &&
         isInsightVizNode(query) &&
@@ -79,7 +82,7 @@ export function ExportedInsight({
 
     return (
         <BindLogic logic={insightLogic} props={insightLogicProps}>
-            <div className="ExportedInsight">
+            <div className={clsx('ExportedInsight', isMetric && 'ExportedInsight--metric')}>
                 {!noHeader && (
                     <div className="ExportedInsight__header">
                         <div>
--- a/frontend/src/exporter/Exporter.tsx
+++ b/frontend/src/exporter/Exporter.tsx
@@ -18,6 +18,8 @@ import { teamLogic } from 'scenes/teamLogic'
 
 import { ExporterLogin } from '~/exporter/ExporterLogin'
 import { ExportType, ExportedData } from '~/exporter/types'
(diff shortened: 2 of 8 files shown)


frontend/src/exporter/Exporter.tsx
```diff
@@ -109,6 +119,7 @@ export function Exporter(props: ExportedData): JSX.Element {
             <div
                 className={clsx('Exporter', {
                     'Exporter--insight': !!insight,
+                    'Exporter--metric': isMetricInsight,
```
sampennington: theres a clear pattern here, lets follow it with !!metric

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
```diff
@@ -64,6 +64,9 @@ export function ExportedInsight({
     const trendsDisplay =
         isInsightVizNode(query) && isTrendsQuery(query.source) ? query.source.trendsFilter?.display : undefined
     const isBoxPlot = trendsDisplay === ChartDisplayType.BoxPlot
+    // A metric has no intrinsically tall content, so give it a square footprint (see ExportedInsight.scss)
+    // instead of letting the filling sparkline stretch to the full exporter height.
```
sampennington: dont need this comment

frontend/src/exporter/Exporter.tsx
sampennington: Done in 6a19f85 — `metric` now holds the insight when it is a Metric display, so the clsx entry reads `!!metric` like the others.

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
sampennington: Removed in 6a19f85.

products/exports/backend/tasks/image_exporter.py
```diff
@@ -192,10 +192,21 @@ def _export_to_png(
...
             funnel_layout = funnels_filter.get("layout")
             is_left_to_right_funnel = is_funnel and (funnel_layout is None or funnel_layout == FunnelLayout.VERTICAL)
+            # A metric renders as a square card whose height derives from the viewport width, so a
+            # narrower viewport keeps it at dashboard-tile size instead of an 800px-tall square.
+            trends_filter = source.get("trendsFilter") or {}
+            is_metric = source.get("kind") == NodeKind.TRENDS_QUERY and (
+                trends_filter.get("display") == ChartDisplayType.METRIC
+            )
             # Set initial window size large enough for wide content like left-to-right funnels with many steps
             # Small funnels will be constrained later.
```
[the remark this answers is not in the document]
sampennington: Good catch — fixed in 34c0332. The backend check now requires the `InsightVizNode` wrapper (mirroring the frontend condition), so a bare `TrendsQuery` keeps the default 800px viewport and full-height layout on both sides.

PR #70791, opened 2026-07-14
feat(data-warehouse): release 166 implemented warehouse sources

Gilbert09:
## Problem

166 warehouse import sources are fully implemented — real `get_schemas`, a client, endpoint config, and passing tests — but still ship hidden behind `unreleasedSource=True`. That flag filters a connector out of the catalog entirely (see `DataWarehouseValuePropVariant.tsx`, `InlineSourceSetup.tsx`, and the "coming soon / Notify me" path in `nonHogFunctionTemplatesLogic.tsx`), so all this finished work is invisible to users.

The flag is doing double duty across the source tree: "still a stub" and "built but not yet released". This PR clears it for the built-and-tested set so those connectors become visible and connectable.

## Changes

- Remove `unreleasedSource=True` from the 166 sources detected as fully implemented (has `get_schemas`, >60 lines, no `NotImplementedError`, and its own tests).
- Drop the 103 per-source tests that asserted `config.unreleasedSource is True` (plus the now-stale gating comments), and rename three test methods that described the hidden state (`aviator`, `mixmax`) / drop a stale comment (`bigmailer`).
- No source logic, schema, or type changes. `unreleasedSource` stays on the `SourceConfig` type and on the ~375 genuine stubs that are still unfinished.

> [!NOTE]
> This is a release action, not just cleanup. Many of these sources were deliberately gated — 103 had a test locking in the hidden state, and 15 carried a comment like "stays hidden until it's been exercised against the live API". Removing the flag exposes them to users on merge + deploy. Reviewers should confirm none of these need live-API validation before they go visible; `releaseStatus` (e.g. `ALPHA`) is untouched, so alpha labelling still applies.

Automated only (I'm Claude, agent-assisted — no manual UI testing done):

- Ran the source-config test across all 166 changed sources: `pytest <166 test_*_source.py>` → **3086 passed**.
- Ran the full `vercel` suite to confirm the previously-failing `unreleasedSource is True` assertion is resolved and the rest still pass.
- `ruff check` + `ruff format` clean on all 269 changed files; `hogli ci:preflight` reports 0 failures.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Tom directed this while auditing scaffolded warehouse sources. The thread started from a mistaken premise — that `unreleasedSource=True` marks unimplemented stubs — which turned out to be wrong: it means "not GA yet", and 166 sources are fully built but still flagged. I (Claude) invoked `/implementing-warehouse-sources`, which documents that removing the flag is the intended "finish the source" step.

Detection used a code-shape heuristic (`get_schemas` present, >60 lines, no `NotImplementedError`, own tests) rather than the flag itself, since the flag was the unreliable signal. I flagged mid-way that 103 of the 166 have tests deliberately asserting the hidden state (15 with an explicit "hidden until live API" reason); Tom chose to release all 166 regardless, relying on this PR's review as the gate. The 63 without such an assertion look like genuinely leftover flags; the other 103 were intentional and are the ones worth a closer look before merge.


danielcarletti (approved):

danielcarletti (approved):
LGTM

PR #70481, opened 2026-07-13
feat(batch-exports): authenticate Snowflake exports via integrations

rossgray:
## Problem

Snowflake batch exports store all their credentials (`user`, `password`, `private_key`, `private_key_passphrase`) inline in each destination's encrypted config. That means credentials are duplicated per export and can't be reused, rotated, or managed independently. Databricks, Azure Blob, Postgres, and the S3 family already authenticate via the shared `Integration` model. This wires Snowflake up to do the same.

Second PR in the Snowflake migration. The first (#69642) added the `snowflake` Integration kind; this one makes batch exports actually use it.

## Changes

- **Runtime resolution (`snowflake_batch_export.py`):** when a destination has an `integration_id`, the Temporal activity resolves account, user, authentication type and the credential material from the linked Integration at run time. Otherwise it falls back to inline config (legacy path). A missing or wrong-kind integration raises a non-retryable error.
- **Serializer (`api/batch_export.py`):** typed `Snowflake` config and request serializers, `integration_id` accepted, and the integration kind validated against the destination in `validate_destination`. An export that already uses an integration can't drop back to inline credentials (matches the S3 family behaviour).
- **`service.py`:** `account` and `user` become optional on the input dataclass, since integration-backed exports resolve them at run time.

> [!NOTE]
> Inline credentials still work at runtime so existing exports keep running. It is also still possible to create Snowflake batch exports without integrations via the API for now. Enforcing integrations and backfilling existing exports is a later PR.

## How did you test this code?

Automated tests only:

- **API** (`test_create_snowflake.py`, `test_update_snowflake.py`, a shared case in `test_create.py`)
- **Temporal** (`snowflake/test_activity.py`, new): un-gated unit tests for the integration resolver — a missing integration and a wrong-kind integration each raise the expected non-retryable error. Plus a live-Snowflake activity test (`test_activity_e2e.py`) that resolves credentials from an integration, gated behind `SKIP_IF_MISSING_REQUIRED_ENV_VARS`.

## Automatic notifications

- [ ] Publish to changelog? _Not yet_
- [ ] Alert Sales and Marketing teams?

## Docs update

No user-facing docs — backend only. The frontend create/edit flow comes with the next PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Built with PostHog Code (Claude). Skills invoked: `/improving-drf-endpoints` (the config/request serializers and `validate_destination` branch) and `/writing-tests` (the API and temporal test additions).

[...]

PR #70794, opened 2026-07-14
feat(batch-exports): add snowflake integration support to frontend

rossgray:
## Problem

The Snowflake backend support for Integrations is already merged ([#70481](https://github.com/PostHog/posthog/pull/70481)); this is the frontend to match, mirroring the S3 work in [#66359](https://github.com/PostHog/posthog/pull/66359).

## Changes

- Add `snowflake` as an integration kind (`INTEGRATION_KINDS`, `IntegrationKind`) with its icon, and an `integration` field on `BatchExportServiceSnowflake`.
- New `SnowflakeSetupModal` + logic to create a snowflake Integration (name, account, user, and either key pair or password auth). Key pair is the default.
- Wire the Snowflake destination form to the integration picker for new exports. Exports created before integrations existed keep their inline credential fields (grandfathered), with a banner noting they'll be migrated automatically.

Automated tests:

- `batchExportConfigFormLogic.test.ts` (67 cases pass). Updated the Snowflake cases to assert the new integration flow: required fields are now `integration_id` + `database`/`warehouse`/`schema`/`table_name`, the create payload carries `destination.integration` with no leaked credentials, and `authentication_type` no longer defaults into the config. The existing password/keypair round-trip fixtures already guard grandfathered inline exports.

Manual testing via the local stack:

- tested creation of a new Snowflake batch export using an Integration
- tested an existing Snowflake batch export still works without an Integration (legacy inline credentials still work)

### New modal

image

### Warning for legacy exports

image

## Automatic notifications

- [x] Publish to changelog?
- [ ] Alert Sales and Marketing teams? No

## Docs update

We should update the docs for Snowflake batch exports to mention the new flow using integrations, as was done for S3 recently.

## 🤖 Agent context

Built with Claude (PostHog Code). Ross directed the work: scope it as a frontend mirror of the S3 integration PR, then a few refinements (key pair as the default auth, modal width tuning for pasting private keys).

Notes for reviewers:

- Structurally this follows `postgres.tsx` more than the S3 files. Snowflake defines its `Fields` inline like Postgres, whereas the three S3 destinations share `S3FamilyFields`.
- Dropped the `authentication_type: 'password'` default on the destination. New exports use the picker and never render the inline auth field, so a default would only leak the key into `config`, which the backend rejects as unknown.
- Skills invoked: `/writing-tests` (before touching the test file) and `/user-skills:create-pr` (this PR).
- The kea `*LogicType.ts` files are gitignored and regenerated in CI, so they're not in the diff.

---

### DIFF
--- a/frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
+++ b/frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
@@ -1,7 +1,7 @@
 import { useActions, useValues } from 'kea'
 import { Form } from 'kea-forms'
 
-import { LemonButton, LemonInput, LemonModal, LemonSelect, LemonTextArea } from '@posthog/lemon-ui'
+import { LemonButton, LemonFileInput, LemonInput, LemonModal, LemonSelect, LemonTextArea } from '@posthog/lemon-ui'
 
 import { LemonField } from 'lib/lemon-ui/LemonField'
 
@@ -53,6 +53,13 @@ export const SnowflakeSetupModal = (props: SnowflakeSetupModalLogicProps): JSX.E
                 </LemonField>
                 {snowflakeIntegration.authentication_type === 'keypair' ? (
                     <>
+                        <LemonField
+                            name="private_key_file"
+                            label="Private key file"
+                            help="Upload the key file you generated, or paste its contents below."
+                        >
+                            <LemonFileInput accept=".p8,.pem,.key" multiple={false} />
+                        </LemonField>
                         <LemonField name="private_key" label="Private key">
                             <LemonTextArea className="ph-ignore-input" placeholder="my-private-key" minRows={8} />
                         </LemonField>
--- a/frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
+++ b/frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
@@ -1,0 +1,71 @@
+import { useActions, useValues } from 'kea'
+import { Form } from 'kea-forms'
+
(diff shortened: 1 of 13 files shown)


frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+                            { value: 'keypair', label: 'Key pair' },
+                            { value: 'password', label: 'Password' },
+                        ]}
+                    />
+                </LemonField>
+                {snowflakeIntegration.authentication_type === 'keypair' ? (
+                    <>
+                        <LemonField name="private_key" label="Private key">
+                            <LemonTextArea className="ph-ignore-input" placeholder="my-private-key" minRows={8} />
+                        </LemonField>
```
tomasfarias, line 65: I think we should support file inputs here. The workflow the snowflake docs describe is to generate a key file: https://docs.snowflake.com/en/user-guide/key-pair-auth. So, it makes sense that if you are following that workflow you can just upload the key file here.

tomasfarias (commented):
What do you think about supporting file inputs?

frontend/src/scenes/integrations/snowflake/SnowflakeSetupModal.tsx
rossgray: yeah, good idea. Will see if I can add that

rossgray: have added that now, and updated the screenshot

tomasfarias (approved):
dope, thanks

Issue #66380, opened 2026-06-26
Workflows: live editing (draft → test → publish) and revision history


mayteio:
# Workflows: live editing (draft → test → publish) and revision history

**TLDR:** follow-live stays: live edits reach people mid-flow, which is how users use the product today and how the market works. Ship the agent unblock first - **draft → test → publish** on the existing `draft` column, zero migrations - then harden follow-live into a documented contract, then **skip-forward past deleted steps** so restructuring never strands people mid-flow, then **revision history** for rollback. All behind the `workflows-revisions` flag: a two-way door.

## The build

**Status (2026-07-27):** `workflows-revisions` and `workflows-timing-reschedule` are at **100% of organizations**. The sweep also covers deleted timing steps and condition edits (#73925), clearing the workflows-side readiness gates for the poll removal (#72541). Remaining: merge the user docs page (PostHog/posthog.com#18506); builder UI parity is tracked as a follow-up below.

### Draft → test → publish, MCP-first

The point of this work: agents get a safe place to iterate and test before anything touches live (and what the support tickets ask for). PATCH on an active workflow routes to the `draft` column; `publish` copies draft → live in one transaction (optimistic guard, strict validation, bytecode compile); `discard` clears it. MCP drops the active-edit rejection, gains `workflows-publish` (which echoes "N runs in flight" so nothing publishes blind); the `building-workflows` skill is rewritten around the cycle. Drafting changes nothing running; publish is the same live edit the UI already does daily, with more guardrails, behind the flag. Manually tested end to end over local MCP - see the test table on #70823. The `workflows-revisions` flag exists in prod, gated to PostHog + the two requesting orgs.

**PRs:**

- #70796
- #70823
- #70860
- #72576 - draft-cycle papercuts found live-testing over MCP: `add_action` wires its `edges` in the same op, `output_variable` coerced to the worker shape

### Trustworthy follow-live

The fix list below ("What's missing"), plus the docs. Verification of what publish inherits; parallelizable with the draft cycle, except the timing reschedule.

**PRs:**

- #71057
- #71060
- #71500
- #71062
- #71290
- #71473
- #71293
- #71298
- #73925 - deleted timing steps and condition edits join the reschedule sweep (workflows-side readiness for the poll removal, #72541)

**Docs:**

- PostHog/posthog.com#18506 - user docs: per-edit live-edit semantics page (`/docs/workflows/editing-live-workflows`)

**Infra & observability:**

[...]

PR #70796, opened 2026-07-14
feat(workflows): expose an in-flight run count for hog flows

mayteio:
## Problem

Live edits to an active workflow reach runs already in flight - parked on waits and delays or mid-execution - but nothing in the API layer can say how many runs that is.
For the workflows draft → test → publish cycle (step 1 of #66380), publish needs to echo a crude "N runs in flight" so nothing publishes blind.
In-flight runs live in the cyclotron jobs table, which only the node CDP service can reach, so Django has no way to ask today.

## Changes

- `countInFlightJobs(teamId, functionId)` on `CyclotronV2Manager` and the `CyclotronV2JobProducer` interface: one indexed count of `available`/`running` jobs by team and function (`idx_queue_function_id` covers the lookup).
- `GET /api/projects/:team_id/hog_flows/:id/in_flight_count` on the CDP API, returning `{count}`.
  Guards mirror the rerun endpoint: 503 when `CYCLOTRON_NODE_DATABASE_URL` is unset, 404 for unknown team or workflow (including cross-team lookups).
- `get_hog_flow_in_flight_count()` in `posthog/plugins/plugin_server_api.py`, same shape as its sibling client helpers.

No callers yet - the publish endpoint that consumes this lands in a follow-up PR.

- `cyclotron-v2.test.ts`: new case proving the count includes available and running jobs and excludes terminal statuses, other functions, and other teams - catches anyone loosening the status or scoping filters.
- `cdp-api.test.ts`: three route cases - 200 with the count and correct `(team_id, function_id)` args (catches a param swap), 404 for an unknown workflow, 503 when the producer isn't configured (locks in fail-on-request rather than fail-at-boot for local dev).

Both suites run green locally.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None - internal service endpoint, no user-facing behavior.

## 🤖 Agent context

Built with Claude Code as PR 1 of 3 for the draft → test → publish cycle in #66380.
Skills invoked: /writing-tests.
Tests were written first against the real cyclotron test database and the supertest app.
The count method went on the existing `CyclotronV2JobProducer` surface (already injected into `CdpApi` for batch dispatch) rather than a new pool or a dedicated manager - the interface is documented as the place API entrypoints grow producer-side needs.

### DIFF
--- a/nodejs/src/cdp/cdp-api.test.ts
+++ b/nodejs/src/cdp/cdp-api.test.ts
@@ -15,7 +15,7 @@ import { UUIDT } from '~/common/utils/utils'
 
 import { createCdpConsumerDeps } from '../../tests/helpers/cdp'
 import { forSnapshot } from '../../tests/helpers/snapshots'
-import { getFirstTeam, resetTestDatabase } from '../../tests/helpers/sql'
+import { createTeam, getFirstTeam, resetTestDatabase } from '../../tests/helpers/sql'
 import { Hub, Team } from '../types'
 import { FixtureHogFlowBuilder } from './_tests/builders/hogflow.builder'
 import { HOG_EXAMPLES, HOG_FILTERS_EXAMPLES, HOG_INPUTS_EXAMPLES } from './_tests/examples'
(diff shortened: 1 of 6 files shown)


nodejs/src/cdp/cdp-api.test.ts
```diff
@@ -1168,6 +1171,67 @@ describe('CDP API', () => {
...
+        it('returns the in-flight job count for a workflow', async () => {
+            const res = await supertest(app).get(
+                `/api/projects/${countHogFlow.team_id}/hog_flows/${countHogFlow.id}/in_flight_count`
+            )
+
+            expect(res.status).toEqual(200)
+            expect(res.body).toEqual({ count: 3 })
+            expect(mockCountInFlightJobs).toHaveBeenCalledWith(countHogFlow.team_id, countHogFlow.id)
+        })
+
```
dmarchuk, line 1214: Would it make sense to add a cross-team case here? 🤔 The `team_id` guard is the real isolation boundary but we only cover missing-workflow and 503. I think requesting another team's workflow and asserting 404 would also make sense here

dmarchuk (approved):
Nice! 🙌

nodejs/src/cdp/cdp-api.test.ts
mayteio: Good call — added in 65183ca: requests the flow under a second team in the same org, asserts the 404 and that the count query is never reached.

PR #70797, opened 2026-07-14
feat(personhog): e2e test harness for ci

z0br0wn:
## Problem
  
The personhog leader write path (router → leader → changelog → writer → Postgres) has per-crate unit and integration tests, but nothing exercises the full five-service stack under real traffic, and nothing at all exercises it under failure: leader crashes, graceful drains, zombie processes, writer lag, coordinator loss. The property the whole design hangs on is simple to state and was previously unverified end to end: **every write acked by the leader path must be visible afterwards** – in strong reads immediately, and in Postgres at exactly the acked version once the writer drains.

## Changes

New `rust/personhog-test-harness` crate (reviving the earlier personhog-cannon draft, #55581) plus a CI job that runs it. The `gate` command spawns a fully isolated local stack against the docker-compose Postgres/Kafka/etcd, seeds persons, drives traffic, disturbs the stack mid-run, then verifies and tears down. Isolation is by port range, etcd prefix, and a per-run changelog topic, so it runs alongside a dev stack.

  ```mermaid
  flowchart LR
      H{{test harness}} -->|spawn + chaos| R
      H -->|seed persons SQL| PG
      H -->|blast writes + probe reads| R
      subgraph S[isolated stack]
          R[leader-mode routers] --> L[leaders]
          L --> K[(changelog topic)]
          K --> W[writer]
          E[(etcd)] --- R
          E --- L
      end
      W --> PG[(posthog_persons)]
      H -->|verify strong reads + rows/versions| PG
      class R,L,W phBlue;
      class H phYellow;
      class K,E,PG phGray;
  ```

Three verification layers share one acked-write journal (every ack recorded under a globally unique property key):

  - **Probers** assert read-your-write recency live, during chaos windows, via write-then-strong-read cycles
  - **End-of-run strong reads** assert every acked key is present and versions never regressed
  - **Postgres quiesce** asserts every acked key landed in the row at exactly the acked version

The gate exits non-zero on any violation, and also on zero acked writes (a broken-but-reachable stack must not pass vacuously).

Chaos scenarios, all validated green locally:

  | scenario | flags | validates |
  |---|---|---|
  | leader crash | `--kill-after` (+ fast lease revoke) | reassignment + warming, stash absorption |
  | crash discovered by TTL | `--kill-fast false --leader-lease-ttl` | coordinator blindness window, recovery |
  | graceful drain | `--shutdown-after` | freeze/stash/drain/warm choreography |
  | scale-up | `--scale-up-after` | rebalance handoffs to a new pod |
  | zombie | `--zombie-after` (SIGSTOP + lease revoke, later SIGCONT) | old owner waking with stale cache corrupts nothing |
  | same-identity restart | `--restart-after` | re-registration + convergence on owned partitions |
  | writer crash / pause | `--writer-crash-after`, `--writer-pause-after` | at-least-once redelivery under the version guard, lag tolerance |
  | coordinator kill | `--routers 2 --router-kill-after` | election failover, handoffs under the new coordinator |
  | mid-handoff target kill | `--kill-handoff-target` | coordinator re-planning when the warming pod dies |
  | eviction pressure | `--cache-capacity` | expected red – reproduces the known eviction-under-lag hazard |

One production change rides along: the router's leader address resolver now dials `host:port` pod names as-is (bare names keep resolving via DNS on the fleet-wide leader port), which is what lets N local leaders listen on distinct ports. K8s behavior is unchanged.

The `personhog-gate` CI job boots only db/kafka/etcd from compose, migrates the persons DB with sqlx directly, and runs five gate configurations (plain + four chaos composites). It is a new job name, so it is not a required check while it soaks. On failure it uploads every service's log as an artifact.

  > [!NOTE]
  > The harness surfaced three real leader-path defects, documented with signatures and fix directions in the crate README: cache eviction under writer lag loses acked writes (gate goes red under `--cache-capacity` pressure), the leader's unordered lifecycle shutdown black-holes a draining pod's partitions (~1%
  failed writes on drain runs, gate stays green since nothing acked is lost), and a crashed coordinator blocks all handoffs for the 10–20s election window. The eviction run is intentionally excluded from CI until fixed.

- 7 unit tests on the harness's own verification logic, run in CI inside the gate job. They cover the one surface e2e runs structurally cannot: a false-negative verifier looks identical to a healthy stack. Decision tables for property verification, journal version tracking and ack-regression flagging, strong-read version checks, the SQL-interpolation table-name guard, and chaos timeline ordering.
- Every chaos scenario in the table above was executed locally against a spawned stack, all green (the coordinator-failover composite three times consecutively after making the election deterministic). The eviction run reproduces the known hazard on demand.
- Both CI gate invocations were rehearsed locally at exact CI parameters before wiring the job.
- The eviction hazard aside, runs sustain roughly 300–700 writes/sec (debug builds) with zero violations across thousands of acked writes per run.
- `cargo clippy -D warnings`, `cargo fmt`, and the affected-services integration test (which the new crate registers with) all pass.

## Docs update

The crate README documents usage, every chaos flag, and the known-defects catalog. Internal dev tooling – suggest the `skip-inkeep-docs` label.
 ## 🤖 Agent context

Built with Claude Code. Skills invoked along the way: `/writing-tests` (gating which unit tests earn their place). Key decisions, in rough order: revived the personhog-cannon draft rather than starting fresh, but dropped its `comfy-table`/`owo-colors` dependencies (new to the workspace) for plain formatting.
Chose the `host:port` pod-name resolver tweak over adding an advertise-address field to the registration schema (smaller, backward compatible). Seeding is direct SQL for now because the create RPC's future is still being settled – `src/seed.rs` is the documented swap seam. The read-your-write probers were added late and immediately exposed that the coordinator-kill scenario was nondeterministic (initial election winner is a race), which led to the deterministic bring-up (first router must win the election) plus election-lease revocation on kill, and to documenting the coordinator failover window as a third known defect. A toxiproxy-based network-fault layer (asymmetric partitions, etcd watch latency) was scoped and deliberately deferred.

Two notes on it: the "three defects" callout is written code-facing (mechanisms and fix directions from this repo only, no operational numbers beyond harness-scale measurements), per the public-repo rules. And per the template's draft-by-default rule, when you're ready I'd push and open it as a draft with you assigned as DRI.


gantoine (commented):
CI change looks good

.github/workflows/ci-rust.yml
```diff
@@ -470,6 +470,143 @@ jobs:
...
+    # write acked by the leader path is visible afterwards — in strong reads
+    # and in Postgres at the acked version — including under chaos: leader
+    # kill (lease-revoked and TTL-expiry), scale-up, graceful drain, zombie
+    # (SIGSTOP/SIGCONT), writer crash and lag, same-identity restart,
+    # coordinator failover, and a mid-handoff target kill. Not a required
+    # check while it soaks.
+    personhog-gate:
+        name: PersonHog e2e gate
+        needs: changes
+        if: needs.changes.outputs.rust == 'true'
```
nickbest-ph: should probably use the affected-services crate to be a bit more deliberate about when this runs.

i think the cleanest way to do this is by using `rust/affected-services/determinator-rules.toml` to mark this crate as changing like so:
```
[[package-rule]]
on-affected = ["personhog-replica", "personhog-router", "personhog-leader", "personhog-writer", etc.]
mark-changed = ["personhog-test-harness"]
```

and then you can just have this be
```
needs: [changes, affected]
if: needs.changes.outputs.rust == 'true' && contains(needs.affected.outputs.affected_crates, 'personhog-test-harness')
```

nickbest-ph:
Great! Lgtm. This is gonna be awesome to have.

Might be worth considering AGENT context and a skill somewhere that tells agents when and how to add scenarios here after it fixes protocol bugs/regressions in the service.
wdyt?

nickbest-ph (approved):

PR #70800, opened 2026-07-14
feat(ingestion): add common ingestion pipeline builder prototype

pl:
## Problem

Every Kafka-fed ingestion pipeline in nodejs (analytics, AI, error tracking, heatmaps, client warnings, session replay) hand-assembles the same skeleton: beforeBatch hooks, messageAware, header parsing, pre-parse restrictions, body parse + team resolution, lifting the team into context, a teamAware processing block, ingestion warning handling, result handling, side effect handling, afterBatch flush hooks. Only the middle differs per product. The boilerplate is duplicated in every pipeline, the invariants (warnings before results, side effects last) rely on convention, and the deep callback nesting makes it hard to see at a glance what a pipeline actually does.

## Changes

Adds a common ingestion pipeline builder (`nodejs/src/ingestion/common/common-ingestion-pipeline.ts`) that owns the skeleton and reads as a flat recipe. Phase markers carry the structure; there is no user-side nesting except batch hooks and grouping:

```ts
newCommonIngestionPipeline<TInput, TContext>({ teamManager, outputs, promiseScheduler, concurrentBatches: 1 })
    .beforeBatch((b) => b.pipe(createEventFiltersBatchAppMetricsBeforeBatchStep(outputs)))
    .parseHeaders()
    .pipe(createAllowEventsStep(['$$heatmap']))
    .pipe(createApplyBasicEventRestrictionsStep(eventIngestionRestrictionManager))
    .parseMessage()
    .resolveTeam()
    .pipe(createValidateHistoricalMigrationStep())
    .pipe(createValidateEventMetadataStep())
    .pipe(createApplyEventFiltersStep(eventFilterManager))
    .pipe(createDropOldEventsStep())
    .gather()
    .pipeChunk(createApplyCookielessProcessingStep(cookielessManager))
    .pipe(createCheckHeatmapOptInStep())
    // ...
    .afterBatch((b) => b.pipe(createFlushEventFiltersBatchAppMetricsStep()))
    .build()
```

Semantics:

- Consecutive `.pipe()` calls coalesce into one sequential per-element block; any chunk-level call (`.pipeChunk`, `.gather`, `.concurrentlyPerGroup`, `.compose`) closes the block. Block boundaries fall out of where the chunk operations sit, which is how the hand-written pipelines were already shaped.
- `.parseHeaders()` / `.parseMessage()` / `.resolveTeam()` pipe the shared steps. `.resolveTeam()` also lifts the team into the element context and opens the team-aware scope: everything after it runs inside `handleIngestionWarnings`. It takes an optional `wrap` to decorate the resolution step (error tracking wraps it in topHog metrics).
- `.build()` wraps the whole thing in `messageAware` and bundles `handleResults`, `handleSideEffects`, and the batching hooks. The built pipeline owns *all* of its side effects: the beforeBatch/afterBatch hook pipelines are wrapped in side-effect handling too (via a new single-item `SideEffectHandlingProcessor` and `PipelineBuilder.handleSideEffects`, mirroring the chunk-level one), so `BatchResult.sideEffects` always comes out empty and drivers only drain results. An `awaitSideEffects` config flag (default false, i.e. schedule on the promise scheduler) applies to element results and hooks alike.
- Stage order is enforced by the type system (body steps don't typecheck before `.parseMessage()`), and redirect kinds are declared up front via the `ROut` type parameter so the outputs must cover every redirect a step can produce.
- `.concurrentlyPerGroup` mirrors the framework method (within-group ordering stays visible at the call site), and `.compose(fn)` is the escape hatch for existing subpipeline functions like `createPostTeamPreprocessingSubpipeline`, whose generics are widened to accept mid-chain builders.
- Internally each stage composes continuations over the chunk builder — no casts and no `any`; the coalescing block's start type stays existential inside closures.

Five pipelines are rewired onto the builder: client warnings, heatmaps, joined analytics (keeps its per-distinct-id `concurrentlyPerGroup` and post-team subpipeline), AI (keeps its topHog metric wrappers and per-step retries), and error tracking. Two mechanical enablers rode along: the schema-enforcement toggle moved into `createValidateEventSchemaStep` (a no-op step when disabled beats conditional chain composition), and the post-team subpipeline generics were widened.

> [!NOTE]
> Error tracking previously built a bare chunk pipeline driven through `ChunkPipelineUnwrapper`; it is now a `BatchingPipeline` like the other consumers, with passthrough batch hooks and `concurrentBatches: 1`, and its custom team-lift `filterMap` became a plain `messageBytes` step. Its `createErrorTrackingPipeline` / `runErrorTrackingPipeline` signatures are unchanged, so the existing suites verify the conversion.

> [!NOTE]
> Sequential-block boundaries shift slightly versus the hand-written versions (e.g. post-team-resolution validation now runs inside the team-aware scope, so its warnings route per-team even for dropped elements). Per-element step order is unchanged everywhere; the end-to-end suites pass unchanged, including all 22 ingestion-consumer snapshots.

Session replay is the one pipeline left (custom team resolution via `TeamService` rather than `TeamManager`); a doc-test chapter for the builder is also still pending.

- `pnpm build` (tsc across nodejs) and eslint on the touched files
- Existing suites pass unchanged, so the rewired pipelines are behaviorally identical: `src/ingestion/ingestion-consumer.test.ts` (57 tests, 22 snapshots, end-to-end joined pipeline), `src/ingestion/pipelines/ai/` (806 tests), `src/ingestion/pipelines/errortracking/` (146 tests, including the end-to-end pipeline and consumer suites, driven through the unchanged `createErrorTrackingPipeline`/`runErrorTrackingPipeline` API), `src/ingestion/pipelines/heatmaps/` (42 tests), `src/ingestion/pipelines/clientwarnings/`
- One new test case in `validate-event-schema.test.ts`: a step built with `enabled: false` passes through an event that violates an enforced schema — guards the flag plumbing, which no existing test covered; the enabled-path cases already existed
- Otherwise no new tests: the builder is exercised end-to-end by the existing pipeline suites; a dedicated doc-test chapter is deferred until the API settles

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None yet, prototype. The framework's doc-test chapters (`nodejs/src/ingestion/framework/docs/`) get a builder chapter if this graduates.

## 🤖 Agent context

Built with Claude Code. Skills used: `ingestion-pipeline-doctor-nodejs`, `writing-tests`. The API went through three shapes on this branch: staged callbacks per phase (preParse/resolveTeam/perTeam), then a collapse of the post-resolution stage, then the current flat recipe — the flat form was chosen because the linear spine is the point of the template and the callback nesting added no information. Key implementation decisions: sequential-block coalescing with the block's start type kept existential inside closures (cast-free), `ROut` fixed at the factory so redirect coverage is checked where outputs are supplied, the post-team subpipeline generics widened instead of adding a special builder hook, and error tracking converted to a `BatchingPipeline` (rather than giving the builder a second, unwrapped build path) to keep all consumers on one pipeline shape.

### DIFF
--- a/nodejs/src/ingestion/common/common-ingestion-pipeline.ts
+++ b/nodejs/src/ingestion/common/common-ingestion-pipeline.ts
@@ -113,88 +113,75 @@ export type AfterBatchCallback<TOutput, TContext, CBatch, ROut extends string> =
 >
 
 /**
- * The accumulated content of the messageAware block before `.resolveTeam()`:
- * a transform over the batch sub-pipeline builder. Each call extends the
- * previous transform, so the skeleton is composed exactly once, at `.build()`.
+ * A phase's accumulated content: a transform over the phase's chunk builder,
+ * from the phase's start element type `TStart` to the current element type,
+ * under the element context `C`. Each call extends the previous transform,
+ * so the skeleton is composed exactly once, at `.build()`.
  */
-type SubpipelineTransform<TInput, TContext, CBatch, TOut, ROut extends string> = (
-    builder: ChunkPipelineBuilder<
-        BatchElement<TInput, CBatch>,
-        BatchElement<TInput, CBatch>,
-        BatchContext<TContext>,
-        BatchContext<TContext>
-    >
-) => ChunkPipelineBuilder<BatchElement<TInput, CBatch>, TOut, BatchContext<TContext>, BatchContext<TContext>, ROut>
+type ChainTransform<TStart, TCurrent, C, ROut extends string> = (
+    builder: ChunkPipelineBuilder<TStart, TStart, C, C>
+) => ChunkPipelineBuilder<TStart, TCurrent, C, C, ROut>
 
-/** The accumulated team-aware content after `.resolveTeam()`, nested inside the team lift at `.build()`. */
-type TeamSubpipelineTransform<TPost, TContext, TCurrent, ROut extends string> = (
-    builder: ChunkPipelineBuilder<TPost, TPost, TeamAwareContext<TContext>, TeamAwareContext<TContext>>
-) => ChunkPipelineBuilder<TPost, TCurrent, TeamAwareContext<TContext>, TeamAwareContext<TContext>, ROut>
+/** The messageAware block before `.resolveTeam()`: runs under the batch context. */
+type SubpipelineTransform<TInput, TContext, CBatch, TOut, ROut extends string> = ChainTransform<
+    BatchElement<TInput, CBatch>,
+    TOut,
+    BatchContext<TContext>,
+    ROut
+>
 
 /**
  * Coalesces consecutive per-element steps into one sequential block. `extend`
  * appends a step to the open block (or opens one); `build` closes it into the
  * committed chunk-level transform. The block's start type stays existential —
  * it lives only in the closures — so stages don't carry it as a type parameter.
+ *
+ * Both builder phases share this mechanism and differ only in instantiation:
+ * the pre-team phase runs under the batch context, the team-aware phase under
+ * the team-lifted context (see the aliases below).
  */
-interface PreTeamChain<TInput, TContext, ROut extends string, CBatch, TCurrent> {
-    build(): SubpipelineTransform<TInput, TContext, CBatch, TCurrent, ROut>
+interface Chain<TStart, C, ROut extends string, TCurrent> {
+    build(): ChainTransform<TStart, TCurrent, C, ROut>
     extend<U, R2 extends ROut>(
         step: ProcessingStep<TCurrent, U, R2>,
         options?: { retry?: RetryOptions }
-    ): PreTeamChain<TInput, TContext, ROut, CBatch, U>
+    ): Chain<TStart, C, ROut, U>
 }
 
-function pendingPreTeamChain<TInput, TContext, ROut extends string, CBatch, TBlock, TCurrent>(
-    committed: SubpipelineTransform<TInput, TContext, CBatch, TBlock, ROut>,
-    pending: (
-        start: StartPipelineBuilder<TBlock, BatchContext<TContext>>
-    ) => PipelineBuilder<TBlock, TCurrent, BatchContext<TContext>, ROut>
-): PreTeamChain<TInput, TContext, ROut, CBatch, TCurrent> {
+function pendingChain<TStart, C, ROut extends string, TBlock, TCurrent>(
+    committed: ChainTransform<TStart, TBlock, C, ROut>,
+    pending: (start: StartPipelineBuilder<TBlock, C>) => PipelineBuilder<TBlock, TCurrent, C, ROut>
+): Chain<TStart, C, ROut, TCurrent> {
     return {
         build: () => (builder) => committed(builder).sequentially(pending),
-        extend: (step, options) => pendingPreTeamChain(committed, (start) => pending(start).pipe(step, options)),
+        extend: (step, options) => pendingChain(committed, (start) => pending(start).pipe(step, options)),
     }
 }
 
-function committedPreTeamChain<TInput, TContext, ROut extends string, CBatch, TCurrent>(
-    committed: SubpipelineTransform<TInput, TContext, CBatch, TCurrent, ROut>
-): PreTeamChain<TInput, TContext, ROut, CBatch, TCurrent> {
+function committedChain<TStart, C, ROut extends string, TCurrent>(
+    committed: ChainTransform<TStart, TCurrent, C, ROut>
+): Chain<TStart, C, ROut, TCurrent> {
     return {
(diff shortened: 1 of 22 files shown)


nodejs/src/ingestion/common/common-ingestion-pipeline.ts
```diff
@@ -0,0 +1,489 @@
+import { Message } from 'node-rdkafka'
```
jose-sequeira, line 1: No tests for the massive class we added. We should add a test file for it as most pipelines do

nodejs/src/ingestion/common/common-ingestion-pipeline.ts
```diff
@@ -0,0 +1,489 @@
...
+    builder: ChunkPipelineBuilder<TPost, TPost, TeamAwareContext<TContext>, TeamAwareContext<TContext>>
+) => ChunkPipelineBuilder<TPost, TCurrent, TeamAwareContext<TContext>, TeamAwareContext<TContext>, ROut>
+
+/**
+ * Coalesces consecutive per-element steps into one sequential block. `extend`
+ * appends a step to the open block (or opens one); `build` closes it into the
+ * committed chunk-level transform. The block's start type stays existential —
+ * it lives only in the closures — so stages don't carry it as a type parameter.
+ */
+interface PreTeamChain<TInput, TContext, ROut extends string, CBatch, TCurrent> {
```
jose-sequeira: The two chain interfaces plus their four factory functions (`pendingPreTeamChain/committedPreTeamChain` and `pendingTeamChain/committedTeamChain`) implement the exact same pending/committed coalescing logic which leads to a good amount of duplicated code with only the type parameters changed (`BatchContext<TContext>` vs `TeamAwareContext<TContext>`, `SubpipelineTransform` vs `TeamSubpipelineTransform`).

Could we generalize and then provide type aliases?
`Chain<TBuilderContext, TTransform>` and then alias based on parameters?

nodejs/src/ingestion/common/common-ingestion-pipeline.ts
```diff
@@ -0,0 +1,489 @@
...
+        // The hooks handle their own side effects, so nothing rides out on
+        // `BatchResult.sideEffects` and drivers only ever drain results.
+        return newBatchingPipeline<TInput, TFinal, TContext, CBatch, TContext, ROut>(
+            (builder) => this.beforeBatchCallback(builder).handleSideEffects(promiseScheduler, sideEffectOptions),
+            (batch) =>
+                batch
+                    .messageAware((b) => this.transform(b))
+                    .handleResults(pipelineConfig)
+                    .handleSideEffects(promiseScheduler, sideEffectOptions),
+            (builder) => afterBatchCallback(builder).handleSideEffects(promiseScheduler, sideEffectOptions),
```
jose-sequeira, line 471: Based on this, the pipeline now drivers the side effects, however, the consumer still does

```
let result = await this.pipeline.next()
while (result !== null) {
    for (const sideEffect of result.sideEffects ?? []) {      // always [] now
        void this.promiseScheduler.schedule(sideEffect)
    }
    result = await this.pipeline.next()
}
```

This can be a bit misleading if folks are investigating since it becomes unclear who is driving side effects

jose-sequeira: Should be a harmless no op, but worth to mention

nodejs/src/ingestion/pipelines/clientwarnings/pipeline.ts
```diff
@@ -5,30 +5,23 @@ import { IngestionOutputs } from '~/common/outputs/ingestion-outputs'
...
 import { createApplyBasicEventRestrictionsStep } from '~/ingestion/common/steps/event-preprocessing/apply-event-restrictions'
 import { createDropOldEventsStep } from '~/ingestion/common/steps/event-processing/drop-old-events-step'
-import { EmitEventStepOutput } from '~/ingestion/common/steps/event-processing/emit-event-step'
 import { createHandleClientIngestionWarningStep } from '~/ingestion/common/steps/event-processing/handle-client-ingestion-warning-step'
 import { createRecordIngestionLagStep } from '~/ingestion/common/steps/record-ingestion-lag'
-import { addTeamToContext } from '~/ingestion/common/subpipelines/helpers'
-import { newBatchingPipeline } from '~/ingestion/framework/builders'
-import { PipelineConfig } from '~/ingestion/framework/result-handling-pipeline'
 
 export interface ClientWarningsPipelineConfig {
```
jose-sequeira, line 26: `consumer.test.ts` in `clientwarnings` mocks this entire file, so none of this gets tested

jose-sequeira (approved):
Approved, overall change seems good to unify code. Think some improvements to testing across the pipelines can be made as well as other smallr call outs

PR #70803, opened 2026-07-14
feat: Add live widget platform

jordanm-posthog:
## Problem
All widgets are currently static / locked to dashboard refreshes. 

## Changes
Example for Web Analytics:

https://github.com/user-attachments/assets/fdd3f710-0a22-40d0-a79d-e45665e680cd

This PR adds the platform for **live** widgets. A live widget seeds once from `run_widgets`, then keeps itself fresh client-side from a live event stream. It does not wait on the dashboard refresh loop.

```mermaid
flowchart LR
    RW[run_widgets seed<br/>carries generatedAt]:::phGray
    SSE{{livestream SSE}}:::phRed
    SW[sliding window<br/>max-merge, drop events at or before generatedAt]:::phBlue
    TILE[tile re-renders<br/>from selectors]:::phYellow
    RW -->|seed once, re-seed on refresh| SW
    SSE -->|batched events| SW
    SW --> TILE
```

1. **The `run_widgets` result is a seed, not the state.** It is a snapshot plus a `generatedAt` server timestamp, and the tile takes over from there.
2. **Seeds must be idempotent.** Manual refresh and dashboard auto-refresh both re-run `run_widgets`. The platform re-seeds live tiles on purpose (a re-seed heals gaps from a dropped connection or a backgrounded tab), so a merge must never double count.
3. **No `dateRange` or `filterTestAccounts`.** A live tile shows a fixed real-time window, and the stream cannot apply test-account filters, so seed and stream would disagree.

### Backend: two `WidgetSpec` fields

In `products/dashboards/backend/widget_specs/registry.py`:

| Field | Purpose |
| --- | --- |
| `is_live: bool` | Marks the type live. Flows into `widget-form-fields.json` via `hogli build:widget-types`, read on the frontend through `isLiveDashboardWidgetType()`, and exposed as `live` on the REST/MCP catalog. Do not hand-maintain frontend live-type lists. |
| `creation_flag: str \| None` | Adds-only rollout gate, resolved in `widget_create.py` via `feature_flags.widget_flag_enabled`. Creating a tile needs the flag; already-placed tiles keep rendering when it is off. Not live-specific, but a new live family should set it as its kill switch. |

### Frontend: the toolkit at `products/dashboards/frontend/widgets/live/`

Compose these. Do not re-wire SSE, batching, or pruning by hand.

| Module | Responsibility |
| --- | --- |
| `LiveWidgetSeedPayload` (`types.ts`) | Base interface for seed payloads (`generatedAt`). Extend it in your result type. |
| `liveWidgetStream(options)` | Kea logic builder. Opens one livestream SSE connection, batches events on a flush interval, ticks every 60s, all through `cache.disposables` (tears down on unmount, pauses on hidden tabs). |
| `LiveWidgetSlidingWindow` | Minute-bucketed window: an overall count plus named breakdown domains. Encodes the seed-merge correctness rules. |
| `useLiveWidgetSeed(payload, seed, guard?)` | The one prop-to-action React bridge. Feeds the tile's `result` prop into your logic's seed action. |
| `LiveWidgetEmptyState` | "No data in the window yet" body with an optional CTA. |
| `LiveWidgetIndicator` | The pulsing "Live" header marker. Platform-rendered, you do not wire this. |
| `isLiveDashboardWidgetType(widgetType)` | Frontend read of `is_live` from the generated manifest. |

### 1. Spec

```python
LIVE_SIGNUPS_WIDGET_TYPE: WidgetSpec(
    widget_type=LIVE_SIGNUPS_WIDGET_TYPE,
    config_model=LiveSignupsWidgetConfig,
    query_fn=run_live_signups_widget,
    required_scopes=("query:read",),
    group_id="my_product",
    group_label="My product",
    label="Live signups",
    description="Signups in the last 30 minutes, updating live.",
    required_product_access=None,
    product_access_denied_message=None,
    availability_requirements=(),
    form_fields=("limit",),          # no dateRange, no filterTestAccounts
    filter_fields=(),
    is_live=True,
    creation_flag="my-product-live-signups",
),
```

### 2. Seed query returning `generatedAt`

The seed is ordinary product code, a `query_fn` that delegates to your product's query module. Stamp `generatedAt` at query time and cap the payload, since it rides the batched `run_widgets` response.

```python
from datetime import UTC, datetime

def run_live_signups_widget(*, team, config, include_total_count=False, **kwargs):
    # include_total_count is part of the generic query_fn convention; accept and ignore it.
    return {
        "generatedAt": datetime.now(UTC).isoformat(),
        "buckets": _signups_by_minute(team=team, window_minutes=30),  # your HogQL
    }
```

### 3. One shared kea logic on `liveWidgetStream` + `LiveWidgetSlidingWindow`

Make the logic **unkeyed** and share it across the family's tiles. Kea ref-counting then keeps a single SSE connection no matter how many live tiles are on the dashboard, and tears it down when the last one unmounts.

The window is a mutable accumulator, so hold it in `cache` and bump a version to drive re-renders. `liveWidgetStream` hands you `(events, logic)` and `(logic)` in its callbacks, so it stays decoupled from your action names (kea-typegen can't see builder-injected symbols).

```ts
import { LiveWidgetSlidingWindow, liveWidgetStream } from 'products/dashboards/frontend/widgets/live'

liveWidgetStream({
    getStreamSpec: () => ({ eventType: 'signed_up', columns: ['properties.$browser'] }),
    onEvents: (events, logic) => logic.actions.ingestEvents(events),
    onMinuteTick: (logic) => logic.actions.pruneWindow(),
})
```

Your logic owns the window and the merge calls (all four are the correctness surface):

```ts
// afterMount: reset so data can't leak across dashboards
cache.window = new LiveWidgetSlidingWindow<'browsers'>({
    windowMinutes: 30,
    breakdowns: {
        browsers: (e) => (typeof e.properties?.$browser === 'string' ? e.properties.$browser : null),
    },
})

// ingestEvents: fold streamed events in, then bump the render version
events.forEach((e) => cache.window.addEvent(e))

// seed (from the tile result): merge, never replace
cache.window.mergeCountSeed(payload.buckets, payload.generatedAt)
// cache.window.mergeBreakdownSeed('browsers', payload.browserBuckets, payload.generatedAt)

// pruneWindow (every minute): drop buckets older than the window
cache.window.prune()

// selectors read snapshots: cache.window.totalCount() / .breakdownTotals('browsers')
```

New to kea plumbing here? See `/writing-kea-logics` and `/using-kea-disposables`.

### 4. Components

Bridge the seed prop with `useLiveWidgetSeed`, render from your selectors, and show `LiveWidgetEmptyState` when the window is empty. The "Live" marker and tag are platform-rendered, you do not add them.

```tsx
useLiveWidgetSeed(result, seedFromResult)   // default guard: only seeds once generatedAt is present

return total > 0 ? (
    <BigNumber value={total} />
) : (
    <LiveWidgetEmptyState message="No signups in the last 30 minutes yet." />
)
```

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context


MattPua:
@jordanm-posthog looking at this today!

do you have screenshots i can look at

products/dashboards/frontend/widgets/live/LiveWidgetIndicator.tsx
```diff
@@ -0,0 +1,12 @@
...
+    return (
+        <span className="flex items-center gap-1 font-medium text-success" data-attr="live-widget-indicator">
+            <span className="relative flex size-2" aria-hidden>
+                <span className="absolute inline-flex h-full w-full animate-ping rounded-full bg-success opacity-75" />
+                <span className="relative inline-flex size-2 rounded-full bg-success" />
+            </span>
+            Live
+        </span>
+    )
+}
```
MattPua: nit: tiniest of nits, not 100% sure if this really needed its own file, technically might be able to group it with the empty state above, and make it like how shadcn does component exports, with multiple exports per file using a generic file name

MattPua:
> No dateRange or filterTestAccounts. A live tile shows a fixed real-time window, and the stream cannot apply test-account filters, so seed and stream would disagree.

I see this mentioned, maybe answered by a screenshot, but does the UI make it clear about the filterTestAccounts not being a thing for live widgets?

products/dashboards/frontend/widgets/live/liveWidgetTypes.ts
```diff
@@ -0,0 +1,8 @@
+import widgetFormFields from '../../generated/widget-form-fields.json'
+
+const WIDGET_MANIFEST = widgetFormFields.widgets as Record<string, { live?: boolean }>
+
+/** SSOT is `WidgetSpec.is_live` on the backend, flowed through `hogli build:widget-types`. */
+export function isLiveDashboardWidgetType(widgetType: string): boolean {
+    return WIDGET_MANIFEST[widgetType]?.live === true
+}
```
MattPua, line 8: nit:

a bit overkill at the moment to have a separate file for just this one extraction here. Would be more sensible atm to be in some shared util function or something.

 might make more sense later if we add more types in here, so feel free to ignore also.

products/dashboards/frontend/widgets/live/types.ts
```diff
@@ -0,0 +1,13 @@
...
+ * The result is a one-shot SEED of the widget's real-time state, not the state itself: the tile
+ * self-updates client-side afterwards. `generatedAt` is the server clock at seed-query time —
+ * streamed events at or before it are already counted in the seed, so client merges use it to
+ * avoid double counting. Any re-run of run_widgets (manual tile refresh, dashboard auto-refresh)
+ * re-seeds, so seed merges must be idempotent.
+ */
+export interface LiveWidgetSeedPayload {
+    /** Server clock at seed-query time (ISO-8601). */
+    generatedAt: string
+}
```
MattPua: similar nit as above: technically could be merged together with the `liveWidgetTypes.ts` file

products/dashboards/backend/widget_specs/openapi.py
```diff
@@ -117,6 +117,12 @@ def _build_openapi_serializers() -> tuple[
                     ),
                 ),
                 "required_product_access": serializers.CharField(required=False, allow_null=True),
+                "live": serializers.BooleanField(
+                    help_text=(
+                        "Whether tiles of this type self-update in real time after load. Live tiles show a "
+                        "fixed real-time window, so their config takes no dateRange."
+                    ),
+                ),
```
MattPua, line 126: so live widgets mention we dont apply the dateRange and filter test accounts, but i think we need the openapi spec to reflect that?

MattPua:
question:

do we do any optimizations if the same live widget is used in multiple places? basically, if we're streaming live data, is it the same source thats streaming it, and if so, are doing the same fetch on the server side twice, or is that handled with cache or something?

.agents/skills/manage-dashboard-widgets/references/live-widgets.md
```diff
@@ -0,0 +1,63 @@
+# Live widgets (self-updating tiles)
+
+A **live** widget seeds once from `run_widgets`, then self-updates client-side — it does not rely on the dashboard refresh loop for freshness.
```
MattPua, line 3: nit: i dont think the skills docs should refer to "this PR" or "separate PR" unless you want to leave explicit PR referencecs with direct links for agent context

products/dashboards/frontend/widgets/live/liveWidgetStream.ts
```diff
@@ -0,0 +1,114 @@
...
+                cache.disposables.add(() => {
+                    batch = []
+                    const connection = createStreamConnection({
+                        url,
+                        token,
+                        onMessage: (data) => {
+                            try {
+                                batch.push(JSON.parse(data) as LiveEvent)
+                            } catch (error) {
+                                console.error('Failed to parse live widget event:', error)
```
MattPua: is it up to the widget to deal with errors if it fails to make a connection?

products/dashboards/frontend/widgets/live/liveWidgetStream.ts
```diff
@@ -0,0 +1,114 @@
...
+                    url.searchParams.append('eventType', spec.eventType)
+                }
+                if (spec.columns?.length) {
+                    url.searchParams.append('columns', spec.columns.join(','))
+                }
+
+                let batch: LiveEvent[] = []
+
+                // pauseOnPageHidden (default) aborts the stream when the tab hides and reconnects on show.
+                cache.disposables.add(() => {
```
MattPua: is we potentially had like 20 live widgets on a page, how bad would that be for performance?

products/dashboards/frontend/widgets/live/LiveWidgetSlidingWindow.ts
```diff
@@ -0,0 +1,150 @@
...
+export interface LiveWidgetSeedBucket {
+    minute: string
+    value: string
+    views: number
+}
+
+const toEpochSeconds = (iso: string): number | null => {
+    const ms = new Date(iso).getTime()
+    return Number.isFinite(ms) ? ms / 1000 : null
+}
```
MattPua, line 15: nit: for what its worth, ifeel like we must have something already for this

MattPua (approved):

PR #70762, opened 2026-07-14
feat(tasks): route slack follow-ups through the queue with per-message actors

VojtechBartos:
## Problem

Follow-ups are queued now (#70806), but everything that needs "who sent this one" — credential resolution, reply @-tagging, the permission broker — still reads shared mutable state the next message overwrites. Two people messaging back-to-back get replies tagged to whoever spoke last.

## Changes

- Each queued message carries its sender: the follow-up signal gains `actor_user_id` + a `context` dict (extension point, holds the sender's Slack id). Additive with defaults, so old histories replay unchanged.
- Delivery pins credentials and the token mint to that sender, and records `message_id → sender` (redis, 2h TTL).
- The relay resolves the @-tag from `message_id` first (echoed by the agent, see below), falling back to the live actor for pre-rollout runs. `mapping.latest_actor_slack_user_id` stays as the pre-rollout fallback, marked for removal.
- One `slack_actor_state_updates` builder in `run_actor.py` for every writer of the load-bearing keys.

```text
U2 sends a message
  ▼ signal [message, artifact_ids, message_id, actor_user_id, context{slack_user_id}]
task queue ─▶ send_followup_to_sandbox   pins creds/token, records message_id → U2
  ▼
agent-server runs the turn ─▶ POST relay_message {text, message_id}   ◀─ echo from code#3434
  ▼
relay_slack_message   tag = actor(message_id) → live actor → mapping
  ▼ "@U2 …"
```

## Paired with PostHog/code#3434

The agent-server echoes the initiating `message_id` on `relay_message` (that PR). Neither side hard-depends on the other: without the echo the relay uses its fallbacks (correct except rapid multi-user bursts), and an older backend ignores the unknown field. Exact per-message tagging switches on once both ship, with the agent release (npm + sandbox image) as the gating leg. Verified end-to-end locally with the two PRs together.

## How did you test this code?

Locally, two-user thread: each reply tags its sender and PostHog attribution follows the speaker. Automated coverage for the pinning, the turn-start actor stamp, and the forwarding/API call shapes.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

Internal behavior only.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Claude Code (Claude Fable 5).

PR #70806, opened 2026-07-14
feat(tasks): route slack follow-ups through the workflow follow-up queue

VojtechBartos:
## Problem

Slack thread follow-ups were delivered straight to the sandbox, bypassing the task workflow's follow-up queue — so Slack and web messages had no ordering against each other, and a busy agent needed a 504 special case.

## Changes

The Slack forward signals `send_followup_message` like the web UI does. The workflow delivers one message at a time and each delivery blocks until the turn completes, so messages execute strictly in order. The deterministic Slack `message_id` rides in the signal, keeping redelivery idempotent. Delivery failures surface through the run's error path; the forward still replies inline when a message can't be queued. The facade's direct `send_user_message` wrapper loses its last caller and is removed.

All tested locally in a two-user Slack thread (rapid alternating commands execute strictly in order), plus the rewritten forwarding tests, delivery tests, and web API call-shape tests — ~360 green.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal behavior only.

## 🤖 Agent context

Claude Code (Claude Fable 5).

### DIFF
--- a/posthog/temporal/ai/slack_app/activities/task_creation.py
+++ b/posthog/temporal/ai/slack_app/activities/task_creation.py
@@ -948,13 +948,6 @@ def forward_posthog_code_followup_activity(
     if user_message_ts:
         safe_react(slack.client, channel, user_message_ts, "eyes")
 
-    auth_token = None
-    if actor_user and actor_user.id:
-        distinct_id = actor_user.distinct_id or f"user_{actor_user.id}"
-        auth_token = tasks_facade.create_sandbox_connection_token(
-            task_run.id, user_id=actor_user.id, distinct_id=distinct_id
-        )
-
     uploaded_attachments, attachment_skips = _upload_prepared_slack_attachments(
         tasks_facade,
         task_run_id=task_run.id,
@@ -973,42 +966,24 @@ def forward_posthog_code_followup_activity(
         or user_text
     )
 
-    send_kwargs: dict[str, Any] = {
-        "auth_token": auth_token,
-        "timeout": 90,
-        # Deterministic across activity retries: a retry after a partial failure
-        # (or the in-line resend below) redelivers with the same id, and the
-        # agent-server drops the duplicate instead of applying the message twice.
-        "message_id": _slack_followup_message_id(channel, user_message_ts, thread_ts),
-    }
-    if uploaded_attachments:
-        send_kwargs["artifacts"] = uploaded_attachments
-
-    result = tasks_facade.send_user_message(task_run.id, user_text, **send_kwargs)
-    if not result.success and result.retryable and result.status_code != 504:
-        result = tasks_facade.send_user_message(task_run.id, user_text, **send_kwargs)
-
-    if not result.success:
+    # Queue on the workflow so delivery is ordered with the web path. The
+    # deterministic message id keeps redelivery idempotent.
+    signal_result = tasks_facade.signal_task_run_user_message(
+        task_run.id,
+        mapping.task_id,
+        task_run.team_id,
+        content=user_text,
+        artifact_ids=_uploaded_attachment_ids(uploaded_attachments),
+        message_id=_slack_followup_message_id(channel, user_message_ts, thread_ts),
+    )
+    if signal_result is not True:
         logger.warning(
-            "posthog_code_followup_forwarding_failed",
+            "slack_app_followup_signal_failed",
             channel=channel,
             thread_ts=thread_ts,
-            error=result.error,
-            status_code=result.status_code,
+            task_run_id=str(task_run.id),
+            signal_result=signal_result,
         )
-        if result.retryable and result.status_code == 504:
-            # Agent is still processing — leave the :eyes: reaction up so the thread
-            # reads as in-progress. relayAgentResponse fires when it finishes,
-            # delivering the correct response to Slack.
-            _delete_followup_progress(
-                integration_id=inputs.integration_id,
-                channel=channel,
-                thread_ts=thread_ts,
-                user_message_ts=user_message_ts,
-                mentioning_slack_user_id=mapping.mentioning_slack_user_id,
-            )
-            return True
-
         _set_followup_done_reaction(slack, channel, user_message_ts, "x")
         slack.client.chat_postMessage(
             channel=channel,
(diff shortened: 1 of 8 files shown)


VojtechBartos:
* **#70454** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#70762** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70762?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#70806** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/70806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#71562** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

products/tasks/backend/temporal/client.py
```diff
@@ -465,10 +465,15 @@ def execute_build_sandbox_image_workflow(image_id: str, team_id: int) -> None:
     )
 
 
-def signal_task_followup_message(workflow_id: str, message: str | None, artifact_ids: list[str]) -> None:
+def signal_task_followup_message(
```
tatoalo, line 468: here we need patching or in-roll workflows will fail, similarly in https://github.com/PostHog/posthog/pull/70762

VojtechBartos: yeah, you are right, interestingly my agent were talling me extra params will be dropped, but looks like TypeError will follow aka `await handler(*input.args)`

tatoalo: if that agent was Fable, Anthropic needs to send us some money back :laughing:

VojtechBartos: hahaha, opus 4.8, but still should refund us right? :D

tatoalo: def! And a beer as well :beers:

VojtechBartos: i have created new base PR0 which just adds defaults to signal handler, we should be good to go https://github.com/PostHog/posthog/pull/71562

cvolzer3 (commented):
Fable comments incoming

posthog/temporal/ai/slack_app/activities/task_creation.py
```diff
@@ -1050,42 +1043,25 @@ def forward_posthog_code_followup_activity(
...
+    # turn runs under this message's actor. The deterministic message id
+    # keeps redelivery idempotent.
+    signal_result = tasks_facade.signal_task_run_user_message(
+        task_run.id,
+        mapping.task_id,
+        task_run.team_id,
+        content=user_text,
+        artifact_ids=_uploaded_attachment_ids(uploaded_attachments),
+        message_id=_slack_followup_message_id(channel, user_message_ts, thread_ts),
+    )
```
cvolzer3: The message now travels in the ordered queue, but the actor still travels out-of-band through run state: this activity writes `slack_actor_user_id` (via `update_task_run_state` above) before signaling, and `send_followup_to_sandbox` re-reads it at delivery time via `get_task_run_credential_user`. Under the exact scenario this PR enables (two users queuing rapid messages), Alice's queued message can be delivered after Bob's forward has already overwritten the state, so her turn runs with Bob's connection token, Bob's `distinct_id`, and Bob's user-scoped MCP servers mounted by the credential refresh. The old path minted the token inline with the correct actor, so this is a new gap, and this comment's claim that "the turn runs under this message's actor" is no longer guaranteed.

Suggestion: carry `actor_user_id` in the signal payload (`send_permission_response` already does exactly this) and thread it through `PendingFollowup` into `SendFollowupToSandboxInput`, falling back to run state when absent. If that's deferred, this comment should at least be reworded so it doesn't document a guarantee the code no longer provides.

products/slack_app/backend/tests/test_followup_forwarding.py
```diff
@@ -1212,68 +1190,27 @@ def test_forwarding_failure_posts_error(self, mock_slack_cls, mock_send, mock_to
         call_kwargs = mock_slack_instance.client.chat_postMessage.call_args.kwargs
         assert "couldn't deliver" in call_kwargs["text"]
```
cvolzer3: `test_forwarding_failure_posts_error` is stale now: both patches target functions the activity no longer calls, so the mocked failure return is inert. What actually happens is the real `signal_task_run_user_message` runs, `sync_connect()` attempts a live Temporal connection inside a unit test, the signal fails, and the activity posts the error reply, so the assertion passes by accident while doing real network I/O on every run. The scenario is properly covered by `test_signal_failure_posts_error_reply` below; this one can be deleted.

posthog/temporal/ai/slack_app/activities/task_creation.py
```diff
@@ -1050,42 +1043,25 @@ def forward_posthog_code_followup_activity(
...
+    if signal_result is not True:
```
cvolzer3, line 979: `signal_task_run_user_message` collapses every exception to `False`, so a transient Temporal connectivity blip now lands the user on an :x: and "The sandbox may have stopped. Please try starting a new task." with no retry, where the old path retried retryable errors once. Since this activity is already designed to be safely re-run under Temporal retries (deterministic `message_id`, upsert-safe attachment upload), consider letting transient signal errors propagate so the activity retries, and reserving this terminal reply for definitive outcomes (run not found, workflow already completed).

posthog/temporal/ai/slack_app/activities/task_creation.py
```diff
@@ -1094,7 +1070,7 @@ def forward_posthog_code_followup_activity(
         )
         return True
 
-    # Message delivered; the agent is now working on it, so leave the :eyes: reaction
+    # Message queued; the agent picks it up next, so leave the :eyes: reaction
```
cvolzer3, line 995: Behavior change worth stating in the PR description: once the message is queued, a delivery failure inside the workflow marks the whole run failed (`_send_followup_to_sandbox`'s except path sets `followup_delivery_failed`), whereas the old direct path posted an inline error and left the run alive for an in-thread retry. And in that async failure path nothing swaps this :eyes: reaction to :x: on the user's message; the failure only surfaces through the generic run-failure Slack update. This is consistent with the web path so it's likely intentional, but it deserves an explicit sign-off.

cvolzer3 (approved):

PR #70807, opened 2026-07-14
fix(experiments): add fairness key

rodrigoi:
## Problem

Stacked on #70801, and closes the last gap left by removing the workflow's concurrency pool in #70793: dispatch ordering. A run now schedules every metric activity up front, and the task queue serves them FIFO, so one org's large run queued first occupies the queue head while another org's small run waits behind it until autoscaling adds pods. The old per-run fan-out cap limited this incidentally; nothing replaced it at the dispatch layer.

## Changes

Dispatches the calc activities with a Temporal task-queue fairness key set to the org id:

- New optional `fairness_key` on `ExperimentMetricsRecalculationWorkflowInputs`, passed by the trigger endpoint as the experiment's organization id, and applied per calc activity via `Priority(fairness_key=...)`. Under backlog, the task queue round-robins dispatch across keys instead of FIFO, so no org's burst can queue-starve another's.
- The key scope deliberately matches the `app_per_org` ClickHouse limiter (#70801): dispatch fairness and query quota act on the same boundary. Interleaving orgs at dispatch also spreads a single org's wave of activities over time, which softens how hard it slams its own quota.
- On a namespace without task-queue fairness support the field is ignored and ordering stays FIFO, so this degrades to a safe no-op. The key is applied on the activities, not the workflow, because activity slots are the contended resource.

```bash
pnpm turbo run backend:test --filter=@posthog/products-experiments
```

cat-type-small

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Model: Fable 5
Manually refactored: **yes**

Skills used:

Relevant decisions:

- Fairness key is the org id rather than the team id or run id, so dispatch fairness shares a boundary with the per-org query limiter; a per-run key would give a bursty org more aggregate share.
- Verified the installed temporalio SDK (1.24.0) supports `Priority(fairness_key=...)` on `execute_activity`, and that unsupported namespaces silently ignore it, before building on the feature.
- No dedicated test: the change is pass-through plumbing into Temporal's dispatcher, and asserting the call kwargs would be a change-detector test with no observable behavior to pin locally.

### DIFF
--- a/products/experiments/backend/presentation/views.py
+++ b/products/experiments/backend/presentation/views.py
@@ -1047,7 +1047,10 @@ class EnterpriseExperimentsViewSet(
                 asyncio.run(
                     temporal.start_workflow(
                         "experiment-metrics-recalculation-workflow",
-                        MetricsRecalcInputs(recalculation_id=recalculation_id),
+                        MetricsRecalcInputs(
+                            recalculation_id=recalculation_id,
+                            fairness_key=str(experiment.team.organization_id),
+                        ),
                         id=f"experiment-metrics-recalculation-{recalculation_id}",
                         task_queue=settings.EXPERIMENTS_RECALCULATION_TASK_QUEUE,
                     )
--- a/products/experiments/backend/temporal/models.py
+++ b/products/experiments/backend/temporal/models.py
@@ -39,6 +39,9 @@ class ExperimentMetricsRecalculationWorkflowInputs:
     """Input to the batch metrics recalculation workflow."""
 
     recalculation_id: str  # UUID as string
+    # Task-queue fairness key for the calc activities (the org id, matching the app_per_org limiter scope).
+    # None falls back to default dispatch order.
+    fairness_key: str | None = None
 
 
 @dataclasses.dataclass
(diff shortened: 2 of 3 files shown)


rodrigoi:
* **#70915** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#70807** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70807?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/70807?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#70801** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#70793** <a href="https://app.graphite.com/github/pr/PostHog/posthog/70793?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#69836** <a href="https://app.graphite.com/github/pr/PostHog/posthog/69836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

jurajmajerik (approved):

PR #70809, opened 2026-07-14
feat(onboarding): add quickstart landing page after onboarding

fercgomes:
## Problem

Finishing onboarding currently drops users on the AI-first home page. For a brand-new project that's a dead end three times over:

- most new projects have no events yet, so nothing on the platform can show value
- an empty chat input assumes you already know what to ask
- nothing on the screen shows what PostHog can actually do, so cross-product discovery never happens

Looking at our own product analytics, a large share of newly signed-up organizations never ingest a first event, and only a small fraction of users who complete onboarding ever submit a query on the AI home. The AI home is great for activated users, but it's the wrong first screen for an empty project.

## Changes

New `Quickstart` scene at `/quickstart` (`frontend/src/scenes/quickstart/`), acting as the landing page right after onboarding:

- **Get your data flowing** hero: the AI wizard command to copy, a link into the framework install guide, a teammate invite, and a "connect a data source" escape hatch. A live indicator polls for the team's first event and collapses into a success bar once data arrives (reuses the existing `useInstallationComplete` hook).
- **Featured product grid** (product analytics, web analytics, session replay, error tracking, feature flags) plus the full catalog below. Every card derives a status from team state plus a 30-day event-signal query: `Active` (the tool's own signal exists, e.g. exceptions captured or flags evaluated), `Ready` (usable but no signal yet, with a nudge toward it), `1-click enable` (patches team opt-ins and records a product intent), `Needs install` (SDK-dependent tools before any events, opening the SDK dialog), or `Needs setup`. A data-signals strip tracks the activation ladder: installed, events in dev, events in production, custom events. Copy and icons come from the existing `availableOnboardingProducts` metadata.
- **Learn the ropes** row bridging to the AI home (Max), docs, and tutorials.

Routing and navigation:

- "Quickstart" is now the first item in the sidebar's Project section.
- The **Configure homepage** modal (the gear on the Home sidebar item) gains a Quickstart option, and completing onboarding for the first time persists Quickstart as the user's homepage (unless one is already configured). The Home button then lands new signups on Quickstart, and switching back to the AI chat home is the existing Launchpad option in the same modal.
- The context-first onboarding completion and the legacy flow's fallback now redirect to `/quickstart` instead of `/`. Per-product `completeRedirectUrl`s are unchanged, so finishing a specific product's onboarding still lands in that product. The home page itself is untouched.

```mermaid
flowchart LR
    A[Onboarding completed] --> B{Product has its own destination?}
    B -- yes --> C[Product page]
    B -- no --> D[Quickstart]
    D -. once data flows .-> E[AI home / user default]
    class D phBlue;
    class A phYellow;
    class C,E phGray;
```

> [!NOTE]
> Behavior change, gated behind the `quickstart-homepage` feature flag (created in production at 0% rollout, off means behavior matches master): users finishing the context-first onboarding (and legacy flows without a product destination) now land on `/quickstart` instead of the AI home, and first-time onboarders get Quickstart persisted as their default homepage. The page also captures `quickstart viewed`, `quickstart action clicked`, and `quickstart product enabled` events so we can measure whether it helps activation.

https://github.com/user-attachments/assets/d6ff684c-c5bf-4002-99df-87af2f9da21f

- `pnpm jest src/scenes/onboarding/legacy/onboardingLogic.test.ts`: 73/73 pass. The experiments fallback assertion now expects `/quickstart` (that flow has no product-specific destination, so it exercises the new fallback).
- `pnpm typescript:check`: zero errors in the new/changed files (the sandbox shows pre-existing failures in unrelated products from unbuilt quill workspace packages).
- `hogli ci:preflight --fix`: no failures.
- Not manually tested in a running app, and the new scene has no dedicated unit tests: the status derivation is a direct mapping from team fields, and the redirect change is covered by the existing onboarding suite.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No published docs describe the post-onboarding destination, so nothing to update.

## 🤖 Agent context

Built with PostHog Code (Claude) directed by Fernando. Research before writing code: recent internal design discussions about the post-onboarding experience, funnels on our own analytics (signup to first event ingested, onboarding completion to AI home usage), and getting-started patterns from Mobbin (Hootsuite, Apollo, June, Pipedrive). A Slack search tool wasn't available in the session.

Key decisions: reuse existing plumbing rather than new state (`availableOnboardingProducts` for copy/icons, `teamLogic.updateCurrentTeam` opt-ins for 1-click enables, `useInstallationComplete` for the live event check, `urls.onboarding()` for setup deep links); keep per-product completion redirects and only change the home fallbacks; the whole experience is now gated behind the `quickstart-homepage` flag (0% rollout); auto-hiding the nav item for activated teams remains the next knob. Guides applied: `frontend/src/AGENTS.md` (Lemon components, business logic in kea) and the generated-types rules; kea typegen run for the new logic.

---


fercgomes:
## 🤖 QA team review (multi-agent)

Six independent review agents (security, performance, frontend, copy, and two generalists) reviewed this branch. All fixable findings were addressed in e725f04a83b1258966633a67b8c5f214be9cb07f; the table below records status.


| Key                 | Value                                                                          |
| ------------------- | ------------------------------------------------------------------------------ |
| **Branch**          | `posthog-code/quickstart-homepage`                                             |
| **Base**            | `master`                                                                        |
| **Files changed**   | 14                                                                              |
| **Agents deployed** | 🔒 security · ⚡ performance · 🎨 frontend · ✏️ copy · 🧑‍💻 generalist-a · 🕵️ generalist-b |
| **Date**            | 2026-07-14                                                                      |

---

## 📋 Summary

- New `/quickstart` landing scene gated on the `test` variant of the `quickstart-homepage` multivariate flag, becoming the default post-onboarding homepage.
- Tool cards derive Active/Ready/Enable/Install statuses from a 30-day HogQL signals query; publications rail streams the posthog.com RSS feed; header surfaces the project API key.
- Flag gating, onboarding redirects, and homepage persistence touch `onboardingLogic`, `ConfigureHomeModal`, and nav.

### Key findings

- 🟠 Flag-off dead end: persisted `/quickstart` homepage rendered "Not found" once the user leaves the `test` variant (3 agents converged)
- 🟠 Unbounded 30-day HogQL aggregate re-ran on every homepage mount (4 agents converged)
- 🟡 RSS `FeedStream` had a concurrency race and O(n²) per-chunk parsing (2 agents each)
- 🟡 "Enable" spinner reset by any unrelated team update; enable failures were silent

---

## 🏁 Verdict

> ⚠️ **REQUEST CHANGES** (now addressed)

One HIGH + three MEDIUM agent risk levels → overall HIGH. The two convergent blockers (flag-off 404 homepage, per-mount full scan) plus all fixable findings were remediated on the branch in the same pass; remaining items are tracked follow-ups.

---

## 👥 Agent summaries

| Agent             | Risk       | Summary                                                                                                          |
| ----------------- | ---------- | ---------------------------------------------------------------------------------------------------------------- |
| 🔒 security       | 🟢 LOW     | No new backend surface, auth, or secret exposure; suggested https-scheme validation on RSS-fed URLs (defense in depth). |
| ⚡ performance    | 🟡 MEDIUM  | 30-day full-scan HogQL query on every mount of the homepage; O(n²) RSS chunk parsing; feed buffer retained for the session. |
| 🎨 frontend       | 🟡 MEDIUM  | Flag-off NotFound strands persisted homepages; silent enable failures; 404 flash while flags load; newsletter seed will rot. |
| ✏️ copy           | 🟢 LOW     | Copy is strong overall; nits on the API key tooltip, MCP command in prose, Active/Ready subtitle, and "Events in dev" label. |
| 🧑‍💻 generalist-a   | 🟡 MEDIUM  | Persisted-homepage/flag lifecycle mismatch; FeedStream not concurrency-safe; enable spinner reset by unrelated team updates. |
| 🕵️ generalist-b   | 🟠 HIGH    | Same two structural risks (flag-off 404, per-mount full scan) plus cache-shape crash path and RSS stream races.  |

**Note:** ✏️ copy findings are always non-blocking nits. 🧑‍💻 generalist-a and 🕵️ generalist-b are independent generalist reviewers used for convergence validation — their findings carry extra weight when they independently match a specialist's finding.

Risk emojis: 🔴 CRITICAL, 🟠 HIGH, 🟡 MEDIUM, 🟢 LOW, ⚪ NONE

---

## 📝 Findings

| #  | Status    | Priority    | Finding                                                        | Location                          | Agents                          | Reasoning                                                                                     | Suggested fix / applied fix                                                                 |
| -- | --------- | ----------- | -------------------------------------------------------------- | --------------------------------- | ------------------------------- | ---------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------- |
| 1  | ✅ Fixed  | 🟠 High     | Flag-off strands persisted quickstart homepage on "Not found"   | `Quickstart.tsx` scene gate       | frontend, generalist-a, generalist-b | Homepage persists server-side but the scene 404s when the flag isn't `test` (rollback, experiment end, flags not yet loaded) | Scene now shows a spinner while flags load and redirects to the project homepage otherwise; ConfigureHomeModal keeps the Quickstart option visible for anyone already using it |
| 2  | ✅ Fixed* | 🟠 High     | Unbounded 30-day events aggregate on every homepage mount       | `quickstartLogic.ts` loadToolSignals | performance, frontend, generalist-a, generalist-b | Scene is the default homepage; every "Home" click re-ran a full 30-day scan with JSON property predicates | 5-minute per-team client-side cache added; *deeper mitigation (sampling / skip for activated teams / server cache) is a follow-up before wide rollout |
| 3  | ✅ Fixed  | 🟡 Medium   | `FeedStream` unsafe under concurrent readers                    | `publications.ts` FeedStream      | generalist-a, generalist-b      | Initial load + eager scroll sentinel can overlap; interleaved `reader.read()` corrupts pages   | `ensureItems` calls now serialized on an internal promise chain; `start()` memoized          |
| 4  | ✅ Fixed  | 🟡 Medium   | O(n²) RSS parsing (full-buffer split per chunk)                 | `publications.ts` countItems      | performance, generalist-b       | Hundreds of multi-MB string splits on the main thread of a new user's landing page             | Item count tracked incrementally per chunk (with boundary-spanning tail)                     |
| 5  | ✅ Fixed  | 🟡 Medium   | Enable spinner cleared by any unrelated team update             | `quickstartLogic.ts` enablingProducts | generalist-a, generalist-b   | Global `updateCurrentTeamSuccess/Failure` reset re-armed the button mid-request → double PATCH | Per-product `productEnableFinished` action fired in the enable listener's `finally`          |
| 6  | ✅ Fixed  | 🟡 Medium   | "Enable" failures completely silent                             | `quickstartLogic.ts` enableProduct | frontend                        | Primary CTA failed with no feedback — the top papercut class                                   | Error toast on failure                                                                        |
| 7  | ⬜ Open   | 🟡 Medium   | Hardcoded newsletter seed will rot under "Fresh from PostHog"   | `publications.ts` NEWSLETTER_ISSUES | frontend                        | Static seed ages visibly ("5 months ago" cards under a "fresh" heading)                        | Serve the Substack feed via a server-side proxy (already tracked as a follow-up)             |
| 8  | ✅ Fixed  | 🟢 Low      | `quickstart viewed` captured before signals loaded              | `quickstartLogic.ts` afterMount   | frontend, generalist-a, generalist-b | `active_products` always reflected the empty fallback, skewing the experiment's own metrics    | Capture moved to `loadToolSignalsSuccess`                                                     |
| 9  | ✅ Fixed  | 🟢 Low      | localStorage cache: no entry-shape validation, `hasMore` hardcoded | `publications.ts` readCache     | generalist-a, generalist-b      | A corrupted entry crashed rendering; single-page feeds re-fetched or mis-advertised more pages | Entries validated (string title/url), `hasMore` stored in the cache, key bumped to v4        |
| 10 | ✅ Fixed  | 🟢 Low      | RSS-fed URLs rendered without scheme validation                 | `publications.ts` parse/fixImageUrl | security                        | Defense in depth against feed compromise / cache poisoning (`javascript:` URIs)               | Only `https://` link and image URLs survive parsing                                          |
| 11 | ✅ Fixed  | 🟢 Low      | Malformed RSS `pubDate` renders "Invalid Date"                  | `Quickstart.tsx` PublicationCard  | frontend                        | Classic date-display bug class                                                                 | Timestamp omitted unless `dayjs(...).isValid()`                                              |
| 12 | ✅ Fixed  | 🟢 Low      | `completeOnboarding` swallowed team-update failures silently    | `onboardingLogic.tsx`             | generalist-a                    | Sibling context flow toasts on the same failure; this path didn't                              | Matching error toast added                                                                    |
| 13 | ✅ Fixed  | 🟢 Low      | Copy nits: API key "safe" claim, MCP command in prose, Active/Ready subtitle, "Events in dev" label | `Quickstart.tsx` | copy                            | Advisory; clarity and consistency                                                              | Tooltip now says write-only; MCP command moved out of prose; subtitle distinguishes Active vs Ready; ladder step renamed "First events" |
| 14 | ⬜ Open   | 🟢 Low      | Polish follow-ups: hero image load feedback, tooltip keyboard a11y, feed buffer retained for session, external image origins on self-hosted | various | frontend, performance, generalist-b | Low-impact polish; none block the flag experiment                                              | Track as follow-ups                                                                           |

frontend/src/scenes/quickstart/Quickstart.tsx
```diff
@@ -0,0 +1,1230 @@
...
+    const installationComplete = useInstallationComplete('ingested_event')
+
+    const quickstartVariant = featureFlags[FEATURE_FLAGS.QUICKSTART_HOMEPAGE]
+    useEffect(() => {
+        // The route can outlive enrollment (it's persisted as some users' homepage),
+        // so a non-test variant gets sent home instead of stranded on a dead page
+        if (quickstartVariant !== undefined && quickstartVariant !== 'test') {
+            router.actions.replace(urls.projectHomepage())
+        }
+    }, [quickstartVariant])
```
MattBro: This redirect loops for anyone who got Quickstart persisted as their homepage and then falls out of the flag. `replace(urls.projectHomepage())` lands on `/home`, but `redirectToConfiguredHomepage` in `sceneLogic.tsx` sees the persisted `/quickstart` homepage and replaces right back here; its loop guard only bails when the current location already equals the target. Since the flag is the rollback lever, that breaks Home navigation for exactly the users a rollback needs to protect.

Related: if the flag is ever deleted outright, the variant stays `undefined`, so this effect never redirects and the page renders `SpinnerOverlay` forever.

Suggestion: once `featureFlagLogic.values.receivedFeatureFlags` is true, treat anything other than `'test'` (including `undefined`) as off, and clear the persisted homepage (`sceneLogic.actions.setHomepage(null)`) before the replace so `/home` sticks.

MattBro:
image

I feel like I'd have a hard time following these instructions because as soon as I click the button, I'd forget the next steps.

frontend/src/layout/scenes/ConfigureHomeModal.tsx
```diff
@@ -26,23 +29,32 @@ export function ConfigureHomeModal({ isOpen, onClose }: ConfigureHomeModalProps)
...
+        null
+    )
     const currentMode = isUsingProjectDefault
         ? 'launchpad'
-        : isUsingNewTabHomepage
-          ? 'search'
-          : isUsingDefaultDashboard
-            ? 'default_dashboard'
-            : null
+        : isUsingQuickstart
```
MattBro: nit: I wonder if there's a more readable way to structure this

frontend/src/scenes/quickstart/publications.ts
```diff
@@ -0,0 +1,326 @@
...
+    if (offset === 0 && publications.length > 0) {
+        writeCache(publications, hasMore)
+    }
+    return { publications, hasMore }
+}
+
+// Substack sends no CORS headers, so the newsletter feed can't be fetched from the
+// browser. Until it's proxied server-side, the rail ships with a static seed of recent
+// issues served through the same paging interface as the live blog feed.
+const NEWSLETTER_ISSUES: QuickstartPublication[] = [
```
MattBro, line 204: nit: Works for now, but this should be dynamic pretty quickly, I might forget to come back to update this if it were me. Is there something we can add so we remember to do that if it ends up being successful?

MattBro (approved):
Looks awesome! Excited to see how it impacts activation!

frontend/src/scenes/quickstart/publications.ts
fercgomes: wtf i did not realize this was being done like this

MattBro: Lol I mean I think it's fine for an initial experiment to roll it out like this then fast follow if it's promising. I guess depends on your level of confidence that we'll want this. Would hate to see a bunch of investment in a better solution if we just end up throwing it away. That being said, might be easy enough to one shot the real deal.

frontend/src/scenes/quickstart/Quickstart.tsx
fercgomes: Fixed in 1d7c2304492. The redirect now waits for feature flags, treats a missing flag as disabled, clears a persisted Quickstart homepage before navigating, and lives in quickstartLogic.

frontend/src/layout/scenes/ConfigureHomeModal.tsx
fercgomes: Refactored this into a typed getHomepageMode helper so the priority is explicit without the nested ternary.

frontend/src/scenes/quickstart/publications.ts
fercgomes: Tracked the server-proxied feed follow-up in [GROW-161](https://linear.app/posthog-team-growth/issue/GROW-161/proxy-the-quickstart-newsletter-feed-server-side).

fercgomes:
Restyled the header installation progress (`ff50f7ba04f`): the bordered card with the brand-blue `LemonProgress` bar is now a quiet chip that height-matches the project token beside it, reusing the page's own status vocabulary (pulsing accent dot, per-step pip strip, hairline "View details"). The whole chip opens the existing details dialog, and it wraps below the token on narrow viewports.

quickstart-chip

In context (`InstallationRunning` story):

quickstart-installation-running

fercgomes:
Iterated on the progress chip (`9f0ea453c1c`): it now sits on its own row directly below the project token, flattened to a single line height-matched with the token chip (34px), so the header stays quiet and nothing stretches across the page.

quickstart-v3-wide

PR #70810, opened 2026-07-14
fix(llm-gateway): free-cap posthog_code users with no seat

adboio:
## Problem

A posthog_code user with no seat skips the $75 free-plan cap and falls through to the default user cost limits ($500/day burst, $3k/30d sustained). That carve-out was deliberate when it shipped in #54983: alpha users had no seats yet, and free limits were meant to kick in "automatically when beta launches and seats are created". The alpha is long over, and every real user has had a seat since then, so the carve-out has been dormant.

#70790 wakes it up: with seat creation retired, every new signup lands seatless and inherits roughly 40x the free monthly allowance, at pure inference cost, until the usage-based cutover.

## Changes

- `PlanInfo` gains `seat_missing`, set only when the seats endpoint definitively returns 404 (the user has no seat). It rides the existing Redis plan cache and threads into `ThrottleContext`.
- The free-plan cap now applies when a seat exists **or** the seat is confirmed missing.
- Both the enforcement path (`dependencies.py`) and `/v1/usage` reporting carry the flag, so displayed limits match enforced limits.
- **Review follow-up:** the Django best-seat endpoint's multi-org fan-out used to return 404 when billing lookups *failed*, not just when they conclusively found no seat, which would have poisoned the `seat_missing` cache during a billing blip and free-capped paying users. `fetch_seat` is now transport-only (it returns each org's forwarded billing response), the caller classifies outcomes in one place, and the endpoint returns 503 when nothing was found but any lookup failed. 404 is reserved for "every org checked, no seat anywhere". The gateway needs no change for this, since non-404 errors already fail open and uncached. Found-seat-with-partial-failure still returns the best available seat, as before.

> [!NOTE]
> Plan-resolution **failures** are unchanged: `seat_missing` stays false and unresolved users keep default limits. This preserves the shipped fail-loose behavior deliberately, so a billing-service outage degrades to looser limits instead of hard-blocking paying users at $75 mid-session. The loose path stays bounded by the product-wide cost cap and the 15-minute plan cache.

Cache entries written before the field existed read as `seat_missing=false` and turn over within the cache TTL.

Should deploy BEFORE #70790.

Automated only, full gateway suite green (1200 passed). New coverage and the regression each piece catches:

- Parameterized pair in `test_cost_throttles.py`: confirmed-404 context blocked at the free cap, failure-shaped context allowed at default limits. Catches both a revert of the fix and an accidental flip to fail-tight.
- Cache round-trip test in `test_plan_resolver.py`: the flag survives write then read. Catches caching it on one side only, which would flip-flop a user's limits between cache hit and miss.
- One-line `seat_missing` assertions folded into the existing 404, error, no-url, and legacy-cache-entry tests.
- `test_seat_api.py` best-seat cases: all-lookups-fail flips from asserting 404 to asserting 503 (the bug the review caught), plus partial-failure-with-no-seat returns 503 and all-conclusive-404s still return 404 (the contract `seat_missing` depends on).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal rate-limiting behavior, no public docs affected.

## 🤖 Agent context

Written by Claude in a PostHog Code session, directed by @adboio, as part of the Code usage-based billing transition. We traced the seatless fallback to #54983's PR description before changing it, confirming the loose limits were an alpha-era bridge rather than an outage-protection design, then made fail-loose-on-failure an explicit decision rather than a side effect. The `/writing-tests` skill gated the new tests. Pre-existing mock assignments in `test_plan_resolver.py` needed `ty: ignore[invalid-assignment]` suppressions (the file predates ty in lint-staged); restructuring the file's mocking pattern was rejected as churn.

---

_Created with_ [_PostHog Code_](https://posthog.com/code?ref=pr)

### DIFF
--- a/products/tasks/backend/presentation/views/seat_api.py
+++ b/products/tasks/backend/presentation/views/seat_api.py
@@ -282,9 +282,6 @@ class SeatViewSet(viewsets.ViewSet):
             return Response(data)
 
         def fetch_seat(org: Organization) -> tuple[Organization, Response]:
-            """Per-org billing lookup, transport only — the caller classifies the
-            response. Header-build failure becomes the same 502 an unreachable
-            billing service produces."""
             headers = self._get_billing_headers_for_org(user, org)
             resp = (
                 self._billing_request("GET", f"/api/v2/seats/{distinct_id}/", headers, query_params=query_params)
@@ -307,8 +304,6 @@ class SeatViewSet(viewsets.ViewSet):
                 if 200 <= drf_resp.status_code < 300 and isinstance(drf_resp.data, dict):
                     results.append((org, drf_resp.data))
                 elif drf_resp.status_code != status.HTTP_404_NOT_FOUND:
-                    # 404 = billing conclusively found no seat in this org; anything
-                    # else means the org couldn't be checked.
                     logger.warning("fetch_seat_failed", org_id=str(org.id), status_code=drf_resp.status_code)
                     any_lookup_failed = True
 
--- a/products/tasks/backend/presentation/views/seat_api.py
+++ b/products/tasks/backend/presentation/views/seat_api.py
@@ -281,29 +281,40 @@ class SeatViewSet(viewsets.ViewSet):
             data = {**drf_resp.data, "organization_id": str(org.id), "organization_name": org.name}
             return Response(data)
 
-        def fetch_seat(org: Organization) -> tuple[Organization, dict[str, Any]] | None:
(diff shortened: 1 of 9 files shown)


brandonleung (approved):

adboio:
### Merge activity

* **Jul 14, 6:32 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/70810).

PR #67117, opened 2026-06-30
feat(eng-analytics): emit CI signals to Signals

rnegron:
> [!NOTE]
> Stacked on [feat(signals): ci signal taxonomy and emit plumbing](https://github.com/PostHog/posthog/pull/70766) — merge that first; this PR then auto-retargets to master.

## Problem

PostHog Code can't see what happens to a PR between opened and deployed. Engineering analytics owns the curated CI/PR read layer; this turns its highest-value CI conditions into Signals the self-driving inbox can group, research, and autonomously fix.

## Changes

```mermaid
flowchart LR
    W[GitHub warehouse tables] --> D{{CI detectors}}
    S([Hourly Temporal coordinator]) --> D
    D --> E[Signals facade emit_signal]
    E --> I[Inbox: group, research, auto-start PR]
    classDef phBlue fill:#1d4aff,stroke:#1d4aff,color:#fff;
    classDef phYellow fill:#f9bd2b,stroke:#f9bd2b,color:#000;
    classDef phGray fill:#e5e7eb,stroke:#c7ccd1,color:#000;
    class D,E phBlue;
    class W phGray;
    class S,I phYellow;
```

- Three deterministic detectors composing the curated queries: `ci_flaky_check` (job failed then passed on a rerun of the same commit), `ci_broken_default_branch` (default branch red), `ci_duration_regression` (p95 up both relative and absolute)
- Hourly coordinator fans out per enrolled team; deterministic workflow ids make emission idempotent; inert until a team opts in
- [Multi-repo GitHub sources](https://github.com/PostHog/posthog/pull/71450) are scanned per synced repo — every repo contributes findings, and dedupe keys are repo-qualified so repos never collide in the ledger
- The sweep is userless, so enrollment alone never authorizes reads: enabling snapshots the requesting user's accessible GitHub source ids, and the coordinator scans only that snapshot under the enabler's live access control (source deletion, access revocation, org removal, a deactivated enabler, and none-level warehouse access all fail closed)
- Flag-gated "GitHub CI" roster card, inbox signal card, one-toggle config endpoint
- Inbox source setup re-keyed by signal source (one `WAREHOUSE_SOURCE_SETUP` registration per source) — GitHub now backs two sources, so the old product-keyed maps and the duplicated `DataSourceSetup` copies collapse into one registry and one shared component

> [!CAUTION]
[...]

PR #70825, opened 2026-07-14
fix(eng-analytics): fail closed on none-level source access

rnegron:
## Problem

We're about to point an hourly, userless sweep at GitHub warehouse data ([CI signals](https://github.com/PostHog/posthog/pull/67117)), which forced the question of who is actually allowed to read which sources. Auditing that turned up a gap that's live on master today, independent of the sweep:

- `filter_queryset_by_access_level` returns the queryset **unfiltered** for a user with "none" `external_data_source` access and no object grants
- Engineering analytics resolves GitHub sources through it, so such a user can enumerate every source on the team

## Changes

- `_github_sources` (the product's access chokepoint): callers without resource-level access see only self-created or explicitly granted sources
- Root cause is the platform filter's none-branch; flagged for the RBAC owners

- New test: a none-level member sees only their own source, and an explicit object grant survives — catches the unfiltered passthrough regressing (needs the `ACCESS_CONTROL` feature, matching how it manifests)
- Engineering-analytics backend suite green

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Extracted by Claude Code from a security sweep of the CI signals stack.

### DIFF
--- a/products/engineering_analytics/backend/logic/sources.py
+++ b/products/engineering_analytics/backend/logic/sources.py
@@ -20,9 +20,10 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING
 from uuid import UUID
 
-from django.db.models import QuerySet
+from django.db.models import Q, QuerySet
 
 from posthog.models.team import Team
+from posthog.rbac.user_access_control import NO_ACCESS_LEVEL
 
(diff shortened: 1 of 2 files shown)


gantoine (approved):
nice

PR #70830, opened 2026-07-14
refactor(inbox): key warehouse source setup by signal source

rnegron:
## Problem

**Adding a signal source should be one entry. Today it's four product-keyed structures plus two copy-pasted `DataSourceSetup` components.**

That shape assumes one warehouse product means exactly one signal source, which breaks the moment GitHub backs a second ([CI signals](https://github.com/PostHog/posthog/pull/67117)). It also already ships a bug: the two copies drifted, so the pganalyze wizard pre-selects `issues`+`servers` while sync only forces `issues`.

## Changes

Before:

```mermaid
flowchart LR
    S[signal source] --> P{{warehouse product}}
    P --> A[DATA_WAREHOUSE_SOURCE_CONFIG]
    P --> B[SIGNAL_SOURCE_REQUIRED_TABLES]
    P --> C[agentRosterMeta.dataWarehouseSource]
    P --> D[inline completion map]
    A --> X[DataSourceSetup in AgentsTab]
    D --> Y[DataSourceSetup in SignalSourcesPanel]
    class S phYellow;
    class P phRed;
    class A,B,C,D phGray;
    class X,Y phBlue;
```

After:

```mermaid
flowchart LR
    S[signal source] --> R[WAREHOUSE_SOURCE_SETUP]
    R --> X[DataSourceSetup]
    class S phYellow;
    class R phGray;
    class X phBlue;
```

- Keyed by signal source, not warehouse product, so one product can back several. Adding a source becomes one registry entry
- pganalyze drift fixed; unused roster-meta field removed
- Enabling an already-connected source now syncs every repo's rows. [Multi-repo GitHub](https://github.com/PostHog/posthog/pull/71274) names rows `owner/repo.issues`, which the exact-name match missed, so the source sat there not syncing. New `schemasToEnableFor`

> [!WARNING]
> First-time setup is still broken for qualified sources, and is **not** fixed here. `sourceWizardLogic` matches `requiredTables` against `schema.table` exactly ([L1520](https://github.com/PostHog/posthog/blob/a880dc83ca5384b7fe793e2f8d596affed2fe97d/products/data_warehouse/frontend/scenes/NewSourceScene/sourceWizardLogic.tsx#L1520)), so connecting GitHub from the inbox fails with `Required tables not found in source: issues`. That match belongs to the data warehouse wizard.

> [!NOTE]
> Backend half of the same naming bug: [the signals qualified-schema PR](https://github.com/PostHog/posthog/pull/71553). Independent, no file overlap, either can merge first.

- Jest: 66 inbox tests. `signalSourcesLogic.test.ts` covers `schemasToEnableFor` — catches a multi-repo source resolving to nothing, and only the first repo being enabled. Nothing covered qualified names before
- kea typegen + tsc clean; the re-key itself is pure re-wiring, no behavior change beyond the pganalyze fix
- Merged master (246 commits). One conflict, where [anomaly investigation](https://github.com/PostHog/posthog/pull/71533) added a listener beside the re-keyed one. Kept both, then checked every renamed symbol for stale callers

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Extracted by Claude Code from the CI signals stack; the stack absorbs this on merge.

The multi-repo fix landed here rather than in its own PR because this PR already rewrites that exact call site. A review pass found the wizard gap called out above; an earlier draft of this description wrongly implied setup was covered.


rnegron:
superseded by https://github.com/PostHog/posthog/pull/67117

PR #70831, opened 2026-07-14
feat(workflows): retry push sends on transient provider errors

dmarchuk:
## Problem

Push sends had no resilience against provider throttling.
A single failed channel send was thrown straight to a hard failure, so a large batch run (say a workflow triggered for 100k people) could silently drop notifications the moment the customer's FCM project or APNs topic started rate-limiting.

Two facts make this bite Android in particular:

- Push uses each customer's **own** credentials (their FCM service account, their APNs `.p8`), so a rate limit is scoped to that customer's project/topic. Their own blast can throttle themselves.
- FCM enforces a hard per-project quota (default 600k messages/min) and returns `429 QUOTA_EXCEEDED` with a `Retry-After`. We ignored it. APNs is more forgiving, but the same gap applies to its `429` and to any `5xx`/network blip.

Email already handles this (the SES throttle-reschedule path). Push never got the equivalent, so it just dropped.

## Changes

The push send path now distinguishes a **transient** failure from a **terminal** one and reschedules instead of dropping.

- **Retriability comes from the error normalizer, not the HTTP layer.** `push-notification-errors.ts` already normalizes every provider failure into a `PushFailureReason`. A new `REASON_RETRIABLE` map marks `rate_limited` / `provider_error` / `network_error` as retriable and everything else (`auth_error`, `invalid_token`, `invalid_payload`, `unregistered`, `unknown`) as terminal. `PushSendError` now carries that `retriable` flag plus the parsed `retryAfterMs`, so the send loop decides purely from the normalized reason and never re-parses a raw response.
- **Reschedule instead of drop.** When nothing was delivered and a channel failed transiently, the whole invocation is rescheduled (`finished = false` + `queueScheduledAt`) rather than failed. The delay honors the provider's `Retry-After` when present (bounded by `MAX_RETRY_AFTER_MS = 5m`), else exponential backoff with jitter. Capped by `CDP_FETCH_RETRIES`, so a persistently-throttled project fails terminally rather than looping forever.
- **Only reschedule when no channel has delivered yet** — otherwise a retry would re-deliver to a channel that already sent. When several channels throttle at once, the reschedule waits out the **longest** `Retry-After` across them, since the retry re-attempts every failed channel.
- **The P1 fix:** the reschedule restores `queueParameters` on the result. `createInvocationResult` clears them, and without restoring, the retry dequeue routes on an undefined queue type, resumes the hog VM instead of re-running the send, pops an empty stack, and drops the notification. There's a round-trip test that feeds the rescheduled invocation back through the executor and asserts the send actually re-runs.
- **Business metrics are deferred to the terminal outcome.** `push_sent` / `push_skipped` / `push_failed` are emitted once when the invocation finishes, with real counts, rather than per-channel on every attempt — so a rescheduled retry doesn't inflate the per-notification counts.
- **New `push_notification_rescheduled_total` counter** (by platform), incremented once per distinct throttled platform. A sustained rate means that customer's project is being rate-limited.
- **Budget wiring:** `PushNotificationFetchUtils` gains `maxRetries` / `backoffBaseMs` / `backoffMaxMs`, wired from the existing `CDP_FETCH_*` settings in `cdp-services.ts` and `hog-transformer.service.ts`, so push shares one retry policy with the generic fetch destinations rather than inventing its own.

Dead-token pruning (FCM `UNREGISTERED`, APNs `410`) is unchanged: still a skip, not a retry.

> [!NOTE]
> This was reconciled onto master after the observability PR (#70913) merged. The retriability signal is derived from that PR's normalizer, so there's no `hog-executor` / `request.ts` churn and no shared fetch-retry module move.

Added and updated Jest cases in `push-notification.service.test.ts`, each aimed at a specific regression this PR must not reintroduce:

- a non-retriable `4xx` fails terminally with no reschedule
- a transient `5xx` reschedules, and the **round-trip** case feeds the rescheduled invocation back in and asserts the send re-runs and reports `push_sent` (this is the P1 `queueParameters` guard)
- retry-budget exhaustion (`tries == maxRetries`) stops rescheduling and fails terminally
- a skipped channel alongside a transient failure reschedules **without** re-counting the skip on the retry attempt (deferred-metrics guard)
- two channels throttling reschedule using the **longest** `Retry-After` (a `429` with `retry-after: 30` parks ~30s out, not FCM's sub-2s backoff)

I (Claude) can't fully run the nodejs toolchain in this sandbox, but CI runs the full `push-notification.service.test.ts` suite and the Node.js test matrix — both are green on the current head. I audited the `PushSendError` constructor call sites and the `PushNotificationFetchUtils` consumers after widening both, so the added fields have no stranded caller.

Scope note on the retry path: the transient-retry behavior can only be exercised by mocking the provider HTTP boundary — the FCM/APNs send URL isn't overridable, so a real provider `429` can't be forced on demand against a live stack. The Jest cases drive the real service with that boundary mocked, which is the faithful level for this logic.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change — this is an internal delivery-reliability improvement with no API or config surface (the retry budget reuses the existing `CDP_FETCH_*` settings).

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Directed by the requester (session owner), who is the PR assignee and DRI.

Authored with Claude Code. This came out of a scalability investigation into what happens when a 100k batch run sends push through APNs vs FCM — the finding was that email got throttle handling and push didn't. Invoked the `/writing-tests` skill before writing the tests to keep them aimed at real regressions.

Decisions along the way: chose reactive `Retry-After`/backoff handling as the first increment (self-contained, mirrors the proven email path) over a proactive per-integration token bucket (more infra, deferred). A first attempt relocated shared retry primitives into a new `fetch-retry.ts` module; that was dropped during reconciliation onto master in favor of deriving retriability from the observability PR's normalizer reason, which removes the module move and the `hog-executor` churn entirely. The reschedule is gated on "nothing delivered yet" specifically to preserve the no-double-delivery invariant the multi-channel fan-out relies on.

### DIFF
--- a/nodejs/src/cdp/services/messaging/push-notification.service.ts
+++ b/nodejs/src/cdp/services/messaging/push-notification.service.ts
@@ -1,4 +1,5 @@
 import { createHash, createSign } from 'crypto'
+import { DateTime } from 'luxon'
 import { Counter, Histogram } from 'prom-client'
 
 import {
@@ -55,6 +56,12 @@ const pushNotificationSendDurationMs = new Histogram({
     buckets: [10, 25, 50, 100, 200, 500, 1000, 2000, 5000, 10000],
 })
 
+const pushNotificationRescheduledCounter = new Counter({
+    name: 'push_notification_rescheduled_total',
+    help: 'Push sends rescheduled after a transient provider failure (throttle / 5xx / network) instead of being dropped. A sustained rate for one platform means that provider is rate-limiting the sending project.',
+    labelNames: ['platform'],
+})
+
 // Apple rate-limits new APNs provider tokens (returns 429 TooManyProviderTokenUpdates if refreshed more
 // than once every ~20 min per key) and accepts a token for up to 1 hour. Cache the signed JWT in Redis
 // keyed by the auth key id so the whole fleet reuses one token per key rather than minting one per send.
@@ -66,15 +73,40 @@ const APNS_JWT_TTL_SECONDS = 45 * 60
 // push targets a handful of provider integrations, not thousands.
 const MAX_PUSH_CHANNELS = 10
 
+// Cap a provider-supplied Retry-After. FCM returns one on QUOTA_EXCEEDED and honoring it is the point,
+// but a hostile or misconfigured value must not park an invocation for hours.
+const MAX_RETRY_AFTER_MS = 5 * 60 * 1000
+
 function stringifyBody(body: unknown): string {
     return typeof body === 'string' ? body : JSON.stringify(body)
 }
 
-function pushSendError(platform: PushPlatform, err: NormalizedPushError): PushSendError {
+// Retry-After is delta-seconds or an HTTP-date. Return a bounded millisecond delay, or undefined if the
+// header is absent or unparseable so the caller falls back to exponential backoff.
+function parseRetryAfterMs(response: FetchResponse | null): number | undefined {
+    const header = response?.headers?.['retry-after']
+    if (!header) {
+        return undefined
+    }
+    const seconds = Number(header)
+    const ms = Number.isFinite(seconds) ? seconds * 1000 : Date.parse(header) - Date.now()
+    if (!Number.isFinite(ms) || ms <= 0) {
+        return undefined
+    }
+    return Math.min(ms, MAX_RETRY_AFTER_MS)
+}
+
+// Exponential-ish backoff with jitter for a retry when the provider didn't give a Retry-After.
+function backoffAt(baseMs: number, maxMs: number, tries: number): DateTime {
+    const delay = Math.min(baseMs * tries + Math.floor(Math.random() * baseMs), maxMs)
+    return DateTime.utc().plus({ milliseconds: delay })
+}
+
+function pushSendError(platform: PushPlatform, err: NormalizedPushError, retryAfterMs?: number): PushSendError {
     // Append the raw provider code so the failure surfaced to the hog template stays debuggable, while
     // the human-readable sentence leads.
     const message = err.code ? `${err.message} [${err.code}]` : err.message
-    return new PushSendError(message, platform, err.reason, err.level, err.code)
+    return new PushSendError(message, platform, err.reason, err.level, err.retriable, err.code, retryAfterMs)
 }
 
 export type PushNotificationFetchUtils = {
@@ -84,6 +116,11 @@ export type PushNotificationFetchUtils = {
         fetchDuration: number
     }>
     maxFetchTimeoutMs: number
+    // Retry budget + backoff for transient provider failures, wired from CDP_FETCH_* so push shares the
+    // same policy as the generic fetch destinations.
+    maxRetries: number
+    backoffBaseMs: number
+    backoffMaxMs: number
 }
 
 export class PushNotificationService {
@@ -106,14 +143,24 @@ export class PushNotificationService {
         const result = createInvocationResult<CyclotronJobInvocationHogFunction>(invocation, {}, { finished: true })
         const addLog = createAddLogFunction(result.logs)
 
-        const pushMetric = (metricName: 'push_sent' | 'push_skipped' | 'push_failed'): void => {
+        // How many times this send has already been rescheduled after a transient failure, carried on the
+        // invocation's queue metadata across reschedules.
+        const metadata = (invocation.queueMetadata as { tries?: number }) || {}
+        const tries = metadata.tries ?? 0
+
+        // Business metrics are emitted once, at the terminal outcome below, rather than per channel — a
+        // rescheduled attempt must not re-count the same notification's skips or failures on every retry.
+        const pushMetric = (metricName: 'push_sent' | 'push_skipped' | 'push_failed', count: number): void => {
+            if (count <= 0) {
+                return
+            }
             result.metrics.push({
                 team_id: invocation.teamId,
                 app_source_id: invocation.parentRunId ?? invocation.functionId,
                 instance_id: invocation.state.actionId || invocation.id,
                 metric_kind: 'push',
                 metric_name: metricName,
-                count: 1,
+                count,
             })
         }
 
@@ -132,8 +179,10 @@ export class PushNotificationService {
             )
         }
 
-        let errorCount = 0
         let successCount = 0
+        let skippedCount = 0
+        let otherErrorCount = 0
+        const errors: PushSendError[] = []
         let firstError: string | undefined
         for (const integrationId of integrationIds) {
             try {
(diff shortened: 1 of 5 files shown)


nodejs/src/cdp/services/messaging/push-notification.service.ts
```diff
@@ -124,15 +176,49 @@ export class PushNotificationService {
...
-        // Retry (by surfacing an error) only when a channel was attempted and failed and nothing was
-        // delivered. A skip (no device token) is not a delivery, so retrying is safe; but once any
-        // channel has delivered we must not retry, or it would re-deliver to that channel.
         const success = successCount > 0 || errorCount === 0
+
+        // Nothing delivered and a channel failed for a transient reason (provider throttle, 5xx, network
+        // blip): reschedule the whole invocation with backoff instead of dropping the notification. Honor
+        // the provider's Retry-After when present, else back off exponentially, capped by maxRetries. Only
+        // safe when nothing was delivered — once a channel has sent, retrying would re-deliver to it.
+        if (!success && successCount === 0 && retriableErrors.length > 0 && tries < this.fetchUtils.maxRetries) {
```
mayteio: I traced the reschedule round trip and I believe this branch never actually re-attempts the send.

`createInvocationResult(invocation, {}, ...)` goes through `cloneInvocation`, which resets `queueParameters` to `undefined` (see `nodejs/src/cdp/utils/invocation-utils.ts`), and this branch never restores it. When the parked invocation is dequeued again, `executeWithAsyncFunctions` routes on `queueParameters?.type` (`hog-executor.service.ts`), finds `undefined`, and falls through to `execute()`, which resumes the hog VM instead of re-entering this service.

That resume is the failure mode: the VM was suspended with `ip` already advanced past the async call (`common/hogvm/typescript/src/execute.ts`, the `frame.ip += 1` branch in `CALL_GLOBAL`), so it expects the host to have pushed the send result onto the stack first, which this branch deliberately skips. The retry dequeue therefore pops an empty stack and throws `HogVMException: Invalid HogQL bytecode, stack is empty, can not pop`.

Net effect: the notification is still dropped, now after the backoff delay, the workflow fails with a bytecode error instead of the push error, and no `push_failed` metric is ever emitted for it.

The fetch retry path handles this exact case by restoring the params before rescheduling, see `result.invocation.queueParameters = params` in `executeFetch` (`hog-executor.service.ts`). The equivalent here should be `result.invocation.queueParameters = invocation.queueParameters` in this branch. It also needs a round-trip test to prove it: feed the rescheduled `result.invocation` back through the executor (or at minimum assert `result.invocation.queueParameters?.type === 'sendPushNotification'`) and check the send is re-attempted. The current tests only assert the shape of the first result, so they cannot catch this.

Worth noting: the SES throttle path in `email.service.ts` that this mirrors appears to have the same latent gap (it also clears `queueParameters` and skips the stack push, and has no round-trip test). Email mostly avoids it because the Valkey bucket gates at dequeue and that path is a rarely-hit safety net, but here it is the primary mechanism. Probably worth a follow-up for email either way.

mayteio: Two smaller notes on this condition:

- If one channel fails terminally (e.g. a 400) while another fails transiently (429), the whole invocation reschedules and the known-terminal channel is re-attempted on every retry. Possibly acceptable, but worth a code comment so the behavior reads as intentional.
- `tries < maxRetries` with `tries` starting at 0 allows `maxRetries` reschedules after the initial attempt, i.e. `maxRetries + 1` total attempts. The fetch path counts the initial attempt within `maxRetries` (`attempts` is incremented before the `attempts < maxRetries` check), so the two paths interpret `CDP_FETCH_RETRIES` slightly differently.

Nit: `!success && successCount === 0` is redundant, `!success` already implies `successCount === 0`.

nodejs/src/cdp/services/messaging/push-notification.service.ts
```diff
@@ -124,15 +176,49 @@ export class PushNotificationService {
...
+            // burning the retry budget without delivering.
+            const retryAfterValues = retriableErrors
+                .map((e) => e.retryAfterMs)
+                .filter((ms): ms is number => ms !== undefined)
+            result.invocation.queueScheduledAt =
+                retryAfterValues.length > 0
+                    ? DateTime.utc().plus({ milliseconds: Math.max(...retryAfterValues) })
+                    : getNextRetryTime(this.fetchUtils.backoffBaseMs, this.fetchUtils.backoffMaxMs, nextTries)
+            result.finished = false
+            pushNotificationRescheduledCounter.labels({ platform: retriableErrors[0].platform }).inc()
```
mayteio: If both FCM and APNs fail transiently in the same invocation, only `retriableErrors[0].platform` is counted. Since the counter's stated purpose is spotting which provider is rate-limiting, consider incrementing once per distinct platform in `retriableErrors`.

nodejs/src/cdp/services/messaging/push-notification.service.ts
```diff
@@ -124,15 +176,49 @@ export class PushNotificationService {
...
+            addLog(
+                'warn',
+                `Push notification throttled or temporarily unavailable; retrying (attempt ${nextTries}/${this.fetchUtils.maxRetries}).`
+            )
+            // Skip the failure metric and the VM-state push — the eventual retry reports the real outcome.
+            return result
```
mayteio, line 260: This early return keeps everything already accumulated in `result.metrics`, and skipped channels emitted `push_skipped` earlier in the loop. A recipient with one skipped channel and one throttled channel therefore re-emits `push_skipped` on every attempt, up to `maxRetries + 1` times for a single notification. Consider suppressing the skip metric on rescheduled attempts (or emitting it only on the terminal attempt), so the counts stay per-notification rather than per-attempt.

nodejs/src/common/utils/fetch-retry.ts
```diff
@@ -0,0 +1,42 @@
+import { DateTime } from 'luxon'
+
+import { FetchResponse, InvalidRequestError, SecureRequestError } from './request'
```
mayteio: Small caveat on the "survives the mock" rationale: this module still imports `SecureRequestError` / `InvalidRequestError` from the mocked `request` module, and the shared mock (`nodejs/tests/helpers/mocks/request.mock.ts`) only defines `fetch` / `internalFetch`. Under that mock these classes are `undefined`, so `error instanceof SecureRequestError` throws a TypeError the moment a test calls `isFetchResponseRetriable` with a non-null error. It holds today only because all current tests pass `fetchError: null`. Consider having the mock `jest.requireActual` the error classes so the full retry-classification path is exercisable under it.

mayteio (commented):
Took a close look at the retry/reschedule path. The main concern (detailed inline on the reschedule branch) is that the rescheduled invocation loses its `queueParameters`, so the retry dequeue resumes the hog VM instead of re-entering the push service, and the notification is still dropped. The remaining inline comments are smaller: metric double-counting on retried skips, the reschedule counter only labeling the first platform, retry-budget semantics vs the fetch path, and a caveat on the fetch-retry module's test-mock rationale.

The Retry-After handling itself (bounded parsing, taking the max across channels, falling back to backoff) and the config wiring look good.

dmarchuk:
Thanks for the thorough review @mayteio — the `queueParameters` trace was spot on, that would have dropped every rescheduled notification with a bytecode error. All points addressed. Since #70913 (observability) merged in the meantime, I also rebased this onto master and rebuilt the retry on top of it, so the inline threads are now outdated — force-pushed as a single commit. Mapping:

- **Reschedule lost `queueParameters` → retry resumed the hog VM.** Fixed: `result.invocation.queueParameters = invocation.queueParameters` before rescheduling, mirroring `executeFetch`. Added a **round-trip test** that feeds the rescheduled `result.invocation` back through the service and asserts the send actually re-runs (and `queueParameters?.type === 'sendPushNotification'`).
- **`push_skipped` re-emitted every retry.** Business metrics (`push_sent`/`push_skipped`/`push_failed`) are now emitted **once at the terminal outcome**, never on a rescheduled attempt. Test asserts a rescheduled attempt carries no `push_skipped`/`push_failed`.
- **Counter labeled only the first platform.** Now increments once per distinct platform in the retriable set.
- **Longest Retry-After.** Already taking the max across channels; kept, with a test (FCM 500 + APNs 429 `Retry-After: 30` → ~30s, not the sub-2s backoff).
- **Terminal channel re-attempted / `maxRetries + 1` semantics / redundant `successCount === 0`.** Added the clarifying comments and dropped the redundant guard.
- **fetch-retry mock caveat.** Moot now — the reconciliation derives retriability from the normalized reason instead of `isFetchResponseRetriable`, so there's no `fetch-retry.ts` and no `SecureRequestError`/`InvalidRequestError` under the mock. The diff is down to the push module (no `hog-executor`/`request.ts` churn).

You also flagged the same latent gap in the SES throttle path (email clears `queueParameters` too, but the Valkey bucket makes it a rarely-hit safety net) — agreed it's worth a follow-up; I'll leave that out of this PR's scope.

---

nodejs/src/cdp/services/messaging/push-notification.service.ts
dmarchuk: Done, the counter now increments once per distinct platform in `retriableErrors` (see the `new Set(...)` loop)

nodejs/src/cdp/services/messaging/push-notification.service.ts
dmarchuk: Good catch! Now it's emitted only after the reschedule early-return, so it fires once on the terminal attempt

dmarchuk:
@mayteio Thanks for the review here! 🙏 Seems Claude posted a response above, which seems right to me. Let me know in case there's anything else 🙏

mayteio (approved):
SEND IT :D

PR #70085, opened 2026-07-10
perf(dashboards): reduce database queries on dashboard load

fercgomes:
## Problem

Every dashboard load issues more Postgres queries than necessary on the read path. On each `GET /api/projects/:id/dashboards/:id` (and `stream_tiles`):

- `last_accessed_at` is written with `instance.save()`, which fires the `FileSystemSyncMixin` `post_save` signal and re-syncs the (unchanged) filesystem entry on every read — an extra SELECT + UPDATE + shortcut lookup.
- `last_accessed_at` is written on *every* read, even for dashboards that are polled/auto-refreshed many times a minute.
- The team's `InsightVariable` rows are loaded even when the dashboard and its insights use no variables (the common case).

These are fixed per-request costs sitting on the dashboard-load critical path.

## Changes

- Bump `last_accessed_at` with a signal-free queryset `.update()` (scoped by pk) instead of `instance.save()`, so a read no longer triggers a FileSystem re-sync. The filesystem representation never depends on `last_accessed_at`.
- Throttle the `last_accessed_at` write to a 60s window: repeat/auto-refresh loads within the window skip the write entirely. It only drives "recently viewed" ordering, which doesn't need per-request precision; the first load still writes.
- Skip the `InsightVariable` lookup when no variables are involved, by short-circuiting `map_stale_to_latest` and passing the already-cached queryset instead of eagerly materialising it. Dashboards/insights that *do* use variables still load them once (verified no N+1).

Measured on a 10-tile dashboard GET: **28 → 22 Postgres queries** (~21% fewer), and the count stays flat as tile count grows (no N+1). Both the classic retrieve path and `stream_tiles` benefit.

## How did you test this code?

- Updated the canonical `test_adding_insights_is_not_nplus1_for_gets` query-count assertions (20→15 empty, 31→25 with insights) and regenerated the affected `@snapshot_postgres_queries` snapshots (`test_dashboard.ambr`, `test_insight.ambr`).
- Ran the affected dashboard/insight query-count and snapshot tests against a locally provisioned Postgres + ClickHouse; they pass, including a second assert-mode run to confirm the snapshots are deterministic.
- Separately verified that a dashboard *with* variables still fires the `InsightVariable` query exactly once (flat at 10 and 20 tiles, no N+1) and that `last_accessed_at` is still set on first load.
- I (the agent) did not run the full CI suite; only the tests directly affected by these changes.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

[...]

PR #70832, opened 2026-07-14
fix(dashboards): prefetch alert threshold and subscribers on read paths

thmsobrmlr:
## TL;DR (eli5)

When a dashboard has alerts on its insights, opening it made the server ask the database two extra questions for *every single alert* (who's subscribed? what's the threshold?). A dashboard with lots of alerted insights asked hundreds of tiny questions one by one. Now it asks once for all of them upfront — same answer, a fraction of the round trips.

## Problem

`AlertSerializer` serializes `threshold` and `subscribed_users` for every alert, but the alert prefetch on dashboard GET (`DashboardSerializer.get_tiles`) and insight list only does `select_related("created_by")`. Each alert on the page costs two extra Postgres queries, so alert-heavy dashboards degrade to O(alerts) queries per load. Found while benchmarking dashboard loads against real dashboards: an alert-per-insight dashboard paid ~3 extra queries per tile, dwarfing its fixed query cost.

**Why:** follow-up from the query benchmarking of #70085 — the biggest per-tile query multiplier on dashboard load turned out to be alerts, not the fixed costs that PR addresses.

## Changes

- Both alert prefetch sites now use `select_related("created_by", "threshold")` + `prefetch_related("subscribed_users")`.
- Measured on a 30-tile dashboard with two alerts per insight (one threshold, one detector, one subscriber each): **176 → 27 queries, wall time −72%**; query count is now flat as tile count grows.

- `test_insight_alerts_do_not_nplus1_dashboard_gets` (dashboard GET) and `test_listing_insights_with_alerts_does_not_nplus1` (insight list): constant-query-count tests that create insights with threshold + detector alerts and assert the per-request query count doesn't grow with more insights. Each catches the regression of dropping `threshold`/`subscribed_users` from its endpoint's prefetch — no existing test creates alerts on dashboard/listed insights, so this N+1 was invisible to the suite. Both fail on the previous prefetch and pass with this change (verified locally by stashing the fix).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change; no docs affected.

## 🤖 Agent context

Authored with PostHog Code (Claude). Skills invoked: /improving-drf-endpoints, /writing-tests. The N+1 was located by benchmarking dashboard GETs per-commit for #70085, then diffing captured SQL against production dashboards; the repeating statements were per-alert `posthog_user` (subscribed_users, serialized twice per alert) and `posthog_threshold` selects. Fix validated with a standalone query-count benchmark (176 → 27 queries at 30 tiles) and the two new regression tests, run locally against Postgres + ClickHouse.

---

### DIFF
--- a/posthog/api/test/dashboards/__snapshots__/test_dashboard.ambr
+++ b/posthog/api/test/dashboards/__snapshots__/test_dashboard.ambr
@@ -2260,9 +2260,17 @@
          "posthog_user"."onboarding_delegated_to_organization_id",
          "posthog_user"."onboarding_delegation_accepted_at",
          "posthog_user"."events_column_config",
-         "posthog_user"."email_opt_in"
+         "posthog_user"."email_opt_in",
+         "posthog_threshold"."created_by_id",
+         "posthog_threshold"."created_at",
+         "posthog_threshold"."id",
+         "posthog_threshold"."team_id",
+         "posthog_threshold"."insight_id",
+         "posthog_threshold"."name",
+         "posthog_threshold"."configuration"
   FROM "posthog_alertconfiguration"
   LEFT OUTER JOIN "posthog_user" ON ("posthog_alertconfiguration"."created_by_id" = "posthog_user"."id")
+  LEFT OUTER JOIN "posthog_threshold" ON ("posthog_alertconfiguration"."threshold_id" = "posthog_threshold"."id")
   WHERE "posthog_alertconfiguration"."insight_id" IN (1,
                                                       2,
                                                       3,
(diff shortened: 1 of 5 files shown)


thmsobrmlr:
## Per-change benchmark of this fix

*(Comment posted by PostHog Code — benchmarked each half of the prefetch change separately.)*

Same harness as before: local PG, dashboard GET with 30 tiles, 2 alerts per insight (one threshold + one detector, 1 subscriber each → 60 alerts, 30 thresholds). Median of 19 requests.

| Variant | Queries | Wall ms |
|---|---|---|
| unfixed (`created_by` only) | 176 | 1388 |
| + `select_related("threshold")` only | 146 (**−30**) | ~1390 (~0) |
| + `prefetch_related("subscribed_users")` only | 57 (**−119**) | 435 (**−69%**) |
| both (this PR) | **27** | **427** |

Cost of the heavier prefetch statement itself (60 alerts, statement-level micro-benchmark): the threshold LEFT JOIN takes the alert prefetch from 3.5ms → 4.7ms; adding the `subscribed_users` prefetch (one extra batched query) brings the pair to ~17ms — replacing 149 individual queries.

Reading: the `subscribed_users` prefetch is the dominant win — each alert was fetching the full ~60-column user row **twice**. The threshold join contributes no local wall time (each eliminated lookup was a ~0.1ms PK hit) but removes 1 query per threshold alert; over a real network round trip to the primary that's meaningful latency and connection load, and it costs +1.2ms of JOIN on the prefetch statement. Both halves keep the query count flat as tiles grow.

PR #70834, opened 2026-07-14
chore(pipeline): render app metrics charts via quill-charts directly

sampennington:
## Problem

`AppMetricSummary`, `AppMetricsTrends`, and the hog flow editor metrics panel render app metrics time series through the SQL insights chart dispatcher (`DataVisualization`'s `LineGraph`), wrapping simple `{labels, series}` data in SQL `AxisSeries` column metadata just to reach the quill `TimeSeriesLineChart` underneath. That chart is meant for SQL insights only, and these external consumers make cleaning up its legacy paths (see [#70675](https://github.com/PostHog/posthog/pull/70675)) harder than it should be.

**Why:** keep the SQL chart scoped to SQL insights so its flag/legacy cleanup can proceed; app metrics surfaces should consume `@posthog/quill-charts` directly like other product charts do.

## Changes

- New shared `AppMetricsTimeSeriesChart` in `lib/components/AppMetrics/` rendering `AppMetricsTimeSeriesResponse` on quill's `TimeSeriesLineChart`, using the app-level `useChartTheme`/`useChartConfig` hooks (same pattern as other direct quill chart consumers).
- `AppMetricSummary` (compact tile: axes/grid hidden, single pinned color), `AppMetricsTrends` (legend + label overrides), and workflows' `HogFlowEditorPanelMetrics` (legend + per-metric label/color) now use it instead of the SQL chart.

This covers every app-metrics surface (data pipelines, hog functions, batch exports, data warehouse, workflows) since they all go through these two shared components.

- `pnpm --filter=@posthog/frontend typescript:check` — no new errors vs master (remaining whole-app errors are the pre-existing missing-kea-typegen noise of a fresh checkout; the two implicit-any hits in `HogFlowEditorPanelMetrics.tsx` exist identically on master).
- `oxlint`/`oxfmt` clean on changed files.
- No automated tests added: this is a like-for-like rendering swap with no logic change, and the quill chart behavior is covered by the charts package's own tests. I was not able to verify visually in a running app.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A — internal chart rendering only.

## 🤖 Agent context

Authored with PostHog Code (Claude Code). Part of scoping the DataVisualization chart to SQL insights only: an audit found six consumers outside SQL insights; this PR migrates the app-metrics family (the other areas follow in separate PRs). Decisions: one shared component rather than three inline charts, since all three render the same `AppMetricsTimeSeriesResponse`; labels stay pre-formatted from `appMetricsLogic`, so the x-axis tick callback parses them as naive local strings; tooltip keeps the previous pin/sort behavior from the SQL chart path.

---


mayteio:
@sampennington got a screenshot you can share? :D

sampennington:
> @sampennington got a screenshot you can share? :D

Here we go:
Screenshot 2026-07-15 at 11 47 32

Should be identical to the new charts currently rolled out at 100%. Just want to stop other products importing the charts from insights

sampennington:
We should consider using the Metric components for the top charts there e.g. 
Screenshot 2026-07-15 at 11 50 02

mayteio (approved):
LGTM 🙏🏻

PR #70835, opened 2026-07-14
chore(endpoints): render usage trends chart via quill-charts directly

sampennington:
## Problem

`EndpointsUsageTrendsNode` renders through the SQL insights chart dispatcher (`DataVisualization`'s `LineGraph`), wrapping its trend points in SQL `AxisSeries` column metadata just to reach the quill `TimeSeriesLineChart` underneath. That chart is meant for SQL insights only, and external consumers make cleaning up its legacy paths (see [#70675](https://github.com/PostHog/posthog/pull/70675)) harder.

**Why:** keep the SQL chart scoped to SQL insights; the endpoints usage chart should consume `@posthog/quill-charts` directly like other product charts do.

## Changes

- `EndpointsUsageTrendsNode` builds quill `Series` directly and renders `TimeSeriesLineChart`, using the app-level `useChartTheme`/`useChartConfig` hooks.
- Extracted the presentational `EndpointsUsageTrendsChart` out of the query node and added a Storybook story (line, area, and breakdown variants) so the chart has visual-regression coverage.
- Behavior preserved: metric unit scaling (bytes/duration/CPU/error rate) with suffixed axis ticks and tooltip values, breakdown grouping into multiple series, area fills for CPU time and bytes read, legend for breakdown series.

- New Storybook story renders the chart with static data — light/dark snapshots are captured by Visual Review, which is the visual verification for this rendering swap.
- `oxlint`/`oxfmt` clean on the changed files.
- No unit tests added: like-for-like rendering swap, and quill chart behavior is covered by the charts package's own tests.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A — internal chart rendering only.

## 🤖 Agent context

Authored with PostHog Code (Claude Code). Part of scoping the DataVisualization chart to SQL insights only — an audit found consumers outside SQL insights; this PR covers the endpoints area (app metrics: [#70834](https://github.com/PostHog/posthog/pull/70834); error tracking follows separately). The data transform helpers were kept, only their output shape changed from `AxisSeries` to quill `Series` + labels.

---


sakce (approved):

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
+            xAxis: { tickFormatter: createXAxisTickCallback({ allDays: labels, timezone }) },
+            yAxis: { scale: 'linear', showGrid: true, startAtZero: true, tickFormatter: formatValue },
+            legend: { show: series.length > 1, position: 'top', interactive: true },
+            tooltip: {
+                placement: 'cursor',
+                pinnable: true,
+                sortedByValue: true,
+                showTotal: series.length > 1,
+                valueFormatter: formatValue,
+                labelFormatter: (label: string) => dayjs(label).format('MMM D, YYYY'),
```
sampennington: 🤖 **sp-ship** · 💻⚛️📐📊 `[code+react+xp+hog-charts]` · **Blocker**

The query's `interval` is dropped, and both date formatters suffer for it. `EndpointsUsageTrendsQuery` carries `interval?: IntervalType`, and `EndpointsUsageFilters.tsx` exposes it as a Hourly/Daily/Weekly picker, with the runner bucketing via `toStartOfHour` etc. Pick **Hourly** and all 24 buckets in a day render the identical header `Jul 1, 2026` — the old chart.js path printed the raw bucket string, which was ugly but lossless. Separately, `dayjs(label)` parses in the browser's zone while the axis ticks (line 94) parse in the team zone, so the two can name different days for the same point.

Quill already solves both. Thread `props.query.interval` into the chart and declare the axis instead of hand-building the callback:

```ts
xAxis: { timezone, interval, allDays: labels },
// drop tickFormatter, tooltip.labelFormatter, and the dayjs import
```

`useXTickFormatter` then builds the tick callback from the real bucket size rather than `inferInterval`'s guess from the first two labels (the runner doesn't zero-fill, so a sparse series can infer wrong), and `useTimeSeriesTooltipConfig` installs `createTooltipDateFormatter({ interval, timezone })` so the header can't drift from the ticks.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -134,29 +161,16 @@ function getScaleFactor(values: number[], metric: string): ScaleFactor {
...
-    const { divisor, label, suffix, decimalPlaces } = getScaleFactor(values, metric)
-    return {
-        values: values.map((v) => v / divisor),
-        label,
-        settings: { formatting: { suffix, decimalPlaces } },
-    }
-}
-
-function transformDataForLineGraph(
+function transformDataForChart(
```
sampennington: 🤖 **sp-ship** · 🧪📊 `[test+hog-charts]` · **Blocker**

The rewritten pure transform ships with no unit test — three canvas snapshots are its only guard, and pixels can't discriminate its outputs. Regressions that would sail through: swapping the `query_duration` min/s divisor-suffix pair (no story uses that metric), dropping `/ scale.divisor` in the breakdown branch (the `Breakdown` story is `metric="requests"` where the divisor is 1, so the bug is invisible), or returning `label: metric` instead of `label: scale.label`. `error_rate` is the one branch whose divisor inverts (`0.01`), which is exactly where an off-by-100 hides.

`transformDataForChart` and `getScaleFactor` are pure, take plain data, and return plain `{ labels, series, scale }` — the cheapest possible test level, and nothing else in the repo covers this file. The analogous quill migrations here all shipped transform tests alongside their stories (`sqlLineGraphAdapter.test.ts`, `stickinessBarChartTransforms.test.ts`). Lift both into a sibling `endpointsUsageTrendsTransforms.ts` and add one `test.each` over the metric/threshold matrix, plus a sparse-breakdown case for the `|| 0` gap fill and `Object.keys(dateGroups).sort()` ordering.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
-    const { xData, yData } = transformDataForLineGraph(results, props.query.metric)
+    return <EndpointsUsageTrendsChart results={results} metric={props.query.metric} />
+}
+
+export function EndpointsUsageTrendsChart({
+    results,
+    metric,
+}: {
+    results: TrendsDataPoint[]
+    metric: string
```
sampennington: 🤖 **sp-ship** · 💻✨📐📊 `[code+quality+xp+hog-charts]` · **Suggestion / follow-up**

`metric: string` widens what the schema already narrows to `'bytes_read' | 'cpu_seconds' | 'requests' | 'query_duration' | 'error_rate'`. Now that this component is exported and consumed by the story, a typo lands on `getScaleFactor`'s default branch and silently renders unscaled with an "Executions" label instead of failing to compile. `metric: EndpointsUsageTrendsQuery['metric']` costs nothing at the one real call site.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
+        [results, metric, isAreaChart]
+    )
+
+    const { timezone } = useValues(teamLogic)
+    const theme = useChartTheme()
+    const config = useChartConfig<TimeSeriesLineChartConfig>(() => {
+        const formatValue = (value: number): string => `${value.toFixed(scale.decimalPlaces)}${scale.suffix}`
+        return {
+            xAxis: { tickFormatter: createXAxisTickCallback({ allDays: labels, timezone }) },
+            yAxis: { scale: 'linear', showGrid: true, startAtZero: true, tickFormatter: formatValue },
```
sampennington, line 97: 🤖 **sp-ship** · 💻📊 `[code+hog-charts]` · **Suggestion / follow-up**

Reusing `formatValue` as the y-axis `tickFormatter` is a behavior change — the old path left y ticks to chart.js, which pinned `ticks: { precision: 1 }`. `LineChart` generates ticks via `d3Scales.y.ticks(...)`, and in this `h-60` box the plot region is ~110-130px, so d3 emits fractional steps on small domains: a low-volume endpoint with a max of 1 request/day renders ticks as `0, 0, 0, 1, 1, 1`. Large counts also lose digit grouping. Consider letting quill format the axis (it adapts precision per tick and groups digits) and keeping `formatValue` for `tooltip.valueFormatter`, where fixed decimals are what you want.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
+            legend: { show: series.length > 1, position: 'top', interactive: true },
+            tooltip: {
+                placement: 'cursor',
+                pinnable: true,
+                sortedByValue: true,
+                showTotal: series.length > 1,
```
sampennington, line 103: 🤖 **sp-ship** · 📊 `[hog-charts]` · **Suggestion / follow-up**

`showTotal` is gated only on series count, so a "Total" row is appended for every breakdown chart regardless of whether the metric is additive. `requests`, `bytes_read`, and `cpu_seconds` sum meaningfully across endpoints; `error_rate` (a per-endpoint percentage) and `query_duration` (a per-endpoint duration) do not — a four-endpoint breakdown shows a summed percentage or a summed latency, presented with the same authority as the real rows. Consider gating on the metric too.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
-            <LineGraph
-                xData={xData}
-                yData={yData}
-                visualizationType={chartType}
-                chartSettings={{
-                    showLegend: yData.length > 1,
-                }}
-            />
+        <div className="border rounded bg-bg-light p-2 h-60 flex flex-col">
+            <TimeSeriesLineChart series={series} labels={labels} theme={theme} config={config} />
```
sampennington: 🤖 **sp-ship** · 📊 `[hog-charts]` · **Suggestion / follow-up**

`onError` is dropped. The comparable adapter this is modelled on passes `onError={makeChartErrorHandler('sql-line-chart')}`, which forwards the `ChartErrorBoundary` catch to `posthog.captureException` tagged with the chart name. Without it a render failure still shows the boundary fallback but only reaches `console.error`, so a chart that breaks on a specific data shape is invisible in telemetry.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -134,29 +161,16 @@ function getScaleFactor(values: number[], metric: string): ScaleFactor {
...
-        settings: { formatting: { suffix, decimalPlaces } },
-    }
-}
-
-function transformDataForLineGraph(
+function transformDataForChart(
     results: TrendsDataPoint[],
-    metric: string
+    metric: string,
+    isAreaChart: boolean
```
sampennington: 🤖 **sp-ship** · ⚛️✨📐 `[react+quality+xp]` · **Suggestion / follow-up**

Two small cleanups in this transform. (1) `isAreaChart` is a presentation flag threaded down purely to spread `...fill`, and it's fully derived from `metric`, which the function already has — deriving `fill` locally drops the parameter and a redundant `useMemo` dependency, and keeps the transform data-only. (2) `getScaleFactor(results.map((r) => r.value), metric)` is now called identically in both branches (lines 196 and 214); hoisting it above `const hasBreakdown` removes the duplication and makes the branch read as what it is — one way of building `labels`/`series`, two ways of grouping.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
-            ? ChartDisplayType.ActionsAreaGraph
-            : ChartDisplayType.ActionsLineGraph
+    const { labels, series, scale } = useMemo(
+        () => transformDataForChart(results, metric, isAreaChart),
+        [results, metric, isAreaChart]
+    )
+
+    const { timezone } = useValues(teamLogic)
+    const theme = useChartTheme()
+    const config = useChartConfig<TimeSeriesLineChartConfig>(() => {
```
sampennington, line 93: 🤖 **sp-ship** · ♻️ `[reuse]` · **Suggestion / follow-up**

This config block is now the third near-verbatim copy of the same quill line-chart setup: `frontend/src/lib/components/AppMetrics/AppMetricsTimeSeriesChart.tsx` has the identical `useChartTheme()` + `useChartConfig<TimeSeriesLineChartConfig>` shape (same xAxis tick callback, same `legend: { show, position: 'top', interactive: true }`, same tooltip block, same `flex flex-col` container), and the error-tracking rate-limit charts hand-roll it again. Importing from `DataVisualization/Components/Charts/` is deliberately off the table here, but three independent copies across the merged siblings is the signal to extract a SQL-free helper into `frontend/src/lib/charts/` next to `hooks.ts` and have them all consume it. Fair as a follow-up rather than a blocker on this PR.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
```diff
@@ -63,24 +68,46 @@ export function EndpointsUsageTrendsNode(props: {
...
+}
+
+export function EndpointsUsageTrendsChart({
+    results,
+    metric,
+}: {
+    results: TrendsDataPoint[]
+    metric: string
+}): JSX.Element {
+    // CPU time and bytes read render better as filled areas; other metrics stay plain lines.
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

This comment restates the condition on the very next line, and "other metrics stay plain lines" narrates the boolean. `isAreaChart` already carries the meaning — suggest deleting it.

products/endpoints/frontend/nodes/EndpointsUsageTrendsChart.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ReactNode } from 'react'
+
+import { EndpointsUsageTrendsChart } from './EndpointsUsageTrendsNode'
+
+const meta: Meta<typeof EndpointsUsageTrendsChart> = {
+    title: 'Endpoints/UsageTrendsChart',
+    component: EndpointsUsageTrendsChart,
+    parameters: {
+        layout: 'centered',
+        testOptions: { snapshotBrowsers: ['chromium'] },
```
sampennington, line 12: 🤖 **sp-ship** · 📊📐 `[hog-charts+xp]` · **Suggestion / follow-up**

These stories opt into snapshotting without `testOptions.skipCanvasDraw`. Quill charts paint asynchronously (ResizeObserver → rAF), which is the race `withChartCanvasSnapshot` exists to defuse (`Dashboards.stories.tsx` and `MCPAnalyticsDashboard.stories.tsx` both set it). The bind is real here though: the plot area *is* the story, so `skipCanvasDraw` would snapshot a blank canvas with only the axis/legend overlays. That may still be the right call — overlay layout is what a snapshot catches reliably anyway — but worth deciding deliberately rather than by omission, otherwise this lands as three new flaky VR checks on every PR. (`snapshotBrowsers: ['chromium']` is already the default and can be dropped.)

products/endpoints/frontend/nodes/EndpointsUsageTrendsChart.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type Story = StoryObj<typeof EndpointsUsageTrendsChart>
+
+const DAYS = ['2026-07-01', '2026-07-02', '2026-07-03', '2026-07-04', '2026-07-05', '2026-07-06', '2026-07-07']
+
+function Stage({ children }: { children: ReactNode }): JSX.Element {
+    return <div className="w-[760px]">{children}</div>
+}
+
+// Single line series — request counts over a week.
```
sampennington: 🤖 **sp-ship** · 🦔✨ `[personal+quality]` · **Suggestion / follow-up**

The three story comments (lines 24, 36, 51) each restate the story below them — the names `Requests` / `BytesReadArea` / `Breakdown`, the `metric` prop, and the `* 1024 * 1024` already say it. Suggest deleting all three. Line 51's is also inaccurate: it says "stacked legend", but nothing about the legend is stacked.

sampennington (commented):

Two things to fix. The chart drops the query's user-selectable `interval` and hardcodes a day-granularity tooltip header, so on the Hourly setting all 24 buckets in a day render the same header — and because the header parses in the browser zone while the axis parses in the team zone, the two can disagree by a day; quill's `xAxis: { timezone, interval }` fixes both and deletes the hand-rolled formatters. Separately the pure scaling/grouping transform was rewritten in this diff with no unit test, and a canvas snapshot can't discriminate a wrong divisor or an off-by-100 on `error_rate`.

The rest is non-blocking: axis tick formatting, a nonsensical Total row on non-additive metrics, a dropped `onError`, and some transform/story tidying. Worth noting two things that checked out and are **not** problems — the `h-60 flex flex-col` container is correct for quill's `flex-1 min-h-0` shell, and area breakdowns still stack (`LineChart` auto-stacks when 2+ series carry `fill`), so this really is like-for-like on layout and stacking.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Fixed in `bfe19d8` — dropped the hand-rolled `tickFormatter`/`labelFormatter` (and the `dayjs` import), threaded `props.query.interval` through, and now declare `xAxis: { timezone, interval, allDays: labels }` so quill derives both the axis ticks and the tooltip header from the real bucket size and team timezone.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Fixed in `bfe19d8` — lifted `getScaleFactor` and `transformDataForChart` into a sibling `endpointsUsageTrendsTransforms.ts` and added a `test.each` over the metric/threshold matrix (including the `error_rate` 0.01 divisor) plus a sparse-breakdown case asserting the `|| 0` gap fill.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Fixed in `bfe19d8` — `EndpointsUsageTrendsChart` now takes `metric: EndpointsUsageMetric` (the query's own narrowed union), landed alongside the transform extraction since both touched the same signature.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Fixed in `fc4efc1` — wired `onError={makeChartErrorHandler('endpoints-usage-trends-chart')}`, matching the SQL/stickiness chart adapters.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Fixed in `fc4efc1` — removed, the boolean already carries the meaning.

products/endpoints/frontend/nodes/EndpointsUsageTrendsChart.stories.tsx
sampennington: 🤖 **sp-triage**

Fixed in `fc4efc1` — removed all three, the story names and props already say it.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Out of scope for this PR — the transform just moved into `endpointsUsageTrendsTransforms.ts` to land the missing-test blocker, and reshaping its internals in the same pass risked conflicting with the new tests. Fair as a follow-up cleanup.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Agreed as a follow-up rather than a blocker on this PR — deferring the shared-helper extraction out of scope here.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Deferring — this is a real UX tradeoff (fixed decimals vs quill's adaptive per-tick precision) that needs a human call on which behavior is wanted for low-volume axes, not a mechanical fix. Leaving open for review.

products/endpoints/frontend/nodes/EndpointsUsageTrendsNode.tsx
sampennington: 🤖 **sp-triage**

Deferring — gating `showTotal` correctly needs a human decision on which metrics are additive (and how that should be expressed, e.g. on `getScaleFactor`'s return or a separate lookup), not a mechanical fix. Leaving open for review.

products/endpoints/frontend/nodes/EndpointsUsageTrendsChart.stories.tsx
sampennington: 🤖 **sp-triage**

Deferring — whether to keep full canvas paint (real plot-area coverage, flakier) or opt into `skipCanvasDraw` (stable, but only snapshots axis/legend overlays) is a deliberate tradeoff for a human to make, not a mechanical fix. Leaving open for review.

PR #70847, opened 2026-07-14
feat(web-analytics): session id set fast path for filtered web overview and paths tile

lricoy:
## Problem

PR #70746 removed the events↔sessions join for *unfiltered* web analytics queries (86% of user-facing traffic). The filtered remainder still pays the full join: the sessions-side subquery aggregates every session in the date range on every shard, even when the filter matches a tiny fraction of sessions. Measured on a large team, a pathname-filtered 30d overview reads the same 384M rows as unfiltered (9.1s, 4.6 GiB peak) because the filter never reaches the sessions side.

The alternative validated in production benchmarks: run the filtered events side first, collect the set of session ids it matches, then aggregate the sessions side over only that id set. With a selective filter (62k of 34M sessions) that ran in 0.95s at 24 MiB peak - 9.6x faster, ~190x less memory, identical results.

> [!NOTE]
> Naming: earlier revisions of this PR called this the "two-phase" fast path. Renamed to **session id set** - the mechanism is the id set linking the two scans, and the tags/settings now say so: `web_overview_session_id_set_query`, `web_overview_session_id_set_preflight`, `WEB_ANALYTICS_SESSION_ID_SET_TEAM_IDS`.

## Changes

Two pieces:

- **HogQL layer** (`sessions_v2.py`): `build_direct_session_id_in_pushdown` - when a direct select over `sessions` carries a top-level `session_id IN (SELECT ...)` filter and the `sessionIdPushdown` modifier is on, the id-set filter is rewritten onto `raw_sessions.session_id_v7` *inside* the lazy subquery, before the per-session GROUP BY, as `GLOBAL IN` (id subquery executes once on the initiator), and the original outer term is neutralized so the id subquery isn't executed a second time above the GROUP BY. Without this rewrite the filter applies after aggregating all sessions in range - correct but slow (measured 27s vs 0.95s for the same shape). The rewrite only fires when the filtered column provably belongs to the query's single sessions occurrence (resolved-type ownership check - a joined table's own `session_id` column, an events property named `session_id`, and sessions self-joins are all left untouched), and string ids are matched only in canonical lowercase UUID form via a `toString` round-trip, exactly like the un-rewritten byte-exact string comparison. Fails open on everything else (NOT IN, literal lists, OR-nested terms, multi-column subqueries): the outer filter still applies untouched, so it is purely a performance rewrite.
- **Overview runner** (`web_overview.py`): `should_use_session_id_set` + `session_id_set_select` - when the team is in `WEB_ANALYTICS_SESSION_ID_SET_TEAM_IDS` and *every* filter is evaluable events-side (event property filters; test-account filters of type event/person; no conversion goal — sampling fields are ignored no-ops across web analytics since #71413), the query becomes the two-scan shape from #70746 with the filters applied events-side and a `session_id IN (SELECT DISTINCT $session_id_uuid FROM filtered events)` predicate sessions-side. `_calculate` flips `sessionIdPushdown` on for this shape. Session/cohort filters keep the join path untouched.

- **Paths tile** (`stats_table.py` + `stats_table_strategies.py`): the same optimization for the heaviest live query family. `SessionIdSetPathBounceStrategy` / `SessionIdSetPathBounceAvgTimeStrategy` reuse the no-join two-scan templates with filters applied: the counts side filters events directly, and the bounce side aggregates sessions only over the id set collected under the **pathname-excluded** filters (mirroring the join path's `_event_properties_for_bounce_rate` semantics - an entry path's bounce rate must not be restricted by which pathname is being viewed). When the only filter is `$pathname`, the bounce side stays a plain unfiltered sessions scan - no id set, no preflight. New tags: `stats_table_session_id_set_path_bounce_query`, `stats_table_session_id_set_path_bounce_and_avg_time_query`, `stats_table_session_id_set_preflight`. The shared eligibility gates and preflight now live on `WebAnalyticsQueryRunner` so both runners stay in sync.

Rollout gating: allowlist only, no percent mechanism. `WEB_ANALYTICS_SESSION_ID_SET_TEAM_IDS` defaults to the Cloud dogfooding team (project 2) on US Cloud - same default pattern as the other web analytics fast-path lists - and empty on EU and self-hosted.

Because the rewrite lives in shared HogQL schema code keyed on a modifier that teams can persist in `team.modifiers`, any HogQL query with a matching shape (SQL editor included) can hit it. That exposure is covered directly: a ClickHouse-backed parity test asserts a direct sessions query returns identical rows with the modifier forced on vs off (including malformed non-UUID session ids), and shape tests pin that the rewrite fires only for the exact supported shape and never copies the id subquery.

Benchmarks (US prod, team 2, 30d, `$pathname LIKE '/blog%'`, stats from `system.query_log`):

| Variant | Duration | Read | Peak memory |
|---|---|---|---|
| Join (today's filtered path) | 9.1s | 70 GiB | 4.6 GiB |
| Id-set filter without the pushdown | 27-34s | 10 GiB | 200 MiB |
| **Session id set with pushdown (this PR's shape)** | **0.95s** | 8.3 GiB | **24 MiB** |

> [!NOTE]
> Selectivity gate: before choosing the session-id-set shape, `_calculate` runs a sub-second preflight (`web_overview_session_id_set_preflight`, `uniq($session_id_uuid)` over the filtered events on materialized columns). Above `SESSION_ID_SET_MAX_MATCHING_SESSIONS` (10M ids; memory scales ~linearly with set size and the join path wins past ~20M) - or on any preflight error - the query falls back to the join path, so unselective filters (e.g. $host on a single-domain team) can't degrade toward the known GLOBAL IN broadcast regression.

Runner tests (`test_web_overview.py`):

- `test_session_id_set_results_match_join_path` (parameterized over compare on/off) and `test_session_id_set_with_test_filters_matches_join_path`: identical results to the join path on a ClickHouse fixture with matching, non-matching, and malformed (non-UUID) session ids. Catches semantic drift in the two-scan shape.
- `test_session_id_set_pushes_id_filter_below_session_aggregation`: asserts the printed SQL contains exactly one `globalIn(raw_sessions.session_id_v7, ...)` plus a full snapshot of the generated query. The pushdown fails open by design, so a silent break would keep results correct while regressing the sessions scan back to aggregating every session in range; a retained outer copy would run the filtered-events scan twice.
- `test_session_id_set_eligibility` / `test_session_id_set_supports_event_and_person_test_account_filters` (parameterized): session-property, cohort, mixed, no-filter, and conversion-goal shapes all fall back to the join path.
- `test_session_id_set_falls_back_to_join_when_filter_is_unselective` and `test_session_id_set_falls_back_to_join_when_preflight_query_fails`: both the over-the-cap and preflight-exception paths fall back to the join and return correct results.

Shared-schema hardening (`test_session_v2_where_clause_extractor.py`, new `TestDirectSessionIdInPushdownV2`):

- Shape lock: `session_id IN (SELECT ...)` rewrites to exactly one `globalIn` with the events scan appearing once (moved, not copied); `session_id_v7 IN (...)` skips the UUID cast.
- Negative shapes (parameterized): literal lists, `NOT IN`, OR-nested terms, and modifier-off are never rewritten; a multi-column subquery fails open without neutralizing the original predicate; v3 sessions are untouched.
- Hijack guards (parameterized): a `session_id`-named column belonging to a joined table (an events property, a replay table column) and sessions self-joins produce byte-identical SQL with the modifier on vs off - the rewrite must never steal another table's filter.
- `test_pushdown_preserves_results_for_direct_sessions_query`: the SQL-editor safety proof - a direct sessions query returns identical rows with the modifier on vs off on a fixture with matching, non-matching, malformed, and uppercase (non-canonical) session ids, plus an executed-SQL assertion that the pushed shape actually fired when (and only when) the modifier is on.

Adversarial review pass: a multi-agent review of the diff (independent reviewers per failure class, each finding double-verified by adversarial agents attempting to refute it) surfaced three real issues, all fixed here with the locking tests above - the joined-table filter hijack, the non-canonical UUID string semantics change, and the preflight running even when the pre-aggregated builder already served the response (now gated). It also flagged that the fallback/parity tests could pass vacuously under fail-open; they now assert the executed query's tag and modifiers instead of internal flags alone.

Paths tile (`test_web_stats_table.py`, new `TestWebStatsTableSessionIdSetFastPath`, 18 tests): join-path parity for bounce and avg-time variants under browser and pathname filters (fixture includes a malformed session id and a time-on-page signal so the avg-time assertion is non-vacuous); strategy-selection eligibility (parameterized incl. sampling, session filters, conversion goals, non-PAGE breakdowns); no-join precedence for unfiltered queries; SQL-shape lock (exactly one `globalIn` for non-pathname filters, zero for pathname-only); execution wiring (paginator call's tag + `sessionIdPushdown`); observable preflight fallback to the join strategy.

Full runs: 247 passed across `test_web_stats_table.py`, `test_web_overview.py`, `test_session_v2_where_clause_extractor.py`, and `test_sessions_v2.py`. The generated query was also rendered with production parameters and validated read-only against US prod: sessions/duration/bounce identical to the join path across teams from 62k to 4.8M matching ids.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change outside the allowlisted dogfood team.

## 🤖 Agent context

- Authored with Claude Code as the follow-up to #70746, from Lucas's proposal to run the left (events) side separately and filter the sessions side by its result. Skills invoked: /writing-tests, /evaluating-web-analytics-performance.
- Key empirical decisions: the naive HogQL-expressible shape (id filter applied after session aggregation) measured 27-34s despite ~5 CPU-seconds, so the pre-GROUP-BY pushdown is the load-bearing piece; the urls-array membership alternative was rejected on semantics (it matches any event URL including autocapture, returning 2.3x the sessions of the event filter); reusing the existing `sessionIdPushdown` modifier avoids schema churn.
- Review feedback round: renamed "two-phase" to "session id set", defaulted the allowlist to the US dogfood team, and added the shared-schema hardening test class covering the `team.modifiers` exposure.
- Patch coverage: the changed lines are exercised by the new parity/eligibility/SQL-shape tests; check the backend coverage PR comment on CI.


jordanm-posthog (approved):
🚀

PR #70858, opened 2026-07-14
feat(metrics): add access control

ReeceJones:
## Problem

Metrics can be opened and queried by any project member even when resource access controls are configured. This makes the product inconsistent with other app monitoring surfaces.

**Why:** Teams need one place to control who can view metrics data, with the same behavior in the API, navigation, and access control settings.

## Changes

- Register metrics as a resource-level access control resource.
- Gate metrics API requests and the metrics scene at viewer access.
- Add metrics to the access control settings table, rollout mapping, icon map, and Storybook context.

- Added an EE permission matrix covering viewer access and `none` access. This guards the previously unrestricted metrics API route.
- Ran `ruff check`, Python compilation, and `git diff --check` successfully.
- Could not run the targeted pytest or frontend checks because this checkout does not have synced Python or Node dependencies. The no-sync pytest attempt could not find `pytest`.

+ Manually testing

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs update is needed. Metrics appears automatically in the existing access control settings UI when its rollout flag is enabled.

## 🤖 Agent context

Implemented with PostHog Code using Codex CLI. I used `/rj-retroactively-adding-access-control`, `/improving-drf-endpoints`, and `/writing-tests`. Metrics has no persisted resource object or mutating endpoint, so I chose resource-level viewer controls rather than object-level controls or serializer access fields.

---


DanielVisca:
Took a look through this. The backend side is solid: every metrics endpoint 403s before validation, and the insight/query runner path is covered too, so the enforcement is real.

One thing worth knowing: the control stops at the product API. The metrics HogQL tables (`posthog.metrics` etc) carry no access scope, so a member set to "none" can still select from them through the generic query endpoint. Same story as logs and tracing today, so I don't think it blocks this PR, but worth a follow-up if we want the setting to mean what admins will assume it means. I added a note to the settings tooltip in the meantime.

Pushed a few fixes on top (f3747ec): registered the scene in `sceneToAccessControlResourceType` so no-access users get the normal access denied page instead of a fully greyed-out UI (deep links skipped the tab click guard otherwise), fixed the "this metrics" grammar in the permission tooltip, made the new `LemonInputSelect` `disabledReason` also disable the value pills and clear button (they were still clickable), and stopped an access-denied `has_metrics` check from reading as "you haven't sent any metrics yet". Tests included.

DanielVisca (approved):

ReeceJones:
> One thing worth knowing: the control stops at the product API. The metrics HogQL tables (`posthog.metrics` etc) carry no access scope, so a member set to "none" can still select from them through the generic query endpoint. Same story as logs and tracing today, so I don't think it blocks this PR, but worth a follow-up if we want the setting to mean what admins will assume it means. I added a note to the settings tooltip in the meantime.

This is also a problem with some other products which have access control. We're aware of it and will fix it sometime this Q.
 
> Pushed a few fixes on top ([f3747ec](https://github.com/PostHog/posthog/commit/f3747ec569cb0dcb45519d76b471fccf017690a1)): registered the scene in `sceneToAccessControlResourceType` so no-access users get the normal access denied page instead of a fully greyed-out UI (deep links skipped the tab click guard otherwise), fixed the "this metrics" grammar in the permission tooltip, made the new `LemonInputSelect` `disabledReason` also disable the value pills and clear button (they were still clickable), and stopped an access-denied `has_metrics` check from reading as "you haven't sent any metrics yet". Tests included.

Thanks!

PR #70878, opened 2026-07-14
fix(replay-vision): render comma-joined timestamp citations as seek chips

arnohillen:
## Problem

Replay Vision renders `(t <seconds>)` citation markers from scanner output as seekable timestamp chips.
The chip extractor only matched the single form `(t 123)`, but the model also leaks comma-joined variants like `(t 1437, 1441, 1444, 1446)` despite the one-moment-per-parens prompt rule.
Those fell through and showed up as raw text in the observation UI:

> Immediately after the switch, the user successfully found and opened several recordings from the top of the newly sorted list (t 1479, 1481).

## Changes

- The signal-description stripper in `scanners/base.py` already matched comma-joined markers leniently. That regex is now shared (as `TIMESTAMP_CITATION_RE`) with the chip extractor in `call_scanner_provider.py`, so the two parsers can't drift apart.
- The extractor emits one chip per cited second inside a marker, keeping the existing past-end-of-recording drop per timestamp.
- `CitedText` now splits any leaked markers client-side via the new `parseCitedSegments` in `utils/citations.ts`. Observations persisted before this fix render chips too, including old rows that never had segments.

| | Rendered reasoning |
|---|---|
| Before | ...changed the date range filter (t 1437, 1441, 1444, 1446), scrolling through... |
| After | ...changed the date range filter ⏪ 23:57 ⏪ 24:01 ⏪ 24:04 ⏪ 24:06, scrolling through... |

> [!NOTE]
> The client-side fallback can't range-check timestamps (the component doesn't know the recording duration). A misread out-of-range citation on an old row becomes a chip that seeks to the end of the recording instead of raw text. New rows keep the backend range check.

- Extended the existing parameterized `TestExtractSegments` cases with three comma-joined variants (plain comma list, repeated `t` prefix, mixed in-range and out-of-range). These fail if the extractor regresses to single-form matching.
- Added a jest test for `parseCitedSegments` covering the leaked-marker split inside persisted segments, the no-segments fallback, plain-text passthrough, and malformed segment filtering. These fail if the client-side fallback is removed. No existing frontend test covered citation parsing.
- All 81 existing scanner tests still pass, which covers the shared regex through the signal-description stripper.
- I (or, actually Claude) did not run the UI manually, so the table above is derived from the rendering code rather than a live screenshot.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, internal rendering fix.

## 🤖 Agent context

Authored with Claude Code, starting from a screenshot of an observation whose model reasoning contained unparsed markers. Skills invoked: /writing-tests.

Decisions: reused the existing lenient stripper regex instead of adding a second pattern, and added the client-side fallback because segments are computed at scan time, so a backend-only fix would leave already-persisted observations rendering raw markers forever. Collapsing clustered timestamps into a single range chip was considered and rejected for now (it needs a new segment kind through the model, API, and renderer).

### DIFF
--- a/products/replay_vision/backend/temporal/activities/call_scanner_provider.py
+++ b/products/replay_vision/backend/temporal/activities/call_scanner_provider.py
@@ -36,6 +36,7 @@ from products.replay_vision.backend.temporal.gemini import gemini_api_key
 from products.replay_vision.backend.temporal.metrics import record_provider_call
 from products.replay_vision.backend.temporal.scanners import scanner_from_snapshot
 from products.replay_vision.backend.temporal.scanners.base import (
+    TIMESTAMP_CITATION_RE,
     BaseScanner,
     BaseScannerOutput,
     ChipSegment,
@@ -57,8 +58,6 @@ logger = structlog.get_logger(__name__)
 _MAX_LLM_ATTEMPTS = 2  # one initial call + one re-prompt with the validation error appended per step
 # Cache TTL: a scan is a handful of turns and finishes in minutes; well under this.
 _VIDEO_CACHE_TTL = "900s"
-# `(t <seconds>)` with optional leading whitespace, so we eat the space when stripping the paren.
-_TIMESTAMP_CITATION_RE = re.compile(r"\s*\(t (\d+)\)")
 
 _OutputT = TypeVar("_OutputT", bound=BaseModel)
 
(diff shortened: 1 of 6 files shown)


ksvat (approved):

arnohillen:
### Merge activity

* **Jul 14, 9:48 PM UTC**: @arnohillen merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/70878).

PR #70883, opened 2026-07-14
fix(cohorts): repair precalculated cohort data made stale by person merges

### LINKED ISSUE #70109: fix(cohorts): realtime cohort pipeline uses stale person_id after distinct_id re-identification
## Bug description

Realtime cohort membership (`cohort_membership`) can end up pinned to a `person_id` that no longer matches the *current* `distinct_id → person` mapping, so a person can be recorded as an "entered" cohort member in ClickHouse while never appearing in the cohort's Persons list in the UI (which always resolves membership against current identity).

## Root cause

`precalculated_events` (populated by the realtime consumer, see `posthog/hogql/database/schema/precalculated_events.py`) stores the `person_id` that was resolved for a `distinct_id` at the moment a behavioral condition was evaluated. `HogQLRealtimeCohortQuery` (`posthog/hogql_queries/hogql_cohort_query.py`) reads that `person_id` straight off `precalculated_events` when building behavioral cohort conditions, and `RealtimeCohortCalculationWorkflow` (`posthog/temporal/messaging/realtime_cohort_calculation_workflow.py`) writes `cohort_membership` from that result.

If the `distinct_id` is later reassigned to a different person (identify/merge — visible as a version bump in `person_distinct_id2`), nothing re-resolves or invalidates the existing `precalculated_events`/`cohort_membership` rows. They keep referencing the old, now-orphaned person.

The event backfill path (`posthog/management/commands/backfill_precalculated_events.py` / `backfill_precalculated_events_workflow.py`, also surfaced via the admin backfill tool) has the same issue: it replays historical events using whatever identity resolution applied *at replay time*, without accounting for merges that happened between the original event and the backfill run.

## How to reproduce

1. Capture an event for `distinct_id` X while it's associated with person A, matching a realtime cohort's behavioral condition — this writes a `precalculated_events` row with `person_id = A`.
2. Re-identify/merge `distinct_id` X onto person B, so `person_distinct_id2` now resolves X to B.
3. Let `RealtimeCohortCalculationWorkflow` run again (or run the event backfill for that cohort/date range).
4. `cohort_membership` gets (or keeps) an `entered` row for person A, who `distinct_id` X no longer belongs to. Person A won't show up in the cohort's Persons list, since that's resolved via current identity — the membership row is effectively orphaned.

This is most likely to surface when a merge happens close in time to the event/backfill that trips the behavioral condition, but can affect backfilled data of any age relative to a later merge.

## Impact

- `cohort_membership` (and anything built on top of it — realtime cohort feature flag targeting, cohort-of-cohort composition) can silently reference stale, non-current person identities with no visible symptom other than a membership row that never surfaces anywhere.
- There's currently no mechanism that re-evaluates `precalculated_events`/`cohort_membership` when a person merge invalidates the identity resolution used to produce them.

## Suggested fix directions

- Re-resolve `person_id` in `precalculated_events` (or at query time in `HogQLRealtimeCohortQuery`) against the current `person_distinct_id2`/overrides mapping instead of trusting the value captured at write time.
- Or: on person merge, emit an invalidation/recompute signal for any `precalculated_events`/`cohort_membership` rows tied to the old `person_id`, so the next realtime cohort calculation run reconciles them.

## Additional context

Found while investigating a report of a `cohort_membership` row whose person never appeared in the cohort's UI Persons list; traced to a `precalculated_events` row keyed to a `person_id` that `person_distinct_id2` no longer resolves that `distinct_id` to.

Related: #53990 (behavioral cohorts event backfilling mega-issue), #39366 (Realtime Cohort Calculation MVP).

gustavohstrassburger:
## Problem

Realtime cohort membership (`cohort_membership`) is computed from `precalculated_events`, which stores the `person_id` that was resolved for a `distinct_id` at write time. When a later identify/merge re-points that `distinct_id` to another person, nothing re-resolves or invalidates the stored rows. The realtime pipeline then permanently keeps the superseded person in the cohort and misses the surviving one, diverging from the batch pipeline the UI reads and affecting anything built on `cohort_membership` (realtime flag targeting, cohort-of-cohort composition).

The sibling table `precalculated_person_properties` has the same bug plus a second layer: each row is a *verdict* (`matches` true/false) evaluated against the person's properties at write time, and a merge changes the surviving person's property set, so re-attributing rows isn't enough. Its backfill can't repair these rows either, since it stamps `distinct_id = person uuid` and therefore never shares a row key with the realtime consumer's rows.

The cohort calculation path deliberately never joins person tables, so the fix has to happen at the data instead of at query time: correct the stored rows when merges happen, and make backfills able to repair history.

Closes #70109

## Changes

**Before** — a merge writes an override, but nothing ever consults it, so both precalculated tables keep the pre-merge person forever:

```mermaid
flowchart LR
    M[person merge] --> O[(person_distinct_id_overrides)]
    O -. "never consulted" .-> PE[(precalculated_events<br/>stale person_id)]
    O -. "never consulted" .-> PP[(precalculated_person_properties<br/>stale person_id + stale verdicts)]
    PE --> CALC{{realtime cohort calculation}}
    PP --> CALC
    CALC --> CM[(cohort_membership<br/>stale forever)]
    class PE,PP,CM phRed;
    class O phGray;
```

**After** — the override row is the invalidation signal; every 15 minutes the workflow repairs each table the way its contents demand (facts get re-attributed, verdicts get re-evaluated):

```mermaid
flowchart LR
    M[person merge] --> O[(person_distinct_id_overrides)]
    O --> R{{reconcile-precalculated-data workflow}}
    R -- "re-attribute: re-emit rows<br/>with current person_id" --> K[clickhouse_prefiltered_events topic] --> PE[(precalculated_events)]
    R -- "re-evaluate: fresh verdicts against<br/>surviving person's properties" --> K2[clickhouse_precalculated_person_properties topic] --> PP[(precalculated_person_properties)]
    PE --> CALC{{realtime cohort calculation}}
    PP --> CALC
    CALC --> CM[(cohort_membership<br/>self-heals via entered/left diff)]
    class R,CALC phBlue;
    class O,PE,PP,K,K2,CM phGray;
```

- **New scheduled Temporal workflow `reconcile-precalculated-data`** (interval `RECONCILE_PRECALCULATED_DATA_INTERVAL_MINUTES`, default 15 min, on the existing messaging task queue). Each run is incremental: it only looks at overrides that landed within a lookback window (`RECONCILE_OVERRIDES_LOOKBACK_HOURS`, default 48h), so a merge is reconciled by the next realtime calculation run. A `full_scan` workflow input ignores the window for first-deploy remediation or recovery after downtime longer than the lookback.
- **`precalculated_events` rows are re-attributed**: the event happened regardless of the merge, so stale rows are re-emitted with the current `person_id` and the `ReplacingMergeTree` replaces them. Corrected rows keep their original `source` so the backfill coordinator's day-skip check stays intact.
- **`precalculated_person_properties` rows are re-evaluated, not just re-attributed**: each row is a verdict computed against pre-merge properties, so the workflow re-runs every realtime person-property condition against the surviving person's current properties and emits fresh verdicts per merged distinct_id — including `matches=false`, which is what overwrites a stale `matches=true` row (the row key is `(team_id, condition, distinct_id)`). Persons absent from the persons table are skipped rather than evaluated against empty properties.
- **`cohort_membership` needs no new machinery**: once the source tables are corrected, the next calculation run's existing FULL OUTER JOIN diff emits `left` for the old person and `entered` for the new one.
- **The event backfill now resolves `person_id` through `person_distinct_id_overrides`** (same argMax + `is_deleted` semantics as the HogQL events↔overrides lazy join) instead of reading it raw off the events table, so backfill runs write current identity and re-runs (with the existing ignore-backfilled-dates option) can repair stale history.
- Filter extraction from cohorts moved to `filter_storage` (`extract_person_property_filters`) used by both the workflow and the person-properties backfill command, so the two can't drift.

> [!NOTE]
> The lookback must comfortably exceed the schedule interval (so no merge falls between runs), and both must stay well inside the person-overrides squash cadence (weekly by default): the squash deletes override rows after folding them into the events table only, and neither precalculated table is part of that squash.

@greptile

The agent sandbox has no Docker, so I (well, Claude) could not run the ClickHouse-backed tests locally; ruff, import, and workflow/schedule registration checks pass, and CI runs the tests. Each test group targets a named regression no existing test catches:

- `test_backfill_resolves_person_id_through_overrides` - reverting the backfill query to raw `events.person_id` would attribute a merged distinct_id's events to the superseded person; the control event also catches a broken join that overrides everyone.
- `test_corrects_stale_rows_and_leaves_current_ones_alone` - catches inverted diff logic, correction churn on already-correct rows, and rewriting `source` (which would break the backfill day-skip check).
- `test_old_override_needs_full_scan` - catches a broken `full_scan` recovery path and an incremental run that unexpectedly reacts to overrides outside the lookback window.
- `test_deleted_override_does_not_correct` - catches dropping the `is_deleted` HAVING clause from the overrides query.
- `test_reevaluates_verdicts_for_merged_distinct_ids` - catches re-attributing person-property rows without re-evaluating verdicts, dropping the `matches=false` emission (stale trues would never be overwritten), and keying emissions by anything other than the real merged distinct_id.
- `test_missing_person_is_skipped_not_guessed` - catches evaluating a missing person against empty properties, which would write confidently wrong verdicts.
- `test_returns_distinct_realtime_cohort_teams_only` - catches team-selection drift that would make the reconciler scan teams with no realtime cohorts.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Nothing to update: the touched settings are internal pipeline knobs not documented under `docs/`.

## 🤖 Agent context

Authored with Claude Code (PostHog Code task). Skills invoked: `/writing-tests`. The investigation that led here traced a `cohort_membership` row whose person never appeared in the cohort UI to a `precalculated_events` row keyed to a person the distinct_id no longer resolves to.

Decisions along the way, directed by the human driver: query-time joins against person tables in the cohort calculation path were explicitly ruled out (the whole point of the precalculated tables is join-free calculation), so correction happens against stored data. A merge-topic consumer (`person_merge_events`) was considered and rejected for this iteration: per-merge correction needs a scan of the team's `precalculated_events` slice anyway (person_id is not in the sort key), so batching against the small, squash-bounded overrides table gets the same result with far less machinery. Corrected rows deliberately preserve their original `source` value so `check_day_already_backfilled_activity` keeps working.

---


gustavohstrassburger:
Pushed d080efa reworking the reconciliation to be merge-driven rather than a slow periodic sweep, per the second fix direction in #70109: the override row written at merge time acts as the invalidation signal, each run only diffs distinct_ids merged within a 48h lookback window, and the schedule now fires every 15 minutes so a merge is reconciled by the next realtime calculation run. A `full_scan` input covers first-deploy remediation and recovery after downtime longer than the lookback. PR description updated to match.

gustavohstrassburger:
Pushed 31fad75 extending the fix to `precalculated_person_properties`, so this PR now repairs all the precalculated cohort data a person merge invalidates. That table needed different treatment than `precalculated_events`: its rows are *verdicts* evaluated against the person's properties at write time, and a merge changes the surviving person's property set — so the workflow re-evaluates every realtime person-property condition against the surviving person's current properties and emits fresh verdicts per merged distinct_id (including `matches=false`, which is what overwrites a stale `matches=true` row). Notably the person-properties backfill cannot repair these rows at all — it stamps `distinct_id = person uuid`, so its rows never share a key with the realtime consumer's — making this reconciliation the only repair path. Renamed the workflow to `reconcile-precalculated-data` accordingly; PR description updated.

gustavohstrassburger:
Addressed the review findings in 3f16c92:

**must_fix — absent survivor evaluated against empty properties** (`reconcile_precalculated_data_workflow.py`): a distinct_id whose surviving person isn't yet queryable in the persons table (replication lag, or a deleted survivor excluded by the `is_deleted` HAVING) is now skipped rather than evaluated against `{}`. The override stays in the lookback window so a later run retries once the person row lands. Added `test_missing_person_is_skipped_not_guessed`.

**should_fix — eval-failure written as matches=false**: a condition absent from the evaluator's result (bytecode threw / non-bool, or the condition is no longer an active realtime filter) now leaves the stored verdict untouched instead of overwriting it with a guessed false. Active-condition evaluation failures are counted separately (`verdicts_skipped_eval_failed`) for observability. Added `test_eval_failure_leaves_verdict_untouched`.

**should_fix — missing execution test for argMax person_id resolution**: added `test_performed_event_uses_latest_person_id_after_merge`, which seeds pre-/post-merge rows for the same key and asserts membership reflects only the survivor for both `performed_event` and `performed_event_multiple` — catches a reverted/broken argMax that the string-assertion tests can't.

**should_fix / consider — optimize_aggregation_in_order**: set on both realtime behavioral subqueries and on the events-backfill overrides subquery, mirroring `select_from_person_distinct_id_overrides_table`.

Intentionally not changed: the chunked-team concurrency (a `consider` — the correctness concern was already handled by the run-wide shared `since` boundary; chunk-barrier latency is absorbed by the 15-min schedule / 48h lookback) and the Kafka flush now already raises on delivery failure and validates settings via `_positive_int_env`, so those earlier findings are resolved. Copilot's import-order notes are false positives (isort here is length-sorted; lint is green).

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+            Cohort.objects.filter(deleted=False, cohort_type=CohortType.REALTIME)
+            .values_list("team_id", flat=True)
+            .distinct()
+            .order_by("team_id")
+        )
+
+    return ReconciliationTeamIdsResult(team_ids=await get_team_ids())
+
+
+def _extract_person_property_filters(cohort: Cohort) -> list[PersonPropertyFilter]:
```
haacked: `suggestion`: `_extract_person_property_filters` here is a near-verbatim copy of [`extract_person_property_filters`](https://github.com/PostHog/posthog/blob/3f16c9230b3bee9aead88ce1993132d7a83a27b3/posthog/management/commands/backfill_precalculated_person_properties.py#L23) in `backfill_precalculated_person_properties.py`: same AND/OR tree walk, same `type == "person"` leaf check, same `conditionHash`/`bytecode`/`key` guard, same `PersonPropertyFilter` construction. Any change to person-filter encoding has to land in both, so they drift independently.

The PR description says filter extraction "moved to a shared `filter_extraction` module used by both the workflow and the person-properties backfill command," but no such module exists. Move the function into `filter_storage.py` (already imported here for `combine_filter_bytecodes`) and have the management command import it too.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+    GROUP BY distinct_id
+    HAVING argMax(is_deleted, version) = 0 AND max(_timestamp) >= toDateTime(%(since)s)
+    FORMAT JSONEachRow
+"""
+
+# Latest version of each precalculated_events row for a batch of overridden distinct_ids.
+# GROUP BY covers the full ReplacingMergeTree sort key (minus the constant team_id) with
+# argMax by _timestamp, so rows already corrected in an earlier pass — but not yet collapsed
+# by a background merge — don't get re-emitted forever.
+PRECALCULATED_EVENTS_BATCH_QUERY = """
```
haacked, line 116: `suggestion`: This query's `GROUP BY condition, date, distinct_id, uuid` matches the `precalculated_events` sort key exactly (`(team_id, condition, date, distinct_id, uuid)`, team_id/condition fixed in the WHERE), but it's missing `SETTINGS optimize_aggregation_in_order = 1`. The PR already added that setting to the structurally identical dedup in `hogql_cohort_query.py` and the backfill's overrides subquery, so ClickHouse falls back to full hash-based GROUP BY here. Since this groups by `uuid`, it's nearly one group per matched row, the highest-cardinality of the four new queries in this file.

[`OVERRIDES_QUERY`](https://github.com/PostHog/posthog/blob/3f16c9230b3bee9aead88ce1993132d7a83a27b3/posthog/temporal/messaging/reconcile_precalculated_data_workflow.py#L86), [`RECENT_OVERRIDES_QUERY`](https://github.com/PostHog/posthog/blob/3f16c9230b3bee9aead88ce1993132d7a83a27b3/posthog/temporal/messaging/reconcile_precalculated_data_workflow.py#L99), and [`PERSON_PROPERTIES_QUERY`](https://github.com/PostHog/posthog/blob/3f16c9230b3bee9aead88ce1993132d7a83a27b3/posthog/temporal/messaging/reconcile_precalculated_data_workflow.py#L159) have the same gap against their tables' sort keys. Add `SETTINGS optimize_aggregation_in_order = 1` before `FORMAT JSONEachRow` to all three.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+                start_to_close_timeout=dt.timedelta(hours=2),
+                heartbeat_timeout=dt.timedelta(minutes=5),
+                retry_policy=activity_retry_policy,
+            )
+            return events_result, properties_result
+
+        failed_teams: list[int] = []
+        total_corrected = 0
+        total_verdicts_corrected = 0
+        for chunk_start in range(0, len(team_ids), run_config.team_concurrency):
```
haacked, line 684: `suggestion`: Nothing exercises `ReconcilePrecalculatedEventsWorkflow.run` itself, so the per-team failure isolation this chunk loop provides has no test coverage. All existing tests use `ActivityEnvironment` to drive a single activity; the `return_exceptions=True` gather, the empty-`team_ids` early return, and the failed-team aggregation never run.

If `return_exceptions=True` were removed, or a change let one team's exception propagate out of the gather, that team aborts its chunk and skips all later chunks in the run, leaving those teams stale while tests stay green. A workflow-level test with Temporal's WorkflowEnvironment and mocked activities (one team's activity raising, others succeeding) would catch this by asserting successful teams' results are still aggregated and the raising team appears in `failed_teams`.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+    return ReconcileTeamResult(
+        overridden_distinct_ids=overridden_distinct_ids,
+        rows_checked=rows_checked,
+        rows_corrected=rows_corrected,
+        duration_seconds=duration_seconds,
+    )
+
+
+@temporalio.workflow.defn(name="reconcile-precalculated-events")
+class ReconcilePrecalculatedEventsWorkflow(PostHogWorkflow):
```
haacked: `suggestion`: This class, the registered workflow-type string (`"reconcile-precalculated-events"`), and every `RECONCILE_EVENTS_*` env var (batch size, lookback, concurrency) all say "events," but these knobs also govern `reconcile_team_precalculated_person_properties_activity`. Someone tuning `RECONCILE_EVENTS_DISTINCT_ID_BATCH_SIZE` for the events table silently changes person-properties batching too.

Since this workflow is new on this branch (no deployed schedule to migrate), renaming to `ReconcilePrecalculatedDataWorkflow` / `reconcile-precalculated-data` / `RECONCILE_PRECALCULATED_DATA_*` is free now and matches the names the PR description already uses.

posthog/temporal/messaging/test/test_reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,572 @@
...
+            result = await ActivityEnvironment().run(
+                reconcile_team_precalculated_person_properties_activity, ReconcileTeamInputs(team_id=team.pk)
+            )
+
+        assert result.distinct_ids_skipped_absent_person == 1
+        assert result.verdicts_checked == 0
+        assert result.verdicts_corrected == 0
+        mock_producer.produce.assert_not_called()
+
+    async def test_eval_failure_leaves_verdict_untouched(self, team):
```
haacked, line 488: `suggestion`: This test patches the evaluator to return `{}`, so it only covers the case where every stored condition is omitted. The verdict loop branches per `condition_hash` (present ones get corrected, absent ones left alone), but this test doesn't exercise that mix: a distinct_id with two stored verdicts where the evaluator returns one hash and omits the other.

A regression that bailed on the whole distinct_id at the first missing hash would still pass this test while silently dropping corrections for conditions that did evaluate. Add a test case with two active condition hashes where the evaluator returns only one, asserting the returned one gets corrected while the omitted one increments `verdicts_skipped_eval_failed` and isn't produced.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+            query_type="precalculated_events_reconciliation",
+        ):
+            async with get_client(team_id=inputs.team_id) as client:
+                kafka_producer = get_producer(topic=KAFKA_CDP_CLICKHOUSE_PREFILTERED_EVENTS)
+                kafka_results: list = []
+                rows_checked = 0
+                rows_corrected = 0
+                overridden_distinct_ids = 0
+
+                async def reconcile_batch(batch_overrides: dict[str, str]) -> None:
```
haacked, line 531: `nit`: Since `distinct_id` is the 4th sort-key column, `WHERE team_id = X AND distinct_id IN (batch)` can't prune on `condition`/`date`, so each batch scans the team's full `precalculated_events` slice. Fine for typical merge volumes, but a team with tens of thousands of merged distinct_ids in the window does that many full-slice scans instead of one.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -0,0 +1,734 @@
...
+            async with get_client(team_id=inputs.team_id) as client:
+                kafka_producer = get_producer(topic=KAFKA_CDP_CLICKHOUSE_PRECALCULATED_PERSON_PROPERTIES)
+                kafka_results: list = []
+                verdicts_checked = 0
+                verdicts_corrected = 0
+                overridden_distinct_ids = 0
+                distinct_ids_skipped_absent_person = 0
+                verdicts_skipped_eval_failed = 0
+
+                async def reconcile_batch(batch_overrides: dict[str, str]) -> None:
```
haacked, line 360: `nit`: This closure nests two streamed queries, three dicts, and several skip branches five levels deep, mutating `nonlocal` counters throughout. Extracting the pure decision (given verdicts + properties, which corrections to emit) into a plain function would make it unit-testable without a Kafka producer or ClickHouse client.

haacked (approved):
Ran another full pass after the latest fixes landed. Everything I flagged before (the empty-properties guess, the eval-failure bug, the missing tests) is genuinely fixed this time, verified by reading the code, not just the test names. A few suggestions inline, the main one being a duplicated filter-extraction function the PR description claims was consolidated but wasn't.

gustavohstrassburger:
Thanks @haacked — addressed in 8a46473:

- **Duplicated filter extraction**: moved `extract_person_property_filters` into `filter_storage.py` (as suggested); both the reconciliation workflow and the backfill management command now import it, and the redundant extraction tests here were dropped since the shared helper's existing suite covers them. (The PR description's `filter_extraction` name was aspirational — updated to point at `filter_storage`.)
- **Missing `SETTINGS optimize_aggregation_in_order`**: added to `PRECALCULATED_EVENTS_BATCH_QUERY`, `OVERRIDES_QUERY`, `RECENT_OVERRIDES_QUERY`, and `PERSON_PROPERTIES_QUERY`.
- **`events` naming**: renamed to `ReconcilePrecalculatedDataWorkflow` / `reconcile-precalculated-data` / `RECONCILE_PRECALCULATED_DATA_*` across the workflow, schedule, constants, and settings.
- **No `run()` coverage**: added workflow-level tests (Temporal `WorkflowEnvironment` + mocked activities) asserting a failing team is isolated (survivors still run both activities, run completes) and that empty team selection is a no-op.
- **eval-failure test only covered all-omitted**: added a mixed case — two active hashes, evaluator returns one and omits the other; asserts the returned one is corrected and only the omitted one increments `verdicts_skipped_eval_failed`.

The two nits (full-slice scan per batch; extracting the pure decision from the nested closure) are fair — leaving them for now since the batch case is bounded by typical merge volumes and the closure change is a larger refactor, but noted.

PR #70731, opened 2026-07-14
feat(autoresearch): speed up backend test suite startup and enable in-shard parallelism

thiagosalvatore:
## Problem

Backend CI spends a meaningful share of every Django shard's wall-clock on fixed startup costs that repeat identically across all ~50 shards: a no-op `migrate` against the schema-cache-primed DB, sqlx subprocess spawns for an already-migrated persons DB, heavyweight SDK imports pulled in by the DRF router, and pytest teardown GC passes that feed warnings we already suppress. Within each shard, pytest also runs single-process even though `pytest-xdist` is installed and the settings already isolate Postgres, Redis, and ClickHouse per worker.

Created with Autoresearch.

## Changes

Fixed-cost reductions, paid once per pytest session and per CI shard:

- **pytest.ini** disables the `unraisableexception` and `threadexception` plugins. Their session-cleanup `gc_collect_harder()` full-heap passes (~3s measured) only exist to raise warnings that `-p no:warnings` already suppresses, so they were pure teardown cost.
- **conftest.py** adds `_skip_noop_migrate()`, a fail-open fast path that skips Django's migrate machinery (~1.3s of graph build and project-state render, measured down to 0.05s) when every on-disk migration, discovered via `pkgutil` without importing the modules, is already recorded as applied. Any targeted invocation (`app_label`, `--fake`, `--run-syncdb`, `--check`, `--prune`) or any error falls through to the real migrate, so cold DBs and PR-delta migrations behave exactly as before.
- **posthog/conftest.py** adds `_persons_migrations_current()`: one psycopg query against `_sqlx_migrations` replaces two sqlx subprocess spawns (~0.5s → ~0.02s) when the persons DB is already fully migrated. Fail-open on any mismatch or error.
- **posthog/api/rest_router.py** drops the stale `import posthog.temporal.ai` circular-import workaround, which pulled the langgraph/chat-agent stack (~4.6s of imports) into every Django process. I verified all previously-circular import orders (`ee.hogai.core.agent_modes`, `chat_agent`, `tools`, both router-first and graph-first) now resolve without it.
- **posthog/api/wizard/** defers the `google.genai` import (~1s of pydantic model construction for its generated types module) and the `openai` SDK import from module level into the request handlers that actually call those APIs. The wizard module previously loaded both SDKs via the DRF router in every Django process. Test mocks now patch the SDK source modules instead of the wizard module attributes.
[...]

PR #70886, opened 2026-07-14
feat(autoresearch): disable pytest unraisable/thread exception plugins

thiagosalvatore:
## Problem

Every backend pytest session (and every CI shard) pays several full-heap `gc.collect()` passes at session cleanup, run by pytest's `unraisableexception` and `threadexception` plugins purely to surface `__del__`/thread exceptions as warnings. With `-p no:warnings` already in our addopts, those warnings can never escalate to failures, so the passes (~3s per session at our heap sizes, measured under cProfile) are pure teardown cost.

## Changes

Adds `-p no:unraisableexception -p no:threadexception` to `pytest.ini` addopts, with a comment explaining why they're safe to disable here.

I (Claude) ran a fixed 320-test benchmark (`posthog/api/test/{test_organization.py,test_team.py,notebooks}`) locally before/after: 24.7s → 21.8s, all green. Also validated as part of the combined branch, whose full Backend CI merge-gate matrix ran green (all Django shards, run 29340497622).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — no user-facing behavior change.

## 🤖 Agent context

Authored by Claude in a PostHog Code autoresearch session; identified via cProfile of a warm test session (gc_collect_harder dominated teardown). Split from #70731 at reviewers' request so each optimization lands independently.

---

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -18,7 +18,11 @@ junit_family = xunit1
 # nothing in this repo uses the `faker` fixture or the library (faker is a transitive dep)
 # -p pytest_boot_gc opens the boot GC window (gc.disable) before pytest-django runs
 # django.setup(), which happens before conftest files load; the root conftest closes it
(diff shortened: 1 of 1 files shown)


webjunkie (approved):
Saving checks out (~3s on a 64-test file here).

Only `unraisableexception` runs `gc_collect_harder` though — `threadexception` doesn't import gc at all, so the comment's "both plugins" is wrong. Disabling it is harmless, just buys nothing.

PR #70888, opened 2026-07-14
feat(autoresearch): skip sqlx spawns when persons test DB is migrated

thiagosalvatore:
## Problem

`run_persons_sqlx_migrations(keepdb=True)` spawns two sqlx subprocesses (`database create` + `migrate run`, ~0.5s) in every pytest session — including every CI shard and every future xdist worker — even when the persons DB is already fully migrated.

## Changes

Adds `_persons_migrations_current()` to `posthog/conftest.py`: one short psycopg query compares the on-disk migration versions in `rust/persons_migrations/` against the `_sqlx_migrations` table, and skips both subprocess spawns when nothing is pending (~0.5s → ~0.02s). Fail-open: any mismatch, missing DB, or error runs the real sqlx path, and `keepdb=False` (fresh-DB) runs are untouched.

I (Claude) verified locally: warm skip (~0.02s), fall-through when the persons DB doesn't exist, and the 320-test API benchmark green. Also validated on the combined branch, whose full Backend CI merge-gate matrix ran green (run 29340497622).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — test-infrastructure only.

## 🤖 Agent context

Authored by Claude in a PostHog Code autoresearch session. Split from #70731 at reviewers' request.

---

### DIFF
--- a/posthog/conftest.py
+++ b/posthog/conftest.py
@@ -185,6 +185,30 @@ def reset_clickhouse_tables():
     run_clickhouse_statement_in_parallel(list(CREATE_DATA_QUERIES()))
 
 
+def _persons_migrations_current(database_url: str, migrations_path: str) -> bool:
+    """True when every migration file on disk is already recorded in _sqlx_migrations.
+
+    One short psycopg query instead of two sqlx subprocess spawns (~0.5s per pytest
+    session, and per xdist worker). Fail-open: any error means "not current" and the
+    real sqlx path runs.
+    """
+    import psycopg  # noqa: PLC0415 — deferred so conftest import stays cheap
+
+    try:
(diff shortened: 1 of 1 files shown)


webjunkie (commented):
Shape's good. Two things — the query ignores `success` and `checksum`.

sqlx writes the row with `success=false` before running and flips it after, so a crashed migration leaves the dirty state sqlx normally refuses to run past, and this reads it as applied. `WHERE success` fixes it.

Checksum: editing an already-applied migration is the normal loop while writing one. Today sqlx errors loudly, after this it's a silent skip against stale schema. Either compare (sha384 of the file bytes) or note the trade in the docstring.

PR #70903, opened 2026-07-14
chore: record error-tracking autocapture receiver in setup baseline

rnegron:
## Problem

Master has been red on Backend CI since [feat(error-tracking): move autocapture opt-in to settings (phase 1)](https://github.com/PostHog/posthog/pull/70701) merged. That PR added a new `post_save` receiver (`mirror_autocapture_opt_in_to_settings`) that connects at `django.setup()`, but didn't record it in the committed startup-receiver baseline. `test_setup_receivers_match_baseline` pins the exact set of first-party receivers wired at setup, so the unrecorded one fails the guard on every master run.

## Changes

- Add the one receiver line to `posthog/test/setup_receivers_baseline.txt`, in sorted position.

The receiver's wiring is already correct (import-light `signals.py` imported from `ErrorTrackingConfig.ready()` with `# noqa: F401`), and `test_no_new_heavy_imports_at_setup` still passes, so recording it is the right fix rather than restructuring.

- Reproduced the guard's `django.setup()` receiver capture in a clean subprocess and confirmed the updated 217-line baseline is a zero-diff exact match for the live receiver set.
- Did not run the full pytest test locally: my dev shell exports `DEBUG=1`, which leaks a settings warning onto the subprocess stdout and would false-fail the assertion. CI has no such artifact.

## 🤖 Agent context

- Raúl asked me (Claude, via Claude Code) to triage a red-master Slack alert and fix if possible.
- Invoked the `/django-startup-time` skill to confirm the guard's intent and that "record in baseline" (not defer/restructure) was the correct fix here.
- Verified the culprit commit, confirmed the receiver is import-light and `ready()`-wired, and validated the baseline diff before committing.


rnegron:
Replaced by https://github.com/PostHog/posthog/pull/70869

PR #70910, opened 2026-07-14
feat(personhog): recover evicted dirty persons from the changelog

z0br0wn:
## Problem

A leader cache miss falls back to Postgres unconditionally. Postgres only reflects what the writer has applied, so whenever the writer lags, the fallback can serve state older than what the leader already acked. Foyer evicts freely under capacity pressure, which makes this a data-loss path: a dirty entry gets evicted, the next operation reloads the stale PG row, and later merges build on the stale base.

The e2e test harness reproduces this deterministically. With a 10-entry cache, 50 persons, and an 8s writer pause, a 15s run acks every write and then loses 8,603 of them.

## Changes

**Before** – every miss trusts PG:

  ```mermaid
  flowchart LR
      R[Request] --> C{Cache hit?}
      C -- yes --> S[Serve]
      C -- miss --> PG[(PG fallback)]
      PG --> S2[Serve, stale if the writer lags]
      class R,S,S2 phYellow
      class C phBlue
      class PG phGray
  ```

**After** – the miss path never trusts a stale source:

  ```mermaid
  flowchart LR
      R[Request] --> C{Cache hit?}
      C -- yes --> S[Serve]
      C -- miss --> D{Dirty index marked?}
      D -- yes --> K[(Changelog point-read at the marked offset)]
      K -- ok --> S
      K -- fail --> U[UNAVAILABLE, retryable]
      D -- no --> PG[(PG fallback, known current)]
      PG --> S
      class R,S phYellow
      class C,D phBlue
      class K,PG phGray
      class U phRed
  ```

- **Dirty index**: every acked produce marks `(team, person)` with the record's version, offset, and partition, taken from the Kafka delivery report. A marked person's latest state may not be in PG yet; an unmarked person's PG row is known current.
- **Changelog recovery**: a miss on a marked person fetches the single record at the marked offset. Records are full state and the topic compacts by key, so one point-read is always sufficient. One long-lived consumer is shared across recoveries with the assignment swapped per fetch. The record must carry the mark's exact key and version, since both derive from the same state a mismatch means acked state and the produced record diverged, and recovery fails loudly rather than serving it.
- **Pruning and seeding**: a periodic loop drops marks once the writer's committed offsets pass them (one batched OffsetFetch per tick). Warming seeds marks for records at or above the committed offset, and releasing a partition clears its marks.
- **Bounded backpressure**: the index is soft-bounded (default 10M entries, roughly 1 GB). At capacity, writes for unmarked persons shed with `RESOURCE_EXHAUSTED` before producing. A durable-but-unmarked write would silently reopen the stale-fallback hole, and unbounded growth during a long writer outage would OOM into a crash loop that replays the same backlog through warming.
- **Metrics**: person loads by source and outcome (cache, changelog, pg) with duration histograms, cache departures by reason, dirty index size, per-partition writer lag, prune and shed counters.
- **CI**: the eviction-under-writer-lag scenario joins the personhog gate as a regression step, and the harness README's known-defect entry moves to fixed-and-gated.

  > [!WARNING]
  > Two behavior changes on the leader path. A failed recovery for a dirty person now returns retryable `UNAVAILABLE` instead of silently serving stale PG state. And when the dirty index is at capacity (a writer outage measured in hours), writes for previously unmarked persons shed with `RESOURCE_EXHAUSTED` until the writer catches up.

  > [!NOTE]
  > Deployment requirement: recovery depends on `compact` being present in the changelog topic's `cleanup.policy`. Dev declares `personhog_updates` as `compact,delete` with 7-day retention, which is compatible (marks only outlive the writer's committed offset, and a mark somehow outliving retention fails loudly, never stale). Production terraform does not declare this topic yet, so topic config belongs on the production rollout checklist.

All automated. 72 leader tests pass (`cargo test -p personhog-leader`), clippy with `-D warnings` and fmt clean.

New tests and the regression each one catches:

- `dirty_index.rs` unit tests: admission denied at capacity except for marked keys, marks keep the newest offset, prune respects the committed-offset boundary, partition scoping, and the size counter staying in step with inserts, updates, and removals.
- `evicted_dirty_person_recovers_from_changelog`: the core fix. The service is built with no PG pool at all, so a recovery path that regressed to PG fallback returns NotFound and fails the test.
- `dirty_person_with_failed_recovery_is_unavailable_not_stale`: a mark pointing past the high-water mark must produce `UNAVAILABLE`, catching any change that reintroduces the stale fallback on recovery failure.
- `recovery_fails_when_record_version_disagrees_with_the_mark`: the version tripwire, catching any future divergence between acked state and the produced record.
- `writes_shed_when_dirty_index_is_full`: a full index sheds new persons with `RESOURCE_EXHAUSTED` while marked persons stay admitted.
- `warming_seeds_dirty_index_when_writer_has_no_commits`: warming must mark every record when the writer has no committed offset, catching a seeding regression that would strand post-warm evictions on stale PG.

End to end with the e2e harness, same command on both sides of the fix:

  ```bash
  personhog-test-harness gate --leaders 3 --partitions 8 --persons 50 --concurrency 10 \
    --duration 15s --cache-capacity 10 --writer-pause-after 3s --writer-pause-duration 8s
  ```

  - Against a leader built from this branch's base commit: 8,603 consistency violations.
  - With this branch: three consecutive runs, zero violations, zero failed writes.

The drain + zombie + writer-lag chaos composite also runs green. The determinator package-rule marks the harness affected by leader changes, so the full gate (including the new eviction step) runs on this PR.

## Docs update

No product docs impact. The harness README's known-defects section is updated in this PR.

## 🤖 Agent context

Built with Claude Code. The `/writing-tests` skill shaped the test selection, and the harness's `extending-personhog-test-harness` skill drove the CI scenario addition (red before the fix, green after, three consecutive green runs before wiring into CI).

Design alternatives we rejected along the way: pinning dirty entries in the cache (unbounded memory during a writer outage), a hybrid disk cache tier (still probabilistic, heavyweight), and gating reads on writer progress (couples leader availability to PG, which the leader design exists to avoid). The bounded shed exists because an OOM restart would replay the same backlog through warming and crash-loop. A hot-path review pass replaced `DashMap::len()` in write admission with an atomic size counter and collapsed the prune loop to a single pass over the index.


rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /// Drop every mark the writer has applied, across all partitions, in a
+    /// single pass: a mark is applied once its partition's committed offset
+    /// (absent = no commit yet) has moved past it. Returns how many were
+    /// pruned.
+    pub fn prune_applied(&self, committed: &HashMap<u32, i64>) -> usize {
+        let mut removed = 0;
+        self.map.retain(|_, mark| {
+            let applied = committed
```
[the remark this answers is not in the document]
z0br0wn: i can add a guard in a follow up pr, but the leader service should only read from the primary db for strongly consistent reads, so this shouldn't be a problem

eli-r-ph (approved):
Good idea, worth trying out.

Nonblocking, but may be worth considering: I'm concerned using a single global serialized consumer to do point lookups will really bog down though under contention. I have had better luck with a pool of single-partition low level consumers for this sort of thing.

The pool could populate as new partitions show up needing a point lookup, registering with a partition (expensive part) and then stick around for reuse on the same partition using fast `seek()`s

rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,295 @@
...
+                max_offset: i64::MIN,
+            });
+            s.count += 1;
+            s.max_offset = s.max_offset.max(entry.offset);
+        }
+        stats
+    }
+
+    /// Highest marked offset for a partition, if any.
+    pub fn max_offset(&self, partition: u32) -> Option<i64> {
```
nickbest-ph, line 336: nit: this doesn't look like its used and is an O(n)/expensive operation

could be commented such or removed

rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Drop marks on `partition` that the writer has applied (offset below
+    /// its committed offset). Returns how many were pruned.
+    pub fn prune_partition(&self, partition: u32, committed: i64) -> usize {
+        self.prune_applied(&HashMap::from([(partition, committed)]))
+    }
+
+    /// Drop every mark on `partition` — used when the partition is released
+    /// to another owner, whose warming rebuilds its own marks.
+    pub fn clear_partition(&self, partition: u32) -> usize {
```
nickbest-ph, line 297: this might introduce a small race between `release_partition` and an inflight read for the person miss path.

`recover_or_load` treats no mark in the dirty index as PG is trusted for this person, but the release clears the marks, so:

1. reader misses the cache, partition still owned, person absent, enters `recover_or_load`
2. `release_parititon` runs, clears marks
3. reader does `dirty_index.get` (just cleared), then loads from pg which is stale

We could re-check partition ownership when the index returns no mark before falling back to PG, then return some error code like `FAILED_PRECONDITION` if the partition is gone

nickbest-ph (approved):
nice, overall lgtm

there might be a small race introduced when there are inflight reads to a missed person while we revoke a partition. worth checking to see if you think its valid/writing a test to see if its valid.

rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,295 @@
...
+///
+/// The index is soft-bounded by `max_entries`: under a sustained writer
+/// outage it would otherwise grow one mark per unique person written —
+/// gigabytes over hours at production churn, and an OOM restart replays
+/// the same backlog through warming. Write admission (`can_admit`) is
+/// checked before producing, so the bound sheds new write volume instead
+/// of ever refusing to record a fact: `mark` itself never fails, because
+/// warming must be able to record marks for records that are already
+/// durable regardless of the bound.
+pub struct DirtyIndex {
```
jose-sequeira, line 84: nit: Max entries could be good to export as a gauge so we can derive percentage full current_size/max_entries and alert on percentage full

rust/personhog-leader/src/recovery.rs
```diff
@@ -0,0 +1,249 @@
...
+        let Some(slot) = self
+            .consumers
+            .get(&partition)
+            .map(|entry| Arc::clone(entry.value()))
+        else {
+            return Err(format!(
+                "no recovery consumer for partition {partition}; partition not owned"
+            ));
+        };
+        let consumer = slot.lock().await;
```
jose-sequeira: This lock feels concerning, we're essentially blocking the consumer slot for all person fetches per partition across all retries and sleeps. Not only this makes every recovery serialized for all persons in a partition but their wait time is amplified by all the retries.

I think that at minimum we should release the lock while we sleep. Additionally we could improve throughput by allowing a pool of consumers per partition, this way we can parallelize reads for a partition until a certain threshold

rust/personhog-leader/src/warming.rs
```diff
@@ -357,6 +378,27 @@ pub async fn warm_from_kafka(
...
 
+    // Records at or above the writer's committed offset are not yet in PG:
+    // seed the dirty index so that, if the cache later evicts them, a miss
+    // recovers from the changelog instead of trusting a stale PG row. With
+    // no committed offset the writer has applied nothing, so every record
+    // is marked. Seeding happens before the install publishes the
+    // partition, so no request can observe the cache without the marks.
+    let mut seeded = 0u64;
+    for (key, cached, offset) in &buffered {
+        if committed_offset.is_none_or(|committed| *offset >= committed) {
```
jose-sequeira, line 389: The `warming_seeds_dirty_index_when_writer_has_no_commits` test covers the `none` invariant of this arm. But there's no test with a commited offset

rust/personhog-leader/src/service.rs
```diff
@@ -145,10 +242,11 @@ impl PersonHogLeaderService {
 
         // Double-check cache -- another request may have loaded it
         if let CacheLookup::Found(person) = self.cache.get(partition, key) {
+            Self::record_cache_hit();
             return Ok(person);
         }
 
-        self.load_from_pg(partition, key).await
+        self.recover_or_load(partition, key).await
```
jose-sequeira, line 272: The other comment I made about keeping a lock through boundaries is amplified here since this method is called while still keeping the per key lock and preserving it through that boundary as well

rust/personhog-leader/src/main.rs
```diff
@@ -302,3 +329,63 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    prune_interval: Duration,
+) {
+    let mut interval = tokio::time::interval(prune_interval);
+    loop {
+        interval.tick().await;
+        // One pass over the index for the read side, one batched
+        // OffsetFetch, one retain pass for the prune — the loop's cost
+        // stays linear in the index even when a lagging writer has made
+        // it large.
+        let stats = dirty_index.partition_marks();
```
jose-sequeira: I'm quite worried in terms of performance of this approach at scale, at 10M entries, `partition_marks` and `prune_applied` require a full `iter()` and a `retain()`.

`iter()` needs a read lock and `retain()` a write lock. This loop runs every 5s and those locks directly clash with `mark()` on the write hot path and `get()` on every cache miss. 

Don't think it's a blocker, but something to note

rust/.config/nextest.toml
```diff
@@ -1,3 +1,18 @@
+# The k3s integration tests each boot a privileged single-node Kubernetes
+# cluster in docker. Their #[serial] attributes only lock within a process,
+# and nextest runs one process per test, so without this group the clusters
+# boot concurrently and starve each other (and the runner) — the K8s-aware
+# assignment paths then stall on a slow API server and time out. max-threads
+# serializes them across processes; threads-required keeps the rest of the
+# suite from saturating the runner while a cluster is up.
+[test-groups]
+k3s = { max-threads = 1 }
```
jose-sequeira, line 9: Is this relevant for the PR?

rust/common/k8s-awareness/src/watcher.rs
```diff
@@ -402,7 +405,8 @@ async fn run_statefulset_watcher(
     let api: Api<StatefulSet> = Api::namespaced(client.clone(), namespace);
     let config = WatcherConfig::default().fields(&format!("metadata.name={}", controller.name));
 
-    let stream = watcher::watcher(api, config);
+    // Same backoff rationale as the deployment watcher above.
+    let stream = watcher::watcher(api, config).default_backoff();
```
jose-sequeira, line 409: Is this related to the PR?

rust/.config/nextest.toml
z0br0wn: not directly - sorry, just fixing a flakey test that i was hitting in ci here (same to the comment below)

rust/personhog-leader/src/cache/dirty_index.rs
z0br0wn: good call out, and will fix. i think this warrants a follow up pr to have the router retry on failed precondition, because right now this will pass through to the client (not the end of the world, but would be nice to obfuscate). will push up a fix in a few

rust/personhog-leader/src/recovery.rs
z0br0wn: yeah great call out - i dont think we need the per partition consumers at all here, it made sense when i was trying out seek instead of assign / fetch, but it ended up being much slower. i'm going to just move this to having a general pool of available consumers (just like a db connection pool) which will be easier to manage.

rust/personhog-leader/src/main.rs
z0br0wn: yup, totally valid. gonna take a pass at adding a separate fifo map index which will make this much lighter weight

rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,506 @@
...
+    /// its committed offset). Pops the reclaim queue only while its head
+    /// is applied, so the cost is proportional to the marks reclaimed.
+    /// The pass releases the queue lock every `PRUNE_CHUNK` pops: a
+    /// catch-up after a long writer outage reclaims millions of marks in
+    /// one call, and an unbroken hold would stall every new-key `mark` —
+    /// the write hot path — for the whole pass. Between chunks the queue
+    /// is re-fetched, so a concurrent `clear_partition` (which detaches
+    /// it) ends the pass instead of racing the drain. Returns how many
+    /// were pruned.
+    pub fn prune_partition(&self, partition: u32, committed: i64) -> usize {
```
nickbest-ph, line 213: i dont think this is covered by any unit test. probably worth covering as logic here is non-trivial

nickbest-ph (approved):

rust/personhog-leader/src/cache/dirty_index.rs
```diff
@@ -0,0 +1,556 @@
...
+            }
+            budget -= 1;
+            let (_, key) = queue.pop_front().expect("front was just observed");
+            // The removal re-checks the map's offset under the shard
+            // lock: between the pop and here the key may have been
+            // re-marked, and removing the newer mark would silently
+            // reopen the stale-fallback hole.
+            if self
+                .map
+                .remove_if(&key, |_, mark| mark.offset < committed)
```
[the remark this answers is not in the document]
z0br0wn: not really a real concern from a security perspective, and not new behavior - but, this means we should introduce leader side validation so we reject up front instead of quietly down stream at the writer. i'll log that as a follow up task, it's a changeset on its own.

rust/personhog-leader/src/recovery.rs
```diff
@@ -0,0 +1,250 @@
...
+            .await
+            .map_err(|_| "recovery pool closed".to_string())?;
+        histogram!("personhog_leader_recovery_pool_wait_ms")
+            .record(wait_start.elapsed().as_secs_f64() * 1000.0);
+        let consumer = self
+            .idle
+            .lock()
+            .expect("recovery pool lock poisoned")
+            .pop_front()
+            .expect("a permit guarantees an idle consumer");
```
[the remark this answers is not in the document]
z0br0wn: fixed

PR #70937, opened 2026-07-15
feat(web-analytics): skip fresh teams and fail loudly on empty audience in eager backfill

lricoy:
## Problem

Two backfill-run failure modes surfaced while running `web_analytics_eager_backfill` across the full active-WA audience:

1. **A relaunched backfill spends hours re-walking already-warm teams.** The op sorts eligible teams by staleness but still processes every one of them — each warm team costs a full tile-matrix verification pass (~24 read queries) before the run reaches the never-warmed tail. At fleet scale that's thousands of teams of pure verification, and in the worst case the runtime budget expires before the cold tail is reached at all.
2. **A transient ClickHouse error silently turns a backfill into a no-op.** The active-audience fetch is best-effort (returns `[]` on any error) so the hourly warmer never skips a cycle — but a backfill run that gets `[]` warms only the small static list and reports SUCCESS. This happened in production: an `ALL_CONNECTION_TRIES_FAILED` blip during the audience scan produced a 14-minute "successful" backfill that warmed nothing (`active_teams=0` in the run's output metadata).

## Changes

- New `skip_fresh_hours` config on `EagerWarmConfig` (default `None` = current behavior): teams whose latest READY `PreaggregationJob` in the baseline window is fresher than the threshold are dropped from the walk instead of merely sorted last. Teams with no precompute jobs at all are never skipped — they're the audience the backfill exists for. The backfill job's default config sets it to 24h; the hourly warmer keeps `None` (its static audience is small and should always be topped up).
- The backfill job now raises `dagster.Failure` when `active_teams_pct > 0` and the audience fetch returns no teams, so a transient ClickHouse error produces a visible failed run to relaunch instead of a silent no-op success. The hourly job's best-effort behavior is unchanged.

- `test_skip_fresh_hours_drops_only_freshly_warmed_teams` (parameterized): with the flag on, a freshly-warmed team is skipped while stale and never-warmed teams still process (guards the sentinel regression where never-warmed teams would be silently skipped, no-opping the backfill); with `None`, all teams process — pins the hourly job's unchanged behavior.
- `test_backfill_fails_when_active_audience_fetch_returns_nothing`: executes the backfill job in-process with the fetch patched to `[]` and asserts the run fails without warming — the exact production incident.
- Full dag suite: 25 passed.

## 🤖 Agent context

- Built during live monitoring of the full-fleet backfill; both failure modes were observed in production runs (the silent no-op via `active_teams=0` in run output metadata, the slow warm-team walk via query_log insert/verify ratios).
- Skills invoked: /writing-tests.

### DIFF
--- a/products/web_analytics/dags/eager_web_analytics_precompute.py
+++ b/products/web_analytics/dags/eager_web_analytics_precompute.py
@@ -468,6 +468,24 @@ def warm_eager_baseline_op(context: dagster.OpExecutionContext, config: EagerWar
     team_ids, gate_reason, diagnostics = _resolve_eager_audience(
         active_teams_pct=config.active_teams_pct, active_lookback_hours=config.active_lookback_hours
     )
+    # The active-audience fetch is best-effort ([] on any ClickHouse error) so the
+    # hourly warmer never skips a cycle. For the backfill that leniency is wrong: a
+    # transient cluster error would silently shrink a fleet-wide run to the static
+    # lists and the run would report SUCCESS having warmed nothing (this happened —
+    # an ALL_CONNECTION_TRIES_FAILED blip produced a 14-minute no-op "backfill").
+    # Fail loudly instead so the launcher knows to relaunch.
+    if (
+        job_name == "web_analytics_eager_backfill"
+        and config.active_teams_pct > 0
+        and not diagnostics.get("active_teams")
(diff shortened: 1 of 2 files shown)


lricoy:
Not needed anymore since we moved to the on-demand warmer

PR #70944, opened 2026-07-15
fix(flags): sync flag definitions into EU HyperCache from US

haacked:
## Problem

Our backend's internal ("dogfood") feature flag checks use `only_evaluate_locally=True` against team 2's flag definitions. Team 2 only exists in the US Postgres, so EU's signal-driven HyperCache writer has nothing to build team 2's entry from, and EU local-eval can't resolve the real dogfood flags. The current stopgap pins `POSTHOG_SELF_TEAM_ID=1` on EU, which evaluates against team 1's flags instead of the actual dogfood flags.

## Changes

Before, EU pods read team 1's cache entry via the override, and team 2's entry was permanently empty:

```mermaid
flowchart LR
  Pods[EU pods] -->|read team 1 entry via POSTHOG_SELF_TEAM_ID=1| HC[(EU HyperCache)]
  HC -.->|team 2 entry: empty, nothing to build from| X[US-only team 2 rows]
  class Pods phBlue;
  class HC,X phGray;
```

After, one periodic task keeps team 2's entry fresh and every pod reads the shared cache exactly as today:

```mermaid
flowchart LR
  Beat[Celery beat task, every 30s] -->|GET /flags/definitions with PSAK + If-None-Match| US[US region]
  Beat -->|write team 2 entry on 200| HC[(EU HyperCache)]
  Pods[EU pods] -->|read team 2 flags| HC
  class Beat,Pods phBlue;
  class US phRed;
  class HC phGray;
```

- New `sync_cross_region_dogfood_flags()` in `products/feature_flags/backend/cross_region_flag_sync.py`: EU-only sync that fetches team 2's flag definitions from the US region's authenticated `/flags/definitions` endpoint with a project secret API key and writes the payload into EU's existing `flag_definitions_hypercache`. Only this one task polls; pods never hit the US API directly, which is the whole point of HyperCache.
- Uses the endpoint's ETag support: sends `If-None-Match` with the locally cached ETag, so an unchanged upstream is a 304 with no payload and no write. On a 200 the write is unconditional, because this bare-int cache key isn't covered by the hourly expiry-tracking sweep, so the write is what re-stamps the Redis TTL.
- Fails safe on network errors, non-200s, bad JSON, and non-dict payloads: logs a warning (with throttled error-tracking capture) and keeps the existing cache entry.
- The Celery task runs on `FEATURE_FLAGS_LONG_RUNNING`, not the fast `FEATURE_FLAGS` queue, so a stalled cross-region HTTP call can't delay the sub-second signal-driven cache rebuilds. Scheduled every 30s (matching the SDK's default poll interval) and registered only in EU.
- New `POSTHOG_FLAGS_PROJECT_SECRET_TOKEN` setting, a PSAK scoped to team 2. Unset disables the sync, so this PR is a safe no-op until the companion charts change provisions the secret and removes the EU `POSTHOG_SELF_TEAM_ID=1` override.

Origin: [Slack thread](https://posthog.slack.com/archives/C0351B1DMUY/p1783626343748319).

New suite `test_cross_region_flag_sync.py` (8 tests, all passing): region and token no-op guards, ETag header sent + 304 skips the write, 200 writes the payload and arms the ETag, fail-safe on request exceptions and bad responses (500, invalid JSON, non-dict payload). Also ran the adjacent suites (`test_local_evaluation.py`, `test_scheduled.py`, `test_tasks.py`), 92 passing, no regressions.

Also verified end to end against a local dev stack, running the sync in a Django shell pointed at the local Rust `/flags/definitions` service with a real project secret API key (scoped to local team 2, matching the hardcoded team id):

- 200 path: full payload (flags, cohorts, group_type_mapping) fetched with PSAK auth and written into HyperCache under key 2, ETag armed.
- 304 path: with the cache warm, the sync sent `If-None-Match` and the upstream answered 304, zero cache writes.
- Auth failure: a bogus token got a real 401; the sync logged a warning and left the existing cache entry intact.
- Upstream 503 (cache miss on the serving side): logged a warning, no write, no crash.

Not covered locally: the Celery beat wiring itself (30s EU-gated registration), which is plain schedule registration.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/products/feature_flags/backend/cross_region_flag_sync.py
+++ b/products/feature_flags/backend/cross_region_flag_sync.py
@@ -1,0 +1,109 @@
+"""
+Cross-region sync of the dogfood team's (team 2) flag definitions into EU's
+HyperCache.
+
+Every other team's `flag_definitions_hypercache` entry is rebuilt locally by
+Django signals reading straight from Postgres (see local_evaluation.py). Team 2
+-- PostHog's internal "self" team, used for our own dogfooded local evaluation
+-- lives only in the US Postgres, so EU has no rows to build it from. Rather
+than have every EU pod poll the US API directly (which HyperCache exists to
+avoid), a single periodic task fetches team 2's flag definitions from the US
+region's authenticated `/flags/definitions` endpoint and writes the result
+straight into EU's HyperCache, so every EU pod still reads from the shared
+cache with zero per-pod polling.
+
+No-op on US, where team 2's HyperCache entry is already populated by the
+normal signal-driven path.
+"""
+
+from django.conf import settings
+
+import requests
+import structlog
+from posthoganalytics.request import US_INGESTION_ENDPOINT
+
+from posthog.utils import capture_exception_throttled, get_instance_region
+
+from products.feature_flags.backend.local_evaluation import flag_definitions_hypercache
+
+logger = structlog.get_logger(__name__)
+
+# PostHog's own internal "self" team, canonically team 2 (see _build_flag_provider
+# in posthog/utils.py). Only its US Postgres rows are authoritative.
+DOGFOOD_SELF_TEAM_ID = 2
+
+_FLAGS_DEFINITIONS_PATH = "/flags/definitions"
+_REQUEST_TIMEOUT_SECONDS = (3, 10)  # (connect, read)
+
+# Throttle window for capturing an unexpected failure (network error, bad JSON) to
+# PostHog error tracking, shared across the 30s schedule so a sustained outage
+# reports once per window instead of once per tick.
+_SYNC_FAILURE_CAPTURE_THROTTLE_KEY = "cross_region_dogfood_flags_sync_capture_throttle"
+_SYNC_FAILURE_CAPTURE_THROTTLE_TTL = 300  # seconds
+
+
+def sync_cross_region_dogfood_flags() -> None:
+    """Refresh team 2's flag-definitions HyperCache entry in EU from the US region.
+
+    No-op outside EU or when the PSAK isn't configured. Uses the endpoint's ETag
+    support: sends `If-None-Match` with the locally cached ETag, so an unchanged
+    upstream is a 304 with no payload transferred and no local write.
+    """
+    # Defense in depth: the beat registration in scheduled.py is also EU-gated.
+    # This keeps direct invocation (shell, tests) safe outside EU.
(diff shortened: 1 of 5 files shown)


products/feature_flags/backend/cross_region_flag_sync.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if response.status_code == 304:
+        return
+
+    if response.status_code != 200:
+        logger.warning(
+            "cross_region_dogfood_flags_sync_bad_status",
+            status_code=response.status_code,
+        )
+        return
```
matheus-vb, line 86: The non-200 and non-dict paths only logger.warning, no `capture_exception_throttled` like the `network`/`JSON` paths, and no metric since they return before `update_cache`. The most likely sustained failure, a rotated or expired PSAK returning a steady 403 (or a US 5xx), would then degrade EU dogfood flags silently until the 7-day TTL empties the entry, with nothing to alert on. Since this task is the only source of team 2’s EU cache entry, probably worth a counter labeled by outcome, or at least a capture on `status_code` >= 500 and 403.

products/feature_flags/backend/cross_region_flag_sync.py
```diff
@@ -0,0 +1,109 @@
...
+    if get_instance_region() != "EU":
+        return
+
+    token = settings.POSTHOG_FLAGS_PROJECT_SECRET_TOKEN
+    if not token:
+        logger.debug("cross_region_dogfood_flags_sync_no_token")
+        return
+
+    headers = {"Authorization": f"Bearer {token}"}
+    local_etag = flag_definitions_hypercache.get_etag(DOGFOOD_SELF_TEAM_ID)
```
matheus-vb, line 63: We recompute the etag locally and send it as `If-None-Match` instead of echoing the server’s ETag header. This works today only because the payload is fully string-keyed (`cohorts` and `group_type_mapping`). If an int-keyed dict is ever added, `sort_keys=True` sorts int keys numerically here but lexicographically after the JSON round-trip ("10" before "2"), so the etag stops matching, the 304 never fires, and we do a full rewrite every 30s (silently), and breaking the “only logs when changed” claim. Echoing the response ETag would remove the coupling.

matheus-vb (approved):
Nice! Two non-blocking suggestions.

PR #70965, opened 2026-07-15
fix(slo): trustworthy delivery metric and NUL-safe snapshot writes

vdekrijger:
## Problem

The **SLO Monitoring** dashboard's "28d Success Rate" insight could misrepresent subscription-delivery health in two ways:

- **Warm-up ramp.** The 28-day rolling window was computed over only 28 base rows, so the earliest ~27 points had partial windows and rendered as an artificial ramp on the left edge of the chart — easy to misread as a recent decline when the current value is actually healthy.
- **Day-boundary phantom failures.** Success/failure was bucketed by calendar day (`greatest(daily_starts − daily_successes, 0)`), so a delivery that started late in a day and finished after midnight was counted as a failure on the start day and dropped on the next. For batch-heavy operations this manufactured a large number of phantom failures.

Separately, a subscription delivery could **fail its entire write** when snapshot content contained a NUL byte (`\x00`): Django's JSONField serializes NUL to a unicode escape that Postgres text/jsonb columns reject, raising a `DataError`. This affected insight query results, and — unguarded — AI report markdown, step diagnostics, the user prompt, and query-error messages.

## Changes

- **SLO success-rate query** (`slo-monitoring/insights.tf`): pair starts to completions by `correlation_id` (the approach the burn-rate query on the same dashboard already uses) so cross-midnight deliveries are attributed to their start day; compute over 56 days but display only the fully-warmed last 28. Uncorrelated events keep the per-day fallback. Also dropped a redundant `sample_rate` round-trip (`weight` is already `1/sample_rate`).
- **NUL scrubbing**: a shared recursive `strip_null_bytes` helper (keys and values) applied at every untrusted ingress that reaches `content_snapshot` — insight results, query-error messages, AI markdown/diagnostics, and the user prompt. The large insight payload stays behind a cheap byte-level gate so clean deliveries skip the walk.

**Why:** the dashboard metric was firing on artifacts rather than real regressions, and the NUL `DataError` was failing real deliveries outright.

- Added `test_serialize_insight_result_strips_null_bytes` covering NUL in result strings, map-derived dict keys, and columns; it round-trips through stdlib `json` (what JSONField uses) and asserts the output is NUL-free.
- Verified the `strip_null_bytes` algorithm (keys + values at all nesting depths, plus the escape-detection gate) with a standalone simulation.
- Validated the rewritten HogQL query against live data: the warm-up ramp is gone and the day-boundary phantom failures no longer appear, while genuine failures are still reflected; the `sample_rate`→`weight` simplification is result-identical.
- Ran `ruff format` / `ruff check` on all changed Python.
- I was **not** able to run the pytest suite in this environment (test dependencies weren't installed here); the new test should be run in CI.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — DRI assigned as the PR assignee.

Investigation started from an SLO snapshot report that flagged a subscription-delivery success-rate drop. Digging into the data showed the "drop" was mostly the dashboard metric's warm-up + day-boundary artifacts (true per-operation success rate is high), which motivated the query rewrite; the NUL `DataError` surfaced as a recurring real delivery failure in the same data.

- Skills invoked: `/simplify` (removed the redundant `sample_rate` round-trip; promoted the NUL helper to a shared function and applied it on the AI-report path) and `/code-review` (high effort). The review caught that `strip_null_bytes` scrubbed dict values but not keys — a NUL in a `Map(String, …)` key would still fail the write — now fixed and tested. A confirmation pass then flagged that query-error messages bypassed the scrub; closed at the source.
- Decision: scrub at each untrusted ingress (gated for the large insight payload) rather than one blanket pre-save walk, to avoid re-walking multi-MB payloads on every delivery.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B9A53J8RF/p1784094998253529?thread_ts=1784094998.253529&cid=C0B9A53J8RF)*

### DIFF
--- a/products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
+++ b/products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
@@ -28,8 +28,8 @@ from products.exports.backend.temporal.subscriptions.delivery_common import (
     auto_disable_and_return,
     deliver_email,
     deliver_slack,
+    strip_null_bytes,
 )
-from products.exports.backend.temporal.subscriptions.insight_snapshot import strip_null_bytes
 from products.exports.backend.temporal.subscriptions.types import (
     AI_REPORT_DIAGNOSTICS_KEY,
     AI_REPORT_PROMPT_SNAPSHOT_KEY,
--- a/products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
+++ b/products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
@@ -29,6 +29,7 @@ from products.exports.backend.temporal.subscriptions.delivery_common import (
     deliver_email,
     deliver_slack,
 )
+from products.exports.backend.temporal.subscriptions.insight_snapshot import strip_null_bytes
 from products.exports.backend.temporal.subscriptions.types import (
     AI_REPORT_DIAGNOSTICS_KEY,
     AI_REPORT_PROMPT_SNAPSHOT_KEY,
@@ -105,13 +106,15 @@ async def _persist_ai_report(delivery_id: uuid.UUID, result: AiReportResult, pro
         # No DoesNotExist guard: create_delivery_record always writes this row before
         # generation runs, so a missing row is a wiring bug — let it raise loudly.
         delivery = SubscriptionDelivery.objects.get(pk=delivery_id)
+        # LLM output and the user prompt can carry NUL bytes that Postgres text/jsonb reject;
+        # scrub them here (payloads are small) as they are the untrusted inputs on this write path.
         delivery.content_snapshot = {
             **(delivery.content_snapshot or {}),
-            AI_REPORT_SNAPSHOT_KEY: result.markdown,
-            AI_REPORT_DIAGNOSTICS_KEY: [dataclasses.asdict(d) for d in result.diagnostics],
+            AI_REPORT_SNAPSHOT_KEY: strip_null_bytes(result.markdown),
+            AI_REPORT_DIAGNOSTICS_KEY: strip_null_bytes([dataclasses.asdict(d) for d in result.diagnostics]),
             AI_REPORT_WINDOW_END_KEY: result.window_end_utc,
             # prompt is None for non-AI subs; "" if cleared — omit either.
-            **({AI_REPORT_PROMPT_SNAPSHOT_KEY: prompt} if prompt else {}),
+            **({AI_REPORT_PROMPT_SNAPSHOT_KEY: strip_null_bytes(prompt)} if prompt else {}),
         }
         delivery.save(update_fields=["content_snapshot", "last_updated_at"])
 
--- a/products/exports/backend/temporal/subscriptions/insight_snapshot.py
+++ b/products/exports/backend/temporal/subscriptions/insight_snapshot.py
@@ -28,32 +28,19 @@ from posthog.models import Team, User
 from products.dashboards.backend.models.dashboard import Dashboard
 from products.dashboards.backend.models.dashboard_tile import DashboardTile
 from products.exports.backend.models.subscription import Subscription
+from products.exports.backend.temporal.subscriptions.delivery_common import strip_null_bytes
 from products.product_analytics.backend.models.insight import Insight
 
 logger = structlog.get_logger(__name__)
+
+# strip_null_bytes now lives in delivery_common (a general-purpose sanitizer shared across delivery
+# paths); imported here for the query-result scrub below and so existing
+# `insight_snapshot.strip_null_bytes` importers keep resolving.
 
 
 def _json_safe_value(val: Any) -> Any:
     """Coerce to JSONField-safe Python (dict/list/scalars); unknown types use fallback, not pass-through."""
     return to_jsonable_python(val, fallback=lambda x: str(x))
-
-
-def strip_null_bytes(value: Any) -> Any:
-    """Recursively remove NUL (\\x00) from strings — Postgres text/jsonb columns cannot store it.
-
-    Anything written to ``SubscriptionDelivery.content_snapshot`` that originates outside a Postgres
-    text column (ClickHouse query results, LLM output, user-supplied prompts) must pass through this
-    first, or the NUL surfaces as a unicode escape that fails the whole delivery write with a DataError.
-    """
-    if isinstance(value, str):
-        return value.replace("\x00", "")
-    if isinstance(value, list):
-        return [strip_null_bytes(v) for v in value]
-    if isinstance(value, dict):
-        # Strip keys too: a Map(String, …) column can produce data-derived keys carrying NUL,
-        # and Postgres rejects it in a jsonb key just as it does in a value.
-        return {strip_null_bytes(k): strip_null_bytes(v) for k, v in value.items()}
-    return value
 
 
 def build_initial_content_snapshot(subscription: Subscription) -> dict[str, Any]:
--- a/products/exports/backend/temporal/subscriptions/insight_snapshot.py
+++ b/products/exports/backend/temporal/subscriptions/insight_snapshot.py
@@ -33,10 +33,6 @@ from products.product_analytics.backend.models.insight import Insight
 
 logger = structlog.get_logger(__name__)
 
-# strip_null_bytes now lives in delivery_common (a general-purpose sanitizer shared across delivery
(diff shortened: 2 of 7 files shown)


products/exports/backend/temporal/subscriptions/insight_snapshot.py
```diff
@@ -38,6 +38,24 @@ def _json_safe_value(val: Any) -> Any:
     return to_jsonable_python(val, fallback=lambda x: str(x))
 
 
+def strip_null_bytes(value: Any) -> Any:
```
vdekrijger: strip_null_bytes is a general-purpose recursive data sanitizer with zero dependency on insight types, yet lives in insight_snapshot.py — a module whose name describes domain-specific serialization, not shared utilities. Future consumers outside the insight path (e.g., email body scrubbing, Slack message sanitization) would import from insight_snapshot, creating a misleading conceptual dependency and weakening module cohesion. Move to delivery_common.py (already imported by both callers, same package) and re-export from insight_snapshot for backward compatibility.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
```diff
@@ -71,22 +89,28 @@ def _serialize_insight_result(result: InsightResult) -> dict[str, Any]:
...
+            "query_status": _json_safe_value(result.query_status),
+        },
+        default=str,
     )
+    parsed = orjson.loads(dumped)
+    # Query results can carry NUL bytes from upstream data. Django's JSONField serializes a NUL to
+    # a unicode escape that Postgres text/jsonb columns reject, which fails the whole delivery write.
+    # orjson emits that same escape sequence for a real NUL, so gate the (rare) recursive strip on
+    # its presence to avoid walking multi-MB payloads on every delivery.
+    if b"\\u0000" in dumped:
```
vdekrijger, line 103: Isn't it safe to always pass it through strip_null_bytes unconditionally?

products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
```diff
@@ -105,13 +106,15 @@ def _write() -> None:
         # No DoesNotExist guard: create_delivery_record always writes this row before
         # generation runs, so a missing row is a wiring bug — let it raise loudly.
         delivery = SubscriptionDelivery.objects.get(pk=delivery_id)
+        # LLM output and the user prompt can carry NUL bytes that Postgres text/jsonb reject;
+        # scrub them here (payloads are small) as they are the untrusted inputs on this write path.
         delivery.content_snapshot = {
             **(delivery.content_snapshot or {}),
-            AI_REPORT_SNAPSHOT_KEY: result.markdown,
-            AI_REPORT_DIAGNOSTICS_KEY: [dataclasses.asdict(d) for d in result.diagnostics],
+            AI_REPORT_SNAPSHOT_KEY: strip_null_bytes(result.markdown),
```
vdekrijger, line 113: In _persist_ai_report, the three new strip_null_bytes calls wrapping result.markdown, the diagnostics list, and prompt (lines ~111-113) have zero test coverage with NUL-bearing inputs — the test test_persist_ai_report_writes_markdown_query_diagnostics_and_prompt in test_activities.py uses plain strings like "# Weekly report" that pass cleanly, so a regression that silently drops the scrub would not be caught.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
```diff
@@ -38,6 +38,24 @@ def _json_safe_value(val: Any) -> Any:
     return to_jsonable_python(val, fallback=lambda x: str(x))
 
 
+def strip_null_bytes(value: Any) -> Any:
+    """Recursively remove NUL (\\x00) from strings — Postgres text/jsonb columns cannot store it.
+
+    Anything written to ``SubscriptionDelivery.content_snapshot`` that originates outside a Postgres
+    text column (ClickHouse query results, LLM output, user-supplied prompts) must pass through this
+    first, or the NUL surfaces as a unicode escape that fails the whole delivery write with a DataError.
+    """
```
vdekrijger: 🤖 _AI-suggested · github-user's review style_

`strip_null_bytes` only checks `isinstance(value, list)` — a `tuple` containing strings with NUL bytes falls through to `return value` at line 52, silently passing the NULs through. The AI diagnostics path at `ai_subscription/activities.py:115` converts the tuple to a list before calling, so the code is safe today, but the `Any` signature makes it easy for a future refactor to pass a tuple directly and lose the NUL scrubbing, which would cause a Postgres `DataError` on `delivery.save()`. Consider adding `isinstance(value, tuple): return tuple(strip_null_bytes(v) for v in value)` for parity, or documenting that tuples are unsupported.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
vdekrijger: Done in e276d146 — moved `strip_null_bytes` into `delivery_common.py` (it's a general-purpose sanitizer with no dependency on insight types) and re-exported it from `insight_snapshot.py` so existing `insight_snapshot.strip_null_bytes` importers keep resolving. The AI path now imports it from `delivery_common` too.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
vdekrijger: It's safe correctness-wise, yes. I kept the gate because `result.result` can be a multi-MB insight payload and `strip_null_bytes` recursively rebuilds the whole dict/list tree — doing that on every delivery when almost none carry a NUL is wasteful. The gate is a single byte-scan over bytes we already have (`dumped`), and orjson always escapes a real NUL, so it never false-negatives. The AI path scrubs unconditionally since those payloads are small. Happy to drop the gate if you'd rather have the uniform simplicity.

products/exports/backend/temporal/subscriptions/ai_subscription/activities.py
vdekrijger: Added `test_persist_ai_report_strips_null_bytes` in `test_activities.py` (e276d146): it passes NUL-bearing markdown, diagnostics (description + hogql), and prompt, then asserts the persisted snapshot is scrubbed. Without the scrub the Postgres write raises `DataError`, so the test fails loudly if the scrub is ever dropped.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
vdekrijger: Good catch — added a `tuple` branch to `strip_null_bytes` (e276d146) for parity with lists, so a future caller passing a tuple gets scrubbed instead of silently passing the NUL through.

products/exports/backend/temporal/subscriptions/insight_snapshot.py
vdekrijger: Sounds good — keeping the gate. Detailing the rationale here for the record:

**What the gate is:** in `_serialize_insight_result` the payload is serialized once (`dumped = orjson.dumps(...)`), and we only run the recursive scrub when that serialized buffer actually contains a NUL:

```python
parsed = orjson.loads(dumped)
if b"\\u0000" in dumped:
    parsed = strip_null_bytes(parsed)
return parsed
```

**Why it's here (cost):** `result.result` is arbitrary ClickHouse output and can reach multiple MB (it's what feeds `content_snapshot`). `strip_null_bytes` recursively rebuilds the entire dict/list tree, allocating a full copy. NUL bytes in query results are very rare, so on the overwhelming majority of deliveries that walk would be pure waste. The gate replaces it with a single C-level `bytes.__contains__` scan over bytes we already have in hand — no extra serialization, no allocation.

**Why it's correct (no false negatives):** orjson always emits a real NUL as the six-byte escape `\u0000` in its output (JSON forbids a raw control byte), including inside object keys. So `b"\\u0000" in dumped` is true whenever a genuine NUL is present anywhere in the structure. It can only *over*-trigger (e.g. if the literal text `\u0000` appears in data), which is harmless — the scrub then just walks a clean tree and removes nothing.

**Why only here:** this gate guards the one large payload. The AI-report path (`_persist_ai_report`) scrubs `markdown` / diagnostics / prompt unconditionally, because those are small and a gate would add complexity for no measurable win.

MattPua (approved):

PR #70967, opened 2026-07-15
feat(autoresearch): cut large dashboard GET wall time ~65% by eliminating cached-result serialization round trips

thmsobrmlr:
## TL;DR (eli5)

Opening a big dashboard was slow even when every chart was already cached. The server was unpacking every cached chart result into Python objects, copying them, and re-packing them into JSON — three times the work for zero change in output. Now it reuses one serializer for all tiles, stops copying the result tree, and passes the cached results straight from Redis into the response as raw JSON. Same bytes out, about 3x faster.

## Problem

Large dashboard GETs (70-80 insight tiles) spend 80%+ of wall time outside Postgres, even with every tile served from the query cache. Profiling a local repro (70 trends tiles, ~450 KiB cached results each, ~29 MB response) showed the non-SQL time concentrated in:

| Phase | ~ms/request |
|---|---|
| orjson.loads of 70 cached result blobs | 389 |
| pydantic `model_dump(by_alias=True)` of the full response | 254 |
| DRF serializer construction per tile (`get_fields` deepcopy) | 300 |
| Response render (orjson.dumps) | ~100 |
| Redis GET + zstd | 76 |
| SQL (26 queries) | 38 |

## Changes

Three independent, behavior-preserving optimizations (one commit each):

1. **Skip the full-response `model_dump`** in `calculate_for_query_based_insight` — build `InsightResult` from response model attributes; only the small nested models (`timings`, `query_status`, `resolved_date_range`) are dumped, matching previous output.
2. **Reuse one `DashboardTileSerializer` across tiles** in `get_tiles` — DRF field construction (deep-copying ~60 declared fields incl. the nested insight serializer, per tile) dominated serializer CPU. Per-tile state flows through the shared context; `(dashboard, insight)` uniqueness keeps the per-instance `insight_result` lru_cache correct.
3. **Serve cached results as raw JSON** — behind `QUERY_CACHE_SPLIT_FORMAT_WRITES` (default **off**): every deploy ships split-capable readers, and split-format writes are enabled only once the reader is fully rolled out (old readers treat split entries as cache misses and would recompute + churn the cache during a mixed-version window). When enabled, cache blobs are written in a split format (header JSON + results JSON segments; legacy blobs stay readable, old readers treat new blobs as cache misses and recompute). On cache hits, the insight serializer opts in to validate only the header and embed the results segment into the response via `orjson.Fragment`, skipping the parse → validate → re-serialize round trip entirely. Gated to orjson renderers (`SafeJSONRenderer`/SSE tile stream), disabled for `?indent=`, the shared-page template (stdlib `json.dumps`), `run_insights` LLM formatting, and whenever the query or cached results carry series `custom_name`s (those need parsed patching).

Results on the local repro (median of 25 GETs, single core; SQL time unchanged at ~38ms / 26 queries):

| Change | median wall | p90 |
|---|---|---|
| baseline | 999 ms | 2130 ms |
| 1. no full model_dump | 873 ms | 2035 ms |
| 2. serializer reuse | 787 ms | 2229 ms |
| 3. raw results pass-through | **347 ms** | **440 ms** |

A fourth commit fixes a bug an adversarial review pass caught in change 1: response classes whose `results` items are pydantic models (Retention, Paths, marketing analytics `list[list[...]]`) need those items dumped to dicts — DRF's JSON encoder otherwise iterates the models into `(field, value)` tuple arrays, corrupting the response shape. Covered by new regression tests in `posthog/caching/test/test_calculate_results.py` (model items, nested lists of models, plain-dict identity pass-through, raw-fragment path) — a gap no existing test covered.

The p90 collapse comes from eliminating ~700 MB of per-request Python object allocation (GC pauses). The same per-tile pipeline serves `stream_tiles` (the app's SSE path), so it benefits equally; the plain `retrieve` path remains the one used by API consumers, mobile, and embeds.

## Where the remaining time goes (why this is close to the floor)

After the three fixes the local repro sits at ~345 ms (from 999 ms), of which ~307 ms is non-SQL. Decomposition of the remainder: Redis fetch + header parse ~80 ms, per-tile query-runner machinery (cache key, async-status lookup, observability contexts, per-tile analytics event) ~60 ms, DRF field iteration across ~45 insight fields per tile ~50 ms, URL routing + middleware + auth ~50 ms, SQL 38 ms, response render ~20 ms. No remaining single-digit-risk change is worth >30 ms; further reduction needs behavior or API changes, e.g.:

- batching the per-tile cache/status Redis GETs needs the cache keys before serialization, but keys are derived inside each tile's query runner (the tiles' stored `filters_hash` is only maintained for legacy filters-based insights) — batching would require restructuring `insight_result` to build runners in a pre-pass;
- skipping the per-tile async-query-status lookup on fresh cache hits would drop 70 Redis round trips but changes `query_status` semantics;
- a "results-only" dashboard payload for consumers that ignore the ~40 insight metadata fields would cut the DRF loop, but that's an API change (`stream_tiles` already exists for the app frontend and benefits from these fixes equally — verified the SSE events remain valid JSON with the fragment path active).

Also evaluated and rejected: bypassing django-redis's pickle framing of cached bytes (isolated A/B: 0.8 ms per 70 blobs — pickle-framing bytes is nearly free; a sampling profiler had misattributed the cost).

## Post-review hardening (commits 5)

Addressed `ty` type errors and review feedback: `Literal` overloads keep `process_query_dict`/`process_query_model` returning `dict | BaseModel` for all existing callers; raw pass-through is additionally gated off for response classes whose `results` are model-typed (so the schema-drift validation path is never bypassed) and requires an explicit `raw_results_supported` capability for the SSE stream instead of trusting content negotiation; model conversion is container-preserving (dict-shaped/scalar results keep their shape); nested funnel steps' `action.custom_name` is detected at write time; split-format writes are flag-gated for rollout safety (with a write-format regression test).

- Byte-identity: sha256 of the full 29.9 MB dashboard response is identical before/after each change (and raw vs parsed path), with only `last_accessed_at` normalized.
- `posthog/api/test/dashboards/test_dashboard.py` — 176 passed, 1 failed (`test_shared_dashboard`) — that failure reproduces identically on master in this environment (no object storage) and is unrelated.
- Canonical N+1 tests pass, including `test_insight_alerts_do_not_nplus1_dashboard_gets` and `test_adding_insights_is_not_nplus1_for_gets` (query counts unchanged).
- `posthog/hogql_queries/test/test_query_runner.py` (89 passed), `posthog/caching/test/test_cache_size_tracker.py`, `posthog/caching/test/test_warming.py`, and `products/product_analytics/backend/api/test/test_insight.py -k "cach or refresh or shared"` (12 passed).
- Two mock-assertion tests updated for the new `allow_raw_results` kwarg (exact-kwargs `assert_called_once_with`).
- New `posthog/caching/test/test_calculate_results.py` (4 tests): catches the model-typed-results regression described above; asserts through the public `calculate_for_query_based_insight` interface with only `process_query_dict` mocked.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Produced in an autoresearch (iterative optimization) session with Claude Code: built a benchmark harness (70 cached trends tiles seeded into the query cache), profiled with pyinstrument + targeted phase timers, and applied fixes one at a time with before/after medians and byte-identity checks between each. Considered and rejected: batching per-tile Redis GETs via MGET (needs cache keys upfront, which only exist after per-tile runner construction; revisit separately), and skipping `apply_dashboard_filters_to_dict` for empty filters (measured at 0.1 ms — not worth the risk). The split cache format records at write time whether results contain custom series names, so the raw path can prove that custom-name patching is a no-op before skipping the parse.

Created with Autoresearch.

---

### DIFF
--- a/posthog/api/services/query.py
+++ b/posthog/api/services/query.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Literal, Optional, overload
 
 import structlog
 import pydantic_core
@@ -55,6 +55,52 @@ class RawCachedQueryResponse:
 
     response: BaseModel
     raw_results: bytes
+
+
+# The overloads keep the public contract at `dict | BaseModel` for the vast majority of
+# callers: only allow_raw_results=True can produce a RawCachedQueryResponse.
+@overload
+def process_query_dict(
+    team: Team,
+    query_json: dict,
+    *,
+    dashboard_filters_json: Optional[dict] = ...,
+    variables_override_json: Optional[dict] = ...,
+    limit_context: Optional[LimitContext] = ...,
+    execution_mode: ExecutionMode = ...,
+    user: Optional[User] = ...,
+    user_access_control: Optional[UserAccessControl] = ...,
+    query_id: Optional[str] = ...,
+    insight_id: Optional[int] = ...,
+    dashboard_id: Optional[int] = ...,
+    is_query_service: bool = ...,
+    cache_age_seconds: Optional[int] = ...,
+    pagination_cursor: Optional[str] = ...,
+    analytics_props: Optional[AnalyticsProps] = ...,
+    allow_raw_results: Literal[False] = ...,
+) -> dict | BaseModel: ...
+
+
+@overload
+def process_query_dict(
+    team: Team,
+    query_json: dict,
+    *,
+    dashboard_filters_json: Optional[dict] = ...,
+    variables_override_json: Optional[dict] = ...,
+    limit_context: Optional[LimitContext] = ...,
+    execution_mode: ExecutionMode = ...,
+    user: Optional[User] = ...,
+    user_access_control: Optional[UserAccessControl] = ...,
+    query_id: Optional[str] = ...,
+    insight_id: Optional[int] = ...,
+    dashboard_id: Optional[int] = ...,
+    is_query_service: bool = ...,
+    cache_age_seconds: Optional[int] = ...,
+    pagination_cursor: Optional[str] = ...,
+    analytics_props: Optional[AnalyticsProps] = ...,
+    allow_raw_results: bool,
+) -> dict | BaseModel | RawCachedQueryResponse: ...
 
 
 def process_query_dict(
@@ -129,6 +175,50 @@ def process_query_dict(
         analytics_props=analytics_props,
         allow_raw_results=allow_raw_results,
     )
+
+
+@overload
+def process_query_model(
+    team: Team,
+    query: BaseModel,
+    *,
+    dashboard_filters: Optional[DashboardFilter] = ...,
+    variables_override: Optional[list[HogQLVariable]] = ...,
+    limit_context: Optional[LimitContext] = ...,
+    execution_mode: ExecutionMode = ...,
+    user: Optional[User] = ...,
+    user_access_control: Optional[UserAccessControl] = ...,
+    query_id: Optional[str] = ...,
+    insight_id: Optional[int] = ...,
+    dashboard_id: Optional[int] = ...,
+    is_query_service: bool = ...,
+    cache_age_seconds: Optional[int] = ...,
+    pagination_cursor: Optional[str] = ...,
+    analytics_props: Optional[AnalyticsProps] = ...,
+    allow_raw_results: Literal[False] = ...,
+) -> dict | BaseModel: ...
+
+
+@overload
+def process_query_model(
+    team: Team,
+    query: BaseModel,
+    *,
+    dashboard_filters: Optional[DashboardFilter] = ...,
+    variables_override: Optional[list[HogQLVariable]] = ...,
+    limit_context: Optional[LimitContext] = ...,
+    execution_mode: ExecutionMode = ...,
+    user: Optional[User] = ...,
+    user_access_control: Optional[UserAccessControl] = ...,
+    query_id: Optional[str] = ...,
+    insight_id: Optional[int] = ...,
+    dashboard_id: Optional[int] = ...,
+    is_query_service: bool = ...,
+    cache_age_seconds: Optional[int] = ...,
+    pagination_cursor: Optional[str] = ...,
+    analytics_props: Optional[AnalyticsProps] = ...,
+    allow_raw_results: bool,
+) -> dict | BaseModel | RawCachedQueryResponse: ...
 
 
 def process_query_model(
--- a/posthog/api/services/query.py
+++ b/posthog/api/services/query.py
@@ -252,7 +252,7 @@ def process_query_model(
             query_runner.apply_pagination_cursor(pagination_cursor)
         query_runner.is_query_service = is_query_service
         if allow_raw_results:
-            query_runner._serve_raw_cached_results = True
+            query_runner.serve_raw_cached_results = True
 
         result = query_runner.run(
             execution_mode=execution_mode,
@@ -263,7 +263,7 @@ def process_query_model(
             cache_age_seconds=cache_age_seconds,
             analytics_props=analytics_props,
         )
-        raw_results = query_runner._raw_cached_results_bytes
+        raw_results = query_runner.raw_cached_results_bytes
         if raw_results is not None and isinstance(result, BaseModel):
             return RawCachedQueryResponse(response=result, raw_results=raw_results)
 
--- a/posthog/api/services/query.py
+++ b/posthog/api/services/query.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Optional
 
 import structlog
@@ -42,6 +43,20 @@ from common.hogvm.python.debugger import color_bytecode
 logger = structlog.get_logger(__name__)
 
 
+@dataclass(frozen=True)
+class RawCachedQueryResponse:
+    """A cached query response whose `results` field is carried as raw JSON bytes.
+
+    `response.results` holds an empty-list placeholder; `raw_results` is the JSON-encoded
+    results segment straight from the cache, ready to be embedded into a JSON response
+    (e.g. via orjson.Fragment) without a parse/re-serialize round trip. Only produced when
+    a caller passes allow_raw_results=True.
+    """
+
+    response: BaseModel
+    raw_results: bytes
+
+
 def process_query_dict(
     team: Team,
     query_json: dict,
@@ -59,7 +74,8 @@ def process_query_dict(
     cache_age_seconds: Optional[int] = None,
     pagination_cursor: Optional[str] = None,
     analytics_props: Optional[AnalyticsProps] = None,
-) -> dict | BaseModel:
+    allow_raw_results: bool = False,
+) -> dict | BaseModel | RawCachedQueryResponse:
     upgraded_query_json = upgrade(query_json)
     try:
         model = QuerySchemaRoot.model_validate(upgraded_query_json)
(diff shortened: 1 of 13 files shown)


posthog/caching/calculate_results.py
```diff
@@ -97,27 +124,61 @@ def calculate_for_query_based_insight(
...
+        return InsightResult(
+            # orjson.Fragment embeds the cached results bytes into the JSON response as-is,
+            # skipping the parse/re-serialize round trip. Callers passing allow_raw_results
+            # guarantee the result feeds an orjson renderer. The parsed path must convert
+            # model-typed result items (e.g. RetentionResult, PathsLink) to dicts like
+            # model_dump did — DRF's JSON encoder would otherwise mangle them into
+            # (field, value) tuple arrays.
+            result=orjson.Fragment(raw_results)
+            if raw_results is not None
+            else _model_list_field_as_dicts(process_response, "results"),
```
pauldambra: 🤖 PostHog Code acting on behalf of @pauldambra: this helper assumes every model response has list-shaped `results`. Supported response models include dictionary-shaped results (and `HogQueryResponse.results` can be arbitrary), so dictionaries become lists of keys, strings become character lists, and scalars can raise. Please use a recursive container-preserving conversion that only dumps nested `BaseModel` values, and add dict/scalar/deeply-nested regression cases.

posthog/hogql_queries/query_cache.py
```diff
@@ -176,7 +179,14 @@ class DjangoCacheQueryCacheManager(QueryCacheManagerBase):
     """
 
     def set_cache_data(self, *, response: dict, target_age: Optional[datetime]) -> None:
-        fresh_response_serialized = OrjsonJsonSerializer({}).dumps(response)
+        from posthog.caching.fetch_from_cache import encode_split_cached_response
+
+        if isinstance(response.get("results"), list):
```
pauldambra: 🤖 PostHog Code acting on behalf of @pauldambra: writing `PHQC2` under the existing cache key is not safe during a rolling deploy or rollback. Old readers treat new entries as misses, recompute expensive queries, and overwrite them with legacy JSON; mixed versions can alternate formats and create a cache-load spike. Please use versioned keys/dual writes, or land a backward-compatible reader first and gate split-format writes until it is fully deployed.

products/product_analytics/backend/api/insight.py
```diff
@@ -1122,6 +1123,16 @@ def insight_result(self, insight: Insight) -> InsightResult:
                 request_user_access_control = (
                     getattr(view, "user_access_control", None) if isinstance(request_user, User) else None
                 )
+                # Raw cached results (orjson.Fragment) are only valid when the response is
+                # rendered by orjson: SafeJSONRenderer directly, or the SSE tile stream which
+                # renders each tile with SafeJSONRenderer. Pretty-printed output (?indent=) would
+                # not indent inside a fragment, so keep the parsed path there.
+                accepted_renderer = getattr(self.context["request"], "accepted_renderer", None)
+                allow_raw_results = (
+                    isinstance(accepted_renderer, SafeJSONRenderer | ServerSentEventRenderer)
```
pauldambra: 🤖 PostHog Code acting on behalf of @pauldambra: `ServerSentEventRenderer` is configured for the whole dashboard viewset, but only `stream_tiles` explicitly re-renders each tile through `SafeJSONRenderer`. A normal action negotiated as `text/event-stream` can therefore enable `orjson.Fragment` and then return the Python structure unchanged. Please gate this on the `stream_tiles` action (or pass an explicit context capability) rather than treating every SSE-negotiated request as an orjson boundary.

posthog/api/services/query.py
```diff
@@ -59,7 +74,8 @@ def process_query_dict(
     cache_age_seconds: Optional[int] = None,
     pagination_cursor: Optional[str] = None,
     analytics_props: Optional[AnalyticsProps] = None,
-) -> dict | BaseModel:
+    allow_raw_results: bool = False,
+) -> dict | BaseModel | RawCachedQueryResponse:
```
pauldambra, line 124: 🤖 PostHog Code acting on behalf of @pauldambra: this public return-type widening is breaking Python quality checks in existing callers such as `posthog/api/query.py`, `products/endpoints/backend/logic/execution.py`, and the AI query executor. Since `allow_raw_results=False` cannot narrow the return statically, please use overloads/a narrower internal entry point (or unwrap before crossing this API) so ordinary callers retain the existing `dict | BaseModel` contract.

pauldambra (dismissed):
🤖 **Automated pr-swarm review from PostHog Code, acting on behalf of @pauldambra.**

The optimization direction is promising, but the swarm converged on blocking correctness and rollout issues. In addition to the inline comments below, I agree with the existing P1 comments about bypassing cached-result schema validation and missing nested `action.custom_name` values. CI is also currently failing because the widened `process_query_dict` return type leaks into callers that mutate the returned mapping.

Please preserve arbitrary result container shapes, make split-cache writes safe during mixed-version deploys/rollbacks, restrict raw fragments to a guaranteed orjson rendering path, and restore a caller-safe type contract before merging.

thmsobrmlr:
🤖 Comment from PostHog Code (on behalf of the PR author).

All items from the review are addressed on the branch — mapping for re-review:

- **Caller-safe type contract**: `Literal` overloads on `process_query_dict`/`process_query_model` keep the `dict | BaseModel` return for every existing caller; only `allow_raw_results=True` widens it (3e2a1cf8). The remaining mypy error (`InsightResult.resolved_date_range`) is fixed by annotating the field as the dict it has always carried (943a10ff). Python code quality is green.
- **Schema-validation bypass (Greptile P1)**: raw pass-through is additionally gated on the response class typing `results` as plain data (`response_results_contain_models`); model-typed results (retention, paths, …) always take the parsed+validated path (3e2a1cf8).
- **Arbitrary result container shapes**: conversion is container-preserving (`_dump_nested_models`) and only runs for model-typed annotations — dict-shaped and scalar results keep their shape; dict/scalar regression cases added in `posthog/caching/test/test_calculate_results.py` (3e2a1cf8).
- **Mixed-version deploy safety**: split-format **writes** are gated behind `QUERY_CACHE_SPLIT_FORMAT_WRITES` (default off), so split-capable readers roll out everywhere before any pod writes the new format; write-format regression test added (3e2a1cf8).
- **SSE renderer gate**: content negotiation alone no longer enables fragments — `stream_tiles` passes an explicit `raw_results_supported` context capability since it renders each tile with `SafeJSONRenderer` itself (3e2a1cf8).
- **Nested `action.custom_name` (Greptile P1)**: write-time detection now checks `action.custom_name` inside nested result lists too (3e2a1cf8).

Byte-identity (raw vs parsed sha256-equal on a 29.9 MB response) and the benchmark (~999→341 ms median, split writes enabled) re-verified after the changes.

pauldambra (commented):
🤖 **Automated rereview from PostHog Code, acting on behalf of @pauldambra.**

The follow-up commits address the original blocking findings: arbitrary result shapes are preserved, nested custom names are detected, split writes default off for safe forward rollout, raw SSE fragments require the explicit tile-stream path, and the public query return type is narrowed with overloads. I am downgrading the previous request-changes review to a non-blocking comment.

Two concerns remain worth addressing before enabling `QUERY_CACHE_SPLIT_FORMAT_WRITES` in production:

- The trailing raw results segment has no length/integrity validation. Empty, truncated, or invalid JSON can reach `orjson.Fragment`, while paths that parse the segment can raise instead of treating the entry as a cache miss. Consider framing it with length/checksum metadata and converting decode failures into cache misses.
- The write gate makes the initial rolling deploy safe, but split entries remain unreadable by an older binary after a rollback. A versioned key, dual-write period, or explicit disable-and-drain rollback procedure would close that gap.

These are non-blocking for merge while split writes remain disabled by default, but they should be resolved or operationally guarded before flipping the setting.

PR #70970, opened 2026-07-15
feat(error-tracking): read and write autocapture opt-in from settings (phase 2)

ablaszkiewicz:
move reads to the new table. Do dual writes

Ran backfills both in EU and US today and there's no drift


hpouillot (approved):

cat-ph (approved):
🚀

products/error_tracking/backend/models.py
```diff
@@ -752,6 +757,21 @@ def sync_autocapture_opt_in(team_id: int, opt_in: bool | None) -> None:
...
+
+    ErrorTrackingSettings is the sole read source: opted-in teams were backfilled, so the
+    Team column is no longer consulted (it's still dual-written for rust, the API
+    serializers, and the frontend toggle until a later phase). A missing row or a null
+    value reads as disabled.
+    """
+    return bool(
+        ErrorTrackingSettings.objects.filter(team_id=team.id)
+        .values_list("autocapture_exceptions_opt_in", flat=True)
+        .first()
```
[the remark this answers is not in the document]
ablaszkiewicz: There are no missing rows. I ran backfills both in EU and US and we have double writes done in phase 1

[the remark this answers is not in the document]
ablaszkiewicz: Huh I didn't know such command existed. It's dev only but I agree this should be updated. Will update the command 🚀

products/error_tracking/backend/presentation/views/settings.py
```diff
@@ -9,6 +9,10 @@
 
 
 class ErrorTrackingSettingsSerializer(serializers.Serializer):
+    autocapture_exceptions_opt_in = serializers.BooleanField(
```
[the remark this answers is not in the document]
ablaszkiewicz: This is expected. We want this to be controlled by error tracking scope. Also right now if RBAC off, anyone can change that field. With our change in RBAC off still anyone can change it. That is the same behavior

ablaszkiewicz:
@PostHog/team-error-tracking I increased scope of this PR to also include writes. This is safe as we have double writes on both sides (if you update it via team, it propagates to ET settings and vice versa but without infinite loop lol) so this will be safe as pods start rolling out

products/error_tracking/backend/logic/__init__.py
```diff
@@ -346,13 +349,22 @@ def get_or_create_settings(team_id: int) -> ErrorTrackingSettings:
...
-    if updates:
-        settings.save(update_fields=list(updates))
+    with transaction.atomic():
+        for key, value in updates.items():
+            setattr(settings, key, value)
+        if updates:
+            settings.save(update_fields=list(updates))
+        # Mirror onto Team while it's still dual-written; updated_at set since update() skips auto_now.
+        if "autocapture_exceptions_opt_in" in updates:
+            Team.objects.filter(id=team_id).update(
```
cat-ph: I think we somehow need to care about the remote config `Team.post_save` 🤔 

also not sure if we need to have a post receiver somewhere here too? https://github.com/PostHog/posthog/blob/master/posthog/models/remote_config.py#L515-L517

products/error_tracking/frontend/scenes/ErrorTrackingConfigurationScene/exception_autocapture/ExceptionAutocaptureSettings.tsx
```diff
@@ -10,11 +10,13 @@ import { ProductIntentContext, ProductKey } from '~/queries/schema/schema-genera
...
 import { disableSurveyLogic } from './disableSurveyLogic'
+import { exceptionAutocaptureLogic } from './exceptionAutocaptureLogic'
 
 export function ExceptionAutocaptureToggle(): JSX.Element {
     const { userLoading } = useValues(userLogic)
-    const { currentTeam } = useValues(teamLogic)
-    const { updateCurrentTeam, addProductIntent } = useActions(teamLogic)
+    const { addProductIntent } = useActions(teamLogic)
+    const { autocaptureOptIn, settingsLoading } = useValues(exceptionAutocaptureLogic)
+    const { setAutocaptureOptIn } = useActions(exceptionAutocaptureLogic)
```
cat-ph, line 19: should we also move the read in places like i.e. https://github.com/PostHog/posthog/blob/d1518c6c0bb59b68058768640c274036ecee7715/products/error_tracking/frontend/components/SetupPrompt/SetupPrompt.tsx#L60-L68 ? 🤔

products/error_tracking/backend/presentation/views/settings.py
```diff
@@ -9,6 +9,10 @@
...
+        required=False,
+        help_text="Whether the SDK autocaptures unhandled exceptions for this project. Enabling starts exception ingestion.",
+    )
```
cat-ph, line 16: also wondering, I don't think we have log_activity for the updates to settings, should we add something? (might be separate/follow-up maybe)

products/error_tracking/backend/logic/__init__.py
ablaszkiewicz: Dang that's right. I'll add a post receiver for et settings

products/error_tracking/frontend/scenes/ErrorTrackingConfigurationScene/exception_autocapture/ExceptionAutocaptureSettings.tsx
ablaszkiewicz: 😿 adding

products/error_tracking/backend/presentation/views/settings.py
ablaszkiewicz: Good idea. Will add in this PR

ablaszkiewicz: Added backend + frontend describer

products/error_tracking/frontend/scenes/ErrorTrackingConfigurationScene/exception_autocapture/ExceptionAutocaptureSettings.tsx
ablaszkiewicz: Fixed

products/error_tracking/backend/logic/__init__.py
ablaszkiewicz: Added

products/error_tracking/backend/presentation/views/settings.py
```diff
@@ -54,5 +59,10 @@ def retrieve_settings(self, request, *args, **kwargs):
     def update_settings(self, request, *args, **kwargs):
         serializer = ErrorTrackingSettingsSerializer(data=request.data, partial=True)
         serializer.is_valid(raise_exception=True)
-        settings = error_tracking_api.update_settings(self.team.id, dict(serializer.validated_data))
+        settings = error_tracking_api.update_settings(
```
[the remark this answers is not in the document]
ablaszkiewicz: Not true.

By default that field is writable by anyone. If RBAC enabled, then it requires admin.

After our changes if RBAC enabled, it requires error tracking write scope which is correct

products/error_tracking/backend/logic/__init__.py
```diff
@@ -381,13 +385,39 @@ def get_or_create_settings(team_id: int) -> ErrorTrackingSettings:
...
+    with transaction.atomic():
+        for key, value in updates.items():
+            setattr(settings, key, value)
+        if updates:
+            settings.save(update_fields=list(updates))
+        # Mirror onto Team via save() so its post_save consumers (remote config, team metadata cache) fire.
+        if "autocapture_exceptions_opt_in" in updates:
+            team = Team.objects.get(id=team_id)
+            team.autocapture_exceptions_opt_in = bool(updates["autocapture_exceptions_opt_in"])
+            team.save(update_fields=["autocapture_exceptions_opt_in", "updated_at"])
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed in 9a12d36. You're right that saving `Team` triggered the mirror back into `ErrorTrackingSettings`, doubling both the write and the remote config rebuild (via `mirror_autocapture_opt_in_to_settings` → `sync_autocapture_opt_in`).

I mirrored onto the deprecated `Team` column with a queryset `update()` instead of `Team.save()`, so Team `post_save` never fires. The settings row is the source of truth and already dispatches the rebuild through `rebuild_remote_config_on_settings_save`, which matches the "the rebuild must hang off this model" comment in `signals.py`. Net result: one settings write, one rebuild. Strengthened `test_update_settings_dispatches_remote_config_rebuild` to assert exactly one dispatch so this can't silently regress.

ablaszkiewicz:
Heads-up on CI: the `Jest test (EE - 1)` shard failed once on `products/product_analytics/frontend/insights/stickiness/StickinessBarChart/StickinessBarChart.test.tsx` with `Exceeded timeout of 5000 ms for a test` (plus a "worker failed to exit gracefully" warning). It's an unrelated rendering test that only got pulled into this PR's run because the diff touches `frontend/src/types.ts`, and this branch's latest commit is Python-only. Re-ran the shard and it passed, so it's a flaky timeout rather than a regression from this change.

PR #70974, opened 2026-07-15
feat(slo): add billing objectives and burn alerts

mcoll-posthog:
## Problem

[PostHog/billing#2052](https://github.com/PostHog/billing/pull/2052) emits SLO lifecycle events for billing status requests and usage reports v2 intake. The shared SLO dashboard does not yet define objectives or alert delivery for those operations.

Malformed usage report messages cannot always be attributed to `US` or `EU`. Without a catch-all series, those failures disappear from the regional burn-rate alerts.

## Changes

- Add 99.95% successful-completion objectives over 28 days for `billing_status` and `usage_report_ingest`.
- Add an `UNKNOWN` usage-report region so malformed messages remain in the intake SLO denominator.
- Add hourly fast-burn alerts at 14.4x the sustainable budget rate.
- Deliver firing alerts to the Billing team Slack channel.
- Support operations with different region counts in the generated dashboard layout.
- Use an alert-only Trends formula because PostHog alerts currently accept `TrendsQuery`, while the dashboard keeps its correlation-aware HogQL query.

- Ran `terraform fmt -check` on the SLO monitoring configuration.
- Ran `./bin/hogli ci:preflight --strict`.
- Ran `git diff --check`.
- I did not run Terragrunt plan or apply.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No product documentation changes. The Billing repository owns the internal SLO objective notes.

## 🤖 Agent context

Codex desktop prepared this PR in the current task. No public session link is available. Skills used: `/marce-development`, `/slo-instrumentation`, `/sending-notifications`, `/write-like-a-human`, and `/github:yeet`. The Slack connector was used only to resolve the stable Billing channel ID.

The main design choice was to keep the dashboard's correlation-aware HogQL calculation and create a small Trends formula only for alert evaluation, since alerts currently reject HogQL-backed insights.

### DIFF
--- a/terraform/us/project-2/team-analytics-platform/slo-monitoring/alerts.tf
+++ b/terraform/us/project-2/team-analytics-platform/slo-monitoring/alerts.tf
@@ -1,0 +1,180 @@
+locals {
+  billing_slo_operation_keys = ["billing_status", "usage_report_ingest"]
+  billing_slo_operation_regions = {
+    for key, operation in local.slo_operation_regions :
+    key => operation if contains(local.billing_slo_operation_keys, operation.operation)
+  }
+
+  # A 14.4x burn for one hour consumes roughly 2% of a 28-day error budget.
+  billing_slo_fast_burn_threshold = 14.4
+}
+
+# PostHog alerts currently require a TrendsQuery, while the dashboard burn-rate
+# insights use HogQL to correlate started/completed events. This alert-only insight
+# uses the same started-minus-successful-completed availability calculation over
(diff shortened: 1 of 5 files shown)


pawel-cebula (approved):
Approving since the overall shape makes sense to me.

I would still take a look at the existing Greptile and PostHog review comments before merging.

One question: should these Slack notifications go to `#team-billing` or `#alerts-billing`? I would lean toward `#alerts-billing`, since these are automated alert notifications and that channel already seems meant for this kind of traffic.

PR #70979, opened 2026-07-15
feat(signals): add team master switch to disable autonomous inbox PRs

VojtechBartos:
## Problem

Teams using the signals inbox get autonomous PRs opened for actionable reports by default. Some teams want reports and notifications, but prefer to review and open PRs manually. There was no way to turn autostart off.

Zendesk: https://posthoghelp.zendesk.com/agent/tickets/61892 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/64118)

## Changes

- Add `SignalTeamConfig.autostart_enabled` (default `true`, with a Postgres `db_default`).
- Guard `maybe_autostart_implementation_task` right after the team config loads, so it gates every autostart path including the reviewer-less fallback used by teams without a linked GitHub account. When off, reports still generate and notify, but nothing auto-starts a task or opens a PR.
- Expose the field on `SignalTeamConfigSerializer` and surface it in the Django admin list view so it can be flipped per team.

No UI yet.

Added one parametrized async test in `test_auto_start.py`: an actionable, prioritized report with no resolvable reviewer auto-starts under the team's signals enabler when the switch is on, and does not when it is off. Catches a regression where the guard is removed or the fallback path bypasses it. Full file passes (21 tests). Also ran `hogli ci:preflight` (lint, format, migrations clean) and `hogli build:openapi` (no drift).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I directed this; Claude implemented it. Invoked the `/django-migrations`, `/improving-drf-endpoints`, and `/writing-tests` skills along the way. The guard is placed at the single choke point (after the team config load) rather than in each path, so it covers reviewer, user-triggered, and reviewer-less fallback autostart uniformly. Used `db_default=True` on the boolean so existing rows and any non-Django writer stay `true`. The `SignalTeamConfig` endpoint is consumed via a handwritten `api.signalTeamConfig` client method rather than the generated pipeline, so there was no frontend type drift to regenerate.


products/signals/backend/models.py
```diff
@@ -128,6 +128,9 @@ class SignalTeamConfig(UUIDModel):
         on_delete=models.CASCADE,
         related_name="signal_team_config",
     )
+    # Master switch for autonomous inbox PRs. When False, actionable reports still generate and
+    # notify, but never auto-start an implementation task — the team reviews and opens PRs manually.
+    autostart_enabled = models.BooleanField(default=True, db_default=True)
```
skoob13: Should be nullable

VojtechBartos: right 👍 ->  updated

skoob13 (approved):

PR #70991, opened 2026-07-15
feat(experiments): render funnel metric results with quill funnel chart

sampennington:
## Problem

The experiments funnel visualization was a DOM-based fork of the insights funnel renderer. To feed it, experiment results had to be converted into synthetic insights funnel data (`FunnelStepWithNestedBreakdown` with per-variant `nested_breakdown`) and run through `scenes/funnels` conversion-metric utilities, dragging ~880 lines of forked rendering code and an impedance-mismatched data layer along with it.

We now have a canvas funnel chart in `@posthog/quill-charts`, and experiments is its first adopter in the app.

## Changes

https://github.com/user-attachments/assets/82ae17cd-4714-43b9-b347-13d9a17f787a

- New `ExperimentFunnelChart` renders funnel metric results via the quill `FunnelChart`: one band per step ("Experiment exposure" first), one bar per variant, valued as percent of that variant's exposures. Variant colors come from the feature flag via `getVariantColor`, matching `VariantTag` elsewhere.
- Custom tooltip keyed to the hovered variant bar: entered/converted count, dropped off, conversion so far, conversion from previous, plus a "Click to view users" hint.
- Click-to-inspect still opens the persons modal through `ExperimentActorsQuery`, preserving the exposure-step offset rules (step 0 not queryable, drop-offs at the first metric step not queryable, drop-off clicks map to negative funnel steps).
- Per-step legend rendered through the chart's `stepFooter`, pixel-aligned under each band: step lettermark and name, converted users with rate from the first step, dropped-off users with rate from the previous step. Unordered funnels still label steps "Completed N steps".
- Deletes the legacy implementation (`FunnelChart`, `FunnelBarVertical`, `StepBar(s)`, `StepLegend`, `FunnelTooltip`, `funnelUtils`) and the `convertExperimentResultToFunnelSteps` shim in `ResultDetails.tsx`, its only consumer. Net -555 lines.

The chart is visible on single-metric experiment results (inline under the metrics table) and in the metric details modal. The `Scenes-App/Experiments > ExperimentWithFunnelMetric` story covers it in visual review.

I'm an agent, so no manual browser testing. What I ran:

- `pnpm --filter=@posthog/frontend typescript:check` - no errors in touched files (pre-existing unrelated failures in `products/tracing` from missing local typegen)
- oxlint + oxfmt on the touched files
- Verified no remaining references to the deleted files or `convertExperimentResultToFunnelSteps`

No new tests: this is a rendering swap with no new logic beyond data mapping, and the existing `ExperimentWithFunnelMetric` storybook snapshot exercises the new chart. The persons-modal step gating logic was ported unchanged from `StepBar.tsx`.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (Fable). Sam directed the work: adopt the new quill funnel chart in experiments. Key decisions:

- Direct replacement rather than a feature flag, matching how other quill chart adoptions landed; the legacy renderer was deleted since `ResultDetails.tsx` was its only consumer.
- The insights-funnel data conversion layer was dropped entirely: the quill chart takes `steps: string[]` + `Series[]` directly, so the mapping shrank to two memos in the new component, with raw counts carried in `series.meta` for the tooltip and click handler.
- Step naming, variant coloring, and persons-modal gating were ported as-is from the deleted code.

No repo skills were invoked (no DRF/migration/test surface touched).


frontend/src/scenes/experiments/stories/ExperimentWithMultiStepFunnelMetric.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    {},
+                [`/api/projects/:team_id/feature_flags/${EXPERIMENT_WITH_MULTI_STEP_FUNNEL_METRIC.feature_flag.id}/status/`]:
+                    {},
+                [`/api/environments/:team_id/default_release_conditions/`]: [],
+            },
+            post: {
+                '/api/environments/:team_id/query/:kind': async ({ request }) => {
+                    const body = (await request.json()) as Record<string, any>
+
+                    if (body.query.kind === NodeKind.ExperimentExposureQuery) {
```
sampennington, line 39: 🤖 **sp-ship** · 🦔 `[personal]` · **Blocker**

Branching logic in a story-body MSW handler (`if (body.query.kind === NodeKind.ExperimentExposureQuery) { return … } return …`). Sam's no-branches-in-tests rule flags this; prefer routing by kind via a lookup map computed up front. Caveat: this `if`/return shape is identical across all 11 sibling experiment `.stories.tsx` files, so staying consistent is a strong counter-argument — reasonable to defer unless the convention changes repo-wide.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,335 @@
...
+        () => [result.baseline, ...(result.variant_results ?? [])].filter(Boolean),
+        [result.baseline, result.variant_results]
+    )
+
+    const numMetricSteps = Math.max(metric.series.length, ...variants.map((v) => v.step_counts?.length ?? 0))
+
+    const steps = useMemo(() => {
+        const isUnordered = metric.funnel_order_type === StepOrderValue.UNORDERED
+        return [
+            'Experiment exposure',
```
sampennington: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion / follow-up**

The `'Experiment exposure'` step label is the semantic pivot the whole exposure/step-offset logic depends on (index 0 = exposure). Pull it into a named constant (e.g. `EXPOSURE_STEP_LABEL`) so the convention is self-documenting.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,335 @@
...
+                                value={percentage(funnelConversionRate(count, meta.counts[0] ?? 0), 2, true)}
+                            />
+                            <TooltipRow
+                                label="Conversion from previous"
+                                value={percentage(funnelConversionRate(count, previous), 2, true)}
+                            />
+                        </>
+                    )}
+                    {!!experimentQuery && stepIndex > 0 && (
+                        <div className="mt-1 pt-1 border-t border-current/25 text-xs opacity-60 text-center">
```
sampennington: 🤖 **sp-ship** · 📊♻️ `[hog-charts+reuse]` · **Suggestion / follow-up**

Two things on this tooltip footer hint: (1) it renders "Click to view users" whenever `experimentQuery && stepIndex > 0`, but `openExperimentPersonsModal` silently no-ops for a step-1 drop-off click (`!converted && backendStepNo === 1`) — so a user can be told to click and get nothing; consider suppressing the hint for that non-actionable case. (2) the class string is byte-identical to quill `DefaultTooltip`'s `footer` slot — reuse that prop (or extract the shared hint) so the styling stays in sync with the other quill tooltips.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,335 @@
...
+    return `Step ${stepNumber}`
+}
+
+/**
+ * Opens the persons modal for a funnel step. The frontend prepends an "Experiment exposure"
+ * step at index 0 that the backend actors funnel doesn't have, so frontend step index N maps
+ * to backend step number N. The exposure step itself can't be queried, and neither can
+ * drop-offs at the first metric step ("exposed but never entered the funnel").
+ */
+function openExperimentPersonsModal({
```
sampennington: 🤖 **sp-ship** · 📊🧪 `[hog-charts+test]` · **Suggestion / follow-up**

`openExperimentPersonsModal` is the subtlest logic in the PR — the FE→backend step-index mapping, the two early-return guards, and the `converted ? backendStepNo : -backendStepNo` sign flip — but the only coverage is the default-render Storybook snapshot (`play: makeDelay(500)` never clicks a bar). A regression that offsets the step or flips the sign would query the wrong persons cohort silently. The quill `FunnelChart` primitive is unit-tested upstream, but this experiment glue isn't — add a focused test asserting the `ExperimentActorsQuery` `funnelStep` for converted vs drop-off clicks across step indices (including the exposure step and the step-1 drop-off early-returns).

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,335 @@
...
+                </div>
+            </Tooltip>
+            {previousCount != null && (
+                <Tooltip title="Users who didn't complete this step, with drop-off rate relative to the previous step">
+                    <div className="flex items-center gap-1.5">
+                        <IconTrendingFlatDown className="text-danger shrink-0" />
+                        <span>
+                            {pluralize(droppedOff, 'user')}{' '}
+                            <span className="text-secondary">
+                                ({percentage(1 - funnelConversionRate(count, previousCount), 2)})
```
sampennington: 🤖 **sp-ship** · 📊 `[hog-charts]` · **Suggestion / follow-up**

Drop-off percentage uses `1 - funnelConversionRate(count, previousCount)`; when `previousCount === 0` that renders "100%" dropped off even though 0 of 0 dropped. Cosmetic edge (only when a prior step has zero users), but a zero guard would keep it consistent with the collapse-to-0 behaviour elsewhere.

sampennington (commented):
🤖 **sp-ship** · 🚫 Blockers

Clean, faithful migration — nine reviewers (code, react, efficiency included) found no logic or perf regressions, and the persons-modal click mapping, variant colors, conversion/drop-off math and tooltip content were all verified preserved against the deleted Chart.js viz (~900 lines removed). The single Blocker is a no-branching-in-tests flag on the story's MSW response handler, which is byte-identical to all 11 sibling experiment stories — a judgment call that's reasonable to defer. Remaining suggestions are minor: a magic-string → constant, a misleading "click to view users" hint on the non-actionable step-1 drop-off, a zero-guard on the drop-off %, and (most useful) a focused unit test for the untested FE→backend step-index mapping.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,318 @@
...
+                data: counts.map((count) => funnelConversionRate(count, counts[0]) * RATE_TO_PERCENT),
+                meta: { variantKey: variant.key, counts },
+            }
+        })
+    }, [variants, experiment, numMetricSteps])
+
+    const stepTotals = useMemo(
+        () => steps.map((_, stepIndex) => series.reduce((sum, s) => sum + (s.meta?.counts[stepIndex] ?? 0), 0)),
+        [steps, series]
+    )
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Already fixed — `hiddenKeys` state is now tracked and threaded into the legend `config` and `stepTotals` filtering in `c594210`, matching the suggested fix exactly. Resolving.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
```diff
@@ -0,0 +1,335 @@
...
+                            />
+                            <TooltipRow
+                                label="Conversion from previous"
+                                value={percentage(funnelConversionRate(count, previous), 2, true)}
+                            />
+                        </>
+                    )}
+                    {!!experimentQuery && stepIndex > 0 && (
+                        <div className="mt-1 pt-1 border-t border-current/25 text-xs opacity-60 text-center">
+                            Click to view users
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Deferring — the tooltip is rendered once per bar/index and quill-charts does not currently surface which sub-region (converted vs. drop-off track) the cursor is over at hover/render time (`TooltipContext.inTrackArea` is declared but only populated during click resolution, never during hover). So a correct per-region hint would need new chart API surface, not a local one-file tweak. Leaving open for a human call on whether to suppress the hint entirely at step 1, add the missing chart plumbing, or accept the minor overstatement.

frontend/src/scenes/experiments/stories/ExperimentWithMultiStepFunnelMetric.stories.tsx
sampennington: 🤖 **sp-triage**

Deferring — confirmed the branching MSW pattern is identical across all 11 sibling `*.stories.tsx` files under `frontend/src/scenes/experiments/stories/`. Blocker bucket, but this is a repo-wide convention call, not a single-file fix — leaving open for a human decision on whether to change the convention everywhere or accept it here.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: 🤖 **sp-triage**

Fixed in `2a0abc1` — extracted `EXPOSURE_STEP_LABEL` and used it in the `steps` array.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: 🤖 **sp-triage**

Deferring — same underlying constraint as the greptile thread on this line: quill-charts doesn't expose which sub-region (converted vs. drop-off) is under the cursor at tooltip-render time, so cleanly gating the hint for the step-1 drop-off case needs new chart API surface, not a local tweak. The `DefaultTooltip` footer-slot reuse suggestion is a smaller, separable ask but ties into the same hint logic — leaving both parts open for a human call.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: 🤖 **sp-triage**

Deferring — agree a focused test on `openExperimentPersonsModal`'s step-index mapping and sign flip would be valuable, but authoring it is a real test-design task (picking assertions across converted/drop-off/exposure/step-1 cases), not a mechanical single-file fix. Leaving open for a human to pick up.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: 🤖 **sp-triage**

Fixed in `2a0abc1` — guarded the drop-off rate to 0 when `previousCount` is 0, extracted into a `droppedOffRate` local for readability.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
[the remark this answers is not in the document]
sampennington: Not a regression, and `stepIndex > 1` would make it worse — closing as intended behavior.

The deleted `StepBar.tsx` gated the hint identically, and its comment spelled out the same carve-out:

```ts
// Only show "Click to inspect actors" hint when clicking will actually work:
// - Step 0 (exposure): can't use actors query (returns early), so don't show hint
// - Step 1 (first metric) drop-offs: can't query (no exposure in backend funnel), conversions work
// - Step 2+: both conversions and drop-offs work
const hasClickableData = stepIndex > 0
```

That flowed into `FunnelTooltipContent`'s `<ClickToInspectActors />`, per hovered bar, not per sub-region — so the pre-existing chart also advertised a click over step 1's drop-off track. The new `stepIndex > 0` is a faithful port.

On the suggestion: at step 1 the converted track *is* clickable, and it's the bar the user is most likely aiming at. `stepIndex > 1` would drop a truthful hint from a working target to avoid an untruthful one on a secondary target — net worse than what shipped before.

One residual difference worth naming: legacy set `cursor: default` on the non-clickable regions, whereas quill sets the pointer cursor for the whole hovered band (`Chart.tsx`: `pointer={hoverIndex >= 0 && !!onPointClick}`). Making either the cursor or the hint region-aware needs quill to populate `TooltipContext.inTrackArea` on hover (today it's only set during click resolution) — worth doing when a second consumer wants it, not worth blocking this swap on.

frontend/src/scenes/experiments/stories/ExperimentWithMultiStepFunnelMetric.stories.tsx
sampennington: Closing as convention. All 11 sibling stories under `frontend/src/scenes/experiments/stories/` route the shared `POST /api/environments/:team_id/query/:kind` mock with the same `if (body.query.kind === …) return …` shape, because one endpoint has to serve both the exposure query and the metric query. Diverging in this one file would cost more than the branch does, and changing it is a repo-wide sweep, not a line in this PR.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: Both parts settled in `37ad444`.

**(2) footer reuse** — fixed by extracting it. `DefaultTooltip`'s `footer` prop isn't reusable here (this tooltip's rows are bespoke per hovered variant, not one row per series), so instead the shared div is now `TooltipFooter` in `@posthog/quill-charts`, rendered by `DefaultTooltip` *and* by this tooltip. One place to change, no drift. Documented in `packages/quill/packages/charts/AGENTS.md`.

**(1) the step-1 drop-off hint** — closing as intended, same as the greptile thread above: the deleted `StepBar.tsx` gated the hint with `const hasClickableData = stepIndex > 0` and its comment already acknowledged that step-1 drop-offs aren't queryable while conversions are, so `stepIndex > 0` is a faithful port rather than a new overstatement. Tightening it to `stepIndex > 1` would remove a truthful hint from step 1's *converted* bar — the likelier click target — to suppress an untruthful one on the drop-off track.

frontend/src/scenes/experiments/charts/funnel/ExperimentFunnelChart.tsx
sampennington: Covered in `cbe5e24`. Pulled the risky part out as `experimentActorsFunnelStep(stepIndex, converted)` — the exposure-step offset, both early returns (now a `null` return), and the drop-off sign flip in one named unit — and added a parameterized `ExperimentFunnelChart.test.ts` over exposure / first metric step / later steps × converted / drop-off, asserting the exact `funnelStep` the actors query gets. An off-by-one or a flipped sign now fails the test instead of silently querying the wrong cohort.

andehen (approved):
Nice, code looks good! Also tested locally, worked fine here 👍

PR #70999, opened 2026-07-15
feat(mcp): add scoped path-cleaning-rules-update tool

pauldambra:
## Problem

Editing a project's path cleaning rules over MCP today means calling `project-settings-update` with the **entire** `path_cleaning_filters` list. That is a sharp edge for an agent surface: the model has to read the current rules, reconstruct all of them, renumber `order` by hand, and re-send the whole array — forgetting one silently drops it, and it can clobber rules the team already configured.

This came from a Slack thread where the in-app assistant could correctly *propose* path cleaning rules but had no safe way to *apply* them.

## Changes

Adds a hand-authored MCP tool **`path-cleaning-rules-update`** (registered in `TOOL_MAP`, metadata in `tool-definitions.json`) that does the read-modify-write for you:

- Granular ordered operations: `append`, `insert`, `replace`, `remove`, `reorder`. Auto-manages `order`.
- Built-in **preview → confirm**: unless `confirm: true`, it returns the resulting rules (and how they rewrite any `sample_paths`) **without saving**.
- Validates every regex and resolves aliases up front, so an invalid op fails atomically with an actionable message rather than half-applying.
- New `updatePathCleaningFilters` method on the projects API client.
- Updates the `managing-path-cleaning-rules` skill to prefer this tool over the whole-list overwrite.

Prefer this over `project-settings-update` for path cleaning. It's a durable home for the capability as the in-app assistant unifies onto the MCP tools.

Automated tests only (the agent did not run the tool against a live project):

- New unit suite `tests/unit/path-cleaning-rules-update.test.ts` (17 tests) covering the pure ops (append/insert/replace/remove/reorder, index clamping, unknown-alias and non-permutation errors, invalid-regex rejection, no input mutation) and the handler's preview-vs-confirm behavior (no write on preview, renumbered write on confirm, read-failure surfaces without a write).
- Ran the affected invariant suites green: `tool-catalog`, `tool-schema-snapshots` (new snapshot added for the tool), `tool-references`, `custom-tool-handlers`, `generate-tools`, `skills-generated`, `tool-name-validation`.
- Confirmed the changed TypeScript files typecheck clean (remaining repo typecheck errors are pre-existing `@posthog/quill` UI-app resolution, untouched by this PR).

These catch the regressions the tool exists to prevent: dropping/clobbering existing rules, mis-numbered `order`, and half-applied edits on a bad operation.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Skill `managing-path-cleaning-rules` updated in the same PR.

## 🤖 Agent context

Authored by the PostHog Slack app (Claude) from a Slack thread. The requester wanted the capability built directly rather than filed as an issue. Skill invoked: `managing-path-cleaning-rules` (to ground the tool's semantics — rule ordering, re2 aliasing, global/historical impact).

Key decision: implement as a **hand-authored custom tool** (like `read-data-schema` / `event-definition-update`) rather than a codegen'd OpenAPI wrapper, because `path_cleaning_filters` is a plain field on the project with no dedicated endpoint — the value is in the read-modify-write + preview logic, which codegen can't express. Chose a built-in `confirm` preview gate so the safety flow doesn't depend on client UI.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C05LJK1N3CP/p1784103523405569?thread_ts=1784103523.405569&cid=C05LJK1N3CP)*

### DIFF
--- a/products/web_analytics/skills/managing-path-cleaning-rules/SKILL.md
+++ b/products/web_analytics/skills/managing-path-cleaning-rules/SKILL.md
@@ -137,13 +137,13 @@ applies granular operations (`append`, `insert`, `replace`, `remove`,
 returns a **preview** of the resulting rules without saving. Pass
 `sample_paths` to see how the resulting set rewrites real paths.
 
-Preview first (no `confirm`), surface it to the user, then apply:
+First call it without `confirm` to get the preview, surface that to the user,
+then re-send the same call with `"confirm": true` to save:
 
 ```json
 {
   "operations": [{ "action": "append", "alias": "/users/<id>/profile", "regex": "/users/\\d+/profile" }],
-  "sample_paths": ["/users/123/profile", "/users/me/profile"],
-  "confirm": true
+  "sample_paths": ["/users/123/profile", "/users/me/profile"]
 }
 ```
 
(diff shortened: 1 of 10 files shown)


lricoy (approved):
LGTM! Thanks for doing it. 

I think the reviews were addressed; I will rebase this on https://github.com/PostHog/posthog/pull/65832 after. I thought we had merged it already 🙈

PR #71000, opened 2026-07-15
feat(integrations): add just-in-time OAuth token refresh to the gateway

benjackwhite:
## Problem

Today Django's Celery beat refreshes every integration's OAuth token, partly in Celery and partly in the model layer. The integration-gateway is meant to own credential lifecycle behind one audited service. This PR makes the gateway the just-in-time refresher for the OAuth2 kinds it owns, so a consumer reading through the gateway always gets a fresh token, and refresh logic lives in one place.

## Changes

Refresh is **opt-in per kind** and coordinated with Django so exactly one system refreshes a given kind.

**Rust (`rust/integration-gateway`)**
- New `refresh/` module: half-life expiry check (mirrors `OauthIntegration.access_token_expired`), a generic `grant_type=refresh_token` flow for the initial provider set (hubspot, salesforce, google-ads/analytics/search-console/sheets), a Redis single-flight lock per integration (`common-redis` `set_nx_ex`) so refresh-token-rotating providers can't double-refresh, re-read-under-lock, write-back of re-encrypted tokens, and `errors = TOKEN_REFRESH_FAILED` on failure (same sentinel Django uses).
- `encrypt_leaf` on the existing Fernet decryptor (primary key = `ENCRYPTION_SALT_KEYS[0]`, byte-compatible with `EncryptedJSONField`). Refresh runs on the DB-load path before decrypt/cache, so the in-proc cache is written through and never serves a stale pre-refresh token.
- Fail-open: any refresh error returns the existing token (still valid, since we refresh at half-life). Enabled via `INTEGRATION_GATEWAY_REFRESH_KINDS`; empty (default) means pure pass-through and no Redis dependency.

**Django**
- The `refresh_integrations` beat excludes `settings.INTEGRATION_GATEWAY_REFRESH_KINDS` (the same env var the gateway reads), so a gateway-owned kind isn't refreshed twice. Empty default = no behavior change.

```diff
- Integration.objects.filter(kind__in=OauthIntegration.supported_kinds).exclude(kind="meta-ads")
+ Integration.objects.filter(kind__in=OauthIntegration.supported_kinds)
+     .exclude(kind="meta-ads")
+     .exclude(kind__in=settings.INTEGRATION_GATEWAY_REFRESH_KINDS)
```

> [!NOTE]
> Stacked on #70992 (base `posthog-code/integration-gateway-cdp-cutover`), which is itself stacked on #70450. Review/merge in order.

> [!WARNING]
> A kind must be added to `INTEGRATION_GATEWAY_REFRESH_KINDS` in the same change that enables the gateway to refresh it — the setting drives both the gateway (what it refreshes) and the Django beat (what it stops refreshing). Enabling one side without the other either double-refreshes or leaves a kind unrefreshed.

**Scoped out for now** (stay on the Django beat): Slack (DB-sourced creds), the HTTP-Basic providers (reddit/pinterest/stripe), TikTok, Bing, Jira (rotating refresh token), Meta, and the service-account/GitHub kinds; the `reload-integrations` publish on refresh (`common-redis` has no publish; staleness is bounded and safe given half-life refresh — the pre-refresh token is still valid, and the gateway write-through cache covers gateway-path readers); and a dedicated write DB pool.

**Why:** move token refresh into the audited gateway, one provider at a time, without a dual-writer race against Django's beat.

I (Claude) have **no Rust toolchain in this environment**, so I could not compile the crate — relying on CI for build/clippy. Automated tests added:

- Rust unit tests: `refresh::expiry` (half-life threshold, missing-fields, Salesforce default), `refresh::providers` (unknown kind / missing creds / override), and a `crypto_parity` encrypt round-trip.
- Crypto write parity is established transitively: the PR-1 fixture proves this salt key decrypts Django/`cryptography`-produced ciphertext (so the key is byte-identical to Django's), and `encrypt_leaf` uses that same key — so by the Fernet spec a value encrypted here is Django-readable. The true end-to-end (gateway writes a row, Django reads it) is the local e2e below.
- Python: `test_refresh_integrations_skips_gateway_owned_kinds` proves the beat excludes gateway-owned kinds (the dual-writer safety property).

Local e2e (not run here): seed an expired-but-refreshable hubspot integration with a valid refresh token, point `REFRESH_TOKEN_URL_OVERRIDE` at a mock token endpoint, set `INTEGRATION_GATEWAY_REFRESH_KINDS=hubspot`, then `/api/v1/credentials/fetch` returns a fresh token, writes it back re-encrypted, and the Django beat skips hubspot; killing the mock sets `TOKEN_REFRESH_FAILED` and the stale token is still served.

## 🤖 Agent context

Ben directed this; I (Claude) implemented it. Key decisions: single per-kind ownership list drives both sides (gateway + beat) so coordination can't drift; refresh only on the cache-miss/DB-load path (write-through) so we never cache a stale token; fail-open on every error path since a gateway blip must not break a consumer; scoped the initial provider set to the generic OAuth2 flow and left exotic/rotating/service-account providers on the Django beat. The `/writing-tests` skill wasn't available in this session — its value-gate was applied by hand (each test names a unique regression). Deferred the `reload-integrations` publish because `common-redis` exposes no publish and half-life refresh makes the staleness safe.

---

### DIFF
--- a/posthog/settings/integrations.py
+++ b/posthog/settings/integrations.py
@@ -1,5 +1,7 @@
+import os
+
 from posthog.settings.base_variables import DEBUG, TEST
-from posthog.settings.utils import get_from_env, str_to_bool
+from posthog.settings.utils import get_from_env, get_list, str_to_bool
 
 # integration-gateway (Rust credential service). URL Django/consumers call, and the dedicated
 # scoped-JWT signing secret (see posthog/integration_gateway_jwt.py). The secret defaults to a known
@@ -11,6 +13,10 @@ INTEGRATION_GATEWAY_JWT_SECRET = get_from_env(
     "INTEGRATION_GATEWAY_JWT_SECRET",
     LOCAL_DEV_INTEGRATION_GATEWAY_JWT_SECRET if DEBUG or TEST else "",
 )
+# Integration kinds the gateway refreshes, sharing the exact env var the gateway reads. These are
+# excluded from the Celery refresh beat (posthog/tasks/integrations.py) so exactly one system
+# refreshes a given kind. Empty (default) = the beat refreshes everything, gateway refreshes nothing.
+INTEGRATION_GATEWAY_REFRESH_KINDS = get_list(os.getenv("INTEGRATION_GATEWAY_REFRESH_KINDS", ""))
 
 HUBSPOT_APP_CLIENT_ID = get_from_env("HUBSPOT_APP_CLIENT_ID", "")
 HUBSPOT_APP_CLIENT_SECRET = get_from_env("HUBSPOT_APP_CLIENT_SECRET", "")
--- a/posthog/tasks/integrations.py
+++ b/posthog/tasks/integrations.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+
 from celery import shared_task
 
 from posthog.models.integration import (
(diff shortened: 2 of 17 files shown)


benjackwhite:
Replaced by #72530 — the token-refresh branch could not accept signed-git writes (every commit/rewrite/merge was refused), so the same change was re-pushed on a fresh branch off the updated PR2 base. This PR's branch is left in place for reference.

PR #71001, opened 2026-07-15
feat(pipeline): render app metric summary tiles as quill metric cards

sampennington:
## Problem

Stacked on #70834. After that PR, the app metric summary tiles render a full `TimeSeriesLineChart` in minimal mode plus hand-rolled header markup (label, total, plain-text percentage diff). Quill has a purpose-built pattern for exactly this shape of tile: the composable `Metric` from `@posthog/quill-components`, already used by the MCP analytics KPI tiles.

## Changes

- `AppMetricSummary` is now a quill `Card` (`size="sm"`, `flush`) rendering the composable `Metric` — same recipe as MCP analytics' `MetricTile`: header with title and change pill, period total as the headline, "vs. N prior" resting subtitle, and a full-bleed gradient sparkline hugging the card's bottom edge.
- Hovering the sparkline swaps the headline to the hovered point's value with its date as the subtitle (Metric's built-in behavior), replacing the old floating tooltip.
- Change pill direction semantics: new optional `goodDirection` prop, defaulting from the tile's `color` — danger/warning tiles (Failure, Dropped, Billing limited, ...) treat an increase as bad, so a drop in failures shows green. No call-site changes needed.
- States Metric doesn't model stay on the card: quill `Skeleton` loading state, "No data" state, `hideIfZero`, the `onClick` drill-down (arrow next to the title, quill tooltip), and the `footer` slot. The tile no longer uses any Lemon components.

At rest / hovering a point (local GeoIP transformation):

composable-metric-tiles

composable-metric-hover

I'm an agent; tested against the local dev stack on the GeoIP transformation metrics page (screenshots above): tiles render with real data, the Failure tile's negative change shows green (goodDirection inversion), the zero-data tile suppresses its pill, and sparkline hover swaps the headline/subtitle. Also ran:

- `pnpm --filter=@posthog/frontend typescript:check` — no errors in touched files
- oxlint + oxfmt on the changed file

No new tests: presentation-only swap; Metric behavior is covered by the quill packages' own tests.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (Fable), directed by Sam while testing #70834 — the tiles looked like a hand-rolled version of quill's metric tiles, so this swaps them over as a follow-up rather than widening the like-for-like parent PR. First cut used the monolithic `MetricCard` from quill-charts inside the old Lemon card shell; Sam compared it to the MCP analytics tiles and asked for that look, so it was redone with the composable `Metric` + quill `Card` recipe those tiles use. Other decisions: derive `goodDirection` from the existing tile `color` tokens instead of editing all eight consumer info-maps; suppress the resting subtitle with `''` when there's no comparison period so the total isn't captioned with the last bucket's date.


sampennington:
Closing — not pursuing this migration; focusing the quill chart work on the experiments funnel chart (#70991).

PR #71005, opened 2026-07-15
feat(capture): measure billing grace period ingestion

pauldambra:
## Problem

We allow trusted organizations to continue sending data for a grace period after they exceed a billing limit, but capture does not expose how much data is admitted during that window.

**Why:** We need a direct measure of grace-period ingestion to understand the volume allowed before billing enforcement starts.

## Changes

- Load the existing billing grace-period Redis set alongside capture quota limits.
- Count admitted analytics and replay events across analytics, capture v1, AI multipart, OTEL, and session replay paths.
- Expose `capture_events_admitted_during_billing_grace_period_total` with a low-cardinality `resource` label.

> [!NOTE]
> The `recordings` series counts accepted `$snapshot` events at capture. It does not deduplicate them into billable replay sessions.

- Added a focused Rust regression test that verifies only analytics events remaining ingestable after scoped quota filtering increment the grace-period counter.
- Added an endpoint-level replay regression test that verifies an accepted snapshot increments the `recordings` series.
- CI Rust lint and the capture/capture-logs test shard pass.
- Ran `git diff --check` successfully.
- I could not run Rust formatting or tests because this runner does not provide `cargo`, `hogli`, or `flox`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change is needed. This adds an internal capture metric.

## 🤖 Agent context

Codex implemented this change using the repository's `/ingestion-pipeline-doctor-nodejs` and `/writing-tests` skills. I reused the existing quota-suspension Redis set rather than introducing new billing state, and placed metric increments at event admission points to avoid counting events rejected by quota paths.

---


rust/capture/src/payload/analytics.rs
```diff
@@ -128,6 +128,10 @@ pub async fn handle_event_payload(
         .quota_limiter
         .check_and_filter(&context.token, events)
         .await?;
+    state
+        .quota_limiter
+        .report_grace_period_ingestion(&context.token, events.len() as u64)
```
pauldambra: > [!NOTE]

**[convergent: logic + reliability + maintainability]** 🟠 HIGH

The counter is emitted at quota admission, before downstream validation, restrictions, processing, or sink delivery. Calling it `ingested` overstates what it measures, especially for AI multipart and OTEL paths that can still reject the request afterward.

Rename the metric and helper to say `accepted` or `admitted`, and document that downstream drops are included.

rust/capture/src/quota_limiters.rs
```diff
@@ -86,13 +94,23 @@ impl CaptureQuotaLimiter {
             ServiceName::Capture,
         )
         .expect(&err_msg);
+        let grace_period_limiter = RedisLimiter::new(
+            redis_timeout,
+            redis_client.clone(),
+            QUOTA_LIMITING_SUSPENDED_CACHE_KEY.to_string(),
```
pauldambra, line 118: > [!NOTE]

**[convergent: logic + reliability]** 🟡 MEDIUM

The grace cache only observes the capture mode's global resource. Independent grace periods for scoped resources such as `exceptions`, `survey_responses`, and `llm_events` are not counted.

Either add scoped grace limiters with explicit overlap semantics, or keep this PR focused on global event/recording admission and make that scope explicit in naming and documentation.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: logic/correctness, performance/reliability, maintainability, and security

The implementation is operationally safe and low-cardinality, but the current metric name claims successful ingestion while the call sites measure quota admission. Scoped billing resources are also intentionally or accidentally outside the current measurement.

### Key findings

- 🟠 HIGH: rename the metric to match quota-admission semantics, or move emission to successful sink delivery.
- 🟡 MEDIUM: clarify or implement coverage for scoped grace resources.

### Convergence

- Three reviewers independently identified the ingestion-versus-admission mismatch.
- Two reviewers independently identified the scoped-resource coverage gap.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 correctness | Fix the semantic mismatch before merge. |
| 📈 reliability | Metric cardinality is safe; coverage and timing need clarification. |
| 📐 maintainability | Rename is the smallest robust fix; centralized outcomes are a larger follow-up. |
| 🛡 security | No security boundary or sensitive-label issue found. |

---

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments and the summary comment.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +238,63 @@ impl CaptureQuotaLimiter {
...
+        if event_infos.is_empty() {
+            return;
+        }
+
+        let mut unreported_indices: Vec<usize> = (0..event_infos.len()).collect();
+
+        // Scoped resources take precedence so each admitted event is counted once,
+        // even when both scoped and global grace periods overlap.
+        for limiter in &self.scoped_limiters {
+            if !limiter.is_in_grace_period(token).await {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + 4 other qa-team reviewers]** 🟡 MEDIUM

When only the global grace period is active, scoped-billable events are counted under the global resource label even though the global limiter never governs them on v0 paths. This loop `continue`s past scoped limiters that are not in grace without partitioning, so their matched events (`$exception`, survey events, `$ai_*`) fall through to `report_global_grace_period_admission` and increment `resource="events"`. But `check_and_filter` deliberately retains scoped-matched events when the global limit trips, meaning those events would have been admitted even with no grace — for exception- or AI-heavy orgs the `events` series can be dominated by traffic that was never at risk, overstating grace-window exposure. It also makes the same series mean different things on v0 vs v1 (v1's global check is all-or-nothing, so the fallback attribution is accurate there).

Suggested fix: partition against every scoped limiter regardless of its grace state — increment the scoped counter only when that scope is in grace, but always remove scoped-matched indices from the global fallthrough — or explicitly document the "everything sent while in grace counts under the global resource" semantics on the metric constant.

rust/capture/src/otel/mod.rs
```diff
@@ -184,7 +184,10 @@ pub async fn otel_handler(
             }
         };
     }
-
+    state
```
pauldambra, line 187: > [!NOTE]

**[qa-team/generalist-b]** 🟡 MEDIUM

"Admitted" is counted before stages that can still reject the counted events, and the ordering is inconsistent across paths: here the report fires before `check_restrictions` (which rejects the whole batch all-or-nothing); the AI endpoint counts before multipart retrieval/parsing and blob upload; v0 recordings counts before session-id validation; and every path counts before the Kafka sink. Any post-count failure returns an error that SDKs retry, re-incrementing the counter on each attempt while events are ingested at most once — during a sink incident an in-grace org's retries inflate the series for data never ingested.

Suggested fix: move the report after the last drop point per path (at minimum after `check_restrictions` here), or document on the metric constant that it means "passed the billing quota gate" and is an upper bound on ingested volume.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +238,63 @@ impl CaptureQuotaLimiter {
...
+        }
+
+        counter!(
+            CAPTURE_EVENTS_ADMITTED_DURING_BILLING_GRACE_PERIOD_TOTAL,
+            "resource" => Self::get_resource_for_mode(self.capture_mode).as_str()
+        )
+        .increment(event_count);
+    }
+
+    pub async fn report_grace_period_admission<T: HasEventName>(&self, token: &str, events: &[T]) {
```
pauldambra, line 308: > [!NOTE]

**[convergent: qa-team/performance + paul + xp]** 🟡 MEDIUM

This runs on every request, and `report_grace_period_admission` always builds a `Vec<EventInfo>` (with a per-event `has_property` probe) plus an index vector before ever consulting the grace sets — even though grace sets hold a handful of orgs, so effectively all capture traffic pays this cost for nothing on the hottest paths in the service. The classification work also duplicates what `check_and_filter` computed moments earlier and discarded.

Suggested fix: hoist the cheap in-memory membership checks (global grace plus each scoped grace) to the top and return early when the token is in no grace period, building event infos only when at least one grace period is active; apply the same guard to the v1 shim's `admitted_event_infos` construction. Longer term, `check_and_filter` is the one place that already knows both "what survived" and "which resource each event maps to" — emitting the grace counter from inside that same walk would say it once.

rust/capture/src/quota_limiters.rs
```diff
@@ -86,13 +94,23 @@ impl CaptureQuotaLimiter {
             ServiceName::Capture,
         )
         .expect(&err_msg);
+        let grace_period_limiter = RedisLimiter::new(
```
pauldambra, line 115: > [!NOTE]

**[convergent: qa-team/performance + paul + xp]** 🟢 LOW

This doubles the background Redis pollers per capture pod (each scoped limiter plus the global now gets a grace twin, each spawning its own periodic `zrangebyscore` loop against the shared billing Redis) — and recordings-mode pods also poll scoped grace keys (exceptions/surveys/llm_events) they can never match. Mitigations verified: the suspended zsets are small, the request path never touches Redis (`is_limited` is a DashMap read), poll failure fails open, and the loaded-tokens gauge distinguishes the grace sets. Acceptable as-is, but this doubles again with each future scoped limiter — consider one shared fetch per resource feeding both the limit and grace maps, and skipping scoped grace limiters in Recordings mode.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +238,63 @@ impl CaptureQuotaLimiter {
         Ok(filtered_events)
     }
 
+    pub async fn report_global_grace_period_admission(&self, token: &str, event_count: u64) {
+        if event_count == 0 || !self.grace_period_limiter.is_limited(token).await {
```
pauldambra: > [!NOTE]

**[qa-team/compatibility]** 🟢 LOW

Counting keys only off the suspended set, but the Python writer's per-org update path (`ee/billing/quota_limiting.py`) adds a token to `@posthog/quota-limits/<resource>` without removing the still-future-scored member from `@posthog/quota-limiting-suspended/<resource>`; only the scheduled full sync restores mutual exclusion. In that window a hard-limited token still looks "in grace" to this reader, and scoped events retained from a globally-limited batch get counted as grace admissions for an org that is actually limited (conversely, an org that upgrades mid-grace keeps counting until the next sync). Cheap defense on the Rust side: skip counting when the corresponding hard limiter also matches the token (limited overrides grace), which also makes the metric robust to writer-side skew. The rest of the cross-language contract checks out: key prefix, resource suffixes, token members, and epoch-second score semantics all match the Python writer exactly.

rust/common/limiters/src/redis.rs
```diff
@@ -35,6 +35,7 @@ use tokio::time::interval;
 // due to historical reasons we use different suffixes for quota limits and overflow
 // hopefully we can unify these in the future
 pub const QUOTA_LIMITER_CACHE_KEY: &str = "@posthog/quota-limits/";
+pub const QUOTA_LIMITING_SUSPENDED_CACHE_KEY: &str = "@posthog/quota-limiting-suspended/";
```
pauldambra, line 38: > [!NOTE]

**[qa-team/generalist-b]** 🟢 LOW

The new suspended-set reader inherits `RedisLimiter`'s stale-on-error behavior: score expiry (`zrangebyscore(now, +Inf)`) is evaluated only server-side in Redis, and on fetch failure the last-known DashMap contents are kept with just a warn log. If Redis is unreachable past a grace expiry, every pod keeps counting that org's full traffic as grace-admitted indefinitely. Fail-open is the right call for enforcement limiters, but for a measurement set it fails toward silently wrong numbers with no staleness signal beyond the existing loaded-tokens gauge — worth noting where the metric is documented, or store scores in the map and filter expiry client-side for the grace limiters.

rust/capture/src/prometheus.rs
```diff
@@ -5,6 +5,8 @@ use metrics::counter;
 use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
 
 pub const CAPTURE_EVENTS_DROPPED_TOTAL: &str = "capture_events_dropped_total";
+pub const CAPTURE_EVENTS_ADMITTED_DURING_BILLING_GRACE_PERIOD_TOTAL: &str =
```
pauldambra, line 45: > [!NOTE]

**[convergent: qa-team/generalist-b + paul]** 🟢 LOW

The unit caveats live only in the PR description. `resource="recordings"` counts raw `$snapshot` events — everywhere else in billing, `recordings` means deduplicated billable sessions, so someone staring at a dashboard could easily read this as "billable recordings admitted" and be off by a lot. The counter as a whole measures "admitted past the billing quota gate" rather than ingested volume. Add a short comment on this constant stating the per-resource units and the measurement point so the caveats survive beyond the PR for the next engineer building a dashboard on it.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +238,63 @@ impl CaptureQuotaLimiter {
...
+            .map(|event| EventInfo {
+                name: event.event_name(),
+                has_product_tour_id: event.has_property("$product_tour_id"),
+            })
+            .collect();
+        self.report_grace_period_admission_for_event_infos(token, &event_infos)
+            .await;
+    }
+
+    pub async fn report_grace_period_admission_for_event_infos(
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-a + xp]** ⚪ NIT

Correct measurement now depends on every admission path remembering to pair `check_and_filter` with the right `report_*` call — five hand-wired sites, including the two-call sequence in the v1 shim and AI endpoint where the gateway-verified split ("wallet-billed, count it globally") is hand-rolled twice in two different ways. A future endpoint that adds quota checking will compile and ship with grace measurement silently missing, and a caller can get the two-method combination wrong. Consider a single entry point on the limiter that takes the events (plus a gateway-verified marker where relevant) and does the whole split internally — or at minimum a doc comment on `check_and_filter` stating that callers must report the admitted set.

rust/capture/src/v1/quota_limiter_shim.rs
```diff
@@ -92,27 +92,51 @@ pub async fn apply_quota_limits(
...
+        .iter()
+        .filter(|event| event.result == EventResult::Ok)
+        .filter_map(|event| {
+            if event.is_gateway_verified {
+                global_only_event_count += 1;
+                None
+            } else {
+                Some(EventInfo {
+                    name: event.event_name(),
+                    has_product_tour_id: event.has_property("product_tour_id"),
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

This looks up `"product_tour_id"` while the v0 path looks up `"$product_tour_id"` — correct, because the two event representations store the flag under different keys, but it's a subtle trap the next person will trip over (the commit log shows it already took a couple of rounds to get right). A one-line comment here noting why the `$` is absent in the v1 representation would earn its place.

rust/capture/src/prometheus.rs
```diff
@@ -5,6 +5,8 @@ use metrics::counter;
...
+    "capture_events_admitted_during_billing_grace_period_total";
```
pauldambra, line 46: > [!NOTE]

**[paul]** ⚪ NIT

love this. measure-first before you ever touch the limiting behaviour is exactly right — every time we've added a limiter and got the condition wrong we've wished we'd shipped a "would have limited" counter first. one question while i'm here: what's the plan once we can see the numbers? is there a follow-up that acts on this, or is it purely to size the grace window for now? not blocking, just want to know if there's a next step i should be looking for.

rust/capture/src/quota_limiters.rs
```diff
@@ -86,13 +94,23 @@ impl CaptureQuotaLimiter {
...
         .expect(&err_msg);
+        let grace_period_limiter = RedisLimiter::new(
+            redis_timeout,
+            redis_client.clone(),
+            QUOTA_LIMITING_SUSPENDED_CACHE_KEY.to_string(),
+            config.redis_key_prefix.clone(),
+            Self::get_resource_for_mode(config.capture_mode),
+            ServiceName::Capture,
+        )
+        .expect("failed to create billing grace period limiter");
```
pauldambra: > [!NOTE]

**[paul]** ⚪ NIT

nit picking — the global limiter just above builds an `err_msg` with the capture mode in it, but the grace limiter `.expect()`s a static string. if this ever blows up at startup it'd be nice to have the same context. tiny thing, ship as you see fit.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists + generalists), paul-reviewer, xp-reviewer, security-audit

## Verdict: 💬 APPROVE WITH NITS

Metrics-only change with sound underlying semantics — the Rust reader's contract with the Python billing writer was verified end to end, no admission decision changes on any path, and security found nothing. The findings are all about the *accuracy* of the new metric (which is the PR's entire deliverable) and hot-path cost; none block merge, but the three MEDIUMs deserve a look before dashboards consume the numbers.

### Key findings

- 🟡 MEDIUM — `quota_limiters.rs:279`: when only the global grace period is active, scoped-billable events (`$exception`, surveys, `$ai_*`) are attributed to the global `events`/`recordings` series even though the v0 global limiter deliberately exempts them — overstates grace exposure and makes v0/v1 series mean different things.
- 🟡 MEDIUM — `otel/mod.rs:187`: "admitted" is counted before later drop points (restrictions, validation, S3, Kafka sink), so SDK retries after post-count failures inflate the series; measurement points are inconsistent across paths.
- 🟡 MEDIUM — `quota_limiters.rs:253`: every request pays per-event allocation and classification before the (almost always negative) grace-membership check — hoist the cheap membership checks and return early.
- 🟢 LOW — doubled background Redis pollers per pod (incl. scoped grace pollers on recordings pods that can never match); limits/suspended set-overlap window in the Python writer can miscount hard-limited orgs; stale-on-error keeps counting expired grace periods during Redis outages; metric unit caveats (raw `$snapshot` events vs billable sessions) should live on the constant, not just the PR body.
- ⚪ NIT — five hand-wired report call sites (future endpoints can silently skip measurement); `product_tour_id` vs `$product_tour_id` key difference deserves a comment; static `.expect` message; and a question on what acts on this metric next.

### Convergence

- Hot-path guard before allocation: **qa-team/performance + paul + xp** independently.
- Doubled Redis pollers: **qa-team/performance + paul + xp** independently.
- Global-fallback attribution: 5 of 6 qa-team sub-reviewers.
- Count-point ordering: 5 of 6 qa-team sub-reviewers.
- Pin metric semantics in code: **qa-team/generalist-b + paul**.
- Single reporting entry point: **qa-team/generalist-a + xp**.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | LOW-to-MEDIUM risk, approve with nits: well-contained observability change, cross-language Redis contract verified; top concerns are scoped-event attribution under the global label and the pre-drop count point. |
| 👤 paul | likes the measure-first shape and the semantics; wants a cheap "in grace at all?" guard on the hot path before stamping, and flags the `recordings` label overloading a word that means billable sessions elsewhere. |
| 📐 xp | Solid, well-tested, right semantics; reservations are say-it-once — event→resource classification walked twice per batch, and the gateway-verified split hand-rolled in two callers instead of owned by the limiter. |
| 🛡 security-audit | No exploitable vulnerabilities: labels come from a static enum, no user input reaches Redis keys or metric labels, all new code is post-auth, post-quota, read-only. |

---

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +263,100 @@ impl CaptureQuotaLimiter {
...
+        if event_infos.is_empty() {
+            return;
+        }
+
+        let mut unreported_indices: Vec<usize> = (0..event_infos.len()).collect();
+
+        // Scoped resources take precedence so each admitted event is counted once,
+        // even when both scoped and global grace periods overlap.
+        for &limiter_idx in scoped_in_grace {
+            let limiter = &self.scoped_limiters[limiter_idx];
```
pauldambra: > [!NOTE]

**[convergent: qa-team (all 4 sub-reviewers) + paul + xp]** 🟡 MEDIUM

`report_grace_period_admission_for_event_infos` is `pub` and does panicking indexed access (`&self.scoped_limiters[limiter_idx]`) on the caller-supplied `scoped_in_grace: &[usize]`, with nothing tying those indices to the instance whose `scoped_limiters_in_grace_period` produced them. A security trace confirmed this is unreachable from user input today (both callers derive indices by enumerating the same instance's startup-fixed Vec), so this is a latent footgun rather than a live bug — but differently-shaped limiter instances already exist (production builds 3 scoped limiters; v1 test utils build 0), and a future caller reusing a cached grace set or copying the shim pattern against a second limiter turns a wrong metric into a per-request panic on the hottest ingestion path.

Suggested hardening (any one of): use `self.scoped_limiters.get(limiter_idx)` and skip-with-`debug_assert!` on miss (a skipped scoped label degrades to the global fall-through the metric already documents); make the method and its two sibling probes `pub(crate)` to shrink the misuse surface to the actual need (only the v1 shim calls them from outside this module); or mint an opaque newtype from `scoped_limiters_in_grace_period` that only this method accepts, so the "same limiter, same call" contract lives in the type system.

rust/capture/src/prometheus.rs
```diff
@@ -6,6 +6,36 @@ use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
...
+/// - **Attribution when only the global grace period is active.** Scoped
+///   grace periods (exceptions, surveys, LLM events) take precedence, but
+///   when a scoped limiter isn't itself in grace, events it would have
+///   matched fall through and are counted under the global resource label
+///   (`events`/`recordings`) instead of the scoped one.
+/// - **Fails toward stale counts, not toward silence.** The underlying
+///   Redis reader fails open: if Redis is unreachable past a grace-period
+///   expiry, pods keep counting that org's traffic as grace-admitted using
+///   the last-known state, with only a warn log and the existing
+///   loaded-tokens gauge as a staleness signal.
```
pauldambra: > [!NOTE]

**[convergent: qa-team/compatibility + 2 other qa-team reviewers]** 🟢 LOW

The new fail-open caveat is accurate except for one claim: it cites "the existing loaded-tokens gauge as a staleness signal", but that gauge is only written on successful Redis fetches — during an outage it freezes at its last healthy value and is indistinguishable from a healthy steady state, so it cannot detect the exact stale-count failure mode this bullet describes; the warn log is the only real signal. Suggest rewording to "with only a warn log as the failure signal (the loaded-tokens gauge freezes at its last value)".

While touching this comment: the attribution bullet names the scoped resources informally ("exceptions, surveys, LLM events") — spelling out the exact `resource` label literals would save a dashboard author from filtering on a label value that silently matches nothing.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +263,100 @@ impl CaptureQuotaLimiter {
...
+        if events.is_empty() {
+            return;
+        }
+
+        // Grace-period membership is a handful of orgs at a time and each check is
+        // an in-memory DashMap read (no Redis round-trip), so it's cheap to probe
+        // every limiter up front and skip the EventInfo allocation entirely when
+        // this token isn't in any grace period — the common case on every request.
+        let scoped_in_grace = self.scoped_limiters_in_grace_period(token).await;
+        if scoped_in_grace.is_empty() && !self.grace_period_limiter.is_limited(token).await {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-b + paul]** 🟢 LOW

Two small asymmetries introduced by the guard refactor, both bounded and both fixable the same way: (1) the guard reads global grace membership here and `report_global_grace_period_admission` re-reads it at count time, while the scoped set is a snapshot taken before reporting — a token flapping in/out of grace between the reads (background poller mutates these DashMaps concurrently) makes one batch's attribution internally inconsistent for up to one poll interval; (2) the global membership is probed twice per reported batch even though the whole point of the guard was frugality. No panic or double-count is possible and the metric is documented as approximate, so this is not blocking — but passing the guard's already-computed global-grace boolean down so one snapshot drives both the guard and the counting would fix both at once.

rust/capture/src/quota_limiters.rs
```diff
@@ -20,6 +26,20 @@ pub struct EventInfo<'a> {
     pub has_product_tour_id: bool,
 }
 
+impl<'a> EventInfo<'a> {
+    /// Builds an `EventInfo` from any `HasEventName` event. `product_tour_id_key`
+    /// is the property key to probe for `has_product_tour_id` — callers differ on
+    /// this because v0 events carry a raw `$`-prefixed property map (key
+    /// `"$product_tour_id"`) while v1's `WrappedEvent` promotes it to a
+    /// structured field checked via the unprefixed `"product_tour_id"` key.
+    pub fn from_event<T: HasEventName>(event: &'a T, product_tour_id_key: &str) -> Self {
```
pauldambra, line 35: > [!NOTE]

**[paul, with xp dissenting]** ⚪ NIT

paul: i'm slightly allergic to the stringly-typed `product_tour_id_key: &str` — it's two literals across four call sites, and a typo compiles fine and just quietly stops matching product-tour events. not a clean bool (prefix *and* map-vs-field semantics differ), so no strong steer — just flagging the allergy. ship as you see fit.

xp's counterpoint from the same review round: the parameter earns its place — v0's raw `$`-prefixed property map and v1's promoted structured field genuinely differ, the two keys are the whole reason the duplication existed, and the doc comment explains the divergence at the one point a reader meets it. Treat as a judgement call for the author.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +263,100 @@ impl CaptureQuotaLimiter {
...
+
+        self.report_global_grace_period_admission(token, unreported_indices.len() as u64)
+            .await;
+    }
+
+    /// True if `token` is in the global grace period for this capture mode's
+    /// resource. Scoped grace periods are checked separately via
+    /// `scoped_limiters_in_grace_period` — combine both to know whether
+    /// there's any grace-admission work to do for `token`.
+    pub async fn is_in_global_grace_period(&self, token: &str) -> bool {
```
pauldambra: > [!NOTE]

**[paul]** ⚪ NIT

nit picking: `is_in_global_grace_period` and `scoped_limiters_in_grace_period` (and the 3-arg report method) are `pub` but the only caller outside this module is the v1 shim, same crate. `pub(crate)` would keep the misuse surface for the index-slice coupling flagged above as small as the actual need. no worries if you'd rather keep it uniform.

pauldambra (commented):
> [!NOTE]

QA Swarm re-review of the two fix commits (5cd6ded, c3cc558) complete. See inline comments and the round-2 summary comment.

pauldambra:
> [!NOTE]
>
> Round-2 re-review of the fix commits (5cd6ded, c3cc558) by qa-team, paul-reviewer, xp-reviewer, security-audit

## Verdict: 💬 APPROVE WITH NITS

The round-1 fixes landed well: all four reviewers independently verified by inspection that the (uncompiled-in-authoring) Rust type-checks — every caller of the re-signatured 3-arg reporting method updated, `EventInfo::from_event` sound at all sites including the `&SpanEvent` OTEL path and `&mut WrappedEvent` coercion, no dangling references, and the existing tests unaffected. The metric doc comment was called out as "intent-revealing code at its best". Security audit traced and disproved all three risk hypotheses in the delta (index panic from user input, key-parameter injection, guard miscounting) — zero findings.

### Key findings

- 🟡 MEDIUM — `quota_limiters.rs:320`: caller-supplied `&[usize]` indexed into `self.scoped_limiters` unchecked; unreachable from user input today but a latent per-request panic footgun for future callers (flagged independently by all six reviewer perspectives; bounds-safe `.get()`, `pub(crate)`, or a newtype all fix it).
- 🟢 LOW — `prometheus.rs:35`: the doc's claim that the loaded-tokens gauge is a staleness signal is factually off (it freezes on outage); reword.
- 🟢 LOW — `quota_limiters.rs:288`: guard/count read the global grace state twice with a scoped snapshot in between; pass the boolean down for one consistent snapshot.
- ⚪ NIT — stringly-typed `product_tour_id_key` (paul allergic, xp fine with it); `pub` vs `pub(crate)` on the new probes; a note that the empty-guard now exists at three layers (fine as-is).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | LOW risk; fixes verified compile-sound and metric-output-equivalent; wants the index footgun hardened and the gauge doc line corrected. |
| 👤 paul | "this round landed well" — approving, ship at will; nits on the index-slice coupling, stringly key, and pub-ness. |
| 📐 xp | Round-1 concerns genuinely resolved; only LOW is the `&[usize]` leaking private indices across a pub boundary; praise for the metric doc comment. |
| 🛡 security-audit | Zero findings; all delta risk hypotheses traced and disproven. |

---

pauldambra:
> [!NOTE]
>
> Multi-perspective review of `6aa2c11a`: correctness and reliability reviewers found no actionable issue in the final grace-state changes. One contracts reviewer was unavailable. A performance reviewer raised broader cache atomicity and admission-snapshot architecture concerns, but these are pre-existing or cross-cutting changes outside this measurement-only PR.

## Verdict: ✅ APPROVE

No new in-scope findings. Hard-limit precedence, scoped/global attribution, gateway exemptions, and documented cache limitations are consistent in the current diff.

---

rust/capture/src/v1/quota_limiter_shim.rs
```diff
@@ -92,27 +91,66 @@ pub async fn apply_quota_limits(
...
+            .collect();
+        limiter
+            .report_grace_period_admission_for_event_infos(
+                &admitted_event_infos,
+                &scoped_in_grace,
+                is_global_grace,
+            )
+            .await;
+        // Gateway-verified events never appear in `admitted_event_infos` above
+        // (they're wallet-billed, not scoped/global-quota eligible), so they're
```
pauldambra: > [!NOTE]

**[convergent: qa-team/data-integrity + qa-team/generalist-a]** 🟢 LOW

This comment says gateway-verified events are "not scoped/global-quota eligible", but the global half is factually wrong: the step-1 check at the top of `apply_quota_limits` short-circuits the whole batch on the global Events quota, gateway-verified events included, and `ai_endpoint.rs` explicitly documents them as still subject to the team's global Events quota. Only the scoped `llm_events` exemption is real — and counting them under the global grace label on the next line is correct precisely *because* they are global-quota eligible. As written, a future reader could take the comment at face value, conclude the `count_global_grace_admission(global_only_event_count)` call is a bug, and remove it, silently dropping gateway-verified traffic from the metric.

The code is right; suggest rewording to match `ai_endpoint.rs`, e.g. "wallet-billed and exempt from the scoped llm_events quota (so they never appear in the scoped partition above), but still subject to the global Events quota, hence counted under the global grace label here".

rust/capture/tests/quota_limiters.rs
```diff
@@ -195,6 +248,81 @@ fn extract_captured_event_names(events: &[ProcessedEvent]) -> Vec<String> {
...
+        .post("/e")
+        .body(create_batch_payload_with_token(&["$exception"], token))
+        .header("Content-Type", "application/json")
+        .header("X-Forwarded-For", "127.0.0.1")
+        .send()
+        .await;
+
+    assert_eq!(response.status(), StatusCode::OK);
+    assert_eq!(sink.events().len(), 1);
+    assert_eq!(grace_period_count(&snapshotter, "events"), None);
```
pauldambra, line 356: > [!NOTE]

**[convergent: qa-team/data-integrity + qa-team/generalist-b]** 🟢 LOW

The decisive assertion of `test_global_limit_takes_precedence_over_global_grace_period` is purely negative (`== None`), which also passes vacuously if the thread-local `DebuggingRecorder` plumbing ever stops observing handler metrics (e.g. someone drops `flavor = "current_thread"` and the server task migrates off the recorder's thread). Today the sibling recordings test proves the recorder mechanism works via its positive `Some(1)` assertion, but that safety net lives in a different test and can be changed or deleted independently.

Consider adding a positive control from this same request — e.g. assert the events-dropped counter incremented for the over-quota drop that fires in `check_and_filter` on exactly this path (verify the actual `cause` label value used there before asserting) — so `None` for the grace counter provably means "suppressed" rather than "unobserved".

rust/capture/src/prometheus.rs
```diff
@@ -6,6 +6,41 @@ use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
...
+///   admitted".
+/// - **Attribution when only the global grace period is active.** Scoped
+///   grace periods (`resource="exceptions"`, `resource="survey_responses"`,
+///   `resource="llm_events"`) take precedence, but when a scoped limiter
+///   isn't itself in grace, events it would have matched fall through and
+///   are counted under the global resource label (`events`/`recordings`)
+///   instead of the scoped one.
+/// - **Hard limits override grace.** A token present in both the hard-limit
+///   and suspended Redis sets (a transient skew while the billing writer
+///   updates them) is treated as limited, not in grace, and is not counted.
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + qa-team/generalist-b]** ⚪ NIT

"is treated as limited, not in grace, and is not counted" is slightly more absolute than the code guarantees: grace and hard-limit membership are read by separate probes over DashMaps refreshed by independent pollers using clear-then-reinsert, so during a refresh window a both-sets token can transiently read as grace=true/limited=false and get counted. Same theme for the "a single snapshot" comment in `quota_limiter_shim.rs:96` and the method doc on `report_grace_period_admission_for_event_infos`: the "snapshot" is two sequential awaits, not an atomic read.

This is metric-only, bounded by the poll interval, and strictly narrower than pre-delta behavior, so no code change is needed — consider softening the wording (e.g. "is not counted (modulo the limiter refresh window)" and "the same values drive the guard and the counting") so nobody later builds an enforcement decision on an atomicity guarantee the code does not provide.

rust/capture/src/prometheus.rs
```diff
@@ -6,6 +6,41 @@ use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
...
+/// - **Fails toward stale counts, not toward silence.** The underlying
+///   Redis reader fails open: if Redis is unreachable past a grace-period
+///   expiry, pods keep counting that org's traffic as grace-admitted using
+///   the last-known state, with only a warn log as the failure signal (the
```
pauldambra, line 42: > [!NOTE]

**[qa-team/data-integrity]** 🟢 LOW

Non-blocking follow-up: the corrected doc now honestly concedes the fail-open path has no machine-readable staleness signal (warn log only, and the loaded-tokens gauge freezes at its last value). Since this counter is meant to inform billing decisions, a pod that misses a grace-period expiry keeps inflating it for that org with nothing alertable — the classic silent-stale-cache pattern. A cheap fetch-error counter (or last-success-timestamp gauge) in the `RedisLimiter` poller's error arm would make staleness alertable for this and the hard limiters alike. That is shared limiter infrastructure used by several services, so a separate PR is reasonable — flagging so it does not get lost.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +263,129 @@ impl CaptureQuotaLimiter {
         Ok(filtered_events)
     }
 
+    pub async fn report_global_grace_period_admission(&self, token: &str, event_count: u64) {
```
pauldambra, line 285: > [!NOTE]

**[qa-team/reliability]** ⚪ NIT

`report_global_grace_period_admission` is now the one remaining probe-at-report-time entry point (used by the gateway-trusted path in `ai_endpoint.rs`), i.e. the residual instance of the guard-vs-count re-probe pattern this delta removed everywhere else. Behavior is correct today because that caller sits directly behind an `is_quota_limited_v1(Events)` check, so the tightened `is_in_global_grace_period` meaning agrees with the guard in front of it. For full consistency the caller could snapshot `is_in_global_grace_period` once and call `count_global_grace_admission` directly; fine to defer.

rust/capture/src/quota_limiters.rs
```diff
@@ -211,6 +263,129 @@ impl CaptureQuotaLimiter {
...
+        }
+    }
+
+    /// True if `token` is in the global grace period for this capture mode's
+    /// resource and not currently hard-limited — a token in both sets (a
+    /// transient skew while the billing writer updates them) counts as
+    /// limited, not in grace. Scoped grace periods are checked separately via
+    /// `scoped_limiters_in_grace_period` — combine both to know whether
+    /// there's any grace-admission work to do for `token`.
+    pub(crate) async fn is_in_global_grace_period(&self, token: &str) -> bool {
```
pauldambra, line 388: > [!NOTE]

**[paul]** ⚪ NIT

nit picking on naming — `is_in_global_grace_period` now quietly means "in grace AND not hard-limited". the doc comment right above spells it out nicely, so i'm not worried, but at the call sites (shim, `report_grace_period_admission`) a reader just sees `let is_global_grace = ...` and won't know it already excludes hard-limited tokens. i wonder if `in_global_grace_and_not_limited` or similar would carry its own weight without needing the reader to remember the doc. not blocking at all, folding the hard-limit exclusion in here is arguably the right home for it.

rust/capture/tests/quota_limiters.rs
```diff
@@ -195,6 +248,81 @@ fn extract_captured_event_names(events: &[ProcessedEvent]) -> Vec<String> {
...
+        })
+}
+
+#[tokio::test(flavor = "current_thread")]
+async fn test_recording_grace_period_counts_admitted_snapshots() {
+    let token = "test_token_recording_grace_period";
+    let (router, sink) =
+        setup_router_with_limits_and_grace(token, CaptureMode::Recordings, false, true, vec![])
+            .await;
+    tokio::time::sleep(Duration::from_millis(50)).await;
```
pauldambra, line 304: > [!NOTE]

**[paul]** ⚪ NIT

love that these are here and they actually exercise the precedence at the integration level (traced it — `$exception` survives the global limit via retained_indices, and grace count stays None, so it's a real test). silly observation: this + the three in-module grace tests share almost identical scaffolding (recorder/snapshotter/guard + 2-event setup) and only differ by which flags are set and the expected counts. `rstest` is already a dev-dep so they'd parameterize cleanly. total nit, ship as you see fit.

pauldambra (commented):
> [!NOTE]

QA Swarm round-3 re-review (delta c3cc558..6aa2c11: hard-limit precedence + round-2 hardening merge) complete. See inline comments and the round-3 summary comment.

pauldambra:
> [!NOTE]
>
> Round-3 re-review of the latest delta (hard-limit precedence fix + round-2 hardening merge, c3cc558..6aa2c11) by qa-team, paul-reviewer, xp-reviewer, security-audit

## Verdict: ✅ APPROVE

All four reviewers independently verified the highest-risk surface of this hand-merged, never-locally-compiled delta and found no compile errors, test breakage, or semantic regressions: every re-signatured call site matches, the separate-crate integration tests touch only `pub` items, all five touched/new tests hand-trace to their asserted counts under the new hard-limit-overrides-grace + snapshot semantics (the new precedence test would genuinely fail against pre-delta code), and the metric docs are factually accurate on the points corrected last round. Remaining findings are comment/doc/test-hardening nits — nothing blocking.

### Key findings

- 🟢 LOW — `v1/quota_limiter_shim.rs:125`: the new gateway-verified comment is factually wrong ("not scoped/global-quota eligible" — the global half is untrue) and could mislead a future reader into deleting a correct counting call; reword (code is right).
- 🟢 LOW — `tests/quota_limiters.rs:323`: the new precedence test's decisive assertion is negative-only; add a positive control so `None` provably means suppressed, not unobserved.
- 🟢 LOW — `prometheus.rs:38`: fail-open staleness has no alertable signal; a fetch-error counter in the shared `RedisLimiter` poller is a reasonable separate PR — flagged so it isn't lost.
- ⚪ NIT — soften three absolute doc claims (refresh-window caveat); the residual probe-at-report-time path in the AI gateway flow; a naming thought on `is_in_global_grace_period`; rstest-parameterizing the grace test scaffolding.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | LOW across all four sub-reviewers; approve — convergent on the wrong shim comment and the negative-only assertion; verified the merged intersection end to end. |
| 👤 paul | "perfetto — stamped, ship it": snapshot threading consistent, hard-limit override symmetric, tests earn their place; two total nits. |
| 📐 xp | "This is fine" — canonical global-grace read routes through one method, per-scoped logic acceptably separate (receivers differ), helpers pull their weight, nothing superfluous. |
| 🛡 security-audit | Zero findings: no amplification (short-circuited DashMap reads), the bounds-safe fallback is provably unreachable, snapshot skew not attacker-steerable, enforcement paths byte-identical. |

---

rust/capture/src/quota_limiters.rs
```diff
@@ -113,13 +151,30 @@ impl CaptureQuotaLimiter {
...
 
+    /// Drops events over quota for `token` and returns the rest.
+    ///
+    /// This only enforces limits — it does not report grace-period
+    /// admissions. Callers on an admission path must separately call
+    /// `report_grace_period_admission` (or the `_for_event_infos` variant)
+    /// with the returned, still-admitted events so the billing-grace
+    /// measurement counter stays accurate; nothing enforces that pairing at
+    /// compile time.
     pub async fn check_and_filter<T: HasEventName>(
```
jose-sequeira, line 172: nit: after every check_and_filter call, we have to manually call the reporting. 

maybe the reporting could be included within the method?

jose-sequeira (approved):
LGTM, just a small nit

pauldambra:
i'm gonna be mindful of how much time i actually have and come back to this at some point....

PR #71010, opened 2026-07-15
chore(frontend): inline all kea logic types

mariusandra:
## Problem

Kea logic types currently live in generated `*LogicType.ts` and `*ModelType.ts` files. That splits a logic from its inferred type and makes a clean typegen run regenerate roughly one thousand separate files.

This PR completes the inline-mode migration started by `typegen-inline` and measures its effect on typegen and TypeScript performance.

## Changes

- Converted all 1,041 discovered Kea logic types to inline `MakeLogicType` declarations.
- Removed generated type-file imports and the remaining tracked generated type files.
- Enabled inline mode in `.kearc` and updated the repo guidance for future logic changes.
- Upgraded to `kea@4.0.0-pre.6` and `kea-typegen@3.8.3`, which include the required metadata and inline import support from [keajs/kea#167](https://github.com/keajs/kea/pull/167) and [keajs/kea-typegen#58](https://github.com/keajs/kea-typegen/pull/58).

Benchmarks used three runs on the same local macOS checkout. Clean typegen deleted generated output before each uncached run. TypeScript used `tsgo --noEmit --incremental false`. Times are wall-clock seconds from `/usr/bin/time -p`.

| Benchmark | Before runs | Before median | After runs | After median | Change |
| --- | --- | ---: | --- | ---: | ---: |
| Clean uncached typegen | 225.25, 205.87, 190.35 | 205.87 | 17.79, 18.87, 18.97 | 18.87 | -90.8% |
| Nonincremental TypeScript | 16.58, 19.80, 15.93 | 16.58 | 21.17, 16.07, 18.24 | 18.24 | +10.0% |

The typegen improvement is clear. The TypeScript measurements do not demonstrate an improvement and remain within the observed run-to-run spread.

- `pnpm --filter=@posthog/frontend typegen:check` using kea-typegen 3.8.3 and kea 4.0.0-pre.6
- `hogli test frontend/src/lib/components/TaxonomicFilter/taxonomicFilterLogic.test.ts`
- `hogli test frontend/src/scenes/cohorts/cohortCountWarningLogic.test.ts`
- `hogli ci:preflight --strict`
- `yarn test --watchman=false` and `yarn build` in `kea-typegen` (51 tests, including the pnpm physical-path regression)
- `pnpm run test:types` and `pnpm run test:tsd` in `kea`
- `pnpm --filter=@posthog/frontend exec tsgo --noEmit --incremental false` reaches only the pre-existing `QuillDateFilter.tsx` error for the removed `DateTimePickerProps.ranges` property. It is not newly green on this branch.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `frontend/src/AGENTS.md` and the repo `writing-kea-logics` skill. No user-facing documentation changes are needed because this only changes generated TypeScript representation and developer workflow.

## 🤖 Agent context

Codex (GPT-5) implemented the migration under human direction. No shareable session URL is available. Repo skills used: `/writing-kea-logics`, `/writing-skills`, `/writing-tests`, and `/running-ci-preflight`.

The required Kea and kea-typegen support was developed and verified in their upstream repositories. I switched the monorepo to the published releases once they became available, leaving no local package patches for this migration.


Gilbert09 (approved):

frontend/src/exporter/ExporterLogin.tsx
```diff
@@ -12,12 +13,85 @@ import { LemonInput } from 'lib/lemon-ui/LemonInput'
 import { ERROR_MESSAGES } from 'scenes/authentication/shared/loginErrorMessages'
 import { SupportModalButton } from 'scenes/authentication/shared/SupportModalButton'
 
-import type { loginLogicType } from './ExporterLoginType'
-
 export interface LoginForm {
     password: string
 }
 
+// Generated by kea-typegen. Update if you're an agent, ignore if you're human.
```
gantoine, line 20: are the block actually generated and injected in this file, or "generated" in the sense that an agent will write/update them? if the former they should probably be placed in their own type files to prevent conflicts, and if the latter i would remove the "generated" language.

gantoine (changes requested):

frontend/src/exporter/ExporterLogin.tsx
mariusandra: they're both. They're autogenerated, like the current typegen does, but the expectation is that when an agent does work around this logic, it'll also update the file here. We would still run typegen at the end of it all in CI, but this time it would be more of a "typecheck". The goal/expectation is that 90% of the time when adding a thing or two to a logic, agents will correctly also update the type here (cause there really isn't much to do here for most things), saving a whole extra CI run in the process.

mariusandra: also, i'd avoid separate files cause this just adds bloat to the codebase. the current generated *Logic.ts files are extremely verbose (duplicate things a lot). we could commit the MakeLogicType stuff to a different file as well, but since all logics are anyway large containers, it doesn't feel like a few extra lines of generated metadata at the top will impact the experience for anyone

gantoine (approved):

mariusandra:
Superseded by the smaller, parallel migration series:

- Foundation: #71446
- Batches: #71455, #71456, #71457, #71458, #71459, #71460, #71461, #71462, #71463

The batch PRs target `master` independently and use reserved `.kearc` slots, so sibling merges do not require restacking.

mariusandra:
Closing in favor of the smaller migration series above.

PR #71015, opened 2026-07-15
feat(aio): batch count-triggered eval report checks per team

carlos-marchal-ph:
## Problem

The count-triggered eval report coordinator fires one ClickHouse `count()` scan **per report** every 5 minutes, scaling linearly with the number of count-triggered reports and adding to `ClickHouseAtCapacity` pressure.

> [!NOTE]
> Stacked on top of #70998 (offline routing). Review/merge that first. This PR's diff is only the batching change.

## Changes

Share one count query per team instead of one per report. Each report becomes a `countIf(properties.$ai_evaluation_id = {evaluation_id} AND timestamp >= {since})` column in a single per-team query, so every count is **identical** to what the single-report query returned — the shared `WHERE` (an `IN` set + `min(since)`) only narrows the scan and never changes a count.

- Postgres gating (deliverability, cooldown, daily cap) and the `count >= threshold` decision are unchanged; only the count query is shared.
- Candidates are fetched ordered by team so a batch keeps a team's reports together.
- The per-query width is capped so a team with many reports splits across a few queries rather than one enormous one.
- The workflow selects the batched path behind `workflow.patched()` so a coordinator already in flight when this deploys replays its original per-report command sequence and doesn't hit a nondeterminism error. The legacy per-report activity stays registered for those runs and can be removed once they drain.

> [!NOTE]
> Incremental/watermark counting was deliberately **not** used: it would drop late-arriving events (the full re-scan picks them up), silently changing which reports fire and when. Batching keeps every count an exact re-scan.

Automated tests added:
- ClickHouse-backed cases (real `$ai_evaluation` events, no query mocking) asserting per-report `since` windows, evaluation isolation, threshold, per-team scoping, and width chunking.
- A Postgres-only case asserting gated reports are resolved with **no** ClickHouse query.
- A workflow case asserting cross-batch aggregation and per-batch failure isolation.

What I (the agent) actually ran here: `ruff check`/`ruff format` (clean), the no-DB workflow tests (pass, including the new batched-aggregation test), and the module-integrity checks (pass — they guard the activity/workflow registration edits). The ClickHouse- and Postgres-backed tests could **not** run in this sandbox (no Postgres/ClickHouse services); they use the existing `ClickhouseTestMixin` + `_create_event` harness and will run in CI.

## Docs update

No user-facing behavior change.

## 🤖 Agent context

Second step of the coordinator capacity work from a Slack thread; the first step (offline routing) is #70998. Incremental counting was considered and rejected for the correctness reason above. Invoked the repo's `/writing-tests` skill before adding tests.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B8ZE81ZT7/p1784102271694409?thread_ts=1784102271.694409&cid=C0B8ZE81ZT7)*


posthog/temporal/ai_observability/eval_reports/activities.py
```diff
@@ -132,30 +156,148 @@ def _check_count_triggered_eval_report_sync(
...
+    now = now or dt.datetime.now(tz=dt.UTC)
+
+    reports = {
+        str(report.id): report
+        for report in EvaluationReport.objects.deliverable()
+        .filter(
+            id__in=report_ids,
+            frequency=EvaluationReport.Frequency.EVERY_N,
+            trigger_threshold__isnull=False,
+            evaluation__output_type="boolean",
```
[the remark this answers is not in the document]
Radu-Raicea: Agreed, needs to be fixed

posthog/temporal/ai_observability/eval_reports/workflow.py
```diff
@@ -190,6 +199,63 @@ async def _check_count_triggered_eval_report_candidates(report_ids: list[str]) -
...
+    failed: list[tuple[str, str]] = []
+    skipped_counts = {
+        "cooldown": 0,
+        "daily_cap": 0,
+        "not_deliverable": 0,
+        # A team's count query failed inside the batch; the report is re-checked next tick.
+        "count_query_error": 0,
+    }
+
+    batches = _batch_report_ids(report_ids, COUNT_TRIGGER_REPORTS_PER_ACTIVITY)
```
Radu-Raicea: Before, we ran at most 5 queries at once, and each query had 60 seconds. Now, one activity may run up to 200 queries one after another, but the whole activity still has only 60 seconds. We also start all activities at once. Should we make one activity per team and limit how many activities run at the same time?

posthog/temporal/ai_observability/eval_reports/activities.py
```diff
@@ -132,30 +156,148 @@ def _check_count_triggered_eval_report_sync(
...
+                        (
+                            report_id,
+                            str(report.evaluation_id),
+                            since,
+                            get_outcome_definition(report.evaluation.output_type).event_predicate,
+                        )
+                        for report_id, report, since in chunk
+                    ],
+                )
+            except Exception:
```
Radu-Raicea: If we catch here we won't be using the Temporal retries for the activity

Radu-Raicea (changes requested):

posthog/temporal/ai_observability/eval_reports/activities.py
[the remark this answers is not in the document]
carlos-marchal-ph: Fixed and added a test for it

posthog/temporal/ai_observability/eval_reports/workflow.py
carlos-marchal-ph: Yeah, did exactly that!

posthog/temporal/ai_observability/eval_reports/activities.py
carlos-marchal-ph: Yep, now it's a bit different too since the queries are team scoped as per the feedback above, so should be easier to handle with Temporal's lifecycles

posthog/temporal/ai_observability/eval_reports/activities.py
```diff
@@ -199,6 +333,66 @@ def _count_eval_results_for_report(report: "EvaluationReport", since: dt.datetim
...
+
+    # evaluation_id and since go in as ast.Constant placeholders (no interpolation to audit);
+    # `since` stays a datetime so HogQL prints toDateTime64(..., 6, <team_tz>) — a bare string
+    # would shift by the team's offset. event_predicate is a trusted internal output-type
+    # definition (never user input), interpolated to match the single-report query exactly.
+    # Columns are read positionally below, so no aliases are needed.
+    select_columns: list[ast.Expr] = [
+        # nosemgrep: hogql-fstring-audit (the predicate comes from fixed internal output-type definitions)
+        parse_expr(
+            f"countIf(properties.$ai_evaluation_id = {{evaluation_id}} AND {event_predicate} AND timestamp >= {{since}})",
```
Radu-Raicea: If we change the target from generation to trace, we might get to the threshold needed for the report too fast.

For example, let's say 80 generation evals have run, and then we switch to target traces, and only 20 are run, it would still hit the 100 threshold for the report, where we probably want to wait until we get 100 traces evals.

We should probably add the target to this `countIf`.

Radu-Raicea (approved):

PR #71018, opened 2026-07-15
fix(frontend): Keep hash params across the / homepage redirect

Twixes:
## Problem

`app.posthog.com/#panel=max:<prompt>` opens PostHog AI pre-filled with the prompt. It broke on the Home scene (`/`) — the prompt was silently dropped — while still working elsewhere. And even once the hash survives, opening a Max *side panel* on top of a homepage that already IS the full-scene PostHog AI is redundant. Came up in a [Slack thread](https://posthog.slack.com/archives/C08499A7REU/p1783548393086039?thread_ts=1783548393.086039&cid=C08499A7REU).

Cause: `/` redirects to the homepage and that redirect dropped the URL hash (other scenes have no such redirect). And the side panel opens from the hash regardless of whether the current scene is already full-scene AI. (This is the current live implementation, separate from the in-progress port to the new PostHog AI.)

## Changes

- `sceneLogic`: the `/` → homepage redirect now forwards the incoming hash (via a shared `withForwardedHashAndSearchParams` helper), so the prompt reaches the destination.
- `aiFirstHomepageLogic`: when the homepage is the full-scene PostHog AI, it intercepts `#panel=max:<prompt>`, prefills its own composer, enters AI mode, and closes the redundant side panel (reusing `parseCommandString` from `maxLogic`).

Added logic tests, each confirmed to fail without its fix:
- `sceneLogic`: the `#panel` hash survives the `/` redirect (with and without a configured homepage).
- `aiFirstHomepageLogic`: a `#panel=max:<prompt>` navigation prefills the full-scene chat and leaves the side panel closed.

Ran the affected suites locally (`sceneLogic`, `maxLogic`, `sidePanelStateLogic`, `aiFirstHomepageLogic`) — all green. No manual browser testing.

## 🤖 Agent context

Authored by the PostHog Slack app agent, directed by Michael Matloka. Invoked `/writing-tests` and `/simplify`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08499A7REU/p1783548393086039?thread_ts=1783548393.086039&cid=C08499A7REU)*


yasen-posthog (approved):
Sorrryyyy

PR #71023, opened 2026-07-15
feat(agent-platform): add native Slack agent messaging experience

benjackwhite:
## Problem

Connecting an agent-platform agent to Slack is a bring-your-own-app flow, and once connected the bot shows up as a plain Slack bot. Slack's June-2026 "Agent messaging experience" lets an app appear as a native, DM-able agent — with suggested starter prompts and a real "thinking" indicator that auto-clears on reply — and we weren't taking advantage of any of it. Progress was faked with a normal message that gets posted, edited, and deleted.

**Why:** we want agents to use Slack's native agent tooling and show up as an agent you can talk to, not just a bot. This is milestone 1 of that effort; a follow-up will auto-create and manage the Slack app itself so users don't have to hand-build it.

## Changes

Adds an opt-in `agent_surface` flag to the `slack` trigger config. When enabled:

- **Manifest** (`slack-manifest.ts`): emits `features.agent_view` (+ `agent_description` and fixed `suggested_prompts`), the `assistant:write` scope, the App Home Messages tab, and the `assistant_thread_started` / `app_home_opened` bot events.
- **Ingress** (`slack.ts`): handles the new lifecycle events — greets a new conversation with a welcome message and pushes suggested prompts via `assistant.threads.setSuggestedPrompts`. The agent surface is DM-first, so the flag enables DMs on its own.
- **Runner** (`slack-reply.ts` + `driver.ts`): reports progress with the native `assistant.threads.setStatus` indicator (auto-clears when the reply lands) instead of the simulated post/edit/delete status message.

Everything is gated behind `agent_surface` (default `false`), so existing bring-your-own Slack apps are byte-for-byte unaffected (verified by a back-compat manifest test). Suggested prompts / `agent_description` / `welcome_message` are new optional fields on the same trigger config.

The agent **did not run these tests locally** — this was done in a sparse, dependency-free checkout where the JS workspace can't be installed. The tests below were added following existing patterns and are expected to run in CI:

- `slack-manifest.test.ts` — `agent_surface=true` emits `agent_view` / `assistant:write` / the new events / App Home; `agent_surface=false` stays byte-identical to today (back-compat guard); empty `agent_view` when no description/prompts.
- `slack-reply.test.ts` — native mode calls `assistant.threads.setStatus` and `clear()` is a no-op; simulated mode keeps post/update/delete; `toPlainStatus` strips emoji + markdown. Catches a regression where native status would fall back to posting/deleting a message.
- `slack-trigger.test.ts` (e2e) — `assistant_thread_started` posts welcome + prompts; `app_home_opened` (Messages tab) refreshes prompts and is ignored on the Home tab; a turn under `agent_surface` uses native `setStatus` (no simulated status message, no delete) while the reply still lands; a DM enqueues under `agent_surface` without `allow_direct_messages`. This is also the wire test for the spec → native-status plumbing.

## Docs update

No user-facing docs under `products/agent_platform/docs/` describe Slack setup today; the trigger's own JSDoc (the authored contract) was extended in `spec.ts`.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — directed by Ben White.

- Tool: PostHog Code (Claude Opus 4.8). Work started as an investigation into improving the Slack configuration and using Slack's native agent tooling, then implemented the first milestone.
- No mandatory repo skills were triggered — the change is node-side (zod spec, manifest generator, ingress/runner TypeScript); it touches no Django serializer/viewset, migration, or generated API types, so `hogli build:openapi` should be a no-op (worth confirming in CI).
- Key decision: the manifest uses `features.agent_view` (not the deprecated `assistant_view`), and suggested prompts are both set on the manifest (fixed) and pushed at runtime — matching Slack's current Agent-experience docs.

---

### DIFF
--- a/products/agent_platform/services/agent-ingress/src/triggers/slack.ts
+++ b/products/agent_platform/services/agent-ingress/src/triggers/slack.ts
@@ -39,6 +39,10 @@ import type { RouteCtx, TriggerDeps, TriggerModule } from './types'
 // enforcement point, but the helper stays available on the package surface.
 export { verifySlackSignature }
 
+/** Fallback greeting for a new agent conversation when the spec sets no
+ *  `welcome_message`. */
+const DEFAULT_AGENT_WELCOME = "Hi! I'm your PostHog agent — ask me anything and I'll get to work."
+
 async function slackEventsHandler(ctx: RouteCtx): Promise<void> {
     const { req, res, deps, resolved } = ctx
     // Signature already verified by the slack_signing guard.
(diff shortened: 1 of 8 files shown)


benjackwhite:
Superseded by #71076, which contains this work plus the follow-up making the native Slack agent surface the standard experience (rather than an opt-in flag). The signed-commit tooling required a fresh branch, so the full feature lives on `posthog-code/slack-agent-surface-standard`.

PR #71026, opened 2026-07-15
feat(self-driving): Fix scout linking on report

Twixes:
## ~~Problem~~

~~The inbox report detail scene names which scout authored a report, but only as plain text. There was no way to jump from a report to the scout that created it, even though we have a scout detail page, as experienced by @pauldambra.~~

## ~~Changes~~

~~In the report detail meta line, the scout name on a scout-authored report now links through to the scout's detail page.~~

Actually, someone already _just_ did this in another PR, so just fixing the way we detect the scout here.

### DIFF
--- a/frontend/src/scenes/inbox/components/detail/ReportDetail.tsx
+++ b/frontend/src/scenes/inbox/components/detail/ReportDetail.tsx
@@ -158,9 +158,10 @@ function MetaSourceStack({
         return null
     }
     // Name the authoring scout on a scout-authored report so it's clear at a glance who wrote it,
-    // and link the name straight to the scout's detail page.
+    // and link the name straight to the scout's detail page. The scout may not sort first among mixed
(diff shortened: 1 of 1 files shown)


frontend/src/scenes/inbox/components/detail/ReportDetail.tsx
```diff
@@ -119,7 +119,13 @@ function ReportDetailMeta({
         </span>
     )
     if (hasSource) {
-        stats.push(<MetaSourceStack sourceProducts={report.source_products} scoutName={scoutName} />)
+        stats.push(
+            <MetaSourceStack
+                sourceProducts={report.source_products}
+                scoutName={scoutName}
+                scoutSlug={report.scout_name}
```
Twixes: This looks weird, Claude! I think to disambiguate we should rename `scoutName` to `displayScoutName`

webjunkie (approved):

PR #70364, opened 2026-07-13
feat(error-tracking): use fingerprints in issue URLs

ablaszkiewicz:
## Changes

Fingerprint is now the primary thing we resolve issues by. It used to be `error_tracking/<issue_id>`. Now it's `error_tracking/fingerprint`. Old format still works as a fallback.

Performance SHOULD remain unchanged. In the issues list query there is one additional subquery which fetches canonical fingerprint for each issue but everything is indexed so should be extremely fast. There are no extra roundtrips anywhere.

Canonical fingerprint means the oldest fingerprint associated with that issue. We make sure to use that so URLs remain stable.

Also:
- cleaned up couple of things regarding alerts,

## How did you test this code?

Quickly checked couple of internal things:
- issues list single issue open - ✅ ,
- issues list bulk issues open - ✅,
- activity tab exception - ✅ ,
- web analytics issues panel - ✅

Then checked if the fallback works - URLs with issue ids in them should still work. It works and redirects you to fingerprint based URL ✅.

Then alerts - new format included in all messages ✅

Then MCP - new format included ✅

And finally - check if new URL encoding works. Created this fingerprint - `$uper/strange#fingerprint`. Both UI and MCP work fine with it ✅

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

PR #71031, opened 2026-07-15
feat(error-tracking): add fingerprint redirect URL for stable issue links

ablaszkiewicz:
## Problem

Issue IDs are unstable: when issues get merged, the merged-away id dies and every alert or external link that embedded it goes stale. #70364 tried making fingerprints the primary identifier in all issue URLs, which touched the whole URL surface; this PR takes the narrower approach from that review — keep issue URLs as they are, add one redirect route that outbound links go through.

## Changes

- **New route** `/error_tracking/fingerprint/:fingerprint?timestamp=<ts>` — a small scene that resolves the fingerprint and `router.replace`s to `/error_tracking/<issue_id>?fingerprint=<fp>&timestamp=<ts>`. `timestamp` is forwarded only when the link carries one.
- **New endpoint** `GET /error_tracking/fingerprints/resolve?fingerprint=<value>` returning the fingerprint record (`issue_id`, ...).

Every durable outbound surface now links through the redirect, using the issue's oldest fingerprint (merges keep it). All fall back to the plain issue-id URL when no fingerprint exists:

- **Chat alerts** (Slack/Teams/Discord created/reopened + onboarding Slack) via one shared Hog helper; fingerprint percent-encoded with `encodeURLComponent`.
- **Spiking alerts**: spiking templates use the same link helper as created/reopened. The cymbal change that attaches `fingerprint` and `exception_timestamp` to the spiking event is split into #71526, which deploys first.
- **Alert-created tracker issues**: Linear/GitHub/GitLab templates get an optional hidden `posthog_issue_url` input; existing configured functions keep working unchanged.
- **UI-created tracker issues**: GitHub/GitLab/Jira issue bodies (frontend) and the Linear attachment (backend, new `get_issue_permalink` facade function).
- **Emails** (weekly digest with batch-fetched fingerprints, issue-assigned) and **in-app assignment notifications**.

Drive-by fixes for two pre-existing bugs: the events table and recording inspector passed the fingerprint *positionally* into `urls.errorTrackingIssue(id, params)` (landing in the URL as garbage), and the session-recording Linear integration attached an error-tracking URL instead of the recording URL — `LinearIntegration.create_issue` now takes the attachment URL explicitly.

Old issue-id URLs keep working; only newly created alerts get the new format.

Robot:
- Backend: resolve happy path with a special-character fingerprint, missing-param 400, unknown 404, cross-team 404; Linear attachment URL uses the encoded fingerprint permalink with issue-URL fallback.
- Frontend: kea logic tests for the redirect (timestamp forwarded/omitted) and the single-decode contract for `%XX` fingerprints (kea-router pre-decodes the pathname).
- Node: GitHub and Linear template tests covering `posthog_issue_url` and the issue-id fallback.
- The cymbal emit path is covered in #71526.
- Manual (local dev stack): resolve endpoint over HTTP, and browser through login → `/error_tracking/fingerprint/verify%2F%24trange%23fp` → issue page.
- Not tested against real Slack/Linear/GitHub/GitLab workspaces or real mailboxes — template/unit tests only.

Me - ran some sanity checks.

### Discord alert

image

link is

```
http://localhost:8000/project/1/error_tracking/fingerprint/69e9bd5685f54c1aa209864738fcded473981c7d4f5e1d9dc87cab38ce59d86fb364924ea9bd7b47b25f65b823bf00ca8863165eb8169d684cb3494f3d4039f4?timestamp=2026-07-16T12:38:05.152Z&utm_source=alert&utm_campaign=error_tracking_alert&utm_medium=discord
```

OPENS ALRIGHT!

### Github alert

CleanShot 2026-07-16 at 16 27 23@2x

OPENS ALRIGHT!

### Github external reference

image

image

OPENS ALRIGHT!

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Alert-authoring skill references under `products/error_tracking/skills/` updated in this PR (deep-link shape, tracker sub-template inputs, spiking fallback guard).

## 🤖 Agent context

Authored with Claude Code (skills: /improving-drf-endpoints, /writing-tests). Replaces #70364 per its review discussion: only outbound links change, all through a single redirect route. The redirect is a frontend scene because app routes are SPA-handled. A multi-agent deep review ran on the branch; its one blocking finding (a stale `LinearIntegration.create_issue` test) is fixed here, and the `app-url-manifest.json` regen was scoped down to just the new route so unrelated drift stays out of this PR.

---


frontend/src/scenes/hog-functions/sub-templates/sub-templates.ts
```diff
@@ -10,6 +10,13 @@ import {
     SurveyEventName,
 } from '~/types'
 
+// Deep link used in error tracking alert messages. Routes through the fingerprint redirect page
+// (/error_tracking/fingerprint/<fingerprint>) so links stay valid even after issues are merged.
+// Also used by the onboarding alert setup (onboardingErrorTrackingAlertsLogic) — keep the format
+// in one place so URL changes can't drift between the two surfaces.
```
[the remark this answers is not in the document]
ablaszkiewicz: It already emits both!

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/errorTrackingFingerprintSceneLogic.ts
```diff
@@ -0,0 +1,117 @@
...
+                return
+            }
+            router.actions.replace(
+                urls.errorTrackingIssue(resolvedFingerprint.issue_id, {
+                    fingerprint: props.fingerprint,
+                    // Only forward a timestamp the link actually carried — without one, the issue
+                    // scene defaults to the newest event via its summary query.
+                    timestamp: props.timestamp,
+                })
+            )
```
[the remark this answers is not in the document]
ablaszkiewicz: Good point. Fixing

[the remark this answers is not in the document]
ablaszkiewicz: Fixed. Now it carries to the next scene

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/errorTrackingFingerprintSceneLogic.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    loaders(({ props }) => ({
+        resolvedFingerprint: [
+            null as ErrorTrackingFingerprintApi | null,
+            {
+                resolveFingerprint: async (): Promise<ErrorTrackingFingerprintApi | null> => {
+                    const teamId = teamLogic.values.currentTeamId
+                    if (!teamId || !props.fingerprint) {
+                        return null
+                    }
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed. We now have a subscription on `currentTeamId`

frontend/src/scenes/hog-functions/sub-templates/sub-templates.ts
```diff
@@ -10,6 +10,13 @@ import {
...
+export const errorTrackingIssueLinkHogTemplate = (medium: string): string =>
+    `{project.url}/error_tracking/fingerprint/{encodeURLComponent(event.properties.fingerprint)}?timestamp={event.properties.exception_timestamp}&utm_source=alert&utm_campaign=error_tracking_alert&utm_medium=${medium}`
```
cat-ph: this skips `)` so it breaks markdown links if there's a `)` 😬 

Image

products/error_tracking/backend/logic/__init__.py
hpouillot, line 275: Don't we need to use attachment url for the other integrations ?

products/error_tracking/backend/notifications.py
```diff
@@ -48,7 +49,8 @@ def dispatch_issue_assigned_realtime(
         issue = assignment.issue
         title = f"{assigner.first_name or assigner.email} assigned an issue to you"[:100]
         body = (issue.name or "")[:200]
-        source_url = f"/project/{team.project_id}/error_tracking/{issue.id}"
+        fingerprint = get_canonical_fingerprint(team_id=assignment.team_id, issue_id=issue.id)
```
hpouillot, line 52: I think I would keep the issue url here, if the issue is merged, the assignment won't propagate so in any case it case it will be confusing.

products/error_tracking/backend/weekly_digest.py
```diff
@@ -320,20 +320,31 @@ def get_new_issues_for_team(team: Team, issue_rows: list | None = None) -> list[
 
 def _build_issues_list(results: list, issues_by_id: dict, team: Team) -> list[dict]:
     """Build issue dicts with sparkline from query results containing issue_id, occurrence_count, daily_counts."""
+    from products.error_tracking.backend.logic import build_issue_permalink_path, list_first_fingerprints
+
+    issue_ids = [issue_id for issue_id, _, _ in results if issue_id]
+    fingerprints_by_issue_id = {
```
hpouillot: nit: maybe `first_fingerpring_by_issue_id` ? I was expecting a list

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/ErrorTrackingFingerprintScene.tsx
```diff
@@ -0,0 +1,36 @@
...
+        ),
+        timestamp,
+    }),
+}
+
+export function ErrorTrackingFingerprintScene(): JSX.Element {
+    const { resolvedFingerprint, resolvedFingerprintLoading } = useValues(errorTrackingFingerprintSceneLogic)
+
+    if (!resolvedFingerprint && !resolvedFingerprintLoading) {
+        return <NotFound object="issue" />
```
hpouillot: Would be cool to add a retry / polling mechanism or explain reasons why it is not found. Event might still be processing when the user clicks the link for instance

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/ErrorTrackingFingerprintScene.tsx
```diff
@@ -0,0 +1,36 @@
...
+} from './errorTrackingFingerprintSceneLogic'
+
+export const scene: SceneExport<ErrorTrackingFingerprintSceneLogicProps> = {
+    component: ErrorTrackingFingerprintScene,
+    logic: errorTrackingFingerprintSceneLogic,
+    paramsToProps: ({ params: { fingerprint }, searchParams: { timestamp } }) => ({
+        // Decode from the raw pathname, not the router's already-decodeURI'd `fingerprint` param —
+        // decoding that a second time crashes on a literal `%` and mangles `%XX` fingerprints.
+        fingerprint: decodeFingerprintSegment(
+            rawFingerprintPathSegment(router.values.location.pathname) ?? fingerprint
```
hpouillot, line 22: Any reason we need this logic ? kea should handle the decoding already ?

hpouillot (approved):
Few comments but otherwise looks good ! It's too bad we cannot use backend redirection but at the same time if we wait for event ingestion might be better to have a scene 🦜

products/error_tracking/backend/notifications.py
ablaszkiewicz: I would argue that it is better to actually see the issue but not assigned to you rather than see a 404. Don't you think?

products/error_tracking/backend/logic/__init__.py
ablaszkiewicz: Nah. Linear is special because it supports native link attachment - that's why we're using `attachment_url`. Other platforms receive fingerprint-based link in the body. I ran sanity checks before opening this PR - there's a screenshot and a body in description which confirms that this works as expected

products/error_tracking/backend/notifications.py
hpouillot: Hum, yes it's probably better, we could also maybe transfer assignment on merge if there is none on the target.

cat-ph (approved):

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/ErrorTrackingFingerprintScene.tsx
cat-ph: I was also thinking about this yeah 🤔

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/ErrorTrackingFingerprintScene.tsx
ablaszkiewicz: default decoding doesn't decode all chars. It leaves some still decoded. We could run another decoding but then we would over-decode.

That's why we run manual decoding here to do just one proper pass.

Example where kea doesn't decode

`payment%2Ffailed` -----kea------> `payment%2Ffailed`
`payment%2Ffailed` ---manual----> `payment/failed`

This happens because default decode does `decodeURI` and not `decodeURIComponent`

products/error_tracking/frontend/scenes/ErrorTrackingFingerprintScene/ErrorTrackingFingerprintScene.tsx
ablaszkiewicz: CleanShot 2026-07-17 at 15 52 48

hpouillot: Perfect

products/error_tracking/backend/weekly_digest.py
ablaszkiewicz: Done

ablaszkiewicz: Agree. Done

frontend/src/scenes/hog-functions/sub-templates/sub-templates.ts
ablaszkiewicz: Fixed

PR #71032, opened 2026-07-15
feat(experiments): optimistic concurrency for experiment updates

### LINKED ISSUE #70852: Bug report: experiments - stale full-array save silently drops metrics
## Please describe
Saving an experiment from a tab with stale state silently deletes metrics that other users added after the tab loaded. The experiment PATCH replaces the full `metrics` / `metrics_secondary` arrays with whatever the client sends (`removeMetric` in `frontend/src/scenes/experiments/experimentLogic.tsx` filters the in-memory array and PATCHes the result), and the server applies it last-write-wins: `update_experiment` in `products/experiments/backend/experiment_service.py` setattrs each field with no version or `updated_at` check, and `_sync_ordering_with_metric_changes` treats every UUID missing from the submitted array as an intentional deletion, purging it from the ordering too. Expected: a stale write gets rejected or at least warned about (notebooks have a conflict warning for this; the skills API does compare-and-swap), not applied as silent data loss.

Observed on Cloud EU on 2026-07-08: a user's experiment tab was loaded a day earlier, when the experiment had 3 secondary metrics. Teammates then added 4 more. The next day the user deleted one of the 3 metrics visible in their stale tab, and the resulting full-array PATCH removed all 4 metrics they had never seen in a single write. The activity log shows a 7 to 2 metrics change nobody intended, which is confusing for customers trying to audit what happened (their first read of it was "PostHog deleted our metrics").

Related: draft PR #70388 serializes metric saves client-side, fixing the same-tab rapid-click variant of this (concurrent PATCHes computed from one snapshot). Per that PR's own notes it does not cover cross-tab or stale-client writes, which need the server-side check; filing this so that follow-up doesn't get lost.

## How to reproduce

1. Open an experiment with secondary metrics in tab A.
2. In tab B, add one or more secondary metrics to the same experiment.
3. Back in tab A (without reloading), delete any one secondary metric.
4. Reload: the metrics added in tab B are gone, removed by tab A's save.

## Additional context

From: https://posthoghelp.zendesk.com/agent/tickets/62318 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/62950)

### Debug info

```shell
Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019f61e2-91cb-79fa-b21e-0e212882da6d?t=1845
Admin (internal): http://go/adminOrgEU/019aa6ad-708b-0000-e868-4cbdabbb22a2 (project ID 110349)
```

mp-hog:
## Problem

Saving an experiment from a tab with stale state silently deletes metrics other users added after the tab loaded. The frontend PATCHes the full `metrics` / `metrics_secondary` / `saved_metrics_ids` arrays computed from in-memory state, and the server applies them last-write-wins with no version check, the ordering-sync then treats every uuid missing from the stale array as an intentional deletion. This can lead to situations like a stale tab's single-metric delete removed four secondary metrics the user had never seen, and the activity log showed a "7 → 2 metrics" change nobody intended.

Closes #70852

## Changes

Follows the established in-repo optimistic-concurrency patterns (feature flags' version + 3-way merge, skills/prompts/notebooks' version CAS returning 409), adapted per-metric:

- New `Experiment.version` integer (nullable, additive migration), bumped on every update and on lifecycle saves (launch, end, reset, archive, ship variant). Excluded from activity-log diffs.
- `ExperimentSerializer` accepts two optional write fields: `version` (last-read token) and `original_experiment` (the base metric collections). Omitting them keeps today's behavior, so API/MCP/scripted callers are unaffected.
- `update_experiment` resolves stale writes under a `select_for_update` row lock. The rule: a stale write that only touches the metric collections merges per metric uuid; anything else 409s.
    - **Merges cleanly:** concurrent add / delete / edit of *different* metrics (inline or shared).
    - **Returns a structured 409** (`detail`, `current_version`, `conflicting_fields` / `conflicting_metric_uuids`): a same-metric double edit, an edit-vs-delete on the same metric, or a stale write that itself carries scalar fields (e.g. a description edit written against metrics someone changed meanwhile).
    - **Never blocked by the other side's scalar changes:** a PATCH that omits a field cannot clobber it, so e.g. a concurrent launch does not fail a stale metric edit.
    - **Server-authoritative reconciliation:** ordering arrays are rebuilt server-side, never trusted from the stale client; metric `fingerprint` keys are stripped from comparisons so launch-time fingerprint recompute never reads as a concurrent edit.
- Frontend sends the handshake on all three experiment write paths (the `updateExperiment` loader and the two raw shared-metric `api.update` calls), sourcing the base from the existing `unmodifiedExperiment` snapshot. On 409 it shows the server's error toast and reloads fresh server state, re-applying the rejected scalar fields locally so the user's edit stays visible for review and retry.
- Post-deploy monitoring: every stale-write resolution emits an **"experiment update concurrency"** analytics event with `resolution` (`merged` / `rejected`), `conflicting_fields` / `conflicting_metric_uuids_count`, `versions_behind`, and `base_snapshot_sent` — chart it broken down by `resolution` to watch the merge/409 ratio after rollout. The pre-existing "experiment updated" event (with `changed_fields`) serves as the volume baseline.

The incident case now merges with **no error and no data loss**: the stale tab's delete applies and the concurrently added metrics survive.

stale-metric-update-toast

- Locally in 2 tabs
- `TestExperimentConcurrency` (17 API tests, `test_presentation_api.py`): each maps to a concurrency scenario no existing test covered — the incident replay (stale delete + concurrent adds → merged, nothing lost), concurrent different-metric adds, same-metric double-edit → 409, edit-vs-delete both directions → 409, concurrent scalar edits (description/stats_config) not blocking stale metric writes, a stale scalar write → 409 with `conflicting_fields`, stale reorder keeping relative order while appending concurrent additions, shared-metric link removal vs concurrent link, link-metadata double-edit → 409, no-version back-compat (last-write-wins preserved), null-version-as-zero, sequential version increments, launch fingerprint churn not producing phantom conflicts, stale metric addition merging past a concurrent launch, and version-without-base strict CAS.
- `TestConcurrentMetricMerge` (5 pure-function tests, `test_experiment_service.py`): pins merge branches awkward to reach via the API — identical edits on both sides, double deletes, fingerprint-only churn, edit-vs-delete.
- Jest (`experimentLogic.test.ts`): payload decoration on the loader and raw write paths, and 409 → toast + fresh reload preserving the rejected scalar edit.
- Full existing suites pass on current master: 739 backend tests (`test_presentation_api.py` + `test_experiment_service.py`), 90 jest tests. `makemigrations --check` clean; migration is a plain nullable `AddField`.

## 🤖 Agent context

- Authored with PostHog Code (Claude). Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`.
- Design was researched against existing in-repo precedents and agreed with the requester: metric collections merge per uuid (each metric's server-assigned uuid preserves intent); a stale write carrying scalar fields conflicts because that edit was likely written against context that has since changed; the other side's scalar changes never block a metric-only write since a PATCH that omits a field cannot clobber it.
- Known follow-ups left out of scope: guarding `ExperimentSavedMetricService.update_saved_metric` (shared-metric row edits propagate to all experiments), consolidating the raw `api.update` paths with #70388's client-side queue, and richer activity-log descriptions for merged writes.

---


jurajmajerik (approved):
Very cool!

Issue #64179, opened 2026-06-17
Feature request: workflows - surface email bounce info


slshults:
### Feature request

Title:
Feature request: workflows - surface email bounce info
## Is your feature request related to a problem?

When a Workflow email campaign reports a number of bounced emails in its metrics, there is no way in the product to see which recipient addresses bounced, or why. 

The "bounced" figure is an aggregate counter only: bounces are not emitted as queryable events, they do not set a person property, and the workflow execution logs sends, but not bounces. The per-bounce detail the email provider returns (the recipient address, bounce type, SMTP status code, and the diagnostic message from the recipient's mail server) is used to increment the counter and then discarded, so it is not stored or surfaced anywhere a user can reach. 

Permanent (hard) bounces are auto-added to the Opt-outs list, but that list is team-wide, commingles hard bounces with explicit unsubscribes, is not scoped to a specific campaign or batch run, and does not include soft (transient) bounces or greylisting bounces. As a result, anyone trying to reconcile bounce reporting, debug why specific addresses bounced, or build their own suppression and reporting cannot retrieve the bounced addresses or bounce reasons for a given campaign.
## Describe the solution you'd like

Surface per-recipient bounce outcomes for a workflow email campaign (per batch run) somewhere, events, and somewhere in the UI ( perhaps drilling down by clicking on the `Bounced` tile of the `Campaign email` view of a workflow's `/metrics` page.) For each bounced recipient, show the address, whether it was a hard (permanent) or soft (transient) bounce, and the bounce detail sent back by the recipient's receiving mail server: the SMTP status / bounce code, the bounce subtype, and the diagnostic message. That would allow customers to reconcile the bounced count against specific addresses, see why each one bounced, and feed hard bounces into their own suppression and reporting.
## Describe alternatives you've considered

- Workflows > Opt-outs list: only contains hard bounces, is team-wide rather than per-campaign, and mixes auto-suppressed bounces with explicit unsubscribes, so it cannot be used to reconcile a single campaign's bounces.
- Workflow execution logs: record "Email sent to" lines per recipient, but contain no bounce entries.
-  Querying via SQL: PostHog emits no bounce event or person property for workflow email bounces (the metric is an aggregate counter only), and there is no customer-accessible bounce table, so the per-recipient data cannot be pulled.

None of these surface the per-recipient bounce list, or the bounce reasons.

## Additional context

[...]

PR #71038, opened 2026-07-15
feat(workflows): suppress repeatedly-bouncing email addresses

meikelmosby:
### Problem

Workflow emails (e.g. the Error tracking weekly digest) keep getting sent to addresses that can't receive mail. Today only permanent bounces are acted on — via the opt-out preferences table, which conflates "user chose not to receive" with "address can't receive". Soft/transient bounces are logged but never suppressed, so a persistently-dead address gets retried on every send. There was also no per-team list a user could see or edit.

Refs #64179

### Changes

A per-team email suppression list — model, REST API, UI, and a unified bounce-driven write path — behind two env flags.

#### Model + migration + API + UI
- New MessageSuppression table (fail-closed team-scoped) with source distinguishing bounce-driven from manual entries, plus a transient_bounce_count, last_bounce_diagnostic, and suppressed/suppressed_at for provenance.
- messaging_suppressions REST API (list / add / remove) and a Suppression list tab in the Workflows product.
- FKs to the hot posthog_team / posthog_user tables use db_constraint=False so CreateModel takes no lock.

#### Write paths (nodejs) — one service, two entry points:
- Soft bounces (Transient): consecutive-bounce counter per address; auto-suppress once the count crosses EMAIL_SUPPRESSION_TRANSIENT_BOUNCE_THRESHOLD (default 5). A successful delivery resets the counter so a one-off outage never accumulates.
- Hard bounces (Permanent): dual-write — the existing recipientsManager.optOut call is untouched, and a suppression row is written alongside with the SMTP diagnostic and immediate suppressed=true. Manual entries are never overwritten. Phase 1 of consolidating deliverability signals into one list; phase 2 (retire the opt-out write, unify the pre-send check) is a follow-up PR.

#### Enforce path
  - The pre-send hook consults the suppression list before sending. Suppression is treated as a deliverability signal, not a preference, so it applies to transactional emails too — before the transactional opt-out bypass.

#### Two env kill switches, both OFF by default (ships dark):
  - EMAIL_SUPPRESSION_WRITE_ENABLED — record soft/hard bounces and populate the list.
  - EMAIL_SUPPRESSION_ENFORCE_ENABLED — skip sends for suppressed recipients.

Independent so writing can be enabled first to observe what would be suppressed before enforcing.

Known: migration + service in one PR

This PR intentionally keeps everything together for end-to-end review. The migration/service-separation check flags it (a DB migration can't ship with plugin-server code in one PR). Will be split into a migration-first PR and a service PR before merge.

### How did you test this code?

Automated tests, grouped by regression each guards:

  - Parser (helpers/ses.test.ts) — Transient and Permanent bounces surface into distinct output arrays (transientBounceRecipients, hardBounceRecipients) with the SMTP diagnostic. Guards a parser change silently dropping either mirror.
  - Service (email-suppression.service.test.ts) — counter arithmetic below/at threshold, delivery resets the counter, isSuppressed short-circuits when enforce is off, recordHardBounces immediate suppression, MANUAL rows preserved, unsuppressed transient counters escalate on a hard bounce.
  - Enforce wiring (recipient-preferences.service.test.ts) — a suppressed row causes shouldSkipAction to return true for both marketing and transactional emails. Guards a refactor moving the suppression check after the transactional bypass.
  - Write wiring (email-tracking.service.test.ts) — POST /public/m/ses_webhook with a Transient or Permanent bounce writes the expected row in posthog_messagesuppression. Guards the full HTTP → parser → dispatch → DB seam.

### 🤖 Agent context

Authored by the PostHog Slack app from a support thread about recurring transient email bounces. Iterated with Claude (Opus 4.7, 1M context) across several sessions.

  Design decisions along the way:
  - Dedicated suppression model rather than reusing the $all opt-out row, so bounce-driven vs manual entries are distinguishable and each carries its own provenance (diagnostic, timestamp, source).
  - Consecutive-bounce counter with reset-on-delivery for soft bounces, so a one-off SMTP outage doesn't accumulate into a suppression.
  - Two env flags separating "populate the list" from "enforce it" — lets write be enabled first to observe what would be suppressed before behavior changes.
  - Dual-write for hard bounces rather than an immediate cutover from the opt-out path. Rationale: the opt-out table has no provenance column, so historical bounce-driven opt-outs can't be reliably backfilled into suppression; a dark dual-write for a couple of weeks lets us build a clean forward record before phase
  2 retires the opt-out write.
  - Suppression check runs before the transactional opt-out bypass since a hard bounce is deliverability, not a preference.

  Skills invoked: /writing-tests (each round of tests passed through the value gate), /django-migrations (hot-table FK guidance → db_constraint=False), /improving-drf-endpoints (schema annotations on the suppression API viewset).

  Iterated on CI feedback: made the model IDOR-compliant (TeamScopedRootMixin + semgrep), fixed a duplicate kea action, and set db_constraint=False on hot-table FKs to clear the migration-risk block.

Created with PostHog (https://posthog.com?ref=pr) from a Slack thread (https://posthog.slack.com/archives/C06GG249PR6/p1784114029881419)


nodejs/src/cdp/services/messaging/helpers/ses.ts
```diff
@@ -234,8 +234,23 @@ export const formatSesEventLogs = (rec: SesEventRecord): SesEventLogLine[] => {
...
+        // Empty set means no restriction (dev/test); prod is expected to configure the workflow SES
+        // topic ARN so we can reject events forged from an attacker-controlled SNS topic.
+        allowedTopicArns: string[] = []
+    ) {
+        this.allowedTopicArns = new Set(allowedTopicArns.map((arn) => arn.trim()).filter(Boolean))
+    }
 
-    constructor(private trackingCodeSigner: EmailTrackingCodeSigner) {}
+    private isTopicAllowed(topicArn: string | undefined): boolean {
+        if (this.allowedTopicArns.size === 0) {
```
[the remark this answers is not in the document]
meikelmosby: thats okay here it should be a no-op following the pattern that we currently have. closing this gap in a follow up PR

frontend/src/lib/api.ts
```diff
@@ -1964,6 +1965,24 @@ export class ApiRequest {
         return this.environments().current().addPathComponent('messaging_preferences').addPathComponent('add_opt_out')
     }
 
+    public messagingSuppressions(): ApiRequest {
+        return this.environments().current().addPathComponent('messaging_suppressions').addPathComponent('suppressions')
```
mayteio: Was this exercised against a real backend? `routes.py` registers `MessageSuppressionViewSet` under `routers.projects` only (deliberately not the dual-route shim, per the comment there), yielding `/api/projects/:id/messaging_suppressions/...` - but these builders use `environments().current()`, so every call should 404. The sibling `messaging_preferences` paths only work from `environments()` because they use `register_legacy_dual_route`. Options: register the dual route, build these from `projects()`, or drop the handwritten builders for the generated `messagingSuppressions*` functions in `products/messaging/frontend/generated/api.ts`, which already encode the correct URL.

products/messaging/backend/api/message_suppression.py
```diff
@@ -0,0 +1,157 @@
...
+        serializer.is_valid(raise_exception=True)
+
+        identifier = serializer.validated_data["identifier"].strip().lower()
+
+        # Soft-delete and un-suppress. Reset the bounce counter so a previously-dead address that
+        # a user deliberately re-enables starts from a clean slate.
+        updated = (
+            MessageSuppression.objects.for_team(self.team_id)
+            .filter(identifier=identifier)
+            .update(
```
mayteio, line 211: This keeps `source='MANUAL'` on the removed row, and both node upserts preserve `suppressed`/`deleted` `WHEN source = 'MANUAL'` - so once a manual entry is removed, the address can never be auto-suppressed again, even on a '550 user unknown' hard bounce, and the row stays hidden from the UI (the list filters `deleted=false`). Suggest resetting `source` to `BOUNCE` here (the counter is already reset), or narrowing the MANUAL guard in the upserts to rows that are still suppressed.

mayteio (commented):
Took a detailed pass over this. The design is solid - dark launch, separate write/enforce flags, counter-with-reset semantics - but I'd hold the merge on the first three items below.

**Blocking**
1. The frontend client calls `/api/environments/...` but the viewset is registered under `/api/projects/` only, so every Suppression tab request 404s (inline).
2. A manually-suppressed address that is later removed keeps `source='MANUAL'`, and both bounce upserts treat MANUAL as authoritative - that address can never be auto-suppressed again, even on hard bounces (inline).
3. Suppression is only enforced on the workflow-action path; email destination hog functions reach `executeSendEmail` with no suppression check (inline).

**Same round, please**
4. Transient bounce counting is not idempotent under SNS at-least-once redelivery (inline).
5. Editor test sends count toward suppression while metrics/logs gate on `isTest` (inline).
6. The list action declares a bare array response but returns a paginated envelope, so all generated clients are mistyped (inline).
7. A deep link to `/workflows/suppression` with the flag off renders a blank tab area (inline).

**Smaller (inline: 8-10), plus cleanup themes with no inline thread:**
- The handwritten `SuppressionEntry` type and `lib/api.ts` client duplicate the generated Orval client this PR regenerates; using the generated `messagingSuppressions*` functions would also have prevented item 1.
- `recordTransientBounces` / `recordHardBounces` are near-clone 60-line upserts that will drift; consider one parameterized private upsert.
- Every bounce write clears the entire cross-team LazyLoader cache; `markForRefresh` on the touched keys is available.
- `recordDeliveries` writes to the Postgres primary on every successful delivery, though it only matches rows for previously-bounced addresses.
- No index matches the list query (`team + suppressed + deleted ORDER BY -updated_at`); a partial index would serve both the page and the COUNT.
- The env flags are hand-parsed (only `'1'`/`'true'` accepted, values frozen at construction) rather than going through `CdpConfig` like sibling services.
- New env flags + REST API + UI tab ship without a `docs/` update, the serializer fields have no `help_text`, and the new viewset actions have unannotated signatures.

nodejs/src/cdp/services/messaging/recipient-preferences.service.ts
```diff
@@ -28,6 +55,41 @@ export class RecipientPreferencesService {
...
+            ...extractEmailsFromAddressList(emailInputs?.bcc),
+        ]
+
+        if (recipients.length === 0) {
+            return false
+        }
+
+        try {
+            const results = await Promise.all(
+                recipients.map((email) => this.emailSuppressionService.isSuppressed(invocation.teamId, email))
```
mayteio, line 83: This is the only path to `isSuppressed`, and its callers are the workflow-action path (hogflow executor + the hog_function action). The email destination template reaches `emailService.executeSendEmail` via the hog executor without passing `shouldSkipAction`, so suppressed addresses still receive mail from email destination functions with enforcement on - and the `EmailSuppressionService` docstring's 'consulted before every email send' doesn't hold. Consider enforcing inside `executeSendEmail`, where teamId and recipient are both in hand, so every send path shares one choke point.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
```diff
@@ -0,0 +1,314 @@
...
+        })
+        params.push(...identifiers)
+
+        const query = `
+            INSERT INTO posthog_messagesuppression
+                (id, team_id, identifier, source, reason, transient_bounce_count, last_bounce_at,
+                 last_bounce_diagnostic, suppressed, suppressed_at, deleted, created_at, updated_at)
+            VALUES ${valueClauses.join(', ')}
+            ON CONFLICT (team_id, identifier) DO UPDATE SET
+                transient_bounce_count = posthog_messagesuppression.transient_bounce_count + 1,
```
mayteio, line 146: SNS HTTP delivery is at-least-once, and nothing on this path keys off the SES MessageId, so a redelivered Bounce notification re-runs this increment: one real bounce can count twice, and an address at threshold-1 gets suppressed off a duplicate. Worth deduping on messageId (even a short-TTL Redis set), especially since the webhook now awaits these writes serially before returning 200, which widens the retry window.

nodejs/src/cdp/services/messaging/helpers/ses.ts
```diff
@@ -560,13 +621,50 @@ export class SesWebhookHandler {
...
+                hardBounceRecipients.push({ teamId, emailAddresses: emails, diagnostic })
+            }
+
+            // Count soft (Transient) bounces toward suppression. These are recipient-side failures
+            // (server unreachable, mailbox full, greylisting); one is harmless but a persistent run
+            // of them means the address can't receive mail.
+            if (teamId && codeIsTrusted && rec.eventType === 'Bounce' && rec.bounce.bounceType === 'Transient') {
+                const emails = rec.bounce.bouncedRecipients.map((r) => r.emailAddress)
+                const diagnostic = rec.bounce.bouncedRecipients.find((r) => r.diagnosticCode)?.diagnosticCode
+                transientBounceRecipients.push({ teamId, emailAddresses: emails, diagnostic })
```
mayteio, line 652: Metrics and logs gate on `isTest` above, but these pushes don't - repeated 'Run test' sends to an address that is greylisting accumulate real `transient_bounce_count` and can suppress a production address from editor test runs. The pre-existing opt-out push was deliberately ungated for hard bounces, but extending that to threshold counting changes the blast radius; suggest gating the transient counting (or all three pushes) on `!isTest`.

products/messaging/backend/api/message_suppression.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+
+    scope_object = "hog_flow"
+    # Custom actions must declare their write status so TeamAndOrgViewSetMixin's AccessControlPermission
+    # checks hog_flow:write on the mutating endpoints; the default 'suppressions' list stays a read.
+    scope_object_write_actions = ["add_suppression", "remove_suppression"]
+    serializer_class = _FallbackSerializer
+
+    @extend_schema(
+        responses={200: MessageSuppressionSerializer(many=True)},
```
mayteio: The action returns `paginator.get_paginated_response(...)` (`{count, next, previous, results}`), but this schema declares a bare array - so the generated `messagingSuppressionsSuppressionsList` is typed `Promise<MessageSuppressionApi[]>` with no page/page_size params (same in the MCP generated client). Any consumer of the generated client gets the envelope where it expects rows. `ListModelMixin` + `pagination_class` (as sibling viewsets do) would make drf-spectacular emit the paginated schema for free, or annotate the paginated response explicitly.

products/workflows/frontend/WorkflowsScene.tsx
```diff
@@ -27,12 +27,13 @@ import { AccessControlLevel, AccessControlResourceType, Breadcrumb } from '~/typ
...
 import { optOutCategoriesLogic } from './OptOuts/optOutCategoriesLogic'
 import { OptOutScene } from './OptOuts/OptOutScene'
+import { SuppressionScene } from './Suppression/SuppressionScene'
 import { MessageTemplatesTable } from './TemplateLibrary/MessageTemplatesTable'
 import { newWorkflowLogic } from './Workflows/newWorkflowLogic'
 import { NewWorkflowModal } from './Workflows/NewWorkflowModal'
 import { WorkflowsTable } from './Workflows/WorkflowsTable'
 
-const WORKFLOW_SCENE_TABS = ['workflows', 'library', 'channels', 'opt-outs'] as const
+const WORKFLOW_SCENE_TABS = ['workflows', 'library', 'channels', 'opt-outs', 'suppression'] as const
```
mayteio, line 37: 'suppression' is accepted by the urlToAction guard unconditionally, but the tab entry only exists when `WORKFLOWS_SUPPRESSION_LIST` is on - a flag-off user opening `/workflows/suppression` gets `activeKey='suppression'` with no matching tab: nothing highlighted and an empty content area, instead of the previous fallback to 'workflows'. Suggest deriving the valid-tab list from the same flag, or falling back when the tab isn't rendered.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
```diff
@@ -0,0 +1,314 @@
...
+            loader: async (keys) => await this.loadSuppressed(keys),
+        })
+    }
+
+    public clear(): void {
+        this.lazyLoader.clear()
+    }
+
+    /** Pre-send check: is this recipient currently on the team's suppression list? */
+    public async isSuppressed(teamId: number, email: string): Promise<boolean> {
```
mayteio, line 87: Manual add/remove via the Django API never reaches this cache - `clear()` only runs on the node's own bounce writes - and the LazyLoader defaults cache positives and negatives for ~5-6 minutes per pod. With enforcement on, a user who removes an address in the UI and immediately re-runs the workflow still gets the send skipped, with nothing surfaced. The plugin-server pubsub (`publish_message` + `markForRefresh`, as `IntegrationManagerService` does) would close this; if the staleness is acceptable for v1, worth a comment stating the window.

products/workflows/frontend/Suppression/suppressionListLogic.ts
```diff
@@ -0,0 +1,226 @@
...
+            __default: EMPTY_SUPPRESSIONS,
+            loadSuppressions: async (): Promise<PaginatedSuppressions> => {
+                try {
+                    return await api.messaging.getSuppressions(1)
+                } catch {
+                    lemonToast.error('Failed to load suppression list')
+                    return EMPTY_SUPPRESSIONS
+                }
+            },
+            loadNextPage: async (): Promise<PaginatedSuppressions> => {
```
mayteio: The catch here returns the previous data, which kea-loaders treats as success, so `loadNextPageSuccess` still fires and `currentPage` increments even when the fetch failed: the UI shows 'Page 2' over page-1 rows and the next click requests page 3, permanently skipping page 2 (mirrored in `loadPreviousPage`). Letting the loader throw (kea-loaders emits Failure actions) or setting `currentPage` only from the fetched page keeps them in sync.

products/workflows/frontend/Suppression/SuppressionList.tsx
```diff
@@ -0,0 +1,175 @@
...
+                    <label htmlFor="suppression-identifier" className="text-sm font-medium">
+                        Email address
+                    </label>
+                    <LemonInput
+                        id="suppression-identifier"
+                        placeholder="email@example.com"
+                        value={newIdentifier}
+                        onChange={setNewIdentifier}
+                        autoFocus
+                        onPressEnter={() => {
```
mayteio, line 181: This Enter handler doesn't check `addSuppressionLoading` (the footer button does), so a second Enter mid-flight fires a duplicate POST; same for the Remove button (no `loading`/`disabledReason`). The backend `get_or_create` makes duplicates harmless, but per the repo rule on double-submission guards, worth wiring the existing loading selectors into both.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
```diff
@@ -0,0 +1,314 @@
...
+        // newer than the last bounce, so leave the counter alone rather than risk erasing a fresh
+        // bounce. In practice the SES Delivery schema requires a timestamp, so this branch only
+        // guards against a caller that forgets to pass it.
+        const query = `
+            UPDATE posthog_messagesuppression
+            SET transient_bounce_count = 0, updated_at = NOW()
+            WHERE team_id = $1
+              AND identifier = ANY($2)
+              AND source <> 'MANUAL'
+              AND suppressed = false
```
dmarchuk, line 273: Is it intended that a delivery never unsuppresses a transient-suppressed address? 🤔 The reset only touches `suppressed = false` rows, so once an address crosses the soft-bounce threshold it stays suppressed even if it starts delivering again, and only a manual removal clears it. Feels a bit off for something "transient" and it skews the write-phase data toward over-suppression since recovered addresses keep counting as suppressed - which is the data we're using to decide enforcement

products/workflows/frontend/Suppression/SuppressionList.tsx
meikelmosby: Addressed in 3c3c5d8bb7d — added `!addSuppressionLoading` guard to the onPressEnter handler and wired `disabledReason`/`loading` on both the Add-modal button and the per-row Remove button.

products/workflows/frontend/Suppression/suppressionListLogic.ts
meikelmosby: Addressed in 3c3c5d8bb7d — the loaders now rethrow on failure so kea-loaders emits the `*Failure` action instead of `*Success`, and `currentPage` stays put. Added a jest test in `suppressionListLogic.test.ts` asserting the counter doesn't advance/roll back on failed loads.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
meikelmosby: Partly addressed in 3c3c5d8bb7d — added a v1 doc block on the service class calling out the ~5-minute per-process staleness window as expected behavior. Pubsub-based invalidation (mirroring `IntegrationManagerService`) is filed as a follow-up rather than blocking this PR.

products/workflows/frontend/WorkflowsScene.tsx
meikelmosby: Addressed in 3c3c5d8bb7d — connected `featureFlagLogic` and added a fallback in `urlToAction` that flips `possibleTab` back to `'workflows'` when `suppression` is requested with the flag off. Kea logic test in `WorkflowsScene.test.ts` covers both flag-on and flag-off deep-link cases.

products/messaging/backend/api/message_suppression.py
meikelmosby: Addressed in 3c3c5d8bb7d — added a `PaginatedMessageSuppressionSerializer` wrapper and pointed the `@extend_schema` response at it, plus declared the `page` / `page_size` query params. Generated client will pick up the correct envelope shape on the next `hogli build:openapi`.

nodejs/src/cdp/services/messaging/helpers/ses.ts
meikelmosby: Addressed in 3c3c5d8bb7d — the three suppression pushes (`hardBounceRecipients`, `transientBounceRecipients`, `deliveredRecipients`) now gate on `!isTest`. The pre-existing opt-out push stays ungated for backward compat. Parameterized jest test in `helpers/ses.test.ts` covers permanent + transient + delivery test-send cases.

nodejs/src/cdp/services/messaging/recipient-preferences.service.ts
meikelmosby: Addressed in 3c3c5d8bb7d — the pre-send suppression check moved into `EmailService.executeSendEmail`, which every send path (workflow action + hog function email destination) shares. `RecipientPreferencesService` still runs its check for early filtering on the workflow action path, but `executeSendEmail` is now the single choke point that can't be bypassed. Added a jest test asserting SES isn't called when the recipient is suppressed.

products/messaging/backend/api/message_suppression.py
meikelmosby: Addressed in 3c3c5d8bb7d (option a) — `remove_suppression` now resets `source='BOUNCE'` alongside the un-suppress + counter reset, so a subsequent hard/soft bounce can auto-suppress the address again. Python API test in `test_message_suppression.py` asserts the source flips back after removal.

frontend/src/lib/api.ts
meikelmosby: Addressed in 3c3c5d8bb7d — swapped `.environments()` → `.projects()` on the three builders to match the `routers.projects.register` registration in `routes.py`. Minimal-diff fix; migrating to the generated `messagingSuppressions*` client per the `/adopting-generated-api-types` skill is a good follow-up but out of scope here.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
```diff
@@ -0,0 +1,314 @@
...
+    private readonly threshold: number
+    private readonly writeEnabled: boolean
+    private readonly enforceEnabled: boolean
+    private readonly lazyLoader: LazyLoader<boolean>
+
+    constructor(private postgres: PostgresRouter) {
+        this.threshold = resolveThreshold()
+        this.writeEnabled = isWriteEnabled()
+        this.enforceEnabled = isEnforceEnabled()
+        this.lazyLoader = new LazyLoader({
```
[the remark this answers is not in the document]
meikelmosby: Same finding as mayteio's comment on the same file — partly addressed in 3c3c5d8bb7d with the v1 staleness comment. Pubsub-based invalidation deferred to a follow-up.

nodejs/src/cdp/services/messaging/recipient-preferences.service.ts
```diff
@@ -28,6 +39,27 @@ export class RecipientPreferencesService {
...
 
+    private async isRecipientSuppressed(
+        invocation: CyclotronJobInvocationHogFunction,
+        action: MessageAction
+    ): Promise<boolean> {
+        // Suppression is driven by email bounces, so it only applies to email sends.
+        if (action.type !== 'function_email') {
+            return false
+        }
+        const email = invocation.state.globals.inputs?.email?.to?.email
```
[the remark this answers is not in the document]
meikelmosby: Addressed in 49e3c28e00a (initial fix scans to + cc + bcc in `isRecipientSuppressed`) and reinforced in 3c3c5d8bb7d where the check moved into `executeSendEmail` — `buildSuppressionSkipReason` there does the same to/cc/bcc scan, so even paths that skip `RecipientPreferencesService` (email destinations) are covered.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
meikelmosby: > the soft-bounce threshold it stays suppressed even if it starts delivering again
how should that happen when we suppress sending? 😅 .. what you might mean is to auto-block suppressed addresses when they are transient bounces.. so a time based expiry?

i'd love to just ship this now and handle this later tbh as the first version i would be happy to just say you need to handle the un-suppressing yourself and we can come up with something like a time based expiry as FUP

nodejs/src/cdp/services/messaging/email-suppression.service.ts
meikelmosby: totally fair but also low likelihood to cause actually problems at bigger scale for now. love to ship this as v2 alongside with the time based expiry for blocked transient email addresses.. using valkey should be fine in the future for this.

nodejs/src/cdp/services/messaging/email-suppression.service.ts
dmarchuk: Yeah agree with something like time based expiry 👍 not a blocker for sure

nodejs/src/cdp/services/messaging/helpers/ses.ts
```diff
@@ -234,8 +234,23 @@ export const formatSesEventLogs = (rec: SesEventRecord): SesEventLogLine[] => {
...
+        // Empty set means no restriction (dev/test); prod is expected to configure the workflow SES
+        // topic ARN.
+        allowedTopicArns: string[] = []
+    ) {
+        this.allowedTopicArns = new Set(allowedTopicArns.map((arn) => arn.trim()).filter(Boolean))
+    }
 
-    constructor(private trackingCodeSigner: EmailTrackingCodeSigner) {}
+    private isTopicAllowed(topicArn: string | undefined): boolean {
+        if (this.allowedTopicArns.size === 0) {
```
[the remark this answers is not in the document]
meikelmosby: wont fix in this PR.

products/messaging/backend/api/message_suppression.py
```diff
@@ -0,0 +1,204 @@
...
+            OpenApiParameter(name="page", type=int, location=OpenApiParameter.QUERY, required=False),
+            OpenApiParameter(name="page_size", type=int, location=OpenApiParameter.QUERY, required=False),
+        ],
+        responses={200: PaginatedMessageSuppressionSerializer},
+        summary="List suppressed email addresses for the team",
+    )
+    @action(detail=False, methods=["get"])
+    def suppressions(self, request: Request, **kwargs: Any) -> Response:
+        """List suppressed recipients for the team, most recently updated first."""
+        suppressions = (
```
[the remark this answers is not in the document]
meikelmosby: addressed

mayteio (approved):
Hell yeah man, #senditfriday

dmarchuk (approved):

PR #71073, opened 2026-07-15
feat(billing): attach actor IP to billing activity logs

yasen-posthog:
## Problem

Billing activity log rows had no IP address: billing is called server-to-server, so it only ever sees PostHog's egress IP, never the end user's.

## Changes

- Forward the end-user IP to billing on authed calls via the `X-PostHog-Actor-IP` header
- Stamp the forwarded IP onto billing activity log rows in the SQS consumer
- System-origin changes (Stripe, dunning) carry no actor IP and stay null

> [!IMPORTANT]
> Needs a companion billing PR to read `X-PostHog-Actor-IP` and include `ip_address` in the SQS message. Until then the field is read null-safely and stays null.

- 2 header-forwarding cases added; extended the consumer test to assert the IP persists on the row
- Agent-authored; only the automated tests above were run.

## Automatic notifications

- [ ] Publish to changelog?


yasen-posthog:
Addressed the security/robustness feedback on this PR in `0ca0a64`:

- **Spoofable actor IP (hex-security + greptile):** switched from `get_ip_address()` to `get_trusted_client_ip()` in `BillingViewset`, so the forwarded client IP is only accepted when the proxy chain is trusted — a spoofed `X-Forwarded-For` can no longer dictate the IP recorded in billing audit records.
- **License activation skipped attribution (greptile):** the `license` action now constructs its `BillingManager` with the trusted actor IP for both the activation request and `update_license_details`, so that path is attributed too.
- **Malformed IP could block the queue (greptile):** `BillingConsumer` now runs the inbound `ip_address` through a new `sanitize_ip_address()` helper, dropping empty/malformed/non-string values to `None` so they can't fail the IP-column write and leave the SQS message looping on redelivery. Added parameterized coverage for these cases.

ee/billing/queue/BillingConsumer.py
```diff
@@ -255,6 +256,9 @@ def _process_billing_activity(self, body: dict[str, Any]) -> None:
             scope="Billing",
             activity=activity,
             detail=Detail(name=detail_data.get("name"), changes=changes),
+            # Sanitize so a malformed/non-string value can't fail the IP column write and
+            # leave the message looping on redelivery forever.
+            ip_address=sanitize_ip_address(body.get("ip_address")),
```
pawel-cebula: This still lets actor IP and actor identity drift apart. The license path now sends `X-PostHog-Actor-IP`, but it builds `BillingManager(license, ip_address=ip_address)` without the authenticated user, so the Billing JWT has no `distinct_id`. If Billing emits an activity message with that IP and no actor, this consumer resolves `user=None`, logs `is_system=True`, and still persists the IP here. Same issue applies to any missing/foreign `distinct_id` message because the IP is not gated on the resolved org-scoped user. Can we only pass an IP to `log_activity` when `user` resolved, and add assertions that the `distinct_id=None` / cross-org actor cases keep `log.ip_address is None`?

pawel-cebula (commented):
Left one comment about keeping actor IP and actor identity coupled in billing audit rows.

ee/billing/queue/BillingConsumer.py
yasen-posthog: Good catch. Fixed in d0e5921 by gating the persisted IP on a resolved org-scoped actor in the consumer:

```python
ip_address = sanitize_ip_address(body.get("ip_address")) if user is not None else None
```

So a system-origin message (no `distinct_id`), a message whose `distinct_id` isn't found, or one whose `distinct_id` belongs to a user outside the org — all of which resolve `user=None` and log `is_system=True` — now drop the IP regardless of what upstream sent. That covers the license-activation path too, since the fix is downstream of every caller.

Extended `test_writes_system_activity_when_distinct_id_absent` and `test_does_not_attribute_distinct_id_from_another_organization` to assert `log.ip_address is None`.

pawel-cebula (approved):

PR #71086, opened 2026-07-15
chore(python): enforce explicit exception chaining

hpouillot:
## Problem

Ruff's `B904` rule is globally ignored, so new exceptions raised inside `except` handlers can omit an explicit cause without lint feedback.

**Why:** Enforcing the rule now prevents the exception-chaining backlog from growing while existing violations can be fixed incrementally.

## Changes

- Enable Ruff `B904` across the Python codebase.
- Add targeted `# noqa: B904` suppressions for existing violations.
- Keep runtime behavior unchanged in this prerequisite PR.

- `ruff check . --no-cache`
- `ruff format --check` on all changed Python files
- `git diff --check origin/master`
- Verified a new unchained raise is rejected by `B904`.

`hogli ci:preflight --fix` could not run because `hogli`, Python, and Flox are unavailable in the task environment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation update is needed because this only changes lint enforcement and adds suppressions for existing code.

## 🤖 Agent context

Codex CLI through PostHog Code converted the existing exception-chaining PR into a baseline-only prerequisite. No repository skill was required because this change only updates Ruff configuration and line-level lint suppressions.

---


ablaszkiewicz (approved):
Looks safe to merge 👍

hpouillot:
While checking some issues, I came across few exceptions that are really hard to debug because we are missing the root cause. Agents are confused and struggles to fix the real problem without the context. IMO we should enforce the "from" clause when applicable.

webjunkie (changes requested):
I think forcing chaining might leak sensitive details if not done carefully.

I'd recommend to check if we instead of `noqa: B904` which invites this should do 

    raise AuthenticationError("Authentication failed") from None

right away here at security boundaries (when translating an internal error into an API-, CLI-, or user-facing error) as a good pattern.

hpouillot:
Closing the PR as python already add the root cause as __context__ when an error is thrown inside an except clause and it is supported by our python SDK.

PR #71093, opened 2026-07-15
feat(conversations): add draft mode for ticket replies

xljones:
## Problem

Replying to a ticket is a single click with no confirmation step, so it's easy to send a customer-facing reply prematurely or to the wrong channel. Draft mode adds an opt-in "are you sure" step for agents who want it.

## Changes

- New "Draft mode" toggle left of the Send button. When on, the composer is tinted green (mirroring how a private note tints it yellow).
- Pressing Send in draft mode first shows a "Ready to send?" confirmation naming the recipient: the customer email address(es) plus CC for email tickets, or a generic phrase for Slack / Teams / GitHub. Private notes skip the confirmation since they're never sent externally.
- A browser-local default lives at the bottom of Conversations > General settings. Each ticket seeds its toggle from that default on open; flipping it for one ticket is a local override and does not change the default.

### Local

Test manually locally, the main new setting lives here
Screenshot 2026-07-15 at 17 51 18

When in draft mode, you can disable (for this ticket only), or leave on and draft the response
Screenshot 2026-07-15 at 17 50 53

If draft mode is enabled, you'll be prompted as follows before the send goes through
Screenshot 2026-07-15 at 17 50 58

I can't test this with email / Teams as I don't have these available to me locally.

### Automated

- Added parameterized Jest cases for the new `replyRecipientDescription` selector (email single recipient, email with CC, and the slack/teams/github/widget phrases). This guards the recipient string shown in the confirmation, in particular that email uses `email_from` (the customer) and not `email_to` (our sending identity). Full file passes 17/17.
- I did not manually exercise the UI.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

n/a

## 🤖 Agent context

Built with PostHog Code (Claude Opus). Skill invoked: `/writing-tests`.

- The draft-mode default is stored per-browser in localStorage (a small dedicated kea logic), not on the team `conversations_settings` blob, to match existing client-persisted UI state and avoid a backend change. The per-ticket toggle is an in-memory override seeded on mount.
- Recipient names for Slack/Teams are generic phrases because only channel IDs are exposed client-side today; email shows the real address(es).
- The composer tint started as blue but was switched to green using the `success-highlight` / `border-success` semantic tokens. Raw `bg-blue-*` utilities weren't in the compiled CSS and read too faint; the success tokens are used app-wide and mirror the private-note `warning` pair exactly.

---


products/conversations/frontend/components/Chat/MessageInput.tsx
```diff
@@ -65,16 +75,29 @@ export function MessageInput({
...
+            }
+            // Private notes are never sent externally, so they skip the draft-mode confirmation.
+            if (draftMode && !isPrivate && sendConfirmationMessage) {
+                LemonDialog.open({
+                    title: 'Ready to send?',
+                    description: sendConfirmationMessage,
+                    primaryButton: { children: 'Send', type: 'primary', onClick: doSend },
+                    secondaryButton: { children: 'Cancel' },
+                })
+            } else {
```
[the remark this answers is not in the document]
xljones: Good catch, fixed in 5161576. In draft mode `messageSending` only flips once Send is confirmed inside `doSend`, so nothing blocked a concurrent submit. Two changes: a `confirmOpenRef` guard blocks opening a second dialog while one is open (reset via the dialog's `onClose`), and `handleSubmit` now early-returns while `messageSending`/`isUploading` so the Cmd+Enter path can't re-submit during an in-flight send (the button was already disabled, but Cmd+Enter bypassed it).

xljones:
Changed the draft mode text to be the same as the attach private note.

image

products/conversations/frontend/components/Chat/MessageInput.tsx
```diff
@@ -58,23 +69,48 @@ export function MessageInput({
...
+            // Private notes are never sent externally, so they skip the draft-mode confirmation.
+            if (draftMode && !isPrivate && sendConfirmationMessage) {
+                // A rapid second Cmd+Enter would otherwise stack a duplicate dialog, since
+                // messageSending only flips once Send is confirmed inside doSend.
+                if (confirmOpenRef.current) {
+                    return
                 }
-            })
+                confirmOpenRef.current = true
+                LemonDialog.open({
```
veryayskiy, line 95: MessageInput.tsx: LemonDialog.open({ onClose }) never runs — LemonDialog only calls onAfterClose, and .open() overwrites that with its unmount destroy. So confirmOpenRef never clears; after one confirm/cancel (or dismiss via X/Esc), draft-mode send silently no-ops until remount. Reset the ref in the dialog button onClicks (and for dismiss, chain onAfterClose in LemonDialog.open or stop using the ref).

veryayskiy: the result: 
Image

veryayskiy (changes requested):

products/conversations/frontend/components/Chat/MessageInput.tsx
xljones: good spot. this was actually a guard for something i wasn't that worried about [here](https://github.com/PostHog/posthog/commit/5161576b7130d673d74e8c74a7ae809759caccc6). Will fix

xljones: Fixed in ad69788 — you're exactly right on both counts. `onClose` is a dead prop on `LemonDialog` (accepted but never invoked; the inner `LemonModal` gets its own `onClose={() => setIsOpen(false)}`), and `LemonDialog.open()` overwrites `onAfterClose` with its unmount handler. So the ref was set on open and never cleared → draft-mode send no-opped after the first confirm/cancel/dismiss until remount.

Rather than resetting only in the button onClicks (which misses Esc/backdrop), I made `LemonDialog.open`/`openForm` chain the caller's `onAfterClose` before unmounting instead of dropping it, and reset the guard from `onAfterClose` so it fires on every close path.

Bonus: that swallowed-`onAfterClose` is a latent bug elsewhere too — `featureFlagLogic` calls `LemonDialog.open({ onAfterClose: () => resolve(false) })` inside an `await new Promise`, so cancelling/Esc-ing the "Change flag key?" dialog currently never resolves and silently hangs the submit. This fix resolves that as well.

Kept it to the chaining fix rather than adding a `.open()` portal/transition test, since that path is flake-prone in jsdom and the existing LemonDialog suite stays green.

xljones: Reworked in 48296471 to avoid touching LemonDialog at all. Backed out both the `confirmOpenRef` guard and the shared LemonDialog change.

The double-dialog scenario the guard was for can't actually happen: LemonModal uses `react-modal`, which traps focus while the dialog is open, so a second Cmd+Enter goes to the modal rather than the tiptap editor and never re-fires `onPressCmdEnter`. So the guard was defending against something the modal already prevents. The `messageSending`/`isUploading` early-return stays and covers the real case (re-submit during an in-flight send on the Cmd+Enter path).

Separately: the swallowed-`onAfterClose` in `LemonDialog.open()` is a real latent bug (e.g. featureFlagLogic's "Change flag key?" dialog resolves its promise via `onAfterClose`, which never runs on cancel/Esc). I've left it out of this PR since it's unrelated to draft mode — happy to raise it separately for the lemon-ui owners.

xljones: Tested in local dev: 

https://github.com/user-attachments/assets/e2d807a0-a0bc-4e0d-a86b-a2a8f9f0b3ad

veryayskiy (approved):
Let's go!

PR #71103, opened 2026-07-15
feat(tracing): add operation detail scene with latency histogram and trace samples

jonmcwest:
## Problem

The Operations tab shows aggregate latency and error stats per operation, but there's no way to drill into one operation: no view of how its latency distributes, and no path from "p95 looks bad" to an actual slow trace. Separately, tracing's read endpoints were only partially instrumented, so we couldn't measure whether anyone uses these surfaces across web, API, and MCP callers.

## Changes

Clicking a row in the Operations tab now opens a per-operation scene:

- A span-level latency distribution histogram on 1-2-5 log buckets. Operations are usually child spans, so the histogram endpoint honors `rootSpans=false` and buckets the operation's own span durations, the same population the samples query counts.
- Drag-select a duration range on the histogram to filter a pager of sampled traces from that range, each rendered as an inline waterfall anchored on the operation's own span.
- Date range, duration selection, and sample index all sync to the URL for deep linking. The date range carries over from the Operations tab, so the detail stats match the row that was clicked.

Backend instrumentation: the duration-histogram, aggregate, and per-trace endpoints now fire `report_user_action` events (`tracing duration histogram queried`, `tracing aggregation queried`, `tracing trace fetched`) matching the existing `tracing query executed` pattern, so usage is segmentable by `access_method` (web, personal API key, MCP).

Review hardening rolled into this PR:

- Superseded requests are aborted server-side (per-loader `AbortController`s via the disposables plugin), instead of just having their results discarded client-side.
- The samples query passes `excludeAttributes` since the scene only reads scalar span fields, and adjacent samples that share a trace reuse the already-loaded waterfall instead of refetching it.
- Deep links with a negative `sample` index are rejected, links missing the service param get the missing-params fallback instead of a silent empty query, and a persisted duration selection that falls entirely off a reshaped histogram axis renders no highlight instead of highlighting everything.
- The trace-list duration histogram now counts the population the active view shows (root spans in Traces view, every matching span in Spans view), matching the sparkline.
- Deduplicated helpers: category duration axis config, error-rate formatting, and the duration-range type each live in one place now.

Screenshots are from a locally seeded run; I can attach them in a comment.

Automated, run locally:

- `products/tracing/backend/tests/test_duration_histogram.py` covers root-level vs span-level bucketing through the endpoint, including the `rootSpans=false` path the scene depends on.
- `products/tracing/frontend/durationBuckets.test.ts` covers the 1-2-5 bucket math (snap, fill, upper bound, selection-to-range).
- `products/tracing/frontend/tracingOperationSceneLogic.test.ts` covers the two regressions most likely to recur: a drag selection must refetch samples scoped to the selected range and reset the pager, and paging to a sample must fetch its containing trace and anchor the waterfall on the sample's own span.
- Full affected suite green: 52 frontend tests across the three tracing test files, 6 backend tests, plus `ruff` and `oxlint`/`oxfmt` on changed files. TypeScript check is clean for every touched file.

Manually verified earlier in the session against a locally seeded dev stack: scene loads from the Operations tab, drag selection filters samples, pager and waterfall behave, deep links restore selection and sample.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with PostHog Code (Claude). Skills invoked while producing this PR: `/improving-drf-endpoints`, `/writing-tests`, `/using-kea-disposables`, plus a simplify and multi-angle code review pass whose confirmed findings were applied here.

Decisions worth knowing about as a reviewer: the histogram endpoint defaults `rootSpans=true` so existing callers keep the trace-level distribution, and the scene opts into span-level explicitly. Instrumentation went into the backend viewset (not `posthog.capture` in the frontend) specifically so usage is visible across web, API, and MCP via `access_method`. The review pass deliberately skipped a few flagged items as out of scope: single-bucket drag selection needs a shared `Sparkline` change, the hand-rolled `api.tracing` client block predates this PR and migrating it to generated types is a separate effort, and the repeated `report_user_action`/filter-group parsing boilerplate in the tracing viewset (7 and 5 copies now) is worth a follow-up helper.

---

### DIFF
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -3016,6 +3016,9 @@ const api = {
                 serviceNames?: string[]
                 statusCodes?: number[]
                 filterGroup?: PropertyGroupFilter
+                // true (default) buckets root spans only (a distribution of traces); false buckets
+                // every matching span — pair with a span name filter for operation-scoped pages.
+                rootSpans?: boolean
             },
             signal?: AbortSignal
         ): Promise<{
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -2955,6 +2955,8 @@ const api = {
                 // false (default) groups by trace_id and returns root spans; true returns every
                 // matching span (root and child) flat. See products/tracing/backend logic.py.
                 flatSpans?: boolean
+                // true omits the per-span attribute maps — use when only scalar span fields are read.
+                excludeAttributes?: boolean
             },
             signal?: AbortSignal
         ): Promise<{
@@ -2972,12 +2974,14 @@ const api = {
                 statusCodes?: number[]
                 filterGroup?: PropertyGroupFilter
                 offset?: number
-            }
+            },
+            signal?: AbortSignal
         ): Promise<{ results: Record<string, any>[]; hasMore: boolean; nextOffset?: number | null }> {
             return new ApiRequest()
                 .tracingSpans()
                 .withAction(`trace/${traceId}`)
                 .create({
+                    signal,
                     data: {
                         ...query,
                         dateRange: query?.dateRange ?? { date_from: '-24h' },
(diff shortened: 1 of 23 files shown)


jonmcwest:
Opened from a branch with a stale master base, so the diff re-presented the already-merged operation detail scene (#67757) as new. Replaced by #71484, which contains only the new work (usage events + hardening) on top of current master.

PR #71111, opened 2026-07-15
feat(error-tracking): add selectable MCP event context

hpouillot:
## Problem

Agents need enough error-event context to debug issues without receiving a large exception payload on every request. Stack traces, code variables, release details, and ingestion diagnostics should be available explicitly when needed.

## Changes

- Add an `include` parameter to `query-error-tracking-issue-events` with compact context groups.
- Keep exception, environment, navigation, and correlation context as the default.
- Make stack traces, SDK-masked code variables, releases, and diagnostics opt-in.
- Select only the event properties required by the requested context groups.
- Regenerate frontend and MCP API schemas and document the new parameter for agents.

- `ruff check` passes for the changed Python files.
- `hogli build:openapi` completes and regenerates the frontend and MCP artifacts.
- `pnpm --filter=@posthog/mcp lint-tool-names` passes.
- `hogli ci:preflight --strict` passes.
- Added API coverage for context filtering and explicit code-variable inclusion. The focused test could not execute locally because the workspace Postgres hostname was unavailable.
- MCP type checking was attempted but the workspace has unresolved local `@posthog/quill` and `@posthog/quill-charts` packages unrelated to this change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The MCP tool prompt documents the selectable context groups and the sensitive-data considerations for code variables.

## 🤖 Agent context

Codex implemented a narrow, backward-compatible extension to the existing issue-events endpoint rather than introducing a new event response shape. The existing `verbosity` parameter remains supported for raw payload compatibility, while `include` controls which context groups are queried and returned. Code variables remain excluded unless explicitly requested through `include`, except for legacy callers already requesting stack or raw verbosity without the new parameter.

---


products/error_tracking/backend/facade/query_utils.py
```diff
@@ -42,27 +42,55 @@ def normalize_volume_resolution(volume_resolution: int) -> int:
...
+        "properties.$browser_version",
+        "properties.$os",
+        "properties.$os_version",
+        "properties.$app_namespace",
+        "properties.$app_version",
+        "properties.$device_type",
+    ],
+    "release": ["properties.$exception_releases"],
+    "navigation": ["properties.$current_url", "properties.$screen_name", "properties.$referrer"],
+    "correlation": [
```
hpouillot, line 70: adding few ids to link with other resources

products/error_tracking/backend/presentation/views/query.py
```diff
@@ -248,14 +264,25 @@ def issue_events(self, request: ValidatedRequest, **kwargs: object) -> Response:
...
+                    .model_dump(mode="json")
+                )
+            except ResolutionError as error:
+                raise ValidationError(str(error)) from error
         raw_columns = data.get("columns")
-        columns = [str(column) for column in raw_columns] if isinstance(raw_columns, list) else EVENT_SELECTS
+        columns = [str(column) for column in raw_columns] if isinstance(raw_columns, list) else event_selects
         raw_results_value = data.get("results")
         raw_results: list[object] = raw_results_value if isinstance(raw_results_value, list) else []
-        verbosity = cast(str, params.get("verbosity", "summary"))
```
hpouillot, line 256: Dropping the verbosity param in favor of include. It won't be backward compatible but this endpoint is only used by agents from what I saw

products/error_tracking/backend/facade/query_utils.py
cat-ph: 💙

products/error_tracking/backend/facade/query_utils.py
```diff
@@ -262,26 +297,29 @@ def normalize_exception_list(value: object, verbosity: str, only_app_frames: boo
...
+def normalize_error_property(
+    name: str, value: object, include_stacktrace: bool, only_app_frames: bool, include_code_variables: bool = True
+) -> object:
     if name == "$exception_list":
-        return normalize_exception_list(value, verbosity, only_app_frames)
+        return normalize_exception_list(value, include_stacktrace, only_app_frames, include_code_variables)
     if name == "$exception_releases":
         return parse_jsonish(value)
     if name == "$exception_types":
         return normalize_string_array(value)
```
cat-ph, line 324: should we add `$cymbal_errors` here too?

products/error_tracking/backend/facade/query_utils.py
```diff
@@ -42,27 +42,55 @@ def normalize_volume_resolution(volume_resolution: int) -> int:
...
-    "properties.$current_url",
-]
+DEFAULT_EVENT_CONTEXT_INCLUDES = ["exception", "environment", "navigation", "correlation"]
+EVENT_CONTEXT_PROPERTY_SELECTS = {
+    "exception": [
+        "properties.$exception_types",
+        "properties.$exception_values",
+        "properties.$exception_list",
+        "properties.$exception_fingerprint",
+        "properties.$exception_issue_id",
```
cat-ph: I know I also added this but thinking out loud whether it's still _that_ useful now with auto-merge and more volatile issue IDs changing, or it might confuse agents 🤔 but I'd err to keeping it, the more context the better 😆 

should we also add $exception_level and maybe $exception_handled? 🤔

products/error_tracking/backend/facade/query_utils.py
```diff
@@ -42,27 +42,55 @@ def normalize_volume_resolution(volume_resolution: int) -> int:
...
+        "properties.$app_version",
+        "properties.$device_type",
+    ],
+    "release": ["properties.$exception_releases"],
+    "navigation": ["properties.$current_url", "properties.$screen_name", "properties.$referrer"],
+    "correlation": [
+        "properties.$session_id",
+        "properties.$trace_id",
+        "properties.$ai_trace_id",
+        "properties.$span_id",
```
cat-ph: I thought we use `$ai_span_id`? 🤔 or is this for traces separately from AIO? I looked at our project and we don't have $trace_id / $span_id in the taxonomy, but there is `$ai_span_id`

cat-ph (approved):
coooool 🔥 I like this quite a bit more than the 'verbosity' levels! 😬 few comments but looks great!

products/error_tracking/backend/facade/query_utils.py
hpouillot: I think trace_id is for tracing and ai_trace_id is for llma but I'll double check

products/error_tracking/backend/facade/query_utils.py
hpouillot: Good point, I think we can remove it. The events are queried for a specific issue_id, if it's not the same it will be confusing 🫨 I'll add level and handled props as well 👍

products/error_tracking/backend/facade/query_utils.py
hpouillot: Good catch, It wasn't the case 😄 I added it here: https://github.com/PostHog/posthog-python/pull/743
We can rollout $trace_id and $span_id in other SDKs as well, if there are not attached the property won't be returned

cat-ph: nice! thanks for adding it 💙

PR #71132, opened 2026-07-15
feat(slack-app): answer analytics questions with charts in slack

sampennington:
## Problem

The Slack agent (`@PostHog`) answered analytics questions as text tables, even when the answer was obviously a chart — users had to literally say "chart".
And the delivery recipe that did exist made the agent the courier: export, download, verify, base64, re-upload — eight tool calls each costing an LLM turn (104s vs 43s for the same question as text, measured locally), with every step a place to fail (sandbox scope 403s, unreachable presigned redirects, one corrupted download crashing the run).

## Changes

Asking `@PostHog` an analytics question now returns a chart in the thread by default when the answer is naturally visual. The whole answer arrives as **one Slack message**: the agent's summary, the chart image titled with the chart name, and an "Open in PostHog" button linking to the live insight.

Under the hood:

- **One-call chart endpoint** — `POST /api/projects/:id/tasks/:task/runs/:run/living_artifacts/chart/` takes `{name, query}` (insight query JSON) or `{name, insight_id}`, renders the PNG server-side via the exports facade, and registers it as a `slack_file` living artifact; bytes never round-trip through the agent. Requires `task:write` **and** `query:read` (the body executes insight queries, so a task-only key must not become a data-read scope), validates against `QuerySchemaRoot` before dispatch, carries the standard ClickHouse throttles, and returns typed 400s. The open-in-PostHog `url` is persisted on the artifact to power the button.
- **Composed Slack delivery** — chart PNGs upload to Slack *without* a channel share and post as Block Kit `image` blocks in a single `chat.postMessage` together with the answer text and a per-chart button. The previous file-share delivery couldn't do this: `files.completeUploadExternal` materializes in the thread asynchronously (racing anything posted after it) and share messages can't carry buttons. A short retry absorbs Slack's upload-processing lag (`invalid_blocks`); if composing fails outright, the text posts plainly and each chart falls back to its own message. Non-image files (xlsx/pdf) keep the share path.
- **Prompt** — chart delivery collapses to the single call, visual answers (trends, funnels, breakdowns, retention) include a chart by default, and the answer text is steered toward the takeaway (trend, inflection points, numbers that matter) instead of announcing the chart, since the card already carries the title.

> [!NOTE]
> `products.tasks` gains a tach dependency on `products.exports`, restricted in practice to the facade added in #70829. `tach check --dependencies --interfaces` passes.

Verified end to end on a full local stack (real Slack workspace + tunnel, Docker sandboxes, temporal, browserless): mention → server-side render → one composed Slack message with title, image, and working button, iterated against live Slack behavior (ordering races and the `invalid_blocks` processing lag were observed empirically, not assumed).

- `TestTaskRunLivingArtifactChartAPI`: keys with only `task:write` or only `query:read` are rejected (pins the scope semantics), a both-scopes key gets a 200 with the url persisted into artifact metadata and the rendered bytes handed to the artifact service, an invalid query 400s before dispatch, and a render failure returns the typed 400 instead of a 500. The exactly-one-of matrix lives on the serializer with no DB.
- Relay/delivery tests pin the composed message shape (answer section + card blocks + button), the upload-without-share behavior, the `invalid_blocks` retry, and the flag/scope gates.
- Prompt snapshot suite (`test_task_creation.py`) passes; snapshots updated.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No public docs cover the task-run living artifacts API today.

## 🤖 Agent context

- Built in Claude Code sessions driving a real Slack workspace against the local stack; each design change corresponds to a failure observed live rather than speculation — including the switch from file shares to composed image-block messages (shares landed out of order and can't carry buttons) and the `invalid_blocks` retry (Slack rejects block references to a just-uploaded file until it finishes processing).
- Decision: composed endpoint in the tasks product calling the exports facade, rather than an MCP tool — the sandbox already talks to these REST endpoints with env-provided ids, so this gets the latency win (1 LLM turn instead of 8) without touching the separately-deployed MCP server.
- Decision: a per-chart `description` request field was added and then removed — with the answer text composing into the same message, it duplicated information; interpretation belongs in the answer prose (now prompted for explicitly).
- Considered and dropped: broadening the sandbox's raw env API key scopes (its most exposed credential) — the endpoint only needs scopes every sandbox token already carries.
- Skills invoked: `/improving-drf-endpoints`, `/writing-tests`, `/debugging-local-task-agent-runs`.

---


products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+        if url:
+            chart_metadata["posthog_url"] = url
+        artifact, error = tasks_facade.create_task_run_living_artifact(
+            self._run_id(),
+            task_id,
+            self.team_id,
+            artifact={
+                "name": name,
+                "artifact_type": TaskArtifactType.FILE,
+                "adapter": TaskArtifactAdapter.SLACK_FILE,
```
sampennington, line 2484: 🤖 **sp-ship** · 💻 `[code]` · **Blocker**

Chart creation still hard-fails without `files:write`: this goes through `SlackFileArtifactAdapter.commit`, which raises `"...missing the files:write scope"` (`living_artifacts.py:729-734`). The scope relaxation was applied only to delivery (`has_file_scope`), so in the rollout this PR targets — flag on, `files:write` not yet approved — every chart call 400s and no chart is ever created. Gate that check on url-delivered image content.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+        if error is not None:
+            return Response(TaskRunErrorResponseSerializer({"error": error}).data, status=status.HTTP_400_BAD_REQUEST)
+        serializer = TaskRunLivingArtifactChartResponseSerializer(
+            {"artifact": artifact, "export_asset_id": asset.id, "url": url}
+        )
+        return Response(serializer.data)
+
+    def _chart_url(self, query: dict | None, asset) -> str | None:
+        if query is not None:
+            return absolute_uri(f"/project/{self.team_id}/insights/new#q={quote(json.dumps(query))}")
```
sampennington: 🤖 **sp-ship** · ♻️ `[reuse]` · **Blocker**

`quote(json.dumps(query))` emits `%5C` for every JSON escape — `\uXXXX` for any non-ASCII value (`ensure_ascii` default), `\"` for a quote inside a string — and `absolute_uri` rejects `%5c` via `has_authority_bypass_chars` (`posthog/security/url_validation.py:166`), raising `PotentialSecurityProblemException` → 500. So a chart on a query with a non-ASCII event or property value 500s *after* the render already succeeded. `_deep_link` (`products/data_catalog/backend/logic/execution.py:200-212`) hit and documented this exact trap — it appends the encoded payload after `absolute_uri` resolves the path; reuse that shape.

sampennington (commented):

Four real defects, three of which only fire after the expensive render has already succeeded: chart *creation* still hard-raises without `files:write` even though delivery no longer needs it; `_chart_url` hands `absolute_uri` a `%5C`-bearing fragment it rejects outright, so any non-ASCII value in the query 500s; the Slack retry budget can outlast the relay activity's timeout and force a replay that reposts the answer and charts; and `answer_posted` reports success when the text post silently swallowed its failure. Nothing pins the view→exports-facade call contract and the failure-path test asserts only artifact bookkeeping, so each of these ships green.

The structural suggestions — answer posting leaking into the file-delivery service, the anonymous card tuple, the section limit declared in two modules — are worth a pass but don't block. Three test bodies also branch (a ternary in an assertion, two `or`/`or {}` helper defaults) and need parameterizing.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+            error = e.response.get("error")
+            if attempt == attempts or error not in ("invalid_blocks", "ratelimited"):
+                raise
+            if error == "ratelimited":
+                headers = getattr(e.response, "headers", None) or {}
+                try:
+                    retry_after = float(headers.get("Retry-After") or _IMAGE_BLOCK_POST_RETRY_INTERVAL_S)
+                except (TypeError, ValueError):
+                    retry_after = _IMAGE_BLOCK_POST_RETRY_INTERVAL_S
+                time.sleep(min(retry_after, _RATE_LIMIT_MAX_WAIT_S))
```
sampennington: 🤖 **sp-ship** · 💻⚡ `[code+efficiency]` · **Blocker**

Worst-case sleep here is 7 × `_RATE_LIMIT_MAX_WAIT_S` = 70s for the composed post alone, plus ~20s per card in the fallback — against the relay activity's `start_to_close_timeout=timedelta(minutes=1)` (`slack_relay/workflow.py:23`). One rate-limited post self-times-out, and `maximum_attempts=3` replays the whole activity; since cards are only marked delivered after `_post_composed_answer_message` returns, the retry reposts the answer text and duplicate charts. Track a wall-clock deadline across attempts and mark each card delivered as soon as its own post succeeds.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+                channel=mapping.channel,
+                thread_ts=mapping.thread_ts,
+                text=artifact.name,
+                blocks=_chart_card_blocks(artifact, file_id),
+                attempts=_IMAGE_BLOCK_FALLBACK_ATTEMPTS,
+            )
+            delivered.append(card)
+        except Exception:
+            logger.warning("task_artifact.slack_file_delivery_failed", artifact_id=str(artifact.id), exc_info=True)
+    return bool(sections), delivered
```
sampennington: 🤖 **sp-ship** · 💻🧪 `[code+test]` · **Blocker**

`answer_posted` is `bool(sections)` even when every `_post_thread_text` swallowed its exception (`living_artifacts.py:1058-1062` logs and returns `None`), so the relay's `if not answer_posted` fallback is skipped and the user's answer is silently dropped — contradicting the "the answer must not be lost" comment three lines up. Have `_post_thread_text` return whether it succeeded and OR those results in. `test_failed_chart_post_leaves_artifact_pending` asserts only artifact bookkeeping, so it passes straight through this.

products/tasks/backend/tests/test_api.py
```diff
@@ -8008,6 +8008,133 @@ def test_scoped_api_key_permissions(self, scope, method, url_template, should_ha
...
+            "https://app.dev/exporter/export-1.png?token=abc"
+        )
+        mock_render.return_value = (mock_asset, b"png-bytes")
+        mock_create.return_value = (self._artifact_response(), None)
+        response = self._post_chart(["task:write", "query:read"], {"name": "Chart", "query": self.CHART_QUERY})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        self.assertEqual(data["export_asset_id"], 321)
+        self.assertIn("/insights/new#q=", data["url"])
+        self.assertEqual(mock_create.call_args.kwargs["artifact"]["content_bytes"], b"png-bytes")
```
sampennington, line 8087: 🤖 **sp-ship** · 🧪 `[test]` · **Blocker**

Nothing asserts *how* the view calls `render_png_export`, so a wrong `export_context` wrapping (`query` unwrapped, or `{"query": ...}`) or a stray `insight_id` alongside it would 400 every real request while this test stays green — the facade rejects any context whose `source` isn't an `InsightVizNode` (`products/exports/backend/facade/api.py:32-38`). Add `mock_render.assert_called_once_with(team=self.team, created_by=self.user, export_context={"source": self.CHART_QUERY}, insight_id=None)`.

products/tasks/backend/logic/services/tests/test_living_artifacts.py
```diff
@@ -626,9 +626,35 @@ def test_pending_file_delivery_skipped_when_flag_off(self, mock_integration_for_
...
+            ("url_over_slack_button_cap", "https://us.posthog.com/project/1/insights/new#q=" + "x" * 3000, False),
+        ]
+    )
+    def test_button_only_added_when_url_fits_slack_cap(self, _name, url, expect_button):
+        from products.tasks.backend.logic.services.living_artifacts import _chart_card_blocks
+
+        artifact = TaskArtifact(name="Chart", metadata={"posthog_url": url})
+        blocks = _chart_card_blocks(artifact, "F123")
+        self.assertEqual(
+            [b["type"] for b in blocks], ["section", "image", "actions"] if expect_button else ["section", "image"]
```
sampennington: 🤖 **sp-ship** · 🦔💻✨ `[personal+code+quality]` · **Blocker**

Ternary in the assertion picks the expected block list. Move `["section", "image", "actions"]` / `["section", "image"]` into the `parameterized.expand` cases instead of branching on `expect_button`. (The function-level imports at 646/655 also go against the repo's module-level import rule.)

products/tasks/backend/tests/test_api.py
```diff
@@ -8008,6 +8008,133 @@ def test_scoped_api_key_permissions(self, scope, method, url_template, should_ha
...
+        api_key_value = generate_random_token_personal()
+        PersonalAPIKey.objects.create(
+            user=self.user,
+            label="chart test key",
+            secure_value=hash_key_value(api_key_value),
+            scopes=scopes,
+        )
+        self.client.force_authenticate(None)
+        return self.client.post(
+            f"/api/projects/@current/tasks/{task.id}/runs/{run_id or run.id}/living_artifacts/chart/",
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Blocker**

`run_id or run.id` is a short-circuit branch in a test helper — a conditional decides which run is exercised. Have every caller pass the run id explicitly, or split the helper in two.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
```diff
@@ -281,34 +263,68 @@ def test_pending_slack_file_upload_uses_final_message(
...
             created_by=self.user,
-            name="report.xlsx",
-            artifact_type=TaskArtifact.ArtifactType.SPREADSHEET,
+            name=name,
+            artifact_type=TaskArtifact.ArtifactType.FILE,
             adapter=TaskArtifact.Adapter.SLACK_FILE,
             status=TaskArtifact.Status.ACTIVE,
             location=location,
-            metadata={"delivery_status": "pending"},
+            metadata={"delivery_status": "pending", **(metadata or {})},
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Blocker**

`**(metadata or {})` branches inside a test helper. Make `metadata` a required kwarg and pass `{}` from the xlsx caller so the fixture is computed with no branching.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -784,69 +785,128 @@ def has_pending_slack_file_artifacts(run: TaskRun) -> bool:
...
 
 
-def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str) -> int:
+@dataclass
+class SlackFileDeliveryResult:
+    answer_posted: bool = False
+    delivered_count: int = 0
+
+
+def has_pending_slack_image_artifacts(run: TaskRun) -> bool:
```
sampennington, line 829: 🤖 **sp-ship** · ♻️✨⚡📐 `[reuse+quality+efficiency+xp]` · **Suggestion / follow-up**

This is the third full scan and re-parse of the same `slack_file` rows per relay tick — after `has_pending_slack_file_artifacts`, before `deliver_pending_slack_file_artifacts` walks them again. One `_pending_slack_files(run)` helper with the predicates derived from it would drop a query and two JSON walks. It also omits the `_living_artifacts_enabled_for_run` gate its identically-named sibling applies, which only happens to be safe because the relay short-circuits on the sibling first.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -784,69 +785,128 @@ def has_pending_slack_file_artifacts(run: TaskRun) -> bool:
...
+        pending = _pending_slack_file_version(artifact)
+        if pending is None:
+            continue
+        content_type = _pending_slack_content_type(artifact, pending[1])
+        if content_type.startswith("image/"):
+            return True
+    return False
+
+
+def deliver_pending_slack_file_artifacts(
```
sampennington, line 840: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion / follow-up**

This function now delivers files *and* posts the agent's answer, handing back `answer_posted` so the relay avoids double-posting — you can't describe it without "and". Exposing the pending chart cards (blocks + a mark-delivered call) and leaving answer posting with the relay's `SlackThreadHandler` would remove `answer_sections`, `answer_posted`, `_post_thread_text` (a duplicate of `post_thread_message`), and the duplicated char limit in one move.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+
+
+# Slack caps a message at 50 blocks; leading answer sections spill into plain
+# messages when a message would exceed it.
+_SLACK_MESSAGE_BLOCK_LIMIT = 50
+
+
+# (artifact, pending version, uploaded file id, upload response) — the last two are
+# None for url-referenced images, which involve no upload.
+_SlackImageCard = tuple[TaskArtifact, dict[str, Any], "str | None", "dict[str, Any] | None"]
```
sampennington, line 1037: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion / follow-up**

`_SlackImageCard` is an anonymous 4-tuple that needs a comment to explain its slots and forces `pending[1]` / `f[2]` / throwaway `_version_payload` unpacks. `SlackFileDeliveryResult` right above shows the better pattern — a `NamedTuple` or small dataclass makes these loops self-describing and lets the image/other split filter on a named attribute.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+            )
+            delivered.append(card)
+        except Exception:
+            logger.warning("task_artifact.slack_file_delivery_failed", artifact_id=str(artifact.id), exc_info=True)
+    return bool(sections), delivered
+
+
+# Section blocks hard-cap at 3000 chars, and mrkdwn conversion can expand text past
+# the relay's pre-conversion split (table column padding) — re-split after conversion.
+_SLACK_SECTION_BLOCK_CHAR_LIMIT = 2900
```
sampennington: 🤖 **sp-ship** · ♻️✨📐🧪 `[reuse+quality+xp+test]` · **Suggestion / follow-up**

The 2900-char section budget is declared twice under two names — here and `SLACK_SECTION_TEXT_LIMIT` in `slack_relay/activities.py:245` — and the two must agree or the relay's pre-split chunks get re-split after conversion. Nothing ties them together in a test either. Export one constant.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+                channel=mapping.channel,
+                thread_ts=mapping.thread_ts,
+                text=fallback_text,
+                blocks=[*kept, *card_blocks],
+            )
+            return bool(sections), list(image_cards)
+        except Exception:
+            logger.warning("task_artifact.slack_composed_message_failed", exc_info=True)
+
+    # Degraded path: the answer must not be lost — post the text plainly, then each
```
sampennington, line 1088: 🤖 **sp-ship** · 📐 `[xp]` · **Suggestion / follow-up**

The degraded path triggers on any exception from the composed post, including ones where Slack accepted the message but the SDK raised (read timeout) — the fallback then re-posts the answer text and every card, duplicating the whole answer in the thread. Narrowing it to failures that prove nothing was delivered would keep the guarantee without the at-least-once side effect.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+# Slack rejects button urls over 3000 chars, and an ad-hoc chart url embeds the whole
+# encoded query JSON — a fat multi-series query can exceed it.
+_SLACK_BUTTON_URL_MAX_CHARS = 3000
+
+
+def _chart_card_blocks(artifact: TaskArtifact, file_id: str | None) -> list[dict[str, Any]]:
+    metadata = artifact.metadata or {}
+    image_url = metadata.get("image_url")
+    if isinstance(image_url, str) and image_url:
+        image_block: dict[str, Any] = {"type": "image", "image_url": image_url, "alt_text": artifact.name}
```
sampennington, line 1185: 🤖 **sp-ship** · ♻️ `[reuse]` · **Suggestion / follow-up**

This duplicates `_block_for_asset` (`ee/tasks/subscriptions/slack_subscriptions.py:79`), which additionally swaps in a fetchable url under `settings.DEBUG` — without that, a chart card in local dev points Slack's image proxy at a localhost export url and renders broken.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+    elif isinstance(posthog_url, str) and posthog_url:
+        logger.warning(
+            "task_artifact.chart_url_too_long_for_button",
+            artifact_id=str(artifact.id),
+            url_length=len(posthog_url),
+        )
+    return blocks
+
+
+def _post_blocks_with_processing_retry(
```
sampennington, line 1222: 🤖 **sp-ship** · ♻️ `[reuse]` · **Suggestion / follow-up**

This re-implements Slack post retry that already exists twice — `_send_slack_message_with_retry` + `_RETRYABLE_SLACK_ERRORS` (`ee/tasks/subscriptions/slack_subscriptions.py:34,286`) and `_slack_retry_after_seconds` (`products/signals/backend/tasks.py:84`) — the latter also handling the lowercase `retry-after` header this copy misses.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+    @action(
+        detail=False,
+        methods=["post"],
+        url_path="chart",
+        # query:read because the body executes arbitrary insight queries — task:write alone
+        # must not become a data-read scope.
+        required_scopes=["task:write", "query:read"],
+        throttle_classes=[ClickHouseBurstRateThrottle, ClickHouseSustainedRateThrottle],
+    )
+    def chart(self, request, *args, **kwargs):
```
sampennington, line 2407: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion / follow-up**

`chart()` is ~75 lines at mixed abstraction levels: task gating, run precheck, query allow-list, schema validation, the blocking render, url + metadata assembly, artifact registration — with five inline `Response(...400)` constructions. Extracting `_validated_chart_query(query)` (raising a DRF `ValidationError` rather than hand-building 400s) and `_register_chart_artifact(...)` would leave a linear happy path and make the allow-list independently testable.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+                TaskRunErrorResponseSerializer({"error": asset.exception or "Chart render failed"}).data,
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        # Persisted on the artifact so Slack delivery can compose the chart card:
+        # image_url lets Slack's image proxy fetch the rendered PNG from us (no
+        # files:write scope needed), posthog_url powers the "Open in PostHog" button.
+        # The delivery-purposed token works for orgs that disallow publicly shared
+        # resources, same as subscription images.
+        url = self._chart_url(query, asset)
+        chart_metadata: dict = {"image_url": asset.get_subscription_delivery_content_url()}
```
sampennington: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion / follow-up**

`get_subscription_delivery_content_url()` mints a 365-day (`PUBLIC_ACCESS_TOKEN_EXP_DAYS`) unauthenticated JWT that deliberately bypasses `allow_publicly_shared_resources` (`posthog/api/sharing.py:935-939`), and it's persisted in artifact metadata returned by the living-artifacts list/retrieve endpoints — so anything that can read the artifact holds a year-long public link to team analytics. Consider passing an `expiry_delta` closer to the living-artifact TTL.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+                team=self.team,
+                created_by=request.user,
+                export_context={"source": query} if query is not None else None,
+                insight_id=request.validated_data.get("insight_id"),
+            )
+        except ValueError as e:
+            return Response(TaskRunErrorResponseSerializer({"error": str(e)}).data, status=status.HTTP_400_BAD_REQUEST)
+        if png is None:
+            return Response(
+                TaskRunErrorResponseSerializer({"error": asset.exception or "Chart render failed"}).data,
```
sampennington: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion / follow-up**

`asset.exception` is the raw `str(e)` from the export pipeline (`posthog/tasks/exporter.py:39`), so ClickHouse or headless-browser internals can be echoed verbatim in the 400 — and the new prompt tells the agent to "read the `error` field and report the failure", relaying it into a Slack channel. Map to a short user-facing reason and log the raw text.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+                    TaskRunErrorResponseSerializer(
+                        {
+                            "error": "Only insight queries wrapped in an InsightVizNode can be charted — "
+                            "SQL and table queries are not supported yet"
+                        }
+                    ).data,
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+            try:
+                QuerySchemaRoot.model_validate(upgrade(query))
```
sampennington: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion / follow-up**

The validated value is discarded — `upgrade(query)` is called for its exception only, then the raw `query` is what gets rendered and embedded in the "Open in PostHog" url. `upgrade_node` mutates nested nodes in place, so a root-level migration would validate one shape and render another. Bind `upgraded = upgrade(query)` and pass that through.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+        operation_id="tasks_runs_living_artifacts_chart",
+    )
+    @action(
+        detail=False,
+        methods=["post"],
+        url_path="chart",
+        # query:read because the body executes arbitrary insight queries — task:write alone
+        # must not become a data-read scope.
+        required_scopes=["task:write", "query:read"],
+        throttle_classes=[ClickHouseBurstRateThrottle, ClickHouseSustainedRateThrottle],
```
sampennington: 🤖 **sp-ship** · 💻⚡ `[code+efficiency]` · **Suggestion / follow-up**

`ClickHouseBurstRateThrottle` (240/min) is calibrated for sub-second queries, but this action holds a web worker and DB connection for up to `RENDER_TIMEOUT` (90s) — a retrying agent at the burst limit can pin a lot of workers. A dedicated low-rate throttle (cf. `SubscriptionTestDeliveryThrottle`, 10/min) fits render work better.

products/tasks/backend/presentation/serializers.py
```diff
@@ -954,6 +955,46 @@ def validate(self, attrs):
...
+
+    def validate(self, attrs):
+        if (attrs.get("query") is None) == (attrs.get("insight_id") is None):
+            raise serializers.ValidationError({"query": "Provide exactly one of query or insight_id."})
+        return attrs
+
+
+class TaskRunLivingArtifactChartResponseSerializer(serializers.Serializer):
+    artifact = TaskRunLivingArtifactResponseSerializer(help_text="The living artifact registered for delivery.")
+    export_asset_id = serializers.IntegerField(help_text="Id of the rendered PNG export backing the chart.")
```
sampennington, line 990: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion / follow-up**

`export_asset_id` has no consumer — the agent prompt never mentions it, the relay doesn't read it, no frontend uses it — and `SlackFileDeliveryResult.delivered_count` is incremented but never read or logged (only asserted in a test). Dropping both shrinks a public response contract you'd otherwise have to keep supporting.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
```diff
@@ -336,6 +352,130 @@ def test_pending_slack_file_upload_uses_final_message(
...
+        self.assertEqual(button["url"], chart_url)
+        self.assertEqual(button["text"]["text"], "Open in PostHog")
+
+        # The composed message carries the answer text — nothing posted via the handler.
+        mock_post.assert_not_called()
+
+        artifact.refresh_from_db()
+        self.assertEqual(artifact.location["delivery_status"], "delivered")
+        self.assertNotIn("slack_file_id", artifact.versions[0])
+        self.assertEqual(artifact.versions[0]["delivery_status"], "delivered")
```
sampennington, line 432: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

The riskiest new bookkeeping has no coverage: partial delivery across multiple cards (one posts, one fails → only the posted one marked), the >50-block spill path, and the `elif answer_sections is not None` all-uploads-failed fallback. A regression that marks undelivered cards delivered — losing the chart permanently — would pass.

products/tasks/backend/tests/test_api.py
```diff
@@ -8008,6 +8008,133 @@ def test_scoped_api_key_permissions(self, scope, method, url_template, should_ha
...
+    )
+    def test_requires_exactly_one_of_query_or_insight_id(self, _name, body):
+        from products.tasks.backend.presentation.serializers import TaskRunLivingArtifactChartRequestSerializer
+
+        serializer = TaskRunLivingArtifactChartRequestSerializer(data=body)
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("query", serializer.errors)
+
+
+class TestTaskRunLivingArtifactChartAPI(BaseTaskAPITest):
```
sampennington, line 8025: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

The `insight_id` half of the request contract has no endpoint test: nothing covers that it's forwarded with `export_context=None`, that `_chart_url` yields `/insights/{short_id}`, or the `url: null` case when the asset has no insight.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -784,69 +785,128 @@ def has_pending_slack_file_artifacts(run: TaskRun) -> bool:
...
+        content_type = _pending_slack_content_type(artifact, pending[1])
+        if content_type.startswith("image/"):
+            return True
+    return False
+
+
+def deliver_pending_slack_file_artifacts(
+    run: TaskRun, *, answer_sections: list[str] | None = None
+) -> SlackFileDeliveryResult:
+    """Deliver pending slack_file artifacts to the mapped thread.
```
sampennington, line 843: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

Twelve-line docstring; only "charts post by public url, no files:write" and "chat.postMessage is synchronous unlike channel shares" are load-bearing — cut the composition walkthrough and the `answer_posted` restatement. Same trim applies to the docstrings at 991 and 1174 (both restate their signatures) and the comments at `views/api.py:2401` (duplicates the `task_run_exists` docstring) and `activities.py:411` (repeats the 3000-char rationale already on `SLACK_SECTION_TEXT_LIMIT`).

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
```diff
@@ -336,6 +352,130 @@ def test_pending_slack_file_upload_uses_final_message(
...
+            content_type="image/png",
+            metadata={"posthog_url": chart_url, "image_url": image_url},
+        )
+        slack = unittest.mock.MagicMock()
+        slack_integration = unittest.mock.MagicMock()
+        slack_integration.client = slack
+        # No files:write — url-referenced charts must deliver without any file scope.
+        slack_integration.missing_scopes.return_value = {"files:write"}
+        mock_integration_for_mapping.return_value = slack_integration
+        # First post rejected transiently to exercise the retry.
```
sampennington, line 394: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

This comment narrates the `side_effect` line below it. Same at 407 and 423 (both restate the `assert_not_called()` calls directly beneath) and 460 (restates the test name). Delete all four.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Confirmed — `answer_posted` was `bool(sections)` regardless of whether `_post_thread_text` actually succeeded, so a silently-swallowed post failure still told the relay the answer went out and skipped its fallback. Fixed locally: `_post_thread_text` now returns whether the post succeeded, and both call sites (the degraded path in `_post_composed_answer_message` and the all-uploads-failed branch in `deliver_pending_slack_file_artifacts`) AND that into `answer_posted`.

This also covers the duplicate finding from the greptile review on the same line.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Confirmed as a real 500: `quote(json.dumps(query))` emits `%5C` for any escaped character (non-ASCII values via `ensure_ascii`, embedded quotes), and `absolute_uri` rejects `%5c` as an authority-bypass pattern — so any chart on a query containing a non-ASCII event/property value 500s after the render already succeeded. Fixed locally by matching `_deep_link`'s pattern (`products/data_catalog/backend/logic/execution.py:200-212`): `absolute_uri` now only resolves the literal path, and the encoded query payload is appended after. Added a regression test posting a query with a non-ASCII series name and asserting 200 — reverting the fix reproduces the 500.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/tests/test_api.py
sampennington: 🤖 **sp-triage**

Confirmed and fixed locally: added `mock_render.assert_called_once_with(...)` pinning `team`, `created_by`, `insight_id`, and the `export_context` shape. While adding it, discovered the assertion can't hardcode `export_context={"source": self.CHART_QUERY}` — `upgrade(query)` (called earlier for validation) mutates the query dict in place, so the query actually passed to `render_png_export` differs from the original by an injected `version` key. That's a real, separate bug (validated value discarded, mutated raw value rendered) tracked at `views/api.py:2423` as its own follow-up — the new assertion checks the call's actual shape instead of coupling to that quirk.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/logic/services/tests/test_living_artifacts.py
sampennington: 🤖 **sp-triage**

Fixed locally: moved the expected block-type lists (`["section", "image", "actions"]` / `["section", "image"]`) into the `parameterized.expand` cases instead of a ternary in the assertion. Also moved the `_chart_card_blocks`/`_section_blocks` imports to module level per the repo's import convention.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/tests/test_api.py
sampennington: 🤖 **sp-triage**

Fixed locally: `run_id` is now a required kwarg on `_post_chart` (no `run.id` fallback branch). Split run creation into a `_create_run` helper so every caller creates the run and passes its id explicitly.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
sampennington: 🤖 **sp-triage**

Fixed locally: `metadata` is now a required kwarg on `_create_pending_slack_file_artifact` (`**(metadata or {})` removed). The one caller that previously relied on the default (`test_pending_slack_file_upload_posts_text_then_file`) now passes `metadata={}` explicitly.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/presentation/views/api.py
```diff
@@ -2359,6 +2369,120 @@ def create(self, request, *args, **kwargs):
...
+            except (pydantic.ValidationError, TypeError, KeyError, ValueError):
+                return Response(
+                    TaskRunErrorResponseSerializer({"error": "Invalid insight query"}).data,
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+        if not isinstance(request.user, User):
+            # Scoped-token auth guarantees a real user; the facade's access checks need one.
+            raise NotAuthenticated()
+        try:
+            asset, png = render_png_export(
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed — `required_scopes=["task:write", "query:read"]` gates token auth, but `AccessControlPermission` only checks the `task` resource, and session-authenticated users carry no scopes at all. A task collaborator without query-viewer RBAC could render and receive chart data through this endpoint. Fixed locally: added an explicit `self.user_access_control.check_access_level_for_resource("query", "viewer")` gate before calling `render_png_export`, matching the existing pattern in `products/data_catalog/backend/presentation/views.py:191` and `ee/api/subscription.py:1041`.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed — `required_scopes=["task:write", "query:read"]` gates token auth, but `AccessControlPermission` only checks the `task` resource, and session-authenticated users carry no scopes at all. A task collaborator without query-viewer RBAC could render and receive chart data through this endpoint. Fixed locally: added an explicit `self.user_access_control.check_access_level_for_resource("query", "viewer")` gate before calling `render_png_export`, matching the existing pattern in `products/data_catalog/backend/presentation/views.py:191` and `ee/api/subscription.py:1041`.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+def _post_blocks_with_processing_retry(
+    slack: Any,
+    *,
+    channel: str,
+    thread_ts: str,
+    text: str,
+    blocks: list[dict[str, Any]],
+    attempts: int = _IMAGE_BLOCK_POST_ATTEMPTS,
+) -> None:
+    for attempt in range(1, attempts + 1):
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed — same finding as the sp-swarm review on this line. Fixed locally: `_post_thread_text` now returns whether the Slack post succeeded, and `answer_posted` is derived from that instead of `bool(sections)` alone, so a swallowed post failure no longer gets reported as delivered.

Push is currently blocked in this session by an unrelated 1Password SSH-signing failure — the fix is staged locally and will land on the next push to this branch. Leaving this thread open until then.

products/tasks/backend/logic/services/living_artifacts.py
```diff
@@ -855,22 +915,227 @@ def deliver_pending_slack_file_artifacts(run: TaskRun, *, initial_comment: str)
...
+    if isinstance(image_url, str) and image_url:
+        image_block: dict[str, Any] = {"type": "image", "image_url": image_url, "alt_text": artifact.name}
+    else:
+        image_block = {"type": "image", "slack_file": {"id": file_id}, "alt_text": artifact.name}
+    blocks: list[dict[str, Any]] = [
+        {"type": "section", "text": {"type": "mrkdwn", "text": f"*{_escape_slack_mrkdwn_text(artifact.name)}*"}},
+        image_block,
+    ]
+    posthog_url = metadata.get("posthog_url")
+    if isinstance(posthog_url, str) and 0 < len(posthog_url) <= _SLACK_BUTTON_URL_MAX_CHARS:
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed real — `posthog_url` in artifact metadata is settable through the generic living-artifact create/edit endpoints, so a task writer could point the "Open in PostHog" button at an external site. Deferring to a human: the fix needs a design decision (validate the parsed URL's scheme+host against the configured PostHog origin at render time, vs. deriving the URL server-side from trusted chart state and dropping caller-supplied `posthog_url` entirely) that's broader than this PR's scope. Leaving this thread open.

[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Confirmed real — `posthog_url` in artifact metadata is settable through the generic living-artifact create/edit endpoints, so a task writer could point the "Open in PostHog" button at an external site. Deferring to a human: the fix needs a design decision (validate the parsed URL's scheme+host against the configured PostHog origin at render time, vs. deriving the URL server-side from trusted chart state and dropping caller-supplied `posthog_url` entirely) that's broader than this PR's scope. Leaving this thread open.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Confirmed real: `SlackFileArtifactAdapter.commit` still hard-raises without `files:write`, so with the flag on but `files:write` not yet approved, every chart call would 400. Deferring to a human: gating this check on url-delivered image content vs. file-upload content is a design decision about the adapter's commit path and the rollout sequencing this PR targets — out of scope for an automated fix. Leaving this thread open.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Confirmed real: worst-case retry sleep can exceed the relay activity's 60s `start_to_close_timeout`, and since cards are only marked delivered after the whole composed post returns, a self-timeout replay can repost the answer and duplicate charts. Deferring to a human: fixing this needs a wall-clock deadline tracked across attempts plus reordering when each card is marked delivered relative to the retry loop — a design change to the retry/delivery-marking contract, not a localized edit. Leaving this thread open.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Confirmed this is a live bug — while fixing the adjacent test-coverage finding (`test_api.py:8085`) I hit this directly: `upgrade(query)` mutates the query dict in place (injects a `version` key into the nested source), so the "validated" shape and the actually-rendered/embedded shape already diverge today. Deferring the fix rather than folding it into this triage pass: binding `upgraded = upgrade(query)` and passing that through changes what gets rendered and what the "Open in PostHog" deep link encodes, which is a small behavior change worth its own review rather than a silent side-fix. Leaving this thread open as a tracked follow-up.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid reuse follow-up — collapsing `has_pending_slack_file_artifacts` and `deliver_pending_slack_file_artifacts` onto one `_pending_slack_files(run)` helper (and picking up the `_living_artifacts_enabled_for_run` gate) is a real simplification, but it's a cross-cutting change to a hot path this PR doesn't need to touch to ship. Resolving as out-of-scope follow-up rather than folding into this triage pass.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid separation-of-concerns follow-up — splitting file delivery from answer-posting would remove `answer_sections`/`answer_posted`/`_post_thread_text` in one move, but it's a meaningful API reshape of this function touching the relay's call site too. Resolving as out-of-scope follow-up rather than folding into this triage pass.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid readability follow-up — a `NamedTuple`/dataclass would make the `_SlackImageCard` slots self-describing instead of positional unpacks. Cosmetic, no behavior change, and touches several call sites — resolving as out-of-scope follow-up rather than folding into this triage pass.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid — the 2900-char budget is duplicated under two names (here and `SLACK_SECTION_TEXT_LIMIT` in `slack_relay/activities.py:245`) with nothing enforcing they agree. Worth exporting one constant, but it's a cross-module change outside this triage pass's localized-fix scope. Resolving as out-of-scope follow-up.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid — the degraded path currently triggers on any exception from the composed post, including cases where Slack actually accepted the message (e.g. an SDK-side read timeout), so the fallback can duplicate an already-delivered answer. Narrowing this needs a decision about which exceptions prove nothing was delivered — a behavior change to the at-least-once guarantee, not a mechanical fix. Resolving as out-of-scope follow-up for this triage pass.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid reuse note — this duplicates `_block_for_asset` (`ee/tasks/subscriptions/slack_subscriptions.py:79`), which additionally swaps in a fetchable url under `settings.DEBUG`. Worth consolidating, but out of scope for this triage pass (cross-module reuse, not a bug in this PR's new code path in prod). Resolving as follow-up.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid reuse note — this re-implements Slack retry logic that already exists in `_send_slack_message_with_retry`/`_RETRYABLE_SLACK_ERRORS` (`ee/tasks/subscriptions/slack_subscriptions.py:34,286`) and `_slack_retry_after_seconds` (`products/signals/backend/tasks.py:84`), the latter also handling the lowercase `retry-after` header this copy misses. Worth consolidating, but a cross-module refactor outside this triage pass's scope. Resolving as follow-up.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Valid readability follow-up — extracting `_validated_chart_query(query)` and `_register_chart_artifact(...)` would flatten `chart()`'s mixed abstraction levels. A worthwhile refactor, but a larger structural change than this triage pass's localized-fix scope. Resolving as follow-up.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Valid concern — `get_subscription_delivery_content_url()`'s default 365-day unauthenticated JWT is persisted in artifact metadata and returned by the living-artifacts list/retrieve endpoints, giving anything that can read the artifact a long-lived public link. Deferring rather than picking a number: choosing an `expiry_delta` requires knowing the living-artifact's actual retention/TTL policy, which is a product decision outside this triage pass. Leaving this thread open for a human to size the expiry against actual artifact lifetime.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Valid concern — `asset.exception` is the raw `str(e)` from the export pipeline, and the new agent prompt tells the agent to relay the `error` field into Slack, so ClickHouse/headless-browser internals could leak into a channel. Deferring: choosing what a "short user-facing reason" should say (and which raw-error patterns map to it) is a product-copy decision, not a mechanical fix — see `/writing-user-facing-copy` if this is picked up. Leaving this thread open.

products/tasks/backend/presentation/views/api.py
sampennington: 🤖 **sp-triage**

Valid — `ClickHouseBurstRateThrottle` (240/min) is calibrated for sub-second queries, not a render that can hold a worker for up to `RENDER_TIMEOUT` (90s); a dedicated low-rate throttle (cf. `SubscriptionTestDeliveryThrottle`, 10/min) fits better. Resolving as follow-up — a rate-limit tuning change, not a bug in this PR's behavior, and picking the right number is a capacity/product call outside this triage pass.

products/tasks/backend/presentation/serializers.py
sampennington: 🤖 **sp-triage**

Valid — `export_asset_id` has no consumer (prompt, relay, or frontend) and `SlackFileDeliveryResult.delivered_count` is only ever asserted in tests, never read or logged. Dropping both would shrink the public response contract. Resolving as follow-up — a scope-shrinking cleanup, not a bug, and safe to defer past this PR.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
sampennington: 🤖 **sp-triage**

Valid coverage gap — partial delivery across multiple cards, the >50-block spill path, and the all-uploads-failed fallback are the riskiest new bookkeeping and currently untested; a regression marking undelivered cards delivered would pass today. Resolving as a follow-up rather than adding tests in this pass — writing these properly needs care per `/writing-tests` (parameterized cases, no change-detector mocking of the composition internals), better done as a focused addition than folded into a triage pass.

products/tasks/backend/tests/test_api.py
sampennington: 🤖 **sp-triage**

Valid coverage gap — the `insight_id` half of the request contract (forwarded with `export_context=None`, `_chart_url` yielding `/insights/{short_id}`, and the `url: null` no-insight case) has no endpoint test. Resolving as a follow-up rather than adding it in this pass — lower risk than the partial-delivery gap (this path is a straightforward pass-through), better done as a focused addition.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: 🤖 **sp-triage**

Valid — per `/writing-code-comments`, only the load-bearing lines ("charts post by public url, no files:write", "chat.postMessage is synchronous unlike channel shares") need to stay; the composition walkthrough and restated signature are prose padding. Resolving as a follow-up cleanup rather than editing comments in this triage pass — low risk, but touches several unrelated spots (also 991, 1174, `views/api.py:2401`, `activities.py:411`) better done as one deliberate pass.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
sampennington: 🤖 **sp-triage**

Valid — these comments narrate the line directly beneath them (the `side_effect` setup, the `assert_not_called()` calls, the test name at 460) and don't explain a non-obvious "why". Resolving as a follow-up cleanup rather than editing in this triage pass, alongside the docstring trims flagged in the sibling thread at `living_artifacts.py:808`.

products/tasks/backend/logic/services/living_artifacts.py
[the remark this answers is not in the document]
sampennington: Fixed in 6629b428. `_post_thread_text` now returns whether the post landed and callers OR the results, so `answer_posted` can't be true when every text post was swallowed. Same root cause as the sp-ship blocker on this file.

products/tasks/backend/presentation/views/api.py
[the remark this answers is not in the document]
sampennington: Fixed in 6629b428. The ad-hoc path now requires `check_access_level_for_resource("query", "viewer")` before rendering, and the action declares a 403. Your reasoning held up: the scopes only gate tokens, session auth skips them entirely, and `render_png_export` only does object-level checks for `insight_id`.

products/tasks/backend/logic/services/living_artifacts.py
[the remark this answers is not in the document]
sampennington: Fixed in 6629b428. The button url is validated against the configured PostHog origin (scheme + host) and skipped with a warning otherwise, plus a parameterized case asserting an off-origin url produces no button. Deriving it server-side rather than from metadata is the fuller fix; the origin check closes the vector without reshaping how the endpoint and delivery share state.

products/tasks/backend/presentation/views/api.py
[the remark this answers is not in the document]
sampennington: Fixed in 6629b428 — same finding as the veria-ai thread on this line. `check_access_level_for_resource("query", "viewer")` now gates the ad-hoc path before any render.

products/tasks/backend/logic/services/living_artifacts.py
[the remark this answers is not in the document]
sampennington: Fixed in 6629b428 — same finding as the veria-ai thread on this line. The url is validated against the PostHog origin before it can become a button.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428. Confirmed by reading `commit` — the relaxation was only applied to delivery via `has_file_scope`, so with the flag on and `files:write` unapproved every chart call 400'd. The creation-side check is now gated on non-image content: images deliver as blocks referencing a public url, and delivery already skips any image it can't reference. Added `test_slack_file_adapter_allows_images_without_file_scope` so re-tightening it can't silently break every chart again.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428, reusing the `_deep_link` shape as suggested — the payload is appended after `absolute_uri` resolves the path. Verified the trap empirically before changing anything: a `Café ☕` event name yields `%5Cu00e9`, which `has_authority_bypass_chars` rejects, so it 500'd after the render had already succeeded.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: Fixed in 6629b428. All posting now runs against a wall-clock deadline (`_SLACK_POST_BUDGET_S = 45.0`) checked before each sleep, so a rate-limited retry raises instead of running past the activity's 1-minute `start_to_close_timeout`. Cards are marked delivered via a callback as soon as each post lands, so a replay can't repost ones that already went out.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: Fixed in 6629b428. `_post_thread_text` returns success and callers OR it, so the "answer must not be lost" comment now holds. You were right that `test_failed_chart_post_leaves_artifact_pending` passes straight through this — it only asserts artifact bookkeeping.

products/tasks/backend/tests/test_api.py
sampennington: Fixed in 6629b428 — added exactly that assertion. Worth it for the reason you gave: the facade rejects any `export_context` whose `source` isn't an `InsightVizNode`, so a wrong wrapping would 400 every real request with the existing assertions still green.

products/tasks/backend/logic/services/tests/test_living_artifacts.py
sampennington: Fixed in 6629b428. The expected block lists moved into the `parameterized.expand` cases (with a third for the off-origin url), and both function-level imports are hoisted to the module-level import block.

products/tasks/backend/tests/test_api.py
sampennington: Fixed in 6629b428 by splitting the helper in two: `_post_chart` creates the task and run, `_post_chart_to_run` takes them explicitly. I first tried just dropping the parameter after grepping for callers and concluding none passed it — mypy caught that the unknown-run test does, on a continuation line the grep missed.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
sampennington: Fixed in 6629b428 — `metadata` is a required kwarg and the xlsx caller passes `{}`.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428 — the url is minted with `expiry_delta=CHART_IMAGE_URL_TTL` (31 days), scoped to the artifact's own 30-day storage TTL rather than the 365-day default.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428. The 400 now returns a flat "Chart render failed" and `asset.exception` goes to the log, so pipeline internals can't be relayed into a Slack channel.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428 — `query = upgrade(query)` is bound and that node is what gets validated, rendered, and deep-linked.

products/tasks/backend/presentation/views/api.py
sampennington: Fixed in 6629b428 — added `TaskRunChartRenderThrottle` (10/minute, keyed per team) and dropped the two ClickHouse throttles from this action.

products/tasks/backend/temporal/slack_relay/tests/test_activities.py
sampennington: Partly addressed in 6629b428, and I want to be straight about what isn't. The per-card marking this asked about is now structural rather than test-covered: cards are marked via a callback as each post lands, so the "marks undelivered cards delivered" shape is much harder to reintroduce. But I did not add the three cases you named — partial delivery across cards, the >50-block spill path, and the `elif answer_sections is not None` all-uploads-failed fallback. They're DB-backed and I have no Postgres in this environment, so I couldn't run them before pushing and won't claim coverage I haven't executed. Still a genuine gap.

products/tasks/backend/logic/services/living_artifacts.py
sampennington: Not fixed in 6629b428 — flagging rather than silently skipping. The degraded path still triggers on any exception from the composed post, including a read timeout where Slack accepted the message, so that case still duplicates the answer. Narrowing it needs a way to distinguish "provably not delivered" from "unknown", which is a real change to the retry contract rather than a tweak; the deadline work in this commit reduces how often the path is entered but doesn't address the at-least-once side effect you identified.

sampennington:
Closing in favor of a 3-PR stack split off this branch for reviewability (this PR had grown past easy review and the comment history got noisy). The stack's tip is byte-identical to this branch's tip, including all review fixes from here:

1. [#75259](https://github.com/PostHog/posthog/pull/75259) — chart render endpoint
2. [#75260](https://github.com/PostHog/posthog/pull/75260) — composed Slack delivery
3. [#75261](https://github.com/PostHog/posthog/pull/75261) — agent prompt exposure

PR #71157, opened 2026-07-15
feat(data-warehouse): implement kubecost import source

Gilbert09:
## Problem

The Kubecost source was scaffolded in #71137 as a hidden stub with no sync logic. Platform and FinOps teams use Kubecost to attribute Kubernetes and cloud spend to namespaces, controllers, and pods, and want that data in the warehouse for chargeback, showback, and efficiency tracking – a k8s cost angle the cloud-billing sources don't cover.

## Why

Fill in the scaffolded Kubecost stub so users can actually connect their Kubecost deployment and sync cost data, shipping it visible as an alpha source.

## Changes

Implements the Kubecost source end-to-end, following the `source.py` / `settings.py` / `kubecost.py` architecture contract (modeled on the Matomo source, the closest self-hosted analog):

- **Tables:** `allocation_by_namespace`, `allocation_by_controller`, `allocation_by_pod` (Allocation API) and `assets` (Assets API). Result sets are flattened into one row per allocation/asset per day, with injected `key`, `window_start`, and `window_end` columns – composite primary key `(key, window_start)` since names are only unique within a window.
- **Sync model:** Kubecost has no pagination or updated-since cursor. The `window` query param is the server-side time filter, so the source walks contiguous one-day windows oldest-first (`ResumableSource`, resume state = next unfetched day, saved after each yielded day). Incremental syncs restart from the watermark minus a 3-day lookback because cloud-billing reconciliation restates recent days. First sync backfills 90 days, days beyond the deployment's ETL retention come back empty and are skipped.
- **Config:** user-supplied base URL (self-hosted deployments must expose the cost-model API, `/model` suffix normalized away) plus an optional API key sent as a bearer token, since self-hosted Kubecost ships with no built-in auth. The host goes through `is_database_host_valid` at validate and sync time, redirects are pinned off, and `connection_host_fields` forces key re-entry when the URL is retargeted.
- **Everything else:** all HTTP via `make_tracked_session()`, tenacity retries on 429/5xx/timeouts, non-retryable 401/403 mapping, `releaseStatus=ALPHA` with `unreleasedSource` removed, `lists_tables_without_credentials` for public docs, canonical table/column descriptions from the official API docs, SOURCES.md moved to Implemented.

> [!NOTE]
> Endpoint behavior was verified against the public Kubecost demo instance (response envelope, RFC3339 window filtering, aggregation params, and the `data: [null]` shape for empty/out-of-retention windows – the latter is explicitly handled and tested). Bearer-token auth couldn't be exercised against a real authenticated deployment, so it's the conservative pass-through header noted in the code.

The user-facing doc (`contents/docs/cdp/sources/kubecost.md`) is written per the docs template and `audit_source_docs` passes for kubecost against a local posthog.com checkout, but it needs to land in the posthog.com repo separately:

<details>
<summary>kubecost.md for posthog.com</summary>

```markdown
---
title: Linking Kubecost as a source
sidebar: Docs
showTitle: true
availability:
  free: full
  selfServe: full
  enterprise: full
sourceId: Kubecost
---

<AlphaRelease />

The Kubecost connector syncs Kubernetes cost data from your Kubecost (IBM / Apptio) deployment into PostHog – workload cost and usage allocation by namespace, controller, and pod, plus infrastructure asset costs like nodes, disks, and load balancers. This is useful for chargeback, showback, and efficiency tracking alongside your product data.

## Prerequisites

Kubecost runs inside your Kubernetes cluster, so its cost-model API must be reachable from PostHog:

- Expose the Kubecost API (served under `/model`, e.g. `http://<kubecost-address>/model/allocation`) through an ingress or load balancer.
- Self-hosted Kubecost ships with no built-in authentication, so we strongly recommend putting an authenticating proxy in front of it. If the proxy (or Kubecost Cloud) requires a token, the connector sends it as a bearer token.

## Adding a data source

<SourceSetupIntro />

When linking Kubecost, you'll need:

- **Kubecost API URL** – the address where the cost-model API is reachable, for example `https://kubecost.example.com`. Entering it with or without the `/model` suffix both work.
- **API key** (optional) – a token accepted by your deployment or the proxy in front of it, sent as `Authorization: Bearer <key>`. Providing a key requires an `https://` URL so it is never sent in plaintext. Leave empty if your endpoint is unauthenticated.

## Sync modes

<SyncModes />

The first sync backfills up to 90 days of history, bounded by your deployment's configured ETL retention – days beyond retention are skipped. Data is pulled one day at a time, and incremental syncs re-pull a 3-day trailing window so costs restated by cloud-billing reconciliation are picked up. Rows are merged on the allocation/asset key and the day's window start.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

## Troubleshooting

- **"Unable to reach the Kubecost API"** – check that the URL is correct and that the endpoint is reachable from the public internet (an in-cluster address like `http://kubecost-cost-analyzer.kubecost:9090` won't work).
- **Tables sync but older history is missing** – Kubecost only serves data within its configured ETL retention. Increase the retention in your Kubecost deployment if you need a longer backfill.

<TroubleshootingLink />
```

</details>

- `hogli test products/warehouse_sources/backend/temporal/data_imports/sources/kubecost/tests` – 50 tests pass. `test_kubecost.py` guards the transport regressions no existing test covers: the `data: [null]` empty-window shape (would crash flattening), composite-key column injection (missing `window_start` would break delta merges), watermark-minus-lookback window math, resume-state saved after (not before) each yielded day, `/model` suffix normalization (would produce `/model/model` URLs), and per-endpoint aggregate params. `test_kubecost_source.py` guards credential validation (SSRF host rejection, invalid URL), schema catalog, and pipeline argument plumbing.
- Registry-wide suites (`test_source_categories.py`, `test_source_config_generator.py`) pass.
- `ruff check` / `ruff format` clean, `hogli ci:preflight --fix` reports no failures.
- Live-API behavior (envelope, window filtering, aggregation, null sets) confirmed with curl against the public Kubecost demo. I did not test against an authenticated Kubecost Cloud deployment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc written for posthog.com (see collapsed section above) – needs a companion PR in that repo since this one can't touch it.

## 🤖 Agent context

- Authored with Claude Code (PostHog Code task). Skills invoked: `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, `/writing-tests`.
- Modeled on the Matomo source as the closest analog (self-hosted instance URL + token, day-window walk, `ResumableSource`). Chose a custom `requests` transport over the declarative `rest_source.RESTClient` because Kubecost has no pagination at all – the paginator abstraction adds nothing when the loop is a time-window walk.
- Chose fixed aggregation levels as separate tables (namespace/controller/pod) instead of a user-configurable aggregate field, so table schemas stay stable and primary keys stay meaningful.
- Verified endpoint behavior against the public Kubecost demo before finalizing: confirmed RFC3339 window pairs filter server-side, `accumulate=true` returns one result set per request, and empty/out-of-retention windows return `data: [null]`.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -259,6 +259,7 @@ the row lists both.
 | klaviyo                   | HTTP                        | requests                                                        | ✅                          |
 | koyeb                     | HTTP                        | requests                                                        | ✅                          |
 | kong_konnect              | HTTP                        | requests                                                        | ✅                          |
+| kubecost                  | HTTP                        | requests                                                        | ✅                          |
 | lago                      | HTTP                        | requests                                                        | ✅                          |
 | lambda_labs               | HTTP                        | requests                                                        | ✅                          |
 | langfuse                  | HTTP                        | requests                                                        | ✅                          |
@@ -722,7 +723,6 @@ doesn't conflict with concurrent PRs.
 - klaus
 - knock
 - koyeb
-- kubecost
 - kyve
 - lacework
 - lambda_labs
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
@@ -2291,7 +2291,8 @@ class KoyebSourceConfig(config.Config):
 
 @config.config
 class KubecostSourceConfig(config.Config):
-    pass
+    host: str
+    api_key: str | None = None
 
 
 @config.config
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/kubecost/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/kubecost/canonical_descriptions.py
@@ -1,0 +1,85 @@
+"""Canonical, documentation-sourced descriptions for Kubecost endpoints and columns.
+
+Sourced from the official Kubecost API reference (https://docs.kubecost.com/apis).
+Keyed by the endpoint names in `settings.py` `KUBECOST_ENDPOINTS`, which match the
+`ExternalDataSchema.name` of a synced Kubecost table. Columns absent here fall back to LLM enrichment.
+"""
+
+from products.warehouse_sources.backend.temporal.data_imports.sources.common.canonical_descriptions import (
+    CanonicalDescriptions,
+)
+
+_ALLOCATION_DOCS_URL = "https://docs.kubecost.com/apis/monitoring-apis/api-allocation"
+
+_ALLOCATION_COLUMNS = {
+    "key": "The allocation's name in the queried aggregation (with special entries like __idle__ for unallocated cluster capacity). Unique per window.",
+    "window_start": "Start of the one-day query window this row covers (RFC3339). Used as the incremental cursor and partition key.",
+    "window_end": "End of the one-day query window this row covers (RFC3339).",
+    "name": "The name of the allocation, matching the aggregation level (namespace, controller, or pod).",
+    "properties": "Kubernetes metadata for the allocation: cluster, node, namespace, controller, pod, container, labels, and annotations.",
+    "window": "The queried time window the allocation was computed over, with start and end timestamps.",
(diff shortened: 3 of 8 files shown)


danielcarletti (approved):
LGTM

PR #71178, opened 2026-07-15
feat(data-warehouse): implement opsgenie import source

Gilbert09:
## Problem

The Opsgenie warehouse source existed only as a scaffolded stub (`unreleasedSource=True`, no fields, no sync logic), so users couldn't pull their Opsgenie alerting and on-call data into the PostHog data warehouse.

**Why:** fill in the scaffolded Opsgenie connector end-to-end so it ships visible and connectable, per the warehouse-sources implementation workflow.

## Changes

Implements Opsgenie as a `ResumableSource` following the standard `source.py` / `settings.py` / `opsgenie.py` split (PagerDuty is the closest in-repo analog and served as the reference):

- **Endpoints:** `alerts`, `incidents`, `users`, `teams`, `schedules`, `escalations`, `services`, `integrations` — matching the streams Airbyte/Fivetran cover for Opsgenie.
- **Auth + region:** `GenieKey` API-key header, with a US/EU region select that routes to `api.opsgenie.com` or `api.eu.opsgenie.com`.
- **Pagination:** limit-offset (max 100/page), terminating on the response's `paging.next` link; resumable via offset checkpoints saved after each yielded batch.
- **Incremental sync:** `alerts` and `incidents` only, via the search `query=createdAt >= <epoch ms>` filter with `sort=createdAt&order=asc`. Opsgenie's search syntax can't filter on `updatedAt`, so these are create-time cursors (same append-style limitation as other connectors for this API); everything else is full refresh.
- **20,000-result search cap:** the alert/incident search rejects `offset + limit > 20000`, so instead of truncating, pagination re-slices into a new `createdAt` window anchored on the last row read (with a guard against a non-advancing window).
- **Everything else per the skill:** tracked HTTP transport, tenacity backoff on 429/5xx, non-retryable 401/403 errors for both regional hosts, 403 accepted at source-create (a key may lack Configuration access for `integrations`), canonical table/column descriptions, `lists_tables_without_credentials` for public docs, `releaseStatus=ALPHA` with `unreleasedSource` removed.

Also regenerates `OpsgenieSourceConfig`, moves opsgenie into the Implemented table in `SOURCES.md`, and adds the user-facing doc in [PostHog/posthog.com#18522](https://github.com/PostHog/posthog.com/pull/18522) (matching `docsUrl`; `audit_source_docs` reports no issues for this source).

> [!NOTE]
> Live-API verification was limited to unauthenticated probes (host reachability, 401/422 auth status codes) since no Opsgenie credentials were available. The `query`/`sort` filtering behavior and the v1 paths for incidents/services follow the current API docs and the Airbyte/Fivetran connectors; code comments note the `>=` boundary re-fetch (deduped by merge on `id`). Worth a smoke test with a real account before promoting past alpha.

Automated tests only (no Opsgenie account available for manual verification):

- `tests/test_opsgenie.py` (transport): the 20k-cap re-windowing (offset resets, query re-anchors on the last row's `createdAt`, state checkpointed) and the non-advancing-window guard — regressions here silently truncate backfills or loop forever; resume-from-saved-state, save-after-yield ordering, pagination termination, region→host routing, credential status mapping incl. Opsgenie's 422 for malformed keys, and 429 retry.
- `tests/test_opsgenie_source.py` (source): schema catalog (incremental only where the API has a server-side filter), 403 accepted at create but rejected per-schema, resume-config binding, and pipeline argument plumbing.

All 60 pass, plus the 1,555 registry-wide source tests (categories, generated configs). `ruff check`/`format` and `hogli ci:preflight --fix` clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc added in [PostHog/posthog.com#18522](https://github.com/PostHog/posthog.com/pull/18522).

## 🤖 Agent context

Authored with Claude Code (PostHog Code task). Skills invoked: `implementing-warehouse-sources`, `documenting-warehouse-sources`, `writing-tests`. Key decisions: used PagerDuty as the structural reference (same limit-offset shape and product domain); chose `requests` + tracked session over the generic `rest_source.RESTClient` because the 20k-cap windowing needs custom pagination state; skipped per-alert fan-out endpoints (alert logs/recipients) since they have no server-side time filter and would cost 2 API calls per alert per sync — can be added later if requested; omitted unverifiable `sort` params on full-refresh endpoints rather than risk 422s. Reverted an env-dependent Stripe hunk the config generator produced locally so the diff stays scoped to opsgenie.

Note: Atlassian has announced Opsgenie's end of support — its APIs shut down April 5, 2027, so this connector has a bounded lifespan (called out in the source caption and doc).

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -331,6 +331,7 @@ the row lists both.
 | openfda                          | HTTP                        | requests                                                        | ✅                          |
 | openrouter                       | HTTP                        | requests                                                        | ✅                          |
 | openweather                      | HTTP                        | requests                                                        | ✅                          |
+| opsgenie                         | HTTP                        | requests                                                        | ✅                          |
 | ortto                            | HTTP                        | requests                                                        | ✅                          |
 | oura                             | HTTP                        | requests                                                        | ✅                          |
 | outbrain                         | HTTP                        | requests                                                        | ✅                          |
@@ -789,7 +790,6 @@ doesn't conflict with concurrent PRs.
 - onepassword
 - onesignal
 - open_data_dc
-- opsgenie
 - opuswatch
 - oracle
 - oracle_ebs
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
@@ -2950,7 +2950,8 @@ class OpinionStageSourceConfig(config.Config):
 
 @config.config
 class OpsgenieSourceConfig(config.Config):
-    pass
+    api_key: str
+    region: Literal["us", "eu"] = config.value(default="us")
 
 
 @config.config
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/opsgenie/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/opsgenie/canonical_descriptions.py
@@ -1,0 +1,129 @@
+"""Canonical, documentation-sourced descriptions for Opsgenie endpoints and columns.
+
+Sourced from the official Opsgenie REST API reference (https://docs.opsgenie.com/docs/api-overview).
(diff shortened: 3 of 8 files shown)


danielcarletti (approved):
LGTM

PR #71186, opened 2026-07-15
feat(data-warehouse): implement uservoice import source

Gilbert09:
## Problem

PostHog can import product analytics data from many SaaS tools, but not from [UserVoice](https://uservoice.com) — a product-feedback, idea-management, and customer-support platform. Teams that collect feature requests, NPS, and helpdesk tickets in UserVoice had no way to join that qualitative feedback with how users actually behave in their product.

This adds UserVoice as a data warehouse source so its feedback and support data lands alongside the rest of a team's data.

## Changes

Implements the UserVoice Admin API v2 source end-to-end following the `implementing-warehouse-sources` skill and the `Aha!` source as the closest reference (subdomain + Bearer token, page/cursor pagination, `updated_after` incremental filter).

- **`source.py`** – `UservoiceSource(ResumableSource)`: form fields (subdomain + API access token), schema catalog, credential validation, non-retryable auth errors, resumable-manager wiring, and pipeline handoff. Ships visible with `releaseStatus=ReleaseStatus.ALPHA` (the scaffold's `unreleasedSource=True` is removed).
- **`settings.py`** – declarative endpoint catalog for 10 resources: `suggestions`, `forums`, `users`, `comments`, `notes`, `nps_ratings`, `tickets`, `ticket_messages` (incremental via `updated_after`), plus `suggestion_statuses` and `labels` (full refresh). Primary key `id`, stable `created_at` partition key.
- **`uservoice.py`** – transport over `make_tracked_session()`: cursor pagination with a page-number fallback, ISO-8601 `updated_after` formatting, subdomain validation/normalization (blocks host retargeting), tenacity retries on 429/5xx, and resumable state saved after each yielded page.
- **`canonical_descriptions.py`** – documentation-sourced table/column descriptions.
- Regenerated `generated_configs.py` (`UservoiceSourceConfig` gains `subdomain`/`api_key`), moved `uservoice` into the Implemented table in `SOURCES.md`.

### Sync semantics

> [!NOTE]
> Incremental endpoints use `sort_mode="desc"`. UserVoice doesn't document a verifiable list-ordering guarantee (and its feedback endpoints tend to return newest-first), so `desc` defers the incremental watermark write to successful job end. A crashed mid-sync run can't advance the watermark past rows it never fetched; the next run re-pulls from the old watermark and merge dedupes on the primary key. This is the same pattern the Klaviyo fan-out uses.

### Unverified against a live account

I could not curl the live UserVoice API (no account/credentials in this environment), so a few behaviors are set conservatively from the public API docs and should be confirmed against a real account:

- Whether `pagination.cursor` preserves the `updated_after` filter across pages. The paginator keeps `updated_after` on every request regardless, and `sort_mode="desc"` + merge dedup keep results correct even if the cursor already encodes it.
- Exact availability of the helpdesk resources (`tickets`, `ticket_messages`) — these only return data on accounts with the Helpdesk feature. Users leave them deselected otherwise.

These are noted in code comments too.

### Docs

A posthog.com checkout wasn't available in this environment, so the user-facing doc isn't in this PR. The doc content (following the `documenting-warehouse-sources` template, with `sourceId: Uservoice` matching `docsUrl`) still needs to land in `posthog.com` at `contents/docs/cdp/sources/uservoice.md`. `lists_tables_without_credentials = True` is set so the `<SourceTables />` catalog renders.

Added two test modules (60 tests, all passing):

- `tests/test_uservoice.py` (transport) – subdomain normalization/injection rejection, `updated_after` formatting, incremental-param gating, cursor vs page-number pagination + termination, non-advancing-cursor guard, resume-from-saved-state (cursor and page), incremental filter plumbing, and the `SourceResponse` `sort_mode`/partition contract.
- `tests/test_uservoice_source.py` (source) – source type, config shape (ALPHA, no `unreleasedSource`, secret token field), `connection_host_fields`, non-retryable errors matching auth but not transient failures, per-endpoint sync modes, credential-validation mapping, resumable-manager binding, and `source_for_pipeline` argument plumbing.

I (Claude) also ran the cross-source guard tests: `test_source_categories.py` (1526 passed) and `test_source_config_generator.py` (15 passed, snapshot matches). Ran `ruff check --fix` and `ruff format` on the changed files. I did not perform any manual/live-API testing.

## 🤖 Agent context

Authored by Claude (PostHog Code). Skills invoked: `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, `/writing-tests`.

Chose the raw-`requests` + `ResumableSource` shape (Aha! as reference) over the declarative `rest_source.RESTClient` path for full control over UserVoice's cursor+page pagination and the `updated_after` filter. Enabled incremental only on the mutable, high-volume resources; the small reference tables stay full refresh. The generated-config regeneration surfaced pre-existing drift in unrelated sources (Stripe, Uptick) — I applied only the deterministic UserVoice change to keep the diff scoped.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -513,6 +513,7 @@ the row lists both.
 | unstructured                     | HTTP                        | requests                                                        | ✅                          |
 | upstash                          | HTTP                        | requests                                                        | ✅                          |
 | uptimerobot                      | HTTP                        | requests                                                        | ✅                          |
+| uservoice                        | HTTP                        | requests                                                        | ✅                          |
 | vantage                          | HTTP                        | requests                                                        | ✅                          |
 | vapi                             | HTTP                        | requests                                                        | ✅                          |
 | vellum                           | HTTP                        | requests                                                        | ✅                          |
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
@@ -4541,7 +4541,8 @@ class UsersnapSourceConfig(config.Config):
 
 @config.config
 class UservoiceSourceConfig(config.Config):
-    pass
+    subdomain: str
+    api_key: str
 
 
 @config.config
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/uservoice/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/uservoice/canonical_descriptions.py
@@ -1,0 +1,119 @@
+"""Canonical, documentation-sourced descriptions for UserVoice Admin API v2 endpoints and columns.
+
+Sourced from the official UserVoice Admin API v2 reference (https://developer.uservoice.com/docs/api/v2/reference/).
+Keyed by the resource names in `settings.py` `ENDPOINTS`, which match the `ExternalDataSchema.name`
+of a synced UserVoice table. Columns absent here fall back to LLM enrichment.
+"""
+
+from products.warehouse_sources.backend.temporal.data_imports.sources.common.canonical_descriptions import (
+    CanonicalDescriptions,
+)
+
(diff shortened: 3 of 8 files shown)


danielcarletti (approved):
LGTM

PR #71187, opened 2026-07-15
feat(data-warehouse): implement dagster_cloud import source

Gilbert09:
## Problem

Dagster+ (Dagster Cloud) was scaffolded as a hidden stub. Data teams running Dagster+ want to warehouse their run history, backfills, and asset catalog to compute reliability, DORA, and SLA metrics. This fills in the stub so the connector actually syncs, and ships it visible (ALPHA).

**Why:** requested as part of the eng/support source push started in the scaffold PR (`tom/scaffold-100-eng-support-sources`), which this stacks on.

## Changes

Dagster+ speaks GraphQL only, so it can't use the generic REST client and needs hand-written queries against the org/deployment endpoint (`https://{org}.dagster.cloud/{deployment}/graphql`) with the `Dagster-Cloud-Api-Token` header.

Three streams, following the `source.py` / `settings.py` / `dagster_cloud.py` split:

| Table | Query | Sync |
| --- | --- | --- |
| `runs` | `runsOrError` + `RunsFilter` | Incremental (`updatedAfter` / `createdAfter`), full refresh |
| `backfills` | `partitionBackfillsOrError` | Full refresh |
| `assets` | `assetsOrError` | Full refresh (catalog) |

- `ResumableSource` with cursor pagination (runId / backfill id / connection cursor). State saved after each page yields so a crash re-fetches the last page rather than skipping it; merge dedupes on the primary key.
- Only `runs` is incremental, because `RunsFilter.updatedAfter`/`createdAfter` are genuine server-side epoch filters. Backfills and assets have no server-side timestamp filter, so they ship full refresh (a client-side cursor would re-walk full history every run).
- All outbound HTTP goes through `make_tracked_session()`; `tenacity` retries 429/5xx and transient network errors with jittered backoff.
- Dagster returns timestamps as epoch-seconds floats; these are normalized to fixed-precision ISO-8601 UTC strings so datetime partitioning and the incremental watermark read them like every other source. `runs`/`backfills` partition on their stable creation time (`creationTime` / `timestamp`), never an updated-at field.
- `runsOrError` returns newest-first with no ascending option, so `runs` reports `sort_mode="desc"`.
- `organization` and `deployment` are validated as plain slugs and both listed in `connection_host_fields`, so the stored token can't be retargeted off `*.dagster.cloud` and editing either forces token re-entry.
- Canonical descriptions + `lists_tables_without_credentials=True` so the public docs render the table catalog.
- `unreleasedSource=True` removed; ships `releaseStatus=ReleaseStatus.ALPHA`.

> [!NOTE]
> I could not curl the live API (Dagster+ is a paid, per-org product and I have no token). Query shapes, field names, and the `RunsFilter` timestamp fields were verified against the current Dagster GraphQL schema source rather than a live call, so the endpoint behavior is unverified against a real deployment. Per the source skill this is exactly what ALPHA is for. The code is conservative (lean field sets, defensive `__typename` handling) given the schema is documented as internal-facing and subject to breaking changes.

### Docs

There's no posthog.com checkout in this environment, so I could not add or run `audit_source_docs` against the user-facing doc. The doc needs to land in the **posthog.com** repo at `contents/docs/cdp/sources/dagster-cloud.md` (matching `docsUrl`). Proposed content, following the `documenting-warehouse-sources` template:

<details>
<summary>dagster-cloud.md</summary>

```md
---
title: Linking Dagster+ as a source
sidebar: Docs
showTitle: true
sourceId: DagsterCloud
beta: true
---

<AlphaRelease />

Sync your Dagster+ (Dagster Cloud) run history, backfills, and asset catalog into the PostHog Data warehouse to build reliability, DORA, and SLA reporting on top of your orchestration data.

## Prerequisites

A paid Dagster+ account and a user API token with access to the deployment you want to sync. Tokens are created under **Organization settings → Tokens** in Dagster+.

## Adding a data source

<SourceSetupIntro />

You'll need:

- **Organization** – your Dagster+ organization name (the `your-org` in `your-org.dagster.cloud`).
- **Deployment** – the deployment name to sync (for example `prod`).
- **API token** – a user token from **Organization settings → Tokens**.

## Sync modes

<SyncModes />

The `runs` table supports incremental sync (via Dagster's `updatedAfter` / `createdAfter` run filters). `backfills` and `assets` are full refresh.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

## Troubleshooting

If the connection fails with an authentication error, confirm the token is still valid in **Organization settings → Tokens** and that the organization and deployment names match your Dagster+ URL.

<TroubleshootingLink />
```

</details>

Automated only — I (Claude) did not do manual testing and cannot connect to a live Dagster+ deployment. Added two test modules under `dagster_cloud/tests/` (50 cases), run green with `pytest`:

- `test_dagster_cloud.py` (transport): cursor pagination + checkpointing for both cursor modes, resume-from-saved-cursor, error `__typename` handling, timestamp normalization and epoch conversion, `RunsFilter` construction, `build_graphql_url` slug rejection (the SSRF guard), and that an incremental run actually sends a server-side filter rather than pulling everything.
- `test_dagster_cloud_source.py` (source class): schema incremental flags per endpoint, released-not-hidden config, secret field shape, non-retryable-error matching (401/403 permanent, 429/5xx/network retryable), documented-tables rendering, and `source_for_pipeline` dropping a stale watermark on non-incremental runs.

Each group targets a concrete regression (wrong cursor field, watermark applied client-side, credential retargeting, a source hidden by `unreleasedSource`) rather than restating implementation. Also ran `test_source_categories` and `test_source_config_generator` green, and `ruff check`/`ruff format`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc content proposed above for the posthog.com repo; not landed in this PR (no posthog.com checkout available).

## 🤖 Agent context

Tom directed this work; I (Claude, via PostHog Code) implemented it. Skills invoked: `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, `/writing-tests`.

Key decisions: chose `ResumableSource` over `SimpleSource` because every stream paginates by a persistable cursor. Verified query shapes and the `RunsFilter` timestamp fields against the Dagster GraphQL schema source since no live token was available, and kept only `runs` incremental (the only stream with a real server-side timestamp filter). Kept backfill/asset field sets deliberately lean because Dagster documents the schema as internal and breaking-change-prone. The generated config class was hand-populated to the deterministic generator output (three required string fields) because the config generator needs a database this environment doesn't have; a CI regen produces identical output.

---


danielcarletti (approved):
LGTM

PR #71204, opened 2026-07-15
feat(data-warehouse): implement railway import source

Gilbert09:
## Problem

Railway was scaffolded as a stub source in #71137 (registered with `unreleasedSource=True`, no sync logic). Users deploying on Railway can't yet pull their deployment and infrastructure data into the PostHog Data warehouse.

**Why:** Railway is a fast-growing PaaS, and its deployment history (status, timing, failure rates) is exactly the kind of engineering data teams want to analyze next to their product data.

## Changes

Implements the Railway source end-to-end over Railway's public GraphQL API (`backboard.railway.com/graphql/v2`) and ships it visible with `releaseStatus=ALPHA` (the `unreleasedSource=True` stub flag is removed).

Architecture follows the standard warehouse-source split:

- `settings.py` – declarative endpoint catalog with per-endpoint GraphQL documents: `projects`, `services`, `environments`, `deployments`, `project_members`, `volumes`
- `railway.py` – transport: tracked-session GraphQL client (all HTTP goes through `make_tracked_session()`), Relay cursor pagination, per-project fan-out, tenacity retries on 429/5xx, resume state (cursor + project-id bookmark saved after each yielded batch)
- `source.py` – `ResumableSource` wiring: `get_schemas`, `validate_credentials`, `source_for_pipeline`, `get_non_retryable_errors`, `lists_tables_without_credentials=True`
- `canonical_descriptions.py` – table/column descriptions from the Railway docs

Design notes:

- Railway is GraphQL-only, so the generic `rest_source.RESTClient` doesn't apply. Plain `requests` with hand-written GraphQL documents (buildbetter is the in-repo precedent).
- No list query has a server-side time filter, so only `deployments` supports incremental sync: connections return newest-first, so we page until an entire page predates the `createdAt` watermark (`sort_mode="desc"`, typeform-style termination), plus a 24h default lookback so recent deployments pick up late status changes. Everything else is full refresh (those tables are small).
- `project_members` uses a composite `[project_id, id]` primary key since member ids are user ids shared across projects.
- Auth is an account or workspace token (`Authorization: Bearer`). Project tokens are deliberately unsupported (different header, scoped to one environment, can't list projects). Railway returns auth failures as HTTP 200 + a GraphQL `Not Authorized` error, so the transport re-raises those with a stable prefix that `get_non_retryable_errors` matches.

> [!NOTE]
> Verified against the live API without credentials: Railway's endpoint allows unauthenticated introspection and validates query documents before auth, so the schema/field/argument shapes and all seven GraphQL documents were confirmed against production. What I could NOT verify without a real token: actual row ordering of `deployments` (docs and dashboard say newest-first; `sort_mode="desc"` plus job-end watermark persistence keeps a wrong assumption from corrupting the cursor) and page-size limits (kept defensively at 100). Also discovered live: Railway's CDN rejects some default Python user agents with a bare 403, so the transport sends an explicit `User-Agent`.
>
> Webhook support (Railway's `webhookCreate` mutation) was deliberately left out: webhooks are per-project (one source spans many projects) and reportedly require a personal token even for team projects, so automatic creation would fail for workspace-token users. Reasonable follow-up once verified with a real account.

The icon (`railway.png`), enum entry, schema entry, and migration already landed with the scaffold PR.

The user-facing doc (`contents/docs/cdp/sources/railway.md`) was written per the docs template and needs to land in the posthog.com repo; `manage.py audit_source_docs` passes for railway against that checkout. `SOURCES.md` moves railway to the Implemented table (HTTP (GraphQL) / requests / tracked).

- `pytest products/warehouse_sources/.../railway/tests/` – 29 tests pass.
  - `test_railway.py` (transport): cursor pagination + resume-state saved only after a yielded batch (guards the crash-skips-a-page bug), fan-out project bookmarking and resume slicing (guards restart-from-scratch and wrong-project resume), deleted-project-mid-sync null handling, `project_members` composite-key row shape (guards duplicate-row merge blowup), incremental watermark termination and its absence on first sync (guards unbounded history re-walks), retryable status mapping, and the auth-error string contract between the transport and `get_non_retryable_errors` (guards endless retries on a revoked token).
  - `test_railway_source.py` (source class): source released + alpha (guards re-hiding the connector), schema catalog and incremental advertisement, credential/pipeline plumbing.
- `pytest products/warehouse_sources/.../sources/tests/` – 1727 pass; the 2 `test_stripe_config` failures pre-exist on the base branch (verified via stash).
- All seven GraphQL documents validated against the live Railway API (resolver-level errors only, no validation errors).
- `hogli ci:preflight --fix` clean (markdown formatter skipped, no node_modules in this environment).
- Not done: no end-to-end sync against a real Railway account (no credentials in this environment) – hence ALPHA.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc written for posthog.com at `contents/docs/cdp/sources/railway.md` (matches `docsUrl`); it needs a companion PR in that repo.

## 🤖 Agent context

Authored by Claude (PostHog Code session) following the `/implementing-warehouse-sources`, `/writing-tests`, and `/documenting-warehouse-sources` skills. Key decisions: bespoke GraphQL transport instead of `rest_source.RESTClient` (GraphQL-only API), `ResumableSource` without `WebhookSource` (per-project webhooks + personal-token requirement make auto-creation unreliable), incremental limited to `deployments` via desc-order watermark termination since Railway exposes no server-side time filters. The Railway schema was introspected live (unauthenticated introspection is enabled) to confirm every query, field, and argument before writing the endpoint catalog.

---


danielcarletti (approved):
LGTM

PR #71207, opened 2026-07-15
feat(data-warehouse): implement tawk_to import source

Gilbert09:
## Problem

The tawk.to warehouse source existed only as a scaffolded stub (`unreleasedSource=True`, no fields, no sync logic), so users couldn't pull their live chat and support data into the Data warehouse.

## Why

tawk.to is a widely used free live-chat and ticketing platform, and support conversations are a natural dataset to analyze alongside product data in the warehouse.

## Changes

Implements the tawk.to source end to end and ships it visible with `releaseStatus=ReleaseStatus.ALPHA` (the `unreleasedSource=True` stub flag is removed):

- **Endpoints** (`settings.py`): `properties`, `chats` (with message transcripts), `tickets`, and `members`, declared with per-endpoint primary keys and a stable `createdOn` partition key for chats.
- **Transport** (`tawk_to.py`): tawk.to's API is RPC-over-POST (`https://api.tawk.to/v1/{method}` with JSON bodies and API-key basic auth), so the generic GET-oriented `rest_source.RESTClient` doesn't fit. A small requests-based client goes through `make_tracked_session()`, retries 429/5xx with tenacity, and paginates with `size`/`offset`.
- **Property fan-out**: an optional `Property ID` field pins the sync to one property; left blank, the source fans out over every property from `property.list` and stamps `propertyId` onto each row.
- **Resumable** (`ResumableSource`): resume state is a stable property-ID bookmark plus the row offset, saved after each yielded batch so a crash re-yields the last page instead of skipping it.
- `validate_credentials` (cheap `property.list` probe), `get_non_retryable_errors` (401/403), `canonical_descriptions.py`, and `lists_tables_without_credentials=True` so public docs render the table catalog.
- `SOURCES.md` updated (scaffolded → implemented) and `TawkToSourceConfig` regenerated.

> [!NOTE]
> All endpoints ship **full refresh only** for now. tawk.to's full API reference is gated behind an access-approval process, and community reports say the `startDate`/`endDate` filters can silently return empty result sets, so I couldn't verify a server-side timestamp filter with a live key. Per the warehouse-sources guidance, unverified filters mean no `supports_incremental`; endpoint paths, auth behavior, and error shapes were verified against the live API (401 vs 404 probes), and pagination against known working client code. A repeated-page guard stops the paginator early if a list method ignores `offset`.

The user-facing doc is in [PostHog/posthog.com#18531](https://github.com/PostHog/posthog.com/pull/18531) (slug `tawk-to`, matching `docsUrl`).

- `tests/test_tawk_to.py` (transport): offset pagination advances and terminates on short pages/total, the repeated-page guard stops a non-paginating endpoint instead of looping, rows carry `propertyId` without overwriting an API-provided one, property fan-out + bookmark advancement, resume-from-saved-state skips completed properties and starts at the saved offset, missing bookmarked property restarts cleanly, `ok:false` bodies raise, and per-endpoint `SourceResponse` metadata. No existing test covered any of this since the source was a stub.
- `tests/test_tawk_to_source.py` (source class): config fields (secret API key, optional property ID), released-with-alpha-status contract (`unreleasedSource is None`), full-refresh-only schemas, non-retryable error matching, credential validation plumbing, resume-manager binding, and blank property ID normalization.
- Ran the full `sources/tests` + `sources/common` registry suites (categories, config generator, schema) locally: 1595 passed. `ruff check`/`format` and `hogli ci:preflight --fix` pass.
- I could not test against a live tawk.to account (REST API access requires an approved request form), so endpoint behavior beyond what's stated above is unverified — hence alpha.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc added in [PostHog/posthog.com#18531](https://github.com/PostHog/posthog.com/pull/18531); `python manage.py audit_source_docs` passes for this source against that checkout.

## 🤖 Agent context

- Authored with Claude Code (PostHog Code cloud task) following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills.
- API research: cross-referenced the tawk.to help center, community threads, and public client implementations; live-probed the RPC paths unauthenticated to confirm which methods exist (401) vs don't (404) and the error body shape.
- Decisions: `ResumableSource` over `SimpleSource` (offset + property bookmark makes resumption deterministic); bespoke POST transport over `rest_source.RESTClient` (RPC-over-POST doesn't fit the GET paginator); full refresh over incremental everywhere (server-side date filters unverifiable while API access is gated — a wrongly trusted filter would silently lose rows); `sort_mode="desc"` declared so a future incremental rollout doesn't inherit a corrupt-watermark default.
- The regenerated `generated_configs.py` hunk was applied surgically: a full generator run in this environment produced unrelated env-dependent drift (Stripe OAuth fields flip required/optional based on missing local env vars), which was reverted.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -487,6 +487,7 @@ the row lists both.
 | taboola                          | HTTP                        | requests                                                        | ✅                          |
 | tailscale                        | HTTP                        | requests                                                        | ✅                          |
 | tavus                            | HTTP                        | requests                                                        | ✅                          |
+| tawk_to                          | HTTP                        | requests                                                        | ✅                          |
 | teamcity                         | HTTP                        | requests                                                        | ✅                          |
 | teamtailor                       | HTTP                        | requests                                                        | ✅                          |
 | teamwork                         | HTTP                        | requests                                                        | ✅                          |
@@ -939,7 +940,6 @@ doesn't conflict with concurrent PRs.
 - swonkie
 - synthesia
 - systeme
-- tawk_to
 - teachable
 - telli
 - tempo
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs.py
@@ -4266,7 +4266,8 @@ class TavusSourceConfig(config.Config):
 
 @config.config
 class TawkToSourceConfig(config.Config):
-    pass
+    api_key: str
+    property_id: str | None = None
 
 
 @config.config
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/tawk_to/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/tawk_to/canonical_descriptions.py
@@ -1,0 +1,42 @@
+from products.warehouse_sources.backend.temporal.data_imports.sources.common.canonical_descriptions import (
+    CanonicalDescriptions,
+)
+
+# tawk.to's full API reference is only provided after REST API access approval, so column
+# coverage is limited to fields verified from public sources; everything else falls back to
+# LLM enrichment.
+CANONICAL_DESCRIPTIONS: CanonicalDescriptions = {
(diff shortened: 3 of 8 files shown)


danielcarletti (approved):
LGTM

PR #71212, opened 2026-07-15
feat(data-warehouse): implement customerly import source

Gilbert09:
## Problem

Customerly was a scaffolded warehouse source stub (registered but hidden with `unreleasedSource=True` and no sync logic). Users who run their support and marketing on Customerly couldn't pull that data into the PostHog Data warehouse.

## Changes

Implements the Customerly source end-to-end and ships it visible with `releaseStatus=ReleaseStatus.ALPHA` (the `unreleasedSource` flag is removed):

- **Endpoints** (declared in `settings.py`): `users`, `leads`, `tags`, `knowledge_base_collections`, and `knowledge_base_articles` (fanned out per collection). These are the syncable list endpoints the public REST API exposes.
- **Transport** (`customerly.py`): `requests` via `make_tracked_session()`, bearer-token auth, page-number pagination (`page`/`per_page`), tenacity retries on 429/5xx, and `ResumableSource` state so long imports resume mid-list (and mid-fan-out for articles) after a heartbeat timeout.
- **Full refresh only**: the API exposes no server-side timestamp filter on any list endpoint, so no table advertises incremental sync.
- **Auth failure handling**: Customerly reports bad tokens as HTTP 500 with an error body, so the transport detects the auth error body before the generic retryable-5xx path and raises a permanent error mapped in `get_non_retryable_errors()`.
- `validate_credentials` probes `/v1/tags` (a cheap single request), `lists_tables_without_credentials = True` plus `canonical_descriptions.py` so the public docs render the table catalog, and SOURCES.md moves customerly into the Implemented table.

Endpoint shapes, pagination style, and route existence were verified against the live API (unauthenticated probes confirm routes and the auth scheme) and the official Postman collection. Two things could not be verified without a real access token, and conservative choices are noted in code comments: the users/leads list sort uses the documented `sort=name&sort_direction=asc` example, and page-boundary duplicates are handled by merge-on-primary-key.

The user-facing doc (`contents/docs/cdp/sources/customerly.md`) is added in companion PR [PostHog/posthog.com#18539](https://github.com/PostHog/posthog.com/pull/18539); `docsUrl` points at `/docs/cdp/sources/customerly`. `audit_source_docs` run against that checkout reports no customerly issues.

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/customerly/tests/` (50 tests) plus `test_source_categories.py` – all pass.
- `test_customerly.py` (transport): pagination terminates on a short page and resumes from saved state (guards infinite-loop/skipped-page regressions), articles fan-out walks collections in ascending id order and skips completed collections on resume, tags normalize from bare strings to rows, and an auth failure disguised as HTTP 500 raises a permanent error instead of being retried (this is Customerly-specific behavior no existing test covers).
- `test_customerly_source.py` (source class): schema catalog is full-refresh only, non-retryable error mapping matches real error strings and not other vendors', credential validation plumbs through, and the source config stays visible (`unreleasedSource is None`).
- Not done: an end-to-end sync against a live Customerly account (no access token available in this environment).
- `ruff check` / `ruff format` and `hogli ci:preflight --fix` pass; `makemigrations --check` reports no changes.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

User-facing doc added in [PostHog/posthog.com#18539](https://github.com/PostHog/posthog.com/pull/18539) following the documenting-warehouse-sources template.

## 🤖 Agent context

Implemented with Claude Code following the `/implementing-warehouse-sources`, `/writing-tests`, and `/documenting-warehouse-sources` skills, using the aircall source as the structural reference. Key decisions: `ResumableSource` over `SimpleSource` because page-number pagination gives a deterministic resume point; full refresh everywhere because no server-side incremental filter exists (per the skill's rule against client-side "incremental"); raw `requests` over `rest_source.RESTClient` because the auth-error-as-500 quirk and per-collection fan-out need custom classification and resume logic that the aircall-style transport expresses more directly.

---


danielcarletti (approved):
LGTM

PR #71213, opened 2026-07-15
feat(data-warehouse): implement drata import source

Gilbert09:
## Problem

Drata was a scaffolded warehouse source with no sync logic. Teams using Drata for compliance automation (SOC 2, ISO 27001, HIPAA) want their controls, monitoring-test results, evidence, personnel and device compliance, vendors, policies, risks, and audit-trail events in the warehouse so they can trend audit readiness and blend GRC posture with the rest of their data.

**Why:** part of implementing the eng/support source batch scaffolded in #71137 — this fills in the Drata stub so the connector actually syncs and is visible to users.

## Changes

Implements the Drata source end to end as a `ResumableSource`, following the standard `source.py` / `settings.py` / `drata.py` split:

- **Auth and regions**: Bearer API key plus a region select (North America / Europe / Asia-Pacific) mapped to Drata's regional API hosts. All outbound HTTP goes through `make_tracked_session()` with the key redacted.
- **14 tables**, declared in `settings.py`: account-level inventories (`workspaces`, `users`, `personnel`, `devices`, `assets`, `vendors`, `policies`), the `events` audit trail, workspace-scoped fan-out tables (`controls`, `monitoring_tests`, `evidence_library`, `frameworks`, each carrying an injected `workspaceId` in a composite primary key), and risk tables (`risk_registers`, `risks`) which are deselected by default because they 403 without Drata's Risk Management Pro feature.
- **Pagination**: v2 cursor pagination (`cursor`/`size`, explicit `sort=createdAt&sortDir=ASC`), with a guard that terminates if the cursor stops advancing. Resume state (cursor + fan-out parent bookmark) is saved after each yielded batch, so Temporal can resume mid-sync without re-walking earlier pages or parents.
- **Incremental sync on `events` only** — the one endpoint with a documented server-side timestamp filter (`createdAtStartDate`). Everything else has no since-filter in the v2 API, so those tables ship full refresh rather than a fake client-side "incremental".
- **Partitioning** on the stable `createdAt` field (monthly), `sort_mode="desc"` on the incremental endpoint so the watermark only commits after a complete sync.
- Non-retryable errors for invalid/expired keys, missing scopes, and unaccepted API terms (412); 403 at source-create is accepted since custom-scoped keys may legitimately lack the probe endpoint's scope. Curated `canonical_descriptions.py` from the official API docs; static table catalog exposed for public docs; `SOURCES.md` updated.

The source ships visible with `releaseStatus=ALPHA` (the scaffold's `unreleasedSource=True` is removed). The icon already existed at `frontend/public/services/drata.png`.

> [!NOTE]
> Endpoint paths, params, pagination, and response shapes were verified against Drata's current public API v2 reference (developers.drata.com). Live curl verification wasn't possible without an API key — Drata returns 401 on every path before route resolution — so two behaviors are declared conservatively and noted in code comments: the `createdAtStartDate` filter on `events` is documented but not live-tested, and response ordering couldn't be confirmed, so the incremental endpoint uses `sort_mode="desc"` (watermark commits only at successful job end, correct regardless of arrival order).

89 unit tests across the two standard modules, all passing, plus the registry-wide source suites (1816 passed; the two pre-existing `test_stripe_config` failures also fail on the base branch with this change stashed). `ruff check` / `ruff format` and `hogli ci:preflight --fix` are clean. No live-credential testing was possible in this environment.

- `tests/test_drata.py` (transport): catches regressions in cursor termination (including a non-advancing cursor, which would otherwise loop forever), resume-from-saved-cursor without re-fetching page one, fan-out parent bookmarks (resume skips completed workspaces, saved cursor applies only to the bookmarked parent, deleted parents fall back to a full walk), `workspaceId`/`riskRegisterId` injection that keeps composite keys unique table-wide, the events server-side filter on every page, region-host routing, retryable-vs-permanent status mapping, URL query scrubbing in error messages, and credential-validation mapping (401 invalid, 403 accepted at create, 412 terms).
- `tests/test_drata_source.py` (source class): catches drift in the schema catalog and incremental flags, the risk tables' default-off state, the placeholder-config docs path, per-region non-retryable error matching, config-to-transport plumbing (including nulling the watermark when incremental is off), and the resume-manager binding.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The user-facing doc below follows the `documenting-warehouse-sources` template and needs to land in the posthog.com repo as `contents/docs/cdp/sources/drata.md` (no posthog.com checkout was available here, so `audit_source_docs` couldn't be run). `docsUrl` already points at `https://posthog.com/docs/cdp/sources/drata`.

<details>
<summary>posthog.com doc: contents/docs/cdp/sources/drata.md</summary>

```markdown
---
title: Linking Drata as a source
sidebar: Docs
showTitle: true
sourceId: Drata
beta: true
---

<AlphaRelease />

The Drata connector syncs your [Drata](https://drata.com/) compliance data, including controls, monitoring tests, evidence, personnel and device compliance, vendors, policies, risks, and audit-trail events, into the PostHog Data warehouse. Use it to trend audit readiness and analyze your compliance posture alongside the rest of your data.

## Prerequisites

- A Drata account with permission to create API keys.
- Accepted Drata API terms and conditions (Drata prompts for this the first time the API is used).
- The `risk_registers` and `risks` tables require Drata's Risk Management Pro feature.

## Adding a data source

<SourceSetupIntro />

The connector authenticates with a Drata API key:

1. In Drata, go to **Settings** and open **API keys**, then create a new API key.
2. Grant the key **read** scopes for the data you want to sync (or the "All read" scope). Write scopes are not needed.
3. Copy the key when it is shown. Drata only displays it once.
4. Pick the **region** that matches your Drata account: North America, Europe, or Asia-Pacific.

If your key uses IP allowlisting, create a key without an allowlist for this connection.

## Sync modes

<SyncModes />

The `events` table supports incremental sync on `createdAt`. The other tables are compliance inventories with no server-side change filter, so they sync as full refresh.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

Tables scoped to a workspace in Drata (`controls`, `monitoring_tests`, `evidence_library`, `frameworks`) sync across every workspace, with a `workspaceId` column identifying the workspace each row belongs to. Rows in `risks` carry a `riskRegisterId` column identifying their risk register.

## Troubleshooting

- **Invalid Drata API key**: the key is wrong, expired, or was revoked. Create a new key under Settings and API keys in Drata, then update the source.
- **Permission errors (403)**: grant the API key the matching read scope for the table you're syncing. The `risk_registers` and `risks` tables also require the Risk Management Pro feature.
- **"You must accept the Drata terms and conditions"**: open Drata and accept the API terms, then retry the sync.

<TroubleshootingLink />
```

</details>

## 🤖 Agent context

- Built with PostHog Code (Claude Code). Skills invoked: `/implementing-warehouse-sources`, `/writing-tests`, `/documenting-warehouse-sources`.
- Endpoint selection cross-referenced the Drata v2 API reference against the streams other connectors ship. Workspace-scoped resources use a fan-out over `/workspaces` with a stable parent-id resume bookmark (klaviyo's pattern) rather than asking the user for a workspace id, since most accounts have more than one and the list is cheap.
- Incremental sync is limited to `events` on purpose: the v2 API has no updated-since filter, and a sorted-DESC client-side walk was rejected in favor of honest full refresh for the inventory tables, which are small.
- Regenerating `generated_configs.py` produced unrelated formatting churn in other sources' configs, so the file was re-formatted with ruff and only the Drata hunk kept.

---


danielcarletti (approved):
LGTM

PR #71224, opened 2026-07-15
feat(data-warehouse): implement featurebase import source

Gilbert09:
## Problem

Featurebase (customer feedback, roadmap, and changelog platform) was scaffolded in #71137 as a hidden stub with no sync logic. Product teams want it warehoused to rank feature demand (posts + upvoters), analyze feedback volume by board and status, and join changelog cadence against adoption metrics.

## Changes

Implements the source end to end and ships it visible with `releaseStatus=ALPHA` (removes `unreleasedSource`).

**Architecture** (per the `source.py` / `settings.py` / `featurebase.py` contract):

- `settings.py` – declarative catalog of 10 endpoints: `posts`, `boards`, `post_statuses`, `comments`, `custom_fields`, `changelogs`, `admins`, `companies`, `contacts`, and an opt-in `post_voters` fan-out (one voters request chain per post, composite `[postId, id]` key).
- `featurebase.py` – transport over `make_tracked_session()`: cursor pagination (`limit`/`cursor`/`nextCursor`), tenacity backoff on 429/5xx, resumable state (cursor + fan-out post bookmark, saved after each yielded batch), and webhook management (create/reconcile/info/delete against `/v2/webhooks`).
- `FeaturebaseSource` – `ResumableSource + WebhookSource`, static schema catalog (`lists_tables_without_credentials`), credential validation, non-retryable auth errors, canonical table/column descriptions from the official API reference.

**Incremental sync.** Featurebase has no `updatedSince`-style filter on posts/comments, so those use the descending-sort early-cutoff pattern (same as the GitHub source): sweep newest-first, stop once a whole page predates the watermark, `sort_mode="desc"` so the watermark only persists at job end. `changelogs` has a genuine server-side lower bound (`startDate`) with an ascending sort, so it stays `asc`. Everything else is full refresh. All tables are merge-only (`supports_append=False`) since every resource is mutable.

**Webhooks.** `POST /v2/webhooks` returns the signing secret on create, so registration is fully automatic (secret lands on the hog function via `extra_inputs`). The template verifies the HMAC-SHA256 `{timestamp}.{body}` signature and routes on `data.item.object` into `posts`/`comments`/`changelogs`. Deleted-object topics are intentionally not subscribed – merging a deletion payload would resurrect the row; deletions reconcile on full refresh.

Docs: companion posthog.com PR [PostHog/posthog.com#18541](https://github.com/PostHog/posthog.com/pull/18541) adds `/docs/cdp/sources/featurebase` (matches `docsUrl`); `audit_source_docs` passes for this source against that checkout.

> [!NOTE]
> Verified unauthenticated against the live API: base URL, auth header shape, and that invalid/missing keys return **403** (not 401) with `{"success": false, "message": ...}` – `validate_credentials` and `get_non_retryable_errors` are built around that. Authenticated behavior (pagination, sort enums, the `startDate` filter, webhook creation) is implemented from the current official API reference (`2026-01-01.nova`) but couldn't be curl-verified without an API key; the riskier assumptions are flagged with code comments (notably `sortBy=recent` ordering for the posts `updatedAt` sweep).

- `tests/test_featurebase_source.py` + `tests/test_featurebase.py` (72 tests, all passing): cursor pagination + resume-from-saved-state, desc-sweep termination at the watermark (and full walk on first sync), server-side `startDate` param shape, fan-out post-ID injection/404 skip/bookmark resume, webhook batch dedup to latest-per-id, webhook create/reconcile/delete including the org webhook cap error, sort-mode/partition/primary-key mapping, and schema capability flags. Each group targets a regression the skill calls out (watermark corruption from wrong sort_mode, duplicate rows from non-unique fan-out keys, unbounded incremental re-walks, endless retries on bad credentials).
- Registry-wide suites (`test_source_categories`, `test_generated_configs`, `test_ci_core_coupled_sources`) pass; the 2 `test_stripe_config` failures pre-exist on the base branch.
- `ruff check --fix`, `ruff format`, and `hogli ci:preflight --fix` pass.
- Not done: no live sync against a real Featurebase account (no API key available in this environment) – hence ALPHA.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Handled in [PostHog/posthog.com#18541](https://github.com/PostHog/posthog.com/pull/18541).

## 🤖 Agent context

Authored with PostHog Code (Claude Code). Skills invoked: `implementing-warehouse-sources`, `writing-tests`, `documenting-warehouse-sources`. Key decisions: chose `ResumableSource + WebhookSource` over the generic `rest_source.RESTClient` path to support the desc-sweep cutoff and fan-out bookmarks; modeled the webhook flow on the Stripe/Customer.io sources; excluded `*.deleted` topics from the webhook subscription to avoid resurrecting deleted rows via merge; kept `post_voters` off by default due to per-post API cost. API behavior was verified from the live API where possible without credentials (auth error shape) and from the versioned API reference otherwise.

---


danielcarletti (approved):
LGTM

PR #71225, opened 2026-07-15
feat(data-warehouse): implement confluent_cloud import source

Gilbert09:
## Problem

The `confluent_cloud` warehouse source was scaffolded in #71137 but had no sync logic, so it was hidden from users. Confluent Cloud's Metrics API only retains about 7 days of data, so teams doing capacity planning or FinOps on Kafka need those operational metrics warehoused to analyze anything beyond the last week.

## Changes

Implements the source end to end against the [Confluent Cloud Metrics API](https://api.telemetry.confluent.cloud/docs) and ships it released with `releaseStatus=ALPHA`:

- **7 tables**: `metric_descriptors` and `resource_descriptors` (full refresh catalogs), plus hourly time-series tables for Kafka clusters, connectors, ksqlDB, Schema Registry, and Flink compute pools. Metric names are discovered from the live descriptor catalog at sync time, so new Confluent metrics flow without code changes.
- **Windowed incremental sync**: the query endpoint takes ISO-8601 `intervals`, so each sync walks day-sized windows from the `timestamp` watermark (minus a 2h restatement overlap, deduped by merge) to now. `sort_mode="desc"` because the API returns points newest-first within groups.
- **Resumable**: the window start is checkpointed to Redis after each completed window, after its rows are yielded. Query page cursors are short-lived, so a retried window restarts fresh.
- **Auth and validation**: HTTP Basic with a Cloud API key + secret. The descriptor endpoints are public, so `validate_credentials` probes the query endpoint; the API requires a resource filter for authorization, so users list resource IDs per type in the config and validation distinguishes 401 (bad key) from 403 (missing MetricsViewer role).
- Standard trimmings: tracked HTTP transport, tenacity backoff on 429/5xx, non-retryable auth errors, canonical table/column descriptions, `lists_tables_without_credentials` for public docs, SOURCES.md moved to Implemented.

The user-facing doc is in [PostHog/posthog.com#18544](https://github.com/PostHog/posthog.com/pull/18544); `audit_source_docs` passes for this source against that checkout.

> [!NOTE]
> Endpoint behavior was verified against the live API where possible without credentials: the descriptor endpoints, pagination (`page_token`/`page_size`), `resource_type` filtering, and the published OpenAPI spec (query schema, filter requirement, point ordering, pagination semantics). The authenticated query path follows the spec exactly but hasn't been run against a real account, hence alpha.

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/confluent_cloud/tests/` — 50 tests pass. Transport tests cover the regressions specific to this source's shape: window iteration off-by-ones, resume-state saved before yield (data loss), query pagination not re-POSTing the identical body, incremental watermark/overlap/retention-floor math, deprecated-metric filtering, and 401/403/no-ids error mapping. Source tests cover schema sync modes, credential probe selection, and pipeline argument plumbing.
- Sources guard suites (`test_source_categories`, `test_generated_configs`) pass; the two `test_stripe_config` failures are pre-existing on the base branch.
- `ruff check`/`ruff format` and `hogli ci:preflight --fix` are clean.
- Not done: an end-to-end sync against a real Confluent Cloud account (no credentials available).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc added in [PostHog/posthog.com#18544](https://github.com/PostHog/posthog.com/pull/18544).

## 🤖 Agent context

- Authored with Claude Code following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills.
- Key decisions: user-supplied resource IDs instead of auto-discovery (the spec states queries require a resource filter for authorization, and the attributes-based discovery path couldn't be verified without credentials); one query per metric per window (the API caps requests at one aggregation); values aggregated per resource rather than grouped by metric labels like topic, to keep cardinality and primary keys stable; whole-window resume checkpoints because query page tokens expire.
- 403 on the credential probe is accepted as valid only when probing a placeholder resource (no IDs configured) — a real key without MetricsViewer gets a clear error instead.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -129,6 +129,7 @@ the row lists both.
 | concord                          | HTTP                        | requests                                                        | ✅                          |
 | configcat                        | HTTP                        | requests                                                        | ✅                          |
 | confluence                       | HTTP                        | requests                                                        | ✅                          |
+| confluent_cloud                  | HTTP                        | requests                                                        | ✅                          |
 | chartmogul                       | HTTP                        | requests                                                        | ✅                          |
 | circleci                         | HTTP                        | requests                                                        | ✅                          |
 | cimis                            | HTTP                        | requests                                                        | ✅                          |
@@ -629,7 +630,6 @@ doesn't conflict with concurrent PRs.
 - cockroachdb
 - codacy
 - codecov
-- confluent_cloud
 - constant_contact
 - copper
 - coralogix
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/confluent_cloud/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/confluent_cloud/canonical_descriptions.py
@@ -1,5 +1,6 @@
 from products.warehouse_sources.backend.temporal.data_imports.sources.common.canonical_descriptions import (
     CanonicalDescriptions,
+    CanonicalEndpoint,
 )
 
 _METRICS_DOCS_URL = "https://api.telemetry.confluent.cloud/docs"
@@ -12,7 +13,7 @@ _TIME_SERIES_COLUMNS = {
 }
 
 
-def _time_series_table(resource_noun: str) -> dict:
+def _time_series_table(resource_noun: str) -> CanonicalEndpoint:
     return {
         "description": f"Hourly time-series values for every Metrics API metric that applies to your {resource_noun}, one row per metric, resource, and hour.",
         "docs_url": _METRICS_DOCS_URL,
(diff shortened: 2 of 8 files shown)


Gilbert09:
Pushed a fix for the `ty` failure in `Python code quality` (the canonical-descriptions helper needed a `CanonicalEndpoint` return annotation) — that was the only diagnostic in this PR's files.

The `Product tests (warehouse-sources)` shards fail on a collection error in `stripe/tests/test_stripe_source.py` (`StripeAuthMethodConfig.__init__() missing 1 required positional argument: 'stripe_secret_key'`). That reproduces on the base branch `tom/scaffold-100-eng-support-sources` without this PR's changes — the branch's Stripe test expects master's optional-field config shape while the branch still carries the older Stripe source — so it needs a master merge on the scaffold branch rather than a fix here. The `test_generated_configs.py::test_stripe_config` failures are the same pre-existing drift.

estefaniarabadan (approved):

PR #71231, opened 2026-07-15
feat(data-warehouse): implement babelforce import source

Gilbert09:
## Problem

babelforce (a cloud contact center / CCaaS platform) existed only as a scaffolded warehouse source stub hidden behind `unreleasedSource=True`, so users couldn't import their contact center data into the PostHog Data warehouse.

## Why

Teams running their voice and SMS support on babelforce want call reporting, agent, and queue data next to their product data in PostHog for analytics.

## Changes

Implements the babelforce source end-to-end following the `source.py` / `settings.py` / `babelforce.py` architecture contract:

- **Endpoints** (`settings.py`): `calls` (call reporting), `agents`, `agent_groups`, `queues`, `numbers`, `recordings`, `sms`, and `conversations` from the Manager API, all keyed on `id`. `calls` and `sms` partition on `dateCreated` (stable creation time).
- **Transport** (`babelforce.py`): `requests` via `make_tracked_session()`, access ID/token header auth, per-environment base URL (`https://{environment}.babelforce.com/api/v2`, usually `services`), `page`/`max` pagination that advances from `pagination.current` in the response and terminates on last page, missing pagination metadata, or a non-advancing page. Retries with tenacity on 429/5xx/transport errors.
- **Incremental sync**: only `calls` advertises it, using the documented server-side `dateCreated.start`/`dateCreated.end` unix-timestamp filters. The reporting API doesn't document a response ordering and has no sort param, so the response is declared `sort_mode="desc"`, which makes the pipeline finalize the incremental watermark only after a fully successful sync – correct for any actual ordering. The upper bound is frozen at sync start so pages stay stable while new calls arrive mid-sync.
- **Resumable**: `ResumableSource` saving the next page number plus the query window after each yielded batch, so Temporal can resume mid-pagination without restarting.
- **Security**: `environment` is declared as a `connection_host_fields` entry (retargeting the subdomain forces re-entering the token) and validated as a plain DNS label before any request is built.
- **Released**: `unreleasedSource=True` removed; ships visible with `releaseStatus=ReleaseStatus.ALPHA`.
- Canonical table/column descriptions from the official OpenAPI spec, `lists_tables_without_credentials=True` so public docs render the table catalog, `SOURCES.md` moved to the Implemented table.

> [!NOTE]
> I could not curl the live API with real credentials (none available), so two behaviors are documented-but-unverified and noted in code comments: the exact filter param spelling (`dateCreated.start`/`.end` per the current official OpenAPI spec; the community Airbyte connector historically used `filters.dateCreated.from`/`.to`) and the first-page index (handled by omitting `page` on the first request and advancing from the response's `pagination.current`, the same strategy the Airbyte connector uses in production). If the filter turns out to be ignored, incremental syncs degrade to full-window reads with correct results (merge dedupes on `id`) rather than losing data.

The user-facing doc was written per the canonical template, passes `audit_source_docs`, and is up as a companion PR: [PostHog/posthog.com#18551](https://github.com/PostHog/posthog.com/pull/18551).

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/babelforce/tests/` (65 passed): transport tests cover pagination termination (last page, missing pagination metadata, non-advancing page without re-yielding duplicates – guards infinite loops and duplicate rows if the API ignores `page`), resume-from-saved-state reusing the saved window, incremental runs windowing the query from the watermark, watermark coercion from ISO strings/datetimes/epochs, environment/subdomain validation (guards credential retargeting via a crafted host), and credential-validation status mapping. Source tests cover the config contract (visible source, alpha status, secret password token field), schema catalog (incremental only on `calls`), non-retryable 401/403 mapping, resumable manager binding, and pipeline argument plumbing.
- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/tests/` (registry-wide guards, 1555 passed).
- `ruff check` / `ruff format` and `hogli ci:preflight --fix` pass. No live-API manual testing was done (no credentials).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc written for posthog.com (`contents/docs/cdp/sources/babelforce.md`, matching the `docsUrl` slug) – companion PR: [PostHog/posthog.com#18551](https://github.com/PostHog/posthog.com/pull/18551).

## 🤖 Agent context

Built with the `implementing-warehouse-sources`, `writing-tests`, and `documenting-warehouse-sources` skills. Endpoint catalog, auth, pagination, and filter params were derived from the official OpenAPI spec embedded in babelforce's developer hub Swagger UI, cross-checked against the community Airbyte connector's manifest (which confirmed the header auth scheme, the `page`/`max` pagination advancing from `pagination.current`, and the region enum). Chose the access ID/token header pair over the OAuth password grant (no password storage, no 12h token refresh); chose `sort_mode="desc"` because the reporting API's ordering is undocumented, trading per-batch watermark checkpointing for correctness; kept `recordings` full-refresh because the mapping of its `filters.time` param to a response field is unverified.

---


danielcarletti (approved):
LGTM

PR #71232, opened 2026-07-15
feat(data-warehouse): implement new_relic import source

Gilbert09:
## Problem

New Relic was a scaffolded warehouse source stub (hidden behind `unreleasedSource=True`, no sync logic), so users couldn't import their observability data into the Data warehouse.

## Changes

Implements the `new_relic` source end to end and ships it visible with `releaseStatus=ReleaseStatus.ALPHA`.

New Relic's actively developed API is NerdGraph (GraphQL), so this source speaks GraphQL over `requests` via `make_tracked_session()` rather than the generic REST client. It follows the standard `source.py` / `settings.py` / `new_relic.py` split:

- **Event tables** (`transactions`, `transaction_errors`, `page_views`, `logs`, `spans`) are read with NRQL `SELECT * FROM <table> SINCE <start> UNTIL <end> LIMIT MAX` through `actor.account.nrql`. NRQL caps results at 5,000 rows per query, so windows are walked in ascending 6-hour chunks and recursively halved when a window fills up (down to a 1s floor, with a truncation warning). Rows are sorted ascending client-side so the declared `sort_mode="asc"` holds. NRQL event rows have no unique identifier, so these tables sync **append-only** on `timestamp` (converted from epoch ms to a datetime, which also powers weekly datetime partitioning). Initial sync reaches back 30 days; `UNTIL` excludes a 5-minute ingest-lag buffer so late-arriving events aren't permanently missed. `logs` and `spans` are off by default since they can be very high volume.
- **Entity/config tables** (`entities` via `entitySearch`, `alert_policies`, `alert_conditions`) use NerdGraph cursor pagination and sync full refresh (New Relic exposes no updated-since filter for them).
- **Resumable**: event tables save the next window start after each yielded window (`ResumableSource`). Entity-style tables intentionally don't save cursor state, since NerdGraph cursors can expire and a stale saved cursor would wedge every retry until the state TTL passes.
- **Regions**: EU-hosted accounts use `api.eu.newrelic.com/graphql`, selectable via a region field.
- **Auth**: User API key + account ID; `validate_credentials` probes `actor.account(id:)` and distinguishes a bad key (401/403) from a valid key without access to the given account. `get_non_retryable_errors()` covers 401/403 on both regional hosts. Retries (tenacity) cover 429/5xx and NerdGraph's HTTP-200 timeout errors.
- Canonical table/column descriptions from New Relic's attribute dictionary, `lists_tables_without_credentials = True` so the public docs render the table catalog, and `SOURCES.md` moved to Implemented.

> [!NOTE]
> I verified auth behavior against the live API without credentials (invalid key → HTTP 401 with a JSON `errors` body) and validated all five GraphQL documents parse, but couldn't run authenticated smoke tests (no New Relic account/key available). The conservative bits that follow from that: entity search queries only universal `EntityOutline` fields, NRQL windows use epoch-ms `SINCE`/`UNTIL`, and everything ships as alpha. Field-level notes are in code comments.

### Docs

The user-facing doc (`contents/docs/cdp/sources/new-relic.md`, matching `docsUrl`) was written per the docs skill and passes `manage.py audit_source_docs` against a posthog.com checkout, but it needs to land in the posthog.com repo separately:

<details>
<summary>new-relic.md content for posthog.com</summary>

~~~md
---
title: Linking New Relic as a source
sidebar: Docs
showTitle: true
availability:
  free: full
  selfServe: full
  enterprise: full
sourceId: NewRelic
---

<AlphaRelease />

The New Relic connector syncs your observability data – APM transactions, transaction errors, page views, logs, distributed tracing spans, monitored entities, and alert configuration – into PostHog, so you can query it alongside your product analytics.

## Prerequisites

You need a New Relic account and a **User API key** (it starts with `NRAK-`). License keys, browser keys, and the deprecated Insights query keys won't work. Create a User API key on the [API keys page](https://one.newrelic.com/admin-portal/api-keys/home) of your New Relic account.

You also need your New Relic **account ID**. It's shown in the URL when viewing your account, or under **Administration** → **Access management** → **Accounts**.

## Adding a data source

<SourceSetupIntro />

When linking New Relic, you'll need:

- **User API key** – the key you created on the API keys page (starts with `NRAK-`).
- **Account ID** – the numeric ID of the New Relic account to sync.
- **Region** – select **EU** if your account is hosted in New Relic's EU data center. Most accounts use **US**.

## Sync modes

<SyncModes />

Event tables (`transactions`, `transaction_errors`, `page_views`, `logs`, and `spans`) are append-only, since New Relic events are immutable. On the initial sync, only the last 30 days of events are imported, and New Relic's own data retention limits how far back data is available on your plan. The `logs` and `spans` tables can be very high volume, so they're disabled by default – toggle them on in the schema configuration if you want them.

Entity and alert configuration tables (`entities`, `alert_policies`, and `alert_conditions`) are synced with a full refresh, since New Relic doesn't expose updated-since filters for them. These are small configuration tables, so full refresh syncs complete quickly.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

## Troubleshooting

- **Your API key is valid but has no access to account …** – the key belongs to a user without access to the account ID you entered. Double-check the account ID and make sure the key was created by a user on that account.
- **Your New Relic API key is invalid or has been revoked** – create a new User API key on the [API keys page](https://one.newrelic.com/admin-portal/api-keys/home) and reconnect.

<TroubleshootingLink />
~~~

</details>

Automated tests only (no live-credential manual testing, see the note above):

- `tests/test_new_relic.py` (transport, 50 cases): window chunking splits recursively at the 5,000-row cap and yields ascending batches (guards watermark corruption from unordered NRQL results); incremental syncs start 1ms past the watermark and stop before the ingest-lag buffer (guards boundary-row duplication in append mode and permanent late-event loss); resume state is saved after each yielded window except the last and honored on restart (guards data loss on mid-sync crashes); cursor pagination terminates on null `nextCursor`; retryable vs fatal error mapping for HTTP statuses and HTTP-200 GraphQL errors; credential validation distinguishes bad key from no-account-access.
- `tests/test_new_relic_source.py` (source class, 25 cases): schema catalog flags (append-only events, full-refresh entities, high-volume tables off by default), non-retryable error matching per region, config/pipeline plumbing including watermark dropped on full refresh, and `unreleasedSource is None` so the source can't silently regress to hidden.

All 75 pass, plus the 1,570 shared source registry/config-generator/category guard tests. `ruff check`/`format` and `hogli ci:preflight --fix` are clean; `makemigrations --check` shows no pending migration (the enum landed with the scaffold).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc content included above; it needs a companion PR in posthog.com (this repo's `audit_source_docs` passes with that file in place).

## 🤖 Agent context

Authored by Claude (PostHog Code cloud task) from a per-source implementation brief. Skills invoked: `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, `/writing-tests`. Key decisions: NerdGraph GraphQL over the legacy REST v2 API (REST v2 doesn't cover event/NRQL data); bespoke transport instead of `rest_source.RESTClient` (GraphQL POST body, not REST); append-only instead of incremental-merge for event tables (NRQL rows have no unique id, and a non-unique merge key would multi-match on every sync); recursive window halving instead of cursor advancement for the 5,000-row cap (NRQL returns newest-first within a window, so a capped page can't be advanced past); no resume state for cursor-paginated tables (stale NerdGraph cursors would poison retries). The Stripe config drift the generator produced locally (env-dependent OAuth fields) was reverted so only the New Relic config changed in `generated_configs.py`.

---


danielcarletti (approved):
LGTM

PR #71233, opened 2026-07-15
feat(data-warehouse): implement clickhouse_cloud import source

Gilbert09:
## Problem

The `clickhouse_cloud` warehouse source was scaffolded in #71137 with `unreleasedSource=True` and no sync logic. FinOps and platform teams want ClickHouse Cloud cost and service data in the warehouse for chargeback, cost-trend, and capacity dashboards.

## Changes

Implements the source end-to-end and ships it visible with `releaseStatus=ALPHA` (the `unreleasedSource` flag is removed).

**Architecture** (per the `source.py` / `settings.py` / transport split):

- `settings.py` – declarative endpoint catalog: `organizations`, `services`, `usage_cost`, `api_keys`, `members`, `activities`, `backups`, with per-endpoint primary keys, stable partition keys (`date`, `createdAt`, `startedAt`), and incremental fields.
- `clickhouse_cloud.py` – transport over `make_tracked_session()` with HTTP Basic auth (API key ID/secret). The API returns complete unpaginated `result` arrays, and scopes each key to exactly one organization.
- `usage_cost` is the flagship stream: the API caps `from_date`/`to_date` at 31 days per request, so the source walks contiguous non-overlapping windows forward from the org's `createdAt` (full refresh) or the watermark (incremental), yielding rows sorted ascending so per-batch watermark checkpointing stays correct. Nested `metrics` (storageCHC, computeCHC, ...) are flattened into columns. Recent records are restated until ClickHouse locks them, so the schema declares a 31-day incremental lookback and is merge-only.
- `ResumableSource`: the usage cost walk bookmarks the next window start in Redis after each yielded window, so Temporal retries resume mid-backfill instead of restarting.
- 429s honor `Retry-After` (the API allows 10 requests per 10s); 401/403 are registered as non-retryable with actionable user messages.
- `canonical_descriptions.py` documents every table/column from the official OpenAPI spec, and `lists_tables_without_credentials=True` publishes the catalog to the public docs.

> [!NOTE]
> Endpoint shapes were verified against the live public OpenAPI spec at `api.clickhouse.cloud` (fetched during implementation) plus unauthenticated probes for error shapes. I couldn't run authenticated calls (no ClickHouse Cloud credentials), so two behaviors are spec-verified but not live-verified: the `activities` `from_date` filter (kept merge-only so a silently-ignored filter can't duplicate rows) and the exact lock cadence of usage cost records (covered by the 31-day re-pull window).

Docs: [PostHog/posthog.com#18554](https://github.com/PostHog/posthog.com/pull/18554) adds `contents/docs/cdp/sources/clickhouse-cloud.md`; `docsUrl` matches the slug and `audit_source_docs` reports no issues for this source.

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse_cloud/tests/` – 52 tests pass.
  - Transport tests catch: >31-day or overlapping usage cost windows (API 400s / duplicate rows within a sync), a future-dated watermark producing `from_date > to_date`, resume state saved before yield (data loss on crash), missing `organizationId`/`serviceId` stamping (breaks composite primary keys), activities arriving unsorted (corrupts the ascending watermark), credential-status mapping, and retry/non-retryable error classification.
  - Source tests catch: incremental advertised on endpoints without a server-side filter, the restatement lookback being dropped, wrong primary keys/partitioning per endpoint, and the source regressing to hidden (`unreleasedSource`).
- Registry-wide suites (`sources/tests/`, `common/test/`) pass except pre-existing MinIO-dependent failures that also fail on the base branch.
- `ruff check` / `ruff format` clean. `generate:source-configs` regenerated `ClickhouseCloudSourceConfig`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

[PostHog/posthog.com#18554](https://github.com/PostHog/posthog.com/pull/18554) adds the source doc.

## 🤖 Agent context

- Authored with Claude Code following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills.
- Modeled on the Anthropic source (closest shape: org-level cost reporting with date windows), but yields plain `list[dict]` batches instead of a source-level `Batcher` per current skill guidance.
- Chose a bespoke transport over `rest_source.RESTClient` because the API has no pagination and needs date-window chunking; skipped the Prometheus metrics endpoint (point-in-time text-format scrapes don't fit warehouse tables well) – can be added later if requested.

---


Gilbert09:
CI triage: the failing `Python code quality` (mypy) and `Product tests (warehouse-sources)` checks all traced to one issue inherited from the base branch — the scaffold branch's regeneration of `generated_configs.py` turned `StripeAuthMethodConfig`'s option sub-fields required (`stripe_secret_key: str`, `stripe_integration_id: int`), which breaks Stripe test collection and mypy. Only the selected auth option's fields exist in a config dict, so they must stay optional.

Pushed 0176148 restoring the exact class shape master has. Note for the base branch (#71137): anyone re-running `generate:source-configs` will reintroduce this, since Stripe's select-option sub-fields are marked `required=True` in `stripe/source.py` and the generator honors that for the dataclass — the generator likely needs to force select-option children optional.

danielcarletti (approved):
LGTM

PR #71236, opened 2026-07-15
feat(data-warehouse): implement coralogix import source

Gilbert09:
## Problem

Coralogix was scaffolded as a stub in #71137 (registered but hidden, no sync logic). Teams centralizing observability data want their Coralogix logs and trace spans in the warehouse for cross-source reliability, error, and cost analysis alongside product data.

## Why

Part of the eng/support sources batch: turn the scaffolded Coralogix stub into a working, visible connector.

## Changes

Implements the source end-to-end on top of the scaffold branch:

- **Transport (`coralogix.py`)**: Coralogix has no entity-list REST surface or cursor pagination – data comes from the DataPrime query API (`POST /api/v1/dataprime/query`), which streams NDJSON and caps rows per query. The source advances a half-open `[start, end)` time window through the range, bisecting windows that hit the row cap and growing sparse ones (1h initial, 10s min, 24h max). Windows that still hit the cap at the minimum size log a warning instead of truncating silently. Rows are sorted per window so the stream is globally ascending (`sort_mode="asc"`).
- **Tables (`settings.py`)**: `logs` and `spans`, one row per record with metadata/labels flattened to columns and the raw body kept as a JSON string in `user_data`. Both are append-only (immutable telemetry, timestamp watermark) – merge-based incremental would pointlessly re-merge huge immutable tables. Initial sync and full refresh pull the last 7 days.
- **Source (`source.py`)**: `ResumableSource` saving the synced-up-to window boundary to Redis after each yielded window; domain select (7 regional clusters), API key (DataQuerying preset), and query tier (frequent search / archive) fields; `connection_host_fields=["domain"]` so retargeting the cluster re-requires the key; non-retryable 401/403 matching; `lists_tables_without_credentials` + canonical descriptions for the public docs catalog. Ships visible with `releaseStatus=ALPHA` (`unreleasedSource` removed).
- `generated_configs.py` regenerated; SOURCES.md row moved to Implemented.

> [!NOTE]
> Verified against the live API without credentials only: all 7 regional domains resolve and return 403 on a bad key (empty body), and the request/response shapes follow the current Coralogix HTTP query API docs. Row-level details docs don't pin down (whether `startDate`/`endDate` are inclusive, timestamp string precision) are handled defensively: a client-side half-open window filter guarantees no boundary double-count either way, and the timestamp parser accepts ISO (incl. nanosecond fractions) and epoch numbers.

The user-facing doc is in [PostHog/posthog.com#18550](https://github.com/PostHog/posthog.com/pull/18550) (`docsUrl` matches; `audit_source_docs` reports no coralogix issues against that checkout).

Automated tests only (44 passing):

- `tests/test_coralogix.py` – transport behavior against a simulated query endpoint: window walking never double-counts boundary rows even if the API treats both bounds inclusive; capped windows bisect and re-fetch instead of dropping rows; min-size capped windows warn and advance (no infinite bisection); resume restarts inclusively from saved state; timestamp parsing survives nanosecond fractions (else every watermark would stall); NDJSON result/warning/error line handling; retryable vs non-retryable status mapping; tracked session keeps sample capture off (telemetry bodies can hold user secrets).
- `tests/test_coralogix_source.py` – source is visible (no `unreleasedSource` – the regression that kept 166 finished sources hidden), append-only schema advertisement, non-retryable matcher against real HTTPError strings, config/inputs plumbing into the transport, docs catalog rendering without credentials.

Also ran the registry-wide suites (`test_source_config_generator`, `sources/tests`) – the two `test_stripe_config` failures there pre-exist on the base branch. `hogli ci:preflight --fix` passes (markdown check skipped, needs node). Not tested: a real sync against a credentialed Coralogix account.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc added in [PostHog/posthog.com#18550](https://github.com/PostHog/posthog.com/pull/18550).

## 🤖 Agent context

Built with PostHog Code (Claude Code) following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills. Key decisions: append-only instead of merge incremental (immutable telemetry; spans have no documented unique id, logs use `logid`); adaptive window slicing over the async background-query API (bounded memory, far less machinery, works on both tiers); synchronous query endpoint only – background queries for very large archive pulls left as a follow-up; per-window ascending sort so `sort_mode="asc"` watermark checkpointing is genuinely safe. Couldn't curl with real credentials, hence ALPHA and the defensive boundary/timestamp handling noted above.

---


danielcarletti (approved):
LGTM

PR #71241, opened 2026-07-15
feat(data-warehouse): implement klaus import source

Gilbert09:
## Problem

Zendesk QA (formerly Klaus) was a scaffolded warehouse source with no sync logic, hidden behind `unreleasedSource=True`. Teams doing support quality assurance in Zendesk QA couldn't pull their reviews, AutoQA results, or CSAT data into PostHog to join against product analytics.

## Changes

Implements the Klaus source end to end and ships it visible with `releaseStatus=ReleaseStatus.ALPHA`:

- **`settings.py`** – declarative catalog of 10 endpoints from the official Public Export API spec (pub.klausapp.com): `reviews`, `autoqa_reviews`, `autoqa_ratings`, `csat`, `users`, `workspaces`, `quizzes`, `scorecards`, `disputes`, `calibration_sessions`, with per-endpoint primary keys, partition keys, and incremental fields.
- **`klaus.py`** – transport over `make_tracked_session()` (no vendor SDK). Bearer auth against `https://{subdomain}.zendesk.com/qa`, with the subdomain normalized and pinned by a strict character allow-list so a crafted value can't splice a different host into the URL.
- **Resumable pagination** – the source is a `ResumableSource`: page/pageSize pagination checkpoints to Redis after each yielded batch, and workspace fan-out endpoints (`scorecards`, `disputes`, `calibration_sessions`) bookmark the workspace id so a crash resumes at the right workspace and page.
- **Rate limiting** – the public API is aggressively rate limited (on the order of one request per minute), so throttled requests are the norm mid-sync. 429s and 5xxs retry via tenacity honoring `Retry-After`, and the resumable checkpoints make long paced syncs survivable.
- **Incremental sync** – endpoints with a server-side `fromDate` filter (`reviews`, `autoqa_reviews`, `autoqa_ratings`, `csat`, `calibration_sessions`) support incremental mode; the rest ship full refresh.
- **`canonical_descriptions.py`** – table and column descriptions from the API spec, plus `lists_tables_without_credentials = True` so the public docs render the table catalog.
- Regenerated `KlausSourceConfig`, moved klaus into the Implemented table in `SOURCES.md`.

> [!NOTE]
> A few behaviors of the Klaus API aren't publicly documented and couldn't be verified without live account credentials; the code makes the conservative choice in each case, with comments at the decision sites:
> - **Page indexing base is unknown (0 or 1).** The first request omits `page` and derives the next index from the response's pagination echo, which is correct for either base; a stuck echo fails loudly instead of looping.
> - **Which timestamp `fromDate` filters on is unknown.** Each incremental endpoint advertises the earliest stable creation-time field on the row as its cursor, so the watermark can only under-advance (re-fetching an overlap that merge dedupes) rather than skip rows, and a 1 hour overlap window guards an undocumented exclusive boundary.
> - **Response ordering is undocumented**, so every endpoint declares `sort_mode="desc"`: the watermark persists only at successful job end instead of checkpointing per batch against an unverified ordering.
> - **`disputes` exposes an optional `fromDate`** but ships full refresh until the filter can be curl-verified against a live account.

The existing `frontend/public/services/klaus.png` icon is kept (no new icon needed).

### Docs

The user-facing doc (below) follows the documenting-warehouse-sources template and matches `docsUrl=https://posthog.com/docs/cdp/sources/klaus`. It needs to land in the posthog.com repo as `contents/docs/cdp/sources/klaus.md` – I validated it against a posthog.com checkout with `manage.py audit_source_docs` (no klaus findings; the command's overall failure is pre-existing drift from other undocumented sources).

<details>
<summary>contents/docs/cdp/sources/klaus.md</summary>

```markdown
---
title: Linking Klaus as a source
sidebar: Docs
showTitle: true
availability:
  free: full
  selfServe: full
  enterprise: full
sourceId: Klaus
---

<AlphaRelease />

The Klaus connector syncs your Zendesk QA (formerly Klaus) quality assurance data – conversation reviews, AutoQA results, CSAT responses, disputes, scorecards, calibration sessions, and quizzes – into PostHog.

## Prerequisites

You need a Zendesk QA account with API access. Only admins and account managers can create the API token the connector uses.

## Adding a data source

<SourceSetupIntro />

When linking Zendesk QA, you'll need:

- **Zendesk subdomain** – the first part of your Zendesk URL. For `yourcompany.zendesk.com`, enter `yourcompany`.
- **API token** – created by an admin or account manager in Zendesk QA under **Settings > Auto QA and integrations > API**.

## Sync modes

<SyncModes />

The Zendesk QA public API is heavily rate limited, so large initial syncs can take a while. Syncs pace themselves around the rate limit and resume automatically if interrupted. We recommend incremental sync where available to keep follow-up syncs small.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

## Troubleshooting

- **Invalid credentials**: check the subdomain matches your Zendesk URL and that the API token was created by an admin or account manager and hasn't been revoked.
- **Slow syncs**: the Zendesk QA public API enforces a strict rate limit, so a full history import can take hours on large accounts. Later incremental syncs only fetch recent data and finish much faster.

<TroubleshootingLink />
```

</details>

Automated tests only (no live Klaus account was available, so I did not test against the real API):

- `tests/test_klaus.py` (transport): subdomain normalization rejects host-splicing values (catches a regression that would send the bearer token to a non-zendesk.com host); self-calibrating pagination advances correctly for both 0- and 1-indexed servers and for a server-clamped pageSize (each of these would otherwise skip or duplicate pages); resume state is saved only after a yielded batch and drives the next request's page; workspace fan-out injects `workspace_id` (guards composite primary key uniqueness) and resumes mid-fan-out; 429 retries honor `Retry-After`; `fromDate` defaults, watermark overlap, and per-endpoint omission; credential validation status mapping including the 429-doesn't-block case.
- `tests/test_klaus_source.py` (source class): schema catalog incremental flags per endpoint, config field shapes, `connection_host_fields`, resumable manager binding, and argument plumbing into the transport.
- Also ran the registry-wide suites (`sources/tests/`, `test_source_config_generator.py`) – all pass – plus `ruff check`/`ruff format` and `hogli ci:preflight --fix` (0 failures).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc content included above; it needs a companion PR in the posthog.com repo (`contents/docs/cdp/sources/klaus.md`).

## 🤖 Agent context

- Built with Claude Code following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills.
- Endpoint catalog was derived from the official Public Export API swagger spec fetched from pub.klausapp.com, cross-checked against Zendesk's export help article. No connector exists in Airbyte/Fivetran/Stitch to cross-reference, so stream selection follows the export API surface.
- Chose a hand-rolled paginator over the generic `rest_source.RESTClient` because the source needs three behaviors the declarative path doesn't model together: self-calibrating page indexing, workspace fan-out with a resumable bookmark, and `Retry-After`-aware pacing for the ~1 req/min rate limit.
- Regenerating `generated_configs.py` in this environment also mutated `StripeAuthMethodConfig` (its shape depends on Stripe OAuth env vars this sandbox lacks); I reverted that hunk and kept only the klaus addition.

---


danielcarletti (approved):
LGTM

Gilbert09:
Addressed the veria-ai finding in 2009bd80d7165f19203f2ad6450a8f5ab92a837d: both tracked sessions now pass `capture=False` so free-text customer content (review comments, CSAT feedback, conversations) stays out of HTTP sample capture, while request metrics and token redaction are retained. Matches the existing pattern in the aviator and browserbase sources.

PR #71243, opened 2026-07-15
feat(aio): add 'generated' delivery status for target-less eval reports

Radu-Raicea:
## Problem

The **Status** column on evaluation reports shows each run's `delivery_status`. When a report has no delivery target configured, the run was left at `pending`, the model default. In the UI that reads as "delivery still in progress" forever, even though the report generated fine and there was simply nothing to deliver. Raised in a Slack thread while digging into what the Status column values actually mean.

## Changes

Add a distinct `generated` delivery status for the no-target case so the column tells the truth.

- New `DeliveryStatus.GENERATED`, set in `deliver_report` when a report has no delivery targets (previously fell through to `pending`).
- Frontend badges in `EvaluationReportsTab` and `EvaluationReportViewer` render it as a green "Generated" pill.
- Serializer help text and the generated OpenAPI enum updated to include the new value.

> [!NOTE]
> Existing runs already stored as `pending` are left as-is; only new runs get `generated`. No backfill.

Added `test_deliver_report_without_targets_marks_generated` to `test_delivery.py`. It catches the regression where a target-less report is marked `pending` (or any other status) instead of `generated` — no existing delivery test exercised the no-target path (the others all cover email/slack success, failure, and partial failure).

I (Claude, via the PostHog Slack app) could not run the suite or linters in this environment (no flox/drf-spectacular/ruff available), so the OpenAPI type change in `api.schemas.ts` was hand-edited to match generator output rather than produced by `hogli build:openapi`. Worth a reviewer running `hogli build:openapi` to confirm no drift.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude via the PostHog Slack app, directed by Radu. Skills invoked while shaping this: `/django-migrations` (for the `AlterField` choices migration) and `/writing-tests` (before adding the delivery test).

Decisions: chose a new enum value over re-labeling `pending` only at display time, because the run row alone can't distinguish "genuinely pending" from "no target" without the new state. Labeled it green ("success") rather than muted, since a target-less report generating is a healthy terminal outcome. Skipped a data backfill of historical `pending` runs to keep the migration a pure metadata `AlterField`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C087XQ7K9K7/p1784131521779119)*

### DIFF
--- a/posthog/temporal/ai_observability/eval_reports/delivery.py
+++ b/posthog/temporal/ai_observability/eval_reports/delivery.py
@@ -430,7 +430,8 @@ def deliver_report(report_id: str, report_run_id: str) -> None:
         increment_delivery("slack", "failed")
 
     if not had_any_target:
-        report_run.delivery_status = EvaluationReportRun.DeliveryStatus.PENDING
+        # Report generated successfully, but no delivery target is configured, so there is nothing to send.
+        report_run.delivery_status = EvaluationReportRun.DeliveryStatus.GENERATED
     elif all_failed:
         report_run.delivery_status = EvaluationReportRun.DeliveryStatus.FAILED
     elif all_errors:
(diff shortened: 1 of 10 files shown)


brandonleung (approved):
does this free reports that are already in pending state?

Radu-Raicea:
@brandonleung no, but it's fine. It would require a backfill, and it's not worth it.

PR #71244, opened 2026-07-15
feat(personhog): ordered shutdown phases for the lifecycle manager

z0br0wn:
## Problem

The lifecycle manager signals every standard component simultaneously when shutdown begins. For the personhog leader that ordering is a correctness problem with two documented consequences (the e2e harness README tracks both as known defects):

  1. **Every graceful drain black-holes the pod's partitions.** The gRPC server and Kafka producer finish shutting down in ~160ms while the coordination drain needs seconds to hand partitions off. For that whole window the pod is still the registered owner of its partitions with a dead server: writes get UNAVAILABLE, the router retries against the same owner, and callers fail. The harness measures ~70 failed writes per drain, which means every production deploy would briefly black-hole every leader's partitions.
  2. **A drain overlapping a pod death wedges convergence for ~36s.** The coordination component only got a 5s graceful window against a 30s worst-case drain, so re-driven handoffs stalled with no DrainedAck until the pod's lifecycle timeout force-exited it, with the affected partitions unreachable throughout.

## Changes

**Before** – one token, everything signalled at once:

  ```mermaid
  flowchart LR
      T[SIGTERM] --> R[Root token cancelled]
      R --> C[Coordination drain, needs seconds]
      R --> G[gRPC server, gone in ~160ms]
      R --> K[Kafka producer, gone in ~160ms]
      C -.->|drain needs these alive| G
      class T phYellow
      class R phBlue
      class C,G,K phGray
  ```

**After** – phased: the drain completes through a still-serving server, then the rest stops:

  ```mermaid
  flowchart LR
      T[SIGTERM] --> R[Root token cancelled]
      R --> C[Phase 0: coordination drains, 35s window]
      C -->|phase 0 finished| P[Phase 1 token cancelled]
      P --> G[gRPC server stops]
      P --> K[Producer stops]
      G --> O[Observability stage]
      K --> O
      class T phYellow
      class R,P phBlue
      class C,G,K,O phGray
  ```

  - `ComponentOptions::with_shutdown_phase(n)`: phases drain in ascending order. A phase's token is only cancelled once every earlier phase's component has finished (completed, died, or timed out its graceful window, measured from when its phase begins). Components within a phase drain concurrently, and the global shutdown timeout bounds the whole sequence, so a stuck phase advances rather than wedging.
  - **Default behavior is byte-identical for the crate's 13 existing consumers.** Phase 0 components keep receiving the root shutdown token itself, exactly as before, with no hop through the monitor thread. The mechanism generalizes the ordering the crate already had (standard components, then observability).
  - The leader drains coordination in phase 0 with its graceful window raised from 5s to 35s (it must cover the pod's 30s drain timeout), stops the gRPC server and producer in phase 1, and grows its global timeout from 30s to 60s to fit the sequence.
  - The component status enum was previously also named "phase"; it is renamed apart (`ComponentStatus`) so the two concepts stay distinct.
  - The harness README's known-defects ledger is updated: the drain black-hole moves to fixed, the convergence wedge to mostly fixed (its residual, draining pods remaining rebalance targets, is now churn rather than a black hole).

  > [!NOTE]
  > Deploys together with a charts change raising the leader's `terminationGracePeriodSeconds` from 45s to 75s, keeping it above the new 60s global timeout. Both values are safe in either merge order, and the leader currently deploys to dev only.
  
All automated:

  - Three new lifecycle tests, each catching a distinct regression no existing test covers: a phase-1 component signalled before phase 0 finished (the core ordering guarantee), a phase-1 component that completes during an earlier phase's drain wedging its own phase (the one-shot `work_completed` pattern), and a timed-out phase failing to advance the sequence. The 46 pre-existing lifecycle tests pass unchanged, which is the evidence that default behavior did not move.
  - The full personhog-leader suite passes; clippy with `-D warnings` and fmt are clean.
  - e2e harness, same scenarios that document the defects: the graceful-drain run drops from ~70 failed writes to **zero** (6,984/6,984 and 6,851/6,851 across runs), the drain-overlapping-pod-death composite settles in **0.0s across three consecutive runs** where it previously wedged for up to 36s, and the drain+zombie+writer-lag composite passes with zero failed writes. Zero consistency violations everywhere.

## 🤖 Agent context

Built with Claude Code. The defect and its fix direction came out of the e2e harness work (the drain scenario's failed-write signature and the convergence wedge's coordinator log trace). Design alternatives considered: a leader-local composite shutdown future gating server stop on drain completion (rejected: awkward producer handle plumbing and not reusable — the router will want the same ordering), and serial one-at-a-time termination from an ordered list (rejected: the requirement is a partial order, and fully serializing turns shutdown time into the sum of graceful windows instead of the max per stage; serial remains expressible as one component per phase). Phase 0 deliberately shares the root token rather than getting a loop-cancelled one so existing consumers keep exact semantics and timing, including readiness flipping 503 at the same instant the first wave drains.

### DIFF
--- a/rust/common/lifecycle/src/handle.rs
+++ b/rust/common/lifecycle/src/handle.rs
@@ -61,7 +61,9 @@ pub(crate) struct HandleInner {
 
 impl Handle {
     /// Future that resolves when shutdown begins for this handle's tier.
-    /// Standard handles: resolves when global shutdown begins.
+    /// Standard handles: resolves when global shutdown begins — or, for a
+    /// component registered with `with_shutdown_phase(n)` for n > 0, once
+    /// every earlier phase's components have finished.
     /// Observability handles: resolves after all standard components finish.
     /// Use in `tokio::select!` to break out of work loops.
     pub fn shutdown_recv(&self) -> tokio_util::sync::WaitForCancellationFuture<'_> {
--- a/rust/common/lifecycle/src/manager.rs
+++ b/rust/common/lifecycle/src/manager.rs
@@ -88,6 +88,7 @@ impl ManagerBuilder {
             health_poll_interval: self.health_poll_interval,
             shutdown_token: self.external_shutdown_token.unwrap_or_default(),
             observability_shutdown_token: CancellationToken::new(),
+            phase_tokens: HashMap::new(),
             event_tx_slot: Arc::new(OnceLock::new()),
             components: HashMap::new(),
             observability_components: HashMap::new(),
@@ -102,18 +103,21 @@ pub struct ComponentOptions {
     graceful_shutdown: Option<Duration>,
     liveness_deadline: Option<Duration>,
     stall_threshold: u32,
+    shutdown_phase: u8,
     pub(crate) is_observability: bool,
     pub(crate) is_advisory: bool,
     config_errors: Vec<String>,
 }
 
 impl ComponentOptions {
-    /// No graceful shutdown timeout, no liveness deadline, stall threshold 1.
+    /// No graceful shutdown timeout, no liveness deadline, stall threshold 1,
+    /// shutdown phase 0.
     pub fn new() -> Self {
         Self {
             graceful_shutdown: None,
             liveness_deadline: None,
             stall_threshold: 1,
+            shutdown_phase: 0,
             is_observability: false,
             is_advisory: false,
             config_errors: Vec::new(),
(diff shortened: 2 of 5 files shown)


eli-r-ph (approved):
LGTM. One potential issue to keep an eye on as you starting using this:

The graceful shutdown timeouts people configured for each handle in the past were based on all the components initiating shutdown at the same time. With this change you'll need to:

1. Successively reduce shutdown timeouts on later-phase handles, to avoid components timing out while enqueued during graceful sequence
2. Account for new total shutdown time a multi-phase app can take as you set the global k8s termination_grace_seconds (etc.) window for your deploys

PR #71245, opened 2026-07-15
feat(data-warehouse): implement spacelift import source

Gilbert09:
## Problem

Spacelift was scaffolded in #71137 as a hidden stub (`unreleasedSource=True`, no fields, no sync logic). Platform teams warehouse Spacelift data to measure deployment frequency, run durations, drift detection, and policy activity across their Terraform/OpenTofu/Pulumi stacks, so the connector needs a real implementation.

**Why:** part of the batch effort to turn the scaffolded engineering/support source stubs into working, released connectors.

## Changes

Implements the Spacelift source end-to-end and ships it visible with `releaseStatus=ReleaseStatus.ALPHA` (the `unreleasedSource` flag is removed).

- **Transport** (`spacelift.py`): Spacelift's API is GraphQL-only with a two-step auth flow, so this is a bespoke client on `make_tracked_session()` rather than the generic REST client. It exchanges the API key id/secret for a ~10h JWT via the `apiKeyUser` mutation (refreshing proactively before expiry and once more on an `unauthorized` mid-sync), retries 429/5xx/transient network failures with capped exponential backoff, and pages Relay-style `search*` connections (`first`/`after`, `pageInfo`), checkpointing the cursor to Redis after each yielded page so Temporal can resume mid-table.
- **Endpoints** (`settings.py`): `stacks`, `runs`, `policies`, `contexts`, `modules`, `worker_pools`, `spaces`, `managed_entities`. `runs` is account-wide (no per-stack fan-out needed) and supports incremental sync on `createdAt` via the server-side `timeInRange` predicate, with a 7-day re-pull window so runs that change state after first sync converge via the delta merge; it reports `sort_mode="desc"` so the watermark only persists at successful job end. Everything else is a small catalog synced as full refresh.
- **Security**: the account name is interpolated into the request host, so it is validated as a plain DNS label and declared in `connection_host_fields` (changing it forces re-entering the API secret).
- `canonical_descriptions.py` from the official docs, `lists_tables_without_credentials=True` so the public docs render the table catalog, `SOURCES.md` moved to Implemented (HTTP GraphQL, tracked transport).

> [!NOTE]
> Field selections, pagination shape, the `timeInRange` predicate, the `searchModules` 50-row page cap, and the null-user token-exchange failure mode were all verified against the live GraphQL endpoint (schema validation runs before auth, so this works without credentials). What could not be verified without an account: whether the `createdAt` predicate on `searchRuns` filters semantically (the Spacelift web UI sends exactly this predicate, so confidence is high) and the API's default run ordering (assumed newest-first like the UI; `sort_mode="desc"` keeps the watermark safe in either order). Both are noted in code comments.

- `tests/test_spacelift_source.py` (source level): released/ALPHA config shape, secret field flags, `connection_host_fields` (guards the credential-retargeting hole), non-retryable error matching for the three stable auth failure messages (and non-matching for transient ones), schema catalog with only `runs` incremental, credential validation and pipeline argument plumbing including dropping the watermark on full refresh.
- `tests/test_spacelift.py` (transport level): account-name host-injection rejection, query document generation per endpoint, epoch coercion of watermark values, lookback and zero-clamp on the incremental predicate, token exchange failing on Spacelift's null-user response, JWT refresh-once-then-permission-error on `unauthorized`, 429/5xx retry, cursor pagination with state saved after yield, resume-from-saved-cursor, `RunWithStack` flattening, predicate presence/absence per sync mode, and the `hasNextPage` without `endCursor` infinite-loop guard.
- All 72 tests pass (`pytest`, 3.3s); `ruff check`/`format` and `hogli ci:preflight --fix` clean. No migration needed (enum landed with the scaffold).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

User-facing doc written for posthog.com (`contents/docs/cdp/sources/spacelift.md`) following the canonical source-doc template; it needs to land in the posthog.com repo. `docsUrl` points at `/docs/cdp/sources/spacelift` and `manage.py audit_source_docs` passes for this source against that checkout.

## 🤖 Agent context

Built with Claude Code following the `implementing-warehouse-sources`, `documenting-warehouse-sources`, and `writing-tests` skills. Key decisions: bespoke GraphQL client instead of `rest_source.RESTClient` (token exchange + GraphQL documents), `ResumableSource` with per-page cursor checkpoints, incremental only on `runs` (the one stream with a verified server-side time filter), full refresh for the small catalogs, and `sort_mode="desc"` for `runs` since ascending order could not be verified without credentials. The live endpoint was used as a schema oracle (GraphQL validation errors precede auth), which caught the `searchModules` page cap and confirmed every field selection.

---


danielcarletti (approved):
LGTM

PR #71253, opened 2026-07-15
chore(conversations): add local Slack setup skill

xljones:
## Problem

There's no written guide for connecting a real Slack workspace to local PostHog Conversations, so testing the SupportHog Slack integration locally means rediscovering the same walls every time: the unconfigured OAuth client id, redirect-uri mismatches, white-screen callbacks, and free-tier tunnel rate limits. The general `docs/internal/slack-local-setup-guide.md` covers the PostHog Code / notifications Slack app, not the conversations `SUPPORT_SLACK_*` flow.

## Changes

Adds `products/conversations/skills/setting-up-support-slack-locally/` — an agent skill that captures the whole path end to end. Key idea it front-loads: OAuth and inbound events have opposite reachability needs, so keep the app and OAuth on `localhost` and point only Event Subscriptions at a public tunnel. `references/troubleshooting.md` catalogs the failure modes we actually hit (unconfigured client id, redirect mismatch, missing bot user, white page = tunnel not rewriting the `Host` header, network errors = free-tier rate limits) and a signed-`url_verification` curl to verify the event path without touching Slack.

Also removes a stray `products/conversations/.gitignore` whose only line (`skills/`) was hiding the skills directory — every other product tracks `products/*/skills/` as the source the build distributes.

I (Claude, via PostHog Code) wrote the skill while actually working through this setup on a local stack, so the steps and the troubleshooting table reflect real failures and fixes rather than theory. Validated with `hogli lint:skills` (118 passed; the skill is clean) and `hogli build:skills` (builds from the conversations product). No automated tests apply to a skill.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The skill itself is the doc; it cross-links the existing `docs/internal/slack-local-setup-guide.md`.

## 🤖 Agent context

Built with PostHog Code (Claude Opus). Skill invoked: `/writing-skills`. The content was distilled from a live debugging session getting the conversations Slack integration working locally, so the ordering of the troubleshooting table mirrors the order the walls appeared. Kept it Slack-focused rather than a broad "support channels" skill — the tunnel/`SITE_URL` mechanics generalize to Teams/email and can be added as `references/` later if useful.

---


products/conversations/skills/setting-up-support-slack-locally/references/troubleshooting.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Verify the event path without Slack
+
+Because `validate_support_request` (`products/conversations/backend/support_slack.py`) is standard Slack
+HMAC signing, you can replay exactly what Slack sends on save — a signed `url_verification` — and confirm
+the whole path (tunnel → Caddy → Django → signature check → handler) works. A `{"challenge": ...}` response
+with `server: granian` proves it end to end and isolates any problem away from Slack's config.
+
+```bash
+SECRET=$(grep '^SUPPORT_SLACK_SIGNING_SECRET=' .env.local | cut -d= -f2)
```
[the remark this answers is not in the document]
xljones: Good point, addressed in 1b4b076. Expanded the `403` bullet to call out that the backend uses the Constance-stored signing secret when one exists, so signing with the `.env.local` value can 403 even against a correctly configured backend. It now says to sign with the active (Constance) value if one is set, or clear it so the env default applies.

products/conversations/skills/setting-up-support-slack-locally/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+2. **OAuth & Permissions → Bot Token Scopes** — the flow requests these (from `SUPPORTHOG_SLACK_SCOPE` in
+   `products/conversations/backend/api/slack_oauth.py`): `channels:history`, `channels:read`, `chat:write`,
+   `chat:write.customize`, `groups:history`, `groups:read`, `reactions:read`, `users:read`,
+   `users:read.email`.
+3. **OAuth & Permissions → Redirect URLs** — add `http://localhost:8010/api/conversations/v1/slack/callback`
+   and Save. If Slack refuses a plain-http localhost URL, use the tunnel URL for the callback too and log in
+   once on the tunnel origin (see [references/troubleshooting.md](references/troubleshooting.md)).
+4. Copy the Client ID, Client Secret, and Signing Secret from Basic Information into `.env.local` (Step 1).
+
+## Step 3 — tunnel for inbound events
```
veryayskiy, line 80: We need to say which Event Subscriptions bot events to enable (message.channels / message.groups, app_mention, reaction_added, member_joined_channel, member_left_channel — matching SUPPORT_EVENT_TYPES). Request URL alone passes url_verification but never delivers tickets.

veryayskiy (approved):
see the comment. stamped not to block

products/conversations/skills/setting-up-support-slack-locally/SKILL.md
xljones: fixed in [`6c98073` (this PR)](https://github.com/PostHog/posthog/pull/71253/changes/6c98073ec3d2a64fda210cbb5258cde7638e6ef5). 

Given the link to where they're defined in case of future changes too.

PR #71254, opened 2026-07-15
fix(ci): preserve selective storybook file filters

pauldambra:
## Problem

Draft Storybook runs can select a reduced shard matrix, but the visual regression container does not include `jq`. The file parser therefore produces no positional test filters and the reduced shards run the full suite.

**Why:** Selective runs should execute only the affected stories so draft CI remains fast and predictable.

## Changes

- Parse the affected-story JSON with the Node runtime already installed in the job.
- Fail explicitly when selective mode produces an empty file list instead of silently running every story.

- Ran a local smoke test covering null-delimited paths, including a path containing a space.
- Ran pinned `actionlint` 1.7.12 against `.github/workflows/ci-storybook.yml`.
- Ran `bin/hogli lint:workflows`.
- Ran `bin/hogli ci:preflight --fix` with the branch even with `master`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation change is needed because this fixes internal CI execution behavior.

## 🤖 Agent context

PostHog Code investigated engineering analytics and GitHub Actions logs, then implemented the targeted workflow fix. The `/authoring-ci-workflows`, `/diagnosing-ci-and-merge-bottlenecks`, and `/querying-posthog-data` skills guided the investigation, workflow conventions, and validation approach.

---

### DIFF
--- a/.github/workflows/ci-storybook.yml
+++ b/.github/workflows/ci-storybook.yml
@@ -458,7 +458,14 @@ jobs:
                   if [ "$MODE" = "selective" ]; then
                       # test-storybook is jest-based; positional path args filter the run to
                       # only those story files before Jest distributes them across shards.
-                      mapfile -d '' -t AFFECTED_FILES < <(jq -j '.[] | ., "\u0000"' <<< "$AFFECTED_FILES_JSON")
+                      affected_files_path=$(mktemp)
+                      node -e 'for (const file of JSON.parse(process.env.AFFECTED_FILES_JSON)) process.stdout.write(`${file}\0`)' > "$affected_files_path"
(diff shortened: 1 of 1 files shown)


pauldambra:
Adding `jq` to the shared Playwright image could make sense if we want it to be a supported container-level dependency. I am not making that change in this PR because it would require changing and rebuilding the shared image, then updating its pinned tag or digest, which broadens the blast radius beyond the selective Storybook regression being fixed here. The current approach uses Node, which this job already installs and requires, and fails closed if parsing produces no files. I would treat adding `jq` to the image as a separate follow-up owned alongside the container.

gantoine (approved):

PR #71272, opened 2026-07-15
fix(replay-vision): prevent daily digest card flash on turn-on

ksvat:
## Problem

On a Replay Vision scanner page, clicking "Turn on daily digest" made the digest hero card flash: it fully disappeared, then reappeared in the "next digest arrives at 8am" state. The blink reads like an error and is jarring right after a deliberate click.

## Changes

The card was unmounting because `createDigest` refetched the whole actions list after the create POST. That refetch drove the card through two `null`-rendering windows in `ScannerDigestCard`:

1. `visionActionsLoading` flips true while `digest` is still absent (the created action isn't in the list until the reload lands) → card returns `null`.
2. Once the list reloads, `loadActionsSuccess` fires `loadLatestRun`, flipping `latestRunLoading` true while `latestRun` is null → card returns `null` again.

The refetch was unnecessary. `visionActionsCreate` already returns the fully serialized created action (including the read-only `next_run_at` the copy needs), and a brand-new digest has no runs. So instead of reloading:

- Added an `addAction` action + reducer to `visionActionsLogic` that appends a created action to the list, mirroring the existing optimistic `toggleActionEnabled` pattern.
- `createDigest` now captures the created object and dispatches `addAction(created)` + `loadLatestRunSuccess(null)` instead of `loadActions()`. `digest` resolves immediately and the run state settles locally, so neither `null` window opens.

The card now swaps in place: the button shows its own loading spinner, then it transitions straight to the "next digest arrives" state.

No backend or serializer changes, so no generated-type regeneration needed.

- Updated the existing one-click-create logic test in `scannerDigestLogic.test.ts` to assert the new dispatch chain (`createDigestSuccess`, `addAction`, `loadLatestRunSuccess`) and, crucially, that after settling `digest` is populated and `latestRunLoading` is false — the two conditions that guarantee no flash. This catches a regression back to the refetch behavior. Ran it: 3/3 pass.
- Typecheck clean for the changed files (regenerated the gitignored kea `*LogicType` files for the new action).
- I (Claude) did not run the app to watch the transition manually — the behavior is covered by the logic-level assertions above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover this card, so nothing to update.

## 🤖 Agent context

Kim reported the flash and directed the fix; I (Claude, via PostHog Code) traced the cause and implemented it. I explored the scanner-page logic to find that the flash came from `createDigest` calling `loadActions()`, then chose the optimistic-insert approach because `visionActionsLogic` already uses that exact pattern for `toggleActionEnabled` and `visionActionsCreate` returns the full created object — so no refetch is needed at all. Considered instead loosening the `visionActionsLoading && !digest` guard in the component, but that only closed one of the two flash windows, so I went with the logic-side fix that closes both.

Skills invoked: `/writing-tests` (before editing the logic test).

---

### DIFF
--- a/products/replay_vision/frontend/replay_scanners/scannerDigestLogic.test.ts
+++ b/products/replay_vision/frontend/replay_scanners/scannerDigestLogic.test.ts
@@ -78,7 +78,7 @@ describe('scannerDigestLogic', () => {
         expect(logic.values.latestRunLoading).toEqual(false)
     })
 
-    it('one-click create sends the digest marker and defaults, then reloads the list', async () => {
+    it('one-click create sends the digest marker and defaults, then settles without a card flash', async () => {
         useMocks(mocksFor([]))
         mountLogic()
         await expectLogic(logic).toFinishAllListeners()
@@ -91,10 +91,12 @@ describe('scannerDigestLogic', () => {
                 },
             },
         })
+        // Optimistic insert + local run resolution instead of a refetch: refetching would blank the
+        // card (digest absent while the list reloads, then latestRunLoading true) and flash it.
         await expectLogic(logic, () => {
             logic.actions.createDigest()
         })
-            .toDispatchActions(['createDigestSuccess', 'loadActions'])
+            .toDispatchActions(['createDigestSuccess', 'addAction', 'loadLatestRunSuccess'])
             .toFinishAllListeners()
         expect(body).toMatchObject({
             name: 'Daily digest: my-scanner',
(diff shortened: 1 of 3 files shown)


ksvat:
* **#71272** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71272?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ksvat:
### Merge activity

* **Jul 15, 6:16 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71272).

PR #71276, opened 2026-07-15
feat(replay-vision): link scanner, observations, and run from alert messages

ksvat:
## Problem

Alert messages delivered by Replay Vision actions were plain text. An alert like "**Alert: AIO Session summaries** — 1 new matching observation since the last check" gave you nothing to click: no way to jump to the scanner, the matching observations, or the full match list. Two smaller gaps in the same flow: the threshold editor said "reaches 1", which doesn't tell you whether that means at least or at most, and on the alert detail page only the timestamp cell of a run row was clickable.

## Changes

Alert messages now link everywhere they can:

- The header's scanner name links to the alert's run page - this alert's message plus every matching observation, with breadcrumbs back to the scanner.
- Each "Most recent matches" example line carries an `[obs N]` citation marker, resolved into an observation link by the existing machinery (the Slack pass server-side, `resolveObservationCitations` in-app). `N` is the observation's 1-based position in the run's persisted `observation_ids`, which is exactly what the run serializer numbers `index` by.
- When more matches fired the alert than the 5 example lines shown, the message appends "See all N matches" linking to the run detail page, which lists every included match.

Example Slack output:

> **Alert: [My scanner](…)** — 12 new matching observations since the last check.
>
> Most recent matches:
> \- (2026-07-15) verdict=yes [[1]](…)
> \- (2026-07-15) verdict=yes [[2]](…)
> …
>
> [See all 12 matches](…)

Implementation notes:

- `_persist_fired` now fetches the observation rows once and derives both the message and `observation_ids` from that single fetch, so the citation numbers can't drift out of alignment with the persisted ids.
- The header and run links are injected after `strip_external_links_markdown` runs (the same pattern the observation citation links already use), because the sanitizer only lets `*.posthog.com` hosts through and would defang `SITE_URL` links on self-hosted or dev instances.
- `_markdown_to_slack` gained a markdown-link to `<url|label>` conversion. Safe by construction: only PostHog-hosted links survive the strip pass, and it runs after mrkdwn escaping.

UI fixes on the alert editor and detail page:

- Threshold copy "reaches" is now "is at least", matching the backend's at-or-above semantics.
- Run rows are fully clickable (pointer cursor plus row-level navigation) and get a right-aligned "View" button, which is a real link so cmd-click works.

- Extended `test_count_alert_fires_and_persists_deterministic_message` to assert the header run-page link, the `[obs N]` markers, the id/label alignment (newest observation = `[1]`), and both resolved Slack links. Catches the links silently dropping or pointing at the wrong page, or a citation pointing at the wrong observation.
- New `test_many_matches_list_only_examples_and_link_to_the_run`: a 7-match alert lists exactly 5 examples and carries a correctly formed run link in both markdown and Slack output, and small alerts don't get the overflow link. Catches the "See all" link regressing into noise or a broken URL.
- Full alert + synthesis suites green (66 tests), oxlint clean on the touched frontend files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Replay Vision actions are still flag-gated (`replay-vision-actions`) and undocumented, so no docs to update.

## 🤖 Agent context

Authored by Claude (PostHog Code) from the PR author's feedback on a live alert she received. Skills invoked: `/writing-tests`. Main decision: reuse the existing `[obs N]` citation pipeline rather than baking URLs into the persisted markdown, and inject the scanner/run URLs post-sanitization so they survive on instances whose `SITE_URL` isn't a posthog.com host. Also considered and rejected listing more examples for big alerts: the message stays capped at 5 examples plus the count, with the run page as the "all matches" destination.

---

### DIFF
--- a/products/replay_vision/backend/temporal/vision_actions/alerts.py
+++ b/products/replay_vision/backend/temporal/vision_actions/alerts.py
@@ -209,11 +209,12 @@ def _persist_fired(
     markdown = strip_external_links_markdown(_alert_markdown(action, alert_config, metric_value, matched_count, rows))
     # Links are added AFTER the strip pass (like the Slack citation links) so the URLs aren't defanged
     # on instances whose SITE_URL isn't a posthog.com host (self-hosted, dev).
-    markdown = _linkify_header_scanner(markdown, action, team.id)
+    run_url = _run_url(team.id, str(action.id), str(run.pk))
+    markdown = _linkify_header(markdown, action, run_url)
     if matched_count > EXAMPLE_LINES:
         # More matches than the message lists as examples — point at the run page, which shows every
         # match this alert included (capped at MAX_OBSERVATIONS).
-        markdown += f"\n\n[See all {matched_count:,} matches]({_run_url(team.id, str(action.id), str(run.pk))})"
+        markdown += f"\n\n[See all {matched_count:,} matches]({run_url})"
     run.synthesized_markdown = markdown
     run.output = {"slack": _markdown_to_slack(markdown, team_id=team.id, observation_ids=observation_ids)}
     run.observation_count = matched_count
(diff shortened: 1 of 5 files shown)


ksvat:
* **#71292** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#71276** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71276?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71276?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ksvat:
### Merge activity

* **Jul 15, 9:25 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71276).
* **Jul 15, 9:26 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71276).

PR #71279, opened 2026-07-15
feat(billing): measure heatmap event volume in daily usage report

pauldambra:
## Problem

Heatmap interactions live in their own `heatmaps` table and never land in `events`, so nothing in the daily usage report has ever counted them. That makes heatmap volume invisible to billing, which is a problem the moment we want to reason about charging for it. Right now we can't even answer "how much heatmap volume does a team generate" from the usage report.

This is measurement first. We want the number before any pricing decision, and we explicitly do **not** want it to show up for customers yet.

## Changes

Add a `heatmap_events_count_in_period` counter to the daily usage report:

- New `get_teams_with_heatmap_count_in_period` query that counts rows in the `heatmaps` table per team for the period (same billing workload/settings as the other counters).
- Wire it through `_get_all_usage_data` and `_get_team_report` into `UsageReportCounters`, and it aggregates to the org level automatically via the existing `dataclasses.fields` loop.
- New `Product.HEATMAPS` tag for query attribution.

> [!NOTE]
> This is a measurement-only counter, like the existing langfuse/helicone event counts. It flows into the internal "organization usage report" telemetry so we can query heatmap volume per team, but there is no billing product mapped to it, so it does not surface to customers. When we're ready to charge, the billing side maps a product onto this field.

Added one ClickHouse-backed test, `TestHeatmapUsageReport::test_heatmap_events_counted_per_team_within_period`. It inserts in-period, out-of-period, and other-team heatmap rows and asserts the team report counts only the in-period rows for the right team, and that the count aggregates to the org report. This catches the regression where the counter reads the wrong table, the wrong time window, or gets wired to the wrong `all_data` key — no existing test exercises the `heatmaps` table in the usage report.

I (actually the PostHog Slack app / Claude) could not run the suite in this environment (no `hogli`, `pytest`, or dev stack available), so I verified the changes compile and lint clean on line length and left CI to run the tests.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior changes, so no docs update.

## 🤖 Agent context

Authored by the PostHog Slack app (Claude) from a Slack thread, directed by Paul. Invoked the repo's `/writing-tests` skill before adding the test to check it against the value gate.

Key decision: keep this measurement-only rather than adding a billable line item. The thread was weighing pricing options (a large divisor like 1000:1 vs parity), and the honest first step is to make the volume visible internally without exposing anything to customers. Modeled the new counter on the existing non-billed measurement counters (langfuse/helicone) so it inherits the "captured in telemetry, invisible to billing UI" behavior for free.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C05LJK1N3CP/p1784124918819529?thread_ts=1784124918.819529&cid=C05LJK1N3CP)*

### DIFF
--- a/posthog/tasks/test/test_usage_report.py
+++ b/posthog/tasks/test/test_usage_report.py
@@ -1293,11 +1293,12 @@ class TestReplayUsageReport(APIBaseTest, ClickhouseTestMixin, ClickhouseDestroyT
 
 
 class TestHeatmapUsageReport(APIBaseTest, ClickhouseTestMixin, ClickhouseDestroyTablesMixin):
-    def _create_heatmap(self, team_id: int, timestamp: datetime, count: int = 1) -> None:
+    def _create_heatmap(self, team_id: int, timestamp: datetime, count: int = 1, session_id: str | None = None) -> None:
+        session_ids = [session_id] * count if session_id else [f"sess_{i}" for i in range(count)]
         rows = ", ".join(
-            f"('sess_{i}', {team_id}, 'user_1', '{timestamp.strftime('%Y-%m-%d %H:%M:%S')}', "
+            f"('{heatmap_session_id}', {team_id}, 'user_1', '{timestamp.strftime('%Y-%m-%d %H:%M:%S')}', "
             f"10, 20, 16, 100, 200, false, 'https://example.com', 'click')"
-            for i in range(count)
+            for heatmap_session_id in session_ids
         )
         sync_execute(
             "INSERT INTO sharded_heatmaps "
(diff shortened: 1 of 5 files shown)


posthog/tasks/test/test_usage_report.py
```diff
@@ -1287,6 +1292,38 @@ def test_usage_report_replay_excludes_deleted_recordings(self) -> None:
         assert report.recording_bytes_in_period == 20  # 2 web * 10 bytes each
 
 
+class TestHeatmapUsageReport(APIBaseTest, ClickhouseTestMixin, ClickhouseDestroyTablesMixin):
+    def _create_heatmap(self, team_id: int, timestamp: datetime, count: int = 1) -> None:
+        rows = ", ".join(
+            f"('sess_{i}', {team_id}, 'user_1', '{timestamp.strftime('%Y-%m-%d %H:%M:%S')}', "
```
pawel-cebula: This test would still pass if the query changed from `count()` to `count(distinct session_id)`, because the three in-period rows all use different `sess_{i}` values. Since the field is meant to measure interaction volume, can we include at least two rows for the same session and keep the expected count at the row count?

pawel-cebula (approved):
Looks good to me. The usage-report wiring is in place, and the US Offline Metabase check makes the raw heatmap query look acceptable for the sampled full day. I left one P3 suggestion to pin the interaction-count semantics in the focused test.

PR #69416, opened 2026-07-08
fix(insights): throttle shared force refresh via cache age override

thmsobrmlr:
## ELI5

Anyone with a shared insight link (no login) can add `?refresh=force_blocking` to make PostHog re-run the query from scratch on ClickHouse. That's expensive, so we only allow it once every 30 minutes. The question is how we remember when the last one happened; today that memory lives in a database table we want to delete.

This PR uses the "best before" label the data already has. Every computed result sits in the cache with a timestamp of when it was made. Instead of asking "when did someone last force a refresh?", we ask the result itself: "are you older than 30 minutes?" If it's younger, you get the existing result; if it's older or missing, we recompute. There's no ticket and no separate memory — the leftovers in the fridge carry their own date. The deleted table was just a duplicate copy of that same timestamp, so semantics are preserved, minus a couple of old bugs (e.g. a freshly created insight no longer waits 30 minutes for its first refresh). In one line: the sibling PR limits *how often you can ask*; this option limits *how often the answer actually gets recomputed, based on how fresh it already is*.

## Problem

The `?refresh=force_blocking` throttle on shared/embedded insights reads its clock from `InsightCachingState.last_refresh` — the last remaining production read of that model, whose original purpose (proactive cache refresh) no longer runs. That Postgres row is a mirror of a timestamp the cached query result already carries.

This is option 2 of 2 candidate approaches — compare with the Redis-slot variant in [#69413](https://github.com/PostHog/posthog/pull/69413).

## Changes

- Shared `force_blocking` now runs as `RECENT_CACHE_CALCULATE_BLOCKING_IF_STALE` with `cache_age_seconds=SHARED_FORCE_BLOCKING_MIN_AGE` (30 min) — the same per-call staleness override Endpoints already use. The cached result's own `last_refresh` becomes the throttle clock: a cache younger than the window is served as-is, anything older (or missing) recomputes synchronously.
- `shared_insights_execution_mode` returns `(mode, cache_age_seconds)`; `calculate_for_query_based_insight` grew a pass-through `cache_age_seconds` param (the `process_query_dict` → `run()` plumbing already existed).
- Removed `_last_refresh_for_shared_gate` and the `InsightCachingState` read from the insight API.

[...]

PR #71280, opened 2026-07-15
chore(insights): drop obsolete posthog_insightcachingstate table

thmsobrmlr:
## Problem

Moving a dashboard tile could crash with an `IntegrityError`. `move_tile` hard-deletes the stale soft-deleted tile occupying the destination's unique slot, and #70453 removed the `InsightCachingState` model from Django state while intentionally keeping the `posthog_insightcachingstate` table and its FK alive for the rolling deploy. With the model gone Django no longer cascades to those rows, so the surviving `dashboard_tile_id` FK (default `NO ACTION`) rejects the delete.

This is the planned follow-up drop to that state-only removal (see the note in #70453). It fixes the crash by removing the FK entirely, rather than working around it in application code.

## Changes

- New `RunSQL` migration dropping `posthog_insightcachingstate` (which also drops the `posthog_insightcachingstate_lookup` partial index). `DROP TABLE IF EXISTS`, no reverse - the table is obsolete.

The model and all readers/writers were removed in #70453 (stacked on #69416). The table has been write-dead since, and it is a leaf table (no inbound FKs), so a plain drop is safe.

> [!NOTE]
> Deploy window is satisfied: #70453 merged 2026-07-14 and has been live in both prod-us and prod-eu for well over a full deploy cycle. Verified current prod is a descendant of its merge commit (`behind_by: 0`).

Compared to patching `DashboardTile.delete()`, dropping the table fixes the `move_tile` crash and the latent bulk/cascade-delete variants at once, and leaves no interim cleanup code to remove later.

No local DB or Django in the authoring sandbox, so I could not run the suite or `sqlmigrate` locally - CI will run the migration. The migration is a pure `RunSQL` drop with no state operations, so `makemigrations` reports no drift by construction. Verified deploy timing via the project's GIT deploy annotations and GitHub commit ancestry, and confirmed no remaining references to the table across Python, `nodejs`, `rust`, `go`, and `services`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed. Internal cache bookkeeping table, no user-facing change.

## 🤖 Agent context

Thomas asked for alternatives to #71266, then asked me (Claude) to investigate whether just dropping the table now was safe, and to implement it. I invoked `/django-migrations` and followed its retire-a-table procedure (state-only `DeleteModel` already shipped in #70453, this is the `RunSQL` drop). I checked deploy safety with `/checking-deploy-timing` against the project's deploy annotations and confirmed prod ancestry of the #70453 merge commit, and grepped for external (non-Django) writers and inbound FKs before concluding a plain drop is safe.

---

### DIFF
--- a/products/product_analytics/backend/migrations/0003_drop_insightcachingstate_table.py
+++ b/products/product_analytics/backend/migrations/0003_drop_insightcachingstate_table.py
@@ -1,0 +1,20 @@
+# Drops the posthog_insightcachingstate table. The model was removed from Django state in
+# 0002_delete_insightcachingstate (#70453, deployed 2026-07-14), which kept the table alive
+# for one deploy cycle so in-flight old pods could keep writing. That window has long passed:
+# nothing reads or writes the table, and the surviving FK on dashboard_tile_id was the last
+# thing making DashboardTile hard-deletes fail. Dropping the table removes it for good.
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("product_analytics", "0002_delete_insightcachingstate"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="DROP TABLE IF EXISTS posthog_insightcachingstate;",
+            reverse_sql=migrations.RunSQL.noop,  # obsolete table, no reverse
+        ),
+    ]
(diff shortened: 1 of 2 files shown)


sampennington (approved):

thmsobrmlr:
### Merge activity

* **Jul 16, 10:00 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71280).
* **Jul 16, 10:01 AM UTC**: @thmsobrmlr merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71280).

PR #71314, opened 2026-07-15
feat(replay-vision): ship pipeline logs into the Logs product

arnohillen:
## Problem

Replay Vision runs a multi-step async Temporal pipeline (rasterize a recording, upload to Gemini, run the scan, emit the event and side effects) on 5-minute schedules. It fails in a lot of ways: Gemini upload errors, rasterize failures, LLM validation retries, quota-exhausted skips, orphaned workflows. Today every one of those signals lands in `structlog` output that only goes to stdout and infra-level log aggregation. None of it is queryable in the product we ship for exactly this: PostHog Logs.

We already dogfood a lot of the stack from Replay Vision (LLM analytics, feature flags, metrics, Signals, surveys, CDP). Logs was the biggest genuinely-unused product, and this pipeline is the ideal first consumer.

## Changes

I (well, Claude) added a first-party log lane that mirrors the pipeline's existing `structlog` output into the Logs product, with **zero new log call sites**.

- **`posthog/otel_logs.py`** — a generic `OtelLogHandler`, the logs counterpart of the existing `posthog/otel_metrics.py`. It ships records over OTLP to the in-cluster `capture-logs` `/i/v1/logs` ingest, keyed per-process (fork-safe under gunicorn/celery) and per-service. It maps a structlog event dict to a Logs record: `event` becomes the body, scalar fields become stringified attributes (the logs pipeline only indexes string-typed values), and `service.name` is pinned on the record so the read side can filter on it.
- **`products/replay_vision/backend/temporal/logs.py`** — attaches that handler to the `products.replay_vision.backend.temporal` logger namespace under `service.name = replay-vision`. Because structlog routes through stdlib logging, every existing pipeline log call is caught by propagation.
- **`products/replay_vision/backend/temporal/__init__.py`** — installs the bridge at import (idempotent), so it activates wherever the worker imports the pipeline.
- **`posthog/settings/ingestion.py`** — adds `OTLP_LOGS_INGEST_TOKEN`, pairing with the existing `OTLP_LOGS_INGEST_ENDPOINT` (mirrors the metrics token pattern).

> [!NOTE]
> This is a hard no-op until both `OTLP_LOGS_INGEST_ENDPOINT` and `OTLP_LOGS_INGEST_TOKEN` are set (via charts, pointed at the internal project). Nothing changes for any deployment that does not opt in. Fully backwards compatible.

Before:

```mermaid
flowchart LR
    A[Temporal pipeline<br/>structlog calls] --> B[stdout / infra log aggregation]
    class A phBlue
    class B phGray
```

After:

```mermaid
flowchart LR
    A[Temporal pipeline<br/>structlog calls] --> B[stdout / infra log aggregation]
    A --> C[OtelLogHandler<br/>service.name=replay-vision]
    C --> D[capture-logs /i/v1/logs]
    D --> E[PostHog Logs product]
    class A phBlue
    class B phGray
    class C phBlue
    class D phRed
    class E phYellow
```

Automated tests only (I could not exercise a live OTLP ingest or verify records landing in a real project, since that needs the charts config above):

- `posthog/test/test_otel_logs.py` (14 cases) — the core handler. Each targets a concrete regression: no-op when un/half-configured (an unauthenticated export would 401 and drop records), the event-dict to body plus stringified-attributes contract, severity mapping, non-scalar attributes dropped, the plain-stdlib-message path, exception info attached (losing stack traces would defeat the point), fail-soft emit (a telemetry throw must never crash the activity), and the `trust_env=False` egress-proxy bypass (the same Smokescreen 407 that shipped unnoticed on the metrics and CI-logs lanes).
- `products/replay_vision/backend/tests/test_logs_bridge.py` (2 cases) — idempotent install attaches exactly one handler under the right service name, and a record from a logger under the pipeline namespace actually propagates into Logs (a wrong namespace string would silently ship nothing).

All 16 pass. Invoked the `/writing-tests` gate before writing them. `hogli ci:preflight --strict` is green (lint, format).

## Automatic notifications

- [ ] Publish to changelog?

Neither. This is internal telemetry plumbing with no user-facing surface.

## Docs update

No user-facing docs. Internal dogfood lane, so `skip-inkeep-docs` applies.

## 🤖 Agent context

Built with Claude Code (Fable 5). I asked it to review the PostHog product stack, find the biggest product Replay Vision does not dogfood yet, and plan before implementing. It ran a repo-wide inventory (Replay Vision already dogfoods LLM analytics, flags, metrics, Signals, surveys, CDP), surfaced Logs as the strongest gap, and I picked it.

Skills invoked: `/writing-tests` (before authoring the tests).

Decisions along the way:

- **Modeled on two proven precedents.** The provider/gating shape mirrors `posthog/otel_metrics.py`; the Logs-specific correctness details (stringify attributes so they are queryable, pin the resource so records do not land under the pod's `OTEL_SERVICE_NAME`, bypass Smokescreen with `trust_env=False`) come from the existing `engineering_analytics` CI-logs emitter, which is the one shipped example of feeding the Logs product from a backend product.
- **Bridge, not new call sites.** Rather than rewrite ~34 files, the handler attaches to the pipeline's logger namespace and catches structlog output by propagation. One install point, reversible, additive to the existing console/JSON handlers.
- **Kept it one PR.** The generic core module has no other consumer yet, so splitting it out would ship dead code and double the CI/review overhead for a single-team change.
- **INFO floor.** The bridge ships INFO and above (you want to see the pipeline working, not only failures). Raising it to WARNING later is a one-line change if volume/cost warrants.


fasyy612 (approved):
sweeeet

PR #71323, opened 2026-07-15
fix(aio): price Anthropic cache writes by TTL

### LINKED ISSUE #71306: Anthropic 1-hour prompt-cache writes use the 5-minute rate in auto-costing
## Bug Description

PostHog's automatic cost calculation cannot distinguish Anthropic 5-minute and 1-hour prompt-cache writes.

Anthropic reports an aggregate cache-write count plus a TTL breakdown, and supports both TTLs in one request:

```json
{
  "cache_creation_input_tokens": 248,
  "cache_creation": {
    "ephemeral_5m_input_tokens": 148,
    "ephemeral_1h_input_tokens": 100
  }
}
```

PostHog currently represents one aggregate count, `$ai_cache_creation_input_tokens`, and one `ModelCost.cache_write_token` price. [`calculateInputCost`](https://github.com/PostHog/posthog/blob/a6ae50156b3535f6f7360d915d332f48b55b63cc/nodejs/src/ingestion/pipelines/ai/costs/input-costs.ts#L180-L185) therefore applies one write rate to all 248 tokens. The configured rate and fallback are the Anthropic 5-minute rate (`1.25 ×` base input price), while 1-hour writes cost `2 ×`.

Let `p` be the base input price. For the synthetic usage above:

- Expected cache-write cost: `148 × 1.25p + 100 × 2p = 385p`
- Current calculation: `248 × 1.25p = 310p`
- Difference: `75p`, or `19.48%` of the correct cache-write cost

For a request containing only 1-hour writes, the cache-write component is under-reported by `37.5%`.

## How to reproduce

1. Use the Anthropic usage payload above.
2. Represent it with the current PostHog event property: `$ai_cache_creation_input_tokens: 248`.
3. Run automatic model-based cost calculation.
4. Compare the resulting cache-write component (`310p`) with Anthropic's TTL-specific calculation (`385p`).

The breakdown cannot be represented by the current [`ModelCost`](https://github.com/PostHog/posthog/blob/a6ae50156b3535f6f7360d915d332f48b55b63cc/nodejs/src/ingestion/pipelines/ai/costs/providers/types.ts#L1-L14) or [custom-pricing properties](https://posthog.com/docs/ai-observability/calculating-costs#option-1-custom-price-per-token), so a single custom cache-write price cannot make mixed-TTL requests exact.

## Additional context

Anthropic documents the response fields, mixed-TTL requests, and the `1.25 ×` / `2 ×` pricing multipliers in its [prompt-caching documentation](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#1-hour-cache-duration).

One backward-compatible direction is to keep `$ai_cache_creation_input_tokens` for existing events and add TTL-specific counts and rates. Existing aggregate-only events would retain today's behavior; events with the breakdown would price each bucket separately. I am happy to contribute the implementation, focused regression tests for 5-minute, 1-hour, mixed, and legacy events, and the corresponding documentation after confirming the preferred property names.

## Debug info

```shell
Source-level reproduction against PostHog master at a6ae50156b3535f6f7360d915d332f48b55b63cc.
Synthetic provider usage only; no project data is required.
```

gouveags:
## Problem

Anthropic prices 5-minute cache writes at `1.25 ×` the input-token rate and 1-hour writes at `2 ×`. PostHog currently collapses both into one aggregate count, so automatic costing applies the 5-minute rate to 1-hour writes.

Closes #71306

## Changes

- Preserve Anthropic's 5-minute and 1-hour cache-creation counts from direct raw usage and `@posthog/ai`'s Vercel wrapper (`usage.raw` and `providerMetadata.anthropic.usage`).
- Extract Vercel Amazon Bedrock's TTL breakdown from `providerMetadata.amazonBedrock.usage.cacheDetails`, with legacy `bedrock` alias support and no double-counting when both aliases are present.
- Price a complete TTL breakdown separately without double-counting the aggregate. Aggregate-only and partial-breakdown events keep the existing path.
- Register, validate, and document the new token-count fields and custom-pricing precedence.

For custom pricing, `$ai_cache_write_1h_token_price` wins for 1-hour writes. If it is absent, `$ai_cache_write_token_price` applies to both TTLs. Catalog pricing keeps the existing 5-minute rate and derives the 1-hour rate from `2 ×` the resolved prompt price. No model prices are hardcoded.

> [!NOTE]
> This corrects core ingestion when an event supplies a complete TTL breakdown through raw Anthropic usage, Vercel Amazon Bedrock metadata, or both explicit properties. SDKs that emit only the aggregate keep the existing pricing path until they propagate the breakdown.

- Focused AI costing, extraction, and pipeline suites: 3 suites, 257 tests passed.
- Shared AI token validator suite: 1 suite, 1,183 tests passed.
- Covered 5-minute, 1-hour, mixed-TTL, aggregate-only, partial-breakdown, custom-price precedence, raw extraction, Vercel Bedrock top-level and nested provider metadata, aliases, and sparse TTL buckets, `$ai_usage` removal, process-level pricing, and invalid-property normalization.
- Ran the Node.js TypeScript check with Node 24.
- Regenerated the frontend taxonomy JSON with the repository generator and verified it parses.
- Ran changed-file Prettier and `git diff --check`.

I did not run a full PostHog stack or a full production build. The change is covered at the calculation, extraction, validation, and `processAiEvent` boundaries.

## Automatic notifications

- [x] Publish to changelog?
- [x] Alert Sales and Marketing teams?

## Docs update

Updated the AI generation event reference with the TTL-specific count fields, complete-pair fallback rule, aggregate invariant, and custom-price precedence. Registered the fields in the event taxonomy.

## 🤖 Agent context

I directed the fix and used Codex for implementation, tests, validation, and review. The repository's `/writing-tests` skill shaped the regression-first test contract. No public session link is available.

The implementation stays in the existing ingestion pipeline and continues using the existing model catalog. It derives the 1-hour rate from the resolved prompt price, preserves aggregate-only events, and leaves SDK-specific propagation as follow-up work.

## Companion Python SDK work

PostHog/posthog-python#746 preserves the TTL breakdown across supported Python integrations. It is intentionally a draft until this server contract merges. The SDK gap is tracked in PostHog/posthog-python#745.


richardsolomou (commented):
Thanks for the contribution! Could we add one TTL-breakdown case using numeric strings? The validator accepts them, but the pricing-path tests only use numbers, so a regression in `hasNumericProperty` could silently fall back to aggregate pricing.

gouveags:
@richardsolomou Thanks for catching this. I added a numeric-string TTL breakdown case in `aa11d718e5` and verified that it still prices the 5-minute and 1-hour counts separately instead of falling back to the aggregate. Does that cover the regression you had in mind?

nodejs/src/ingestion/pipelines/ai/costs/modality-tokens.ts
```diff
@@ -210,6 +233,7 @@ export const extractModalityTokens = (event: EventWithProperties): EventWithProp
             if (!isObject(metadata)) {
                 return
             }
+            extractAnthropicCacheCreation(metadata)
```
richardsolomou, line 280: Could you handle the usage shape emitted by `@posthog/ai`’s Vercel wrapper here? It captures `$ai_usage` as `{ usage: result.usage, providerMetadata }`, while Anthropic V3 stores the TTL breakdown under `usage.raw.cache_creation` and `providerMetadata.anthropic.usage.cache_creation`.

Neither path is traversed before `$ai_usage` is deleted, so mixed TTL writes are priced entirely at the 5-minute rate. Please add support for these paths and an end-to-end regression case.

Wrapper source: https://github.com/PostHog/posthog-js/blob/da1d781999407b72420f6ee9fdb9e9e5c46d4170/packages/ai/src/vercel/middleware.ts#L517-L542

richardsolomou (changes requested):

nodejs/src/ingestion/pipelines/ai/costs/modality-tokens.ts
gouveags: Thanks again for catching another inconsistency. You’re right, I was only covering the direct Anthropic usage shape because that was the shape behind the issue that originally prompted this PR. I hadn’t realized the Vercel wrapper exposed the same breakdown through these other shapes.

I've added support for the TTL breakdown under both `usage.raw.cache_creation` and `providerMetadata.anthropic.usage.cache_creation`. The `processAiEvent` regression test now uses a wrapper-shaped `$ai_usage` payload with the same breakdown in both places, so it covers extraction, cleanup, and pricing without counting the tokens twice. The original direct usage shape remains covered, and the focused extractor and `processAiEvent` test suites pass.

Does this cover the case you had in mind, or is there anything else you'd like me to include?

gouveags:
@richardsolomou Your last review prompted me to take another close pass over the provider payloads, and I noticed we weren't covering Vercel's Amazon Bedrock `cacheDetails` shape.

I've added support for its 5-minute and 1-hour TTL buckets under both `amazonBedrock` and the legacy `bedrock` alias. When both aliases are present we read the canonical one once, and sparse responses fill the missing bucket with zero so pricing doesn't fall back to the aggregate. The extractor and `processAiEvent` tests now cover the aliases, sparse responses, and end-to-end mixed-TTL pricing.

Let me know if you'd like me to cover anything else.

richardsolomou (approved):
Could we also cover `rawUsage.providerMetadata.amazonBedrock`? That supported nested path currently has no regression test.

gouveags:
I added the missing `rawUsage.providerMetadata.amazonBedrock` case with both 5-minute and 1-hour `cacheDetails` entries. It verifies that both TTL buckets are extracted before `$ai_usage` is removed, and the focused extractor suite passes.

gouveags:
@richardsolomou Thanks again for the reviews. It seems like the remaining step is for you or another maintainer to approve the check workflows so the required pipeline can run, and then merge if everything passes.

I also opened the companion draft PR PostHog/posthog-python#746. I have been running into this exact issue in our application, and we need both the ingestion and Python SDK sides covered so the same reporting bugs do not continue through that path. We use PostHog to track our users' LLM usage and inform product and business decisions, so accurate cost data is important to us. If there is someone on the Python SDK or LLM analytics side I should coordinate with, I would really appreciate a pointer.

Please let me know if there is anything else you have in mind or any other way I can help.

Thanks for the whole review process and for the chance of contributing!

richardsolomou:
@gouveags Ran CI, couple small issues to sort out then should be good to go

gouveags:
@richardsolomou No worries! It looks like I can’t do much from my side, so I’ll wait for the checks to be rerun. Thanks again for all the help!

gouveags:
@richardsolomou this seems finally ready to be merged

richardsolomou:
@gouveags Last thing, our repository only accepts verified commits. Could you verify your commits and update the branch? GitHub’s guide [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

gouveags:
@richardsolomou I didn't know that all commits were required to be verified, really sorry. I'll keep this in mind for future contributions. I've fixed the branch history. Can you check again?

PR #71331, opened 2026-07-15
chore: add saml_relay_state to idp config mdoel

ReeceJones:
## Problem
Right now the RelayState parameter for SAML is derived from the org domain model's primary key. This was missed during the initial decoupling of idp config fields from the org domain model.

## Changes

- **Database migration**: Added `saml_relay_state` CharField (max 36 chars, nullable) to `IdentityProviderConfig` model in migration 1257
- **Model field**: Added `saml_relay_state` field to `IdentityProviderConfig` with appropriate validators and defaults
- **Backfill command**: Created `backfill_identity_provider_config_saml_relay_state` management command that:
  - Populates `saml_relay_state` from the linked `OrganizationDomain.id` for configs linked to exactly one domain
  - Skips configs with zero or multiple linked domains (ambiguous cases)
  - Supports `--dry-run` flag for safety
  - Is idempotent (skips already-populated values)
- **Admin interface**: Updated `IdentityProviderConfigAdmin` to:
  - Include `saml_relay_state` in the read-only fields list
  - Add `saml_relay_state` to the "SAML Configuration" fieldset for visibility
- **Tests**: Added comprehensive test coverage for the backfill command covering:
  - Successful backfill when linked to exactly one domain
  - Skipping configs with no linked domains
  - Skipping configs linked to multiple domains
  - Dry-run mode verification
  - Idempotency (not overwriting existing values)

Automated tests verify:
- The backfill command correctly identifies and updates configs linked to a single domain
- Ambiguous cases (zero or multiple linked domains) are properly skipped
- The dry-run flag prevents database writes
- Existing values are not overwritten (idempotent behavior)

The migration follows standard Django patterns and the admin interface changes are display-only (read-only fields).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Human-driven (agent-assisted)

---


posthog/models/identity_provider_config.py
```diff
@@ -35,11 +35,14 @@ class IdentityProviderConfig(ModelActivityMixin, UUIDModel):
     updated_at = models.DateTimeField(auto_now=True)
 
     # ---- SAML attributes ----
-    # Field shapes intentionally mirror `OrganizationDomain` (including nullability) so
-    # values can be copied verbatim while domains remain the source of truth.
     saml_entity_id = models.CharField(max_length=512, blank=True, null=True)
     saml_acs_url = models.CharField(max_length=512, blank=True, null=True)
     saml_x509_cert = models.TextField(blank=True, null=True)
+    saml_relay_state = models.CharField(
```
yasen-posthog, line 41: **Configs created after this merges will never get a relay state.**

`saml_relay_state` isn't in `IDP_CONFIG_SYNCED_FIELDS`, and `sync_identity_provider_config_from_domain` builds new configs from exactly that tuple (line 173 in this file). Omitting it from the tuple is correct — `OrganizationDomain` has no matching column to mirror from — but the consequence is that every config created from here on is NULL, and the backfill is a point-in-time fix for today's rows only.

So whatever lands next either needs a write path (set it at creation, or give the field a `default`) or an explicit fallback to `config.id` in the reader. Worth deciding now rather than discovering it when the reader ships and newly configured SAML orgs can't log in.

posthog/models/identity_provider_config.py
```diff
@@ -35,11 +35,14 @@ class IdentityProviderConfig(ModelActivityMixin, UUIDModel):
...
+        max_length=36,
+        blank=True,
+        null=True,
```
yasen-posthog, line 44: **No uniqueness constraint or index.**

If this becomes the key that routes an assertion to a tenant, two configs sharing a value is an ambiguity in a pre-auth path. The codebase already has to defend against exactly this shape for `saml_entity_id` — `_recover_idp_initiated_relay_state` in `ee/api/authentication.py:129` refuses to act unless the issuer match is unique, with a comment noting that `saml_entity_id` has no uniqueness constraint.

Backfilled values inherit uniqueness from domain PKs, but nothing preserves that afterward. Postgres allows multiple NULLs under a unique constraint, so `unique=True` costs nothing here, and the table is small enough that adding it now is free — much less so once the reader depends on it.

posthog/models/identity_provider_config.py
```diff
@@ -35,11 +35,14 @@ class IdentityProviderConfig(ModelActivityMixin, UUIDModel):
     updated_at = models.DateTimeField(auto_now=True)
 
     # ---- SAML attributes ----
-    # Field shapes intentionally mirror `OrganizationDomain` (including nullability) so
-    # values can be copied verbatim while domains remain the source of truth.
     saml_entity_id = models.CharField(max_length=512, blank=True, null=True)
     saml_acs_url = models.CharField(max_length=512, blank=True, null=True)
     saml_x509_cert = models.TextField(blank=True, null=True)
+    saml_relay_state = models.CharField(
+        max_length=36,
```
yasen-posthog, line 42: `max_length=36` is UUID-string length, and the backfill only ever writes a UUID. If it's always a UUID, `UUIDField` says so and stores it more cheaply; if it isn't always a UUID, then 36 is an arbitrary cap that deserves an explanation.

posthog/models/identity_provider_config.py
```diff
@@ -35,11 +35,14 @@ class IdentityProviderConfig(ModelActivityMixin, UUIDModel):
...
+        blank=True,
+        null=True,
+    )
```
yasen-posthog, line 45: No `help_text` or comment, unlike every neighbor on this model. This field's semantics are the least obvious of the lot — a reader's first question is "why isn't this just `config.id`?" and there's nothing in the code that answers it. The admin renders the field too, so `help_text` earns its keep here.

posthog/models/identity_provider_config.py
```diff
@@ -35,11 +35,14 @@ class IdentityProviderConfig(ModelActivityMixin, UUIDModel):
     updated_at = models.DateTimeField(auto_now=True)
 
     # ---- SAML attributes ----
-    # Field shapes intentionally mirror `OrganizationDomain` (including nullability) so
-    # values can be copied verbatim while domains remain the source of truth.
     saml_entity_id = models.CharField(max_length=512, blank=True, null=True)
```
yasen-posthog, line 38: The deleted comment ("Field shapes intentionally mirror `OrganizationDomain` (including nullability) so values can be copied verbatim while domains remain the source of truth") still applies to `saml_entity_id`, `saml_acs_url`, and `saml_x509_cert` — only the new field escapes it. Repo convention is to preserve comments unless the change makes them obsolete; scope it to the three mirrored fields rather than dropping it.

posthog/management/commands/backfill_identity_provider_config_saml_relay_state.py
```diff
@@ -0,0 +1,54 @@
...
+            if len(domain_ids) > 1:
+                skipped_multiple_domains += 1
+                self.stdout.write(f"  config={config.pk} linked from multiple domains; skipping (not 1:1)")
+                continue
+
+            updated += 1
+            self.stdout.write(f"  config={config.pk} -> saml_relay_state={domain_ids[0]}")
+            if not dry_run:
+                config.saml_relay_state = str(domain_ids[0])
+                config.save(update_fields=["saml_relay_state"])
```
yasen-posthog, line 48: **`config.save()` here fires the domain mirror as a side effect.**

`IdentityProviderConfig.save()` is overridden to always call `sync_domains_from_identity_provider_config(self)` regardless of `update_fields`, which issues an `UPDATE` writing all eight synced IdP fields onto every linked domain. So a command whose stated job is "set one column" also force-pushes the config's SAML/SCIM/XAA values onto its domains. In a consistent state that's just a redundant write per config; if the config is stale relative to the domain, it silently overwrites the domain's IdP settings.

Sidestep it with a queryset update:

```python
IdentityProviderConfig.objects.filter(pk=config.pk).update(saml_relay_state=str(domain_ids[0]))
```

posthog/management/commands/backfill_identity_provider_config_saml_relay_state.py
```diff
@@ -0,0 +1,54 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from posthog.models.identity_provider_config import IdentityProviderConfig
+
+logger = logging.getLogger(__name__)
```
yasen-posthog, line 7: This logger is never used — the command writes through `self.stdout` throughout. If you do want logging, the sibling `sync_identity_provider_configs.py` and this model both use structlog rather than stdlib `logging`.

posthog/management/commands/backfill_identity_provider_config_saml_relay_state.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class Command(BaseCommand):
+    help = (
+        "Backfill IdentityProviderConfig.saml_relay_state from the linked OrganizationDomain's id. "
+        "Only populates configs linked from exactly one domain; configs linked from zero or "
+        "multiple domains are left untouched. Idempotent (skips configs already populated)."
+    )
+
+    def add_arguments(self, parser):
```
yasen-posthog, line 17: Missing type annotations here and on `handle` below. The sibling `sync_identity_provider_configs.py` annotates both (`def add_arguments(self, parser) -> None` and `def handle(self, *args: Any, **options: Any) -> None`), and CLAUDE.md asks for mypy-strict-shaped signatures on new code.

yasen-posthog (changes requested):
The migration is safe and the backfill is careful about ambiguous links, but I'd like the field's design settled — uniqueness, and what populates it for configs created after this lands — and the backfill's unintended domain-mirror write removed, before this merges.

PR #71335, opened 2026-07-15
feat(hogql): tag queries with a structural ast fingerprint

andyzzhao:
## Problem

Triaging poor-performing agent/human-written HogQL queries is difficult because queries with the same pattern are not easily grouped together. The current mechanism to group queries together uses ClickHouse's `normalizeQuery`, which is lexical, so it only collapses literals but nothing else. 

So the following two queries would not collapse:

```sql
SELECT event, count() AS cnt
FROM events
GROUP BY event
ORDER BY cnt DESC
```

```sql
select e.event, COUNT() total
from events e
group by 1
order by 2 desc
```

Alias names, keyword case, table-alias qualification (`e.event` vs `event`), and positional `GROUP BY 1` / `ORDER BY 2` references all keep otherwise-identical queries apart. Measured over 7 days of US traffic, ad-hoc HogQL fragments into ~800k near-singleton shapes, so pattern-level pain ranking structurally cannot see it.

## Changes

- `posthog/hogql/fingerprint.py`: rewrites the query AST into a canonical form and hashes it. Literals become `?`, alias and CTE names are dropped or renamed positionally (references to them, including `GROUP BY 1` / `ORDER BY 2`, are resolved to the real expression first), function names are lowercased when it is safe (`COUNT()` merges with `count()`), and table qualifiers are stripped when the select has one source (`e.event` merges with `event`) or renamed positionally when it has joins, so alias spellings merge without losing which source a column came from. Literal lists and template placeholders collapse to one placeholder (`IN ('a','b')` merges with `IN ('a','b','c')` and `IN {ids}`), `properties['foo']` folds into the field chain so it merges with `properties.foo` while keeping the property name, and `count(*)` merges with `count()`. The version string is part of the hash, so changing the rules later starts a new series instead of silently mixing with the old one.
- `posthog/hogql/query.py`: computes the fingerprint at query time, where the AST is already in hand, and adds it to `log_comment` via `tag_queries`. Wrapped in try/except so it can never break a query.
- `posthog/clickhouse/query_tagging.py`: new optional `hogql_fingerprint` tag field.

On 2,000 real production queries, 1,548 distinct query texts become 746 fingerprints, at about 0.1 ms per query once the process is warm.

Follow-ups in separate PRs: add the field to the query_log_archive export allowlist, then group the query-pain dashboard by it.

- New `posthog/hogql/test/test_fingerprint.py`: parameterized pairs that must share a fingerprint and pairs that must not, plus determinism, version, and lazy-join guards.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed: internal observability tagging; no user-facing behavior change.

## 🤖 Agent context

Written by Claude Code, directed by the assignee, as part of a query-pain prioritization effort: a pain dashboard over query_log_archive showed the top-50 normalized shapes capture only ~29% of user-facing wait time because ad-hoc SQL fragments under lexical normalization. Design reuses existing machinery on purpose: `CloningVisitor`, `Expr.to_hogql()`, the function registry for case canonicalization, and the same execution-time hook as `extract_hogql_features` (which is also how the fingerprint reaches log_comment with no new pipeline). The `/writing-tests` skill gate was applied to the new test file. Known limitation discussed and deferred: identifier-level differences (e.g. filtering on different properties) intentionally remain distinct patterns in v1.

### DIFF
--- a/posthog/clickhouse/query_tagging.py
+++ b/posthog/clickhouse/query_tagging.py
@@ -491,6 +491,9 @@ class QueryTags(BaseModel):
 
     hogql_features: Optional[HogQLFeatures] = None
 
+    # Structural fingerprint of the HogQL AST (literals/aliases/positional refs canonicalized) for grouping query patterns.
+    hogql_fingerprint: Optional[str] = None
+
     modifiers: Optional[object] = None
     number_of_entities: Optional[int] = None
     person_on_events_mode: Optional[str] = None  # PersonsOnEventsMode
--- a/posthog/hogql/fingerprint.py
+++ b/posthog/hogql/fingerprint.py
@@ -1,13 +1,6 @@
-"""Structural fingerprint for HogQL queries.
-
-Two queries share a fingerprint when they do the same work with different
-values: literals (including the length of literal lists), alias and CTE
-names, formatting, single-source table-alias qualification, bracket vs dot
-property access, count(*) vs count(), and positional GROUP BY / ORDER BY
-references are canonicalized away, while tables, columns, functions,
-operators, and which join source a column comes from are kept. Tagged into query_log
-so query patterns can be grouped for performance triage regardless of how
-each caller happened to write the SQL.
+"""Structural fingerprint for HogQL queries: spelling differences (literals,
+alias names, formatting, qualification) are canonicalized away and the result
+hashed, so query_log rows can be grouped into patterns for performance triage.
 """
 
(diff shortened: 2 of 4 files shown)


andyzzhao:
on a random sample of queries clickhouses `normalizeQuery` gets within 0.2% of collapse with this fingerprint mechanism so I don't think this is worth it

PR #49129, opened 2026-02-25
fix(insights): add confirmation dialog and dashboard redirect on insight delete

sampennington:
## Problem

https://posthog.slack.com/archives/C0ACRAMJUAG/p1771954132325999

When coming from a dashboard to delete an insight there are two problems:
- No confirmation 
- No feedback it's deleted and you don't return to dashboard. 

## Changes

- Adds a confirmation dialog
- Redirect to dashboard on delete

https://github.com/user-attachments/assets/d48040b9-87e3-473d-a964-d6ba370ed69f

## How did you test this code?
Playwright test

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Publish to changelog?
no

PR #21146, opened 2024-03-25
fix: allow user to go to settings before a product is set up

zlwaterfield:
## Problem

Right now there is a bug where a user can't go to any settings page except `/settings/project` when they don't have a product set up. This is because the `url.settings()` returns `/settings/project` by default and doesn't take into account all the different settings pages that user could be trying to hit. 

## Changes

The fix is to just check for any settings pages.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Does this work well for both Cloud and self-hosted?

Yes

PR #17857, opened 2023-10-08
fix: early access management cancel button

kiran-4444:
## Problem
Fixes: #17841 

## Changes

- image

- Also, fixed "cancel" button behaviour while creating and editing early access feature

## How did you test this code?
Locally in my browser

PR #4564, opened 2021-06-02
[Fix] Annotations list not refreshing after creating annotation

alexkim205:
## Problem

See #4277.

## Changes

- Fix cypress test that was asserting on existence of text in `<textarea>` and not annotations `<table>`.
- Refactor `createGlobalAnnotation` into its own loader to expose `createGlobalAnnotationSuccess` to `annotationLogic`
- Typescriptify `annotations/logic.js`

## Checklist

- [X] N/A All querysets/queries filter by Organization, by Team, and by User
- [X] N/A Django backend tests
- [X] N/A Jest frontend tests
- [X] N/A Cypress end-to-end tests
- [X] N/A Migrations are safe to run at scale (e.g. PostHog Cloud) – present proof if not obvious

## Testing

- Once the test was fixed, it correctly failed because the annotation didn't show up in the table (reproduces #4277). This PR fixed that test.
- `./bin/e2e-test-runner`

Before

https://user-images.githubusercontent.com/13460330/120408728-73803680-c304-11eb-8d50-8c1b058eac7c.mov

After

https://user-images.githubusercontent.com/13460330/120408559-200de880-c304-11eb-8eb4-577e07d79586.mov

PR #71342, opened 2026-07-15
chore(ci): prune redundant playwright e2e specs

gantoine:
## Problem

PostHog is moving AI-first. Developers increasingly work through the MCP server and API via coding agents, which lowers the value of end-to-end tests that click through general product UI. Meanwhile the Playwright suite had grown skewed and expensive: 58 spec files with ~37% of test cases concentrated in insights/dashboards, several suites that mock the entire backend (so they prove nothing a Jest component test wouldn't), and a real flake burden (retries=3 plus a nightly flake-audit job, and a cluster of dirty-state retry workarounds in the heaviest specs).

This prunes the suite against `playwright/README.md`'s own bar: test what only a browser can prove; smoke checks and API-mocked flows belong in Jest, Storybook, or backend tests.

> [!IMPORTANT]
> **Scope narrowed after team review.** The first iteration also trimmed the insight/dashboard/persons/notebooks lifecycle tests (save/discard/edit state machines, dashboard delete/template/duplicate flows, side-panel actions). Reviewers pushed back: those flows exercise the interplay of multiple kea logics mounted together on a live scene against a real backend, which single-logic Jest tests can't reproduce — and an internal analysis mapping recent bug fixes to the test layer that would have caught them confirmed the trimmed lifecycle flows correspond to real, recently shipped regressions. Those tests are now fully restored; only cuts with no mapped bug and a clear lower-layer owner remain.

## Changes

Suite goes from **58 spec files to 37**. Net diff is ~18.5k lines removed, most of it giant mock JSON fixtures.

**Deleted outright**

| Spec(s) | Why |
| --- | --- |
| `toolbar`, `web-analytics` | Both only exercised the shared `AuthorizedUrlList` form. Near-identical, no product behavior asserted. |
| `annotations`, `project-home-page`, `before-onboarding` | Shallow "page renders / create via form / nav reachability" smoke checks. |
| `batch-exports` (5), `cdp/hog-function-backfills`, `logs` (2), `billing` (2), `events` | Fully API-mocked via `page.route` — they never hit the real backend, so they test React wiring a Jest+RTL+MSW test would cover faster. Backend truth is already covered by product pytest suites and kea logic tests. |
| `api-insight-creation`, `insight-json-tab` | API-create-then-render check (with a disabled screenshot) and a JSON-tab display check; no mapped regressions, and the render pipeline is covered by the kept exact-number insight tests. |
| `experiments/experiment-metric-validation` | A pure API test (no page navigation). Its empty/whitespace event-name regression is already covered by `test_blank_event_names_pass_validation_on_create/update` in the experiments backend suite, so nothing was lost. |
| `insights/lifecycle` | Render/save smoke check with no computed-output assertions. |

**Restored after review** (byte-identical to master): `dashboards.spec.ts`, `trends.spec.ts`, `funnels.spec.ts` (minus one test, below), `insight-side-panel.spec.ts`, `persons.spec.ts`, `notebooks.spec.ts`, and the create→search→open test from `insights-list.spec.ts`.

**Still trimmed** (each cut has no mapped bug in the recent-regressions analysis and a clear lower-layer owner):

- `funnels`: dropped only the left-to-right/top-to-bottom layout toggle test — a purely visual chrome check.
- `insights-list`: kept create→search→open (insight search has broken before); dropped delete-from-list and duplicate-from-list row-menu CRUD.
- `cohorts`: kept create. Dropped the 120s dynamic→static duplicate (serializer-level behavior).
- `surveys/crud`: kept one focused test for the browser-only editor behavior (question type → scale → live preview re-render, plus display-condition property targeting). Dropped the serial create→launch→stop→delete chain and two config-persistence tests covered by `test_survey.py` and the surveys Jest files.

**Replacement coverage for the deleted billing specs**

Billing is a flagged critical path, so rather than lose the two mocked billing specs I re-created their intent at cheaper, non-flaky layers. Between them these cover everything the deleted specs did, plus the wiring:

- `billingProductLogic.test.ts` (kea logic) — the data/state logic the specs implicitly relied on:
  - `customLimitUsd` / `hasCustomLimitSet` treat a **$0 limit as a real limit** (drop all usage), not as "no limit" (unbounded spend). This is the `=== 0` handling the deleted "set limit to $0" e2e test guarded.
  - Unsubscribe survey response state stays **isolated per product type** (the logic is keyed by `product.type`) — what the deleted "maintains unique state between product types" e2e test guarded.
- `BillingLimit.test.tsx` (RTL render) — the DOM-to-API wiring the logic test can't see:
  - Clicking "Set a billing limit", entering a value, and saving PATCHes `/api/billing` with the value under `custom_limits_usd`.
  - Clicking "Remove limit" PATCHes `/api/billing` with a null limit.

I mutation-checked all of it: breaking the `=== 0` guard, replacing `key()` with a constant, and removing the Save/Remove submit binding each turns the relevant assertion red — so none of the assertions are vacuous.

**Coverage layering after this PR**

| Layer | Covers | Billing |
| --- | --- | --- |
| Storybook (image snapshot) | does it render | ✅ billing page + unsubscribe modal render |
| kea logic test (`billingProductLogic.test.ts`) | data/state logic | ✅ $0 handling, per-product survey isolation |
| RTL render test (`BillingLimit.test.tsx`) | click → action → API wiring | ✅ set + remove limit PATCH |

**Housekeeping**

- Removed the orphaned fixtures under `playwright/mocks/` (billing, events, logs) left behind by the deleted specs.
- Updated `tools/playwright_area_map.json` so affected-only spec selection no longer points at deleted specs, and every surviving spec stays reachable (the restored `insights-list.spec.ts` got its `saved-insights` scene entry back).

> [!NOTE]
> New critical-path coverage (2FA login, invite accept, API key lifecycle, project deletion danger zone, flag evaluation, real event capture → queryable) is stacked on this PR in #71488. Remaining known gaps for future work: SSO/SAML, session expiry, org/account deletion, member role/permission boundaries, onboarding first-event verification. The recent-regressions analysis also suggests a chart tooltip interaction spec (hover/dismiss/unmount lifecycle) would have caught a cluster of real bugs — none of the current suite covers it.

- Ran the full restored set locally against the real dev stack (`dashboards`, `trends`, `funnels`, `insight-side-panel`, `persons`, `notebooks`, `insights-list`) with `--retries 0`. All pass except four tests blocked by local-stack gaps rather than spec issues: my ClickHouse lacked the versioned funnel UDFs CI installs (loading `latest_user_defined_function.xml` fixed the funnel-math tests), the CSV/XLSX export download and the funnel persons-modal actors query never resolve on my stack, and the "view source" test is blocked by the MCP-hint toast, which never auto-closes in dev builds (`AUTO_DISMISS_MS` is `false` when `NODE_ENV === 'development'`) — CI's production build dismisses it after 15s. All four run green in CI on master, and the restored files are byte-identical to master; the only in-file edits are two test deletions, which can't affect sibling tests.
- `hogli test` on both new Jest tests — `billingProductLogic.test.ts` (5/5) and `BillingLimit.test.tsx` (2/2). Mutation-checked each: reverting the `=== 0` guard, the `key()`, and the Save/Remove submit binding each turned the relevant assertion red, confirming the tests catch the regressions the deleted billing specs caught.
- `tools/test_playwright_spec_selection.py` — passes. This enforces that every area-map target resolves to a real spec and every surviving spec stays reachable, so it's the direct guard on the map edits.
- `oxlint` + `oxfmt` on all edited specs and both new tests — clean.
- `hogli ci:preflight --strict` — 0 failures.
- CI runs the complete surviving suite once this PR leaves draft (drafts skip the e2e workflow; editing `playwright/` and `tools/` paths is in the area map's `force_full` set), which is the authoritative validation for this change.

No Playwright coverage was lost without a cheaper replacement, a confirmed existing lower-layer test, or (for the remaining trims) zero mapped regressions in the recent bug-fix analysis.

## 🤖 Agent context

Georges-Antoine asked for an audit of the Playwright suite through the lens of PostHog going AI-first, then to execute the elimination side of it. I (Claude) did the exploration, audit, and this prune. After team review, he shared reviewer feedback plus a bug-fix-to-test-layer analysis, and asked me to restore the contested trims.

Skills invoked: `/writing-tests` (the mandatory gate before changing tests — its value bar and cost-ladder framed every keep/trim/delete decision, and drove the billing replacements down to the cheapest layer that catches each regression).

- The first iteration over-indexed on "covered by kea logic tests" for the insight/dashboard lifecycle trims. Reviewers correctly pointed out that single-logic tests can't reproduce the interplay of the entangled logics mounted together on those scenes, and the bug-fix analysis backed that up with real shipped regressions in exactly those flows — so those trims are reverted wholesale rather than selectively.
- The remaining cuts were each checked against that same analysis: no mapped bug, plus an identified lower-layer owner, before staying deleted.
- Before deleting the mislabeled `experiment-metric-validation` API test, I confirmed its regression is already parameterized in the experiments backend suite, so it's a safe delete rather than a port.
- Both `logs` specs turned out to be fully API-mocked (one even has an empty `E2E tests` placeholder), so they went with the other mocked suites rather than being kept.
- Billing was the one deletion I judged worth replacing rather than deferring, since it's a flagged critical path. The two mocked e2e specs became a kea logic test (data logic) plus a small RTL test (DOM wiring), which together cover more than the originals did and run in ~2s instead of a browser boot.

---


rnegron (commented):
the "Why" table makes sense to me... and I'm fully in favor of removing not-useful tests.

hmm ultimately I think this should be up to the respective teams? like if batch-exports finds their removed tests important for a good reason, then likely they should stay.

frontend/src/scenes/billing/BillingLimit.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        await userEvent.type(input, '2000')
+        await userEvent.click(screen.getByTestId('save-billing-limit-product_analytics'))
+
+        await waitFor(() => expect(patchedBody).toEqual({ custom_limits_usd: { product_analytics: 2000 } }))
+    })
+
+    it('removing an existing limit PATCHes billing with a null limit', async () => {
+        await seedBilling({ product_analytics: 500 })
+        render(
+            <Provider>
```
pawel-cebula, line 95: This loses the part of the old billing-limit spec that proved the UI refreshes to the saved value. The PATCH mock returns the original `billingSeed`, and the test stops at `patchedBody`, so a regression where the success/reload path no longer renders the updated limit, especially `$0`, would still pass. Can we have the mock return the updated billing payload and assert the rendered state after save/remove?

playwright/e2e/billing/billing.spec.ts
```diff
@@ -1,84 +0,0 @@
...
-                status: 200,
-                body: billingContent,
-            })
-        })
-
-        // Visit the billing page
-        await page.goto('/organization/billing')
-    })
-
-    test('Show and submit unsubscribe survey', async ({ page }) => {
```
pawel-cebula, line 32: The replacement coverage keeps the per-product survey state, but the deleted test also covered the actual submit path: modal -> reason -> `/api/billing/deactivate` -> modal closes. I don't see a Jest/RTL/logic replacement for that path in this PR, and it wires into `deactivateProductSuccess` / survey submission / billing reload. Can we keep a focused submit test or add a lower-layer equivalent for this flow?

gantoine: thanks for flagging these! added tests to cover the gaps 🫡

rnegron (approved):

playwright/e2e/early-access-management.spec.ts
```diff
@@ -1,66 +0,0 @@
-import { delay } from 'lib/utils/async'
-
-import { randomString } from '../utils'
-import { PlaywrightWorkspaceSetupResult, expect, test } from '../utils/workspace-test-base'
-
-test.describe('Early Access Management', () => {
```
haacked, line 6: `suggestion`: This removes the only E2E test for `earlyAccessFeatureLogic.ts`'s create/save/edit/delete flow. While backend tests (`test_deleting_early_access_feature_removes_feature_enrollment_from_flag`, `test_cant_soft_delete_flag_with_early_access_feature`) cover the flag-coupling behavior, the UI wiring has broken before: PR #56424 fixed a flaky save in `earlyAccessFeatureLogic.ts`, and PR #56513 fixed the edit button's timing in `EarlyAccessFeature.tsx`. The spec also doesn't appear in the PR description's deletion table, unlike the other specs cut.

Add a kea-logic test for `earlyAccessFeatureLogic` covering create → save-as-draft → edit → delete, mirroring how `billingProductLogic.test.ts` replaced the billing specs elsewhere in this same PR.

haacked (approved):
Reviewed only the feature-flags-relevant parts of this PR (early access features), not the full e2e prune. One suggestion inline.

thmsobrmlr (approved):

gantoine:
## Coverage audit: verifying every pruned case has a lower-layer owner

I traced each deleted/trimmed test back to its origin commit and then checked whether the behavior it asserted is actually covered by a backend test, a kea logic / RTL+MSW test, or a Storybook snapshot. Two categories needed action; both are now addressed in this branch.

### 1. Regression guards that were added by a bug fix and had no replacement — restored

Four removed tests were originally added by `fix` commits to lock in a specific regression, with no equivalent lower-layer coverage. Restored in `e27c5b7`:

| Test | Guards | Origin fix |
| --- | --- | --- |
| `early-access-management.spec.ts` (cancel flow) | cancel on a new EAF returns to the list | #17857 |
| `annotations.spec.ts` — "Create annotation" | list refreshes after create | #4564 |
| `before-onboarding.spec.ts` | settings reachable before a product is set up | #21146 |
| `insights-list.spec.ts` — "Can delete an insight from the list" | the "Delete insight?" confirmation dialog | #49129 |

The new `earlyAccessFeatureLogic.test.ts` covers save/edit/delete at the logic layer, so only the cancel path (which lives in the component's `onClick`, not the logic) needed the e2e back.

### 2. Deletions whose behavior had no real lower-layer coverage — gaps closed at the cheapest level

The audit found three behaviors the "covered elsewhere" rationale didn't actually hold for. Closed in `9b908f7`, each mutation-checked (breaking the guarded source line turns the assertion red):

- **Backfill creation** — `batchExportBackfillModalLogic.tsx` had **no test file at all**; its submit path (date transform → `api.createBackfill` → success toast → required-date validation) was only exercised by the deleted `batch-export-runs`/`batch-export-backfills` specs. New kea logic test `batchExportBackfillModalLogic.test.ts` (8 cases).
- **Feature-flag properties tab** — `test_property_definition.py` covers the backend data split, but nothing asserted the frontend surfaces a separate taxonomic group. Added a case to `taxonomicFilterLogic.test.ts` asserting the `EventFeatureFlags` group endpoint filters `is_feature_flag=true` and event properties filter `is_feature_flag=false`.
- **Survey responses limit** — no test set `responses_limit` to a non-null value and read it back. Added a round-trip create case to `test_survey.py`.

### 3. One browser-only test restored

`toolbar.spec.ts` "Toolbar loads" asserts the real toolbar iframe (`#__POSTHOG_TOOLBAR__ .Toolbar`) injects into the page — genuinely browser-only, reproduced at no lower layer. Restored (with its area-map entries). The add-URL form/save logic in that spec remains covered by `authorizedUrlListLogic.test.ts`.

### Everything else checks out

The remaining deletions have a confirmed lower-layer owner and are safe:

- **Billing** (both specs, incl. the `$0` limit case and set/remove PATCH) → `billingProductLogic.test.ts` + `BillingLimit.test.tsx`
- **experiment-metric-validation** → `test_blank_event_names_pass_validation_on_create/_on_update` (a superset)
- **cohorts / insights-list duplicate** → `cohortEditLogic.test.ts` + `test_cohort.py`, `savedInsightsLogic.test.ts`
- **funnels layout, batch-export config/logs/metrics, cdp backfills** → logic tests + Storybook snapshots
- **api-insight-creation, events filter, project-home-page, logs facets, surveys lifecycle** → backend/logic tests, or the kept exact-number insight specs

The `logs-alerts` save-then-enable regression logic is covered by `logsAlertDetailSceneLogic.test.ts`; only the browser-level 400/toast behavior is dropped, which is acceptable.

PR #71348, opened 2026-07-15
fix(agent-platform): keep MCP startup from initiating OAuth

dmarticus:
## Problem

The Agent Builder declares the first-party PostHog MCP as its authoring transport. MCP clients open during session startup, but identity resolution previously combined credential lookup with authorization initiation. Simply entering the builder could therefore generate an OAuth link even though PostHog Code should reuse the signed-in user's authentication.

The supplied PostHog bearer was also restricted to the regional API host and rejected when sent to the matching first-party MCP host.

OAuth callback routing is a separate concern. [#71535](https://github.com/PostHog/posthog/pull/71535) and [PostHog/charts#13238](https://github.com/PostHog/charts/pull/13238) handle intentional OAuth flows by routing callbacks through each agent's own host.

## Changes

- Make MCP startup identity resolution lookup-only. It can reuse trigger or previously linked credentials, but cannot initiate OAuth.
- Preserve the identity provider on passive auth failures so an agent can offer `@posthog/identity-connect` later, only after the user explicitly asks to use or connect that capability.
- Allow PostHog bearers on the matching first-party regional MCP host.
- Pin the Agent Builder auth order so the user-scoped PostHog bearer is tried before the internal service fallback.
- Leave callback URL construction and deployment routing to the companion PRs.

- Ran TypeScript checks, formatting checks, and lint for `@posthog/agent-runner`, `@posthog/agent-shared`, and `@posthog/agent-tests`.
- Ran the focused MCP client regression suite: 28 tests passed.
- Ran the focused Agent Builder bundle regression suite: 13 tests passed.
- Added prompt coverage for passive startup plus explicit later connection. That test uses the existing S3-backed prompt harness, which could not run locally because object storage was unavailable at `localhost:8333`.
- Ran `git diff --check`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the Agent Builder example documentation to explain the load-bearing auth order and the distinction between the user bearer and the internal service fallback.

## 🤖 Agent context

Codex CLI traced the Agent Builder trigger credential, identity resolution, MCP startup, and callback paths. I used the `/debugging-ci-failures`, `/writing-tests`, and `/writing-skills` skills. An earlier version also changed callback configuration, but that work was removed after reconciling with the agent-host routing implemented by the companion PRs. The final scope keeps startup passive while retaining enough provider context for a user-requested connection later.

---

### DIFF
--- a/products/agent_platform/services/agent-tests/src/examples/agent-builder/agent.md
+++ b/products/agent_platform/services/agent-tests/src/examples/agent-builder/agent.md
@@ -117,12 +117,10 @@ one, refuse and explain why.
 
 1. **Act as the asking user — never as PostHog.** Every PostHog MCP
    call runs with the asking user's PostHog identity. You hold
-   no fallback credential. A `posthog` trigger passes the user's bearer
-   through. A trusted `posthog_internal` trigger proves the caller may open
-   the builder, but may still need a one-time PostHog connection before MCP
-   tools can act as that user (see "Acting as the user"). If a call returns
-   403, that is the user's permissions speaking — surface it, don't work
-   around it.
+   no fallback credential. In PostHog Code / MCP the bearer passes
+   through from the trigger (see "Acting as the user"). If a call
+   returns 403, that is the user's permissions speaking — surface
+   it, don't work around it.
 2. **Never accept raw secrets in chat.** API keys, OAuth tokens,
    passwords. If the user pastes one, tell them not to and reset
    the secret to whatever you'd have used the punch-out flow for.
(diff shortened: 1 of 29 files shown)


products/agent_platform/services/agent-tests/src/examples/agent-builder/agent.md
```diff
@@ -162,7 +162,11 @@ account. Every `posthog__*` MCP call is signed with that user's PostHog
 identity, so what you can see and change is exactly what they can.
 On both of your surfaces (PostHog Code and MCP / IDE) the user's PostHog
 bearer passes through from the trigger — they're already authenticated,
-nothing to link.
+nothing to link. Never tell a PostHog Code user to connect the PostHog MCP;
```
daniloc, line 165: heh

daniloc (approved):
shipppp

PR #70124, opened 2026-07-10
refactor(flags): holdout attachment capability and experiment rule accessor

andehen:
## Problem

Two remaining leaks in the flag facade boundary series (prep for PostHog/requests-for-comments-internal#1097):

1. `holdout_filters_for_flag` in the experiment model hand-built the flag's `holdout` filters field.
2. "What is this experiment's flag config" had no single home: call sites dug into `filters.multivariate` / `groups[0]` independently. Under the upcoming rule-level model, an experiment's config becomes a rule on the flag, so resolution needs exactly one place to swap the v2 arm in.

**Stacked on #70123** (review only the last commit).

## Changes

Pure move and consolidation, no behavior change:

- `set_holdout(current_filters, *, holdout_id, exclusion_percentage)` joins the facade's pure filters transforms. It takes plain values, so flags learns no experiment concepts; the experiment-holdout shape knowledge (first release group's rollout percentage) stays experiments-side as an `ExperimentHoldout.exclusion_percentage` property. The holdout viewset propagation in `ee/clickhouse/views/experiment_holdouts.py` uses it too and stops importing an experiments-model internal.
- New `products/feature_flags/backend/facade/rules.py`: minimal frozen dataclasses `ExperimentRuleConfig` (variants, rollout percentage, assignment unit, holdout ref) and `HoldoutRef`, plus `experiment_rule_from_filters(filters)`, the v1 derivation (flag-wide `multivariate` + `groups[0]` + flag-level holdout/aggregation). Deliberately excludes seed handling and reason mapping, which are open design items. The module is Django-free so model modules can import it without cycles.
- `get_experiment_rule(experiment)` next to the Experiment model is now the one home for experiment-to-rule resolution. The mechanical deep reads route through it (`_ensure_feature_flag`'s existing-flag variants dig, the summary data service). Analysis and query-runner reads are deliberately untouched: the holdout marker representation under the new model is an open decision.

## How did you test this code?

Automated tests only, no manual testing.

- `test_facade.py`: 9 passed (6 base + a 3-case parameterized test for the rule-config derivation: full v1 filters, empty filters, group-without-rollout with null holdout — the derivation had no direct coverage before)
- `test_experiment_service.py`: 455 passed; experiments holdout subset 10 passed
- `ee/clickhouse/views/test/test_experiment_holdouts.py`: 7 passed (attach/update/detach end to end, unchanged)
[...]

PR #70123, opened 2026-07-10
refactor(flags): move enrollment-freeze filter transforms behind the facade

andehen:
## Problem

Freezing experiment enrollment performs group-structure surgery on the flag's `filters`: it ANDs a cohort condition into every release group, stamps marker metadata, and the inverse strips it back out. The predicate "is enrollment frozen" re-derives state from those stamps, and the can-freeze preconditions (super groups and holdouts evaluate before release conditions, group-aggregated flags can't use person cohorts) were recomputed independently by the frontend. All of that flag-evaluation knowledge lived in `products/experiments/`. Part of the flag facade boundary series (prep for PostHog/requests-for-comments-internal#1097).

**Stacked on #70122** (review only the last commit).

## Changes

Pure move plus one additive serializer field:

- New `products/feature_flags/backend/facade/filters.py` with four pure functions: `restrict_groups_to_cohort`, `strip_group_cohort_restriction`, `groups_restricted_to_cohort` (backs `is_exposure_frozen`), and `group_cohort_restriction_blocker` (the consolidated can-freeze flag-shape preconditions). The `EXPOSURE_FROZEN_*` marker keys stay defined in experiments and are passed in as opaque values, so flags learns no experiment concepts.
- The four experiments call sites (freeze, unfreeze, reset, ship) call the facade; `is_exposure_frozen` delegates; a new `can_freeze_exposure` property combines lifecycle state with the flag-side blocker.
- Additive `can_freeze_exposure` field on `ExperimentSerializer` (with `help_text` and `@extend_schema_field`), so the frontend can stop recomputing evaluation-order knowledge. The frontend swap is deferred: `canFreezeExposure` in `experimentActions.tsx` reads optimistically-mutated local state, so switching it to the serialized field needs a staleness audit first.
- The facade `__init__.py` re-exports were removed: `facade.api` imports the flag serializer, whose module imports the experiment model, which now imports the facade. Importing from `facade.api` / `facade.filters` directly breaks the cycle and matches the tasks/notifications facade pattern.

> [!NOTE]
> The only response-shape change is the new `can_freeze_exposure` field. OpenAPI types regenerated (`api.schemas.ts`, MCP generated client, purely additive).

## How did you test this code?

Automated tests only, no manual testing.

- `test_experiment_service.py`: 455 passed (446 base + 9-case parameterized `can_freeze_exposure` property test covering the blocker branches); freeze/frozen/unfreeze/reset subset 61 passed
[...]

PR #67329, opened 2026-07-01
fix(approvals): make feature-flag approval policies non-bypassable

yasen-posthog:
## Problem

Feature-flag approval policies were only enforced on the direct edit path. Every sibling write path — scheduled changes, experiment launch/pause/ship-variant, copy, and create — could change a policy-gated field with no approval.

## Changes

- Make the approval gate fail closed instead of silently skipping when org context is missing
- Detect gated changes from the validated change rather than the request body — closes the experiment and ship-variant paths
- Gate feature-flag create/copy when the new flag lands in a policy-guarded state
- Gate scheduled feature-flag changes at creation; expire ones whose window passes before approval
- Route experiment launch/pause/resume flag flips through the gate so they need approval
- Add a bypass-matrix regression suite covering every enable/disable/rollout path

## How did you test this code?

- ~60 backend unit + integration tests added, incl. a parameterized bypass matrix across all mutation paths
- Verified live on a local stack: prior bypasses blocked, approve→apply flow works
- Agent-authored; only the automated tests above were run.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #71353, opened 2026-07-15
feat(flags): support system writes through the facade

andehen:
## Problem

The feature-flag facade (`products/feature_flags/backend/facade/api.py`) is the single gated write path for flags, but every write required an acting user.
Non-HTTP callers (experiment reset from service code, tasks) had no user to pass, so they fell back to raw `flag.filters = ...; flag.save()` writes that bypass serializer validation and activity logging.
That raw-write escape hatch is exactly what the `feature-flags-no-raw-filters-access` semgrep guard (#70128) is trying to burn down.

## Changes

**System writes through the facade (`user=None`):**

- `create_flag` / `update_flag` (and everything built on them, e.g. `set_flag_active`) now accept `user=None`, meaning a system write:
  - the serializer context gets `ServiceRequest(None)`, so validation, cache refresh, and versioning all still apply
  - activity is logged with `is_system=True` and no user (`log_activity` already attributes user-less entries as system), with `created_by` / `last_modified_by` left null
  - the approval gate does not engage: the facade's request shim explicitly declares `is_system=True`, and `@approval_gate` passes through only requests that declare it — it never infers a system write from a merely absent user (a pre-auth `HttpRequest` also lacks one), so a system write can never raise `ApprovalRequired` while a user-less request without the marker still engages the gate. Approval policies target human-driven changes, and a request-less caller (beat task, pre-delete hook) has no way to surface a 409 or a change request. The skip is explicit in `products/approvals/backend/decorators.py` with a comment saying why.
  - passing `user=None` together with a user-bearing `request` is rejected with a `ValueError` — the request's user would otherwise silently win (gate engages, attribution goes to `request.user`), turning an intended system write into a human-attributed one.
- One tiny null-guard in `EvaluationTagsChecker.is_enabled` so a user-less request short-circuits to disabled instead of raising. Everything else on the serializer path already tolerated a null user (nullable FKs, `report_user_action` no-ops without a user, dashboard template helpers default `user=None`).

**First consumer, `products/experiments/backend/experiment_service.py`:**

- `_clear_frozen_exposure` non-HTTP branch: the raw `flag.filters = stripped; flag.save(update_fields=["filters"])` write is now `update_flag(..., user=None)`. Resets from non-HTTP callers get validation and a system activity-log entry instead of a silent write. Since the gate is skipped, this is safe inside the caller's transaction.
- The variants read in `_ensure_feature_flag` now uses the null-safe model accessor (`existing_flag.variants or list(DEFAULT_VARIANTS)`) instead of a raw `filters` dig, dropping the last `experiment_rule_from_filters` import here.

> [!NOTE]
> Part of the `feature-flags-no-raw-filters-access` baseline burndown (#70128). Originally stacked on #70124 / #70123; those and #70128 have since merged, so this PR now sits directly on master and the rule-yaml edit here removes the baseline exclude lines for `experiment_service.py` and `experiment_summary_data_service.py` for real. Local semgrep with the excludes removed shows 0 findings on both files.

- New facade tests in `products/feature_flags/backend/test/test_facade.py`:
  - `test_system_create_logs_system_activity` / `test_system_update_logs_system_activity`: catch any reintroduced `request.user` dereference on the serializer write path and any regression in system attribution (`is_system=True`, null user) — no existing test exercises a user-less write.
  - `test_system_write_never_raises_approval_required`: approvals feature enabled plus an enabled `feature_flag.disable` policy matching the change; a system disable must pass through and create no `ChangeRequest`. Guards against removing the gate skip, which would make non-HTTP resets raise inside a transaction (rolling back the just-created change request).
- New consumer test `test_reset_experiment_clears_freeze_without_request` in `test_experiment_service.py`: the non-HTTP reset branch strips freeze stamps, persists via the facade, and logs a system activity entry — previously this branch wrote raw and logged nothing, so no existing test could catch the routing regressing.
- Review-feedback tests: `test_system_write_rejects_user_bearing_request` (facade `ValueError` guard); `test_gate_passes_through_for_declared_system_request` and `test_userless_request_without_system_marker_still_engages_gate` in `test_decorators.py` (the gate skips only an explicitly declared system request, and a user-less request without the marker must not apply the write); `test_system_disable_under_policy_applies_without_change_request` in the bypass matrix as the control for the one sanctioned skip.
- Ran locally: `test_facade.py` (35 passed, rerun after the `ValueError` guard and the rebase onto master), full `test_experiment_service.py` (457 passed), frozen-exposure tests in `test_contracts.py` / `test_presentation_api.py` (5 passed), full `products/approvals/backend` suite (248 passed, since the decorator changed).
- Semgrep verification with the baseline excludes for both experiment service files removed: 0 findings on `experiment_service.py`, `experiment_summary_data_service.py`, and the facade.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude Code, directed by @andehen as part of the flag-facade baseline burndown.
Design decision: extended `create_flag`/`update_flag` to accept `user=None` rather than adding a parallel system-write function, keeping one write path as the invariant.
The serializer accommodated `user=None` almost entirely as-is; the only serializer-side change needed was a one-line null-guard in the evaluation-tags feature check.
The approval-gate skip lives in the `@approval_gate` decorator itself (rather than the facade) so every gated write path treats user-less requests consistently.


andehen:
Restacked onto master after #70124 merged. The baseline exclude lines for `experiment_service.py` and `experiment_summary_data_service.py` are now removed in this PR (semgrep: 0 findings on both). One rebase note: the `_validate_existing_flag` variants-accessor swap was dropped — master's `experiment_eligibility_error(feature_flag.variants)` already covers it.

products/approvals/backend/decorators.py
```diff
@@ -499,6 +499,12 @@ def wrapper(self, *args, **kwargs):
                 request = args[0] if args else kwargs.get("request")
                 _, team, organization = _extract_context(self, request)
 
+            # System writes (no acting user, e.g. facade calls from tasks or service code)
+            # bypass the gate: approval policies target human-driven changes, and a
+            # ChangeRequest cannot exist without a requester to attribute it to.
+            if getattr(request, "user", None) is None:
```
haacked: `suggestion`: The approval gate infers a system write from `request.user is None`. That's safe today because only `ServiceRequest(None)` produces it, but the inference is fragile: a plain Django `HttpRequest` also has `request.user is None` before `AuthenticationMiddleware` runs. If a future `@approval_gate` target gets invoked outside DRF's request pipeline, it would silently skip approval, the same bypass class that #67329 built `test_feature_flag_bypass_matrix.py` to catch.

Carrying an explicit signal instead of inferring one from absence would close that: e.g. `ServiceRequest(user, is_system=True)` and gate on `getattr(request, "is_system", False)`, so only a request that deliberately declares itself a system write can take this path.

products/feature_flags/backend/facade/api.py
```diff
@@ -35,10 +41,16 @@ def _serializer_context(team: Team, user: Any, request: Any | None) -> dict:
...
     request shim carrying the acting user — FeatureFlagSerializer needs request.user.
+    ``user=None`` makes this a system write: ``ServiceRequest(None)`` carries no user,
+    which skips the approval gate and logs activity with ``is_system=True``.
 
     Pass BOTH get_team and get_organization so the approval gate resolves the policy
     from context rather than falling back to instance derivation.
     """
+    # A user-bearing request would silently override user=None (gate engages, attribution
+    # goes to request.user) — reject the contradiction instead.
+    if user is None and getattr(request, "user", None) is not None:
```
haacked: `suggestion`: `getattr(request, "user", None) is not None` gets evaluated twice back to back here, once in the `ValueError` guard and once when picking the shim, so a reader has to check both match to trust the guard actually covers the branch below it.

```python
request_has_user = getattr(request, "user", None) is not None
if user is None and request_has_user:
    raise ValueError("user=None is a system write; do not pass a user-bearing request with it")
flag_request = request if request_has_user else ServiceRequest(user)
```

products/feature_flags/backend/facade/api.py
```diff
@@ -35,10 +41,16 @@ def _serializer_context(team: Team, user: Any, request: Any | None) -> dict:
 
     Callers without a real request (internal service paths) fall back to a minimal
     request shim carrying the acting user — FeatureFlagSerializer needs request.user.
+    ``user=None`` makes this a system write: ``ServiceRequest(None)`` carries no user,
```
haacked: `nit`: This docstring restates the gate-skip and `is_system` mechanics in full, where `create_flag` and `update_flag` just point back at the module docstring for the same thing. Worth pointing here too, so the mechanism only lives in one place.

products/feature_flags/backend/facade/api.py
haacked: `suggestion`: No current caller passes `user=None` alongside a user-bearing `request`, so nothing exercises this guard. If it regressed, the call would silently fall through to the user-bearing branch: the write gets attributed to `request.user` and the approval gate engages, the opposite of the system write the caller asked for, with no error raised.

Add a case in `test_facade.py` asserting `update_flag(flag, {...}, team=self.team, user=None, request=<a request with .user set>)` raises `ValueError`.

products/approvals/backend/decorators.py
haacked: `suggestion`: This None-user bypass has no test at the decorator level and no entry in [`test_feature_flag_bypass_matrix.py`](https://github.com/PostHog/posthog/blob/525f2fb223ceb7d2c8a019a2e14e7fc0fda9da90/products/approvals/backend/tests/test_feature_flag_bypass_matrix.py#L494), the catalog built specifically to lock down approval-gate regressions after #67329. A future change to this condition, whether narrowing it to miss a real system write or broadening it to let a human-driven request slip through, wouldn't fail anywhere.

Add a case beside [`test_gate_passes_through_when_no_policy`](https://github.com/PostHog/posthog/blob/525f2fb223ceb7d2c8a019a2e14e7fc0fda9da90/products/approvals/backend/tests/test_decorators.py#L92) in `test_decorators.py` that builds the serializer with `request.user is None` over an enabled policy and asserts `save()` doesn't raise, plus a control entry in the bypass matrix mirroring `test_enable_without_policy_applies_normally` that does a `user=None` disable and asserts 0 `ChangeRequest`s.

haacked (approved):
Nice fix! A couple suggestions inline, nothing blocking.

products/approvals/backend/decorators.py
andehen: Good call. The gate now checks getattr(request, "is_system", False) is True and the facade's ServiceRequest shim declares is_system=True when user=None. A request that merely lacks a user (like a pre-auth HttpRequest) engages the gate again, same as before this PR. The is True comparison is deliberate so attribute-forwarding objects like MagicMock can't accidentally take the bypass.

products/feature_flags/backend/facade/api.py
andehen: Done, pulled it into a request_has_user variable as suggested.

products/feature_flags/backend/facade/api.py
andehen: Trimmed, it now points at the module docstring like the write functions do.

products/feature_flags/backend/facade/api.py
andehen: Added test_system_write_rejects_user_bearing_request in test_facade.py. It calls update_flag with user=None plus a request carrying a real user, and asserts the ValueError and that the flag is untouched.

products/approvals/backend/decorators.py
andehen: Added both. test_decorators.py gets test_gate_passes_through_for_declared_system_request (system request over an enabled policy saves fine, no ChangeRequest) and test_userless_request_without_system_marker_still_engages_gate (user is None but no system marker, the write must not apply). The bypass matrix gets test_system_disable_under_policy_applies_without_change_request as the control for the one sanctioned skip.

PR #71362, opened 2026-07-15
Replace forked action with original action

Piccirello:
The upstream action now pins its dependencies, so we can resume using it.


Piccirello:
@stamphog I created the damn fork 😭

rnegron (approved):

PR #71363, opened 2026-07-15
refactor(surveys): route survey flag maintenance through the flag facade

andehen:
## Problem

The two survey flag-maintenance Celery beat tasks (`update_survey_adaptive_sampling`, `update_survey_iteration`) mutate `FeatureFlag.filters` JSON directly and save the model raw.
That bypasses the flag serializer's validation, cache updates, and activity logging, and it digs into the `filters` schema that the feature flags product owns (flagged by the `feature-flags-no-raw-filters-access` semgrep rule, #70128).

## Changes

Two new pure transforms in `products/feature_flags/backend/facade/filters.py`:

- `set_first_release_condition_rollout(current_filters, rollout_percentage)` sets `groups[0].rollout_percentage` and preserves everything else. It raises on missing/empty groups, matching the invariant the adaptive-sampling caller already relies on (survey flags are always created with a release condition).
- `replace_release_conditions(current_filters, groups)` swaps `groups` wholesale while preserving `multivariate`, `payloads`, aggregation, and any other keys.

Both beat tasks now write through the facade:

- `posthog/tasks/update_survey_adaptive_sampling.py` routes the rollout bump through `update_flag(..., user=None)`, the validated, activity-logged system-write path. System writes skip the approval gate by design, so an enabled approval policy can't 409 a cron job.
- `posthog/tasks/update_survey_iteration.py` routes the iteration-suffix targeting-flag update through `replace_release_conditions` + `update_flag(..., user=None)`, and creates missing targeting flags through `create_flag(..., user=None)` with `creation_context: "surveys"` and no usage dashboard (matching the previous raw create). The create has to be a system write — a user-bearing write would engage the approval gate, which a cron job can't surface as a 409 — but system writes null `created_by`, and access control grants a flag's creator guaranteed highest access, so losing it would silently change behavior vs the old raw create (which passed `created_by=survey.created_by`). The task therefore restores `created_by = survey.created_by` after the gated create via `QuerySet.update()`, which bypasses `save()`/signals: the write stays system-attributed (`is_system` activity log, null `last_modified_by`) and cannot re-engage the gate.

> [!NOTE]
> Flag writes now go through the serializer, which normalizes `aggregation_group_type_index` (None = person-aggregated) onto every condition set and the top-level filters, exactly like flag writes from the UI/API. The tests pin this down.

These are cron jobs, so a validation regression in the facade path would break them silently. That's what the new tests target:

- Round-trip tests (parameterized) push the filters shapes surveys actually persist (the `$survey_dismissed`/`$survey_responded`/`$survey_last_seen_date` internal-property conditions) through the new path and assert they persist unchanged apart from the intended rollout/groups change. No existing test exercised these shapes against serializer validation.
- System-write tests run both tasks with the approvals feature and an enabled `feature_flag.update` policy, asserting the tasks succeed, log activity with `is_system=True`, and never create a change request. Catches anyone re-engaging the approval gate on these paths. The update-path test also covers `survey.created_by=None`; the create-path test asserts the recreated flag's `created_by` is restored to the survey's creator, catching anyone dropping the attribution patch-up.
- Transform unit tests (parameterized, no DB) pin the two new transforms' contracts, including the raise-on-missing-groups behavior and input non-mutation.
- Existing iteration tests upgraded from subset to full-equality filter assertions, so serializer-added keys are visible rather than silently ignored.

Ran locally: `pytest posthog/tasks/test/test_update_survey_adaptive_sampling.py posthog/tasks/test/test_update_survey_iteration.py products/feature_flags/backend/test/test_facade.py` (37 passed).
Also verified the semgrep rule from #70128 reports 0 findings on both task files with their baseline excludes removed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change, no docs affected.

## 🤖 Agent context

Authored by Claude Code as part of the `feature-flags-no-raw-filters-access` baseline burndown. Skills invoked: `/writing-tests`. Verified with ruff, tach boundary check, and a local semgrep run of the rule.

Decisions:

- `set_first_release_condition_rollout` deliberately raises when no release group exists instead of synthesizing one, because the caller guarantees the group exists at creation time and a missing group would mean a broken invariant worth surfacing.
- The flag create keeps `_should_create_usage_dashboard: False` to match the previous raw `objects.create` (no dashboard), and passes `creation_context: "surveys"`, mirroring how the survey product itself creates targeting flags.

Notes for reviewers:

- Stacked on #71353 → #70124 → #70123.
- The two survey-task baseline exclude lines in the semgrep rule get removed when #70128 (the rule PR) restacks, local verification already shows 0 findings for these files.


posthog/tasks/update_survey_iteration.py
```diff
@@ -68,47 +69,55 @@ def _has_final_iteration_ended(survey: Survey) -> bool:
...
+    return create_flag(
+        {
+            "key": str(survey.id),
+            "active": True,
+            "filters": {"groups": user_submitted_dismissed_groups},
+            "creation_context": "surveys",
+            "_should_create_usage_dashboard": False,
+        },
+        team=survey.team,
+        user=None,
```
haacked, line 120: `blocking`: This is a breaking change to already-shipped behavior, not just refactor mechanics, and the PR description doesn't mention it. Before this PR, `FeatureFlag.objects.create(created_by=survey.created_by, ...)` preserved the real creator; now this forces `created_by=None` unconditionally, as [documented in the facade itself](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/products/feature_flags/backend/facade/api.py#L70-L71).

It only fires as a fallback, when `survey.internal_targeting_flag` is missing outside the normal flow: the survey API already creates the flag eagerly with the real author on every create/update ([`_add_user_survey_interacted_filters`](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/products/surveys/backend/api/survey.py#L1815)). Rare, not the primary path, but it still silently changes what happens when it fires.

Not something to revert: passing `survey.created_by` would flip `is_system` to `False` ([api.py:56](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/products/feature_flags/backend/facade/api.py#L56)) and let the approval gate 409 this cron job, which is what commit `e18500c9f06` fixed. Concrete cost of leaving it as-is: `UserAccessControl` grants a flag's creator "highest access" ([user_access_control.py:741](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/posthog/rbac/user_access_control.py#L741)), lost on beat-task-recreated flags but kept on UI-created ones.

A fix exists that doesn't reopen the gate: capture the flag this call returns and patch just the column afterward, since `QuerySet.update()` bypasses `.save()` and its signals so it can't touch `is_system`, `last_modified_by`, or the activity log:

```python
if survey.created_by is not None:
    FeatureFlag.objects.filter(pk=flag.pk).update(created_by=survey.created_by)
    flag.created_by = survey.created_by
```

Worth generalizing into an `attributed_to` kwarg on `create_flag` for reuse elsewhere, but that's a facade change outside this PR's scope. At minimum, this needs to be called out as an accepted breaking change in the PR description before merge.

posthog/tasks/test/test_update_survey_iteration.py
```diff
@@ -214,14 +167,94 @@ def test_iteration_change_updates_flag_with_new_iteration_suffix(self) -> None:
...
+            update_survey_iteration()
+
+        self.recurring_survey.refresh_from_db()
+        flag = self.recurring_survey.internal_targeting_flag
+        assert flag is not None
+        self.assertEqual(flag.filters, self._expected_iteration_filters(3))
+        self.assertFalse(ChangeRequest.objects.filter(team=self.team).exists())
+        log = ActivityLog.objects.get(scope="FeatureFlag", item_id=str(flag.id), activity="updated")
+        self.assertTrue(log.is_system)
+        self.assertIsNone(log.user)
```
haacked, line 231: `suggestion`: This checks the filters, the skipped approval gate, and the system activity log, but never `flag.created_by`. The flag [starts with `created_by=self.user`](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/posthog/tasks/test/test_update_survey_iteration.py#L47), and `update_flag` shouldn't touch it, only the create path forces it to `None`:

```suggestion
        self.assertIsNone(log.user)
        self.assertEqual(flag.created_by, self.user)
```

products/feature_flags/backend/test/test_facade.py
```diff
@@ -443,6 +446,68 @@ def test_does_not_alias_input(self):
...
+        result = replace_release_conditions(filters, new_groups)
+
+        assert result == {
+            "groups": new_groups,
+            "multivariate": {"variants": [{"key": "control", "rollout_percentage": 100}]},
+            "payloads": {"control": "{}"},
+            "aggregation_group_type_index": 1,
+        }
+        assert result["groups"] is not new_groups
+        assert filters["groups"][0]["properties"][0]["key"] == "old"  # input not mutated
```
haacked, line 512: `suggestion`: This asserts `result["groups"] is not new_groups` but not that the groups are actually deep-copied; swapping `deepcopy(groups)` for `list(groups)` would still pass, and a later mutation would corrupt the caller's `new_groups`. Mirror [`test_does_not_alias_input`](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/products/feature_flags/backend/test/test_facade.py#L435) by mutating a nested field and checking the source is untouched:

```suggestion
        result["groups"][0]["properties"][0]["key"] = "mutated"
        assert new_groups[0]["properties"][0]["key"] == "new"
        assert filters["groups"][0]["properties"][0]["key"] == "old"  # input not mutated
```

products/feature_flags/backend/facade/filters.py
```diff
@@ -125,6 +125,28 @@ def replace_variant_distribution(current_filters: dict, variants: list[dict]) ->
...
+    return new_filters
+
+
+def replace_release_conditions(current_filters: dict, groups: list[dict]) -> dict:
+    """Replace the release ``groups`` wholesale.
+
+    Every other key (``multivariate``, ``payloads``, aggregation, holdout/super groups)
+    is preserved — only who the flag releases to changes, not what it serves.
+    """
+    return {**deepcopy(current_filters), "groups": deepcopy(groups)}
```
haacked: `nit`: This deep-copies `current_filters` including `groups`, which gets thrown away on the next line. [`replace_variant_distribution`](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/products/feature_flags/backend/facade/filters.py#L116) above only deep-copies the value it's replacing:

```suggestion
    return {**current_filters, "groups": deepcopy(groups)}
```

posthog/tasks/update_survey_iteration.py
```diff
@@ -68,47 +69,55 @@ def _has_final_iteration_ended(survey: Survey) -> bool:
...
+                "filters": replace_release_conditions(
+                    existing_targeting_flag.get_filters(), user_submitted_dismissed_groups
+                )
+            },
             team=survey.team,
-            created_by=survey.created_by,
-            active=True,
-            key=str(survey.id),
-            filters=user_submitted_dismissed_filter,
+            user=None,
```
haacked, line 106: `nit`: The `create_flag` call below explains why `user=None` (skips the approval gate), but this one doesn't:

```suggestion
            # system write: skips the approval gate, same as the create_flag call below
            user=None,
```

posthog/tasks/test/test_update_survey_iteration.py
```diff
@@ -214,14 +167,94 @@ def test_iteration_change_updates_flag_with_new_iteration_suffix(self) -> None:
...
+        )
+        self.recurring_survey.internal_targeting_flag = None
+        self.recurring_survey.start_date = now() - timedelta(self.iteration_frequency_days * 3)
+        self.recurring_survey.save()
+
+        with self.captureOnCommitCallbacks(execute=True):
+            update_survey_iteration()
+
+        internal_flag = FeatureFlag.objects.get(key=self.recurring_survey.id)
+        self.assertIsNone(internal_flag.created_by)
```
haacked: `nit`: [`self.recurring_survey.created_by`](https://github.com/PostHog/posthog/blob/2f77951ca76b35800b2a2ed313131d2c0d87a1d0/posthog/tasks/test/test_update_survey_iteration.py#L37) is `self.user`, never nulled here, so this is really asserting a non-null creator gets dropped. Make that explicit:

```suggestion
        self.assertIsNotNone(self.recurring_survey.created_by)
        self.assertIsNone(internal_flag.created_by)
```

haacked (commented):
One blocking issue inline: a breaking change to creator attribution that isn't called out. A few other non-blocking suggestions too.

posthog/tasks/update_survey_iteration.py
andehen: Good catch. Applied your suggested fix in 01388812b04 and updated the PR description. The create-path test now asserts created_by gets restored to the survey's creator.

posthog/tasks/test/test_update_survey_iteration.py
andehen: Added in 8a3bb18ca43.

products/feature_flags/backend/test/test_facade.py
andehen: Applied in 8a3bb18ca43.

products/feature_flags/backend/facade/filters.py
andehen: Fixed in 8a3bb18ca43.

posthog/tasks/update_survey_iteration.py
andehen: Added in 8a3bb18ca43.

posthog/tasks/test/test_update_survey_iteration.py
andehen: Superseded by the fix for your blocking comment: created_by is now restored on this path, so the assertion became assertEqual(internal_flag.created_by, self.user) in 01388812b04.

posthog/tasks/test/test_update_survey_iteration.py
```diff
@@ -214,14 +167,96 @@ def test_iteration_change_updates_flag_with_new_iteration_suffix(self) -> None:
...
+        self.org.save()
+        ApprovalPolicy.objects.create(
+            organization=self.org,
+            team=self.team,
+            action_key="feature_flag.update",
+            conditions={},
+            approver_config={"quorum": 1, "users": [self.user.id]},
+            created_by=self.user,
+        )
+        self.recurring_survey.created_by = None
```
haacked: `suggestion`: This line doesn't affect anything the test checks. I think you can drop it.

```suggestion
```

posthog/tasks/update_survey_adaptive_sampling.py
```diff
@@ -31,8 +33,16 @@ def _update_survey_adaptive_sampling(survey: Survey) -> None:
...
-        internal_response_sampling_flag.save(update_fields=["filters"])
+        update_flag(
+            internal_response_sampling_flag,
+            {
+                "filters": set_first_release_condition_rollout(
+                    internal_response_sampling_flag.get_filters(), today_entry["rollout_percentage"]
+                )
+            },
+            team=survey.team,
+            user=None,
```
haacked, line 46: `nit`: This `user=None` is the system-write pattern from `update_survey_iteration.py`, but it needs a comment here explaining it skips the approval gate.

```suggestion
            # system write: skips the approval gate, a beat task cannot surface an ApprovalRequired change request
            user=None,
```

haacked (approved):
Thanks for fixing up the `created_by` issue!. A couple tiny non-blocking suggestions inline.

PR #71366, opened 2026-07-15
refactor(early-access-features): move enrollment filters behind the flag facade

andehen:
## Problem

Early access features manipulate their linked feature flag's enrollment filters by reading and writing `flag.filters` directly and, on some paths, saving the flag with a raw `.save()`.
That bypasses the flag product's write boundary: no serializer validation, no activity logging, and flag-schema knowledge (`feature_enrollment` marker, legacy `super_groups`) living in a consumer product.
Part of the baseline burndown for the `feature-flags-no-raw-filters-access` semgrep rule (#70128).

## Changes

New facade transform `set_feature_enrollment(current_filters, enrolled, *, groups=None)` in `products/feature_flags/backend/facade/filters.py`.
It preserves the exact semantics of the local EAF helper it replaces: spreads the filters, writes the `feature_enrollment` marker (`None` clears it without removing the key), pops legacy `super_groups`, and optionally replaces `groups`.
The flag product now owns the enrollment-marker knowledge.

All EAF flag-write sites route through the facade (`create_flag` / `update_flag`):

| Site | Before | After |
|---|---|---|
| GA + rollout-to-all stage change | direct `FeatureFlagSerializer` | `update_flag(..., user=request.user)` (gated) |
| stage → active | direct `FeatureFlagSerializer` | `update_flag(..., user=request.user)` (gated) |
| stage → inactive | raw `.save()` | `update_flag(..., user=None)` (system write) |
| link existing flag on create (active stage) | direct `FeatureFlagSerializer` | `update_flag(..., user=request.user)` (gated) |
| new flag on create | direct `FeatureFlagSerializer` | `create_flag(..., user=request.user)` (gated) |
| EAF destroy | raw `.save()` | `update_flag(..., user=None)` (system write) |
| file-system pre-delete hook (`apps.py`) | raw `.save()` | `update_flag(..., user=None)` (system write) |

Raw `.filters` reads at those sites now go through `get_filters()`, and the local `_set_enrollment_filters` helper is deleted.
The `apps.py` hook keeps its lazy `facade.api` import (heavy import chain off startup) while the import-light `facade.filters` transform is imported at module level.

> [!NOTE]
> The destroy, deactivate, and pre-delete-hook writes are deliberate system writes (`user=None`): a cleanup on destroy or stage demotion must never surface a 409 approval requirement, preserving today's ungated behavior while gaining serializer validation and activity logging.
> The tradeoff is that these three writes' flag activity is now attributed as system instead of the acting user (previously they logged no flag activity at all).

Ran locally: `pytest products/early_access_features/backend products/feature_flags/backend/test/test_facade.py` (94 passed, including all pre-existing EAF tests unchanged).

New tests and the regression each catches:

- `test_stage_transitions_round_trip_legacy_flag_filters` (parameterized, EAF suite): a legacy-shaped flag (`super_groups`, payloads, custom release conditions) driven through stage transitions round-trips with only the enrollment keys changed. Catches the newly-serializer-routed writes rejecting or mangling historically-unvalidated filter shapes, which no existing test covered since they only used freshly-created flags.
- `test_destroy_and_demote_never_require_approval` (parameterized, facade suite, placed there because `early_access_features` has no tach dependency on `approvals`): with approvals enabled and an enabled `feature_flag.update` policy, EAF destroy and stage demotion still succeed with enrollment cleared and no change request. Catches these cleanups being flipped back to gated user writes.
- `test_file_system_deletion_clears_enrollment_and_logs_system_activity`: the file-system pre-delete hook clears enrollment and logs flag activity as system. Catches the hook's lazy facade import breaking (which would only surface at deletion time) or reverting to a raw save.
- `TestSetFeatureEnrollment` (parameterized transform unit tests, next to the other facade transform tests): catches the marker being removed instead of written as `None`, `super_groups` surviving, or the groups override being dropped.

Semgrep: ran the `feature-flags-no-raw-filters-access` rule with the two EAF baseline excludes removed against `api.py` and `apps.py` – 0 findings.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior, API, or config changes – internal write-path refactor only.

## 🤖 Agent context

Authored by Claude Code as part of the flag-facade boundary work.
Skills invoked: `/improving-drf-endpoints` and `/writing-tests`.
Decisions: the destroy/deactivate/hook writes were made system writes rather than user-gated ones so cleanup can never 409 (see the note above); the approvals-interaction test lives in the flags facade suite instead of the EAF suite because tach forbids `early_access_features` importing `approvals`, and the test only drives HTTP endpoints anyway.

Notes for reviewers:

- Stacked on #71353 → #70124 → #70123 as part of the `feature-flags-no-raw-filters-access` baseline burndown; base branch is `feat/flags-facade-system-writes`, not master.
- The two EAF baseline exclude lines in the semgrep rule get removed at restack once #70128 (the rule) merges – local verification against the rule already shows 0 findings for these files.


products/early_access_features/backend/api.py
```diff
@@ -388,8 +392,14 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
                 resource_scope="early_access_feature:write",
             )
             assert_feature_flag_rbac_access(self.user_access_control, feature_flag=related_feature_flag)
-            related_feature_flag.filters = _set_enrollment_filters(related_feature_flag.filters, enrolled=None)
-            related_feature_flag.save()
+            # System write: clearing enrollment when the feature is destroyed is cleanup, so an
+            # enabled approval policy must never block it with a 409. Activity is logged as system.
+            update_flag(
```
haacked: `blocking`: This turns "destroy always succeeds" into "destroy can fail." The old cleanup write bypassed validation entirely with a raw model save, so it always went through no matter what was stored in the flag. Routing it through `update_flag` now validates the flag's entire filters, because `set_feature_enrollment` copies the whole existing filters (not just the enrollment key) into every write, which defeats the [partial-update shortcut](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/feature_flags/backend/api/feature_flag.py#L1241) that would otherwise skip validation.

That has two consequences. A flag using group aggregation while linked to an EAF now hits a [hard rejection rule](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/feature_flags/backend/api/feature_flag.py#L1367) and destroy returns a 400 where it used to succeed. Worse, a flag with a legacy property missing its `key` field [crashes with a raw `TypeError`](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/feature_flags/backend/api/feature_flag.py#L1380) instead of a clean validation error, and the [file-system pre-delete hook](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/early_access_features/backend/apps.py#L45) has no try/except around this call, so that crash surfaces as an uncaught exception instead of a delete.

This isn't hypothetical. A flag can get linked to an EAF via `feature_flag_id` without ever passing through a validating write, and the PR's own test suite builds a flag in exactly that shape (`_create_feature_with_legacy_flag`). The same validation now runs on the [demote path](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/early_access_features/backend/api.py#L214) and the pre-delete hook too.

Fall back to the raw write when validation rejects, so cleanup keeps its "never fails" guarantee for legacy flags while still validating well-formed ones:

```python
try:
    update_flag(
        related_feature_flag,
        {"filters": set_feature_enrollment(related_feature_flag.get_filters(), None)},
        team=self.team,
        user=None,
    )
except (ValidationError, TypeError):
    related_feature_flag.filters = set_feature_enrollment(dict(related_feature_flag.filters or {}), None)
    related_feature_flag.save(update_fields=["filters"])
```

Same guard needed at `apps.py:45` and `api.py:214`.

products/feature_flags/backend/facade/filters.py
```diff
@@ -138,6 +138,23 @@ def set_holdout(current_filters: dict, *, holdout_id: int | None, exclusion_perc
...
+    matching what ``FeatureFlag.has_feature_enrollment`` treats as not enrolled.
+    Legacy ``super_groups`` (the pre-marker enrollment encoding) are dropped by the
+    same write either way. ``groups`` optionally replaces the release conditions —
+    early access uses it to roll a generally-available feature out to everyone.
+    Everything else is spread through untouched.
+    """
+    new_filters = {**current_filters, "feature_enrollment": enrolled}
+    new_filters.pop("super_groups", None)
+    if groups is not None:
+        new_filters["groups"] = groups
```
haacked: `suggestion`: `set_feature_enrollment` stores the `groups` list by reference instead of copying it. `replace_variant_distribution` deepcopies `variants` specifically to prevent aliasing; `test_does_not_alias_input` guards against regression. Every caller here passes a fresh literal today, so nothing breaks yet, but a caller that reused a list would find the flag's stored filters mutating under it.

```suggestion
        new_filters["groups"] = deepcopy(groups)
```

products/feature_flags/backend/facade/api.py
```diff
@@ -53,7 +58,10 @@ def _serializer_context(team: Team, user: Any, request: Any | None) -> dict:
     # goes to request.user) — reject the contradiction instead.
     if user is None and request_has_user:
         raise ValueError("user=None is a system write; do not pass a user-bearing request with it")
-    flag_request = request if request_has_user else ServiceRequest(user, is_system=user is None)
+    if isinstance(request, ServiceRequest):
+        # Caller-built shims default to POST — rebuild so the method matches this write.
+        request = ServiceRequest(request.user, is_system=request.is_system, method=method)
```
haacked: `suggestion`: This rebuilds a new `ServiceRequest` just to correct its `method`, discarding any other state the caller set (`path`, `data`, `GET`, `headers`, `session`; none are set today, which is exactly what makes the loss silent). Since the type is already narrowed to `ServiceRequest`, mutating the field in place is simpler and keeps the request intact.

```suggestion
        request.method = method
```

products/early_access_features/backend/api.py
```diff
@@ -347,19 +352,18 @@ def create(self, validated_data):
             if validated_data.get("stage") in EarlyAccessFeature.ActiveStage:
                 filters["feature_enrollment"] = True
```
haacked: `suggestion`: This branch still writes the raw `feature_enrollment` marker directly (`filters["feature_enrollment"] = True`), which is exactly the kind of flag-schema knowledge this PR moves into the facade. This is the only raw-filters-access the semgrep rule was meant to catch that still passes.

```suggestion
                filters = set_feature_enrollment(filters, True)
```

products/feature_flags/backend/facade/api.py
```diff
@@ -74,16 +82,47 @@ def create_flag(data: dict, *, team: Team, user: Any, request: Any | None = None
...
     (fields it omits are untouched) applied as-is — nothing is silently dropped.
     Raises ApprovalRequired when a policy requires approval; the flag is left untouched.
     ``user=None`` is a system write (see module docstring): ``last_modified_by`` is
-    cleared, activity is logged as system, the approval gate is skipped."""
-    serializer = FeatureFlagSerializer(flag, data=data, partial=True, context=_serializer_context(team, user, request))
+    cleared, activity is logged as system, the approval gate is skipped.
+
+    Encrypted payload values carried over unchanged from ``flag.get_filters()`` are
+    preserved as-is, never re-validated or re-encrypted."""
+    data = _redact_unchanged_encrypted_payloads(flag, data)
```
haacked, line 114: `suggestion`: `_redact_unchanged_encrypted_payloads` turns stored ciphertext into `REDACTED_PAYLOAD_VALUE` before it reaches the serializer, so [`FeatureFlagSerializer.update()`'s existing "preserve on placeholder" branch](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/feature_flags/backend/api/feature_flag.py#L1952) fires instead of re-encrypting it. The [`refresh_from_db`](https://github.com/PostHog/posthog/blob/aaf91b7526983633ccda83cd8c28a7824fb08510/products/feature_flags/backend/facade/api.py#L124) a few lines down exists because that same branch mutates the in-memory instance's payloads to their response form, so a caller reusing the returned instance for a second write would otherwise persist the placeholder over the real ciphertext.

Both mechanisms work around the same gap: the serializer only recognizes its own placeholder, not a byte-identical ciphertext echo. Widening that check in the serializer (`v == existing_value or v == REDACTED_PAYLOAD_VALUE`) would let facade callers pass `get_filters()` straight through and could remove the need for both the helper and the reload. That touches the serializer every HTTP caller goes through, so it's a trade-off worth weighing.

Separately, the reload adds a SELECT on every encrypted-payload `update_flag` call across all ~13 facade callers, not just the ones that reuse the instance; if that shows up in write-latency traces, an opt-out param (`reload_after: bool = True`) would scope it.

products/warehouse_sources/backend/temporal/data_imports/workflow_activities/tests/test_import_data_sync.py
```diff
@@ -218,7 +218,7 @@ async def test_schema_column_type_changed_routes_through_handler_without_source_
 
     handle_mock.assert_awaited_once()
     assert handle_mock.await_args is not None
-    assert handle_mock.await_args.args[3] is error
+    assert handle_mock.await_args.args[5] is error
```
haacked, line 223: `suggestion`: This only asserts `args[5] is error`, confirming the argument count is correct but not the argument order. A regression that swapped `team_id`/`source_id`/`run_id` would still pass. Catching that needs the anonymous `mock.MagicMock()` at the call site (line 217) bound to a variable so the assertions have something to compare against:

```python
job_inputs = mock.MagicMock()
with pytest.raises(NonRetryableException):
    await module._handle_import_error(job_inputs, logger, error)

handle_mock.assert_awaited_once()
args = handle_mock.await_args.args
assert args[0] is job_inputs.team_id
assert args[1] == str(job_inputs.source_id)
assert args[2] is job_inputs.run_id
assert args[5] is error
```

products/feature_flags/backend/facade/api.py
```diff
@@ -74,16 +82,47 @@ def create_flag(data: dict, *, team: Team, user: Any, request: Any | None = None
...
+    byte-identical to the stored ones are swapped for the redacted placeholder, which the
+    serializer preserves. Genuinely new payload values pass through and are encrypted."""
+    filters = data.get("filters")
+    if not flag.has_encrypted_payloads or not isinstance(filters, dict):
+        return data
+    stored = (flag.filters or {}).get("payloads") or {}
+    incoming = filters.get("payloads")
+    if not isinstance(incoming, dict):
+        return data
+    redacted = {k: REDACTED_PAYLOAD_VALUE if k in stored and v == stored[k] else v for k, v in incoming.items()}
```
haacked, line 98: `suggestion`: The only test for this (`test_update_preserves_encrypted_payloads_carried_from_stored_filters`) covers the all-unchanged case. There's no test for a payload value that actually changed, or a brand-new key absent from `stored`; if the comparison were inverted, a real payload update would be silently redacted instead of encrypted.

```python
def test_redact_unchanged_encrypted_payloads_only_redacts_byte_identical(self):
    flag = FeatureFlag.objects.create(
        team=self.team, created_by=self.user, key="redact-flag",
        has_encrypted_payloads=True,
        filters={"payloads": {"same": "CIPHER", "changed": "OLD"}},
    )
    out = _redact_unchanged_encrypted_payloads(
        flag, {"filters": {"payloads": {"same": "CIPHER", "changed": "NEW", "fresh": "PLAIN"}}}
    )
    payloads = out["filters"]["payloads"]
    assert payloads["same"] == REDACTED_PAYLOAD_VALUE
    assert payloads["changed"] == "NEW"
    assert payloads["fresh"] == "PLAIN"
```

products/feature_flags/backend/facade/api.py
```diff
@@ -74,16 +82,47 @@ def create_flag(data: dict, *, team: Team, user: Any, request: Any | None = None
...
     ``user=None`` is a system write (see module docstring): ``last_modified_by`` is
-    cleared, activity is logged as system, the approval gate is skipped."""
-    serializer = FeatureFlagSerializer(flag, data=data, partial=True, context=_serializer_context(team, user, request))
+    cleared, activity is logged as system, the approval gate is skipped.
+
+    Encrypted payload values carried over unchanged from ``flag.get_filters()`` are
+    preserved as-is, never re-validated or re-encrypted."""
+    data = _redact_unchanged_encrypted_payloads(flag, data)
+    serializer = FeatureFlagSerializer(
+        flag, data=data, partial=True, context=_serializer_context(team, user, request, method="PATCH")
```
haacked, line 116: `question`: The `method="PATCH"` fix changes what `_extract_resource_id` and the logged `http_method` see for approval change-requests generated from facade updates (previously always `"POST"`). Does anything verify that metadata? The current tests cover the serializer's create-only-validation effect and the system-write approval skip, but not a gated update that enforces a policy and checks the resulting `ChangeRequest`.

products/early_access_features/backend/test/test_early_access_feature.py
```diff
@@ -1788,3 +1792,144 @@ def test_migrate_command_skips_already_responded_emails(self, mock_query, mock_c
...
+
+        flag.refresh_from_db()
+        assert flag.filters["feature_enrollment"] == expected_enrollment
+        assert "super_groups" not in flag.filters
+        assert flag.filters["payloads"] == self.LEGACY_FLAG_FILTERS["payloads"]
+        groups = [
+            {key: value for key, value in group.items() if key != "aggregation_group_type_index"}
+            for group in flag.filters["groups"]
+        ]
+        assert groups == expected_groups
```
haacked, line 2109: `nit`: This test uses a non-encrypted flag and checks the DB via `refresh_from_db`, but never verifies the API response. Since it's the one round-trip test on a non-encrypted flag, it's the natural place to confirm the new redaction override doesn't fire when there's nothing to redact.

```python
assert response.json()["feature_flag"]["filters"]["payloads"] == self.LEGACY_FLAG_FILTERS["payloads"]
```

haacked (commented):
Solid refactor, thanks for cleaning this up. One blocking issue inline on the cleanup paths, plus a handful of suggestions.

products/early_access_features/backend/api.py
andehen: Good catch, fixed in 25aa98ba708.

I pulled the guard into one helper (clear_feature_enrollment) used at all three sites, so the cleanup contract lives in one place. apps.py defer-imports it from api.py like it did with _set_enrollment_filters before.

Two deviations from your snippet:

The helper catches Exception instead of (ValidationError, TypeError). The validator can throw other types on stored legacy JSON, e.g. KeyError from a variant dict missing "key", so enumerating types is whack-a-mole. The fallback is the exact old write, and it logs a warning with the flag id so we can see how often legacy shapes actually hit it.

There's also one failure shape a try/except can't catch: if the stored filters have no "groups" key at all, the serializer's partial-PATCH shortcut returns the stored filters and the cleanup silently no-ops, no exception raised. The helper checks for that and goes straight to the raw write.

Added a parameterized destroy test covering all three shapes: keyless property, group-aggregated condition, no release groups.

products/feature_flags/backend/facade/api.py
andehen: Good spot that both mechanisms work around the same gap, but widening the update() check alone can't replace them. A byte-identical ciphertext echo is rejected before update() ever runs, in validate_filters (json.loads on every payload value, a raw Fernet token isn't valid JSON; the placeholder only survives validation because it's a quoted JSON string). So the serializer-side version would need the instance-aware comparison in both validate_filters and update(), on the path every HTTP caller goes through, versus one helper on the only path that needs it. The reload is also needed either way, the in-memory instance holds the response form after save regardless of which branch preserved the ciphertext.

On the SELECT: it only fires for has_encrypted_payloads flags (remote-config only), after a write that already does validation + activity logging, so I'd hold off on a reload_after param until it shows up in traces. Added a test pinning the changed/fresh-key redaction behavior in the meantime.

products/feature_flags/backend/facade/filters.py
andehen: Done, groups is now deepcopied.

products/feature_flags/backend/facade/api.py
andehen: Done, now just mutating the method in place.

products/early_access_features/backend/api.py
andehen: Done, this now goes through set_feature_enrollment as well.

products/warehouse_sources/backend/temporal/data_imports/workflow_activities/tests/test_import_data_sync.py
andehen: Done, the test now pins the argument order.

products/feature_flags/backend/facade/api.py
andehen: Added a test covering changed and new payload keys.

products/feature_flags/backend/facade/api.py
andehen: Nothing verified it before, added a test asserting the change request stores PATCH and the flag id as resource_id.

products/early_access_features/backend/test/test_early_access_feature.py
andehen: Done, added the response assertion.

products/early_access_features/backend/api.py
```diff
@@ -63,6 +59,35 @@ def assert_feature_flag_rbac_access(
...
+            return
+        except Exception as exc:
+            # Stored legacy JSON can raise arbitrary exception types through the flag
+            # validator (ValidationError, TypeError, KeyError, ...), so catch broadly.
+            logger.warning(
+                "early_access_feature_enrollment_cleanup_fell_back_to_raw_write",
+                feature_flag_id=feature_flag.id,
+                error=str(exc),
+            )
+    feature_flag.filters = cleared_filters
```
haacked: `suggestion`: `clear_feature_enrollment` has a raw `feature_flag.filters = cleared_filters` write that breaks the `feature-flags-no-raw-filters-access` semgrep rule. The rule currently doesn't fire because `products/early_access_features/backend/api.py` is [still in the rule's exclude list](https://github.com/PostHog/posthog/blob/88f983fe886fc26a0ee2b685bc55a3aad96a23e2/.semgrep/rules/devex/feature-flags-no-raw-filters-access.yaml#L38), but removing that exclusion is the whole point of the burndown. When the exclusion is removed, this line will fire, and the PR description's "excludes removed → 0 findings" check (which predates this fallback) will no longer hold.

Add a nosemgrep comment explaining this is a deliberate never-fail cleanup fallback when the gated write can't validate stored legacy filter shapes:

```suggestion
    feature_flag.filters = cleared_filters  # nosemgrep: feature-flags-no-raw-filters-access -- deliberate never-fail cleanup fallback when the gated write can't validate stored legacy filter shapes
```

products/feature_flags/backend/test/test_facade.py
```diff
@@ -443,6 +545,96 @@ def test_does_not_alias_input(self):
...
+                    "super_groups": [],
+                },
+                None,
+                [{"properties": [], "rollout_percentage": 100}],
+                {"groups": [{"properties": [], "rollout_percentage": 100}], "feature_enrollment": None},
+            ),
+        ]
+    )
+    def test_transform(self, _name, current_filters, enrolled, groups, expected):
+        assert set_feature_enrollment(current_filters, enrolled, groups=groups) == expected
```
haacked, line 638: `nit`: `set_feature_enrollment` now `deepcopy`s the `groups` override ([`filters.py:154`](https://github.com/PostHog/posthog/blob/88f983fe886fc26a0ee2b685bc55a3aad96a23e2/products/feature_flags/backend/facade/filters.py#L154)) so the returned filters can't alias the caller's list, but `TestSetFeatureEnrollment.test_transform` only compares by equality and doesn't test that guard. Mirror [`test_does_not_alias_input`](https://github.com/PostHog/posthog/blob/88f983fe886fc26a0ee2b685bc55a3aad96a23e2/products/feature_flags/backend/test/test_facade.py#L534) in this file (which covers `_roll_out_variant`): pass a `groups` list, then assert mutating a nested dict in the returned filters doesn't touch the input.

haacked (approved):
All the previous feedback is addressed, nice work! One small suggestion and a nit inline.

products/early_access_features/backend/api.py
andehen: Done, added the nosemgrep annotation noting the deliberate never-fail cleanup fallback.

products/feature_flags/backend/test/test_facade.py
andehen: Done, added test_does_not_alias_groups_input mirroring test_does_not_alias_input.

PR #42884, opened 2025-12-05
chore(flags): removing some legacy features (so we can delete the corresponding flags)

dmarticus:
## Problem

Sometimes if we love something, we have to let it go.

(these features are both tied to flags have been disabled for years, I think if we ever planned to revive them, we should approach them anew rather than try and use this old code).

## Changes

removing `ff-dashboard-templates` and `auto-rollback-feature-flags` (two feature flag features that never saw the light of day).

PR #12485, opened 2022-10-27
feat(flags): Auto rollback beta

EDsCODE:
## Problem

Hackathon from team experiments october offsite. Key motivation here is based on long term vision discussion we want feature flags to have more automated functionality. As a starter for our hackathon, we've decided to create an auto rollback trigger based on sentry metrics or insight metrics. This is a beta rollout for internal testing

Relevant links:
Long term vision: https://github.com/PostHog/posthog.com/pull/4516
Related RFC for extension: https://github.com/PostHog/product-internal/issues/369

## Changes
- add ability to auto rollback feature flags based on sentry errors or insight changes
- insight metric is based on 7 day rolling average
- sentry metric is based on a simple threshold of error count
- sentry metric condition is only usable if a sentry api key is provided in the instance settings of the organization

Screen Shot 2022-11-01 at 1 46 16 PM

👉 *Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review.*

PR #71381, opened 2026-07-15
chore(flags): remove legacy auto-rollback backend

haacked:
## Problem

The feature flag auto-rollback beta shipped in 2022 (#12485) and its UI was removed as legacy in #42884, but the backend stayed behind. An hourly Celery task kept evaluating `rollback_conditions` for every flag that still had them, and a single malformed condition failed the entire run every hour.

## Changes

- Removes the `check_flags_to_rollback` Celery task, its hourly schedule, and `ee/tasks/auto_rollback_feature_flag.py`
- Removes the `sentry_stats` endpoint, which only existed for this feature, plus the frontend leftovers that referenced it (the MSW mock and the `SENTRY_AUTH_TOKEN`/`SENTRY_ORGANIZATION` instance settings entries)
- Removes `rollback_conditions` and `performed_rollback` from the `FeatureFlag` model, serializers, and generated frontend/MCP types

> [!NOTE]
> The flags API no longer accepts or returns `rollback_conditions` and `performed_rollback`.

The migration removes the fields from Django state only (`SeparateDatabaseAndState`); the Postgres columns stay in place so code from the previous deploy can still read them, and can be dropped in a follow-up migration after a full deploy cycle.

Removal only, no new behavior. Deleted the tests covering the removed task and API fields, and retargeted the two version history tests that used `performed_rollback` to `ensure_experience_continuity` so their coverage of nullable field diffs survives. Regenerated snapshots and types cover the rest.

## Automatic notifications

- [ ] Publish to changelog?


haacked:
Skipping the compatibility stub here. `check_flags_to_rollback` was already failing on every run in prod (a malformed `rollback_conditions` value broke the whole task), so losing its last invocation or two during the rollout changes nothing functionally.

When a worker receives a message for an unregistered task, Celery logs it and rejects the message without requeueing — no retry loop, no worker crash. Worst case here is a single harmless log line during this specific deploy. Given that, keeping a no-op stub around just to swallow that message isn't worth the extra code.

matheus-vb (approved):

PR #68636, opened 2026-07-06
feat(self-driving): Show PR CI checks and comments on reports

Twixes:
## Problem

So users can now view PostHog pull requests in PostHog, including the devs, but that's about it. You can't merge, and you can't even tell the status of the work. Is it any good to be merged? Who knows until you're going to GitHub - and that is friction.

The desktop PostHog Code app recently gained a native PR-review surface (files, comments, CI checks) so a PR can be reviewed without leaving the app ([reference PR](https://github.com/PostHog/code/pull/3174)). The cloud inbox only had the files-changed diff and an "Open in GitHub" link, forcing reviewers to switch to GitHub to see CI status or PR discussion. This brings the same checks and comments review surface to the cloud inbox.

## Changes

Let's allow folks to actually see the CI status of the work. This means we fetch live data from GitHub on this.

We also show general comments. Not inline, just that's more involved, but it adds trust to show e.g. bot comments.

Backend:

- Two new methods on`GitHubIntegrationBase`: `get_pull_request_checks` and `get_pull_request_comments`.
- Two `@action`s on the signal report viewset: `pr_checks` and `pr_comments`. They resolve the report's implementation PR URL and the GitHub integration that can read the repository, using the same connection-boundary scoping as the existing `diff` action, with typed response serializers.

**Checks and comments together in the report detail** (conversation comment + inline review comment on a file, alongside the CI checks)

inbox-pr-checks-comments-storybook

**Mixed CI states are expanded and prioritized**

inbox-pr-checks-mixed-storybook

**Fully successful checks start collapsed**

inbox-pr-checks-passing-storybook

## How did you test this code?

_Had Claude do a bunch here:_

Automated tests added:

- `posthog/models/test/test_integration_model.py`: guards the normalized checks and comments responses, including chronological merging, comment-path handling, partial upstream failures, status mapping, and unavailable PRs.
- `products/signals/backend/test/test_signal_report_api.py::TestSignalReportPrEndpoints`: guards endpoint wiring, PR URL parsing, connection-boundary scoping, and upstream error mapping.

Manual visual verification in local dev with seeded historical reports and deterministic GitHub-bound mocks:

- Failing, cancelled, running, successful, and neutral CI states sort and render with the appropriate indicators.
- A fully successful PR starts with CI collapsed; a failing or running PR starts expanded.
- A manual expansion of successful checks survives the 15-second poll.
- Conversation and inline review comments render together, including a review-file path.
- The files-changed tab and its unified diff still render.

[...]

PR #71392, opened 2026-07-16
feat(signals): address Signals PR feedback by @-mentioning the bot on GitHub

Twixes:
## Problem

Signals opens autonomous PRs. When someone reviews one and wants changes, the loop is broken: they leave a comment and nothing happens, so they have to leave GitHub and go back to PostHog to get the agent to act. This lets a reviewer stay in GitHub, drop notes on the PR, @-mention the PostHog bot, and have the agent address the feedback and push commits to the same PR.

The routing worry that usually comes with "one repo, many projects" does not apply here: a Signals PR is already tied to exactly one project through the task that opened it, so we inherit the project, repo, and report from that task instead of guessing from the repo.

## Changes

New front door, same engine. A GitHub `issue_comment` mention funnels into the same `create_and_run_task` path Slack and Signals autostart already use.

```mermaid
flowchart TD
    GH[GitHub issue_comment]:::phRed --> WH{{github_webhook}}:::phYellow
    WH -->|bot mention on a Signals PR| C[Celery: process_github_mention]:::phBlue
    C --> ID[resolve commenter identity]:::phBlue
    ID -->|eligible| RUN{{agent run: push to PR head}}:::phBlue
    ID -->|not connected| GATE[connect link + pending row]:::phGray
    RUN --> RELAY[terminal comment on PR]:::phRed
    GATE -.->|user connects GitHub| REPLAY[replay last 12h]:::phBlue
    REPLAY --> C
```

- **Identity is the security boundary.** The commenter is resolved on their immutable GitHub account id, never the login string (which is renamable and spoofable). To run they must be an org member with a usable personal GitHub connection that covers the repo. Commits are then authored as them (`GITHUB_MENTION` origin resolves to USER authorship).
- **Connect-gate, no dead end.** An un-connected commenter gets a one-click connect link and a `GitHubPendingMention` row keyed by account id. When they connect (or re-scope) GitHub, a replay job re-runs their still-pending mentions from the last 12 hours through the normal path. The account-id key makes the replay binding verified: the OAuth flow proves they control that account.
- **Feedback scope + injection.** The agent addresses the PR description plus org-member conversation comments only. Non-member comments (possible on public repos) are passed as context, never instructions, and the prompt frames all of it as data.
- **Result relay.** On a terminal run the tasks completion path posts one outcome comment back to the PR. Outcome only, no project data, since PR comments can be public.
- **Guards.** Per-PR mentions queue behind an in-flight run rather than spawning a parallel one. There is a global kill switch (`SIGNALS_GITHUB_MENTIONS_ENABLED`). Follow-up runs are recorded against the report as billing-exempt so an existing report's PR is not re-billed.

> [!NOTE]
> Two deliberate follow-ups. First, pushing onto the existing PR head without opening a new PR needs a flag in the external `@posthog/agent` harness; in-repo we already pass `branch=<head>` + `create_pr=False` + USER authorship and check out the head, so the backend side is ready. Second, the kill switch is global for now, not per-org. Both are called out in the plan.

I (Claude) drove this end to end but could not exercise the live GitHub webhook, the sandbox, or a real agent run. Verification is automated tests plus `manage.py check`, `tach check`, `ruff`, and `hogli ci:preflight`.

Automated tests added, and the regression each group catches:

- `test_github_mention_models` - fail-closed team scoping isolates pending mentions across teams, and the 12h/pending replay lookup filter. Catches a scoping regression that would leak another team's rows.
- `test_identity` - the eligibility matrix. Catches matching on login instead of account id (spoofing), a non-member being ruled eligible, and a repo-uncovered connection being treated as push-capable.
- `test_webhook` - only genuine bot mentions on Signals PRs enqueue: no enqueue for edits, plain issues, the bot's own comments, non-mentions, non-Signals PRs, or duplicate deliveries.
- `test_process` - the launch contract (`GITHUB_MENTION`, `create_pr=False`, `branch=head`, `interaction_origin=github`, billing-exempt report record) and the gate/reject routing.
- `test_replay` - the 12h window and idempotency: in-window pending replays and is marked processed, expired is skipped, already-processed and other accounts are untouched.
- `test_result_relay` - the terminal comment posts to the mapped PR for completed/failed and posts nothing without a mapping.
- `test_sandbox_credentials` (extended) - `GITHUB_MENTION` resolves to USER authorship while `SIGNAL_REPORT` stays BOT. Catches attribution regressions.

Existing `test_sandbox_credentials` (27) and `test_user_integration` (72) still pass, confirming the authorship and connect-callback edits are backward compatible.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Fully authored by Claude (Claude Code, Opus 4.8) under Michael's direction. We ran the compound-engineering flow first: `ce:brainstorm` to settle scope and the identity model, `document-review` (five reviewer personas) to pressure-test it, then `ce:plan`. Repo skills invoked while building: `/django-migrations` (the two new models plus the origin enum, with both team FKs `db_constraint=False` to keep the hot `posthog_team` table lock-free) and `/writing-tests` (the value gate on every test above).

Key decisions along the way. The brainstorm's naive "require a linked identity" idea both broke first-use and, per the review, would still produce bot-authored commits because most commenters lack a push token. That reframed into the connect-gate plus 12h replay, which keeps real attribution and removes the dead end. The review also caught two things the plan had treated as free reuse: the shared webhook dispatcher terminates on the Conversations handler (so this is wired as a side effect alongside it, not a passive add), and "run as the commenter" needs a push-capable personal connection, not just a login match. A codebase spike confirmed `GITHUB_MENTION` falls through to BOT authorship by default, hence the two small enum-set edits in tasks. The push-to-existing-branch capability turned out to need an external agent-harness flag, which is why that piece is scoped as a follow-up rather than faked here.

Planning docs are in the branch under `docs/brainstorms/` and `docs/plans/`.


Twixes:
* **#72447** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#71392** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71392?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71392?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#72242** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72242?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#71563](https://github.com/PostHog/posthog/pull/71563) <a href="https://app.graphite.com/github/pr/PostHog/posthog/71563?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#68636** <a href="https://app.graphite.com/github/pr/PostHog/posthog/68636?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

Twixes:
Closing as stale: I'm leaving PostHog and this never picked up a reviewer. Happy for anyone to reopen or re-raise if it's still useful.

PR #71394, opened 2026-07-16
feat(tasks): expose resume-chain presigned log urls on run detail

charlesvien:
## Problem

Clients bootstrap a task run's session history through the paginated `session_logs/` endpoint. Every page re-downloads the run's entire log chain from S3 and re-parses every JSONL line just to slice out ~2 MB, so one cold load of a run with N bytes of history costs O(N²). Once a single page crosses the client's fetch timeout, the run becomes unloadable: retry restarts at offset 0 and dies the same way.

The run detail response already exposes a presigned `log_url`, but only for the run's own log, not its resume ancestors, so clients can't fetch the full history from object storage directly.

## Changes

- Add `log_urls` to the run detail response: presigned S3 URLs for every JSONL log in the run's resume chain, oldest first, cached the same way as `log_url`.
- Returns an empty list when presigning is unavailable so clients can fall back to `session_logs/`.
- Regenerated OpenAPI types.

The PostHog Code client prefers downloading history straight from these URLs and keeps the paginated endpoint as fallback (PostHog/code#3489), which also preserves compatibility for old clients.

Added API tests to `TestTaskRunAPI`: a chainless run returns exactly its own presigned log URL, a resumed run returns the chain oldest first, and presigning being unavailable yields an empty list instead of a 500. All three pass locally via `hogli test`. `ruff` and `hogli ci:preflight --fix` are clean. I did not test against a live deployment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A

## 🤖 Agent context

Authored by Claude Code (directed by @charlesvien) while debugging why very long-running cloud task runs fail to load their history in PostHog Code. The investigation traced the failure to `session_logs` doing a full chain read and full JSONL parse per page. We chose presigned chain URLs over byte-cursor ranged pagination in the endpoint because the presigned `log_url` pattern already exists on run detail and this removes Django from the history byte path entirely. A follow-up should address the `append_log` full read-modify-write on the write path. Skills used: `/improving-drf-endpoints` and `/writing-tests` (read as repo skill files).

### DIFF
--- a/products/tasks/backend/facade/api.py
+++ b/products/tasks/backend/facade/api.py
@@ -366,12 +366,14 @@ def _task_run_chain_log_urls(run: TaskRun) -> list[str]:
     return urls
 
 
-def _task_run_detail_to_dto(run: TaskRun) -> contracts.TaskRunDetailDTO:
+def _task_run_detail_to_dto(run: TaskRun, *, include_log_urls: bool = False) -> contracts.TaskRunDetailDTO:
     """Map a ``TaskRun`` to its HTTP detail DTO.
 
     Reproduces the SMF-derived fields ``TaskRunDetailSerializer`` computed: ``log_url`` does
     presigned-URL I/O (with caching), and ``runtime_adapter`` / ``provider`` / ``model`` /
-    ``reasoning_effort`` are parsed off the run ``state``.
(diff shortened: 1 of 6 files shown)


charlesvien:
@greptileai

PR #71400, opened 2026-07-16
fix(experiments): view-recordings filter fails for server-side flag exposure

concernedsebster:
## Problem

Clicking "View recordings" on an experiment variant returns zero recordings when the experiment's feature flag is evaluated server-side: backend SDKs (posthog-python, posthog-node, ...) emit the default `$feature_flag_called` exposure event without a `$session_id`, so the auto-populated event filter can only ever match zero sessions.

#71639 (merged) made these surfaces explain themselves instead of landing on an empty list, and the "Recordings" tab from #71919 renders a warning banner in that case - but there is still nothing to watch, even though recordings of the experiment audience exist and are reachable through the `$feature/<flag_key>` property that `posthog-js` stamps on client-side events after flags load. Surfaced by a customer support ticket, where the equivalent property filter (added manually) was verified to return recordings.

## Changes

Per the [review discussion](https://github.com/PostHog/posthog/pull/71400#issuecomment-5056266512): keep the exposure event as the primary filter (it is what the analysis counts), and use `$feature/<flag_key>` only as a fallback for the default-exposure branch when the linkability check says the exposure event can't match any session - labelled as "flag was active in this session" rather than "exposed", since the property reflects the flag's value per event, not the enrollment moment.

- `getExposureFallbackFilter` builds the fallback property filter for the default-exposure branch. Custom exposure criteria get no fallback (a flag-value filter can't stand in for their semantics) and keep the blank-with-explanation behavior from #71639.
- `applySessionLinkability` accepts the fallback and reports `usedExposureFallback`, so surfaces can adjust their labelling.
- The per-variant "View recordings" buttons now use the fallback instead of disabling, with a tooltip explaining the approximation.
- The experiment "Recordings" tab now uses the fallback instead of the warning banner, with an info banner explaining the approximation.

The legacy experiment view is left unchanged; it has no linkability check to key the fallback off.

Automated (run by the agent): the three touched Jest suites pass (108 tests). New coverage: `getExposureFallbackFilter` cases guard that the fallback never applies to custom exposure criteria and that unknown variants degrade to an "is set" check; `applySessionLinkability` cases guard the substitution (including still dropping unlinkable metric steps) and that a linkable exposure event is never replaced; `experimentReplayTabLogic` cases guard that a server-side default exposure builds the flag-value filter instead of blanking the tab, while a server-side custom exposure still blanks with the explanation. Frontend typecheck clean.

No manual UI testing done by the agent. The `$feature/<flag_key>` property filter shape was verified against real data in the original investigation that produced this PR.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs to update.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Original diagnosis and PR by @concernedsebster; revised per @mp-hog's [review feedback](https://github.com/PostHog/posthog/pull/71400#issuecomment-5056266512) by PostHog Code (Claude), which also resolved the merge conflict with the #71639/#71919 refactor by rebasing onto master.

Decisions along the way:

- The fallback is keyed off the existing `seenTogether` linkability check instead of unconditionally replacing the `$feature_flag_called` filter (the PR's original approach), so client-side experiments keep true exposure semantics.
- `applySessionLinkability` returns `exposureUnlinkable: false` when the fallback substitutes, so existing consumers' blanking logic needed no changes; the new `usedExposureFallback` flag only drives labelling.
- Left `getViewRecordingFiltersLegacy` (legacy experiment view) unchanged: that surface has no linkability check, so a check-gated fallback can't be expressed there, and unconditional replacement was what the review asked to avoid.

---


jurajmajerik (approved):
Nice, thank you! Tested and works well.

mp-hog:
Thanks @concernedsebster ! There's a parallel track integrating recordings into the experiment page (right now behind feature flags replay-experiment-context and experiment-recordings-tab):

- #71639 (merged) fixes the same class from the other side: it runs replay's seen_together check when building the links, drops metric steps that are never session-linked, and disables the button when the exposure event itself isn't linkable. So today (since it's merged) the behavior is: The button explains itself instead of landing empty.
- #71919 added a "Recordings" tab on the experiment page whose whole list is exposure-defined — and it inherits the same limitation: when the exposure event isn't session-linkable it renders a warning banner instead of a list.
- The session_context endpoint behind the player already merges exposure events with the stamped $feature/<key> property, for exactly the reason you found.

Suggestion: keep the exposure event as the primary filter, since it's what the analysis counts (custom exposure criteria, run window, variant handling), and use $feature/<key> as a fallback for the default-exposure branch when the linkability check says the exposure event can't match any session — labelled as "flag was active in this session" rather than "exposed", because the property reflects the flag's current value per event rather than the enrollment moment. That turns a blank surface into an approximate but honest one, and it's the change that also unblanks the recordings tab.

So in short: Keep the behavior of #71639 and add your logic as a fallback.

Would you be up for adjust the PR that way? Happy to jump into your PR if you prefer

concernedsebster:
got it, thanks for the context here @mp-hog ! i'm aligned with your approach, this makes a lot more sense. feel free to jump into my PR.

mp-hog:
Revised per the [suggestion above](https://github.com/PostHog/posthog/pull/71400#issuecomment-5056266512): the `$feature_flag_called` exposure filter stays primary, and `$feature/<flag_key>` now kicks in only as a fallback when the linkability check (from #71639) marks the default exposure event as never session-linked. The per-variant buttons get a tooltip explaining the approximation, and the recordings tab (from #71919) shows the list with an info banner instead of the warning banner. Custom exposure criteria keep the existing blank-with-explanation behavior, and the legacy view is untouched.

mp-hog:
Follow-up ([beb66aee](https://github.com/PostHog/posthog/commit/beb66aee42444bc671fbc1980694ec53328394f4)): the fallback filter is now typed as an event property (`type: 'event'`) instead of `type: 'feature'`. A self-review traced the feature-typed shape through the recordings query backend: `is_event_property` in `posthog/session_recordings/queries/utils.py` only routes `type == 'event'` into the optimized events subquery, so a feature-typed filter skips it, survives `_strip_person_and_event_and_cohort_properties`, and fires `capture_exception(UnexpectedQueryProperties)` on every query — landing in replay's error tracking — before being applied via the slower replay-scope events lazy join. Event-typed is also how the taxonomic picker and the experiments exposure filters already shape `$feature/<flag_key>` properties, and it's the shape the recordings backend tests cover.

PR #71407, opened 2026-07-16
feat(capture): swappable custom-key map in global rate limiter

eli-r-ph:
## Problem

The capture global rate limiter's custom per-key threshold overrides are baked in from a CSV env var at construction and can't change without a redeploy. To move those overrides to a dynamically refreshed source (later PRs), the limiter first needs a way to swap its custom-key map at runtime without hurting the hot read path.

This is PR 1 of a stack. It only adds the seam. Behavior is unchanged.

## Changes

- Store `custom_keys` as `Arc<ArcSwap<HashMap<String, u64>>>` so the whole map can be replaced atomically with a lock-free load on the read path.
- Add an optional `CustomKeyResolver` closure and a single `resolve()` helper; route `is_custom_key`, `check_limit_internal`, and `process_read_results` through it.
- Add `replace_custom_keys` to the trait and impl for whole-map CAS swaps.
- With no resolver injected, `resolve()` is an exact `map.get(key)` lookup, identical to today. Capture keeps seeding the ArcSwap from the CSV at construction, so the feature-off path is byte-for-byte the same.

> [!NOTE]
> No behavior change on its own. The dynamic source and the hierarchical resolver land in the stacked capture PR.

Automated only (I'm an agent, no manual testing):

- `cargo clippy -p limiters --all-targets` clean.
- `cargo test -p limiters` — 55 unit + 10 integration tests pass, including new coverage for the injected resolver and a concurrent read-during-swap consistency test (`test_replace_custom_keys_concurrent_read_is_consistent`).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude, via the Cursor agent) wrote this at Eli's direction. The design decision worth flagging: we chose `ArcSwap` over a `DashMap` or an `RwLock<HashMap>` so the hot path stays lock-free and readers never observe a partially-updated map — the refresh loop swaps the whole map in one store. The resolver is a closure injected by capture rather than logic in `common`, so capture's `token:distinct_id` key structure never leaks into the shared crate.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -6086,6 +6086,7 @@ name = "limiters"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "arc-swap",
  "async-trait",
  "chrono",
  "common-redis",
--- a/rust/common/limiters/Cargo.toml
+++ b/rust/common/limiters/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+arc-swap = { workspace = true }
 async-trait = { workspace = true }
 anyhow = { workspace = true }
 common-redis = { path = "../redis" }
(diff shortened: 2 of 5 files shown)


eli-r-ph:
Folding this into #71408. An API change and its sole caller are an atomic unit — the common `custom_keys` type change here can't compile standalone without the capture caller update in #71408 (CI builds the whole workspace). Combined diff is ~656 LOC, so merging the two keeps a single reviewable, buildable PR.

Issue #13223, opened 2022-12-08
Regressions in new data table component


Twixes:
## Bug description

Noticed a few problems in the new feature-flagged events table in Persons (e.g. on [this marius@posthog.com page](https://app.posthog.com/person/pI40e5w84oZfhqdX4Sqf9XVmH5vVM3pyOMY1KRfe9XT#activeTab=events)):
1. Property value columns don't show anything
1. There's no "Time" column
1. There's a redundant "Person" column when on the person page
1. Similarly, there's no date separators
1. _Also_, the ClickHouse query made by the new UI takes very long (literally 20 s on the person page linked above for me, as opposed to 0.5 s with the old UI), because there no longer is a minimum timestamp filter to avoid going through too many CH parts in one query
1. The event buffer notice is missing ("Note that some events with a never-before-seen distinct ID are deliberately delayed by 60 seconds ..."
1. "More"/kebab button column is too wide, it should have `width: 0`

### `data-exploration-live-events` disabled
<img width="1050" alt="on" src="https://user-images.githubusercontent.com/4550621/206511932-545d39af-fdbf-43a8-9b45-39b42acb633b.png">

### `data-exploration-live-events` enabled
<img width="1033" alt="off" src="https://user-images.githubusercontent.com/4550621/206511899-f85868df-e5b3-48ac-b19b-6dd6d57bafd5.png">

mariusandra:
I'm fixing these issues in https://github.com/PostHog/posthog/pull/13251 and have some questions.

> 1. Property value columns don't show anything

I'm not able to reproduce this

> 1. There's no "Time" column

It was also there for me

> 1. There's a redundant "Person" column when on the person page

Fixed

> 1. Similarly, there's no date separators

Do we actually want them? I fear we'll need to remove these when we move to a [virtualized data table](https://github.com/PostHog/posthog/pull/11817), so I didn't implement them yet. Not sure if we even want them in PostHog 3000. Also, if the table is not sorted by time, should there be another label for the other sort key?

Because of these questions, I decided to skip this for now, and I'd gladly just remove this. 🤔 

> 1. _Also_, the ClickHouse query made by the new UI takes very long (literally 20 s on the person page linked above for me, as opposed to 0.5 s with the old UI), because there no longer is a minimum timestamp filter to avoid going through too many CH parts in one query

Will fix.

> 1. The event buffer notice is missing ("Note that some events with a never-before-seen distinct ID are deliberately delayed by 60 seconds ..."

[...]

PR #71408, opened 2026-07-16
feat(capture): redis-backed dynamic custom rate-limit thresholds

eli-r-ph:
## Problem

Custom per-key rate-limit thresholds in capture come from a CSV env var fixed at boot. Ops can't retune a noisy `token` or `token:distinct_id` without a redeploy. This PR makes the limiter refresh those thresholds from Redis on a timer, so the Django admin (PR #71409) can update them live.

Self-contained on `master`: includes both the `common` limiter changes (swappable custom-key map + Redis-backed refresh) and the capture wiring.

## Changes

- `common/limiters` owns the dynamic-threshold machinery:
  - `custom_keys` is an `Arc<ArcSwap<HashMap>>` so the map swaps atomically without locking hot-path readers.
  - New `custom_key_source` module: a `CustomKeyThresholdSource` trait, a `RedisCustomKeyThresholdSource` (single `GET` + JSON parse that owns its connection and reconnects after a failure), and the JSON parser.
  - `new()` spawns a lifecycle-aware refresh loop (`tokio::time::interval` + `select!` on a stop channel) when a source is configured; `shutdown()` stops it. The interval is clamped to a 1s floor so a misconfigured `0` can't panic the task. Each fetch swaps via `custom_keys.store()`.
  - Empty-state contract: only an explicit Redis blob is authoritative. A Redis/parse error **or** an absent key is fail-static (keep the last-good map) — a fresh rollout, an eviction, or an accidental delete never silently drops the hot overrides. A deliberate clear is an explicit empty blob (`{}`) written by the admin, which arrives as `Ok(Some(empty))`.
- `capture` builds the Redis source from config and injects a hierarchical resolver — look up the full `token:distinct_id` key, then fall back to the `token`-level threshold — so capture's key structure stays out of `common`. Applied to both the CSV seed and the dynamic map.
- Config: `GLOBAL_RATE_LIMIT_CUSTOM_THRESHOLD_KEY` (feature toggle) and `GLOBAL_RATE_LIMIT_CUSTOM_THRESHOLD_REFRESH_SECS` (default 60). Reuses the existing event-restrictions Redis URL — no new pool. Dormant unless the key is set, so the CSV path is unchanged.

> [!WARNING]
> Feature stays off until the Redis key env var is set (charts #13223) and the writer is deployed (Django #71409). Until then the limiter uses only the CSV seed.

Automated only (agent, no manual testing):

- `cargo fmt` + `cargo clippy --all-features` clean for `limiters` and `capture`.
- `cargo test -p limiters` (unit + integration) and `cargo test -p capture --lib` pass. Coverage includes:
  - JSON parse: flat / empty / malformed.
  - Refresh path: applies thresholds; **absent key is fail-static** (keeps current map); **explicit empty blob clears**; Redis error is fail-static.
  - Refresh task lifecycle: `new()` spawns + ticks + applies from a mock source; `shutdown()` stops it; zero interval is clamped, not panicking.
  - `RedisCustomKeyThresholdSource`: success, `NotFound`→none, connection-error→reset+rebuild, parse-error→keep-connection.
  - Hierarchical resolver (rstest table): token scope matches all of that token's `token:distinct_id` keys; `token:distinct_id` scope matches only the exact key — for both the CSV seed and the dynamic map.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude, via the Cursor agent) wrote this at Eli's direction, modeled on the existing event-restrictions repository/refresh pattern. Key decisions: (1) `common` owns the Redis fetch/timer/parse/swap while capture supplies only a source + resolver, keeping capture's key structure out of `common`; (2) reuse the event-restrictions Redis rather than a second pool; (3) fail-static on error **and** on an absent key — a Redis blip or a missing/evicted key never drops current thresholds; only an explicit empty `{}` blob (paired with the Django writer in #71409) clears them.


rust/capture/src/global_rate_limiter.rs
```diff
@@ -111,14 +138,62 @@ impl GlobalRateLimiter {
...
+        // separate store from the traffic-count Redis in `redis_instances`) and
+        // owns its own connection/reconnect; the common limiter runs the loop.
+        let custom_key_source: Option<Arc<dyn CustomKeyThresholdSource>> = if enable_dynamic_source
+        {
+            match (
+                config.global_rate_limit_custom_threshold_key.as_ref(),
+                config.event_restrictions_redis_url.as_ref(),
+            ) {
+                (Some(key), Some(redis_url)) => {
+                    let response_timeout = (config.redis_response_timeout_ms != 0)
```
z0br0wn, line 160: is it worth throwing a warn on either of these to isolate a bad configuration?

rust/common/limiters/src/global_rate_limiter.rs
```diff
@@ -571,6 +647,80 @@ impl GlobalRateLimiterImpl {
...
+    ) {
+        match source.fetch().await {
+            Ok(Some(map)) => {
+                let count = map.len();
+                config.custom_keys.store(Arc::new(map));
+                metrics::gauge!(CUSTOM_THRESHOLDS_LOADED_GAUGE, "scope" => scope).set(count as f64);
+                metrics::gauge!(CUSTOM_THRESHOLDS_LAST_REFRESH_GAUGE, "scope" => scope)
+                    .set(Utc::now().timestamp() as f64);
+            }
+            Ok(None) => {
```
z0br0wn, line 711: I think the Ok(None) → clear semantics can wipe the CSV seed at boot, contradicting the seed comment in capture.

The chain: an absent Redis key returns Ok(None) (custom_key_source.rs:139), this arm stores an empty map over the ArcSwap — which is the same map the CSV seed lives in (capture/src/global_rate_limiter.rs:187) — and the interval fires immediately at startup. So a rollout where the threshold key is configured but Django hasn't written the blob yet clears the CSV overrides seconds after boot. The seed comment says "until then (and if Redis is unreachable) the CSV seed applies", but that only holds for the Err arm — reachable-but-keyless lands here and clobbers the seed.

Same failure mode if the key is ever deleted by accident or evicted (this is the event-restrictions Redis — if its maxmemory policy is allkeys-*, an idle config key is fair game). All silent, fail-open, and the keys ops bothered to override are exactly the hot ones.

Suggestion: treat NotFound as fail-static too (keep current map, keep the not_found counter), and make an explicit empty blob the clear signal — Django writes {}, which parses as Ok(Some({})) and flows through the Ok(Some) arm. Admin keeps a deliberate clear path, absence (the ambiguous state) becomes safe, and zero new code beyond changing this arm. The loaded gauge + last_refresh timestamp already distinguish the two cases for observability.

z0br0wn (approved):
one concern, lgtm after that's addressed

rust/capture/src/global_rate_limiter.rs
eli-r-ph: we have monitors for that already (separate PR) we pick that up elsewhere

rust/common/limiters/src/global_rate_limiter.rs
eli-r-ph: hmm I thought we dealt with this specific scenario already and in tests...I'll review it, thanks!

PR #71409, opened 2026-07-16
feat(capture): admin for dynamic global rate-limit thresholds

eli-r-ph:
## Problem

Custom per-key thresholds for the capture global rate limiter need an ops-facing way to be set and updated without a redeploy. This adds the write side: a model plus staff-only admin that publishes the thresholds to the Redis key capture reads (the stacked capture PR #71408).

Pairs with #71408 — the empty-state contract is split across both PRs (see below), so they should land together.

## Changes

- New `GlobalRateLimitThresholdConfig` model: `token`, optional `distinct_id`, `threshold`, `note`, plus `created_at`/`updated_at` audit columns; `unique_together = (token, distinct_id)`.
- `post_save`/`post_delete` signals regenerate one Redis key holding a JSON blob `{ resolved_key: threshold }` on `PLUGINS_RELOAD_REDIS_URL`.
  - Publishes via `transaction.on_commit`, so a rolled-back admin save (admin wraps `save_model` in a transaction) never leaks a threshold capture would read.
  - Writes an explicit empty blob (`{}`) when the table is empty rather than deleting the key: capture treats an absent key as fail-static, so a deliberate clear must be a written state (matches #71408).
- `resolved_key` matches capture's lookup exactly: `token`, or `token:distinct_id` with `distinct_id` truncated to the same 128-char cap capture uses (verified against `MAX_DISTINCT_ID_CHARS` in `rust/capture/src/global_rate_limiter.rs`).
- `threshold = 0` `help_text` documents that it rate-limits the key to zero, but the cache-based limiter can still leak an isolated event on a cold cache miss, so it is not a hard block — use an `EventRestriction` in the Django admin to block a key permanently (addresses veria-ai review).
- Staff-only `ModelAdmin` (team id resolved from token via subquery, resolved-key display, `created_at`/`updated_at` shown read-only). No customer API or frontend.

> [!NOTE]
> Token-keyed ops table with no `team_id`, so it isn't team-scoped and needs no IDOR baseline entry (same as `EventIngestionRestrictionConfig`).

Automated only (I'm an agent, no manual testing):

- `bin/hogli test posthog/models/test/test_global_rate_limit_threshold_config.py` and `posthog/admin/test_admin.py::TestGlobalRateLimitThresholdConfigAdmin` — 10 pass: resolved-key format + 128-char truncation, blob regeneration on save, threshold update, delete regenerates, **delete-last-row writes an empty blob** (not a delete), **rollback discards the Redis write**, plus admin team-id annotation and resolved-key display.
- `ruff check`/`format` clean; `makemigrations --check` clean and `sqlmigrate` verified (`max_migration.txt` at `1259`).

These catch the regressions that matter here: the Redis key format must stay byte-identical to what capture looks up; the clear path must write `{}` (not delete) to satisfy capture's fail-static-on-absent contract; and a rolled-back save must not publish.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs — internal ops tooling only.

## 🤖 Agent context

I (Claude, via the Cursor agent) wrote this at Eli's direction. Invoked the `/django-migrations` skill for the migration. Modeled on `EventIngestionRestrictionConfig` (same Redis-writer + signals shape), but used `UUIDModel` instead of the deprecated `UUIDTModel`, and a single JSON blob keyed by resolved key rather than per-type arrays. The 128-char distinct_id truncation was confirmed against capture's Rust constant so the written key and the lookup key match. The `transaction.on_commit` deferral and the write-`{}`-on-empty contract came out of review with #71408.


posthog/models/global_rate_limit_threshold_config.py
```diff
@@ -0,0 +1,77 @@
...
+        return self.token
+
+
+def regenerate_redis_thresholds() -> None:
+    """Rebuild the Redis threshold blob from all rows; delete the key when empty."""
+    redis_client = get_client(PLUGINS_RELOAD_REDIS_URL)
+
+    configs = list(GlobalRateLimitThresholdConfig.objects.all().order_by("id"))
+    if not configs:
+        redis_client.delete(CUSTOM_THRESHOLDS_REDIS_KEY)
```
z0br0wn: I think delete-on-empty is the wrong half of the empty-state contract, given how capture consumes this key.

On the capture side (#71408), an absent key returns Ok(None) from the source (custom_key_source.rs:139) and the refresh loop maps that to "clear all custom thresholds" — including the CSV seed, which lives in the same swapped map (capture/src/global_rate_limiter.rs:187). Since the refresh interval fires immediately at startup, the combined behavior of the two PRs as written is:

  - Enable GLOBAL_RATE_LIMIT_CUSTOM_THRESHOLD_KEY in capture before any row exists in this table → key absent → capture wipes its CSV overrides seconds after boot.
  - The key gets evicted (this Redis's maxmemory policy permitting) or deleted by accident → all overrides silently clear, and since sync here is signal-driven only, nothing rewrites the blob until the next admin edit.

Both cases are silent and fail-open, and the keys ops bothered to override are by definition the hot ones.

Since both PRs are open, suggest fixing the contract on both sides now: here, write an explicit empty blob instead of deleting —

  if not configs:
      redis_client.set(CUSTOM_THRESHOLDS_REDIS_KEY, json.dumps({}))
      return

— and on the capture side, treat NotFound as fail-static (keep the current map) the same as Err. Then "cleared" is always an intentional, written state ({} parses as Ok(Some({})) and flows through the normal replace path), while "absent" — the ambiguous state — becomes safe. Cross-linking my comment on the capture PR where the Ok(None) arm lives.

posthog/models/global_rate_limit_threshold_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Each row maps a token (optionally scoped to a single distinct_id) to a
+    threshold that overrides the limiter's global default. All rows serialize to a
+    single Redis key that capture refreshes on a timer. Internal ops tooling only:
+    no customer-facing API or UI.
+    """
+
+    # Matches Team.api_token's max_length so any valid capture token can get an override.
+    token = models.CharField(max_length=200)
+    distinct_id = models.CharField(
```
z0br0wn, line 34: distinct_id stores 450 chars but resolved_key truncates to 128, so two rows differing only past char 128 pass unique_together yet produce the same key, and the blob silently keeps the higher id. Truncate in clean()/save() so the stored row matches what's enforced.

posthog/models/global_rate_limit_threshold_config.py
```diff
@@ -0,0 +1,77 @@
...
+        max_length=450,
+        blank=True,
+        default="",
+        help_text=(
+            "Optional. Empty applies the threshold to the whole token; "
+            "set applies it only to this token and distinct_id."
+        ),
+    )
+    threshold = models.PositiveBigIntegerField(help_text="Max events allowed per rate-limit window for this key.")
```
z0br0wn: PositiveBigIntegerField allows 0, which on the capture side means a total block for the key. If that's an intended kill switch, document it in help_text; if not, MinValueValidator(1).

posthog/migrations/1259_globalratelimitthresholdconfig.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 5.2.14 on 2026-07-16 03:19
+
+from django.db import migrations, models
+
+import posthog.models.utils
+
+
+class Migration(migrations.Migration):
```
z0br0wn, line 8: Worth adding created_at/updated_at while migration 1257 is unmerged — free now, a second migration later. For an ops table, "when did this limit change" matters.

posthog/models/global_rate_limit_threshold_config.py
```diff
@@ -0,0 +1,77 @@
...
+    configs = list(GlobalRateLimitThresholdConfig.objects.all().order_by("id"))
+    if not configs:
+        redis_client.delete(CUSTOM_THRESHOLDS_REDIS_KEY)
+        return
+
+    thresholds = {config.resolved_key: config.threshold for config in configs}
+    redis_client.set(CUSTOM_THRESHOLDS_REDIS_KEY, json.dumps(thresholds))
+
+
+@receiver(post_save, sender=GlobalRateLimitThresholdConfig)
```
z0br0wn, line 79: Django's admin wraps change views in transaction.atomic, so these signals fire mid-transaction — a rollback after save_model leaves Redis updated with a row the DB doesn't have. Wrap the regenerate in transaction.on_commit(). A small periodic reconcile task would also self-heal failed writes and eviction — fine as a follow-up.

z0br0wn (approved):
claude had some feedback, approving now and trusting you to make any changes if they're needed

posthog/models/global_rate_limit_threshold_config.py
eli-r-ph: this is an intended feature (kill switch) I'll review but should be docuemented as such already... 👀 will update if that disappeared

posthog/migrations/1259_globalratelimitthresholdconfig.py
eli-r-ph: good call, updating

posthog/models/global_rate_limit_threshold_config.py
eli-r-ph: this was already called out by AI review and went with this, its not an issue IMO.

128 cap is a feature of rate limiter backend to ensure huge keys can't clog limiter caches/Redis. It's an intentional decision to keep local limiter caches + Redis from being DDoS'd vs. blocking colliding keys if the team has many IDs with a very long common prefix. Either way we want the end user's attention on it if this is happening, but we do want the full granularity of the key in storage here and in backend logs/warnings, to surface which ID caused the wide blocking in that case.

eli-r-ph: 🤔 450 max size is a bit larger than I expected though; I think from planning step this was a size in parity with prior art in Django admin, but I'll double check. We do see extreme case dist IDs in excess of this in the DB (but few)

posthog/models/global_rate_limit_threshold_config.py
eli-r-ph: good callout, this came up in planning/greptile too. this is a known thing in parity with event restrictions handling in Django; it's not likely to be a problem at human speed of updates (hasn't been yet for restrictions either)

could make both more robust in a follow-on but seems OK for now?

PR #71417, opened 2026-07-16
fix(mcp-analytics): span tool report charts over the selected window

gesh:
## Problem

On an individual tool's report, the tile sparklines and the reliability trend charts built their day axis from only the days that had data. When a tool had data on a single day in the selected window (common arriving from the Tool quality tab's 7-day default on a low-traffic tool), the series was one point, so the line couldn't draw and every tile lost its sparkline.

## Changes

Pad the day axis to the selected window (mirroring the Tool quality tab), gap-filling empty days, so the charts always match the chosen range and have enough points to render a line. The no-data empty state is preserved.

Added a unit case to `mcpAnalyticsToolDetailLogic.test.ts` covering the single-active-day window (guards the "sparklines lose their line" regression the earlier tests didn't). Ran `hogli test products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.test.ts` — 4/4 pass. I (Claude) did not exercise this in a running frontend.

## 🤖 Agent context

Georgis spotted the missing sparkline lines while reviewing the tool-quality date-persist work. I traced it to `dailyChartData` spanning only data-days instead of the selected window, then applied the same range-padding the Tool quality tab already uses. Invoked the `/writing-tests` skill before adding the regression case.

### DIFF
--- a/products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.test.ts
+++ b/products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.test.ts
@@ -53,4 +53,19 @@ describe('buildDailyChartData', () => {
             sessions: [3, 0, 1],
         })
     })
+
+    // Guards the "sparklines lose their line" regression: with only one active day, an axis that spans
+    // just the data would be a single point (no line). A window range pads it out so the line renders.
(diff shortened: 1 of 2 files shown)


gesh:
Covered in https://github.com/PostHog/posthog/pull/71435

PR #71432, opened 2026-07-16
chore: Inline the switch user UX

luke-belton:
## Problem

Switching user is something I've needed so infrequently, it doesn't make sense to make it one of the primary actions on the impersonation notice.

## Changes

Inline the switch user menu with the currently impersonated user's email.

https://github.com/user-attachments/assets/b7fe446e-1b89-4e5f-b50f-09414a432765

Verified locally

### DIFF
--- a/frontend/src/layout/navigation/ImpersonationNotice/ImpersonationNotice.scss
+++ b/frontend/src/layout/navigation/ImpersonationNotice/ImpersonationNotice.scss
@@ -167,6 +167,7 @@
     // so it reads as the (clickable) email rather than a standalone button.
     &__user-trigger.LemonButton {
         display: inline-flex;
+        max-width: 100%;
         min-height: auto;
         padding: 0 0.125rem;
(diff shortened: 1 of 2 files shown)


frontend/src/layout/navigation/ImpersonationNotice/ImpersonationNotice.scss
```diff
@@ -162,6 +162,18 @@
         line-height: 1.5;
         color: var(--text-primary);
     }
+
+    // The change-user menu trigger sits inline in the "Signed in as …" sentence,
+    // so it reads as the (clickable) email rather than a standalone button.
+    &__user-trigger.LemonButton {
```
xljones, line 168: a border to make this more button-like would be good.

xljones (approved):
Nice. I'd love to see the upgrade to r/w button promoted here unless you've other ideas for something else.

PR #71435, opened 2026-07-16
fix(mcp-analytics): render tool report charts across the selected window

gesh:
## Problem

The Tool quality tab and an individual tool's report weren't rendering their activity / latency charts correctly across the selected window. The headline tile numbers were right, but the charts read empty or flat:

- **Flat at zero for cross-timezone viewers.** Day buckets were parsed in the *browser's* timezone and then converted to the project timezone, shifting them off midnight so they matched no axis point — the whole series drew flat at zero for anyone not sitting in the project's timezone.
- **Sub-day windows collapsed.** "Last 12 hours" landed in a single daily bucket → one point, no trend.
- **No zero-fill across the window.** Empty buckets weren't filled, so a window with sparse data didn't span — a tool with one active day drew no continuous line.

Screenshot 2026-07-16 at 11 57 36

## Changes

Screenshot 2026-07-16 at 11 57 21

- **Timezone.** Read the bucket keys in the project timezone (`dayjs.tz(s, tz)` instead of `dayjs(s).tz(tz)`), so points land on the axis. This is what unfroze the flat-zero series. Same one-line fix to the dashboard's own copy of the helper, which had the identical latent bug.
- **Zero-fill and span the window.** Count series (calls / errors / users / sessions) fill empty buckets with 0 so the line runs across the whole selected range. Latency (p50 / p95) intentionally leaves gaps (NaN) rather than a misleading 0 ms, so periods with no calls aren't drawn as real dips — with genuinely sparse data that shows as points rather than a continuous line, which is honest.
- **Bucket by the window's interval.** Group by `getDefaultInterval` (the app-wide auto-interval the dashboard already uses), so a sub-day window buckets by hour (minute under ~2h) instead of one day point; day+ windows are unchanged. A shared `timeBuckets.ts` builds the gap-filled axis, and the `MCPToolDailyStatsQuery` runner takes an `interval` and `dateTrunc`s to match the frontend keys.

> [!NOTE]
> Supersedes #71417 (the tool-report sparkline padding fix) — it rewrote the same `buildDailyChartData`, and the window-spanning axis is now produced by the interval-aware bucket keys. #71417 can be closed.

Automated only — I (Claude) don't have a browser here.

- `timeBuckets` unit tests including a cross-timezone regression guard (runs under a non-UTC ambient tz to prove buckets don't shift — the flat-zero bug).
- Tab + tool-report logic tests: rows project onto day **and** hourly bucket keys (guards the "12 hours → one point" collapse).
- Backend runner test: `interval=hour` splits same-day calls into separate hourly buckets.
- Full `products/mcp_analytics/frontend` Jest suite (142) and the tool-tables backend suite (28) green; dashboard is otherwise a zero-diff and its 87 tests still pass. Frontend typecheck clean for the changed files.

Reviewer should eyeball a sub-day range (last 12 hours) on both the tab and a tool report, and confirm a custom absolute range still buckets sensibly. (Local seed data lands on a single day, so the latency chart there legitimately shows one point.)

## 🤖 Agent context

Georgis drove this from the tool report while testing date ranges, where the charts read flat/empty. The timezone bug surfaced from the live app returning one bucket that never matched a key; fixed by parsing the bucket string as the project tz. Main decision: bucket by the standard `getDefaultInterval` rather than a bespoke interval picker — it's what the dashboard and web/marketing analytics already use, so all three MCP surfaces stay consistent (trade-off: "last 1 hour" shows a single hourly point, same as everywhere else in the app). Custom-range label formatting and tile-layout polish were deliberately left for a follow-up.

Invoked `/writing-tests` before adding the regression cases.

### DIFF
--- a/products/mcp_analytics/backend/hogql_queries/tool_tables.py
+++ b/products/mcp_analytics/backend/hogql_queries/tool_tables.py
@@ -375,10 +375,14 @@ class MCPToolDailyStatsQueryRunner(AnalyticsQueryRunner[MCPToolDailyStatsQueryRe
         return mcp_query_date_range(self.team, self.query.dateRange)
 
     def to_query(self) -> ast.SelectQuery | ast.SelectSetQuery:
+        # Bucket granularity comes from the frontend's getDefaultInterval, so a sub-day window buckets
+        # by hour/minute instead of collapsing to a single day point. dateTrunc respects the team
+        # timezone, so the buckets line up with the frontend's gap-fill keys. Defaults to day.
+        interval = self.query.interval.value if self.query.interval else "day"
         return parse_select(
             """
             SELECT
-                toString(toDate(timestamp)) AS day,
+                toString(dateTrunc({interval}, timestamp)) AS day,
                 count() AS calls,
                 {_IS_ERROR} AS errors,
                 {_P50} AS p50,
@@ -391,6 +395,7 @@ class MCPToolDailyStatsQueryRunner(AnalyticsQueryRunner[MCPToolDailyStatsQueryRe
             ORDER BY day
             """,
             placeholders={
+                "interval": ast.Constant(value=interval),
                 "_CONVERSATION_ID": parse_expr(_CONVERSATION_ID),
                 "_IS_ERROR": parse_expr(_IS_ERROR),
                 "_P50": parse_expr(_P50),
--- a/products/mcp_analytics/backend/hogql_queries/tool_tables.py
+++ b/products/mcp_analytics/backend/hogql_queries/tool_tables.py
@@ -378,6 +378,9 @@ class MCPToolDailyStatsQueryRunner(AnalyticsQueryRunner[MCPToolDailyStatsQueryRe
         # Bucket granularity comes from the frontend's getDefaultInterval, so a sub-day window buckets
         # by hour/minute instead of collapsing to a single day point. dateTrunc respects the team
         # timezone, so the buckets line up with the frontend's gap-fill keys. Defaults to day.
+        # Explicit generous LIMIT so a fine interval over a wide window (reachable via the interval
+        # field, e.g. hourly over a quarter from an MCP caller) isn't silently cut to the default 100
+        # rows — which, with ORDER BY day ASC, would drop the most recent buckets.
         interval = self.query.interval.value if self.query.interval else "day"
         return parse_select(
             """
@@ -393,6 +396,7 @@ class MCPToolDailyStatsQueryRunner(AnalyticsQueryRunner[MCPToolDailyStatsQueryRe
             WHERE {where}
             GROUP BY day
             ORDER BY day
+            LIMIT 10000
             """,
             placeholders={
                 "interval": ast.Constant(value=interval),
--- a/products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.ts
+++ b/products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.ts
@@ -4,7 +4,7 @@ import { urlToAction } from 'kea-router'
 
 import api from 'lib/api'
 import { dayjs } from 'lib/dayjs'
-import { dateFilterToText, dateStringToDayJs } from 'lib/utils/dateFilters'
+import { dateFilterToText, dateStringToDayJs, getDefaultInterval } from 'lib/utils/dateFilters'
 import { teamLogic } from 'scenes/teamLogic'
 import { urls } from 'scenes/urls'
 
@@ -20,8 +20,10 @@ import {
     MCPToolTopUserItem,
     NodeKind,
 } from '~/queries/schema/schema-general'
+import { IntervalType } from '~/types'
 
 import type { mcpAnalyticsToolDetailLogicType } from './mcpAnalyticsToolDetailLogicType'
+import { buildBucketKeys, normalizeBucket } from './timeBuckets'
 
 export interface ToolSummary {
     calls: number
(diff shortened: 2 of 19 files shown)


products/mcp_analytics/frontend/mcpDashboardOverviewLogic.ts
```diff
@@ -343,8 +343,11 @@ export function lastBucketIsInProgress(
 }
 
 export function normalizeBucket(raw: unknown, timezone: string): string {
+    // The raw value is a project-timezone wall clock (dateTrunc runs in the team timezone), so parse
+    // it AS that timezone. `dayjs(s).tz(tz)` reads it in the browser tz then converts, shifting day
+    // buckets off midnight so they match no key — the chart reads flat for non-project-tz viewers.
     const s = String(raw ?? '')
-    return s ? dayjs(s).tz(timezone).format(BUCKET_FORMAT) : ''
+    return s ? dayjs.tz(s, timezone).format(BUCKET_FORMAT) : ''
```
pauldambra: 🤖 **Automated review note from PostHog Code, acting on the requester’s behalf.**

**[P1] Preserve serialized bucket wall-clock values outside UTC**

The query API serializes `dateTrunc` buckets as ISO strings ending in `Z` (the existing tests in this file document that contract). PostHog’s patched Day.js intentionally treats those inputs to `dayjs.tz` as instants, so `2026-06-19T00:00:00Z` becomes `03:00` in `Europe/Athens` instead of remaining local midnight. The generated zero-fill key is still `00:00`, so the join misses and the dashboard/tool-quality series can remain flat for non-UTC projects. Please parse the serialized bucket as a project-timezone wall clock before formatting, and cover an ISO `Z` input with a non-UTC timezone. The duplicate `normalizeBucket` in `timeBuckets.ts` needs the same treatment.

products/mcp_analytics/frontend/mcpAnalyticsToolQualityLogic.ts
```diff
@@ -383,6 +360,12 @@ ORDER BY day
             (dateFilter: DateFilter): string =>
                 dateFilterToText(dateFilter.dateFrom, dateFilter.dateTo, 'the selected range') ?? 'the selected range',
         ],
+        // Grouping interval for the daily chart — PostHog's standard auto-choice, so a sub-day window
+        // (e.g. last 12 hours) buckets hourly instead of collapsing to a single day point.
+        interval: [
+            (s) => [s.dateFilter],
+            (dateFilter: DateFilter): IntervalType => getDefaultInterval(dateFilter.dateFrom, dateFilter.dateTo),
```
pauldambra, line 363: 🤖 **Automated review note from PostHog Code, acting on the requester’s behalf.**

**[P2] Select minute buckets for the offered sub-hour ranges**

`McpDateFilter` offers Last 5/15/30 minutes, but those presets are emitted as absolute ISO timestamps. `getDefaultInterval` returns `day` for an absolute range shorter than one hour, so each of those visible presets still collapses to a single daily point. Please derive `minute` for sub-hour absolute windows (and apply the same fix to the tool-detail selector), or otherwise preserve the preset’s intended interval.

products/mcp_analytics/frontend/timeBuckets.ts
```diff
@@ -0,0 +1,80 @@
...
+    interval: IntervalType
+): string[] {
+    const { start, end } = resolveWindow(dateFrom, dateTo, timezone)
+    const last = startOfBucket(end, interval).valueOf()
+    const keys: string[] = []
+    let cursor = startOfBucket(start, interval)
+    // Bounded windows keep this small; the cap is just a guard against a pathological range.
+    for (let i = 0; cursor.valueOf() <= last && i < 100000; i++) {
+        keys.push(cursor.format(BUCKET_FORMAT))
+        cursor = cursor.add(1, interval)
```
pauldambra: 🤖 **Automated review note from PostHog Code, acting on the requester’s behalf.**

**[P2] Recalculate timezone offsets while stepping across DST**

Day.js retains the original UTC offset when `add` advances a timezone-aware value. For example, a New York daily range from March 7 through March 9, 2026 emits only March 7 and March 8 because the cursor remains at `-05:00` while the end bucket is `-04:00`; hourly ranges also create the nonexistent spring-forward `02:00` bucket. Advance buckets in a way that reapplies the timezone/DST offset, and add a range crossing a DST transition to the helper tests.

pauldambra (commented):
🤖 Automated review submitted by PostHog Code, acting on the requester’s behalf.

products/mcp_analytics/frontend/mcpDashboardOverviewLogic.ts
gesh: Fixed. `normalizeBucket` now reads the bucket in UTC (`dayjs.utc(s)`) instead of `dayjs.tz(s, tz)` — a Z-stamped bucket keeps its wall-clock digits and matches the generated keys on non-UTC projects (`dayjs.tz` was converting the instant by the project offset). While here, I removed the duplication you're pointing at: the dashboard no longer carries its own `normalizeBucket`/`buildBucketKeys`/etc. — both surfaces now use the single `timeBuckets` interface, so the fix lives in one place. Added non-UTC Z-input test coverage.

products/mcp_analytics/frontend/timeBuckets.ts
gesh: Fixed. `buildBucketKeys` now re-anchors each bucket from the window start (`startOfBucket(first.add(i, interval), interval)`) instead of cumulatively adding to a timezone-aware cursor, so a DST-crossing range no longer drops (or repeats) a bucket. Added a New York spring-forward test — Mar 7–9 2026 now yields all three days. Same fix in the consolidated helper (dashboard shares it now).

products/mcp_analytics/frontend/mcpAnalyticsToolQualityLogic.ts
gesh: Confirmed valid — the 5/15/30-min presets emit absolute ISO, and `getDefaultInterval` returns `day` for a sub-hour absolute range. Decision after discussion: keep using the standard `getDefaultInterval` directly (consistent with the dashboard, web analytics, and insights, none of which auto-select minute) rather than add a bespoke MCP interval picker. So the presets stay, and a sub-hour window follows `getDefaultInterval`'s standard bucketing. A dedicated sub-hour interval would be a separate change if we decide we want it.

products/mcp_analytics/frontend/mcpAnalyticsToolDetailLogic.ts
```diff
@@ -314,8 +316,23 @@ export const mcpAnalyticsToolDetailLogic = kea<mcpAnalyticsToolDetailLogicType>(
...
+            (
+                dailyStats: DailyToolStat[],
+                dateFilter: DateFilter,
+                interval: IntervalType,
+                timezone: string
+            ): DailyChartData => {
+                const bucketKeys = buildBucketKeys(dateFilter.dateFrom, dateFilter.dateTo, timezone, interval)
+                return buildDailyChartData(dailyStats, bucketKeys, timezone)
+            },
+        ],
```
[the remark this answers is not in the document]
gesh: Fixed. `loadDailyStats` now uses a kea-loaders `breakpoint` (matching the sibling tab loader), so a superseded response fetched at a stale interval can't overwrite `dailyStats` and collapse a day into one hourly bucket. The other loaders render their own payload directly (no interval-dependent re-bucketing), so a stale response there is merely stale and self-corrects rather than misleading — left those as-is.

products/mcp_analytics/backend/hogql_queries/tool_tables.py
```diff
@@ -375,10 +375,14 @@ def query_date_range(self) -> QueryDateRange:
         return mcp_query_date_range(self.team, self.query.dateRange)
 
     def to_query(self) -> ast.SelectQuery | ast.SelectSetQuery:
```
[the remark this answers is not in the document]
gesh: Fixed. Added an explicit `LIMIT 10000` to `MCPToolDailyStatsQuery`, so a fine interval over a wide window (reachable via the `interval` field / the MCP tool) no longer hits the implicit 100-row cap that, with `ORDER BY day ASC`, silently dropped the most recent buckets. 10000 covers realistic chart windows. The sibling tab runner in the stacked follow-up (#71465) gets the same cap.

PR #71442, opened 2026-07-16
fix(auth): normalize pasted login codes to digits

eleftheriatrivyzaki:
## Problem

The emailed 6-digit login verification code was rejected as invalid even when correct. Copy-pasting the code out of an email client can drag in characters the user never sees - a trailing newline, a leading space, a non-breaking or zero-width space, a soft hyphen - and the code was compared with no normalization at any layer, so a visually-correct code failed the exact match.

## Changes

- `CodeBasedVerificationSerializer.validate_code` normalizes the submitted code: NFKC (so fullwidth digits fold to ASCII), then strips the noise an email client or manual grouping injects (whitespace, zero-width family, word joiner, BOM, soft hyphen, and a grouping hyphen like `123-456`). It then requires exactly 6 digits and rejects anything else, so garbage isn't mined for digits.
- The verify endpoint now routes through the serializer (`serializer.validated_data`) instead of reading `request.data.get("code")` raw, so normalization can't be bypassed.
- The login form applies the same normalization before submit and validates against 6 digits.

Stripping down to 6 digits is safe: it can't turn a wrong code into a right one, since the `constant_time_compare` still has to match exactly.

Added backend tests in `test_code_based_verification.py`: a no-DB serializer matrix (newline, surrounding/embedded/grouped spaces, hyphen grouping, non-breaking space, zero-width space, word joiner, BOM, soft hyphen, fullwidth digits), a negative test that non-6-digit input is rejected (letters around digits, digits embedded in text, too short, too long), and one endpoint case that a whitespace-wrapped correct code logs in. Ran `hogli test posthog/api/test/test_code_based_verification.py` - 25 passed. Frontend typecheck and lint clean. Not manually tested in the browser.

## 🤖 Agent context

---


yasen-posthog (approved):
LGTM, nice!

feliperalmeida (approved):

PR #71443, opened 2026-07-16
feat(review_hog): default urgency threshold to consider

sortafreel:
## Problem

ReviewHog's urgency threshold defaulted to "Should fix", which drops every
`consider`-level finding before it reaches the pull request. But `consider`
findings pass the same validation pass as everything else, so they are real,
worth-surfacing issues. We want the out-of-the-box experience to show all of
them.

## Changes

Flip the default `urgency_threshold` from `should_fix` to `consider` ("All
issues"). New users now publish every validated finding by default; they can
still move the slider stricter whenever they want.

The change lands in one place per layer:

- **Model + constant:** `ReviewUserSettings.urgency_threshold` `default`/`db_default`
  and `DEFAULT_URGENCY_THRESHOLD` both move to `consider`.
- **Migration 0017:** alters the column default and resets existing rows to
  `consider` (safe, the feature has no persisted rows yet, done for correctness).
- **Temporal inputs:** the backward-compat dataclass defaults
  (`ResolveActingUserResult` / `BuildBodyInput` / `PublishInput`) follow the model
  so "defaults match the model's" stays true.
- **Frontend:** the slider moves the "default" annotation onto "All issues",
  rewords "Should fix", and the two `?? 'should_fix'` fallbacks now fall back to
  `consider`.
- **Copy:** the settings serializer `help_text` and the validator prompt no longer
  claim `consider` is always suppressed (whether it surfaces now depends on the
  reader's threshold).

The `should_fix`-threshold test fixtures were renamed `_DEFAULT_PUBLISHED` →
`_SHOULD_FIX_PUBLISHED`, since they exercise the publish gate at that threshold and
never meant "the product default".

No screenshot: I could not stand up the app in this environment. The only visual
change is the slider's helper text and which stop reads as the default.

I could not run the Django test suite (no Django install) or the frontend
typecheck/Jest (no `node_modules`) in this environment, so I did not execute them.
I ran `ruff check` and `ruff format --check` on the touched Python (clean) and
byte-compiled every changed module. I updated the affected assertions to the new
default rather than adding tests, since the behavior these lock in already had
coverage:

- `test_settings_api` / `test_acting_user`: the auto-created row and the no-row
  resolve fallback now expect `consider`.
- `test_constants`: `DEFAULT_URGENCY_THRESHOLD` now mirrors `CONSIDER`.
- `test_publish_review` / `test_prepare_validation_markdown`: unchanged behavior,
  fixtures renamed only.

Please run the review_hog `backend:test` scope and `hogli build:openapi` before
merge. I hand-synced the two generated type description strings to match the new
`help_text` because codegen was not runnable here.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs reference this default (checked `docs/`); the product doc is
`products/review_hog/ARCHITECTURE.md`, updated in this PR.

## 🤖 Agent context

I (Claude) made this change for Alex, who asked to switch the default urgency
threshold to `consider`. Before writing anything I confirmed one open question:
the settings row is created eagerly on first tab open, so a default change alone
would not move anyone who had already opened the tab. Alex confirmed there are no
existing users yet (just merged, demoing soon), so I both changed the default and
added a data migration that resets any existing rows (variant 2).

Skills invoked: `/django-migrations` (the migration is hand-written because Django
was not installed to run `makemigrations`; I matched the `AlterField` output shape
and added a `RunSQL` data reset), `/improving-drf-endpoints` (serializer `help_text`
edit), and `/writing-tests` (the test edits are mechanical expectation updates, not
new coverage).

One judgment call worth flagging: I moved the temporal dataclass defaults to
`consider` too. They only matter for deserializing pre-field payloads, of which
there are none, but keeping them at `should_fix` would have made the "defaults
match the model's" comment false.

---


Twixes (approved):
A little odd that just changing just the default is this many changes, but they do seem relevant ✅

PR #71444, opened 2026-07-16
fix(billing): tighten posthog code limit UX

pawel-cebula:
## Problem

PostHog Code startup customers should not be able to remove their usage limit or raise it above the allowed cap. The old UI treated Code like other metered products, so it allowed confusing edits before Billing rejected them.

PostHog Code usage-based billing launched on July 16, 2026, so the seats notice should no longer say it is launching shortly.

Demo: https://www.loom.com/share/e07dbb38b8e34c13bd4512b4ac62f41e

## Changes

- Add product-specific billing limit config for PostHog Code, gated by startup-program status.
- Cap startup Code billing limits at $3,000 and disable limit removal actions.
- Show clearer help text and next-period cap messaging when the current limit is above the allowed cap.
- Add Storybook states for the Code limit, edit mode, and next-period capped state.
- Render the Code billing Storybook states at a billing-page width so Visual Review matches the real layout.
- Update the Code seats notice and retired-seat API message to say usage-based billing launched on July 16, 2026, and existing seats expire on the same date.

- Verified locally on the devbox with local Billing connected using a paid Code customer on the default $50 limit and startup metadata. Confirmed the $3,000 cap, disabled removal actions, and the July 16 launch copy.
- Added Storybook states covering the Code billing limit, edit mode, next-period capped state, and billing-page-width visual layout.
- Updated the retired Code seat API assertion so the launch-date warning stays covered.

## Automatic notifications

- [ ] Publish to changelog?


pawel-cebula:
* **#71444** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

frontend/src/scenes/billing/billingLimitConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+                billingLimitNextPeriod !== null &&
+                billingLimitNextPeriod <= POSTHOG_CODE_BILLING_LIMIT_MAX
+                    ? `Current usage is already above the Code billing limit cap, so this period stays at $${customLimitUsd.toLocaleString()}. The $${billingLimitNextPeriod.toLocaleString()} limit starts next period.`
+                    : null,
+        }
+    },
+}
+
+export const getBillingLimitConfig = (context: BillingLimitConfigContext): BillingLimitConfig => {
+    const productConfig = BILLING_LIMIT_CONFIG_BY_PRODUCT[context.product.type]?.(context)
```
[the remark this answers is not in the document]
pawel-cebula: I think this is a false positive after checking the stacked Billing PRs. `posthog_code_usage` is the new Billing product key from PostHog/billing#2030, and Billing serializes product keys into `product.type` (`CustomerProduct(type=product_key, ...)`). The `usage_key` for that product is `posthog_code_credits`, which is the usage meter, not the billing-limit product key.

PostHog/billing#2049 also stores and validates the startup cap under `custom_limits_map["posthog_code_usage"]`. The serializer still accepts legacy usage-key-shaped limit input for compatibility, but the canonical key for this product is the product key. I reverted the temporary `usage_key` fallback here: PostHog/posthog@82ffb458fec.

frontend/src/scenes/billing/billingProductLogic.ts
```diff
@@ -772,31 +794,40 @@ export const billingProductLogic = kea<billingProductLogicType>([
...
             submit: async ({ input }) => {
-                if (props.product.current_amount_usd && input < props.product.current_amount_usd) {
+                const currentAmountUsd = props.product.current_amount_usd
+                    ? parseFloat(props.product.current_amount_usd)
+                    : null
+                const projectedAmountUsd = props.product.projected_amount_usd
+                    ? parseFloat(props.product.projected_amount_usd)
+                    : null
+
+                if (input !== null && currentAmountUsd !== null && input < currentAmountUsd) {
```
abhischekt, line 830: If I understand this right then we won't trigger the `Billing limit warning` anymore? Just double checking it's intentional because we fire `updateBillingLimits({[type]: null})`

frontend/src/scenes/billing/billingLimitConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+            max: POSTHOG_CODE_BILLING_LIMIT_MAX,
+            help: 'Code billing limits can be set from $0 to $3,000 per month.',
+            removalDisabledReason: "Code billing limits can't be removed. Set the limit to $0 instead.",
+            maxExceededError: "Code billing limits can't exceed $3,000 per month.",
+            currentAboveMaxNotice:
+                customLimitUsd !== null &&
+                customLimitUsd > POSTHOG_CODE_BILLING_LIMIT_MAX &&
+                billingLimitNextPeriod !== null &&
+                billingLimitNextPeriod <= POSTHOG_CODE_BILLING_LIMIT_MAX
+                    ? `Current usage is already above the Code billing limit cap, so this period stays at $${customLimitUsd.toLocaleString()}. The $${billingLimitNextPeriod.toLocaleString()} limit starts next period.`
```
abhischekt: we say usage but have the `customLimitUsd` interpolated in the message. Should it not be the usage value?

abhischekt (approved):
Reads sensibly to me - left 2 questions

frontend/src/scenes/billing/billingProductLogic.ts
pawel-cebula: Yep, intentional. `null` means “remove the limit”, so we skip the lower-than-current-usage warning for that path. For startup Code customers the remove action is disabled in the UI, and Billing rejects removal too, since removing the limit would mean unlimited usage.

frontend/src/scenes/billing/billingLimitConfig.ts
pawel-cebula: Yep, good call. Clarified that this is the current-period limit in PostHog/posthog@45b8ee4a4e3.

tatoalo (approved):

PR #71465, opened 2026-07-16
refactor(mcp-analytics): move Tool quality tab queries to gated runners

gesh:
> Follow-up to #71435 (now merged) — rebased onto master, standalone. The tab daily runner also carries the LIMIT 10000 cap discussed in the #71435 review.

## Problem

The Tool quality tab ran **four inline HogQL queries straight from the browser** (`tool_quality.sql` + three inline strings), while the tool *detail* page uses typed, flag-gated query runners. That inconsistency meant the tab's queries:

- weren't behind the `mcp-analytics` feature flag at the API layer (the detail runners gate specifically so the generic `/query/` endpoint can't bypass it),
- returned positional `unknown[][]` tuples the frontend hand-mapped by index,
- built category/tool filters as property-filter objects in the UI.

## Changes

Four new gated query runners mirroring the detail page's pattern:

| Runner | Replaces |
|---|---|
| `MCPToolQualityRowsQuery` | the `tool_quality.sql` per-tool table |
| `MCPToolQualityDailyStatsQuery` | the inline interval-bucketed daily series |
| `MCPToolCategoryCountsQuery` | the scope-share counts |
| `MCPToolCategoriesQuery` | the distinct-categories scope selector |

Each is gated by `validate_mcp_analytics_access`, returns typed rows, and takes `categories` / `toolName` as typed params bound server-side as `ast.Constant` (never interpolated). The frontend drops the raw SQL and `?raw` import for typed `api.query({ kind })` calls with the same result shapes — no UI change.

**Near-pure refactor:** the tab still scopes to the same population (by `$mcp_tool_name`, category-filtered). One deliberate behavior change: the runners build the window from `QueryDateRange` instead of the old `{filters}` placeholder, so a fixed custom range now truncates `date_from` to start-of-day and includes the whole `date_to` day (a null `date_to` caps at end-of-today) — matching the detail-page runners and insights generally. Relative ranges like `-7d` and the numbers they produce are unchanged for practical purposes; only fixed custom ranges shift by the boundary. The detail page's *effective-tool* scoping stays separate on purpose; unifying the two surfaces is a follow-up.

New `NodeKind`s were wired into every registry: the `get_query_runner` dispatch, the facade re-exports, **and** `kind_fallback_tags` (the exhaustive query-tagging match — missing this one throws `assert_never` at runtime when the query executes through `/query/`, caught during local testing).

Automated only — I (Claude) don't have a browser here.

- New ClickHouse-backed runner tests (`test_tool_quality_tables.py`, 6 cases): per-tool aggregation + error rate, category filter, hourly bucketing, tool scoping, category counts incl. uncategorized, distinct sorted categories.
- Tab logic test updated to assert the typed query kinds/params (incl. that a malicious tool name rides as a `toolName` param, never raw SQL).
- Verified `kind_fallback_tags` resolves all four new kinds to `Product.MCP_ANALYTICS` (the exact `assert_never` path that failed before the fix).
- Full `products/mcp_analytics/frontend` Jest suite (142) + backend `test_tool_quality_tables` / `test_tool_tables` (34) green; frontend typecheck clean; `ci:preflight` passing.

## 🤖 Agent context

Georgis flagged that the tab's `dailyStats` (and its siblings) should be backend logic, not inline HogQL in the UI, and greenlit a runner migration as its own PR. Decided against an MCP tool here — a query runner and an MCP tool are different layers; matching the detail page means runners, no tool (a tool-quality MCP tool would be a separate DRF-endpoint decision). Kept it a pure refactor so the diff is reviewable and the numbers don't move. Georgis caught the missing `kind_fallback_tags` arm by running the live app.


products/mcp_analytics/backend/hogql_queries/tool_quality_tables.py
```diff
@@ -0,0 +1,352 @@
...
+                count() AS total_calls,
+                {_IS_ERROR} AS errors,
+                round({_IS_ERROR} * 100.0 / count(), 1) AS error_rate_pct,
+                {_P50} AS p50_duration_ms,
+                {_P95} AS p95_duration_ms,
+                {_P99} AS p99_duration_ms,
+                uniq(distinct_id) AS users,
+                countDistinctIf(toString(properties.$session_id), toString(properties.$session_id) != '') AS sessions,
+                toString(min(timestamp)) AS first_seen,
+                toString(max(timestamp)) AS last_seen
```
lucasheriques: blocking: `toString(min(timestamp))` returns a naive project-timezone string, and `ToolQualityTable` feeds `last_seen` straight into `<TZLabel>`, which parses naive strings as browser-local time. So "Last seen" shifts by the timezone delta for anyone viewing from a different tz than the project (it can even show a future time).

The old inline query returned the raw DateTime and the API serialized it with the offset. `MCPToolTopUsersQueryRunner` does the same: raw `max(timestamp)`, offset preserved by `str(row[7] or "")`. Dropping the two `toString()` wrappers here matches it.

(The daily runner's `toString(dateTrunc(...))` is fine, `normalizeBucket` wants the naive wall clock.)

products/mcp_analytics/backend/hogql_queries/tool_quality_tables.py
```diff
@@ -0,0 +1,352 @@
...
+    date_range: QueryDateRange,
+    categories: list[str] | None,
+    *,
+    tool_name: str | None = None,
+) -> ast.Expr:
+    """WHERE for tool-name-bearing $mcp_tool_call events in the window, filtered by category/tool."""
+    exprs: list[ast.Expr] = [
+        parse_expr("event = {event}", placeholders={"event": ast.Constant(value=MCP_TOOL_CALL_EVENT)}),
+        parse_expr("timestamp >= {date_from}", placeholders={"date_from": date_range.date_from_as_hogql()}),
+        parse_expr("timestamp <= {date_to}", placeholders={"date_to": date_range.date_to_as_hogql()}),
```
lucasheriques, line 75: question: moving from the `{filters}` placeholder to `QueryDateRange` changes the window. `-7d` now truncates to start-of-day (was exactly now minus 168h), an explicit `date_to` now includes the whole end day (was `timestamp < midnight`), and a null `date_to` gets capped at end-of-today.

Numbers move for fixed custom ranges, so this isn't a pure refactor in the strict sense. I think the new semantics are the right call (they match the detail-page runners and insights generally), just confirming it's intentional and not something you'd want to preserve.

products/mcp_analytics/frontend/mcpAnalyticsToolQualityLogic.ts
```diff
@@ -7,41 +7,17 @@ import { dateFilterToText, getDefaultInterval } from 'lib/utils/dateFilters'
...
+import {
+    MCPToolCategoryCountItem,
+    MCPToolCategoryItem,
+    MCPToolQualityDailyStatItem,
+    MCPToolQualityRowItem,
+    NodeKind,
+} from '~/queries/schema/schema-general'
+import { IntervalType } from '~/types'
 
-import toolQualityQueryTemplate from '../backend/templates/tool_quality.sql?raw'
```
lucasheriques, line 15: question: `tool_quality.sql` lost its last consumer here, but the file survived. And `test_tool_quality_sql.py` still asserts its `__ORDER_BY__`/`__ORDER_DIRECTION__` placeholders, whose substitution logic this PR deletes. CI is now guarding a query nobody runs, and it will drift from `MCPToolQualityRowsQueryRunner.to_query()` silently.

Can we delete both files in this PR? Two comments still point at the template as source of truth and would need repointing too: `ToolQualityTable.tsx:41` and `mcpDashboardOverviewLogic.ts:89`.

products/mcp_analytics/backend/tests/test_tool_quality_tables.py
```diff
@@ -0,0 +1,147 @@
...
+    _create_event(
+        team=team,
+        event="$mcp_tool_call",
+        distinct_id=distinct_id,
+        timestamp=timestamp or datetime.now(tz=UTC),
+        properties=properties,
+    )
+
+
+class TestMCPToolQualityRowsQueryRunner(_MCPAnalyticsTeamScopedTestMixin, ClickhouseTestMixin, APIBaseTest):
```
lucasheriques, line 60: question: the point of the migration is putting these kinds behind `validate_mcp_analytics_access`, but no test exercises the gate. The suite calls `runner.calculate()` directly, so a runner that lost the override would stay reachable through `/query/` with everything green.

`test_harness_breakdown.py:211-220` has the pattern (flag on allows, flag patched off raises `UserAccessControlError`). One parameterized test over the four runners would pin it.

lucasheriques (approved):
Nice migration. The runners mirror the detail-page pattern faithfully, `categories`/`toolName` ride as `ast.Constant` so user input never touches query source, and catching the `kind_fallback_tags` gap saved a runtime blow-up.

Two things worth fixing before merge: a timezone bug on the Last seen column, and a date-boundary change the description says doesn't exist. Details inline. The rest can be follow-ups.

Approving so you're not blocked on a re-review once those land.

products/mcp_analytics/backend/hogql_queries/tool_quality_tables.py
gesh: Fixed — dropped both `toString()` wrappers so `first_seen`/`last_seen` return raw datetimes (offset preserved by `str(row[...])`), matching `MCPToolTopUsersQueryRunner`. TZLabel now gets offset-aware values.

products/mcp_analytics/backend/hogql_queries/tool_quality_tables.py
gesh: Intentional — the `QueryDateRange` boundaries match the detail runners and insights, and I want to keep them. You're right it isn't strictly identical for fixed custom ranges though; tightened the PR description to spell out the deliberate boundary shift instead of claiming "identical numbers".

products/mcp_analytics/frontend/mcpAnalyticsToolQualityLogic.ts
gesh: Done — deleted `tool_quality.sql` and `test_tool_quality_sql.py` (the `templates/` dir is gone with them), and repointed both comments (`ToolQualityTable.tsx:41`, `mcpDashboardOverviewLogic.ts:89`) at `MCPToolQualityRowsQueryRunner`.

products/mcp_analytics/backend/tests/test_tool_quality_tables.py
gesh: Added `TestMCPToolQualityGate` — parameterized over the four runners: flag on allows, flag patched off raises `UserAccessControlError`. Thanks for the `test_harness_breakdown.py` pointer.

products/mcp_analytics/backend/hogql_queries/tool_quality_tables.py
```diff
@@ -0,0 +1,352 @@
...
+    def _calculate(self) -> MCPToolQualityRowsQueryResponse:
+        with tags_context(
+            product=Product.MCP_ANALYTICS,
+            feature=Feature.QUERY,
+            team_id=self.team.id,
+            name="mcp_tool_quality_rows_query",
+        ):
+            response = execute_hogql_query(
+                query=self.to_query(),
+                team=self.team,
```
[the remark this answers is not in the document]
gesh: Good catch — fixed in 416a995. Threaded `user=self.user` through all four runners (rows, daily-stats, category-counts, categories), so property-level access control applies role/user-specific deny rules instead of only the default. Matches the core analytics runners (events, actors, retention, sessions).

PR #71468, opened 2026-07-16
fix(replay): keep score export queries below size limit

nicowaltz:
## Problem

The surfacing-score export sends the full AI-training opted-in team list as a ClickHouse query parameter in two places. Client-side substitution makes the rendered SQL grow with every opted-in team and can exceed ClickHouse `max_query_size` before the query runs.

Moving the list to a query-scoped external table also requires the HTTP ClickHouse adapter to forward `external_tables`. That adapter currently drops them.

```mermaid
flowchart LR
    A[Score export activity] -->|team IDs substituted twice| B[Rendered SQL]
    B --> C[ClickHouse parser]
    class A phBlue;
    class B phRed;
    class C phYellow;
```

## Changes

- Send opted-in team IDs as a query-scoped external table and filter with `GLOBAL IN` on the distributed replay table.
- Translate the existing clickhouse-driver external-table format to clickhouse-connect `ExternalData` for HTTP queries, including typed values accepted by the native driver.
- Add regression coverage for bounded SQL, external-table contents, typed serialization, and end-to-end HTTP forwarding.

```mermaid
flowchart LR
    A[Score export activity] -->|bounded SQL| B[HTTP ClickHouse client]
    A -->|external table data| B
    B --> C[ClickHouse]
    class A phBlue;
    class B phRed;
    class C phYellow;
```

- `hogli test posthog/temporal/tests/session_replay/surfacing_score_export_sweep/test_export.py posthog/clickhouse/client/test/test_connection.py::test_http_client_forwards_external_tables posthog/clickhouse/client/test/test_connection.py::test_http_client_serializes_typed_external_table_values` - 12 passed.
- `ty check` on all changed Python files.
- `hogli ci:preflight --strict --against origin/master`.

The new tests catch query growth from inlined team IDs, HTTP transport loss of query-scoped external data, and serialization failures for valid typed values.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change. This is an internal query transport fix with no user-facing API or configuration change.

## 🤖 Agent context

Codex investigated the failed export, implemented and tested the fix, and used the repository `/writing-tests` skill. The implementation keeps the existing external-table contract and adds the missing HTTP conversion so both native and HTTP ClickHouse transports preserve bounded query text.

### DIFF
--- a/posthog/clickhouse/client/connection.py
+++ b/posthog/clickhouse/client/connection.py
@@ -1,4 +1,5 @@
 import os
+import json
 import logging
 from collections.abc import Mapping
 from contextlib import contextmanager
@@ -13,6 +14,9 @@ from clickhouse_pool import ChPool
 
 if TYPE_CHECKING:
     from clickhouse_connect.driver import Client as HttpClient
+    from clickhouse_connect.driver.external import ExternalData
+
+    from posthog.clickhouse.client.execute import ClickHouseExternalTable
 
 from posthog.clickhouse.workload import Workload
 from posthog.settings import data_stores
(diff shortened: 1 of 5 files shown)


posthog/clickhouse/client/connection.py
```diff
@@ -132,6 +140,47 @@ def get_clickhouse_creds(user: ClickHouseUser) -> tuple[str, str]:
     return __user_dict[user] if user in __user_dict else __user_dict[ClickHouseUser.DEFAULT]
 
 
+def _to_clickhouse_connect_external_data(
```
robbie-c, line 143: nit: this function name reads like it was written by Yoda

PR #71477, opened 2026-07-16
feat(data-warehouse): point source authoring at the shared REST framework + add missing helpers

Gilbert09:
## Problem

We have ~432 implemented warehouse REST sources and a lot more in flight. A survey found heavy code duplication, but the root cause is **adoption, not a missing abstraction**: a full shared REST framework already exists in `common/rest_source/` (paginators, auth classes, a tracked+retrying `RESTClient`, declarative endpoints, fan-out), and `make_tracked_session()` already retries 429/5xx. Only ~31 of ~432 sources use it.

The reason the rest hand-roll everything: the `implementing-warehouse-sources` skill and the source-generation prompt both point new sources at `klaviyo`, which hand-rolls its client because it has genuine edge complexity (fan-out + resume, non-percent-encoded query strings, JSON:API flattening). Every generated source copied that boilerplate — including a redundant second retry layer on top of the transport's built-in retries. The skill also had factual drift and omitted patterns every real source uses.

## Changes

**Skill + template (root cause):**
- Make `chargebee` the canonical reference (declarative `rest_source` + framework auth + resume); demote `klaviyo`/`github` to a bespoke-transport fallback. New "Prefer the shared REST framework" section listing what the framework gives for free and when hand-rolling is justified.
- Fix the retry guidance: the tracked transport and `RESTClient` already retry 429/5xx honoring `Retry-After` — don't stack a second `tenacity` layer (the most-copied mistake).
- Fix drift: `schema-general.ts` is PascalCase not kebab-case; icon convention is PNG not SVG; correct the `_is_host_safe` cross-reference; reconcile the `Batcher` rule with reality (a smaller-footprint batcher is a valid choice); document `redact_values`, the base-class capability flags (`supports_column_selection` / `supports_row_filters` / `has_managed_hogql_schema`), and the API-versioning skill. Refresh stale `source.template` TODOs.

**New shared helpers** (for boilerplate with no shared home, copy-pasted across ~400 sources; `rest_source` already covers fetch/retry/paginators/auth so nothing is added there):
- `build_endpoint_schemas()` — the `get_schemas` ENDPOINTS loop + `names` filter, with `append_only`/`merge_only`/metadata overrides.
- `incremental_field()` — one `IncrementalField` builder that keeps `type`/`field_type` in sync (a common copy-paste bug).
- `auth_non_retryable_errors()` / `AUTH_401_403_ERRORS` — the 401/403 non-retryable map, host-scoped or generic.
- `validate_via_probe()` — the "GET probe -> status -> bool" credential check.

Adopted `build_endpoint_schemas` in `chargebee` as a worked example. Retrofitting the rest of the existing sources is a separate, test-gated, batched follow-up.

> [!NOTE]
> The batch source-generation prompt (`build_prompts.py`) was also updated to prefer the framework and drop the redundant-retry instruction. That file lives outside this repo, so it's not in this diff.

- New unit tests `common/test/test_source_helpers.py` (18 cases): `build_endpoint_schemas` output matches the hand-written loop, `names`/`append_only`/`merge_only`/metadata behavior, `incremental_field` type/field_type sync, `auth_non_retryable_errors` key format (must match the real `requests` error text or non-retryable matching silently breaks), `validate_via_probe` status/exception mapping.
- `chargebee/` suite (20 tests) passes with the helper adopted — confirms identical `get_schemas` output.
- `hogli ci:preflight --fix` clean (ruff lint/format, markdown).

## Automatic notifications

- [ ] Publish to changelog?


products/warehouse_sources/backend/temporal/data_imports/sources/common/schema.py
```diff
@@ -100,3 +101,65 @@ def build_default_schemas(source_schemas: list[SourceSchema]) -> list[dict]:
...
+    - ``descriptions`` / ``should_sync_default`` / ``supports_webhooks``: per-endpoint metadata.
+
+    ``names`` (the schema-picker filter) keeps only the requested endpoints when set.
+    """
+    descriptions = descriptions or {}
+    should_sync_default = should_sync_default or {}
+    schemas = []
+    for name in endpoints:
+        fields = incremental_fields.get(name) or []
+        has_incremental = bool(fields)
```
estefaniarabadan: (reply from PHC) An endpoint mapped to an explicit `[]` here becomes non-incremental (`bool([]) == False`), whereas the old per-source loop used `.get(endpoint) is not None`, so an explicit `[]` counted as incremental.
Harmless today (no source maps to `[]`), but it is a subtle behavior change from the loop this replaces. Worth a short note, or matching the old semantics so the "exact reproduction" claim holds.

estefaniarabadan (approved):

products/warehouse_sources/backend/temporal/data_imports/sources/common/schema.py
estefaniarabadan: this is a nit that can be ignored, testint to add comments from PHC (it's getting better)

Gilbert09: Good catch — matched the old loop's `.get(name) is not None` semantics exactly so a present-but-empty `[]` still counts as incremental, keeping the "exact reproduction" claim honest. Fixed in 7c09353.

PR #71481, opened 2026-07-16
feat(data-warehouse): close rest_source gaps (item-transform + fan-out resume)

Gilbert09:
## Problem

A full shared REST framework (`common/rest_source/`) exists, but two real capability gaps forced ~complex sources to hand-roll their client instead:

1. **No per-item reshaping.** `data_selector` extracts a list but can't reshape each row (e.g. flatten a JSON:API `attributes` block into the row root), so JSON:API-shaped sources hand-rolled.
2. **Fan-out + resume were mutually exclusive.** A dependent (parent→child) resource raised `NotImplementedError` if resume was requested, so any source that both fans out and needs resumable backfill (e.g. list→profile fan-out) couldn't use the framework.

## Changes

- **Declarative `data_map`.** New optional `data_map: Callable[[dict], dict]` on the endpoint config, wired to `Resource.add_map` (already the transform mechanism) — applied after `data_selector`, before type coercion. 1:1 reshape; use `data_selector` to extract first.
- **Fan-out + resume.** Resume is now routed to the dependent (child) resource: each parent's child pagination is checkpointed under that parent's resolved child path. A restart skips parents already fully synced and resumes the one in progress. The parent list re-paginates each run (it's usually small) and completed parents are skipped by path. State: `{"completed": [...], "current": path|None, "child_state": {...}|None}`. When a config has a dependent resource, the non-dependent parent no longer consumes the resume hook.

`paginate()` already deep-copies the paginator per call, so each parent's child pagination starts clean — which is what makes per-parent resume correct.

- New tests: `test_fanout.py` (3 cases — first-run checkpoints completed parents; resume skips completed and resumes the in-progress parent's child cursor; no-hook path unchanged) and `test_data_map.py` (declarative reshape yields flattened rows through `rest_api_resource`).
- Additive + backward compatible: all **114 existing `rest_source` tests** pass, plus the **chargebee / zendesk / attio** consumer suites (66 tests) — chargebee exercises single-resource resume, unaffected.
- `hogli ci:preflight --fix` clean.

> [!NOTE]
> Converting live connectors (e.g. klaviyo's list fan-out, JSON:API sources) onto these hooks is a separate, per-source, test-gated follow-up — a live-connector rewrite belongs in its own PR guarded by that source's tests, not bundled with the framework change. The skill's "Prefer the shared REST framework" section (PR #71477) will be extended to document these two hooks once it lands.

## Automatic notifications

- [ ] Publish to changelog?


estefaniarabadan (approved):
looks good, not sure why the agents do not read the directive for not creating long comments :_(

PR #71483, opened 2026-07-16
fix(aio): keep low-confidence sentiment in neutral

carlos-marchal-ph:
## Problem

The AI observability sentiment eval (which classifies user messages to Max) misclassified two kinds of generation, confirmed against production eval runs:

- **Terse, task-focused queries flagged negative.** Short commands like "retention graph for these people" split near-evenly between negative and neutral, and the argmax promoted those coin-flips to `negative`. A meaningful share of negatives sit within a hair of neutral.
- **No-user-message generations counted as neutral.** Generations with no user turn defaulted to a `neutral` label, polluting the neutral bucket.

Context in the originating Slack thread.

## Changes

- **Neutral-margin gate** (`select_sentiment_label`): a polar label (`negative`/`positive`) must beat `neutral` by more than `SENTIMENT_NEUTRAL_MARGIN` (0.15); otherwise the message resolves to `neutral`. Applied consistently across the model's per-message parse, the per-generation aggregate, and the read-side trace aggregation (`_aggregate_trace_sentiment`), so a low-confidence polar score never surfaces as polar on either the write or the read path. The threshold is a single tunable constant and should be recalibrated against a labeled sample.
- **Exclude `message_count = 0` runs** from the two sentiment read queries so generations with no user message stop landing in the neutral distribution. The filter keys off the latest evaluation via `argMax`, so a later empty re-evaluation can't be bypassed.

The label-selection helper now lives in `posthog/hogql_queries/ai/sentiment_labeling.py`, so the write path (temporal eval) and the read path (query runners) share one definition without the read path pulling in the AI-observability workflow graph.

Deliberately out of scope (flagged for follow-up): swapping the sentiment model, and the per-generation re-classification that re-scores the same user turn on every generation in a trace (an architectural change to *when* evals fire).

Per review, we deliberately do **not** persist the raw classified message text: it would sit in the long-lived events table past the AI-event retention window, and the read path never surfaced it anyway.

Automated tests (run in CI):

- `sentiment/tests/test_utils.py`: parameterized cases for `select_sentiment_label` (coin-flip splits resolve to neutral, confident labels kept, exact-margin ties stay neutral) plus a `build_generation_result` case asserting the gate is applied. Catches regressions where the margin gate is removed or weakened.
- `sentiment/tests/test_model.py`: a near-tie that argmaxes to negative comes back `neutral` through `classify()`, guarding the gate's wiring into the model's result parsing.
- `hogql_queries/ai/test/test_sentiment_evaluations.py`: a low-confidence polar aggregate resolves to neutral through `load_trace_sentiment_evaluations`, guarding the gate on the read path.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Investigated from a Slack thread: built an empirical taxonomy of the eval's error runs against production data, then read the source to locate the root causes. The terse-query misclassification is the sentiment model resolving no-sentiment text into a razor-thin negative, not a message-role problem (tool results are already filtered upstream). Chose a single symmetric margin gate over per-pattern heuristics to avoid over-fitting to edge cases. Review feedback applied: reuse the `message_count` alias in the read queries, apply the gate on the read-side aggregation too, and drop the persisted message text on retention grounds.

Skills invoked: `/exploring-llm-evaluations`, `/writing-tests`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B8ZE81ZT7/p1784171902751399?thread_ts=1784171902.751399&cid=C0B8ZE81ZT7)*


posthog/temporal/ai_observability/sentiment/utils.py
```diff
@@ -44,12 +62,13 @@ def build_generation_result(
             "label": result.label,
             "score": result.score,
             "scores": result.scores,
+            "text": item.text,
```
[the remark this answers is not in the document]
Radu-Raicea: Yep, we shouldn't put it

posthog/temporal/ai_observability/sentiment/utils.py
```diff
@@ -2,9 +2,27 @@
 
 from typing import Any
 
+from posthog.temporal.ai_observability.sentiment.constants import SENTIMENT_NEUTRAL_MARGIN
 from posthog.temporal.ai_observability.sentiment.schema import PendingClassification
 
 
+def select_sentiment_label(scores: dict[str, float], neutral_margin: float = SENTIMENT_NEUTRAL_MARGIN) -> str:
```
Radu-Raicea: We should probably do this in `_aggregate_trace_sentiment` too from `posthog/hogql_queries/ai/sentiment_evaluations.py`

posthog/hogql_queries/ai/sentiment_evaluations.py
```diff
@@ -79,6 +79,7 @@ def get_sentiment_for_generation(
   AND length(toString(properties.$ai_target_event_id)) > 0
   AND {generation_filter}
 GROUP BY trace_id, toString(properties.$ai_target_event_id)
+HAVING argMax(toString(properties.$ai_sentiment_message_count), timestamp) != '0'
```
Radu-Raicea: Could we use the already made alias at the top (`message_count`)?

posthog/hogql_queries/ai/sentiment_evaluations.py
```diff
@@ -101,6 +102,7 @@ def get_sentiment_for_generation(
       AND trace_id IN {trace_ids}
       AND length(toString(properties.$ai_target_event_id)) > 0
     GROUP BY trace_id, toString(properties.$ai_target_event_id)
+    HAVING argMax(toString(properties.$ai_sentiment_message_count), timestamp) != '0'
```
Radu-Raicea: Could we use the already made alias at the top (`message_count`)?

Radu-Raicea (approved):
Approving, but the comments should be tackled!

carlos-marchal-ph:
I distinctly remember telling the bot NOT to store anything on the eval LMAO don't know how it sneaked that past me should have reviewed better 🙏

PR #71484, opened 2026-07-16
feat(tracing): report usage events and harden operation detail scene

jonmcwest:
## Problem

The operation detail scene shipped in #67757, but its backend endpoints (duration-histogram, aggregate, per-trace) fire no `report_user_action`, so we can't tell whether anyone uses the scene — or any of these endpoints — across web, API, and MCP callers. A review pass over the scene also surfaced a handful of state-sync bugs and duplicated helpers.

## Changes

Instrumentation: the duration-histogram, aggregate, and per-trace endpoints now fire `tracing duration histogram queried`, `tracing aggregation queried`, and `tracing trace fetched`, matching the existing `tracing query executed` pattern so usage segments by `access_method` (web, personal API key, MCP). The now-unreferenced timeseries request serializer is dropped.

Hardening from the review pass:

- The date range carries through the operation URL, so detail stats match the row clicked on the Operations tab, and deep links restore it.
- A persisted duration selection that falls entirely off a reshaped histogram axis renders no highlight (previously it highlighted the whole chart).
- Links missing the service param get the missing-params fallback instead of silently querying `serviceNames: ['']`.
- A failed samples refetch reloads the waterfall for the retained samples instead of leaving it permanently blank.
- Superseded requests are aborted server-side via per-loader `AbortController`s (matching `tracingDataLogic`), adjacent samples sharing a trace reuse the loaded waterfall, and the samples query passes `excludeAttributes` since the scene reads only scalar span fields.
- The trace-list duration histogram counts the population the active view shows: root spans in Traces view, every matching span in Spans view.
- Shared helpers replace copies: category duration axis config, error-rate formatting, and the duration-range type.

Automated, run locally against current master:

- `products/tracing/backend/tests/test_duration_histogram.py` (3 tests) — endpoint bucketing incl. the `rootSpans=false` path.
- 56 frontend tests across `durationBuckets.test.ts`, `tracingOperationSceneLogic.test.ts`, and `tracingDataLogic.test.ts` — includes the drag-selection refetch/pager-reset and sample-paging waterfall-anchor regressions.
- `ruff` and `oxlint`/`oxfmt` clean on changed files.

I have not manually re-verified against the live app since rebasing onto current master; the earlier iteration of these changes was manually verified on a seeded local stack.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with PostHog Code (Claude). Skills invoked: `/improving-drf-endpoints`, `/writing-tests`, `/using-kea-disposables`, plus simplify and multi-angle code-review passes whose confirmed findings are the hardening list above.

This PR replaces #71103, which was accidentally opened from a branch cut from a stale master and re-presented the whole (already merged) feature as new. This branch cherry-picks only the genuinely new work onto current master. Decisions worth knowing: instrumentation lives in the backend viewset rather than frontend `posthog.capture` so usage is visible across all access methods, and the histogram endpoint keeps `rootSpans=true` as its default so existing callers keep trace-level distributions. Deliberately skipped as out of scope: single-bucket drag selection (needs a shared `Sparkline` change), migrating the handwritten `api.tracing` client block to generated types, and deduplicating the repeated `report_user_action`/filter-group boilerplate in the tracing viewset.

---

### DIFF
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -2901,6 +2901,8 @@ const api = {
                 // false (default) groups by trace_id and returns root spans; true returns every
                 // matching span (root and child) flat. See products/tracing/backend logic.py.
                 flatSpans?: boolean
+                // true omits the per-span attribute maps — use when only scalar span fields are read.
+                excludeAttributes?: boolean
             },
             signal?: AbortSignal
         ): Promise<{
@@ -2918,12 +2920,14 @@ const api = {
                 statusCodes?: number[]
                 filterGroup?: PropertyGroupFilter
                 offset?: number
-            }
+            },
+            signal?: AbortSignal
         ): Promise<{ results: Record<string, any>[]; hasMore: boolean; nextOffset?: number | null }> {
             return new ApiRequest()
                 .tracingSpans()
                 .withAction(`trace/${traceId}`)
                 .create({
+                    signal,
                     data: {
                         ...query,
                         dateRange: query?.dateRange ?? { date_from: '-24h' },
--- a/frontend/src/products.tsx
+++ b/frontend/src/products.tsx
@@ -1258,8 +1258,19 @@ export const productUrls = {
     slackTaskContext: (): string => '/slack-task-context',
     toolbarLaunch: (): string => '/toolbar',
     tracing: (): string => '/tracing',
-    tracingOperation: (serviceName: string, spanName: string): string =>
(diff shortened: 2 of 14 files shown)


jonmcwest:
Split into focused PRs: #71492 (backend usage events), #71494 (histogram population fix), #71495 (dedupe refactor), #71496 (date range through the operation URL), #71497 (request lifecycle hardening), #71498 (off-axis highlight fix).

PR #70424, opened 2026-07-13
feat(mcp): add posthog and project skill discovery

skoob13:
## Problem

PostHog publishes more than 100 bundled skills, while teams also keep their own workflows in the current project's Skills store. Embedding either corpus in the `exec` schema would exceed Claude's per-tool character limit, and returning a long skill or reference can exceed the MCP client's output limit.

## Changes

- Keep the compact tool-domain index and fixed `learn` grammar inline so tool discovery remains available within the 17,500-character schema budget.
- Discover both sources with `learn skills` and `learn -s <query>`.
- Qualify bundled skills as `posthog:<skill>` and current-project skills as `project:<skill>`.
- Gate PostHog and project skill discovery with the `mcp-exec-skills` feature flag for every cli-mode client.
- Keep Claude web and desktop's built-in `learn analytics`, `learn visualizations`, and `learn feedback` guides available independently of the skills flag.
- Give Claude Code and other exec clients only the skill commands when the flag is enabled. Plugin consumers continue receiving no `learn` command because they supply skills directly.
- Load the published PostHog skill archive through Redis with stale-while-revalidate behavior.
- Read project skills through a request-scoped API adapter with `llm_skill:read`; project content is not cached in the MCP service.
- Add bounded project full-text search across names, descriptions, bodies, file paths, and Markdown references. Results are limited to 10 skills, two snippets per skill, and a five-second database timeout.
- Preserve scoped `-s` search, `--lines` reads, file manifests, and heading outlines for references that exceed the 44,000-character response budget.
- Expose only latest, active, uncategorized project skills. Category-specific skills remain on their existing surfaces.
- Regenerate the frontend and MCP API clients and document the merged discovery flow.

Before:

```mermaid
flowchart LR
    Exec[exec learn] --> Guides[Bundled guides]
    classDef phYellow fill:#f9bd2b,stroke:#f9bd2b,color:#000;
    classDef phGray fill:#e5e7eb,stroke:#c7ccd1,color:#000;
    class Exec phYellow;
    class Guides phGray;
```

After:

```mermaid
flowchart LR
    Exec[exec learn] --> Guides[Claude built-in guides]
    Flag[mcp-exec-skills] --> PostHog[PostHog catalog]
    Flag --> Project[Project adapter]
    Exec --> PostHog
    PostHog --> Redis[Redis SWR cache]
    Redis --> Release[Published skills.zip]
    Exec --> Project
    Project --> API[Current-project Skills API]
[...]

PR #71491, opened 2026-07-16
feat(mcp): improve exec learn skill triggering and efficiency

skoob13:
## Problem

Stacked on #70424. Reviewing how agents discover and read skills through `exec learn` against the Claude Code harness's native skill mechanics surfaced triggering and efficiency gaps: an agent browsing `learn skills` sees bare names with no way to learn what a skill does short of reading its whole body, project results always sort below PostHog results regardless of relevance, exact-substring matching misses simple word-form variations (analyzing vs analysis), zero-hit searches dead-end, search snippets leak YAML frontmatter, every read costs a full round-trip, repeated project-skill reads re-fetch the same API resources within one request, and the skill archive re-downloads up to 32 MB on every soft-TTL refresh even when unchanged.

## Changes

Triggering:

- `learn -d <source>:<skill> [...]` describes up to 20 skills per call (one line each) without reading bodies, so the names-only listing stays cheap while descriptions remain reachable.
- `learn -s` merges PostHog and project results into one relevance order. Project matches are scored client-side with the same weights the PostHog ranker uses (name 1000, description 300, plus per-`matched_field` tier bonuses), so an exact project match outranks a weak PostHog content match.
- Query tokens now match light stem variants (suffix stripping with a 5-char minimum), so "analyzing retention" finds "retention analysis" while "sessions" still doesn't false-match "assessing".
- Zero-hit searches explain that no query word matched, suggest broader keywords, and inline the available names on small catalogs (or counts plus `learn skills` on large ones). Scoped file searches that find nothing point at the read and `--lines` recovery commands.
- SKILL.md frontmatter is excluded from snippet extraction and content scoring, so snippets stop showing the redundant `description:` line and descriptions stop double-counting.

Efficiency:

- Batch reads: `learn <source>:<skill> <path> [path...]` and `learn <source>:<skill> [<source>:<skill>...]` read up to 10 targets in one call, under the same 44k output cap.
- `ProjectSkillCatalog` memoizes skill, file, and listing fetches per request (promise-memoized with eviction on rejection), so batches and repeated reads dedupe to one API call each.
- `SkillArchiveCache` stores the response ETag and revalidates with `If-None-Match`; a 304 bumps freshness and re-extends TTLs without downloading or re-parsing the zip.

The eval scorers in `skill_distribution_scorers.py` recognize the new command forms (`-d` as a skill command, all-qualified batch reads as loads) so `eval_skill_distribution` keeps measuring correctly.

- `services/mcp`: full unit suite (2,162 passed) and hono suite (305 passed), including new cases for stemming (with a "sessions" ≁ "assessing" over-match guard), frontmatter-free snippets with preserved raw line numbers, relevance-merge ordering (a name-matching project skill outranking a content-only PostHog match), batch describe (happy path, unknown names, cap), both batch-read forms, flag-with-multiple-paths rejection, both zero-hit recovery shapes, project-catalog memoization (dedupe, concurrent dedupe, retry-after-error, single list fetch), and archive 304 revalidation (conditional header sent, cached bytes reused, new-etag replacement, no-validator full download). Each group guards a behavior nothing previously covered because the behaviors are new.
- Instruction snapshots regenerated; the claude.ai 16,384-char inputSchema budget test passes with the extended syntax line (+16 chars).
- `pytest test_skill_distribution_scorers.py` (35 passed) with new parameterized rows: batch reads count as skill loads, `-d` counts as a skill command but not a load.
- `ruff check`/`format --check` clean; typecheck clean for all touched files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Grammar prose updated in `services/mcp/README.md` and the handbook's implementing-mcp-tools page in this PR.

## 🤖 Agent context

Planned and verified by Claude Code (Fable 5) in a live session with skoob13 directing; implementation was delegated to three parallel Opus subagents (skill-catalog ranking, project-catalog memoization, archive ETag) plus one for the exec-learn grammar phase, each on disjoint files. A design-review pass before implementation corrected two plan flaws: the original batch-read disambiguation rule broke on `--lines 40:120` and on paths containing colons (replaced with a strict `^(posthog|project):` all-or-nothing rule), and the original stemming spec both under-matched its own motivating example and over-matched "sessions" onto "assessing" (replaced with longest-suffix-first variants and a 5-char minimum). Backend rank exposure for the relevance merge was considered and rejected to avoid serializer/OpenAPI churn since client-side scoring was validated numerically. Skills invoked: /writing-tests. The orchestrating session re-ran every suite independently before committing.

https://claude.ai/code/session_01U2LTg3j8bRQQzrmukgUdk1


skoob13:
Responses to the review findings — fixes landed in 6602c17cc94 (`fix(mcp): address review findings on skill cache, search, and gate`).

**Greptile: Payload and ETag can diverge (`SharedBlobCache.ts`)** — valid, with a narrower failure mode than described. A reader racing a writer only sees a transient in-memory mismatch (the refresh lock keeps it from acting on it), but a partial failure of the three parallel `SET`s could durably persist the new ETag over the old bytes — after which every conditional refresh 304s and `touchCache()` keeps re-marking stale bytes fresh. Fixed by landing the bytes `SET` before the validator/freshness writes, so both partial-failure orders self-heal (old+old is consistent; new bytes + old etag forces a full 200 re-download on the next refresh). Covered by a test asserting a failing bytes write leaves the old etag untouched.

**Greptile: Fallback drops content-only matches (`project-skill-catalog.ts`)** — valid. The backend matches the whole query via `icontains` over name/description/body/files, so the listing-only fallback lost body/file matches that PostHog skills (searched locally over full content) would surface, biasing the merged ranking. Fixed: on zero whole-query hits, the client retries the backend with up to 3 informative single tokens (same tokenizer and length floor the scorer uses, longest first, `Promise.allSettled` so one failure can't sink the rest), merges hits per skill, and scores them against the original query; the listing rank remains the last resort.

**ReviewHog: `-d` describe misreports real skills past the 200-skill listing cap** — valid. `descriptions()` was replaced with `describe(names)`: it seeds from the memoized listing and, only when the listing is truncated, resolves the missing names through the uncapped, memoized exact-name endpoint (bounded by the caller's 20-name cap). A miss on a complete listing stays authoritative with no extra fetches, and describe/read can no longer contradict each other.

**ReviewHog: `isSkillLoad()` naive whitespace split** — valid. It now reuses the dispatcher's quote-aware `tokenizeLearnInput` and the shared `QUALIFIED_IDENTIFIER` regex (tokenizer errors return false), closing both divergences: a quoted flag no longer opens the gate, and a quoted qualified identifier now does. Both cases are covered by gate tests.

**ReviewHog: batch reads over-report `skill invoked` on partial failure** — valid. Batch paths now fetch all content first and report invocations only after the whole `Promise.all` resolves, keeping the all-or-nothing output contract; a failing sibling means zero events. Tested for both multi-file and multi-skill batches.

**ReviewHog: file-only reads never count as `skill invoked`** — valid; the "follow-ups within an already-loaded skill" assumption is indeed falsified by the batch-file grammar this PR adds. `trackSkillInvoked` now captures every read kind with an additive `skill_read_kind` property, deduped per skill identifier per request via the per-request catalog closure — a 5-file batch of one skill counts once, separate commands still each count.

**ReviewHog: skills gate adds a Redis round-trip to every `call`** — valid observation, deliberately not changed. It is a constant O(1) GET that fails open and is dominated by the inner tool's PostHog API round-trip on the same path; folding it into the session-context batch couples the resolver to gate state for a negligible win. Happy to revisit if the gate ever grows more session lookups.

PR #71493, opened 2026-07-16
fix(tasks): stop escalating already-resolved permission responses

VojtechBartos:
## Problem

Slack `@PostHog` runs in `full_auto` were spamming threads with approval cards and "Approval recorded / No further action is needed" replies, even though the user never should have been asked.

Root cause: a `permission_response` sent to the sandbox agent can be delivered twice (the Modal tunnel retries the request on a stale upstream connection). The first delivery resolves the request. The second gets back the agent-server's `No pending permission request found for id: X` JSON-RPC error. The broker read that as a failed send and fell through to posting a Slack approval card for a request it had already auto-approved. Each card was dead on arrival, and every click swapped it for the "Approval recorded" notice, hence the flood.

Origin thread (behind PostHog auth): https://posthog.slack.com/archives/C09SK2PAGKF/p1784197051162259

## Changes

Treat the "already resolved" error as a delivered response on both paths:

- **Auto-responder** (`permission_broker.try_auto_respond_permission_request`): dedupes and returns handled instead of escalating to a human card.
- **Human-response activity** (`send_permission_response_to_sandbox`): records the decision without retrying or posting a delivery-failure notice.

Detection lives in one place, `is_permission_already_resolved_error()` in `agent_command.py`, mirroring the existing `NO_ACTIVE_SESSION_ERROR` handling.

> [!NOTE]
> This stops the Slack symptom but not the underlying duplicate delivery. The source-level fix (making the agent-server's `permission_response` idempotent so a duplicate returns success) belongs in the `posthog/code` repo and is a planned follow-up.

Automated only (no manual Slack run). New regression tests, both run locally and passing:

- `test_permission_broker.py::test_already_resolved_request_is_answered_not_escalated` — the "already resolved" error is answered and deduped, never escalated to a human prompt. No existing test covered a non-success send that should still count as handled.
- `test_send_permission_response_to_sandbox.py::test_already_resolved_records_state_without_raising` — the human-response activity records state and returns instead of raising (which would retry the activity and post a delivery-failure notice).

Ran the two affected suites (49 tests) plus the relay permission tests, all green. Did not run mypy locally per repo convention (CI covers it).

## 🤖 Agent context

I (Vojta) directed this; Claude did the investigation and the change. Diagnosis started from the Slack thread and used the PostHog MCP (`tasks-runs-retrieve`, `query-logs`, `tasks-runs-session-logs-retrieve`) to correlate the sandbox session log with the tasks-agent worker logs. That correlation is what pinned it to duplicate `permission_response` delivery rather than a bad permission mode: the run state was a clean `full_auto`, and every escalated request had a matching `permission_broker_auto_allow_failed` carrying the "No pending permission request found" error a few ms before the card was posted.

Considered fixing it in the agent-server (idempotent responses) but kept this PR to the Python side to stop the user-visible spam quickly; the agent-server hardening is noted above as a follow-up. No repo skills were invoked.

### DIFF
--- a/products/tasks/backend/logic/services/agent_command.py
+++ b/products/tasks/backend/logic/services/agent_command.py
@@ -40,6 +40,18 @@ BLOCKED_IP_RANGES = [
 
 NO_ACTIVE_SESSION_ERROR = "No active session for this run"
 
+# The agent-server raises this (as a JSON-RPC error) when a permission_response
+# targets a request that is no longer pending — it was already resolved, most
+# often by a duplicate delivery of the same response through the Modal tunnel.
+# The message carries the request id as a suffix, so match on the stable prefix.
+# A response that gets this back has effectively landed: callers should treat it
+# as delivered rather than retry it or escalate the request to a human.
+NO_PENDING_PERMISSION_REQUEST_ERROR = "No pending permission request found"
+
+
+def is_permission_already_resolved_error(error: str | None) -> bool:
+    return error is not None and NO_PENDING_PERMISSION_REQUEST_ERROR in error
+
 
 @dataclass
 class CommandResult:
--- a/products/tasks/backend/logic/services/permission_broker.py
+++ b/products/tasks/backend/logic/services/permission_broker.py
@@ -27,7 +27,11 @@ from django.core.cache import cache
 
 import structlog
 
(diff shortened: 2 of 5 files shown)


cvolzer3 (approved):

VojtechBartos:
closing in favour of https://github.com/PostHog/posthog/pull/71616

PR #71501, opened 2026-07-16
feat(capture): pass snapshot_host as a kafka header

robbie-c:
## Problem

The ml-mirror anonymizer ([#70528](https://github.com/PostHog/posthog/pull/70528)) gates its hostname classification on the recorded page's hostname, which the browser SDK stamps on `$snapshot` events ([posthog-js#4143](https://github.com/PostHog/posthog-js/pull/4143)). Earlier iterations forwarded the stamp inside the serialized message body, which forced the anonymizer to scan the payload for it and made JSON key ordering a cross-service contract - review feedback (rightly) called that too delicate to rely on.

## Changes

- The stamp rides as an optional `snapshot_host` Kafka header - the existing mechanism (`CapturedEventHeaders`) for fields consumers read without parsing the payload. `RecordingProperties` deserializes `$snapshot_host` from the incoming event, string values ride `CapturedEvent.snapshot_host` (marked `skip_serializing`, so the wire body is unchanged) into `to_headers()`, and non-string stamps are dropped. The serialized message body does not carry the property at all.
- `serialize_snapshot_data_sync` builds derived `Serialize` structs instead of a `json!` map. This is a substantial perf win on capture's hot path: the `json!` macro deep-copies the entire `snapshot_items` array into an intermediate `Value` tree before serializing, while the structs serialize borrowed references straight to the output string. Release-mode benchmark, byte-identical output, medians of 10 runs:

| Payload | `json!` map (old) | structs (new) |
|---|---|---|
| 100 KB, 354 events | 0.92 ms | 0.12 ms |
| 1 MB, 3.6k events | 11.5 ms | 2.2 ms |
| 5 MB, 18k events | 31.8 ms | 7.6 ms |

This function is the piece capture deliberately runs synchronously on the async runtime because sync serialization benchmarked faster than `spawn_blocking` under high concurrency (per the comment on it), so a ~4x reduction in its CPU time is a direct win on request latency under load.

Version skew is fail-safe in both directions: a missing header (old capture, old SDK, mobile) lands on the anonymizer's collapse-every-hostname fallback, and old consumers ignore unknown headers.

- `test_snapshot_host_passes_through_as_a_kafka_header`: runs the full `process_replay_events` path and asserts a string stamp lands in `to_headers()`, non-string/absent stamps don't, and the serialized body never contains the property. This is the regression a multi-agent review of #70528 originally found (capture silently dropping the stamp while both neighbors' suites pass), now pinned at the header.
- Extended the existing `CapturedEventHeaders` → `OwnedHeaders` roundtrip test with a `snapshot_host` value.
- `cargo test -p capture --lib`: 905 tests pass; `cargo test -p common-types` passes.
- The benchmark ran as a temporary `#[ignore]` test (old implementation copied inline, identical inputs, output length asserted equal); it is not committed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal message-shape change, no user-facing docs affected.

## 🤖 Agent context

Authored with PostHog Code (Claude), companion to [#70528](https://github.com/PostHog/posthog/pull/70528). Robbie (AI research, which owns the ml-mirror pipeline) directed the work. The design went through three shapes on review feedback: body property relying on serde_json's alphabetical key order (accidental contract), explicit struct field order with a byte-order test (still an ordering contract), and finally this Kafka header per @pl's review - which also let the anonymizer delete its payload scan entirely. The struct serializer survives from round two purely for the perf win.

---


pl:
I don't mind the optimization, but we should never rely on ordering in JSON, unless we put it in arrays. Someone's gonna forget it one day, because it's a very delicate and uncommon contract, and will break the implementation.

I would suggest adding a Kafka header for that specific `$snapshot_host` field. We do that for several fields already, so that we can defer parsing the message.

robbie-c:
@pl nice, ok I didn't realise that! Will do that instead.

PR #71509, opened 2026-07-16
feat(prompts): Fetch prompts by label

jurajmajerik:
## Problem

https://github.com/PostHog/posthog/pull/70862 added prompt labels (movable pointers to versions) and the endpoints to manage them, but there is no way to fetch a prompt through a label yet. Until that exists, moving a label has no effect on the apps that consume prompts.

## Changes

Adds `?label=production` to the prompt fetch endpoint. The response is the version the label points to, plus a `label` field showing what was resolved. Stacked on #70862.

- `?label=` on `GET .../llm_prompts/name/{name}/`, mutually exclusive with `?version=` (400 if both are passed)
- Label results get their own HyperCache entry (`{team}:{name}:label:{label}`). We cache the resolved prompt content, so a warm fetch is two Redis reads (the latest entry for metadata, then the label entry) with no Postgres lookup
- Signals on `LLMPromptLabel` clear that cache entry whenever a label is created, moved, or deleted
- New `prompt_label` property on the `$llm_prompt_fetched` event
- The get-prompt MCP tool description now explains the `label` parameter
- Fetching without a label or version still returns the latest version, same as before

Design notes:

- **Why signals for cache invalidation:** the existing invalidation runs when a prompt version row is saved, and a label move doesn't save any version row — without its own trigger, a moved label would serve the old version until the cache expires (up to a day). A signal on the label model covers every path that changes labels (set, delete, archive's cleanup) in one place.
- **Creating a label also needs invalidation:** fetching a missing label caches a short-lived "not found", and the create signal clears it so a new label works right away.

New tests in `test_llm_prompt.py`, all through the public API with the cache in the loop: fetching by label after a move returns the new version (this test fails if the invalidation breaks, and it caught a real gap during development), a label created after a cached 404 works immediately, label+version together returns 400, and fetches 404 after label delete and after prompt archive. Full prompt API + cache suites pass locally (119 tests), mypy clean on changed files.

### DIFF
--- a/posthog/api/llm_prompt_serializers.py
+++ b/posthog/api/llm_prompt_serializers.py
@@ -112,6 +112,12 @@ class LLMPromptGetByNameQuerySerializer(LLMPromptFetchQuerySerializer):
         help_text=CONTENT_MODE_HELP,
     )
 
+    def validate_label(self, value: str) -> str:
+        # Fetching also writes to the cache (miss sentinels under caller-controlled keys),
+        # so impossible label names are rejected before any cache touch — and a caller
+        # sending 'Production' gets the naming rules instead of a confusing 404.
+        return validate_prompt_label_name_value(value)
+
     def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
         if attrs.get("version") is not None and attrs.get("label") is not None:
             raise serializers.ValidationError("Use either version or label, not both.")
--- a/posthog/api/llm_prompt_serializers.py
+++ b/posthog/api/llm_prompt_serializers.py
@@ -101,8 +101,10 @@ class LLMPromptFetchQuerySerializer(serializers.Serializer):
 class LLMPromptGetByNameQuerySerializer(LLMPromptFetchQuerySerializer):
     label = serializers.CharField(  # type: ignore[assignment]
         required=False,
+        max_length=PROMPT_LABEL_NAME_MAX_LENGTH,
         help_text=(
-            "Fetch the version this label currently points to, e.g. 'production'. Mutually exclusive with version."
+            "Fetch the version this label currently points to, e.g. 'production'. "
+            "Lowercase letters, numbers, dots, hyphens and underscores. Mutually exclusive with version."
         ),
     )
     content = serializers.ChoiceField(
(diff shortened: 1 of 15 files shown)


posthog/storage/llm_prompt_cache.py
```diff
@@ -113,6 +144,31 @@ def get_prompt_by_name_from_cache(team: Team, prompt_name: str, version: int | N
         except Exception as err:
             capture_exception(err)
 
+    if label is not None:
+        label_key = prompt_label_cache_key(team.id, prompt_name, label)
+        labeled_prompt = llm_prompts_hypercache.get_from_cache(label_key)
```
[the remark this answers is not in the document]
jurajmajerik: Fixed

posthog/storage/llm_prompt_cache.py
```diff
@@ -113,6 +159,31 @@ def get_prompt_by_name_from_cache(team: Team, prompt_name: str, version: int | N
         except Exception as err:
             capture_exception(err)
 
+    if label is not None:
+        label_key = prompt_label_cache_key(team.id, prompt_name, label)
+        labeled_prompt = llm_prompts_label_hypercache.get_from_cache(label_key)
```
Radu-Raicea, line 164: Should we make this fetch skip the object storage?

posthog/api/llm_prompt_serializers.py
```diff
@@ -99,13 +99,30 @@ class LLMPromptFetchQuerySerializer(serializers.Serializer):
 
 
 class LLMPromptGetByNameQuerySerializer(LLMPromptFetchQuerySerializer):
+    label = serializers.CharField(  # type: ignore[assignment]
```
Radu-Raicea, line 102: We should make this aware of the constraints from `validate_prompt_label_name_value`

Radu-Raicea (approved):

posthog/api/llm_prompt_serializers.py
jurajmajerik: Added: https://github.com/PostHog/posthog/pull/71509/commits/ec2d81684c4173bcec1db6a29754085af463f57b

posthog/storage/llm_prompt_cache.py
jurajmajerik: Yep, will do in a follow up!

PR #71520, opened 2026-07-16
feat(data-warehouse): resumable rest_source paginators + first source migration (alguna)

Gilbert09:
## Problem

Most warehouse REST sources hand-roll their transport (fetch/retry/pagination/auth) instead of using the shared `rest_source` framework, because the framework couldn't express a few behaviors those sources rely on. This PR closes the last generic gap and migrates the first source as proof.

## Changes

**Framework — resumable built-in paginators.** `OffsetPaginator`, `PageNumberPaginator`, `JSONResponseCursorPaginator`, and the next-URL paginators now implement `get_resume_state`/`set_resume_state` (and resume-aware `init_request`). Previously only a hand-written `BasePaginator` subclass could resume — which is *the* main reason sources rolled their own transport. Now any resumable source uses a built-in paginator.

**Migration — `alguna` onto `rest_source`.** End to end: `OffsetPaginator` + resume, Bearer auth, the date-version header, `data_selector_required` (fail-loud on shape change), `validate_via_probe`, and `build_endpoint_schemas`. The hand-rolled fetch/retry/pagination is gone.

> [!NOTE]
> **Stacked PR.** This branch contains the commits from #71477 (helpers + skill) and #71481 (rest_source gaps: data_map, fan-out resume, fail-loud selector) as well. Merge those first; the only net-new commit here is the top one — `resumable built-in paginators + migrate alguna` — and the diff collapses to just that once the two foundations land. Review that commit.

This is the base for the source-migration batches that follow, stacked on top.

- **alguna: all 25 tests pass** — behavior preserved vs the old hand-rolled transport: offset progression, resume-from-saved-offset, short-page termination (one request, no extra call), sort param present/absent per endpoint, fail-loud `ValueError` on a missing `data` key, credential probe (status mapping + swallows exceptions), and `SourceResponse` partition/primary-key metadata.
- **Paginators: 26 tests pass**, incl. 5 new resume round-trip tests (state out → seed a fresh paginator → first request targets the resumed page) for offset/page/cursor/next-url.
- Full `rest_source` suite + chargebee/zendesk/attio consumers still green. `hogli ci:preflight --fix` clean.

## Automatic notifications

- [ ] Publish to changelog?


estefaniarabadan (approved):
should SOURCES.md be updated to reflect alguna's migration to rest_source?

PR #71530, opened 2026-07-16
feat(warehouse_sources): support Klaviyo API version 2026-07-15

danielcarletti:
## Problem

Klaviyo's `2024-10-15` API revision (the one our data-warehouse Klaviyo source pins today) is in its deprecated phase and retires around 2026-10-15, after which requests fall forward to the next revision or return 410. We need to move onto the current `2026-07-15` revision without breaking any source still pinned to a supported version.

## Changes

- Add `2026-07-15` to `KlaviyoSource.supported_versions` and make it the new `default_version`, so new sources start on the current revision.
- Mark `2024-10-15` deprecated with `sunset_at=2026-10-15`. The generic in-product deprecation warning lights up from this metadata, no source-specific UI.
- Thread the resolved pin into Klaviyo's `revision` request header. It was hardcoded to `2024-10-15`; now `source_for_pipeline` passes `resolve_api_version(inputs.api_version)` down through `klaviyo_source` → `get_rows` → `_get_headers`, so a source pinned to either supported revision syncs against exactly that revision. Old pins keep working verbatim.
- Add version labels as constants in a new `constants.py` so the version declaration and the request layer share one label (mirrors the Stripe pattern).
- Data migration `0079_repin_klaviyo_api_version` repins the source-level `ExternalDataSource.api_version` from `2024-10-15` to `2026-07-15`.

> [!NOTE]
> The migration is written but not run. A human reviews and executes it.

### Why this repin is a safe automatic transform

I checked the Klaviyo changelog for every GA revision from `2024-10-15` through `2026-07-15`. None of the breaking changes touch fields this source reads:

- We flatten each object's `attributes` generically into columns and never request `?include=conversation` / `relationships`, so the `2026-07-15` profile "conversations became plural" break does not affect us.
- The `campaigns`, `flows`, `lists`, `metrics`, `profiles`, and list-membership responses are backward compatible for the fields we read (incremental cursors `updated_at` / `created_at` / `updated` / `created` / `joined_group_at` are unchanged).
- The one behavior difference: under `2026-07-15` the events endpoint returns events whose metric can't be resolved by default (opt-out via `filter=has(metric)`). That is additive (potentially more event rows), not a renamed or removed field, so no data or schema transform is needed. Repinned sources will start picking up those events.

Because no field-level transform is required, the repin is a single idempotent `UPDATE`. It filters on the exact old value (`api_version="2024-10-15"`), so re-runs are no-ops, and it leaves schema-level `ExternalDataSchema.api_version` overrides alone since those are intentionally customer-pinned. Reverse is a no-op: once repinned, a source on `2026-07-15` is indistinguishable from one natively created on the new default.

Automated tests I (Claude) actually ran, all green:

- `test_klaviyo.py` (46 passed) including two new tests:
  - `TestApiVersionThreadsToRevisionHeader` (parameterized over both supported revisions) — catches a regression where the version stops reaching the `revision` header and the sync silently uses the wrong revision.
  - `TestVersionDeprecation` — asserts `2024-10-15` is deprecated with the exact `2026-10-15` sunset and the current default carries no deprecation. The registry invariant test checks set relationships but not the specific sunset date this PR pins.
- `test_source_versions.py -k Klaviyo` (1 passed) — the registry invariants (default in supported, deprecated ⊆ supported, default not deprecated).

I could not run the migration or `makemigrations --check` in this environment (no database). The migration makes no model-field changes (pure `RunPython` data migration), so there is no schema state to drift.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change: the source setup and table list are unchanged.

## 🤖 Agent context

Authored by Claude (Opus 4.8) via PostHog Code, directed by @danielcarletti.

Skills invoked: `/warehouse-source-new-version` (version-declaration + dispatch conventions), `/django-migrations` (migration safety + idempotency), `/writing-tests` (test value gate).

Decisions: kept dispatch minimal — the version is just a header value and the endpoints/fields we read are backward compatible across the two revisions, so no per-version modules or column-hint gating were needed (Klaviyo auto-infers its schema; it has no managed HogQL schema). Ruled out a data/schema transform after confirming the only in-range field breaks (`campaigns` push-notification changes in 2025-01-15, profile `conversations` in 2026-07-15) do not touch fields this source consumes. Checked for a duplicate PR first: #63201 bumps the Klaviyo CDP *destination* revision (different product, revision 2026-04-15) and is unrelated; no open PR addresses this source's version bump.

---


danielcarletti:
Re: Greptile's P2 on cross-revision resume state — I looked at this and don't think it needs a code change in this PR, but flagging my reasoning for a human to confirm rather than resolving the thread.

Resume state (`ResumableSourceManager`) is keyed by `job_id` with a 24h TTL and is cleared on successful completion. It only survives across retries of a single sync job, not across scheduled runs. So the scenario needs a `2024-10-15` job to crash mid-pagination, the repin migration to run within 24h, and that same job to retry and re-resolve the pin.

Even then the worst case is one failed attempt: the next scheduled sync is a new `job_id`, finds no resume state, and starts fresh under `2026-07-15` — it self-heals, it can't get stuck. And Klaviyo pins the revision via the `revision` header (not the cursor URL), and the sort/filter semantics on the endpoints we sync are unchanged between these revisions, so a continuation cursor created under the old revision should still be accepted.

This property is inherent to the shared `resolve_api_version(inputs.api_version)` path every versioned source uses (Stripe's repin has the same shape), so tagging resume state with its revision would be a framework-level change beyond this version bump. Happy to do it as a separate change if a maintainer wants the belt-and-suspenders.

Gilbert09 (approved):

PR #71551, opened 2026-07-16
feat(review_hog): scope custom review skills to their author

sortafreel:
## Problem

Custom review skills (perspectives, validators, blind-spots checks) are team-level `LLMSkill` rows, and all three Code review config menus listed every prefixed team skill. A custom perspective one user authored showed up in every teammate's menu, with its body readable and enableable by exact name. As more people author custom skills, everyone's menu fills up with everyone else's customs.

**Why:** users should only see the canonical skills plus the customs they created themselves, so the review menus stay theirs.

## Changes

- New `visible_skill_names()` in `skill_loader.py`: a skill is visible to a user iff it's canonical (by name, so team-edited canonicals stay visible to everyone) or the earliest live version's `created_by` is that user. Every edit mints a new version stamped with the editor, so only the first version identifies the author.
- All three config viewsets (perspectives, validators, blind-spots) filter `list` by visibility, and `partial_update` 404s on a non-visible skill exactly like a missing one, closing the enable-any-team-skill-by-name backdoor.
- Loader backstop instead of a data migration: `load_perspectives_for_run` warn-and-skips enabled names outside the acting user's visible set (hole-preserving `pass_number` slots, same as the dead-skill skip), and the single-active loaders fall back to the canonical on a foreign selection. Leftover foreign-enabled configs stop running instead of running invisibly.
- Decision and rationale recorded in `products/review_hog/ARCHITECTURE.md` (dated entry under "Customizable perspectives").

> [!NOTE]
> The team-level Skills page (`/skills`) intentionally still shows all team skills. This is a menu-scoping fix, not a confidentiality boundary. Signals scouts share the same all-team-menu pattern and are deliberately out of scope.

Automated tests added/updated (each guards a regression no existing test caught):

- Per surface: a teammate's custom is hidden from `list` and PATCHing it by name 404s (previously the exact reported behavior).
- Perspective loader skips a foreign-enabled config while keeping `pass_number` holes stable (resume-key safety).
- Single-active validator loader falls back to the canonical on a foreign selection.
- Existing custom-skill fixtures now stamp `created_by`, so the whole existing suite runs under the new visibility rule.

I (Claude via PostHog Code) could not execute the test suite in the cloud sandbox (no service stack for pytest), so these tests are authored but not yet run. Please run locally before marking ready:

```
hogli test products/review_hog/backend/tests/
```

`ruff check` and `ruff format` pass. `hogli ci:preflight` flags OpenAPI type drift (endpoint description text changed); run `hogli build:openapi` locally and commit any regenerated files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No `docs/` pages cover the review-hog config menus yet.

## 🤖 Agent context

Authored by Claude in a PostHog Code session directed by the assignee. Skills invoked: `/improving-drf-endpoints`, `/writing-tests`. Design was settled interactively before implementation: chose earliest-live-version `created_by` as the authorship signal (no schema change) over adding an owner field to `LLMSkill` or a config-row-subscription model; chose strict enforcement with no grandfathering (feature just released) with loader-level soft-skip as the backstop instead of a data migration.

---

### DIFF
--- a/products/review_hog/ARCHITECTURE.md
+++ b/products/review_hog/ARCHITECTURE.md
@@ -2929,9 +2929,9 @@ Locked with the user (AskUserQuestion, 2026-07-16):
 
 - **Visibility rule:** a skill is visible to a user iff its name is canonical (`CANONICAL_*_SKILL_NAMES` — by name, so a team-edited canonical stays visible to everyone)
   OR the **earliest live version's `created_by`** is that user.
-  This refines the 2026-06-29 "`created_by` is audit-only" lock: still audit-only for identity and editing (edits mint new versions stamped with the *editor* and do NOT transfer visibility);
+  This refines the 2026-06-29 "`created_by` is audit-only" lock: still audit-only for identity and editing (edits mint new versions stamped with the _editor_ and do NOT transfer visibility);
   the first version's author now governs **menu visibility only**.
-  Earliest *live* version rather than literally `version=1`: deleted rows are excluded, so an archived name later recreated by another user belongs to the new author
+  Earliest _live_ version rather than literally `version=1`: deleted rows are excluded, so an archived name later recreated by another user belongs to the new author
   (deleted `(name, version)` pairs can duplicate — the unique constraints only cover `deleted=False`).
 - **Strict, no grandfathering** (decided while the feature had no external adopters): `partial_update` 404s on a non-visible skill exactly like a missing one
   (a distinct error would leak that the name exists), closing the enable-any-team-skill-by-name backdoor. No data migration; the loaders backstop instead.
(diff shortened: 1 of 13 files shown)


Twixes (approved):

dkarasiewicz (approved):

sortafreel:
Note on 0e3576dd: master's Frontend typechecking was failing for every PR — the recent inline-kea-types batches left four files stale against what `typegen:write` regenerates in CI (kea-typegen rewrites interfaces but doesn't remove orphaned type imports, so tsgo fails with TS6196). This commit regenerates those four files and drops the dead imports so this PR (and master after merge) typechecks again. Overlaps with the broader catch-up in #71587, which will want a rebase after this lands.

PR #66053, opened 2026-06-25
feat(conversations): log in as customer directly from support ticket

luke-belton:
## Problem

When working a support ticket as PostHog staff, logging in as the customer meant leaving the ticket, opening the Django admin user search, finding the user, and impersonating from there. Slow, easy to lose the ticket's context, and there was no way back to the ticket afterwards.

This continues the work started in #48334, rebuilt on the current support layout, and builds on the `identity_verified` trust signal from #54628 (now merged).

## Changes

The ticket "Staff actions" tile's **Login as customer** button now starts a read-only impersonation session **directly**, and the impersonation notice knows how to get you back to the ticket when you're done.

### Direct login from the ticket

- **Backend** – new `/admin/impersonation/from-ticket/` endpoint (`ee/admin/loginas_views.py`) resolves the ticket's customer email to a user, auto-fills the reason (`Support ticket #<n>`), and starts a read-only session. If the ticket originated in another region it returns a redirect to that region's admin instead of failing.
- **Region inference** – for PostHog's internal support project, the widget derives a `region` trait from the app URL (`products/conversations/backend/api/widget.py`), which drives the cross-region routing above.
- **Frontend** – `AdminLoginButtons` is a single direct-login button wired to a new `initiateImpersonation` flow (`loginAsFromTicket` reuses the existing admin OAuth2 handshake from `adminLoginAs.ts`). The old multi-region admin-redirect machinery was removed.

Verified identity, US region:

https://github.com/user-attachments/assets/f2dd0ca3-cd27-4286-9dd0-4dc3d5ed84e7

Identity unknown:

CleanShot 2026-07-03 at 11 21 36

Unverified identity:

CleanShot 2026-07-03 at 11 23 28

### Return to ticket

- The originating ticket id is stored in the session, and `/logout` from a ticket-initiated impersonation lands staff back on the ticket rather than the admin user page.
- The impersonation notice shows a **Return to ticket #X** button (with loading state) instead of the usual logout dropdown when the session was started from a ticket. A small `{ticketNumber, email}` context is persisted across the post-login reload and dropped whenever the stored ticket no longer matches the active impersonation (staff load, or a different customer impersonated via Django admin).
- The **session expired** overlay offers the same "Return to ticket #X" action, and the re-impersonate / upgrade-to-read-write modals prefill the reason with `Investigating ticket #X`.

### Identity verification gate

[...]

PR #71561, opened 2026-07-16
fix: staff actions

veryayskiy:
## Problem

Screenshot 2026-07-16 at 16 05 22

## Changes

Screenshot 2026-07-16 at 16 05 31


luke-belton (approved):
this will change slightly anyway when I get a review on https://github.com/PostHog/posthog/pull/66053 from my team, but looks good for now 👍

PR #70018, opened 2026-07-10
feat(data-warehouse): add vendor API versioning framework for warehouse sources

danielcarletti:
## Problem

Vendors ship new API versions and our warehouse sources silently rot on old ones. We're building automation that detects new vendor API versions and opens PRs adding support for them, but sources have no first-class notion of an API version today: what a source implements is buried in hardcoded URLs and headers, nothing records which version a customer's source syncs with, and there is no way to add a new version without changing behavior for everyone at once.

## Changes

**Versioning declarations on every source.** `_BaseSource` gains `supported_versions` (opaque vendor labels, never parsed or ordered), `default_version`, `api_docs_url` (the vendor's API changelog page, distinct from the posthog.com `docsUrl`), and `deprecated_versions` (per-version `deprecated` + `sunset_at` metadata). Unversioned vendors (SQL databases, file stores) inherit the `"v1"` default untouched. All 217 released sources are seeded with the version their code currently implements, read from their URLs/headers/SDK pins (e.g. Stripe `2024-09-30.acacia`, Klaviyo `2024-10-15`, Shopify `2025-10`, Meta Ads `v25.0`), plus the vendor API docs URL.

**A version pin per source instance.** `ExternalDataSource` gets a nullable `api_version` column. A dedicated column rather than `job_inputs` because `job_inputs` is encrypted at rest, and the pin must be queryable: it's exposed in the `data_warehouse_sources` HogQL system table. New sources are stamped with the source type's `default_version` at creation, a data migration backfills every existing row with its source's current version (frozen mapping, idempotent, never overwrites an existing pin), and a NULL pin resolves to the default at runtime. Net effect: flipping a source's default later never changes behavior under an existing customer.

**Resolution in the sync pipeline.** `import_data_sync` resolves the pin through the source's `resolve_api_version` and hands it to the source as `SourceInputs.api_version`. A present pin is honored verbatim even if no longer declared: silently moving a customer to another version is the exact failure mode this framework exists to prevent. Stripe demonstrates the dispatch pattern by threading `inputs.api_version` down to its `StripeClient` (same value it hardcoded before, so no behavior change). Other sources keep their single hardcoded version until their first real version bump introduces dispatch.

**Public catalog.** `GET /api/public_source_configs/` entries now include `versions`, `defaultVersion`, `apiDocsUrl`, and `deprecatedVersions` (additive; consumed by external automation).

[...]

PR #71570, opened 2026-07-16
feat(data-warehouse): seed vendor API version metadata for remaining sources

danielcarletti:
## Problem

The vendor API versioning framework (#70018) shipped with metadata seeded on the ~197 flagship connectors, but it left the long tail of connectable REST sources on the bare unversioned default with no docs link. Those sources have been in the tree for a while — this isn't a batch of new connectors, it's the coverage gap the original PR didn't close. Without `api_docs_url` and a real version pin, the deprecation UI has nothing to show and the `public_source_configs` payload reports an empty version story for most sources.

## Changes

Seeds vendor API version metadata on 236 connectable REST sources that were missing it. For each source:

- `api_docs_url` — the vendor's API reference / changelog page (always https).
- `supported_versions` + `default_version` — only when the vendor exposes a real version token (a URL path segment like `/v3/`, a dated version header, a named release), pinned to **the version the source's own code actually calls**, not the vendor's newest. Sources with no meaningful version, or with per-endpoint mixed versions that can't be pinned to one token, keep the unversioned default and get docs-only.

Database sources (Postgres, MySQL, MSSQL, Snowflake, Redshift, Supabase, MongoDB, Elasticsearch) are **excluded** — there's no vendor API version to pin, same reasoning the original PR used to skip them.

Also updates the authoring guidance so new sources declare this up front:
- `implementing-warehouse-sources` skill — a new "Vendor API version metadata" section (the two cases, the "pin what the code calls" rule, the CI invariants, and a pointer to `/warehouse-source-new-version` for later bumps).
- `source.template` — an inline TODO on the class body where the attributes go.

> [!NOTE]
> These are pure metadata class attributes. No framework, serializer, migration, or generated-config change. `supported_versions=("v1",)` is identical to the framework default, so a source pinned to bare `v1` and a docs-only source resolve the same way at runtime.

The one source deliberately left uncovered is `custom` — the generic user-defined REST source has no vendor, so a docs URL would be misleading.

Automated only (I'm Claude; I did not manually connect any source):

- `test_source_versions.py` — 858 registry-wide invariant assertions pass (default ∈ supported_versions, https `api_docs_url`, no duplicate/empty versions) across every registered source, including all 236 seeded here.
- `test_public_source_configs.py` — 8 pass (the endpoint now surfaces the seeded metadata).
- `ruff check` + `ruff format` clean on all changed files; the diff is pure additions (zero deletions in source files).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Daniel directed this as the follow-up to #70018.

I (Claude) ran the research as a parallel fan-out: an audit script walked the registry to find every connectable source (`unreleasedSource` falsy) still on the unversioned default, excluded DB sources by base class (`SQLSource` + Mongo/Elasticsearch), then 16 subagents each web-researched a batch of ~15 sources and edited their `source.py`. Every version token was grounded in the source's own client code first, then web-verified against the vendor docs.

- **Bare `/v1` is treated inconsistently on purpose.** Some sources pin `supported_versions=("v1",)`, others stay docs-only. It's functionally identical (the framework default *is* `v1`), and the existing corpus already mixes both, so I let each source's judgment stand rather than force a normalization that would risk wrongly pinning `v1` on mixed-version APIs (mux, scaleway, vercel, inngest…).
- A handful of sources (campayn, eventee, justsift, papersign, persistiq, qualaroo, simplesat, smartreach, zenloop) already had `supported_versions` from a prior pass but no docs link; those got only `api_docs_url` added.
- A couple of low-confidence docs URLs where no vendor-owned developer site exists: `campayn` points at the community `Campayn-API` GitHub repo, `hellobaton` at the vendor integrations page. Called out here so a reviewer can swap them if they know better.
- Master merged in a brand-new `coveralls` source during this work; it's left unseeded (arrived after the audit). The audit script is repeatable, so a follow-up can sweep any newly-added sources.

Skills referenced: `implementing-warehouse-sources`, `warehouse-source-new-version`.


estefaniarabadan (approved):
Small finds:
- Anthropic docs URL: verify https://platform.claude.com/docs/en/api/versioning actually resolves (historically docs.anthropic.com).
- Source template formatting: missing blank line between the TODO metadata comment block and @property in source.template.

PR #71574, opened 2026-07-16
chore(data-warehouse): finish source-agnostic presentation layer

Gilbert09:
## Problem

After the direct-query migration, the DWH source/schema views still named concrete source types in five more families: xmin gating, GitHub multi-repo reconciliation, custom-source OAuth/manifest/limits, and SQL schema reconciliation. This PR migrates all of them, driving the source-agnostic guard baseline to **zero**.

## Changes

> [!NOTE]
> Stacked on #71558 (direct-query) → #71543 (guard). Review those first. Behaviour-preserving throughout. One commit per family.

Each family moves to whichever mechanism fits the concern (per `presentation/README.md`):

- **xmin** → a `supports_xmin` flag on `_BaseSource` (Postgres sets it True). The views read `SourceRegistry.get_source(...).supports_xmin`.
- **Server-managed job inputs** (Custom's `auth_oauth2_integration_id` pointer, GitHub's legacy `repository` marker) → `_BaseSource.server_managed_job_input_fields(incoming, existing)`; Custom and GitHub override it.
- **GitHub multi-repo reconciliation** → a `data_warehouse` namespaced-resource registry (`namespaced_resources.py`) keyed on source type, with a GitHub adapter wrapping the existing repo helpers. The view dispatches via `get_namespaced_resource_adapter(source_type)`.
- **Custom source** (per-team cap, manifest-host gate, row-backed OAuth2 credential-preservation gate, post-create integration-row claim) → capabilities on `_BaseSource` (`max_instances_per_team`, `job_inputs_add_connection_host`, `has_preserved_row_backed_credentials`, `on_source_created`) implemented on `CustomSource`, which owns `CustomOAuth2Integration`. The per-team cap now generalizes to any capped source type.
- **SQL schema reconciliation** → extends the direct-query engine adapter with `reconcile_schemas` (all four engines) and `refresh_name_substitutions` (Postgres only; `None` elsewhere).

The last commit regenerates the guard baseline to empty. It's now a hard regression check: any new `ExternalDataSourceType.<NAME>` / `is_direct_<engine>` in the presentation layer fails CI.

Subtlety worth a reviewer's eye: `is_multi_schema_capable_sql_source` = {Postgres, **MSSQL**, Snowflake, Redshift}, which differs from the `direct_engine` set {Postgres, **MySQL**, Snowflake, Redshift}. So the Postgres name-substitution uses a `None`-vs-`{}` sentinel: only Postgres returns a dict (handled, suppresses the generic path); the other engines return `None` and fall through to the already-source-agnostic `is_multi_schema_capable` migration, which still covers MSSQL. Traced per engine to preserve exact behaviour.

- Extended `test/test_direct_query_engines.py` (20 cases): adds the `refresh_name_substitutions` `None`-sentinel for non-Postgres engines. All pass.
- Verified each capability resolves via the facade.
- `ruff`, `tach`, import-linter (`presentation must use facade` KEPT), and `oxfmt` all pass. The guard passes on the empty baseline and fails when a probe `ExternalDataSourceType.POSTGRES` reference is added.
- The `tests/api/` end-to-end suites can't run in my local env (pre-existing Python 3.12-vs-3.13 conftest mismatch, unrelated). CI runs them.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I directed this with Claude Code (Opus 4.8), continuing the source-agnostic sweep to completion. Each family picked the lightest mechanism that fit: a flag where a boolean sufficed, a source method where behaviour depended on the source, an engine/registry adapter where the work was `data_warehouse`-domain (touched `DataWarehouseTable`/schema rows). The main correctness risk was the reconciliation family's MSSQL-vs-MySQL set mismatch, handled with the `None` sentinel and traced per engine. Skill invoked: `implementing-warehouse-sources`.


posthog/test/setup_receivers_baseline.txt
```diff
@@ -63,6 +63,7 @@ post_delete:posthog.storage.team_access_cache_signal_handlers.team_deleted_auth_
 post_delete:products.access_control.backend.property_access_control._invalidate_restriction_cache_on_change
 post_delete:products.actions.backend.models.action.action_deleted
 post_delete:products.ai_observability.backend.models.llm_prompt.invalidate_llm_prompt_cache
+post_delete:products.ai_observability.backend.models.llm_prompt.invalidate_llm_prompt_label_cache
```
danielcarletti, line 66: I think this file changes are not meant to be here 🤔

danielcarletti (approved):

PR #71586, opened 2026-07-16
feat(stamphog): capture provenance trailers and familiarity in review events

gantoine:
## Problem

First step of the review-at-scale RFC (#71584).

Two signals stamphog already has access to are currently thrown away:

- Every agent-authored commit is stamped with `Generated-By:` / `Task-Id:` trailers, and nothing in the repo reads them. In a 30-PR sample of recently merged PRs, 22 carry them, so agent authorship is a well-populated dimension we simply never capture.
- `familiarity.py` computes how familiar the PR author is with the code the PR touches (git-blame overlap, prior PRs in the paths, recency), but only on the T1-agent path, and the result is never persisted. That makes a per-subsystem knowledge trend impossible.

## Changes

- `github.py`: new `CommitProvenance` plus `parse_provenance_trailers()` / `pr_provenance()`, which read the PR's head commits from the local checkout (`git log base..head`). Head commits, never the squash-merge commit, since squash merges can drop or rewrite trailers. No new API calls; `ensure_commits` already fetches the objects.
- `review_pr.py`: familiarity now computes on every LLM-reviewed run instead of only T1-agent. The reviewer prompt is unchanged outside T1 (the signal is still attached to the classification only there); this is telemetry, not a behavior change.
- `stamphog_review_completed` gains: `stamphog_owner_teams`, `stamphog_familiarity_band` / `_blame_overlap_pct` / `_prior_prs_in_paths` / `_days_since_last_touch`, `stamphog_agent_authored`, `stamphog_agent_commit_count`, `stamphog_commit_count`, `stamphog_generated_by`, `stamphog_task_ids`.
- The `--output-json` payload gains a `provenance` block, and its `familiarity` field is now populated on every tier (nothing downstream parses it; the workflow reads `head_sha`, `review_body`, and `final_verdict`).

> [!NOTE]
> No gate, verdict, or prompt behavior changes outside what T1 already did. Early-exit paths (bot author, WAIT, pending migration check) emit the new properties as null/empty.

- Full stamphog suite: 431 passed locally (`pytest tools/pr-approval-agent/`).
- New tests and the regression each catches:
  - `test_parse_provenance_trailers` (parameterized over the message shapes git actually emits, including squash-style bodies and case variants): a parsing regression would silently zero the agent-authored dimension rather than fail anything.
  - `test_familiarity_computed_on_every_tier_but_prompted_only_on_t1`: reintroducing the old T1-only guard would silently drop non-T1 familiarity telemetry; attaching the signal to non-T1 classifications would change those prompts.
  - `test_capture_review_completed_includes_familiarity_and_provenance` (populated and absent variants): locks the event property names and null/empty defaults downstream queries key on, and guards the early-exit paths where both signals are still None.
- Verified the parser against real data: PR #71542's commits parse to `agent_commit_count=1, generated_by=('PostHog Code',)` with the correct task UUID.
- I (the agent) did not run a live labeled review end to end; the event capture path is covered by the tests above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed; stamphog's README describes gates and verdicts, which are unchanged. The motivation is documented in the RFC (#71584).

## 🤖 Agent context

I directed this from a PostHog Code session; Claude (Fable) wrote the implementation and tests. Skills invoked: `/writing-tests`. Decisions along the way: familiarity stays out of the reviewer prompt for non-T1 tiers so this PR is purely additive telemetry (the conservative option, chosen to keep stamphog behavior reviewable separately from the RFC direction); trailers are parsed from head commits via local git rather than a new API call; a trivial `agent_authored` property test was written and then deleted per the testing value gate.

---

### DIFF
--- a/tools/pr-approval-agent/github.py
+++ b/tools/pr-approval-agent/github.py
@@ -5,6 +5,7 @@ else (PR metadata, reviews, comments, check runs) from the GitHub API.
 Also handles team membership checks for the ownership gate.
 """
 
+import re
 import json
 import subprocess
 from collections.abc import Callable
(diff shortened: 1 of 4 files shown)


rnegron (approved):

feliperalmeida (approved):

Piccirello:
**Agent-driven security review** — all findings are **low severity or informational. Nothing here blocks merge.** Posting so they're on record rather than because anything needs fixing before this lands.

First, the important part: the "telemetry, not a behavior change" claim holds up under checking. `provenance` never enters `self.classification`, so it reaches neither the reviewer prompt nor `_render_review_body`. `_format_familiarity` still reads `cl.get("familiarity")`, which is still only populated on `T1-agent`, so non-T1 prompts and every posted comment are byte-identical to before. Gates, tier assignment, and verdict logic are untouched, the new `git log` args aren't user-interpolated (shas come from the API), and every new failure path returns `None` and degrades to null/empty properties. No approval-bypass or injection path.

The rest are hardening notes on the new untrusted-input channel (commit-message text).

### Low — two superlinear parses over unbounded commit-message text, run before any gate

`pr_provenance()` runs inside `_fetch()`, ahead of the bot-author gate and every deterministic gate, over the full unbounded output of `git log --format=%B`. Two independent quadratics, both measured against this branch:

1. `^generated-by:[ \t]*(.+?)[ \t]*$` backtracks quadratically on a long run of spaces before the line end — 40k spaces parses in 3.0s, and it's O(n²) from there.
2. `g not in generated_by` / `t not in task_ids` scan a growing list: 10k trailers → 0.19s, 20k → 0.63s, 40k → 2.8s, 80k (a 1.2 MB message) → 11.5s.

A large enough commit message therefore eats the job's `timeout-minutes: 20` and the run is killed. That's fail-closed — no verdict means no approval, the sticky comment says so and the label is retained — so the impact is runner minutes plus stamphog being unable to review that PR. Only reachable by someone with write access on a same-repo PR, since fork PRs die at the app-token step.

Capping `log_output` bytes (and the collected trailer count) before parsing bounds both at the root. A linear regex (`^generated-by:(.*)$` plus `.strip()`) and set-based dedup alongside the ordered list are cheap defense in depth.

### Low — an author can silently drop their own `stamphog_review_completed` event

`stamphog_generated_by` / `stamphog_task_ids` are captured with no length or count cap. `posthoganalytics`'s consumer drops any event over `MAX_MSG_SIZE = 900 KiB` in its background thread with only a `log.error` — nothing surfaces to the caller. Around 900 KB of distinct short `Task-Id:` lines parses fast enough to reach `capture()` and then takes the *whole* event with it, including `stamphog_gate_verdict`, `stamphog_llm_verdict`, `stamphog_final_verdict` and the reasoning. Since that event is the monitoring record for an unattended approver, losing it silently is worth avoiding.

Worth noting this needs a different shape of input than the item above (many distinct short values vs. one long line), so fixing one doesn't cover the other. `_sanitize_untrusted(text, max_len)` in `policy.py` already does the right thing for untrusted logins in the review body and would fit here, plus a cap on list length.

### Low, but the one worth writing down — provenance is self-asserted

Nothing verifies that a `Generated-By:` trailer was written by an agent; anyone can type it, and the regex matches anywhere in the body rather than only in the trailer block. The record separator is spoofable too — git preserves a literal `\x1e` byte in a commit message, so one commit crafted with an embedded separator parses as `commit_count=2, agent_commit_count=1`, i.e. arbitrary control over the agent-authored ratio in both directions.

Harmless for telemetry, and I don't think it needs changing here. It matters because #71584 sequences later work to consume this pipeline. If `agent_authored` ever weights a tier, a gate, or reviewer assignment, it becomes a forgeable input. A line in the `CommitProvenance` docstring (and in the RFC) saying the dimension is self-asserted and must not gate anything without verified attribution — verified-committer identity, or the pushing app identity from the API — would be enough to keep that from being rediscovered later.

### Low — familiarity now runs on every tier with no aggregate time budget

Per-call timeouts are all there, but nothing bounds the total: 1 `gh` call + `merge-base` + up to 30 `git blame` + 2 `git log`, each at 30s, is ~17 minutes worst case inside a 20-minute job that already reserves 5 minutes for the bot-review wait plus the LLM roundtrip. That cost used to be confined to `T1-agent`; it now applies to `T0-deterministic` and to `T2-never` denials that never use the result. Same fail-closed outcome as above. A single wall-clock deadline across `compute_familiarity`, returning partial or `None` when exceeded, would cover it.

### Info — per-author familiarity now lands in the public artifact on every tier

`to_dict()` moved from `classification["familiarity"]` to `self.familiarity`, so `review.json` — uploaded with 30-day retention from a public repo — now carries `blame_overlap_pct`, `days_since_last_touch`, `prior_prs_in_paths` and `top_prior_authors` on T0 and T2 runs too. It's all derived from public git history and the posted comment is unchanged, so nothing secret is exposed. Flagging it only because it's per-person data attached to a named individual, published by default, and the RFC specifies atrophy signals should be subsystem-level rather than per-person.

PR #71597, opened 2026-07-16
feat(mcp): add skill usage suite for learn search quality and reference pulling

skoob13:
## Problem

`eval_skill_distribution` scores the discovery protocol (search first, load by qualified name, tool after load) but leaves two gaps: search quality is only evaluated indirectly, and pulling nested reference files isn't evaluated at all — an agent that ignores every `references/*.md` scores identically to one that follows them. None of the `learn` improvements on the base branch (stem-tolerant matching, cross-source relevance merge, zero-hit recovery, batch reads) had eval-level signal.

## Changes

New sandboxed suite `eval_skill_usage` (experiment `sandboxed-cli-mcp-skill-usage-cli`) with four cases and three new scorers, reusing the distribution scorers for the shared protocol checks:

- **`ExpectedReferencePulled`** — did the agent read one of the skill's reference files (exec: `learn <skill> <path>` incl. batch/`-s`/`--lines` forms; bundled: native file read of the mounted reference)? A bare skill load or all-qualified batch load doesn't count.
- **`SearchRecoveryAfterZeroHit`** — opportunistic: after a zero-hit `learn -s`, the next command must be another `learn`, not a product tool. Self-skips when no zero-hit occurred.
- **`SkillAnswerCorrectness`** — binary judge: the final message states every fact of an expected answer (paraphrase ok, numbers must match).

Cases: experiment bias thresholds that live only in `bias-and-skew.md`; error-tracking system-table query routed through the `querying-posthog-data` dispatcher; a paraphrased-vocabulary bot-traffic prompt exercising stem-tolerant search; and an exec-only case whose `setup` seeder plants an `LLMSkill` project skill (`project:hedgebox-revenue-policy` with a needle exclusion rule in a reference file) to cover cross-source merged search, `learn project:` loads, and project reference reads end to end.

Supporting changes:

- `skill_distribution_scorers.py` — expectations optionally carry `source: posthog|project` (default posthog, fully backward compatible).
- `eval_skill_distribution.py` — drive-by: removed the explicit `ExitCodeZero()`; the harness auto-adds it and rejects the duplicate with a `ValueError`, so the suite could not construct.
- `harness/services.py` — carries the `pnpm dev` launch fix, also submitted separately as #71583 against master; this copy falls away on rebase once that merges.

- Unit: 62 scorer tests pass (`test_skill_usage_scorers.py` + extended `test_skill_distribution_scorers.py`). New tests catch: reference-pull grammar parsing (flag truncation, batch-load vs reference-read distinction, project source), zero-hit recovery ordering, and judge `_prepare` short-circuits — none covered before.
- Live smoke runs (docker provider):
  - `project_enterprise_revenue_policy` exec: reference pull, answer judge, load-before-tool all 1.0. **Found a real gap on the first run:** `expected_skill_discovered` is 0 because the backend `llm_skills/search/` matches the whole query string via `icontains`, so natural multi-word queries never return project skills and the cross-source merge gets nothing to merge. The agent recovered via `learn skills`. Known-red until that's fixed (MCP-side tokenized fallback over the memoized listing is the cheap option).
  - `bias_threshold_reference_pull` exec: the agent grepped the sandbox's PostHog checkout for the thresholds instead of running `learn -s` — a genuine triggering failure the suite is designed to expose (`skill_search_first` 0), since the exec instructions mandate skills-first.
  - `bias_threshold_reference_pull` bundled: 100% — native skill loaded, reference read, judge passed; also validates the bundled-arm reference matching.
  - Regression: `eval_skill_distribution --eval paid_bill_revenue_by_account_plan --skill-delivery exec` scores 100% after the `source` generalization and `ExitCodeZero` fix.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — evals are internal tooling; suite discovery and running are covered by the existing eval harness README conventions.

## 🤖 Agent context

Planned and built with Claude Code in the same session as the base branch. Skills invoked: /writing-evals, /writing-tests (plus /improving-drf-endpoints earlier in the session; no DRF changes here). Implementation was delegated to an Opus subagent against an approved plan; the orchestrating session independently re-ran tests, ruff, suite discovery, and all four live smoke runs, and reviewed every diff. Decisions: new suite instead of extending `eval_skill_distribution` (keeps its Braintrust history comparable); client-side `source` generalization instead of new scorer classes; the `expected_answer` for the bias case was recalibrated after the first smoke run (the seeded $multiple share is ~0.9%, not the ~2% flip probability). The two known-red scorers on the project case are deliberate: they document the backend search gap described above.

### DIFF
--- a/products/posthog_ai/eval_harness/harness/services.py
+++ b/products/posthog_ai/eval_harness/harness/services.py
@@ -99,7 +99,7 @@ def start_mcp_server(
     Pointed at the in-process Django live server (which uses the test DB).
     Uses a non-default port to avoid conflicts with a running dev MCP server.
 
-    Runs the Node-native Hono server via ``pnpm dev:hono``. In production the
+    Runs the Node-native Hono server via ``pnpm dev``. In production the
     Cloudflare Worker is now a proxy that forwards to a regional Hono
     deployment, so Hono is what real users hit.
     """
@@ -114,7 +114,7 @@ def start_mcp_server(
     api_url = live_server_url
 
     # The Hono server reads config directly from process env — no wrangler
-    # --var wiring needed. PORT picks the listen port; the dev:hono script
+    # --var wiring needed. PORT picks the listen port; the dev script
     # bundles via esbuild then spawns Node on the bundle.
     env = {
         **os.environ,
@@ -145,7 +145,7 @@ def start_mcp_server(
     _, stop = LONG_LIVED_SUBPROCESSES.start(
         name="MCP server",
         port=MCP_PORT,
-        cmd=["pnpm", "dev:hono"],
+        cmd=["pnpm", "dev"],
         cwd=mcp_dir,
         env=env,
         log_prefix="mcp",
--- a/products/posthog_ai/eval_harness/test/test_skill_distribution_scorers.py
+++ b/products/posthog_ai/eval_harness/test/test_skill_distribution_scorers.py
@@ -18,8 +18,10 @@ from products.posthog_ai.evals.cli_mcp.skill_distribution_scorers import (
 
 SKILL = "querying-posthog-data"
 QUALIFIED_SKILL = f"posthog:{SKILL}"
+PROJECT_QUALIFIED_SKILL = f"project:{SKILL}"
 EXEC_EXPECTED = skill_distribution_expectations(SKILL, ["execute-sql"], "exec")
 BUNDLED_EXPECTED = skill_distribution_expectations(SKILL, ["execute-sql"], "bundled")
+PROJECT_EXEC_EXPECTED = skill_distribution_expectations(SKILL, ["execute-sql"], "exec", source="project")
 
 
 def _acp_line(update: dict) -> str:
@@ -252,3 +254,23 @@ def test_exec_batch_read_counts_as_load_but_describe_does_not(load_command: str,
 )
 def test_delivery_specific_scorers_skip_the_other_arm(scorer: Scorer, expected: dict[str, dict[str, object]]) -> None:
     assert _score(scorer, _happy_path(), expected).score is None
+
+
+@pytest.mark.parametrize(
+    "scorer,output",
+    [
+        (
+            ExpectedSkillDiscovered(),
+            _output(_exec("search", "learn -s revenue", f'{{"name":"{PROJECT_QUALIFIED_SKILL}"}}')),
+        ),
+        (
+            ExpectedSkillLoaded(),
+            _output(
+                _exec("search", "learn -s revenue", PROJECT_QUALIFIED_SKILL),
+                _exec("load", f"learn {PROJECT_QUALIFIED_SKILL}"),
+            ),
+        ),
+    ],
+)
+def test_project_source_discovery_and_loading(scorer: Scorer, output: dict[str, object]) -> None:
+    assert _score(scorer, output, PROJECT_EXEC_EXPECTED).score == 1.0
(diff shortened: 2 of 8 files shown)


services/mcp/src/hono/analytics.ts
```diff
@@ -221,3 +222,34 @@ export async function trackToolsList(toolNames: string[], state: ResolvedState):
...
+ */
+export async function trackSkillInvoked(state: ResolvedState, invocation: SkillInvocation): Promise<void> {
+    try {
+        const analyticsContext = await state.reqCtx.safelyGetAnalyticsContext(state.context)
+        const sessionUuid = await state.reqCtx.getEffectiveSessionUuid(state.requestContext)
+        const { properties, groups } = buildBaseProperties(state, analyticsContext)
+
+        getPostHogClient().capture({
+            distinctId: state.distinctId,
+            event: 'llma skill invoked',
```
skoob13:
```suggestion
            event: 'skill invoked',
```

services/mcp/src/hono/analytics.ts
```diff
@@ -221,3 +222,34 @@ export async function trackToolsList(toolNames: string[], state: ResolvedState):
...
+            groups,
+            properties: {
+                ...properties,
+                ...(sessionUuid ? { $session_id: sessionUuid } : {}),
+                skill_source: invocation.source,
+                skill_name: invocation.skill,
+                skill_identifier: `${invocation.source}:${invocation.skill}`,
+                skill_read_kind: invocation.readKind,
+                ...(invocation.path ? { skill_file_path: invocation.path } : {}),
```
skoob13: Make sure nested file reads are not tracked

services/mcp/src/lib/instructions-formatter.ts
```diff
@@ -70,13 +71,31 @@ export class InstructionsFormatter {
...
-    buildExecInstructions(ctx: InstructionsContext): string {
-        return this.compose([COMPACT_INSTRUCTIONS], ctx, { compact: true })
+    buildExecInstructions(ctx: InstructionsContext, opts: { skillsEnabled?: boolean } = {}): string {
+        const instructions = this.compose([COMPACT_INSTRUCTIONS], ctx, { compact: true })
+        if (!opts.skillsEnabled) {
+            return instructions
+        }
+        // Session-start reinforcement for clients that surface instructions — the
+        // description-level mandate alone doesn't reach agents that skim the tool list.
+        return `SKILLS FIRST: run \`learn -s "<task keywords>"\` on the posthog exec tool and load matching skills before any PostHog product or data work; never answer PostHog-behavior questions from cloned source or the web.\n\n${instructions}`
```
skoob13: Does it fit into 2048 characters? I doubt so. It's enough to have this in the command prompt.

services/mcp/src/lib/instructions-formatter.ts
```diff
@@ -70,13 +71,31 @@ export class InstructionsFormatter {
...
+     *  uncapped top-level `description`, so the skills text costs no schema budget.
+     *  The skills mandate LEADS the description: it is the only signal that reaches
+     *  an agent before its first tool call, and agents that answer PostHog-behavior
+     *  questions by cloning the public repo never make a call for the gate to catch. */
+    buildExecToolDescription(opts: { skillsEnabled?: boolean } = {}): string {
+        const blurb = EXEC_TOOL_BLURB.trim()
+        if (!opts.skillsEnabled) {
+            return blurb
+        }
+        const skillsFirst =
```
skoob13: Move out to templates

PR #71599, opened 2026-07-16
fix(batch-exports): resume S3 staging reads from last offset on retry

rossgray:
## Problem

The batch export producer streams Arrow IPC files from the internal S3 staging area onto a queue for destination writers. Each file read is wrapped in a retry with backoff.

On a mid-file failure (e.g. a connection reset), the retry re-opened the file with a plain GET from byte 0 and re-emitted any batches already queued, so customers could receive duplicate rows.

## Changes

Track how far we've read into each file and resume from there on retry.

- `asyncpa.py`: `AsyncMessageReader` exposes `bytes_consumed` (always at an IPC message boundary); `AsyncRecordBatchReader` accepts a pre-parsed `schema` so a resumed stream, which starts mid-file with no schema message, parses directly.
- `producer.py`: per-key `S3FileResumeState` (offset, cached schema, object size, ETag) survives across retries in the read closure. On resume it issues a ranged GET (`Range: bytes=<offset>-`) pinned with `IfMatch` to the original ETag, so a changed object 412s instead of yielding garbage. The offset only advances after a batch's slices are all enqueued, so a batch is never half-counted.

This relies on stage files being plain uncompressed Arrow IPC streams (ClickHouse `FORMAT ArrowStream`), so message-boundary offsets are valid resume points. Side benefit: `records_produced`/`bytes_produced` no longer double-count on retried files.

I (Claude, agent-assisted) did not test manually. Automated coverage:

- `test_asyncpa.py` — reader resumes from a captured byte offset with a seeded schema (parameterized, incl. resuming exactly at the EOS marker).
- `test_producer.py` (new) — against a fake S3 client: single mid-file failure resumes with a ranged, ETag-pinned GET (and a clean full re-read when the failure precedes any batch); consecutive failures resume from successive boundaries; the fully-consumed short-circuit. Plus one e2e test that runs the real staging activity into real MinIO and reads it back through the real producer with a mid-file failure injected, asserting every row is delivered exactly once.

I confirmed the tests catch the bug: against the unfixed producer, the mid-file cases fail on the exactly-once assertion (a batch is delivered twice); with the fix they pass.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Ross directed this work; I (Claude, via PostHog Code) explored, planned, and implemented it. Skill invoked: `/writing-tests`.

Notes for reviewers:

- Rejected the simpler "re-read from 0 and skip N emitted batches" alternative: it re-downloads the consumed prefix each retry and makes no forward progress on a persistently flaky connection. Ranged resume needs only a small `asyncpa.py` change (a byte counter + optional schema arg).

---

### DIFF
--- a/products/batch_exports/backend/temporal/pipeline/producer.py
+++ b/products/batch_exports/backend/temporal/pipeline/producer.py
@@ -1,8 +1,11 @@
 import typing
 import asyncio
+import collections
+import dataclasses
 
 from django.conf import settings
 
+import pyarrow as pa
 from aiobotocore.response import StreamingBody
 from opentelemetry import trace
 
@@ -20,6 +23,22 @@ if typing.TYPE_CHECKING:
 
 LOGGER = get_write_only_logger(__name__)
 TRACER = trace.get_tracer(__name__)
+
+
+@dataclasses.dataclass
+class S3FileResumeState:
+    """Tracks how far into an S3 staging file we have fully enqueued record batches.
+
+    Used to resume from the last record batch boundary when retrying a failed
+    read, instead of re-reading (and re-emitting) the whole file.
+    """
+
+    # Absolute byte offset of the next unread IPC message.
+    offset: int = 0
+    # Cached schema, so a resumed stream (which lacks the schema message) can be parsed.
+    schema: pa.Schema | None = None
+    # Full object size, from the first (non-ranged) GET.
+    object_size: int | None = None
 
 
 class Producer:
@@ -96,7 +115,6 @@ class Producer:
                     if not keys:
                         return
 
-                    # Read in batches
                     await self._stream_record_batches_from_s3(
                         s3_client, keys, queue, max_record_batch_size_bytes, min_records_per_batch
                     )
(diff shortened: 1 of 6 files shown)


products/batch_exports/backend/tests/temporal/pipeline/test_producer.py
```diff
@@ -0,0 +1,133 @@
...
+        self.get_object_requests: list[tuple[str, str | None]] = []
+
+    async def get_object(self, Bucket: str, Key: str, Range: str | None = None) -> dict:
+        self.get_object_requests.append((Key, Range))
+        self._get_counts[Key] += 1
+
+        offset = int(Range.removeprefix("bytes=").removesuffix("-")) if Range is not None else 0
+        assert offset < len(self._objects[Key]), "Invalid range requested"
+
+        fail_after_bytes = self._fail_first_read_after_bytes if self._get_counts[Key] == 1 else None
```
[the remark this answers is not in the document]
rossgray: yeah suppose adding another test here doesn't hurt

tomasfarias:
Surprised to see you did end up going with the bytes thing rather than just skipping record batches based on our last convo.

Will review later today.

rossgray:
> Surprised to see you did end up going with the bytes thing rather than just skipping record batches based on our last convo.

Yeah, I planned through the whole thing with Claude which helped a lot (particularly around the Arrow IPC format and how to request a range of data using the `Range` argument in `s3_client.get_object`)

products/batch_exports/backend/temporal/pipeline/producer.py
```diff
@@ -123,6 +145,45 @@ async def _list_keys(self, s3_client: "S3Client", stage_folder: str) -> list[str
         self.logger.info(f"Producer found {len(keys)} files in S3 stage, with prefix '{stage_folder}'")
         return keys
 
+    async def _open_staging_file(
```
tomasfarias: I think we should split up this function into two as there is one mode that uses the resume state and one mode that doesn't. So, just to make it clear, there is `_open_staging_file` (the else branch + returns a state object) and `_resume_staging_file` (or a different name, which takes a state object, and doesn't return one).

I don't like the side effect of updating the state object only when the file is new. I prefer clearer input/output boundaries.

products/batch_exports/backend/temporal/pipeline/producer.py
```diff
@@ -147,6 +218,10 @@ async def stream_from_s3_file(key):
                     self.records_produced += record_batch_slice.num_rows
                     self.bytes_produced += record_batch_slice.nbytes
 
+                # Only advance the resume point once every slice of this batch is enqueued.
+                state.offset = base_offset + reader.bytes_consumed
+                state.schema = reader.schema
+
```
tomasfarias, line 244: _enqueued_ does not mean fully processed...

tomasfarias: If -for some reason- an error is raised when reading the next batch, but before the previous batch from this file has been fully processed, then we would be missing it on a retry.

But that means that downstream should fail processing the batch for it to be missed, as there is no intermediate state: Batches are either processed successfully or they fail the entire batch export. 

Given that there cannot be a partial failure downstream, as scary as this looks, I think it's safe.

tomasfarias (commented):
I'd make just 1 change around how the state is created/preserved. I don't like that we can have an empty state before the download even starts. That requires a bunch of `None` or `0` checks.

I'll open a PR on this.

tomasfarias:
Here's a lil PR with my review comments applied: https://github.com/PostHog/posthog/pull/73579. If you are okay, then we can just merge on Monday. Also fixes that stupid mypy type error (by just ignoring it).

products/batch_exports/backend/temporal/pipeline/producer.py
rossgray: yeah that's a fair point

tomasfarias (approved):

PR #71611, opened 2026-07-16
feat(flags): add etag support for remote config endpoint

### LINKED ISSUE #60314: feat(flags): Add etag support for remote config flags
### Bug Description

This is a follow-up to #60313

We should ensure that client and server for serving up remote config flags support etags like we do for `/flags/definitions`

gustavohstrassburger:
## Problem

SDKs poll `GET /api/projects/{project_id}/feature_flags/{key}/remote_config` for payloads that rarely change, and every poll ships the full body. `/flags/definitions` already supports `If-None-Match`/304 conditional requests; the remote config endpoint (now served by the Rust feature-flags service) should too.

Closes #60314

## Changes

- The Rust `remote_config` handler now answers `If-None-Match` with a 304 when the client's etag matches, and attaches `ETag` (weak) plus `Cache-Control: private, no-cache` and `Vary: Authorization` to all responses (200, empty-body 200, and 304), so caches revalidate on every reuse and key entries by credential.
- Unlike `/flags/definitions` there is no cache (and no Redis-stored etag) backing this endpoint, so the etag is content-derived: sha256 of the exact response body, truncated to 16 hex chars, computed after payload resolution. Same etag shape as HyperCache's `_compute_etag`, so SDKs see one format across both endpoints.
- Deriving the etag from the resolved body makes it caller-dependent by construction. For encrypted flags, a secret-key caller's redacted body and a personal-key caller's decrypted body carry different etags, so a 304 never validates one credential class's cached body against the other's.
- A match saves the body transfer, not the DB read, since the payload has to be resolved to compute the etag.
- Reuses the sibling endpoint's etag helpers (header parsing, weak formatting, 304 builder) by widening them to `pub(crate)`, and adds a `flags_remote_config_etag_total` counter with hit/miss/none labels.

> [!NOTE]
> Server side only. SDK clients still need to send `If-None-Match` from `getRemoteConfigPayload()` to benefit, which is per-SDK-repo work.

Automated tests, each catching a regression no existing test covered:

- `test_remote_config_etag_roundtrip`: 200 carries a weak etag and revalidation cache headers, a matching `If-None-Match` gets 304 with no body, a stale one gets a full 200. Catches the 304 path breaking or the headers being dropped.
- `test_remote_config_encrypted_etag_differs_by_credential`: redacted and decrypted responses for the same encrypted flag carry different etags, and the redacted etag does not 304 a decrypting caller. Catches an etag computed from the stored row or shared across credential classes, which would validate a cached redacted body where plaintext should be served.
- Extended `test_remote_config_empty_payload_returns_empty_body` to assert the empty-body branch also carries an etag and can 304. Catches that branch losing etag support, since it builds its response separately.

Ran locally: `cargo check -p feature-flags`, `cargo clippy -p feature-flags --tests` (clean, one pre-existing warning on master), `cargo fmt`, and the crate's unit tests (35 passed). I could not run the integration tests locally because the sandbox has no Postgres/Redis, so they need CI to verify.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (PostHog Code cloud task) from issue #60314. Skills invoked: `/writing-tests`. Key decision: a content-derived per-request etag instead of a Redis-stored one like `/flags/definitions`, because no cache backs this endpoint and the response varies by credential class for encrypted flags, which a single stored etag can't represent. Considered and rejected adding a HyperCache for per-flag payloads as out of scope.

@greptile

---

### DIFF
--- a/rust/feature-flags/src/api/remote_config.rs
+++ b/rust/feature-flags/src/api/remote_config.rs
@@ -15,12 +15,20 @@
 //! feature flag access, even though this endpoint can return decrypted payloads. OAuth
 //! access tokens are also not accepted — any non-`phs_` bearer goes through personal-key
 //! validation and gets 401; only `phs_` and `phx_` credentials work.
+//!
+//! Supports `If-None-Match` conditional requests like `/flags/definitions`, but with a
+//! content-derived ETag: no cache backs this endpoint (the payload is read from Postgres
+//! per request), so the etag is a hash of the exact response body computed after payload
+//! resolution. That makes it caller-dependent by construction — a secret-key caller's
+//! redacted body and a personal-key caller's decrypted body hash to different etags, so a
+//! 304 never validates one credential class's cached body against the other's. A match
+//! saves the body transfer, not the DB read.
 
 use crate::{
     api::{auth, errors::FlagError, flag_definitions},
     database::get_connection_with_metrics,
     flags::flag_payload_decryptor::REDACTED_PAYLOAD_VALUE,
-    metrics::consts::REMOTE_CONFIG_AUTH_COUNTER,
+    metrics::consts::{REMOTE_CONFIG_AUTH_COUNTER, REMOTE_CONFIG_ETAG_COUNTER},
     router::State as AppState,
     team::team_models::Team,
 };
@@ -28,11 +36,12 @@ use axum::{
     debug_handler,
     extract::{Path, Query, State},
     http::{HeaderMap, Method, StatusCode},
-    response::{IntoResponse, Json, Response},
+    response::{IntoResponse, Response},
 };
 use common_metrics::inc;
 use serde::Deserialize;
 use serde_json::Value;
+use sha2::{Digest, Sha256};
 use tracing::warn;
 
 /// Query params. SDKs pass `?token=phc_...` (the project key) and call this with `@current`
@@ -192,10 +201,63 @@ pub async fn remote_config(
     // Django applies `or None` to the final value and renders None as an empty body, not
     // the JSON literal `null`. Apply the falsy check after decryption so an empty decrypted
     // string nulls out too.
-    match payload.filter(|v| !is_falsy(v)) {
-        Some(v) => Ok(Json(v).into_response()),
-        None => Ok(empty_ok_no_content_type()),
+    let payload = payload.filter(|v| !is_falsy(v));
+
+    // `Value::to_string` produces the same compact JSON `Json(v)` would serialize, so the
+    // etag hashes the exact bytes the client receives (empty payloads hash the empty body).
+    let body = payload.as_ref().map(Value::to_string).unwrap_or_default();
+    let current_etag = compute_etag(&body);
+
+    let client_etag = flag_definitions::extract_etag_from_header(headers.get("if-none-match"));
+    if client_etag.as_deref() == Some(current_etag.as_str()) {
+        inc(
+            REMOTE_CONFIG_ETAG_COUNTER,
+            &[("result".to_string(), "hit".to_string())],
+            1,
+        );
+        return Ok(flag_definitions::not_modified_response(&current_etag));
     }
+    inc(
+        REMOTE_CONFIG_ETAG_COUNTER,
+        &[(
+            "result".to_string(),
+            if client_etag.is_some() {
+                "miss"
+            } else {
+                "none"
+            }
+            .to_string(),
+        )],
+        1,
+    );
+
+    match payload {
+        Some(_) => Ok(json_ok_with_etag(body, &current_etag)),
+        None => Ok(empty_ok_no_content_type(&current_etag)),
+    }
+}
+
+/// Content-derived ETag: sha256 of the exact response body, truncated to 16 hex chars —
+/// the same shape HyperCache's `_compute_etag` produces for `/flags/definitions`, so SDKs
+/// see one etag format across both endpoints.
+fn compute_etag(body: &str) -> String {
+    let digest = Sha256::digest(body.as_bytes());
+    hex::encode(digest)[..16].to_string()
+}
+
+/// 200 with a pre-serialized JSON body plus ETag and Cache-Control headers, mirroring the
+/// sibling endpoint's `ok_response_with_etag`.
+fn json_ok_with_etag(body: String, etag: &str) -> Response {
+    (
+        StatusCode::OK,
+        [
+            ("content-type", "application/json".to_string()),
+            ("etag", flag_definitions::format_weak_etag(etag)),
+            ("cache-control", "private, must-revalidate".to_string()),
+        ],
+        body,
+    )
+        .into_response()
 }
 
 /// Decrypts the stored ciphertext on the personal-key path. Returns `None` when there is no
@@ -223,9 +285,17 @@ fn resolve_decrypted_payload(
 }
 
 /// 200 with an empty body and no Content-Type (NOT a JSON `null`), matching DRF's `Response(None)`:
-/// the renderer emits no bytes and DRF then deletes the Content-Type header.
-fn empty_ok_no_content_type() -> Response {
-    StatusCode::OK.into_response()
+/// the renderer emits no bytes and DRF then deletes the Content-Type header. Still carries the
+/// ETag/Cache-Control pair so a client polling a not-yet-set payload can 304 too.
+fn empty_ok_no_content_type(etag: &str) -> Response {
+    (
+        StatusCode::OK,
+        [
+            ("etag", flag_definitions::format_weak_etag(etag)),
+            ("cache-control", "private, must-revalidate".to_string()),
+        ],
+    )
+        .into_response()
 }
 
 /// Mirrors Python truthiness for `payloads.get("true") or None`. In practice the payload
--- a/rust/feature-flags/src/api/remote_config.rs
+++ b/rust/feature-flags/src/api/remote_config.rs
@@ -49,6 +49,12 @@ use serde::Deserialize;
 use serde_json::Value;
 use tracing::warn;
 
+/// The body varies by `Authorization` at the same URL (decrypted vs redacted), so caches
+/// must revalidate on every reuse (`no-cache`, stronger than the sibling endpoint's
+/// `must-revalidate`) and key entries by credential (`Vary: Authorization`).
+const CACHE_CONTROL: &str = "private, no-cache";
+const VARY: &str = "Authorization";
+
 /// Query params. SDKs pass `?token=phc_...` (the project key) and call this with `@current`
 /// as the URL segment; the token resolves the project, matching Django. `api_key` is Django's
 /// accepted alias for `token` (see `get_token`).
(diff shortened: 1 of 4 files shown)


rust/feature-flags/src/api/remote_config.rs
```diff
@@ -192,10 +201,63 @@ pub async fn remote_config(
...
+    }
+}
+
+/// Content-derived ETag: sha256 of the exact response body, truncated to 16 hex chars —
+/// the same shape HyperCache's `_compute_etag` produces for `/flags/definitions`, so SDKs
+/// see one etag format across both endpoints.
+fn compute_etag(body: &str) -> String {
+    let digest = Sha256::digest(body.as_bytes());
+    hex::encode(digest)[..16].to_string()
+}
```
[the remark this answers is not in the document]
gustavohstrassburger: Valid — fixed in 60b272e: the local `compute_etag` now delegates to `common_hypercache::writer::compute_etag`, and the `sha2` import is gone.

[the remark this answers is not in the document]
gustavohstrassburger: Fixed in 60b272e by reusing `common_hypercache::writer::compute_etag` (which encodes only the first 8 digest bytes into a pre-sized string), removing the slice and extra allocation.

rust/feature-flags/src/metrics/consts.rs
```diff
@@ -329,6 +329,12 @@ pub const REMOTE_CONFIG_REQUESTS_COUNTER: &str = "flags_remote_config_requests_t
 // split decides redact-vs-decrypt, so the mix is worth watching during the phase 2/3 cutover.
 pub const REMOTE_CONFIG_AUTH_COUNTER: &str = "flags_remote_config_auth_total";
 
+// Remote config ETag metrics
+// Labels: result (hit = 304, miss = 200 with stale etag, none = 200 without etag). Unlike flag
+// definitions there is no redis_error case: no cache backs this endpoint, so the etag is
+// content-derived per request and can always be computed.
```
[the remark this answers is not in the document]
gustavohstrassburger: Fixed in 60b272e — the comment now reads `none = request sent no If-None-Match` and notes that 200s always carry an etag on this endpoint.

gustavohstrassburger:
valid — the `compute_etag` duplication finding was correct and has been fixed in 60b272e by reusing `common_hypercache::writer::compute_etag`.

rust/feature-flags/src/api/remote_config.rs
```diff
@@ -192,10 +200,62 @@ pub async fn remote_config(
...
-        None => Ok(empty_ok_no_content_type()),
+    let payload = payload.filter(|v| !is_falsy(v));
+
+    // `Value::to_string` produces the same compact JSON `Json(v)` would serialize, so the
+    // etag hashes the exact bytes the client receives (empty payloads hash the empty body).
+    let body = payload.as_ref().map(Value::to_string).unwrap_or_default();
+    let current_etag = compute_etag(&body);
+
+    let client_etag = flag_definitions::extract_etag_from_header(headers.get("if-none-match"));
+    if client_etag.as_deref() == Some(current_etag.as_str()) {
```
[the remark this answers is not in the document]
gustavohstrassburger: Valid, but this is a pre-existing gap in the shared parser rather than something this PR introduces: `extract_etag_from_header` is the same helper `/flags/definitions` uses, and its own tests explicitly document that `*` and comma-separated lists are treated as a single tag and fall back to a 200 (see `test_extract_etag_from_header_multiple_etags_no_special_handling`). The fallback is safe (full response, never a wrong 304), and in practice our SDK clients send back the single weak etag we emitted. Worth fixing with an RFC 7232 list-aware matcher, but in a follow-up covering both endpoints so `/flags/definitions` and `/remote_config` stay consistent.

rust/feature-flags/src/api/remote_config.rs
```diff
@@ -192,10 +200,62 @@ pub async fn remote_config(
...
+
+/// 200 with a pre-serialized JSON body plus ETag and Cache-Control headers, mirroring the
+/// sibling endpoint's `ok_response_with_etag`.
+fn json_ok_with_etag(body: String, etag: &str) -> Response {
+    (
+        StatusCode::OK,
+        [
+            ("content-type", "application/json".to_string()),
+            ("etag", flag_definitions::format_weak_etag(etag)),
+            ("cache-control", "private, must-revalidate".to_string()),
```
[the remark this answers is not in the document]
gustavohstrassburger: Valid — fixed in 8cab14d. All three responses (200, empty 200, 304) now send `Cache-Control: private, no-cache` so caches revalidate on every reuse rather than only once stale, plus `Vary: Authorization` so entries are keyed by credential and a personal-key body can never be replayed to a secret-key caller. The 304 uses a local builder instead of the sibling endpoint's so it doesn't re-advertise the weaker `must-revalidate` policy when a cache updates stored headers from a 304.

rust/feature-flags/src/api/remote_config.rs
```diff
@@ -192,10 +206,83 @@ pub async fn remote_config(
...
+        ],
+        body,
+    )
+        .into_response()
+}
+
+/// 304 carrying the same ETag/Cache-Control/Vary trio as the 200s — not the sibling
+/// endpoint's `not_modified_response`, whose `must-revalidate` would weaken the stored
+/// entry's policy when a cache updates its headers from the 304.
+fn not_modified(etag: &str) -> Response {
```
haacked, line 283: `suggestion`: `json_ok_with_etag`, `not_modified`, and `empty_ok_no_content_type` each build the etag/cache-control/vary header trio from scratch. `not_modified` and `empty_ok_no_content_type` are byte-identical except for the status code, so if an edit tweaks one and misses the other, the credential-isolation guarantee (the `no-cache` plus `Vary: Authorization` fix) silently breaks on one response path.

Pull the trio into a shared helper so they can't drift:

```rust
fn revalidation_headers(etag: &str) -> [(&'static str, String); 3] {
    [
        ("etag", flag_definitions::format_weak_etag(etag)),
        ("cache-control", CACHE_CONTROL.to_string()),
        ("vary", VARY.to_string()),
    ]
}

fn not_modified(etag: &str) -> Response {
    (StatusCode::NOT_MODIFIED, revalidation_headers(etag)).into_response()
}

fn empty_ok_no_content_type(etag: &str) -> Response {
    (StatusCode::OK, revalidation_headers(etag)).into_response()
}
```

`json_ok_with_etag` can destructure the array and prepend its own `content-type` header.

rust/feature-flags/tests/test_remote_config.rs
```diff
@@ -839,8 +839,204 @@ async fn test_remote_config_empty_payload_returns_empty_body() {
...
 
     assert_eq!(response.status(), 200);
+    // Empty responses still carry an etag so a client polling a not-yet-set payload can 304.
+    let etag = response
+        .headers()
+        .get("etag")
+        .expect("empty response should carry an etag")
+        .to_str()
+        .unwrap()
+        .to_string();
```
haacked, line 849: `suggestion`: This test asserts the empty 200 carries an etag but never checks `cache-control` or `vary`, the headers that prevent a private cache from serving one credential class's empty body to another. If those headers were dropped from just `empty_ok_no_content_type`, this test would still pass.

Add the same assertions the roundtrip test uses right after capturing the etag:

```rust
assert_eq!(
    response.headers().get("cache-control").unwrap().to_str().unwrap(),
    "private, no-cache"
);
assert_eq!(
    response.headers().get("vary").unwrap().to_str().unwrap(),
    "Authorization"
);
```

rust/feature-flags/tests/test_remote_config.rs
```diff
@@ -839,8 +839,204 @@ async fn test_remote_config_empty_payload_returns_empty_body() {
...
+            .get("cache-control")
+            .unwrap()
+            .to_str()
+            .unwrap(),
+        "private, no-cache"
+    );
+    assert_eq!(
+        response.headers().get("vary").unwrap().to_str().unwrap(),
+        "Authorization"
+    );
```
haacked, line 931: `nit`: This test checks `cache-control` and `vary` on the 200 but not `content-type`, which now requires an explicit override since the body switched from `Json(v)` to a raw `String` (which axum defaults to `text/plain`).

```rust
assert_eq!(
    response.headers().get("content-type").unwrap().to_str().unwrap(),
    "application/json"
);
```

rust/feature-flags/src/api/remote_config.rs
```diff
@@ -192,10 +206,83 @@ pub async fn remote_config(
...
+/// that produces `/flags/definitions` etags, so SDKs see one etag format across both
+/// endpoints.
+fn compute_etag(body: &str) -> String {
+    common_hypercache::writer::compute_etag(body)
+}
+
+/// The body varies by `Authorization` at the same URL (decrypted vs redacted), so caches
+/// must revalidate on every reuse (`no-cache`, stronger than the sibling endpoint's
+/// `must-revalidate`) and key entries by credential (`Vary: Authorization`).
+const CACHE_CONTROL: &str = "private, no-cache";
```
haacked: `nit`: `CACHE_CONTROL` and `VARY` sit between `compute_etag` and `json_ok_with_etag` rather than with the other module-level items near the top. Move them just under the `use` block so constants don't get scattered between functions.

rust/feature-flags/tests/test_remote_config.rs
haacked, line 849: `nit`: `empty_ok_no_content_type`'s doc comment says the empty 200 carries no `content-type`, matching DRF, but the test doesn't verify this. Add `assert!(response.headers().get("content-type").is_none());` right after the etag check to catch regressions.

haacked (approved):
Nice fix! Solid test coverage on the credential-isolation edge case. A couple non-blocking suggestions inline.

rust/feature-flags/src/api/remote_config.rs
gustavohstrassburger: Done in d4e9fb5 — extracted `revalidation_headers` and all three builders use it; `json_ok_with_etag` prepends its content-type as a separate header part.

rust/feature-flags/tests/test_remote_config.rs
gustavohstrassburger: Done in d4e9fb5 — the empty-payload test now asserts `cache-control: private, no-cache` and `vary: Authorization`.

rust/feature-flags/tests/test_remote_config.rs
gustavohstrassburger: Done in d4e9fb5 — the roundtrip test now asserts `content-type: application/json` on the 200.

rust/feature-flags/src/api/remote_config.rs
gustavohstrassburger: Done in d4e9fb5 — `CACHE_CONTROL`/`VARY` now sit directly under the `use` block.

rust/feature-flags/tests/test_remote_config.rs
gustavohstrassburger: Done in d4e9fb5 — added `assert!(response.headers().get("content-type").is_none())` to the empty-payload test.

PR #71614, opened 2026-07-16
fix(hogql): guard groups-join prefilter against group-type-name aliases

xljones:
## Problem

A HogQL query over `events` that selects a group-type-name property crashes with a `RecursionError` once `{filters}` is expanded, but works fine without `{filters}`:

```sql
SELECT organization.properties.name AS organization_name, count()
FROM events
WHERE {filters} AND organization.properties.name IS NOT NULL
GROUP BY organization_name
```

`organization` is a per-project `FieldTraverser` alias to `group_0`. `{filters}` injects a `timestamp` bound into the WHERE, which switches on the groups-join prefilter. That prefilter clones the outer WHERE into an inner `SELECT $group_0 FROM events WHERE ...` subquery, and its guard is supposed to skip cases where the WHERE references a lazy join (cloning one re-triggers the same join and blows the stack). The guard only listed the physical `group_0`..`group_4` names, so the friendly alias slipped past.

## Changes

Derive the group-type-name aliases from the events table at query time and union them into the guarded set, so all current and future group types are treated like `group_N`. Over-guarding is safe: a false positive only skips a perf optimization, never changes results.

Workaround before this ships: alias the events table, which bypasses the prefilter.

```sql
SELECT organization.properties.name AS organization_name, count()
FROM events AS e
WHERE {filters} AND organization.properties.name IS NOT NULL
GROUP BY organization_name
```

Parameterized the lazy-join-guard test to add the group-type-name case. It fails with `RecursionError` on the pre-fix guard and passes after. Ran `test_groups_join_prefilter`, `test_groups`, `test_filters` locally, all green. No manual UI testing.

## 🤖 Agent context

Driven by @xljones, implemented by Claude (PostHog Code). Invoked `/writing-tests`. Chose to read the `FieldTraverser` aliases off the events table rather than hardcode names, since group types are per-project.

---


posthog/hogql/database/schema/groups.py
```diff
@@ -269,10 +287,8 @@ def _outer_events_prefilter(node: SelectQuery):
 
     if not _references_timestamp(where):
         return None
-    if any(
-        chain and isinstance(chain[0], str) and chain[0] in EVENTS_LAZY_JOIN_ALIASES
-        for chain in find_field_chains(where)
-    ):
+    guarded_aliases = EVENTS_LAZY_JOIN_ALIASES | _group_type_name_aliases(context)
+    if any(chain and isinstance(chain[0], str) and chain[0] in guarded_aliases for chain in find_field_chains(where)):
         return None
```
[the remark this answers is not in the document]
xljones: resolved in [`c4ff3d1` (this PR)](https://github.com/PostHog/posthog/pull/71614/commits/c4ff3d13802bea27dcd825e9db4e17ad7d5daa34).

posthog/hogql/database/schema/groups.py
```diff
@@ -211,6 +212,10 @@ def join_with_group_n_table(
 # resolve the same lazy join inside our inner subquery — producing unbounded recursion or a
 # `ResolutionError: Select query must have a type`. Skip the optimization in that case;
 # we'd rather pay the original groups-hash-table cost than crash.
+#
+# Per-project group-type-name aliases (e.g. `organization`, `company`) are FieldTraversers to
+# `group_N` added dynamically per project, so they're not listed here — `_group_type_name_aliases`
+# reads them off the events table at query time and unions them into the guarded set.
```
arthurdedeus: quite an essay we have here, huh?

posthog/hogql/database/schema/groups.py
```diff
@@ -233,7 +238,18 @@ def join_with_group_n_table(
...
-def _outer_events_prefilter(node: SelectQuery):
+def _group_type_name_aliases(context: HogQLContext) -> frozenset[str]:
+    """Per-project group-type-name aliases on the events table (FieldTraversers to `group_N`)."""
+    if context.database is None:
+        return frozenset()
+    try:
+        events_table = context.database.get_table(["events"])
+    except Exception:
+        return frozenset()
+    return frozenset(name for name, field in events_table.fields.items() if isinstance(field, FieldTraverser))
```
arthurdedeus: this returns all `FieldTraversers` in the events table's `fields`, right? not only `group_n` aliases.

posthog/hogql/database/schema/groups.py
```diff
@@ -269,14 +287,21 @@ def _outer_events_prefilter(node: SelectQuery):
 
     if not _references_timestamp(where):
         return None
-    if any(
-        chain and isinstance(chain[0], str) and chain[0] in EVENTS_LAZY_JOIN_ALIASES
-        for chain in find_field_chains(where)
-    ):
+    guarded_aliases = EVENTS_LAZY_JOIN_ALIASES | _group_type_name_aliases(context)
```
arthurdedeus: this guard should happen inside `_group_type_name_aliases`, otherwise the method is misleading (see commend I've added there). `_group_type_name_aliases` is not used in isolation anywhere, so we'd always want to intersect it with the lazy join aliases.

arthurdedeus (approved):
The fix looks sound, good stuff.

One thing concerning me, though, is the growing complexity of this code.. it's getting harder and harder to read and understand.. the comments aren't very helpful either (to me, at least. hope it is for the robots). No need to change anything in this regard, though, just something for us to keep in mind.

posthog/hogql/database/schema/groups.py
xljones: agree wholeheartedly on that. made a [change](https://github.com/PostHog/posthog/pull/71614/changes/8f4a36aa65ab8a137880413aac56c9e3bb161c47) to move it into one method that returns the whole list rather than this half and half approach, good call.

posthog/hogql/database/schema/groups.py
xljones: true – the union with the EVENTS_LAZY_JOIN_ALIASES should dedupe this now in the reworked function

posthog/hogql/database/schema/groups.py
xljones: verbose af. made some tweaks here with my robot pal. hopefully this makes it a bit clearer and useful with a real example of what i'm trying to achieve.

PR #71617, opened 2026-07-16
chore(qa-team): prompt-cache-aware agent launch protocol

robbie-c:
## Problem

I had a go at optimizing the /qa-team skill. Previously it was not KV-cache aware, but with some tweaks I was able to dramatically reduce the startup costs by making sure that each agent launched with a long shared prefix (the entire diff + prompt). Dogfooding this PR with Fable 5 high, a review went from ~$9.50 to ~$5.60, startup costs went from ~$5.30 to ~$1.40.

Claude continues:

The `/qa-team` skill fans out ~10 review agents over the same diff, but its launch protocol defeated prompt caching entirely. Each agent's prompt started with divergent persona text and had the full diff pasted in by hand. Prompt caching is strict-prefix, so one divergent byte at the top meant every agent re-ingested the ~25k tokens of shared harness context plus the entire diff at cache-write prices, and the orchestrator re-emitted the diff once per agent at output-token prices.

We measured this in a live session: follower agents got zero additional cache hits regardless of how the prompt template was ordered, because any divergent byte in the single prompt block spoils the hit for everything around it. For a 10-agent run over a 20k-token diff the launch phase costs roughly $19.50, nearly all of it avoidable.

## Changes

Step 3 of the skill is now a cache-aware launch protocol, with the moving parts shipped as real files in the skill's `scripts/` directory instead of inline snippets the orchestrator would re-generate each run:

- **One byte-identical prompt for every agent, with the diff embedded.** `scripts/build_launch_scripts.js` JSON-encodes the shared review prompt (claim instruction, changed files, commits, full diff) into two generated Workflow scripts in the run directory, so the diff rides in the shared cached prefix, never passes through the orchestrating model as output tokens, and the prompts cannot drift apart by a stray byte. Diffs over ~200 KB automatically fall back to a read-from-disk instruction.
- **Per-agent divergence arrives via a tool result, not the prompt.** Each agent's first action is `scripts/claim_persona.sh`, which atomically (`mv`) hands it the next persona from a queue on disk. Tool results come after the shared prefix, so the cache hit survives.
- **Two-phase staggered launch, no warm-up agent.** `launch_first.js` starts reviewer 1 alone in the background; its persona claim appearing in `claimed/` signals that its first request finished (~15s) and the prefix is cached; `launch_rest.js` then fans out the remaining reviewers in parallel. Reviewer 1 keeps reviewing throughout, so all reviewers run concurrently apart from the brief warm-up window.
- **Failure visibility:** failed reviewers surface as explicit sentinel entries, a reconciliation step compares reviews against the persona queue before synthesis, and the claim script distinguishes queue exhaustion from a broken run dir.
- **Fallback documented** for environments without the Workflow tool (direct Agent-tool launch with the same claim-based stagger).

### Token cost comparison

Measured per-agent first requests (exact `cache_creation` / `cache_read` token counts from subagent transcripts), and the worked example of 10 agents reviewing a 20k-token diff. Pricing used: cache read 0.1× base input, 1-hour-TTL cache write 2× base input, output tokens 5× base input.

| Launch protocol | Diff delivery | Follower's first request (measured) | Launch cost, 10 agents × 20k-token diff |
| --- | --- | --- | --- |
| Original (per-agent prompts) | orchestrator re-types it into all N prompts (output tokens) | read 18.9k, re-write 24.8k + diff | **~$19.50** (≈ $10 of it orchestrator output) |
| Intermediate: claim script + per-agent diff read | each agent reads `diff.patch` (tool result, write-priced) | read 43.1k, write ~0 at launch; diff written per agent afterwards | **~$5.40** |
| Final: builder-embedded shared prompt | baked into the byte-identical prompt by `build_launch_scripts.js` (no model tokens) | read full prefix incl. diff, write 0 (42,476/0 in calibration; 48,782/0 live) | **~$1.90** |

**Whole-run cost, for honesty:** launch is only part of total spend — the review work itself (each agent reading surrounding code across 5–8 requests and writing findings) costs the same under any protocol, so total savings are smaller than the launch-only comparison suggests. The end-to-end dogfood run (6 agents, ~5k-token diff) landed at:

| 6 agents, ~5k-token diff | Launch | Review work | Total |
| --- | --- | --- | --- |
| Old protocol (modeled) | ~$5.30 | ~$4.20 | **~$9.50** |
| This PR, as run (includes the since-fixed divergence bug) | ~$2.30 | ~$4.20 | **~$6.50** |
| This PR, with the fix | ~$1.40 | ~$4.20 | **~$5.60** |

Agent-side totals measured from transcripts: 38 requests, 1.94M cache-read tokens, 216k cache-write tokens, ~5k output tokens. Transcripts appear to undercount some output/thinking tokens, so absolute numbers are floors (true total likely ~$7–8 as run), and orchestrator-side work (persona files, synthesis) adds ~$1–2 on top under either protocol. The protocol's savings scale with agent count and diff size — at 10 agents × 20k-token diff the launch gap is $19.50 vs $1.90 against $6–17 of review work, i.e. roughly a half to two-thirds off the total, not 10×.

The live dogfood run also demonstrated the failure mode the protocol exists to prevent: a builder bug made the follower prompts diverge from reviewer 1's by one token, and the five followers promptly stopped reading his 48,797-token cache entry — two of them re-wrote 31,111 tokens each in a race. That bug is fixed on this branch; with byte-identity restored, followers read everything.

Before:

```mermaid
flowchart LR
  ORC{{Orchestrator}} -->|"re-types persona + diff per agent"| R{{"Reviewers 1..N"}}
  R --> MISS["divergent prefix: each agent re-writes ~25k context + diff"]
  class ORC,R phBlue;
  class MISS phRed;
```

After:

```mermaid
flowchart LR
  BUILD["build_launch_scripts.js<br/>(diff baked in via shell)"] --> F["launch_first.js"]
  BUILD --> RST["launch_rest.js"]
  F -->|"identical prompt"| R1{{"Reviewer 1"}}
  R1 --> CACHE["shared prefix cached (diff included)"]
  R1 -->|"claim appears in claimed/"| ORC{{Orchestrator}}
  ORC --> RST
  RST -->|"identical prompt, parallel"| REV{{"Reviewers 2..N"}}
  CACHE -->|"cache read"| REV
  Q["persona queue + claim_persona.sh"] -->|"tool result"| R1
  Q -->|"tool result"| REV
  class ORC,R1,REV phBlue;
  class BUILD,F,RST phYellow;
  class CACHE,Q phGray;
```

Approaches measured and rejected along the way: reordering the prompt (shared content first) had zero effect because cache hashing is block-granular; delivering the persona via a follow-up message to a finished agent rebuilt the request completely cold; `@file` mentions are not expanded in agent prompts; a dedicated warm-up agent worked but wasted a slot once the two-phase launch made reviewer 1 its own warm-up. The builder started as Python and was ported to Node so the whole launcher is one language.

**Calibration experiments** — spawning agents and reading exact `cache_creation_input_tokens` / `cache_read_input_tokens` from subagent transcripts:

| Experiment | Result |
| --- | --- |
| Old ordering (persona first) | followers read only the 18.9k system/tools segment, re-wrote ~24.8k + payload |
| Reordered prompt (shared first, persona last) | identical to old ordering, no additional hits |
| Byte-identical prompt | followers read the full ~43k prefix, wrote 0 |
| Persona via follow-up message | resumed request read 0, re-wrote ~41k (rejected) |
| Claim script in a live conversation | full prefix hit preserved, distinct personas assigned atomically |
| Workflow-composed identical prompts with embedded payload | first agent wrote 42,466; parallel followers read 42,466, wrote 0 |
| Two workflow runs, staggered (final design) | run 2's parallel agents read the full 42,476-token prefix written by run 1, wrote 0 |
| Two agents launched simultaneously, cold prefix | each wrote the full ~42k tokens (why the stagger exists) |

**End-to-end dogfood run** — the full skill (6 reviewers: 2 generalists + security, reliability, performance, compatibility) reviewing this branch's own diff. The protocol worked end to end: run dir, persona queue, generated scripts, two-phase launch, atomic claims, synthesized report. It also caught a real CRITICAL bug in itself — **all six reviewers independently found** that chained `replaceAll` substitution corrupted the diff shown to reviewers 2..N whenever the reviewed diff contains a placeholder token (this branch's own diff does), breaking the byte-identical invariant; several reviewers proved it by diffing their own prompt against `git show` from inside the run. Fixed with a single-pass template fill, verified byte-identical against this branch's diff. The remaining converged findings (contradictory oversized-diff instructions, claim-script failure modes, silent dropping of failed reviewers) were fixed alongside Greptile's two P1s — including an unquoted run-dir path in the generated claim command, verified fixed by claiming through a directory containing a space.

`build_launch_scripts.js` is also exercised against a synthetic run directory: byte-identical prompts, correct baked-in fan-out, `$&`/`$1` and placeholder-shaped diff content passing through verbatim.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The changed skill is itself the documentation (`.agents/skills/qa-team/`); no pages under `docs/` describe the launch mechanics.

## 🤖 Agent context

- Authored in a Claude Code session (Claude, agent-assisted): https://claude.ai/code/session_01DVoXnZSUtFRNMD7UXXgZ4E. The human driver asked whether the skill was prompt-cache friendly, then directed the measurement, the fix, dropping the warm-up agent in favor of the claim-poll stagger, shipping the launcher as files, consolidating on Node, the end-to-end dogfood run, and the whole-run cost framing.
- The design was reached by elimination through measured experiments (see the tables above): prompt reordering, follow-up-message persona delivery, `@file` mentions, and a dedicated warm-up agent were each tried and rejected on measured cache numbers or wasted work.
- Skills invoked: `/qa-team` (the dogfood run), `/claude-api` (cache pricing reference), `/setup-web-tests` (venv restoration so pre-commit hooks run in the remote environment).
- Known tradeoffs called out in the skill text: the persona queue on disk necessarily contains all personas (independence rests on the prompt instruction not to inspect the run dir), and persona claims are not idempotent under tool retries (the reconciliation step catches the resulting starvation).


pauldambra (approved):

gantoine (approved):
yeah nice!

PR #71622, opened 2026-07-16
fix(ci): fail hogbox preview open so infra errors don't red the PR

rnegron:
## Problem

- A tailnet timeout ([example run](https://github.com/PostHog/posthog/actions/runs/29505141291/job/87645712955)) failed `deploy preview` and painted the PR red. Previews are optional and never blocked merge, they just looked broken.

## Changes

- `deploy` fails open: step-level `continue-on-error` from the tailnet join down, each step gated on the previous step's `outcome`. Status lives in the Deployment plus sticky comment, where preview status belongs.
- All jobs get the `[optional, does not block merge]` prefix (same convention as the Depot shadow workflows). A check run is named after the job alone, so renaming only the workflow would be invisible in the checks list.
- `success()`/`failure()` are useless under continue-on-error, so both reporters now hang off one `ok` flag as exact negations.
- `build PR frontend` stays hard-failing: it runs PR code, and `deploy` reads its conclusion to tell a broken FE build from a backend-only PR.

> [!NOTE]
> There is no yellow. Jobs are success/failure/cancelled/skipped only, and job-level `continue-on-error` still renders a red X (it only flips `needs.<job>.result`).

- actionlint, `lint:workflows` and `ci:preflight --strict` pass.
- Traced the `ok` flag across 7 outcome paths in a shell harness (happy FE, backend-only, tailnet death, bring-up fail, FE build fail, dist fail, swap fail). No live preview run.
- No ruleset requires these check names, so the rename orphans nothing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None.

## 🤖 Agent context

- Claude Code (Opus 4.8), `/authoring-ci-workflows` skill.
- Rejected job-level `continue-on-error` first: confirmed it still reds the PR.
- Kept the `jobName` string in deploy's wait step in sync with the renamed job, verified programmatically. That coupling is the one thing a future rename can silently break.

### DIFF
--- a/.github/workflows/hogbox-preview-env.yml
+++ b/.github/workflows/hogbox-preview-env.yml
@@ -64,7 +64,7 @@
 # red. There is no "yellow" to use instead, and job-level `continue-on-error` still
 # reds the PR, so fail-open means step-level. The name prefix is on the JOBS because
 # a check run is named after the job, not the workflow.
-name: '[optional, does not block merge] Hogbox Preview Environment'
+name: '[Optional] Hogbox Preview Environment'
 
 on:
     pull_request:
@@ -114,7 +114,7 @@ jobs:
     # unlabeled-cancel trap: a no-op event runs only this job, never touches the
     # `preview`/`teardown` concurrency group, so it can't cancel an active build.
     decide:
-        name: '[optional, does not block merge] decide eligibility'
+        name: '[Optional] decide eligibility'
         runs-on: ubuntu-24.04
         timeout-minutes: 5
         permissions:
@@ -215,7 +215,7 @@ jobs:
     # per-PR concurrency group: announcing must never cancel an in-flight
     # deploy; build-frontend/deploy below do the coalescing.
     announce:
-        name: '[optional, does not block merge] announce build'
+        name: '[Optional] announce build'
         needs: decide
         if: needs.decide.outputs.build == 'true'
         runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


.github/workflows/hogbox-preview-env.yml
```diff
@@ -57,7 +57,14 @@
...
 
-name: Hogbox Preview Environment
+# --- OPTIONAL BY DESIGN ---------------------------------------------------------
+# A preview never gates a PR: all preview infra (tailnet, hogland, GitHub API, the
+# reporters) fails open and reports via the Deployment + sticky comment. Only
+# `build-frontend` (PR code — deploy reads its conclusion) and the fork guards still
+# red. There is no "yellow" to use instead, and job-level `continue-on-error` still
+# reds the PR, so fail-open means step-level. The name prefix is on the JOBS because
+# a check run is named after the job, not the workflow.
+name: '[optional, does not block merge] Hogbox Preview Environment'
```
gantoine: weird to have this in the name field, can we reduce it to just `[Optional]`?

.github/workflows/hogbox-preview-env.yml
```diff
@@ -206,7 +215,7 @@ jobs:
     # per-PR concurrency group: announcing must never cancel an in-flight
     # deploy; build-frontend/deploy below do the coalescing.
     announce:
-        name: announce build
+        name: '[optional, does not block merge] announce build'
```
gantoine: same with these

gantoine (changes requested):

.github/workflows/hogbox-preview-env.yml
rnegron: it's to be extra clear 🙈 the depot CI shadow ones are named this way. but OK, will change

gantoine (approved):

PR #71649, opened 2026-07-16
feat(salesforce): support Salesforce API version v67.0

Gilbert09:
## Problem

Salesforce's current stable REST API is v67.0, but our data-warehouse Salesforce source only ever declared and used v61.0, hardcoded into every SOQL query path. New connections should start on the current version, without disturbing sources already pinned to v61.0.

## Changes

- Add `v67.0` to `SalesforceSource.supported_versions` and make it the new `default_version`, so new sources are created on it.
- Thread the resolved API version through `salesforce_source` → `get_resource`, building the query path as `/services/data/<version>/query` instead of the hardcoded `v61.0`. The source resolves override → pin → default via `resolve_api_version`, so existing v61.0-pinned sources keep hitting v61.0 byte-for-byte.
- v61.0 stays fully supported; nothing is deprecated and no existing sources are repinned.

Between v61.0 and v67.0 the SOQL semantics, pagination, and response shapes we rely on are unchanged, so the only version-dependent piece is the URL path segment.

Automated tests only (I'm an agent — no manual Salesforce sync run):

- `test_get_resource_path_uses_api_version` and `test_salesforce_source_threads_api_version_into_request_path` — parameterized over `v61.0`/`v67.0`, catch a regression where the version stops reaching the query path (e.g. a reintroduced hardcode).
- `TestSalesforceSourceVersions` — locks in `v67.0` as the default for new sources and that a `v61.0` pin is still honored after the default bump.
- The registry invariant suite (`test_source_versions.py`) covers default ∈ supported automatically.

Ran the Salesforce source tests plus the version registry suite locally: 887 passed.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (Claude Code). Invoked the `/warehouse-source-new-version` skill for the version-declaration + dispatch conventions and `/writing-tests` before adding tests. The dispatch is minimal by design: only the query-path segment varies across these versions, so I threaded the resolved version string down rather than branching into per-version modules. Ruled out a duplicate by scanning open PRs (version string, module path, and the maintainer's own open PRs) before opening this.

---


danielcarletti:
No comments

danielcarletti (approved):
LGTM

PR #71650, opened 2026-07-16
fix(flags): don't let required evaluation contexts block experiment launch or flag auto-creation

phillram:
## Problem

Launching an experiment failed with `At least one evaluation context is required to create a new feature flag`, even though the linked flag already had evaluation contexts configured and the team requires them. Reported from a Slack thread; it blocked launching affected experiments until the "require evaluation contexts" team setting was temporarily disabled.

Root cause: launching an experiment flips the linked flag's `active` state through the gated `FeatureFlagSerializer` (via the `set_flag_active` / `update_flag` facade), and the launch endpoint is a `POST`. `FeatureFlagSerializer.validate` decided create-vs-update from `request.method`, so it read the flip as a *create* and demanded `evaluation_contexts` in the payload. The update payload is just `{"active": True}`, so validation raised - despite the flag already having contexts.

Any gated flag update routed through the serializer during a `POST` request hits this, not only experiment launch.

There is a second, related failure: features that automatically *create* a flag (experiments, early access features) never populate `evaluation_contexts` in the auto-created flag's payload, so a genuine create through those paths also fails when the team requires contexts - and those creation forms have no field to supply contexts.

Affected users:
- https://us.posthog.com/project/2/support/tickets/66156
- https://us.posthog.com/project/2/support/tickets/66122

Reproduction:
- Enable [Require feature flag evaluation contexts](https://us.posthog.com/settings/project-feature-flags#feature-flag-require-evaluation-contexts)
- Attempt to create an early access feature. Leave the flag blank.
- Notice the error popup:
 ```
 Could not save early access feature: At least one evaluation context is required to create a new feature flag.
 ```

CleanShot 2026-07-21 at 14 47 03@2x

## Changes

1. **Update path:** decide create vs update from `self.instance` (the DRF-native signal) instead of the HTTP verb. When an instance is present it's an update, regardless of the request method, so the "already has contexts, can't remove them all" guard applies and a no-op-for-contexts update passes.
2. **Create path (with defaults configured):** auto-created flags now include the team's default evaluation contexts. Extracted the default-context resolution from Max's flag tool (`_get_default_evaluation_contexts`) into the feature flags facade as `get_default_evaluation_contexts`, and wired it into both auto-create payloads: experiment flag creation (`experiment_service.py`) and early access feature flag creation (`early_access_features/backend/api.py`). This matches the manual flag-creation UI, which already applies these defaults.
3. **Create path (no defaults configured):** closes the previously-known gap - experiment and early access feature flags are now exempt from the evaluation-context requirement, the same way survey flags already were, since their creation forms have no field to supply contexts. Default contexts still apply automatically whenever a team has them configured (see point 2); this exemption only covers the case where none exist.

- `test_update_existing_flag_via_gated_write_under_post_request` mirrors the launch path: an existing flag with an evaluation context is flipped `active` through the `update_flag` facade while carrying a `POST` request, with the team requiring contexts. It fails on the old `request.method` logic and passes now.
- `test_create_experiment_applies_default_evaluation_contexts_when_required` and `test_create_applies_default_evaluation_contexts_when_required`: with the requirement on and team defaults configured, creation succeeds and the auto-created flag carries the default contexts.
- `test_create_experiment_flag_without_tags_when_required`, `test_create_early_access_feature_flag_without_tags_when_required`, `test_create_experiment_without_default_evaluation_contexts_when_required`, and `test_create_without_default_evaluation_contexts_when_required` (early access feature): cover the closed gap - creation now succeeds with the requirement on and no defaults configured, exercised both through the raw feature flags endpoint and through the real experiment/early-access-feature creation paths. The last one reproduces the exact "create an early access feature, leave the flag blank" repro from this PR's description.
- Ran `ruff check` and `ruff format --check` on the changed files. Couldn't run the Django suite in this environment (no Postgres/Redis available), so these are unverified against a live test run - flagging that explicitly rather than claiming a pass.

## 🤖 Agent context

Diagnosed and fixed by the PostHog Slack app agent from a support thread. Traced the error string to `products/feature_flags/backend/api/feature_flag.py`, then followed the experiment launch path (`experiment_service.launch_experiment` → `set_flag_active` → facade `update_flag` → `FeatureFlagSerializer`) to confirm the launch endpoint's `POST` verb was driving the create branch. Considered gating on `request.method in [...] and self.instance` but keying off `self.instance` alone is the correct, verb-independent signal and matches how DRF itself picks `create` vs `update`.

The create-path gap (defaults applied when configured) was flagged by ReviewHog on this PR and closed in a follow-up session with PostHog Code: rather than writing new default-resolution logic, the existing logic from Max's flag tool was extracted into the feature flags facade so all three consumers (manual UI parity, Max tool, auto-create paths) share one implementation.

The no-defaults gap - explicitly called out as a known limitation in this PR, and independently confirmed by ReviewHog against the PR's own "create an EAF with a blank flag" repro - was closed in a later session prompted by a PostHog Signals inbox report: extended the existing survey exemption to experiment and early access feature flags, since all three are auto-created flows with no context-selection field. No serializer fields or schemas changed, so generated types are unaffected.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BAB645UBA/p1784220619388399)*
*Updated with [PostHog Desktop](https://posthog.com/code?ref=pr) from [this inbox report](https://us.posthog.com/project/2/inbox/reports/019fa374-4f34-76a5-bf0c-051da306c812)*


gustavohstrassburger:
hey @phillram, this is in draft for a couple of weeks, do you want us to take it?

phillram:
Hey @gustavohstrassburger , I believe @haacked was looking into this one, as there were more architectural decisions to make.

haacked:
Yeah, the issue is when creating an experiment, we need to be able to select an evaluation context if the team has evaluation contexts required. We can't rely on the default because the default might not apply to the experiment context.

I think a good solution would be to show the evaluation context drop-down when creating an experiment IF the team has contexts required.

I had originally suggested the experiments team build that, but I can't find that discussion. Maybe one of us should just take it on?

haacked:
I'm still working on this. But its waiting on another PR to merge first.

PR #71653, opened 2026-07-16
feat(braintree): support Braintree API version 2026-07-14

Gilbert09:
## Problem

Braintree's GraphQL API is date-versioned via the required `Braintree-Version` header. The source was pinned to `2019-01-01` with the version hardcoded in the request path. Braintree's newest changelog-dated version is `2026-07-14`, and new sources should start there while every existing source keeps working exactly as before.

## Changes

- Declare `2026-07-14` alongside `2019-01-01` in `supported_versions`, and make it the `default_version` (new sources are stamped with it at creation).
- Thread the resolved API version (`resolve_api_version(inputs.api_version)`) from `source_for_pipeline` down through `braintree_source` → `get_rows` → the session's `Braintree-Version` header, instead of a hardcoded constant. Credential validation resolves through the same path.
- No behaviour change for existing pins: a source pinned to `2019-01-01` resolves to and sends `2019-01-01` exactly as before. Existing rows are not migrated or repinned.

The version metadata is surfaced automatically via the public source-config API — no per-source UI work.

Automated tests only (run by the agent):

- `test_braintree.py` — parameterized the version-header test over both `2019-01-01` and `2026-07-14`, asserting the passed version reaches the `Braintree-Version` header. Threaded the version through the existing `get_rows`/`braintree_source`/`validate_credentials` call sites.
- `test_braintree_source.py` — added tests for `supported_versions`/`default_version`, `resolve_api_version` (pin honored verbatim; NULL/empty fall back to the new default), and that `source_for_pipeline` dispatches the resolved version to `braintree_source`.
- Registry invariants in `sources/tests/test_source_versions.py` (default ∈ supported, https docs URL, etc.) pass.

All 906 tests in the touched test dirs pass. Ran `ruff check --fix` and `ruff format` over the touched files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Invoked the `/warehouse-source-new-version` skill and followed its version-declaration + dispatch conventions (thread the opaque version label to where the client/header is built; resolve through `resolve_api_version` at the source class; never hardcode a fallback). Braintree fit the "version is just a header, response shapes compatible" case, so the change is a thread-through rather than per-version modules. Ruled out a duplicate by searching open PRs for the version string, the module path, and the maintainer's own open PRs — the only "braintree" matches were unrelated `rest_source` migration batches that don't touch this source's version.

---


danielcarletti:
## Review: PR #71653 — feat(braintree): support Braintree API version 2026-07-14

**ELI10:** Braintree makes you tell it which "menu version" of its API you want by sending a date. This PR adds a newer date (2026-07-14) so brand-new connections use it, while everyone already connected keeps their old date and nothing about their data changes.

**Goal achieved:** yes — new sources default to `2026-07-14`, existing pins resolve to and send their stored version unchanged.
**Sensible approach:** yes — pure thread-through of `resolve_api_version(...)` to the `Braintree-Version` header, matching the versioning framework; no per-version modules needed since response shapes are compatible.
**Breaks dependents:** no — `get_rows`/`braintree_source`/`validate_credentials` gained a positional `api_version`, but every call site (source.py + tests) is updated and there are no importers outside the braintree package. Creation stamps `api_version=source.default_version` and backfill migration 0075 pinned all pre-existing Braintree sources to `2019-01-01`, so no NULL-pinned source silently jumps to the new default.

### Findings
- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/braintree/source.py:41 — `supported_versions`/`default_version` hardcode the version strings as literals while `braintree.py:23-24` defines `BRAINTREE_VERSION_2019_01_01`/`BRAINTREE_VERSION_2026_07_14` that are unused in production code (imported only by tests). Two sources of truth, and it diverges from the established convention (Stripe defines `STRIPE_API_VERSION_ACACIA` and references it in `supported_versions`/`default_version`). Fix: import the two constants into source.py and use them in `supported_versions`/`default_version` so the constants are the single source of truth and are actually referenced by production code.

### Verdict
Clean, well-scoped version bump that correctly reuses the versioning framework, updates all call sites, and preserves existing pins — no correctness or compatibility concerns. The only nit is the version strings living as unused constants in braintree.py while source.py re-hardcodes them; wiring source.py to reference the constants (as Stripe does) removes the duplication. Not a blocker.

products/warehouse_sources/backend/temporal/data_imports/sources/braintree/source.py
```diff
@@ -38,8 +38,8 @@
 @SourceRegistry.register
 class BraintreeSource(ResumableSource[BraintreeSourceConfig, BraintreeResumeConfig]):
     lists_tables_without_credentials = True  # static endpoint catalog — safe for public docs
-    supported_versions = ("2019-01-01",)
-    default_version = "2019-01-01"
+    supported_versions = ("2019-01-01", "2026-07-14")
```
danielcarletti: [should-fix] supported_versions/default_version hardcode the version strings as literals, while braintree.py:23-24 defines BRAINTREE_VERSION_2019_01_01/BRAINTREE_VERSION_2026_07_14 that are unused in production code (imported only by tests) — two sources of truth, and it diverges from the Stripe convention (STRIPE_API_VERSION_ACACIA is referenced directly in supported_versions/default_version). Fix: import the two constants and use them here so the constants become the single, production-referenced source of truth.

danielcarletti (commented):
One should-fix quality nit; the version bump itself is clean and compatibility-safe.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/braintree/source.py
Gilbert09: Addressed — `supported_versions` and `default_version` now reference `BRAINTREE_VERSION_2019_01_01` / `BRAINTREE_VERSION_2026_07_14` directly (source.py:43-44), so the constants are the single production-referenced source of truth, matching the Stripe convention.

PR #71654, opened 2026-07-16
chore(warehouse_sources): correct omnisend version-bump guidance

Gilbert09:
## Problem

This PR started as a version bump to advertise Omnisend's `2026-03-15` as a new supported API version, selected via an `Omnisend-Version` header on the existing `/v3` base path. That premise was wrong.

Verifying the change (docs-only, no live credentials) showed `2026-03-15` is a **different API surface**, not a header flip on `/v3`:

- **Base path moved.** `2026-03-15` list endpoints live at `https://api.omnisend.com/api` (e.g. `GET /api/contacts`), not `/v3`. Keeping `OMNISEND_BASE_URL=.../v3` and only adding the header means requests never reach the new handlers.
- **Pagination reshaped to cursors** (`paging.cursors.after` / `hasMore`); there's no `paging.next`, which is exactly what `get_rows` follows. Pagination would silently stop after page one.
- **Contact PK renamed** `contactID` → `id`.
- **`/orders` and `/carts` no longer exist as list endpoints** — they're reshaped into events. 2 of the 6 declared schemas can't be list-synced on the new version.

So flipping `default_version` to `2026-03-15` would have broken every newly created source at sync time.

## Changes

Reverted the Omnisend production files to master (v3-only). A real `2026-03-15` implementation is a source rewrite (new `/api` base, cursor pagination, renamed PKs, orders/carts-as-events, changed schema set), and it can't be verified or shipped without live credentials — building it against docs alone and flipping the default to it is the failure the versioning framework exists to prevent.

What's left is a docs/skill correction:

- `warehouse-source-new-version` skill: fix the conventional-title scope (`feat(warehouse_sources)`, not `feat(<dir>)`) and record the Omnisend learning — a vendor "new version" can be a whole new API surface, so verify each declared resource still exists as a list endpoint with the same envelope before advertising a version or flipping the default.

The in-repo `products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/api_inventory.md` already noted that v5 / `2026-03-15` reshape these endpoints into event-centric ones and that v3 is the right fit for a list-and-sync source. The revert keeps that accurate.

> [!NOTE]
> No production behavior changes. Existing and new Omnisend sources stay on `v3`. This PR is now a skill/docs correction; the `2026-03-15` rewrite is deferred until credentials are available to verify it end to end.

Net production diff is empty (reverted to master), so this is confirming no regression rather than covering new code. I (Claude) ran the affected suites locally, all green:

```
products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/tests/ ..... 38 passed
products/warehouse_sources/backend/temporal/data_imports/sources/tests/test_source_versions.py ..... passed
896 passed in 32.77s
```

`test_source_versions.py` is the registry invariant test (default in supported, deprecated ⊆ supported, https docs URL) — it confirms the reverted `OmnisendSource` declaration is well-formed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None. No user-facing behavior changed.

## 🤖 Agent context

Authored by Claude (Claude Code). Invoked `/warehouse-source-new-version` for the versioning conventions and `/writing-tests` while assessing the tests.

Decision trail: an earlier pass added the header-based version bump. Verifying it against the Omnisend docs (no live key) showed `2026-03-15` is a distinct API surface, not a header on `/v3`, and that orders/carts have no list endpoint there. A proper implementation is a source rewrite that can't be verified without credentials, and flipping the default to an unverified surface would break new sources at sync time. So I reverted the production files to master and kept only the skill correction that captures the finding. The `feat` title no longer fit, hence the retitle to `chore`.

---


danielcarletti:
## Review: PR #71654 — feat(omnisend): support Omnisend API version 2026-03-15

**ELI10:** Omnisend (an email tool) started numbering its API by date. This PR teaches our importer about the new date-numbered version, makes brand-new connections use it, and tells the importer to send a little "which version" note on each request.

**Goal achieved:** yes — new sources declare/default to `2026-03-15`, the resolved version is threaded down to the request headers, and `v3` requests stay byte-for-byte identical.
**Sensible approach:** yes — threading the resolved version string down to `_get_headers` (rather than per-version modules) is the minimal, framework-consistent shape; header gated on the legacy label.
**Breaks dependents:** no — both new params (`get_rows`, `omnisend_source`, `_get_headers`) default to `LEGACY_API_VERSION`; the only production callers are in-module, and `validate_credentials`'s `_get_headers(api_key)` call is unchanged. Existing pins (incl. NULL) still resolve exactly as before.

### Findings
- [blocker] products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/source.py:39 — The default flips to `2026-03-15` on the unverified claim that `Omnisend-Version: 2026-03-15` on the `/v3` base path returns the same `{<data_key>: [...], "paging": {"next": ...}}` envelope. This directly reverses the prior in-repo caveat (removed in this PR) that said "v5 / v2026-03-15 reshape several of these into event-centric endpoints; v3 is the right fit for a list-and-sync warehouse source." `api_inventory.md` itself states "Verify against the live API before changing endpoint behavior," yet the PR body confirms no live-API run was done and every added test mocks the session. If the dated version changes the envelope key or pagination, `data[config.data_key]` KeyErrors (or pagination silently stops) for every newly created source at sync time. Fix: verify each endpoint against the live Omnisend API with the `2026-03-15` header before flipping the default; if it does reshape, keep `default_version = "v3"` and only add `2026-03-15` to `supported_versions`.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/tests/test_omnisend_source.py:116 — `resolve_api_version(None)` / `("v3")` assertions duplicate the generic base-class resolution already exercised end-to-end by the parameterized `test_source_for_pipeline_passes_resolved_version`; combined with the literal `default_version` / `supported_versions` equality checks, the test is largely a change-detector. Fix: drop the two `resolve_api_version` assertions here and keep only the declaration-shape check.

### Verdict
The plumbing is clean, minimal, and backwards-compatible — blast radius is genuinely contained. The one must-fix is that the load-bearing behavior change (flipping the default to `2026-03-15`) rests on an unverified API-contract claim that contradicts the repo's own prior note, with no live-API evidence and only session-mocking tests. Confirm the dated version keeps the resource-based list envelope on the `/v3` path before merging; otherwise add the version but leave the default on `v3`.

products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/source.py
```diff
@@ -35,8 +35,8 @@
 
 @SourceRegistry.register
 class OmnisendSource(ResumableSource[OmnisendSourceConfig, OmnisendResumeConfig]):
-    supported_versions = ("v3",)
-    default_version = "v3"
+    supported_versions = ("v3", "2026-03-15")
+    default_version = "2026-03-15"
```
danielcarletti: [blocker] Default flips to `2026-03-15` on the unverified assumption that `Omnisend-Version: 2026-03-15` on the `/v3` base path returns the same `{<data_key>: [...], "paging": {"next": ...}}` envelope. This reverses the prior in-repo caveat (removed in this PR): "v5 / v2026-03-15 reshape several of these into event-centric endpoints; v3 is the right fit for a list-and-sync warehouse source." api_inventory.md itself says "Verify against the live API before changing endpoint behavior," but no live-API run was done and all added tests mock the session. If the dated version reshapes the response, `data[config.data_key]` KeyErrors (or pagination silently stops) for every newly created source at sync time. Fix: verify each endpoint against the live Omnisend API with the 2026-03-15 header before flipping the default; if it reshapes, keep `default_version = "v3"` and only add 2026-03-15 to `supported_versions`.

products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/tests/test_omnisend_source.py
```diff
@@ -108,6 +108,14 @@ def test_get_non_retryable_errors_covers_auth(self) -> None:
         assert any("401" in key for key in errors)
         assert any("403" in key for key in errors)
 
+    def test_new_default_version_is_the_dated_version(self) -> None:
+        source = OmnisendSource()
+        assert source.default_version == "2026-03-15"
+        assert set(source.supported_versions) == {"v3", "2026-03-15"}
+        # An unpinned source resolves to the new default; existing v3 pins are honored verbatim.
+        assert source.resolve_api_version(None) == "2026-03-15"
```
danielcarletti: [nit] These `resolve_api_version(None)` / `("v3")` assertions duplicate the generic base-class resolution already exercised end-to-end by the parameterized `test_source_for_pipeline_passes_resolved_version`. Combined with the literal `default_version` / `supported_versions` equality checks above, this test is largely a change-detector. Fix: drop the two `resolve_api_version` assertions and keep only the declaration-shape check.

danielcarletti (commented):
Plumbing is clean and backwards-compatible (blast radius contained). One must-fix: the default flip to 2026-03-15 rests on an unverified API-contract claim that contradicts the repo's prior note, with only session-mocking tests.

danielcarletti:
AI reply: Reverted the version bump. The PR assumed `2026-03-15` is selected by sending an `Omnisend-Version` header on the same `/v3` list paths, returning the same `{<key>: [...], "paging": {"next": ...}}` envelope. Omnisend's current API docs show it is a different API surface, so that assumption is false on every count:

- **Base path changes:** `2026-03-15` list endpoints live at `https://api.omnisend.com/api` (e.g. `GET /api/contacts`, `GET /api/products`), not `https://api.omnisend.com/v3`. The source keeps `OMNISEND_BASE_URL = .../v3` and only adds a header, so requests never reach the new handlers.
- **Pagination reshaped:** contacts/campaigns/segments use cursor pagination (`paging.cursors.after` + `paging.hasMore`); there is no `paging.next`. `get_rows` reads `data.get("paging", {}).get("next")`, so pagination would silently stop after page 1.
- **Primary keys renamed:** e.g. contact PK is `id`, not `contactID`.
- **Resources dropped:** `2026-03-15` has **no** `/orders` or `/carts` list endpoints at all — those were reshaped into events (`placed order`, `added product to cart`, …). Two of this source's six declared schemas cannot be list-synced on the new version.

So flipping the default to `2026-03-15` would break every newly created Omnisend source at sync time (KeyError / truncated sync / missing endpoints). Adopting it is a source rewrite (new `/api` base path, cursor pagination, orders/carts-as-events, changed schema set), not a header flip — which is exactly what the in-repo caveat this PR removed already said. Advertising it as a `supported_version` the source can't serve would be inert broken scaffolding, so I dropped it entirely rather than leaving it declared.

This change reverts the source to v3-only (production files now byte-identical to master, `test_omnisend*` + `test_source_versions` green — 896 passed), restores the accurate `api_inventory.md` caveat, and records the "a new vendor version can be a whole new API surface" lesson in the versioning skill.

Docs evidence: https://api-docs.omnisend.com/reference/get_contacts.md (server `https://api.omnisend.com/api`, cursor envelope, PK `id`), https://api-docs.omnisend.com/reference/pagination (cursor-based), and the full `2026-03-15` reference listing at https://api-docs.omnisend.com/llms.txt (no orders/carts list endpoints).

Note: the `feat` title no longer matches — this is now effectively a docs/caveat correction. Happy to close instead if you'd rather land the real `2026-03-15` adoption as a separate rewrite.

products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/source.py
Gilbert09: Agreed, and taken to its conclusion: rather than flip the default, this PR now drops the 2026-03-15 version bump entirely. Verifying against the surface showed 2026-03-15 is not a header flip on `/v3` — it is a different API surface (base path `https://api.omnisend.com/api`, cursor pagination, renamed PKs, dropped `/orders` and `/carts` list endpoints). So `default_version` stays `v3`, the source.py and test changes are reverted, and the only remaining change is a skill learning documenting that a version bump here would be a source rewrite, not a supported-version add. Resolving as addressed.

products/warehouse_sources/backend/temporal/data_imports/sources/omnisend/tests/test_omnisend_source.py
Gilbert09: Moot now — the omnisend source and test changes have been reverted (the 2026-03-15 bump is dropped, since it is a different API surface rather than a header flip), so these `resolve_api_version` assertions are no longer part of the PR. The remaining change is a skill-only learning. Resolving.

Gilbert09:
Closing this. The per-source version-bump PRs each also appended a learnings note to `.agents/skills/warehouse-source-new-version/SKILL.md`, and they all conflicted with each other on that file. I've stripped the SKILL.md edit from every one of those PRs. Since this PR's only change was the SKILL.md guidance edit, there's nothing left for it to carry, so closing rather than leaving it empty.

PR #71656, opened 2026-07-16
feat(servicenow): support ServiceNow API version v2

Gilbert09:
## Problem

ServiceNow's Table API ships both `v1` and `v2` (v2 is the newer stable version, available since the Geneva release). This source only ever hit the versionless `/api/now/table` path and implicitly labeled it `v1`. This adds first-class `v2` support so new sources use the current stable version, while every existing source keeps working exactly as before.

Vendor references: [Table API docs](https://www.servicenow.com/docs/r/washingtondc/api-reference/rest-apis/c_TableAPI.html), [KB0551763](https://support.servicenow.com/kb?id=kb_article_view&sysparm_article=KB0551763).

## Changes

- Declared `supported_versions = ("v1", "v2")` and flipped `default_version` to `v2` on `ServiceNowSource`, so new sources are created on v2.
- Added a version → URL mapping in the request layer. `v1` keeps the versionless `/api/now/table/{table}` path it has always used; `v2` targets `/api/now/v2/table/{table}`. The resolved pin is threaded `source_for_pipeline` → `servicenow_source` → `get_rows` via `resolve_api_version(inputs.api_version)`.
- Pointed `api_docs_url` at the vendor Table API reference (where versions are announced).

> [!NOTE]
> No migration or repin. Existing sources were stamped with their `v1` pin at creation, so the default flip leaves them on v1 and their synced data is unchanged. With our request params (`sysparm_display_value=false`, `sysparm_exclude_reference_link=true`) v1 and v2 row shapes are identical, so only the URL segment differs. v1 is not deprecated.

Added/extended automated tests (I, Claude, could not run the suite in this environment — pytest isn't installed here, so these run in CI):

- `test_servicenow.py`: a pure-function `_table_api_url` test covering v1, v2, and the unknown-pin fallback; and a parameterized `get_rows` test asserting the version selects the correct request URL. Catches a broken version → URL mapping (e.g. a v2 source silently hitting the versionless path, or a v1 source getting moved to `/v2/`).
- `test_servicenow_source.py`: asserts `default_version == "v2"` / `supported_versions == ("v1", "v2")`, and that `source_for_pipeline` resolves an unpinned source to v2 while honoring an explicit `v1`/`v2` pin. Catches a regression where the resolved pin stops reaching the request layer.

The registry invariant test (`test_source_versions.py`) also picks up any declaration mistakes automatically. Verified the URL-mapping logic in isolation and ran `ruff check`/`ruff format` over the touched files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude) implemented this following the repo's `/warehouse-source-new-version` skill for the version-declaration and dispatch conventions, and `/writing-tests` for the test additions. The key decision was preserving byte-for-byte behavior for existing sources: since `v1` currently uses the versionless Table API path, I mapped `v1` → versionless and `v2` → the explicit `/v2/` segment rather than moving v1 onto a `/v1/` path (which could change responses on instances where versionless resolves to v2). Checked open PRs (including the maintainer's own) for a pre-existing ServiceNow v2 fix before starting — none found.

---


danielcarletti:
## Review: PR #71656 — feat(servicenow): support ServiceNow API version v2

**ELI10:** ServiceNow's data can be fetched from an old address or a newer one. This teaches PostHog about the newer address (`v2`) and makes brand-new connections use it, while every connection that already exists keeps using the old address so nothing they've synced changes.

**Goal achieved:** yes — new sources default to `v2` (explicit `/api/now/v2/table/`), existing sources stay on the versionless `v1` path.
**Sensible approach:** yes — follows the existing `supported_versions`/`default_version`/`resolve_api_version` framework exactly like `attio`, `drip`, etc.; maps `v1`→versionless (not `/v1/`) which is the correct conservative choice.
**Breaks dependents:** no — the only prod caller of `get_rows` is `servicenow_source`, and the only caller of that is `source_for_pipeline`; both are updated to thread `api_version`. Verified via grep. Existing rows were stamped `api_version="v1"` at creation (old base default `UNVERSIONED_API_VERSION = "v1"`, set in `_create_external_data_source`), so they resolve to the versionless path unchanged — no migration needed.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/servicenow/source.py:50 — the no-migration safety net rests on two assumptions that are stated but not verified in the diff: (a) every existing ServiceNow source row has an explicit `v1` pin, and (b) v1/v2 row shapes are byte-identical under `sysparm_display_value=false` + `sysparm_exclude_reference_link=true`. Any row with a NULL `api_version` (the column is nullable with no DB default) now resolves to `v2` instead of `v1`, silently moving it to `/api/now/v2/table/`. It's harmless only if (b) holds. Fix: confirm no existing ServiceNow rows have NULL `api_version` (query the `data_warehouse_sources` system table / prod), or add a written-not-run backfill pinning NULL→`v1` before merge; the byte-identical claim is load-bearing, so keep it explicit.

### Verdict
Solid, well-scoped PR that reuses the versioning framework correctly, keeps existing syncs byte-for-byte unchanged, and is backed by targeted tests (pure-function URL mapping + wiring assertions) that aren't redundant. The SKILL.md note lands in the intended `### Learnings` section, and leaving `validate_credentials` on the versionless probe is fine (the skill explicitly sanctions this at line 53). No blockers; the only thing worth a sanity check before merge is the NULL-pin edge and the load-bearing v1/v2 byte-identical assumption.

danielcarletti (commented):
One nit only — the PR is clean and the versioning-framework reuse is correct. The single flag is a confirm-not-fix around the no-migration safety net.

products/warehouse_sources/backend/temporal/data_imports/sources/servicenow/source.py
```diff
@@ -39,7 +41,13 @@
...
 @SourceRegistry.register
 class ServiceNowSource(ResumableSource[ServiceNowSourceConfig, ServiceNowResumeConfig]):
-    api_docs_url = "https://developer.servicenow.com"
+    api_docs_url = "https://www.servicenow.com/docs/r/washingtondc/api-reference/rest-apis/c_TableAPI.html"
+
+    # Both Table API versions stay supported; new sources default to v2 (stable since Geneva).
+    # Existing sources were stamped with their v1 pin at creation, so the default flip leaves
+    # them on v1 (the versionless path) and their syncs unchanged.
+    supported_versions = (SERVICENOW_API_VERSION_V1, SERVICENOW_API_VERSION_V2)
+    default_version = SERVICENOW_API_VERSION_V2
```
danielcarletti, line 53: [nit] The no-migration guarantee rests on two assumptions not verified in the diff: (a) every existing ServiceNow row carries an explicit v1 pin, and (b) v1/v2 rows are byte-identical under sysparm_display_value=false + sysparm_exclude_reference_link=true. api_version is nullable with no DB default, so any row with NULL now resolves to v2 (not v1) and moves to /api/now/v2/table/ — harmless only if (b) holds. Fix: confirm no existing ServiceNow rows have NULL api_version (query the data_warehouse_sources system table), or add a written-not-run backfill pinning NULL->v1; the byte-identical claim is load-bearing, so keep it explicit.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/servicenow/source.py
Gilbert09: Good callout — the byte-identical guarantee here is load-bearing, so let me make the reasoning explicit.

Assumption (a) holds: no existing ServiceNow row has a NULL `api_version`.
- **Rows created before the version framework:** migration `0075_backfill_externaldatasource_api_version` (already on master) backfills every row. ServiceNow isn't in its `API_VERSION_BY_SOURCE_TYPE` map, so it falls to `DEFAULT_API_VERSION = "v1"`. The `api_version__isnull=True` guard means it only touches previously-NULL rows, and it's idempotent.
- **Rows created after that backfill but before this PR:** stamped at creation with `default_version` (`api_version=source.default_version` in `_create_external_data_source`), which for ServiceNow was the inherited `UNVERSIONED_API_VERSION = "v1"` prior to this flip.

So there is no NULL-pin window: every existing row carries an explicit `"v1"` pin, and `resolve_api_version` returns v2 only for genuinely new sources created after this deploys. The default flip never repaths an existing row.

On (b): the URL-segment-only claim is the reason v1 stays on the versionless path (`/api/now/table`) rather than being moved to `/api/now/v1/` — with `sysparm_display_value=false` + `sysparm_exclude_reference_link=true` the row shapes are identical across versions, so v1 syncs are byte-for-byte unchanged. As belt-and-suspenders, `_table_api_url` also falls back to the versionless path for any unrecognized pin.

I've made the guarantee explicit in the code comment in `source.py` (ca372f5) so it doesn't rest on tribal knowledge. Given 0075 already covers the backfill, no additional migration is needed here.

PR #71657, opened 2026-07-16
feat(intercom): support Intercom API version 2.15

Gilbert09:
## Problem

Intercom's latest documented API version is 2.15. Our Intercom warehouse source only spoke 2.13, and there was no version dispatch at all - the `Intercom-Version` header was a hardcoded module constant. New sources should start on the current version, but existing customers pinned to 2.13 must stay exactly where they are.

## Changes

- Declare `2.15` in `supported_versions` alongside `2.13`, and flip `default_version` to `2.15` so new sources are created on it. Existing pinned rows are untouched.
- Introduce minimal version dispatch: the resolved pin (`resolve_api_version(inputs.api_version)`) is threaded from `source_for_pipeline` into `intercom_source` and down to the `Intercom-Version` header on every request path - the framework REST client, the companies scroll session, and the substream session.
- `validate_credentials` (which runs at source-create time with no row pin) keeps using the legacy constant, so credential validation behavior is unchanged.

Response shapes are auto-inferred per sync, so 2.15's breaking response changes are absorbed without per-version field mapping. Intercom (a `SimpleSource`) has no managed HogQL schema / canonical column hints, so there's nothing to gate per version.

> [!NOTE]
> This is a plain version add, not a sunset. 2.13 is not deprecated and existing syncs pinned to it send the identical `Intercom-Version: 2.13` header as before.

Automated only (I'm an agent - no manual/live Intercom calls). I ran:

- The full Intercom source test suite (`tests/test_intercom.py`, `tests/test_intercom_source.py`) - 126 passed.
- The registry invariant suite (`sources/tests/test_source_versions.py`) - 858 passed.

New tests added:

- `TestVersionDispatch` in `test_intercom.py` - parameterized over `IntercomSource.supported_versions`, asserts the pinned version reaches the wire as the `Intercom-Version` header on both the session path (scroll/substream) and the framework REST path. Catches a regression where a new version is declared but never threaded to requests.
- `test_default_version_is_latest`, `test_resolve_api_version`, and `test_source_for_pipeline_resolves_version` in `test_intercom_source.py` - lock the new default, the pin-wins-over-default resolution, and that the resolved version is what `source_for_pipeline` hands to `intercom_source`.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude) implemented this. I invoked the repo's `/warehouse-source-new-version` skill and followed its version-declaration + dispatch conventions (declare in `supported_versions`, flip `default_version`, resolve through `resolve_api_version`, keep old versions working, never hardcode a fallback version in the request layer). I checked open PRs from several angles (the version string, the module path, and the maintainer's own open PRs) and found no existing work on Intercom 2.15, so this is not a duplicate.

Decision worth flagging: I left `validate_credentials` on the legacy `2.13` constant rather than the new default, since validation runs pre-pin and the skill treats changing it as optional per bump - keeping it stable avoids any risk of 2.15 rejecting the `/me` probe.

---


danielcarletti:
## Review: PR #71657 — feat(intercom): support Intercom API version 2.15

**ELI10:** Intercom (a service PostHog imports data from) released a newer version of its API, 2.15. This PR teaches the Intercom importer to speak 2.15 for new connections while leaving everyone already on 2.13 exactly where they are, by passing the right version label along to every request.

**Goal achieved:** yes — the resolved pin is threaded from `source_for_pipeline` → `intercom_source` → the `Intercom-Version` header on all three request paths (framework REST, companies scroll session, substream session).
**Sensible approach:** yes — matches the reference source (Stripe) exactly: declare in `supported_versions`, flip `default_version`, re-resolve via `resolve_api_version` in `source_for_pipeline`. No new abstraction, response shapes auto-inferred per sync so no per-field mapping needed.
**Breaks dependents:** no — the only production caller of `intercom_source` is `source_for_pipeline` (updated); it's not a standalone Temporal activity so no workflow-determinism concern. Critically, existing unpinned Intercom sources are safe: backfill migration `0075_backfill_externaldatasource_api_version` already stamped every existing row with an explicit `"2.13"` pin, so flipping `default_version` to `2.15` only affects newly created sources.
**Undisclosed side effects:** none — `validate_credentials` deliberately staying on the `2.13` constant is disclosed in the PR body and the code comment.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/intercom/tests/test_intercom_source.py:37 — `test_resolve_api_version` exercises the un-overridden base-class `resolve_api_version` (trivially `pinned or default_version`); its `None/2.13/2.15` cases are already covered end-to-end by `test_source_for_pipeline_resolves_version` just below, and the framework's own `test_source_versions.py` covers the base method. Fix: drop it (optionally fold the extra empty-string `("", "2.15")` case into the pipeline test) to avoid a duplicate change-detector.

### Verdict
Clean, well-scoped version add that follows the established Stripe pattern faithfully and threads the version to every request path with matching test coverage (parameterized over `supported_versions`, so future bumps stay honest). Blast radius is genuinely contained — the pre-existing backfill migration means no silent version jump for current customers. No blockers. The one thing that can't be verified in code and warrants a human eye: `default_version` now flips every *new* Intercom source onto 2.15, which was only tested with mocked responses (agent-authored, no live Intercom calls) — worth a single manual smoke sync against a real 2.15 workspace before/after merge.

danielcarletti (commented):
One nit only — the PR is clean, faithfully follows the Stripe version pattern, and existing sources are protected by the 0075 backfill (explicit 2.13 pin), so no silent version jump.

products/warehouse_sources/backend/temporal/data_imports/sources/intercom/tests/test_intercom_source.py
```diff
@@ -20,6 +20,44 @@ def setup_method(self):
...
+    @pytest.mark.parametrize(
+        "pinned,expected",
+        [
+            (None, "2.15"),
+            ("", "2.15"),
+            ("2.13", "2.13"),
+            ("2.15", "2.15"),
+        ],
+    )
+    def test_resolve_api_version(self, pinned, expected):
```
danielcarletti: [nit] test_resolve_api_version exercises the un-overridden base-class resolve_api_version (trivially `pinned or default_version`); its None/2.13/2.15 cases are already covered end-to-end by test_source_for_pipeline_resolves_version just below, and the base method has its own coverage in test_source_versions.py. Fix: drop this test (optionally fold the extra empty-string ("", "2.15") case into the pipeline test) to avoid a duplicate change-detector.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/intercom/tests/test_intercom_source.py
Gilbert09: Good call — done. `test_resolve_api_version` is dropped, and the None/2.13/2.15 cases (plus the empty-string case) are now folded into `test_source_for_pipeline_resolves_version`, which exercises resolution end-to-end. Base-class coverage stays in `test_source_versions.py`.

PR #71658, opened 2026-07-16
feat(github): support GitHub API version 2026-03-10

Gilbert09:
## Problem

GitHub released a newer date-based REST API version, `2026-03-10`. The GitHub data-warehouse source only knew about `2022-11-28`, so new sources couldn't be created on the newer stable version. `2022-11-28` is still supported by GitHub (no forced migration), so this is a plain update, not a sunset.

## Changes

- Declare `2026-03-10` in `GithubSource.supported_versions` and make it the `default_version`, so new sources are stamped with it at creation.
- Thread the source's resolved API version down to the `X-GitHub-Api-Version` header. `source_for_pipeline` passes `self.resolve_api_version(inputs.api_version)` into `github_source`, which threads it through `get_rows` / `_fan_out_get_rows` to `_get_headers`. The header is the only version-dependent part for the endpoints we sync (response shapes are compatible across these versions).
- No version is deprecated and no rows are repinned. This PR does not migrate existing sources.

## Why existing sources are safe

A source's effective version is its `api_version` pin, resolved through `default_version` only when the pin is NULL. Flipping the default would move a NULL-pinned row to `2026-03-10` — but no GitHub row reaches sync time on a NULL pin:

- The versioning-framework backfill (migration `0075`, already on master) pinned every pre-existing GitHub source to `2022-11-28`. It is an ancestor of this branch's migrations, so it always runs first.
- Since that framework shipped, `_create_external_data_source` stamps `api_version=default_version` on every new source at creation — no path leaves the column NULL.

So every existing GitHub source carries a concrete `2022-11-28` pin that `resolve_api_version` honors verbatim. The default flip only affects brand-new sources, which is the intended behavior. (An earlier revision of this branch carried a GitHub-specific repin migration; it was redundant with `0075`, which already clears the NULL cohort before any later migration runs, so it was dropped.)

> [!NOTE]
> Flipping `default_version` only changes what brand-new sources are stamped with. Pinned rows keep their pin by design, so this is not a migration.

Added automated tests (run locally, green):

- `test_source_for_pipeline_threads_resolved_api_version` (parameterized): guards that the resolved pin reaches `github_source` — a source pinned to `2022-11-28` stays on it, `2026-03-10` is honored, and an unpinned row falls back to the current default. Catches a regression where the default flip silently moves a source to a new version, which no existing test covered.
- `test_get_rows_sends_pinned_api_version_header` (parameterized): guards that the version actually lands in the outgoing `X-GitHub-Api-Version` header for each supported version, and that the omitted case falls back to the legacy default. Catches a broken thread-through in `github.py` that would sync every source on one version.
- The registry invariant test (`test_source_versions.py`) automatically checks default ∈ supported and the docs URL.

Also verified against GitHub's published API-versions documentation that `2026-03-10` is a real, currently-supported `X-GitHub-Api-Version` value and that `2022-11-28` remains supported (GitHub rejects unknown version labels, so this matters).

## 🤖 Agent context

Authored by Claude (Opus 4.8) via PostHog Code. Invoked the `/warehouse-source-new-version` skill for the version-declaration and dispatch conventions, and `/writing-tests` before adding tests. The GitHub source had no version dispatch yet (the header was hardcoded), so I introduced a minimal one: a single `GITHUB_DEFAULT_API_VERSION` constant defaulting the legacy paths, plus an `api_version` parameter threaded through the sync request path only.

---


danielcarletti:
## Review: PR #71658 — feat(github): support GitHub API version 2026-03-10

**ELI10:** GitHub added a new "menu version" for its data API. This PR teaches the GitHub importer about the new version, makes it the one brand-new imports use, and passes each import's chosen version all the way down to the actual web request header.

**Goal achieved:** yes — the resolved version is threaded end-to-end (`source_for_pipeline` → `github_source` → `get_rows`/`_fan_out_get_rows` → `_get_headers`), covering both the direct and fan-out poll paths.
**Sensible approach:** yes — single `api_version` param + one legacy-default constant, no per-version modules, matches how the versioning framework is used elsewhere (e.g. Stripe `source.py:488`).
**Breaks dependents:** at risk — sources pinned to `2022-11-28` are unchanged, but **unpinned (`api_version = NULL`) GitHub sources created after the 0075 backfill silently move `2022-11-28` → `2026-03-10`**. Harmless only if the response shapes really are compatible (see finding).
**Undisclosed side effects:** none beyond the NULL-cohort behavior change noted below.

### Findings
- [should-fix] `source.py:91` — The PR states existing sources are "unaffected" and that a NULL pin "still sends the exact header it did before," and the agent context says new sources are "stamped [with the default] at creation." All three are inaccurate. There is **no** creation-time stamping — `ExternalDataSource.api_version` is created NULL and resolved to `default_version` at sync time (model comment at `models/external_data_source.py:64`; the only consumers of `default_version` are the frozen 0075 backfill and `resolve_api_version`). So every currently-unpinned GitHub row — all sources created after migration 0075 (~2026-07-10) and before this deploy — sent the hardcoded `2022-11-28` before this PR and will resolve to `2026-03-10` after it. That is precisely the "silently moving a customer to another version" the `resolve_api_version` docstring says the framework exists to prevent, and the `(None → 2026-03-10)` test case codifies the move as intended. Fix: either add a data migration pinning existing NULL `Github` rows to `"2022-11-28"` *before* flipping `default_version` (making the "unaffected" claim actually true), or correct the description to state unpinned sources float to the new default — and, if you keep floating, confirm the synced endpoints' response shapes are genuinely identical across the two versions for those in-flight NULL sources.
- [nit] `source.py:90` — Confirm `2026-03-10` is a real, published `X-GitHub-Api-Version` value in GitHub's REST API-versions changelog. GitHub rejects requests carrying an unknown version string, so a typo'd / not-yet-released date would break every new source at sync time; there's no test that exercises a live request.

### Verdict
The threading is clean and complete, and pinned sources are safe. The one thing to resolve before merge is the NULL-pin cohort: the PR's central safety claim ("existing sources unaffected") does not hold for unpinned rows, which float onto the new default — either pin them via a small backfill or restate the guarantee and confirm cross-version response compatibility. Also worth a 30-second check that the `2026-03-10` label is a version GitHub actually accepts.

products/warehouse_sources/backend/temporal/data_imports/sources/github/source.py
```diff
@@ -87,8 +87,8 @@ class GithubSource(
     OAuthMixin,
 ):
     lists_tables_without_credentials = True  # static endpoint catalog — safe for public docs
-    supported_versions = ("2022-11-28",)
-    default_version = "2022-11-28"
+    supported_versions = ("2022-11-28", "2026-03-10")
+    default_version = "2026-03-10"
```
danielcarletti, line 95: [should-fix] The PR/agent-context claim that existing sources are "unaffected", that a NULL pin "still sends the exact header it did before", and that new sources are "stamped at creation" is inaccurate. There is no creation-time stamping — `ExternalDataSource.api_version` is created NULL and resolved to `default_version` at sync time (see model comment at models/external_data_source.py:64; the only consumers of `default_version` are the frozen 0075 backfill and `resolve_api_version`). So every currently-unpinned GitHub row (all sources created after migration 0075 ~2026-07-10 and before this deploy) sent hardcoded `2022-11-28` before this PR and now resolves to `2026-03-10` — the silent version move the `resolve_api_version` docstring says the framework exists to prevent, codified by the `(None -> 2026-03-10)` test. Fix: either add a data migration pinning existing NULL Github rows to "2022-11-28" before flipping `default_version`, or correct the description to state unpinned sources float to the new default and confirm the synced endpoints' response shapes are identical across both versions.

danielcarletti (commented):
Threading is clean and complete; pinned sources are safe. Main thing to resolve: the "existing sources unaffected" claim does not hold for unpinned (NULL api_version) rows, which float onto the new default at sync time. Plus a quick check that the version string is one GitHub actually accepts.

products/warehouse_sources/backend/temporal/data_imports/sources/github/source.py
```diff
@@ -87,8 +87,8 @@ class GithubSource(
     OAuthMixin,
 ):
     lists_tables_without_credentials = True  # static endpoint catalog — safe for public docs
-    supported_versions = ("2022-11-28",)
-    default_version = "2022-11-28"
+    supported_versions = ("2022-11-28", "2026-03-10")
```
danielcarletti, line 94: [nit] Confirm `2026-03-10` is a real, published `X-GitHub-Api-Version` value in GitHub's REST API-versions changelog. GitHub rejects requests carrying an unknown version string, so a typo or not-yet-released date would break every new source at sync time; no test exercises a live request.

danielcarletti:
AI reply: Root-caused the "existing sources float to the new default" concern and corrected the PR accordingly.

The premise that a NULL-pinned GitHub row "still sends the exact header it did before" was wrong — a NULL pin resolves through `default_version`, so after this flip it would resolve to `2026-03-10`. But that cohort does not exist at sync time:

- Migration `0075` (the versioning-framework backfill, already on master) pins every pre-existing GitHub source to `2022-11-28`. It is an ancestor of this branch's migrations, so it always runs first — any later migration sees zero NULL GitHub rows.
- Since that framework shipped, `_create_external_data_source` stamps `api_version=default_version` at creation, so no new source is left NULL either.

Because of that, the branch-added GitHub-specific repin migration (`0079`) matched zero rows in every possible ordering — it was inert scaffolding duplicating `0075`. Removed it (plus its test), reverted `max_migration.txt` to `0078`, and repointed the resolution test's comment at `0075`. `makemigrations --check` reports no drift; the dispatch is unchanged.

Also confirmed against GitHub's published API-versions docs that `2026-03-10` is a real, currently-supported `X-GitHub-Api-Version` value (released 2026-03-10) and `2022-11-28` remains supported — so the new label is valid.

Ran the GitHub dispatch tests + the version-registry invariant locally: green.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/github/source.py
Gilbert09: Verified this holds and the description was corrected to match. Every existing GitHub source is already non-NULL pinned before this deploy, via two mechanisms that both shipped in the versioning-framework PR (#70018):

1. Migration `0075` backfilled every pre-existing GitHub row to `2022-11-28`.
2. Since that same PR, the create path (`_create_external_data_source`) stamps `api_version=source.default_version` on every new source at creation — so rows created after `0075` are also concrete-pinned (to `2022-11-28`, the default at the time), not NULL.

So no GitHub row reaches sync time on a NULL pin, and `resolve_api_version` honors each existing pin verbatim. Flipping `default_version` only changes what brand-new sources are stamped with. That's why the earlier branch-added repin migration (`0079`) matched zero rows and was correctly dropped. The "Why existing sources are safe" section of the PR description now states this accurately rather than claiming creation-time stamping doesn't exist.

products/warehouse_sources/backend/temporal/data_imports/sources/github/source.py
Gilbert09: Confirmed against GitHub's published REST API-versions changelog (https://docs.github.com/en/rest/about-the-rest-api/api-versions): the "Supported API versions" table lists `2026-03-10` (currently supported, sunset "not yet scheduled") and `2022-11-28` (supported, end of support March 10, 2028). So both the new default and the legacy pin are real, accepted `X-GitHub-Api-Version` labels — no typo or not-yet-released date.

PR #71660, opened 2026-07-16
feat(warehouse_sources): support Stripe API version 2026-06-24.dahlia

Gilbert09:
## Problem

Stripe's latest stable vendor API version is `2026-06-24.dahlia`. Our Stripe warehouse source only declared `2024-09-30.acacia`, so new sources were stamped on an older (still-working but superseded) version and dahlia couldn't be pinned at all.

## Changes

- Declare `2026-06-24.dahlia` in `StripeSource.supported_versions` and make it the `default_version`, so **new** sources are created on it.
- `2024-09-30.acacia` stays supported. Existing sources keep their pin and are byte-for-byte unaffected — the sync path already resolves override → pin → default and threads the resolved version down to `StripeClient(stripe_version=...)`, so no request-layer dispatch change was needed.
- Apply the canonical column hints (`external_table_definitions`) for **every** version, unchanged. Because Stripe has `has_managed_hogql_schema=True`, that same canonical schema also drives the read-side HogQL fields (`hogql_definition`), and the read path isn't `api_version`-aware. Gating only the write hints for dahlia (an earlier draft did) would desync write from read — dahlia's physical columns auto-inferred, but still queried through acacia's canonical mapping. Keeping the hints identical across versions keeps write and read consistent (see the review thread). Real per-version divergence would need the read path versioned too, which needs live dahlia data to author.
- Webhook management (`create`/`delete`/`update`/`get_external_webhook_info`) stays pinned to acacia: the per-instance pin isn't reachable at source-setup time and endpoint CRUD responses are version-stable. Replaced the hardcoded literal with the acacia constant and documented why.

This is a plain update, not a sunset — acacia is **not** deprecated, and no repin migration is included.

> [!NOTE]
> Existing acacia-pinned sources are untouched. Only sources created after this ships default to dahlia. Repinning an existing source remains a manual, human-run action.

Automated only (no manual sync run). Added `TestStripeApiVersioning` to `test_stripe_source.py`:

- `test_dahlia_is_default_and_acacia_still_supported` — guards the "new default + keep old" requirement (catches a default flip-back or a dropped acacia label).
- `test_pinned_version_reaches_stripe_client` (parameterized over acacia + dahlia) — asserts the resolved pin reaches `StripeClient(stripe_version=...)`, catching a regression that hardcodes the version in the request layer.
- `test_column_hints_consistent_across_versions` (parameterized over acacia + dahlia) — both versions get the same canonical hints; catches a regression that would drop a version's write hints and desync it from the version-blind read schema.

Ran the full `test_stripe_source.py` file and the registry-invariant `test_source_versions.py` — 929 passed.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (Opus 4.8) via PostHog Code, directed by @Gilbert09. Invoked the `/warehouse-source-new-version` skill (dispatch + pinning conventions) and `/writing-tests` (test value gate) while producing this PR.

The dispatch plumbing was already in place from the earlier "set your Stripe API version" work, so the change reduced to declaring the label and flipping the default. Column hints stay identical across versions: an earlier draft gated the write hints for dahlia, but review surfaced that the managed HogQL read schema is version-blind, so gating only the write side would desync write from read — keeping the hints the same for every version keeps both consistent. I considered threading the sync pin into webhook creation but rejected it: it would touch the shared `WebhookSource` ABC and, without the per-instance pin available there, could only use `default_version` — which would flip existing acacia sources' webhooks to dahlia and break the "existing sources unaffected" guarantee. Appended these learnings to the skill for the next version bump.

---


danielcarletti:
## Review: PR #71660 — feat(warehouse_sources): support Stripe API version 2026-06-24.dahlia

**ELI10:** PostHog copies data out of Stripe. Stripe released a newer version of its data format ("dahlia"). This PR teaches PostHog to use dahlia for brand-new connections while old connections keep using the version they were already on.

**Goal achieved:** partial — new sources now default to dahlia and old ones stay on acacia, but the read-side managed schema still forces acacia's field shape onto dahlia data.
**Sensible approach:** yes — declaring the label + gating write-side hints is the minimal change, and the existing dispatch plumbing is reused correctly.
**Breaks dependents:** no — existing sources were backfilled to `2024-09-30.acacia` in migration `0075`, so `resolve_api_version` returns acacia for them; webhook clients now use the acacia constant with unchanged value; registry-invariant test still passes.
**Undisclosed side effects:** dahlia-pinned sources are still queried through the acacia canonical HogQL schema (see finding below) — the write-side change is decoupled from the read side.

### Findings

- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/stripe/stripe.py:632 — The version gate only covers the write-side `column_hints`, but Stripe has `has_managed_hogql_schema=True`, and at query time `WarehouseTable.hogql_definition` (products/warehouse_sources/backend/models/table.py:740-744) unconditionally **replaces** the table's fields with the acacia-based `external_tables["stripe_<endpoint>"]` canonical schema — there is no `api_version` awareness in `get_dlt_mapping_for_external_table` or `external_tables`. For a dahlia source, the physical delta columns are now auto-inferred from dahlia's reshaped data (`column_hints=None`), yet the queryable HogQL fields still come from acacia's definitions, which rename/wrap raw columns (`customer` -> `customer_id`, `created` -> `created_at`, per the table.py:344 comment). If dahlia renamed, removed, or retyped any canonically-referenced raw column, those HogQL columns will read null or fail — and setting `column_hints=None` actually widens the gap versus acacia (where write hints and read schema were both acacia-consistent). This was never exercised end-to-end (description: "Automated only, no manual sync run"). Fix: either author dahlia-specific `external_table_definitions` entries and version-gate the read path in `hogql_definition`, or verify that dahlia leaves every canonically-referenced raw column unchanged (and if so, keep the acacia hints for dahlia too so write/read stay consistent), or at minimum disclose the limitation that dahlia sources are queried through the acacia canonical schema.

### Verdict
The version-declaration, default flip, backfill-backed compatibility for existing sources, webhook-pin reasoning, and tests are all sound — this is a clean, well-scoped change on the write side. The one real gap is that the managed HogQL read schema (`has_managed_hogql_schema` / `external_tables`) is not version-aware, so newly created dahlia sources are still exposed through acacia's canonical column mapping. That is the must-address item before dahlia sources can be trusted to query correctly; confirm dahlia's canonically-referenced columns are unchanged or gate/branch the read path.

products/warehouse_sources/backend/temporal/data_imports/sources/stripe/stripe.py
```diff
@@ -623,8 +625,12 @@ def stripe_source(
     api_version: str,
     should_use_incremental_field: bool = False,
 ):
-    column_mapping = get_dlt_mapping_for_external_table(f"stripe_{endpoint.lower()}")
-    column_hints = {key: value.get("data_type") for key, value in column_mapping.items()}
+    # Canonical column hints were built against specific API versions; newer versions reshape some
+    # fields, so only apply hints for hint-compatible versions and let others auto-infer the schema.
+    column_hints: dict[str, Any] | None = None
+    if api_version in HINT_COMPATIBLE_VERSIONS:
+        column_mapping = get_dlt_mapping_for_external_table(f"stripe_{endpoint.lower()}")
```
danielcarletti: [should-fix] The gate only covers write-side column_hints, but Stripe has has_managed_hogql_schema=True: at query time hogql_definition (models/table.py:740-744) replaces the table fields with the acacia-based external_tables["stripe_<endpoint>"] canonical schema, and neither get_dlt_mapping_for_external_table nor external_tables is api_version-aware. So a dahlia source gets auto-inferred physical delta columns but is still exposed through acacia's canonical mapping (which renames raw columns, e.g. customer -> customer_id, created -> created_at). If dahlia renamed/removed/retyped any canonically-referenced column, those HogQL columns read null or fail — and column_hints=None actually widens the gap versus acacia, where write hints and read schema were both acacia-consistent. This path was never exercised end-to-end (automated only, no manual sync run). Fix: author dahlia-specific external_table_definitions entries and version-gate the read path, or verify dahlia leaves every canonically-referenced raw column unchanged (and if so keep the acacia hints for dahlia too so write/read stay consistent), or at minimum disclose the limitation that dahlia sources are queried through the acacia canonical schema.

danielcarletti (commented):
One should-fix: the write-side column_hints version gate is not matched by the read-side managed HogQL schema, so newly created dahlia sources are still queried through acacia canonical field definitions. Details inline and in the PR comment.

danielcarletti:
AI reply: Docs-only blast-radius check for the read-path finding (Stripe changelog, acacia → basil → 2026-06-24.dahlia). Answering the open question from the review: **does dahlia leave every canonically-referenced raw column unchanged? No — it does not.** So the "keep the acacia hints for dahlia too" resolution in `b8577d8` is correct for write/read *consistency* but rests on a premise that doesn't hold for dahlia: the read path is still broken for the columns dahlia removed.

**Why it's still broken.** `has_managed_hogql_schema=True` means `WarehouseTable.hogql_definition` overrides each Stripe table's HogQL fields with the acacia-based canonical map in `external_table_definitions.external_tables`, which is not `api_version`-aware. `column_hints` only *types* columns that exist in the synced data — it can't materialise columns dahlia no longer returns. `structure` in `hogql_definition` is built from the table's real synced columns, but the canonical `fields` still reference the removed physical names, so selecting a removed field (or `SELECT *`) resolves e.g. `DatabaseField(name="charge")` against a `structure` that lacks it → ClickHouse "unknown column", not a silent null. Sync itself succeeds (incremental fields use `created`/`date`, still present), so it only surfaces at query time — and because this PR flips `default_version` to dahlia, every new Stripe source is affected.

**dahlia is chronologically after basil (2025-03-31), so it inherits basil's Invoice/Subscription/InvoiceItem restructuring.** Confirmed removals of raw columns the acacia canonical map references:

- `stripe_invoice`: `subscription` (→ HogQL `subscription_id`), `subscription_details`, `charge`, `payment_intent`, `paid`, `paid_out_of_band` — removed in basil (moved to `parent` / `payments` / the InvoicePayment object).
- `stripe_subscription`: `current_period_start`, `current_period_end` — removed at the subscription level (moved to subscription items).
- `stripe_invoiceitem`: `price`, `unit_amount`, `unit_amount_decimal` — removed (replaced by the polymorphic `pricing` object).

These are three of the most-used billing tables. The other synced endpoints (charge, refund, payout, creditnote, coupon, etc.) need the same per-endpoint audit — I only verified invoice/subscription/invoiceitem against the changelog.

**Fixing it correctly means versioning the READ path** (per-version `external_table_definitions` + threading the source's `api_version` into `hogql_definition`), and the dahlia schema (new `parent` / `payments` / `pricing` / item-level periods) can only be authored and validated against a real dahlia sync — exactly the CAUTION note this PR already added to the skill. A single shared canonical map can't serve both: dropping the removed fields breaks existing acacia sources; keeping them breaks dahlia.

I deliberately did **not** push a speculative read-path change to the shared `table.py` / `external_table_definitions.py`. Guessing dahlia's reshaped schema from docs and shipping it into code every warehouse source hits is the wide-blast-radius risk to avoid, and a fully correct version-aware schema needs a live dahlia sync to author and verify (per the skill note). Two safe options, both a checkpoint decision:

1. Keep dahlia in `supported_versions` but leave `default_version` on acacia until the read path is versioned and validated against a live dahlia sync — no new Stripe source hits the broken managed schema in the meantime.
2. Author the version-aware read schema from a real dahlia sync (all 16 endpoints), gate `hogql_definition` on the source's pinned version, then flip the default.

Changelog sources:
- Invoice `parent` (removes `subscription`, `subscription_details`, `quote`, `subscription_proration_date`): https://docs.stripe.com/changelog/basil/2025-03-31/adds-new-parent-field-to-invoicing-objects
- Invoice partial payments (removes `charge`, `payment_intent`, `paid`, `paid_out_of_band`): https://docs.stripe.com/changelog/basil/2025-03-31/add-support-for-multiple-partial-payments-on-invoices
- Subscription-level period removal: https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-subscription-current-period-start-and-end
- InvoiceItem pricing restructure (removes `price`, `plan`, `unit_amount`, `unit_amount_decimal`): https://docs.stripe.com/changelog/basil/2025-03-31/invoice-pricing-configurations

products/warehouse_sources/backend/temporal/data_imports/sources/stripe/stripe.py
Gilbert09: Good catch — this is exactly why the write-side-only gate was wrong. The managed HogQL read schema (`external_table_definitions` / `hogql_definition`) isn't `api_version`-aware, so gating only `column_hints` for dahlia desynced write from read: dahlia rows got auto-inferred physical columns while still being queried through acacia's canonical mapping, and `column_hints=None` widened the gap rather than closing it.

Resolved in b8577d83: the `HINT_COMPATIBLE_VERSIONS` gate is gone and the acacia canonical hints are now applied for **every** version, so write and read stay consistent (both go through acacia's canonical schema) — the second fix you suggested ("keep the acacia hints for dahlia too so write/read stay consistent"). This matches how acacia behaves today; dahlia is no worse off than acacia is.

Genuine per-version schema divergence would need the read path versioned too (dahlia-specific `external_table_definitions` entries + a version-gated `hogql_definition`), which needs real dahlia sync data to author and verify — out of scope for this label-declaration change, and called out as the follow-up.

PR #71661, opened 2026-07-16
feat(mailerlite): support MailerLite API version v2

Gilbert09:
## Problem

MailerLite's current API (connect.mailerlite.com) is date-versioned through the optional `X-Version` header. When the header is absent the API serves whatever is "latest", so our syncs silently track vendor changes instead of a fixed response shape. The source had no version dispatch at all — it inherited the unversioned `v1` default.

## Changes

Added a `v2` version to the MailerLite source and made it the default for newly created sources. Existing sources keep their `v1` pin and are byte-for-byte unaffected.

- `v1` sends no `X-Version` header (the exact behaviour existing syncs run under).
- `v2` pins `X-Version: 2038-01-19`, the version-pin value MailerLite's docs and official SDK publish, so responses stay on a fixed shape rather than tracking "latest".
- The label → header mapping lives in `settings.API_VERSION_HEADERS`; the resolved version is threaded from the source through `mailerlite_source` → `get_rows` → request headers via `resolve_api_version`.
- Credential validation stays on the unversioned probe path — it runs at creation time before a pin exists, and a list-endpoint probe works regardless of version.

No other sources or the shared retry/error handling are touched.

> [!NOTE]
> `default_version` flips to `v2`, so only new sources start on it. This is a plain version add, not a deprecation — `v1` stays fully supported and nothing is repinned or migrated.

**Why:** MailerLite shipped a newer API behaviour behind `X-Version`, and new sources should pin it explicitly instead of drifting with "latest".

Automated tests only (I'm an agent — no manual sync run):

- `test_mailerlite_source.py`: default is `v2`, version resolution (`resolve_api_version` for null/empty/`v1`/`v2`), and `source_for_pipeline` threads the resolved version into `mailerlite_source` across pins.
- `test_mailerlite.py`: `v1` and the `get_rows` default send no `X-Version` header (guards the byte-for-byte-unchanged invariant for existing syncs), `v2` pins the documented header value.
- Registry invariant suite (`test_source_versions.py`) passes — default ∈ supported, https docs URL, etc.

Ran the MailerLite tests plus the registry invariant test: 908 passed.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (Claude Code). Invoked the `warehouse-source-new-version` skill and followed its version-declaration + dispatch conventions (opaque labels, `resolve_api_version` at the source, keep old versions working, one PR per source). The main decision was how to map the framework's opaque `v1`/`v2` labels onto MailerLite's header-versioned API: `v1` → no header (preserve legacy default-tracking), `v2` → pin the documented `X-Version` date. Credential validation was deliberately left on the unversioned path since it runs before any pin exists. Searched open PRs (including the maintainer's) for existing MailerLite v2 work and found none.

---


danielcarletti:
## Review: PR #71661 — feat(mailerlite): support MailerLite API version v2

**ELI10:** MailerLite's API changes its shape over time. This PR lets us "lock" new connections to a fixed version by sending a date header, so our data imports stop silently drifting, while every connection made so far keeps behaving exactly as before.

**Goal achieved:** yes — adds a `v2` pinned-header version, makes it the default for new sources, leaves existing (`v1`-pinned) sources sending no header.
**Sensible approach:** yes — mirrors the established version-dispatch pattern (`supported_versions`/`default_version`/`resolve_api_version` threaded through `source_for_pipeline`), identical to Stripe.
**Breaks dependents:** no — existing sources were backfilled to `api_version="v1"` (migration `0075`, MailerLite falls to `DEFAULT_API_VERSION="v1"`) and new sources are stamped `default_version` at creation (`external_data_source.py:2168`), so `resolve_api_version` never returns `v2` for a pre-existing row. `mailerlite_source`/`get_rows`/`_get_headers` all default to `v1`, so the unversioned callers (credential validation) are unchanged.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/settings.py:89 — `API_VERSION_HEADERS.get(api_version)` returns `None` for any unmapped label, and `None` also legitimately means "send no header". A future `v3` added to `SUPPORTED_VERSIONS` but forgotten here would silently send no `X-Version` and track "latest" — the exact drift this framework exists to prevent. Fix: assert the map covers `SUPPORTED_VERSIONS` (e.g. a module-level `assert set(API_VERSION_HEADERS) == set(SUPPORTED_VERSIONS)`), or raise on a missing mapping for non-legacy labels so the foot-gun surfaces loudly.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/tests/test_mailerlite_source.py:94 — `test_resolve_api_version` duplicates the generic invariant suite, which already asserts `resolve_api_version(None)`/`("")`/undeclared-label for every source (`test_source_versions.py:50-52`), and `test_source_for_pipeline_plumbing` already covers the `None→v2`/`v1→v1`/`v2→v2` threading. It catches no MailerLite-specific regression. Fix: drop it.

### Verdict
Solid, low-risk change that faithfully follows the existing source-versioning convention; the blast-radius claim ("existing sources byte-for-byte unaffected") holds because of the `0075` backfill and creation-time stamping. No blockers — only a defensive guard on the header map and one redundant test worth trimming. Note: the two red CI checks (Unit Tests / MCP Tests) are an unrelated MCP snapshot drift ("MCP unit test artifacts are out of date"), not caused by this diff.

products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/settings.py
```diff
@@ -71,3 +71,22 @@ class MailerLiteEndpointConfig:
...
+# fixed shape instead of silently tracking "latest".
+MAILERLITE_V1 = "v1"
+MAILERLITE_V2 = "v2"
+
+SUPPORTED_VERSIONS: tuple[str, ...] = (MAILERLITE_V1, MAILERLITE_V2)
+DEFAULT_VERSION = MAILERLITE_V2
+
+# `None` means "send no `X-Version` header". `2038-01-19` is the version-pin value MailerLite's
+# own docs and official SDK publish for locking the API version.
+API_VERSION_HEADERS: dict[str, str | None] = {
```
danielcarletti, line 89: [nit] `API_VERSION_HEADERS.get(api_version)` returns `None` for any unmapped label, and `None` also legitimately means "send no header". A future `v3` added to `SUPPORTED_VERSIONS` but forgotten here would silently send no `X-Version` and track "latest" — the exact drift this framework exists to prevent. Fix: assert the map covers `SUPPORTED_VERSIONS` (e.g. `assert set(API_VERSION_HEADERS) == set(SUPPORTED_VERSIONS)`), or raise on a missing mapping for non-legacy labels.

products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/tests/test_mailerlite_source.py
```diff
@@ -77,10 +81,28 @@ def test_get_resumable_source_manager_bound_to_resume_config(self) -> None:
...
+        source = MailerLiteSource()
+        assert source.default_version == MAILERLITE_V2
+        assert set(source.supported_versions) == {MAILERLITE_V1, MAILERLITE_V2}
+        assert source.default_version in source.supported_versions
+
+    @pytest.mark.parametrize(
+        ("pin", "expected"),
+        [(None, MAILERLITE_V2), ("", MAILERLITE_V2), (MAILERLITE_V1, MAILERLITE_V1), (MAILERLITE_V2, MAILERLITE_V2)],
+    )
+    def test_resolve_api_version(self, pin: str | None, expected: str) -> None:
```
danielcarletti: [nit] `test_resolve_api_version` duplicates the generic invariant suite, which already asserts `resolve_api_version(None)`/`("")`/undeclared-label for every source (`test_source_versions.py:50-52`), and `test_source_for_pipeline_plumbing` already covers the None->v2 / v1->v1 / v2->v2 threading. It catches no MailerLite-specific regression. Fix: drop it.

danielcarletti (commented):
Two nits only. Blast radius is safe (existing sources are v1-pinned via the 0075 backfill + creation-time stamping), and the approach matches the Stripe versioning pattern.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/settings.py
Gilbert09: Done in 430cca04 — added `assert set(API_VERSION_HEADERS) == set(SUPPORTED_VERSIONS)` right below the map so a supported version missing its header mapping fails loudly at import time instead of silently falling through to `None` and tracking "latest".

products/warehouse_sources/backend/temporal/data_imports/sources/mailerlite/tests/test_mailerlite_source.py
Gilbert09: Dropped in 430cca04 — `test_resolve_api_version` was removed since the generic invariant suite plus `test_source_for_pipeline_plumbing` already cover the None->v2 / v1->v1 / v2->v2 threading with no MailerLite-specific regression left uncovered.

PR #71663, opened 2026-07-16
feat(kustomer): support Kustomer API version v2

Gilbert09:
## Problem

Kustomer's API reference exposes both a `v1.0` and a newer `v2` version. The `kustomer` data-warehouse source only declared and supported `v1`, so newly created sources couldn't be created on the current vendor version.

## Changes

- Declare `v2` in `supported_versions` alongside `v1`, and make `v2` the `default_version` so **new** sources are created on it. Existing sources keep their current pin and are unaffected — nothing is migrated or repinned, and `v1` is **not** deprecated.
- **The version pin is declaration-only; it does not change request paths.** Kustomer serves the six declared list resources (customers, conversations, users, teams, tags, brands) under `/v1/` for **both** the v1.0 and v2 doc toggles — the "v2" toggle is a docs-site navigation version, not an API path segment. Building `/v2/<resource>` would 404 every stream. So the request layer is untouched: every version pin resolves to the same `/v1/<resource>` requests. The pin is still recorded and exposed by the framework without any request-layer branch.
  - `v1`-pinned and `v2`-pinned/new sources produce byte-for-byte identical requests (`/v1/customers`, …).
- Credential validation still probes the `/v1/customers` endpoint — it is version-agnostic.

If any of these resources genuinely moves under `/v2/` on the vendor side in future, that would need a per-version path branch in the endpoint catalog (see the note added to the `warehouse-source-new-version` skill).

Automated tests only (no manual/live API calls). Paths verified against the vendor reference (developer.kustomer.com) and the Singer `tap-kustomer` client, which hardcodes `API_VERSION = "v1"` for these resources.

- `test_kustomer_source.py::test_source_requests_v1_rest_paths_for_every_version` — parameterized over `v1`/`v2`/unpinned **and all six resources**, asserts each stream requests `/v1/<resource>`. Catches both a version leaking into the path (a `/v2/` regression) and a per-resource `/v2/` typo in the endpoint catalog.
- `test_kustomer_source.py::test_version_declaration_keeps_v1_and_defaults_to_v2` — guards that `v1` stays supported, `v2` is the default, and nothing is deprecated.
- Full Kustomer suite (62 passed) and the registry-wide version-invariant test (858 passed).

## Docs update

No in-repo docs reference the source's API version; the version metadata is exposed automatically via the source config API.

## 🤖 Agent context

Invoked the `/warehouse-source-new-version` skill for the version-declaration conventions and `/writing-tests` before touching tests. Ruled out a duplicate by searching open PRs (by version string, module path, and the maintainer's own open PRs) — none address Kustomer v2.

---


danielcarletti:
## Review: PR #71663 — feat(kustomer): support Kustomer API version v2

**ELI10:** Kustomer's website shows a "v1" and a "v2" label for its API. This PR tells PostHog to build new Kustomer connections against "v2" by swapping the version number in the web address. But the actual v2 pages still fetch data from the old `/v1/` addresses, so pointing new connections at `/v2/` may send them to a door that doesn't exist.

**Goal achieved:** no — the goal is to let new sources use the current vendor version, but the six declared resources are all still served at `/v1/`, so defaulting to `v2` points new sources at paths that Kustomer's own reference does not document.
**Sensible approach:** alternative worth considering — the version-as-URL-segment dispatch is a fine mechanism, but flipping `default_version` to `v2` is premature until a `/v2/<resource>` endpoint is verified to exist.
**Breaks dependents:** at risk — no external callers of `get_rows`/`kustomer_source` (checked, all in-module and updated), but newly created sources are at risk of 404ing on every stream.
**Undisclosed side effects:** none — existing `v1` pins are untouched; behavior change is confined to new/`v2`-pinned sources.

### Findings

- [blocker] `products/warehouse_sources/backend/temporal/data_imports/sources/kustomer/source.py:36` — `default_version = "v2"` makes every newly created source resolve to `v2`, and `kustomer.py:101` then builds `/{api_version}/{config.resource}` = `/v2/customers`, `/v2/conversations`, etc. Kustomer's API reference documents all six declared resources (customers, conversations, users, teams, tags, brands) as `GET /v1/<resource>` — even under the "v2" docs toggle the List Customers and List Conversations endpoints are shown as `https://api.kustomerapp.com/v1/...` (verified against the vendor reference). So new sources would request paths the vendor does not serve and 404 on every stream. All tests are mocked, so nothing catches this. The PR's own agent note flags that these resources "still serve ... under `/v1/` paths" but then concludes it is "version-compatible" — that conflates response-shape compatibility with the path actually existing at `/v2/`. Fix: keep `default_version = "v1"` (and leave the `v2` dispatch dormant) until a `/v2/<resource>` endpoint is confirmed to exist against the live API; only flip the default once at least one real request to `/v2/customers` is verified to return data, not a 404.

### Verdict
Mechanically the change is clean — signature threading is complete, no external callers, `v1` pins stay byte-for-byte identical, and the security/resume paths are untouched. But the central decision — making `v2` the default and constructing `/v2/<resource>` URLs — rests on an assumption the vendor reference contradicts: these six list resources are documented at `/v1/`, so new sources would likely 404. This must be verified against the live API (or the default reverted to `v1`) before merge; the mocked tests only prove the code assembles a `/v2/` URL, not that Kustomer serves one.

products/warehouse_sources/backend/temporal/data_imports/sources/kustomer/source.py
```diff
@@ -32,8 +32,8 @@
 
 @SourceRegistry.register
 class KustomerSource(ResumableSource[KustomerSourceConfig, KustomerResumeConfig]):
-    supported_versions = ("v1",)
-    default_version = "v1"
+    supported_versions = ("v1", "v2")
+    default_version = "v2"
```
danielcarletti: [blocker] default_version = "v2" makes new sources resolve to v2, and kustomer.py:101 builds /{api_version}/{resource} = /v2/customers, /v2/conversations, etc. Kustomer's API reference documents all six declared resources (customers, conversations, users, teams, tags, brands) as GET /v1/<resource> — even under the "v2" docs toggle the List Customers/Conversations endpoints are shown as https://api.kustomerapp.com/v1/.... So new sources would hit paths the vendor does not serve and 404 on every stream. All tests are mocked, so nothing catches this; the agent note conflates response-shape compatibility with the /v2/ path actually existing. Fix: keep default_version = "v1" (leave v2 dispatch dormant) until a /v2/<resource> endpoint is confirmed live to return data, not a 404.

danielcarletti (commented):
One blocker: flipping default_version to v2 builds /v2/<resource> URLs, but Kustomer serves these six list resources at /v1/ even under its "v2" docs toggle (verified against the vendor reference). New sources would likely 404 on every stream, and all tests are mocked so nothing catches it.

danielcarletti:
AI reply: Verified the /v1-vs-/v2 path concern against the live Kustomer reference (developer.kustomer.com and the Singer tap-kustomer client, which hardcodes `API_VERSION = "v1"`): all six declared list resources — customers, conversations, users, teams, tags, brands — are served under `/v1/` for both the v1.0 and v2 doc toggles; the "v2" toggle is a docs-site nav version, not an API path segment. The branch HEAD already keeps every version pin on the `/v1/<resource>` paths (no `/{version}/{resource}` construction), so a v2 default cannot 404 the streams. I strengthened the existing guard test to parameterize over all six resources (not just customers), so a per-resource `/v2/` typo in the endpoint catalog is caught. Full Kustomer suite (62) and the registry-wide version-invariant test (858) pass.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/kustomer/source.py
Gilbert09: Good catch, and addressed. Two things now guard this:

1. **Default pinned back to v1.** `default_version = "v1"` — v2 stays in `supported_versions` but dormant, so new sources are not stamped onto a version whose dedicated `/v2/<resource>` path has never been confirmed to return data. That is exactly the conservative posture you asked for.
2. **Every resource is hardcoded to `/v1/`.** The endpoint catalog in `settings.py` pins all six resources to `/v1/<resource>` regardless of the version pin, and a new parametrized test (`test_source_requests_v1_rest_paths_for_every_version`, over `None`/`v1`/`v2`) asserts the request path stays `/v1/{resource}` for all of them — so even a v2-pinned source cannot 404 on a `/v2/` path.

Net effect: the PR is a declaration-only add of v2 to `supported_versions`; the request layer is untouched and continues to hit `/v1/`.

PR #71664, opened 2026-07-16
feat(warehouse_sources): support Lightspeed Retail API version 2026-01

Gilbert09:
## Problem

Lightspeed Retail (X-Series) moved from semantic API versions to date-based versioning (`YYYY-MM`). The legacy `2.0` version this source implements is deprecated, and deprecated endpoints increasingly answer `410 Gone` before an eventual retirement. The source needs to support the current `2026-01` version and steer customers off `2.0`, without moving anyone who is pinned to a supported version.

Vendor references: [introduction](https://x-series-api.lightspeedhq.com/docs/introduction), [versioning strategy](https://x-series-api.lightspeedhq.com/reference/versioning-strategy).

## Changes

- **New version `2026-01`** in `LightspeedRetailSource.supported_versions`, with `default_version` flipped to it so new sources start there. Pinned rows keep their pin.
- **Dispatch.** X-Series carries the version as an `/api/<version>` URL path segment, not a header, so the rest_source client's `base_url` is built from the resolved pin. `source_for_pipeline` passes `self.resolve_api_version(inputs.api_version)` into `lightspeed_retail_source`, which builds `_base_url(domain_prefix, api_version)`. No per-version modules: response shapes, the `version` keyset cursor, and pagination are identical across the two versions.
- **The credential probe now uses the pin too.** `validate_credentials` takes the resolved version and probes `/api/<pin>/outlets`. Previously it built its URL from a default, so a source pinned to `2.0` was validated against the wrong base URL. `get_schemas` is a static endpoint catalog with no vendor call, so it stays version independent.
- **Deprecation.** `2.0` goes into `deprecated_versions` with `sunset_at=None`, since the vendor publishes no firm date. The in-product warning and the API fields come from that metadata, so there is no source-specific UI work. I also added `410 Client Error: Gone` to `get_non_retryable_errors`, because a retired version answers 410 forever and would otherwise retry until the source looks merely slow.
- **Migration** `0092_repin_lightspeed_retail_api_version`, written and not run. Idempotent `RunPython` that repins source-level `ExternalDataSource.api_version` from `2.0` to `2026-01`, filtered on the exact old value. `NULL` pins already resolve to the new default, so they are left alone. The reverse is a no-op, since a blanket downgrade would clobber rows natively created on the new default. Schema-level `ExternalDataSchema.api_version` overrides are customer-managed and are not touched.

The source auto-infers its schema and no read field, cursor, or pagination behavior differs between the versions, so the repin needs no data or schema transform. Only the pin moves.

> [!NOTE]
> I (actually Claude) rewrote this branch from scratch on current master. The earlier version of this PR threaded `api_version` through hand-rolled helpers (`_build_url`, a custom `get_rows` generator) that the rest_source migration on master deleted, so the branch was conflicting against code that no longer exists. The migration was renumbered from 0079 to 0092 against the current leaf.

Automated tests, all run locally in the Flox env and green:

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/lightspeed_retail/` (70 passed)
- `pytest .../sources/tests/test_source_versions.py` (1246 passed, the registry-wide version invariants)
- `python manage.py makemigrations warehouse_sources --check --dry-run` (no changes detected, so the migration graph is consistent)
- `mypy --cache-fine-grained .` repo-wide (clean), since the request-layer signatures changed
- ruff check and ruff format

What the new tests catch:

- Both request paths honor the pin instead of a hardcoded or defaulted version: the sync path (`base_url` reaching the actual request URL) and the credential probe (`/api/<pin>/outlets`). Parameterized over `2.0` and `2026-01`. The probe case is the bug this PR fixes, so it is the one test I would not drop.
- `source_for_pipeline` and `validate_credentials` resolve a `None` pin to the new default and pass a present pin verbatim. Existing tests covered neither, since neither surface took a version before.

I did not run the migration, and I did not sync against a live Lightspeed account. There are no stored credentials for this vendor here, so the version semantics are verified against the vendor docs only.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Claude Code (Opus 4.8) did the rewrite at my direction. Skills invoked: `/warehouse-source-new-version`, `/django-migrations`, `/writing-tests`.

The starting point was the stale branch, and the call we made early was to not salvage it. The hand-rolled URL builders it patched are gone from master, so a merge would have been a reconstruction anyway. Reimplementing on rest_source kept the diff to the two places the version actually matters: the client `base_url` and the probe URL.

Two deltas from the earlier attempt, both deliberate. The probe now takes the resolved pin rather than defaulting, which fixes a pinned `2.0` source being validated against the `2026-01` base URL. And `_base_url` takes the version as a required argument rather than a defaulted one, so a future caller cannot silently fall back to the newest version.

### DIFF
--- a/products/warehouse_sources/backend/migrations/max_migration.txt
+++ b/products/warehouse_sources/backend/migrations/max_migration.txt
@@ -1,1 +1,1 @@
-0091_alter_externaldatasource_source_type_and_more
+0092_repin_lightspeed_retail_api_version
--- a/products/warehouse_sources/backend/migrations/0092_repin_lightspeed_retail_api_version.py
+++ b/products/warehouse_sources/backend/migrations/0092_repin_lightspeed_retail_api_version.py
@@ -1,0 +1,34 @@
+from django.db import migrations
+
+# Lightspeed Retail (X-Series) deprecated the legacy 2.0 version; deprecated endpoints increasingly
+# return 410 Gone before an eventual retirement. Repin source-level pins onto the current 2026-01
+# date-based version. The endpoints/fields this source reads carry no documented breaking changes
+# across the two versions (it auto-infers its schema and reads the same `version` keyset cursor), so
+# the repin needs no data/schema transform — only the pin moves.
+LIGHTSPEED_RETAIL_SOURCE_TYPE = "LightspeedRetail"
+DEPRECATED_VERSION = "2.0"
+TARGET_VERSION = "2026-01"
+
+
+def repin_lightspeed_retail_api_version(apps, schema_editor):
+    ExternalDataSource = apps.get_model("warehouse_sources", "ExternalDataSource")
+
+    # Only touch source-level pins still on the deprecated version. NULL pins already resolve to the
+    # source's `default_version` (now 2026-01), so they need no update. Filtering on the exact old
+    # value keeps this idempotent — a re-run matches nothing. Schema-level overrides
+    # (`ExternalDataSchema.api_version`) are intentionally customer-pinned and are left untouched.
+    ExternalDataSource.objects.filter(source_type=LIGHTSPEED_RETAIL_SOURCE_TYPE, api_version=DEPRECATED_VERSION).update(
+        api_version=TARGET_VERSION
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("warehouse_sources", "0091_alter_externaldatasource_source_type_and_more"),
+    ]
+
+    operations = [
+        # Reverse is a no-op: once repinned, these rows are indistinguishable from natively-created
+        # ones, so a blanket downgrade would clobber legitimate 2026-01 pins.
+        migrations.RunPython(repin_lightspeed_retail_api_version, migrations.RunPython.noop, elidable=False),
+    ]
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/lightspeed_retail/constants.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/lightspeed_retail/constants.py
@@ -1,0 +1,5 @@
+# Lightspeed Retail (X-Series) API versions are date-based snapshots (YYYY-MM), replacing the
+# legacy semantic labels. One constant per version so the source's version declaration and the
+# request layer (the `/api/<version>` path segment) share a single label — never parsed or ordered.
+LIGHTSPEED_RETAIL_API_VERSION_2_0 = "2.0"
+LIGHTSPEED_RETAIL_API_VERSION_2026_01 = "2026-01"
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/lightspeed_retail/lightspeed_retail.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/lightspeed_retail/lightspeed_retail.py
@@ -38,8 +38,9 @@ def _clean_domain_prefix(domain_prefix: str) -> str:
     return prefix
 
 
-def _base_url(domain_prefix: str) -> str:
-    return f"https://{_clean_domain_prefix(domain_prefix)}.retail.lightspeed.app/api/2.0"
+def _base_url(domain_prefix: str, api_version: str) -> str:
+    # X-Series carries the pinned vendor version as the `/api/<version>` path segment.
+    return f"https://{_clean_domain_prefix(domain_prefix)}.retail.lightspeed.app/api/{api_version}"
 
 
 def _to_version(value: Any) -> Optional[int]:
@@ -116,6 +117,7 @@ def lightspeed_retail_source(
     team_id: int,
     job_id: str,
     resumable_source_manager: ResumableSourceManager[LightspeedRetailResumeConfig],
+    api_version: str,
     should_use_incremental_field: bool = False,
     db_incremental_field_last_value: Optional[Any] = None,
 ) -> SourceResponse:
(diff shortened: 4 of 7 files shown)


danielcarletti:
## Review: PR #71664 — feat(warehouse_sources): support Lightspeed Retail API version 2026-01

**ELI10:** Lightspeed's shop-data service is retiring its old "2.0" doorway, so this teaches PostHog to knock on the new "2026-01" doorway instead, points new connections at it, and quietly moves existing connections over — without disturbing anyone who deliberately picked a specific version.

**Goal achieved:** yes — new version added, made default, threaded into the `/api/<version>` request path, `2.0` deprecated, existing source-level pins repinned via migration.
**Sensible approach:** yes — single path-segment thread with no per-version modules matches the vendor's versioning model; mirrors the Klaviyo pattern. Version is passed as keyword args everywhere, so the param insertion is safe.
**Breaks dependents:** no — `get_rows`/`lightspeed_retail_source`/`_build_url`/`_base_url` are only called within the source module (grep found no external importers), all callers use keyword args, the migration filter `source_type="LightspeedRetail"` matches the enum value in `types.py:219`, and `2.0`-pinned sources still resolve verbatim.

### Findings
- [should-fix] products/warehouse_sources/backend/migrations/max_migration.txt:1 — Migration number collides with the sibling open PR #71530 (Klaviyo), which also adds `0079_repin_klaviyo_api_version` depending on `0078` and rewrites this file to its own `0079`. Whichever merges second becomes a conflicting leaf migration and a `max_migration.txt` conflict, and `ci:preflight` will block it. Fix: sequence the merges and renumber this migration to `0080` (and update `max_migration.txt`) once the other lands, or coordinate so only one holds `0079`.

### Verdict
Solid, well-scoped version bump that follows the established framework machinery and the Klaviyo precedent; logic, blast radius, and tests are clean. The only actionable item is the `0079` migration-number collision with the parallel Klaviyo PR — not a defect in this PR's logic, but it will hard-block the second of the two to merge until renumbered. The two failing CI checks (`Unit Tests` on `tests/integration/manifest.test.ts`, `MCP Tests`) are unrelated frontend/MCP failures, not caused by this change.

products/warehouse_sources/backend/migrations/max_migration.txt
```diff
@@ -1 +1 @@
-0078_alter_externaldatasource_source_type_and_more
+0079_repin_lightspeed_retail_api_version
```
danielcarletti: [should-fix] Migration number collides with sibling open PR #71530 (Klaviyo), which also adds 0079_repin_klaviyo_api_version depending on 0078 and rewrites this file to its own 0079. Whichever merges second becomes a conflicting leaf migration + max_migration.txt conflict, and ci:preflight will block it. Fix: sequence the merges and renumber this migration to 0080 (and update max_migration.txt) once the other lands.

danielcarletti (commented):
One should-fix: the migration number collides with the parallel open Klaviyo PR #71530 (both add 0079 depending on 0078). Everything else — version threading, blast radius, and tests — is clean.

Gilbert09:
Thanks — agreed on the migration-number collision with #71530. Both branches currently add `0079` off `0078`, so it's a structural conflict (two children of `0078`): renaming mine to `0080` now wouldn't remove it, and pointing the dependency at `0079_repin_klaviyo_api_version` isn't possible until that PR lands on master. I'll sequence behind whichever of the two merges first and `rebase_migration` this one onto the new leaf (fixing both the number and the dependency) at that point.

Separately, the failing `Unit Tests` / `MCP Tests Pass` check is unrelated to this PR — it's `services/mcp/tests/integration/manifest.test.ts` fetching the Context-Mill manifest archive from GitHub releases, where a manifest-referenced file is missing from the published archive. This PR touches only Python warehouse-source files, a migration, and a skill doc (nothing under `services/mcp`), so I've left that check alone as out of scope.

danielcarletti:
AI reply: Reviewed the migration-numbering concern (collision with #71530 Klaviyo).

State: `origin/master` is at `0078_alter_externaldatasource_source_type_and_more`. This PR adds `0079_repin_lightspeed_retail_api_version` depending on `0078` — the correct next number after master, linear and mergeable against master today. #71530 is still open and also adds a `0079_repin_klaviyo_api_version` on `0078`.

No code change made here. Renumbering Lightspeed to `0079` is already the state, and it cannot depend on Klaviyo's `0079` because that migration is not on master yet (the dependency would fail the graph and turn CI red). The two are sibling branches from the same parent, so whichever merges second forks the graph + conflicts on `max_migration.txt`.

Resolution is standard: whichever of #71664 / #71530 merges second runs `python manage.py rebase_migration warehouse_sources` to bump to `0080_*` depending on the first-merged `0079_*`. Both repins are idempotent no-op-reverse `api_version` UPDATEs, so ordering is safe either way.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/migrations/max_migration.txt
Gilbert09: Good catch on the sequencing risk. Confirmed the current state: `master` is still at `0078` (`max_migration.txt` = `0078_alter_externaldatasource_source_type_and_more`), and #71530 (Klaviyo) hasn't merged yet, so `0079` remains the correct next number here and there's no live conflict.

This is a merge-ordering matter that can only be settled at merge time — I can't preemptively renumber to `0080` without leaving a gap if Klaviyo doesn't land first. If the Klaviyo PR merges before this one, I'll rebase, renumber this migration to the next free slot, and update `max_migration.txt` accordingly (the pre-push `ci:preflight` migration-conflict check will catch it either way).

PR #71665, opened 2026-07-16
feat(linkedin_ads): support LinkedIn Ads API version 202606

Gilbert09:
## Problem

LinkedIn's Marketing (LMS) API uses monthly date-based versioning (YYYYMM) sent as a request header. The current active version is 202606, and older versions sunset on a rolling ~1-year clock. The `LinkedinAds` source was pinned to a single hard-coded header behind the default `v1` label, so new sources weren't landing on the latest supported version.

**Why:** keep the pinned vendor API version current so newly created LinkedIn Ads sources track the latest monthly release, without disturbing existing connections.

## Changes

- Declare `supported_versions = ("v1", "202606")` on `LinkedInAdsSource` and set `default_version = "202606"`, so new sources are created on 202606.
- Add minimal version dispatch: the source maps its opaque version label to the LinkedIn `version_string` header and threads it through `LinkedinAdsClient`. The legacy `v1` pin keeps sending the header it always sent (202508), so existing syncs are byte-for-byte unchanged. An unrecognized pin is honored verbatim.
- Credential-validation and account-listing paths keep the legacy default header.
- Point `api_docs_url` at the vendor versioning page.

Existing sources keep their current pin — no migration, repin, or deprecation. Per the vendor changelog, 202606 is a header-only bump for the ad endpoints we sync (accounts, campaigns, campaign_groups, creatives, analytics): no field renames/removals, pagination changes, or new required headers, and we request a fixed `fields` set — so no schema or request-path branching is needed.

Automated tests (run locally, all green):
- `test_source_for_pipeline_dispatches_resolved_api_version` — locks the label→header mapping: `v1`→`202508` (byte-for-byte legacy compat), `202606`→`202606`, no-pin→`202606` (new default), and verbatim passthrough of an undeclared pin.
- `test_request_sends_configured_api_version` — the configured version reaches the Restli `version_string` at the client boundary.
- `test_defaults_new_sources_to_202606` — guards the default bump and supported-version set.
- The registry invariant suite (`test_source_versions.py`) validates the declaration automatically.

Did not exercise a live LinkedIn sync (no credentials available in this environment).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No doc change needed — the default-version bump is transparent to users and the source's posthog.com docs page is unaffected.

## 🤖 Agent context

Invoked the `/warehouse-source-new-version` skill and followed its version-declaration + dispatch conventions, and `/writing-tests` before adding tests. The source had no prior version dispatch — the framework version label was the default `v1` while the client hard-coded the real YYYYMM header — so the main decision was introducing dispatch that keeps `v1` mapped to the existing header for byte-for-byte compatibility while adding 202606. Verified against the vendor changelog that 202606 needs no schema/request-path branching for our endpoints. Appended both learnings to the skill for the next version bump.

---


danielcarletti:
## Review: PR #71665 — feat(linkedin_ads): support LinkedIn Ads API version 202606

**ELI10:** LinkedIn changes the "date stamp" it wants on every request every month. This teaches the LinkedIn source about the new June 2026 stamp and makes brand-new connections use it, while leaving already-set-up connections exactly as they were.

**Goal achieved:** yes — new sources default to `202606`; the `v1` label still sends `202508`, so existing syncs are unchanged.
**Sensible approach:** yes — follows the repo's version-dispatch convention (label→header map, `resolve_api_version`, `supported_versions`/`default_version`).
**Breaks dependents:** no — all changed signatures gained defaulted params; existing sources were backfilled to `"v1"` in migration `0075` (LinkedinAds absent from the map → `DEFAULT_API_VERSION="v1"`), which maps to `202508`, so no in-flight sync changes. Callers verified.
**Undisclosed side effects:** none — the account-listing/probe path staying on the legacy header is disclosed; see finding below for the consequence that isn't spelled out.

### Findings
- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/source.py:163 — `get_oauth_accounts` builds its client via `linkedin_ads_client_for_integration(integration_id, team_id)` with no `api_version`, so account listing uses the default `API_VERSION="202508"`. This PR moves new-source *syncs* to `202606` but leaves the OAuth account-picker on `202508`, the oldest declared header. Per the description LinkedIn sunsets versions on a ~1-year clock, so `202508` (Aug 2025) is due to sunset ~mid-2026 — after which the account picker breaks for even brand-new `202606` sources, blocking new-source setup, while their syncs would still work. Fix: thread the source's current default header into the listing/probe path, e.g. `linkedin_ads_client_for_integration(integration_id, team_id, api_version=_API_HEADER_BY_VERSION[self.default_version])`, so listing tracks the default rather than the oldest label.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/source.py:265 — `resolve_api_version(inputs.api_version)` is a redundant no-op: `SourceInputs.api_version` is already resolved upstream at `import_data_sync.py:220` (`resolve_api_version(schema.api_version or model.pipeline.api_version)`), and `resolve_api_version` is idempotent. Either drop it and map `inputs.api_version` directly, or add a one-word note that it's belt-and-suspenders, so a reader doesn't assume `source_for_pipeline` receives a raw pin.

### Verdict
Solid, convention-following change with genuine byte-for-byte compat for existing sources (verified against the backfill migration and call sites), and appropriately scoped tests. No blockers. The one thing worth acting on before merge is the account-listing path staying pinned to the soon-to-sunset `202508` header — that's a latent break for new-source setup that this PR introduces by advancing the sync version but not the listing version. The double-resolution is a harmless nit.

products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/client.py
```diff
@@ -23,6 +23,8 @@
 LINKEDIN_SPONSORED_URN_PREFIX = "urn:li:sponsored"
 MAX_PAGE_SIZE = 1000
 CREATIVES_PAGE_SIZE = 100  # creatives backend returns transient 500s on heavier requests
+# Default LinkedIn Marketing API version header. Callers pass a resolved version through the
+# source's version dispatch; this default backs the legacy `v1` pin and credential-probe paths.
```
danielcarletti, line 27: [should-fix] This default (202508) also backs the OAuth account-listing path: `get_oauth_accounts` (source.py:163) builds its client via `linkedin_ads_client_for_integration(integration_id, team_id)` with no `api_version`. This PR advances new-source *syncs* to 202606 but leaves the account picker on 202508 — the oldest declared header. Per the description LinkedIn sunsets versions on a ~1yr clock (Aug 2025 -> ~mid-2026), so once 202508 sunsets the account picker breaks for brand-new 202606 sources, blocking setup even though their sync works. Fix: thread the current default header into the listing/probe path, e.g. `linkedin_ads_client_for_integration(integration_id, team_id, api_version=_API_HEADER_BY_VERSION[self.default_version])`.

danielcarletti (commented):
Reviewed the 202606 version bump. Compat for existing sources checks out (backfill -> v1 -> 202508). One should-fix: the OAuth account-listing path stays pinned to the soon-to-sunset 202508 header while syncs advance to 202606, a latent break for new-source setup. Plus a redundant re-resolve nit.

products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/source.py
```diff
@@ -247,12 +261,16 @@ def source_for_pipeline(
         resumable_source_manager: ResumableSourceManager[LinkedInAdsResumeConfig],
         inputs: SourceInputs,
     ) -> SourceResponse:
+        # Honor the instance's pin verbatim when it isn't a label we map, letting LinkedIn validate it.
+        resolved_version = self.resolve_api_version(inputs.api_version)
```
danielcarletti, line 270: [nit] `resolve_api_version(inputs.api_version)` is a redundant no-op: `SourceInputs.api_version` is already resolved upstream at `import_data_sync.py:220`, and `resolve_api_version` is idempotent. Fix: map `inputs.api_version` directly, or note it's belt-and-suspenders so a reader doesn't assume a raw pin arrives here.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/client.py
Gilbert09: Good catch — fixed in `324511003c`. `get_oauth_accounts` now builds its client with `api_version=_API_HEADER_BY_VERSION[self.default_version]`, so the account picker tracks the version new sources are created on (202606) rather than the oldest declared header. See source.py:164-167.

products/warehouse_sources/backend/temporal/data_imports/sources/linkedin_ads/source.py
Gilbert09: Agreed it is idempotent. Kept the call and updated the comment (source.py:268-269) to note it is belt-and-suspenders — so a reader does not assume a raw pin arrives here, and it stays correct if the entry point ever changes. Thanks!

PR #71667, opened 2026-07-16
feat(warehouse_sources): support Notion API version 2026-03-11

Gilbert09:
## Problem

Notion shipped a newer stable API version, `2026-03-11`, that is newer than the `2025-09-03` version our Notion data-warehouse source currently pins. We want new sources to start on the current stable version while every existing source keeps working exactly as before.

## Changes

- Declare `2026-03-11` in `NotionSource.supported_versions` next to `2025-09-03`.
- Make `2026-03-11` the `default_version`, so new sources are created on it. Existing sources are not touched or repinned.
- Thread the resolved API version down to the `Notion-Version` request header (`source_for_pipeline` -> `notion_source` -> `get_rows` -> `_build_session`/`_get_headers`), and validate credentials under the default version. Previously the header was a hardcoded constant.

For the read-only endpoints we consume (`/v1/search`, `/v1/users`, `/v1/blocks/{id}/children`, `/v1/comments`), `2026-03-11` only renames response fields (`archived` -> `in_trash`, the `transcription` block type -> `meeting_notes`) and reshapes the append-block-children request, which we never call. The search object filter stays `data_source`. So the only behavioral difference for a sync is the header value, and the auto-inferred schema absorbs the renamed response fields.

> [!NOTE]
> This is a plain version add, not a deprecation. `2025-09-03` stays fully supported and is not marked deprecated. A source pinned to `2025-09-03` resolves to the same header value it always sent, so its syncs are unchanged.

Automated tests only (I, Claude, ran these; no manual sync against live Notion):

- `test_notion_source.py`: added `test_new_sources_default_to_latest_version` (guards a silent revert of the default back to the old version) and `test_source_for_pipeline_threads_resolved_version` (parameterized over no-pin/legacy-pin/new-pin — guards the resolved version actually reaching the request layer).
- `test_notion.py`: parameterized the header test over both versions to lock in that the header carries whatever version is requested.
- Ran the notion test suite plus the registry invariant test `sources/tests/test_source_versions.py` — all green.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (Claude Code). Invoked the `/warehouse-source-new-version` skill for the version-declaration and dispatch conventions, and the `/writing-tests` skill before adding tests.

Approach: I checked the Notion changelog and confirmed the version bump is a header-only change for our read paths, so I kept the dispatch minimal — resolve the pin at the source class and thread the version string into the request headers rather than introducing per-version modules. Ruled out a duplicate by searching open PRs by version string, the notion module, and the maintainer's own open PRs; none address this.

---


danielcarletti:
## Review: PR #71667 — feat(warehouse_sources): support Notion API version 2026-03-11

**ELI10:** Notion released a newer version of its data feed. This teaches PostHog's Notion importer that the new version exists, uses it for brand-new connections, and quietly leaves every existing connection on the version it already had.

**Goal achieved:** yes — `2026-03-11` is declared and made default, the resolved version is threaded into the `Notion-Version` header, and existing pins keep resolving unchanged.
**Sensible approach:** yes — for read-only endpoints the delta is header-only, so threading the resolved version (rather than adding per-version modules) is the minimal correct dispatch and matches the framework's design.
**Breaks dependents:** no — checked all consumers. `NOTION_VERSION` and the widened function signatures (`_get_headers`, `_build_session`, `validate_credentials`, `get_rows`, `notion_source`) are only used inside the notion package, all updated. Existing sources carry a stored `api_version` (backfill migration `0075` pins Notion at `2025-09-03`; new sources are stamped with `default_version` at creation in `external_data_source.py:2168`), so `resolve_api_version` never falls through to the new default for them — the "existing sources unchanged" claim holds.

### Findings
- [nit] `products/warehouse_sources/backend/temporal/data_imports/sources/notion/canonical_descriptions.py:20` — flipping the default to `2026-03-11` renames `archived` → `in_trash` in responses. `_COMMON_COLUMNS` only describes `archived`, and only the `pages` entry adds an `in_trash` override — so new-default `databases` and `blocks` tables lose the canonical description for that column (falls back to LLM enrichment) while the `archived` entry goes dead for them. Fix (optional): add `in_trash` alongside `archived` in `_COMMON_COLUMNS` so both versions stay covered. (Not part of this diff, so noted here rather than inline.)

### Verdict
Solid, minimal, and correct. The blast radius is contained and the design follows the versioning framework's intent. The only observation is a minor, optional canonical-description gap for the two shared-column tables under the new default — nothing blocking.

danielcarletti (approved):
LGTM

Gilbert09:
Thanks for the review. The canonical-description gap you flagged is now covered — added an \`in_trash\` entry alongside \`archived\` in \`_COMMON_COLUMNS\` (commit is in the branch), so both the pre- and post-2026-03-11 response shapes stay described for the shared-column tables. Rebased on latest master to clear the merge conflict.

PR #71668, opened 2026-07-16
feat(warehouse_sources): support Simplecast API version 2.0

Gilbert09:
## Problem

Simplecast's live REST API is labeled **2.0** (`https://api.simplecast.com`, Bearer auth). The `SimpleCast` warehouse source had no explicit version declaration, so new sources were created on the framework's unversioned placeholder. Legacy 1.0 still works after account migration, but new episode and analytics data is only available via 2.0, so new sources should be created against 2.0.

## Changes

- Declare `supported_versions` and `default_version` on `SimpleCastSource`; **2.0 is the new default** for newly created sources.
- Keep the existing unversioned (`v1`) label supported so already-pinned rows resolve to their current, unchanged wire behaviour. Existing sources are **not** migrated or repinned.
- Thread the resolved `api_version` from `source_for_pipeline` down through `simplecast_source` → `get_rows` → the request headers. Legacy-pinned rows send no version header (requests stay byte-for-byte identical); an explicit vendor version is forwarded via the header Simplecast reserves for version selection.

Note on the header mechanism: Simplecast currently exposes a single live 2.0 API and has announced (but not yet shipped) header-based version selection, promising no breaking changes meanwhile. Forwarding the version is therefore safe today — the API ignores the unknown header — and correct once selection lands. The legacy path sends nothing, so nothing changes for it.

Automated tests only (agent-run); no manual sync against Simplecast.

- `test_simplecast_source.py`: added a case asserting `supported_versions`/`default_version` (catches the default not being bumped or the legacy label being dropped), and a parameterized `source_for_pipeline` test that unpinned/empty pins resolve to 2.0 while `v1` and `2.0` pins are preserved (catches a broken or hardcoded dispatch).
- `test_simplecast.py`: added a parameterized test that `get_rows` builds the HTTP session with the version header for 2.0 and **without** one for the legacy version (guards the byte-for-byte legacy path and the version reaching the request layer).
- Ran the source's test suite (52 passed) plus the registry invariant suite `test_source_versions.py`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing doc changes — this bumps the default vendor API version transparently.

## 🤖 Agent context

Followed the `warehouse-source-new-version` skill. Grounded the versioning mechanism in Simplecast's own docs/help (base URL + Bearer auth is 2.0; header-based version selection announced but not live), which drove the decision to forward the version via a header only for non-legacy pins rather than fabricating a wire change or altering the legacy path. Ruled out a duplicate by scanning open PRs (source path, "simplecast", "2.0", and the maintainer's own open PRs) — none touched this source.

---


danielcarletti:
## Review: PR #71668 — feat(warehouse_sources): support Simplecast API version 2.0

**ELI10:** Simplecast's real API is called "2.0". This PR tells PostHog to label new Simplecast connections as 2.0 and pass that label along, while leaving already-connected accounts exactly as they were.

**Goal achieved:** yes — new sources resolve to 2.0, existing pins are preserved, legacy path is untouched.
**Sensible approach:** alternative worth considering — declaring the versions is right; forwarding a guessed, not-yet-live header today is speculative and could be deferred.
**Breaks dependents:** no — verified via backfill migration `0075` that pre-existing Simplecast rows are pinned to `"v1"` (Simplecast isn't in `API_VERSION_BY_SOURCE_TYPE`, so it falls to `DEFAULT_API_VERSION="v1"`), so `resolve_api_version` returns `v1` → no header → byte-for-byte identical. Only newly created sources (default 2.0) change behavior. `check_access`/`validate_credentials` send no version header. Mirrors the Stripe reference threading pattern.
**Undisclosed side effects:** none — the one behavior change (new sources send `X-Api-Version: 2.0`) is disclosed.

### Findings
- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py:41 — The forwarded `X-Api-Version` header is speculative: Simplecast hasn't shipped header-based version selection, so the header name is a guess and the header is a no-op today (both v1 and 2.0 hit the same live API). It ships an unverified header to production for every new source with no present benefit, and if the eventual header name/format differs this code is revisited anyway — so the "correct once selection lands" upside isn't guaranteed. Fix: consider keeping the `supported_versions`/`default_version` declaration but deferring the wire threading until the real header contract is known (or verify the header name against the live API before merge). Judgment call — the risk is low and it's disclosed.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py:38 — Comment "the header Simplecast reserves for version selection" states the header name as established fact, but per the PR body it's an assumption for an unshipped feature. Fix: soften to reflect that the header name is an expectation, not a confirmed contract.

### Verdict
Small, well-scoped, well-tested change that correctly follows the versioning framework and the Stripe threading precedent; blast radius is genuinely safe (existing sources pinned to `v1` by an already-shipped backfill). The only substantive question is whether to forward the speculative `X-Api-Version` header now versus deferring until the vendor's header contract is real — a disclosed judgment call, not a blocker. Note: the 4 failing CI checks (Frontend formatting/Tests, MCP Tests, Unit Tests) are all frontend/MCP and unrelated to this pure-backend diff (e.g. "MCP unit test artifacts out of date", a `manifest.test.ts` assertion) — likely a stale branch or pre-existing master flake; worth a rebase/master-merge to confirm.

products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py
```diff
@@ -31,8 +32,14 @@ class SimpleCastResumeConfig:
...
 
-def _headers(api_key: str) -> dict[str, str]:
-    return {"Authorization": f"Bearer {api_key}", "Accept": "application/json"}
+def _headers(api_key: str, api_version: str = UNVERSIONED_API_VERSION) -> dict[str, str]:
+    headers = {"Authorization": f"Bearer {api_key}", "Accept": "application/json"}
+    # Legacy-pinned rows send no version header, so their requests stay byte-for-byte identical.
+    # For an explicit vendor version we forward it via the header Simplecast reserves for version
+    # selection; the live 2.0 API ignores an unknown header today and honors it once shipped.
+    if api_version != UNVERSIONED_API_VERSION:
+        headers["X-Api-Version"] = api_version
```
danielcarletti: [should-fix] The forwarded `X-Api-Version` header is speculative: Simplecast has not shipped header-based version selection, so the name is a guess and the header is a no-op today (v1 and 2.0 hit the same live API). It ships an unverified header to production for every new source with no present benefit, and if the eventual name/format differs this code is revisited anyway. Fix: consider keeping the supported_versions/default_version declaration but deferring the wire threading until the real header contract is known (or verify the name against the live API first). Judgment call — low risk and disclosed.

products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py
```diff
@@ -31,8 +32,14 @@ class SimpleCastResumeConfig:
     offset: int = 0
 
 
-def _headers(api_key: str) -> dict[str, str]:
-    return {"Authorization": f"Bearer {api_key}", "Accept": "application/json"}
+def _headers(api_key: str, api_version: str = UNVERSIONED_API_VERSION) -> dict[str, str]:
+    headers = {"Authorization": f"Bearer {api_key}", "Accept": "application/json"}
+    # Legacy-pinned rows send no version header, so their requests stay byte-for-byte identical.
+    # For an explicit vendor version we forward it via the header Simplecast reserves for version
```
danielcarletti: [nit] The comment states the header name as established fact ("the header Simplecast reserves for version selection"), but per the PR it is an assumption for an unshipped feature. Fix: soften to reflect that the header name is an expectation, not a confirmed contract.

danielcarletti (commented):
Two low-severity notes on the speculative version header; blast radius verified safe (existing sources pinned to v1 by backfill 0075). Not a blocker — the header-forwarding-vs-defer question is a disclosed judgment call.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py
Gilbert09: Done — the speculative `X-Api-Version` wire threading is removed. The branch keeps the `supported_versions`/`default_version` declaration (both "v1" and "2.0" resolve to the one live API, so no request branches on the version) but no longer forwards any version header, deferring the wire dispatch until Simplecast ships the real header contract.

products/warehouse_sources/backend/temporal/data_imports/sources/simplecast/simplecast.py
Gilbert09: Done — that comment is removed along with the header threading. The `_headers` helper no longer references a Simplecast version header at all, so there's no longer any claim about a reserved header name.

PR #71669, opened 2026-07-16
feat(warehouse_sources): support Hubspot API version 2026-03

Gilbert09:
## Problem

HubSpot has moved to date-based API versioning (`YYYY-MM`). The version is a URL path segment that replaces the legacy `/crm/v3/` and `/crm/v4/` segments, and new integrations should target the latest stable date version (`2026-03`). Our Hubspot warehouse source only implemented `v3`.

**Why:** keep the Hubspot source current with the vendor's versioning scheme so newly created sources use the supported date version, without disturbing existing sources.

Refs the vendor changelog: <https://developers.hubspot.com/changelog/introducing-date-based-api-versioning>

## Changes

- Add `2026-03` to `HubspotSource.supported_versions` and make it the `default_version` (new sources start on it). `v3` stays fully supported; nothing is deprecated.
- Thread the resolved `api_version` from `source_for_pipeline` down to where CRM URLs are built. A new `apply_crm_api_version` helper rewrites the `/crm/v3/` and `/crm/v4/` path segments onto the date label; for a `v3` pin it returns the path unchanged, so existing syncs are byte-for-byte unaffected.
- Point `api_docs_url` at HubSpot's versioned API reference overview.

Existing sources keep their current pin and are untouched — no migration or repin.

Automated tests only (agent-run):

- `test_source_routing.py::TestApiVersion` — new default resolves to `2026-03`, a `v3` pin is honored verbatim, both versions are supported, `apply_crm_api_version` segment rewriting, and `source_for_pipeline` threads the resolved version into `hubspot_source`. Catches a regression where the default flips back or version threading breaks.
- `test_hubspot_search.py` — parameterized over `v3` and `2026-03` that the search URL and the v4 association batch-read URL carry the pinned version segment. Catches the version failing to reach the HTTP boundary.
- Ran the touched source's tests plus the registry-wide `test_source_versions.py` invariants (973 passed).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by PostHog Code (Claude). Invoked the `warehouse-source-new-version` skill for the version-declaration and dispatch conventions, and `writing-tests` before adding tests. Confirmed via the vendor changelog that date versioning is a URL-path-segment mechanism (`/crm/2026-03/`), which drove the minimal `apply_crm_api_version` rewrite rather than per-version modules — response shapes are built on the latest v3/v4 APIs and stay compatible. OAuth token endpoints (`/oauth/v1/`) are version-independent and left untouched. Checked open PRs (including the maintainer's own) for a duplicate; none addresses this.

---


danielcarletti:
## Review: PR #71669 — feat(warehouse_sources): support Hubspot API version 2026-03

**ELI10:** HubSpot changed how you pick which version of its data API to talk to — instead of `v3`/`v4` in the web address, you now put a date like `2026-03`. This PR teaches PostHog's HubSpot importer to use the new date version for brand-new connections, while leaving every existing connection exactly as it was.

**Goal achieved:** yes — new sources default to `2026-03`, existing pins keep `v3`, and every CRM URL is rewritten through one helper.
**Sensible approach:** yes — a single `apply_crm_api_version` path-segment rewrite (returning `v3` paths unchanged) is the minimal implementation and matches the framework's "pin, never move" design.
**Breaks dependents:** no — verified existing sources are pinned to `v3`: migration `0075_backfill_externaldatasource_api_version.py` stamps `Hubspot → v3`, and `external_data_source.py:2168` stamps `api_version=source.default_version` at creation, so no source is left `None`. `resolve_api_version("v3")` → `"v3"`, and the helper returns v3 paths byte-for-byte. `source_api_versions.py` reads `supported_versions`/`default_version` dynamically, so no other surface needs updating. Only `hubspot_source` is called cross-module and it threads `api_version`.

### Findings
- [should-fix] `products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/source.py:54` — `default_version` flips every newly created HubSpot source to `2026-03`, but the change is exercised only against mocked HTTP; the central assumption that HubSpot serves the v4 association batch-read at `/crm/2026-03/associations/.../batch/read` (v4 collapsing onto the date segment) is unverified against the live API. If wrong, association sync breaks for all new sources. Fix: run one live sync against HubSpot `2026-03` (objects, `/search`, and the batch-read association path) before merge, or land `2026-03` as `supported` while keeping `default_version = v3` until a live sync confirms it.
- [nit] `products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/hubspot.py:260` — the request-issuing functions (`get_rows` here, plus `get_rows_via_search`, `_batch_read_associations`, and `_get_property_names` in helpers.py) default `api_version` to `HUBSPOT_API_VERSION_V3`. A future call site that forgets to thread the resolved version would silently downgrade a `2026-03` source to `v3` instead of failing loudly. Fix: make `api_version` a required keyword-only argument on the functions that actually build/issue CRM URLs.

### Verdict
Clean, well-scoped, and correctly conservative on backward compatibility — the byte-for-byte claim for existing `v3` sources holds up under the migration + creation-stamp check, and all live CRM URLs route through the single rewrite helper. The one real risk is that `2026-03` becomes the default for all new sources with zero live-API validation (only mocked tests), and it hinges on the v4-associations-path-collapse assumption; validate that against a real HubSpot account before merging, or defer the default flip until it's confirmed.

products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/source.py
```diff
@@ -48,9 +50,9 @@ class HubspotSourceOldConfig(config.Config):
 
 @SourceRegistry.register
 class HubspotSource(ResumableSource[HubspotSourceConfig | HubspotSourceOldConfig, HubspotResumeConfig], OAuthMixin):
-    supported_versions = ("v3",)
-    default_version = "v3"
-    api_docs_url = "https://developers.hubspot.com/docs"
+    supported_versions = (HUBSPOT_API_VERSION_V3, HUBSPOT_API_VERSION_2026_03)
+    default_version = HUBSPOT_API_VERSION_2026_03
```
danielcarletti: [should-fix] default_version flips every newly created HubSpot source to 2026-03, but the change is exercised only against mocked HTTP. The central assumption — that HubSpot serves the v4 association batch-read at /crm/2026-03/associations/.../batch/read (v4 collapsing onto the date segment) — is unverified against the live API; if wrong, association sync breaks for all new sources. Fix: run one live sync against HubSpot 2026-03 (objects, /search, and the batch-read association path) before merge, or land 2026-03 as supported while keeping default_version = v3 until a live sync confirms it.

products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/hubspot.py
```diff
@@ -247,6 +257,7 @@ def get_rows(
     include_custom_props: bool = True,
     selected_properties: list[str] | None = None,
     source_id: str | None = None,
+    api_version: str = HUBSPOT_API_VERSION_V3,
```
danielcarletti: [nit] This and the other request-issuing functions (get_rows_via_search, _batch_read_associations, and _get_property_names in helpers.py) default api_version to HUBSPOT_API_VERSION_V3. A future call site that forgets to thread the resolved version would silently downgrade a 2026-03 source to v3 instead of failing loudly. Fix: make api_version a required keyword-only argument on the functions that actually build/issue CRM URLs.

danielcarletti (commented):
Two findings — one should-fix (default flips new sources to 2026-03 with only mocked-HTTP validation of the v4-associations-path assumption), one nit (silent v3 fallback default on request-issuing functions). Backward-compat for existing v3 sources is verified safe.

danielcarletti:
AI reply: docs verification surfaced a real bug in the `2026-03` path rewrite, now fixed.

HubSpot's date-based versioning does **not** replace the `/crm/v3/` or `/crm/v4/` segment in place. Per the vendor docs, the version segment moves to *after* the API resource name:

- `/crm/v3/objects/contacts` -> `/crm/objects/2026-03/contacts`
- `/crm/v3/objects/deals/search` -> `/crm/objects/2026-03/deals/search`
- `/crm/v3/properties/deals` -> `/crm/properties/2026-03/deals`
- `/crm/v4/associations/{from}/{to}/batch/read` -> `/crm/associations/2026-03/{from}/{to}/batch/read`

The previous helper produced `/crm/2026-03/objects/...`, which HubSpot 404s. Since `2026-03` is now the default, **every newly created source would have failed** on the first request. The bug was invisible because the tests asserted the same (wrong) in-place-rewrite shape.

Sources for each path:
- Migration playbook: https://developers.hubspot.com/blog/date-based-api-versioning-migration-playbook ("In a path like `/crm/objects/2026-03/contacts`, `2026-03` is the API version name" and "`/crm/objects/2026-03/{objectType}/search`")
- Developer guide: https://developers.hubspot.com/blog/a-developers-guide-to-hubspots-date-based-api-versioning (pattern `/{api-name}/2026-03/{resource}`)
- Associations batch-read reference: https://developers.hubspot.com/docs/api-reference/latest/crm/associations/associate-records/batch/get-associations (documented path `/crm/associations/2026-03/{fromObjectType}/{toObjectType}/batch/read`)

Fix (`ac28a19`): `apply_crm_api_version` now re-inserts the date segment behind the resource name via a small regex (`/crm/<v3 |v4>/<resource>/` -> `/crm/<resource>/<date>/`); the `v3` pin still returns the path unchanged, so existing syncs stay byte-for-byte. Response shapes are unchanged (date versions are based on the latest v3/v4 APIs), so no per-version parsing needed. Corrected the test expectations and the skill learning note that described the wrong mechanism. All 998 hubspot + `test_source_versions.py` tests pass.

Left the silent-downgrade nit alone: `hubspot_source` already requires `api_version` (the guarded entry point), and making every internal request helper required would churn many call sites for little gain — not low-risk.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/source.py
Gilbert09: Good catch — since this environment has no way to run a live 2026-03 sync (no stored HubSpot credentials / live-sync harness), I've taken the safe fallback you suggested: kept `2026-03` in `supported_versions` for opt-in but reverted `default_version` back to `v3`, so newly created sources stay on the verified v3 paths. The comment in code notes to flip the default once a real end-to-end 2026-03 sync (objects, `/search`, and the association batch-read path) has been confirmed.

products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/hubspot.py
Gilbert09: Done — `api_version` is now a required keyword-only argument (no default) on the functions that actually build/issue CRM URLs: `get_rows`, `get_rows_via_search`, `_batch_read_associations`, and `_get_property_names` in `helpers.py`. A caller that forgets to thread the resolved version now fails loudly (TypeError) instead of silently downgrading a pinned source to v3. The single guarded entry point (`hubspot_source`) already required it, and all internal callers thread it through; the direct-call tests now pass the version explicitly.

PR #71670, opened 2026-07-16
feat(warehouse_sources): support Pipedrive API version v2

Gilbert09:
## Problem

Pipedrive deprecated the API v1 endpoints that have direct v2 replacements, with a vendor sunset date of **2025-12-31**. The Pipedrive source pinned customers to `v1`, which still routed the `activities` collection through the deprecated `/api/v1/activities` endpoint.

## Changes

- Declare `v2` as a supported version and make it the `default_version` (new sources start on v2).
- Mark `v1` deprecated via `deprecated_versions` with `sunset_at=2025-12-31`. The generic in-product deprecation warning keys off this metadata — no source-specific UI.
- Version dispatch is minimal: `v1` and `v2` differ only in the `activities` endpoint (`/api/v1/activities` offset → `/api/v2/activities` cursor; same `id`/`add_time` fields). `endpoints_for_version()` swaps that one config; every other endpoint is version-independent. `notes`, `leads`, `users`, and the `*_fields` endpoints have **no** v2 replacement and stay on v1 under both labels. The resolved version is threaded from `source_for_pipeline` through `self.resolve_api_version(...)`. `v1` remains fully functional; unknown pins fall back to the legacy (v1) activities endpoint.
- **Data migration (written, not run):** `0079_repin_pipedrive_api_version_v2` repins the **source-level** `ExternalDataSource.api_version` from `v1` to `v2` for Pipedrive. It is idempotent (matches only `api_version="v1"`; NULL already resolves to the new default). Schema-level `ExternalDataSchema.api_version` overrides are intentionally left untouched — those are customer-pinned and covered by the schema-level deprecation warning. A human reviews and runs it.
  - No scripted data/schema transform is needed for the `activities` table: it syncs full-refresh with an auto-inferred schema, so the next sync after repinning re-fetches from the v2 endpoint and Delta handles schema evolution.

Automated tests I (the agent) ran (`.venv/bin/pytest`, 926 passed):

- `sources/pipedrive/tests/` — added `TestEndpointsForVersion` (activities resolves v1→offset, v2→cursor, unknown→v1; shared endpoints identical across versions), parameterized `source_for_pipeline` over the resolved version reaching `pipedrive_source`, and a deprecation-metadata test (v1 deprecated with the 2025-12-31 sunset, default is v2). Guards against the version pin being ignored or the wrong endpoint/pagination being used per version.
- `sources/tests/test_source_versions.py` — registry invariants (default in supported, deprecated ⊆ supported, default not deprecated).

The migration was **not** executed (no DB in this environment, and it is written-not-run by design). It is a pure `RunPython` data migration, so it introduces no model-state drift.

## Docs update

Updated the co-located `api_inventory.md` with the source-version split.

## 🤖 Agent context

Invoked skills: `/warehouse-source-new-version` (version declaration, dispatch, pinning semantics, deprecation + migration conventions), `/django-migrations` (data-migration safety), `/writing-tests` (test value gate).

Confirmed against the vendor changelog that `activities` is the only endpoint in the source's set that was on a deprecated v1 endpoint with a v2 replacement, so dispatch is scoped to it alone rather than a blanket path rewrite. Checked open PRs (including the maintainer's) for a pre-existing Pipedrive v1 deprecation fix — none found; #68112 touches Pipedrive field-schema primary keys, a different root cause. Appended a learning to the version skill for the next update.

---


danielcarletti:
## Review: PR #71670 — feat(warehouse_sources): support Pipedrive API version v2

**ELI10:** Pipedrive is turning off an old way of asking for "activities" data at the end of 2025. This PR teaches the Pipedrive importer a new way (v2), makes new connections use it by default, and includes a script to move existing connections over — while carefully leaving alone the few endpoints that have no v2 replacement.

**Goal achieved:** yes — v2 declared + default, v1 deprecated with sunset, dispatch swaps only the `activities` endpoint, migration written to repin existing source-level pins.
**Sensible approach:** yes — mirrors the existing Stripe version-dispatch pattern (`resolve_api_version(inputs.api_version)`) and the shared `deprecated_versions`/`get_version_deprecation` framework; no source-specific UI added.
**Breaks dependents:** no — `pipedrive_source`/`get_rows` gained an `api_version` arg but are only called from `source.py` and Pipedrive-local code (all updated); `PIPEDRIVE_ENDPOINTS` fully renamed with no leftover refs on head; migration `source_type="Pipedrive"` matches the enum value; resume state is keyed per `job_id`, so a repin between runs can't replay a stale v1 offset URL against v2 cursor logic.
**Undisclosed side effects:** none — NULL-pinned existing sources auto-resolve to v2 on next sync (safe: `activities` is full-refresh), which is the intended effect and disclosed.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/pipedrive/settings.py:56 — `endpoints_for_version` falls back to `_ACTIVITIES_BY_VERSION["v1"]` for an unrecognized pin, but v1 is the deprecated endpoint (sunset 2025-12-31, already past). An unknown label therefore resolves to a dead endpoint rather than the current default. Fix: fall back to `_ACTIVITIES_BY_VERSION["v2"]` (the source's `default_version`) so an unrecognized pin degrades to the live endpoint, and update the docstring/comment accordingly.

### Verdict
Solid, tightly-scoped change that reuses the established source-versioning framework and Stripe's dispatch pattern; blast radius is contained to the Pipedrive source and verified against all call sites. Tests are focused and assert observable behavior (per-version endpoint/pagination, deprecation metadata, arg plumbing). No must-fix items — only the low-severity unknown-pin fallback nit.

products/warehouse_sources/backend/temporal/data_imports/sources/pipedrive/settings.py
```diff
@@ -39,4 +42,21 @@ class PipedriveEndpointConfig:
...
+_ACTIVITIES_BY_VERSION: dict[str, PipedriveEndpointConfig] = {
+    "v1": PipedriveEndpointConfig(name="activities", path="/api/v1/activities", pagination="offset"),
+    "v2": PipedriveEndpointConfig(name="activities", path="/api/v2/activities", pagination="cursor"),
+}
+
+
+def endpoints_for_version(api_version: str) -> dict[str, PipedriveEndpointConfig]:
+    """Endpoint configs for a resolved source version. Unknown pins fall back to the legacy
+    (v1) activities endpoint, the conservative choice for a label we don't recognise."""
+    activities = _ACTIVITIES_BY_VERSION.get(api_version, _ACTIVITIES_BY_VERSION["v1"])
```
danielcarletti: [nit] Unknown pin falls back to _ACTIVITIES_BY_VERSION["v1"], but v1 is the deprecated endpoint (sunset 2025-12-31, already past), so an unrecognized label resolves to a dead endpoint rather than the current default. Fix: fall back to _ACTIVITIES_BY_VERSION["v2"] (the source default_version) so an unknown pin degrades to the live endpoint; update the docstring/comment to match.

danielcarletti (commented):
One nit only — the PR is clean and tightly scoped, reusing the existing Stripe version-dispatch pattern and shared deprecation framework. The unknown-pin fallback lands on the now-sunset v1 endpoint.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/pipedrive/settings.py
Gilbert09: Good catch — addressed in the latest commit. `endpoints_for_version` now resolves any label other than an explicit `v1` pin (including unrecognised pins and the default `v2`) to the live v2 cursor `activities` endpoint, so an unknown pin degrades to the current default rather than the sunset v1 one. The docstring was updated to match.

PR #71671, opened 2026-07-16
feat(warehouse_sources): support Brex API version v2

Gilbert09:
## Problem

The Brex data-warehouse source had no version declaration, so it implicitly ran on the framework's unversioned `v1` label with no way to pin or evolve the vendor API version. We want new Brex sources to start on `v2` (Brex's core API base path is `/v2/`) while every existing source keeps working untouched.

## Changes

- Declare `supported_versions = ("v1", "v2")` and set `default_version = "v2"` on `BrexSource`, so new sources are stamped `v2` at creation. Existing rows are NULL-pinned and now resolve to `v2`, which is safe (see below).
- Thread the resolved `api_version` from the source class (`resolve_api_version(inputs.api_version)`) through `brex_source` → `get_rows` → a per-endpoint path resolver.
- Add a `versioned_paths` override map to `BrexEndpointConfig` plus `_resolve_path(config, api_version)` as the consumed dispatch seam. It stays empty today.
- Point `api_docs_url` at the Brex changelog, where versions are announced.
- Extend tests to exercise version resolution, the new default, per-version path dispatch, and that a pinned version reaches the request layer.

> [!NOTE]
> Brex versions each product API independently in the URL path (transactions/users on `/v2`, expenses/vendors on `/v1`) with no global version header or segment. There is no `/v2/expenses` or `/v2/vendors` — those APIs only exist on `/v1`. The source already targets each API's current path, so `v1` and `v2` resolve to byte-for-byte identical requests. That is why flipping the default to `v2` does not repin or migrate existing customers: their requests do not change. The `versioned_paths` seam is where a genuine future per-API bump hooks in. No existing version is deprecated.

Automated only (I, Claude, did not run a live Brex sync). Ran the full Brex source suite plus the registry version-invariant test in the local venv: `pytest products/warehouse_sources/backend/temporal/data_imports/sources/brex/ products/warehouse_sources/backend/temporal/data_imports/sources/tests/test_source_versions.py` — 958 passed.

- `test_supported_versions_and_default` / `test_resolve_api_version_*` — guards against a silent revert of the default back to `v1` or broken pin resolution.
- `test_source_for_pipeline_passes_pinned_api_version` and the added assertion in the plumbing test — guards that the resolved version actually reaches `brex_source` (a drop would sync on the wrong version silently).
- `TestPathResolution` — guards that both versions keep hitting each API's current path (so a naive `/v2/` rewrite of expenses/vendors can't slip in), and that `_resolve_path` honors a per-version override.

## Docs update

No repo docs changed; the Brex `docsUrl` points at posthog.com docs (separate repo).

## 🤖 Agent context

Authored by Claude (Opus 4.8) via PostHog Code, directed by Tom. Invoked the `/warehouse-source-new-version` skill (followed its declaration + dispatch + pinning conventions) and `/writing-tests` before touching tests.

Key decision: the detection note suggested a global `/v2/` bump, but verifying against Brex's API reference and changelog showed versioning is per-API and `/v2/expenses` / `/v2/vendors` do not exist — a blind path rewrite would 404. Chose a nominal default bump with a consumed-but-currently-inert dispatch seam, and confirmed both versions produce identical requests so existing sources need no migration. Appended both findings to the skill's learnings.

---


danielcarletti:
## Review: PR #71671 — feat(warehouse_sources): support Brex API version v2

**ELI10:** Brex (a company card service) numbers its data endpoints like "v1" and "v2" in the web address. This PR teaches PostHog's Brex importer to remember which version a customer is on and gives it a place to point different endpoints at different versions later — but right now it changes nothing about the actual requests, it just labels new connections "v2".

**Goal achieved:** yes — `BrexSource` now declares `supported_versions = ("v1", "v2")`, defaults new sources to `v2`, and threads the resolved version to the request layer.
**Sensible approach:** yes — mirrors the shipped Stripe/Shopify pattern (module-level version constants + `resolve_api_version` + a per-endpoint dispatch seam) and the `warehouse-source-new-version` skill convention.
**Breaks dependents:** no — `get_rows`/`brex_source` are module-internal; the only external caller is `source.py` (updated) and the tests (updated via wrapper). All call sites pass args by keyword, so inserting the required `api_version` param mid-signature breaks nothing. `versioned_paths` is empty, so `_resolve_path` returns `config.path` for every version → v1 and v2 dispatch byte-identical URLs, so NULL-pinned existing sources resolving to `v2` need no migration.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/brex/tests/test_brex.py:451 — `test_dispatched_url_reflects_resolved_version` is parametrized over v1 and v2 but asserts the identical URL for both; `test_resolve_path_matches_current_paths_for_every_version` already proves `_resolve_path` returns the same path per version, so the second version case here catches no distinct regression (it only needs to prove `_resolve_path` is wired into `get_rows`/`_build_url`). Fix: drop the version parametrization and run this wiring guard once.

### Verdict
Solid, low-risk change: the design is the sanctioned framework pattern, the plumbing is correctly scoped (api_version resolved once into `path` and not threaded deeper than needed), and the migration-safety reasoning holds because the seam is inert. The only note is a redundant test parametrization. Worth being aware that the entire change is currently behavioral no-op scaffolding for a future per-API version bump — that is the explicit product intent here, not an accident, so it is fine, but the value lands only when a real per-endpoint version divergence arrives.

products/warehouse_sources/backend/temporal/data_imports/sources/brex/tests/test_brex.py
```diff
@@ -410,3 +425,34 @@ def test_partition_keys_are_stable_posted_dates(self, config):
...
+        # Both versions must keep hitting each API's current path — expenses/vendors have no v2.
+        assert _resolve_path(BREX_ENDPOINTS[endpoint], api_version) == expected_path
+
+    def test_resolve_path_honors_per_version_override(self):
+        config = BREX_ENDPOINTS["users"]
+        overridden = dataclasses.replace(config, versioned_paths={BREX_API_VERSION_V2: "/v3/users"})
+        assert _resolve_path(overridden, BREX_API_VERSION_V2) == "/v3/users"
+        assert _resolve_path(overridden, BREX_API_VERSION_V1) == config.path
+
+    @pytest.mark.parametrize("api_version", [BREX_API_VERSION_V1, BREX_API_VERSION_V2])
```
danielcarletti: [nit] test_dispatched_url_reflects_resolved_version is parametrized over v1 and v2 but asserts the identical URL for both, and test_resolve_path_matches_current_paths_for_every_version already proves _resolve_path returns the same path per version — so the second version case here catches no distinct regression (this test only needs to prove _resolve_path is wired into get_rows/_build_url). Fix: drop the version parametrization and run this wiring guard once.

danielcarletti (commented):
One nit on test redundancy; otherwise a clean, well-tested change that follows the established Brex/Stripe versioning convention and is migration-safe because the version seam is inert (v1 and v2 dispatch identical requests).

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/brex/tests/test_brex.py
Gilbert09: Addressed — the version parametrization on the wiring guard was dropped. `test_resolved_path_is_wired_into_request_url` now runs the get_rows/_build_url wiring check once (v2 only), and per-version path coverage lives solely in `test_resolve_path_matches_current_paths_for_every_version`. No duplicated coverage remains.

PR #71672, opened 2026-07-16
feat(warehouse_sources): support Mixpanel API version 2.0

Gilbert09:
## Problem

The Mixpanel source only ever ran against an implicit unversioned pin (PostHog's
`v1` placeholder), even though Mixpanel's published API version is `2.0`. New
sources should start on the real current version, and we want a version-dispatch
seam so future Mixpanel API versions are a small change rather than a rewrite.

## Changes

- Declare `supported_versions = ("v1", "2.0")` on `MixpanelSource` and flip
  `default_version` to `2.0`, so new sources are stamped with `2.0` at creation.
- Thread the resolved `api_version` from `source_for_pipeline` down through
  `mixpanel_source` → `get_rows` → the raw-export URL builder. The export path
  segment is now derived from the pin via a small `EXPORT_API_PATH_SEGMENT`
  mapping instead of a hardcoded `/api/2.0/export` literal.
- Both supported labels currently map to Mixpanel's `2.0` export path, so
  `v1`-pinned sources are byte-for-byte unaffected. Existing sources keep their
  pin; nothing is migrated or repinned.

The query APIs (engage, cohorts, annotations) stay on Mixpanel's current
`/api/query/` and `/api/app/` paths per the authoritative vendor docs — those
are not version-carrying, so they are untouched.

> [!NOTE]
> This is a plain version add. No version is deprecated and no migration script
> is included.

Automated only (I'm Claude, an agent — I did not run a live Mixpanel sync):

- Extended `test_mixpanel_source.py`: new `TestApiVersions` asserts the `2.0`
  default and the `{v1, 2.0}` supported set, and a parameterized test that
  `source_for_pipeline` threads the resolved version (no pin → `2.0`, `v1` pin →
  `v1`, `2.0` pin → `2.0`). Catches a regression where the version plumbing or
  `resolve_api_version` call is dropped.
- Extended `test_mixpanel.py`: parameterized `test_export_hits_versioned_path`
  verifies the export URL is built from the pin for each supported version.
  Adapted the existing `_iter_export` call sites to the new required arg.
- Ran the full `sources/mixpanel/` suite (90 passed) plus the registry invariant
  test `sources/tests/test_source_versions.py` — all green.
- Ran `ruff check`/`ruff format` over the touched files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change — the version metadata surfaces automatically through
the existing public source-config API.

## 🤖 Agent context

Authored by Claude Code (Claude Opus 4.8). Invoked the `warehouse-source-new-version`
skill for the version-declaration + dispatch conventions and `/writing-tests` for
the test gate.

Key decision: Mixpanel doesn't have a clean `v1` → `2.0` API-path story. The live
vendor docs confirm the query API is served under `/api/query/` (current) while the
raw export lives under `/api/2.0/export`. So the only genuinely version-carrying
path is the export segment. I threaded the resolved version into just that URL
builder and left the query paths alone, which keeps the change minimal and keeps
`v1` behaviour identical. The `EXPORT_API_PATH_SEGMENT` map (both labels → `2.0`
today) is the documented seam for a future export-API version.

---


danielcarletti:
## Review: PR #71672 — feat(warehouse_sources): support Mixpanel API version 2.0

**ELI10:** Mixpanel's data-import connector never had a real "version" label. This PR adds one (`2.0`), makes new connections start on it, and builds a tidy seam so a future Mixpanel API version is a one-line change — without changing anything about how today's syncs actually run.

**Goal achieved:** partial — the version declaration + seam are correct, but the PR ships with red CI (mypy) caused by its own test reformatting.
**Sensible approach:** yes — mirrors the reference Stripe source and the `warehouse-source-new-version` skill; only the export path is version-carrying and only it was threaded.
**Breaks dependents:** no — `mixpanel_source`/`get_rows` are module-local, both gained defaulted args; `_iter_export`'s new required kwarg has all call sites updated; flipping `default_version` to `2.0` maps to the same `/api/2.0/export` path so unpinned/`v1` sources are byte-for-byte unchanged; `test_public_source_configs` asserts only the generic `defaultVersion ∈ versions` invariant (still holds).

### Findings
- [blocker] products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/test_mixpanel.py:289 — CI "Check static typing" is red. The reformatted multi-line `mp._iter_export(  # type: ignore[arg-type]` calls put the ignore on the call's opening line, but mypy attributes the `FakeManager` arg-type error to the line holding the `manager` argument, so each ignore is now flagged `unused-ignore` **and** the real `arg-type` error is unsuppressed. Affects lines 289/290, 361/367, 396/402, 424/430. Fix: move each `# type: ignore[arg-type]` to the line containing the `manager` argument (or keep the whole call on one line as before). Run mypy on the touched files (CI-only per repo convention) before pushing.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/source.py:46 — `api_docs_url` is left unset, so `GET /api/public_source_configs/` will surface `apiDocsUrl: null` for Mixpanel while the reference Stripe source populates it. Fix: set `api_docs_url` to Mixpanel's raw-export/API reference (or changelog) URL so the "where new versions are announced" metadata is populated.

### Verdict
The design is sound and consistent with the established versioning framework, and blast radius is genuinely nil (both labels resolve to the same export path). The one must-fix is the mypy failure the PR introduced: the reformatted `_iter_export` calls stranded their `# type: ignore[arg-type]` comments on the wrong line, turning a green suppression into two errors per call site across four spots. Fix the ignore placement and re-run static typing before merge. (The failing Unit Tests / MCP checks are MCP-manifest drift + a 403, unrelated to this change — worth confirming but not this PR's defect.)

products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/test_mixpanel.py
```diff
@@ -272,14 +276,31 @@ def test_payment_required_gets_actionable_message(self) -> None:
...
+        manager: FakeManager,
+        start: date,
+        end: date,
+        responses: list[FakeResponse],
+        api_version: str = MIXPANEL_API_VERSION_V1,
+    ) -> list[dict]:
         with patch.object(mp, "_request", side_effect=responses) as mock_request:
             batches = list(
-                mp._iter_export("us", "u", "s", "123", LOGGER, manager, start_date=start, end_date=end)  # type: ignore[arg-type]
+                mp._iter_export(  # type: ignore[arg-type]
```
danielcarletti: [blocker] CI "Check static typing" is red because of this reformatting. mypy attributes the FakeManager arg-type error to the line holding the `manager` argument (290/367/402/430), not to this `mp._iter_export(  # type: ignore[arg-type]` opening line — so each ignore is flagged `unused-ignore` AND the real `arg-type` error is unsuppressed. Same pattern at lines 361, 396, 424. Fix: move each `# type: ignore[arg-type]` to the line containing the `manager` argument (or keep the whole call on one line). Run mypy on the touched files (CI-only per repo convention) before pushing.

products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/source.py
```diff
@@ -40,6 +42,9 @@
 class MixpanelSource(ResumableSource[MixpanelSourceConfig, MixpanelResumeConfig]):
     api_docs_url = "https://developer.mixpanel.com"
 
+    supported_versions = SUPPORTED_API_VERSIONS
+    default_version = DEFAULT_API_VERSION
```
danielcarletti, line 46: [nit] `api_docs_url` is left unset, so `GET /api/public_source_configs/` surfaces `apiDocsUrl: null` for Mixpanel while the reference Stripe source populates it. Fix: set `api_docs_url` to Mixpanel's raw-export/API reference (or changelog) URL so the version-announcement metadata is populated.

danielcarletti (commented):
Skeptical fresh-eyes review. One blocker: the PR ships red CI — reformatted _iter_export test calls stranded their `# type: ignore[arg-type]` comments on the wrong line, so mypy "Check static typing" fails with unused-ignore + arg-type errors across four call sites. Design and blast radius are otherwise clean.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/test_mixpanel.py
Gilbert09: Fixed. Each `mp._iter_export(...)` call is now expanded so the `# type: ignore[arg-type]` sits on the `manager,` argument line (test_mixpanel.py:295, 375, 410, 438) — where mypy attributes the `FakeManager` arg-type error — rather than on the opening call line. This clears both the `unused-ignore` and the unsuppressed `arg-type` errors.

products/warehouse_sources/backend/temporal/data_imports/sources/mixpanel/source.py
Gilbert09: Done — `api_docs_url` now points at Mixpanel's raw-data-export API reference (`https://developer.mixpanel.com/reference/raw-data-export-api`), so the version-announcement metadata is populated instead of surfacing `apiDocsUrl: null`.

PR #71673, opened 2026-07-16
feat(bigquery): support BigQuery API version v2

Gilbert09:
## Problem

The BigQuery data-warehouse import source never declared a real vendor API version. It sat on the framework's `UNVERSIONED_API_VERSION` placeholder, even though the google-cloud-bigquery client has always talked to BigQuery's stable v2 REST API (every resource path is served under `/bigquery/v2/`). That left the version metadata surfaced in the API and HogQL out of step with reality, and gave us no seam to pin against if BigQuery ever ships a new REST path version.

**Why:** bring BigQuery in line with the other versioned sources so new sources record the real vendor API version and future version bumps have a place to dispatch.

## Changes

- Declare `v2` as a supported version and make it the **default** for new sources. Existing sources keep their current pin and are not migrated or repinned.
- Add a minimal version dispatch in the source's own directory: `_bigquery_rest_api_version` maps every supported label (and any fallback) onto BigQuery's single stable REST path segment, threaded into `bigquery_client` and applied to the client's connection.
- Because both the legacy label and `v2` resolve to the same `/bigquery/v2/` endpoint, the default bump is a no-op on the wire — existing pinned syncs are byte-for-byte unchanged.
- Append a BigQuery entry to the `warehouse-source-new-version` skill learnings.

> [!NOTE]
> No old version is deprecated and no migration script is included — this is a plain version add, not a sunset.

- `products/.../sources/bigquery/tests/test_source.py` (full file, 999 passing with the registry invariant suite) plus 11 new cases:
  - new sources default to `v2` and a missing pin resolves to `v2` — guards the default bump.
  - `v1`/`v2` pins are honored verbatim — guards pin semantics after the default flip.
  - `_bigquery_rest_api_version` maps every label to the `v2` segment — a wrong mapping would point the client at a nonexistent `/bigquery/<x>/` endpoint.
  - `build_pipeline` threads the resolved REST segment into `_build_source_response` for both pins — guards that a future version actually dispatches.
- `sources/tests/test_source_versions.py` registry invariants pass (default in supported, etc.).

No manual/live BigQuery testing was performed.

## 🤖 Agent context

Authored by Claude (Claude Code). Invoked the `/warehouse-source-new-version` skill for the version-declaration and dispatch conventions, and `/writing-tests` before adding tests. Chose to map both version labels to a single REST segment (rather than per-version branching) because BigQuery's REST API is stable at v2 and response shapes don't diverge, so no column-hint gating or separate request path was needed. Ruled out a duplicate by scanning open PRs (including the maintainer's own) for the version string and the source module path before opening this.


danielcarletti:
## Review: PR #71673 — feat(bigquery): support BigQuery API version v2

**ELI10:** BigQuery's importer never wrote down which version of Google's API it uses. This PR writes down "v2" (which it was always secretly using anyway) so the label the app shows is honest, and adds a tiny switch so a future version could be swapped in. Nothing about how it actually talks to Google changes.

**Goal achieved:** yes — declares `v2` as supported + default, threads a resolved segment through the read path; verified no-op on the wire.
**Sensible approach:** yes — matches the maintainer's own `warehouse-source-new-version` skill convention (declare label, map via dict, set `_connection.API_VERSION` fail-soft). Mechanism is real: the installed `google-cloud-bigquery` `Connection` uses `API_VERSION` in `API_URL_TEMPLATE` (`{api_base_url}/bigquery/{api_version}{path}`), so the seam would actually dispatch if a new label ever mapped elsewhere.
**Breaks dependents:** no — checked all 7 `bigquery_client` call sites and the single `_build_source_response` caller; every new param (`api_version`, `rest_api_version`) has a default, so un-threaded callers are unchanged. Unpinned/legacy-`v1` rows now resolve to `v2` in metadata but map to the same `/bigquery/v2/` endpoint (intended correction, no wire change). `connect()` correctly stays on the default (no row pin in scope, per the skill's documented discovery-path exception).

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/bigquery/tests/test_source.py:1616 — `test_bigquery_new_sources_default_to_v2` asserts a class constant (`default_version == "v2"`) plus base-class `resolve_api_version(None)` behavior; combined with `test_bigquery_supported_pin_is_honored` (which also just re-exercises the base-class verbatim-pin logic already covered by the cited registry-invariant suite), the four new version tests overlap heavily — all reduce to "everything resolves to v2." Fix: drop `test_bigquery_new_sources_default_to_v2` and keep the two BigQuery-specific guards (`_bigquery_rest_api_version` mapping + `build_pipeline` threading), or fold the default/supported assertions into a single parametrized case, per the repo's `/writing-tests` value gate.

### Verdict
Small, well-documented, correct, and safe — the metadata declaration cleanly achieves the goal and the client-threading, while inert on the wire today (by design), is the maintainer's own prescribed seam so it's not over-engineering to flag. No blockers or should-fix items. The only cleanup is trimming the redundant version tests, which lean on base-class/framework logic rather than BigQuery-specific behavior.

products/warehouse_sources/backend/temporal/data_imports/sources/bigquery/tests/test_source.py
```diff
@@ -1610,3 +1610,45 @@ def test_bigquery_resources_exceeded_is_non_retryable():
 
     assert matching, "resourcesExceeded query failure should be recognised as non-retryable"
     assert all(non_retryable_errors[key] is not None for key in matching)
+
+
+def test_bigquery_new_sources_default_to_v2():
+    # New sources (no pin) must be stamped with the current vendor API version, not the legacy label.
```
danielcarletti: [nit] test_bigquery_new_sources_default_to_v2 asserts a class constant (default_version == "v2") plus base-class resolve_api_version(None) behavior, and test_bigquery_supported_pin_is_honored just re-exercises the base-class verbatim-pin logic already covered by the registry-invariant suite this PR cites. All four new version tests reduce to "everything resolves to v2." Fix: drop test_bigquery_new_sources_default_to_v2 and keep the two BigQuery-specific guards (_bigquery_rest_api_version mapping + build_pipeline threading), or fold the default/supported assertions into one parametrized case, per the /writing-tests value gate.

danielcarletti (commented):
One nit only: the new version tests overlap and lean on base-class/framework logic. Mechanism, blast radius, and no-op-on-wire claim all verified sound.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/bigquery/tests/test_source.py
Gilbert09: Good call — addressed. I folded the default/supported assertions into a single BigQuery-specific declaration guard (`test_bigquery_source_declares_v2_as_default`, asserting `supported_versions == ("v1", "v2")` and `default_version == "v2"`) and dropped the two cases that only re-exercised base-class pin resolution already covered by the registry-invariant suite. The two BigQuery-specific guards remain: `test_bigquery_rest_api_version_maps_labels_to_v2` (label→REST-segment mapping) and `test_bigquery_build_pipeline_threads_resolved_rest_api_version` (build_pipeline threading).

PR #71674, opened 2026-07-16
feat(emailoctopus): support EmailOctopus API version v2

Gilbert09:
## Problem

EmailOctopus launched API v2 (recommending it over the now-legacy v1), but the source declared no explicit vendor API versions — it relied on the framework's implicit unversioned `"v1"` default. New sources should be created on v2, while existing sources stay exactly where they are.

## Changes

- Declare `supported_versions = ("v1", "v2")` and bump `default_version` to `"v2"` so **new** sources are created on v2.
- Thread the resolved `api_version` from `source_for_pipeline` into `emailoctopus_source` → `get_rows` via a base-URL-per-version seam (`_base_url_for_version`), mirroring the Salesforce/Stripe version-dispatch pattern.
- EmailOctopus serves every API version from the same host (`api.emailoctopus.com`) with the same bearer auth, resource paths, and cursor pagination — the version isn't carried in the request. Both labels therefore resolve to that one host, so existing `"v1"`-pinned sources are byte-for-byte unaffected by the default bump. The seam exists so a genuinely divergent future version can branch there.
- Existing sources keep their current pin; nothing is migrated or repinned, and no version is deprecated.

Automated only (no manual sync run):

- `test_emailoctopus_source.py::TestEmailOctopusSourceVersions` — locks in `default_version == "v2"` (new sources land on v2) and that both `"v1"` and `"v2"` pins are honored verbatim after the default bump. Catches a regression that drops v1 or reverts the default, which would mislabel or repin existing sources.
- `test_emailoctopus.py::TestApiVersionDispatch` — asserts `_base_url_for_version` resolves every supported label (and an undeclared pin) to the REST host, and that `get_rows(api_version=...)` threads the version end-to-end into the request URL for both v1 and v2. The v1 path was not exercised by any existing test.
- Existing `test_emailoctopus.py` call sites updated for the new required `api_version` argument.
- The registry invariant suite (`sources/tests/test_source_versions.py`) passes for the new declarations.

## 🤖 Agent context

Authored by PostHog Code (Claude Opus 4.8). Invoked the `/warehouse-source-new-version` skill for the version-declaration + dispatch conventions and `/writing-tests` for the test gate.

Key decision: the existing implementation already speaks EmailOctopus's real v2 protocol, and the vendor doesn't encode the API version in requests (confirmed against the v2 docs). So rather than inventing a divergent request path, both supported labels resolve to the same host and the version is threaded purely as a forward-compatible dispatch seam — keeping existing `"v1"`-pinned syncs byte-for-byte unchanged.

---


danielcarletti:
## Review: PR #71674 — feat(emailoctopus): support EmailOctopus API version v2

**ELI10:** EmailOctopus made a newer version of its data connection (v2). This PR tells PostHog to create brand-new EmailOctopus imports on v2, while leaving everyone's existing imports exactly as they were. Because v2 actually uses the same web address as before, nothing about how data is fetched really changes — it's mostly relabeling plus a bit of plumbing for a possible future version.

**Goal achieved:** yes — new sources default to `v2`, existing pins are honored verbatim, and no request behavior changes.
**Sensible approach:** yes — mirrors the established Stripe version-dispatch pattern (`resolve_api_version(inputs.api_version)`) and the `/warehouse-source-new-version` skill.
**Breaks dependents:** no — `get_rows` and `emailoctopus_source` gained a required `api_version` param; the only callers are `source.py` (updated) and the tests (updated). No external callers. `_base_url_for_version` maps both labels to the same host, so `"v1"`-pinned rows are byte-for-byte unchanged.

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/emailoctopus.py:36 — The `_base_url_for_version` seam plus `base_url` threaded through `_iter_list_ids` / `_get_top_level_rows` / `_get_contact_rows` is behaviorally a no-op today: both labels resolve to the same host, so it's pure speculative forward-compat indirection (against the "no unrequested flexibility" guideline). It's defensible since the skill prescribes exactly this and it reduces churn when a divergent version lands — just flagging that the whole seam could be deferred to that future PR, with only the `source.py` version declarations shipping now. Fix (optional): drop the emailoctopus.py threading and keep only `supported_versions`/`default_version` until a version genuinely diverges.
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/tests/test_emailoctopus.py:404 — `TestApiVersionDispatch` can't fail while every version resolves to one host: `test_base_url_resolves_to_rest_host` asserts a constant, and `test_get_rows_threads_version_into_request_host` would still pass even if `get_rows` ignored `api_version` (the URL is `EMAILOCTOPUS_BASE_URL == BASE` regardless). So it doesn't actually guard the "end-to-end thread" it claims to. Fix: either accept it as a low-value smoke test, or leave the dispatch tests for the PR that introduces a real version-varying request (where they can meaningfully fail).

### Verdict
Clean, well-scoped, and low-risk — the change does what the description says, follows the codebase's existing version-dispatch convention, and cannot affect existing `"v1"` syncs. No must-fix items. The only observations are that the base-URL seam and its two tests are inert given both versions share a host; both are reasonable to keep as forward-compat scaffolding, so treat them as judgment calls rather than defects.

products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/emailoctopus.py
```diff
@@ -22,6 +22,20 @@
...
+# or query param). Both supported labels therefore resolve to the one REST host the source has
+# always used; the seam exists so a genuinely divergent future version can branch here, and pinned
+# instances keep hitting the same host regardless of the default.
+_BASE_URL_BY_VERSION = {
+    "v1": EMAILOCTOPUS_BASE_URL,
+    "v2": EMAILOCTOPUS_BASE_URL,
+}
+
+
+def _base_url_for_version(api_version: str) -> str:
```
danielcarletti, line 36: [nit] The _base_url_for_version seam plus base_url threaded through _iter_list_ids / _get_top_level_rows / _get_contact_rows is behaviorally a no-op today: both labels resolve to the same host, so it is pure speculative forward-compat indirection. Defensible since the skill prescribes it and it cuts churn when a divergent version lands. Fix (optional): defer the emailoctopus.py threading to that future PR and ship only the source.py version declarations now.

products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/tests/test_emailoctopus.py
```diff
@@ -380,10 +391,38 @@ def test_primary_keys_and_partitioning(self, endpoint: str, expected_pks: list[s
...
         )
         assert response.name == endpoint
         assert response.primary_keys == expected_pks
         assert response.sort_mode == "asc"
         assert response.partition_mode == "datetime"
         assert response.partition_keys == [EMAILOCTOPUS_ENDPOINTS[endpoint].partition_key]
         assert EMAILOCTOPUS_ENDPOINTS[endpoint].partition_key == "created_at"
+
+
+class TestApiVersionDispatch:
```
danielcarletti, line 404: [nit] TestApiVersionDispatch cannot fail while every version resolves to one host: test_base_url_resolves_to_rest_host asserts a constant, and test_get_rows_threads_version_into_request_host would still pass even if get_rows ignored api_version (the URL is EMAILOCTOPUS_BASE_URL == BASE regardless). So it does not guard the end-to-end thread it claims to. Fix: accept it as a low-value smoke test, or defer the dispatch tests to the PR that introduces a real version-varying request where they can meaningfully fail.

danielcarletti (commented):
Clean, low-risk version-declaration PR that follows the established Stripe/skill dispatch pattern and cannot affect existing v1 syncs. Two nits only, both stemming from the fact that both labels resolve to the same host so the seam is inert today. No must-fix items.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/emailoctopus.py
Gilbert09: Thanks — agreed it's a no-op at runtime today, and that's deliberate. The `warehouse-source-new-version` skill prescribes threading the resolved `api_version` down to the request layer for every source (step 3), and its EmailOctopus learning is explicit that when a vendor doesn't encode the version in the request you "map every supported label to the same base URL and thread the version purely as a forward-compat dispatch seam." Keeping the seam here means the pin genuinely reaches the request layer, so `default_version="v2"` is provably honored rather than asserted, and a future divergent version branches in one obvious place instead of re-threading the call chain. I'd rather keep it consistent with the other sources than special-case EmailOctopus, so keeping as-is.

products/warehouse_sources/backend/temporal/data_imports/sources/emailoctopus/tests/test_emailoctopus.py
Gilbert09: Fair — these are intentionally smoke/guard tests, matching skill step 5 ("at minimum that the version label reaches the client/request layer for each supported version"). `test_get_rows_threads_version_into_request_host` asserts the resolved version flows through `get_rows` into the request host for both supported labels; you're right it can't fail today because both labels resolve to one host, but it locks in the contract so the day a version *does* diverge, the test starts exercising a real branch instead of having to be written from scratch. Keeping them as low-value smoke coverage rather than deferring.

PR #71675, opened 2026-07-16
feat(zonka_feedback): support Zonka Feedback API version v2.1

Gilbert09:
## Problem

Zonka Feedback's current stable API is v2.1 ([apidocs.zonkafeedback.com](https://apidocs.zonkafeedback.com/)); the source shipped under the framework's default unversioned label. New sources should be created on the vendor's current stable version.

**Why:** keep the source aligned with the vendor's current stable API so newly connected sources start on v2.1, while every source already in production stays on the version it was created against.

## Changes

- Declare `supported_versions = ("v1", "v2.1")` and set `default_version = "v2.1"` on `ZonkaFeedbackSource`, so new sources are stamped v2.1 at creation.
- Thread the resolved API version from `source_for_pipeline` (`resolve_api_version`) down through `zonka_feedback_source` → `get_rows` → the URL builder.
- Zonka versions its REST API through the account's region subdomain — not a URL path segment or version header — so v1 and v2.1 both target `{data_center}.apis.zonkafeedback.com`. The host template is keyed by version so a future divergence (host family or path prefix) is a one-line change. An undeclared pin falls back to the current host rather than failing the sync, matching `resolve_api_version`'s honor-the-pin behavior.
- No migration/repin: existing sources keep their pin. v1 and v2.1 build byte-for-byte identical requests, so sources resolving to either (pinned v1, or unpinned resolving to the new default) are unaffected. v1 remains fully supported.

Automated only (I did not run a live sync against Zonka). Ran the source's unit tests plus the registry version-invariant suite:
`pytest products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/tests/ products/warehouse_sources/backend/temporal/data_imports/sources/tests/test_source_versions.py` — all pass.

New/changed tests and the regression each guards:
- `test_version_declaration_defaults_to_v2_1` — locks in v2.1 as the resolved default for new/unpinned sources and that an explicit `v1` pin is honored verbatim (guards a silent repin of existing sources if the default flip regressed).
- `test_source_for_pipeline_plumbs_arguments` (now parameterized over v1/v2.1) — the resolved pin reaches the request layer; catches the version being dropped or hardcoded.
- `test_threads_api_version_to_url_builder` — `get_rows` passes the version to the URL builder; guards the request layer against silently ignoring the pin.
- `test_base_url_resolves_every_version_to_region_host` — each declared version, plus an undeclared pin, resolves to the region host; guards the dispatch table and the fallback.

## 🤖 Agent context

Authored with Claude Code (Opus 4.8). Invoked the `warehouse-source-new-version` skill for the version-declaration + dispatch conventions and `writing-tests` for the test gate. Followed the Stripe source as the threading reference (`resolve_api_version` at the source class, version passed to the request layer). Confirmed against vendor docs that Zonka versions via region subdomain rather than a path segment/header, so no per-version request divergence exists today — the change declares the versions, flips the default, and centralizes the (currently shared) host so a future divergence is localized. Checked open PRs (including the maintainer's own) for an existing Zonka v2.1 change and found none.

---


danielcarletti:
## Review: PR #71675 — feat(zonka_feedback): support Zonka Feedback API version v2.1

**ELI10:** Zonka Feedback (the survey tool PostHog imports data from) now has a newer official version, v2.1. This PR tells PostHog to start new connections on v2.1, while leaving anyone already connected exactly where they are. It turns out v1 and v2.1 send the identical request, so nothing about how data is actually fetched changes.

**Goal achieved:** yes — new sources are stamped `default_version = "v2.1"`; existing pins are honored verbatim.
**Sensible approach:** alternative worth considering — the request-layer version threading + host dict are behaviorally inert since Zonka versions by region subdomain, so the declaration alone meets the goal.
**Breaks dependents:** no — verified `SourceInputs.api_version` is pre-resolved in `import_data_sync.py:220` (double-resolve is idempotent, matches Stripe `source.py:488`); every `base_url` caller still compiles (`check_access`, `get_non_retryable_errors`, tests); v1/v2.1 build byte-for-byte identical requests, so NULL-pin sources flipping v1→v2.1 change nothing observable.

### Findings
- [should-fix] `products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/zonka_feedback.py:28` — `_HOST_TEMPLATE_BY_VERSION` maps both versions to the identical host, and the version threaded through `base_url` → `get_rows` → `zonka_feedback_source` changes no request. Zonka versions via the region subdomain, which is version-independent, so this is speculative machinery for a divergence that doesn't exist; the goal (new sources on v2.1) is met purely by `supported_versions`/`default_version` on the source class. The skill's own rule is "thread only if the version is a header/URL segment" — Zonka is neither — so dropping the dict + request-layer threading (and the two request-layer tests) is arguably *more* skill-aligned, not less. Fix: keep the class-attribute declaration + default flip; re-add threading when a version actually diverges the request. (Judgment call — the author documented the opposite intent; flagging for an explicit decision.)
- [nit] `zonka_feedback.py:27` — the comment (mirrored in SKILL.md and the PR body) claims a future host divergence is "a localized change here." It isn't fully: `get_non_retryable_errors` (`source.py:107`) independently builds hosts via `base_url(data_center)` with no version and would silently keep matching only the default-version host, so a diverged v1-pinned source's 401/403 wouldn't map to the friendly message. Fix: if the threading is kept, either thread the version into `get_non_retryable_errors` too, or soften the comment to note that call site also needs updating on a real divergence.

### Verdict
Correct and safe — no dependents break, existing pins are preserved, and the v1/v2.1 request-identical claim checks out. The only substantive point is that the request-layer threading and the two-identical-values host dict do nothing today; whether to carry that scaffolding for a hypothetical future divergence is a judgment call worth making explicitly rather than by default. Nothing here is a blocker.

danielcarletti (commented):
Two design-level notes on the version scaffolding: the request-layer threading and two-identical-values host dict are behaviorally inert today (Zonka versions by region subdomain), and the "localized one-line change" claim overlooks get_non_retryable_errors. Both are judgment calls, not blockers.

products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/zonka_feedback.py
```diff
@@ -13,9 +13,23 @@
...
+ZONKA_API_VERSION_V2_1 = "v2.1"
+
 # Zonka Feedback hosts data per region; the account's data center is the subdomain of the API host.
 # US=us1, EU=e, IN=in are the documented, verifiable identifiers.
 DATA_CENTER_IDS: tuple[str, ...] = ("us1", "e", "in")
+
+# Zonka versions its REST API through the account's region subdomain, not a URL path segment or a
+# version header, so every supported version targets the same host template. Keyed by version so a
+# future version that moves host families or adds a path prefix is a localized change here.
+_HOST_TEMPLATE_BY_VERSION: dict[str, str] = {
```
danielcarletti: [should-fix] This dict maps both versions to the identical host, and the version threaded through base_url -> get_rows -> zonka_feedback_source changes no request. Zonka versions via the region subdomain (version-independent), so this is scaffolding for a divergence that doesn't exist; the goal (new sources on v2.1) is met purely by supported_versions/default_version on the source class. The skill's rule is 'thread only if the version is a header/URL segment' -- Zonka is neither -- so dropping the dict + request-layer threading is arguably more skill-aligned. Fix: keep the declaration + default flip; re-add threading when a version actually diverges the request. Judgment call flagged for an explicit decision, not a blocker.

products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/zonka_feedback.py
```diff
@@ -13,9 +13,23 @@
...
+ZONKA_API_VERSION_V1 = "v1"
+ZONKA_API_VERSION_V2_1 = "v2.1"
+
 # Zonka Feedback hosts data per region; the account's data center is the subdomain of the API host.
 # US=us1, EU=e, IN=in are the documented, verifiable identifiers.
 DATA_CENTER_IDS: tuple[str, ...] = ("us1", "e", "in")
+
+# Zonka versions its REST API through the account's region subdomain, not a URL path segment or a
+# version header, so every supported version targets the same host template. Keyed by version so a
+# future version that moves host families or adds a path prefix is a localized change here.
```
danielcarletti: [nit] This 'localized change here' claim (also in SKILL.md and the PR body) isn't fully accurate: get_non_retryable_errors in source.py builds hosts via base_url(data_center) with no version, so on a real host divergence a v1-pinned source's 401/403 would keep matching only the default-version host and miss the friendly message. Fix: if threading is kept, thread the version into get_non_retryable_errors too, or soften this comment to note that call site also needs updating.

danielcarletti:
AI reply: Removed the inert per-version request scaffolding. Confirmed against Zonka's API docs that it versions purely via the region subdomain (`{data_center}.apis.zonkafeedback.com`) — same host, same paths (`/responses`, `/surveys`, `/contacts`), no version segment or version header — so v1 and v2.1 build byte-for-byte identical requests. With no request divergence, threading `api_version` into the request layer and keying `_HOST_TEMPLATE_BY_VERSION` on version changed nothing, so per the versioning skill's "don't add inert scaffolding" rule I dropped both and kept only the `supported_versions`/`default_version` declaration.

Changes: removed `_HOST_TEMPLATE_BY_VERSION` and reverted `base_url`, `get_rows`, and `zonka_feedback_source` to their single-host signatures; dropped the `api_version` threading in `source.py` (declaration unchanged); removed the two tests that only exercised the removed threading/map and de-parameterized `test_source_for_pipeline_plumbs_arguments`. The version-declaration test (default flip + honor-pin) stays. Also corrected the SKILL.md learning, which had claimed the version-keyed host made a future divergence a one-line change — it overlooked that `base_url` also feeds `check_access`/`get_non_retryable_errors`, so that was never true.

Tests green: `zonka_feedback/tests/` + `test_source_versions.py` — 915 passed.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/zonka_feedback.py
Gilbert09: Agreed — the version-threading scaffolding has been dropped. `base_url` no longer takes a version, the `_HOST_TEMPLATE_BY_VERSION` dict and the `api_version` params threaded through `get_rows`/`zonka_feedback_source` are gone. The change is now just the `supported_versions`/`default_version` declaration plus the default flip, as you suggested. The skill's Learnings entry was updated to record why threading is inert for a subdomain-versioned vendor.

products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/zonka_feedback.py
Gilbert09: Good catch — this is moot now that the request-layer threading is gone. `get_non_retryable_errors` (and `check_access`) already build hosts via `base_url(data_center)` with no version, and since every version issues byte-for-byte identical requests there's no divergence for them to miss. The "localized change here" comment that made the inaccurate claim was removed with the scaffolding, and the skill's Learnings entry now explicitly notes that `base_url` also feeds `check_access`/`get_non_retryable_errors`, so a real future divergence wouldn't be a one-line change.

PR #71676, opened 2026-07-16
feat(google_sheets): support Google Sheets API version v4

Gilbert09:
## Problem

The Google Sheets data-warehouse source declared no meaningful vendor API version — it sat on the framework's `UNVERSIONED_API_VERSION` (`"v1"`) default, which was never Google's retired API v3. Google's current (and only) stable REST API is v4, which is what the connector already talks to via `gspread`. New sources should record that they run on v4.

## Changes

- Add `"v4"` to the Google Sheets source's `supported_versions` and make it `default_version`, so new sources are created on v4.
- Keep `"v1"` (the framework's legacy `UNVERSIONED` default) in `supported_versions` so sources created before this change stay pinned and unaffected — no repin, no migration.
- Introduce a minimal version dispatch: `source_for_pipeline` resolves the instance pin (falling back to the default) via `resolve_api_version` and threads it down to where the `gspread` client is built.

`gspread` pins the Sheets v4 base URL for every request and exposes no per-client version knob, so every supported label resolves to identical calls. Flipping the default is therefore byte-for-byte safe for existing syncs; the threaded version is the dispatch seam for any future version whose request shape genuinely diverges.

Automated only (agent-run):

- Extended `test_google_sheets.py` with one parameterized test (`test_source_for_pipeline_threads_resolved_api_version_to_client`, 3 cases: unpinned→v4 default, legacy `"v1"` honored, explicit `"v4"`). It guards that `source_for_pipeline` resolves the pin and threads the version to the client build — a regression that drops the resolve/thread would send the wrong (or no) version to the request layer. No existing test exercised version resolution.
- Updated the two existing `google_sheets_source(...)` calls to pass the now-required `api_version`.
- Ran the source's test module plus the registry-wide `test_source_versions.py` invariants (default ∈ supported, deprecated ⊆ supported, default not deprecated, https `api_docs_url`) — all pass.

## Docs update

No user-facing behavior change — new sources simply record `v4` as their pinned version.

## 🤖 Agent context

Authored with Claude Code. Invoked the `/warehouse-source-new-version` skill (version declaration + dispatch conventions) and `/writing-tests` (test value gate).

Key decision: Google Sheets has no request-path divergence between labels because `gspread` bakes in the v4 base URL, so this is a declarative version bump plus a threaded dispatch seam rather than per-version branching. Kept the change inside the source's own directory; did not touch shared layers, other sources, credential validation, or the global retry/error handling.

Checked open PRs (including the maintainer's own) for an existing Google Sheets v4 fix before starting — none found.

---


danielcarletti:
## Review: PR #71676 — feat(google_sheets): support Google Sheets API version v4

**ELI10:** The Google Sheets connector never wrote down which version of Google's data API it uses. This PR labels it "v4" (the real one) so new connections record it, while old connections keep working exactly as before — the actual requests don't change at all.

**Goal achieved:** yes — `default_version`/`supported_versions` now record `v4`, and the resolved pin is threaded to the client build.
**Sensible approach:** yes — one alternative worth considering: the stated goal ("record v4") is met by the 2-line metadata bump alone; the ~30 lines threading `api_version` through five functions is a deliberate, skill-mandated dispatch seam that is a no-op today.
**Breaks dependents:** no — every caller of `google_sheets_source`, `_get_worksheet`, `get_schemas`, `get_schema_incremental_fields`, `google_sheets_client` lives inside the `google_sheets` module; `source_for_pipeline` and both test call sites are updated. New required `api_version` on `google_sheets_source` sits before the only defaulted arg and all callers pass it by keyword. `test_source_versions.py` invariants hold (default ∈ supported, none deprecated, https docs URL).
**Undisclosed side effects:** none — the `_get_worksheet` cache key widens to include `api_version`, but it is constant per source instance, so it only ever adds cache entries; behavior is unchanged (documented in the PR/skill).

### Findings
- [nit] products/warehouse_sources/backend/temporal/data_imports/sources/google_sheets/google_sheets.py:39 — `google_sheets_client(api_version=...)` accepts `api_version` but never references it in the body; the dispatch seam terminates in a dead parameter. This follows the documented `warehouse-source-new-version` convention, so it's defensible. Fix: keep only if the cross-source seam consistency is the intent; otherwise let the seam end one level up at `_get_worksheet`/`get_schemas` (where a future divergent client would actually be constructed) rather than passing an unused arg into `google_sheets_client`.

### Verdict
Clean, low-risk change: contained blast radius, consistent with the framework's versioning convention, and heavily documented. The only observation is that the threaded `api_version` is entirely speculative today and dead-ends as an unused parameter in `google_sheets_client` — acceptable as a deliberate seam, not a defect. No must-fix items. Note: the red "Unit Tests"/"MCP Tests Pass" checks are an unrelated MCP snapshot artifact drift plus a vitest permission test failure — neither is touched by this PR (it only changes google_sheets Python and one skill doc).

products/warehouse_sources/backend/temporal/data_imports/sources/google_sheets/google_sheets.py
```diff
@@ -27,8 +27,16 @@
...
+# every request (its base URL is pinned in gspread.urls and is not per-client configurable); v3 and
+# earlier are retired. The version is therefore threaded purely as metadata recording which vendor
+# GA version a sync declares: every supported label resolves to the same v4 calls. The framework's
+# legacy UNVERSIONED default ("v1") predates this metadata and was never Google's v3, so sources
+# created before this change keep issuing identical requests.
+GOOGLE_SHEETS_API_VERSION_V4 = "v4"
 
-def google_sheets_client() -> gspread.Client:
+
+def google_sheets_client(api_version: str = GOOGLE_SHEETS_API_VERSION_V4) -> gspread.Client:
```
danielcarletti: [nit] `google_sheets_client` accepts `api_version` but never references it in the body — the dispatch seam terminates in a dead parameter. This follows the documented `warehouse-source-new-version` convention, so it is defensible. Fix: keep only if cross-source seam consistency is intended; otherwise let the seam end one level up at `_get_worksheet`/`get_schemas` (where a future divergent client would actually be built) rather than passing an unused arg into `google_sheets_client`.

danielcarletti (commented):
One nit only — the change is clean, blast radius is contained to the google_sheets module, and it follows the versioning-framework convention. Flagging the speculative dispatch seam.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/google_sheets/google_sheets.py
Gilbert09: Addressed in the follow-up refactor (`refactor(google_sheets): end version seam at the worksheet cache key`). `google_sheets_client()` no longer accepts `api_version` — the dispatch seam now terminates at `_get_worksheet`, where `api_version` participates in the memoization key so a future version whose client differs can't reuse another version's cached handle. This matches your suggestion to end the seam one level up rather than passing a dead arg into the client builder.

PR #71677, opened 2026-07-16
feat(incident_io): support incident.io API version v2

Gilbert09:
## Problem

The incident.io data warehouse source declared no vendor API versions, so it rode the framework's implicit `UNVERSIONED_API_VERSION` default. incident.io versions its API per-resource and runs versions in parallel, and several core resources now recommend their v2 list endpoints. We want new sources to be created on v2 while keeping every previously supported version working, with existing syncs unaffected.

**Why:** requested as a plain version update for the incident.io source — add v2, make it the default for new sources, keep the current behaviour intact.

## Changes

- Declare `supported_versions = ("v1", "v2")` and `default_version = "v2"` on `IncidentIoSource`.
- Model each endpoint's path as a `{version: path}` map with a `path_for(pin)` resolver. incident.io exposes many resources on a single version, so the resolver honours the pin where the resource offers it and otherwise falls back to the version it does offer.
- Thread the resolved API version from `source_for_pipeline` through `incident_io_source` → `get_rows` into URL building. The creation-time credential probe validates against the default version (no pin exists yet).

> [!NOTE]
> v2 is the new default and resolves to **exactly** the paths the source already requested, so existing unpinned sources (which re-resolve NULL → the new default) are byte-for-byte unaffected. A `v1` pin only drops back to the `/v1` list where incident.io genuinely offers one — `incident_roles` and `custom_fields`. Resources that exist solely on `/v2` (incidents, alerts, follow_ups, escalations, users, schedules, incident_updates) stay on `/v2` under either pin. `incidents` in particular stays on v2 because the v1 incidents list is deprecated and lacks the `sort_by` + `[gte]` filters our incremental sync depends on.

No existing sources are migrated or repinned, and no version is deprecated — this is a plain update.

Automated only (I, Claude, ran these — no manual/staging testing):

- `test_incident_io.py`, `test_incident_io_source.py`, and the registry-wide `test_source_versions.py` — 1002 passed.
- New coverage catches: (1) the default version reproducing today's per-endpoint paths byte-for-byte (a regression that moved any URL under the default would fail); (2) a v1 pin resolving to `/v1` only for the dual-version config resources and staying on `/v2` elsewhere; (3) an undeclared pin falling back to the default path instead of crashing; (4) the resolved version actually reaching the request URL through `get_rows`; (5) `source_for_pipeline` resolving the pin (NULL/"" → v2, explicit pin honoured) and passing it down.

## 🤖 Agent context

Authored by Claude (via PostHog Code). Invoked the `/warehouse-source-new-version` skill for the version-declaration + dispatch conventions and `/writing-tests` before adding tests. Grounded the v1-vs-v2 endpoint availability in the vendor docs (`docs.incident.io/llms.txt`) rather than the untrusted detection notes: only `incidents`, `incident_roles`, and `custom_fields` expose both v1 and v2 list endpoints. Rejected a "prefer v1 everywhere" v1 variant once the docs showed the v1 incidents list is deprecated and can't do incremental — downgrading it would have broken sync. Appended three incident.io-specific learnings to the skill for the next version bump.

---


danielcarletti:
## Review: PR #71677 — feat(incident_io): support incident.io API version v2

**ELI10:** incident.io lets you ask its computer using an old menu (v1) or a new menu (v2). This PR makes new setups use the new menu and lets each dish come from whichever menu has it. The catch: everyone already set up got secretly labeled "old menu," so turning this on quietly moves two of their dishes back to the old menu instead of leaving them alone.

**Goal achieved:** partial — v2 is wired as the default and threaded to the request layer, but the "existing sources unaffected" guarantee does not hold.
**Sensible approach:** yes with a fix — the `{version: path}` map + `path_for` resolver is a reasonable shape; it just needs to reconcile with how existing rows were already pinned.
**Breaks dependents:** yes — checked `import_data_sync.py` (resolves override→pin→default) and migration `0075` (on master). Existing `IncidentIo` sources are pinned `"v1"`, not NULL, so `incident_roles` and `custom_fields` silently move from `/v2` to `/v1`.
**Undisclosed side effects:** the PR's NOTE claims existing sources are unpinned (NULL) and byte-for-byte unaffected; in reality every existing source is pinned `"v1"` and two endpoints change.

### Findings

- [blocker] products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/settings.py:142 (and :160) — Exposing `/v1` paths for `incident_roles` and `custom_fields` regresses existing sources. The PR's safety argument assumes existing incident.io sources have `api_version = NULL` and re-resolve to the new default `v2`. That premise is false: migration `0075_backfill_externaldatasource_api_version` (already on master, from versioning-framework PR #70018) backfills every source type absent from its snapshot map to `DEFAULT_API_VERSION = "v1"`, and `"IncidentIo"` is not in that map — so all pre-existing rows became `api_version = "v1"`. New rows are stamped `source.default_version` at creation (`external_data_source.py:2168`), which was `UNVERSIONED_API_VERSION = "v1"` before this PR. So no incident.io source is NULL; they are all pinned `"v1"`. Once this PR activates dispatch, `path_for("v1")` returns `/v1/incident_roles` and `/v1/custom_fields`, silently switching existing syncs off the `/v2` endpoints they use today (schema/field drift on live customer tables — exactly the silent-migration failure this framework exists to prevent). Fix: add a written-not-run data migration repinning `ExternalDataSource` rows where `source_type = "IncidentIo"` and `api_version = "v1"` to `"v2"` (idempotent, reviewable per the skill's migration-script rule) so the backfilled sentinel stops colliding with the now-meaningful `"v1"` label; alternatively keep `incident_roles`/`custom_fields` as v2-only in the map so a `"v1"` pin falls back to `/v2` and matches today's behavior.

- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/tests/test_incident_io.py:485 — `test_v1_pin_prefers_v1_list_where_available` asserts `/v1/incident_roles` and `/v1/custom_fields` under a `v1` pin, i.e. it encodes the regression above as the expected result, so the suite stays green while existing syncs break. No test covers the actual production scenario (rows already pinned `"v1"` from the 0075 backfill). Fix: after adding the repin migration, assert that an existing `"v1"`-pinned source resolves these two endpoints to `/v2` (byte-for-byte), and drop/adjust the expectation that `"v1"` genuinely serves `/v1` for them.

- [should-fix] .agents/skills/warehouse-source-new-version/SKILL.md:64 — The recorded learning states existing incident.io rows are "unpinned (NULL)"; they are not — migration 0075 backfilled them to `"v1"`. This wrong premise will mislead the next unversioned→versioned bump into the same trap. Fix: rewrite the learning to note that formerly-unversioned sources were backfilled to `"v1"` (the framework's `DEFAULT_API_VERSION`), so any version-add that gives `"v1"` real endpoint meaning must repin existing rows in the same PR.

### Verdict
The dispatch mechanics and byte-for-byte-under-default reasoning are sound, but they rest on a false assumption about existing data. Because the 0075 backfill and creation-time stamping already pinned every incident.io source to `"v1"`, turning on real v1/v2 dispatch silently downgrades `incident_roles` and `custom_fields` for current customers. Must-fix: add a written-not-run migration repinning existing `IncidentIo` `"v1"` → `"v2"` (or make those two endpoints v2-only), fix the test that currently blesses the regression, and correct the skill learning. Blast radius is bounded by the number of live incident.io sources (ALPHA), but the change is silent, so it should not ship as-is.

products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/settings.py
```diff
@@ -86,50 +110,56 @@ class IncidentIoEndpointConfig:
...
         name="severities",
-        path="/v1/severities",
+        paths={INCIDENT_IO_API_VERSION_V1: "/v1/severities"},
         data_key="severities",
     ),
     "incident_roles": IncidentIoEndpointConfig(
         name="incident_roles",
-        path="/v2/incident_roles",
+        paths={
+            INCIDENT_IO_API_VERSION_V1: "/v1/incident_roles",
```
danielcarletti: [blocker] Exposing /v1 paths for incident_roles and custom_fields regresses existing sources. The PR assumes existing incident.io rows are api_version=NULL, but migration 0075 (on master) backfilled every source type absent from its snapshot map to DEFAULT_API_VERSION="v1", and "IncidentIo" is not in that map. Creation also stamps default_version (was UNVERSIONED="v1" pre-PR) at external_data_source.py:2168. So every existing source is pinned "v1", not NULL — and once dispatch is live, path_for("v1") returns /v1/incident_roles (line 142) and /v1/custom_fields (line 160), silently moving existing syncs off the /v2 endpoints they use today. Fix: add a written-not-run data migration repinning ExternalDataSource rows where source_type='IncidentIo' and api_version='v1' -> 'v2' (idempotent), or keep these two endpoints v2-only so a v1 pin falls back to /v2.

products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/tests/test_incident_io.py
```diff
@@ -435,4 +438,70 @@ def test_partition_keys_are_stable_creation_fields(self, config):
...
+            ("incident_roles", "/v1/incident_roles"),
+            ("custom_fields", "/v1/custom_fields"),
+            ("severities", "/v1/severities"),
+            ("incident_statuses", "/v1/incident_statuses"),
+            ("incidents", "/v2/incidents"),
+            ("follow_ups", "/v2/follow_ups"),
+            ("users", "/v2/users"),
+        ],
+    )
+    def test_v1_pin_prefers_v1_list_where_available(self, endpoint, expected_path):
```
danielcarletti: [should-fix] This test asserts /v1/incident_roles and /v1/custom_fields under a v1 pin — it blesses the regression as correct, so the suite stays green while existing (backfilled-to-v1) syncs break. No test covers the real production scenario. Fix: after adding the repin migration, assert an existing "v1"-pinned source resolves these two endpoints to /v2 byte-for-byte, rather than to /v1.

.agents/skills/warehouse-source-new-version/SKILL.md
```diff
@@ -60,3 +60,6 @@ After you finish a version-update or deprecation PR using this skill, **append w
 ### Learnings
 
 - (seed) Stripe: response shapes differ enough across date versions that canonical column hints must be gated per version; newer versions auto-infer schema instead.
+- incident.io: versions per-resource, not globally — some resources are v2-only, some v1-only, a few expose both. Model paths as a `{version: path}` map per endpoint with a `path_for(pin)` that falls back to the version the resource actually offers, so a coarse source pin never invents a nonexistent URL.
+- incident.io: when the source's implicit version was the `UNVERSIONED_API_VERSION` default and existing rows are unpinned (NULL), flipping the default re-resolves those rows — so the NEW default must reproduce today's exact per-endpoint paths byte-for-byte, and the older label becomes the only one that may legitimately differ.
```
danielcarletti: [should-fix] This learning states existing incident.io rows are "unpinned (NULL)" — they are not. Migration 0075 backfilled formerly-unversioned sources to "v1" (the framework's DEFAULT_API_VERSION), and creation stamps default_version. This false premise will mislead the next unversioned->versioned bump into the same trap. Fix: rewrite to note that a version-add giving "v1" real endpoint meaning must repin existing "v1" rows in the same PR.

danielcarletti (commented):
Three findings, one root cause: existing incident.io sources are already pinned "v1" (via migration 0075 backfill + creation-time stamping), not NULL as the PR assumes — so activating v1/v2 dispatch silently downgrades incident_roles and custom_fields from /v2 to /v1. Inline notes cover the code, the test that blesses it, and the skill learning that misstates the premise.

Gilbert09:
Thanks @danielcarletti — verified and fixed in d399681.

Confirmed independently: migration `0075_backfill_externaldatasource_api_version` sets `DEFAULT_API_VERSION="v1"` and `IncidentIo` is absent from its snapshot map, so existing rows were backfilled to `api_version="v1"` (not NULL). My original v1 map would have downgraded `incident_roles` and `custom_fields` to /v1 for those already-pinned sources.

Fix taken (your second option, and the one consistent with "don't repin existing sources"): both endpoints are now v2-only, so a `"v1"` pin falls back to /v2 — byte-for-byte with today, no data migration. The per-version map stays as the seam for a future bump that genuinely moves an endpoint (which would ship its own repin migration).

- `test_v1_pin_matches_default_byte_for_byte` now asserts a `"v1"` pin resolves every endpoint identically to the v2 default (guards the exact regression you flagged).
- Skill learning corrected to state the 0075 backfill and the repin requirement.

Leaving the threads for you to resolve.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/settings.py
Gilbert09: Fixed in `5a6ba4a` (rebased). Rather than add a repin migration, I kept `incident_roles` and `custom_fields` v2-only, so a `"v1"` pin falls back to `/v2` for them — existing sources (backfilled to `"v1"` by migration 0075) keep hitting the exact `/v2` endpoints they sync from today, byte-for-byte, with no source repinned. The per-version map remains as the seam for a future bump that genuinely moves a resource's endpoint; that change would carry its own written-not-run repin migration.

products/warehouse_sources/backend/temporal/data_imports/sources/incident_io/tests/test_incident_io.py
Gilbert09: Fixed in `5a6ba4a`. The v1-pin test no longer blesses the `/v1` downgrade — `test_v1_pin_matches_default_byte_for_byte` now asserts that a `"v1"` pin resolves *every* endpoint to the identical path the v2 default does (parametrized over all endpoints), so `incident_roles` and `custom_fields` staying on `/v2` is what the suite enforces. The request-URL test likewise asserts a `"v1"`-pinned source still hits `https://api.incident.io/v2/custom_fields`.

.agents/skills/warehouse-source-new-version/SKILL.md
Gilbert09: Fixed in `5a6ba4a`. The learning is rewritten to state that formerly-unversioned sources are NOT NULL-pinned — migration `0075_backfill_externaldatasource_api_version` backfilled them to `DEFAULT_API_VERSION="v1"` and creation stamps `default_version` — so giving `"v1"` differing endpoint meaning silently moves those rows, and the fix is to either repin in a written-not-run migration or keep the diverging resources on the version existing rows actually sync from.

Gilbert09:
We dont wanna version this - the API is versioned on a per schema basis and they're all correct

PR #71679, opened 2026-07-16
feat(warehouse_sources): support ShipStation API version v2

Gilbert09:
## Problem

ShipStation now runs two active vendor APIs: the original **v1** (`ssapi.shipstation.com`, HTTP basic auth, US Pacific DateTimes) and the newer ShipEngine-based **v2** (`api.shipstation.com/v2`, API-Key header, ISO 8601 UTC). The source only implemented v1. v1 is not deprecated, so both need to keep working.

## Changes

- Declare `supported_versions = ("v1", "v2")` and flip `default_version` to **v2**, so new sources are created on v2. Existing sources keep their pin and are unaffected.
- Thread the resolved `api_version` through the transport (`source_for_pipeline` → `shipstation_source` → `get_rows`), dispatching per-version differences from a single frozen `ShipStationDialect` table: base URL, auth scheme (basic auth vs `API-Key` header), page-size param spelling (`pageSize` vs `page_size`), date-filter format (Pacific vs ISO 8601 UTC), and the credential probe path.
- `validate_credentials` now probes the host of the default version (what a new source is created on).
- Added the v2 host variants to the non-retryable auth-error matchers.
- v1 stays byte-for-byte identical: the internal transport helpers default to v1 and only the source layer passes the resolved version; an undeclared pin falls back to the v1 transport.

Version-blind parts of the framework (`get_schemas`, `SourceConfig.fields`) mean per-version schema/credential-field divergence isn't expressed here — this bump scopes to the request path, per the `warehouse-source-new-version` skill.

Automated only (agent could not exercise the live vendor API):

- `products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/tests/` — 73 passed.
- `sources/tests/test_source_versions.py` (shipstation + resolution cases) — passed.

- `test_declares_both_versions_defaulting_to_v2` + `test_source_for_pipeline_resolves_api_version` — catch a dropped version, a default that isn't v2, or the resolved pin not reaching the request layer.
- `TestVersionDispatch` (host/pagination per version, auth scheme per version, v2 probe host, unknown-pin → v1 fallback) — catch the transport not switching host/auth/pagination for v2, or silently promoting an undeclared pin to a newer scheme.
- `test_format_values_v2_uses_iso_utc` — catches v2 sending Pacific-space timestamps the v2 API rejects.

`test_public_source_configs.py` needs Postgres, unavailable in this sandbox; it's unaffected by this change (asserts general version invariants + Stripe specifics).

## Docs update

No user-facing doc change: the version pin is surfaced generically by existing API/UI fields.

## 🤖 Agent context

Authored by Claude Code. Invoked skills: `/warehouse-source-new-version` (version-declaration + dispatch conventions) and `/writing-tests` (test gate).

Considered a full v2 resource catalog with its own schema discovery, but rejected it: `get_schemas` and `SourceConfig.fields` are version-blind, so a divergent v2 schema/credential model can't be expressed without shared-layer changes the skill scopes out. Chose the minimal request-path dispatch (dialect table) instead, keeping v1 untouched. Ruled out a duplicate PR — no open PR (including from the source maintainer) touches ShipStation v2. Appended learnings to the skill.

---


danielcarletti:
## Review: PR #71679 — feat(warehouse_sources): support ShipStation API version v2

**ELI10:** ShipStation has two different websites for its robots to talk to: an old one and a newer one. This PR teaches PostHog to dial the new one and switches new customers to it by default — but it still asks the new one for pages (orders, customers, products) that only exist on the old one, so those requests will come back empty-handed.

**Goal achieved:** no — v2 transport is wired, but the shared endpoint catalog is v1-only, so new v2 sources can't actually sync any table.
**Sensible approach:** alternative worth considering — given the framework's version-blind schemas/fields (which the PR documents), v2 can't be expressed functionally yet; don't ship it as the default (or at all) until the resource catalog can diverge per version.
**Breaks dependents:** at risk — existing sources are safe (migration 0075 backfilled all ShipStation rows to `"v1"`, since ShipStation isn't in `API_VERSION_BY_SOURCE_TYPE`), but *new* sources are stamped `default_version` (`external_data_source.py:2168`), now `"v2"`, and will fail every sync.
**Undisclosed side effects:** none beyond the above.

### Findings
- [blocker] products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/settings.py:132 — Flipping `SHIPSTATION_DEFAULT_VERSION` to v2 makes every newly created ShipStation source resolve to v2 (`resolve_api_version(None)` → default; stamped at creation in `external_data_source.py:2168`). But `get_rows` always uses the v1 endpoint catalog (`SHIPSTATION_ENDPOINTS`: paths `/orders`,`/customers`,`/products`,`/fulfillments`,`/stores`, camelCase primary keys `orderId`/`shipmentId`, data keys, `data.get("pages")` pagination). ShipStation API v2 (ShipEngine) is a different resource API (shipments/labels/rates/carriers, snake_case) that does not serve those v1 paths — so new sources will 404 on every table while `validate_credentials` still passes (it probes `/carriers`). Fix: keep `SHIPSTATION_DEFAULT_VERSION = "v1"` and don't expose v2 as a selectable version until the endpoint catalog / schema / credential fields can diverge per version (exactly the limitation the PR's "considered and rejected a v2 catalog" note describes).
- [blocker] products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/shipstation.py:63 — The v2 dialect sets `credentials_probe_path="/carriers"` precisely because v1's `/stores` isn't valid on v2 — an in-PR admission that the two versions have different resource sets. Yet `get_rows` still requests the v1 paths (including `/stores`, `/orders`) for a v2 pin. The probe passing while the sync fails is a false-green: creation validates, first sync errors on every schema. Fix: don't route the shared v1 endpoint catalog through the v2 host; v2 needs its own resource/schema definitions before it can be offered.
- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/source.py:42 — `supported_versions` now advertises v2 as a user-selectable version, and the v1-only `incremental_params` (`createDateStart`), `sort_by` enums (`CreateDate`), data keys, and camelCase primary keys are still used for a v2 pin, so even the two resources that do overlap (shipments/warehouses) would be filtered/parsed with v1 spellings. Also, `api_secret` is `required=True` in `get_source_config`, but v2 auth uses only the API key in the `API-Key` header — a v2 user has no secret to enter. Fix: fold into the blocker above — hold v2 out of `supported_versions` until per-version schema/credential divergence is expressible.

### Verdict
The transport-only "dialect" abstraction is clean and v1 stays byte-for-byte identical, but it's the wrong shape for this vendor: ShipStation v1→v2 is a full resource-API replacement, not a host/auth/pagination/date reskin, and the PR's own probe-path swap (`/stores`→`/carriers`) proves the resource sets differ. Shipping v2 as the new default therefore breaks new-source creation end-to-end while the mocked tests stay green (the v2 test feeds a v1-shaped `{"orders":[{"orderId":1}]}` response and only asserts host/param, never that the endpoint exists). Must-fix before merge: revert `default_version` to `"v1"` and drop v2 from `supported_versions` until the shared endpoint catalog, `get_schemas`, and credential fields can vary per version.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/settings.py
```diff
@@ -122,3 +122,11 @@ class ShipStationEndpointConfig:
...
     name: config.incremental_fields for name, config in SHIPSTATION_ENDPOINTS.items() if config.incremental_fields
 }
+
+# Vendor API versions. v1 is the original ssapi.shipstation.com API (HTTP basic auth,
+# US Pacific DateTimes); v2 is the newer, ShipEngine-based api.shipstation.com/v2 API
+# (API-Key header, ISO 8601 UTC). Both are actively supported by the vendor.
+SHIPSTATION_API_VERSION_V1 = "v1"
+SHIPSTATION_API_VERSION_V2 = "v2"
+SHIPSTATION_SUPPORTED_VERSIONS = (SHIPSTATION_API_VERSION_V1, SHIPSTATION_API_VERSION_V2)
+SHIPSTATION_DEFAULT_VERSION = SHIPSTATION_API_VERSION_V2
```
danielcarletti: [blocker] Flipping the default to v2 breaks new sources. New sources are stamped `default_version` at creation (external_data_source.py:2168), so they resolve to v2 — but `get_rows` always uses the v1 endpoint catalog (`/orders`,`/customers`,`/products`,`/fulfillments`,`/stores`, camelCase primary keys, `data.get("pages")`). ShipStation API v2 (ShipEngine) doesn't serve those v1 paths, so every table sync 404s while `validate_credentials` (probing `/carriers`) still passes — a false green. Existing sources are safe (migration 0075 backfilled ShipStation to `v1`). Fix: keep `SHIPSTATION_DEFAULT_VERSION = "v1"` until the endpoint catalog/schemas/fields can diverge per version.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/shipstation.py
```diff
@@ -29,6 +30,47 @@
...
+        uses_pacific_time=True,
+        uses_api_key_header=False,
+        credentials_probe_path="/stores",
+    ),
+    SHIPSTATION_API_VERSION_V2: ShipStationDialect(
+        base_url="https://api.shipstation.com/v2",
+        page_size_param="page_size",
+        uses_pacific_time=False,
+        uses_api_key_header=True,
+        credentials_probe_path="/carriers",
```
danielcarletti, line 63: [blocker] Setting the v2 probe path to `/carriers` (instead of v1's `/stores`) is an in-PR admission that v1 and v2 expose different resources — yet `get_rows` still requests the v1 paths (including `/stores`, `/orders`) for a v2 pin. Creation validates against `/carriers` while the first sync errors on every schema. Fix: don't route the shared v1 endpoint catalog through the v2 host; v2 needs its own resource/schema definitions before it can be offered.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/source.py
```diff
@@ -37,6 +39,9 @@
 class ShipStationSource(ResumableSource[ShipStationSourceConfig, ShipStationResumeConfig]):
     api_docs_url = "https://www.shipstation.com/docs/api/"
 
+    supported_versions = SHIPSTATION_SUPPORTED_VERSIONS
```
danielcarletti, line 42: [should-fix] Advertising v2 in `supported_versions` exposes a non-functional version. The v1-only `incremental_params`/`sort_by`/data keys/camelCase primary keys are used for a v2 pin, so even overlapping resources (shipments/warehouses) filter and parse with v1 spellings. Also `api_secret` is `required=True` in `get_source_config`, but v2 auth uses only the API key in the `API-Key` header — v2 users have no secret to enter. Fix: hold v2 out of `supported_versions` until per-version schema/credential divergence is expressible.

danielcarletti (commented):
Skeptical fresh-eyes review. One design-level concern dominates: the PR wires v2 transport (host/auth/pagination/date) and flips the default to v2, but the shared endpoint catalog is v1-only, so new sources will fail every sync. See inline comments; the full report is in a separate PR comment.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/settings.py
Gilbert09: Addressed — the default is back on v1. `SHIPSTATION_DEFAULT_VERSION = SHIPSTATION_API_VERSION_V1` and `SHIPSTATION_SUPPORTED_VERSIONS = (SHIPSTATION_API_VERSION_V1,)`, so new sources are stamped v1 and keep hitting the v1 endpoint catalog. The v2 transport dialect stays as landed groundwork, reachable only by an explicit `ExternalDataSource.api_version` pin (honored verbatim by `resolve_api_version`), until the framework can express per-version schemas (`get_schemas`) and credential fields (`SourceConfig.fields`). A comment at the declaration spells out the constraint.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/shipstation.py
Gilbert09: Right — the divergent probe path was the tell that v1 and v2 serve different resources. Since v2 is no longer advertised or defaulted (`supported_versions`/`default_version` are v1-only now), `get_rows` only ever runs the v1 catalog against the v1 host for any source created through the product. The `/carriers` probe and the rest of the v2 dialect are unreachable except by an explicit `api_version="v2"` pin, and v2 won't be advertised until its own resource/schema definitions exist.

products/warehouse_sources/backend/temporal/data_imports/sources/shipstation/source.py
Gilbert09: Done — v2 is held out of `supported_versions` (`SHIPSTATION_SUPPORTED_VERSIONS = (SHIPSTATION_API_VERSION_V1,)`), so no non-functional version is exposed and the `api_secret required=True` field mismatch you flagged never reaches a user. v2 stays as unadvertised transport groundwork behind an explicit pin until per-version schema and credential-field divergence is expressible.

Also fixed the related security finding on the same path: the v2 `API-Key` custom header isn't stripped by `requests` across a cross-host redirect (only `Authorization` is), so the v2 session is now built with `allow_redirects=False`.

PR #71680, opened 2026-07-16
feat(replay-vision): persist a recovery run when an alert clears

ksvat:
## Problem

When a threshold alert fires and later clears, run history shows the firing but nothing marks the end of the breach.
You can't tell from the history whether a breach is still ongoing, and the alert silently re-arms with no visible record of when things recovered.

## Changes

- When an `on_breach` alert's condition clears after a firing, the check now persists a "Recovered" run: the metric's current value vs the threshold, plus how long the alert had been firing (when the original firing is still within the state scan window).
- New `AlertStatus.RECOVERED`. The workflow completes the run without delivering anything - the bookend is visible run history, not a notification.
- Recovery runs carry an `output.recovered` marker. State resolution (`_last_evaluated_run`) reads them as "cleared", so the next breach fires again instead of being suppressed as still-breached (both firings and recoveries persist a message, so the marker is what keeps them apart).
- An unmeasurable metric (average score over an empty window) re-arms quietly without a recovery row - with no measurement there's no "back at X" to claim.
- API: new `is_recovery` field on the run serializers so the UI can distinguish the bookend from a firing. Generated types regenerated.

Automated only, no manual testing. Two new tests in `test_vision_action_alerts.py`:

- fire → recover → fire again: the recovery run persists a visible message, reads as "cleared" in state resolution, and the next breach fires (misread as a breach it would suppress firing forever).
- unmeasurable window after a breach re-arms without a recovery row.

Full alert test file (20 tests) green locally.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Implemented with Claude (PostHog Code), directed by Kim. Skills invoked: /writing-tests, /improving-drf-endpoints. The quiet re-arm behavior for unmeasurable averages was a deliberate choice over emitting a recovery with no value in it.

---

### DIFF
--- a/products/replay_vision/backend/api/vision_actions.py
+++ b/products/replay_vision/backend/api/vision_actions.py
@@ -621,6 +621,12 @@ class VisionActionRunListSerializer(serializers.ModelSerializer):
     error_reason = serializers.SerializerMethodField(
         help_text="Short human-readable reason a run skipped or failed; null on success.",
     )
+    is_recovery = serializers.SerializerMethodField(
+        help_text=(
+            "True for the run recording an alert's condition clearing after a breach (the recovery "
+            "bookend in run history). False for alert firings and summaries."
+        ),
+    )
 
     class Meta:
         model = VisionActionRun
@@ -630,6 +636,7 @@ class VisionActionRunListSerializer(serializers.ModelSerializer):
             "scheduled_at",
             "observation_count",
             "error_reason",
+            "is_recovery",
             "created_at",
             "updated_at",
         ]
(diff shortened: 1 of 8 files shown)


ksvat:
* **#71681** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#71680** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71680?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ksvat:
### Merge activity

* **Jul 16, 10:08 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71680).
* **Jul 16, 10:09 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71680).

PR #71684, opened 2026-07-16
feat(warehouse-sources): support Google Ads API v24

Gilbert09:
## Problem

Google Ads API v24 is the current GA major version and v23 is now one major behind. The `GoogleAds` warehouse source only implemented v23, so new sources had no path onto v24 and the source would fall behind as Google's ~12-month support window advances.

## Changes

- Declare `v24` in `GoogleAdsSource.supported_versions` and make it the `default_version`, so **new** sources are created on v24. Existing sources keep their pin: their **sync** resolves to v23 and is unchanged. One nuance (raised in review): the periodic schema-**discovery** path (`get_schemas`) can't see a source's pin today (the shared framework signature doesn't carry it), so it runs at `default_version` (v24) for every source. Impact is low — discovery is table-name granularity over the static `RESOURCE_SCHEMAS`, identical across v23/v24 here — and threading the pin into discovery is a framework-wide change left out of this PR's scope.
- Bump `google-ads` 29.2.0 → 30.1.0. The Google Ads version is baked into the SDK's generated `vNN.*` modules, and v24 first ships in 30.1.0 with v23 retained — so a package bump is required for v24 to actually work at the wire level. Blast radius is contained: only this source imports the versioned SDK modules (the CDP Google Ads Conversions destination uses the Data Manager API).
- Thread the resolved `api_version` from `source_for_pipeline` down through `google_ads_source` and `get_schemas`, and pass it to `get_service(..., version=...)` for **both** `GoogleAdsService` and `GoogleAdsFieldService`. This matters because the SDK's client default is the newest bundled version, which the bump flips to v24 — pinning explicitly keeps v23 sources on v23.
- Make the protobuf type-URL resolver dispatch on the URL's version segment instead of a hardcoded `v23`, so v23 and v24 responses each decode against their own protos.

> [!NOTE]
> This is a plain version add, not a sunset. v23 is not deprecated and existing v23 **syncs** are unaffected — the sync path resolves each source's stored pin and threads it end to end. Schema discovery runs at the default version (see the discovery nuance above).

- The Google Ads source unit suite (`test_google_ads_source.py`, 143 tests) plus the registry version-invariant suite (`test_source_versions.py`).
- New tests, parameterized over `v23` and `v24`:
  - version resolution honors a pin and falls back to the new v24 default; v24 is the default and both versions are supported.
  - the resolved version reaches `get_service` for both `GoogleAdsService` and `GoogleAdsFieldService` (mocked boundary).
  - the type-URL resolver decodes both a v23 and a v24 enum URL against the matching protos, and raises on an unknown version.
- Verified `google-ads==30.1.0` bundles both v23 and v24 modules and that they import; confirmed the source-registration path still does not import the SDK.
- `uv lock --check`, `ruff check`, and `ruff format` clean on touched files.

I did not run the credential-gated e2e tests (no live Google Ads credentials in this environment); I updated their call sites to the new signature and confirmed they still collect.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing doc change: version pins are internal metadata surfaced automatically via the public source-config API.

## 🤖 Agent context

Authored by Claude Code (Opus 4.8) from a task to add v24 support while keeping every prior version working.

Skills invoked: `/warehouse-source-new-version` (the repo's version-add playbook), which I also appended two Google-Ads-specific learnings to.

Key decision: the skill's minimal pattern is "thread the version string to the request layer" (as Stripe does with an opaque header). Google Ads is different — the version is compiled into the pinned SDK's generated modules, so a genuine v24 needs an SDK bump. I picked the smallest bump that bundles both v23 and v24 (30.1.0, which removes nothing) to keep the v23 path behaviorally identical, and threaded the resolved version to both service builders plus the type-URL resolver rather than hardcoding a fallback. I checked open PRs (including the maintainer's own) first and found no existing v24 work.

---


danielcarletti:
## Review: PR #71684 — feat(warehouse-sources): support Google Ads API v24

**ELI10:** PostHog can pull data from Google Ads. Google shipped a newer version (v24) of their data language, so this teaches PostHog to speak v24 for new connections while still speaking v23 for the ones already set up.

**Goal achieved:** partial — v24 is added and threaded through sync, but the schema-discovery path is hardcoded to the default version, so an existing v23 source does not stay fully on v23.
**Sensible approach:** yes — SDK bump to a release bundling both versions, version threaded to both service builders and the type-URL resolver; matches the versioning framework.
**Breaks dependents:** no — grep confirms no other module imports the versioned `google.ads.googleads.vNN.*` SDK modules (the CDP Google Ads destination uses the Data Manager API); v23 modules are retained in 30.1.0. The 29→30 major bump is a real risk surface but is covered by the 143-test unit suite at the mocked SDK boundary; live-network behavior is untested (no creds, disclosed).
**Undisclosed side effects:** schema discovery for existing v23-pinned sources now runs against v24 — see finding.

### Findings
- [should-fix] products/warehouse_sources/backend/temporal/data_imports/sources/google_ads/source.py:130 — `get_schemas` (the discovery/reconciliation path, run periodically by `sync_new_schemas.py`) hardcodes `self.default_version` (now v24), so an existing **v23-pinned** source discovers its schemas against v24. This contradicts the PR's "existing sources keep their pin and are untouched" claim: before this PR discovery ran at v23 (the only bundled version); after it, all sources — including v23 ones — discover at v24, while their actual sync correctly uses the pin. Impact is low in practice (the discovered output is table-name granularity over the static `RESOURCE_SCHEMAS`), but it is a genuine version divergence between discovery and sync for the same source. Fix: thread the source's resolved pin into discovery — `source.api_version` is available at the caller (`sync_new_schemas.py:72`, from the `ExternalDataSource` fetched at line 52) — instead of hardcoding `self.default_version`; if the framework's `get_schemas` signature can't carry the pin yet, soften the PR claim to acknowledge discovery moves to the default version.

### Verdict
Solid, well-tested version add: the SDK bump, dual-version proto resolver, and end-to-end threading of the pin through the sync path are all correct, and the new tests target the right seams (resolved version reaching `get_service`, per-version type-URL decoding, unknown-version raising). The one thing worth addressing before merge is that the schema-discovery path still uses `default_version` rather than the source's pin, which means an existing v23 source is not quite "untouched" as the description states — either thread the pin or adjust the claim.

products/warehouse_sources/backend/temporal/data_imports/sources/google_ads/source.py
```diff
@@ -127,6 +127,7 @@ def get_schemas(
         google_ads_schemas = get_google_ads_schemas(
             config,
             team_id,
+            self.default_version,
```
danielcarletti: [should-fix] Discovery hardcodes `self.default_version` (now v24), so an existing v23-pinned source discovers schemas against v24 while its sync correctly uses the pin — a version divergence that contradicts the PR's "existing sources are untouched" claim (before this PR discovery ran at v23, the only bundled version). Impact is low (output is table-name granularity over static RESOURCE_SCHEMAS) but real. Fix: thread the source's resolved pin into discovery — `source.api_version` is available at the caller (sync_new_schemas.py:72, from the ExternalDataSource fetched at line 52) — instead of `self.default_version`; if the framework signature can't carry it yet, soften the claim to say discovery uses the default version.

danielcarletti (commented):
Schema discovery (get_schemas) still uses default_version rather than the source pin, so an existing v23 source is not fully "untouched" — either thread the pin or soften the claim. Otherwise the v24 add is correct and well-tested.

Gilbert09:
Thanks @danielcarletti — good catch, and you're right that discovery diverges from sync for an existing v23 source.

I went with your second option (adjust the claim) rather than threading the pin, because the pin isn't reachable from `get_schemas` today: the base `_BaseSource.get_schemas(config, team_id, ...)` signature has no version parameter, and it's called from several places (`sync_new_schemas.py`, `get_documented_tables`, source setup). Threading `source.api_version` through it is a framework-wide signature change across every source, which is outside this PR's scope.

So the sync path honors each source's pin end to end (v23 stays v23), while discovery runs at `default_version` (v24) for all sources. In practice the discovered output is table-name granularity over the static `RESOURCE_SCHEMAS`, so it's identical across v23/v24 here. I've updated the PR description to state this accurately instead of claiming existing sources are fully "untouched". Threading the pin into discovery would be a good follow-up when the framework signature grows a version parameter.

danielcarletti:
AI reply: Addressed the discovery-path divergence raised in review. Schema discovery (`sync_new_schemas_activity`) now threads the source pin into `get_schemas` for multi-version sources (`source.api_version`, gated on `len(supported_versions) > 1` so single-version sources are untouched), and `GoogleAdsSource.get_schemas` resolves it via `self.resolve_api_version(api_version)` instead of hardcoding `self.default_version`. So a v23-pinned source now reconciles schemas under v23, matching its sync path. Added `api_version` to the base `get_schemas` signature (documented contract), plus tests: the sync activity threads the pin only for multi-version sources, and the Google Ads override discovers against the resolved pin. Also appended a learning to the versioning skill. Note: the user-initiated "refresh schemas" view paths still discover at default — out of scope here, flagging for a follow-up if we want them pinned too.

danielcarletti (approved):
LGTM

products/warehouse_sources/backend/temporal/data_imports/sources/google_ads/source.py
Gilbert09: Fixed — discovery now threads the source's resolved pin instead of hardcoding \`default_version\`. \`GoogleAdsSource.get_schemas\` accepts \`api_version\` and resolves it via \`self.resolve_api_version(...)\`, and \`sync_new_schemas_activity\` passes \`source.api_version\` for multi-version sources, so a v23-pinned source discovers schemas under v23 (matching its sync path) rather than the new v24 default.

To keep the base \`get_schemas\` signature Liskov-compatible across every single-version source, the \`api_version\` param lives only on multi-version overrides (not on \`_BaseSource.get_schemas\`); the call site narrows via a small \`Protocol\`.

PR #71697, opened 2026-07-16
fix(exports): stop ai subscription planner emitting invalid HogQL

slshults:
## Problem

PostHog AI prompt subscriptions have a planner LLM write 1-25 HogQL steps from the user's report prompt.
When a prompt requires enumerating property KEYS (e.g. "are properties being sent that aren't in our spec?"), the planner reliably reaches for ClickHouse Map idioms that HogQL can't run. In a user-reported example prompt subscription, 10 of 11 generated queries failed, So, this seems worth fixing prior to the migration to sandboxing.

Three failing patterns, all reproduced read-only during the investigation of support ticket [#62614](https://posthoghelp.zendesk.com/agent/tickets/62614):

1. `mapKeys(properties)`. In HogQL, `properties` (and `person.properties`) is a JSON String column, not a Map, so this fails at ClickHouse execution ("Function mapKeys requires at least one argument of type Map"). Correct: `JSONExtractKeys(properties)`.
2. `flatten(...)`. Not in HogQL's function set, fails validation. Correct: `arrayFlatten(...)`.
3. Nested aggregates, e.g. `countIf(x) / max(count())`.

The pipeline can't self-heal these today. ClickHouse runtime errors aren't in `_RETRYABLE_QUERY_ERRORS` in `report_pipeline.py`, so the HogQL fix loop never gets a chance to repair them, and the step just fails with "This query failed to run due to an internal error."

This is an interim, prompts-only fix to improve this until the migration to sandboxes.

## Changes

All in `products/exports/backend/temporal/subscriptions/ai_subscription/prompts.py`:

- `PLAN_GENERATION_PROMPT`: three new bullets in the "HogQL syntax constraints" list covering the patterns above, plus a "Property-keys audit" reference pattern (`arrayJoin(JSONExtractKeys(properties))` with `{{date_range}}`) that was verified against live data.
- `HOGQL_FIX_PROMPT`: mirrors the same three rules so the fix loop can repair these when it does see them (parse-time failures like `flatten`).

Note: both prompts are team-overridable at runtime via the LLMPrompt product (`ai-subscription-planner`, `ai-subscription-hogql-fix`), so this only changes the code defaults used by teams without managed overrides. That's expected.

- The replacement query patterns were verified read-only against live production data during the ticket investigation (the `JSONExtractKeys`/`arrayFlatten` forms ran successfully; the `mapKeys`/`flatten` forms reproduced the failures).
- `ruff check` and `ruff format --check` clean.
- Existing suite `test_prompts.py` passes (8 passed). No existing tests assert prompt text, and no tests were added: an assertion that the prompt string contains certain words would be a pure change detector, and the prompts are LLM instructions whose effect can't be meaningfully unit-tested.
- Not done: an end-to-end LLM run of the planner with the new prompt (no local LLM harness for this pipeline).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior or documented workflow changes, prompt tuning only.

## 🤖 Agent context

Authored by Claude (Fable 5) in PostHog Code, directed by @slshults, from an investigation handoff for ticket [#62614](https://posthoghelp.zendesk.com/agent/tickets/62614). The `code-reviewer` agent reviewed the diff before commit and verified the HogQL claims against the codebase (`JSONExtractKeys` and `arrayFlatten` are registered HogQL functions; `flatten` is not; `properties` is a `StringJSONDatabaseField`).

Decision: kept this prompts-only rather than also widening `_RETRYABLE_QUERY_ERRORS` in `report_pipeline.py` to let ClickHouse runtime errors into the fix loop. That widening crosses a deliberate trust boundary (`_safe_error_message` intentionally withholds ClickHouse error text, which can echo team-scoped identifiers) and deserves its own PR if we want it.

---


andyzzhao (approved):
approving as a stopgap for now but some prompt consolidation would be nice so we don't have drift

PR #71713, opened 2026-07-16
fix(llm-gateway): keep bedrock fallback alive by scrubbing unsupported request shape

cvolzer3:
## Problem

When Anthropic returns 5xx/429 (e.g. `overloaded_error` / 529 during capacity crunches), the LLM Gateway fails over to Bedrock-hosted Claude. But the Bedrock request was being built with shapes Bedrock rejects, so the fallback returned a 400 immediately instead of serving the request — defeating the point of having a fallback. Two concrete causes:

- `output_config.format` was in the Bedrock allowlist and forwarded verbatim. Bedrock-hosted Claude rejects it with `output_config.format: Extra inputs are not permitted`. Bedrock's error names the *nested* path, so it accepts `output_config` itself (e.g. `{"effort": ...}`) but not the `format` sub-key.
- Server-side tool *definitions* (`web_search`, `web_fetch`, `code_execution`) were already stripped, but references to them left elsewhere were not: `server_tool_use` / `*_tool_result` blocks in the message history, and a `tool_choice` naming a stripped tool. Bedrock then 400s with `Tool 'web_search' not found in provided tools`.

Context (auth-gated): [Slack thread](https://posthog.slack.com/archives/C0ASEL8GJQM/p1784227905060309?thread_ts=1784227905.060309&cid=C0ASEL8GJQM)

## Changes

In `services/llm-gateway/src/llm_gateway/api/anthropic.py`, `sanitize_for_bedrock` now additionally:

- Prunes unsupported `output_config` sub-keys (`format`) while preserving the rest of `output_config`, so structured-output requests degrade gracefully on the fallback instead of 400-ing.
- Scrubs dangling server-side tool references from the message history (`server_tool_use`, `mcp_tool_use`, and `*_tool_result` blocks — plain client `tool_result` is kept) and drops a `tool_choice` that can no longer be satisfied — a named tool that was stripped, or any `tool_choice` at all once no tools remain (even `auto`/`none`, which are no-ops without tools but still risk rejection).

Each strip emits the existing `BEDROCK_PARAM_STRIPPED` warning/metric so we keep visibility into what's being dropped.

Not included: surfacing the swallowed Bedrock error to the client. On fallback failure the gateway re-raises the original (retryable) Anthropic error; returning the downstream Bedrock 4xx/5xx instead would mislead clients into treating a transient overload as a permanent bad request. Happy to add it if we'd rather trade that off — flagging for a decision rather than shipping silently.

Automated unit tests only (I did not run the service against live Bedrock). Added to `tests/test_anthropic.py`:

- `TestStripStructuredOutputFormat` — `format` pruned, `effort` kept, `output_config` dropped when emptied, no-op when absent/not-a-dict.
- `TestStripServerSideToolUsesFromMessages` — server-side blocks stripped, client `tool_result` kept, message dropped when content empties, string content untouched.
- `TestReconcileToolChoice` — drops any `tool_choice` once no tools remain (named, `any`, `auto`, `none`); keeps a valid named choice and `any`/`auto`/`none` while client tools remain.
- Integration asserts through `sanitize_for_bedrock` in `TestSanitizeForBedrock`, including that stripping the only (server-side) tool also drops an `auto` tool_choice, and that the message-scrub and tool_choice strips increment `BEDROCK_PARAM_STRIPPED`.

These catch regressions where a Bedrock-incompatible field (nested `output_config.format`, dangling server-side tool reference, or unsatisfiable `tool_choice`) survives sanitization — the exact shapes seen 400-ing the fallback in prod, none of which existing tests covered.

`uv run pytest tests/test_anthropic.py` → 158 passed. `ruff check` / `ruff format` clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Directed by Chris Volzer via Slack while triaging an LLM Gateway high-error-rate alert. The alert was driven by upstream Anthropic 529 overloads; investigation of the `llm-gateway` service logs showed the Bedrock fallback was itself 400-ing on request shape, so it wasn't absorbing the load. This PR fixes the two request-shaping bugs behind those fallback 400s.

Authored by the PostHog Slack app (Claude Opus 4.8). No repo skills were required for these changes (pure Python service logic + unit tests). Considered dropping `output_config` wholesale, but Bedrock's nested error path indicated only the `format` sub-key is unsupported, so pruning sub-keys preserves `effort`. Deliberately scoped out changing the client-facing error on fallback failure (see Changes) to avoid turning a retryable error into a terminal one.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ASEL8GJQM/p1784227905060309?thread_ts=1784227905.060309&cid=C0ASEL8GJQM)*

### DIFF
--- a/services/llm-gateway/src/llm_gateway/api/anthropic.py
+++ b/services/llm-gateway/src/llm_gateway/api/anthropic.py
@@ -281,6 +281,7 @@ def strip_server_side_tool_uses_from_messages(data: dict[str, Any], *, model: st
     if stripped:
         data["messages"] = new_messages
         logger.warning("Stripping server-side tool blocks from messages for Bedrock", model=model, product=product)
+        BEDROCK_PARAM_STRIPPED.labels(param="messages.server_tool_blocks", product=product).inc()
 
 
 def _is_server_side_tool_block(block: Any) -> bool:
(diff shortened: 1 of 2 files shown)


brandonleung (approved):
LGTM

PR #71722, opened 2026-07-16
fix(shortcuts): match ⌘⌥ shortcuts by layout-aware key, not physical position

haacked:
## Problem

On non-QWERTY keyboard layouts (Dvorak, AZERTY, etc.), `⌘⌥` (command+option) shortcuts fired the wrong action because the shortcut engine resolved the pressed letter from the **physical key position** (`event.code`) instead of the layout-aware `event.key`. On Dvorak the key labelled "C" sits where QWERTY has "I", so a `⌘⌥C` press was matched as if it were `⌘⌥I` — landing the user on the wrong destination.

Since every app shortcut uses `command+option` (`baseModifier`), this affected all of them on non-QWERTY layouts.

## Changes

`shortcutLogic` now prefers `event.key` (correct on every layout) when matching the `⌘⌥` chord. It only consults the physical `event.code` for the macOS quirk the branch was originally written for — where holding Option turns `event.key` into a special glyph or dead key (e.g. `⌥K` → "˚") — and never when `event.key` already gives a usable letter or digit. This brings the engine in line with the older `useKeyboardHotkeys` hook, which already matches on `event.key`.

Added `shortcutLogic.test.ts` with two cases:
- **Layout-aware matching** — a Dvorak-style event (`event.key: 'c'`, `event.code: 'KeyI'`) triggers the `⌘⌥C` shortcut and not `⌘⌥I`. This fails if the engine reverts to matching on `event.code`.
- **macOS Option-glyph fallback** — an event whose `event.key` is the glyph "˚" with `event.code: 'KeyK'` still triggers `⌘⌥K`. This fails if the physical-key fallback is dropped, which would break Option shortcuts on macOS.

No existing test covered `shortcutLogic`, so neither regression was previously caught. I was unable to run the JS test suite in this environment (no `node_modules`/toolchain available) — CI will execute the tests.

## Docs update

No docs change — internal keyboard-handling behavior only.

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread. Traced the regression: the current shortcut engine (`shortcutLogic`, introduced in #41677) reimplemented keyboard handling from scratch and reintroduced physical-key (`event.code`) matching in its Option branch, whereas the older `useKeyboardHotkeys` hook it sits alongside matches on the layout-aware `event.key`. Fix keeps the macOS Option-dead-key fallback (the branch's original purpose) but gates it so it only triggers when `event.key` isn't already a usable letter/digit.

Skill invoked: `/writing-tests` (to gate the added regression tests).

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1784230051235469)*

### DIFF
--- a/frontend/src/lib/components/Shortcuts/shortcutLogic.test.ts
+++ b/frontend/src/lib/components/Shortcuts/shortcutLogic.test.ts
@@ -1,0 +1,60 @@
+import { initKeaTests } from '~/test/init'
+
+import { shortcutLogic } from './shortcutLogic'
+
+jest.mock('lib/posthog-typed', () => ({
+    __esModule: true,
+    default: { __loaded: true, capture: jest.fn() },
+}))
+
+// The ⌘⌥ matching path branches on isMac(), which is captured once at module load.
+jest.mock('lib/utils/dom', () => ({
+    ...jest.requireActual('lib/utils/dom'),
+    isMac: () => true,
+}))
+
+describe('shortcutLogic', () => {
+    let logic: ReturnType<typeof shortcutLogic.build>
+
+    beforeEach(() => {
+        initKeaTests()
+        logic = shortcutLogic()
+        logic.mount()
+    })
+
+    afterEach(() => {
+        logic.unmount()
+    })
+
+    function register(name: string, keybind: string[][]): jest.Mock {
+        const callback = jest.fn()
+        logic.actions.registerShortcut({ name, keybind, intent: name, interaction: 'function', callback })
+        return callback
+    }
+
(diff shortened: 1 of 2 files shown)


haacked:
Closing — this was opened against the wrong repo. The reported bug (Dvorak Cmd+C navigating to the inbox) is in PostHog/code, where the Inbox shortcut is `mod+i` matched by react-hotkeys-hook's physical-key (event.code) logic. The real fix is PostHog/code#3525. The change here was an incidental, unrelated hardening of this app's own shortcut engine and isn't needed.

PR #71735, opened 2026-07-16
fix(clickhouse): retry 429 rate limit during connect inline

Gilbert09:
## Problem

Error tracking surfaced a `ClickHouseConnectionError` from the ClickHouse import source:

> HTTPDriver for `<redacted host>` returned response code 429

Issue: https://us.posthog.com/project/2/error_tracking/019f6c8b-ecb2-7302-89ea-cedf3f72a9e3

The failure originates in this source's `_get_client` (`products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py`), which wraps connect-time driver errors as `ClickHouseConnectionError`. A `429 Too Many Requests` is a rate limit: the server (or a proxy in front of it) is explicitly asking us to slow down and try again. Today the first 429 during connect fails the connect immediately.

## Changes

A 429 is transient, not a config error, so I retry it inline with exponential backoff, mirroring the existing transient connect-drop handling (`_TRANSIENT_CONNECT_DROP_SUBSTRINGS`). The backoff base is a bit longer than the connect-drop retry to give the rate limit room to clear.

> [!NOTE]
> This is **not** added to `get_non_retryable_errors` - retrying is the correct response to a rate limit. It also remains retryable at the Temporal level as a backstop if the local retries are exhausted. Matching is on the stable `returned response code 429` phrase only, so the volatile per-request URL never enters the comparison. Only 429 is matched: 404 stays deterministic-non-retryable and 5xx stay retryable, unchanged.

Added automated tests (no manual/live testing - I can't reach a rate-limited ClickHouse host):

- `TestIsRateLimited` - `_is_rate_limited` matches the stable 429 phrase and does **not** match 404 / 502 / auth errors (which have their own handling). Catches a regression that broadens or breaks the match.
- `TestGetClientTransientRetry::test_retries_rate_limit_then_succeeds` - a 429 on the first connect is retried and the second attempt's client is returned. Catches a regression where a 429 is no longer retried inline.
- Added a 429 case to `test_transient_errors_are_retryable` - locks in that 429 is never treated as non-retryable.

Ran the affected tests locally with `pytest` on `test_clickhouse.py`; all pass (one pre-existing DB-backed reconcile test errors only because this sandbox has no Postgres - unrelated to this change).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Triaged from an error-tracking webhook by Claude (Claude Code). I read the stack into `_get_client`, confirmed the 429 is raised at connect time and wrapped as `ClickHouseConnectionError`, and classified it as a transient rate limit rather than a user/upstream config error or a code bug. Considered and rejected adding it to `get_non_retryable_errors` (that would stop retrying a rate limit, which is wrong). Chose inline retry with backoff to match the maintainer's established pattern for transient HTTP statuses and this file's own connect-drop retry. Invoked the `/writing-tests` skill before adding tests. Checked open PRs (including the maintainer's own) for a duplicate - none addresses this.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
@@ -135,12 +135,21 @@ def _is_transient_connect_drop(error_message: str) -> bool:
 # slow down. A 429 is explicitly "retry later", so a brief backed-off re-attempt
 # often clears a short rate-limit burst; if it doesn't, the failing Temporal
 # activity stays retryable and recovers later. We match only 429 — other 4xx
-# response codes are deterministic (e.g. 404 stays non-retryable).
+# response codes are deterministic (e.g. 404 stays non-retryable). Matching the
+# stable status phrase keeps the volatile per-request URL out of the comparison.
 _TRANSIENT_RATE_LIMIT_SUBSTRING = "returned response code 429"
+
+# Backoff base between connect retries after a 429. Longer than the connect-drop
+# retry (which just re-dials) to give the rate limit room to clear.
+_RATE_LIMIT_BACKOFF_BASE_SECONDS = 2
+
+
+def _is_rate_limited(error_message: str) -> bool:
+    return _TRANSIENT_RATE_LIMIT_SUBSTRING in error_message
 
 
 def _is_retryable_connect_error(error_message: str) -> bool:
-    return _is_transient_connect_drop(error_message) or _TRANSIENT_RATE_LIMIT_SUBSTRING in error_message
+    return _is_transient_connect_drop(error_message) or _is_rate_limited(error_message)
 
 
 def _get_client(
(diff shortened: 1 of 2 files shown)


Gilbert09:
Rebased onto master to resolve the conflict. In the meantime master independently landed the base of this fix — `_get_client` now already retries a 429 during connect (via `_is_retryable_connect_error`), with the 429 case tested. Rather than duplicate that, I narrowed this branch to its genuine remaining delta:

- 429 now backs off **exponentially** (base 2 → 2s, 4s) instead of the linear `attempt` wait, to give the rate limit room to clear; a plain connect-drop still uses the short linear wait.
- Kept the `_is_rate_limited` helper (reused inside `_is_retryable_connect_error`) and its `TestIsRateLimited` coverage.
- Dropped the now-redundant/duplicate 429 retry tests that master already carries and replaced them with `test_rate_limit_backs_off_exponentially`, which locks in the distinct backoff behavior.

Ran the affected unit tests locally (29 passed) plus ruff lint/format.

PR #71738, opened 2026-07-16
fix(clickhouse): stop capturing transient rate-limit noise in partition probe

Gilbert09:
## Problem

Error tracking surfaced a handled `OperationalError` from the ClickHouse data-warehouse import source:

> `HTTPDriver for <redacted host> returned response code 429`

The exception originates in `_get_partition_settings` in `products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py`.
That function is a best-effort optimization: it reads `system.tables.total_bytes` to size Delta partitions, and on any failure it already falls back to default partitioning and returns `None`.

The problem is that its `except ClickHouseError` block calls `capture_exception` on every failure, including a transient HTTP 429 (rate limited) from the source (or a proxy/LB in front of it).
The sync itself does not fail here, the error is caught and recovered from (`$exception_handled: true`).
So the only effect is error-tracking noise for something that is not actionable on our side and clears on its own.

This is the same class of issue the sibling duplicate-primary-key probe already handles via its expected-probe-failure allowlist.

## Changes

`_get_partition_settings` now skips `capture_exception` for transient HTTP responses (429 and the gateway codes 502/503/504), while still capturing any other unexpected `ClickHouseError`.
Behavior is otherwise unchanged: it still logs at debug and returns `None`, so the pipeline continues with default partitioning.

A `Code: NNN` from a real ClickHouse query error does not match the transient-status substrings, so genuine failures still surface.

This is deliberately not a `NonRetryableErrors` change: the error is swallowed inside a best-effort probe and never reaches the pipeline retry machinery, and transient rate limits should stay retryable everywhere else.

Added `TestGetPartitionSettings` with two cases:

- transient HTTP responses (429/502/503/504) return `None` **without** calling `capture_exception` — this is the regression the fix targets, and no existing test covered `_get_partition_settings` capture behavior.
- a genuine `Code: 62` error still calls `capture_exception` — guards against over-suppressing real failures.

Ran locally:

```
pytest .../clickhouse/test_clickhouse.py::TestGetPartitionSettings          # 5 passed
pytest .../clickhouse/test_clickhouse.py::TestHasDuplicatePrimaryKeys \
       ::TestGetIncrementalRowCount ::TestClickHouseSourceNonRetryableErrors # 31 passed
ruff check + ruff format --check                                            # clean
```

## 🤖 Agent context

Triaged from an error-tracking webhook by Claude (Claude Code). I confirmed the issue in error tracking (pulled the stack trace, exception type, and that it was handled), traced the call path into `_get_partition_settings`, and confirmed via the `clickhouse-connect` exception hierarchy that `OperationalError` is a `ClickHouseError` subclass so it hits that `except` block.

Decision: this is a fixable bug (over-eager capture in a graceful-fallback path), not a `NonRetryableErrors` case, because the error is already swallowed and transient rate limits should remain retryable. Chose to mirror the file's existing `_is_expected_probe_failure` pattern rather than remove capturing entirely, so genuine unexpected errors still surface. Checked open PRs (including the maintainer's) for a matching fix and found none.

Invoked the `/writing-tests` skill before adding the regression test.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
@@ -899,6 +899,24 @@ def _get_incremental_row_count(
     return int(count) if count is not None else None
 
 
+# clickhouse-connect surfaces a non-2xx HTTP status from the server (or a
+# proxy/LB in front of it) as `HTTPDriver for <url> returned response code <N>`.
+# 429 (rate limited) and the transient gateway codes mean the endpoint can't
+# serve us right now, not that anything we sent was wrong — they clear on their
+# own. A real ClickHouse query error carries a `Code: NNN` instead. We match
+# only these transient statuses so genuine failures still surface.
+_TRANSIENT_HTTP_RESPONSE_SUBSTRINGS: tuple[str, ...] = (
+    "returned response code 429",
+    "returned response code 502",
+    "returned response code 503",
+    "returned response code 504",
+)
+
+
+def _is_transient_http_response(message: str) -> bool:
+    return any(substring in message for substring in _TRANSIENT_HTTP_RESPONSE_SUBSTRINGS)
+
+
 def _get_partition_settings(
     client: ClickHouseClient, database: str, table_name: str, logger: FilteringBoundLogger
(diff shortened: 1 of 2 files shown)


danielcarletti (approved):
LGTM

PR #71741, opened 2026-07-16
fix(clickhouse): tolerate readonly source profiles rejecting settings

Gilbert09:
## Problem

The ClickHouse data-warehouse import source crashes at connect time for customers whose ClickHouse user has a readonly profile, which is common on managed offerings. Error tracking surfaced it as `ClickHouseConnectionError: Setting max_block_size is unknown or readonly`, raised from `_get_client`.

Root cause: `get_rows` opens the streaming client with a batch of performance and Arrow-output tuning settings (`max_block_size`, `output_format_arrow_string_as_string`, `optimize_read_in_order`, and so on) passed at client construction. On a readonly profile, `system.settings` reports every setting as readonly, so clickhouse-connect refuses to send them and raises `ProgrammingError` ("Setting `<x>` is unknown or readonly"). Since that happens during construction, `_get_client`'s connect-error handler rewraps it as a fatal `ClickHouseConnectionError` and the whole sync fails.

These settings are tuning hints, not correctness requirements (modern ClickHouse defaults already match what we ask for). The sync should degrade to the server defaults, not die. The other settings-bearing paths in this file (`_has_duplicate_primary_keys`, `_get_incremental_row_count`) already catch `ClickHouseError` and fall back. Only client construction was unguarded.

## Changes

Apply the tuning settings after connect via `set_client_setting`, one at a time, tolerating a per-setting readonly rejection and logging a warning instead of failing. Settings the server accepts still take effect; the rest fall back to the server default.

```diff
- return get_client(..., settings=settings or {}, compress=True)
+ client = get_client(..., compress=True)
+ ...
+ _apply_session_settings(client, settings or {})  # skips readonly-rejected settings
+ return client
```

Added `TestGetClientSessionSettings` with two cases, both of which I confirmed fail on the pre-fix code and pass after:

- `test_readonly_setting_does_not_fail_connection` — a client whose `set_client_setting` raises `ProgrammingError` for one setting no longer fails the connection, and the accepted settings are still applied. This is the exact regression from the error-tracking issue.
- `test_settings_applied_after_connect_not_at_construction` — guards that tuning settings are applied post-connect (where they can be dropped individually) rather than at construction (where one readonly setting kills the whole connect).

I (Claude) ran the source's unit tests locally: all pass except one pre-existing DB-backed test (`TestClickHouseReconcileSchemaMetadata`) that needs Postgres, unrelated to this change. `ruff check` and `ruff format --check` are clean on both files.

## 🤖 Agent context

Triaged from a PostHog error-tracking issue for the ClickHouse import source. I (Claude) confirmed the issue in error tracking, traced the stack (`get_rows` → `_get_client` → clickhouse-connect's `_validate_settings`), read the clickhouse-connect internals to confirm the default `invalid_setting_action='error'` behavior and that `system.settings.readonly` reflects the user profile, and judged this a fixable fragility rather than a user/upstream error worth adding to `NonRetryableErrors` (the settings are optional tuning; we should degrade, not stop retrying). Considered flipping the process-global `invalid_setting_action` to `drop` but rejected it — `clickhouse_connect` is also used for our internal ClickHouse in `posthog/clickhouse/client/connection.py`, so a global change would weaken validation there. The per-setting `set_client_setting` approach keeps the fix scoped to this source. Invoked the `/writing-tests` skill before adding tests.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/clickhouse/clickhouse.py
@@ -13,7 +13,7 @@ import pyarrow as pa
 import structlog
 from clickhouse_connect import get_client
 from clickhouse_connect.driver.client import Client as ClickHouseClient
-from clickhouse_connect.driver.exceptions import ClickHouseError
+from clickhouse_connect.driver.exceptions import ClickHouseError, ProgrammingError
 from dlt.common.normalizers.naming.snake_case import NamingConvention
 from structlog.types import FilteringBoundLogger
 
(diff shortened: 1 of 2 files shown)


danielcarletti (approved):
LGTM

Issue #66176, opened 2026-06-25
chore(flags): Phase 2 — accept PSAK on the Django flag surfaces


haacked:
Phase 2 of #63111.

Accept PSAK on the Django surfaces that authenticate with the legacy team secret token today. Keep `TeamSecretTokenAuthentication` in place for back-compat throughout.

## Tasks

- [x] `remote_config`: add `ProjectSecretAPIKeyAuthentication` and `psak_allowed_actions = ["remote_config"]`, keep `TeamSecretTokenAuthentication`, and confirm scope enforcement and team binding.
- [ ] Conversations external API: add PSAK support alongside (or replacing) the direct `Q(secret_api_token=…)` lookup. This product chose the team secret token deliberately because `api_token` is public, so loop in its owners and confirm PSAK semantics fit.
- [x] Resolve the header-only vs body-token difference. Legacy auth accepts the token in the request body; PSAK is header-only. Audit whether anything sends the secret token in the body for these endpoints; if so it stays on the legacy path until it moves to the header.

## Acceptance

- A `feature_flag:read` PSAK authenticates `remote_config` (and any other migrated surface) with correct scope and team enforcement.
- Existing team secret tokens keep working unchanged.

patricio-posthog:
Body token audit is done. I checked all our server SDKs (python, node, php, go, ruby, rust, dotnet and java) and every single one sends the key as an Authorization header, none of them ever puts it in the request body. On top of that, the counter we've had in production since March (`api_auth_secret_api_key_body`) has never fired once, in [US](https://grafana.prod-us.posthog.dev/d/flags-secret-token-migration/0db7ec3?orgId=1&from=now-90d&to=now&timezone=utc&var-datasource=victoriametrics&refresh=5m) or [EU](https://grafana.prod-eu.posthog.dev/d/flags-secret-token-migration/0db7ec3?orgId=1&from=now-6h&to=now&timezone=utc&var-datasource=eflgazs0l40e8d&refresh=5m). Also all the endpoints that accept this token are GET or read the header only, so there is nowhere a body token could even arrive.

Issue #63111, opened 2026-06-11
chore(flags): Migrate Feature Flag Secure API Keys to Project Secret API Keys


haacked:
## Summary

Feature flag local evaluation and remote config are authenticated with a per-project team secret token (`Team.secret_api_token` / `secret_api_token_backup`). These are `phs_`-prefixed credentials stored in plaintext in Postgres, and compared in plaintext in both Django and the Rust feature-flags service.

Now that Project Secret API Keys (PSAK) are live, we should move these surfaces over to PSAK and delete the plaintext columns. PSAKs are SHA256-hashed at rest, carry explicit scopes, get query-log attribution, and have a proper CRUD UI (up to 50 per project).

The catch is we can't break existing customers. Their SDKs are already configured with a team secret token, so the plan is additive and opt-in first. We only deprecate and drop the columns once usage hits zero.

## What's there today (legacy)

Storage, plaintext, no encryption:

- `Team.secret_api_token` and `Team.secret_api_token_backup`, both `CharField(max_length=200, null=True)` with unique indexes (`posthog/models/team/team.py`, migrations `0726`/`0727`).
- Generated by `generate_random_token_secret()` as `"phs_" + generate_random_token(35)` (`posthog/models/utils.py`).
- One active plus one backup per project. Rotation promotes active to backup via `Team.rotate_secret_token_and_save()` / `delete_secret_token_backup_and_save()`, exposed as `PATCH /api/environments/{id}/rotate_secret_token` (admin only).
- Cached in Redis as plaintext (`get_team_in_cache` / `set_team_in_cache`).

Auth (`posthog/auth.py`):

- `TeamSecretTokenAuthentication` pulls a `phs_` token from the Authorization header or the request body (`allow_body_token=True`), looks the team up by plaintext match, and returns a synthetic `TeamSecretTokenUser(team)`.
- Gated by `TeamSecretTokenPermission` (`posthog/permissions.py`), which blocks cross-team access.
- Tagged in the query log as `AccessMethod.TEAM_SECRET_TOKEN`.

Consumers:

- Django: `remote_config` (`products/feature_flags/backend/api/feature_flag.py:3756`), the live debugger (`products/live_debugger/backend/api.py:248`), and the conversations external API (`products/conversations/backend/api/external.py`, which does a direct `Q(secret_api_token=…) | Q(secret_api_token_backup=…)` lookup rather than going through the auth class).
[...]

PR #52333, opened 2026-03-25
fix: improve secret API key validation

haacked:
## Problem

`ProjectSecretAPIKeyAuthentication.find_secret_api_token` enforces the `phs_` prefix for tokens in the `Authorization` header but accepts any value from `data["secret_api_key"]` in the request body. This means an attacker can use a public project token (`phc_*`) — exposed in frontend JS snippets — to authenticate against secret-key-gated endpoints (`local_evaluation`, `remote_config`, `active_breakpoints`).

All three affected endpoints are read-only (GET), so the impact is data exposure (feature flag definitions, remote config values, debugger breakpoints), not integrity or availability.

## Changes

- **Enforce `phs_` prefix on body-extracted tokens** (`posthog/auth.py`): Apply the same `phs_` prefix + alphanumeric regex validation to body tokens that already exists for header tokens. Non-`phs_` tokens in the body are now silently rejected.
- **Normalize header token extraction** (`posthog/auth.py`): Refactor the header regex to capture the full token first, then validate format separately — consistent with the body path.
- **Add deprecation metric** (`posthog/auth.py`): `SECRET_API_KEY_BODY_COUNTER` Prometheus counter tracks body-based secret key usage for future deprecation planning.
- **Parameterized tests** (`posthog/api/test/test_authentication.py`): Covers public tokens, non-prefixed tokens, empty strings, and integer values in the body — all correctly rejected.

## How did you test this code?

- Added parameterized auth tests for invalid body tokens (public token, non-prefixed, empty, integer) — all rejected
- Ran existing `TestProjectSecretAPIKeyAuthentication` suite — all pass
- Manually tested with curl.

Do not publish to changelog

## Docs update

No docs changes needed.

PR #36150, opened 2025-08-04
fix: Remove support for `secret_api_key` in the query string

haacked:
## Problem

It's not a good idea to put sensitive information in the query string where it can be cached by browsers, proxies, etc.

See this discussion for more context: https://posthog.slack.com/archives/C02E3BKC78F/p1753312954817879

## Changes

Removed the option to retrieve `secret_api_key` from the query string.

## How did you test this code?

Unit tests

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

PR #71753, opened 2026-07-16
fix(auth): remove secret token body support

patricio-posthog:
## Problem

The legacy `TeamSecretTokenAuthentication` accepted the `phs_` secret token in the request body (`secret_api_key` field) as well as the Authorization header. PSAK auth is header-only, so the body path was an obstacle to migrating customers off the legacy token (#63111).

The audit tracked in #66176 confirmed the body path is unused: all nine official server SDKs send the key only as an `Authorization: Bearer` header, every Django endpoint accepting the token is GET-only or header-only, and the production counter `api_auth_secret_api_key_body` (live since #52333, March 2026) never fired once on US or EU. Full evidence: https://github.com/PostHog/posthog/issues/66176#issuecomment-4897855736

## Changes

- `_extract_phs_token` now reads the Authorization header only. Removed the `allow_body_token` parameter, the body-parsing branch, the `secret_api_key` body field constant, and the now-purposeless audit counter.
- Same pattern as the query-string removal (#35606 measured, #36150 removed).

Updated the auth unit tests. The previous "valid token in body authenticates" test is inverted into a regression lock: a valid token in the body must no longer authenticate. Tests that exercised the removed parsing branches (non-dict bodies, invalid body tokens) are deleted since the code path no longer exists; header-token tests are unchanged. All affected tests pass locally (20/20 in `test_authentication.py` for the token classes).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change needed. The docs never documented sending the token in the body.

### DIFF
--- a/posthog/api/test/test_authentication.py
+++ b/posthog/api/test/test_authentication.py
@@ -1977,8 +1977,8 @@ class TestTeamSecretTokenAuthentication(APIBaseTest):
             access_method=AccessMethod.TEAM_SECRET_TOKEN,
         )
 
-    def test_authenticate_with_valid_secret_api_key_in_body(self):
-        # Simulate a request with a valid team secret token
+    def test_authenticate_with_valid_secret_api_key_in_body_not_supported(self):
+        # Body tokens were removed after the audit in #66176: even a valid token must not authenticate.
         wsgi_request = self.factory.post(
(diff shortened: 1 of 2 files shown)


Piccirello (approved):
Love this

PR #60474, opened 2026-05-28
fix(mcp): emit body for tools with anyOf/oneOf request schemas

tomasfarias:
## Problem

Our shiny new file download batch export MCP tool was not working. Apparently the body is not being correctly passed along.

## Changes

I got the robot to fix the robot, and it produced this patch.

## How did you test this code?

I haven't

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Publish to changelog?

No.

## Docs update

No.

## 🤖 Agent context

I got PostHog code to fix posthog code.

PR #71765, opened 2026-07-16
fix(mcp): flatten union tool input schemas for client compatibility

rafaeelaudibert:
## Problem

A customer reported that the `file-download-batch-exports-create` MCP tool fails to load in GitHub Copilot Chat with:

```
Tool mcp_posthog_file-download-batch-exports-create failed validation: object has unsupported top-level schema keyword 'anyOf'
```

The failure happens client side, at tool registration, before any call is made. Copilot Chat translates MCP tool schemas into a restricted JSON Schema subset that requires the root of `inputSchema` to be a plain object. A union keyword (`anyOf`/`oneOf`) at the root gets the whole tool rejected, so it's unusable from Copilot Chat entirely.

Our schema ends up with a root `anyOf` through this chain:

1. The backend endpoint declares its request body with drf-spectacular's `PolymorphicProxySerializer` (three variants: `events`, `persons`, `sessions`, discriminated on `model`), so the OpenAPI spec describes the body as a union of object schemas.
2. Codegen carries that through faithfully: the generated `FileDownloadBatchExportsCreateBody` is a `zod.union`, used directly as the tool's schema.
3. `toMcpInputSchema` converts it to JSON Schema for `tools/list`, producing a top-level `anyOf`. An earlier fix (#60474) made the generated handler send union bodies correctly, and a partial workaround spliced in `type: 'object'` next to the `anyOf` to satisfy the MCP spec, but the root `anyOf` remained and that's exactly what Copilot rejects.

Today only this one tool has a union body, but any future polymorphic endpoint would hit the same wall.

## Changes

In `toMcpInputSchema`, when the converted schema is a root `anyOf`/`oneOf` whose variants are all object schemas, merge them into one flat object schema instead of splicing `type: 'object'` next to the union:

- properties are unioned across variants
- discriminator-style fields (same property, different single-value enums per variant) get their enums merged, so `model` advertises as `enum: ['events', 'persons', 'sessions']`
- `required` keeps only the fields required by every variant

If any variant isn't a plain object, the previous `type: 'object'` splice remains as the fallback.

> [!NOTE]
> The flatten is advertisement-only and intentionally lossy (variant-specific requirements like "only the events model accepts `include`/`exclude`" are no longer schema-enforced). Call-time validation in the executor still parses arguments against the original zod union, so invalid variant mixes are rejected server side before reaching the API. The backend serializer and generated files are untouched; the precise union stays the source of truth everywhere else.

- Added two vitest cases to `tests/unit/tool-catalog.test.ts`. The first catches the regression where a union body would again advertise a root `anyOf` (asserts flat object shape, merged discriminator enum, intersected `required`). The second guards the fallback so non-object-variant unions still advertise `type: 'object'`. No existing test covered the advertised wire shape of union schemas.
- Ran a throwaway test against the real generated `FileDownloadBatchExportsCreateBody` and verified the advertised schema comes out as a flat object with all six properties, the merged `model` enum, and no root union keyword (deleted afterwards, since the unit tests cover the logic).
- `pnpm --filter=@posthog/mcp run typecheck` and the tool-catalog, render-ui, and exec unit suites all pass.
- I did not manually verify against a live Copilot Chat session.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover the MCP tool schema wire format, so nothing to update.

## 🤖 Agent context

Diagnosed and implemented with Claude Code (Fable 5). The `/writing-tests` skill was invoked before adding the unit tests.

Two approaches were considered: flattening in codegen (`generate-tools.ts`) so the tool's zod schema itself becomes a flat object, versus flattening at advertisement time in `toMcpInputSchema`. I (well, Claude) chose the latter because it keeps strict call-time validation against the original union, avoids generated-file churn, and automatically covers any future polymorphic tool. The formatter incidentally rewrote unrelated generated `api.ts` files during the run; those were reverted to keep the diff scoped to the two files.

### DIFF
--- a/services/mcp/src/hono/tool-catalog.ts
+++ b/services/mcp/src/hono/tool-catalog.ts
@@ -25,6 +25,43 @@ export type { PreBuiltTool }
 
 const EMPTY_OBJECT_JSON_SCHEMA = { type: 'object' as const, properties: {} }
 
+type JsonSchema = Record<string, unknown>
+
+/**
+ * Merge a top-level anyOf/oneOf of object variants into a single object schema,
+ * or return null if any variant is not a plain object schema.
+ *
+ * A root union keyword breaks clients that translate tool schemas into a
+ * restricted JSON Schema subset: GitHub Copilot Chat rejects the whole tool at
+ * registration with "object has unsupported top-level schema keyword 'anyOf'".
+ * The merge is advertisement-only and lossy (a variant's extra requirements are
+ * not enforced by the merged schema); call-time validation in the executor still
+ * runs against the original zod union, so invalid variant mixes are rejected.
+ */
+function flattenTopLevelUnion(variants: JsonSchema[]): JsonSchema | null {
+    if (variants.length === 0 || !variants.every((v) => v['type'] === 'object' && v['properties'])) {
+        return null
+    }
+    const properties: Record<string, JsonSchema> = {}
+    for (const variant of variants) {
+        for (const [name, prop] of Object.entries(variant['properties'] as Record<string, JsonSchema>)) {
+            const existing = properties[name]
+            if (!existing) {
+                properties[name] = prop
+            } else if (Array.isArray(existing['enum']) && Array.isArray(prop['enum'])) {
+                // Discriminator fields carry one enum value per variant; merge them
+                // so every variant stays expressible.
+                properties[name] = { ...existing, enum: [...new Set([...existing['enum'], ...prop['enum']])] }
+            }
+        }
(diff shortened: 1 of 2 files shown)


fercgomes (approved):

rafaeelaudibert:
### Merge activity

* **Jul 20, 10:00 PM UTC**: @rafaeelaudibert merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/71765).

PR #71768, opened 2026-07-16
fix(agent-platform): lazy-load seed secret contract

dmarticus:
## Problem

The Agent Builder seed workflow is the automation that keeps the canonical production Agent Builder aligned with the checked-in bundle and current kernel skills. It runs after relevant changes merge and on a daily schedule to pick up backend-delivered kernel skills and correct deployment drift.

The shared seed script began eagerly loading a generated backend secret contract during startup. The production workflow intentionally checks out only the example bundles, so that contract was absent and every seed run crashed before making an API request. As a result, Agent Builder changes could merge without reaching production, kernel-skill updates could not converge after backend deployment, and the scheduled drift-repair path was [failing](https://github.com/PostHog/posthog/actions/runs/29535216730). Production remained on its previous live revision until I noticed and seeded it manually.

The generated contract is needed only by the opt-in local `SEED_DUMMY_SECRETS=1` mode. Normal production seeding should not depend on it.

## Changes

Load the trigger-secret contract only when `SEED_DUMMY_SECRETS=1`. Keep the production workflow's narrow sparse checkout unchanged.

Add a regression test that copies only the production sparse-checkout subtree and verifies the shared seeder starts successfully.

### DIFF
--- a/.github/workflows/cd-agent-builder-seed.yml
+++ b/.github/workflows/cd-agent-builder-seed.yml
@@ -55,9 +55,7 @@ jobs:
             - name: Checkout
               uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
               with:
-                  sparse-checkout: |
-                      products/agent_platform/services/agent-tests/src/examples
-                      products/agent_platform/backend/logic/trigger_required_secrets.generated.json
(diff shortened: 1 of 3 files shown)


daniloc (approved):
🌰

PR #71785, opened 2026-07-16
feat(replay-vision): link the summary header to the run page

ksvat:
## Problem

Summary reports lead with "**Summary for {scanner name}**", but nothing in that header is clickable.
In Slack (and on the digest card) the only links are the cited observations - there's no way to click through to the summary run itself or back to the scanner.
Alert messages already link their header to the run page; summaries never got the same treatment.

## Changes

- The scanner name in the summary header is now a link to this summary's run page (`/replay-vision/actions/{id}/runs/{id}`), which shows the full report, every recording it covered, and breadcrumbs back to the scanner. Same target the alert headers link to.
- The link is added after the external-link strip pass (the citation-link pattern), so the run URL isn't defanged on instances whose `SITE_URL` isn't a posthog.com host.
- The Slack payload picks it up automatically via the existing markdown-link conversion (`<url|scanner name>`).
- Scanner-name sanitization for the header now also strips `[]()` so a free-text name can't break out of the link label (mirrors the alert header sanitizer). A name the strip pass rewrote (e.g. one containing a bare URL) won't match the expected header and stays unlinked rather than guessed at.

Updated the existing header assertions in `test_vision_action_synthesis.py` to lock the linked header in both the stored markdown and the Slack rendering; the name-sanitization and link-defang tests cover the new `[]()` stripping unchanged.
Full synthesis test file (39 tests) green locally.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Implemented with Claude (PostHog Code), directed by Kim ("make the scanner clickable or at least allow the user to click through to the summary run itself"). Linked to the run page rather than the scanner page for consistency with the alert headers, which made the same call - the run page carries breadcrumbs back to the scanner, so both destinations are one click away.

---

### DIFF
--- a/products/replay_vision/backend/temporal/vision_actions/synthesis.py
+++ b/products/replay_vision/backend/temporal/vision_actions/synthesis.py
@@ -175,6 +175,9 @@ def _synthesize(inputs: SynthesizeGroupSummaryInputs) -> SynthesizeGroupSummaryR
     markdown = strip_external_links_markdown(
         _summary_header(action, batch.window_start, len(batch.lines), batch.window_total) + markdown
     )
+    # Linkify AFTER the strip pass (like the citation links) so the run URL isn't defanged on
+    # instances whose SITE_URL isn't a posthog.com host (self-hosted, dev).
+    markdown = _linkify_summary_header(markdown, action, _run_url(team.id, str(action.id), str(run.pk)))
     slack_text = _markdown_to_slack(markdown, team_id=team.id, observation_ids=batch.observation_ids)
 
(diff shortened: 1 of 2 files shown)


ksvat:
* **#71785** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71785?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71785?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #71787, opened 2026-07-17
feat(retention): scan first-time retention in two arms on the query variant

andyzzhao:
## Problem

First-time retention (first occurrence matching filters, and first-ever occurrence) has to find each actor's first-ever start event, so the runner deliberately drops the window bound from the events WHERE clause. On the single-scan shape that means one scan over the team's entire event history, and because the start and return entity conditions sit in a single OR, ClickHouse has to decode the expensive columns (`person_properties`, property maps) for every historical row just to evaluate the filter. Column reads don't short-circuit through an OR.

Replaying a representative production first-time retention insight (8-day window, refreshed several times a day) with direct IO and the condition/uncompressed caches disabled: it reads 2.44 TiB and burns ~192 CPU-seconds per run. Retention is currently the single largest CPU consumer among user-facing query patterns.

## Changes

All on the flag-gated dwh-variant path (`retention-fixed-interval-base-query-dwh-variant`); the legacy default path is untouched.

- `_can_single_scan` now excludes first-time modes, so they build the two-arm UNION ALL shape instead of one unbounded scan.
- Each events arm's WHERE is scoped to its own entity via the new `RetentionQueryRunner.arm_event_filters` instead of sharing the global OR-of-both-entities filter:
  - the start arm scans all time (required for the first-occurrence anchor) but pre-filters to the start entity's event names, so the sort key prunes granules and the expensive columns are only decoded near those rows,
  - the return arm is always bounded to the query window on raw `events.timestamp` (return events outside the window never enter any aggregate, in every retention mode), and since it emits `[] AS start_event_timestamps` it never references start-entity expressions at all.
- `events_where_clause` gains an optional `entities` parameter so the event-name pre-filter can be built per arm; the cohort-breakdown filter is extracted into `_cohort_breakdown_filters` and applied per arm.

Benchmark on the production query above, byte-identical results on every run:

| shape | wall | CPU | read | memory |
|---|---|---|---|---|
| single scan (today) | 5.0 s | 192 s | 2.44 TiB | 3.5 GiB |
| two arms, per-arm WHERE | 1.4 s | 12.7 s | 17 GiB | 630 MB |

A control arm that only added the window bound inside the OR (keeping one scan) read the same 2.4 TiB, confirming the scan split is what matters, not the filter logic.

> [!WARNING]
> Two rollout facts reviewers should know. First, the gating flag `retention-fixed-interval-base-query-dwh-variant` does not exist yet in the internal flag project, so the events-only two-arm path in this PR is unreachable in production until the retention team creates and ramps that flag. Second, retention queries with a data warehouse entity route to the variant builder unconditionally, so the per-arm WHERE rewrite here ships to that surface on deploy regardless of the flag, and turning the flag off does not fully revert this change (a code revert does).

### Multi-agent QA review

A six-reviewer QA pass (database, performance, data-integrity, reliability, two generalists) on the initial version of this branch surfaced findings that are now fixed in follow-up commits:

- First-ever retention with a breakdown could silently mis-bucket actors: per-arm narrowing broke the invariant that both arms resolve the same `argMinIf` breakdown value before the outer `max()` collapse (reproduced empirically with an overlapping action target). The return arm now always degrades its first-ever breakdown value to the empty bucket, making the start arm the sole authority, and a red-green parity test pins it.
- The per-arm filters had dropped the entity property predicate from arm WHERE clauses, un-pruning selective filters on the unbounded start arm. Each arm now pushes its own mode-safe predicate (no-props on the first-ever start arm, with-props elsewhere).
- Same-entity and all-events start entities gained nothing from the split while paying for two scans and two override joins, so `_can_single_scan` now keeps the single pass whenever the split cannot shrink the scan (which also revives the single-scan first-time path for those shapes).
- A zero-step action entity could produce `event IN tuple()` and silently blank the insight; the event-name prefilter is now skipped when an entity resolves to no events.

- Full `test_retention_query_runner.py` suite (145 passed): every test in it runs both the legacy and the variant path and asserts equal results, so the changed variant is checked against the unchanged legacy output across recurring, first-time, first-ever, breakdowns, groups, actions, minimum occurrences, and sampling.
- `test_retention_data_warehouse.py` (21 passed) for the arm construction shared with data-warehouse entities.
- `test_retention_first_time_ever_breakdown_with_overlapping_action_target`: parity + behavioral test for the first-ever breakdown fix, verified to fail without the fix and pass with it.
- New `TestRetentionFirstTimeTwoLegScan` AST-shape tests: first-time modes must produce the two-arm shape with an unbounded start arm and a bounded return arm scoped to their own event names, recurring mode must keep the single scan, same-entity and all-events first-time shapes must keep the single scan (the carve-out), and group-aggregated first-time queries must carry the `$group_N` filter in both arms. No existing test pinned the scan topology, which is the thing this PR changes and the thing a future refactor could silently regress.
- Two SQL snapshots updated; the diff shows exactly the intended structure change.
- Production benchmark: replayed the heaviest real first-time retention query from the query log with the rewritten shape via readonly Metabase access, three arm variants, interleaved runs, caches disabled, identical result matrices.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed: no user-facing behavior change, query results are identical.

## 🤖 Agent context

Written by Claude Code, directed by the assignee, following a query-pain analysis that ranked RetentionQuery as the top fleet CPU pattern. The investigation went: pull the heaviest retention runs from the query log, read the generated SQL, find the unbounded scan, then benchmark three rewrites against production before touching the runner. A bound-inside-the-OR fix was tried first and rejected with data (no read reduction). The two-arm implementation reuses the existing variant UNION architecture rather than adding a new path, and ships behind the same variant flag, so rollout and parity verification ride the existing migration. The `/query-clickhouse-via-metabase` and `/writing-tests` skills were used along the way.

### DIFF
--- a/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
+++ b/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
@@ -282,10 +282,8 @@ class RetentionFixedIntervalBaseQueryBuilder(RetentionBaseQueryBuilder):
         # start and return timestamp arrays can be computed in one pass. Property aggregation has a known
         # legacy/variant discrepancy and stays on the UNION; a data-warehouse entity is a genuinely
         # different source and cannot collapse here.
-        # First-time modes also stay on the UNION: they drop the window bound to find each actor's
-        # first-ever start event, and a single scan then reads every column over all history, while
-        # the two arms confine the unbounded scan to the start entity's rows and keep the return arm
-        # window-bounded.
+        # First-time modes also stay on the UNION: their single scan is unbounded in time (first-ever
+        # anchor), so splitting confines the all-time read to the start entity's rows.
         return (
             not self.has_property_aggregation
             and not self.is_first_occurrence_matching_filters
@@ -503,8 +501,6 @@ class RetentionFixedIntervalBaseQueryBuilder(RetentionBaseQueryBuilder):
 
         table_name = entity.table_name if entity_is_dwh else "events"
         assert table_name
-        # Arm-scoped filters instead of the OR of both entities: each arm pre-filters to its own
-        # entity's event names, and in first-time modes only the start arm scans all time.
         where_expr = (
             None
             if entity_is_dwh
--- a/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
+++ b/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
@@ -119,17 +119,12 @@ class RetentionFixedIntervalBaseQueryBuilder(RetentionBaseQueryBuilder):
 
         if self.is_first_ever_occurrence:
             # First-ever buckets each actor by the breakdown value on their earliest START
-            # event, resolved here via argMinIf(..., start_entity_expr_no_props). The events
-            # return arm can only do that when it shares the events source with the start
-            # entity. When the start entity is a DWH table, start_entity_expr_no_props
-            # collapses to a truthy constant, so on the events return arm argMinIf would
-            # instead read the actor's earliest RETURN event's value. Degrade to the empty
-            # bucket — the start (DWH) arm already emits "", so the outer max() yields "".
-            if (
-                query_kind == "return"
-                and self.start_event.type == EntityType.DATA_WAREHOUSE
-                and self._breakdown_extract_targets_events_table()
-            ):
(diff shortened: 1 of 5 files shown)


thmsobrmlr:
Superseded by #74218

PR #71790, opened 2026-07-17
feat(surveys): show URL reach estimate in the full survey editor

darkopia:
## Problem

The guided survey wizard estimates reach for the URL display condition ("About X unique users viewed matching URLs in the last 30 days"), but the full editor never shows that estimate. The only number visible there is the "Filters match: ~N users" count inside Audience filters, which is a person-pool estimate that ignores runtime display conditions like URL and device targeting. A customer setting up in-app surveys (where one project tracks several surfaces, e.g. an app plus a marketing site) reported via a support request that stakeholders read the audience-filter count as the survey's actual reach and get misled, since the URL condition usually narrows the audience a lot.

## Changes

- Moved the wizard's existing URL reach estimate out of `WhereStep` into a shared `SurveyUrlAudienceEstimate` component. The core implementation is the wizard's code relocated: same HogQL query, same 500ms debounce, same abort handling, same state machine, same copy. It reads the URL condition from `surveyLogic`, so any editor rendered inside a `surveyLogic` `BindLogic` can show it.
- The full editor's "URL targeting" field under Display conditions now renders the same estimate as the wizard. The estimate keeps the wizard's original positive-only scope (contains / exact / regex): negative match types ("is not", "not contains", "not regex") intentionally show no estimate, because a count of users NOT matching a URL is roughly everyone, which costs a maximal scan and misleads more than it informs.
- Guards skip the query for inputs that cannot produce a useful result: empty URL, an exact match against a bare path, and regex patterns rejected by RE2JS. RE2JS (already a dependency) matches ClickHouse's RE2 engine, so JS-only syntax like lookaheads, which is a valid survey condition, no longer sends a query that is guaranteed to fail.
- Hardening from a multi-agent QA review pass: the ClickHouse query is cancelled server-side (via `clientQueryId` + `api.insights.cancelQuery`) when the component unmounts or the condition changes mid-flight; malformed query responses surface the error state instead of "About 0 users"; the loading state shows immediately so a stale count never lingers through the debounce window; the estimate text sits in an `aria-live="polite"` region; and a zero count gets guidance copy ("No pageviews matched this URL condition...") instead of "About 0 unique users".
- Added an info tooltip on "Audience filters" clarifying that its user counts estimate people matching the filters project-wide, and that actual reach is usually smaller because display conditions still apply.
- Dropped the redundant `team_id = <id>` clause from the estimate query when moving it; HogQL queries are always scoped to the current project (enforced server-side by the HogQL printer).

No screenshots, sorry: I authored this in a headless environment. The PR's Storybook visual review run captures the estimate rendering in the full editor (`scenes-app-surveys--new-survey-targeting-section`).

- `SurveyUrlAudienceEstimate.test.ts`: one parameterized suite (14 cases) over the pure `getUrlAudienceEstimateParams` helper, which holds all the gating logic. It catches regressions in the match-type-to-operator mapping, the negative-operator skip, and the guards for empty URLs, exact-match bare paths, and regexes invalid in RE2 — including a lookahead pattern that is valid JS regex but must not reach ClickHouse, which pins the validation to the RE2 engine.
- Ran the existing surveys wizard jest suites to confirm the `WhereStep` refactor is behavior-preserving; all pass alongside the new suite (35 tests total).
- A six-agent QA review (security, performance, frontend, copy, two generalists) ran against the diff; its convergent findings drove the hardening commit, and reviewers independently verified the injection safety of the `{filters}` mechanism and the server-side `team_id` guard.
- `pnpm --filter=@posthog/frontend typescript:check` (no errors in touched files; remaining failures are pre-existing unbuilt-workspace issues in unrelated products), oxlint/oxfmt on touched files, and `hogli ci:preflight --fix` (0 failures).
- Not done: no manual browser verification beyond the Storybook captures above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The estimate line and tooltip are self-describing editor UI; user-facing survey docs live in posthog.com and can pick this up via the automatic docs update.

## 🤖 Agent context

Authored by Claude (PostHog Code) from a support-request investigation directed by a PostHog employee; I could not verify the driver's GitHub handle from this session, so the PR is left unassigned for them to pick up. Skills invoked: `/debugging-surveys`, `/adopting-generated-api-types`, `/writing-tests`, and `/qa-team`. Three iterations shaped this PR: the first commit extracted the estimate into a new keyed kea logic, which the driver flagged as over-built; the second commit deleted that logic and relocated the wizard's original `useState`/`useEffect` implementation into the shared component; the third commit hardens the estimate based on convergent findings from a six-agent QA review (RE2 regex validation, negative-operator skip, server-side query cancellation, plus smaller UX/a11y fixes). Two low-severity QA findings were deliberately left open as product decisions: the full editor stays silent (rather than warning) for exact-match-with-bare-path conditions, and the error state remains a passive message without a retry affordance.

---


darkopia:
CI triage for the failing Storybook / Visual Review gate ([run d783ff2e](https://us.posthog.com/project/2/visual_review/runs/d783ff2e-56e9-4106-b566-8ab3390b8ba0)): all code checks are green, and the 4 changed snapshots break down as follows. I (the agent) inspected all baseline/current image pairs but have not approved or finalized anything, since committing a new baseline is a human decision.

**Intended (from this PR), recommend approve:**
- `scenes-app-surveys--new-survey-targeting-section` (light + dark): the URL targeting field now shows "About 0 unique users viewed matching URLs in the last 30 days." and Audience filters gained an info tooltip icon. That is exactly this PR's change (page grows 36px, everything else identical).

**Pre-existing flake, unrelated to this PR:**
- `scenes-app-settings-authentication-domains--boost-needs-upgrade` (light + dark): the committed baseline is a mis-capture showing the project home screen instead of the settings page, while this run's capture renders the settings page correctly. The current image is hash-identical to what master's latest run produced (also flagged `changed` there), and the story's history flips between changed and unchanged across master runs since April. Worth tolerating or re-baselining on master rather than handling in this PR.

---

lucasheriques (approved):

PR #71804, opened 2026-07-17
fix(data-management): make virtual property definition pages load instead of 404

andyzzhao:
## Problem

The property definitions page lists virtual properties (Bot name, Traffic type, Initial channel type, Total MRR, ...), but clicking any of them lands on "Page not found". Reported in [this Slack thread](https://posthog.slack.com/archives/C09ALCCKESZ/p1784246592950579).

Two layers were broken:

1. Virtual property ids look like `$builtin_$virt_bot_name`, and kea-router's url-pattern matcher rejects `$` in path segments (its charset is `a-zA-Z0-9-_~ %`), so the detail route never even matched and users hit the app-level 404.
2. Even with routing fixed, the detail page fetches the definition from the API by id, and virtual properties have no database row (they're fabricated inside the list endpoint from `CORE_FILTER_DEFINITIONS_BY_GROUP`), so the fetch 404s.

## Changes

**Routing**

- `urls.propertyDefinition` / `urls.propertyDefinitionEdit` percent-encode the id (the `:id` route registration passes through raw), following the existing `urls.site` pattern. Encoding is a no-op for real UUID ids.
- Both definition scenes decode the id in `paramsToProps` and in the component. The component half matters because `App.tsx` passes raw route params to scene components while `BindLogic` gets the `paramsToProps` output; decoding in only one place produces two logic instances with different keys.
- The legacy `/events/properties/:id` redirect decodes before re-encoding to avoid double encoding.

**Virtual property detail page**

- New `getVirtualPropertyDefinition` in `taxonomy/helpers.ts` resolves `$builtin_*` ids into a read-only definition from the taxonomy (event, then person, then group properties, mirroring the backend's injection groups).
- `definitionLogic` short-circuits to it before calling the API, and bounces the `/edit` URL back to the view.
- `DefinitionView` hides delete/edit for virtual definitions, shows a "Source: Virtual" block explaining the property is computed at query time, resolves labels across taxonomy groups (so `$virt_initial_channel_type` titles as "Initial channel type"), and renders rich taxonomy descriptions (links work).
- The definition not-found page now says "Property not found" instead of "Event not found" on property routes.

**Screenshots**

The virtual property page for "Bot name" (`$virt_bot_name`), with the Source: Virtual panel and no edit/delete actions:

pr-virt-bot-name

A person-taxonomy virtual property ("Initial channel type") resolving its label and rich linked description:

pr-virt-initial-channel-type

Verified in the local dev app end to end (I drove it via Claude with a headless browser):

- Properties list -> click "Bot name" -> detail page renders label, description, type, and the Virtual tag, with no delete/edit buttons.
- `$builtin_$virt_initial_channel_type` (person taxonomy) resolves the right label and renders its linked description.
- Direct navigation to the `/edit` URL of a virtual property redirects to the read-only view.
- Real property and event definition pages are unchanged (delete/edit intact, API loading as before).

Automated tests added:

- `definitionLogic.test.ts`: a `$builtin_*` id resolves from the taxonomy without an API call. Catches removal of the loader short-circuit, which would regress the page to a 404 toast.
- `urls.test.ts`: encode round-trip for property definition ids, plus the `:id` passthrough. Catches someone "simplifying" the `encodeURIComponent` away, which would break routing for every virtual property link.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs changes needed. This fixes broken navigation rather than changing documented behavior.

## 🤖 Agent context

Authored with Claude Code. Skills invoked: /writing-kea-logics, /adopting-generated-api-types, /writing-tests.

Decisions along the way: I first assumed the 404 came from the API detail fetch and planned a frontend-only taxonomy short-circuit. Live verification showed the route itself never matched because of the `$` charset issue, which is why the encode/decode plumbing exists. I chose a frontend-only fix over teaching the backend `retrieve()` about `$builtin_*` ids since the REST 404 for a non-record is arguably correct, and nothing else consumes those ids by detail fetch today. The dual decode (paramsToProps plus component) copies the established `Site.tsx` pattern for ids that need encoding.

### DIFF
--- a/frontend/src/scenes/data-management/definition/DefinitionEdit.tsx
+++ b/frontend/src/scenes/data-management/definition/DefinitionEdit.tsx
@@ -46,13 +46,15 @@ import { getEventDefinitionIcon, getPropertyDefinitionIcon } from '../events/Def
 export const scene: SceneExport<DefinitionLogicProps> = {
     component: DefinitionEdit,
     logic: definitionLogic,
-    paramsToProps: ({ params: { id } }) => ({ id }),
+    paramsToProps: ({ params: { id } }) => ({ id: id ? decodeURIComponent(id) : id }),
 }
 
-export function DefinitionEdit(props: DefinitionLogicProps): JSX.Element {
+export function DefinitionEdit(rawProps: DefinitionLogicProps): JSX.Element {
+    // The app renders scene components with raw route params, so decode the id like paramsToProps does
+    const props = { ...rawProps, id: rawProps.id ? decodeURIComponent(rawProps.id) : rawProps.id }
     const logic = definitionEditLogic(props)
     const definitionLogicInstance = definitionLogic(props)
-    const { definitionLoading, definitionMissing, isProperty } = useValues(definitionLogicInstance)
+    const { definitionLoading, definitionMissing, isProperty, singular } = useValues(definitionLogicInstance)
     const { editDefinition } = useValues(logic)
     const { saveDefinition } = useActions(logic)
     const { tags, tagsLoading } = useValues(tagsModel)
@@ -80,7 +82,7 @@ export function DefinitionEdit(props: DefinitionLogicProps): JSX.Element {
     const mediaPreviewDragTarget = createRef<HTMLDivElement>()
 
     if (definitionMissing) {
-        return <NotFound object="event" />
+        return <NotFound object={singular} />
     }
     return (
         <Form logic={definitionEditLogic} props={props} formKey="editDefinition">
--- a/frontend/src/scenes/data-management/definition/DefinitionEdit.tsx
+++ b/frontend/src/scenes/data-management/definition/DefinitionEdit.tsx
@@ -26,7 +26,11 @@ import { lemonToast } from 'lib/lemon-ui/LemonToast/LemonToast'
 import { getPrimaryPropertyForEvent, hasTaxonomyPrimaryProperty } from 'lib/utils/events'
 import { eventUsageLogic } from 'lib/utils/eventUsageLogic'
 import { definitionEditLogic } from 'scenes/data-management/definition/definitionEditLogic'
-import { DefinitionLogicProps, definitionLogic } from 'scenes/data-management/definition/definitionLogic'
+import {
+    DefinitionLogicProps,
+    decodeDefinitionId,
+    definitionLogic,
+} from 'scenes/data-management/definition/definitionLogic'
 import { PropertyAccessControl } from 'scenes/data-management/definition/PropertyAccessControl'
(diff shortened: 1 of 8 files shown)


andyzzhao:
## QA team review

Ran the multi-agent QA review (six independent reviewers: two generalists, security, frontend, performance, copy) on this branch. Verdict: approve with nits, overall MEDIUM (three reviewers at MEDIUM, none higher). Security and performance came back clean; every reviewer independently confirmed the encode/decode mechanism is correct (kea-router returns raw undecoded segments, and scene components receive raw params while the bound logic gets the `paramsToProps` output, so both decodes are required and decode exactly once per path).

All findings are addressed in 771419eb792:

| Priority | Finding | Fix |
| --- | --- | --- |
| Medium (3 reviewers) | `$virt_mrr`/`$virt_revenue` exist in both person and group taxonomy under one shared `$builtin_*` id, and the label/description were resolved with a different group order than the definition itself, so the page could show "The current MRR for this group." on the person variant | `getVirtualPropertyDefinition` now returns the taxonomy group it matched, and the view plus breadcrumbs consume that instead of a second independent lookup, so all fields come from one entry |
| Medium (4 reviewers) | Five unguarded `decodeURIComponent` calls could throw `URIError` on malformed percent sequences (e.g. `/properties/100%25off`, which the route charset accepts), crashing the scene where the old code 404ed gracefully | New shared `decodeDefinitionId` helper wraps the existing `tryDecodeURIComponent`, falling back to the raw id which then 404s into the normal not-found state; used at all five sites |
| Low | Decode expression duplicated across four sites, risking logic-key drift and silent duplicate fetches | Same shared helper |
| Low | Delete dialog rendered a literal `$` ("recreated if the $property is ever seen") due to `${singular}` in JSX text (pre-existing, in a block this PR moved) | `{singular}` |
| Low | Empty description state said "Add a description for this property" on virtual definitions, which can't be edited | Virtual definitions show "No description" instead |
| Low | The `:id` sentinel in `urls.ts` only recognized that exact placeholder, so a future different param name would silently register a dead route | Sentinel broadened to any `:`-prefixed placeholder |
| Low | The edit-page bounce only fired on mount, so a future client-side push to `/edit` with the logic already mounted would reopen the editable form for a read-only virtual definition | Added a `urlToAction` guard on the edit route, id-matched to the logic instance, alongside the mount-time bounce |

Also added a regression test pinning `decodeDefinitionId` on malformed percent sequences. All 13 tests across the touched suites pass.

The one residual, deliberately not fixed here: `$builtin_$virt_mrr` is genuinely ambiguous between the person and group variant because the backend builds the synthetic id from the name alone; the page now consistently shows the person variant. Disambiguating fully needs the group baked into the id server-side, which is a backend contract change worth its own PR.

Gilbert09 (approved):

PR #71810, opened 2026-07-17
feat(retention): two-arm first-time scan on the legacy path

andyzzhao:
## Problem

PR #71787 restructured first-time retention into a two-arm scan on the DWH-capable variant builder, but the flag that routes events-only retention to that builder does not exist yet, so production events-only traffic still runs `build_base_query_legacy`. That is the path behind the most expensive retention queries we see in the query log today: in first-time modes the legacy scan has no timestamp bound (the first-occurrence anchor needs all history), and when the returning entity is "All events" it has no event-name prefilter either, so one merged scan reads every aggregate input column over the team's entire event history.

Only a small slice of those rows needs to be read that wide. The start entity's rows are the only ones the anchor needs all-time, and they need very few columns. Everything else is only consumed inside window-bounded aggregates.

## Changes

Inside `build_base_query_legacy`, when the split can shrink the scan, the single `FROM events` is replaced by a `UNION ALL` of two lean arms:

- the start arm scans all-time but only the start entity's event names, emitting just the actor, the timestamp, and role tags
- the return arm is scoped to the returning entity and bounded to the query window

The union is aliased back as `events` with tag columns (`is_start`, `matches_start`, `is_return`), so every aggregate expression in the outer query stays byte-identical and only the entity conditions swap to tag comparisons. A row matching both entities appears once per arm and contributes to exactly the aggregates it did before.

The rewrite applies only when it can pay for itself and stay provably equivalent:

- first-time or first-ever mode, concrete start event names, start and return entities differ (`_first_time_split_shrinks_scan`, shared with #71787)
- no value breakdown and no property aggregation, since those read event properties in the outer query and the tag arms don't carry them

Everything else emits exactly the same SQL as before (existing snapshots prove this mechanically). Recurring retention, breakdowns, property aggregation, cohorts, and the DWH-routed variant are untouched.

- New AST-shape tests (written first, verified failing): first-time and first-ever modes produce the two-arm union with an unbounded entity-scoped start arm and a bounded return arm; recurring, same-entity, all-events-target, breakdown, and property-aggregation shapes keep the single scan. These catch a regression where the gate silently widens or the arms lose their bounds, which no existing test pinned for the legacy path.
- Full retention suite passes in both events-schema modes (198 tests each), including the legacy-vs-variant comparison harness which asserts result parity on every test, so all existing first-time behavioral tests now exercise the new SQL end to end.
- Benchmarked the generated SQL against a production first-time retention query (concrete start event, all-events return, person and event property filters, test-account filters), interleaved runs with condition/uncompressed caches off and direct IO: about 231x fewer bytes read, 18x less CPU, 6x lower latency, 7x less memory, byte-identical result rows on both interleaved pairs.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, no user-facing behavior change.

## 🤖 Agent context

Built with Claude Code in the same session as #71787. Skills invoked: /writing-tests, /query-clickhouse-via-metabase (for the production benchmark), plus the superpowers TDD and debugging skills.

Decisions along the way:

- We considered routing the beneficial shapes to the variant builder instead (one-line routing change), but chose to optimize legacy in place: the variant's rollout flag doesn't exist yet, and swapping whales onto a whole different builder inherits every divergence the parity suite might still miss. This change alters one dimension of the live path (the scan structure) and keeps legacy's own aggregation and breakdown code.
- Design choice: raw row-stream arms with role tags plus an unchanged outer query, rather than the variant's per-arm pre-aggregation. This sidesteps the breakdown-authority bug class QA found in #71787 (arms disagreeing on a per-actor value before the outer collapse) because per-actor resolution still happens in one place.
- Breakdown and property-aggregation shapes are deliberately excluded rather than ported, keeping the equivalence argument airtight; they can follow if the query log shows expensive queries of that shape.

### DIFF
--- a/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
+++ b/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
@@ -630,7 +630,28 @@ class RetentionFixedIntervalBaseQueryBuilder(RetentionBaseQueryBuilder):
             source=ast.Field(chain=["events", "timestamp"])
         )
 
-        event_filters = self._event_filters()
+        two_arm = self._legacy_first_time_two_arm_scan()
+        if two_arm:
+            # Aliasing the UNION as `events` keeps every aggregate expression below identical;
+            # only the entity conditions swap to the arms' role tags.
+            select_from = ast.JoinExpr(
+                table=ast.SelectSetQuery.create_from_queries(
+                    [
+                        self._legacy_scan_arm(self.start_event, "start"),
+                        self._legacy_scan_arm(self.return_event, "return"),
+                    ],
+                    "UNION ALL",
+                ),
+                alias="events",
+            )
+            where: ast.Expr | None = None
+            start_condition: ast.Expr = parse_expr("events.matches_start = 1")
+            return_condition: ast.Expr = parse_expr("events.is_return = 1")
+        else:
+            select_from = ast.JoinExpr(table=ast.Field(chain=["events"]))
+            where = ast.And(exprs=self._event_filters())
+            start_condition = self.start_entity_expr
+            return_condition = self.return_entity_expr
 
         start_event_timestamps = parse_expr(
             """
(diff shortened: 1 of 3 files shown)


andyzzhao:
Superseded by #71817, rebuilt as a standalone PR on master.

PR #71811, opened 2026-07-17
docs(engineering-analytics): update the insights skill for warehouse views, reviews, and new tools

webjunkie:
## Problem

The `turning-engineering-analytics-into-insights` skill predates several things engineering analytics now exposes, so agents following it give stale answers:

- It claims CI **cost** cannot back an insight — but the cost model is now rendered into the `engineering_analytics_job_costs` warehouse view, so cost insights are plain SQL.
- It doesn't mention the other two exposed views (`engineering_analytics_ci_job_history`, `engineering_analytics_ci_failures` — the fingerprinted/grouped CI failure lines).
- It says reviews are "not in the data yet" — the `github_reviews` warehouse table now exists (webhook-fed).
- It names only the original MCP tools, missing `broken-tests`, `run-failure-logs`, `team-ci-health`, and the `investigating-ci-failures` sibling skill.

It also never states the design rule for the views: they exist **only** where product code (cost model, fingerprint recipe, commit attribution) must be rendered into SQL so it can't drift; for everything else, pure HogQL over the raw tables is always enough and no further views should be created or requested.

## Changes

- Rework the SKILL.md intro into the two-substrate model (raw tables + the three fixed-name views), with the views-from-code rule stated explicitly, and update the "what backs an insight" table (cost and grouped failure lines move to **yes**).
- Extend table discovery with `github_reviews`, the optional-endpoint probe, and the views' no-prefix / union-all-sources / gating behavior.
- Move cost out of "What NOT to rebuild in SQL" (point at the view instead) and refresh the product-logic list and MCP tool names.
- Add recipes: weekly CI cost by workflow (`job_costs` view), time to first review (`github_reviews`), distinct failures per day (`ci_failures` view), plus the `created_at_raw` scan-pruning floor for `ci_job_history`.
- Refresh the caveats list (reviews start at source-connect; NULL cost semantics; failure counts are absolute, never rates).

Docs-only change. Cross-checked every claim against the current code: `mcp/tools.yaml` (enabled tools and their descriptions), `backend/facade/warehouse_views.py` and `backend/logic/views/{job_costs,ci_job_history,ci_failures}.py` (view names, gating, column contracts, NULL semantics), SPEC.md §"Exposed warehouse views", and the GitHub source's `settings.py` (reviews endpoint: webhook-fed, `pr_number` injection, PENDING drop, no history backfill).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Fully autonomous

---

### DIFF
--- a/products/engineering_analytics/skills/turning-engineering-analytics-into-insights/SKILL.md
+++ b/products/engineering_analytics/skills/turning-engineering-analytics-into-insights/SKILL.md
@@ -4,40 +4,69 @@ description: >
   Converts engineering analytics (PR / CI) data into saved PostHog insights, dashboards, and subscriptions, and
   explains what data the product reads so it can be queried directly with SQL. The engineering analytics dashboard
   and MCP tools run curated HogQL privately over per-team GitHub warehouse tables; this skill teaches discovering
-  those tables via engineering-analytics-sources, replicating the curated column semantics in HogQL, saving the
-  query with insight-create, and scheduling delivery with subscriptions-create. Use when asked to "save this as an
-  insight", "put CI health / merge times on a dashboard", "email me PR throughput weekly", "subscribe to these
-  numbers", "alert on CI success rate", or "what data/tables does engineering analytics read". For ad-hoc CI and
-  merge questions use diagnosing-ci-and-merge-bottlenecks instead.
+  those tables via engineering-analytics-sources, replicating the curated column semantics in HogQL, reading the
+  exposed engineering_analytics_* warehouse views where product logic is involved (CI cost, fingerprinted failure
+  lines, commit attribution), saving the query with insight-create, and scheduling delivery with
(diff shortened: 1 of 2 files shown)


webjunkie:
Addressed two of Copilot's four findings in 4357ed3:

- **View scoping claim**: the skill no longer says each view "unions every connected source" — it now says the views cover all the team's sources at once and names the repo-filter columns per view, noting `engineering_analytics_ci_failures` reads the Logs product rather than the per-source tables.
- **Prefix substitution**: the recipes intro now lists every `github_*` table that needs the source prefix, including `github_workflow_jobs` and `github_team_members`.

The other two findings (re-adding "optional / may not be synced" caveats for `workflow_jobs` / `reviews` / `teams` tables and the probe guidance) are deliberately not applied: the PR author explicitly asked for those point-in-time availability hedges to be removed so the skill describes the data model timelessly — agents discover actual availability at query time. Leaving those threads for humans to resolve.

PR #71813, opened 2026-07-17
fix(mcp): support ID-JAG authorization metadata

dmarticus:
## Problem

> [!NOTE]
> The ID-JAG token is already valid and PostHog successfully authenticates it. This PR does not make the token valid. It fixes how MCP obtains the token's permissions when building the tool catalog.

MCP startup needs two separate answers:

1. **Authentication:** Who does this token represent?
2. **Authorization:** Which scopes, organizations, and tools may that user access?

Authentication already worked through `/api/users/@me/`. Authorization metadata did not.

The old authorization path assumed every credential was either a personal API key or a database-backed OAuth token:

```text
Valid ID-JAG token
  -> personal API key lookup: not found
  -> OAuth introspection: 403 because the JWT has no OAuth database row
  -> no scopes returned
  -> MCP cannot build the scoped tool catalog
  -> PostHog tools are missing
```

ID-JAG access tokens are signed, short-lived JWTs. Their scopes and organization restriction are encoded in the token itself rather than stored in the OAuth token table.

## Changes

For ID-JAG access tokens, MCP now:

```text
Valid ID-JAG token
  -> authenticate the token through PostHog
  -> read scope and organization claims from that validated token
  -> return authorization metadata without calling OAuth introspection
  -> pass the scopes into tool catalog filtering
  -> expose the PostHog tools allowed by those scopes
```

The important behavior change is the early authorization-metadata return. Calling `/api/users/@me/` alone would only verify the token and would not fix the missing tools.

The resolver is also named `getAuthorizationMetadata()` rather than `getApiKey()` so it is clear that it supports personal API keys, OAuth tokens, and ID-JAG tokens.

- Added StateManager regressions using the real ID-JAG access-token claim shape. They prove that validated ID-JAG scopes and organization restrictions are returned without personal-key lookup or OAuth introspection, and that claims are rejected when PostHog authentication fails.
- `pnpm --dir services/mcp test --run tests/unit/StateManager.test.ts` – 59 tests pass.
- Focused MCP unit suites covering authorization metadata and scope-based tool filtering – 149 tests pass.
- `pnpm --dir services/mcp typecheck`
- `pnpm --dir services/mcp lint`
- `pnpm --dir services/mcp format:check`

`hogli ci:preflight --fix` was unavailable in this checkout.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation change. This restores the existing ID-JAG authentication and authorization contract for MCP clients.

## 🤖 Agent context

Codex CLI traced the credential from agent ingress into MCP and inspected the authorization bootstrap path. It used the `/security-audit`, `/writing-tests`, and `/exploring-mcp-sessions` skills.

The key design decision is to keep PostHog as the authentication trust boundary. MCP only reads authorization claims after PostHog has accepted the same token, then uses those claims instead of the database-backed OAuth introspection path.

---

### DIFF
--- a/services/mcp/src/hono/request-state-resolver.ts
+++ b/services/mcp/src/hono/request-state-resolver.ts
@@ -112,9 +112,9 @@ export class RequestStateResolver {
         const flagAnalyticsContext = await reqCtx.safelyGetAnalyticsContext(context)
         const flagGroups = flagAnalyticsContext ? buildMCPAnalyticsGroups(flagAnalyticsContext) : undefined
 
-        const [allFlags, _apiKey, distinctId] = await Promise.all([
+        const [allFlags, authorizationMetadata, distinctId] = await Promise.all([
             this.resolveAllFlags(reqCtx, allFlagKeys, flagGroups),
-            context.stateManager.getApiKey(),
+            context.stateManager.getAuthorizationMetadata(),
             reqCtx.getDistinctId(),
         ])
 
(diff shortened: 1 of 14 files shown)


dmarticus:
this is an incorrect approach – closing

PR #71814, opened 2026-07-17
fix(quill-charts): keep pinned tooltip open while selecting text to copy

sampennington:
## Problem

Feedback on the new quill chart tooltip: you can pin it, but you can't copy text out of it — the tooltip dismisses on click.

Two root causes:

1. The pinned tooltip's outside-dismiss listener keyed off the document `click` event. A text-selection drag that starts inside the tooltip and releases outside it fires its `click` on the common ancestor of the two targets (usually `<body>`), which read as "outside", so the tooltip was dismissed and the selection wiped mid-copy. Separately, releasing a selection inside a clickable row fired `onRowClick`, opening the drill-down (e.g. persons modal) instead of letting the user copy.
2. The tooltip is portaled (`FloatingPortal`), and React portals bubble synthetic events through the **React tree**, not the DOM tree — so a click inside the tooltip still reached the chart wrapper's `onClick`, whose pinned branch is `clearTooltip()`. Any click inside the pinned tooltip dismissed it, even with fix 1 in place. The same bubbling let a selection drag inside the tooltip reach the wrapper's `onMouseDown` and activate drag-to-zoom, whose `onDragActivate` also clears the tooltip.

## Changes

- `useTooltipLifecycle`: decide "outside" at `pointerdown` — where the press *started* — instead of at `click` (same approach as floating-ui's `useDismiss`). A selection drag that starts inside the tooltip can now release anywhere without dismissing the pin. Escape and genuine outside presses still dismiss; pie charts share the same lifecycle hook, so they're covered too.
- `useChartInteraction`: guard the wrapper's `onClick` and drag-to-zoom `onMouseDown` against events originating inside `[data-hog-charts-tooltip]`, so portal-bubbled events from the pinned tooltip can't unpin it or start a zoom drag.
- `DefaultTooltip`: skip `onRowClick` when the click completes a text selection, so selecting a row's label or value no longer triggers drill-down.
- Updated the charts `AGENTS.md` consumer guide with the copy-friendly pinned-tooltip behavior.

- Added a lifecycle test simulating the exact gesture (pointer-down inside the tooltip, click landing outside → pin survives) — catches a revert to click-based outside-dismissal, which no existing test did.
- Added two `useChartInteraction` cases for the portal-bubbling path: a click originating inside the tooltip keeps the pin, and a press-drag starting inside the tooltip doesn't activate drag-to-zoom (which would clear the pin mid-copy). Existing tests called `handlers.onClick()` with no event; they now pass a real event since the handler inspects `e.target`.
- Added a `DefaultTooltip` case asserting `onRowClick` doesn't fire when the click completes a text selection — catches removal of the selection guard.
- Updated the existing "click outside dismisses" / "click inside wrapper doesn't dismiss" cases to the new pointer-down semantics.
- Ran the full `@posthog/quill-charts` jest suite and the `FunnelLineChart` product test that pins the tooltip and clicks rows — all pass. Not yet manually re-verified in the browser after the portal-bubbling fix.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `packages/quill/packages/charts/AGENTS.md` in this PR.

## 🤖 Agent context

Authored with PostHog Code (Claude Code) from user feedback that the pinned quill chart tooltip couldn't be copied from. Skills invoked: /writing-tests. Considered guarding the existing click listener with a "did the press start inside?" flag, but moving dismissal to `pointerdown` is simpler and matches the floating-ui convention. The portal-bubbling guard was found when manual testing showed clicks inside the tooltip still dismissed it — React portals propagate synthetic events through the React tree, so the wrapper's own handlers were a second, untested dismiss path.

**Why:** Users pin the tooltip specifically to interact with its content, and copying values/labels is the main reason to do so — dismissing on the selection gesture defeats the feature.

---


packages/quill/packages/charts/src/core/hooks/useChartInteraction.ts
```diff
@@ -23,6 +23,13 @@ import { useDragToZoom } from './useDragToZoom'
 import { useLatest } from './useLatest'
 import { useTooltipLifecycle } from './useTooltipLifecycle'
 
+/** The tooltip is portaled out of the wrapper's DOM tree, but React portals still bubble
+ *  synthetic events through the React tree — so a click or drag that starts inside the pinned
+ *  tooltip reaches the wrapper's handlers and would dismiss the pin or start a zoom drag. */
+function originatesInTooltip(e: React.SyntheticEvent): boolean {
+    return e.target instanceof Element && !!e.target.closest('[data-hog-charts-tooltip]')
```
sampennington, line 30: 🤖 **sp-ship** · ♻️📐 `[reuse+xp]` · **Suggestion / follow-up**

The "originates inside the tooltip" predicate now lives here as `originatesInTooltip` and is reimplemented inline in `useTooltipLifecycle.ts` (`handlePointerDownOutside`, and the scroll guard), and the raw `'[data-hog-charts-tooltip]'` marker is hardcoded across several production sites plus `Tooltip.tsx` where it's set. A `HOG_CHARTS_TOOLTIP_SELECTOR` constant already exists in `testing/tooltip.ts` (with a "kept in sync with Tooltip.tsx" comment) but can't be imported from production. Consider one small production module holding the marker constant + an `isInsideTooltip(target)` helper that all call sites route through — the two variants differ only in synthetic-vs-native `e.target`, which an `EventTarget | null` signature absorbs.

packages/quill/packages/charts/src/core/hooks/useTooltipLifecycle.ts
```diff
@@ -154,13 +154,17 @@ export function useTooltipLifecycle<Meta = unknown>({
...
+    // that starts inside the tooltip and releases outside it fires its click on the common
+    // ancestor of the two targets (usually <body>), and dismissing on that would clear the
+    // tooltip and wipe the selection mid-copy.
     useEffect(() => {
         if (!isPinned) {
             return
         }
 
-        const handleClickOutside = (e: MouseEvent): void => {
+        const handlePointerDownOutside = (e: PointerEvent): void => {
```
sampennington, line 167: 🤖 **sp-ship** · 💻📊 `[code+hog-charts]` · **Suggestion / follow-up**

Moving outside-dismiss from `click` to `pointerdown` shifts two edges beyond the copy-drag fix: `pointerdown` fires for right/middle-click and at the start of a touch-scroll (where `click` was effectively primary-button only), and a press that *starts* outside the chart now dismisses immediately even if the user drags back in before releasing. Likely acceptable/intended, but if only primary-button dismissal is wanted, gate on `e.button === 0` / `e.isPrimary`; either way worth a line in the doc note or a test pinning the intended behavior.

packages/quill/packages/charts/src/overlays/DefaultTooltip.tsx
```diff
@@ -161,7 +160,18 @@ export function DefaultTooltip<Meta = unknown>({
                             data-attr="hog-chart-tooltip-row"
                             data-closest={isClosest ? 'true' : undefined}
                             className={`flex items-center gap-2 min-w-0 py-0.5 px-1.5 rounded transition-colors duration-150${isClosest ? ' font-semibold bg-current/[.1]' : ''}${clickable}`}
-                            onClick={onRowClick ? () => onRowClick(s) : undefined}
+                            onClick={
+                                onRowClick
+                                    ? () => {
+                                          // A click that completes a text selection is a copy
+                                          // gesture, not a drill-down.
+                                          if (window.getSelection()?.toString()) {
```
sampennington, line 168: 🤖 **sp-ship** · 💻📊 `[code+hog-charts]` · **Suggestion / follow-up**

`window.getSelection()?.toString()` reads a document-global selection, so a lingering selection anywhere else on the page could silently suppress a legitimate row drill-down. It's mostly masked because a plain click's mousedown collapses existing selections before `click` fires, but a selection the click doesn't collapse would make the row appear dead. Consider scoping the check to a selection that actually intersects this row/tooltip (e.g. `getSelection().containsNode(rowEl, true)` or comparing `anchorNode`).

packages/quill/packages/charts/src/core/hooks/useChartInteraction.ts
```diff
@@ -295,74 +302,96 @@ export function useChartInteraction<Meta = unknown>({
...
-    const onClick = useCallback(() => {
-        // A click that closes out a drag-to-zoom gesture must not also pin/unpin or fire onPointClick.
-        if (shouldSwallowClick()) {
-            return
-        }
-        const currentIndex = hoverIndexRef.current
-        if (currentIndex < 0) {
-            return
-        }
+    const onClick = useCallback(
```
sampennington, line 305: 🤖 **sp-ship** · ✨📐 `[quality+xp]` · **Suggestion / follow-up**

`onClick` now carries four guard clauses plus a three-way pin / resolve-nearest / drill decision tree (~55 lines, 16-entry dep array) — a ComposedMethod candidate; extracting the pinnable-multi-series block into a named helper would make the body read like a table of contents. Flagging only in passing: most of the length is pre-existing code re-indented to slot in the new guard, so not worth churning now unless you're already in here.

sampennington (commented):

Solid, well-tested fix for a real bug — pinned tooltips staying put through a copy-selection drag. The pointer-event reasoning holds up (dismissal keys off where the press started, Escape/scroll/second-click paths all still work, and the portaled-DOM guard is correct), so no blockers. The main theme across reviewers is light cleanup: the `.closest('[data-hog-charts-tooltip]')` origin check is now duplicated across both hooks (and the marker string is hardcoded in several places), and the two behavioral edges of the `click`→`pointerdown` swap and the document-global `getSelection()` check are worth a second look. Details inline.

packages/quill/packages/charts/src/core/hooks/useChartInteraction.ts
sampennington: 🤖 **sp-triage**

Acknowledged — non-blocking suggestion (constant/helper reuse for the tooltip marker check). Follow-up if we revisit this area.

packages/quill/packages/charts/src/core/hooks/useTooltipLifecycle.ts
sampennington: 🤖 **sp-triage**

Acknowledged — non-blocking suggestion (pointerdown vs click edge-case behavior). Follow-up if we revisit this area.

packages/quill/packages/charts/src/overlays/DefaultTooltip.tsx
sampennington: 🤖 **sp-triage**

Acknowledged — non-blocking suggestion (scoping the selection check to the row/tooltip). Follow-up if we revisit this area.

packages/quill/packages/charts/src/core/hooks/useChartInteraction.ts
sampennington: 🤖 **sp-triage**

Acknowledged — non-blocking suggestion (extracting the pinnable-multi-series block for readability). Follow-up if we revisit this area.

PR #71816, opened 2026-07-17
fix(funnels): correct breakdown bar widths in top-to-bottom layout

sampennington:
## Problem

In a funnel with optional steps and "conversion rate relative to previous step", the top-to-bottom (horizontal bars) layout rendered breakdown bar widths that contradicted the percentages printed under each step: a step with a ~5% conversion could render a full-width bar. The left-to-right layout of the same insight rendered correctly.

**Why:** the breakdown path in the top-to-bottom chart recomputed each segment's basis from the immediately preceding step's count. The precomputed conversion rates treat the last *non-optional* step as "previous", so an optional step that out-counts its immediate (optional) predecessor produced fractions above 100%, clamped to a full bar.

## Changes

- `buildBreakdownSegments` now distributes the step's precomputed aggregate rate (`conversionRates.fromBasisStep`) across breakdown variants by count, so the stacked bar total always equals the rate shown in the step footer — for any step reference, with or without optional steps.
- Removed the now-unused `stepReference` build option (the precomputed rate already encodes it).

- Added a regression test where an optional step out-counts its immediate predecessor with a breakdown: previously the recomputed fraction exceeded 100% and clamped to a full bar; the test asserts segments are sized from the precomputed aggregate rate and the stack always closes the 0–100 track. No existing test covered a breakdown step whose neighbor-derived basis diverges from `fromBasisStep`.
- Removed two tests that existed only to exercise the deleted `stepReference` option.
- Ran the full `funnelBarHorizontalTransforms` Jest suite (32 passed) and the frontend TypeScript check (remaining errors are pre-existing and unrelated).
- Not able to verify visually in a running app in this environment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — rendering bug fix, no behavior/API/config change.

## 🤖 Agent context

Diagnosed from two screenshots of the same funnel in both layouts: bar widths in the top-to-bottom view matched count-over-immediately-preceding-step while the footers matched count-over-last-non-optional-step, which pointed at `buildBreakdownSegments` recomputing the basis via `getReferenceStep` instead of using the precomputed `conversionRates`. Considered threading the optional-steps list into the transform, but deriving segment widths from the already-correct `fromBasisStep` is simpler and keeps the bar consistent with the footer by construction. Skills invoked: /writing-tests.

---

### DIFF
--- a/products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.test.ts
+++ b/products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.test.ts
@@ -1,4 +1,4 @@
-import { EntityTypes, FunnelStepReference, type FunnelStepWithConversionMetrics } from '~/types'
+import { EntityTypes, type FunnelStepWithConversionMetrics } from '~/types'
 
 import {
     buildFunnelBarHorizontalCompareData,
@@ -43,7 +43,6 @@ function makeStep({
 }
 
 const options = {
-    stepReference: FunnelStepReference.total,
     getColor: () => '#1d4aff',
     getLabel: (variant: FunnelStepWithConversionMetrics) => String(variant.breakdown_value ?? variant.name),
     fillerColor: '#eef0f3',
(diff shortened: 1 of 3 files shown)


sampennington (commented):

Nothing blocks merge: the fix is arithmetically sound (distributing the precomputed rate by count share exactly reconstructs the footer percentage), the stepReference removal is complete, and the regression test verifiably fails against the old neighboring-step basis. Two small cleanups are suggested inline — reusing the existing `funnelConversionRate` guard and renaming a now-stale `describe` block.

products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.ts
```diff
@@ -66,13 +65,15 @@ function buildBreakdownSegments(
     options: BuildOptions
 ): Series<FunnelBarHorizontalSegmentMeta>[] {
     const step = steps[stepIndex]
-    const basisCount = getReferenceStep(steps, options.stepReference, stepIndex).count
+    // Split the precomputed rate across variants by count so the stack matches the footer.
+    // Deriving the basis from the neighboring step would ignore optional steps and overflow the bar.
+    const fractionPerCount = step.count > 0 ? step.conversionRates.fromBasisStep / step.count : 0
```
sampennington, line 70: 🤖 **sp-ship** · ♻️📐 `[reuse+xp]` · **Suggestion / follow-up**

This open-codes the guarded ratio that `funnelConversionRate(count, basisCount)` already provides — and it's already imported in this file: `const fractionPerCount = funnelConversionRate(step.conversionRates.fromBasisStep, step.count)`. Also worth knowing: the "stack matches the footer" property holds exactly only when `sum(variant.count) === step.count` (true for `aggregateBreakdownResult` today); if a top-N/"Other" bucketing ever breaks that invariant the proportions drift slightly, though the filler still closes the bar.

products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.test.ts
```diff
@@ -212,42 +211,41 @@ describe('buildFunnelBarHorizontalData', () => {
     })
 
     describe('reference step', () => {
```
sampennington, line 213: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

`describe('reference step')` is now a misnomer — this PR deletes the `stepReference` concept from these transforms. Consider renaming to what the block now covers, e.g. `describe('breakdown stack sizing with optional steps')`.

products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.ts
sampennington: 🤖 **sp-triage**

Agreed this is a valid follow-up, but tagged Suggestion/follow-up rather than Blocker, so leaving as a non-blocking cleanup rather than auto-fixing in this pass — deferring to a human/follow-up PR to swap in `funnelConversionRate(fromBasisStep, count)` here.

products/product_analytics/frontend/insights/funnels/FunnelBarHorizontalChart/funnelBarHorizontalTransforms.test.ts
sampennington: 🤖 **sp-triage**

Good catch, the describe block name is stale. Tagged Suggestion/follow-up rather than Blocker, so leaving as a non-blocking naming cleanup rather than auto-fixing in this pass.

PR #71817, opened 2026-07-17
feat(retention): two-arm first-time scan on the legacy path

andyzzhao:
## Problem

First-time and first-ever retention are the most expensive insight queries in the query log. In those modes the base scan has no timestamp bound, because the first-occurrence anchor has to look at all history, and when the returning entity is "All events" it has no event-name prefilter either. The result is one merged scan that reads every aggregate input column over the team's entire event history.

Only a small slice of those rows needs to be read that wide. The start entity's rows are the only ones the anchor needs all-time, and they need very few columns. Everything else is only consumed inside window-bounded aggregates. Bounding the whole scan with an OR doesn't help, since ClickHouse still reads every referenced column for every granule the unbounded branch can match.

## Changes

Inside `build_base_query_legacy`, when the split can shrink the scan, the single `FROM events` is replaced by a `UNION ALL` of two lean arms:

- the start arm scans all-time but only the start entity's event names, emitting just the actor, the timestamp, and role tags
- the return arm is scoped to the returning entity and bounded to the query window

The union is aliased back as `events` with tag columns (`is_start`, `matches_start`, `is_return`), so every aggregate expression in the outer query stays byte-identical and only the entity conditions swap to tag comparisons. A row matching both entities appears once per arm and contributes to exactly the aggregates it did before.

The rewrite applies only when it can pay for itself and stay provably equivalent:

- first-time or first-ever mode, concrete start event names, start and return entities differ
- no value breakdown and no property aggregation, since those read event properties in the outer query and the tag arms don't carry them

- New AST-shape tests (written first, verified failing): first-time and first-ever modes produce the two-arm union with an unbounded entity-scoped start arm and a bounded return arm; recurring, same-entity, all-events-target, breakdown, and property-aggregation shapes keep the single scan; group aggregation carries the `$group_N` filter into both arms. These catch a regression where the gate silently widens or the arms lose their bounds, which no existing test pinned.
- Full retention suite passes in both events-schema modes, so all existing first-time behavioral tests exercise the new SQL end to end.
- Benchmarked the generated SQL against a production first-time retention query (concrete start event with person and event property filters, all-events return with an event-name exclusion, test-account filters, 8 day intervals). Interleaved runs on production replicas with condition and uncompressed caches off and direct IO, results byte-identical on both pairs:

| metric | before (single scan) | after (two arms) | ratio |
| --- | --- | --- | --- |
| bytes read | 2.44 TiB | 10.79 GiB | 231x |
| CPU time | 165 s | 9.1 s | 18x |
| wall time | 4.8 s | 0.8 s | 6x |
| peak memory | 3.5 GiB | 0.5 GiB | 7x |

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, no user-facing behavior change.

## 🤖 Agent context

Built with Claude Code. Skills invoked: /writing-tests, /query-clickhouse-via-metabase (for the production benchmark), plus the superpowers TDD and debugging skills.

Decisions along the way:

- Design choice: raw row-stream arms with role tags plus an unchanged outer query, rather than per-arm pre-aggregation. Per-actor resolution (first-occurrence anchor, interval indexing) still happens in one place, which keeps the equivalence argument simple and avoids arms disagreeing on per-actor values.
- The first-ever anchor compares a no-props minimum against a with-props minimum, so the start arm filters on the props-stripped matcher and evaluates the full matcher into the `matches_start` column instead.
- Breakdown and property-aggregation shapes are deliberately excluded rather than ported, keeping the equivalence argument airtight; they can follow if the query log shows expensive queries of that shape.

### DIFF
--- a/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
+++ b/posthog/hogql_queries/insights/retention/retention_base_query_fixed.py
@@ -610,7 +610,28 @@ class RetentionFixedIntervalBaseQueryBuilder(RetentionBaseQueryBuilder):
             source=ast.Field(chain=["events", "timestamp"])
         )
 
-        event_filters = self._event_filters()
+        two_arm = self._legacy_first_time_two_arm_scan()
+        if two_arm:
+            # Aliasing the UNION as `events` keeps every aggregate expression below identical;
+            # only the entity conditions swap to the arms' role tags.
+            select_from = ast.JoinExpr(
+                table=ast.SelectSetQuery.create_from_queries(
+                    [
+                        self._legacy_scan_arm(self.start_event, "start"),
+                        self._legacy_scan_arm(self.return_event, "return"),
+                    ],
+                    "UNION ALL",
+                ),
+                alias="events",
+            )
+            where: ast.Expr | None = None
+            start_condition: ast.Expr = parse_expr("events.matches_start = 1")
+            return_condition: ast.Expr = parse_expr("events.is_return = 1")
+        else:
+            select_from = ast.JoinExpr(table=ast.Field(chain=["events"]))
+            where = ast.And(exprs=self._event_filters())
+            start_condition = self.start_entity_expr
+            return_condition = self.return_entity_expr
 
         start_event_timestamps = parse_expr(
             """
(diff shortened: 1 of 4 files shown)


thmsobrmlr (approved):
Makes sense. Quick note that I'm planning to retire the whole legacy query soonish - it should be replaced by the data warehouse compatible version once I've backported all features onto it. I'll make a not to use this improvement there as well, or let you do it in a follow up, if you like.

andyzzhao:
@thmsobrmlr thanks! I was just going to ask you about that code path; please go ahead and do the port to the new path whenever you like

PR #71605, opened 2026-07-16
feat(data-warehouse): migrate 4 sources to rest_source (batch 10)

Gilbert09:
## Problem

Tenth batch of hand-rolled source → `rest_source` migrations, stacked on #71598.

## Changes

**Four sources migrated** (behavior-preserving, suites green): `clockodo` (35 tests), `cloudflare` (33), `codefresh` (57), `coassemble` (46).

**Vetted and skipped**: `clickup`, `clockify`, `coda` (all two-level fan-out — this cluster is now the largest recurring gap and the next planned framework feature), `cloudbeds` (HTTP-200 body-level errors classified retryable-vs-permanent).

> [!NOTE]
> Stacked on #71598 → #71589 → #71580 → #71572 → #71565 → #71559 → #71552 → #71540 → #71524 → #71520 → #71481/#71477.

## How did you test this code?

- Merged verification: **311 tests pass** across the 4 migrated sources + the full `rest_source` suite.
- Each source's suite passed in isolation first; coverage preserved. Worktree isolation verified before merge. `hogli ci:preflight --fix` clean.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #71824, opened 2026-07-17
refactor(data-warehouse): migrate coda, aiven, asana, clickup, clockify to rest_source (batch 11)

Gilbert09:
## Problem

Batch 11 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

This batch also lands a generic framework capability — **multi-level chained dependent resources** — so fan-out sources whose child endpoints depend on a resource that is itself a dependent resource (e.g. `orgs → projects → tasks`) become expressible. An endpoint may also now declare multiple resolve params bound from the same parent row.

## Changes

Framework:
- Multi-level chained dependent resources in `common/rest_source` (a child's parent may itself be a dependent resource), plus multiple resolve params from a single parent row. With 2+ dependent resources in a chain, resume is disabled (retries re-fetch, merge dedupes).

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green and asserts the same observable behavior via the framework):
- **coda** — docs → tables → rows chained fan-out
- **aiven** — multi-resolve-param fan-out (project + invoice number from the same parent row)
- **asana** — workspaces → projects → tasks/sections, with parent filtering via `data_map`
- **clickup** — spaces → folders → lists chained fan-out
- **clockify** — workspaces → projects/users → child resources

Each reuses the shared helpers (`validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, framework `auth` for secret redaction).

Not migrated this batch (left hand-rolled, valid blockers): **circleci** (client-side dedup before hydration pinned to an exact request count, plus per-parent page caps with warning logs), **bugsnag** (interleaved multi-level resume that pins "never re-fetch a completed fan-out parent", which the chain-level no-resume tradeoff can't honor).

- Merged run of all 5 migrated dirs + `common/rest_source/tests/` — 376 passed.
- Smoke-run of previously-migrated fan-out/paginator consumers (adroll, bigmailer, chargebee) — 95 passed.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-10 PR (#71605); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/aiven/aiven.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/aiven/aiven.py
@@ -1,9 +1,5 @@
-from collections.abc import Iterator
+from collections.abc import Callable
 from typing import Any
-
-import requests
-from structlog.types import FilteringBoundLogger
-from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_exponential_jitter
 
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.pipeline.typings import SourceResponse
 from products.warehouse_sources.backend.temporal.data_imports.sources.aiven.settings import (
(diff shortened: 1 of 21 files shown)


Gilbert09:

Issue #19167, opened 2023-12-07
Notebook (and maybe generally?) comments


benjackwhite:
## Is your feature request related to a problem?

A feature request that has already come up internally is comments on Notebooks a la Google docs etc. 

This is one of those "iceberg" features that is simple on the face of it but has endless potential for scope and complexity creep so I wanted to outline what we could do to determine if it is worth the effort.

## Describe the solution you'd like

1. **Comment on Notebooks** It would be great if I can add a comment to any line of the document, including a "widget" that would somehow be indicated in the UI, opening the right hand side panel to show the comments when clicked
2. **Mentions** - Possibly underpinning this is the idea of mentions so that you can @mention someone and have a way for that person to be aware of all places they've been mentioned
3. **Replies** The next feature that would immediately be requested is replies - this would mean that any "comment" can get a follow up reply to make a thread like system
4. **Notifications** - Comments are only really useful if I would be made aware of them - this would be either via an @mention in the comment, or if it is my Notebook
  * We already have the activity panel - a somewhat okay solution would be to have some representation there for me to see all the comments I would be interested in
  * TRICKY - longer term people would want notifications which of course gets more complicated...
5. **Comments anywhere** - why keep this constrained to Notebooks? Could I attach comments to a feature flag? To an Insight? Massive scope creep...

It _feels_ like it would make a lot of sense to keep comments separate from Notebooks themselves.

```yaml
comment:
  - user: User
     timestamp: Date
     message: String
     notebook: String # ID of the notebook or "resource" to make it generic
     notebookIdentifier: String # Could be a nodeId or a line number
     deleted: boolean
```

daibhin:
Agreed on all the thinking here. Some additional thoughts as I read through:

### Inline comments
It's likely, possibly most likely, that people will want to leave comments on the inline text. Text won't have id's like our own nodes do. I'd imagine inline comments would be implemented as a mark (similar to bold or italic) that then references a comment object (as indicated above). That would make it super easy to reference / render the comment from the document but not as easy to do it the other way around. In the activity feed for example, rendering the text part of the document the comment relates to would be difficult. Perhaps we could store a `preview:string` on the comment model to solve for those cases. 

### Rich text editing
[...]

PR #71827, opened 2026-07-17
feat(comments): improve discussion threading and reply UX

luke-belton:
## Problem

Threaded replies in the discussions side panel are easy to miss. Replies get a small indent with no other visual grouping, so threads blend into neighboring discussions. Replying takes two clicks through the ⋯ menu, and the reply composer sits pinned at the bottom of the panel, visually disconnected from the thread being replied to.

Refs #19167

## Changes

- Each thread is one contained card: the root comment, an "N replies" bar, the replies, and the inline composer. The card border tints accent only for the thread being replied to; a deep-linked comment gets a subtle background wash and its thread auto-expands, even while a reply is open elsewhere.
- Threads are collapsible and collapsed by default. Clicking a thread card toggles it: opening shows the replies with the inline composer focused beneath them; closing puts the composer away and returns the footer composer. Each composer keeps its own draft (one per thread, plus the footer), so text never carries between threads and returning to a thread restores what you typed. The replies row also toggles via Enter/Space. While an anchored notebook comment is in progress, clicking a thread only expands it - the anchor survives.
- Threads show a small "Reply" button whenever the composer is closed: in the replies row while collapsed, and at the thread's bottom (where the composer opens) while expanded; it disappears while composing. After sending, the composer stays open and focused for follow-ups. There is no Cancel button - closing the thread is the cancel.
- While a reply composer is open, the footer swaps to a "New comment" button that closes the reply and brings back the top-level composer, focused - so starting a new thread never requires collapsing the current one first.
- Slimmer cards: the add-reaction button moved to the comment header and shows on hover, the bottom row renders only when it has content (reactions, recording link, edited marker), and reply cards use a smaller avatar.
- Reply mode ends when the reply target disappears (deleted or dropped on reload) and when the discussions surface closes, so reopening always starts fresh. Sends resolve against send-time state (a composer you switched to mid-request is never cleared), clicks bubbling out of popovers such as the emoji picker cannot toggle the thread, and "Start discussion" on LLM traces exits any reply first so the quote lands in a fresh footer composer.
- New `Scenes-App/Discussions` Storybook stories (threads + replying state) with mocked data, so the thread UI has visual regression coverage - previously no story rendered comments with data.

https://github.com/user-attachments/assets/7fb16321-1913-46db-985a-017b88c6dc80

(there's a bug in the recording where clicking the emoji expands the thread - I fixed that in cf63d5e2015be011909710c34e2dd6e9d69fd659)

- `commentsLogic.test.ts` (17 cases) covering the reply flow end to end: draft capture and per-thread draft isolation, reply/anchored-context mutual exclusion, focus on editor remount and after send, `source_comment` targeting the thread root, the reply/root scroll gate, stale-reply clearing, thread expansion pinning and collapse rules, deep-link reveal for replies and roots, the "New comment" focus handoff, the mid-flight composer-swap race, isEmpty syncing to a freshly registered composer, reveal-on-selection being scoped to deep links (clicking a comment's controls, e.g. its emoji reaction button, no longer expands the thread), startNewComment leaving an already-mounted footer editor registered, and empty sends leaving stashed drafts untouched. Each guards behavior introduced in this PR (the logic previously had zero tests).
- `Discussions.stories.tsx` so visual regression snapshots the collapsed thread list, the expanded rail, reply button, task tag, emoji reactions, and the inline reply composer.
- `pnpm --filter=@posthog/frontend typescript:check` - no new errors introduced (compared against a clean checkout).
- oxlint + oxfmt clean; kea typegen for `commentsLogic.ts` up to date.
- I (the agent) wasn't able to run the full app in this environment. Remaining manual QA: reply on a long thread (composer scrolled into view), a thread with a deleted root comment, and the logs drawer comments tab, which reuses the same composer.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs changes needed - UI-only change.

## 🤖 Agent context

- Built with PostHog Code (Claude), directed by Luke Belton. Assigned as DRI (handle verified via the PR author).
- Skills invoked: `/writing-tests` (gating the test additions), `/triaging-visual-review-runs` (verifying the new story snapshots).
- Design direction per Luke: distinctly PostHog rather than a Slack/Linear clone. The thread rail reuses the repo's own `border-l-2` nested-content idiom and the existing `border-accent` highlight pattern. A second iteration from his hands-on review added collapse-by-default, the per-thread reply button, hover-only reactions, and the singular active-thread rule.
- Key decisions: the footer composer hides itself while a reply is in progress, so the logs drawer (`LogComments.tsx`), which reuses it, needed no changes. A new `composerDraft` value in `commentsLogic` carries typed content across the composer remount. The old `focusComposer` subscription fired before the inline editor mounted, so it's replaced by a `setRichContentEditor` listener. A `commentsWithReplies` subscription clears stale reply state (Greptile P1). Also removed an unattached scroll ref in `CommentBottomRow` and gated `scrollToLastComment` to root comments (that scroll is currently a no-op anyway - its ref target has `display: contents` - left for a follow-up).

---


frontend/src/layout/navigation-3000/sidepanel/panels/discussion/SidePanelDiscussion.tsx
```diff
@@ -92,9 +93,11 @@ const DiscussionContent = ({
...
             </div>
 
-            <div className="border-t px-3 pb-3">
-                <CommentComposer {...logicProps} />
-            </div>
+            {!replyingCommentId ? (
+                <div className="border-t px-3 pb-3">
+                    <CommentComposer {...logicProps} />
+                </div>
+            ) : null}
```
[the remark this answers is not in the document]
luke-belton: Good catch — fixed in 62925bd4: a subscription on `commentsWithReplies` now clears `replyingCommentId` whenever its target thread stops rendering (root deleted with no replies, or dropped by a reload), so the footer composer returns with any typed draft preserved. Covered by a regression test in `commentsLogic.test.ts`.

yasen-posthog (commented):
Would you be able to upload a video or some screens of the new UX please?

luke-belton:
> Would you be able to upload a video or some screens of the new UX please?

yep of course - just making a few more tweaks!

frontend/src/scenes/comments/commentsLogic.ts
```diff
@@ -541,8 +616,18 @@ export const commentsLogic = kea<commentsLogicType>([
...
     })),
 
-    listeners(({ values, actions }) => ({
-        setReplyingComment: ({ commentId }) => {
+    listeners(({ values, actions, selectors, cache }) => ({
+        startNewComment: () => {
+            // Bring the footer composer back focused - it registers on its next mount
+            cache.focusEditorOnRegister = true
+            actions.setReplyingComment(null)
+            actions.clearRichContentEditor()
```
a-lider, line 635: Fabel flagged this while tracing the "Start discussion" flow, could you double-check?

> When the discussion tab is already open and no reply is active, `handleStartDiscussion` still calls `startNewComment()`, which deregisters the *mounted* footer editor. Nothing re-registers it (the composer never remounts - `key` doesn't change), so `insertQuoteIntoEditor` retries out and the footer composer can no longer send. An early return here when `replyingCommentId` is already null would cover it

frontend/src/scenes/comments/Comment.tsx
```diff
@@ -203,13 +200,18 @@ const CommentTopRow = ({ comment }: { comment: CommentType }): JSX.Element => {
                     </span>
                 ) : null}
 
+                <span className="opacity-0 transition-opacity group-hover:opacity-100 focus-within:opacity-100">
```
a-lider: nit: if you open the emoji picker and move the pointer off the comment, does the button fade out while the popover is still open? focus likely lands in the panel, so `focus-within` won't keep it visible

a-lider (approved):
Looks great, a few nits

PR #71829, opened 2026-07-17
refactor(data-warehouse): migrate airtable, cohere, configcat, convertkit, copper to rest_source (batch 12)

Gilbert09:
## Problem

Batch 12 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

## Changes

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green and asserts the same observable behavior via the framework, reusing `validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, and framework `auth` for secret redaction):

- **airtable** — bases → tables → records chained fan-out (now expressible via the multi-level fan-out capability landed in batch 11)
- **cohere**
- **configcat**
- **convertkit**
- **copper**

Not migrated this batch (left hand-rolled, valid blockers — all documented framework gaps):
- **coingecko** — HTTP 200 body carries a retryable rate-limit envelope (`{"status":{"error_code":429}}`); the framework retries on HTTP status only.
- **coin_api** — three transport modes in one source (bare-array reference, exchange-rate envelope flatten, forward-cursor timeseries) plus an envelope-sibling field injected into each row.
- **customerly** — reports auth failures as HTTP 500 with an auth-marker body, classified non-retryable vs retryable by body content; the framework retries on HTTP status only.

- Merged run of all 5 migrated dirs + `common/rest_source/tests/` — 338 passed.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-11 PR (#71824); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/airtable/airtable.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/airtable/airtable.py
@@ -1,35 +1,23 @@
-import time
-from collections.abc import Iterator
 from datetime import UTC, date, datetime
 from typing import Any, Optional
-from urllib.parse import quote, urlencode
-
-import requests
-from structlog.types import FilteringBoundLogger
-from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_exponential_jitter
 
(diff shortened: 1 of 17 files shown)


Gilbert09:

PR #71831, opened 2026-07-17
fix(insights): stop value labels dropping every other bar

sampennington:
## Problem

On bar charts, value labels were skipped on every other bar even when there was plenty of room to show them all — the exact complaint from a user in the linked thread. The label overlap check was too aggressive.

## Changes

`ValueLabels` collision avoidance ran two passes: a per-series pass that dropped a label using **only its categorical-axis position**, and then a precise 2D box-overlap pass on whatever survived. Because value labels sit at each bar's tip, two neighbours on bars of different heights have plenty of clearance along the value axis yet still share an x-projection — so the first pass discarded the second before the accurate 2D check ever saw it. That produced the "keep, skip, keep, skip" pattern.

This collapses the two passes into a single greedy pass over the full 2D label box (with `minGap` on every side), so a label is dropped **only** when it would actually overlap a higher-priority one. Ties resolve toward earlier series and left-to-right / top-to-bottom, matching the previous ordering.

Also adds two dense-label `BarChart` stories (vertical + horizontal) so the storybook snapshots cover a busy chart and catch overlap regressions.

Added regression tests in `ValueLabels.test.tsx`:
- vertical + horizontal: neighbours that overlap on the categorical axis but are separated along the value axis are **both kept** (these fail on the old 1D logic — the regression this PR fixes).
- vertical + horizontal: neighbours overlapping on **both** axes still drop one.

Ran the `ValueLabels` Jest suite in an isolated harness — all 33 tests pass (29 existing + 4 new). I (the agent) did not run the full frontend Jest job or view the rendered storybook; the new stories are for the maintainers' snapshot pipeline to pick up.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated from a Slack thread asking whether value-label conflict avoidance was too aggressive on bar charts. Traced it to the per-series 1D pre-pass in `applyCollisionAvoidance` (quill charts). Considered keeping the pass but making it 2D-aware; chose to remove it entirely since the existing `rectsOverlap` pass already checks both axes with `minGap` and the 1D pass was a strictly-more-aggressive superset of its drops. A secondary factor — the default `minGap` stacks on top of the label's built-in chrome — was left alone to keep the change focused; it can be tuned separately if labels still read as too tight.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ASU43UAMS/p1784264171802999?thread_ts=1784264171.802999&cid=C0ASU43UAMS)*


sampennington (commented):
🤖 sp-ship — reviewed, no blockers.

PR #71832, opened 2026-07-17
feat(subscriptions): reuse the SQL assistant's HogQL guidance in reports

vdekrijger:
## Problem

Scheduled AI reports can generate invalid or inefficient HogQL because planner and recovery guidance can drift from the ChatAgent SQL rules.

Reports also miss useful company and product context already stored in core memory.

## Changes

- Planning and recovery apply the authoritative ChatAgent HogQL rules.
- Portable query-writing learnings and subscription constraints remain separate code-owned constants.
- The ChatAgent rules win on conflicts.
- Fresh and frozen report flows include sanitized core memory when enabled.
- Subscription queries default to `LIMIT 250`.
- Recovery lowers result size only for memory, result-size, or timeout pressure.
- Planning and recovery keep separate task prompts while sharing one query-writing base.
- Version-5 frozen plans lazily regenerate under version 6.
- The rendered general ChatAgent SQL prompt remains unchanged.
- This PR changes no skill files and contains no code evaluation harness.

- The focused planner, recovery, and prompt suite passes with 161 tests.
- Targeted mypy, Ruff lint, Ruff format, and strict preflight pass.
- The rendered ChatAgent SQL prompt matches master byte-for-byte.
- The final diff contains only five production and regression-test files.
- Two active UI judges sample 10% of initial-query and recovery runs using the configured Production `gpt-5-mini` key, with results delivered through their configured Slack subscriptions.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None.

## 🤖 Agent context

Codex assisted with implementation and verification. `/qa-team` reviewed an earlier revision; later human feedback narrowed the scope and guided focused revalidation.

The shipped design keeps ChatAgent as the source of truth, removes skill integration, and moves ongoing quality monitoring to PostHog UI evaluations.


products/exports/backend/temporal/subscriptions/ai_subscription/test/test_spec_generator.py
```diff
@@ -879,6 +879,13 @@ def _stored_plan(self) -> dict:
             ).model_dump(),
         }
 
+    @patch(f"{_SG}.build_context_blob", return_value="blob")
+    def test_rejects_v5_plan_after_hogql_prompt_change(self, _mock_blob: MagicMock) -> None:
```
vdekrijger: Honestly this test feels very moot and implementation details, the function is already validated to work so lets just nuke this test.

products/exports/backend/temporal/subscriptions/ai_subscription/test/test_prompts.py
```diff
@@ -1,12 +1,53 @@
...
+    SQL_EXPRESSIONS_DOCS,
+    SQL_SUPPORTED_AGGREGATIONS_DOCS,
+    SQL_SUPPORTED_FUNCTIONS_DOCS,
+)
 
 _P = "products.exports.backend.temporal.subscriptions.ai_subscription.prompts"
 
 
+@parameterized.expand([("planner", PLAN_GENERATION_PROMPT), ("hogql_fix", HOGQL_FIX_PROMPT)])
+def test_prompt_embeds_shared_hogql_rules(_name: str, prompt: str) -> None:
```
vdekrijger: Honestly all these added tests are super moot and implementation detail, we lose a lot of flexibility. Lets just remove all these tests, as I don't think they are valuable.

products/exports/backend/temporal/subscriptions/ai_subscription/prompts.py
```diff
@@ -434,4 +470,10 @@ def render_prompt(template: str, substitutions: dict[str, str]) -> str:
 
 Original query (failed):
 {{{original_hogql}}}
-""".strip()
+""".strip(),
+    {
+        "hogql_casing_rules": HOGQL_FUNCTION_CASING_RULES,
+        "hogql_dialect_rules": HOGQL_DIALECT_RULES,
+        "hogql_window_function_naming_rules": HOGQL_WINDOW_FUNCTION_NAMING_RULES,
+    },
```
vdekrijger: 🤖 _AI-suggested · github-user's review style_

`HOGQL_WINDOW_FUNCTION_NAMING_RULES` repeats the `lagInFrame()`/`leadInFrame()` casing guidance already embedded in `HOGQL_DIALECT_RULES`, leaving two sources of truth for the same rule. Remove one copy so they can't drift.

**Fix:** Drop the `lagInFrame()`/`leadInFrame()` line from `HOGQL_WINDOW_FUNCTION_NAMING_RULES` (the shared dialect block already covers it) and keep only the lowercase window-function names, or move the whole block into the shared module next to the dialect rules.

products/exports/backend/temporal/subscriptions/ai_subscription/prompts.py
```diff
@@ -132,15 +145,19 @@ def render_prompt(template: str, substitutions: dict[str, str]) -> str:
   The pattern `WHERE event = (SELECT … FROM (WITH cte AS (…) SELECT …))` fails to parse. If you
   reach for a CTE, rewrite the whole query as one flat SELECT with conditional aggregation
   (`countIf`/`sumIf`/`uniqIf`) — see the reference patterns below.
-- Do NOT use window functions (`ROW_NUMBER() OVER`, `LAG`, `LEAD`, `RANK`). Use `argMax`/`argMin`
-  or `ORDER BY … LIMIT N` instead.
+- Window functions are supported (HogQL forms are in the dialect notes above); for a single winner per
```
vdekrijger: Do we really still need all these now that we share them from the main app? 

Also it feels like there is redundancy in for example the join section, that is something we can use from the chat_agent/sql_prompts.py. 

Generally I would prefer if we can reduce the size of this custom prompt and heavily rely upon our shared / core sql prompt instruction / skills! That way we have a single source of truth that gains all the benefits and allows ths to freely improve as well.

vdekrijger: Centralized casing, dialect, JOIN, syntax, and user-identity guidance in HOGQL_QUERY_WRITING_RULES. Planner and fixer defaults and managed overrides prepend it once; report-only window and output rules stay local.

products/exports/backend/temporal/subscriptions/ai_subscription/test/test_spec_generator.py
vdekrijger: Removed the explicit v5 prompt-change regression. Stale-plan behavior remains covered by the existing generic validation.

products/exports/backend/temporal/subscriptions/ai_subscription/prompts.py
vdekrijger: Moved all window-function naming guidance into the canonical shared casing block. The subscription prompt no longer maintains a second copy.

products/exports/backend/temporal/subscriptions/ai_subscription/test/test_prompts.py
vdekrijger: Removed the PR-added prompt-content tests. Verification now uses the existing behavioral suites, prompt construction checks, skill build and lint, and QA.

products/exports/backend/temporal/subscriptions/ai_subscription/prompts.py
vdekrijger: Follow-up: narrowed the scope further. The general SQL assistant and execute-SQL prompts now render byte-for-byte as master. Subscription planner and recovery prompts combine reusable general SQL fragments with the canonical query-writing skill rules, while report-only constraints stay local.

vdekrijger: Follow-up: removed the skill integration entirely. Planner and recovery now prepend the code-owned HOGQL_QUERY_WRITING_RULES exported by ChatAgent. Compatible report-only rules stay local and explicitly defer to ChatAgent on conflicts. Skill files and execute-SQL have no diff, and the rendered ChatAgent prompt still matches master byte-for-byte.

products/exports/backend/temporal/subscriptions/ai_subscription/spec_generator.py
```diff
@@ -577,7 +602,12 @@ def build_enriched_prompt(
 ) -> EnrichedPromptSpec:
     cleaned = sanitize_prompt(prompt)
     relevant_events = _select_relevant_events(team, user, cleaned, trace_correlation_id)
-    context_blob = build_context_blob(team, window, relevant_events=relevant_events)
+    context_blob = build_context_blob(
+        team,
+        window,
+        relevant_events=relevant_events,
+        core_memory_text=_load_core_memory_text(team, user),
```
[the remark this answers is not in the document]
vdekrijger: Keeping this as-is because the requested boundary is already enforced at the consuming prompts. Core memory is nested inside `<project_context>`, and `PLAN_GENERATION_PROMPT` explicitly says all content in that wrapper is user-generated data, must be treated as data rather than instructions, and must never override the system rules. `sanitize_core_memory_text` strips structural framing tags, so memory cannot escape that wrapper. The synthesis prompt has the equivalent untrusted-data rule, while the fixer receives no core-memory payload. Core memory may inform relevant factual selection by design, but it is also constrained to the supplied schema and the report objective. A second adjacent copy of the same instruction would be redundant rather than a stronger boundary.

PR #71830, opened 2026-07-17
refactor(data-warehouse): migrate confluence, crunchbase, deel to rest_source (batch 13)

Gilbert09:
## Problem

Batch 13 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

## Changes

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green and asserts the same observable behavior via the framework, reusing `validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, and framework `auth` for secret redaction):

- **confluence**
- **crunchbase** — POST-body (`param_location="json"`) search pagination
- **deel**

Not migrated this batch (left hand-rolled, valid blockers — all documented framework gaps):
- **convex** — a 400 body-error envelope must raise a specific non-retryable "resync" error, plus two incompatible cursor transport modes (snapshot string-cursor vs deltas int-cursor).
- **coupa** — bespoke OAuth2: scope-fallback token minting (400 `invalid_scope` body classification) and mid-sync 401 inline token re-mint.
- **cursor** — multiple transport modes with client-side date-window chunking and response-envelope field injection.
- **dixa** — dual transport: date-windowed conversation export with client-side window-walking and rate-limit sleeps, plus cursor-paginated lists.
- **delighted** — SSRF guards (disabled redirects, off-host link/resume URL rejection) pinned by tests.

## How did you test this code?

- Each migrated source's own `tests/` suite passes in isolation, asserting the same pagination progression, resume round-trips, termination, fail-loud paths, and `validate_credentials` mapping as before.
- Merged run of all 3 migrated dirs + `common/rest_source/tests/` — 276 passed.
- `hogli ci:preflight --fix` clean (0 failures).

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-12 PR (#71829); merge in order.

PR #71840, opened 2026-07-17
refactor(data-warehouse): migrate deepgram, docuseal, drip, elevenlabs to rest_source (batch 14)

Gilbert09:
## Problem

Batch 14 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

## Changes

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green and asserts the same observable behavior via the framework, reusing `validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, and framework `auth` for secret redaction):

- **deepgram**
- **docuseal**
- **drip**
- **elevenlabs**

Not migrated this batch (left hand-rolled, valid blockers — all documented framework gaps):
- **datadog** — retries HTTP 408 (framework retries 429/5xx only), two-secret-header auth, and SSRF host-pinning of cursor next/resume URLs.
- **dockerhub** — custom JWT login-exchange auth with mid-sync 401 re-auth, next-link host-pinning, and 200-body-shape retryable classification.
- **emailoctopus** — 2D per-list × per-status(query-param) fan-out with per-pair interleaved resume; resolve binds only to path.
- **everhour** — client-side dedup + ignored-offset early termination, plus an always-on computed from/to date window (including full refresh).

- Merged run of all 4 migrated dirs + `common/rest_source/tests/` — 326 passed.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-13 PR (#71830); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/deepgram/deepgram.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/deepgram/deepgram.py
@@ -1,16 +1,32 @@
 import dataclasses
-from collections.abc import Iterator
+from collections.abc import Callable
 from datetime import UTC, date, datetime
 from typing import Any, Optional
 from urllib.parse import urlsplit, urlunsplit
 
-import requests
-from structlog.types import FilteringBoundLogger
-from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_exponential_jitter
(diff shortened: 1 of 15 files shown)


Gilbert09:

PR #71843, opened 2026-07-17
feat(conversations): enrich cross-region orgs via trusted Slack attribution

abhischekt:
## Problem

PostHog's support desk runs on US Cloud but serves customers in both regions. The Conversations Slack enrichment (`ee/billing/salesforce_enrichment/conversations_signals.py`) only trusted a ticket's org attribution if it could re-derive a local `OrganizationMembership`. A customer whose members are registered in the sibling region can never satisfy that, so every EU-Cloud customer's Slack signals were silently dropped and their Salesforce fields froze. Measured in prod: 0 of 99 EU-Cloud orgs enriched since the feature launched.

## Approach

Trust the attribution instead of re-verifying it region-locally, and make the attribution itself trustworthy at the point it's created:

- **Enrichment (`conversations_signals.py`):** `_tickets_with_verified_org` now trusts a ticket whose `organization_id` is set and whose identity is platform-attested (`identity_verified=True`), dropping the local-membership requirement. `organization_id` is stamped once at ticket creation — from an authoritative membership lookup, or (for cross-region customers) from the analytics `$groups` signal their own product usage stamps into the support project, which is cross-region by construction.
- **Attribution hardening (`events.py`):** `_resolve_groups_from_analytics` now requires the org group to recur across at least `MIN_ANALYTICS_ATTRIBUTION_EVENTS` (3) of the customer's events — picking the most-frequent org, not merely the most recent — so a lone event carrying a spoofed group can't dictate attribution.

## Relationship to the person-lookup fix

For the specific customer that surfaced this, the ticket's `organization_id` is currently `NULL` because of a person-lookup bug fixed in #71821. This PR makes the enrichment *trust* that attribution once it's stamped; #71821 makes it get stamped. Both are needed to unfreeze that customer; this PR alone fixes every EU-Cloud customer whose attribution already resolves.

- `ee/billing/test/test_conversations_signals.py`: an identity-verified ticket with a resolved org but **no local membership** now joins the verified set (the cross-region fix), and a not-identity-verified ticket is excluded (the attestation gate). Rewrote the old membership-based cases to the new gate; retired two tests whose mechanism (the uuid-cast guard and enrichment-time membership re-verification) no longer exists.
- `products/conversations/backend/api/tests/test_events.py`: attribution requires the durability floor — an org backed by 2 events resolves to nothing, 3+ resolves (real ClickHouse-backed), plus a mocked below-floor case. Updated the existing mock rows to the new `(org, count, customer)` shape.
- All suites green locally; `ci:preflight` passes with 0 failures; the two changed production modules are mypy-clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal enrichment path; no user-facing docs affected.

## 🤖 Agent context

- Authored with PostHog Code (Claude). Skills invoked: `/writing-tests` (test-value gate).
- This PR pivoted mid-flight. It first implemented a signed cross-region membership probe (dedicated endpoint + secret + HMAC). On a "can we simplify this?" step-back, we realized the probe only re-verified an attribution the ticket already carried, so the whole apparatus was avoidable by trusting identity-verified attribution and hardening the analytics resolution at its source. The simpler version is what's here; the cross-region machinery was removed.
- Main trade discussed and chosen deliberately: relaxing the enrichment-time trust model (see the security callout above) in exchange for deleting a permanent cross-region coupling. Called out for a security opinion rather than decided silently.

---


abhischekt:
## Rollout impact analysis

Sized the two effects of this change against production (support project, trailing 90 days for tickets / 30 days for event activity). Summary: net positive, low risk. Figures below are aggregates; employee identities and internal org IDs are omitted since this repo is public.

### Data-model note (correcting an assumption)

The durability floor affects **both** the emitted `$conversation_ticket_created` `$groups` **and** `Ticket.organization_id` in Postgres — both are written from the same `_resolve_org_groups` result in `capture_ticket_created` (`products/conversations/backend/events.py:290-292`). It's one-shot at creation, so existing rows are untouched; only new tickets are affected.

### Current attribution (baseline)

Org attribution happens almost entirely on Slack and widget; email/teams/github rarely resolve an org:

| Channel | Tickets | Attributed | % |
|---|---|---|---|
| slack | 845 | 543 | 64% |
| widget | 151 | 103 | 68% |
| email | 1624 | 23 | 1.4% |
| teams | 55 | 2 | 3.6% |
| github | 16 | 0 | 0% |

Sync population: **151** distinct orgs via Slack, **195** via Slack+widget.

### Enrichment relaxation (trust `identity_verified`)

Strictly additive — the old rule required attestation **and** local membership; the new rule keeps attestation and drops the membership condition. No previously-synced ticket is lost, and it unblocks the cross-region (EU-Cloud) cohort (~99 orgs) that was getting zero sync.

### Attribution durability floor (`≥3` events): real-world drops = 0

Of **476** attributed Slack/widget filers: **446** have a durable signal (≥3 events), **4** fall in the sparse 1–2-event band the floor would newly reject, and **26** have zero org-tagged events in the current window (dormant — they fail the old `≥1` rule too, so not a floor effect).

All **4** sparse candidates are internal (employee) filers attributed to PostHog's own org, and all hold a local membership. Since the membership path runs **before** the analytics fallback in `_resolve_org_groups`, the floor is never reached for them — real attribution loss is **zero**. No customer and no membership-less filer is affected.

### Floor sensitivity

Drops among the 450 filers with any org signal, by threshold:

| Floor | Dropped |
|---|---|
| 2 | 1 |
| **3 (chosen)** | **4** |
| 4 | 6 |
| 5 | 6 |
| 6 | 8 |

The org signal is bimodal (a customer either barely emits it or emits it heavily), so the threshold sits in a sparse valley and the exact constant barely matters. Every filer in these bands is a membership-backed internal user.

### Most-recent → most-frequent org selection

**108 of 450** filers (24%) have multi-org event histories, and **26** (5.8%) would get a different attributed org under most-frequent vs most-recent. Mostly moot in practice: the affected population skews internal (attributed via membership regardless), and most-frequent is more robust for real customers (it resists a stray or spoofed single event). The only downside is a slight, self-correcting lag for a customer who genuinely switched orgs mid-window.

### Net

Additive on the enrichment side (unblocks ~99 EU-Cloud orgs, loses nothing), ~0 real attribution loss on the hardening side.

PR #71846, opened 2026-07-17
refactor(data-warehouse): SSRF host-pinning + migrate delighted, culture_amp, deno_deploy, e2b, firehydrant, fleetio (batch 15)

Gilbert09:
## Problem

Batch 15 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

This batch also lands a generic framework capability — **SSRF host-pinning + redirect control** — so sources that pin pagination/resume URLs to the expected host no longer need bespoke transport code.

## Changes

Framework:
- `client={"allowed_hosts": [...], "allow_redirects": False}` in `common/rest_source`. When `allowed_hosts` is set (even to `[]`, meaning base-host-only), every outgoing request URL — including paginator next-page links and seeded resume URLs — must resolve to an allowed host (the base_url host is always implicitly allowed); an off-host URL raises before the request (and its Authorization header) leaves the process. `allow_redirects=False` stops the session following redirects and rejects any 3xx. This closes the credential-exfiltration path a spoofed `next` link would otherwise open.

- **delighted** — now expressible via the new SSRF host-pinning (off-host link/resume URL rejection + disabled redirects)
- **culture_amp** — OAuth2 client-credentials with per-endpoint scopes + per-employee demographics fan-out
- **deno_deploy**
- **e2b**
- **firehydrant**
- **fleetio**

- **concord** — five transport modes in one source (single, recursive folders-tree, page, offset, and a date-windowed audit-log download).
- **easypost** — client-side incremental watermark filtering with early termination (stops paginating despite `has_more=true`).

- New framework capability: 7 parameterized cases in `test_ssrf_host_pinning.py` covering same-host pinning, explicit allowlists, off-host rejection, disabled enforcement, and redirect rejection/allow.
- Merged run of all 6 migrated dirs + `common/rest_source/tests/` — 462 passed. Additive framework change: all existing `rest_source` tests plus fan-out/paginator consumer smoke (adroll, bigmailer, chargebee, airtable, clickup) stay green.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-14 PR (#71840); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/__init__.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/__init__.py
@@ -268,6 +268,8 @@ def create_resources(
             paginator=create_paginator(client_config.get("paginator")),
             session=client_config.get("session"),
             max_retry_attempts=client_config.get("max_retries", DEFAULT_RETRY_ATTEMPTS),
+            allowed_hosts=client_config.get("allowed_hosts"),
+            allow_redirects=client_config.get("allow_redirects", True),
         )
 
         hooks = create_response_hooks(endpoint_config.get("response_actions"))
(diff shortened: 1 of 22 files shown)


Gilbert09:

PR #71847, opened 2026-07-17
fix(tasks): restore branch after fresh resume fallback

tatoalo:
## Problem

**Why:** A resumed cloud run should continue from the same branch. When snapshot restoration falls back to a fresh sandbox, the workflow clones the repository but skips branch checkout, leaving the run on the repository default branch.

## Changes

- Preserve the current branch when a resume snapshot is restored.
- Check out the requested branch when a resume uses a fresh sandbox instead.


charlesvien (approved):
nice

PR #71853, opened 2026-07-17
feat(data-warehouse): auto-sync new schemas on warehouse sources

estefaniarabadan:
## Problem

When schema discovery finds a new table on an existing warehouse source (a new Postgres table, a new Slack channel, ...), the row is created disabled and someone has to notice it and enable syncing by hand.
Teams whose upstream schemas grow regularly (for example anything named `raw_*`) want those tables to start syncing on their own.

## Changes

Adds a per-source opt-in, off by default, configured in the source's Configuration tab:

- `auto_sync_new_schemas` toggle and `auto_sync_schema_patterns` (fnmatch globs like `raw_*`, matched case-insensitively against the qualified and bare table name; empty means all new tables qualify) on `ExternalDataSource`.
- A single enablement choke point, `auto_enable_new_schemas`, called by both discovery paths: the 6h `discover-schemas` schedule and the manual "Pull new schemas" action. Only rows still in the untouched discovery state (disabled, no sync type) are considered, so revived rows keep prior config and retries are idempotent.
- Auto-enabled schemas reuse the one-shot setup defaults (`build_default_schemas`): incremental with a preferred tracking column, else append, else full refresh, never CDC. Webhook-only and default-off tables stay disabled. Each enabled schema gets a Temporal schedule and an immediate first sync, exactly like flipping the toggle in the UI.
- `refresh_schemas` now reports `auto_enabled` in its response and the toast ("2 added / 1 auto-enabled").
- Direct-query sources are excluded (schemas resolve at query time): the serializer rejects the toggle and the UI hides the section.

```mermaid
flowchart LR
    D{{Discovery: 6h schedule or Pull new schemas}} --> R[Reconcile rows]
    R --> N[New schema row: disabled]
    N --> H[User enables by hand]
    class D phBlue;
    class R,N,H phGray;
```

```mermaid
flowchart LR
    D{{Discovery: 6h schedule or Pull new schemas}} --> R[Reconcile rows]
    R --> N[New schema row: disabled]
    N --> A{Auto-sync on and name matches patterns?}
    A -- yes --> E[Smart defaults + schedule + first sync]
    A -- no --> H[User enables by hand]
    class D phBlue;
    class E phYellow;
    class R,N,A,H phGray;
```

> [!NOTE]
> No behavior change for existing sources: the field defaults to false and new rows keep today's disabled-by-default behavior until a source opts in.

Ran locally (no Postgres/ClickHouse in the sandbox, so DB-backed suites run in CI):

- `test_sync_new_schemas.py` (5 passed): activity wiring, catches the auto-enable call being dropped from the discovery activity.
- `TestSchemaNameMatchesAutoSyncPatterns` (13 parameterized cases): the matching contract (case-insensitivity, bare-tail matching of qualified names, `?` globs, empty patterns match all).
- `test_source_schema.py` (15 passed): untouched smart-defaults behavior, sanity that reuse didn't change it.
- Serializer smoke test without DB: both fields accepted on PATCH, null patterns allowed, blank entries rejected.
- `ruff`, `oxlint`/`oxfmt`, `tach check`, `makemigrations --check` (no drift), full `hogli build:openapi` regen; `tsgo` reports no errors in touched files.

Added for CI (DB-backed):

- `TestAutoEnableNewSchemas`: enablement end state (sync type, incremental config, primary keys, schedule call), opt-out and direct-query no-ops, revived-row skip, webhook-only/default-off skip, one schema's Temporal failure not blocking the rest, seeded `schema_metadata` surviving the config merge. No existing test covered any of these paths.
- API tests: `refresh_schemas` auto-enables matching schemas and reports `auto_enabled`, PATCH persists the new fields, direct-query PATCH rejected with 400.

Not manually tested end-to-end against a live source; the manual flow to verify after merge is: enable the toggle with a pattern on a Postgres source, create a matching table upstream, click "Pull new schemas", and confirm the toast plus an incremental sync starting.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (PostHog Code cloud task) from Estefania's request to explore and build auto-syncing of newly discovered schemas with "name starts with" matching.
Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/adopting-generated-api-types`, `/writing-tests`.
Key decisions: glob patterns over regex (covers "starts with" via `raw_*`), a single shared helper next to `sync_old_schemas_with_new_schemas` rather than changing that function's return contract (keeps GitHub and direct-query callers untouched), candidates filtered to disabled-and-untyped rows for idempotency, schedule creation kept outside the `refresh_schemas` atomic block so Temporal calls never run under the source row lock, and no feature flag since the off-by-default field is the gate.
External data sources use handwritten frontend types, so the frontend follows that existing pattern; the regenerated OpenAPI/MCP artifacts are included.

---


Gilbert09 (approved):

estefaniarabadan:
Thanks @greptile-apps — both findings in `auto_enable_new_schemas` are addressed in the latest commit:

**P1 (enabled schema left unscheduled):** if `sync_external_data_job_workflow(...)` fails after we've already persisted `should_sync=True` + a sync type, the row is now rolled back to the untouched discovery state (`should_sync=False`, `sync_type=None`, original `sync_type_config` restored). Because this helper only reconsiders disabled, untyped rows, the rollback is what makes the next discovery pass retry it — so a Temporal hiccup no longer strands a row as "enabled but never syncing". This also makes the docstring's "retries are idempotent" promise actually hold. Covered by `test_schedule_failure_rollback_lets_next_pass_retry`, and `test_schedule_failure_does_not_block_remaining_schemas` now asserts the rolled-back state.

**P2 (concurrent refresh → duplicate first sync):** each row is now claimed under a `select_for_update` lock inside a short transaction before it's enabled, re-reading and bailing if it's no longer the untouched discovery row. A second concurrent discovery pass (another "Pull new schemas" click, or the 6h schedule firing alongside a manual refresh) sees `should_sync=True` and skips, so only one pass creates the schedule and fires the first sync. The lock is released before the Temporal call, so scheduling still happens outside any row lock.

PR #71857, opened 2026-07-17
fix(personhog): fail team/person deletion fast when personhog unconfigured

hpouillot:
## Problem

Error tracking issue: [personhog client not configured](https://us.posthog.com/project/2/error_tracking/019f28dd-b7d7-7801-a561-094cb5620784)

A Temporal worker without `PERSONHOG_ADDR` set raises `RuntimeError("personhog client not configured")` from `require_personhog_client()`. The team- and person-deletion activities run under an infinite retry policy (`maximum_attempts=0`) intended for idempotent bulky deletes that should retry through transient DB blips.

A missing `PERSONHOG_ADDR` is not transient though. It cannot change without a redeploy, so the activity retried the same failure forever. Each attempt is captured by the Temporal exception interceptor with a fresh distinct id, which is why this single deterministic misconfiguration shows up in error tracking as a flood of events rather than one clear signal.

The underlying fix is operational: the general-purpose Temporal worker running these deletion workflows needs `PERSONHOG_ADDR`. This PR makes the code fail fast and loud when that config is missing, so the problem surfaces cleanly instead of looping.

## Changes

- Add `PersonHogNotConfiguredError` in `posthog/personhog_client/client.py`, a `RuntimeError` subclass raised by `require_personhog_client()` (and the manual `get_personhog_client() is None` checks in the delete paths). Being a `RuntimeError` subclass, every existing `except RuntimeError` handler and fallback keeps working unchanged.
- Mark `PersonHogNotConfiguredError` non-retryable in the `delete-teams` and `delete-persons` retry policies. A permanent misconfiguration now fails the activity fast and surfaces, rather than retrying indefinitely.

A genuinely transient personhog outage is unaffected: that surfaces as a gRPC error from inside the RPC call, not as "not configured", and keeps retrying as before.

> [!NOTE]
> This changes deletion behavior in a misconfigured environment from "retry forever" to "fail fast". Once `PERSONHOG_ADDR` is set, the deletion is re-triggered and proceeds normally.

Added `posthog/personhog_client/test_require_client.py` (no-DB `SimpleTestCase`):

- `require_personhog_client()` raises `PersonHogNotConfiguredError` when the client is unconfigured.
- the error is a `RuntimeError` subclass, so existing handlers still catch it.
- the class name matches the string in `DELETE_RETRY_POLICY.non_retryable_error_types` — Temporal matches `non_retryable_error_types` on the exception class name, so a rename that desyncs the two would silently bring the infinite-retry flood back. This test guards that coupling.

I (Claude) ran the new tests and the existing `test_group_type_mapping.py` fallback test (which patches an unconfigured client) — both pass. The broader deletion suites need a Postgres test DB that isn't available in this sandbox, so I did not run them here.

## Docs update

- [ ] No docs change needed.

## 🤖 Agent context

Investigated from an error-tracking issue (`personhog client not configured`) whose stack trace pointed at `delete_misc_small_tables_activity`. Traced the raise to `require_personhog_client()` and found the delete-teams and delete-persons activities retry it forever under `maximum_attempts=0`, which matches the retry policy's own documented intent to make deterministic failures non-retryable (it already lists `ProtectedError` and `RecursionError`).

Considered an ORM fallback but rejected it: persons/groups/hash-key-overrides live in the persons DB and personhog is the sole path by design, so there's no ORM route to fall back to. The typed-error + non-retryable approach fixes the observable harm (retry flood) and surfaces the real cause without violating that boundary. Changing `retry_policy` on existing `execute_activity` calls is safe for in-flight workflows — it's an activity-schedule attribute, not a workflow command-sequence change.

---

### DIFF
--- a/posthog/personhog_client/client.py
+++ b/posthog/personhog_client/client.py
@@ -92,6 +92,17 @@ from posthog.personhog_client.proto import (
 
 logger = structlog.get_logger(__name__)
 
+
+class PersonHogNotConfiguredError(RuntimeError):
+    """Raised when the personhog client is required but ``PERSONHOG_ADDR`` is unset.
+
+    This is a permanent misconfiguration, not a transient outage: it can only change with a
+    redeploy, so callers that retry (e.g. Temporal deletion activities) treat this type as
+    non-retryable to fail fast instead of looping forever. Subclasses ``RuntimeError`` to keep
+    existing ``except RuntimeError`` handlers working.
+    """
+
+
 # -- Channel-level metrics --
 
 PERSONHOG_DJANGO_CHANNEL_STATE = Enum(
@@ -416,7 +427,7 @@ def get_personhog_client() -> Optional[PersonHogClient]:
 def require_personhog_client() -> PersonHogClient:
     client = get_personhog_client()
     if client is None:
-        raise RuntimeError("personhog client not configured")
+        raise PersonHogNotConfiguredError("personhog client not configured")
     return client
 
 
--- a/posthog/personhog_client/test_require_client.py
+++ b/posthog/personhog_client/test_require_client.py
@@ -1,0 +1,24 @@
+from unittest.mock import patch
+
+from django.test import SimpleTestCase
+
+from posthog.personhog_client.client import PersonHogNotConfiguredError, require_personhog_client
+
+
+class TestRequirePersonHogClient(SimpleTestCase):
+    def test_raises_typed_error_when_not_configured(self) -> None:
+        with patch("posthog.personhog_client.client.get_personhog_client", return_value=None):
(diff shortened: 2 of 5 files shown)


hpouillot:
Infra root-cause fix (the actual reason the error fired): the dev general-purpose temporal worker was missing `PERSONHOG_ADDR` — prod-us/prod-eu and the dev Django workers all set it, only the dev temporal-worker shared config didn't. Fixed in PostHog/charts#13260.

This PR is the defensive complement: once the env var lands, the flood stops; if any environment loses personhog config again, these deletion activities now fail fast instead of retrying forever and spamming error tracking.

hpouillot:
fixed in dev

PR #71859, opened 2026-07-17
fix(tasks): only fire PR babysitting follow-up when the PR is actionable

VojtechBartos:
## Problem

The PR babysitting loop (CI follow-up) wakes the agent on any PR fingerprint change - including the first poll after the PR opens and pending-to-passing transitions. Each pointless wake-up posts a full "everything is green, nothing to fix" summary into the originating Slack thread and burns one of the 3 CI repetitions. Meanwhile review comments were never a wake-up signal at all, so bot findings and human inline feedback were only addressed if a CI transition happened to coincide. Context: [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1784233097090179).

## Changes

- Dispatch a CI follow-up only when the changed PR state is actionable: failing CI or a changes-requested review. Green, pending, or check-less changes record the fingerprint silently.
- Key the fingerprint on the head SHA too (found by Greptile): without it, a new commit failing the same way as its predecessor hashed identically and the follow-up never re-fired after the agent's own failed fix attempt.
- Wake the agent when new unresolved review threads appear (suggested by Georgiy): the loop tracks the last-seen thread count and fires on increase, so resolving threads or the agent's own replies can't re-trigger it.
- Applied to both `process_task` and `task_management` via a shared `is_pr_actionable` predicate, gated with `workflow.patched` for replay determinism (gate lives in a shared `temporal/patches.py`).

Parameterized decision tests for both workflow copies (fire on failing/changes-requested/new feedback, quiet skip on green/pending/thread-resolve) plus a fingerprint case pinning the repeated-failure scenario - catches a regression back to fire-on-any-change, SHA-less hashing, or comment-blind polling, which no existing test did. Ran the affected suites locally: 83 passed. The Temporal time-skipping e2e tests run in CI only.

## 🤖 Agent context

Authored with Claude Code, /writing-tests invoked. Earlier revisions tried a NO_UPDATES reply sentinel matched at the Slack relay (removed as fragile prose matching) and a pending-defer that kept the old fingerprint while CI ran (removed - it caused the repeated-failure suppression Greptile flagged; the head SHA in the fingerprint makes it unnecessary). A determinism review against the temporalio SDK source confirmed the patch gate replays pre-rollout histories safely; the review-thread signal rides the same gate.

### DIFF
--- a/products/tasks/backend/temporal/process_task/activities/get_pr_context.py
+++ b/products/tasks/backend/temporal/process_task/activities/get_pr_context.py
@@ -32,14 +32,22 @@ class GetPrContextOutput:
     changes_requested: bool = False
 
 
-def is_pr_actionable(ci_status: str, changes_requested: bool) -> bool:
-    """Whether the PR state gives the agent real work to do.
+def decide_ci_follow_up(pr: GetPrContextOutput) -> tuple[bool, str | None]:
+    """Decide whether a changed PR snapshot warrants waking the agent.
 
-    A green or check-less PR is not actionable: waking the agent for it produces a
-    "nothing to report" turn that spams the originating Slack thread and burns one
-    of the limited CI follow-up repetitions.
+    Returns ``(fire, fingerprint_to_store)``; a ``None`` fingerprint means the
+    caller keeps its previous one. Only an actionable state — failing CI or a
+    changes-requested review — fires: waking the agent for a green or check-less
+    PR produces a "nothing to report" turn that spams the originating Slack
+    thread and burns one of the limited CI follow-up repetitions. Pending CI
+    keeps the old fingerprint so the settled state (which may be failing) still
+    registers as a change on the next tick.
     """
-    return changes_requested or ci_status == "failing"
+    if pr.changes_requested or pr.ci_status == "failing":
+        return True, pr.fingerprint
+    if pr.ci_status == "pending":
+        return False, None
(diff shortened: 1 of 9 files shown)


products/tasks/backend/temporal/process_task/activities/get_pr_context.py
```diff
@@ -27,6 +27,22 @@ class GetPrContextOutput:
...
+    """Whether a changed PR snapshot warrants waking the agent.
+
+    Only failing CI or a changes-requested review gives the agent real work to
+    do. Waking it for a green, pending, or check-less PR produces a "nothing to
+    report" turn that spams the originating Slack thread and burns one of the
+    limited CI follow-up repetitions. Pending needs no special handling: the
+    head SHA and CI status are both in the fingerprint, so the settled state
+    (which may be failing) registers as its own change on a later tick.
+    """
+    return pr.changes_requested or pr.ci_status == "failing"
```
skoob13, line 46: What about comments (bots/humans)? I think it's also a very good signal.

skoob13 (approved):
We should probably consider comments as the actionability signal too.

products/tasks/backend/temporal/process_task/activities/get_pr_context.py
VojtechBartos: Good point, adding

PR #71860, opened 2026-07-17
feat(llm-gateway): add modal backend for glm with weighted routing

tatoalo:
## Problem

GLM 5.2 (`@cf/zai-org/glm-5.2`)  is currently served exclusively through Cloudflare Workers AI. We want to move that traffic to a Modal-hosted vLLM endpoint, gradually and reversibly, without changing the model id any client uses.

## Changes

- add a new Modal provider with OpenAI-compatible routing, Modal proxy authentication, and model ID mapping for vLLM-hosted GLM models.

- centralize GLM backend routing across the Anthropic, Chat Completions, and Responses APIs. Routing is controlled by a feature flag or a deterministic traffic fraction, both disabled by default, with flag failures falling back to fraction-based routing.

- avoid cross-backend retries so Modal and Cloudflare health remain independently visible. Rollback is handled by reducing the traffic fraction.

- align Modal pricing and metrics with the existing Cloudflare GLM model, keeping user billing unchanged while labeling metrics by provider.

- continue advertising GLM in `/v1/models` whenever either backend is available.

- add `glm_modal_smoke.py` to validate the Modal integration end-to-end.

### DIFF
--- a/services/llm-gateway/src/llm_gateway/flags.py
+++ b/services/llm-gateway/src/llm_gateway/flags.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+
+import asyncio
+
+import structlog
+from cachetools import TTLCache
+from posthoganalytics import Posthog
+
+from llm_gateway.config import get_settings
+
+logger = structlog.get_logger(__name__)
+
+GLM_MODAL_FLAG = "tasks-glm-modal-inference"
+
+_flag_cache: TTLCache[tuple[str, str], bool] = TTLCache(maxsize=10_000, ttl=60)
+# Global per-flag backoff so an evaluation outage doesn't stack one blocking roundtrip per new user.
+_flag_unavailable_cache: TTLCache[str, bool] = TTLCache(maxsize=100, ttl=5)
+_client: Posthog | None = None
+
+
+def _get_client() -> Posthog | None:
+    global _client
+    settings = get_settings()
+    if not settings.posthog_project_token:
+        return None
+    if _client is None:
+        _client = Posthog(
+            settings.posthog_project_token,
+            host=settings.posthog_host,
+            sync_mode=True,
+            enable_local_evaluation=False,
+            feature_flags_request_timeout_seconds=2,
+        )
+    return _client
+
+
+async def evaluate_flag(flag_key: str, distinct_id: str) -> bool | None:
+    """None when evaluation is unavailable — callers fall back to their own default."""
+    cache_key = (flag_key, distinct_id)
+    if cache_key in _flag_cache:
+        return _flag_cache[cache_key]
+    if flag_key in _flag_unavailable_cache:
+        return None
+    client = _get_client()
+    if client is None:
+        return None
+    try:
+        # Blocking /flags roundtrip — keep it off the event loop. send_feature_flag_events=False:
+        # with sync_mode the default would also block on a $feature_flag_called capture upload
+        # (15s SDK timeout) for every uncached user.
+        enabled = await asyncio.to_thread(client.feature_enabled, flag_key, distinct_id, send_feature_flag_events=False)
+    except Exception as exc:
+        logger.warning("flag_evaluation_failed", flag=flag_key, error=str(exc))
+        _flag_unavailable_cache[flag_key] = True
+        return None
+    if enabled is None:
+        _flag_unavailable_cache[flag_key] = True
+        return None
+    _flag_cache[cache_key] = bool(enabled)
+    return bool(enabled)
--- a/services/llm-gateway/src/llm_gateway/glm_routing.py
+++ b/services/llm-gateway/src/llm_gateway/glm_routing.py
@@ -1,0 +1,187 @@
+"""Backend selection for GLM (`@cf/...`) traffic during the Cloudflare -> Modal migration.
+
+Modal takes traffic opted in by the server-side `tasks-glm-modal-inference` flag or by the
+env-configured fraction (caller-forwarded flag headers are deliberately ignored — they must not
+force a backend operators turned off). No cross-backend retries — rollback is turning the
+flag/fraction back down.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from fastapi.responses import StreamingResponse
+
+from llm_gateway.api.handler import (
+    CLOUDFLARE_ANTHROPIC_CONFIG,
+    CLOUDFLARE_OPENAI_CONFIG,
+    CLOUDFLARE_OPENAI_RESPONSES_CONFIG,
+    MODAL_ANTHROPIC_CONFIG,
+    MODAL_OPENAI_CONFIG,
+    MODAL_OPENAI_RESPONSES_CONFIG,
+    ProviderConfig,
+    handle_llm_request,
+)
+from llm_gateway.auth.models import AuthenticatedUser
+from llm_gateway.cloudflare import (
+    ensure_cloudflare_configured,
+    ensure_cloudflare_model_allowed,
+    is_cloudflare_configured,
+    make_cloudflare_anthropic_call,
+    make_cloudflare_completion_call,
+    make_cloudflare_responses_call,
+)
+from llm_gateway.config import Settings, get_settings
+from llm_gateway.flags import GLM_MODAL_FLAG, evaluate_flag
+from llm_gateway.modal import (
+    ensure_modal_configured,
+    ensure_modal_model_allowed,
+    is_modal_configured,
+    is_modal_served_model,
+    make_modal_anthropic_call,
+    make_modal_completion_call,
+    make_modal_responses_call,
+    should_route_glm_to_modal,
+)
+
+LlmCall = Callable[..., Awaitable[Any]]
+
+
+async def _route_to_modal(model: str, user: AuthenticatedUser, product: str, settings: Settings) -> bool:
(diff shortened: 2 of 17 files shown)


services/llm-gateway/src/llm_gateway/modal.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+
+def _traffic_bucket(user_key: str) -> float:
+    digest = hashlib.sha256(f"{_TRAFFIC_BUCKET_SALT}:{user_key}".encode()).digest()
+    return int.from_bytes(digest[:8], "big") / 2**64
+
+
+def modal_traffic_fraction(product: str, settings: Settings) -> float:
+    return settings.glm_modal_product_traffic_fractions.get(product, settings.glm_modal_traffic_fraction)
```
richardsolomou, line 80: this needs to resolve product aliases before the lookup. with `posthog_code=1.0` and global `0.0`, valid `twig` and `array` requests still read `0.0`, so the per-product ramp and rollback silently dont apply.

services/llm-gateway/src/llm_gateway/glm_routing.py
```diff
@@ -0,0 +1,191 @@
...
+    should_route_glm_to_modal,
+)
+from llm_gateway.request_context import get_posthog_flags
+
+LlmCall = Callable[..., Awaitable[Any]]
+
+
+async def _modal_flag_enabled(user: AuthenticatedUser) -> bool:
+    # A caller-forwarded flag header wins; otherwise evaluate against PostHog.
+    forwarded = (get_posthog_flags() or {}).get(GLM_MODAL_FLAG)
```
richardsolomou: id push back on letting this header decide the backend. these public routes accept `x-posthog-flag-*` from any authenticated caller, so a client can force Modal even after the server flag and fraction are off. can we evaluate this flag server-side, or only trust a signal thats authenticated as internal?

services/llm-gateway/src/llm_gateway/flags.py
```diff
@@ -0,0 +1,52 @@
...
+        return _flag_cache[cache_key]
+    client = _get_client()
+    if client is None:
+        return None
+    try:
+        # Blocking /flags roundtrip — keep it off the event loop.
+        enabled = await asyncio.to_thread(client.feature_enabled, flag_key, distinct_id)
+    except Exception as exc:
+        logger.warning("flag_evaluation_failed", flag=flag_key, error=str(exc))
+        return None
```
richardsolomou, line 55: could we cache failed flag evaluations for a few seconds? rn every uncached user retries the blocking two-second request, so a flag outage turns a burst of new GLM users into repeated thread-pool work and request latency.

richardsolomou (changes requested):
could we add a regression for `twig` and `array` using the canonical product fraction, then cover Modal routing across Anthropic, Chat Completions, and Responses? the focused tests all pass without exercising those paths.

services/llm-gateway/src/llm_gateway/glm_routing.py
```diff
@@ -0,0 +1,191 @@
...
+            settings=settings,
+            flag_enabled=await _modal_flag_enabled(user),
+        )
+
+    if use_modal:
+        return await _send_via_modal(request_data, user, is_streaming, product, modal_config, make_modal_call, settings)
+
+    return await _send_via_cloudflare(
+        request_data, user, is_streaming, product, cloudflare_config, make_cloudflare_call, settings
+    )
```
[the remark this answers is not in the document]
tatoalo: Fixed in the squashed head: routing now checks backend availability first — when Cloudflare isn't configured (and the model has a Modal equivalent), Modal serves the request regardless of flag/fraction, so an advertised model can't land on a 503. Covered by `test_modal_only_configuration_routes_to_modal`.

services/llm-gateway/src/llm_gateway/glm_routing.py
```diff
@@ -0,0 +1,191 @@
...
+
+    use_modal = False
+    if is_modal_served_model(model) and is_modal_configured(settings):
+        use_modal = should_route_glm_to_modal(
+            model,
+            product=product,
+            user_key=str(user.user_id),
+            settings=settings,
+            flag_enabled=await _modal_flag_enabled(user),
+        )
```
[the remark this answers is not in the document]
tatoalo: Fixed: the env-fraction decision (pure, no I/O) now runs first and the flag is only evaluated when the fraction didn't already opt the user in — a full rollout via fraction never touches the flag service. Covered by `test_routes_to_modal_when_fraction_one_without_flag_roundtrip`.

services/llm-gateway/src/llm_gateway/modal.py
tatoalo: Done — `modal_traffic_fraction` now normalizes the product through `_normalize_cost_key` (and the settings validator canonicalizes keys at parse time), so twig/array read posthog_code's fraction. Covered by `test_product_aliases_resolve_in_fraction_lookup`.

services/llm-gateway/src/llm_gateway/glm_routing.py
tatoalo: Agreed — dropped the forwarded-header path entirely; the flag is now evaluated server-side only, so a caller can't force a backend the flag/fraction turned off. `test_forwarded_flag_header_cannot_force_modal` locks it in.

services/llm-gateway/src/llm_gateway/flags.py
tatoalo: Done — added a global per-flag 5s backoff cache: after one failed/absent evaluation, subsequent requests skip the roundtrip and fall back until it expires. Covered by `test_evaluate_flag_backs_off_globally_after_failure`.

services/llm-gateway/src/llm_gateway/modal.py
```diff
@@ -0,0 +1,131 @@
...
+    return _traffic_bucket(user_key) < fraction
+
+
+def _inject_modal_params(kwargs: dict[str, Any], api_base: str, modal_key: str, modal_secret: str) -> None:
+    kwargs["api_base"] = api_base
+    kwargs["api_key"] = _MODAL_PLACEHOLDER_API_KEY
+    # Never forward caller-supplied extra_headers: they arrive as extra-allowed request body fields,
+    # and e.g. a caller Host header would steer this authenticated request (and the proxy-token
+    # pair) to an attacker-controlled endpoint.
+    kwargs["extra_headers"] = {"Modal-Key": modal_key, "Modal-Secret": modal_secret}
```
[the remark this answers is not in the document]
tatoalo: Verified against litellm 1.84.0 (`headers = kwargs.get("headers") or extra_headers` + `headers.update(extra_headers)` — caller `headers` is the base dict) and fixed: the Modal path now pops `headers` and overwrites `extra_headers`, so only the server-controlled proxy-auth pair reaches the upstream request. Regression test feeds both dicts (mixed-case) and asserts `headers` is gone and `extra_headers` is exactly the auth pair.

services/llm-gateway/src/llm_gateway/flags.py
```diff
@@ -0,0 +1,58 @@
...
+    if cache_key in _flag_cache:
+        return _flag_cache[cache_key]
+    if flag_key in _flag_unavailable_cache:
+        return None
+    client = _get_client()
+    if client is None:
+        return None
+    try:
+        # Blocking /flags roundtrip — keep it off the event loop.
+        enabled = await asyncio.to_thread(client.feature_enabled, flag_key, distinct_id)
```
richardsolomou: in the pinned `posthoganalytics` 7.4.2, `feature_enabled()` defaults to `send_feature_flag_events=True`; with `sync_mode=True`, every uncached user does the `/flags` lookup and then blocks on a `$feature_flag_called` upload with the SDK's 15s timeout. can we pass `send_feature_flag_events=False` here?

services/llm-gateway/src/llm_gateway/glm_routing.py
```diff
@@ -0,0 +1,186 @@
+"""Backend selection for GLM (`@cf/...`) traffic during the Cloudflare -> Modal migration.
+
+Modal takes traffic opted in by the `tasks-glm-modal-inference` flag (caller-forwarded header, else
```
richardsolomou: can we update this docstring? it still says caller-forwarded headers opt traffic into Modal, but `_route_to_modal()` now ignores them and only uses the server-side flag/fraction.

richardsolomou (changes requested):
one test gap left: could we pin known user IDs to both sides of an intermediate fraction? the monotonicity test still passes if `_traffic_bucket()` always returns `0` or always returns `1`.

services/llm-gateway/src/llm_gateway/flags.py
tatoalo: great point, we would lose `$feature_flag_called` events from the gateway, so we won't reflect gateway evaluations right? I guess we don't care much as long as we don't block ofc

richardsolomou: yeah but imo thats fine here, this is an operational routing flag, and the provider metrics show where traffic actually went, so id rather drop the exposure event than keep capture on the request path.

richardsolomou (approved):
small test follow-up: could we parameterize the definitive cache case over `True` and `False`, and cover an SDK `None` result in the unavailable-backoff case? the disabled flag is the normal rollout state, and both branches can regress independently.

PR #71856, opened 2026-07-17
refactor(data-warehouse): migrate freshcaller, freshdesk, freshservice, front to rest_source (batch 17)

Gilbert09:
## Problem

Batch 17 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

## Changes

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green, reusing `validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, and framework `auth` for secret redaction):

- **freshcaller**
- **freshdesk**
- **freshservice**
- **front**

Not migrated this batch (left hand-rolled, valid blockers — all documented framework gaps):
- **exchange_rates_api** — client-side date-windowing + watermark-derived range, an HTTP-200 functional error envelope, and dict-to-rows normalization.
- **flowlu** — api_key travels in the query param, so it appears in the raised error's URL; the framework redacts logs but not exception messages.
- **formbricks** — self-hosted-instance security controls (streaming body size/time caps, dynamic DNS/private-IP SSRF guard) beyond static host-pinning, plus a 200-body retryable envelope.
- **freshsales** — view-based endpoints need a preparatory `/filters` request plus a global client-side view reduction (prefer the "All"-named view, else first) before listing.

## How did you test this code?

- Each migrated source's own `tests/` suite passes in isolation, asserting the same pagination progression, resume round-trips, termination, fail-loud paths, and `validate_credentials` mapping as before.
- Merged run of all 4 migrated dirs + `common/rest_source/tests/` — 352 passed.
- `hogli ci:preflight --fix` clean (0 failures).

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-16 PR (#71852); merge in order.

PR #71864, opened 2026-07-17
refactor(data-warehouse): exception-secret redaction + migrate flowlu, fly_io, fullstory, giphy, gitbook, gocardless (batch 18)

Gilbert09:
## Problem

Batch 18 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

This batch also lands a generic framework capability — **auth-secret redaction in raised exception messages** — so a source whose API key rides in a query param no longer leaks it into the user-visible error.

## Changes

Framework:
- `common/rest_source` now scrubs the configured auth's secret values out of every raised error message (`raise_for_status` HTTPError, `HTTP {status} for {url}`, malformed-JSON, redirect, and off-host errors). An API that carries its key in the URL query string (e.g. `?api_key=...`) previously leaked it via the request URL embedded in those messages; the secret is now replaced with `***` while the status/prefix text is preserved. Reuses the same `auth_secret_values` already registered for log redaction.

- **flowlu** — now expressible: its api_key rides in the query param, and the new exception redaction keeps the key out of `latest_error`
- **fly_io**
- **fullstory**
- **giphy** — query-param api_key, redaction-covered
- **gitbook**
- **gocardless**

- **gladly** — bulk data ships as streamed JSONL export files (`iter_lines` + per-line decode tolerance) with client-side job-watermark filtering and two transport modes.
- **gnews** — classifies a 403 by page number (graceful-stop on page > 1, fail-loud on page 1), not expressible via status/content-only response actions.

- New framework capability: 3 cases in `test_exception_redaction.py` — a query-param secret is scrubbed from both a 4xx HTTPError and a 5xx retryable error (status prefix preserved), and messages are untouched when no auth secret is registered.
- Merged run of all 6 migrated dirs + `common/rest_source/tests/` — 429 passed. Additive framework change: all existing `rest_source` tests plus auth-carrying consumer smoke (alguna, apify_dataset, delighted, checkout_com) stay green.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-17 PR (#71856); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/rest_client.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/rest_client.py
@@ -10,6 +10,7 @@ from requests import Request, Response, Session
 from requests.auth import AuthBase
 from requests.exceptions import (
     ChunkedEncodingError,
+    HTTPError,
     JSONDecodeError as RequestsJSONDecodeError,
 )
 from tenacity import RetryCallState, retry, retry_if_exception_type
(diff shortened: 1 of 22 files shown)


Gilbert09:

PR #71868, opened 2026-07-17
feat(autoresearch): cache anonymous preflight to cut login TTFB

pauldambra:
## Problem

Loading the login page should have a very low TTFB. Off cloud (self-hosted, devbox, local dev), the anonymous SPA-shell render embeds the preflight payload, and building it fanned out to roughly ten serial blocking service round trips per request (Redis twice, Postgres four times, ClickHouse, Kafka, a plugin-server HTTP call, object storage). All of them ran before the first byte was flushed, so their summed latency landed directly on TTFB, on every anonymous page load.

**Why:** the payload is identical for every anonymous visitor, so recomputing it per render buys nothing.

## Changes

`preflight_check` gains an `allow_cached` opt-in used only by the SPA-shell template render. The live `/_preflight/` endpoint, authenticated requests, and `settings.TEST` keep today's uncached behavior, so the self-hosted setup page still reflects service status immediately.

Layers, from hottest to coldest, for the anonymous login render:

1. **Per-worker in-memory copy** (15s TTL, single-slot atomic swap, copy-in/copy-out so per-request `auth_brand` mutation can't poison it) – steady state does zero external I/O.
2. **Stale-while-revalidate** – past the TTL (under a 300s cap) the stale copy is served instantly while a single-flight daemon thread refreshes both layers, so expiry never blocks a render.
3. **Shared cache** (60s TTL, Django cache) – cross-worker warm-up, one cache read.
4. **Rebuild** – the 10 independent checks (liveness probes plus the Postgres reads) run as one concurrent `ThreadPoolExecutor` batch, so even an inline cold build costs ~max(check) instead of sum. Cloud keeps its cheap serial short-circuits.
5. **Boot warm** – the wsgi/asgi post-fork first-request hooks kick the background refresher, so a worker's first real anonymous render finds the cache hot.

Net effect on the anonymous login render: warm-path blocking round trips 10 → 0, and the inline cold rebuild goes from 10 serial round trips to 1 parallel batch.

Created with Autoresearch.

- I (Claude) could not run the full stack or pytest in the sandbox this was authored in (no database available), so local verification was `ruff check`, `ruff format --check`, `py_compile`, and a throwaway AST-based round-trip counter used during the optimization run (not committed). CI runs the new tests.
- Added `TestAnonymousPreflightCache` (no-DB `SimpleTestCase`, all service probes mocked at the view's import site). Regressions each case catches, which no existing test did:
  - warm anonymous render re-running service probes (cache wiring silently dropped or broken)
  - one visitor's `auth_brand` cookie leaking into every anonymous visitor's payload (copy-out removed)
  - the live `/_preflight/` endpoint accidentally flipped to cached, freezing service status on the setup page
  - TTL expiry blocking the render inline again, or the background refresher failing to swap in a fresh payload (uses the single-flight lock as the deterministic completion signal)

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, internal performance change with no user-facing behavior or API shape change.

## 🤖 Agent context

Authored by Claude (PostHog Code) during an Autoresearch optimization run targeting login page TTFB. Skills invoked: `/setting-up-devbox`, `/writing-tests`. The sandbox had no Docker or devbox connectivity, so instead of a live CDP TTFB number the run optimized a deterministic proxy: the count of serial blocking external I/O round trips on the anonymous login render path, measured by a throwaway AST harness (kept out of the repo). Decisions along the way: cache opt-in only for the SPA-shell render (rejected caching the `/_preflight/` endpoint itself to keep the self-hosted setup page live), 60s shared TTL plus 15s in-process layer (rejected in-process only, which would lose cross-worker warm-up), copy-in/copy-out after spotting that the response dict is mutated per request, per-probe contextvars copies to keep OTel span parenting across the thread pool, and worker threads closing their DB connections. The boot warm reuses the existing post-fork first-request hooks because threads must not start pre-fork under Nginx Unit's prototype model.

---


pauldambra:
complexity that only benefits hobby and then only on the login page

PR #71873, opened 2026-07-17
feat(managed-migrations): add trial run support to batch-import-worker

jose-sequeira:
## Problem

Users setting up a managed migration have no way to test their data before committing to a full import. If their export has schema problems (malformed lines, missing distinct ids, wrong timestamps), they only find out after the import has started ingesting, or when it pauses partway through.

This PR adds worker-side support for trial runs: a bounded job that runs the exact same source, format, and transform code as a real import, but pairs each source record with the event(s) it would produce (or the reason it failed) and writes those results to an internal S3 bucket instead of capturing anything. The API and UI on top of this are in a follow-up PR; this one is safe to merge and deploy standalone since nothing creates `trial_s3` jobs yet.

## Changes

- A job whose sink is `trial_s3` takes a new sequential trial path (`src/trial/`) instead of the import pipeline, stopping after a configured record limit.
- The trial parser wraps the existing per-content-type transforms so parse and transform failures become per-record results instead of pausing the job. The chunk's trailing remainder still defers via `newline_delim_lines` (a new variant of `newline_delim` that tells the inner closure whether the line was newline-terminated), so lines split at chunk boundaries are re-read whole rather than misreported as errors.
- Results are written as seq-numbered JSONL pages plus a running summary (event name counts, error counts, dropped/skipped totals, timestamp range) to a worker-configured bucket (`TRIAL_BUCKET_*` env). Job config can never point output at an arbitrary bucket.
- Trial progress persists in the job's `state` JSON after pages are durably written, so a resume after backoff or lease loss overwrites the same page indices instead of starting over.
- Shared plumbing extracted rather than duplicated: `handle_fetch_error` (backoff/pause classification) and `load_or_init_state` are now used by both paths, and `select_and_fetch_next_chunk` is generic over the parsed record type.
- No schema changes: the sink variant rides `import_config` and progress rides `state`.

> [!NOTE]
> Error semantics are unchanged for real imports. Trials record data errors per line and keep going; transient errors (rate limits, timeouts, storage weather) back off and retry exactly like imports, so any error in trial results is by construction a non-retriable data problem.

Automated, all run by the agent:

- Unit tests for the trial parser (`src/trial/parser.rs`): per-line pairing and error tolerance, chunk-boundary remainder deferral (the regression that would silently corrupt output), and valid unterminated final lines.
- Unit tests for the sink (`src/trial/sink.rs`): page size bounding and continuous seq numbering across chunks, summary content.
- Unit tests for the summary (`src/trial/mod.rs`): record classification, timestamp range, event-name cap.
- End-to-end test (`tests/trial_e2e_test.rs`) running the real `TrialJob` against local Postgres and SeaweedFS with a chunk size small enough to split lines at boundaries: asserts lease-guarded progress flushes, the completed transition, and reads pages back from the store. Skips locally when the dev stack is down, hard-fails in CI if SeaweedFS is unreachable (same convention as the temp-bucket integration test).
- Full crate suite: 417 tests green; `cargo fmt`, `clippy --all-targets --all-features -D warnings`, and `cargo shear` clean.

Also verified manually end to end together with the follow-up API/UI branches against the local dev stack (SeaweedFS + Postgres): a 200-record trial with two planted bad records produced the expected pages, summary, and error reasons.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

User-facing docs will go with the API/UI PR that actually exposes the feature.

## 🤖 Agent context

Authored with Claude Code, directed by @jose-sequeira. Skills invoked: `/writing-tests`. Notable decisions: a dedicated sequential trial loop instead of genericizing the pipelined two-stage-commit `Job` (trials don't need transactional sinks or byte-precise resume, and the blast radius on the import path stays near zero); page writes ordered before progress flushes so resume is idempotent by re-fetch and overwrite; the trial output bucket is worker-env-configured rather than job-config-supplied for safety. The e2e test was deliberately switched from a DB-free harness copy of the loop to the real `TrialJob` against Postgres + SeaweedFS so the glue (store client, lease-guarded flushes, status transition) is covered.

### DIFF
--- a/rust/batch-import-worker/src/trial/job.rs
+++ b/rust/batch-import-worker/src/trial/job.rs
@@ -1,0 +1,236 @@
+use std::sync::Arc;
+
+use anyhow::{Context, Error};
+use lifecycle::Handle;
+use tokio::sync::Mutex;
+use tracing::{info, warn};
+use uuid::Uuid;
+
+use crate::{
+    context::AppContext,
+    error::ensure_user_message,
+    job::{
+        cleanup_committed_part_if_done,
+        config::SinkConfig,
+        handle_fetch_error, load_or_init_state,
+        model::{JobModel, JobState},
+        select_and_fetch_next_chunk,
+    },
+    parse::{format::ParserFnFor, Parsed},
+    source::DataSource,
+    staging::s3_client::create_trial_output_store,
+    trial::{parser::build_trial_parser, sink::TrialSink, TrialRecord},
+};
+
+/// A bounded trial run of an import: same source, format, and transforms as the
+/// real thing, but each source record is paired with its would-be output (or the
+/// reason it failed) and written to the trial output bucket instead of being
+/// captured.
+///
+/// Unlike [`crate::job::Job`], processing is sequential — no pipelined
+/// fetch/commit, no two-stage offset commit. Durability comes from ordering:
+/// pages are written to object storage before the progress that counts them is
+/// flushed, so a worker death re-fetches at most one chunk and overwrites the
+/// same page indices. Re-transformed records can differ in generated values
+/// (fresh UUIDs for events without one, identify-cache dedup state), which is
+/// acceptable for a preview.
+pub struct TrialJob {
+    context: Arc<AppContext>,
+    handle: Handle,
+    job_id: Uuid,
+    chunk_size: usize,
+    state: Mutex<JobState>,
+    model: Mutex<JobModel>,
+    source: Box<dyn DataSource>,
+    transform: Arc<ParserFnFor<TrialRecord>>,
+    sink: TrialSink,
+    record_limit: u64,
+}
+
+impl TrialJob {
+    pub async fn new(
+        mut model: JobModel,
+        context: Arc<AppContext>,
+        handle: Handle,
+    ) -> Result<Self, Error> {
+        let SinkConfig::TrialS3 { record_limit } = &model.import_config.sink else {
+            anyhow::bail!(
+                "TrialJob constructed for job {} without a trial_s3 sink",
+                model.id
+            );
+        };
+        let record_limit = (*record_limit).clamp(1, context.config.trial_max_record_limit);
+
+        // Fail fast (pausing the job with a user-facing message) when this
+        // worker has no trial bucket configured, before any source work starts.
+        let store = create_trial_output_store(&context.config)
+            .await
+            .map_err(|e| {
+                ensure_user_message(
+                    e,
+                    "Trial runs are temporarily unavailable. Please try again later.",
+                )
+            })?;
+        let prefix = format!(
+            "{}/team_{}/job_{}",
+            context.config.trial_bucket_prefix.trim_matches('/'),
+            model.team_id,
+            model.id
+        );
+        let sink = TrialSink::new(store, prefix, context.config.trial_records_per_page);
+
+        let is_restarting = model.state.is_some();
+        let source = model
+            .import_config
+            .source
+            .construct(&model.secrets, context.clone(), is_restarting, model.id)
+            .await
+            .with_context(|| "Failed to construct data source for trial job".to_string())?;
+        source.prepare_for_job().await?;
+
+        let transform = Arc::new(build_trial_parser(&model, context.clone()).await?);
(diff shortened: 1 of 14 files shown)


rust/batch-import-worker/src/trial/parser.rs
```diff
@@ -0,0 +1,269 @@
...
+    Ok(trial_parser(
+        &model.import_config.data_format,
+        transform_context,
+    ))
+}
+
+/// The trial-mode parser: same source format and transforms as the real import
+/// ([`FormatConfig::get_parser`]), but each source line becomes a
+/// [`TrialRecord`] pairing the line with its outputs, and parse/transform
+/// failures are recorded on the line instead of failing the chunk.
```
nickbest-ph, line 63: nice

rust/batch-import-worker/src/trial/parser.rs
```diff
@@ -0,0 +1,269 @@
...
+/// delegates to [`trial_parser`].
+pub async fn build_trial_parser(
+    model: &JobModel,
+    context: Arc<AppContext>,
+) -> Result<ParserFnFor<TrialRecord>, Error> {
+    let transform_context = TransformContext {
+        team_id: model.team_id,
+        token: context.get_token_for_team_id(model.team_id).await?,
+        job_id: model.id,
+        identify_cache: context.identify_cache.clone(),
```
nickbest-ph: We probably want to build a fresh cache instance for these trial jobs.  A production migration that ends up using the same cache as the trial job will end up with missing $identify events.

rust/batch-import-worker/src/trial/job.rs
```diff
@@ -0,0 +1,236 @@
...
+                Ok(None) => break, // Source exhausted before the record limit
+                Err(e) => Err(e),
+            };
+
+            if let Err(e) = outcome {
+                // Same classification as the import path: transient errors
+                // (source rate limits, object-store weather) back off and the
+                // trial resumes from its persisted progress; anything else
+                // pauses with a user-facing message.
+                handle_fetch_error(
```
nickbest-ph, line 160: nit, we are folding both the fetched/parse-side errors and the commit chunk errors into one outcome, then sending everything to `handle_fetch_error`. That function assumes that errors came from the parse side. If committing the chunk to s3 fails because for some infra reason on our side, we quarantine the users data and show them a message suggesting something is wrong with their data.

nickbest-ph (approved):
lgtm!
stoked for this, great addition.

small observation about trial jobs sharing the global identify cache and possibly polluting it

rust/batch-import-worker/src/trial/parser.rs
jose-sequeira: Good catch — fixed in b702ce6: trial parsers now build fresh in-memory identify and group caches instead of sharing the process-wide ones (the group cache had the same pollution problem, so both are isolated). A fresh cache also better simulates what a real import would emit.

rust/batch-import-worker/src/trial/job.rs
jose-sequeira: Fixed in 74ed5cc: commit-side failures (page writes, progress flushes) now get the "Trial runs are temporarily unavailable" message attached before classification, so transient ones still back off but a pause surfaces the infra message instead of implying a data problem.

rust/batch-import-worker/src/trial/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+/// whether the count was recorded.
+fn bounded_count(counts: &mut HashMap<String, u64>, key: &str, cap: usize) -> bool {
+    if let Some(count) = counts.get_mut(key) {
+        *count += 1;
+        return true;
+    }
+    if counts.len() >= cap {
+        return false;
+    }
+    counts.insert(key.to_string(), 1);
```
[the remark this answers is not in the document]
jose-sequeira: Valid finding — fixed in bd4372c. Summary keys (event names and error messages) are now truncated to 200 bytes at a char boundary before counting, so distinct oversized names merge into one truncated entry, and output timestamps longer than 64 bytes (impossible for the RFC 3339 strings the transforms produce) are ignored rather than cloned into state. Worst-case persisted summary is now ~120KB regardless of input. Covered by a unit test.

PR #71876, opened 2026-07-17
feat(impersonation): show current user in the user switcher

xljones:
## Problem

The impersonation user switcher only listed *other* members, so while impersonating a customer you couldn't see the current user's access level (member/admin/owner) at a glance.

## Changes

The current user now appears in the "Change user" dropdown in its normal power/name sort position (not hoisted to the top), disabled and marked "(you)" with their access-level pill, so their level is visible at a glance. Other members stay clickable to switch into.

When the current user is the only member, a "No other members" hint shows underneath, so a lone disabled row doesn't look like a bug.

impersonation-user-switcher

When you're the only member of the org:

impersonation-user-switcher-alone

`impersonationNoticeLogic.test.ts` passes (39 tests). The `orderedMembers selector` block covers the current user being sorted in place among other members (guards against re-hoisting or dropping them) and being listed on its own both before the members list loads and when they're the only member (guards the instant-display and "No other members" data conditions). The screenshots came from a throwaway Storybook story (not committed) rendering the real widget with mocked data. I (Claude) did not run a live end-to-end impersonation session.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover this staff-only widget.

## 🤖 Agent context

Built with PostHog Code (Claude), directed by Xander. A single `orderedMembers` selector builds the dropdown list: it filters the current user out of the raw members, prepends the `me` row (sourced from `userLogic`, so the current user shows instantly with an authoritative level before the members list loads), and sorts everyone by power level then name. The component marks the current user's row disabled and derives the "no other members" hint from that same list. An earlier revision hoisted the current user to the top and kept a second `changeableMembers` list; a later pass collapsed to the single selector for a cleaner result with identical behavior. Invoked `/writing-tests` before the test changes. Screenshots produced via a temporary Storybook story plus Playwright, then removed.

---


frontend/src/layout/navigation/ImpersonationNotice/ImpersonationNotice.tsx
```diff
@@ -163,16 +172,27 @@ function ImpersonationNoticeContent(): JSX.Element {
...
-            : changeableMembers.map((member) => ({
-                  key: member.user.uuid,
-                  label: <ChangeUserMenuItemLabel member={member} />,
-                  disabledReason: isChangingUser ? 'Switching user…' : undefined,
-                  // Always confirm via the modal (reason pre-filled) rather than switching silently.
-                  onClick: () => setPendingUserId(member.user.id),
-              }))
+    // Everyone is listed in power/name order. The current user's row is disabled and marked, so their
+    // access level is visible without hoisting them out of order.
+    const hasOtherMembers = orderedMembers.some((member) => member.user.uuid !== user?.uuid)
```
luke-belton: nit but should this live in the Kea logic?

luke-belton (approved):
nice - left one minor comment. Also could/should the code comments be a bit more succinct?

frontend/src/layout/navigation/ImpersonationNotice/ImpersonationNotice.tsx
xljones: yeah good call, fixed that [here](https://github.com/PostHog/posthog/pull/71876/commits/d92a264d6a3cdcb0b2538f092bb7f60b4fba31fb) and also tidied up the comments that were unnecessarily verbose

PR #60259, opened 2026-05-27
fix(aiobs): stop trace-clustering self-perpetuating loop

carlos-marchal-ph:
## Problem

The daily trace-clustering coordinator was emitting `$ai_generation_clusters` and `$ai_trace_clusters` events on behalf of teams that had nothing to cluster, and those empty events kept the same teams re-eligible for clustering the next day. The result was a self-perpetuating loop that inflated the apparent set of AI-active teams by ~50% versus reality, and roughly 70% of cluster events in the last week carried `$ai_total_items_analyzed = 0`.

Three colluding bugs:

1. The `DailyTraceClusteringWorkflow` did not short-circuit when the compute activity returned an empty result — it still ran labeling, aggregates, and emission, writing a cluster event with zero items.
2. Team discovery derived its trigger-event list from `AIEventType`, which includes the cluster events themselves. Any team that had an empty cluster event written on its behalf stayed eligible for the next 30 days regardless of customer activity.
3. The discovery window (30 days) was decoupled from the workflow window (7 days), so even without bug #1 discovery would enroll teams whose data the workflow then ignored.

## Changes

- **Fix 1 — Workflow short-circuit** (`workflow.py`): after the compute activity returns, if `compute_result.items` is empty, return a zero-item `ClusteringResult` immediately and skip labeling, aggregates, and emission. Stops the loop at the source.
- **Fix 2 — Decoupled discovery trigger list** (`llm_analytics_usage_report.py`, `team_discovery.py`): added `LLM_ANALYTICS_DISCOVERY_TRIGGER_EVENTS` containing only customer-emitted signals (no `$ai_*_clusters`, no `$ai_*_summary`, no `$ai_tag`). `get_teams_with_ai_events` takes the trigger list as a required parameter; usage-report and discovery pass their respective constants explicitly.
- **Fix 3 — Aligned discovery window** (`team_discovery.py`, clustering coordinator): `DEFAULT_DISCOVERY_LOOKBACK_DAYS` is now 7 (matches `DEFAULT_LOOKBACK_DAYS`). The clustering coordinator passes `lookback_days=inputs.lookback_days` through `TeamDiscoveryInput`, so discovery scopes its eligibility query to the workflow's own window.

## How did you test this code?

I'm an agent. Automated tests only — no manual verification.

[...]

PR #71878, opened 2026-07-17
fix(billing): don't read AI event count as billing user

carlos-marchal-ph:
## Problem

The usage report's per-team AI event count (`ai_event_count_in_period` on the `organization usage report` event) dropped sharply and permanently on **2026-05-28** — for low-volume teams only. This surfaced as a large drop in a saved LLM-analytics insight ([Orgs exclusively using LLMA and no other PostHog product](https://us.posthog.com/project/2/insights/JsGYFeni)), which went 742 → 275 orgs May → June and stayed flat since.

The AI usage itself did not fall. This is a counting regression in the usage report, not real churn.

### Chain of evidence

1. **The drop is a single-day cliff, not gradual.** Teams reporting `ai_event_count > 0` per day fell from ~10–11k to ~4k between May 27 and May 28, and stayed there.
2. **It's AI-count-specific.** On the same day, `event_count`, `recording_count`, `billable_feature_flag_requests`, and even `ai_credits_used` were all unchanged. Only the AI *event count* moved.
3. **Real AI usage continued.** Total AI event volume kept growing across the boundary, and AI-credit-consuming teams were flat — teams were still using LLM analytics.
4. **Only low-volume teams vanished.** Comparing the weeks either side of May 28, teams with 1–99 AI events fell ~60% while every larger bucket stayed flat. Those small teams are exactly the population the affected insight counts.
5. **The only code change to this query in that window** is when `get_teams_with_ai_event_count_in_period` was switched to run as the billing ClickHouse user (merged May 27; first daily usage-report run under it was May 28 — matching the cliff exactly).

### Root cause

AI events (`$ai_generation` and the rest of `AI_EVENTS`) live on the satellite AI-observability ClickHouse cluster, exposed through a distributed shim. The billing user does not resolve that shim completely, so the count comes back short — dropping low-volume teams to zero while high-volume teams (and every main-cluster counter) are unaffected. The sibling `ai_credits` query uses the same user and table but wasn't visibly hit, because credits only cover paying (higher-volume) teams — precisely the group that didn't drop.

## Changes

Remove the `ch_user=ClickHouseUser.BILLING` override from `get_teams_with_ai_event_count_in_period` so it reads as the default user, which resolves the satellite shim fully. This restores the behavior that produced correct counts before May 28. No other query is touched.

> [!NOTE]
> This is the minimal, provably-correct restore. If the billing user's isolation is wanted here, the alternative is to route this query through the dedicated `MAX_AI` (`llm/a`) user provisioned for the satellite cluster, or to grant the billing user full satellite access. I've left that call to the billing/LLM-analytics owners — flagging because I confirmed the trigger and the data symptom, but the exact ClickHouse access difference between users is infra-level and not visible from app code.

I (Claude, via the PostHog Slack app) did **not** run the test suite locally — this environment has no ClickHouse/flox stack. Verification so far is the data analysis above (queried via the PostHog MCP) plus reading the query path. The change is a one-line removal of a `ch_user` override; existing tests for `get_teams_with_ai_event_count_in_period` don't assert the billing user (the `ch_user == BILLING` assertion in the test file targets `_execute_split_query`, which is unchanged). CI should run the affected `posthog/tasks/test/test_usage_report.py` tests before this is marked ready.

## 🤖 Agent context

Investigated from a Slack thread asking whether the insight was broken. I traced the monthly drop to a May 28 daily cliff, isolated it to the AI event counter (all other billing counters and AI credits were unaffected), confirmed real AI usage continued, and narrowed it to the code change that switched this query to the billing ClickHouse user. Considered and rejected: the June gateway-exemption change (#64808, too late to explain May 28) and the billing-user change as a systemic cause (ruled out — only the AI counter moved). Chose the minimal restore over changing the CH user to `MAX_AI`, since the default reader is the exact pre-regression behavior; noted the alternative for owners.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C087XQ7K9K7/p1784236052906459?thread_ts=1784236052.906459&cid=C087XQ7K9K7)*

### DIFF
--- a/posthog/tasks/usage_report.py
+++ b/posthog/tasks/usage_report.py
@@ -1380,9 +1380,7 @@ def get_teams_with_ai_event_count_in_period(
             {"begin": begin, "end": end, "ai_events": AI_EVENTS},
             workload=Workload.OFFLINE,
             settings=CH_BILLING_SETTINGS,
-            # AI events live on the satellite AI-observability cluster, which the billing
-            # user does not read completely — routing this count through it drops low-volume
-            # teams to zero. Use the default reader, which resolves the satellite shim fully.
+            ch_user=ClickHouseUser.BILLING,
         )
 
 
--- a/posthog/tasks/usage_report.py
+++ b/posthog/tasks/usage_report.py
@@ -1380,7 +1380,9 @@ def get_teams_with_ai_event_count_in_period(
             {"begin": begin, "end": end, "ai_events": AI_EVENTS},
             workload=Workload.OFFLINE,
             settings=CH_BILLING_SETTINGS,
-            ch_user=ClickHouseUser.BILLING,
+            # AI events live on the satellite AI-observability cluster, which the billing
+            # user does not read completely — routing this count through it drops low-volume
+            # teams to zero. Use the default reader, which resolves the satellite shim fully.
         )
 
 


carlos-marchal-ph:
Closing — this was based on a wrong hypothesis. The drop in AI-only orgs was the intended effect of #60259 (stopping the trace-clustering discovery loop, which had been emitting $ai_*_clusters events back into teams and keeping dormant orgs counted as AI-active), not a usage-report bug. No change to usage_report.py is warranted.

PR #71892, opened 2026-07-17
feat(insights): quill in-chart legend for pie charts and exports

sampennington:
## Problem

The legacy `InsightLegend` component was replaced by the quill in-chart legend for trends line/bar charts, but two render paths were never migrated and still depend on it:

1. **Pie charts** in the app – `InsightVizDisplay` still rendered the legacy side legend next to the pie, because the quill `PieChart` had no legend support.
2. **Exported/shared insight images** – `ExportedInsight` rendered the legacy horizontal legend below the chart for all trends displays.

As long as these two consumers exist, `InsightLegend` (and its SCSS, shared with `BoxPlotLegend`) can't be deleted.

## Changes

- **quill `PieChart`** now takes `config.legend: ChartLegendConfig`, wired through the same `useChartLegend`/`ChartLegend` plumbing as every other multi-series chart. Toggling a row off removes the slice and the rest rescale to the full circle; the row stays listed but dimmed so it can be restored. Added a `WithLegend` story and updated the charts AGENTS.md.
- **`TrendsPieChart`** feeds `useInsightsLegendConfig` into the pie, so legend show/position/hidden-series state persists through `trendsDataLogic` like the line and bar charts, including the isolate/show-all context menu. `ActionsPie` joins `DISPLAYS_WITH_IN_CHART_LEGEND`, which gives pie the legend position select in the options menu and suppresses the legacy side legend.
- **`ExportedInsight`** maps the `legend` export option onto the chart's own in-chart legend (pinned to the bottom to match the old exported layout) for covered displays, instead of rendering the legacy horizontal `InsightLegend` below the chart. Non-covered displays (e.g. box plot) keep the legacy path.

After this, nothing renders `InsightLegend` except the box plot path, so a follow-up can delete it.

Automated tests only (no manual testing was possible in this environment):

- `PieChart.test.tsx`: legend hidden by default; clicking a legend row removes the slice (aria slice count drops), keeps the row dimmed, and restores on second click – covers the new quill legend wiring.
- `TrendsPieChart.test.tsx`: pie renders the in-chart legend without the legacy side legend, and toggling a series through `trendsDataLogic` removes the slice while keeping it restorable – covers the app-level wiring and hidden-series handoff to `config.legend.hiddenKeys`.
- `ExportedInsight.test.tsx` (new): with the legend export option on, the exported insight renders exactly one legend (the in-chart one, no legacy `InsightLegendMenu`); with it off, no legend at all – catches the double-legend/no-legend regressions in image exports.
- `InsightDisplayConfig.test.tsx`: pie now gets the legend position select in the options menu.

Full runs of the affected suites pass (quill PieChart + Legend, TrendsPieChart, ExportedInsight, InsightDisplayConfig, SqlPieGraph). Frontend typecheck produces a byte-identical error set to master (pre-existing, unrelated errors only).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs reference the exported-image or pie legend layout.

## 🤖 Agent context

Authored with Claude Code (PostHog Code). Skills invoked: `/writing-tests`. Notable decision: master removed the `PRODUCT_ANALYTICS_QUILL_LEGEND` feature flag mid-task (the quill legend is now unconditional), so the pie and exporter paths were wired directly to `usesInChartLegend`/`DISPLAYS_WITH_IN_CHART_LEGEND` rather than flag-gated as originally planned. In the exporter, the in-chart legend is forced to `position: 'bottom'` to preserve the legacy exported layout regardless of the insight's saved legend position.

---


packages/quill/packages/charts/src/charts/PieChart/PieChart.test.tsx
```diff
@@ -164,6 +164,41 @@ describe('PieChart', () => {
         })
     })
 
+    describe('interactive legend', () => {
+        function legendButtons(container: HTMLElement): HTMLButtonElement[] {
+            const legend = container.querySelector('[data-attr="hog-chart-pie-legend"]')
+            return legend ? Array.from(legend.querySelectorAll('button')) : []
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Blocker**

Branching logic in a test helper — the ternary is control flow inside a test file. Every test that calls `legendButtons` renders with `legend: { show: true }` (the absence case asserts directly and never calls it), so the fallback branch can't be exercised. Drop the branch:

```suggestion
            return Array.from(container.querySelectorAll<HTMLButtonElement>('[data-attr="hog-chart-pie-legend"] button'))
```

sampennington (commented):
🤖 **sp-ship** · 🚫 Blockers

One blocker: a ternary used as control flow in the new `legendButtons` test helper (branching logic in tests). Otherwise the migration cleanly reuses the established in-chart legend plumbing across the library, adapter, and exporter — the suggestions cluster on `ExportedInsight`, which re-derives the legend-display predicate locally and writes `legendPosition` even when no legend shows. One thing worth a quick visual check rather than a comment: the bottom legend's 40% height cap only resolves against an explicit container height, so a many-breakdown export could grow the legend at the chart's expense.

cc @sampennington

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
```diff
@@ -72,6 +63,25 @@ export function ExportedInsight({
         !SINGLE_SERIES_DISPLAY_TYPES.includes(trendsDisplay as ChartDisplayType) &&
         !DISPLAY_TYPES_WITHOUT_LEGEND.includes(trendsDisplay as ChartDisplayType)
 
+    // Displays covered by the quill in-chart legend draw the legend inside the chart itself.
+    const usesQuillInChartLegend =
```
sampennington, line 70: 🤖 **sp-ship** · ♻️✨📐 `[reuse+quality+xp]` · **Suggestion / follow-up**

`usesQuillInChartLegend` re-implements the predicate that `insightVizDataLogic`'s `usesInChartLegend` selector owns (the comment there calls it the single source of truth), so the classification now lives in two places that can drift. The exporter can't read the selector (it mutates the query before binding the logic), but consider extracting a pure `displayUsesInChartLegend(display)` next to `DISPLAYS_WITH_IN_CHART_LEGEND` and calling it from both — and while here, note the "which legend renders" decision is spread across `showLegend`, this block, and the JSX guard; a single `legendMode: 'in-chart' | 'side' | 'none'` derivation would keep the three in sync.

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
```diff
@@ -72,6 +63,25 @@ export function ExportedInsight({
...
+        !trendsDisplay || DISPLAYS_WITH_IN_CHART_LEGEND.includes(trendsDisplay as ChartDisplayType)
+
+    if (isInsightVizNode(insight.query) && isTrendsQuery(insight.query.source)) {
+        if (usesQuillInChartLegend) {
+            // The export `legend` option decides whether the chart shows its in-chart quill legend,
+            // pinned to the bottom to match the legacy exported layout (legend below the chart).
+            insight.query.source.trendsFilter = {
+                ...insight.query.source.trendsFilter,
+                showLegend: !!showLegend,
+                legendPosition: 'bottom',
```
sampennington, line 80: 🤖 **sp-ship** · 📊📐 `[hog-charts+xp]` · **Suggestion / follow-up**

This branch writes `legendPosition: 'bottom'` even when `showLegend` is false, so the exported query carries a position for a legend that never renders (and always discards a user-saved position, which is intentional per the comment). Harmless today, but setting `legendPosition` only when `showLegend` is true would make the intent — "pin to bottom *because* we're showing it" — self-evident.

frontend/src/exporter/ExportedInsight/ExportedInsight.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    cleanupJsdom()
+    cleanup()
+})
+
+describe('ExportedInsight', () => {
+    function renderExported(legend: boolean): { container: HTMLElement } {
+        return renderWithInsights({
+            component: (
+                <ExportedInsight
+                    insight={__trendsLineMulti as unknown as InsightModel}
```
sampennington, line 34: 🤖 **sp-ship** · 💻 `[code]` · **Suggestion / follow-up**

Both cases exercise only the line-chart fixture, yet the PR's headline change is the exported *pie* path. Since the export legend logic branches on the shared `usesQuillInChartLegend`, a pie case (e.g. the `trendsPie` fixture, asserting `hog-chart-pie-legend` renders and the legacy `InsightLegendMenu` doesn't) would directly cover the migration this PR is about.

packages/quill/packages/charts/src/charts/PieChart/PieChart.test.tsx
sampennington: 🤖 **sp-triage**

Applied — dropped the ternary fallback so `legendButtons` always queries `[data-attr="hog-chart-pie-legend"] button"` directly, matching the suggestion. Verified the full `PieChart.test.tsx` suite (16 tests) still passes. Fixed in fbd8508c818441811d09ea2d25940b4f6a288f35.

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
sampennington: 🤖 **sp-triage**

Out of scope for this PR — good follow-up idea, but extracting a shared `displayUsesInChartLegend(display)` (and the broader `legendMode` unification across `showLegend`/this block/the JSX guard) is a design change touching the selector, the exporter, and potentially other consumers, not a localized fix. Leaving as a follow-up rather than actioning here.

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
sampennington: 🤖 **sp-triage**

Intentional trade-off — setting `legendPosition` unconditionally is harmless (a position with no rendered legend is inert) and keeps the object shape simple. Making it conditional on `showLegend` is a minor clarity nice-to-have, not a correctness fix, so leaving as-is rather than actioning.

frontend/src/exporter/ExportedInsight/ExportedInsight.test.tsx
sampennington: 🤖 **sp-triage**

Good catch that pie coverage is missing here, but adding a new pie fixture case is a genuine test addition (not a mechanical single-line fix) and should go through the `/writing-tests` value-gate rather than being auto-added during triage. Leaving as a follow-up.

frontend/src/exporter/ExportedInsight/ExportedInsight.tsx
```diff
@@ -72,6 +63,25 @@ export function ExportedInsight({
...
+            }
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-triage**

Verified: `getQueryFromInsightLike` returns the caller's `query` object by reference when one already exists, so `insight.query.source.trendsFilter = {...}` here does write back into the original model. This mutation pattern predates this PR (the previous code also mutated `trendsFilter.showLegend` directly), but a correct fix means deciding where cloning should happen — locally in this component, or in `getQueryBasedInsightModel` for all callers — which is a design call beyond a single-line patch. Deferring to a human rather than auto-fixing.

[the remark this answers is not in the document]
sampennington: Fixed in 552a5ca0 — the exporter now clones `insight.query` (structuredClone) right after `getQueryBasedInsightModel`, so the export-only legend settings and table-control tweaks can no longer write back into the shared model. Left the shared utility untouched to keep the change scoped to the exporter.

PR #71899, opened 2026-07-17
feat(replay-vision): create a PostHog Task from an observation

arnohillen:
## Problem

Replay Vision finds things in recordings (a scanner flags a session), but there's no one-click path from a finding to getting it fixed. The observation just sits there. Meanwhile PostHog already has a product for tracked, agent-fixable work: Tasks (PostHog Code). Nothing connects the two.

## Changes

Add a `create_task` action on the observations viewset that turns a scanner finding into a PostHog Task. Title and description are derived from the scanner and its result, with the session and observation ids for traceability.

```mermaid
flowchart LR
    A[Observation<br/>scanner finding] --> B[POST .../observations/:id/create_task/]
    B --> C[has_tasks_access gate]
    C --> D[tasks facade<br/>create_task_without_run]
    D --> E[PostHog Task]
    class A phYellow
    class B phBlue
    class C phBlue
    class D phRed
    class E phYellow
```

- **Additive only.** A new POST action on an existing viewset. No changes to existing endpoints, models, or behavior, and no migrations (reuses the `USER_CREATED` task origin).
- **Record-only.** Uses `create_task_without_run` through the tasks facade, so no repository or GitHub integration is required. The coding agent isn't started (that's a natural follow-up once repo selection is wired).
- **Gated** on `has_tasks_access` (PostHog Code access), on top of the existing `replay_scanner:read` + `session_recording:read` scopes.
- Mirrors how Signals (`auto_start.py`) and Experiments (`experiment_service.py`) already create tasks via the same facade. Added the `products.tasks` dependency edge in `tach.toml` (facade-only import), same as `review_hog` and `experiments`.
- Regenerated OpenAPI types, so the endpoint is also exposed as an MCP tool automatically.

> [!NOTE]
> Backwards compatible and safe to merge independently. The endpoint only does something when called, requires PostHog Code access, and creates a plain task record.

I'm an agent (Claude Code). Added `test_observation_create_task_api.py` with two cases, both passing:
- creating a task returns 201 with the task id, and the facade is called with the right team, a title carrying the scanner name, and a description carrying the session (guards the observation→task mapping and the cross-product contract);
- without `has_tasks_access` the endpoint returns 403 and creates nothing (guards the access gate, so an observation reader can't mint tasks).

The tasks facade is mocked as the cross-product boundary. Also ran `ruff`, `tach check --dependencies --interfaces` (clean after the dependency edge), and `hogli build:openapi`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs yet (backend action; frontend button is a follow-up). `skip-inkeep-docs` applies.

## 🤖 Agent context

Built with Claude Code (Opus 4.8). Scoped for feasibility first (the concern was that Tasks might require a GitHub repo, which would make this heavy) and confirmed it's additive and repo-free via `create_task_without_run`. Skills invoked: `/improving-drf-endpoints` (new viewset action + serializer) and `/writing-tests` (the value gate for the two tests).

Decisions:

- **Record-only, not create-and-run.** Keeps it repo-free and backwards compatible. Running the agent (`create_and_run_task` with a repository) is a clean follow-up.
- **Reused the `USER_CREATED` origin** rather than adding a `REPLAY_VISION` choice, to avoid a migration in another team's isolated product. A dedicated origin is an optional follow-up for cleaner attribution.
- **Deep link deferred.** The description carries the session and observation ids; a rendered replay URL is a small follow-up.


ksvat (approved):
fix the scope but otherwise good to go!

PR #68427, opened 2026-07-05
fix(data-warehouse): require the API key when connecting Stripe by key

Gilbert09:
## Problem

When onboarding Stripe as a data warehouse source with a restricted API key, the wizard let users click Next with the API key field left blank. Validation only ran later when the sync tried to authenticate, so the failure surfaced far from where it was caused. Session recordings of real onboarding runs showed users proceeding past this step empty-handed and then having to backtrack once the connection errored.

## Changes

The `stripe_secret_key` field (shown when "Restricted API key" auth is selected) was marked `required=False`. This flips it to `required=True`, so the wizard validates it inline on submit like it already does for the Postgres connection fields and for the API key on RevenueCat, Cloudflare, and Attio.

The nested-select validation path only checks the fields under the currently selected auth option, so the OAuth path is unaffected. Editing an existing source still allows a blank key, because the update flow passes `allowBlankSensitiveFields` and skips blank secret fields.

## How did you test this code?

I (Claude) verified the frontend validation wiring by reading the wizard logic: `buildKeaFormDefaultFromSourceDetails` seeds the `auth_method` selection to its `api_key` default, and `getErrorsForFields` walks into the selected option's fields and errors on empty required fields. So an empty key now produces an inline error on Next instead of a late backend auth failure. Ran `ruff check` on the touched file (passes). I did not run the app manually.

## 🤖 Agent context

**Autonomy:** Fully autonomous

Authored by Claude Code as part of a pass over data-warehouse new-source onboarding friction observed in session recordings. I checked that RevenueCat, Cloudflare, and Attio already mark their key field required, confirming Stripe was the odd one out rather than an intentional pattern, and confirmed the edit path still tolerates a blank secret before changing the flag. No skills were required for this one-line config change.

PR #71915, opened 2026-07-17
chore(data-warehouse): split generated_configs into per-source modules

Gilbert09:
## Problem

Every new-source implementation PR regenerates `generated_configs.py` — a single ~5,700-line file holding one generated config class per source (~860 sources). That file is a `backend:contract-check` input for warehouse_sources, so each of those PRs triggers the full Django CI matrix (~38 sharded jobs) plus the dependent-product cascade, even when the PR only implements one source (e.g. #71198 touched only its source dir, SOURCES.md, and generated_configs.py). The single shared file is also a merge-conflict hotspot across parallel source PRs.

Stacked on #71908, which narrows the rest of the contract inputs.

## Changes

- `generated_configs.py` becomes a `generated_configs/` package with one generated module per source (`stripe.py`, `postgres.py`, …, named `ExternalDataSourceType.<MEMBER>.name.lower()`), plus a hand-written, never-regenerated `__init__.py` exposing `get_config_for_source`, which derives module and class names from the enum member at runtime — adding a source touches no shared file at all.
- `manage.py generate_source_configs` now emits per-source modules, asserts module-name uniqueness, and deletes stale modules for removed sources. Output is idempotent (verified by regenerating twice).
- All import sites (~1,300 files, overwhelmingly `sources/*/source.py` and their tests) are repointed to the per-source module (`from ...generated_configs.stripe import StripeSourceConfig`), keeping full static typing — no `__getattr__` re-exports returning `Any`. External importers updated: the four `posthog/hogql/direct_sql/*_adapter.py` files, `snowflake_connection_cache.py`, two runtime test imports, and the five facade `_LAZY` re-export targets.
- Contract-check inputs (`products/warehouse_sources/turbo.json`) now list only the config modules core actually reaches: the facade-re-exported `custom`/`github`/`mysql`/`postgres`/`snowflake`, the directly-imported `redshift`, and the static `__init__.py`. Changing any other source's generated config no longer triggers the full Django matrix.
- `test_ci_core_coupled_sources.py` is hardened to per-module granularity for generated_configs: if core starts consuming a config module that isn't in the contract inputs, the guard fails instead of silently accepting whole-package coverage.
- A new test asserts `get_config_for_source` resolves for every registered source — catching drift between the generator's file-naming rule and the resolver's runtime derivation, which are maintained in two places by design.
- `[tool.ty.src]` exclude updated to the package glob; `source.template` and the implementing-warehouse-sources skill updated to reference per-source modules.

**Pre-existing drift note:** regenerating from today's source definitions changes the Stripe config classes (auth fields became required in the form definition in #68427/#68437 without a regen, but the stored-config dataclass stayed optional so legacy configs keep parsing). This PR preserves the checked-in (optional) behavior byte-for-byte; reconciling that drift is a separate follow-up. All other 856 modules match a fresh regen exactly (after formatting).

- `pytest` on `sources/tests/test_generated_configs.py` (29, incl. the new every-source resolver test — this loads the full source registry, exercising every repointed import), `sources/tests/test_ci_core_coupled_sources.py`, `common/test/test_source_config_generator.py` (17, snapshot regenerated: output no longer embeds the 860-entry mapping function), and `sources/stripe/tests/` (66).
- `bin/hogli product:lint warehouse_sources`, `tach check --dependencies --interfaces`, `lint-imports`, `ruff check`/`format` — all green.
- Runtime smoke: `django.setup()` then imported all four direct-SQL adapters, the five facade lazy re-exports, and resolved `get_config_for_source(STRIPE)`.
- Generator idempotency: ran `manage.py generate_source_configs` twice — second run produced a zero diff.
- Affectedness proofs via `.github/scripts/turbo-discover.js` exactly as CI invokes it: a `generated_configs/logz_io.py`-only change → `run_legacy: false`, matrix = warehouse-sources only; control on `generated_configs/postgres.py` → `run_legacy: true` with the full 20-product cascade, unchanged.

New tests justification: the resolver loop test catches generator/resolver naming-rule drift (the rule now lives in two files by design); the guard-test hardening catches a core-consumed config module missing from contract inputs — a coverage hole that couldn't exist before the split.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `.agents/skills/implementing-warehouse-sources/SKILL.md` and `sources/source.template` to the per-source module layout. No user-facing behavior changes.

## 🤖 Agent context

Claude Code session continuing the CI fan-out audit from #71908. Skills invoked: /writing-tests (for the guard-test hardening and resolver test). Design decisions: a static `__init__.py` with runtime name derivation was chosen over a generated eager re-export `__init__` (which would be regenerated per source, keeping both the contract trigger and the merge-conflict hotspot) and over PEP 562 lazy re-exports (which type every config as `Any` at ~1,300 sites); the one-time import repoint was done with an AST codemod. The fresh-regen Stripe drift was detected by the existing stripe config tests failing against regenerated output; the checked-in behavior was deliberately preserved.

Note for in-flight source PRs: rebase over this and re-run `pnpm generate:source-configs` to resolve any `generated_configs.py` conflict — the regenerated per-source module replaces the old flat-file hunk.


estefaniarabadan (approved):
nice to haves that I didn't see (and claude couldn't find either):
- remove also __pycache__/<module>.*.pyc files, or clearing the __pycache__ directory entirely during regeneration

danielcarletti (approved):

georgemunyoro (approved):

Gilbert09:
Rebased onto `master` and resolved the conflicts. A heads-up on one non-mechanical decision, since it changes this PR's scope:

**`master` removed the `turbo.json` contract-check mechanism this PR was narrowing.** [#71486](https://github.com/PostHog/posthog/pull/71486) (facade wiring-couplings doctrine) deleted `products/warehouse_sources/turbo.json` entirely, moving `warehouse_sources` to "watch everything" (fail-safe) until its facade stops handing out unsanctioned classes, with `hogli product:lint` now enforcing it. So the PR's turbo.json narrowing and the `test_ci_core_coupled_sources.py` hardening no longer have a mechanism to hook into, and re-adding a narrowed `turbo.json` would fight that doctrine (and likely trip the new lint).

How I resolved it:
- **Kept** the valuable, self-contained part — splitting `generated_configs.py` into per-source modules (generator, `__init__` resolver, ~1,300 import repoints, facade re-exports, docs). This stands on its own and is a prerequisite for eventually earning narrowing back under the new doctrine.
- **Dropped** the now-obsolete `turbo.json` narrowing (accepted `master`'s deletion) and took `master`'s `test_ci_core_coupled_sources.py` (which already no-ops when `turbo.json` is absent).

Mechanical resolution notes:
- `master` added 7 new sources (axiom, crisp, dataforseo, kommo, plivo, promptwatch, sleekplan) and evolved 14 config classes (bigquery, stripe auth, datadog, sentry, …) since this PR's base. I re-derived the affected per-source modules from `master`'s authoritative generated output so the package matches `master`'s source definitions byte-for-byte (verified), and repointed every new/merged source + test import to its per-source module.
- Also fixed a latent lint issue in the generator: emitted imports now follow ruff/isort grouping (stdlib `typing` first, then a blank line, then first-party), which was the original CI ruff failure.

Please sanity-check the turbo.json call — if you'd rather pursue narrowing under the new doctrine, that's follow-up work (move the facade-coupled classes into wiring locations first), and I can help.

Gilbert09:
@estefaniarabadan good call on the `__pycache__` cleanup — implemented in 87f170a: when the generator removes a module for a source that left the registry, it now also deletes that module's stale `.pyc` from `__pycache__` so a leftover bytecode file can't keep the deleted config importable.

PR #69542, opened 2026-07-09
chore(data-catalog): setup skill for cataloging a project

thiagosalvatore:
## Problem

The data catalog already has REST, SQL, and MCP surfaces plus a `setting-up-data-catalog` skill that teaches agents the workflow. Two gaps were left:

1. The skill's maintenance flow (review the proposal queue, promote items on a human's say-so) needs the pending queue to be discoverable through the SQL read path. It wasn't. `information_schema.relationships` listed only real joins, and the `certification` column on `information_schema.tables` showed only settled marks. So an agent had no way to find pending proposals, or the ids the promotion tools require. Flow 3 pointed at queries that returned nothing.
2. Only about a quarter of customers have the skill installed. Whatever we expose has to be safe and useful for agents without it, so they lean on what is verified and approved instead of guessing.

## Changes

Three pieces.

### Surface the review queue via `information_schema` (verified-first)

Two new self-contained catalog tables:

- `information_schema.certifications` — the full trust-mark lifecycle, including `proposed`, with the id the certify/deprecate tools need.
- `information_schema.relationship_proposals` — the pending join review queue.

The verified/usable surfaces stay verified-only, so an agent that queries them never meets an unreviewed item:

| What an agent asks | Verified surface (unchanged) | Review queue (new) |
| --- | --- | --- |
| How do these tables join? | `information_schema.relationships` — real joins only | `information_schema.relationship_proposals` |
| Which sources should I trust? | `tables.certification` — settled marks only | `information_schema.certifications` |
| What is the canonical metric? | `information_schema.metrics` — status-labeled | same table, `WHERE status = 'proposed'` |

> [!NOTE]
> A pending join proposal never appears in `information_schema.relationships`. A skill-less agent can't copy unreviewed join keys into a manual `JOIN` from the joins table, because the proposal isn't there. That safety is structural, not dependent on the agent reading a `status` label. Every new surface is flag-gated and stripped when the data-catalog flag is off, so the flag-off catalog is byte-identical to before (asserted by the `serialize_database` snapshots).

### Enable `refresh-from-insight`

Agents can now clear metric drift by re-snapshotting the linked insight. If that changes the definition of an approved metric, it resets to `proposed` for a fresh human approval. I also corrected the `metric-update` tool description: approval resets on definition, description, or unit edits, not just the definition.

### The `setting-up-data-catalog` skill

Teaches the three flows, with Flow 3 now pointing at the working queue queries:

[...]

PR #69538, opened 2026-07-09
feat(hogql): add system.information_schema.metrics table

thiagosalvatore:
## Problem

The catalog is only useful if agents can find a metric without special tooling. The RFC's read path is SQL-first: an agent already discovering the schema through `system.information_schema` should find governed metrics the same way. This PR adds that surface on top of the metric flow (#69527-#69530).

## Changes

- **`system.information_schema.metrics`** — a new HogQL LazyTable, one row per metric, discoverable with `SELECT ... FROM system.information_schema.metrics WHERE name ILIKE '%mrr%'`.
- **Feature flag `product-data-catalog`** — the semantic layer is not GA. The Database build prunes the metrics table node when the flag is off (one org-keyed evaluation per build, `products/data_catalog/backend/logic/flags.py`), so for non-flagged teams the table is absent from `information_schema` listings and direct queries raise `Unknown table`. RBAC row-gating stays as defense in depth.
- **Loader** reads the `data_catalog` facade (fail-soft), computes drift in one bulk pass shared with the REST serializer, hides metrics whose definition references a table the caller is denied (set intersection over the cached `referenced_table_names`), and supports `name` pushdown (equality / IN).
- **One vocabulary across SQL and REST**: persisted `status` plus computed `is_drifted`, so an agent filters `WHERE status = 'approved' AND NOT is_drifted`.
- **Agent guidance, restructured per review** — the `execute-sql` description now has two subsections: "Regular schema discovery" (unchanged pre-PR content) and "Metric discovery (semantic layer)" (new `metric-discovery.md`, ~0.8k chars vs the original ~2.7k addition). The metric section renders only when `product-data-catalog` is enabled for the project — the MCP threads `toolFeatureFlags` into the description render, so a flag-off render is byte-identical to the un-gated prompt. A unit test pins both flag states plus a char budget on the section.
- **Governed-metrics eval suite** — `products/data_catalog/evals/eval_governed_metrics.py` (first product-owned suite root): five sandboxed cases covering approved-metric adoption/citation, proposed-never-canonical, drifted-distrusted, empty-catalog-no-stalling, and no-catalog-detour-on-ad-hoc-questions. Deterministic catalog-lookup scorers plus one graded judge.
- **tach**: `posthog` core `depends_on` `data_catalog` because the loader inline-imports the facade.

> [!NOTE]
[...]

PR #71921, opened 2026-07-17
feat(mcp): route metric discovery to the governed-metrics catalog

thiagosalvatore:
## Problem

Asking a PostHog MCP client "what metrics do I have available?" gets answered with saved insights instead of the governed-metrics catalog (`system.information_schema.metrics`, live since #69538). Two gaps in the prompt steering cause this:

- `entity-schema-discovery.md` routes every "find / which / do we have X" question to `system.*` entity search, with `system.insights` named first, and nothing distinguishes an artifact search from a schema/catalog question.
- `metric-discovery.md` (the only guidance mentioning the metrics catalog) covers only named-number lookups ("what's our MRR?"), not listing questions, and is spliced only into the `execute-sql` tool description. Chat hosts that discard the MCP `instructions` payload (Claude web/desktop) may never load it.

## Changes

- `metric-discovery.md`: lead with the listing case ("what metrics do we have?" is a catalog listing, not an insight search) before the existing named-number lookup guidance.
- `entity-schema-discovery.md`: one un-gated sentence splitting saved-artifact search (`system.insights` et al.) from "what data / tables / events do we have available?" schema questions (`system.information_schema.tables` + `read-data-schema`). Deliberately does not mention the metrics table, which only exists when the data catalog flag is on.
- `InstructionsContext` gains `dataCatalogEnabled` (set once in `InstructionsBuilder.buildContext` from the `product-data-catalog` flag), and the metric-discovery section is spliced into the three surfaces chat hosts actually read: tools-mode instructions, the `exec learn analytics` topic, and the full exec command reference. The analytics topic description also advertises governed metrics when the flag is on.
- Flag-off renders stay byte-identical (existing convention); `formatExecuteSqlDescription` keeps its unchanged splice.

> [!NOTE]
> The `learn analytics` topic content is served at runtime by `ExecHelpCatalog.get()`, so the spliced section doesn't count against claude.ai's ~16,384-char `inputSchema` cap. The only budget-counted change is the ~45-char longer topic description, and the budget test now measures the flag-on worst case.

- New parameterized test in `instructions-formatter.test.ts`: catches removal of the flag splice on any of the three surfaces (guidance silently vanishing), and catches un-gating (flag-off renders must be byte-identical to a context without the field, so flag-off orgs are never steered at a table that doesn't exist for them). No existing test covered these surfaces; `execute-sql-description.test.ts` only covers the SQL tool description.
- Budget test now sets `product-data-catalog: true` so the claude.ai registry cap is measured against the new worst case.
- Ran the full MCP unit suite (74 files, 2169 tests) plus `pnpm --filter=@posthog/mcp run typecheck` - all green. Snapshot updates are exactly the new un-gated routing sentence.
- Not manually re-tested against a live chat host in this PR.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code after debugging why a claude.ai session connected to a local MCP answered a metric-availability question from `system.insights`. Traced the steering surfaces per client mode (tools-mode vs single-exec, and which hosts drop the MCP `instructions` payload) before choosing where to splice the guidance. Skills invoked: /writing-tests. Kept this PR independent of the in-flight data-catalog stack (#69540 → #69541 → #69542 → #69544) since everything it touches is already on master; a companion eval case lands on the stack's eval branch separately.

### DIFF
--- a/services/mcp/src/hono/instructions.ts
+++ b/services/mcp/src/hono/instructions.ts
@@ -65,6 +65,7 @@ export class InstructionsBuilder {
             renderUiEnabled: state.renderUiEnabled,
             metadata: state.metadata,
             groupTypes: state.groupTypes,
+            dataCatalogEnabled: state.toolFeatureFlags?.[PRODUCT_DATA_CATALOG_FLAG] === true,
         }
     }
 
--- a/services/mcp/src/lib/instructions-formatter.ts
+++ b/services/mcp/src/lib/instructions-formatter.ts
@@ -23,6 +23,7 @@ import ENV_CONTEXT from '@/templates/sections/env-context.md'
 import EXAMPLES from '@/templates/sections/examples.md'
 import EXEC_LEARN from '@/templates/sections/exec-learn.md'
 import EXEC_TOOL_BLURB from '@/templates/sections/exec-tool-blurb.md'
+import METRIC_DISCOVERY from '@/templates/sections/metric-discovery.md'
 import RETRIEVING_DATA from '@/templates/sections/retrieving-data.md'
 import SCHEMA_WORKFLOW from '@/templates/sections/schema-workflow.md'
 import TOOL_SEARCH from '@/templates/sections/tool-search.md'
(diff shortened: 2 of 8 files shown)


thiagosalvatore:
The new coverage does not exercise the production feature-flag wiring. The parameterized test constructs `InstructionsContext` with `dataCatalogEnabled` directly and calls the formatter, bypassing the new mapping in `InstructionsBuilder.buildContext` (`services/mcp/src/hono/instructions.ts:68`). If that mapping is removed or mis-keyed, all of the new assertions remain green while real clients lose the guidance.

Could we add a builder-level flag-on/flag-off test that renders from a `ResolvedState`? Since the core change is prompt routing, I would also add a sandboxed listing case for “what metrics are available?” that asserts `system.information_schema.metrics` is queried and `system.insights` is not; the current governed-metrics eval suite only covers named-number lookups.

PR #71925, opened 2026-07-17
feat(signals): rich scanner-finding evidence cards in inbox

andrewm4894:
@Twixes here - I'm the actual initiator via PostHog in Slack. This is not well-tested, which I think is fine.

---

## Problem

Replay-vision **scanner findings** in the Inbox rendered with the generic fallback card — a markdown blurb plus a raw-JSON metadata dump — with no way to jump to the recording at the moment the finding was observed. "Session problem" evidence, by contrast, already gets a rich card with a thumbnail that plays the recording seeked to the observation. This closes that gap so both kinds of replay-derived evidence read the same way.

## Changes

- New `ScannerFindingSignalCard` for `replay_vision` / `scanner_finding` signals, modeled on `SessionReplaySignalCard`:
  - 16:9 thumbnail (from the finding's exported asset) that is itself the play affordance — clicking opens the recording in the modal player, seeked to the finding via `recording_start_time + start_time`.
  - Recording-existence check so the play button disables (instead of opening an empty player) when the recording wasn't captured.
  - Problem-type and confidence tags, plus an affected-user / finding-window / active-total-duration meta line.
- Registered the card in the signal-card registry.
- Added `replay_vision` to the source-product icon metadata, the Inbox source filter, and the "Product · Signal type" header line so the source is recognized rather than falling back to a generic dot/label.

No backend or serializer changes — the finding payload already carried everything needed (`session_id`, `exported_asset_id`, offsets, `recording_start_time`).

Static review against the sibling `SessionReplaySignalCard` and the generated `ReplayVisionScannerFindingSignalExtraApi` type; the field usage and cast pattern mirror the existing card. I was not able to run the frontend typecheck/lint or Storybook in this environment (no `node_modules` present), so those still need to run in CI. No automated tests were added — the sibling rich card has none, and a component test would require heavy kea setup for marginal value.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread. The task: rework scanner-finding evidence cards in the Inbox to match the rich, seek-to-recording "session problem" cards. Explored the Inbox card registry, found `replay_vision` had no dedicated renderer (fell through to `GenericSignalCard`) and was missing from the source-icon and filter maps, then added a card built on the same `useRecordingButton` / thumbnail / existence-check primitives as `SessionReplaySignalCard`. Key difference handled: scanner findings carry `start_time`/`end_time` as float second-offsets plus an absolute `recording_start_time`, so the seek instant is computed with `dayjs(recording_start_time).add(start_time, 'second')` rather than the colon-delimited-offset math the session-problem card uses. No skills were invoked (no migration/DRF/API-type/test surface touched).

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1784288579393159?thread_ts=1784288579.393159&cid=C09SK2PAGKF)*


Twixes:
Going to merge this without testing really, as it's a surface not seen by users - low blast ardius, and I have 98% in the solution here.

PR #71936, opened 2026-07-17
feat(aio): verify gateway-relayed OTLP traces

richardsolomou:
## Problem

Gateway-relayed OTLP traffic needs trusted provenance without letting clients forge sponsorship or bypass billing controls. Sponsorship must remain bounded across reporting periods, retries, delayed gateway generations, and exhausted allowances.

## Changes

Verify gateway signatures through shared capture primitives and bind relayed batches to signed gateway request IDs. A chronological ledger applies bounded trace and evaluation allowances across reporting periods. Legacy ingestion strips client-supplied gateway properties, and verified relays obey the global events quota while bypassing only the scoped LLM quota.

The native-JSON property declarations are isolated in prerequisite PR #78109.

- Ran the focused signed OTLP global-quota endpoint regression.
- Ran Clippy with warnings denied and Rust formatting checks.
- Started the local stack and verified OTLP evaluation output through the live capture HTTP endpoint and Kafka.
- Regenerated the billing snapshot against live Postgres and ClickHouse.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the billing and relay contract in the companion gateway PR.

## 🤖 Agent context

PostHog Code implemented the provenance path. `/rs-address-pr-review`, `/clickhouse-migrations`, `/writing-tests`, `/writing-code-comments`, and `/run-posthog` shaped the review round.

---


richardsolomou:
* **#71937** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#71936** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71936?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

posthog/tasks/usage_report.py
```diff
@@ -1390,6 +1415,78 @@ def get_teams_with_ai_event_count_in_period(
...
+                      AND event IN %(ai_events)s AND timestamp >= %(sponsor_begin)s AND timestamp < %(sponsor_end)s
+                    GROUP BY team_id, event, verified, relay, trace_id, span_id
+                )
+                WHERE verified AND relay AND trace_id != '' AND span_id != ''
+            ) AS relays
+            INNER JOIN (
+                SELECT
+                    team_id,
+                    trace_id,
+                    uniqExact(request_id) AS generation_count
```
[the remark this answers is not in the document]
richardsolomou: done, each request ID is assigned to one trace before sponsorship is calculated, with a regression for cross-trace replay.

posthog/tasks/usage_report.py
```diff
@@ -1390,6 +1416,86 @@ def get_teams_with_ai_event_count_in_period(
...
+                    SELECT
+                        team_id,
+                        {verified_expr} IN ('true', '1') AS verified,
+                        {relay_expr} IN ('true', '1') AS relay,
+                        if(verified AND NOT relay, {trace_id_expr}, '') AS trace_id,
+                        if(verified AND NOT relay, {request_id_expr}, '') AS request_id
+                    FROM {events_read_table(use_new)}
+                    WHERE team_id IN %(relayed_team_ids)s
+                      AND event = '$ai_generation'
+                      AND timestamp >= %(sponsor_begin)s AND timestamp < %(sponsor_end)s
```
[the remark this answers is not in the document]
richardsolomou: done, sponsorship now only counts generations at or before each relay, with a regression for later generations.

rust/capture/src/otel/filtering.rs
```diff
@@ -24,9 +25,22 @@ pub async fn check_quota(
     limiter: &CaptureQuotaLimiter,
     token: &str,
     span_events: &[SpanEvent],
+    gateway_verified: bool,
 ) -> Result<(), QuotaOutcome> {
+    let count = span_events.len();
+
+    if gateway_verified {
```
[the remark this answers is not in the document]
richardsolomou: id push back on this one: verified relay batches intentionally skip the scoped AIO limiter but still hit the global Events quota. the 100-event sponsorship only decides which accepted events are free; replays and overage remain billable.

posthog/tasks/usage_report.py
```diff
@@ -1390,6 +1416,101 @@ def get_teams_with_ai_event_count_in_period(
...
+                    relays.relay_rank AS relay_rank,
+                    countIf(sponsors.sponsor_timestamp <= relays.relay_timestamp) AS generation_count
+                FROM (
+                    SELECT
+                        team_id,
+                        event,
+                        trace_id,
+                        span_id,
+                        relay_timestamp,
+                        row_number() OVER (
```
pawel-cebula: This still looks off as an allowance ledger. We assign `relay_rank` across every verified relay in the trace before we know whether any sponsor existed for that relay timestamp, then only later check `sponsors.sponsor_timestamp <= relays.relay_timestamp`. That prevents direct future sponsorship, but ineligible rows still burn rank slots. For example, with allowance 100: generation A, 101 relays, generation B, then 100 relays means the 101st pre-B relay keeps rank 101 and one of the post-B relays becomes billable even though B should sponsor 100 future relays. It also means ClickHouse has to join every relay to every sponsor in a busy trace before the timestamp/allowance filters apply. Could we model this as a time-ordered sponsor allowance ledger instead, and add regressions for pre-sponsor relays plus over-allowance relays before a later sponsor?

richardsolomou: fixed this as a time-ordered allowance ledger! pre-sponsor and over-allowance relays stay billable without consuming later sponsor allowance, and the relay/sponsor cross join is gone. added the exact regression with pre-sponsor debt, exhausted allowance, and a later sponsor.

pawel-cebula (approved):
Looks good.

posthog/tasks/usage_report.py
```diff
@@ -1390,6 +1416,114 @@ def get_teams_with_ai_event_count_in_period(
...
+        # nosemgrep: clickhouse-fstring-param-audit - property document/table expressions are internal fragments
+        sponsored_counts = sync_execute(
+            f"""
+            SELECT
+                team_id,
+                countIf(
+                    entry_kind = 1
+                    AND ledger_timestamp >= %(begin)s
+                    AND ledger_timestamp < %(end)s
+                    AND cumulative_balance >= least(ifNull(previous_minimum, 0), 0)
```
brandonleung, line 1461: we should consider real traffic where a relayed span's timestamp could be earlier than the generation's timestamp. looks like we're sponsoring span's that come after a generation

maybe we need some bounded backdating grace/watermarking here

rust/capture/src/otel/filtering.rs
```diff
@@ -24,9 +25,22 @@ pub async fn check_quota(
...
+        if limiter
+            .is_quota_limited_v1(token, &QuotaResource::Events)
```
brandonleung, line 34: when over the global events limit, a signed gateway batch 400 while an unsigned batch goes through

verified should not be stricter than unverified

rust/capture/src/otel/provenance.rs
```diff
@@ -0,0 +1,181 @@
...
+        content_type,
+        content_encoding,
+        &body_digest,
+        &signed_at,
+    ]);
+    shared::verify(secret.as_bytes(), &message, &signature, &signed_at, now)
+}
+
+pub fn apply(span_events: &mut [SpanEvent], provenance: Provenance) {
+    let trusted = provenance == Provenance::Verified;
```
brandonleung, line 60: can you add a test here to check what happens fi a real signature is too old?

posthog/tasks/usage_report.py
```diff
@@ -1390,6 +1416,114 @@ def get_teams_with_ai_event_count_in_period(
...
+                                team_id,
+                                event,
+                                min(timestamp) AS relay_timestamp,
+                                {verified_expr} IN ('true', '1') AS verified,
+                                {relay_expr} IN ('true', '1') AS relay,
+                                if(verified AND relay, {trace_id_expr}, '') AS trace_id,
+                                if(verified AND relay, {span_id_expr}, '') AS span_id
+                            FROM {events_read_table(use_new)}
+                            WHERE team_id IN %(relayed_team_ids)s
+                              AND event IN %(ai_events)s
```
brandonleung, line 1545: this seems to be a very wide search range. can we add a filter to improve performance?

brandonleung (changes requested):

posthog/tasks/usage_report.py
richardsolomou: hmm good point. added a bounded five-minute backdating grace for trace events recorded before the gateway generation completes. spans inside the grace can use the trace allowance, while older relays remain billable. lmk if u think theres more we can do here

rust/capture/src/otel/filtering.rs
richardsolomou: fixed. verified relay batches now match direct AI ingestion at the global event limit while still bypassing the scoped AIO limiter. added an integration regression for the signed path.

rust/capture/src/otel/provenance.rs
richardsolomou: should be covered by rejects_stale_signature, it uses the real known-answer HMAC and verifies that the same valid signature becomes Stale once now moves past the five-minute window.

posthog/tasks/usage_report.py
richardsolomou: yeah good point, i added the verified and relay predicates to the inner event scans so unrelated rows are filtered before grouping and window evaluation. the bounded lookaround remains for cross-period accounting.

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1437,139 @@ def get_teams_with_ai_event_count_in_period(
...
+                                    request_id,
+                                    argMin(raw_trace_id, (raw_generation_timestamp, raw_trace_id)) AS trace_id,
+                                    min(raw_generation_timestamp) AS request_sponsor_timestamp
+                                FROM (
+                                    SELECT
+                                        team_id,
+                                        timestamp AS raw_generation_timestamp,
+                                        {verified_expr} IN ('true', '1') AS verified,
+                                        {relay_expr} IN ('true', '1') AS relay,
+                                        if(verified AND NOT relay, {trace_id_expr}, '') AS raw_trace_id,
```
brandonleung, line 1510: it doesn't seem like these sides come from the same place (relay's trace_id is generated by OTLP capture, generation's come from gateway request). does anything guarantee they're the same string?

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1437,139 @@ def get_teams_with_ai_event_count_in_period(
...
+                                GROUP BY team_id, request_id
+                            )
+                            GROUP BY team_id, trace_id
+                        )
+                        ARRAY JOIN [0, 1] AS allowance_kind
+                        UNION ALL
+                        SELECT
+                            team_id,
+                            trace_id,
+                            if(event = '$ai_evaluation', 1, 0) AS allowance_kind,
```
brandonleung, line 1529: can a relayed event ever be an ai evaluation? OTLP only produces spans, generations, and embeddings. can you check if this part ever fires and if the evaluation allowance isn't dead weight?

rust/capture/tests/integration_otel_endpoint.rs
```diff
@@ -844,6 +1058,95 @@ async fn test_quota_limit_exceeded_returns_400_with_no_events() {
...
+        .body(body)
+        .send()
+        .await;
+
+    assert_eq!(response.status().as_u16(), 400);
+    assert!(sink.get_events().await.is_empty());
+}
+
+#[tokio::test]
+async fn test_verified_gateway_batch_matches_direct_ai_global_quota_behavior() {
```
brandonleung, line 1325: set the LLM key here too so we only pass if there's a legitimate bypass

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1437,139 @@ def get_teams_with_ai_event_count_in_period(
...
+                            ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+                        ) AS cumulative_balance
+                    FROM (
+                        SELECT
+                            team_id,
+                            trace_id,
+                            allowance_kind,
+                            if(
+                                allowance_kind = 0,
+                                sponsor_timestamp - toIntervalSecond(%(backdate_seconds)s),
```
brandonleung: where is this 5 minutes coming from? does it need to be measured against live traffic or at least logged so we can adjut this later?

can we make sure that the same backdating is applied to all types?

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1437,139 @@ def get_teams_with_ai_event_count_in_period(
...
+                                    team_id,
+                                    request_id,
+                                    argMin(raw_trace_id, (raw_generation_timestamp, raw_trace_id)) AS trace_id,
+                                    min(raw_generation_timestamp) AS request_sponsor_timestamp
+                                FROM (
+                                    SELECT
+                                        team_id,
+                                        timestamp AS raw_generation_timestamp,
+                                        {verified_expr} IN ('true', '1') AS verified,
+                                        {relay_expr} IN ('true', '1') AS relay,
```
brandonleung, line 1509: these properties should be declared in `events_json.py` to make the filter cheaper (like trace_id and span_id)

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1437,139 @@ def get_teams_with_ai_event_count_in_period(
             ch_user=ClickHouseUser.BILLING,
         )
 
+        base_counts = [(team_id, count) for team_id, count, _ in base_rows]
+        relayed_team_ids = [team_id for team_id, _, has_verified_relay in base_rows if has_verified_relay]
+        if not relayed_team_ids:
+            return base_counts
+
+        # nosemgrep: clickhouse-fstring-param-audit - property document/table expressions are internal fragments
+        sponsored_counts = sync_execute(
```
brandonleung, line 1453: this runs on the update_all_orgs_billing_quotas which is every 15 minutes. is this intended? that's a lot of calls without a memory ceiling ... could see this causing an OOM issue

rust/capture/src/otel/filtering.rs
```diff
@@ -24,9 +24,18 @@ pub async fn check_quota(
     limiter: &CaptureQuotaLimiter,
     token: &str,
     span_events: &[SpanEvent],
+    gateway_verified: bool,
 ) -> Result<(), QuotaOutcome> {
+    let count = span_events.len();
+
+    if gateway_verified {
```
brandonleung, line 32: `ai_endpoint` and v1 shim still check the global events limiter for verified events; but this skips it. will leave to you what the right decision is here but it should be consistent

rust/capture/src/otel/provenance.rs
```diff
@@ -0,0 +1,181 @@
...
+        let Some(properties) = span_event.properties.as_object_mut() else {
+            continue;
+        };
+        properties.retain(|key, _| !key.starts_with(shared::GATEWAY_PREFIX));
+        if trusted {
+            properties.insert(shared::VERIFIED_PROPERTY.to_string(), Value::Bool(true));
+            properties.insert(RELAY_PROPERTY.to_string(), Value::Bool(true));
+        }
+    }
+    counter!(PROVENANCE_METRIC, "reason" => reason).increment(1);
```
brandonleung, line 84: every unsigned request would land here as `invalid` (most traffic), so this quickly gets drowned in noise

posthog/tasks/test/test_usage_report.py
```diff
@@ -5905,6 +5992,347 @@ def ai_count() -> int:
...
+        from posthog.tasks.usage_report import (
+            GATEWAY_SPONSORED_TRACE_EVENTS_PER_TRACE,
+            get_teams_with_ai_event_count_in_period,
+        )
+
+        def ai_count() -> int:
+            return dict(get_teams_with_ai_event_count_in_period(self.begin, self.end)).get(self.team.id, 0)
+
+        baseline_count = ai_count()
+        trace_id = "sponsored-trace"
```
brandonleung, line 5997: can we add a cae where a second team has the same trace id?

posthog/tasks/usage_report.py
brandonleung: also this query is not in the snapshots because fixtures don't et ai_gateway_relay

docs/onboarding/ai-observability/convex.tsx
```diff
@@ -82,12 +82,10 @@ export const getConvexSteps = (ctx: OnboardingComponentsContext): StepDefinition
...
                               }),
                               spanProcessors: [
-                                new SimpleSpanProcessor(
-                                  new PostHogTraceExporter({
-                                    projectToken: process.env.POSTHOG_PROJECT_TOKEN!,
-                                    host: process.env.POSTHOG_HOST,
-                                  })
-                                ),
+                                new PostHogSpanProcessor({
+                                  projectToken: process.env.POSTHOG_PROJECT_TOKEN!,
```
brandonleung, line 86: please verify if the span process takes project token or api key

brandonleung (changes requested):
please make sure we're handling edges consistently. also not sure how the RFC you linked relates

posthog/tasks/usage_report.py
richardsolomou: the gateway adopts the caller's traceparent trace ID for the generation, and capture extracts that same ID from the relayed OTLP batch. ive also verified the join through the full E2E flow.

posthog/tasks/usage_report.py
richardsolomou: the upstack evaluation PR creates verified relayed $ai_evaluation events from OTLP logs, so this allowance becomes active when the stack lands.

rust/capture/tests/integration_otel_endpoint.rs
richardsolomou: done, the signed-path test now sets both the scoped LLM and global event quota keys so it only passes through the intended verified bypass.

posthog/tasks/usage_report.py
richardsolomou: done, the five-minute grace is tied to the gateway provider timeout and now applies consistently to both RFC allowance ledgers.

posthog/tasks/usage_report.py
richardsolomou: done, the gateway verified, relay, and request ID properties are declared native JSON subcolumns now.

posthog/tasks/usage_report.py
richardsolomou: done, billing queries now have a hard 4 GiB ClickHouse memory ceiling and the sponsorship query stays covered by its query-shape tests and CI snapshot generation.

rust/capture/src/otel/filtering.rs
richardsolomou: done, verified OTLP traffic now performs the same global quota lookup while bypassing only the scoped LLM restriction.

rust/capture/src/otel/provenance.rs
richardsolomou: done, missing signatures are recorded as unsigned separately from malformed invalid signatures.

posthog/tasks/test/test_usage_report.py
richardsolomou: done, added a second-team regression using the same trace and span IDs to prove sponsorship remains team-scoped.

docs/onboarding/ai-observability/convex.tsx
richardsolomou: verified, this example uses direct capture where the processor takes projectToken. projectSecret is only accepted inside the explicit aiGateway option.

jose-sequeira (approved):

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1440,135 @@ def get_teams_with_ai_event_count_in_period(
...
+                                event,
+                                min(timestamp) AS relay_timestamp,
+                                {verified_expr} IN ('true', '1') AS verified,
+                                {relay_expr} IN ('true', '1') AS relay,
+                                if(verified AND relay, {trace_id_expr}, '') AS trace_id,
+                                if(verified AND relay, {span_id_expr}, '') AS span_id
+                            FROM {events_read_table(use_new)}
+                            WHERE team_id IN %(relayed_team_ids)s
+                              AND event IN %(ai_events)s
+                              AND timestamp >= %(sponsor_begin)s AND timestamp < %(sponsor_end)s
```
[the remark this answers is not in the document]
richardsolomou: fixed. the relay scan now starts one backdating interval before sponsor_begin, so grace-window relays consume the same allowance even when they sit just outside the normal lookaround. added the boundary regression too.

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1440,135 @@ def get_teams_with_ai_event_count_in_period(
...
+                            PARTITION BY team_id, trace_id, allowance_kind
+                            ORDER BY ledger_timestamp, entry_kind, entry_id
+                            ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+                        ) AS cumulative_balance
+                    FROM (
+                        SELECT
+                            team_id,
+                            trace_id,
+                            allowance_kind,
+                            sponsor_timestamp - toIntervalSecond(%(backdate_seconds)s) AS ledger_timestamp,
```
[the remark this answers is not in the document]
richardsolomou: fixed. the relay scan now starts one backdating interval before sponsor_begin, so grace-window relays consume the same allowance even when they sit just outside the normal lookaround. added the boundary regression too.

posthog/clickhouse/events_json.py
```diff
@@ -33,6 +33,9 @@ def _nullable_json_subcolumn_types(subcolumns: dict[str, str]) -> dict[str, str]
 EVENTS_PROPERTIES_JSON_SUBCOLUMN_DECLARED_TYPES: dict[str, str] = {
     "$active_feature_flags": "Array(String)",
     "$ai_experiment_id": "Nullable(String)",
+    "$ai_gateway_relay": "Nullable(Bool)",
+    "$ai_gateway_request_id": "Nullable(String)",
+    "$ai_gateway_verified": "Nullable(Bool)",
```
brandonleung: should 36 and 38 be `nullable(string)`? nullable bools wouldn't play nice with `IN ('true','1')` checks in clickhouse

posthog/tasks/usage_report.py
brandonleung: i went and looked: `resolveTraceID` on main only reads the trace id header and otherwise mints a random id, and on the relay branch the header still takes precedent over trace parent. 

a caller sending both wouldn't get sponsorship. should we flip that precedence or require they agree? up to you

rust/capture/src/otel/provenance.rs
```diff
@@ -0,0 +1,182 @@
...
+    };
+    let Some(signature) = shared::header_str(headers, shared::SIGNATURE_HEADER) else {
+        return Provenance::Invalid;
+    };
+    let Some(signed_at) = shared::header_str(headers, shared::SIGNED_AT_HEADER) else {
+        return Provenance::Invalid;
+    };
+
+    let body_digest = hex::encode(Sha256::digest(body));
+    let message = shared::canonical(&[
```
brandonleung, line 42: the otel tuple drops the request_id that v1 signs - a retried batch gets double counted and we mint a fresh uuid per event but the ledger only debits once. let's stamp it here like v1 does

posthog/tasks/usage_report.py
```diff
@@ -101,6 +107,7 @@ class TableSizes(TypedDict):
 
 CH_BILLING_SETTINGS = {
     "max_execution_time": 5 * 60,  # 5 minutes
+    "max_memory_usage": 4 * 1024 * 1024 * 1024,
```
brandonleung, line 114: this is on the shared dict, so all other billing queries get the ceiling and not just the new one. let's scope it to this query?

posthog/tasks/usage_report.py
richardsolomou: oopsie, my b, youre right, the explicit trace header could override a different `traceparent` and split the generation from its OTLP records. updating ai-gateway#340 to now give remote W3C context precedence, so both sides keep the same trace ID.

posthog/tasks/usage_report.py
richardsolomou: confirmed this against both branches. ai-gateway #340 relays `gen_ai.evaluation.result` logs, and #71937 uses `$ai_evaluation_run_id` as their replay-stable ledger identity. this becomes active when the full stack lands.

posthog/clickhouse/events_json.py
richardsolomou: yep, these need to match the string extraction used by the billing query. both gateway markers are now declared as `Nullable(String)` and the schema snapshots are updated.

rust/capture/src/otel/provenance.rs
richardsolomou: youre right. the gateway now adds its request ID to the signed OTLP tuple, and capture requires that signed header before stamping it. relay retries are deduplicated by stable span IDs and evaluation run IDs, while relay request IDs are excluded from the direct exemption.

posthog/tasks/usage_report.py
richardsolomou: yep, the memory ceiling was leaking into every billing query through the shared settings. its now scoped to the gateway sponsorship query only.

posthog/tasks/usage_report.py
```diff
@@ -1378,26 +1389,50 @@ def get_teams_with_ai_event_count_in_period(
...
-            -- Perf: extract request_id only for verified rows (normally ≈0), so
-            -- non-gateway rows pay just the verified-bool extraction, not a string
-            -- extraction each, across the full AI event volume.
+            -- Perf: extract request_id and relay only for verified rows (normally ≈0),
+            -- so non-gateway rows pay just the verified-bool extraction, not string
+            -- extractions across the full AI event volume.
             SELECT
                 team_id,
-                COUNT() - uniqExactIf(request_id, verified AND request_id != '') as count
+                COUNT() - uniqExactIf(request_id, verified AND request_id != '') AS count,
```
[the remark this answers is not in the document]
richardsolomou: good catch. relay request IDs are now excluded from the direct gateway exemption, so they only receive the bounded ledger deduction. the sponsorship test now gives every relayed batch a signed request ID and keeps the unmatched one billable.

rust/capture/tests/integration_otel_endpoint.rs
```diff
@@ -227,6 +231,69 @@ async fn send_request_with_client(client: &TestClient, request: &ExportTraceServ
...
+    let mut message = Vec::new();
+    for field in fields {
+        message.extend_from_slice(&(field.len() as u32).to_be_bytes());
+        message.extend_from_slice(field.as_bytes());
+    }
+
+    let mut mac = Hmac::<Sha256>::new_from_slice(secret.as_bytes()).unwrap();
+    mac.update(&message);
+    hex::encode(mac.finalize().into_bytes())
+}
```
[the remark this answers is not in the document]
richardsolomou: yep, the integration helper was still signing the old tuple. ive added the request ID to the canonical message and request headers, and all six focused gateway endpoint tests pass against the production verifier now.

posthog/tasks/usage_report.py
```diff
@@ -1409,6 +1444,136 @@ def get_teams_with_ai_event_count_in_period(
...
+                                SELECT
+                                    team_id,
+                                    request_id,
+                                    argMin(raw_trace_id, (raw_generation_timestamp, raw_trace_id)) AS trace_id,
+                                    min(raw_generation_timestamp) AS request_sponsor_timestamp
+                                FROM (
+                                    SELECT
+                                        team_id,
+                                        timestamp AS raw_generation_timestamp,
+                                        {verified_expr} IN ('true', '1') AS verified,
```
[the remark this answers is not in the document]
richardsolomou: i checked both legacy paths and this is already fail-closed. `/e` and `/batch` call `apply_gateway_provenance`, while the older AI endpoint has the same strip step in `ai_endpoint.rs`; both remove the full `$ai_gateway*` namespace from unsigned AI events. there are unit and HTTP integration cases covering a forged `$ai_gateway_verified` marker too, so these rows cant reach the billing query.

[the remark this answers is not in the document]
richardsolomou: youre right, i traced the wrong processor before. `/e` and `/batch` go through the legacy analytics path, which wasnt stripping these properties. ive added fail-closed stripping for the full `$ai_gateway*` namespace before serialization, with regressions covering both routes.

posthog/tasks/test/__snapshots__/test_usage_report.ambr
```diff
@@ -1491,12 +1491,16 @@
   
   SELECT team_id,
```
brandonleung, line 1492: can you regenerate snapshots?

posthog/clickhouse/events_json.py
```diff
@@ -33,6 +33,9 @@ def _nullable_json_subcolumn_types(subcolumns: dict[str, str]) -> dict[str, str]
 EVENTS_PROPERTIES_JSON_SUBCOLUMN_DECLARED_TYPES: dict[str, str] = {
     "$active_feature_flags": "Array(String)",
     "$ai_experiment_id": "Nullable(String)",
+    "$ai_gateway_relay": "Nullable(String)",
+    "$ai_gateway_request_id": "Nullable(String)",
+    "$ai_gateway_verified": "Nullable(String)",
```
brandonleung: do we need a migration-only PR for this to land first?

rust/capture/src/otel/filtering.rs
```diff
@@ -24,9 +25,20 @@ pub async fn check_quota(
...
+        // Match direct AI ingestion's global lookup while bypassing only the
+        // scoped LLM restriction. Direct AI events are retained at this limit.
+        limiter
+            .is_quota_limited_v1(token, &QuotaResource::Events)
```
brandonleung, line 34: we're discarding the global quota answer but `i/v0/ai` rejects on it

brandonleung (approved):
small comments. no need to re-request review

posthog/tasks/test/__snapshots__/test_usage_report.ambr
richardsolomou: yep, regenerated this through the ClickHouse-backed test. all 86 snapshots passed and there was no additional diff.

posthog/clickhouse/events_json.py
richardsolomou: yeah, split the declarations and ALTER into migration-only PR #78109. ive removed the declarations from this branch so the schema change can land first.

rust/capture/src/otel/filtering.rs
richardsolomou: youre right, verified relays should only bypass the scoped LLM quota. the global events result now rejects the batch like `/i/v0/ai`, with the drop metric preserved.

PR #71937, opened 2026-07-17
feat(aio): ingest OpenTelemetry evaluation results

richardsolomou:
## Problem

Externally computed evaluations sent through OpenTelemetry do not appear alongside AI traces. Compact OTLP batches also need bounded decode, expansion, and Kafka enqueue work.

## Changes

- Import valid `gen_ai.evaluation.result` logs and attach them to target spans.
- Return OTLP partial success for rejected AI telemetry while preserving full success for unrelated records.
- Count ingested events after quota and restriction filtering.
- Bound structural entries, candidate records, and expanded bytes, including discarded records.
- Keep wire-owned IDs fail-closed across trace and log ingestion.

- The OTLP endpoint integration tests cover JSON and protobuf responses, partial filtering, and mixed telemetry batches.
- The focused billing-query tests cover evaluation sponsorship accounting.
- The focused frontend utility tests cover imported and native evaluation rendering.
- Rust formatting and Clippy checks passed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Documented required OTLP evaluation fields, span targeting, and the imported-evaluation investigation workflow.

## 🤖 Agent context

PostHog Code implemented evaluation ingestion and rebased it after the lower stack layer merged. `/stacking-prs`, `/clickhouse-migrations`, `/writing-tests`, `/writing-code-comments`, `/writing-user-facing-copy`, `/rs-rebase`, `/rs-address-pr-review`, and `/writing-pr-descriptions` shaped this update.

---


richardsolomou:
* **#71937** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/71937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#71936** <a href="https://app.graphite.com/github/pr/PostHog/posthog/71936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,415 @@
...
+        .sum()
+}
+
+pub fn expand_into_events(
+    request: &ExportLogsServiceRequest,
+    request_fallback_distinct_id: &str,
+) -> Result<Vec<SpanEvent>, CaptureError> {
+    let mut events =
+        Vec::with_capacity(count_records(request).min(super::MAX_AI_EVENTS_PER_REQUEST));
+    let mut expanded_bytes = 0usize;
```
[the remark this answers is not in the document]
richardsolomou: i think this is already bounded before conversion: prost rejects nested messages past its DecodeContext recursion limit, and serde_json applies its own recursion limit for JSON. any_value_to_json only sees successfully decoded values, so I dont think the unbounded stack-overflow path is reachable here.

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,415 @@
...
+                    continue;
+                };
+                if events.len() == super::MAX_AI_EVENTS_PER_REQUEST {
+                    return Err(CaptureError::RequestParsingError(format!(
+                        "Too many evaluation events: {} exceeds limit of {}",
+                        events.len() + 1,
+                        super::MAX_AI_EVENTS_PER_REQUEST
+                    )));
+                }
+                let event_bytes = serde_json::to_vec(&event.properties)
```
pawel-cebula, line 123: I think this logs path needs a per-event Kafka-size guard before `send_batch`. Right now `expand_into_events` only checks cumulative expanded properties against the 8 MiB OTLP cap, but the main producer defaults `message.max.bytes` to 1,000,000 and `send_batch` can enqueue earlier records before returning `EventTooBig` for a later oversized record. That gives clients a 413 for the whole OTLP logs export while a prefix of the batch may already exist in Kafka. Could we reject any expanded evaluation that can exceed the effective producer message ceiling before enqueueing the batch, and cover it with a producer/mock test rather than `CapturingSink`?

products/ai_observability/skills/exploring-llm-evaluations/SKILL.md
```diff
@@ -59,12 +59,14 @@ Every run of an evaluation emits an `$ai_evaluation` event. Key properties:
 
 | Property                     | Meaning                                                         |
 | ---------------------------- | --------------------------------------------------------------- |
-| `$ai_evaluation_id`          | UUID of the evaluation config                                   |
+| `$ai_evaluation_id`          | Evaluation config UUID or deterministic imported evaluator UUID |
```
pawel-cebula, line 62: Now that this skill documents imported evaluator IDs, the workflow below still assumes every `$ai_evaluation_id` maps to a native `Evaluation` row. `llma-evaluation-list` won't find an imported deterministic ID, and `llma-evaluation-summary-create` 404s unless `Evaluation.objects.get(...)` succeeds. The verification SQL is also boolean/generation-only, so it misses label/number imported results and span-targeted runs. Could we split the skill into native-configured evaluations vs imported OTLP evaluations, and route imported ones through `execute-sql` over `$ai_evaluation_type = 'imported'` / `$ai_evaluation_runtime = 'otel'` with `$ai_evaluation_run_id`, `$ai_trace_id`, and `$ai_target_span_id`?

docs/onboarding/ai-observability/opentelemetry.tsx
```diff
@@ -261,7 +261,7 @@ export const getOpenTelemetrySteps = (ctx: OnboardingComponentsContext): StepDef
...
+                            **Direct OTLP export.** If you run an OpenTelemetry Collector, or want to export from a language that isn't Python or Node.js, point any OTLP/HTTP exporter directly at PostHog's AI ingestion endpoint. PostHog accepts OTLP traces and logs over HTTP in both \`application/x-protobuf\` and \`application/json\`, authenticated with a \`Bearer\` token. Exporters append \`/v1/traces\` and \`/v1/logs\` to the base endpoint. Log EventRecords named \`gen_ai.evaluation.result\` become \`$ai_evaluation\` events. PostHog groups records with the same evaluation name as one imported evaluator while retaining a separate replay-stable identity for each record.
```
pawel-cebula: The direct OTLP section says `gen_ai.evaluation.result` logs become `$ai_evaluation` events, but the parser has a stricter contract than that: it drops records without `gen_ai.evaluation.name`, and without either `gen_ai.evaluation.score.label` or a finite `gen_ai.evaluation.score.value`. It also only attaches the imported run to a target span when the OTLP log has valid `trace_id` and `span_id`. Could we add a minimal log example and call those required fields out? Otherwise users can get a successful export with no imported evals and not have much to debug from.

richardsolomou: fixed! the direct OTLP section now calls out the required evaluation name and label/finite numeric score, explains the trace/span ID requirement for span targeting, and includes a minimal JSON log example.

products/ai_observability/skills/exploring-llm-evaluations/SKILL.md
richardsolomou: fixed. the skill now separates native evaluations from imported OTLP evaluations and routes imported runs through execute-sql using the imported/runtime markers, run ID, trace ID, target span ID, result type, and result.

rust/capture/src/otel/logs.rs
richardsolomou: fixed this in the Kafka sink rather than only the OTLP logs path, lmk if you disagree with that!

send_batch now checks every prepared record against message.max.bytes before enqueueing anything, including payload, key, headers, and framing. added a producer-mock test where the final record is oversized and confirmed zero records are enqueued.

pawel-cebula (approved):
Looks good.

brandonleung (approved):
might be worth doing a pass over the existing tasks and workflow logic branches to make sure we have good coverage but looks good

richardsolomou:
@brandonleung i did a pass over the existing branches and added explicit regressions for both LLM judge task and Hog workflow evaluations. they confirm the existing online boolean result mapping stays unchanged alongside the new imported OTLP paths, so we should be good

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,6 +98,322 @@ pub fn parse_request(
...
+) -> Result<(), CaptureError> {
+    while resource_logs.has_remaining() {
+        let (tag, wire_type) =
+            prost::encoding::decode_key(resource_logs).map_err(invalid_logs_protobuf)?;
+        if tag == 2 && wire_type == prost::encoding::WireType::LengthDelimited {
+            increment_log_node_count(node_count, node_limit)?;
+            let mut scope_logs = take_message(resource_logs).map_err(invalid_logs_protobuf)?;
+            count_scope_log_records(&mut scope_logs, node_count, node_limit)?;
+        } else {
+            prost::encoding::skip_field(
```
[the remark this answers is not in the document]
richardsolomou: fixed. the protobuf and JSON preflights now count resource/scope/log attributes plus nested array and key-value-list elements before full decode. added both wire-format regressions for each amplification path.

rust/capture/src/otel/mod.rs
```diff
@@ -26,14 +28,18 @@ use crate::token::validate_token;
...
 
-/// Maximum raw spans accepted before filtering. Set well above MAX_SPANS_PER_REQUEST
+/// Maximum raw spans accepted before filtering. Set well above
+/// MAX_AI_EVENTS_PER_REQUEST
 /// to accommodate mixed-content batches (e.g. Next.js sending HTTP + AI spans together)
 /// while still bounding the cost of attribute scanning.
-const MAX_RAW_SPANS_PER_REQUEST: usize = 1000;
+const MAX_RAW_OTEL_SPANS_PER_REQUEST: usize = 1000;
+const MAX_RAW_OTEL_LOG_RECORDS_PER_REQUEST: usize = 1000;
+const MAX_RAW_OTEL_LOG_NODES_PER_REQUEST: usize = 3000;
```
brandonleung, line 61: where are these constants coming from? can you see if the two values above are tuned properly now that we're counting attributes too?

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,12 +98,401 @@ pub fn parse_request(
...
+    node_count: &mut usize,
+    node_limit: usize,
+) -> Result<(), CaptureError> {
+    increment_log_node_count(node_count, node_limit)?;
+    let mut key_value = take_message(message).map_err(invalid_logs_protobuf)?;
+    while key_value.has_remaining() {
+        let (tag, wire_type) =
+            prost::encoding::decode_key(&mut key_value).map_err(invalid_logs_protobuf)?;
+        if tag == 2 && wire_type == prost::encoding::WireType::LengthDelimited {
+            increment_log_node_count(node_count, node_limit)?;
```
brandonleung: seem like we have two definitions of what a node is. let's pick one and be consistent

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,12 +98,401 @@ pub fn parse_request(
...
+        let (tag, wire_type) =
+            prost::encoding::decode_key(record).map_err(invalid_logs_protobuf)?;
+        match (tag, wire_type) {
+            (5, prost::encoding::WireType::LengthDelimited) => {
+                increment_log_node_count(node_count, node_limit)?;
+                let mut body = take_message(record).map_err(invalid_logs_protobuf)?;
+                count_any_value_nodes(&mut body, node_count, node_limit)?;
+            }
+            (6, prost::encoding::WireType::LengthDelimited) => {
+                count_key_value_field(record, node_count, node_limit)?;
```
brandonleung: test this branch please

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,12 +98,401 @@ pub fn parse_request(
...
+                .map_err(serde::de::Error::custom)?;
+        }
+        Ok(())
+    }
+
+    fn visit_map<A>(self, mut map: A) -> Result<(), A::Error>
+    where
+        A: MapAccess<'de>,
+    {
+        while map.next_key::<String>()?.is_some() {
```
brandonleung, line 468: JSON counts array elements but not object keys so a body of flat keys would score zero here

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,12 +98,401 @@ pub fn parse_request(
...
+        tag,
+        message,
+        prost::encoding::DecodeContext::default(),
+    )
+    .map_err(invalid_logs_protobuf)
+}
+
+fn take_message<'a>(buf: &mut &'a [u8]) -> Result<&'a [u8], prost::DecodeError> {
+    let len = prost::encoding::decode_varint(buf)? as usize;
+    if len > buf.len() {
```
brandonleung, line 361: can you add tests for this (non-happy flow cases)

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,415 @@
...
+        );
+        properties.insert(
+            "$ai_evaluation_result_type".to_string(),
+            Value::String("label".to_string()),
+        );
+    }
+    if let Some(value) = result_value {
+        if let Some(number) = serde_json::Number::from_f64(value) {
+            if result_label.is_none() {
+                properties.insert("$ai_evaluation_result".to_string(), Value::Number(number));
```
brandonleung, line 264: there's some type messiness here: if a team passes in a numeric value here, we don't have anything to protect us here

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,415 @@
...
+    if record.span_id.len() == 8 {
+        properties.insert(
+            "$ai_target_span_id".to_string(),
+            Value::String(hex::encode(&record.span_id)),
+        );
+    }
+    if let Some(label) = result_label.as_deref() {
+        properties.insert(
+            "$ai_evaluation_result".to_string(),
+            Value::String(label.to_string()),
```
brandonleung, line 254: there are a lot of consumers that classify this property against true or "true". can we make sure we aren't rendering imported results as unknown or throw?

rust/capture/src/sinks/kafka.rs
```diff
@@ -441,7 +460,19 @@ impl<P: KafkaProducer> KafkaSinkBase<P> {
             producer: Arc::new(producer),
             topics: Arc::new(topics),
             replay_envelope_compression,
+            message_max_bytes: usize::MAX,
+        }
+    }
+
+    fn validate_record_size(&self, record: &ProduceRecord) -> Result<(), CaptureError> {
```
brandonleung: we'd be discarding the entire analytic batch rather than just the event itself on an over-1mb event

rust/capture/src/sinks/producer.rs
```diff
@@ -20,6 +24,25 @@ pub struct ProduceRecord {
     pub headers: CapturedEventHeaders,
 }
 
+impl ProduceRecord {
+    /// Returns a conservative upper bound for librdkafka's encoded message size.
+    pub(crate) fn max_wire_size(&self) -> usize {
+        let key_size = self.key.as_ref().map_or(0, String::len);
+        let headers: rdkafka::message::OwnedHeaders = self.headers.clone().into();
```
brandonleung: is there a better way to do this? seems like a lot of work just to sum key and value lengths

rust/capture/src/router.rs
```diff
@@ -392,6 +392,14 @@ pub fn router<TZ: TimeSource + Send + Sync + 'static, R: Client + Send + Sync +
             "/i/v0/ai/otel/",
             post(otel::otel_handler).options(otel::options),
         )
+        .route(
+            "/i/v0/ai/otel/v1/traces",
```
brandonleung, line 381: should we test the alias here?

posthog/tasks/usage_report.py
```diff
@@ -1536,16 +1543,20 @@ def get_teams_with_ai_event_count_in_period(
...
                             WHERE team_id IN %(relayed_team_ids)s
                               AND event IN %(ai_events)s
                               AND timestamp >= %(sponsor_begin)s AND timestamp < %(sponsor_end)s
                               AND {verified_expr} IN ('true', '1')
                               AND {relay_expr} IN ('true', '1')
-                            GROUP BY team_id, event, verified, relay, trace_id, span_id
+                            GROUP BY team_id, event, verified, relay, trace_id, relay_id
                         )
-                        WHERE verified AND relay AND trace_id != '' AND span_id != ''
+                        WHERE verified AND relay AND trace_id != '' AND relay_id != ''
```
brandonleung, line 1660: introduces two holes

1. trace-less evaluations skip quota and get dropped
2. non-OTLP relayed evaluations would also lose sponsorship

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,415 @@
...
+            if result_label.is_none() {
+                properties.insert("$ai_evaluation_result".to_string(), Value::Number(number));
+                properties.insert(
+                    "$ai_evaluation_result_type".to_string(),
+                    Value::String("number".to_string()),
+                );
+            }
+        }
+    }
+    if let Some(explanation) = properties.get(EXPLANATION).and_then(Value::as_str) {
```
brandonleung: a resource-level explanation would get stamped on every evaluation. can we be consistent and read from attributes like the similar flows?

brandonleung (changes requested):
aside from my comments, whenever we discard a whole OTLP batch we should be very careful that the whole batch needs to be dropped. if it HAS to be dropped then let's make sure it visible and not dropped quietly

rust/capture/src/otel/mod.rs
richardsolomou: done, documented why the structural limit is separate from the 4 MiB wire limit and added boundary coverage for resource, record, and nested attributes.

rust/capture/src/otel/ingestion.rs
richardsolomou: done, protobuf and JSON now count each attribute or collection entry once. the extra protobuf AnyValue wrapper increment is gone.

rust/capture/src/otel/ingestion.rs
richardsolomou: done, added direct protobuf coverage for oversized log-record attributes.

rust/capture/src/otel/ingestion.rs
richardsolomou: done, JSON map entries count against the same structural budget now, with a flat-object regression.

rust/capture/src/otel/ingestion.rs
richardsolomou: done, added malformed length-delimited protobuf coverage for the buffer-underflow path.

rust/capture/src/otel/logs.rs
richardsolomou: numeric results are already handled separately through Number::from_f64, while labels only accept strings, so malformed numeric values cant reach the event. the numeric-result regression covers that contract.

rust/capture/src/otel/logs.rs
richardsolomou: done, imported label and numeric results keep explicit result types, while the existing task and workflow boolean branches remain covered.

rust/capture/src/sinks/kafka.rs
richardsolomou: agreed that one oversized analytic event shouldnt discard unrelated events. validation still happens before enqueue, but oversized records are reported and removed before valid siblings are sent.

rust/capture/src/sinks/producer.rs
richardsolomou: this intentionally covers more than key + payload because Kafka's limit includes headers and record framing too. the conservative bound keeps librdkafka from rejecting a later record after earlier ones were queued.

rust/capture/src/router.rs
richardsolomou: done, added integration coverage for the trailing-slash logs alias.

posthog/tasks/usage_report.py
richardsolomou: done, trace-less evaluations no longer receive the signed quota shortcut, and sponsorship falls back to the stable span ID when an evaluation run ID isnt present.

rust/capture/src/otel/logs.rs
richardsolomou: done, evaluation reasoning now comes only from log-record attributes, so resource-level explanations arent stamped onto every evaluation.

rust/capture/src/sinks/producer.rs
jose-sequeira: Why do we need this instead of just acknowledging the kafka error when producing? Would this run on every pipeline? 

This seems unnecessary overhead for most of the current cases, if you need this specifically for your use case I'd implement it there and not on the kafka sink

rust/capture/src/sinks/kafka.rs
```diff
@@ -224,6 +225,7 @@ pub struct KafkaSinkBase<P: KafkaProducer> {
     producer: Arc<P>,
     topics: Arc<KafkaTopicConfig>,
     replay_envelope_compression: EnvelopeCompression,
+    message_max_bytes: usize,
```
jose-sequeira: I would refrain from adding this to the kafka sink itself, specifically when not set as optional

rust/capture/src/sinks/kafka.rs
richardsolomou: moved this out of the Kafka sink. OTLP evaluation expansion now reports and removes only the oversized evaluation before valid siblings reach Kafka.

rust/capture/src/sinks/producer.rs
richardsolomou: yeah agreed, this doesnt belong on every Kafka pipeline. the per-event guard now lives in OTLP evaluation expansion and the generic sink/producer preflight is removed.

rust/capture/src/sinks/kafka.rs
richardsolomou: done, this is no longer part of KafkaSinkBase. only OTLP evaluation expansion applies the per-event guard now.

rust/capture/src/otel/ingestion.rs
```diff
@@ -95,12 +98,402 @@ pub fn parse_request(
...
+    array: &mut &[u8],
+    node_count: &mut usize,
+    node_limit: usize,
+) -> Result<(), CaptureError> {
+    while array.has_remaining() {
+        let (tag, wire_type) = prost::encoding::decode_key(array).map_err(invalid_logs_protobuf)?;
+        if tag == 1 && wire_type == prost::encoding::WireType::LengthDelimited {
+            increment_log_node_count(node_count, node_limit)?;
+            let mut value = take_message(array).map_err(invalid_logs_protobuf)?;
+            count_any_value_nodes(&mut value, node_count, node_limit)?;
```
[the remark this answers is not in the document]
richardsolomou: fixed with an explicit 64-level nesting limit in the streaming preflight, separate from the total structural-entry budget. added protobuf and JSON boundary regressions so deeply nested values are rejected before full decode.

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,471 @@
...
+    let explanation = attributes
+        .get(EXPLANATION)
+        .and_then(Value::as_str)
+        .map(str::to_string);
+    attributes.remove(SCORE_LABEL);
+    attributes.remove(SCORE_VALUE);
+
+    let distinct_id =
+        extract_distinct_id_for_span(&record.attributes, resource, request_fallback_distinct_id);
+    let mut properties = resource_attributes.clone();
```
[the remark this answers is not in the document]
richardsolomou: fixed by sizing the shared resource map once before expanding its records. if the resource alone exceeds the per-event limit, the whole resource group is counted and skipped before any per-record cloning, while valid resource groups in the same request still process.

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,509 @@
...
+        .filter(|value| value.is_finite());
+    if result_label.is_none() && result_value.is_none() {
+        return None;
+    }
+    let explanation = attributes
+        .get(EXPLANATION)
+        .and_then(Value::as_str)
+        .map(str::to_string);
+    attributes.remove(SCORE_LABEL);
+    attributes.remove(SCORE_VALUE);
```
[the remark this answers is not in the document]
richardsolomou: fixed. the raw explanation attribute is removed after normalization, so the event only carries $ai_evaluation_reasoning. added a regression for this.

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,511 @@
...
+                        events.len() + 1,
+                        super::MAX_AI_EVENTS_PER_REQUEST
+                    )));
+                }
+                let event_bytes = serde_json::to_vec(&event.properties)
+                    .map_err(|error| CaptureError::InternalError(error.to_string()))?
+                    .len()
+                    + event.event_name.len()
+                    + event.distinct_id.len();
+                if event_bytes > super::MAX_EXPANDED_AI_EVENT_BYTES_PER_EVENT {
```
brandonleung, line 128: a record rejected here charges nothing to expanded_bytes or the event count, but we've already paid the resource clone and serialize. let's fire something on a body that we reject

rust/capture/src/otel/mod.rs
```diff
@@ -196,10 +209,204 @@ pub async fn otel_handler(
...
+        logs::expand_into_events(&request, &request_fallback_distinct_id).map_err(|err| {
+            report_internal_error_metrics(err.to_metric_tag(), "otel_logs_validation");
+            err.into_response()
+        })?;
+    provenance::apply(
+        &mut events,
+        gateway_provenance,
+        headers.contains_key(crate::gateway_provenance::SIGNATURE_HEADER),
+    );
+    let quota_provenance = if events.iter().all(|event| {
```
brandonleung: one trace-less eval downgrades the whole batch here. can this be per-event? we're also stamping verified on that record before this runs

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,511 @@
...
+    properties.insert(
+        "$ai_evaluation_type".to_string(),
+        Value::String(EVALUATION_TYPE.to_string()),
+    );
+    properties.insert(
+        "$ai_ingestion_source".to_string(),
+        Value::String("otel".to_string()),
+    );
+
+    if record.trace_id.len() == 16 {
```
brandonleung, line 239: a client-set ai_trace_id survives here when the wire id is malformed. can we stamp unconditionally like the traces path?

jose-sequeira (approved):
Approved from the ingestion side

rust/capture/src/otel/logs.rs
richardsolomou: fixed on the current head. numeric results only survive `as_f64`, the finite check, and `Number::from_f64`; anything else is filtered before we build the evaluation event.

rust/capture/src/sinks/kafka.rs
richardsolomou: fixed on the current head. the batch-level size precheck is gone, so one oversized record follows the producer error path without rejecting its valid siblings during preparation.

rust/capture/src/sinks/producer.rs
richardsolomou: fixed on the current head by removing the custom wire-size calculation with the batch-level precheck that needed it.

rust/capture/src/otel/logs.rs
richardsolomou: evaluation candidates now consume the record budget before semantic parsing, so a body full of malformed evaluation records gets rejected after the same bounded count as valid ones.

rust/capture/src/otel/mod.rs
richardsolomou: this is per event now. provenance only stamps evaluations with a valid wire trace ID, and quota checking splits verified and ordinary events inside the batch so a trace-less sibling cant downgrade the rest.

rust/capture/src/otel/logs.rs
richardsolomou: fixed. `$ai_trace_id` is removed from the merged attributes first, then restored only from a valid 16-byte wire trace ID, matching the traces path.

rust/capture/tests/integration_otel_endpoint.rs
```diff
@@ -533,6 +603,270 @@ async fn test_signed_json_body_mismatch_rejects_trusted_provenance() {
...
+            DEFAULT_TEST_TIME,
+            "otel-logs-v1",
+        );
+        let mut request = client
+            .post(LOGS_ENDPOINT)
+            .header("Content-Type", content_type)
+            .header("Authorization", format!("Bearer {TOKEN}"))
+            .header("PostHog-Ai-Gateway-Signature", signature)
+            .header("PostHog-Ai-Gateway-Signed-At", DEFAULT_TEST_TIME);
+        if !content_encoding.is_empty() {
```
brandonleung, line 684: verify bails early if we don't send the header. same further down this file. can we make sure our tests are passing for the right reasons?

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,552 @@
...
+    );
+
+    properties.remove("$ai_trace_id");
+    if record.trace_id.len() == 16 {
+        properties.insert(
+            "$ai_trace_id".to_string(),
+            Value::String(hex::encode(&record.trace_id)),
+        );
+    }
+    if record.span_id.len() == 8 {
```
brandonleung, line 245: client set `ai_target_span_id` would survive malformed span id. same problem for span id and parent id. please sweep when solving

rust/capture/src/otel/logs.rs
```diff
@@ -0,0 +1,552 @@
...
+                        events.len() + 1,
+                        super::MAX_AI_EVENTS_PER_REQUEST
+                    )));
+                }
+                let event_bytes = serde_json::to_vec(&event.properties)
+                    .map_err(|error| CaptureError::InternalError(error.to_string()))?
+                    .len()
+                    + event.event_name.len()
+                    + event.distinct_id.len();
+                if event_bytes > super::MAX_EXPANDED_AI_EVENT_BYTES_PER_EVENT {
```
brandonleung, line 128: can every drop in this loop charge the budget it consumed? same for resource-group skip above

rust/capture/src/otel/mod.rs
```diff
@@ -199,20 +213,183 @@ pub async fn otel_handler(
...
+
+    debug!(
+        "OTEL endpoint request processed successfully: {} spans",
+        span_count
+    );
+
+    Ok(Json(json!({})))
+}
+
+#[instrument(skip(state, body), fields(record_count, body_size))]
```
brandonleung: headers isn't skipped here, so the bearer token and the gateway signature get recorded as span fields and exported. every other handler in capture uses skip_all

rust/capture/src/otel/mod.rs
```diff
@@ -199,20 +213,183 @@ pub async fn otel_handler(
...
     // All-or-nothing quota check: reject the entire batch if any span is over quota
-    if let Err(outcome) = filtering::check_quota(
-        &state.quota_limiter,
-        &token,
-        &span_events,
-        gateway_provenance == provenance::Provenance::Verified,
-    )
-    .await
-    {
+    if let Err(outcome) = filtering::check_quota(&state.quota_limiter, &token, &span_events).await {
```
brandonleung: classification is per event now but the outcome isn't. can we be consistent?

brandonleung (changes requested):
can you do a sweep across all asks to make sure we aren't just solving them locally? this applies for issues you've resolved as well.

rust/capture/tests/integration_otel_endpoint.rs
richardsolomou: yep, these were bailing before the behavior they claimed to cover. every signed trace and logs case now sends the request ID covered by its signature.

rust/capture/src/otel/logs.rs
richardsolomou: youre right. i swept both logs and traces: all four wire-owned ID properties are removed first, then only correctly sized wire values are added back.

rust/capture/src/otel/logs.rs
richardsolomou: yep, every evaluation candidate now charges the expansion budget before it can be dropped. oversized resource groups charge once per candidate too, so skipped records cant avoid the aggregate cap.

rust/capture/src/otel/mod.rs
richardsolomou: youre right, both OTLP handlers had the same issue. both use `skip_all` now, and the live request logs contain none of the auth or gateway signature fields.

rust/capture/src/otel/mod.rs
richardsolomou: yep, the outcome now follows the per-event classification. scoped quotas and restrictions retain unaffected siblings, while genuinely request-wide failures like the global quota still reject the batch.

products/ai_observability/frontend/utils.ts
```diff
@@ -1231,7 +1249,7 @@ export async function queryEvaluationRunsStats(params: {
         SELECT
             count() as total,
             countIf(properties.$ai_evaluation_result IS NOT NULL) as applicable,
-            countIf(${hogql.raw(EVALUATION_PASSED_HOGQL)}) as passed
+            countIf(properties.$ai_evaluation_result = 1) as passed
```
brandonleung: this is a revert of a fix from an earlier round. there are other call sites that still use the constant

rust/capture/src/otel/mod.rs
```diff
@@ -222,63 +236,233 @@ pub async fn otel_handler(
...
+    }
+    process_events(&state, ip, token.to_string(), events, received_at).await?;
+
+    counter!("capture_ai_otel_log_records_accepted").increment(event_count as u64);
+    counter!("capture_ai_otel_logs_events_ingested").increment(event_count as u64);
+    counter!("capture_ai_otel_logs_requests_success").increment(1);
+    debug!(
+        "OTEL logs request processed successfully: {} records",
+        event_count
+    );
```
brandonleung, line 463: do exporters need partial success here and can we move the ingested counter after process_events?

products/ai_observability/frontend/utils.ts
richardsolomou: sorted, it was a rebase issue

rust/capture/src/otel/mod.rs
richardsolomou: yep, exporters need partial success here. `process_events` now returns the actual ingested and rejected counts, so the response and counters reflect quota and restriction filtering without treating unrelated records as rejected.

brandonleung (changes requested):
i think this scope seems largely obsolete since our de-productization. some hardening and fail-closed behaviour could be worth salvaging but i'll leave that to your decision

richardsolomou:
Closing this because we are not moving forward with the AI gateway OpenTelemetry integration right now. The existing direct OTLP ingestion path remains supported, and this work can be recovered if the gateway integration becomes a concrete requirement later.

PR #36803, opened 2025-08-18
fix(insights): adjust week interval query to be after date_from

nima-karimi:
## Problem

Insights with a week interval are truncated to start at the `date_from` instead of week start day when we display them on charts. However, the data returned included records that were from the week start day. This was causing discrepancies between the insights where the data is grouped by week and the rest.

For example, if you query the events from July 1, 2025 to July 31, 2025 with a `week` interval, it will include the records from June 29, 2025 (Sunday) because that is the start of the week. When in fact, the date_from is set to July 1 so we should only fetch the records from and after July 1.

https://posthog.slack.com/archives/C045L1VEG87/p1755525666030529

## Changes

- Update the where query used to filter events to include special handling for `week` intervals. This will set the date_from to the start of day instead of start of week for week intervals.

## How did you test this code?

Create an insight with a week interval from July 1, 2025 to July 31, 2025. Create another insight with a `day` interval but with the same dates. The total number of events must be the same in both insights.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

PR #71938, opened 2026-07-17
fix(insights): stop month/quarter/year intervals including events before date_from

sampennington:
## Problem

A trend insight grouped by **month** (also **quarter** / **year**) didn't respect the date range when the range was shorter than the bucket. For a relative range like "last 7 days" grouped by month, the `WHERE` clause rounded `date_from` back to the start of the interval bucket (the 1st of the month), so events from before `date_from` were counted. The grouped total then disagreed with a table view of the same range.

Weeks were already fixed for exactly this in [#36803](https://github.com/PostHog/posthog/pull/53711) — the guard just never got generalized to the coarser intervals. A previous attempt ([#53711](https://github.com/PostHog/posthog/pull/53711)) was reverted in [#53739](https://github.com/PostHog/posthog/pull/53739); this redoes it and additionally covers quarter (which #53711 missed).

Context: [Slack thread](https://posthog.slack.com/archives/C0368RPHLQH/p1784285277644629?thread_ts=1784285277.644629&cid=C0368RPHLQH).

## Changes

- Generalize the week guard in `date_from_with_adjusted_start_of_interval_hogql` to cover **week / month / quarter / year**: the `WHERE` lower bound now truncates to the start of `date_from`'s day rather than the start of the interval bucket, matching how the day interval already behaves.
- Explicitly calendar-picked dates are unaffected — they set `explicitDate`, which bypasses the start-of-interval adjustment, so only relative ranges change.

Note the bucket *label* still comes from the start of the interval (e.g. the first bar reads "Jul"), while the data is filtered from `date_from` — the same partial-leading-bucket behavior weeks already have, now applied consistently.

I (the agent) could not run the ClickHouse-backed suite or regenerate snapshots in this environment (no local ClickHouse/pytest). Automated changes made:

- Added a parameterized unit test in `test_query_date_range.py` asserting the adjusted lower bound uses day granularity (`toIntervalDay`) for week/month/quarter/year — guards against the guard being narrowed back to a subset of intervals (the exact gap #53711 shipped).
- Updated two existing behavior tests (`test_trends.py`, `test_trends_query_runner.py`) that asserted the old over-counting behavior, mirroring the corrected expectations from #53711.
- `ruff check` / `ruff format` pass; all four Python files compile.

The `.ambr` snapshot files that render month-interval `WHERE` clauses (`toIntervalMonth(1)` → `toIntervalDay(1)`) still need regeneration — expected to be handled by the snapshot-regeneration CI bot, as on #53711.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by the PostHog Slack app (Claude Opus 4.8) at @sampennington's direction from a Slack thread. Investigated the revert history of #53711/#53739 and the original week fix (#36803) to confirm the mechanism, then generalized the guard rather than re-patching a single interval. Invoked the repo `/writing-tests` skill before touching tests; chose a cheap unit-level SQL-generation assertion over duplicating the expensive ClickHouse-backed interval test for quarter/year. Snapshot regeneration deferred to CI.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0368RPHLQH/p1784285277644629?thread_ts=1784285277.644629&cid=C0368RPHLQH)*

### DIFF
--- a/posthog/hogql_queries/utils/query_date_range.py
+++ b/posthog/hogql_queries/utils/query_date_range.py
@@ -365,15 +365,17 @@ class QueryDateRange:
         return self.date_to_start_of_interval_hogql(self.date_from_as_hogql())
 
     def date_from_with_adjusted_start_of_interval_hogql(self) -> ast.Call:
-        if self.interval_name == "week":
-            # in `where` queries with week intervals, we need to return the date_from instead of the start of the week
-            # this ensures that we only fetch records after the date_from
+        if self.interval_name in ("week", "month", "quarter", "year"):
+            # in `where` queries with intervals coarser than a day, filter from the start of `date_from`'s day
+            # rather than the start of the interval bucket. This ensures we only fetch records at or after
+            # date_from, matching how the day interval already behaves and keeping grouped totals consistent
+            # with the selected date range.
             return ast.Call(
                 name="toStartOfInterval",
                 args=[
                     self.date_from_as_hogql(),
                     ast.Call(
-                        name=f"toIntervalDay",
+                        name="toIntervalDay",
                         args=[ast.Constant(value=1)],
                     ),
                 ],
--- a/posthog/hogql_queries/utils/test/test_query_date_range.py
+++ b/posthog/hogql_queries/utils/test/test_query_date_range.py
@@ -118,6 +118,29 @@ class TestQueryDateRange(APIBaseTest):
 
     @parameterized.expand(
         [
+            ("week", IntervalType.WEEK),
+            ("month", IntervalType.MONTH),
+            ("quarter", IntervalType.QUARTER),
+            ("year", IntervalType.YEAR),
+        ]
+    )
+    def test_adjusted_start_of_interval_filters_from_start_of_day(self, _name, interval):
+        # For intervals coarser than a day, the WHERE lower bound must round `date_from` down to the start of
+        # its day, not the start of the interval bucket — otherwise a relative range shorter than the bucket
+        # (e.g. "last 7 days" grouped by month) pulls in events from before date_from. Guards against the guard
+        # being narrowed to only some of these intervals again.
+        now = parser.isoparse("2021-08-25T00:00:00.000Z")
+        adjusted = QueryDateRange(
+            team=self.team, date_range=DateRange(date_from="-3d"), interval=interval, now=now
+        ).date_from_with_adjusted_start_of_interval_hogql()
+
+        self.assertEqual(adjusted.name, "toStartOfInterval")
+        interval_period = adjusted.args[1]
+        assert isinstance(interval_period, ast.Call)
+        self.assertEqual(interval_period.name, "toIntervalDay")
+
+    @parameterized.expand(
+        [
             # now is 2021-08-25T10:00Z (a Wednesday): clip to the end of the last complete interval
             ("day", IntervalType.DAY, "-7d", "2021-08-24T23:59:59.999999Z"),
             ("week_sunday_start", IntervalType.WEEK, "-30d", "2021-08-21T23:59:59.999999Z"),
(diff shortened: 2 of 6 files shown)


sampennington (commented):

No blockers — the change correctly generalizes the week-only WHERE lower-bound day-rounding to all coarser-than-day intervals. It only tightens the filter (bucket labels and zero-fill stay snapped to the interval start via a separate placeholder), is a no-op for already-aligned ranges and for the hour/day/explicitDate paths, and scans equal-or-less data. Two non-blocking follow-ups are inline: expressing the "coarser than a day" guard via the existing `compare_interval_length` helper, and adding behavioural (not just AST-level) coverage for the quarter/year intervals the fix now touches.

posthog/hogql_queries/utils/query_date_range.py
```diff
@@ -365,15 +365,17 @@ def date_from_to_start_of_interval_hogql(self) -> ast.Call:
         return self.date_to_start_of_interval_hogql(self.date_from_as_hogql())
 
     def date_from_with_adjusted_start_of_interval_hogql(self) -> ast.Call:
-        if self.interval_name == "week":
-            # in `where` queries with week intervals, we need to return the date_from instead of the start of the week
-            # this ensures that we only fetch records after the date_from
+        if self.interval_name in ("week", "month", "quarter", "year"):
```
sampennington, line 368: 🤖 **sp-ship** · ♻️✨📐 `[reuse+quality+xp]` · **Suggestion / follow-up**

The guard enumerates `("week", "month", "quarter", "year")` — the literal spelling of "coarser than a day," a concept this file already names via `compare_interval_length(self.interval_type, ">", IntervalType.DAY)` over `ORDERED_INTERVALS`. Expressing it that way says it OnceAndOnlyOnce and stays correct if a coarser interval is ever added (the tuple would silently exclude it). While here, the inline `toStartOfInterval(..., toIntervalDay(1))` could also route through the existing `date_to_start_of_interval_hogql` helper / `interval_spec("day").interval_func` instead of hardcoding the name. Non-blocking — the branch already existed for `week`; this PR only broadens it.

posthog/hogql_queries/utils/test/test_query_date_range.py
```diff
@@ -116,6 +116,29 @@ def test_all_values_quarter_year(self, _name, interval, date_from, expected_star
...
 
+    @parameterized.expand(
+        [
+            ("week", IntervalType.WEEK),
+            ("month", IntervalType.MONTH),
+            ("quarter", IntervalType.QUARTER),
+            ("year", IntervalType.YEAR),
+        ]
+    )
+    def test_adjusted_start_of_interval_filters_from_start_of_day(self, _name, interval):
```
sampennington, line 127: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

The fix extends day-rounding to `month`, `quarter`, and `year`, but only `month` has end-to-end behavioural coverage; `test_trends_quarter_and_year_intervals` uses `date_from="2020-01-01"` (already bucket-aligned), so it can't catch quarter/year being dropped from the guard. This new test locks the interval set at the AST level, but a behavioural test with a mid-quarter `date_from` (asserting pre-`date_from` events are excluded) would prove the user-visible effect. It also only asserts `args[1]` is `toIntervalDay`, not that `args[0]` is the `date_from` value.

posthog/hogql_queries/utils/query_date_range.py
sampennington: 🤖 **sp-triage**

Agreed this is more OnceAndOnlyOnce, but out of scope here — this PR only broadens an existing branch to match `week`'s behavior, and re-deriving the guard from `compare_interval_length`/`ORDERED_INTERVALS` (plus routing through `date_to_start_of_interval_hogql`) is a refactor of already-reviewed, already-tested logic rather than a fix for this change. Leaving as a follow-up rather than touching it in this PR.

posthog/hogql_queries/utils/test/test_query_date_range.py
sampennington: 🤖 **sp-triage**

Good suggestion, but the behavioral effect this test guards is already covered end-to-end elsewhere in this PR: `test_trends.py::test_interval_rounding_monthly` uses a mid-month `date_from` and asserts the pre-`date_from` event is now excluded from the bucketed count, and `test_trends_query_runner.py::test_exact_time_range_with_larger_interval` asserts the same for the WHERE-clause path. This unit test is intentionally the narrower AST-level guard against the interval tuple regressing. Resolving as already-addressed rather than expanding further in this PR.

PR #71939, opened 2026-07-17
feat(ingestion-consumer): deterministic aperture routing strategy

jose-sequeira:
## Problem

Stacked on #71923 (peer awareness) and #71914 (shared discovery). Both existing routing strategies (p2c, binpack) answer "who has the least work" per key group, so small batches scatter tiny sub-batches across the whole worker pool — the measured dev baseline was a mean sub-batch of 3.5 messages with 96% at ≤10, defeating worker-side batching. The interim static cap (#71890) fixed consolidation but load testing showed its blind spot: a fixed fan-out formula can't react to worker saturation (6.7% worker-busy send retries and 9x worker imbalance under load while other workers idled).

## Changes

- New `INGESTION_ROUTING_STRATEGY=aperture` (Finagle-style deterministic aperture): every dispatcher sorts the full worker set into the same ring, takes the slice starting at `peer_index * ring_len / peer_count`, `INGESTION_MIN_APERTURE` wide (default 3), and routes unpinned keys within it via p2c. Peer indices come from #71923, so the fleet's slices tile the pool with zero coordination.
- The configured width is a minimum: the effective width is floored at `ceil(pool / dispatchers)`, the narrowest equal slices that can cover the ring, so the union of slices covers every worker even when workers outnumber `width x dispatchers`. Every peer computes the same floor from the same inputs, keeping the property coordination-free (addresses the review callout on coverage gaps).
- Unroutable ring positions (unhealthy/draining workers) are skipped and the slice extends around the ring, holding the effective width under partial failure.
- Falls back to plain p2c over the full healthy pool while the peer set is unknown (startup, peer awareness disabled), with a startup warning when the strategy is enabled without `PEER_SERVICE_NAME`.
- Pinned keys are honored regardless of slice — per-distinct_id ordering is untouched, and pins migrate into the slice naturally as they expire and re-pin.
- Existing strategies are byte-for-byte unchanged; the default remains binpack.

A follow-up stage makes the width feedback-driven (widen on worker-busy retries / slice saturation, narrow when idle) — the load-test data above is the design target for that controller.

- `cargo test -p ingestion-consumer` (182 tests), fmt, clippy `--all-targets --all-features`, shear — all clean.
- New tests and the regression each catches: ring-slice math (tiling without overlap across peers, width flooring on oversized rings and uneven divisions, wrap-around, extension past unhealthy workers, width capping, every fallback-to-`None` case — each would silently mis-partition the pool); dispatcher-level: all unpinned keys stay within the slice (cap not applied), two complementary peers route into disjoint slices (fleet coverage), a pinned key keeps routing to its out-of-slice worker (per-key ordering), and unknown self index falls back to the full pool without dropping messages (startup stall).
- Not manually tested against a live cluster yet. Verification plan once the stack merges: enable in dev, re-pull `/debug/state`, and compare sub-batch distribution and per-worker spread against the captured baseline and load-test snapshots.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Config documented in `config.rs` doc comments; no user-facing docs.

## 🤖 Agent context

I directed this; Claude Code (Fable 5) implemented it. Third PR in the aperture plan (see the stack). Design grounded in live dev measurements: an idle-traffic baseline, a target-cap comparison (16x sub-batch size improvement), and a load test that exposed the static cap's saturation blind spot. Decisions: aperture implemented as candidate narrowing in the dispatcher with p2c unchanged inside the slice (not a separate selector); integer ring slices rather than Finagle's fractional-boundary weighting (unnecessary at our pool sizes); ring built from the full registered worker set (same EndpointSlice view for all dispatchers) with health filtering applied during slice extension. Skills invoked: /writing-tests.


rust/ingestion-consumer/src/aperture.rs
```diff
@@ -0,0 +1,151 @@
+//! Deterministic aperture: ring-slice candidate selection.
+//!
+//! Every dispatcher sorts the full worker set into the same ring, takes the
+//! slice starting at `peer_index * ring_len / peer_count`, and routes its
+//! unpinned groups only within that slice (via the configured selection
+//! strategy). Because peers agree on both the ring order (same EndpointSlices,
+//! same sort) and their own indices ([`k8s_awareness::PeerTracker`]), the
+//! slices tile the ring without any coordination — each dispatcher talks to
```
eli-r-ph, line 8: "Slices tile the pool" only holds when `min_aperture × peer_count ≥ worker_count`. With default width 3, any fleet where workers > 3×consumers leaves healthy workers in the gaps between slice starts with **zero** unpinned traffic — capacity idles while covered workers absorb everything. Reword the invariant to "tiles only when `min_aperture × dispatchers ≥ pool`", and gate/warn (or size default width) until the feedback controller lands.

rust/common/k8s-awareness/src/endpoints.rs
```diff
@@ -0,0 +1,299 @@
...
+                                        *current = desired;
+                                        true
+                                    }
+                                });
+                            }
+                        }
+                        Some(Err(e)) => {
+                            warn!(error = %e, service = %service_name, "EndpointSlice watcher error, stream will retry");
+                            counter!(
+                                "k8s_awareness_endpoint_watch_errors_total",
```
eli-r-ph, line 113: Discovery error/stream counters are dropped and replaced by `k8s_awareness_endpoint_watch_{errors,restarts}_total{service}`. Any existing dashboard/alert on the old names goes silently to zero. Update dashboards/alerts to the new `k8s_awareness_*` metric names in the same rollout.

rust/ingestion-consumer/src/dispatcher.rs
```diff
@@ -352,8 +368,29 @@ impl Dispatcher {
...
         }
 
+        // Aperture: narrow the candidates to this dispatcher's ring slice, so
+        // the fleet's slices tile the pool and each batch consolidates onto
+        // few workers. Falls back to the full healthy pool while the peer set
+        // is unknown (startup, peer awareness disabled).
+        let narrowed = if router.strategy() == RoutingStrategy::Aperture {
+            self.aperture.as_ref().and_then(|(tracker, width)| {
+                let peers = tracker.snapshot();
+                let ring = aperture::sorted_ring(self.registry.workers());
```
eli-r-ph, line 378: `aperture::sorted_ring(self.registry.workers())` clones the entire registry and sorts it on **every** `assign` (per batch); p2c/binpack don't. Cache the sorted ring and rebuild only on membership change.

rust/ingestion-consumer/src/dispatcher.rs
```diff
@@ -1321,6 +1358,80 @@ mod tests {
         assert!(b.is_empty());
     }
 
+    // ---- deterministic aperture ----
```
eli-r-ph, line 1361: PR claims "pinned keys are honored regardless of slice," but no aperture test asserts a pinned key routes **outside** the dispatcher's slice — the ordering guarantee is unverified. Add a test: pin a key to an out-of-slice worker, assert it still routes there under `Aperture`.

rust/ingestion-consumer/src/aperture.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn wid(n: usize) -> WorkerId {
+        WorkerId::from(format!("http://10.0.0.{n}:6738").as_str())
+    }
+
+    fn ring(n: usize) -> Vec<WorkerId> {
+        sorted_ring((1..=n).map(wid).collect())
+    }
+
+    #[test]
+    fn test_slices_tile_the_ring_across_peers() {
```
eli-r-ph, line 94: Tiling tests use only exact-fit widths, so the general claim is never exercised for under-coverage (gaps) or over-coverage (overlap). Add a case with `peer_count × width < ring.len()` asserting the (expected) gap, so the coverage limit is documented in tests.

rust/common/k8s-awareness/src/peers.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// Watch `service_name`'s EndpointSlices in `namespace` and keep this
+    /// tracker in sync until `cancel` fires.
+    pub fn watch(
+        self: &Arc<Self>,
+        client: Client,
+        namespace: String,
+        service_name: String,
+        cancel: CancellationToken,
+    ) -> JoinHandle<()> {
+        let (members, _watch_task) =
```
eli-r-ph, line 81: The `_watch_task` `JoinHandle` is dropped/detached; the watch task is never awaited on shutdown and relies solely on the cancel token. Fine as-is if cancellation is guaranteed; optionally retain the handle for a clean join on shutdown.

eli-r-ph (approved):
Nice change, and good idea! Left a few non-blocking callouts to consider, but overall this LGTM 🚀

PR #71941, opened 2026-07-17
feat(tasks): per-actor sandbox GitHub identity on multiplayer transitions

VojtechBartos:
## Problem

In multiplayer Slack threads, several people send follow-ups to one shared sandbox that is authenticated as whoever started it. Without per-actor identity, a follow-up from a different user runs with the **starter's** GitHub token, so a user with no repo access could push or open PRs as the starter.

```mermaid
sequenceDiagram
    actor A as User A (has access)
    actor B as User B (no access)
    participant S as Sandbox (shared)
    participant G as GitHub
    A->>S: starts task, clones repo
    Note over S: bound to A's token
    B->>S: follow-up: commit and open a PR
    Note over S: transition A to B
    S->>S: B has no access, so LOG OUT (strip remote + env file)
    B->>S: agent tries to push
    Note over S,G: no token to fall back on, push blocked (was: pushed as A)
```

## Changes

- **Per-follow-up gate** (`_refresh_sandbox_github`): resolve the sender; if they have repo access, **rebind** the sandbox to their token; if not, **log out** (strip the git remote + GitHub env file) so the prior actor's identity can't be reused. USER-authored runs only.
- **Runtime hardening**: the token also survived in the agent-server's frozen process env and was re-applied by the owner-keyed refresh loop. Remove it from the launch env (delivered per command via the live `/tmp/agent-github-env` file instead), and skip refresh for runs bound to a different actor.

```text
one running sandbox, the token lived in 3 places:
  git remote URL          cleared on logout   OK
  /tmp/agent-github-env   cleared on logout   OK
  agent-server env        frozen at launch    removed by this PR (+ refresh loop no longer re-adds it)
```

## Companion PR (required together)

PostHog/code#3611 (agent-server): treats an emptied env file as an explicit logout (no fallback to process env) and runs `gh` calls as the current actor.

End-to-end on a **dedicated GitHub org where both test users have app-mediated access to the same repo**:
- Follow-up from a user **without** access -> blocked; no fallback to the starter's token (verified in a live sandbox: emptied file + patched resolver yields no token).
- Follow-up from a user **with** access -> **rebinds** to their token and pushes as them. Example PR opened by the follow-up actor: https://github.com/VojtechBartosTesting/ph/pull/1

## Agent context

Model: Claude Opus 4.8.

### DIFF
--- a/products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
+++ b/products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
@@ -515,8 +515,18 @@ def _refresh_sandbox_github(task_run: TaskRun, actor_user: Any, state: dict[str,
             actor_user=actor_user,
             repository=repository,
         )
-    except ReauthorizationRequired:
-        token = None  # new actor lacks usable access → log out (reauth surfaced elsewhere)
+    except ReauthorizationRequired as e:
+        # New actor has no usable GitHub access to this repo (missing/invalid integration,
+        # no app installation, or no repo permission), so we log the sandbox out rather than
+        # run under the prior actor's creds. Reauthorization is surfaced elsewhere.
+        logger.info(
+            "refresh_github_actor_reauthorization_required",
+            run_id=run_id,
+            user_id=actor_user.id,
+            repository=repository,
+            reason=str(e),
+        )
+        token = None
 
     if token:
         try:
--- a/products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
+++ b/products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
@@ -502,7 +502,13 @@ def _refresh_sandbox_github(task_run: TaskRun, actor_user: Any, state: dict[str,
 
     sandbox = _resolve_live_sandbox(state)
     if sandbox is None:
-        return True  # no live handle; the periodic refresh loop reconciles identity
+        # We are past the same-actor fast path, so this is an unconfirmed transition. The
+        # follow-up can still reach a live agent through the saved sandbox URL, so proceeding
+        # would run it under the prior actor's retained credentials. A missing handle (dead
+        # sandbox, or a transient control-plane lookup failure) is not proof the sandbox is
+        # safe, so fail closed rather than deliver without a confirmed rebind or clear.
+        logger.info("refresh_github_no_sandbox_handle_fail_closed", run_id=run_id, user_id=actor_user.id)
+        return False
 
     repository = task.repository
     token: str | None = None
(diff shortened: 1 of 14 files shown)


products/tasks/backend/temporal/process_task/sandbox_credentials.py
```diff
@@ -156,6 +182,13 @@ def _live_sandboxes_for_user_integration(user_integration_id: int) -> list[tuple
             continue
         if is_caller_token_run(str(run.id), run.state):
             continue
+        # A per-message actor transition may have rebound (or logged out) this sandbox's GitHub
+        # identity to someone other than the run owner. This loop carries the owner's token, so
+        # re-applying it would undo that transition and resurrect the owner's identity for the
+        # current actor. Skip when the sandbox is bound to a different actor.
+        bound_actor = get_sandbox_github_identity_user(sandbox_identity_scope(str(run.id), run.state))
```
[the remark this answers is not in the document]
VojtechBartos: fixed

products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
```diff
@@ -429,6 +446,106 @@ def _refresh_sandbox_mcp(
...
+            run_id=run_id,
+            user_id=actor_user.id,
+            repository=repository,
+            reason=str(e),
+        )
+        token = None
+
+    if token:
+        try:
+            apply_github_credentials_to_sandbox(sandbox, repository, token)
```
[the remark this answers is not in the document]
VojtechBartos: fixed

products/tasks/backend/temporal/process_task/activities/send_followup_to_sandbox.py
```diff
@@ -429,6 +446,106 @@ def _refresh_sandbox_mcp(
...
+    if get_sandbox_github_identity_user(scope) == actor_user.id:
+        return True  # sandbox already reflects this actor — cheapest check first
+
+    task = task_run.task
+    if get_pr_authorship_mode(task, state) != PrAuthorshipMode.USER:
+        return True
+
+    sandbox = _resolve_live_sandbox(state)
+    if sandbox is None:
+        return True  # no live handle; the periodic refresh loop reconciles identity
```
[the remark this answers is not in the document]
VojtechBartos: fixed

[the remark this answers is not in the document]
VojtechBartos: fixed

andrewm4894 (approved):

products/tasks/backend/temporal/process_task/sandbox_credentials.py
```diff
@@ -111,12 +113,27 @@ def replace_sandbox_credentials(
...
-    if repository:
-        set_git_remote_token(sandbox, repository, github_token)
+def apply_github_credentials_to_sandbox(sandbox: "SandboxBase", repository: str | None, github_token: str) -> bool:
+    """Re-inject a GitHub token into both places a running sandbox reads it from.
+
+    Returns ``True`` only when every applicable write succeeded. A caller enforcing per-actor
+    identity must treat a partial write as an unconfirmed rebind: leaving one location on the
+    previous actor's token would let a follow-up actor act as them.
+    """
+    remote_applied = set_git_remote_token(sandbox, repository, github_token) if repository else True
```
tatoalo, line 125: Transitions can race scheduled refresh or token propagation because the marker check and credential writes aren’t atomic, so the owner token could overwrite the new actor. 

Can we serialize all per-sandbox credential writers around check, writes, and marker update? Just want to avoid regression upon token refresh mechanism, people get pissed (rightfully so) whenever they lose work in their cloud workflows :sad

products/tasks/backend/temporal/process_task/sandbox_credentials.py
```diff
@@ -137,6 +156,81 @@ def _loop_owner_credentials_revoked(task: Task, state: dict | None) -> bool:
...
+    propagation) carry the owner's token, so re-applying it would resurrect the owner's identity
+    over the current actor's session — callers skip when this returns a value. Distinct from
+    `_loop_owner_credentials_revoked`, which gates on owner *eligibility* rather than session rebind."""
+    bound_actor = get_sandbox_github_identity_user(sandbox_identity_scope(run_id, state))
+    return bound_actor if bound_actor is not None and bound_actor != owner_id else None
+
+
+# Only the fast in-sandbox writes run under this lock (token resolution stays outside), so a short
+# TTL comfortably covers them; the wait stays well under the refresh activity's 2 min timeout.
+_CREDENTIAL_LOCK_TTL_SECONDS = 30
```
tatoalo: This lock also protects normal cloud-run credential refreshes, but it can expire while the protected writes are still running because the remote update and chmod may each take 30 seconds. 
CAn we renew the lock or increase its TTL to cover the complete worst-case operation so concurrent refreshes cannot interleave, it was a big issue back a month or so ago

VojtechBartos: ahh interesting, yeah, we can, what TTL you are suggesting, like 5 minutes would be enough?

tatoalo: Five minutes seems okay fort now, yup. Could we also document how it’s derived and add a test covering a writer that exceeds the old 30-second lease? Longer term, we would need some lock renewal but for now I think that;s okay

VojtechBartos: yup, will do!

tatoalo (approved):

PR #71058, opened 2026-07-15
chore(dashboards): remove dashboard quick filters experiment

thmsobrmlr:
## TL;DR

The dashboard quick filters A/B test lost, so the code behind it is being deleted. Dashboards keep the regular filter edit bar everyone in the control group already sees. Nothing changes for users.

## Problem

The `dashboard-quick-filters-experiment` feature flag gated a quick-filters bar on the dashboard scene. The experiment failed, so the test-variant code should be removed and the control experience (the standard edit bar) kept.

## Changes

- Remove the `DASHBOARD_QUICK_FILTERS_EXPERIMENT` flag constant
- Delete the dashboard-scene experiment surface: `DashboardQuickFiltersButton`, `dashboardQuickFiltersLogic`, `dashboardQuickFiltersSelectionLogic`, `dashboardFiltersLogic` (advanced-options toggle), and the debounce test
- Simplify `DashboardFilters.tsx` to always render `DashboardEditBar` and strip quick-filter rows/advanced options
- Remove quick filter wiring from `dashboardLogic` (URL restore deferral, commit-debounce refresh, filter merging selectors, `dashboardFiltersEnabled`)

The shared `lib/components/QuickFilters` library, `QuickFilterContext`, and the backend `QuickFilter` model/API are untouched — they're used by error tracking and dashboard widgets. The now-unused `Dashboard.quick_filter_ids` backend field is left in place; dropping the column safely is a separate staged migration if wanted.

## Why

The experiment failed, so the losing variant's code is dead weight and the flag can be cleaned up.

## How did you test this code?

No manual testing (headless environment without node_modules). Ran oxfmt and oxlint on the changed files; verified via repo-wide search that no references to the removed flag, components, selectors, or actions remain. Existing dashboard tests and CI typecheck cover the touched paths. Deleted `dashboardQuickFiltersDebounce.test.ts` since it only tested removed behavior.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

None needed — experiment was never documented.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Authored with PostHog Code (Claude Code). Scoped the removal by tracing what the `dashboard-quick-filters-experiment` flag actually gated (`dashboardFiltersEnabled` selector); deliberately kept the shared QuickFilters library and backend model/API because error tracking and dashboard widgets consume them, and left the `quick_filter_ids` dashboard field for a separate column-drop migration.

---
[...]

PR #71957, opened 2026-07-17
feat(dashboards): add test account filter override to edit bar

thmsobrmlr:
> Follow-up to #71845, which made interval-only dashboard overrides survive opening an insight.

## TL;DR

Dashboards can now force "filter out internal and test users" on or off for every insight on them, or leave each insight's own setting alone. The new switch lives in a collapsible "Advanced overrides" section of the dashboard edit bar, together with the property and breakdown overrides. A switch showing a dash in the middle means "not overriding anything".

## Problem

The backend already applies a tri-state `filterTestAccounts` dashboard override (null/absent = inherit, true = force on, false = force off) via `QueryRunner.apply_dashboard_filters`, but there was no way to set it from the UI. Meanwhile the edit bar was a single flat row where the rarely-used property and breakdown overrides sat next to the date and interval controls.

## Changes

- `TestAccountFilterSwitch` accepts `checked: boolean | 'indeterminate'` and an optional `onReset`. With `onReset` set it renders as an override control: "No override set" while indeterminate, an "Unset override" link (back to inherit) once resolved. Existing boolean callers are unchanged.
- New `DashboardTestAccountFilter` in the dashboard edit bar maps `effectiveEditBarFilters.filterTestAccounts ?? 'indeterminate'` onto the switch and dispatches the new `setFilterTestAccounts` action.
- The edit bar is restructured into a main row (date, interval, variables) plus a collapsible "Advanced overrides" row (property filters, breakdown, test account switch), collapsed by default with an active-override count badge. This revives the section pattern removed in #71058.
- `dashboardLogic` wires `setFilterTestAccounts` exactly like the existing `setInterval` override: intermittent filter state, auto-preview refresh, `query_filters` URL sync, and a `test_account` change-type on the `dashboard filters changed` event.
- `isDashboardFilterEmpty` now treats `filterTestAccounts: false` as a real constraint, so a force-off-only override survives opening an insight from a dashboard (extends the interval fix in #71845).
- The insight card's "Show insight details" popover now mentions the test account override (and the interval override, which had the same gap), with the same layer tags and struck-through "was ... from Insight" pattern used for the date, property, and breakdown overrides. Resolution is tile-beats-dashboard via a shared scalar helper in `insightDetailsFilterOverrides.ts`; the backend already carried both fields through `filter_override_context`.

No backend or schema changes; the API side shipped previously.

I couldn't run the UI in this environment, so no screenshots. The switch's indeterminate state renders as the centered-handle dash shipped from the prototype exploration in #68644.

Automated only, I (Claude) didn't test manually:

- `dashboardFilterEmpty.test.ts`: `filterTestAccounts: true` and `false` are non-empty (the `false` case catches a truthiness-refactor regression that would drop force-off overrides), `null` stays empty.
- `TestAccountFiltersSwitch.test.tsx`: clicking "Unset override" calls `onReset` without toggling the switch. The link sits inside the switch's `<label htmlFor>`, so without the click-guard the reset would also flip the switch; verified the test fails when the guard is removed.
- `InsightDetails.test.tsx`: parameterized cases for the new scalar override resolution — force-off (`false`) still resolves as an override (same truthiness-regression guard as above), tile beats dashboard, and backend `filter_override_context` layers win over raw props.
- Kea typegen for `dashboardLogic` reconciles cleanly; frontend typecheck reports no errors in changed files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None in this repo; dashboard filter docs live on posthog.com.

## 🤖 Agent context

Authored with PostHog Code (Claude Code). Skills invoked: `/writing-tests`. Decisions: reused `LemonSwitch`'s shipped indeterminate state and the "Unset override" affordance from the prototype in #68644 rather than a 3-state click cycle (keeps stock switch semantics); expand/collapse state lives as a plain reducer in `dashboardLogic` instead of reviving the separate keyed logic deleted in #71058; `combineDashboardFilters`, the save path, and the backend needed no changes since they are key-generic. Originally built stacked on #71845; rebased onto master once that PR merged.

---


MattPua:
@thmsobrmlr is there a screenshot

thmsobrmlr:
> @thmsobrmlr is there a screenshot

Here you go:
Screenshot 2026-07-17 at 18 21 40
Screenshot 2026-07-17 at 18 22 58
Screenshot 2026-07-17 at 18 23 15

I've marked this as ready to get the full CI running, but annoyingly that triggers the auto reviewer assignment as well. Currently testing this myself, but looks promising.

PR #71965, opened 2026-07-17
feat(workflows): add MessageSuppression model + migrations

meikelmosby:
## Problem

Base of a two-PR stack. Splits the `posthog_messagesuppression` schema change out of #71038 so the table can land and bake independently of the write / enforcement paths that consume it.

The suppression list is a per-team record of email addresses workflow email sends should skip — populated automatically on repeated soft bounces or hard bounces, and manageable via the Suppression list UI / API. All that consumer code ships in the child PR.

## Changes

- New model `MessageSuppression` (`products/messaging/backend/models/message_suppression.py`) plus its export from `products.messaging.backend.models.__init__`. Team-scoped via `TeamScopedRootMixin`, `db_constraint=False` on the hot-table FKs to `posthog.Team` / `posthog.User`.
- Migration `0003_message_suppression` — creates `posthog_messagesuppression` with the standard team-scoped shape (source, reason, transient_bounce_count, last_bounce_*, suppressed/suppressed_at, deleted, timestamps), plus the `UNIQUE (team_id, identifier)` constraint the ON CONFLICT upserts in the child PR depend on.
- Migration `0004_add_active_suppression_index` — partial index on `(team_id, updated_at DESC)` filtered by `suppressed=true AND deleted=false`. Sized for the Suppression list UI query (which reads exactly that filter, in that order) and its paginator COUNT, so both stay off the full team scan. Run via `SafeAddIndexConcurrently` with `atomic = False`.

- `./manage.py sqlmigrate messaging 0003` and `./manage.py sqlmigrate messaging 0004` — confirmed `0004` emits `CREATE INDEX CONCURRENTLY "pmsg_supp_active_by_updated" ON "posthog_messagesuppression" ("team_id", "updated_at" DESC) WHERE (NOT "deleted" AND "suppressed")`.
- Full end-to-end coverage of read/write behavior against the table lives in the child PR (Node service tests + DRF endpoint tests) — this base PR is schema-only.

## Docs update

No user-facing behavior in this PR — the docs update for the suppression feature ships alongside the consumers in the child PR.

## 🤖 Agent context

Split out of #71038 after the full stack was CI-green, so the schema change can merge and bake independently of the write / enforcement paths. Feature branch `posthog-code/email-suppression-list` (PR #71038) has been rebased onto this branch.

Skills invoked while producing this split: `/django-migrations` (partial-index safety + `SafeAddIndexConcurrently` usage + hot-table FK handling).

Tool: Claude Code (claude-opus-4-7).


mayteio (approved):
Schema and migration safety all check out: `db_constraint=False` on both hot-table FKs so `CreateModel` takes no lock on `posthog_team`/`posthog_user`, `SafeAddIndexConcurrently` with `atomic = False` for the partial index (predicate matches the model Meta and the list-scene query), and the `UNIQUE (team_id, identifier)` constraint the child PR's `ON CONFLICT` upserts need is in place. I also verified the Node inserts in #71038 list every NOT NULL column explicitly, so the absence of Postgres-level defaults doesn't bite.

One suggestion before merge: `MessageSuppression` extends `UUIDTModel`, whose docstring marks it deprecated for new models in favor of `UUIDModel` (UUIDv7). Since this is a brand-new table with no data, now is the cheapest moment to swap - one-line model change plus regenerating 0003, and time-ordered PKs are a better fit for an upsert-heavy table.

Optional hardening: `db_default=` on `deleted`/`suppressed`/`source`/`transient_bounce_count` would protect any future raw-SQL writer that forgets a column, but it's not required for the current consumers.

PR #71963, opened 2026-07-17
refactor(data-warehouse): migrate papersign, plausible, postmark, productboard, ramp (batch 29)

Gilbert09:
## Problem

Batch 29 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

## Changes

Migrated sources (behavior-preserving, coverage-first — each keeps its full test suite green, reusing `validate_via_probe`, `build_endpoint_schemas`, resumable built-in paginators, and framework `auth` for secret redaction):

- **papersign**
- **plausible**
- **postmark**
- **productboard**
- **ramp**

Not migrated this batch (left hand-rolled, valid blockers):
- **planhat** — a non-list HTTP-200 body is classified retryable; the framework's `data_selector_required` makes it a permanent error and `response_actions` can't content-match "not a list".
- **printify** — same HTTP-200 body-shape retryable classification (bare string / dict-without-`data` / `{"data": <object>}`), which the framework can't reproduce without silent-corruption or permanent-error regressions.

## How did you test this code?

- Each migrated source's own `tests/` suite passes in isolation, asserting the same pagination progression, resume round-trips, termination, fail-loud paths, and `validate_credentials` mapping as before.
- Merged run of all 5 migrated dirs + `common/rest_source/tests/` — 380 passed.
- `hogli ci:preflight --fix` clean (0 failures).

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-28 PR; merge in order.

PR #71969, opened 2026-07-17
refactor(data-warehouse): malformed-body-retryable + migrate groq, planhat, pretix, raygun (batch 30)

Gilbert09:
## Problem

Batch 30 of the ongoing effort to migrate hand-rolled data-warehouse source connectors onto the shared `rest_source` framework, removing duplicated pagination, auth, resume, and schema-building code.

This batch also lands a generic framework capability — **`data_selector_malformed_retryable`** — so sources that defensively treat an unexpected HTTP-200 body shape as a transient error no longer need bespoke transport code.

## Changes

Framework:
- New endpoint key `data_selector_malformed_retryable: True` in `common/rest_source`. When set, a 200 whose parsed body isn't the expected list shape (the `data_selector` matches nothing, or the matched value / selector-less body isn't a list) raises a retryable error and the request is reissued — the check runs inside the retry-wrapped send. This is the retryable counterpart of `data_selector_required` (which fails loud permanently). Wired for simple resources.

- **groq** — now expressible: non-dict body / non-list `data` classified retryable via the new capability
- **planhat** — non-list 200-body classified retryable via the new capability
- **pretix**
- **raygun**

- **printify** — its malformed-body-retryable classification lives on the per-shop fan-out endpoints; `data_selector_malformed_retryable` is currently wired for simple resources only, not the dependent/fan-out path.
- **pylon** — multiple transport modes (date-windowed `/issues` backfill + paged lists + static object_type fan-out) with client-side window/watermark filtering.
- **pypi** — config-list fan-out with no parent API resource, deriving three differently-reshaped streams (including a version-keyed dict-of-lists explode) from one JSON document.

- New framework capability: 5 cases in `test_malformed_body_retryable.py` — malformed-then-valid recovers, three persistent-malformed shapes (bare string / dict-without-key / key-present-but-not-a-list) reraise retryable, and a valid list body is not retried.
- Merged run of all 4 migrated dirs + `common/rest_source/tests/` — 322 passed. Additive framework change: all existing `rest_source` tests plus `data_selector_required` consumer smoke (apify_dataset, alguna, coingecko, aviationstack, greenhouse) stay green.

## 🤖 Agent context

Part of the stacked rest_source migration program. Base branch is the batch-29 PR (#71963); merge in order.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/__init__.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/__init__.py
@@ -314,6 +314,9 @@ def create_resources(
                     None if has_dependent_resource else initial_paginator_state
                 ),
                 data_selector_required: bool = bool(endpoint_config.get("data_selector_required")),
+                data_selector_malformed_retryable: bool = bool(
+                    endpoint_config.get("data_selector_malformed_retryable")
+                ),
             ) -> Iterator[list[Any]]:
                 if incremental_object:
                     params = _set_incremental_params(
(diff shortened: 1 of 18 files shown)


Gilbert09:

Issue #64016, opened 2026-06-16
[mega-issue] MCP Analytics


lucasheriques:
MCP analytics helps teams see how agents and users actually use their MCP servers. It captures sessions, tool calls, errors, and intent, so you can debug, measure real usage, and understand what people are trying to do.

It answers the questions a team building MCP tools actually has:
- which tools get called, and which are slow or erroring?
- what are people trying to do, at the level of "find a churn cohort", not "ran query_run 14 times"?
- where do conversations stall or fall through to feedback?

Not a goal: replacing LLM analytics/AI o11y. Generation traces, prompt/response inspection, and cost belong there.

## Want this for your MCP server?

If you're building an MCP server and want analytics on it, drop a comment here and we'll reach out. We're looking for teams to pilot with and build alongside.

A few things stay internal (a Slack canvas, PostHog only): mainly the list of folks we're talking to for pilots.

## Current status

SDKs
- TypeScript: `@posthog/mcp` on npm at 0.4.x (alpha). `instrument(server, posthog)` wraps a `Server`/`McpServer`; `PostHogMCP` covers custom dispatchers (hono, edge). Source: https://github.com/PostHog/posthog-js/tree/main/packages/mcp
- Python: merged into posthog-python (https://github.com/PostHog/posthog-python/pull/691). Plain `pip install posthog`, then `from posthog.mcp import instrument` — the `mcp` SDK is a peer dependency, so there's nothing extra to install. Wraps FastMCP and the low-level server with event parity with the TS SDK.

Install & onboarding
- `npx -y @posthog/wizard@latest mcp-analytics` instruments an existing TS/JS server automatically.
- In-app onboarding is event-gated: the product detects `$mcp_initialize` / `$mcp_tool_call` and flips to the dashboard when data arrives.

Product
- Dashboard (KPIs, activity, per-tool usage/error charts, client harness breakdown), sessions with LLM intent summaries, tool quality (per-tool error rate, latency percentiles, failures, description drift), and intent clustering with per-cluster journeys.
- The whole product is queryable over MCP too (session list, tool stats, intent clusters, missing-capability reports).
- Known gap, explicit goal: robust session tracking for stateless, serverless, and multi-replica deployments — where in-process session state doesn't survive between tool calls, sessions today can degrade to one call each. Server-side stitching plus `$mcp_conversation_id` is the direction.

Docs & landing page
- docs: https://posthog.com/docs/mcp-analytics
[...]

PR #71977, opened 2026-07-17
feat(mcp_analytics): add access control

ReeceJones:
## Problem

MCP analytics has no RBAC coverage today: any team member can view and interact with it regardless of how the org has access control configured. Sibling analytics products (AI observability, metrics, tracing, revenue analytics) already support viewer/editor/manager access levels.

Refs #64016

## Changes

- Backend: register `mcp_analytics` in `ACCESS_CONTROL_RESOURCES` (`posthog/rbac/user_access_control.py`). The viewsets already declare `scope_object = "mcp_analytics"` with correct custom-action scope lists, and `AccessControlPermission` already runs on every `TeamAndOrgViewSetMixin` viewset, so this one registration is what turns enforcement on. No `RESOURCE_INHERITANCE_MAP` entry (no child resources) and no `AccessControlViewSetMixin` (this is a flat, resource-level product like metrics/tracing rather than an individually-owned object type, so per-object access rules don't apply).
- Frontend: wire the resource into the access control settings table (icon, rollout-flag gating behind the existing `mcp-analytics` flag, scene mapping for both the dashboard scene and the tool-detail scene) and the storybook mock, and gate the product's one mutating action (the "Summarize session intent" button) behind `AccessControlAction`.

- Added `products/mcp_analytics/backend/tests/test_access_control.py`: parameterized over none/viewer/editor access levels against the feedback viewset's list/create actions. This is the regression no existing test caught - `mcp_analytics` wasn't wired into RBAC at all before, so nothing could show enforcement actually happening. Generic level-comparison and org-admin-bypass mechanics are already covered for every `ACCESS_CONTROL_RESOURCES` entry by the property-based suite in `posthog/rbac/test/test_user_access_control_pbt.py`, which now covers `mcp_analytics` too since it derives its resource list from that same tuple.
- `ruff check` / `ruff format` pass on the changed Python files.
- Ran `mypy --cache-fine-grained .` repo-wide: 1 pre-existing error in `posthog/personhog_client/converters.py`, unrelated to this change and not touched by it. Zero new errors.
- I could not run the new test (or the wider Django suite) against a live database: this sandbox has no Postgres/ClickHouse, no docker, no flox, and no `hogli` binary, so there's no way to bring up the dev stack. `pytest` did successfully collect all 6 parameterized cases before failing at the DB-connection step, which at least confirms the file is free of import/syntax errors.
- I could not run the frontend typecheck or lint (`node_modules` isn't installed and a fresh `pnpm install` isn't feasible in this sandbox). I hand-verified each frontend edit against the exact patterns sibling resources (`llm_analytics`, `Endpoint`, `Metrics`, `RevenueAnalytics`, `Tracing`) already use in the same files.
- I did not do any manual/browser testing of the UI.

CI should be treated as the source of truth for the gaps above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs page enumerates which products support access control by name, so nothing needs updating there.

## 🤖 Agent context

Reece asked me (Claude, via PostHog Code) to retroactively add access control to the MCP analytics product, pointing at the repo's `rj-retroactively-adding-access-control` skill, which documents the standard backend/frontend checklist and a reference `access-control.md`.

Skills invoked: `/writing-tests` (before adding the new test file).

Decisions along the way:
- Considered adding `AccessControlViewSetMixin` (per-object `/access_controls/` endpoints) to the mcp_analytics viewsets, and decided against it: two of the four viewsets (sessions, intent clusters) resolve their data from ClickHouse/a facade rather than a real ORM row, so `get_object()` wouldn't resolve to anything the mixin could operate on; the other two (feedback/missing-capability submissions) are one-way telemetry records where per-row sharing isn't a real use case. This mirrors how AI observability (the closest sibling, also flag-gated) treats its trace data as resource-level-only.
- The actual enforcement turned out to need only a single-line registration change - the viewsets already had `scope_object` set correctly, and `AccessControlPermission` is unconditionally part of every `TeamAndOrgViewSetMixin`'s permission chain, so it silently no-ops until the resource is registered.
- Added `Scene.MCPAnalyticsToolDetail` to the frontend `Scene` enum since the product's manifest already registers that scene but it was missing from the enum, which meant the tool-detail page couldn't be wired into the access-control scene map at all.

---


ReeceJones:
👋 Status update on required checks:

- **Python code quality**, **Frontend CI**, **Security** — all green as of the latest commit.
- **MCP Tests Pass** is currently blocked, but not by anything in this PR's diff: `services/mcp/tests/unit/execute-sql-description.test.ts` (`keeps data-catalog discovery within its character budget`) is failing against `master`'s current tip because [#71921](https://github.com/PostHog/posthog/pull/71921) grew the metric-discovery prompt template past its 1200-character budget guard. This is reproducible on a clean `master` merge with none of this PR's files touched, so it needs a fix in that PR/area rather than here — flagging so it doesn't get mistaken for something this change broke.

This PR should be mergeable once that's resolved upstream and merged in.

products/mcp_analytics/frontend/sessions/MCPSessionDetail.tsx
```diff
@@ -219,15 +222,20 @@ export function MCPSessionDetail(): JSX.Element {
...
+                        <LemonButton
+                            type="primary"
+                            size="xsmall"
+                            icon={<IconSparkles />}
+                            loading={isSelectedSessionGenerating}
+                            onClick={() => generateIntent(selectedSession.session_id)}
+                        >
+                            {isSelectedSessionGenerating ? 'Thinking…' : 'Summarize session intent'}
+                        </LemonButton>
+                    </AccessControlAction>
```
lucasheriques, line 238: blocking: the clustering page needs this same treatment. It has three triggers of the write-scoped `recompute` action ("Recompute" at `MCPAnalyticsClustering.tsx:471`, "Compute intent clusters" in the empty state, and the banner Retry), and none are gated. A viewer-level user sees them fully enabled and every click fails with the "Could not start clustering" toast.

Those are quill Buttons rather than LemonButtons, so if they don't take `disabledReason`, `AccessControlAction`'s function-child form gives you `{ disabled, disabledReason }` to spread. `ai_observability` gates its equivalent "Run clustering" this way (`ClustersView.tsx:113`).

posthog/rbac/user_access_control.py
```diff
@@ -79,6 +79,7 @@ class AccessSource(Enum):
     "activity_log",
     "error_tracking",
     "logs",
+    "mcp_analytics",
```
lucasheriques, line 87: question: this registration also makes `intent_digest` editor-only (it's in `scope_object_write_actions`), and `mcpEarlyDataLogic.ts` auto-fires that POST on dashboard mount plus every 60s refresh. So a viewer-level user now sends a guaranteed-403 request on every load, the bare `catch` (written for the 503 LLM-unconfigured case) swallows it, and the digest card silently falls back with no hint it's a permissions thing.

Two ways out: reclassify it as a read action if it's effectively a cached-digest fetch (viewers keep the card, matches `activity_overview`), or keep it editor-level and guard the loader with `userHasAccess` so viewers never fire the request. Which reading is right for what `intent_digest` actually does?

frontend/src/lib/utils/accessControlUtils.ts
```diff
@@ -61,6 +61,8 @@ export const pluralizeResource = (resource: APIScopeObject): string => {
         return 'customer analytics'
     } else if (resource === AccessControlResourceType.LlmAnalytics) {
         return 'AI observability'
+    } else if (resource === AccessControlResourceType.McpAnalytics) {
+        return 'MCP analytics'
```
lucasheriques, line 65: nit: `resourceTypeToString` (below, ~line 116) needs a McpAnalytics branch too. Without it the disabled-button tooltip falls through to the underscore-replace default and reads "You don't have sufficient permissions for this mcp analytics." Siblings all have the explicit case (`Metrics` -> "metrics resource", `LlmAnalytics` -> "AI observability resource").

lucasheriques (approved):
Solid work. The one-line registration doing all the enforcement is well spotted, closing the `/query/` runner path in the same PR was the right call, and the test file covers the transitions that matter.

One theme inline: the PR treats "Summarize session intent" as the product's only mutating action, but `recompute` and `intent_digest` are also in `scope_object_write_actions`, and both have viewer-visible UI that now 403s. Server-side enforcement is correct and fails closed everywhere, so nothing here is a security problem, and only orgs that explicitly restrict `mcp_analytics` below the editor default hit it. Approving so you're not blocked on a re-review.

products/mcp_analytics/frontend/sessions/MCPSessionDetail.tsx
ReeceJones: I have a separate PR open (draft right now) to separate clustering into its own access control resource. Should the actions you reference in `MCPAnalyticsClustering.tsx` be governed by mcp analytics access control or clustering? Or possibly both? Looking at the code it is not clear to me which product would take precedence for access control.

posthog/rbac/user_access_control.py
ReeceJones: It appears to be a read action. I think it makes sense to update the viewset to require `viewer` for `intent_digest`.

a-lider (approved):

PR #71979, opened 2026-07-17
feat(aio): add access control to taggers

ReeceJones:
## Problem

Taggers (`/ai-evals/taggers`, part of AI observability) already had backend access control wired up — `Tagger`/`TaggerViewSet` inherits from the `llm_analytics` resource via `RESOURCE_INHERITANCE_MAP`, same as Evaluations and Datasets. But the frontend scene registration was never finished, so the page-level and button-level gating that its sibling scenes have didn't carry over to Taggers.

## Changes

- Added `Scene.AIObservabilityTags`/`Scene.AIObservabilityTag` to `sceneToAccessControlResourceType` (`frontend/src/scenes/sceneTypes.ts`), mapped to `AccessControlResourceType.LlmAnalytics`. Without this, visiting the Taggers list or detail page directly bypassed the route-level access check that every sibling AI observability page already has (e.g. Evaluations, Datasets) — a user with no access to AI observability could still view Taggers content.
- Wrapped the Save/Create and Delete buttons on the tagger detail/edit scene (`AIObservabilityTag.tsx`) in `AccessControlAction`, matching the pattern already used on the Taggers list page and the sibling Evaluation detail scene.

This sandbox doesn't have frontend dependencies installed (no `node_modules`), so I couldn't run `pnpm typecheck`/lint here. I manually traced the exact pattern from the sibling `AIObservabilityEvaluation.tsx` scene (which already has this wiring) and mirrored it precisely, including how `AccessControlAction`'s prop composition interacts with the existing `disabledReason` logic on the Save button. No new tests added — the underlying `AccessControlAction`/`sceneToAccessControlResourceType` mechanisms are shared, untested-in-isolation primitives used the same way in ~100 other call sites across the codebase, so covering just these two usages wouldn't catch a regression that isn't equally true everywhere else.

## Docs update

N/A — no documented workflow describes Taggers' access control model.

## 🤖 Agent context

Reece asked me to retroactively add access control to the "tagging" product using the repo's `rj-retroactively-adding-access-control` skill. Tagging turned out to be ambiguous — there's no dedicated tagging product; I initially assumed the generic `Tag`/`TaggedItem` model used across dashboards/insights/flags, but a clarifying round revealed he meant AI observability's "Taggers" feature (`/ai-evals/taggers`) instead.

From there I audited Taggers against the skill's checklist and against its sibling resources (Evaluations, Datasets) rather than assuming access control was entirely missing — the backend, in fact, was already fully wired. The two frontend gaps above were the only real deltas. I deliberately did not add new tests (see previous section) and did not touch the backend, Storybook mocks, or the access-control settings table, since those were already correct and consistent with siblings.

One adjacent issue surfaced during the audit but is out of scope for this PR: `llmTaggersLogic.ts`'s `loadTaggers` listener auto-seeds default tagger templates via an unconditional `api.create(...)` call the first time a team's list is empty, with no try/catch and no access-level check — a viewer-level (non-editor) user visiting the page for the first time would trigger a request the backend correctly rejects, but the frontend doesn't handle the failure gracefully. Flagging for a follow-up rather than fixing here since it's a pre-existing behavior question (should seeding require editor access, move server-side, or fail silently?) unrelated to the missing scene registration this PR addresses.


a-lider (commented):
Greptile + veria flagged this, looks legit to me. Dropping the `llm_analytics` inheritance means any existing restrictive `llm_analytics` grant no longer gates taggers, and since `default_access_level` falls back to `editor`, taggers end up editable by every project member with no explicit grant. 

Ideally, customers who currently have `llm_analytics` restricted (none/viewer) should keep that restriction on `taggers`. But maybe it's not critical if there are just a few orgs that have configured it. Could you check?

ReeceJones:
> Greptile + veria flagged this, looks legit to me. Dropping the `llm_analytics` inheritance means any existing restrictive `llm_analytics` grant no longer gates taggers, and since `default_access_level` falls back to `editor`, taggers end up editable by every project member with no explicit grant.
> 
> Ideally, customers who currently have `llm_analytics` restricted (none/viewer) should keep that restriction on `taggers`. But maybe it's not critical if there are just a few orgs that have configured it. Could you check?

I'll add a django management command to backfill the access control rules based on the existing llma rules.

a-lider (approved):

posthog/management/commands/test/test_backfill_tagger_access_control.py
```diff
@@ -0,0 +1,83 @@
...
+
+from django.core.management import call_command
+
+from posthog.models.organization import OrganizationMembership
+from posthog.models.user import User
+
+try:
+    from ee.models.rbac.access_control import AccessControl
+except ImportError:
+    pass
```
carlos-marchal-ph: `test_taggers.py` in this same PR imports the ee model unconditionally, this one guards it. Either works, but I'd stick to one.

posthog/management/commands/backfill_tagger_access_control.py
```diff
@@ -0,0 +1,97 @@
...
+            created += 1
+            if not dry_run:
+                AccessControl.objects.create(
+                    team_id=source.team_id,
+                    resource="tagger",
+                    resource_id=None,
+                    organization_member=source.organization_member,
+                    role=source.role,
+                    access_level=source.access_level,
+                    created_by=source.created_by,
```
carlos-marchal-ph: `created_by=source.created_by` fetches the User per created row. `created_by_id=source.created_by_id` avoids it, and switching the lookup and create to `organization_member_id`/`role_id` would let you drop the `select_related` above entirely.

posthog/rbac/user_access_control.py
carlos-marchal-ph, line 1: The three bots flagged the same issue on this file, I think they're right

carlos-marchal-ph (approved):
Looks good to me. I'd just do a sync migration for the backfill though, I think it's something like 400 records which should take no time at all to run and it simplifies ordering and removes the need to add a management command and run it. Approving to not block.

PR #71984, opened 2026-07-17
feat(aio): add access control to clustering

ReeceJones:
## Problem

AI observability already enforces `llm_analytics` access control across datasets, evaluations, prompts, tags, and review queues — but the LLM trace clustering feature (config, saved jobs, and manual runs) was never wired in. Any team member could edit saved clustering jobs or trigger a clustering run regardless of role, and the run-trigger endpoint bypassed the access-control framework entirely (`scope_object = "INTERNAL"`) even though the frontend button that calls it was already gated at editor level.

## Changes

- `ClusteringConfigViewSet`, `ClusteringJobViewSet`, and the "Run clustering" trigger viewset now enforce access control end to end (the trigger previously bypassed it entirely via `scope_object = "INTERNAL"`).
- Clustering is its own independent access-control resource (`llm_clusters`), rather than folded under the shared `llm_analytics` umbrella that datasets/evaluations/prompts use. It gets its own row in the access-control settings table with its own Viewer/Editor/Manager level, its own row in the personal-API-key/OAuth scope picker, and its own scene-level gate on the two clustering pages.
- Added a `backfill_llm_clusters_access_control` management command (idempotent, `--dry-run` supported) that mirrors every existing `llm_analytics` grant onto a matching `llm_clusters` grant, so an org's existing customization carries over instead of silently changing the moment clustering stops inheriting.
- Fixed a real display bug at its source: both the frontend (`pluralizeResource`/`resourceTypeToString`) and backend (`resource_to_display_name`) pluralization helpers had no case for the new resource, so it fell through to naive pluralization and rendered as "llm clusterss" — this affected the access-control table, a HogQL "results may exclude…" warning, and a project-settings validation message.
- Extended the product's access-control test suite with clustering-specific coverage, including tests proving clustering access is now independent of `llm_analytics` access in both directions.

- Ran `ruff check` / `ruff format --check` and a full repo-wide `mypy --cache-fine-grained .` locally (clean, aside from one pre-existing and unrelated error in `personhog_client/converters.py` that CI itself doesn't flag).
- Could not run the Django/Jest test suites locally (sandbox has no Postgres/ClickHouse/`node_modules`) — CI ran them instead. All checks are green: `Django Tests Pass`, `Frontend typechecking`, `Frontend Tests Pass`, `Python code quality checks`, and `Validate OpenAPI types`.
- New/changed tests: the management command has 7 tests covering member-scoped, role-scoped, and org-default backfill, idempotency on re-run, dry-run (no writes), and leaving other resources/existing rows untouched. The access-control suite has new tests asserting `llm_analytics` editor access no longer implies clustering access, and vice versa — the actual regression this design change needed to guard against.

+ manually tested

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documented workflow describes per-resource access-control coverage today, so nothing to update.

## 🤖 Agent context

- Tool: PostHog Code (Claude).
- Skills invoked: `/improving-drf-endpoints`, `/writing-tests`.
- This PR went through two design iterations. It started by reusing the existing `llm_analytics` scope for clustering, since two of the three clustering viewsets already declared it and a new dedicated scope seemed like unnecessary plumbing. After further review, clustering needed to be independently manageable in the access-control table rather than folded under the AI observability umbrella — so I promoted it to its own top-level resource instead, which meant adding the backfill command above to keep existing grants working, and fixing the pluralization helpers that hadn't been exercised with a real child-turned-independent resource before.
- Confirmed no internal/scheduled caller depends on the clustering-run endpoint's previous `scope_object = "INTERNAL"` bypass — the scheduled clustering pipeline runs through a separate Temporal schedule, not this REST endpoint.

---


a-lider (commented):
Same Q as for `taggers`. Any org that restricted `llm_analytics` to viewer/none falls back to editor for `clustering` resource. Could the backfill run automatically as a django migration?

ReeceJones:
> Same Q as for `taggers`. Any org that restricted `llm_analytics` to viewer/none falls back to editor for `clustering` resource. Could the backfill run automatically as a django migration?

This PR includes a django management command to backfill the access control rules for clustering based on the existing llma rules.

a-lider (approved):
> This PR includes a django management command to backfill the access control rules for clustering based on the existing llma rules.

I meant that the command has to be run manually after deploy, so there's a window where a user with no `llm_analytics` access ends up with access to clustering. 
But probably not that critical.

posthog/management/commands/backfill_llm_clusters_access_control.py
```diff
@@ -0,0 +1,59 @@
...
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = (
+        "Mirror every existing 'llm_analytics' AccessControl row onto a matching 'llm_clusters' row. "
+        "Clustering used to inherit its access level from llm_analytics; now that it's an independent "
+        "resource, this backfill keeps existing grants working instead of silently defaulting to none. "
```
carlos-marchal-ph: This has the failure mode backwards: a resource with no rows defaults to editor, not none, so without the backfill restricted users get more access, not less. The taggers command's help text states it correctly.

frontend/src/lib/scopes.tsx
```diff
@@ -115,6 +115,7 @@ export const API_SCOPES: APIScope[] = [
     { key: 'link', objectName: 'Link', objectPlural: 'links' },
     { key: 'live_debugger', objectName: 'Live debugger', objectPlural: 'live debugger' },
     { key: 'llm_analytics', objectName: 'AI observability', objectPlural: 'AI observability' },
+    { key: 'llm_clusters', objectName: 'Cluster', objectPlural: 'clusters' },
```
carlos-marchal-ph: Current naming doesn't make what it's referring to clear:
```suggestion
    { key: 'llm_clusters', objectName: 'AI trace cluster', objectPlural: 'AI trace clusters' },
```

frontend/src/lib/utils/accessControlUtils.ts
```diff
@@ -61,6 +61,8 @@ export const pluralizeResource = (resource: APIScopeObject): string => {
         return 'customer analytics'
     } else if (resource === AccessControlResourceType.LlmAnalytics) {
         return 'AI observability'
+    } else if (resource === AccessControlResourceType.LlmClusters) {
+        return 'clusters'
```
carlos-marchal-ph: Same as my other suggestion (will stop tagging these, an LLM can probably get them all in one go)
```suggestion
        return 'AI trace clusters'
```

posthog/management/commands/backfill_llm_clusters_access_control.py
```diff
@@ -0,0 +1,59 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
```
carlos-marchal-ph: Is this ever used?
```suggestion
```

carlos-marchal-ph (approved):
Approving to not block but there's a couple of high level things I'd like to see addressed.

First, this references LLMA or `llm_analytics` everywhere due to inertia in the surrounding code, but our team is now AIO or `ai_observability`. In general we've been trying to rename as much as possible, so even though we won't backfill the existing rows (for now), we'd like to use the new name on new ones.

Second thing that's been mentioned by other reviewers is the gap in which everyone gets cluster access before the migration gets triggered manually. I agree that it's probably okay, but have we considered just running this as a sync migration instead of an async backfill? No idea how many rows we have, maybe it's not feasible, but just something to consider.

posthog/management/commands/backfill_llm_clusters_access_control.py
ReeceJones: Moved to a sync migration, so no longer relevant.

frontend/src/lib/utils/accessControlUtils.ts
ReeceJones: Let the robots run loose and update everything to use `AI Trace Clusters` terminology

frontend/src/lib/scopes.tsx
ReeceJones: ditto

posthog/management/commands/backfill_llm_clusters_access_control.py
ReeceJones: ditto

ReeceJones:
> Second thing that's been mentioned by other reviewers is the gap in which everyone gets cluster access before the migration gets triggered manually. I agree that it's probably okay, but have we considered just running this as a sync migration instead of an async backfill? No idea how many rows we have, maybe it's not feasible, but just something to consider.

Fair push, I ran the numbers and there eare roughly 450 rows that need to be migrated, so a sync migration should be nearly instant.

PR #71993, opened 2026-07-17
feat(max): share internal AI chats via direct link only

bs1180:
## Problem

Support agents can create PostHog AI chats while impersonating a customer (for example, to build a dashboard). Those chats are marked **internal** and hidden from the customer. But the internal flag also blocked the direct-link (retrieve) path, so when an agent shared a chat URL with the customer to show them how to use the result, it 404'd for them.

The ask: let a support agent hand over a link that a project member can open, **without** that chat ever appearing in the customer's chat history — and without adding a second, confusing "share link" button to the menu.

Raised in a Slack thread (link in footer).

## Changes

- New `shared_via_link` boolean on `Conversation` (+ migration).
- `retrieve` (direct link) now opens an internal conversation for any project member **once it has been shared**; `list` (history) keeps excluding internal conversations entirely. Net result: a shared internal chat is reachable only by people who have the link and never shows up in anyone's history.
- New impersonation-only `share` action that flips the flag.
- Frontend: for internal chats the actions menu shows a **single** clearly-labeled action (`Share chat with project` / `Copy shared link`) with a tooltip explaining link-only visibility, instead of piling a second copy-link item next to the existing one. Non-internal chats are unchanged.

_No screenshots — this environment can't run the app._

Automated backend tests added to `ee/api/tests/test_conversation.py` (run by CI; the agent could not execute them locally — no Python env in this environment):

- `test_retrieve_internal_conversation_requires_sharing` (parameterized) — a project member gets 404 for an unshared internal chat and 200 once it's shared. Catches regressions that would leak unshared internal chats or break shared-link access.
- `test_shared_internal_conversation_never_appears_in_history` — a shared internal chat stays out of the history list. Catches anyone accidentally applying the shared-link condition to `list` too.
- `test_share_conversation_requires_impersonation` / `test_share_conversation_during_impersonation_marks_shared` — the `share` action is 403 outside impersonation and flips the flag during it.

## 🤖 Agent context

Authored by the PostHog Slack app (Claude) from a Slack thread. Skills invoked while shaping this PR: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`.

Design decisions: considered simply un-marking chats as internal, but that would surface them in the impersonated user's own history — rejected. Chose to decouple direct-link retrieval from history listing via a dedicated `shared_via_link` flag, so link access and history visibility are independent. The migration was hand-written (Django output style) because no Python env was available to run `makemigrations`.

**Follow-up required:** `hogli build:openapi` needs to be run to regenerate the conversations OpenAPI client (`products/conversations/frontend/generated/*`) for the new `shared_via_link` field and `share` operation — the environment lacked a Python runtime to run codegen. The app itself uses the hand-written `Conversation` type and `api.conversations.share`, so nothing fails to compile; only the generated client/spec need reconciling (CI's OpenAPI drift check will confirm).

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C07TQR0V16U/p1784297018306229?thread_ts=1784297018.306229&cid=C07TQR0V16U)*


bs1180:
Hi @PostHog/team-self-driving 👋 — when you have a moment, could you take a look at this one? It's a small change to how internal (impersonation-created) AI chats can be shared: a `shared_via_link` flag that lets a support agent hand a customer a working direct link, while keeping the chat out of everyone's history. Touches `ee/api/conversation.py`, the conversation serializer + migration, and the chat actions menu in `scenes/max`. Backend tests are included. No rush, and thank you! 🙏

PR #71997, opened 2026-07-17
fix: don't render a broken side action on the empty retry button

gustavohstrassburger:
## Problem

`RetryButton` in the shared insight empty states always sets a `sideAction` object on its `LemonButton`, defaulting to `{}` when there's no `query` to link to a query debugger. Since `{}` is truthy, `LemonButton` still renders the side-action UI (a divider plus an empty side button) even though there's nothing to show, leaving a stray blank clickable area next to "Try again."

I ran into this while testing #67776 locally: `InsightErrorState`'s retry button on the person profile's feature flags tab calls `onRetry` without a `query`, and the button rendered with a visible empty side action instead of just "Try again."

## Changes

`sideAction` is now `undefined` when there's no `query`, so `LemonButton` only renders the side action when there's an actual query debugger link to offer.

image

Verified locally against #67776's flags-service error state (the button previously showed a stray side-action divider; after the fix it renders as a plain button). No behavior change for the `query` case, which already worked correctly.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

I found this while manually testing #67776's new error state in the browser. Claude Code diagnosed the root cause (an always-truthy `sideAction` default in `RetryButton`) and applied the fix.


frontend/src/scenes/insights/EmptyStates/EmptyStates.tsx
```diff
@@ -173,24 +173,23 @@ const RetryButton = ({
...
-                                to: urls.debugQuery(query),
-                            },
-                        ]}
-                    />
-                ),
-                placement: 'bottom-end',
-            },
-        }
-    }
+    const sideAction = query
```
haacked, line 176: `suggestion`: No test covers this fix, and `RetryButton` isn't exercised anywhere today. The two tests in `InsightErrorStates.test.tsx` never pass `onRetry`, so the retry button (and its `sideAction` branch) never renders. Reintroducing the old `sideAction = {}` default would render a stray `.LemonButtonWithSideAction` wrapper with no query, and nothing would fail.

Mirror the `reports "insight error message shown" when a server error renders` test above with a case that passes `onRetry` and no `query`, asserting `container.querySelector('.LemonButtonWithSideAction')` is null; add a second case with a `query` asserting it's present.

haacked (approved):
Nice fix! One non-blocking suggestion inline about test coverage.

PR #72000, opened 2026-07-17
fix(insights): rename 'Year to date' date preset to 'This year'

luke-belton:
## Problem

Someone went looking for a "This year" option in the date range picker and scanned the list several times before realizing it was there all along, just labeled "Year to date". The preset sits right between "This week" and "This month", but its odd-one-out naming made it hard to spot.

## Changes

Renames the date preset from "Year to date" to "This year" so it reads consistently with its neighbors. "This week", "This month", and this preset all run from the start of the period to now, so the naming now matches the behavior.

The change is label-only: the persisted query value (`yStart`) is unchanged, so saved insights, dashboards, and shared URLs keep working. Because the label doubles as a lookup key in several preset lists and chip/deny/allow lists, I updated every place that referenced the string.

Revenue analytics is deliberately left alone: it already has a separate "This year" (the full calendar year) alongside "Year to date" (start of year to now), so those two are genuinely different there.

Updated the `dateRangeSelection` round-trip test that asserts `yStart` maps to the preset name. I (the PostHog Slack app) did not run the app manually. Automated coverage: the round-trip test in `dateRangeSelection.test.ts` verifies the `yStart` ↔ preset-name mapping still holds after the rename.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Raised from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1784300639369879?thread_ts=1784300639.369879&cid=C0ACRAMJUAG): a user found "Year to date" hard to find because they were scanning for "This year".

The label is used as an internal lookup key (`DateMappingOption.key`), not just display text, so I traced every reference before editing — the canonical `dateMapping`, the per-product duplicates (retention, customer analytics, endpoints, engineering analytics), and the chip/deny/allow lists plus the `fixedRange` switch in the quill date picker. Left revenue analytics as-is since it distinguishes full-year from year-to-date, and left quill's generic component example fixtures alone.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1784300639369879?thread_ts=1784300639.369879&cid=C0ACRAMJUAG)*

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -1713,9 +1713,9 @@ snapshots:
     components-property-key-info--property-key-info--light:
         hash: v1.k794b7964.2f0245e1b92aba6695884bdf695034bf1efd85caca619940965a08338f15e41c.5Z7YHy2M9ko_15_blb0NaI7WmQeXytUWAntBAs03Qr0
     components-quill-date-filter--date-filter--dark:
-        hash: v1.k794b7964.95c2626b3630789e4e2d50e6e7b079f9ef08687394eb554aa9b71b63c49013ba.G6jYIZtuERuM2gcpHIzsC0DnqeyFryQvG3ufGAwtDN4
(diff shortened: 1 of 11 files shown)


thmsobrmlr (approved):
TY!

PR #72016, opened 2026-07-17
fix(batch-exports): group schedule fields in config form

rossgray:
## Problem

The `timezone` field in the batch export config form is potentially confusing — it could be interpreted as controlling the timezone of the data written to the destination (e.g. BigQuery), when it only affects when scheduled runs fire. The schedule-related fields (interval, start time, timezone) weren't visually grouped, so it wasn't obvious they all describe the run schedule.

## Changes

- Added a "Schedule" section heading and short description to the card that already holds the interval, timezone, and start-time fields, so they read together as scheduling settings.
- Added a matching "Data" heading to the adjacent model/filters card for consistency.
- Sharpened the timezone field tooltip to make clear it only affects run scheduling, not the timezone of the exported data.
- Moved the timezone field below the start time, so it's slightly more obvious what the timezone is associated with.

Ran the stack locally

## Screenshot

image

## 🤖 Agent context

Small copy/layout change scoped to `frontend/src/scenes/data-pipelines/batch-exports/BatchExportConfiguration.tsx`. No skills required (pure `.tsx` change, no serializer/API types touched). The schedule fields were already in a single bordered card but lacked a header; the fix makes that grouping explicit and disambiguates the timezone field via its tooltip.

---

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -5187,9 +5187,9 @@ snapshots:
     scenes-app-batchexports--backfills-with-estimates--light:
         hash: v1.k794b7964.bdb71e856369ccdaacaf893d0e9427b9bfeb6513333c4af98ca37cb9fb687896.kQuW_-2hLUIZa0vOFnn_FuNuy3d7ALzGXm5P7zamdVc
     scenes-app-batchexports--existing-big-query-export--dark:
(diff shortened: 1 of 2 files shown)


tomasfarias (approved):
lgtm

PR #72021, opened 2026-07-17
feat(cohorts): processor seed consumer

matheus-vb:
## Problem

The behavioral cohort streaming pipeline only tracks membership from events it sees live, so a newly streamed cohort starts with no history. A new separate seeder will emit per-person day-tiles that carry that history, but nothing on the processor side consumed them. This PR adds that consumer.

## Changes

Adds a seed consumer to the cohort stream processor, entirely behind `COHORT_SEED_CONSUMER_ENABLED` (default off). It consumes the `cohort_stream_seed_events` topic as a follower co-partitioned with the live event topic, routes each seed tile to the owning partition worker, and folds it into stateful cohort membership.

The apply path re-checks each leaf's time window, slides it forward before evaluating, max-merges the tile's day count into the person's bucket history, recomputes the eviction deadline, runs stage-2 evaluation, and emits membership changes tagged with `origin: "seed"` and the seeding `run_id`. Tiles for a person whose partition moved are redirected and re-produced to the new owner, with a hop cap to bound re-routing. Application is gated by an apply fence: a tile is held until a per-partition live watermark passes the tile's chunk boundary plus a safety margin, so backfill never races ahead of live ingestion and double-counts.

Supporting pieces:

- A pure apply core that is idempotent, convergent across tile orderings, and equivalent to applying the same events live.
- `LiveWatermarks` fed at worker fold time, plus an idle probe so quiet partitions still advance their watermark and open the fence.
- Fenced or backpressured tiles enter a per-partition holdover and pause the seed partition, reusing the events backpressure path; seed offsets commit on their own tracker only after the tile is durably applied.
- Additive, byte-stable wire changes: a `redirect_hops` field on the seed tile and optional `origin` / `run_id` on the membership change, both skipped when unset so existing serialized bytes are unchanged.
- The seed topic wired into the durability checkpoint manifest and revoke teardown.

New Rust tests cover the pure apply core (idempotency, convergence over permutations, live-equivalence with the event path, dedup untouchability, and slide-drop / resurrect-flap regressions), the fence decision and watermark monotonicity, the handler (inline, cross-partition, and hop-capped redirect, origin tagging, in-order skip marks), the worker loop order barrier, and producer byte stability. A real-broker integration suite is `#[ignore]`d.

## Docs update

No.


rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+                warn!(
+                    partition_id,
+                    team_id = tile.team_id().0,
+                    "seed tile re-key produce failed; holding the seed offset for redelivery",
+                );
+                hold(&merge.seed_tracker, partition_id, offset);
+            }
+            return;
+        }
+        TileRoute::CapExhausted { person } => {
```
gustavohstrassburger, line 536: `question`: The `CapExhausted` arm in `handle_seed` applies the tile inline at the current partition, but `person` here is `target_person` from a cross-partition resolution — a person the tombstone resolver says is owned by a different partition. The live event path for that person reads/writes state under the survivor's owning partition, so this inline write lands in a shard the live path never touches. Not a double-count, but the seed's history for this person is effectively orphaned: neither merged with, nor visible to, live evaluation. That diverges from the PR's "equivalent to applying the same events live" claim in an observable way.

The PR frames this as a bounded degraded arm for a "corrupt tombstone cycle" that "shouldn't happen," and it's warned + counted (`SEED_REKEY_HOP_CAPPED_TOTAL`), so this is likely acceptable-by-design. Is orphaning the tile the intended behavior when the hop budget is exhausted, versus dropping it outright (which a later reconcile pass could heal cleanly)? An inline apply at the wrong partition writes durable state that reconcile then has to reconcile against; a drop leaves nothing to clean up. Worth a one-line comment on the arm stating the tradeoff either way.

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+            hold(&merge.seed_tracker, partition_id, offset);
+            return;
+        }
+    }
+
+    tag_seed(&mut changes, tile.run_id());
+    if !changes.is_empty() {
+        let errors = produce_membership(sink, changes).await;
+        if errors > 0 {
+            // Accepted at-most-once: the state committed above, so the redelivery lands
```
gustavohstrassburger: `question`: In the seed-membership produce-failure path, the code holds the offset for redelivery, but state was already committed a few lines earlier. On redelivery the tile lands `Unchanged`, so the stage-1 transition isn't re-minted, only the stage-2 recompose re-runs. The comment here acknowledges this: a single-leaf membership change lost here isn't re-emitted, and "the reconcile snapshot heals this class." So a produce failure on a seed flip is at-most-once for the stage-1 signal even though the offset is held, the one place where "held offset means nothing lost" isn't strictly true.

Consistent with documented at-most-once semantics elsewhere, so not blocking. Is relying on the reconcile snapshot the intended recovery here, or should the replay path re-emit the flip it could reconstruct from prev-vs-current state?

rust/cohort-stream-processor/src/consumers/seeds.rs
```diff
@@ -0,0 +1,904 @@
...
+            Err(err) => {
+                debug!(topic, partition, error = %err, "idle probe: watermark fetch failed; partition stays fenced");
+                continue;
+            }
+        };
+        let frontier = folded
+            .get(&partition)
+            .copied()
+            .or_else(|| committed.get(&partition).copied());
+        let is_idle = match frontier {
```
gustavohstrassburger: `suggestion`: This "is this partition idle" classification has zero unit coverage. The tested half is `advance_probed_idle`, which only re-checks ownership, the classification feeding it is never exercised directly, and the integration suite can't cover it either (a comment there notes the probe "effectively never" fires in that suite since every test drives the watermark through live folds instead).

This sits on the fence's fail-open direction: if the predicate is wrong (a folded frontier one short of `high` gets called idle, or the `high == low` boot-edge fallback misfires), a partition with unfolded live events still in flight gets declared idle, its watermark jumps to `now_ms`, the fence opens early, and a backfill tile double-counts against live ingestion that hasn't landed yet, the exact failure the fence exists to prevent.

Extract the predicate so it's testable without a live broker, and table-test it:

```suggestion
        let is_idle = frontier_is_caught_up(frontier, low, high);
```

with a standalone `fn frontier_is_caught_up(frontier: Option<i64>, low: i64, high: i64) -> bool` mirroring the current match arms. Cases worth pinning: folded frontier == high (idle), == high-1 (not idle), no frontier + empty partition high==low (idle boot edge), no frontier + non-empty high>low (not idle). `probe_advance_never_recreates_a_watermark_revoked_mid_probe` is a good sibling test to mirror for structure.

rust/cohort-stream-processor/src/partitions/router.rs
```diff
@@ -247,8 +247,8 @@ impl PartitionRouter {
         // Refuse (→ hold → pause) once the partition holds its event ceiling, before the batch reaches
         // the mpsc slot. Count events only, so an event-less batch reserves 0 and stays balanced with
         // the receiver's release.
-        let events = count_events(&batch);
-        if channel.intake.try_admit(events) == Admission::Rejected {
+        let counted = count_intake(&batch);
+        if channel.intake.try_admit(counted) == Admission::Rejected {
```
gustavohstrassburger, line 251: `suggestion`: Seeds and live events share one per-partition intake budget (cap 1024) and one worker mpsc channel, with no priority for live events. Both `dispatch_events_nonblocking` and `dispatch_seeds` admit through this same `try_admit` and both `try_send` into the same channel from `ensure_worker`. On a hot partition during a backfill, a seed sub-batch of up to `recv_batch_size` (1000) tiles can be enqueued ahead of a live-event batch. Since the channel is FIFO and the worker drains a whole batch before its next `recv()`, live events behind that seed batch wait for every queued seed tile to process first, each doing a Maintenance-lane read, a state commit, a stage-2 compose, and a membership produce. That's real head-of-line latency on live cohort membership updates during an active backfill.

The apply fence and Maintenance read lane bound correctness and read contention, but neither bounds this queueing delay. Two low-cost options: reserve a fraction of the intake cap for events so seeds can't consume the whole budget, or cap the per-cycle seed dispatch per partition so a seed batch can't monopolize the channel ahead of live events.

Worth confirming with a load test before shipping at high backfill volume: run a backfill against a partition with live traffic and check p99 live-event fold-to-produce latency against the seed-only baseline. If it stays flat, the shared budget is fine as-is.

rust/cohort-stream-processor/src/consumers/seeds.rs
```diff
@@ -0,0 +1,904 @@
...
+                .seed_tracker
+                .committable_offsets(),
+            &self.dispatcher.owned_partitions(),
+        )
+    }
+}
+
+/// Decode one payload into ordered work. Every non-tile outcome is a channel-riding skip, never a
+/// drop: its offset must mark in order (see the module doc).
+fn decode_payload(payload: Option<&[u8]>, partition: i32, offset: i64) -> SeedWork {
```
gustavohstrassburger, line 428: `question`: The unsupported-schema handling here silently and permanently drops data on a forward-incompatible schema bump. This classifies any tile with a schema version the deserializer rejects as `SeedSkipReason::UnsupportedSchema`, an in-order skip: the offset commits and the tile is gone, no replay path. Right call for not wedging a partition, but it means if the seeder ever ships a new schema version before the processor is upgraded to understand it, every one of those tiles is permanently lost the moment the offset commits.

Since both the seeder and this consumer are new and the whole path is behind a default-off flag, this isn't a regression today. But is the intended deploy ordering "processor understands the new schema before the seeder emits it," and if so, is that constraint documented anywhere a future on-call engineer doing a coordinated rollout would find it? A one-line note in the module doc or near the schema-version skip path would be enough.

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+        ) {
+            LeafMergeOutcome::Merged {
+                record,
+                transition,
+                deadline_ms,
+            } => {
+                counter!(SEED_TILES_APPLIED_TOTAL, "variant" => record.state.variant().as_str())
+                    .increment(1);
+                staged.put::<Behavioral>(&key, &record.encode());
+                // The kind is a placeholder for the no-flip case: compose_stage2 reads only the
```
gustavohstrassburger: `suggestion`: The `stage2_leaves` pushes here synthesize a fake `TransitionKind::Entered` to smuggle `(leaf, person)` into `compose_stage2`, which only ever reads `leaf_state_key` and `person_id`. On the `Unchanged` arm, this reads as pure noise, a future maintainer debugging membership will see "Entered" on a leaf that provably didn't change and lose time chasing it. The comment here flags it honestly, but a comment guarding a landmine is weaker than removing the landmine.

The cleaner fix is having `compose_stage2` accept what it actually consumes, `&[(LeafStateKey, Uuid)]`, dropping the placeholder kind entirely. That touches every call site (event/sweep callers would map `transitions.iter().map(|t| (t.leaf_state_key, t.person_id))`), so it's fair to defer, but the current shape leaves a synthesized-transition idiom that reads as a lie.

rust/cohort-stream-processor/src/producer/mod.rs
```diff
@@ -39,6 +47,21 @@ pub struct CohortMembershipChange {
...
+    pub origin: Option<ChangeOrigin>,
+    /// The backfill run that emitted this change; set iff `origin` is.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub run_id: Option<RunId>,
+}
+
+/// Non-live provenance of a membership change, for downstream batching/deprioritization.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ChangeOrigin {
```
gustavohstrassburger, line 61: `question`: `ChangeOrigin::Reconcile` is defined but never constructed anywhere in this PR, the only producer is `ChangeOrigin::Seed`. The doc comment says it's "Reserved for the reconcile snapshot." Small YAGNI: if the reconcile consumer lands in a later PR, adding the variant then is itself an additive, byte-stable change by the same `skip_serializing_if` reasoning already relied on here. Is there a concrete reason to land it now (e.g. freezing the wire enum's serde spelling ahead of a downstream schema)? If so, worth a one-line note; otherwise consider dropping it until the reconcile path exists.

gustavohstrassburger (approved):
Solid, well-tested addition. A few questions and one real test-coverage gap inline, nothing blocking.

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+                "seed stage 2 composition failed; holding the seed offset for redelivery",
+            );
+            hold(&merge.seed_tracker, partition_id, offset);
+            return;
+        }
+    }
+
+    tag_seed(&mut changes, tile.run_id());
+    if !changes.is_empty() {
+        let errors = produce_membership(sink, changes).await;
```
haacked, line 646: `blocking`: When both `COHORT_CASCADE_ENABLED` and `COHORT_SEED_CONSUMER_ENABLED` are on, flipping a base cohort leaves composite cohorts permanently stale. `handle_seed` doesn't call `first_cascades`/`produce_cascades`, unlike the event-batch flush, `handle_sweep`, and `produce_merge_transitions`.

`compose_stage2` only resolves direct references; cohort-of-cohort composites are reached only through the cascade topic and are never re-evaluated by sweep or live events. The cascade machinery is already available here (`MergeWorkerDeps` carries a `cascade_sink`), but no test covers this path.

Add a `first_cascades`/`produce_cascades` call after `produce_membership` succeeds, holding the seed tracker on cascade-produce failure to match this function's existing fail-stop posture. If composite healing is deliberately deferred instead, document that explicitly in code near the produce call so it doesn't read as an oversight, and add a test with cascade enabled asserting the cascade fires (or is deliberately empty).

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+    let Some(team_filters) = snapshot.team(tile.team_id()) else {
+        counter!(SEED_TILES_DROPPED_TOTAL, "reason" => "team_absent").increment(1);
+        mark_processed(&merge.seed_tracker, partition_id, offset);
+        return;
+    };
+    let filters: &TeamFilters = team_filters;
+
+    // Same resolver as the event path. A read failure is fail-stop — a mis-applied tile for a
+    // merged-away person is durable state reconcile cannot retract.
+    let resolution = match tombstone_redirect::resolve_offloaded(
```
haacked, line 494: `suggestion`: This tombstone preflight read (`resolve_offloaded` -> `read_tombstone_offloaded` -> `get_tombstone`) hardcodes `ReadLane::Event`, while the behavioral read and stage-2 compose right after it correctly use `ReadLane::Maintenance`. At backfill volume this puts thousands of Event-lane read permits per second against the live fold's permit pool, undoing the lane isolation the surrounding code exists to guarantee.

Thread a `ReadLane` parameter through `resolve_offloaded`/`read_tombstone_offloaded`/`get_tombstone`, the same way it was already done for `get_stage2`/`multi_get_stage2`/`get_person_record`. The live event call site in `worker.rs` would pass `ReadLane::Event`; here it'd pass `ReadLane::Maintenance`.

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+        record,
+        transition: kind.map(|kind| identity.transition(kind)),
+        deadline_ms,
+    }
+}
+
+/// Handle one seed message on its owning partition worker (the imperative shell). Runs after the
+/// worker's buffered-changes order barrier; marks or holds the **seed** tracker only.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn handle_seed(
```
haacked, line 464: `suggestion`: This function runs about 267 lines, roughly triple its siblings `handle_apply` and `handle_event`, because it inlines the per-leaf accumulation loop alongside the ordered commit/produce/mark sequence whose order is the actual contract.

Extracting the per-leaf loop (building the staged batch and transition lists) into a helper, e.g. `apply_tile_to_leaves`, returning a small struct of staged/transitions/stage2_leaves/schedules, would bring this back in line while keeping the ordered commit -> stage2 -> produce -> mark sequence inline.

rust/cohort-stream-processor/tests/seed_consumer.rs
```diff
@@ -0,0 +1,811 @@
...
+    tasks.push(tokio::spawn(events_consumer.process()));
+
+    Instance { dispatcher, tasks }
+}
+
+/// A tile produced with the seeder's keying lands on its owning worker via the 5-topic
+/// co-assignment, applies behind an already-open fence, emits an `origin: seed`-tagged membership
+/// change, and the seed group's committed offsets reach the produced high-water mark — the
+/// run-completion precondition.
+#[tokio::test]
```
haacked, line 633: `suggestion`: This real-broker E2E suite (811 lines) is `#[ignore]`d, and `ci-rust.yml` only runs `--run-ignored only` for the cymbal package, not `cohort-stream-processor`, so these tests never execute in CI. The seed consume loop (fence-split, dispatch, pause, commit, rebalance co-assignment) has no automated signal beyond unit tests.

A gated CI step for `cohort-stream-processor` mirroring the existing cymbal `--run-ignored only` block would get the broker-backed suite running in CI instead of only locally.

rust/cohort-stream-processor/src/consumers/seeds.rs
```diff
@@ -0,0 +1,904 @@
...
+            }
+        }
+    }
+    split
+}
+
+/// Per-partition holdover of fence-closed or backpressured seeds. Mirrors
+/// [`Backpressure`](crate::partitions::backpressure::Backpressure): a present entry is always
+/// non-empty, so `held_partitions()` is exactly the paused target.
+#[derive(Debug, Default)]
```
haacked: `suggestion`: `SeedHoldover` is a near-verbatim copy of the existing `Backpressure` type: same `pending: HashMap<i32, Vec<T>>`, same `prune_revoked`/`take_held`/`held_partitions`/`absorb`/`held_partition_count` shape, differing only in element type and whether `take_held` flattens results. The doc comment even says it mirrors `Backpressure`.

Extracting a generic `PartitionHoldover<T>` with the shared methods would let `Backpressure` become `PartitionHoldover<ShuffleMessage>` and `SeedHoldover` become `PartitionHoldover<ConsumedSeed>`, with `take_held` returning the grouped `Vec<(i32, Vec<T>)>` shape and `seeds.rs` flattening at the call site.

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+                stage2_leaves.push(identity.transition(TransitionKind::Entered));
+            }
+            LeafMergeOutcome::Dropped(reason) => {
+                counter!(SEED_TILES_DROPPED_TOTAL, "reason" => reason.as_str()).increment(1);
+            }
+        }
+    }
+
+    // Order copied from the event path: state commit → stage 2 compose → produce → schedule → mark.
+    if !staged.is_empty() {
```
haacked, line 595: `question`: This function commits state and produces membership once per tile with no cross-tile batching, capping backfill throughput at one RocksDB commit plus one Kafka produce per seed message, unlike the live event path which batches via `OutputBuffer`. Is per-tile commit the intended steady-state throughput ceiling for backfill, or is a batched seed-apply planned as a follow-up?

rust/cohort-stream-processor/src/workers/seed_path.rs
```diff
@@ -0,0 +1,1965 @@
...
+        counter!(SEED_TILES_DROPPED_TOTAL, "reason" => "no_referencing_leaves").increment(1);
+        mark_processed(&merge.seed_tracker, partition_id, offset);
+        return;
+    }
+
+    let prefix = PersonPrefix::new(partition_id, tile.team_id().0 as u64, person);
+    let keys: Vec<BehavioralKey> = lsks.iter().map(|&lsk| prefix.behavioral_key(lsk)).collect();
+    // The sweep's lane: backfill must not contend with live event reads.
+    let values = match handle
+        .multi_get_behavioral(keys.clone(), ReadLane::Maintenance)
```
haacked: `nit`: This clones the key vector only so `keys` can be re-borrowed in the zip a few lines later. Each key can be reconstructed from `prefix.behavioral_key(lsk)` inside the loop instead, avoiding the clone.

rust/cohort-stream-processor/src/observability/metrics.rs
```diff
@@ -431,6 +431,48 @@ pub const STAGE2_ORPHAN_GC_UNDECODABLE_KEYS_TOTAL: &str = "stage2_orphan_gc_unde
 /// `popped == evicted + dropped`.
 pub const SWEEP_KEYS_DROPPED_TOTAL: &str = "sweep_keys_dropped_total";
 
+/// `cohort_stream_seed_events` payloads consumed and successfully decoded — tiles and ordered
```
haacked: `nit`: The ~15 new seed metrics have no `seed_metric_names_are_stable` test, though every other metric family in this file does, and the module comment states these names are a dashboard contract that a rename cannot silently break.

rust/cohort-stream-processor/src/observability/metrics.rs
```diff
@@ -431,6 +431,48 @@ pub const STAGE2_ORPHAN_GC_UNDECODABLE_KEYS_TOTAL: &str = "stage2_orphan_gc_unde
...
+/// Tile re-key produces that failed; the seed offset is held for redelivery (counter).
+pub const SEED_REKEY_PRODUCE_FAILURE_TOTAL: &str = "cohort_seed_rekey_produce_failure_total";
+/// Tile redirects that exhausted `redirect_hops` and were applied inline at the best-known target
+/// (counter). Non-zero means a corrupt tombstone cycle.
+pub const SEED_REKEY_HOP_CAPPED_TOTAL: &str = "cohort_seed_rekey_hop_capped_total";
+/// The seed commit floor pinned by a sticky offset hold, labelled by `partition` (gauge). Alert on
+/// a sustained non-zero level — same fail-stop semantics as [`MERGE_HELD_OFFSET_GAUGE`].
+pub const SEED_HELD_OFFSET_GAUGE: &str = "seed_held_offset";
+/// Seed partitions currently held (fence-closed or backpressured) — the paused target size (gauge).
+pub const SEED_FENCED_PARTITIONS: &str = "seed_fenced_partitions";
```
haacked: `nit`: These two fence gauges are named `seed_fenced_partitions` / `seed_fence_deficit_ms` without the `cohort_seed_` prefix the rest of the apply-side seed metrics use, reading as an accidental drop of the prefix.

rust/cohort-stream-processor/src/producer/mod.rs
```diff
@@ -29,8 +32,13 @@ pub use merge::{
     CaptureStreamEventSink, CaptureTransferSink, KafkaStreamEventSink, KafkaTransferSink,
     StreamEventSink, TransferSink,
 };
+pub use seed::{CaptureSeedTileSink, KafkaSeedTileSink, NoopSeedTileSink, SeedTileSink};
 
 /// One per-cohort membership change on `cohort_membership_changed_shadow`.
+///
+/// `origin`/`run_id` are additive and absent on the live path (`skip_serializing_if`), so live
```
haacked: `nit`: This says downstream JSONEachRow consumers read with `input_format_skip_unknown_fields`, but that setting isn't configured anywhere in the repo. The only consumer of the shadow topic is a Python parity drain using plain tolerant `json.loads`; the ClickHouse tables read the non-shadow topic and never see these fields.

Rewording to describe the actual consumer (the parity drain's tolerant JSON decode) would keep a future reader from assuming a ClickHouse strict-schema guarantee that doesn't exist for this topic.

rust/cohort-stream-processor/src/consumers/seeds.rs
```diff
@@ -0,0 +1,904 @@
...
+            self.dispatcher.clone(),
+            self.idle_probe_interval,
+            self.handle.clone(),
+        ));
+
+        let mut holdover = SeedHoldover::default();
+        let mut prev_paused_target: HashSet<i32> = HashSet::new();
+        let mut commit_deadline = tokio::time::Instant::now() + self.offset_commit_interval;
+
+        loop {
```
haacked, line 261: `nit`: `SeedFollowerConsumer::process`/`consume_batch` restate the commit-deadline block and `owned_committable_offsets` wrapper from the generic `FollowerConsumer` template almost verbatim, on top of the seed-specific fence/holdover/idle-probe logic that justifiably isn't shared.

haacked (commented):
Nice work landing a hard problem cleanly. One blocking issue on cascade fan-out, a handful of suggestions inline.

rust/cohort-stream-processor/src/workers/seed_path.rs
matheus-vb: Intended, it mirrors the event path’s hop cap degrade exactly, so both inputs treat a corrupt cycle the same way. The orphaned slice is bounded and ages out via its eviction deadline, whereas a drop would silently lose the tile with nothing for reconcile to even see. Added a comment on the arm stating the tradeoff.

rust/cohort-stream-processor/src/workers/seed_path.rs
matheus-vb: Relying on the reconcile snapshot is intended. Re-emitting from prev vs current state on replay would need a second read pass plus flip reconstruction for a rare failure window, and the reconcile snapshot already heals exactly this class deterministically. Marked as a conscious acceptance in a comment at the hold site.

rust/cohort-stream-processor/src/partitions/router.rs
matheus-vb: Deferring to the live traffic protection slice. The designed control there is pausing seed partitions on live lag, which bounds this end to end; an intake reservation or per cycle cap wouldn’t, since the worker channel drain is the actual bottleneck and seeds would still fill it across cycles. Agreed on the acceptance criterion: we’ll compare live p99 fold to produce latency against the seed off baseline.

rust/cohort-stream-processor/src/consumers/seeds.rs
matheus-vb: Yes, the intended ordering is processor understands the new schema before any seeder emits it. Now documented on the `UnsupportedSchema` variant, including that the skip commits and never replays.

rust/cohort-stream-processor/src/producer/mod.rs
matheus-vb: No concrete reason to land it now. Dropped the variant; the enum notes that new origins are additive by the same serde reasoning, so reconcile adds its own when its producer lands.

rust/cohort-stream-processor/src/workers/seed_path.rs
matheus-vb: Per tile is the intended v1 ceiling. A tile is one or two store ops, so the throughput budget was sized on per tile commits and fits the target rate; cross tile batching would also blur the per offset hold and redelivery semantics. If measured tile throughput falls short in validation, batching is the follow-up.

PR #72025, opened 2026-07-17
feat(autoresearch): cut metrics rate-query p50 latency 93ms to 38ms

DanielVisca:
## Problem

The Metrics product is in alpha (recently promoted to the visible nav, alerting MVP shipped). Its dashboard tiles and its alerting checks (which evaluate every returned series on each check interval) both go through `MetricQueryRunner`'s `rate`/`increase`/`histogram_quantile` aggregations — the highest-traffic query path in the product.

This started as an autoresearch loop (investigate the alpha product, find the first bottleneck, iterate on it with measurements). Created with Autoresearch. Benchmarked against real accumulated local dev telemetry (a 181k-row counter metric with 18 label variants over 6 days), the rate + group-by query started at **92.9ms p50** wall / ~75ms ClickHouse-side, with a tail to 159ms. After six measured iterations it sits at **38.4ms p50** / ~31ms ClickHouse-side.

## Changes

Each change was measured in isolation (median of 7 runs, `system.query_log` cross-checks); regressing experiments were reverted (one is documented below).

**1. Series identity is hashed, never stringified** (`cityHash64` instead of `toString(attributes)` as the window partition / group key). A Map-to-String conversion per row existed purely to get a comparable key; the fixed-width hash mirrors what the table's own `resource_fingerprint` column already does. `92.9 → 88.6ms`

**2. rate/increase hash-aggregates samples into (series, bucket) rows before windowing.** The old shape sorted every raw sample (~181k rows) through a `lagInFrame` window just to diff consecutive samples. Consecutive clamped deltas within a bucket are computable from a per-bucket timestamp-sorted value array (`groupArray` + `arraySort` + `arrayDifference` with the same reset-clamp semantics element-wise), so the window now only carries each bucket's boundary — its first sample diffs against the previous bucket's last — over series×buckets rows (~4.9k). Counter resets, first-sample, delta-temporality, and cross-bucket boundary semantics are identical (all pinned by existing tests). Window memory dropped 133MB → 21MB. `~60 → 51.6ms`

**3. Group labels are computed once per (series, bucket), not once per input sample.** Aggregate arguments are evaluated for every row, so the previous `any(toString(<map lookups>))` label paid two map lookups + `toString` on every sample (~15ms of the server time, per direct probe decomposition). The innermost `GROUP BY` now carries `any()` copies of only the raw columns the requested labels need (`__label_svc`/`__label_res`/`__label_attrs`), and the label expressions run one layer up. The series key also hashes the raw `attributes_map_str` column (newly exposed as a hidden HogQL field) instead of the `attributes` ALIAS, which mapApply-strips key tags per row. `48.4 → 40.7ms`

**4. Metrics queries execute against a cached static-schema context.** These queries are authored in the runner against `posthog.metrics` only, with every user input bound as a constant — they can never reference warehouse tables, views, joins, or group mappings. Both `execute_hogql_query` call sites now pass a prebuilt `HogQLContext` whose `Database` is `lru_cache`d per `(timezone, week_start_day)`, skipping several Postgres/personhog round trips plus a ~5ms static-schema reconstruction per query (profiled). The executor explicitly supports caller-provided contexts, and the path already ran with `user=None`, so the ACL posture is unchanged. `51.6 → 48.4ms and 40.7 → 38.4ms`

Also tried and reverted: a redundant `time_bucket` bound in the WHERE clause — `EXPLAIN` + `query_log` showed byte-identical reads (ClickHouse's key analyzer already derives it from the `timestamp` filter).

> [!NOTE]
> One measured tradeoff: the `any(map)` label carriers raise `read_bytes` (~22MB → ~66MB on the benchmark query) while cutting CPU/wall time, because whole map columns flow into the aggregation instead of per-row derived strings. Locally page-cached this is a clear win; on cold prod data the balance could differ. Worth revisiting once the query runs on the prod logs cluster.

Not in this PR (needs a ClickHouse migration + human review, documented from the same investigation): the table's `ORDER BY` doesn't prefix cleanly on `metric_name`, so single-metric queries read ~2.5x more rows than match (457k read / 181k matching); the `idx_metric_name_set` skip index is `set(100)` while local dev alone already has 217 distinct metric names; and the table's `projection_aggregate_counts` projection is never used by any current query shape.

Automated only:
- All 126 tests across `test_metric_query_runner.py` (105), `test_anomaly.py`, and `test_metrics_query_runner.py` pass at every iteration — including the counter-reset, cross-bucket-boundary, delta-temporality, per-series-isolation, group-by, and histogram cases that pin the exact semantics the query rewrite had to preserve.
- 119 HogQL database serialization tests pass — the new `attributes_map_str` field is `hidden=True`, so no schema snapshot drift.
- 2 SQL-shape regression tests (added in iteration 1, updated as the shape evolved) assert the printed ClickHouse SQL keys series identity by `cityHash64(...)` and never by `toString(<map>)`; both were verified to fail against the pre-fix code before being finalized.
- Benchmarks: median-of-7 timed runs via `manage.py shell` against real accumulated dev-stack telemetry in local ClickHouse (5.8M rows / 972 series), cross-checked against `system.query_log` (`query_duration_ms`, `read_rows`, `read_bytes`, `memory_usage`) after every iteration. Server-side shape probes (`clickhouse-client --time --format Null`) were used to decompose costs before choosing each change.
- Not run against production-scale or cold-cache data; the read-amplification finding above suggests prod behavior will be dominated by the schema-level issues this PR deliberately leaves untouched.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A — internal query implementation detail, no user-facing behavior change.

## 🤖 Agent context

Autoresearch optimization loop: investigate the Metrics alpha product, identify the first bottleneck, set the metric, iterate with one measured change per iteration. Created with Autoresearch.

- Skills invoked: `/optimizing-clickhouse-and-hogql-queries` (SQL extraction, EXPLAIN verification, measure-before-suggesting discipline, probe decomposition) and `/writing-tests` (the SQL-shape regression guards).
- Method: every iteration = hypothesis → one focused change → median-of-7 benchmark + `system.query_log` cross-check; correctness gated by the full consuming test suites before any measurement counted. Probes against ClickHouse directly (`--time --format Null`) decomposed server cost to target changes at the dominant term instead of guessing.
- Decisions along the way: an `IN (SELECT DISTINCT service_name ...)` prefilter was designed and rejected before implementation (distributed-table `IN` semantics differ between the single-node local bench and the prod logs cluster — `distributed_product_mode=deny` risk); a prepared-SQL/value-rebinding cache was rejected as wrong-data-risk; `groupArraySorted` was rejected because its N-cap silently truncates dense series.
- Commits are unsigned (1Password signing agent unavailable in-session, matching prior precedent on this repo).

---


DanielVisca:
Im concerned this way cut query latency at the cost of disk space

DanielVisca:
Concerned this is cutting 1 metric at the cost of another. Deprioritizing investigation

PR #72030, opened 2026-07-17
feat(replay-vision): ship pipeline logs to Logs via a structlog mirror

arnohillen:
## Problem

Replay Vision's "ship pipeline logs into the Logs product" bridge shipped nothing in production.
It attached a stdlib `logging.Handler` to the pipeline's logger namespace,
but the Temporal worker reconfigures structlog with a non-stdlib logger factory
([`posthog/temporal/common/logger.py`](https://github.com/PostHog/posthog/blob/master/posthog/temporal/common/logger.py))
that bypasses stdlib logging entirely.
So the handler was never invoked, no records reached the Logs product, and the failure was silent (no export attempts, no errors).

I confirmed the gap against the live Logs product: the deployed worker was running the bridge code and actively emitting the target loggers, but they only reached Logs via stdout scraping (service `temporal-worker-replay-vision`), never under `service.name = replay-vision`.
The original unit test passed because it logged through a stdlib logger directly, a path the worker never uses.

```mermaid
flowchart LR
    A[pipeline log] --> B{{worker structlog chain, non-stdlib factory}}
    B --> R[renderer to stdout scrape]
    D[stdlib OtelLogHandler] -. never fires .-> E[Logs product: replay-vision]
    class A,E phYellow;
    class B phBlue;
    class R,D phGray;
```

## Changes

Mirror the logs with a structlog processor instead of a stdlib handler, integrated into the worker's real log chain.

- `posthog/otel_logs.py`: `otel_log_mirror_processor(...)` returns a fail-soft structlog processor that mirrors matching records to the Logs product over OTLP. It is scoped by logger-name prefix and always fail-closed (only allowlisted operational fields ship, and an exception contributes just its type). It reuses the existing lazy per-process OTLP provider.
- `posthog/temporal/common/logger.py`: a new opt-in `otel_log_mirror` param on `configure_logger`, inserted before the terminal renderer so it sees the structured event dict. No change for any other worker.
- `posthog/management/commands/start_temporal_worker.py`: wire the mirror for the Replay Vision task queue only.
- `products/replay_vision/backend/temporal/logs.py`: `build_vision_log_mirror()` with the service name, namespace prefix, and the fail-closed allowlist. Drops the dead `install_vision_log_bridge()` stdlib-handler path.

The OTLP provider is warmed at worker startup so the exporter's background thread is never spawned lazily inside the Temporal workflow sandbox.

```mermaid
flowchart LR
    A[pipeline log] --> B{{worker structlog chain}}
    B --> M[otel mirror tap]
    B --> R[renderer to stdout scrape]
    M --> E[Logs product: replay-vision]
    class A,E phYellow;
    class B,M phBlue;
    class R phGray;
```

> [!NOTE]
> The generic stdlib `OtelLogHandler` was removed because this bridge was its only caller. It is trivially re-addable if a real stdlib-logging context ever needs it.

- A new integration test drives a log through the real `configure_logger` chain and asserts the record reaches Logs under `service.name = replay-vision` with only allowlisted attributes. This is the coverage the original bridge lacked, and it fails if the mirror is not wired into the worker's chain (the exact bug here).
- Unit tests for the processor cover: the allowlist dropping a payload-derived field (`response_preview`), logger-prefix scoping (a framework logger is not mirrored), severity mapping, exception-type-only, fail-soft on emit error, the unconfigured no-op, and the egress-proxy bypass on the real exporter.
- I (Claude) ran `pytest posthog/test/test_otel_logs.py products/replay_vision/backend/tests/test_logs_bridge.py` (16 passed), `ruff`, and `uv run mypy --cache-fine-grained .` (clean for these files).
- I did not exercise this against a live worker. Delivery under `service.name = replay-vision` is verifiable post-deploy in the Logs product.

## Docs update

No user-facing docs. Added `skip-inkeep-docs`.

## 🤖 Agent context

Built with Claude Code (Opus 4.8). Skills invoked: `superpowers:brainstorming` (design and approval before code) and `writing-tests` (the value gate on each test).

Decisions along the way: the root cause came from querying the live Logs product, which showed the worker emitting the target loggers only via stdout scraping, pointing at the stdlib-vs-structlog gap. I considered wiring via `configure_logger`'s existing `extra_processors`, but those append after the terminal renderer so the mirror would receive rendered output, not the structured event dict, hence the dedicated insertion point before the renderer. I also considered keeping the stdlib `OtelLogHandler` as a general utility but removed it as dead code. The processor was made always fail-closed (allowlist required) rather than supporting a forward-everything mode, since the only caller uses an allowlist and default-safe is better.

Design spec: `docs/superpowers/specs/2026-07-17-replay-vision-logs-structlog-mirror-design.md`.


fasyy612 (approved):
nice 👌

PR #72032, opened 2026-07-17
perf(activity-logs): index org/team activity log list ordering

yasen-posthog:
## Problem

The advanced activity logs list endpoints have a slow p95 tail. In the 24h window ending 2026-07-17T12:00Z, p95 for `advanced_activity_logs` list requests reached **10.76s** (8.69s the prior 24h, up from a 2.27s prior scan and a 6.68s 28d baseline), while p50 held at **0.177s**. So this is a slow tail on specific request shapes, not a uniform regression.

Attributing the tail from Loki access logs (contour), the slow requests are almost entirely the **organization-scoped** endpoint:

| Shape | Duration |
| --- | --- |
| `GET /api/organizations/<org>/advanced_activity_logs/?limit=100` (large org) | 7.7s – 25.9s |
| `GET /api/organizations/<org>/advanced_activity_logs/?start_date=…&page=1&page_size=100` (large org) | 10.7s – 11.1s |
| Same org endpoint **with** a `users=<uuid>` filter | 43ms – 211ms |
| `GET /api/projects/<id>/advanced_activity_logs/?scope=FeatureFlag&page=1` | ~150ms (a few outliers 1–2.4s) |
| `GET /api/projects/<id>/advanced_activity_logs/?start_date=…&page=1` (large team) | 4.6s – 6.0s |

The pattern: the tail is the **ordered list with no narrowing filter** (no scope, no user). Adding a `users=` filter collapses the same org request to sub-200ms.

### Root cause

Every activity row carries `organization_id` (the signal handlers set `organization_id = team.organization_id` alongside `team_id`), so the org-wide endpoint's baseline query is:

```sql
WHERE organization_id = <uuid> [AND created_at >= <lookback>]
ORDER BY created_at DESC
LIMIT 100
```

The `ActivityLog` model had **no** index that serves this. Its two `organization_id` indexes are both **partial** on `detail IS NOT NULL AND jsonb_typeof = 'object'` — a predicate the list query never carries — so Postgres can't use them, and the remaining org index is `(organization_id, scope, -created_at)`, which can't provide `created_at` order without a `scope` constraint. The planner falls back to scanning a large org's rows and sorting them, producing the multi-second tail.

The project-scoped endpoint has the same gap when no `scope` is supplied: the team indexes are all scope-led or partial on `was_impersonated=False AND is_system=False`, so a plain `WHERE team_id = X ORDER BY created_at DESC LIMIT n` on a large team also has no clean supporting index (the 4.6–6.0s project outliers above).

Neither suspected merge (#70340 MCP field selection, #70652 surfacing billing activity) touches the query path — this is a pre-existing missing-index that surfaced as large orgs/teams accumulated activity.

## Changes

Add two plain (non-partial) B-tree indexes to `posthog_activitylog` that directly serve the default ordered list scans:

- `idx_alog_org_created_at` on `(organization_id, created_at DESC)` — organization-scoped list endpoint.
- `idx_alog_team_created_at` on `(team_id, created_at DESC)` — project-scoped list endpoint when no `scope` filter is present.

Both are built with `SafeAddIndexConcurrently` (`atomic = False`) so the build takes only a `SHARE UPDATE EXCLUSIVE` lock and does not block reads/writes on this large, write-heavy table.

```mermaid
flowchart LR
  Q["WHERE organization_id = X<br/>ORDER BY created_at DESC LIMIT 100"] --> B["no usable index:<br/>partial org indexes skipped,<br/>scope-led index can't order"]
  B --> S["scan org rows + sort<br/>7–26s tail"]
  class S phRed
```

```mermaid
flowchart LR
  Q["WHERE organization_id = X<br/>ORDER BY created_at DESC LIMIT 100"] --> I["idx_alog_org_created_at<br/>(organization_id, created_at DESC)"]
  I --> R["index range scan,<br/>walk created_at, stop at 100"]
  class R phBlue
```

- `./manage.py sqlmigrate posthog 1258` emits the expected `CREATE INDEX CONCURRENTLY … ("organization_id", "created_at" DESC)` and `… ("team_id", "created_at" DESC)`, with `lock_timeout`/`statement_timeout` disabled and no transaction wrapper.
- `./manage.py makemigrations posthog --dry-run` reports no model/state drift.
- Applied the migration against a local Postgres: it applied cleanly (exit 0) and both `idx_alog_org_created_at` / `idx_alog_team_created_at` are present in `pg_indexes`.

No new automated test: the change is a pure additive index whose effect is only observable as query latency, not through the public API (results are byte-identical with or without it). A functional test can't catch its regression, a latency assertion would be flaky, and an assertion over `Meta.indexes` would be a change-detector. Migration correctness is covered by the existing migration CI jobs plus the `sqlmigrate`/apply checks above.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigation started from a production latency signal on `advanced_activity_logs` list p95. Confirmed the tail live in Grafana Loki (contour access logs) and attributed it to the organization-scoped endpoint's unfiltered ordered query by splitting p95 across `/api/organizations/` vs `/api/projects/` paths and inspecting the slowest individual request shapes. Read the viewset, filter manager, `ActivityLog` model indexes, and the activity-log signal handlers to establish that every row carries `organization_id` and that no index serves the `organization_id`/`team_id` + `-created_at` ordered scan (the existing org indexes are partial on a `detail` predicate the list query never carries).

Rejected alternatives: the two suspected PRs (#70340, #70652) were read and ruled out — neither touches the query path. Considered narrowing the query in application code, but the correct, minimal fix is the missing index. Chose plain (non-partial) indexes because the list query carries none of the partial predicates on the existing indexes.

Skills invoked: `/django-migrations` (used `SafeAddIndexConcurrently`, `atomic = False`), `/writing-tests` (concluded no test warranted, justified above).


yasen-posthog:
Heads-up on the extra file in this diff: `posthog/test/setup_receivers_baseline.txt` picked up three entries.

After rebasing onto master, `test_setup_receivers_match_baseline` was failing because three first-party signal receivers now connect at `django.setup()` but weren't recorded in the baseline:

- `post_delete`/`post_save` `products.ai_observability.backend.models.llm_prompt.invalidate_llm_prompt_label_cache`
- `post_save` `products.early_access_features.backend.signals.create_waitlist_survey_on_concept_stage`

These came in via unrelated feature work (prompt version labels and gated waitlist surveys) and only trip the baseline once combined on master — master's own baseline is currently missing them. Recording them here (per the test's documented regeneration path) unblocks this PR's required checks. Happy to drop the change if a dedicated baseline fix lands on master first.

yasen-posthog:
Rebased onto master to resolve the migration-number collision: master landed `1259_globalratelimitthresholdconfig`, so this PR's index migration is renumbered to `1260_activitylog_idx_alog_org_created_at_and_more` and now depends on it. No schema change — same two concurrent indexes.

The `setup_receivers_baseline.txt` addition from my earlier note is gone: during the rebase git dropped that commit because its contents are already upstream on master. So the diff is back to just the migration and the model `Meta.indexes` change.

yasen-posthog:
CI note: one Django Core shard (16/19) hit a transient failure in \`posthog/models/async_deletion/test_delete_person.py::test_delete_person\` (\`assert 2 == 1\`). This is an unrelated ClickHouse flake — it counts persons after a synchronous mutation and has no dependency on this PR (which only adds two Postgres btree indexes to \`posthog_activitylog\`); that shard also logged a "Schema restore failed" environment warning. CI auto-retried the shard and it passed. All required checks are now green.

posthog/migrations/1260_activitylog_idx_alog_org_created_at_and_more.py
```diff
@@ -0,0 +1,34 @@
...
+                name="idx_alog_org_created_at",
+            ),
+        ),
+        SafeAddIndexConcurrently(
+            model_name="activitylog",
+            index=models.Index(
+                fields=["team_id", "-created_at"],
+                name="idx_alog_team_created_at",
+            ),
+        ),
```
ReeceJones, line 33: Non-blocking: adding these indexes will increase write latency for activity logs. My (untested) assumption is that activity logs is a high-write, low-read data model. Given that, it may be better to find a way to optimize the queries without adding new indexes, or find a way to combine these indexes into a single index to minimize the write costs. Up to you, trust your judgement.

ReeceJones (approved):

posthog/migrations/1260_activitylog_idx_alog_org_created_at_and_more.py
yasen-posthog: Thanks — this is a fair concern, so I dug into both alternatives.

**Combining into one index isn't possible here.** The two indexes serve two different endpoints with different leading equality columns: the org-scoped list filters `WHERE organization_id = X ORDER BY created_at DESC`, the project-scoped list filters `WHERE team_id = X ORDER BY created_at DESC`. A B-tree can only range-scan `created_at` for order once its equality-filtered column is the prefix, so a single `(organization_id, team_id, -created_at)` (or the reverse) can't serve both — one endpoint would still fall back to a scan+sort. They're genuinely two distinct access patterns.

**Optimizing the query instead doesn't apply.** The slow shape is a plain ordered `LIMIT` with no narrowing predicate (no scope, no user) — that's the endpoint's default list contract, so there's nothing to push down or rewrite. The `org/team + -created_at` sort is inherent; only an index that matches it removes the scan+sort.

**On the write cost** — you're right that it's non-zero, but I think it's a good trade here. Both are narrow, non-partial 2-column B-trees. `posthog_activitylog` already maintains two GIN indexes on the `detail` jsonb column, and GIN maintenance dominates per-insert cost far more than a 2-column B-tree; two more small B-trees are a modest marginal add against that, and they remove a user-facing multi-second (7–26s) p95 tail on the org endpoint.

I did consider making them partial (mirroring the existing `was_impersonated=False AND is_system=False` indexes) to shrink them, but the list path applies visibility rules via `.exclude(...)` (a `NOT` predicate), not a positive equality, so a partial predicate wouldn't reliably match the query and the planner would likely skip it. Keeping them plain is what actually gets used.

Keeping the two indexes as-is.

PR #72042, opened 2026-07-17
fix(flags): key the EU flags mirror with a sentinel cache key

haacked:
## Problem

The cross-region flag sync (#70944) fetches PostHog's own (US team 2) flag definitions and mirrors them into EU's `flag_definitions_hypercache` so EU pods can dogfood local evaluation without polling the US API. It wrote that mirror under the plain per-team cache key `2`. EU allocates team ids independently of US, so key `2` on EU belongs to a real, unrelated team, and the sync overwrote that team's flag-definitions cache entry. The sync token was pulled as a stopgap, and EU deploys currently pin `POSTHOG_SELF_TEAM_ID` to route reads away from the collision.

## Changes

The mirror now lives under a dedicated string sentinel, `EU_CROSS_REGION_MIRROR_CACHE_KEY = "cross-region-flags-mirror"`. A real `Team.id` is always an integer, so the sentinel's rendered key (`cache/teams/cross-region-flags-mirror/...`) can never collide with a real team's key in any region, regardless of how team ids get allocated.

- `sdk_cache_provider.py` defines the sentinel and loosens `HyperCacheFlagProvider`'s key type to `int | str`. The constant lives here rather than in `local_evaluation.py` because `_build_flag_provider` imports it during `AppConfig.ready()`, and importing `local_evaluation` at that point triggers the circular import chain that `_get_hypercache` already documents and defers.
- `cross_region_flag_sync.py` reads the etag and writes the payload under the sentinel. `DOGFOOD_SELF_TEAM_ID = 2` stays as the team requested from the US API, which is still correct.
- `local_evaluation.py`'s `load_fn` short-circuits the sentinel to `HyperCacheStoreMissing()`. Nothing in the DB backs that key, so a cold cache waits for the next sync tick instead of attempting (and failing) a `Team` lookup by api_token.
- `_build_flag_provider()` in `posthog/utils.py` gains an EU branch that resolves reads to the sentinel, which makes the `POSTHOG_SELF_TEAM_ID` deploy override unnecessary once this ships.

Rollout after merge, in order: re-provision the sync token in the EU worker deploy, confirm the sentinel key is populating, then remove the `POSTHOG_SELF_TEAM_ID` override from the EU deploy configs.

Automated tests, all passing locally:

- `test_cross_region_flag_sync.py` (10 tests): sync writes land under the sentinel's rendered cache key; a real team's existing cache entry is untouched by a sync write (the regression that motivated this fix); cold-cache read of the sentinel returns `None` without querying `Team`; etag conditional GET flows through the sentinel key.
- `test_utils.py::TestBuildFlagProvider` (8 tests): region routing parameterized over `None`/`US`/`EU`; the EU provider reads back a payload written under the sentinel via the public `get_flag_definitions()` path; the env override still wins over the EU branch.

Also verified repo-wide `mypy --cache-fine-grained .` is clean for the touched files, and smoke-tested that importing `sdk_cache_provider` does not pull in `local_evaluation` (guarding the startup circular-import constraint).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with PostHog Code from an approved plan. The plan was independently re-verified against the current code before implementation; the one amendment was moving the sentinel constant out of `local_evaluation.py` into `sdk_cache_provider.py` to avoid re-introducing the circular import at `AppConfig.ready()`. Skills invoked: `/writing-tests` (test gating), `/simplify` (merged three near-identical region-routing tests into one parameterized test). Tests were written from the spec by a separate agent before the implementation landed and verified red without it.

---

### DIFF
--- a/products/feature_flags/backend/cross_region_flag_sync.py
+++ b/products/feature_flags/backend/cross_region_flag_sync.py
@@ -1,19 +1,21 @@
 """
-Cross-region sync of the dogfood team's (team 2) flag definitions into EU's
+Cross-region sync of PostHog's own (team 2) flag definitions into EU's
 HyperCache.
 
 Every other team's `flag_definitions_hypercache` entry is rebuilt locally by
 Django signals reading straight from Postgres (see local_evaluation.py). Team 2
--- PostHog's internal "self" team, used for our own dogfooded local evaluation
+-- PostHog's internal "self" team, used for our own local evaluation
 -- lives only in the US Postgres, so EU has no rows to build it from. Rather
 than have every EU pod poll the US API directly (which HyperCache exists to
 avoid), a single periodic task fetches team 2's flag definitions from the US
 region's authenticated `/flags/definitions` endpoint and writes the result
-straight into EU's HyperCache, so every EU pod still reads from the shared
-cache with zero per-pod polling.
+into EU's HyperCache under EU_CROSS_REGION_MIRROR_CACHE_KEY — a string
+sentinel, because EU allocates team ids independently of US and a numeric key
+would collide with a real EU team's entry. Every EU pod still reads from the
+shared cache with zero per-pod polling.
 
 No-op on US, where team 2's HyperCache entry is already populated by the
-normal signal-driven path.
+normal signal-driven path under its real team id.
 """
 
 from django.conf import settings
@@ -24,13 +26,10 @@ from posthoganalytics.request import US_INGESTION_ENDPOINT
 
 from posthog.utils import capture_exception_throttled, get_instance_region
 
+from products.feature_flags.backend.cache_keys import EU_CROSS_REGION_MIRROR_CACHE_KEY
 from products.feature_flags.backend.local_evaluation import flag_definitions_hypercache
 
 logger = structlog.get_logger(__name__)
-
-# PostHog's own internal "self" team, canonically team 2 (see _build_flag_provider
-# in posthog/utils.py). Only its US Postgres rows are authoritative.
-DOGFOOD_SELF_TEAM_ID = 2
 
 _FLAGS_DEFINITIONS_PATH = "/flags/definitions"
 _REQUEST_TIMEOUT_SECONDS = (3, 10)  # (connect, read)
@@ -39,12 +38,12 @@ _REQUEST_TIMEOUT_SECONDS = (3, 10)  # (connect, read)
 # unexpected request error) to PostHog error tracking, shared across the 30s schedule
 # so a sustained outage reports once per window instead of once per tick. Expected
 # transient network blips (connection/timeout/proxy errors) are logged, not captured.
-_SYNC_FAILURE_CAPTURE_THROTTLE_KEY = "cross_region_dogfood_flags_sync_capture_throttle"
+_SYNC_FAILURE_CAPTURE_THROTTLE_KEY = "cross_region_flags_sync_capture_throttle"
 _SYNC_FAILURE_CAPTURE_THROTTLE_TTL = 300  # seconds
 
 
-def sync_cross_region_dogfood_flags() -> None:
-    """Refresh team 2's flag-definitions HyperCache entry in EU from the US region.
+def sync_cross_region_flags() -> None:
+    """Refresh EU's mirror of US team 2's flag definitions from the US region.
 
     No-op outside EU or when the PSAK isn't configured. Uses the endpoint's ETag
     support: sends `If-None-Match` with the locally cached ETag, so an unchanged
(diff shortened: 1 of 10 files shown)


products/feature_flags/backend/test/test_cross_region_flag_sync.py
```diff
@@ -65,16 +71,23 @@ def test_sends_stored_etag_and_skips_write_on_304(self):
         # 304 means unchanged -- a write here would defeat the point of sending the etag.
         mock_set.assert_not_called()
 
-    def test_writes_fetched_payload_on_200(self):
-        payload = {"flags": [{"key": "new-flag"}], "group_type_mapping": {}, "cohorts": {}}
+    def test_writes_fetched_payload_under_sentinel_key_without_clobbering_real_team(self):
+        # The collision this change fixes: before the sentinel key, the sync wrote
+        # under the plain team-id key 2, which is EU's own real, unrelated team.
+        # A real team's cache entry must be untouched by this sync.
+        real_team = Team.objects.create(organization=self.organization, id=2, name="Real EU team 2")
```
matheus-vb: `Team.objects.create(id=2)` also forces its Project to id=2, which collides with BaseTest’s own self.project when the id sequence is low - `IntegrityError: duplicate key … posthog_project_pkey`. I reproduced it failing when this file runs early in a batch and passing in isolation, so it flakes by order (breaks hogli test <file> / early shards). The collision under test is at the cache-key level, not the DB, so you can drop the row and write the int key directly:
```python
flag_definitions_hypercache.set_cache_value(2, real_team_payload)
...
assert flag_definitions_hypercache.get_from_cache(2) == real_team_payload
```

This removes the `Team` import and can’t flake.

products/feature_flags/backend/local_evaluation.py
```diff
@@ -377,10 +384,20 @@ def _apply_flag_dependency_transformation(
     else "replica"
 )
 
+
+def _load_flag_definitions_with_cohorts(key: KeyType) -> dict[str, Any] | HyperCacheStoreMissing:
+    if key == EU_CROSS_REGION_MIRROR_CACHE_KEY:
+        # No DB row backs this key — it's a pure cache mirror written only by
+        # cross_region_flag_sync's periodic task. A cold cache should wait for the
+        # next sync tick, not attempt (and fail) a Team lookup.
+        return HyperCacheStoreMissing()
```
matheus-vb: Returning `HyperCacheStoreMissing()` here makes a cold read cache `__missing__` (1-day miss TTL) and delete the etag, which can shadow S3 or clobber a concurrent 30s sync write, EU’s internal dogfood flags then read empty until the next tick. You can consider raising `HyperCacheDependencyUnavailable` instead: that path returns a miss without caching a sentinel or touching the etag, which fits “not synced yet, retry next tick” and removes the race. Internal-only and self-healing, so not really a blocker.

matheus-vb (approved):
Nice! Two non-blocking suggestions.

products/feature_flags/backend/test/test_cross_region_flag_sync.py
haacked: Fixed!

products/feature_flags/backend/local_evaluation.py
haacked: Fixed!

products/feature_flags/backend/cross_region_flag_sync.py
```diff
@@ -96,21 +95,21 @@ def sync_cross_region_dogfood_flags() -> None:
...
-    # changed, and a bare int key isn't tracked in the expiry sorted set (unlike a
-    # Team key), so this write is what re-stamps the Redis TTL. On a long run of 304s
-    # the entry can still expire; that self-heals within one tick, because the etag
-    # expires with it, so the next poll sends no If-None-Match and gets a full 200.
-    flag_definitions_hypercache.update_cache(DOGFOOD_SELF_TEAM_ID, data=payload)
+    # changed, and a non-Team key isn't tracked in the expiry sorted set, so this
+    # write is what re-stamps the Redis TTL. On a long run of 304s the entry can
+    # still expire; that self-heals within one tick, because the etag expires with
+    # it, so the next poll sends no If-None-Match and gets a full 200.
+    flag_definitions_hypercache.update_cache(EU_CROSS_REGION_MIRROR_CACHE_KEY, data=payload)
```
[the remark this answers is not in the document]
haacked: Good catch, and it's a real gap. The old plain-team-id key does collide with a live EU team already sitting on id 2, so its cache entry got overwritten with mirrored US data before this fix.

I already tracked this down when the collision first went live: checked that team's local-eval usage telemetry and it showed no measurable traffic, so the corrupted entry almost certainly never got served to anyone in practice.

Since it's one specific known key rather than a general pattern, I'll clear and rebuild that team's cache entry manually through the staff cache tools after this deploys, instead of adding migration code for a one-off cleanup.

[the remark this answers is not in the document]
haacked: Good catch, and it's a real gap. The old plain-team-id key does collide with a live EU team already sitting on id 2, so its cache entry got overwritten with mirrored US data before this fix.

I already tracked this down when the collision first went live: checked that team's local-eval usage telemetry and it showed no measurable traffic, so the corrupted entry almost certainly never got served to anyone in practice.

Since it's one specific known key rather than a general pattern, I'll clear and rebuild that team's cache entry manually through the staff cache tools after this deploys, instead of adding migration code for a one-off cleanup.

PR #72045, opened 2026-07-17
fix(persons): stop pagination controls staying active with no rows

gustavohstrassburger:
## Problem

The person profile's "Feature flags" tab table reuses `featureFlagsLogic`'s server-controlled pagination. That reuse is intentional — the flags list is server-paginated at 100 per page, and this table pages through it via the `page` URL param (`urlToAction` forwards it back to `featureFlagsLogic`). But the pagination's entry count reflects the full flags list, so when the `evaluation_reasons` request fails and the table renders its error/empty state, the next/previous arrows stay active over "No entries."

I ran into this while testing #67776's new error state locally.

## Changes

`relatedFeatureFlagsLogic` now derives its own `pagination` value: it passes through the server-controlled pagination normally, and returns `undefined` while the `evaluation_reasons` load has failed, so an errored (empty) table shows no pagination controls at all. Once a retry succeeds, server-side paging comes back untouched.

Extended the existing load-error case in `relatedFeatureFlagsLogic.test.ts`: pagination is suppressed while `loadError` is set, and the server-controlled pagination (with the full flags count) returns after a successful retry — the latter guards against replacing server paging with client-side paging over a single server page, which would make flags beyond the first 100 unreachable. Full suite passes locally. Also verified against #67776's error state in the browser: arrows previously stayed active over "No entries"; the errored table now shows no pagination.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Found while manually testing #67776's new error state. A first agent-assisted fix swapped in client-side pagination; on review the human flagged that the solution didn't make sense, and Claude Code re-diagnosed the pagination wiring and replaced it with the error-state suppression above.


gustavohstrassburger:
@greptile review

matheus-vb (approved):
🚀

PR #72046, opened 2026-07-17
feat(review_hog): scope effectiveness stats with the page-level switch

sortafreel:
## Problem

The Code review page's "For you / Entire project" switch only rescoped the recent-reviews list. The hero proof card and the per-skill effectiveness cards stayed personal, so flipping to Entire project showed one scope's list over another scope's numbers.

**Why:** switching scopes was confusing during ReviewHog dogfooding – the stats silently stayed personal, and there was no way to see project-wide effectiveness at all.

## Changes

- `GET review_hog/reviews/perspective_stats/` accepts `scope=mine|everyone` (default `mine`), reusing the list endpoint's existing `_reports` scope plumbing. Bad values 400.
- The switch moves from the Recent reviews section header to the top of the page (hero overline row) as one page-level control. It now rescopes the hero proof card and all three effectiveness cards (perspectives, blind spots, validation) together with the list. On a scope flip the stale stats drop synchronously so cards show skeletons instead of the other scope's numbers.
- Skill lists and their enable toggles deliberately stay per-user – only the stats rescope. On project scope the effectiveness rows can therefore list skills beyond your own cards (e.g. a teammate's custom perspective), and hero totals equal the sum of rows.
- Copy adapts per scope ("findings worth the team's time", "This project's last N reviews", and the validator card reads "Validation" instead of "Your quality bar", since the project aggregate spans every author's validator).
- Regenerated API types for the new query param; DECISIONS.md records the design choices.

**For you (default):**

scope-for-you

**Entire project – hero proof card, reviews list, and all effectiveness cards rescope together:**

scope-entire-project

- Jest: all 9 `reviewHogSettingsLogic` tests pass, including a new one proving a scope flip re-requests the stats with the new `scope` param and nulls the stale value first – the regression it catches is the switch moving the list but leaving the stats on the old scope, which is exactly the bug this PR fixes.
- Backend: extended the existing `perspective_stats` test with everyone-scope aggregation (teammate's report folds in) and a bad-scope 400 as the params-serializer wiring guard. All 11 tests in `test_reviews_api.py` pass locally against a real database.
- ruff check/format clean, kea typegen up to date, oxfmt no-op, `tsgo --noEmit` clean for all changed files (the only errors are pre-existing on master in `products/tracing`, untouched here). A local `hogli build:openapi` run after checkout produced no diff, confirming the generated files match the generator.
- Verified manually in local dev (screenshots above): the initial load requests `reviews/?scope=mine` and `perspective_stats/?scope=mine`, flipping the switch fires both with `scope=everyone` and re-phrases all four stat surfaces, and the scope persists in the URL (`?reviews_scope=everyone`). As a user with no reviews, `scope=mine` returns empty on both endpoints while `scope=everyone` still returns the project's reviews, so the empty For-you state renders with its copy pointing at the top-of-page switch.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

ReviewHog's design record (`products/review_hog/DECISIONS.md`) is updated in this PR. No pages under `docs/` cover this alpha surface yet.

## 🤖 Agent context

- Authored with Claude Code as a PostHog Code cloud task, directed by the assignee. Requirements were pinned down via a four-question interview before coding: scope means this project (not org-wide), the switch lives at the top of the page, project-scope effectiveness rows show every skill that raised findings, and all four stat surfaces flip together. Applying the scope to skill lists was explicitly deferred to keep this iteration small.
- Skills invoked: `/improving-drf-endpoints`, `/writing-tests`, `/adopting-generated-api-types`.
- The cloud environment had no database, so the backend tests and `hogli build:openapi` first ran during local follow-up: tests pass, and the generator run was a no-op over the hand-synced files.

---


sortafreel:
The failing **Unit Tests** / **MCP Tests Pass** checks are master breakage, not this PR: `tests/unit/execute-sql-description.test.ts` fails with the same assertion (`expected 1444 to be less than 1200`) on master's own MCP CI ([run 29603784864](https://github.com/PostHog/posthog/actions/runs/29603784864), commit `caef05f8`). The character-budget overflow comes from e1250a44 (#71921) growing the flag-gated data-catalog section; this branch doesn't touch `instructions.ts`, and the full MCP suite passes locally at this branch's HEAD (which predates e1250a44). It only appears here because PR CI runs on the merge ref with current master. Re-run or update the branch once master is green.

sortafreel:
Second unrelated red check, same pattern as the MCP one above: **Django tests – Core** fails on `test_setup_receivers_match_baseline` over receivers from `products/ai_observability` and `products/early_access_features` — products this PR doesn't touch. This branch's `setup_receivers_baseline.txt` is byte-identical to master's, the branch's diff vs merge-base is exactly the 9 review_hog/mcp files, and master's own Backend CI is red on its last three completed runs. Both failures should clear on re-run once master is green.

cvolzer3 (approved):

PR #72064, opened 2026-07-17
feat(surveys): route survey summarization through the ai-gateway

brandonleung:
Moves survey response summarization off the direct Gemini call and onto a cheap Anthropic model (`claude-haiku-4-5`) routed through the internal ai-gateway, matching the summarization / eval-report / archetype gateway migrations. Draft: the model and billing choices below need team-surveys + billing sign-off.

## Changes
- New `llm/anthropic.py` provider: builds the client via `build_openai_client("survey_summary")`, calls `claude-haiku-4-5` with `response_format=json_object`, validates into the existing `SurveySummaryResponse`.
- Removes `llm/gemini.py`; `call.py` / `__init__.py` now wire the Anthropic provider. Public `summarize_responses(...)` signature is unchanged.
- `models.py` / `constants.py`: `GeminiModel` becomes `AnthropicModel`, default `claude-haiku-4-5`.
- Registers `survey_summary` in the gateway `Product` literal and the Python-gateway product config.
- Rewrites the summarization tests for the gateway path.

## How attribution is preserved
The gateway captures `$ai_generation` itself, so the client isn't wrapped with `posthoganalytics.ai`. Per-call dims (`team_id`, `survey_id`, `question_id`, `response_count`, `ai_product`/`ai_feature`) ride the `X-PostHog-Properties` blob; the feedback-mapping trace id rides `X-PostHog-Trace-Id`; `distinct_id` rides `user=`. `team_id` is what keys per-team spend in the usage report. One change from the Gemini path: the event's owning project (`$groups`) is now derived from the gateway `phs_` bearer rather than set client-side, same as every other gateway migration. Translation is untouched and stays on Gemini.

## Rollout
Activates on merge: the summarize endpoint runs in `posthog-web-django`, which already has `AI_GATEWAY_URL` set. No charts change needed.

## Open decisions (need sign-off)
- **Model:** `claude-haiku-4-5` as the cheap equivalent. Confirm summary quality is acceptable versus Gemini.
- **Billing:** the old path set `$ai_billable=True`; I mapped the product to `credit_bucket=AI_CREDITS` to preserve that. Billing now moves from the event property to gateway metering. Billing team to confirm the bucket.
- **Event tag:** `ai_product` moves from `surveys` to `survey_summary` (survey translation still emits `surveys`). Confirm no cost dashboard keys on the old value, or keep `surveys` + `ai_feature`.

## Verification
- `test_summarization.py`: 16 passed. Pre-commit (ruff, ty) pass.
- Not yet exercised against a live gateway; validate haiku's structured-output fidelity versus the Gemini JSON-schema path before un-drafting.

## Related
- PostHog/posthog#72061 is the sibling migration (customer archetype classification).


adboio (approved):
approving since things are broken but the robot comments do look like they are worth a look

PR #72088, opened 2026-07-17
feat(ai): surface event definition descriptions to Max taxonomy

sufi-ph:
## Problem

PostHog AI's `read_taxonomy` tool returned custom events as bare names with no description, even when a user had written a description on the event definition. The taxonomy path (`_process_events_data` in `ee/hogai/utils/helpers.py`) only attached descriptions from the built-in core taxonomy or from events explicitly added to the conversation context — it never read the user-authored `description` field off event definitions. So Max couldn't use those descriptions for context.

**Why:** A user pointed out that when they ask PostHog AI about a custom event, it can't see the description they set on the event definition — a gap worth closing so the descriptions people maintain actually inform the agent.

## Changes

Added a batched, EE-gated fallback in `_process_events_data`: for custom events not covered by the core taxonomy or the conversation context, it looks up `EnterpriseEventDefinition.description` in a single query and uses it as the description. Precedence is unchanged where a description already exists (core taxonomy first, then context), with the stored definition as the final fallback. No-op on non-EE builds, where the model has no `description` field.

Added two cases to `test_format_events_prompt.py`:
- a custom event with only a stored event-definition description now surfaces that description (guards the new fallback being dropped);
- a context-supplied description still wins over the stored one (guards the precedence order).

The agent could not run the suite in its environment (no Postgres reachable — every test in the file errors at DB setup). Ran `ruff check`/`ruff format` (clean) and `py_compile` (passes) on both files.

## 🤖 Agent context

Investigated the `read_taxonomy` path with the `Explore` agent to confirm where descriptions come from (`CORE_FILTER_DEFINITIONS_BY_GROUP` and `MaxEventContext`) and that the user-editable `description` field lives only on the EE `EnterpriseEventDefinition` model, gated by `EE_AVAILABLE`. Chose a single batched query over per-event lookups to avoid N+1, and kept the change behind the existing `elif` chain so it only affects events that currently get no description. Invoked the `/writing-tests` skill before adding tests.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BALLBBKAB/p1784312695386939)*


Twixes (approved):
LGTM

PR #72099, opened 2026-07-17
feat(bi): add scatter plot chart type for sql insights

andyzzhao:
## Problem

SQL insights can only aggregate. Every existing chart type (line, bar, pie, 2d heatmap) buckets rows before drawing, so there is no way to look at a per-row distribution over time, like per-run query latency, and then ask "which run is that outlier, and whose was it". Investigating that today means exporting results and hand-building a one-off chart outside PostHog.

## Changes

Adds a `Scatter plot` chart type to the SQL visualization stack. Every result row becomes one dot.

- x-axis takes a datetime column (rendered on a time axis) or a numeric column, y-axis takes a numeric column, both auto-filled from the result shape
- optional color-by column groups dots into colored series with a legend, capped at 10 series total: the largest values keep their own series and the tail folds into `Other`. SQL `NULL` groups as `(null)`, distinct from a literal `"null"` string
- logarithmic y-axis toggle for latency-style tails
- dots are translucent with no border, so overlapping points compound into visibly darker spots and dense regions read as density; the hovered dot goes opaque
- hovering a dot (with a forgiving hit radius) shows the row in a tooltip, clicking opens a row-detail modal with the full row
- a person column (auto-detected when named `distinct_id`) renders in the modal as a link to that person's profile, so a dot resolves to an actual user

Auto-filled columns after picking the chart type in the SQL editor (demo data: two normally distributed latency clusters plus a slow tail):

2-scatter

Color-by plus logarithmic y-axis; overlapping points inside the cluster cores render darker:

4-colorby-log

Hovering a dot shows the row without clicking (the hovered dot goes opaque):

5-hover-tooltip

Clicking a dot, with the person column linked:

3-row-modal

Wiring follows the `TwoDimensionalHeatmap` precedent end to end: `ChartDisplayType.Scatter` plus a `ScatterSettings` block on `ChartSettings` (schema regenerated), a pure row-to-datasets adapter, a Chart.js component, a sidebar config tab, and render branches in both `DataVisualization.tsx` and the SQL editor's `OutputPane` fork. Rows with null, non-numeric, or non-finite values are skipped with a visible count, points cap at 10k with a truncation note.

- `scatterChartAdapter.test.ts` covers the regressions the adapter can ship: null, non-numeric, and non-finite rows becoming broken dots (Infinity breaks Chart.js axis scaling), non-positive values breaking the log scale render, off-by-one in the top-10 color bucketing, the point cap not flagging truncation, and missing columns crashing instead of showing the empty state
- two new `dataVisualizationLogic` cases cover column auto-fill on picking scatter (including person-column detection) and `mergeChartSettings` wiping sibling scatter settings on partial update, a bug shape the heatmap settings already guard against for their block
- frontend typecheck, oxlint/oxfmt, ruff, repo-wide mypy, and `hogli ci:preflight` all pass (the one mypy error is pre-existing on master in an untouched file)
- I also drove the real flow with a scripted headless browser against a local dev stack: ran a synthetic latency query in the SQL editor, switched to scatter, toggled color-by and log scale, hovered a dot for the tooltip, clicked it, and confirmed the modal opened with a working person profile link. The screenshots above are from that run, with demo data drawn from Box-Muller normal distributions so the clusters and the translucency-on-overlap effect look like real latency data
- a multi-agent QA review (security, performance, frontend, compatibility, copy, plus two generalists) was run over the diff; all its actionable findings were fixed in the follow-up commit, notably: `person_id` removed from person-column auto-detection (it holds a UUID the distinct_id route cannot resolve), cleared column settings now use an empty-string marker so the insight save path's `exclude_none` cannot resurrect them, the row modal snapshots the clicked row so a background query re-run cannot silently swap its contents, and the hover tooltip is capped so `SELECT *` results cannot overflow the canvas

> [!NOTE]
> Deploy ordering: backend before frontend, as with prior chart-type additions. `ChartSettings` is `extra="forbid"` server-side, so during the rolling-deploy window a new frontend saving `chartSettings.scatter` against an old pod gets a 400, and a backend rollback strands saved scatter insights until re-deploy. Transient and precedented (`TwoDimensionalHeatmap` shipped identically), but worth knowing before a rollback decision.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No in-repo doc covers SQL insight chart types (the 2d heatmap has none either), so nothing to update here.

## 🤖 Agent context

Built with Claude Code in a single session. Skills invoked: superpowers brainstorming for the design phase, `/writing-tests` before adding tests, and the repo's qa-team multi-agent review skill after the feature was complete. Design decisions: rejected a points-only mode on the existing line graph path (fights its aggregated-series data model) and a bespoke canvas chart (duplicates axis/tooltip/legend machinery); quill-charts was considered as the preferred home for new charts but has no scatter chart yet and only renders behind a feature flag, so this uses the same Chart.js path every other SQL chart defaults to. Mid-implementation Claude found the SQL editor's `OutputPane` maintains its own render fork of the visualization branch chain, which needed its own scatter branch or the editor would render nothing. The translucent-dot rendering was matched against a reference visualization the author had built by hand (0.45 alpha fills, no stroke, opaque on hover), and the hover hit radius was widened after headless testing showed the default 3px target made tooltips nearly impossible to trigger. Known deferred item from the QA review: `Scatter` (like `TwoDimensionalHeatmap` before it) leaks into the assistant trends display enum without a trends renderer; curating that enum is left as a follow-up. Also deferred by design: brush pan/zoom, group click-through links, and scatter as a Trends display type.


thmsobrmlr (commented):
Hey @andyzzhao, this would definitely be super useful to have. More generally, we've been thinking about adding additional chart types and expanding our BI capabilities.

The first step was migrating to a new chart library insight quill, but this PR still appears to use the legacy `useChart` hook and custom tooltips. I'll leave it to @sampennington to provide more context when he's back from PTO

PR #72117, opened 2026-07-18
chore(metrics): add product guardrails doc and fix stale internal docs

DanielVisca:
## Problem

The metrics product is mid private alpha, with two ClickHouse layouts already live (`metrics1` plus the `metric_series`/`metric_samples` split) and a raw SQL tab exposing `posthog.metrics` to users. The decisions that keep this evolvable - what stays stable, what gates wider access, what we deliberately defer - lived in chat threads and heads, not in the repo. Meanwhile the existing internal docs had drifted from the code in two ways that could mislead someone building against them.

## Changes

**New: `docs/internal/metrics/product-guardrails-and-roadmap.md`** - records the load-bearing decisions with the code seams they were validated against:

- Two invariants: user-visible schema never equals physical schema (re-register `posthog.metrics` as a `LazyTable` so the storage port becomes a reversible expansion flip), and series identity is assigned once at ingest (no second fingerprint implementation anywhere).
- The pre-beta gate: a per-team budget on new `series_fingerprint` creation at the `rust/capture-logs` seam - bytes limiting can't see series identity, and label-stripping is called out as the failure mode to never ship.
- An operability principle (dynamic ingest config via ingestion's existing Redis/PG patterns, not feature flags), a review checklist for the future logs-to-metrics rules work, parked workstreams with explicit triggers (saved-metric registry, rollup tiers), and anti-patterns we commit to avoiding.

**Fixes to existing docs:**

- `deployment-layout.md` said the series fingerprint would be `cityHash64(...)`; the shipped implementation is SipHash-1-3 assigned at ingest (`rust/capture-logs/src/metric_record.rs`). Dangerous drift, since a ClickHouse-side recomputation built from the doc would diverge from stored identity.
- Both docs referenced `nodejs/src/metrics-ingestion/...`; the consumer lives at `nodejs/src/ingestion/pipelines/metrics/...`. Also noted that migrations 0283/0285 have landed the series/samples schema, which the status text predated.

Docs-only. Both corrections were verified against source before writing: `compute_series_fingerprint` in `rust/capture-logs/src/metric_record.rs` (SipHasher13) and the actual consumer path on disk. Every claim in the new doc cites the file it was validated against so reviewers can spot-check.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal docs only (`docs/internal/`), no public docs surface.

## 🤖 Agent context

Written by Claude (PostHog Code session) under Daniel's direction, distilled from a competitive-architecture review whose conclusions were validated against this repo's actual seams before being written down as our own decisions. Daniel scoped which recommendations became commitments, which became parked-with-triggers, and which were dropped (a dynamic-limiter workstream was folded into the existing INFRA-B item instead of becoming new work).

---


DanielVisca:
need to remove these docs anyway

PR #72119, opened 2026-07-18
fix(experiments): hide idle recalculation status and 0/0 rows flash

rodrigoi:
## Problem

The recalculation status bar renders in every state, but it only carries useful information while a run is actually executing. When loading cached results from `/latest` it shows a redundant "Loading metrics…" placeholder, and after a run completes it sits there permanently repeating what the metric rows already show.

The "rows read" counter also flashes "0 / 0 rows read". The backend's live progress read legitimately samples all-zeros (queries not started yet, or finished but not flushed to `system.query_log`), and the component rendered whatever the poll returned.

## Changes

Two behavior fixes to `RecalculationStatus`, no visual redesign.

### Hide the bar unless a run is executing

`RecalculationStatus` now renders only in the `cold` and `refreshing` display states and returns null for `initial`, `partial`, and `resting`. It returns null instead of being unmounted so `experimentResultsNotificationLogic` stays alive and the finish-edge browser notification still fires. The `mb-2` wrapper in `ExperimentView` moved onto `StatusBar` itself, so hidden states leave no stray margin.

- `frontend/src/scenes/experiments/MetricsView/shared/RecalculationStatus.tsx`
- `frontend/src/scenes/experiments/ExperimentView/ExperimentView.tsx`

### Never show 0/0 rows read

New `liveRowsProgress` reducer in `experimentMetricsLogic`: it keeps the last nonzero rows-read sample for the current run and ignores all-zeros polls, clearing only when a different run starts. The component renders the counter exclusively from this value, so it either shows nothing (no sample yet) or the latest real number.

- `frontend/src/scenes/experiments/experimentMetricsLogic.ts`
- `frontend/src/scenes/experiments/experimentMetricsLogic.test.ts`

```bash
pnpm --filter=@posthog/frontend jest frontend/src/scenes/experiments/experimentMetricsLogic.test.ts
```

```bash
pnpm --filter=@posthog/frontend typescript:check
```

cat-type-small

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Model: Fable 5
Manually refactored: **yes**

Skills used:

- /writing-kea-logics (local)
- /writing-tests (local)

Relevant decisions:

- Rows-counter retention lives in a `liveRowsProgress` reducer keyed by run id, not component state, so the all-zeros-gap behavior is testable at the logic level and the component stays render-only.
- The idle states return null from inside the component rather than gating the mount in `ExperimentView`, deliberately keeping the notification logic mounted.
- The retained sample is matched against `recalculation.id` before rendering, so a new run never briefly shows the previous run's row count.

### DIFF
--- a/frontend/src/scenes/experiments/ExperimentView/ExperimentView.tsx
+++ b/frontend/src/scenes/experiments/ExperimentView/ExperimentView.tsx
@@ -90,11 +90,7 @@ const MetricsTab = (): JSX.Element => {
                 <MultiVariantBiasWarning />
             </div>
 
-            {showRecalculationStatus && (
-                <div className="mb-2">
-                    <RecalculationStatus experiment={experiment} />
-                </div>
-            )}
+            {showRecalculationStatus && <RecalculationStatus experiment={experiment} />}
 
             {/* Modern metrics view */}
             {!hasMetrics ? (
--- a/frontend/src/scenes/experiments/MetricsView/shared/RecalculationStatus.tsx
+++ b/frontend/src/scenes/experiments/MetricsView/shared/RecalculationStatus.tsx
@@ -4,10 +4,8 @@ import { IconBell, IconCheck, IconInfo } from '@posthog/icons'
 import { LemonButton, Tooltip } from '@posthog/lemon-ui'
 import { useAnimatedNumber } from '@posthog/quill-charts'
 
-import { LemonDivider } from 'lib/lemon-ui/LemonDivider'
 import { Spinner } from 'lib/lemon-ui/Spinner'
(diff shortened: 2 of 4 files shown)


rodrigoi:
* **#72119** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/72119?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #72135, opened 2026-07-18
fix(quick-start): keep setup button visible when navigating to scenes without configured setup

### LINKED ISSUE #70204: Quick start button disappears from the top bar mid-flow - you have to go back to where it respawns to continue
### Bug Description

*Please describe.*

Working through the Quick start checklist as a first-time user. Completing or skipping a step populates fine — the checklist tracks progress correctly. The problem is the Quick start button doesn't persist in the top bar: click into a task, end up on another page, and the button is gone. To carry on with the checklist you have to navigate back to a page where it respawns.

If you save on the page you're configuring and the checklist doesn't resurface, and you're mid-flow, it breaks that flow. The checklist works as a tracker but not as a guide — the thread you're meant to be following keeps physically disappearing, and a new user mid-setup is exactly the person with the least spare attention for re-finding it. Seventeen steps is seventeen chances to lose the thread.

Suggested fix: keep the Quick start button pinned in the top bar on every screen while the checklist is incomplete — or resurface the panel when a step's task is saved so the flow carries you to the next one.

Screenshot attached.

Image

Image

## How to reproduce

1. As a new user, open the Quick start checklist from the top bar and click into any step (e.g. "Create your first insight")
2. Complete or save the task on the page it takes you to
3. Note the Quick start button is no longer in the top bar — navigate back toward the home/insights screen to find where it reappears

### Debug info

```shell
- [X] PostHog Cloud, Debug information: Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019f4fc6-a528-71f7-b12f-4382700a7bd0?t=0
Admin (internal): http://go/adminOrgEU/019f3d5d-3e60-0000-a7f3-58579a754bba (project ID 218385)
```

okxint:
## Problem

Closes #70204

When a new user works through the Quick Start checklist and navigates to a page whose
`productKey` isn't in `PRODUCTS_WITH_SETUP` (e.g. settings, billing, the SQL editor),
the Quick Start button disappears from the top bar. To continue the checklist they have
to navigate back to a page where the button reappears — breaking the flow at exactly the
moment they're least equipped to hunt for it.

## Changes

Removed the `sceneHasNoSetup` guard from `shouldShowButton` in `ProductSetupButton`.

The selector `sceneHasNoSetup` evaluates to `true` when the current scene has a `productKey`
that isn't registered in `PRODUCTS_WITH_SETUP`. Gating the button's visibility on this meant
the button disappeared mid-flow whenever a checklist task sent the user to a scene we don't
have setup tasks for. The button should stay pinned as long as there are remaining tasks,
regardless of which scene the user is currently on.

I wasn't able to manually test this against a running PostHog instance. The logic change is
a one-line removal of a condition; the selector `sceneHasNoSetup` remains defined in
`globalSetupLogic` and the behavior in `shouldShowButton` now reads:

```ts
const shouldShowButton = isCurrentOrganizationNew && (remainingCount > 0 || isGlobalModalOpen)
```

Regression being caught: button no longer disappears when navigating from a setup-task target
page (e.g. Insights) to a page whose product doesn't have setup configured.

## 🤖 Agent context

Authored with Claude Code (claude-sonnet-4-6). No skills were invoked for this change — it was a
targeted one-line fix to a visibility condition with no backend, API type, or test-writing surface.

Traced the bug from the issue report → `ProductSetupButton.tsx:35` → `globalSetupLogic.ts` selector
`sceneHasNoSetup`. The selector itself is correct; the error was using it to gate the button when
it should only gate product auto-selection in the popover. Removed the guard; `sceneHasNoSetup`
remains in the logic for any future use.

Closes #70204

### DIFF
--- a/frontend/src/lib/components/ProductSetup/ProductSetupButton.tsx
+++ b/frontend/src/lib/components/ProductSetup/ProductSetupButton.tsx
@@ -21,7 +21,7 @@ import { ProductSetupPopover } from './ProductSetupPopover'
  * handled by SceneContent when a productKey is provided.
  */
 export function ProductSetupButton(): JSX.Element | null {
-    const { selectedProduct, isGlobalModalOpen, sceneHasNoSetup } = useValues(globalSetupLogic)
+    const { selectedProduct, isGlobalModalOpen } = useValues(globalSetupLogic)
     const { openGlobalSetup, closeGlobalSetup, setSelectedProduct } = useActions(globalSetupLogic)
     const { isCurrentOrganizationNew } = useValues(organizationLogic)
     const showPulseIndicator = useFeatureFlag('QUICK_START_PULSE_INDICATOR', 'test')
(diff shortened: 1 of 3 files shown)


vmaineng (commented):
Logic looks good and good catch with the changes to remove the 'sceneHasNoSetUp' guard from shouldShowButton in ProductSetUpButton

PR #72138, opened 2026-07-18
fix(experiments): allow sorting experiments list by conclusion

### LINKED ISSUE #63621: Bug Report: Filtering the experiment list by `Result` or `Created By` causes the sort to break.
### Bug Description

- You are allowed to sort the experiments table by column.
- However, if you click on the `Result` column, you get an error.
- You get a slightly different error, but still a failure, when you sort by `Created By`.

## How to reproduce
- Navigate to any experiment's page
- Click on the `Result` tab in the table.
- An error pops up: `Load experiments failed: Invalid order field: 'conclusion'`
- Notice the URL changes to include `order=conclusion`

Image

- Click on the `Created By` field in the table.
- See the error: ``Load experiments failed: A server error occurred.``
- Notice the URL changes to include `order=created_by`.

Image

## Additional context
- From https://posthoghelp.zendesk.com/agent/tickets/60328 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61455)

### Debug info

```shell
Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019ecc1e-bff1-7633-999a-577bf474771c?t=1472
Admin (internal): http://go/adminOrgUS/019c4ed6-add0-0000-998a-30c999608349 (project ID 311650)
```

ayaangazali:
## Problem

Sorting the experiments list by the Result column breaks the page. The column's sort key is `conclusion`, so clicking it makes the frontend request `order=conclusion`, but the list endpoint's order allowlist never included that field. The API responds with `Invalid order field: 'conclusion'` and the list fails to load.

The issue also reports a 500 when sorting by Created By. That looks like the bug already fixed in #60532 (`created_by` ordering now goes through a display-name annotation), and I could not reproduce it on master, so this PR only fixes the conclusion half.

Closes #63621

## Changes

- Add `conclusion` / `-conclusion` to `EXPERIMENT_ORDER_ALLOWLIST` in `ExperimentService`
- Order by a `Case`/`When` rank that matches the frontend Result column's client-side sorter (won, lost, inconclusive, stopped early, invalid, then experiments without a conclusion), the same pattern the existing `status` ordering uses
- Update the `order` parameter description on the list endpoint (it was also missing `created_by`) and regenerate the OpenAPI-derived types, which only changes that description string in 3 generated files

- New parameterized test `test_filter_experiments_queryset_orders_by_conclusion` (ascending + descending). The fixture set (won, lost, invalid, no conclusion) is picked so plain alphabetical ordering would fail the assertion, and it checks that experiments without a conclusion sort last ascending. That's the regression no existing test catches: dropping the allowlist entry or the rank mapping breaks Result sorting again.
- Added `conclusion` / `-conclusion` to the existing parameterized `test_order_by_valid_fields_works`.
- Verified both new tests fail on master without the service change (rejected with `ValidationError`) and pass with it.
- Full `products/experiments/backend/test/test_experiment_service.py` passes locally (471 tests).
- `hogli ci:preflight --fix` green, repo-wide `mypy --cache-fine-grained .` clean.
- I did not click through the UI against a live stack; verification is automated at the service layer, which is where the failure originates.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs under `docs/` describe experiments list ordering; the API parameter description is updated in code and flows into the generated docs.

## 🤖 Agent context

Built with Claude Code. Skills invoked: `/writing-tests` and `/improving-drf-endpoints`. Main decision: sort by a rank that mirrors the frontend column's own comparator instead of ordering alphabetically on the raw column, mirroring how `status` ordering is implemented. I also considered removing the column's sorter on the frontend instead, but the column is clearly meant to be sortable. The OpenAPI regen was needed because the `order` parameter description is embedded in the generated types.

---

small disclaimer: i'm a freshman in college trying to contribute properly to projects i use. claude code did a lot of the heavy lifting here and i went through the logic and the tests myself, so if anything looks off i would honestly love to hear what i missed :)

### DIFF
--- a/products/experiments/backend/experiment_service.py
+++ b/products/experiments/backend/experiment_service.py
@@ -596,6 +596,8 @@ class ExperimentService:
         "-duration",
         "status",
         "-status",
+        "conclusion",
+        "-conclusion",
     }
 
     ELIGIBLE_FLAGS_ORDER_ALLOWLIST = {
@@ -3701,6 +3703,23 @@ class ExperimentService:
                     queryset = queryset.order_by(F("status_sort_key").desc())
                 else:
                     queryset = queryset.order_by(F("status_sort_key").asc())
+            elif order_value in ["conclusion", "-conclusion"]:
+                # Match the frontend Result column's sorter: won, lost, inconclusive,
+                # stopped early, invalid, then experiments without a conclusion.
+                queryset = queryset.annotate(
+                    conclusion_sort_key=Case(
+                        When(conclusion="won", then=Value(1)),
+                        When(conclusion="lost", then=Value(2)),
+                        When(conclusion="inconclusive", then=Value(3)),
+                        When(conclusion="stopped_early", then=Value(4)),
+                        When(conclusion="invalid", then=Value(5)),
+                        default=Value(6),
+                    )
+                )
+                if order_value.startswith("-"):
+                    queryset = queryset.order_by(F("conclusion_sort_key").desc())
+                else:
+                    queryset = queryset.order_by(F("conclusion_sort_key").asc())
             elif order_value in ["created_by", "-created_by"]:
                 # Match the frontend column's `first_name || email` sorter — treat an
                 # empty `first_name` as missing and fall back to `email`, so users with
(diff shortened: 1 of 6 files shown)


ayaangazali:
Closing this one, I am no longer working in this repo. Feel free to pick up the change if it is still useful.

PR #72140, opened 2026-07-18
fix(hogvm): treat replaceOne/replaceAll replacement literally

winklemad:
## Problem

`replaceOne` / `replaceAll` in the TypeScript HogVM (`common/hogvm/typescript/src/stl/stl.ts`) and the Hog→JS compiler (`posthog/hogql/compiler/javascript_stl.py`) pass the replacement to JS `String.replace` / `.replaceAll` as a plain string. JS then interprets `$$`, `$&`, `` $` ``, `$'` as **special replacement patterns**, so the result diverges from the Python VM, the Rust VM, and ClickHouse — all of which treat the replacement literally.

The same Hog program returns different output depending on the runtime:

- `replaceAll('price', 'price', '$$')` → `'$'` on TS (dollar silently halved), `'$$'` on Python / Rust / ClickHouse
- `replaceAll('a_b', '_', '$&')` → `'a_b'` on TS, `'a$&b'` correct

The TS VM runs in plugin-server (production ingestion) while the Python VM backs previews/tests, so a hog-function transformation can **preview correctly and run wrong** — a silent data-integrity divergence, reachable via ordinary currency/price strings.

## Changes

- `stl.ts`: `replaceOne` / `replaceAll` now use a replacer function (`() => replacement`), whose return value is never `$`-processed, so the replacement is always literal.
- `javascript_stl.py`: same fix in the two generated function bodies; regenerated the committed `stl.js` / `hogqlx.js` snapshots to match.
- Added a regression test through the real bytecode path covering `$$`, `$&`, `` $` ``, `$'` plus an ordinary-replacement guard. The full HogVM TS suite passes (97/97), and the output matches the Python VM byte-for-byte.

## 🤖 Agent context

Investigated with the help of an AI coding agent (Claude Code). I verified the cross-runtime divergence and the fix myself against the real HogVM code (Python vs TS vs Rust), ran the HogVM TypeScript jest suite (97/97), and confirmed red→green on the new regression test through the real bytecode path. I understand and own this change.

### DIFF
--- a/common/hogvm/__tests__/__snapshots__/hogqlx.js
+++ b/common/hogvm/__tests__/__snapshots__/hogqlx.js
@@ -1,4 +1,4 @@
-function replaceAll (str, searchValue, replaceValue) { return str.replaceAll(searchValue, replaceValue) }
+function replaceAll (str, searchValue, replaceValue) { return str.replaceAll(searchValue, () => replaceValue) }
 function print (...args) { console.log(...args.map(__printHogStringOutput)) }
 function jsonStringify (value, spacing) {
     function convert(x, marked) {
--- a/common/hogvm/__tests__/__snapshots__/stl.js
+++ b/common/hogvm/__tests__/__snapshots__/stl.js
@@ -75,8 +75,8 @@ function startsWith(str, prefix) {
 }
 function round(a) { return Math.round(a) }
 function reverse (value) { return value.split('').reverse().join('') }
-function replaceOne (str, searchValue, replaceValue) { return str.replace(searchValue, replaceValue) }
-function replaceAll (str, searchValue, replaceValue) { return str.replaceAll(searchValue, replaceValue) }
+function replaceOne (str, searchValue, replaceValue) { return str.replace(searchValue, () => replaceValue) }
+function replaceAll (str, searchValue, replaceValue) { return str.replaceAll(searchValue, () => replaceValue) }
 function range(...args) {
     if (args.length === 1) {
         const end = args[0];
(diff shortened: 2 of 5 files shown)


winklemad:
Still relevant and ready on my end — CI's green and the fix plus tests are in good shape, so I'd rather keep it open than let it close. Quick recap of why it matters: the TypeScript (prod) HogVM passes the replacement string to JS `String.replace`, so `$$`, `$&`, `` $` `` and `$'` get treated as special replacement patterns, while the Python and Rust VMs — and ClickHouse — treat the replacement literally. That's a real prod-vs-preview divergence for anyone using `replaceOne`/`replaceAll` with a `$` in the replacement. Happy to rebase if it's drifted from master; would appreciate a review whenever someone has a cycle.

winklemad:
Rebased onto master — the conflict was just placement: `sortable semver arrays use numeric component ordering` landed in the same spot in `execute.test.ts`. Both tests are independent, so I kept both.

Full hogvm suite is green locally on the rebased branch (`pnpm test` in `common/hogvm/typescript`, 6 suites / 174 tests), including the new semver test and this one:

```
✓ sortable semver arrays use numeric component ordering (7 ms)
✓ replaceOne/replaceAll treat the replacement literally (1 ms)
```

Still happy to keep this open — the fix is small and the behaviour it corrects (TS HogVM interpreting `$$`, `$&`, `` $` `` and `$'` in the replacement, unlike the Python interpreter and ClickHouse) is a real divergence between the two VMs.

PR #72146, opened 2026-07-18
feat(error-tracking): drop autocapture settings mirror

ablaszkiewicz:
Mistakes were made... I got really excited to move the `autocapture` field away from team model to our dedicated settings. Turned out just moving that field sounds pretty easy but then you've got some legacy consumers reading it. It would be pretty complex to migrate it.

I took a step back and realized that the amount of work required to move that field is simply not worth it.

This PR removes all the stuff which was added to make that migration happen.

Plan was to migrate that field to use error tracking scope but I think the way it works now (it requires project admin if RBAC on) is the right call. I added a special handling in ET settings where if you have ET write scope but you are not an admin, you get a nice message:
CleanShot 2026-07-18 at 14 38 45@2x

Migration you see in this PR is a PostHog's recommended way of getting rid of some column. You have to first do the `SeparateDatabaseAndState` and in the next migration you can remove the column. Otherwise migration risk analysis CI will block it.


hpouillot (approved):
the greptile comment looks worth fixing, otherwise 🧹

PR #72164, opened 2026-07-18
fix(feature-flags): normalize flag dependency keys to strings

### LINKED ISSUE #71795: Feature flag API accepts a numeric flag-dependency key, but /flags then returns 500 for the entire project
## Bug description

There's a validation asymmetry between the feature flag **write** path and the flag **evaluation** service for flag-dependency properties (`{"type": "flag", "key": <referenced flag id>, "operator": "flag_evaluates_to", ...}`):

- The REST API accepts the dependency `key` as a JSON **number**: `_validate_flag_reference` runs the reference through `safe_int`, so `"key": 226357` and `"key": "226357"` both validate and are stored as submitted.
- The flag evaluation service parses the project's flag configuration strictly. With a numeric `key` stored, `POST /flags?v=2` returns **HTTP 500 for every flag in the project**:

```
Failed to parse flag configuration data. This may indicate a misconfigured feature flag. Please check your flag definitions or contact support.
```

One wrongly-typed field in one flag takes down flag evaluation for the whole project — that's how we hit it in production (EU Cloud, project 218683, 2026-07-16→17): 4 flags written via the API with numeric dependency keys made all 81 flags evaluate as unavailable until each flag was re-saved in the UI (which rewrites the key as a string).

## How to reproduce

1. Create flag B. Create flag A with a release condition depending on flag B, via the **API**, with the dependency property's `key` as a JSON number (integer id of B) — the write is accepted (200).
2. Call `POST https://<region>.i.posthog.com/flags?v=2` with the project API key and any distinct_id.
3. Observe HTTP 500 with the message above; every flag in the project now fails to evaluate. SDKs receive no flag data.

## Expected behavior

Either of:

- **Write-time normalization/rejection**: the serializer already parses the reference with `safe_int` — it could store the canonical string form (or reject non-string keys with a 400 naming the field).
- **Resilient evaluation parsing**: skip the malformed flag (surface it in the flag's status/UI) and serve the rest of the project instead of failing the entire configuration.

Ideally both; the blast radius of one malformed flag being "all flags in the project silently gone" is the painful part, and the error body names neither the flag nor the field.

## Additional context

Encountered while migrating flags US→EU with a patched [posthog-migrate-meta](https://github.com/PostHog/posthog-migrate-meta) (remapping dependency IDs from the state file, which stores IDs as numbers — an easy trap for any API client, since the write API accepts it). Debug environment: PostHog Cloud EU.

igennova:
## Problem

Flag dependency filters can be saved with a numeric `key` (e.g. `226357`) via the REST API. Rust flag evaluation expects `PropertyFilter.key` to be a string, so hypercache deserialization fails and `/flags` returns 500 for the whole project. Re-saving the flag in the UI fixes it because the frontend writes string keys.

Closes #71795

## Changes

- Normalize flag-dependency property keys to stringified flag IDs in `FeatureFlagSerializer.validate_filters` on create/update
- Heal existing bad rows when building the flags cache (`_get_feature_flags_for_service` and batch path)
- Accept string or integer keys in Rust `PropertyFilter` deserialization as defense-in-depth for stale Redis cache entries
- Add regression tests for API save path and cache normalization

- `hogli test products/feature_flags/backend/test/test_flags_cache.py::TestNormalizeFlagDependencyKeys` 
- `cargo test -p feature-flags test_deserialize_property_filter_key_from_number` 
- API test `test_numeric_flag_dependency_key_is_stored_as_string` 

## Docs update

No docs change — internal serialization fix, no user-facing API contract change.

## 🤖 Agent context

Cursor agent assisted with root-cause analysis and implementation. Skills considered: `/improving-drf-endpoints`, `/writing-tests`. Chose stringified flag ID as canonical form (matches existing validation via `safe_int`) rather than resolving to flag name. Rust deserializer kept for rollout safety against stale cache; Python normalization is the primary fix.

### DIFF
--- a/products/feature_flags/backend/api/feature_flag.py
+++ b/products/feature_flags/backend/api/feature_flag.py
@@ -1575,7 +1575,22 @@ class FeatureFlagSerializer(
                 f"Please simplify conditions or reduce payload sizes."
             )
 
+        self._normalize_flag_dependency_property_keys(filters)
+
         return filters
+
+    def _normalize_flag_dependency_property_keys(self, filters: dict) -> None:
+        """Store flag-dependency property keys as strings (flag IDs) for Rust eval."""
+        from posthog.utils import safe_int
+
+        for flag_prop in _get_flag_properties_from_filters(filters):
+            flag_reference = flag_prop.get("key")
+            if flag_reference is None:
+                continue
+            self._validate_flag_reference(flag_reference)
+            flag_id = safe_int(flag_reference)
+            if flag_id is not None:
+                flag_prop["key"] = str(flag_id)
 
     def _validate_flag_reference(self, flag_reference):
         """Validate and convert flag reference to flag key."""
--- a/products/feature_flags/backend/api/test/test_feature_flag.py
+++ b/products/feature_flags/backend/api/test/test_feature_flag.py
@@ -419,6 +419,43 @@ class TestFeatureFlag(APIBaseTest, ClickhouseTestMixin):
         )
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
 
+    def test_numeric_flag_dependency_key_is_stored_as_string(self) -> None:
+        base_flag = FeatureFlag.objects.create(
+            team=self.team,
+            created_by=self.user,
+            key="base-flag-numeric-dep",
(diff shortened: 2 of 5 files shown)


igennova:
fixed by #71802, close as duplicate

PR #72223, opened 2026-07-20
feat(mcp): stateless protocol alongside legacy dialect

skoob13:
## Problem

The 2026-07-28 MCP specification revision ([SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575), finalizing July 28) removes the `initialize`/`notifications/initialized` handshake and protocol-level sessions from the Streamable HTTP transport. New-spec clients self-describe every request via reserved `_meta` keys, discover capabilities through a mandatory `server/discover` RPC, and never send or echo `Mcp-Session-Id`. Our MCP server only understood the legacy handshake dialect, so 2026-07-28 clients would be forced onto the fallback path forever, and their traffic would lose client attribution in analytics (we only extracted `clientInfo` from `initialize` bodies).

The Hono dispatcher was already architecturally stateless (hand-rolled per-request JSON-RPC dispatch, per-user state in Redis keyed by token hash), so this is a wire-dialect compat layer rather than an architectural change.

## Changes

The server now serves both dialects side by side, detecting the dialect per request from the presence of the `io.modelcontextprotocol/protocolVersion` `_meta` key.

```mermaid
flowchart LR
    C1{{Legacy client ≤2025-11-25}} --> I[initialize handshake]
    I --> S[Mcp-Session-Id minted + echoed]
    S --> D[dispatcher]
    C2{{Stateless client 2026-07-28}} --> M[per-request _meta: version + clientInfo]
    M --> V{version supported?}
    V -- no --> E[error -32022]
    V -- yes --> D
    D --> R1[legacy: byte-identical result envelope]
    D --> R2[stateless: resultType + serverInfo _meta + cache hints]
    class C1,C2 phYellow;
    class D phBlue;
    class E phRed;
    class I,S,M,V,R1,R2 phGray;
```

- New `src/lib/stateless-protocol.ts` with the shared constants: reserved `_meta` keys, the supported-version list (2026-07-28 plus every legacy SDK version), the `-32022` error code, and a `_meta` parser used by both runtimes.
- `server/discover` handler in the Hono dispatcher returning `supportedVersions`, capabilities, and per-user instructions. It shares the initialize side effects (context-mill revalidation, init metrics, the `$mcp_initialize` analytics event) via an extracted `recordDiscoveryRequest` helper, so stateless clients stay visible in init analytics. It never mints a session id.
- Stateless results get `resultType: "complete"` and the server identity in `_meta` (merged with any tool-supplied `_meta` such as the ext-apps resource URI). The five `CacheableResult` methods (`server/discover`, `tools/list`, `resources/list`, `resources/read`, `prompts/list`) additionally get `ttlMs`/`cacheScope: "private"`, since everything we serve varies per token or per user.
- Requests carrying an unknown `_meta` protocol version are rejected with `UnsupportedProtocolVersionError` (`-32022`).
- Client attribution fallback in both client-info extractors (Hono analytics path and the Worker wide-logs path): when no `initialize` message is in the body, client name/version/protocol version are read from per-request `_meta`. Initialize wins in mixed batches.
- Legacy behavior is untouched by design. `initialize` still negotiates from the pre-stateless version list (a new-spec client falling back gets the newest legacy version, per the spec's fallback negotiation) and still mints `Mcp-Session-Id`, and requests without protocol `_meta` get byte-identical result envelopes.
- `services/mcp/ARCHITECTURE.md` gained a "Protocol dialects" section documenting both dialects and the correlation story for stateless traffic.

> [!NOTE]
> Deliberately deferred to follow-ups: `Mcp-Method`/`Mcp-Name` header validation and Worker routing on them, the conversation-id-first session/analytics rework for stateless traffic, resource-not-found `-32602` alignment, and deterministic tool ordering.

Automated tests only (no manual testing against a live 2026-07-28 client, since the v2 client SDKs are still in beta):

- New `defineStatelessProtocolTests` in the shared protocol suite, speaking the 2026-07-28 wire format directly over fetch. Each test guards a distinct regression no existing test caught: `server/discover` full contract with no handshake and no session minted (handler removed/shape broken → new-spec clients cannot connect), a `tools/call` with only per-request `_meta` (stamping dropped, or handshake accidentally required), the `_meta` merge through the exec + PostHog Code path (serverInfo stamping clobbering the ext-apps resource URI), cache hints on `tools/list`, the `-32022` rejection, and a legacy-wire guard asserting requests without protocol `_meta` see no new fields.
- Two cases added to the existing `mcp-client-info` unit test: `_meta` attribution fallback (stateless traffic would silently lose client attribution) and initialize-over-`_meta` precedence in mixed batches.
- The stateful dialect is regression-proofed by the existing suites, which run the real v1 SDK `Client` + `StreamableHTTPClientTransport` through the initialize handshake, session-id lifecycle, tools, resources, and exec mode against the refactored dispatcher.

Results: full Hono suite 302 passed / 1 skipped (needs a second API token), all 2197 unit tests pass, lint and typecheck clean for `services/mcp`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

`services/mcp/ARCHITECTURE.md` updated in this PR.

## 🤖 Agent context

I (actually Claude Code) reviewed the 2026-07-28 SDK-betas announcement and the draft spec changelog/schema, mapped them against `services/mcp`, and implemented this compat layer. Skills invoked: `/writing-tests`. Key decisions: stay on the v1 SDK for types and own the supported-version list rather than adopting the beta v2 packages (the dispatcher hand-rolls JSON-RPC, so the SDK v2 migration largely does not apply); gate all new result fields per request dialect so legacy clients see byte-identical envelopes rather than trusting them to tolerate unknown fields; test the new dialect at the wire level since the v2 stateless client SDK is not vendored, while relying on the existing v1 SDK client suites as the stateful regression net. A first version of the `_meta`-merge test called a plain tool and failed, which surfaced that only the exec + PostHog Code consumer path puts `_meta` on call results; the test now goes through that real path.


services/mcp/src/hono/dispatcher.ts
```diff
@@ -166,62 +196,136 @@ class McpDispatcher {
...
 
+        const protocolMeta = parseRequestProtocolMeta(params)
+        if (protocolMeta.protocolVersion && !isSupportedProtocolVersion(protocolMeta.protocolVersion)) {
+            return jsonRpcMethodError(
+                id,
+                UNSUPPORTED_PROTOCOL_VERSION_ERROR_CODE,
+                `Unsupported protocol version: ${protocolMeta.protocolVersion}. Supported versions: ${ALL_SUPPORTED_PROTOCOL_VERSIONS.join(', ')}`
+            )
+        }
+        const stateless = protocolMeta.protocolVersion === STATELESS_PROTOCOL_VERSION
```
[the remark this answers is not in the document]
skoob13: Valid catch — the divergence was real, and worse than described: our own `supportedVersions` advertisement plus the version list in the -32022 error message actively steered conforming modern clients into retrying with a legacy version via `_meta`.

Fixed in b7a84fc, but in the opposite direction from the inline suggestion. Making dialect detection presence-based (`protocolVersion != null`) would statelessly decorate responses for a request declaring a legacy revision — stamping fields (`resultType`, `ttlMs`) that don't exist in the revision the request declared. The spec scopes per-request `_meta` to "revision 2026-07-28 and later", so legacy versions simply aren't selectable through this mechanism; they're only implemented behind the `initialize` handshake.

What changed:
- `server/discover` and `_meta` validation now share a modern-only version list (`MODERN_PROTOCOL_VERSIONS = ['2026-07-28']`) — the answer to the open question in the follow-up comment. Legacy clients never call discover, and legacy support remains discoverable via the era-fallback (`initialize`).
- A legacy or unknown `_meta` version is rejected with `UnsupportedProtocolVersionError`, now carrying the spec-required `data: { supported, requested }` payload (previously missing — conforming clients auto-retry from `error.data.supported`, so a message-only error broke that loop).
- Tests updated: discover asserts modern-only `supportedVersions`, and the rejection test is parameterized over an unknown and a legacy version, asserting the `data` payload.

services/mcp/src/hono/dispatcher.ts
```diff
@@ -50,7 +60,27 @@ const Method = {
...
+// cacheScope). Everything here varies per token (scopes, flags, staff gating)
+// or per user (discover instructions), so shared intermediaries must not cache.
+const CACHEABLE_METHODS: Set<string> = new Set([
+    Method.Discover,
+    Method.ToolsList,
+    Method.ResourcesList,
+    Method.ResourcesRead,
+    Method.PromptsList,
+])
+const CACHEABLE_RESULT_TTL_MS = 60_000
```
rafaeelaudibert: I wonder whether this will not confuse people locally because stuff will be cached? Can we somehow not have this locally?

rafaeelaudibert (approved):

services/mcp/src/hono/dispatcher.ts
skoob13: Yeah, good point. Wondering how they handle it in the SDK.

PR #72225, opened 2026-07-20
feat(conversations): link tickets table channel tag to slack thread

luke-belton:
## Problem

In the support tickets table, the Channel column shows a Slack tag but it is not clickable. The single-ticket detail view already makes its Channel tag link out to the originating Slack thread. This closes that gap so you can jump from the list straight into Slack without opening each ticket first.

## Changes

The `ChannelsTag` component already supported a `to` prop that renders an external-link glyph, an "Open in Slack" tooltip, and wraps the tag in a link. The table column just never passed it.

- Moved the `getChannelThreadUrl` helper out of `SupportTicketScene` (where it was private) and into `ChannelsTag`, next to the component that consumes its output, and exported it.
- The detail view now imports the shared helper instead of defining its own. Behavior there is unchanged.
- The table's Channel column now passes `to={getChannelThreadUrl(ticket)}`.

Only Slack rows that have both `slack_channel_id` and `slack_thread_ts` become clickable. Other channels (email, widget, teams, github) and Slack rows missing those IDs fall back to the existing non-clickable tag, so there is no regression. The table rows already carry those fields, so no API or serializer changes were needed.

CleanShot 2026-07-20 at 10 58 34

I (Claude) ran `pnpm --filter=@posthog/frontend typescript:check`. The three changed files report no type errors. I did not run the app or manually click through the UI, so the visual and click behavior in the tickets list has not been verified by a human yet.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Luke asked whether it would be much work to make the tickets-table Slack tag open the ticket in Slack, the same way the Customer panel in the ticket detail view does. It turned out to be a few lines: the shared `ChannelsTag` component already handled the clickable case, so the work was sharing the URL builder (`getChannelThreadUrl`) rather than duplicating it, then passing `to` from the table column. No new component or logic was introduced.

Tools: Claude Code (Explore agents for codebase search, then direct edits). No repo skills were invoked.

---

### DIFF
--- a/products/conversations/frontend/components/Channels/ChannelsTag.tsx
+++ b/products/conversations/frontend/components/Channels/ChannelsTag.tsx
@@ -60,7 +60,8 @@ export function ChannelsTag({ channel, detail, to }: ChannelsTagProps): JSX.Elem
         const tooltip = channelOpenLabel[channel] ?? `${channel}${detailText ? ` · ${detailText}` : ''}`
         return (
             <Tooltip title={tooltip}>
-                <Link to={to} target="_blank">
+                {/* Stop propagation so clicking the tag opens Slack without triggering a row/parent click. */}
+                <Link to={to} target="_blank" onClick={(e) => e.stopPropagation()}>
                     {tag}
                 </Link>
             </Tooltip>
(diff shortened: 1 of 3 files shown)


xljones (approved):
nice!

veryayskiy (approved):

PR #72227, opened 2026-07-20
fix(activity): key event-detail row expansion by event identity

pauldambra:
## Problem

On `/activity/explore`, expanding event rows tracked expansion by **positional row index**. A query refresh that inserted or reordered rows transferred the expanded state to a *different* event and remounted its detail subtree, and stale expansion state accumulated across query changes. This drove unnecessary remounts (and detached-element churn) on every poll/refresh.

## Changes

- Track expanded rows by a **stable identity** — `event.uuid` → `event.id` → positional-index fallback — via a new `ExpandedRowKey = string | number` type threaded through `tabUiStateLogic` and `dataTableLogic`. Expansion now stays on the same event across refreshes and reordering.
- **Reconcile** stale expansion keys in kea logic on each `loadDataSuccess`: drop keys for events no longer in the result set, so the collection can't grow unbounded. Reconciliation lives in the logic, not a component effect, and is guarded so a `null`/empty result can't wipe persisted expansion.
- **Scope the React `rowKey`** to tables that actually drive expansion by identity (the same guard as `expandableConfig`: `expandable && columnsInResponse` includes `*`). Every other DataTable consumer (error tracking, sessions, traces, LLM/marketing analytics) keeps LemonTable's positional keys, so their reconciliation is unchanged. Within that scope, duplicate event ids are disambiguated (`dedupeRowKeys`) so repeated uuids — ClickHouse at-least-once delivery / pre-merge `ReplacingMergeTree` state, undeduped `EventsQuery` results — can't produce colliding React keys.
- Multiple expansion is preserved (not converted to an accordion). Non-event DataTables keep positional behavior. No changes to generic `LemonTable` internals.

Scoped strictly to the expansion logic — no `PropertiesTable` or `EventDetails` changes. The property-rendering follow-up (windowing large property lists) is a separate PR: #72402.

Agent-run automated tests only (no manual/browser verification):

- `frontend/src/queries/nodes/DataTable/dataTableLogic.test.ts` — `stable expansion identity` group: UUID keys survive inserts/reorder; reconciliation drops departed events; positional fallback when no uuid; and `dedupeRowKeys` gives duplicate event ids distinct keys.
- `frontend/src/lib/logic/tabUiStateLogic.test.ts` — `ExpandedRowKey[]` behavior plus existing isolation/persistence cases.
- `pnpm --filter=@posthog/frontend fix` (oxlint + oxfmt) clean; `typescript:check` reports no errors in changed files.

The agent could not run the Playwright/browser path or verify production detached-DOM metrics; that verification is left for post-deployment.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Produced with Claude Code (PostHog Code). Originally bundled a property-rendering change; at the author's request the rendering work was split into #72402 so this PR is only the keyed-expansion fix. Skills invoked: `/writing-tests`, `/writing-kea-logics`.

Decisions:
- Keyed expansion on `event.uuid` (then `id`, then positional fallback) so non-event tables and events lacking both ids keep working.
- Reconciliation placed in a kea listener on `loadDataSuccess` (not a `useEffect`) so it's testable through the logic and stays out of render; guarded against `null` results so a transient empty load can't erase expansion.
- Identity-based React `rowKey` limited to expandable event tables and made collision-safe against duplicate event ids, so the change can't affect reconciliation for other DataTable consumers.

---


frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -364,6 +500,41 @@ export const dataTableLogic = kea<dataTableLogicType>([
...
         ],
+        getExpandedRowKey: [
+            (s, p) => [s.columnsInResponse, p.query],
+            (
+                columnsInResponse: string[] | null,
+                query: DataTableNode
+            ): ((row: DataTableRow, rowIndex: number) => ExpandedRowKey) => {
+                const starColumnIndex = columnsInResponse?.includes('*')
+                    ? columnsInResponse.indexOf('*')
+                    : isEventsQuery(query.source)
```
pauldambra: **[convergent: qa-team/data-integrity + paul + xp]** 🟡 MEDIUM

The `starColumnIndex` fallback `columnsInResponse?.includes('*') ? indexOf('*') : isEventsQuery(query.source) ? 0 : null` guesses that column 0 of an events query is "the event". That isn't guaranteed — `select: ['person', 'event', ...]` would put the person object at column 0, and `visibleExpandedRowKeys` would mint duplicate, non-unique keys (same person uuid across many rows). Latent today because the expansion UI requires `*` in the response, but the reconciliation path already consumes these bogus keys.

Safer to only trust `indexOf('*')` and fall back to the positional index otherwise, dropping the `isEventsQuery ? 0` guess. Paul/xp also flagged the `0` as an unexplained/magic value that's effectively dead in the expand path.

frontend/src/queries/nodes/DataTable/DataTable.tsx
```diff
@@ -742,25 +743,46 @@ export function DataTable({
...
                         expandedRowRender: function renderExpand({ result }: DataTableRow) {
                             if (
                                 (isEventsQuery(query.source) || isRevenueExampleEventsQuery(query.source)) &&
                                 Array.isArray(result)
                             ) {
-                                return <EventDetails event={result[columnsInResponse.indexOf('*')] ?? {}} />
+                                return (
+                                    <EventDetails
+                                        event={result[columnsInResponse.indexOf('*')] ?? {}}
+                                        tableProps={{ pagination: { pageSize: 50, hideOnSinglePage: true } }}
```
pauldambra: **[qa-team/frontend]** 🔴 HIGH

The new `tableProps={{ pagination: { pageSize: 50, hideOnSinglePage: true } }}` enables LemonTable's router-backed pagination, but `PropertiesTable` passes no `id` to `LemonTable`, so `usePagination` stores the current page in the shared `?page=` search param. Concrete failure: expand two events that each have >50 properties and click "next" on one — the scene URL becomes `/activity/explore?page=2` (a new history entry per click, so Back walks pagination states), and every other paginated properties table jumps to page 2 too (the second expanded event, plus the `$set`/`$set_once`/flags tabs in the same EventDetails all share the one `page` param). The stale `?page=` also persists after collapsing and leaks into shareable URLs.

Give each embedded table a unique `id` (e.g. derived from event uuid + tab key) or avoid router-backed pagination for expanded-row detail tables.

**[convergent: paul + xp]** 🟢 LOW/NIT — the `{ pageSize: 50, hideOnSinglePage: true }` literal is inlined twice (here and line 768). Hoist a `const EVENT_DETAIL_TABLE_PROPS = { pagination: { pageSize: 50, hideOnSinglePage: true } }` so the bound lives in one named place and the two call-sites can't drift.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -236,19 +350,41 @@ export const dataTableLogic = kea<dataTableLogicType>([
...
-        toggleRowExpanded: ({ rowIndex }) => {
-            actions.toggleExpandedRow(undefined, props.vizKey, rowIndex)
+    listeners(({ props, actions, values }) => ({
+        toggleRowExpanded: ({ rowKey }) => {
+            actions.toggleExpandedRow(undefined, props.vizKey, rowKey)
+        },
+        // Reconcile stale expansion whenever a fresh result set arrives: drop keys for
+        // events no longer visible so the collection can't grow unbounded across
+        // pagination/query changes, and a reordered result can't inherit expansion.
+        loadDataSuccess: () => {
```
pauldambra, line 382: **[qa-team/reliability]** 🟠 MEDIUM

The `loadDataSuccess` listener reconciles unconditionally. `dataNodeLogic.loadData` can resolve `null`/`undefined` on several paths (no `currentTeamId` in shared/export contexts, empty HogQL query string, `validateQuery` failure, malformed override query), and kea-loaders still fires `loadDataSuccess(null)`. In that case `dataTableRows` is `null`, `visibleExpandedRowKeys` returns `[]`, and `reconcileExpandedRows` permanently wipes every persisted expansion key for the vizKey — treating "no data received" as "no rows are visible". Same fail-silent erasure on a transient zero-result load: expand an event, change the filter to no matches, revert — the event is back but its expansion key was destroyed in between.

Guard with `if (values.dataTableRows === null) { return }` so reconciliation only runs against an actual result set.

frontend/src/lib/logic/tabUiStateLogic.ts
```diff
@@ -93,12 +113,24 @@ export const tabUiStateLogic = kea<tabUiStateLogicType>([
...
-                toggleExpandedRow: (state, { tabId, vizKey, rowIndex }) => {
+                toggleExpandedRow: (state, { tabId, vizKey, rowKey }) => {
+                    const tabState = state[tabId] ?? {}
+                    const existing = tabState[vizKey] ?? []
+                    const next = existing.includes(rowKey)
+                        ? existing.filter((k: ExpandedRowKey) => k !== rowKey)
+                        : [...existing, rowKey]
+                    return { ...state, [tabId]: { ...tabState, [vizKey]: next } }
+                },
+                reconcileExpandedRows: (state, { tabId, vizKey, visibleKeys }) => {
```
pauldambra, line 142: **[qa-team/reliability]** 🟢 LOW

The unbounded-growth fix only prunes within a vizKey while that table is mounted and reloading. Ad-hoc DataTables (no `uniqueKey`) get a fresh vizKey every mount, so any row expanded there leaves an entry under the `NO_TAB` bucket keyed by a vizKey that will never exist again — `reconcileExpandedRows` never fires for it after unmount, and `clearTabUiState(tabId)` never deletes `NO_TAB`. Per-session, `expandedRowsByTabAndVizKey[NO_TAB]` still grows monotonically with every ad-hoc table a user expands a row in.

Worth dropping the vizKey slot when reconciliation empties it (the reducer currently keeps `[vizKey]: []` forever), or scoping entries to real tab ids.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -364,6 +500,41 @@ export const dataTableLogic = kea<dataTableLogicType>([
             // keeping the previous array lets memoized rows skip re-rendering.
             { resultEqualityCheck: objectsEqual },
         ],
+        getExpandedRowKey: [
+            (s, p) => [s.columnsInResponse, p.query],
+            (
+                columnsInResponse: string[] | null,
+                query: DataTableNode
+            ): ((row: DataTableRow, rowIndex: number) => ExpandedRowKey) => {
+                const starColumnIndex = columnsInResponse?.includes('*')
```
pauldambra: **[paul]** 🟢 LOW

the `uuid` then `id` derivation reads well and the tests cover it. one edge to poke at: `candidate.id` — for an events query is `id` stable across reloads the same way `uuid` is? if `id` can be a per-response positional/synthetic value for some source, two reloads could hand the same event different `id`s and we'd quietly lose expansion (harmless) — or worse, collide. if you're confident `id` is always the real event id here, ignore me. mostly flagging because "fall back to a second identity field" is the kind of thing that's fine until the day it isn't.

pauldambra (commented):
> [!NOTE]

Four-reviewer pass (qa-team specialists + generalists, paul-reviewer, xp-reviewer, security-audit). The core mechanism is sound — uuid-keyed expansion with positional fallback, reconciliation in a kea listener, good regression tests — and security-audit found **no exploitable impact** (keys only gate a boolean; event-derived uuids reach array/Set membership, never computed object keys, so no prototype-pollution or collision sink; reconciliation is prune-only within an in-memory, non-persisted store). Two real defects need fixing before merge, plus polish.

### Key findings

**🔴 HIGH — URL pagination collision across property tables** (`DataTable.tsx:760`) — `PropertiesTable` passes no `id` to `LemonTable`, so router-backed pagination stores page in the shared `?page=` param; multiple expanded events' tables (and `$set`/flags tabs) share one page state and clicks push history entries onto the scene URL.

**🟠 MEDIUM — `starColumnIndex` `0` fallback mints duplicate keys** (`dataTableLogic.ts:511`) — convergent (3 reviewers). The `isEventsQuery ? 0` guess assumes column 0 is the event; `select: ['person', ...]` breaks that. Drop the guess, only trust `indexOf('*')`, fall back to positional index.

**🟠 MEDIUM — reconciliation wipes all keys on `loadDataSuccess(null)`** (`dataTableLogic.ts:375`) — `loadData` resolves `null` on several paths; the listener treats "no data" as "no rows visible" and erases all persisted expansion. Guard with `if (values.dataTableRows === null) return`.

**🟢 LOW — ad-hoc vizKey growth in the `NO_TAB` bucket** (`tabUiStateLogic.ts:124`).

**🟢 LOW — `candidate.id` reload-stability** (`dataTableLogic.ts:509`).

**⚪ NIT — test `tabId` props are misleading** (`tabUiStateLogic.test.ts:79`).

### Convergence
- `starColumnIndex` `0` fallback → qa-team/data-integrity + paul + xp (3 reviewers, highest confidence)
- duplicated `pageSize: 50` literal → paul + xp (2 reviewers)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Sound core; URL-pagination collision + reconcile-on-null are real defects; ad-hoc vizKey growth + misleading test tabId are residuals |
| 👤 paul | Approve with caveats — magic `50` duplicated, unexplained `0` fallback, nudge to make the bound observable, question on `id` stability |
| 📐 xp | Clean and well-tested; LOW/NIT polish only |
| 🛡 security-audit | No findings — no auth/IO/injection boundary touched |

---

frontend/src/lib/logic/tabUiStateLogic.test.ts
```diff
@@ -76,7 +76,7 @@ describe('tabUiStateLogic', () => {
         })
         logic.mount()
         logic.actions.toggleRowExpanded(4)
-        expect(logic.values.expandedRows).toEqual([4])
+        expect(logic.values.expandedRowKeys).toEqual([4])
```
pauldambra, line 79: **[qa-team/generalist]** ⚪ NIT

The dataTableLogic-based tests pass `tabId: TAB_A` in props, but `DataTableLogicProps` has no `tabId` field and dataTableLogic hard-codes `undefined` (→ `NO_TAB`) in both `toggleExpandedRow` and `reconcileExpandedRows`. The test passes with any tabId value and documents tab-scoped isolation that the production wiring doesn't actually provide — all DataTables across scene tabs share the `NO_TAB` bucket, distinguished only by vizKey. Either thread the real scene tab id through, or remove the misleading `tabId` from the test props so the shared-bucket behavior is explicit.

frontend/src/queries/nodes/DataTable/DataTable.tsx
pauldambra: Fixed in b72c359ae539faf643b0341d2b35a0e365b1594a: EventDetails now derives a per-tab, per-event `id` (`event-details-${eventId}-${tabKey}`) and passes it through to LemonTable so each embedded properties table gets its own `?<id>_page=` param instead of sharing `?page=`. Also hoisted the pagination props into a shared `EVENT_DETAIL_TABLE_PROPS` constant per the convergent nit.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
pauldambra: Fixed in b72c359ae539faf643b0341d2b35a0e365b1594a: `getExpandedRowKey` now only trusts an explicit `*` column for `starColumnIndex`; dropped the `isEventsQuery ? 0` guess entirely, so a non-`*` events query falls straight through to the positional-index path instead of risking a bogus key off column 0.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
pauldambra: Fixed in b72c359ae539faf643b0341d2b35a0e365b1594a: `loadDataSuccess` now returns early when `values.dataTableRows === null`, so reconciliation only runs against an actual (possibly empty) result set and can't wipe persisted expansion keys on a null/undefined load.

frontend/src/lib/logic/tabUiStateLogic.ts
pauldambra: Fixed in b72c359ae539faf643b0341d2b35a0e365b1594a: `reconcileExpandedRows` now deletes the vizKey slot entirely once it empties, instead of leaving `[vizKey]: []` around forever, so ad-hoc tables (fresh vizKey per mount) don't leak entries into the NO_TAB bucket for the rest of the session.

frontend/src/lib/logic/tabUiStateLogic.test.ts
pauldambra: Fixed in b72c359ae539faf643b0341d2b35a0e365b1594a: removed the misleading `tabId` prop from the dataTableLogic test fixtures since `DataTableLogicProps` has no such field — the tests already assert against the real (shared NO_TAB bucket) behavior via `expandedRowsFor(undefined, vizKey)`, so this just stops implying tab-scoped isolation that doesn't exist in production.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
pauldambra: Leaving this open for a human call: whether `id` is guaranteed stable across reloads (vs. a per-response synthetic value) for every source that reaches this fallback depends on backend/query-source knowledge I don't have high confidence in from the diff alone. Deferring rather than guessing at a fix — happy to follow up once someone can confirm `id` stability for the relevant sources.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -16,6 +16,7 @@ import { FEATURE_FLAGS } from 'lib/constants'
 import { dayjs } from 'lib/dayjs'
 import { featureFlagLogic } from 'lib/logic/featureFlagLogic'
 import { tabUiStateLogic } from 'lib/logic/tabUiStateLogic'
+import { ExpandedRowKey } from 'lib/logic/tabUiStateLogic'
```
pauldambra: **[qa-team/generalist]** ⚪ NIT (round 2)

Duplicate import lines from `'lib/logic/tabUiStateLogic'` (lines 18-19):
```ts
import { tabUiStateLogic } from 'lib/logic/tabUiStateLogic'
import { ExpandedRowKey } from 'lib/logic/tabUiStateLogic'
```
Merge into one statement and make `ExpandedRowKey` a type-only import: `import { tabUiStateLogic, type ExpandedRowKey } from 'lib/logic/tabUiStateLogic'`.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -364,6 +504,37 @@ export const dataTableLogic = kea<dataTableLogicType>([
             // keeping the previous array lets memoized rows skip re-rendering.
             { resultEqualityCheck: objectsEqual },
         ],
+        getExpandedRowKey: [
+            (s) => [s.columnsInResponse],
+            (columnsInResponse: string[] | null): ((row: DataTableRow, rowIndex: number) => ExpandedRowKey) => {
+                // Only column 0 holding `*` guarantees the full event is at that position — an
```
pauldambra: **[xp]** 🟢 LOW (round 2)

The comment says "Only column 0 holding `*` guarantees the full event is at that position" but the code below does `columnsInResponse.indexOf('*')` — it accepts `*` at any position, not just column 0. The code is correct (an events query can select `person` first, then `*`); the comment is stale and even contradicts itself two sentences later. Trim the "column 0" phrasing so intent and code agree.

frontend/src/scenes/activity/explore/EventDetails.tsx
```diff
@@ -30,15 +31,15 @@ interface EventDetailsProps {
...
-    }
+    const getEventId = (event: ErrorPropertyTabEvent): string => String(eventIdentityKey(event) ?? '')
+    const eventId = getEventId(event)
+
+    // Each tab's PropertiesTable paginates independently, so give every embedded LemonTable a
+    // unique id — otherwise LemonTable's router-backed pagination stores the current page under
+    // the shared `?page=` search param, and paging one expanded event's table pages every other
+    // expanded table (and every tab of this same EventDetails) at once.
+    const tablePropsForTab = (tabKey: string): Partial<LemonTableProps<Record<string, any>>> | undefined =>
+        tableProps ? { ...tableProps, id: `event-details-${eventId}-${tabKey}` } : undefined
```
pauldambra: **[qa-team/frontend]** 🟢 LOW (round 2)

The per-event pagination ids fix the collision but trade it for one-way URL param accumulation: paging an expanded event's table pushes `event-details-<uuid>-<tabKey>_page=N` into the search params (usePagination only ever `push`es, never removes), and because the key embeds the event uuid it's effectively single-use. Collapsing the row, refreshing, or navigating leaves the stale param behind, so a session of browsing expanded events steadily grows the URL and junk keys travel into shared links. Consider clearing the param on collapse/unmount, or using in-memory (non-router) pagination for these embedded tables since URL persistence has no value for a transient expanded row. (Design decision — not a merge blocker.)

pauldambra (commented):
> [!NOTE]

## Round 2 — Verdict: 💬 APPROVE WITH NITS

Re-review after round-1 fixes + the simplify pass. **All six round-1 fixes verified as sound** — the `${id}_page` namespacing isolates each expanded event's tab pagination, the `starColumnIndex` `0` guess is gone with no memo regression, the null-rows guard prevents the wipe, the emptied-slot drop works on the reconcile path, and `eventIdentityKey` is falsy-for-falsy equivalent to the explicit `if`s it replaced. **security-audit again found no exploitable impact** — the new event-derived id reaching `usePagination`'s search-param key is neutralized: kea-router percent-encodes keys on write and parses into a null-prototype object on read, and the `event-details-` prefix + `_page`/`_order` suffix structurally prevent collision with the default `page` param or dangerous key names (`__proto__`, `constructor`).

Remaining findings are all LOW/NIT polish:

**🟢 LOW — stale `_page` URL params accumulate over a session** (`EventDetails.tsx:42`) — usePagination only `push`es, never removes; per-uuid keys are single-use and linger in the URL/shared links. Design decision (non-router pagination) — not a merge blocker.
**🟢 LOW — `eventId === ''` collision for id-less events** (`EventDetails.tsx:42`) — rare (events always carry a uuid); convergent across qa-team/paul/xp.
**⚪ NIT — duplicate import line** (`dataTableLogic.ts:19`) — merge into one `type` import.
**🟢 LOW — stale "column 0" comment** (`dataTableLogic.ts:510`) — code accepts `*` at any position; comment contradicts it.
**⚪ NIT — empty-slot cleanup can't fire for toggle-collapse** (`tabUiStateLogic.ts`) — tiny in-memory growth; partial completion of the round-1 growth fix.

The deferred human question from round 1 (reload-stability of `candidate.id` across all event-query sources; making the bound observable) remains open.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | All round-1 fixes hold; residual is URL param accumulation + id-less-event edge + duplicate import |
| 👤 paul | Approving — round-1 concerns addressed; `eventIdentityKey` reads well; noted `''` edge and id `0` falsy (unrealistic) |
| 📐 xp | Clean XP work — the extraction is the right abstraction, not a wrong one; stale comment + twin EventDetails blocks are cosmetic |
| 🛡 security-audit | No findings — the new id→search-param surface is not exploitable (percent-encoding + null-proto parse + prefix/suffix guards) |

---

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
pauldambra: currently we're keying on row index which is absolutely not stable, so this is an OK trade-off

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -71,6 +71,45 @@ export interface DataTableRow {
...
+    return !!row.result
+}
+
+/**
+ * A stable identifier for an event-like object: its `uuid`, falling back to `id`. `EventDetails`'s
+ * `getEventId` uses the same fallback to key per-event UI state, though on a narrower event type.
+ */
+export function eventIdentityKey(candidate: any): ExpandedRowKey | null {
+    if (candidate && typeof candidate === 'object') {
+        return candidate.uuid || candidate.id || null
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟡 MEDIUM

Identity keys and positional fallback keys share one namespace, so they can collide. `eventIdentityKey` returns `candidate.uuid || candidate.id`, and `id` may be numeric. Then `getRowExpansionKey` falls back to `rowIndex` (also numeric). Failure scenario: a row whose `result.id` is `5` gets key `5`; another row without uuid/id at index `5` also gets key `5` — `expandedRowKeys.includes(5)` matches both, so expanding one renders both expanded (the exact wrong-row bleed this PR fixes). React also stringifies keys, so numeric id `5` and positional `5` collide into the same sibling key.

Parked as MEDIUM because it needs an event with a numeric `id` and no `uuid` — near-impossible for star-column event tables in practice. But it's cheap to make impossible by construction: namespace the two spaces and constrain types, e.g. return `id:${uuid ?? id}` only for string/number values and `ix:${rowIndex}` for the fallback.

frontend/src/queries/nodes/DataTable/DataTable.tsx
```diff
@@ -1036,9 +1053,11 @@ export function DataTable({
...
                                 }
                                 dataSource={dataTableRows ?? NO_ROWS}
-                                rowKey={(_, rowIndex) => {
-                                    return rowIndex
-                                }}
+                                // Key rows by the same stable identity used for expansion, so a
+                                // refresh that reorders events keeps each row's React key (and its
+                                // expanded EventDetails subtree) instead of remounting it. Label
+                                // rows fall back to their positional index via getExpandedRowKey.
+                                rowKey={getExpandedRowKey}
```
pauldambra: > [!NOTE]

**[qa-team/reliability]** 🟡 MEDIUM

`rowKey={getExpandedRowKey}` changes React row keys for *every* DataTable consumer (error tracking, sessions, traces, LLM/marketing analytics), not just the activity table this PR targets — and unlike the old positional keys, identity keys are only unique if the data cooperates. ClickHouse can return duplicate event rows with the same uuid (at-least-once ingestion / pre-merge ReplacingMergeTree state); EventsQuery results aren't deduped. Two rows with the same uuid get identical React keys → duplicate-key warnings, rows dropped/mis-rendered on update, and coupled expansion. Consider scoping the identity-based rowKey to when expansion identity is actually in play (`expandable && columnsInResponse?.includes('*')`) and keeping positional keys elsewhere, or uniquifying repeats at the table boundary.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -236,19 +345,52 @@ export const dataTableLogic = kea<dataTableLogicType>([
             }),
             ['response', 'responseLoading', 'responseError'],
         ],
-        actions: [tabUiStateLogic, ['toggleExpandedRow']],
+        actions: [
+            tabUiStateLogic,
+            ['toggleExpandedRow', 'reconcileExpandedRows'],
+            dataNodeLogic({
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/reliability]** 🟢 LOW

The `dataNodeLogic({ key, query, dataNodeCollectionId, loadPriority })` config is built twice — once in the `values` connect array and once in this `actions` array — with identical config. kea dedupes to the same instance so it's correct today, but the two copies must stay byte-identical: drift and you connect `loadDataSuccess` to a *different* instance and the reconcile silently stops firing. Hoist it to a `const dataNodeLogicProps = {...}` above the connect and reference it twice.

frontend/src/queries/nodes/DataTable/dataTableLogic.ts
```diff
@@ -236,19 +345,52 @@ export const dataTableLogic = kea<dataTableLogicType>([
...
+        loadDataSuccess: () => {
+            // Nothing to reconcile against an empty expansion set — skip computing
+            // visibleExpandedRowKeys (an O(rows) pass) on every load/poll when it can't matter.
+            if (values.expandedRowKeys.length === 0) {
+                return
+            }
+            // A `null` result (no team in export/shared context, an empty query, a validation
+            // failure) means "no data received", not "no rows are visible" — reconciling against
+            // it would wipe every persisted expansion key for this vizKey.
+            if (values.dataTableRows === null) {
```
pauldambra: > [!NOTE]

**[qa-team/reliability]** 🟢 LOW

An empty-but-successful result set permanently wipes persisted expansion. Scenario: user expands two events, then narrows a filter/date range matching zero rows → `loadDataSuccess` with `results: []` → `dataTableRows` is `[]` (passes the null guard) → reconcile against `[]` deletes every key for the vizKey; reverting the filter restores the rows but not the expansion. If intended, fine — but consider also skipping reconciliation when `dataTableRows.length === 0`, and add a one-line comment on this case. Also worth a test pinning the load-bearing null guard — deleting it would silently wipe all persisted expansion on any null response, and nothing currently asserts that.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments + summary. Verdict: 💬 approve with nits.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers (qa-team/reliability on fable, paul-reviewer on opus)

## Verdict: ✅ APPROVE <sub>(round 2 @ f96b7e9)</sub>

Round-1 findings were addressed in commits `988f145` (key-namespace + connect-config fixes) and `f96b7e9` (simplify pass). Round 2 re-reviewed the changed code and found **no new issues and no regressions** — key prefixing is applied at the single `getExpandedRowKey` chokepoint and consumed consistently (no unprefixed producers/comparators remain anywhere in the repo), the hoisted `dataNodeLogicProps` const binds both connect arrays to the same instance, the inlined reconcile loop is equivalent to the removed selector, and both test suites pass (37/37).

### Disposition of round-1 findings

- 🟡 **Numeric-id vs positional-index key collision** — ✅ fixed: keys now namespaced `id:` / `ix:`, collision impossible by construction.
- 🟢 **Duplicated `dataNodeLogic` connect config** — ✅ fixed: hoisted to a single const.
- 🟢 **Empty-result wipes expansion** — resolved as intentional (reviewer hedged "if intended, fine"; wiping on a genuine empty result set reads as reasonable UX).
- 🟡 **`rowKey={getExpandedRowKey}` affects all DataTable consumers / duplicate-uuid collision** — ⏸️ **deferred to the author**: applies at the shared component across error tracking, sessions, traces, LLM/marketing analytics; more than one reasonable fix with cross-product behavioral tradeoffs. Thread left open — see the inline comment on `DataTable.tsx:1060`.
- ⚪ NITs (comment overclaim, falsy fallthrough, session-resident dead vizKeys) — informational, no change needed.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔧 qa-team/reliability (fable) | Round 1: sound machinery, weaknesses in key-namespace design. Round 2: all fixes verified, no regressions, no new findings |
| 👤 paul (opus) | "lovely one to review — literally deleting my least favourite pattern." Only the duplicated connect config worth tidying before merge (now done) |

<details>

round 1 @ 178427f — 💬 APPROVE WITH NITS: 2 MEDIUM + 2 LOW + NITs; 1 actioned, 1 promoted (key namespacing), 1 nit resolved, 1 deferred (rowKey scoping).

</details>

---

frontend/src/queries/nodes/DataTable/DataTable.tsx
pauldambra: Addressed in 94cb73c. The identity-based `rowKey` is now scoped to the same guard as `expandableConfig` (`!context.expandable && expandable && columnsInResponse?.includes('*')`), so every other DataTable consumer (error tracking, sessions, traces, LLM/marketing analytics) keeps LemonTable's positional keys — reconciliation there is unchanged.

Within that scope, `dedupeRowKeys` disambiguates repeated event ids: the first occurrence keeps its key, later duplicates get a positional `#n` suffix, so duplicate-uuid rows (ClickHouse at-least-once / pre-merge ReplacingMergeTree, undeduped EventsQuery results) can no longer produce colliding React keys. Covered by a new unit test in `dataTableLogic.test.ts`.

PR #72231, opened 2026-07-20
feat(hogql): detect events queries with no timestamp filter

pauldambra:
## Problem

An events query with no `timestamp` filter scans the full history of the events table. These queries dominate query cost on shared/embedded dashboards, where every anonymous view re-runs them, and there is currently no way to know when one runs. Compiled insight queries always inject a date range, so the offenders are all hand-written HogQL.

Raised from a [Slack thread](https://posthog.slack.com/archives/C076E99B152/p1784539559707949?thread_ts=1784539559.707949&cid=C076E99B152) investigating why a few customers' sharing-token query cost was disproportionate: a handful of custom SQL insights on public dashboards were doing full-table events scans on every view.

Review on this PR then surfaced a deeper problem worth fixing at the root rather than per call site: there was no way to capture backend telemetry that is both non-blocking (safe on the synchronous request path) and reliably delivered from Celery workers. The global client's event queue was never flushed at worker-child exit — billiard calls `os._exit()` right after the `worker_process_shutdown` signal, so the SDK's background flush and atexit hooks never run — which is why the repo guidance banned plain `posthoganalytics.capture()` in Celery tasks, and why every alternative kept getting flagged in review: `ph_scoped_capture`'s unbounded per-call flush can block the request thread, and a per-query scoped client inside the query-worker pool is a worker-exhaustion hazard when the ingestion endpoint degrades.

## Changes

**Detector** (unchanged since the last review round)

- `posthog/hogql/unbounded_events_query.py`: an AST visitor (`query_reads_events_without_timestamp_filter`) that flags any `SELECT` reading the events table directly while no filter position that can prune the scan (`WHERE`/`PREWHERE`, respecting boolean structure; not `HAVING`) references the events `timestamp` column. Type-aware with a chain-matching fallback; handles subqueries, CTEs, joins, and predicate pushdown from an enclosing bounded SELECT.
- `posthog/hogql/query.py`: on the ClickHouse execution path, when `query_type == "HogQLQuery"` and the events table is referenced, capture an `unbounded events query` event with attribution from the current query tags (team, kind, access method, product/feature, insight/dashboard/session ids). Best-effort — wrapped so telemetry can never affect query execution — and gated on `is_cloud()`.

**Telemetry capture — fixed systemically instead of per call site**

- `posthog/celery.py`: `on_worker_process_shutdown` now flushes the SDK's **event queue** alongside the existing metrics-window flush — bounded, in parallel daemon threads against one shared deadline, abandoned on timeout. The bound (10s) must exceed the SDK's 5s flush interval or healthy recycles would drop the last task's events. This is the only delivery opportunity in a prefork child, and it makes plain `posthoganalytics.capture()` safe from every Celery task repo-wide — not just this PR's call site.
- `posthog/hogql/query.py`: `_capture_unbounded_events_query` is now a single unconditional non-blocking `posthoganalytics.capture()`. The query-kind branch and the per-query scoped client are gone — nothing on the query path can flush or block, on any execution path. The event gains a `kind` property so the alerting insight can segment request vs worker emissions.
- `posthog/ph_client.py`: `ph_scoped_capture`'s exit is now bounded — `shutdown()` runs on a daemon thread joined for 10s and abandoned on timeout — and the scoped client is built with `flush_interval=0.3` so a healthy exit is sub-second instead of stalling ~5s waiting out the SDK's default first-batch hold. This removes the unbounded-wait hazard at every existing call site.
- `CLAUDE.md`/`AGENTS.md`: the "never capture in Celery" rule is replaced with the full backend-telemetry decision rule — when a signal is an event vs a first-party metric, plain capture is safe everywhere (including Celery), call sites only enqueue while lifecycle hooks deliver, and `ph_scoped_capture` is for short-lived processes without a lifecycle hook or when events must be flushed before moving on. Also routes agents to the `instrumenting-first-party-metrics` skill when in that area.

**Why an event rather than a first-party metric**

Review kept circling capture reliability, which raises the fair question of whether this should be a metric instead. It shouldn't: the signal's value is per-occurrence attribution — insight/dashboard/session ids are how you find the offender — which is exactly the cardinality metric attributes must not carry, and the intended consumer is an insight threshold alert, which counts events just as well. Metrics also buy no delivery reliability: their Celery-safety comes from the same `worker_process_shutdown` window flush this PR extends to the event queue, so both sinks now share symmetric delivery guarantees. The decision rule is written into `AGENTS.md` so future telemetry doesn't re-litigate this per call site.

Once this lands, the event supports a normal insight + threshold alert (e.g. routed to Slack) so an unbounded events query can be caught whenever it runs.

- `posthog/test/test_celery.py`: extended the worker-shutdown flush tests — asserts the event queue is flushed with the bounded timeout, and (parameterized over event/metrics flush) that a **hung flush is abandoned within the bound** rather than stalling worker recycling — the regression test review asked for. Added an `event_flush_raises` case to the never-breaks-shutdown test.
- `posthog/test/test_ph_client.py`: new `test_hung_shutdown_does_not_block_context_exit` — locks in the bounded exit so `ph_scoped_capture` can't regress to an unbounded wait on request threads or worker slots.
- `posthog/hogql/test/test_query.py`: `test_unbounded_events_capture_routes_by_query_kind` replaced by `test_unbounded_events_capture_is_single_nonblocking_enqueue` — asserts one enqueue with the right payload and that nothing on the query path calls flush/shutdown (the property that kept regressing across review rounds). The span test is unchanged.
- `posthog/hogql/test/test_unbounded_events_query.py`: detector suite unchanged.

Ran locally: detector suite, ph_client tests, and celery hook tests all pass (49 tests). The DB-backed `test_query.py` classes need Postgres, which is unavailable in this environment — CI runs them.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

`CLAUDE.md`/`AGENTS.md` capture guidance updated in this PR (see Changes).

## 🤖 Agent context

Authored by the PostHog Slack app from the linked thread; capture mechanism reworked after review. Invoked the repo `writing-tests` gate before adding/changing tests.

The review rounds on the capture mechanism (global client drops events in workers → scoped capture blocks the request path → kind-branch still leaves a per-query unbounded flush on the worker path) converged on fixing the platform instead: flush the event queue, bounded, at Celery child exit — after which the correct call-site code is the trivial one. `posthog/celery.py` and `posthog/ph_client.py` are platform-shared files riding along in a hogql PR for that reason; the query-path change is only correct together with the worker-exit flush.

---


posthog/hogql/query.py
```diff
@@ -640,6 +643,10 @@ def _execute_clickhouse_query(self):
                 ),
             )
 
+            if "events" in hogql_features.tables:
+                with self.timings.measure("capture_unbounded_events_query"):
+                    self._capture_unbounded_events_query()
```
georgemunyoro: Compiled insight queries (trends/funnels/retention) nearly all touch events, so on cloud this adds a 2nd full `TraversingVisitor` walk to every insight query, and compiled queries always inject a date range, so there's nothing to catch. Could we short-circuit on `get_query_tag_value("contains_user_hogql")` first?

posthog/hogql/query.py
```diff
@@ -699,6 +706,45 @@ def run_clickhouse_query() -> Any:
...
+        try:
+            if not is_cloud():
+                return
+            if not query_reads_events_without_timestamp_filter(self.select_query):
+                return
+            tags = get_query_tags()
+            # ph_scoped_capture (not the global client): query execution also runs in Celery
+            # workers, where the global SDK's background flush may never run before the worker
+            # exits and would silently drop the event.
+            with ph_scoped_capture() as capture:
```
georgemunyoro: `ph_scoped_capture()`'s exit path is an unbounded wait, and this call site is predominantly the synchronous web request thread, running before the ClickHouse query: `finally: shutdown()` -> `flush(timeout_seconds=None)` -> bare `queue.join()` -> `consumer.join()` (no timeout). The consumer does `task_done()` even on failed uploads, but with `retries=10` × 15s timeout + capped backoff + Retry-After sleeps, a degraded ingestion endpoint blocks a flagged query for minutes, and flagged queries are exactly the hot shared-dashboard ones that re-run per anonymous view. The `try/except` cannot catch this: the blocking happens in the context manager's `finally`, not as an exception.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,130 @@
...
+            self.found = True
+
+    def visit_select_query(self, node: ast.SelectQuery) -> None:
+        # A timestamp bound inside a nested SELECT doesn't constrain this one — don't descend.
+        pass
+
+
+def _field_is_events_timestamp(expr: ast.Field, events_ids: set[str]) -> bool:
+    """``True`` iff ``expr`` references the events table's ``timestamp`` column. Uses type info
+    when resolved; otherwise falls back to the chain, attributing a qualified ``x.timestamp`` only
```
georgemunyoro, line 189: this is is a near line-for-line copy of `feature_extractor._looks_like_event_field` (only the column name differs); a shared parameterized helper would keep them from drifting

georgemunyoro (approved):

andyzzhao (approved):
lgtm. This makes me wonder if we can have an AST check that looks for all known anti-patterns

posthog/hogql/query.py
```diff
@@ -699,6 +708,45 @@ def run_clickhouse_query() -> Any:
...
+            if not is_cloud():
+                return
+            if not query_reads_events_without_timestamp_filter(self.select_query):
+                return
+            tags = get_query_tags()
+            # Non-blocking capture on the global client: this runs on the synchronous query path,
+            # before the ClickHouse query, so it must never block. The SDK enqueues to a background
+            # consumer and returns immediately — a scoped client's flush-on-exit would instead do an
+            # unbounded wait that could stall a flagged query for minutes if ingestion is degraded.
+            posthoganalytics.capture(
```
pauldambra: > [!NOTE]

**[convergent: router + paul]** 🟠 HIGH

`_capture_unbounded_events_query` uses the **global** `posthoganalytics.capture()` client, but this method also runs inside the Celery task `process_query_task` (process_query_task → execute_process_query → process_query_dict → `HogQLQueryExecutor._execute_clickhouse_query`). That's the async/cached recalculation path shared and embedded dashboards ride on — the exact scenario this telemetry targets. Both CLAUDE.md and `posthog/ph_client.py` document that `posthoganalytics.capture()` in Celery tasks silently loses events because the global client's background flush may never run before the worker exits, and `posthog/celery.py` has no flush hook at task/worker teardown. An earlier commit on this branch (72c7a1847e5) reverted a working `ph_scoped_capture()` implementation to unblock the sync path, without preserving Celery-safety — so the events we most want get dropped on the one path where the global client is unreliable.

Not a hard block (best-effort telemetry, well-guarded by try/except + `is_cloud()`), but cheap to fix and there's precedent: `posthog/caching/warming.py` already does scoped capture on the Celery query path, and `execute.py` already branches on `tags.kind`. `get_query_tags()` is already fetched at line 724, so the branch is free. Suggested shape — default to the safe scoped client and special-case only the live request, so a future offline `kind` can't silently fall into the lossy path:

```python
if tags.kind == "request":
    posthoganalytics.capture(...)      # non-blocking, latency matters here
else:
    with ph_scoped_capture() as capture:  # celery/temporal/dagster — flush before exit
        capture(...)
```

posthog/hogql/query.py
```diff
@@ -699,6 +708,45 @@ def run_clickhouse_query() -> Any:
...
+        chase down unbounded queries (they dominate cost on shared/embedded dashboards, where every
+        anonymous view re-runs them). Best-effort: never let telemetry affect query execution.
+        """
+        try:
+            if not is_cloud():
+                return
+            if not query_reads_events_without_timestamp_filter(self.select_query):
+                return
+            tags = get_query_tags()
+            # Non-blocking capture on the global client: this runs on the synchronous query path,
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

This comment asserts the global non-blocking client is chosen "because this runs on the synchronous query path" — but it also runs under Celery (see the HIGH finding below), so the premise is wrong. Rewrite it as part of the `kind`-branch fix, or it will actively mislead the next reader into thinking the Celery path was considered and ruled safe.

posthog/hogql/query.py
```diff
@@ -699,6 +708,45 @@ def run_clickhouse_query() -> Any:
...
+            # before the ClickHouse query, so it must never block. The SDK enqueues to a background
+            # consumer and returns immediately — a scoped client's flush-on-exit would instead do an
+            # unbounded wait that could stall a flagged query for minutes if ingestion is degraded.
+            posthoganalytics.capture(
+                distinct_id=str(tags.team_id or self.team.pk),
```
pauldambra: > [!NOTE]

**[paul]** ⚪ NIT

`distinct_id=str(tags.team_id or self.team.pk)` but the `team_id` property uses `self.team.pk`. Same team in practice so harmless, but the inconsistency makes you look twice — just use `self.team.pk` for both.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+    operand does (``timestamp > x OR event = 'y'`` still admits all history); a leaf bounds when it
+    references the events ``timestamp`` column."""
+    if expr is None:
+        return False
+    if isinstance(expr, ast.And):
+        return any(_expr_bounds_events_timestamp(e, events_ids) for e in expr.exprs)
+    if isinstance(expr, ast.Or):
+        return bool(expr.exprs) and all(_expr_bounds_events_timestamp(e, events_ids) for e in expr.exprs)
+    finder = _TimestampFieldFinder(events_ids)
+    finder.visit(expr)
```
pauldambra, line 168: > [!NOTE]

**[router]** ⚪ NIT

The timestamp finder treats any reference to the events `timestamp` in a filter position as "bounded," regardless of operator — e.g. `WHERE timestamp != '2024-01-01'` counts as bounded even though it barely constrains the scan. The module docstring calls this out as intentional ("we deliberately don't pre-judge cost here"), so not a functional bug — just flagging so the tradeoff (a nearly-full-table scan reported as bounded) is a conscious one.

pauldambra (commented):
> [!NOTE]

Multi-perspective review complete (router first-pass + paul-reviewer delegation). See inline comments and the summary comment.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 3 @ e84a8d6b)</sub>

The delta since the last round (docstring clarifications + hoisting the `is_cloud()` gate above the tracer span) is clean — verified line-by-line against the code it documents, no delegation warranted. Four detector-precision design questions from round 2 remain open as deferred threads for the author (JOIN `ON` bounds, `NOT` inversion, aggregate-subquery pushdown, per-scope CTE shadowing); they degrade only telemetry-signal precision, not query correctness.

### Key findings

**NIT**
- `query.py:736` — the new "span skipped when not cloud" ordering isn't pinned by a test (both capture tests mock `is_cloud=True`). Optional, not blocking.

### Convergence

None this round.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | 47-line delta (docstrings + gate hoist): accurate docs, safe reordering (`is_cloud()` is a pure settings check), danger LOW, confidence HIGH, nothing delegated. |

<details>
<summary>Previous rounds (2)</summary>

round 2 @ f9098a25 — 💬 APPROVE WITH NITS: systemic capture rework verified correct against the SDK source (concurrency delegation found nothing); 3 MEDIUM detector-precision gaps (JOIN ON false positive; NOT-inversion and aggregate-pushdown false negatives) + 2 LOW + 2 NIT posted; docstring nits fixed in eed778f1, four design questions deferred to the author.
round 1 @ 8dfb6492e3 — ⚠️ REQUEST CHANGES: telemetry silently dropped on the Celery path (global client, no worker flush); fixed since by the worker-exit event flush + plain-enqueue capture rework.

</details>

---

posthog/hogql/query.py
pauldambra: Resolved in 95fea83. The capture now branches on the query-tag kind: the synchronous web-request path keeps the non-blocking global client (never touches `ph_scoped_capture`, so it can't block before the ClickHouse query), and only the worker kinds (Celery/Temporal/Dagster) use `ph_scoped_capture` — where a flush-on-exit is what we want, since the global client's background flush may never run before the worker exits and the event would be dropped. Unknown kinds default to the flushing path so a future offline kind can't silently fall into the lossy one.

posthog/hogql/query.py
pauldambra: Fixed in 95fea83 using the suggested `tags.kind` branch: `kind == "request"` keeps the non-blocking global client, everything else uses `ph_scoped_capture` so the event flushes before the worker exits. Defaulted the else-branch to the safe path so an unknown/offline kind can't silently drop the event. Added `test_unbounded_events_capture_routes_by_query_kind` to lock this in.

posthog/hogql/query.py
```diff
@@ -713,6 +723,51 @@ def run_clickhouse_query() -> Any:
...
+
+                # On the synchronous web-request path capture must never block — this runs before the
+                # ClickHouse query — so use the global client, which enqueues to a background consumer and
+                # returns immediately. Everything else (Celery/Temporal/Dagster) runs where that background
+                # flush may never happen before the worker exits, silently dropping the event — exactly the
+                # shared/embedded-dashboard recalc path this telemetry targets — so flush before exit there.
+                if tags.kind == "request":
+                    posthoganalytics.capture(distinct_id=distinct_id, event=event, properties=properties, groups=groups)
+                else:
+                    with ph_scoped_capture() as capture:
```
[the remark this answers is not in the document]
pauldambra: Resolved at the root in f9098a2 + 2295696: the worker path no longer creates a per-query client or flushes at all. `_capture_unbounded_events_query` is now a single non-blocking enqueue on the global client for every query kind, so there is nothing left on the query path for an attacker to tie up — enqueueing a flagged query can no longer hold a worker slot regardless of ingestion-endpoint health.

Delivery from workers moved to where it amortizes correctly: `on_worker_process_shutdown` (posthog/celery.py) now flushes the event queue once per worker-child lifetime, bounded at 10s in a daemon thread and abandoned on timeout. `ph_scoped_capture` itself is also bounded now (daemon-thread shutdown joined for 10s), so its other call sites can't block on a degraded endpoint either.

posthog/hogql/query.py
```diff
@@ -713,6 +723,46 @@ def run_clickhouse_query() -> Any:
...
+                    "client_query_id": tags.client_query_id,
+                    "scene": tags.scene,
+                    "$process_person_profile": False,
+                }
+                groups = {"organization": str(tags.org_id)} if tags.org_id else None
+
+                if tags.kind == "request":
+                    posthoganalytics.capture(distinct_id=distinct_id, event=event, properties=properties, groups=groups)
+                else:
+                    with ph_scoped_capture() as capture:
```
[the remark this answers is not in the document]
pauldambra: Resolved at the root in f9098a2 + 2295696, going one step further than the suggested bounded per-query flush: the per-query client and its flush are gone entirely. The capture is now a single non-blocking enqueue on the global client for every query kind, so `process_query_task` slots can no longer be held by telemetry at all — bounded or otherwise.

Worker delivery moved to `on_worker_process_shutdown` (posthog/celery.py), which now flushes the event queue alongside the metrics window: once per worker-child lifetime, bounded at 10s (the bound must exceed the SDK's 5s flush interval or healthy recycles would drop events), abandoned on timeout via daemon threads.

The requested regression tests exist at the layers where a hang is still possible: `test_hung_flush_does_not_stall_worker_recycling` (parameterized over event/metrics flush) in posthog/test/test_celery.py, `test_hung_shutdown_does_not_block_context_exit` in posthog/test/test_ph_client.py, and `test_unbounded_events_capture_is_single_nonblocking_enqueue` in posthog/hogql/test/test_query.py asserting the query path never calls flush/shutdown.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+    type_ = getattr(expr, "type", None)
+    while isinstance(type_, (ast.TableAliasType, ast.ColumnAliasedTableType)):
+        type_ = type_.table_type
+    if isinstance(type_, ast.TableType) and isinstance(type_.table, EventsTable):
+        return True
+    # No type info — match the FROM identifier, but a CTE named "events" shadows the real table.
+    return isinstance(expr, ast.Field) and bool(expr.chain) and expr.chain[0] == "events" and "events" not in cte_names
+
+
+def _has_events_timestamp_filter(node: ast.SelectQuery, events_ids: set[str]) -> bool:
```
pauldambra, line 146: > [!NOTE]

**[router]** 🟡 MEDIUM

`_has_events_timestamp_filter` only inspects `node.where` and `node.prewhere` for a timestamp bound — it never looks at a JOIN's `ON`/`USING` constraint (`ast.JoinExpr.constraint`). A hand-written query like `SELECT * FROM events AS e JOIN persons AS p ON p.id = e.person_id AND e.timestamp > now() - interval 1 day` (verified empirically) is reported as unbounded even though the JOIN condition does prune the events side for an INNER JOIN. This runs counter to the module's stated goal ("err toward precision so a genuinely bounded query isn't mislabeled") and produces false positives in the alerting signal. Non-blocking (telemetry only), so low urgency — but worth either a doc note scoping this out explicitly or a follow-up walking `join.constraint` the same way WHERE/PREWHERE are walked (careful with LEFT/RIGHT/FULL JOIN semantics, where an ON-clause filter on the non-preserved side does not bound the preserved side).

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+    # HAVING is evaluated after aggregation over the full scan, so it never bounds the read —
+    # only WHERE / PREWHERE do.
+    return _expr_bounds_events_timestamp(node.where, events_ids) or _expr_bounds_events_timestamp(
+        node.prewhere, events_ids
+    )
+
+
+def _expr_bounds_events_timestamp(expr: ast.Expr | None, events_ids: set[str]) -> bool:
```
pauldambra, line 154: > [!NOTE]

**[qa-team/generalist-a]** 🟡 MEDIUM

**`NOT` inverts boolean structure but the detector ignores it, producing false negatives.** `_expr_bounds_events_timestamp` models `And` (any operand bounds) and `Or` (all operands must bound) explicitly, but has no `Not` case — `NOT` is descended via the inherited `TraversingVisitor`, so the leaf finder locates a bare `timestamp` underneath it and reports the query as bounded. This breaks De Morgan equivalence: `WHERE NOT (event = 'x' AND timestamp > now())` returns bounded while its logical equivalent `WHERE event != 'x' OR timestamp <= now()` correctly returns unbounded. The `NOT(AND(...))` form genuinely admits all history, so this is a real false negative in the same class the OR-handling was built to prevent. Repro: `query_reads_events_without_timestamp_filter(parse_select("SELECT count() FROM events WHERE NOT (event = 'x' AND timestamp > now())"))` → `False`. Suggested fix: add an `isinstance(expr, ast.Not)` branch that recurses with AND/OR semantics flipped, or conservatively treat any `Not` wrapping the timestamp reference as non-bounding.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+        self.found = False
+        # Whether the SELECT currently being entered is fed by a timestamp-bounded outer query. Set
+        # only while descending into a bounded SELECT's single FROM source / CTEs, since ClickHouse
+        # pushes an outer predicate down into the table feeding a query — but not across a join or
+        # into an independent WHERE / SELECT-list subquery (``... IN (SELECT … FROM events)``).
+        self._from_bound = False
+
+    def visit_select_query(self, node: ast.SelectQuery) -> None:
+        events_ids = _events_table_identifiers(node, self.cte_names)
+        bounded = self._from_bound or _has_events_timestamp_filter(node, events_ids)
```
pauldambra, line 77: > [!NOTE]

**[qa-team/data-integrity]** 🟡 MEDIUM

**An outer "pushdown" bound is credited without verifying the enclosing bound is on a column the subquery projects raw from the events scan — false negatives on aggregate/GROUP BY subqueries.** Any bare-`timestamp` reference in the outer WHERE makes `bounded` true, and the bound propagates into the single FROM subquery — but ClickHouse can only push the predicate into the events scan if the subquery exposes that column pre-aggregation. It does not for: (a) an aggregate subquery that doesn't select timestamp at all — `SELECT event, c FROM (SELECT event, count() c FROM events GROUP BY event) WHERE timestamp > now()`; (b) a subquery exposing post-aggregation `max(timestamp) AS timestamp` — the realistic "latest event per user, then filter recent" pattern `SELECT distinct_id FROM (SELECT distinct_id, max(timestamp) AS timestamp FROM events GROUP BY distinct_id) WHERE timestamp > now() - interval 7 day`. Both scan all of events but are reported bounded. This is the subquery-boundary analog of the HAVING case the module deliberately excludes. Suggested fix: when the subquery/CTE GROUP BYs (or the projected `timestamp` originates from an aggregate call), don't treat an outer timestamp bound as pushing into it; at minimum document this as a known false-negative class alongside the HAVING note.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+    """``True`` iff some SELECT in the tree reads the events table with no timestamp bound able to
+    constrain its scan — its own filters or an enclosing SELECT's (which push down into it)."""
+    if node is None:
+        return False
+    detector = _UnboundedEventsQueryDetector(_collect_cte_names(node))
+    detector.visit(node)
+    return detector.found
+
+
+def _collect_cte_names(node: ast.Expr) -> set[str]:
```
pauldambra, line 45: > [!NOTE]

**[qa-team/data-integrity]** 🟢 LOW

**CTE names are collected globally across the whole tree, so a CTE literally named `events` in one scope shadows the real events table in unrelated scopes — false negative.** `_collect_cte_names` flattens every CTE name anywhere in the query into one set, and `_is_events_table` treats FROM `events` as "not the real table" whenever `"events" in cte_names`, regardless of visibility. Repro: `WITH inner_q AS (WITH events AS (SELECT 1 AS x) SELECT x FROM inner-scope events) SELECT count() FROM events` → reported bounded even though the outer SELECT reads the real table unbounded; the production gate (`"events" in hogql_features.tables`) still passes, so it's reachable. Requires a user CTE named exactly `events`, so impact is limited. Suggested fix: resolve CTE shadowing per-scope (track the CTE names visible at each SELECT) rather than one global set.

posthog/hogql/unbounded_events_query.py
```diff
@@ -0,0 +1,200 @@
...
+    def visit_field(self, node: ast.Field) -> None:
+        if _field_is_events_timestamp(node, self.events_ids):
+            self.found = True
+
+    def visit_select_query(self, node: ast.SelectQuery) -> None:
+        # A timestamp bound inside a nested SELECT doesn't constrain this one — don't descend.
+        pass
+
+
+def _field_is_events_timestamp(expr: ast.Field, events_ids: set[str]) -> bool:
```
pauldambra, line 187: > [!NOTE]

**[qa-team/generalist-b]** 🟢 LOW

Any reference to the events `timestamp` in WHERE/PREWHERE counts as a bound, including references that cannot prune the scan — e.g. `WHERE timestamp IS NOT NULL`, `WHERE toString(timestamp) != ''`, `WHERE timestamp != now()` all report bounded despite scanning all history. This looks deliberate (the docstring commits to "a leaf bounds when it references the events timestamp column"; full range-pruning analysis would require semantic evaluation), so advisory only: worth a one-line acknowledgment in the docstring that a non-range timestamp reference is intentionally treated as bounded, so future maintainers don't mistake these for detector bugs and the alert consumer understands the precision floor.

posthog/celery.py
```diff
@@ -231,32 +234,52 @@ def on_worker_process_shutdown(**kwargs) -> None:
```
pauldambra, line 236: > [!NOTE]

**[router]** ⚪ NIT

`on_worker_process_shutdown`'s docstring still says only "Remove metric files for this child so recycled workers don't leak stale data," but the function's primary responsibility has grown substantially — it now performs the parallel, deadline-bounded flush of both the event queue and the metrics window, and per the updated AGENTS.md it is the load-bearing reason plain `posthoganalytics.capture()` is safe in Celery tasks. A one-line docstring update would keep this discoverable without reading the body.

posthog/hogql/test/test_query.py
```diff
@@ -2191,3 +2195,56 @@ def test_non_transient_errors_are_not_retried(self):
...
+        self.assertEqual(kwargs["distinct_id"], str(self.team.pk))
+        self.assertEqual(kwargs["event"], "unbounded events query")
+        self.assertEqual(kwargs["properties"]["team_id"], self.team.pk)
+        self.assertEqual(kwargs["properties"]["kind"], "celery")
+        # The query path must never flush — a degraded ingestion endpoint would block
+        # request threads and worker slots.
+        mock_analytics.flush.assert_not_called()
+        mock_analytics.shutdown.assert_not_called()
+
+    def test_unbounded_events_capture_emits_span(self) -> None:
```
pauldambra, line 2224: > [!NOTE]

**[router]** ⚪ NIT

`test_unbounded_events_capture_emits_span` mainly guards that a tracing span gets created around `_capture_unbounded_events_query` — a low-value regression to lock in (losing the span costs observability, not correctness) relative to the rest of the suite added here. Not wrong to keep, just the thinnest test in the set.

pauldambra (commented):
QA Swarm review complete. See inline comments.

posthog/hogql/query.py
```diff
@@ -713,6 +723,47 @@ def run_clickhouse_query() -> Any:
...
+        """Emit a telemetry event when a query reads the events table with no ``timestamp`` bound.
+
+        A full-history events scan is never intended — compiled insight queries always inject a date
+        range, so this only trips on hand-written HogQL. Emitting it as an event lets us alert on and
+        chase down unbounded queries (they dominate cost on shared/embedded dashboards, where every
+        anonymous view re-runs them). Best-effort: capture is a non-blocking enqueue on every path —
+        no flush happens here, so telemetry can never delay or block query execution. Delivery from
+        Celery workers is handled by the bounded flush in ``on_worker_process_shutdown``.
+        """
+        if not is_cloud():
```
pauldambra, line 739: > [!NOTE]

**[router]** ⚪ NIT

Hoisting `is_cloud()` above the tracer span/try-block is a real (disclosed, intentional) behavior change — self-hosted instances now skip creating the `_capture_unbounded_events_query` span entirely. This is safe (`is_cloud()` is a pure settings check, no exception risk) and matches the commit message's intent, but no test locks in the new "span skipped when not cloud" ordering — both existing tests mock `is_cloud` to `True`. Optional: patch `is_cloud` to `False` and assert the span/capture are not emitted, if this ordering matters going forward; not blocking given how low-risk the change is.

pauldambra (commented):
QA Swarm re-review of the new commits complete. See inline comment.

pauldambra:
Closing in favor of a simpler approach: the signal this PR derives at query time already exists as data in the query log (`query_log` / `raw_query_log` in HogQL, backed by `query_log_archive`), which records every executed query's text, kind, actual rows/bytes read, and attribution (insight/endpoint id, user, access method). A SQL insight over that table — filtered to hand-written HogQL reading events with no timestamp reference, floored on `read_rows` so only scans that actually hurt surface — expresses the same detection with zero hot-path code, works retroactively, and measures real cost instead of predicting prunability from the AST (which is where all four open detector-precision threads came from).

The telemetry-delivery fixes that grew out of review here (bounded event-queue flush at Celery worker shutdown in `celery.py`; bounded `ph_scoped_capture` exit in `ph_client.py`) are independently valuable and can be resurrected from this branch as a standalone platform PR if wanted.

Monitoring continues as a saved SQL insight + subscription instead.

PR #72248, opened 2026-07-20
feat(conversations): hoist current user to top of ticket assignee dropdown

xljones:
## Problem

When viewing a ticket, assigning yourself is the most common action, but the assignee dropdown lists roles first and buries the current user inside the alphabetical Users section.

## Changes

Hoists the current user to the top of the ticket assignee dropdown: below "Remove assignee" when that option is present, above the Roles section, and deduplicated out of the Users section below. The hoisted row is labeled with a muted "(you)" so it's obviously the current user. It is hidden while searching, following the existing `membersLogic` convention of dropping me-first ordering during search. The Users section now also hides entirely when there is nobody else to show. The workflow job input assignee picker reuses the same dropdown and inherits the ordering. Also swaps `resolveAssignee`'s lookup pool from `meFirstMembers` to plain `sortedMembers`, since ordering is irrelevant to a find-by-id.

Current user not assigned:

not-selected

Current user assigned (hoisted row active, "Remove assignee" above it):

selected

Verified live in local dev (screenshots above): unassigned and self-assigned states, assigning via the hoisted row, and search hiding the hoisted entry. Conversations jest suites and the frontend typecheck pass with no errors in the changed files. No new tests: this is presentation ordering over already-tested `membersLogic` selectors, and a logic test would need heavy membersLogic/rolesLogic mocking for little regression value.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed.

## 🤖 Agent context

Authored with PostHog Code (Claude), directed by @xljones. Skills invoked: /run-posthog for live verification, /debugging-ci-failures for triaging a kea typegen drift CI failure. The implementation follows the existing `meFirstMembers`/`selectableMembers` conventions in `membersLogic`, with the shared `AssigneeItem` row extracted as a separate refactor commit. The "(you)" label was added on review feedback; no such suffix convention existed in the codebase, so the first cut relied on position alone. One CI fix commit syncs the inline kea typegen block with full-typegen output (single-file typegen emitted `any` for a connected-logic value). The multi-select assignee filter on the tickets list and error tracking's parallel copy of this component intentionally keep their existing ordering.

---


luke-belton (approved):

luke-belton (commented):
should 'you' still appear in the list of assignees if it's already assigned to you?

xljones:
> should 'you' still appear in the list of assignees if it's already assigned to you?

Yes. Right now you do appear (but down the list a bit), and I think you should. It's still a valid option to reselect yourself but also might be even more confusing if you suddenly disappeared!

Can reconsider this later if we find it odd now it's hoisted.

PR #72253, opened 2026-07-20
chore: Make all notifications archivable

jordanm-posthog:
## Problem
Notifications should always be archivable since you can now see archived notifications

## Changes
Removes the archivable column

Locally

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context


products/notifications/backend/migrations/0019_remove_notificationevent_archivable.py
```diff
@@ -0,0 +1,18 @@
...
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[
+                migrations.RemoveField(
+                    model_name="notificationevent",
+                    name="archivable",
+                ),
+            ],
+        ),
+    ]
```
arthurdedeus, line 18: hmm I think this can be a bit problematic. have you checked https://posthog.com/handbook/engineering/safe-django-migrations#dropping-columns?

jordanm-posthog: I made sure to use `SeparateDatabaseAndState` so it's actually a no-op in SQL :) The column stays around, we just detached the field from the model

arthurdedeus: aaah, makes sense! I missed the `SeparateDatabaseAndState

arthurdedeus (approved):

PR #72008, opened 2026-07-17
refactor(alerts): add reusable alert editor components

MattPua:
## Problem

Insight and logs alert editors duplicate definition layout, notifications, advanced options, evaluation history, and editor structure. The shared pieces also need to work inside modals, scenes, and embedded sections without owning their container.

## Changes

Before:
CleanShot 2026-07-17 at 13 06 07@2x
CleanShot 2026-07-17 at 13 06 47@2x

After:
CleanShot 2026-07-17 at 13 05 57@2x
CleanShot 2026-07-17 at 13 06 30@2x

https://github.com/user-attachments/assets/c9d118af-cfdf-4570-831b-09dbca486d21

- Add reusable alert editor, definition, notification destination, advanced-options, and evaluation-history components
- Adopt the shared components in insight and logs alert editing while keeping product-specific state and behavior in each product
- Add consolidated Storybook coverage for the shared alert components
- Update the product alerting skill with frontend architecture and adoption guidance

## How did you test this code?

- Ran targeted alert Jest suites, 13 tests
- Built Storybook successfully
- Ran Oxfmt and Oxlint on changed frontend files
- Ran `./bin/hogli lint:skills` and `./bin/hogli build:skills`
- Ran `git diff --check`

> [!NOTE]
> The full frontend typecheck reports 179 existing errors outside the changed files. No errors reference the files changed by this PR.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

No public docs update needed. The product alerting skill now documents the shared frontend architecture and adoption path.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

- Built with Codex CLI through iterative human review
- Invoked `/qa-team`, `/writing-skills`, and the Skills Store `xp-reviewer`
- Kept product API calls, form state, validation, destination payloads, and scheduling behavior product-owned
- Removed pass-through abstractions and kept only components with shared presentation or behavior

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #66355, opened 2026-06-26
feat(billing): add billing alerts backend

willwearing:
## 3-PR stack

Billing alerts are split across 3 PRs:

1. **#66353:** schema and migration.
2. **#66355 (this PR):** backend, API, Temporal, destinations, generated clients, and MCP.
3. **#72260: shared billing alert UI.** A separate frontend PR built from Matt's reusable alert components in #72008.

Merge order is #66353, #66355, then the shared UI PR #72260. This PR targets `will/billing-alerts-migration`.

## Problem

Customers need organization-wide billing alerts when spend jumps and the notification needs to land in an operational channel.

The first version supports spend thresholds with Slack, Microsoft Teams, and webhook destinations. Email, quiet hours, weekend skipping, project envelopes, and country envelopes stay outside this version.

## Changes

This PR now owns backend behavior only:

- Billing metric evaluation and lifecycle persistence
- Shared alert state machine and scheduling integration
- Destination creation, validation, deletion, and HogFunction dispatch
- DRF API and generated clients
- Temporal activities, workflow, schedules, and worker registration
- MCP tools and backend coverage

I removed 2,031 lines of handwritten billing frontend and Kea code. The frontend is now isolated in the third PR and built from the reusable alert editor, definition, destination, advanced option, next evaluation, and history components merged in #72008.

## How did you test this code?

Automated, run locally through Flox against the current head:

- **200 backend tests pass** — 81 in `products/billing_alerts/backend/test`, plus 119 in the shared `products/alerts/backend/test`, which this PR's failure-policy change reaches.
- **`uv run mypy --cache-fine-grained .`** clean repo-wide (17,749 files).
- `ruff check` / `ruff format --check`, `tach check --dependencies --interfaces`, and `services/mcp` `pnpm lint-tool-names` all clean.

Two tests were added. Each was confirmed by reverting its fix and watching it fail, so it pins the defect rather than the current implementation:

- `test_transient_failure_with_attempts_remaining_does_not_advance_the_counter` — without the fix `consecutive_failures` reaches 5 across same-day retries, which trips BROKEN and disables the alert.
- `test_transient_fetch_failure_commits_groups_already_prepared` — without the fix an earlier organization's claim stays `evaluating` behind a live 15-minute lease, so the Temporal retry skips it.

[...]

PR #66353, opened 2026-06-26
feat(billing): add billing alerts schema

willwearing:
## 3-PR stack

Billing alerts are split across 3 PRs so each layer can be reviewed and deployed on its own:

1. **#66353 (this PR): schema and migration.**
2. **#66355: backend, API, Temporal, destinations, generated clients, and MCP.**
3. **#72260: shared billing alert UI.** The frontend is built from Matt's reusable alert components in #72008.

Merge order is #66353, #66355, then the shared UI PR #72260

## Problem

Billing alerts need durable tables before the service and UI code can deploy. Migration jobs aren't orchestrated with service deployments, so the schema has its own deployment boundary.

## Changes

Adds the Billing alerts Django app shell, organization-scoped alert configuration and event models, indexes, constraints, and the initial migration.

Team references use `db_constraint=False` and `DO_NOTHING` to avoid locking `posthog_team`. Historical events keep the original team ID, and the event-to-alert foreign key stays constrained inside the new billing alert tables.

## How did you test this code?

Codex ran:

```bash
bin/hogli test products/billing_alerts/backend/test/test_models.py
bin/hogli product:lint billing_alerts
uv run ruff check products/billing_alerts/backend
DEBUG=1 uv run python manage.py sqlmigrate billing_alerts 0001
DEBUG=1 uv run python manage.py makemigrations billing_alerts --check --dry-run
DEBUG=1 uv run python manage.py analyze_migration_risk --fail-on-blocked
```

The model test passed, migration drift was empty, migration risk returned 0, and `sqlmigrate` confirmed there is no FK constraint to `posthog_team`.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

No docs update in this migration-only PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

I asked Codex to rebase and validate the schema as the first part of a 3-PR stack. Codex used the repository migration, product lint, test-writing, and GitHub publishing guidance. The migration remains deliberately separate from backend and frontend behavior.

PR #72260, opened 2026-07-20
feat(billing): add billing alert management

willwearing:
## Problem

The first billing alerts PR had 2,031 lines of custom frontend and Kea code bolted onto an already large backend change.

Matt's #72008 shipped a reusable alert editor, definition row, notification editor, advanced options, next evaluation status, and history chart. Billing alerts should use those components and keep its own code focused on billing rules and API adapters.

## Changes

This is the third PR in the stack:

1. #66353 adds the schema and migration.
2. #66355 adds the backend, API, Temporal, destinations, generated clients, and MCP.
3. This PR adds the billing UI on top of Matt's shared alert components.

The replacement has 1,452 lines of product frontend. It supports list, create, edit, pause, delete, check now, Slack destination selection, advanced options, next evaluation status, and billing event history.

It adds 2 small shared component seams:

- A `workspaceSelector` slot in `AlertNotificationDestinationEditor`
- Billing value formatting and current-rule classification in `AlertEvaluationHistoryChart`

### Before

```mermaid
flowchart LR
    A[Billing tab] --> B[Custom billing wizard]
    B --> C[Billing-specific Kea and UI]
    C --> D[Billing alert API]
    class A phYellow;
    class B,C phGray;
    class D phRed;
```

### After

```mermaid
flowchart LR
    A[Billing tab] --> B[Shared alert editor]
    B --> C[Billing rule adapters]
    C --> D[Billing alert API]
    B --> E[Shared destinations and history]
    class A phYellow;
    class B,E phBlue;
    class C phGray;
    class D phRed;
```

### Shared editor in use

Shared billing alert editor

### Firing state after the API trigger

Billing alert firing after API trigger

### Delivered to Slack

Billing alert delivered to Slack

Automated, run locally through Flox against the current head:

- **18 Jest tests pass** across 4 suites in `products/billing_alerts/frontend`.
- Frontend TypeScript check clean for `billing_alerts`. The repo reports pre-existing errors elsewhere on this branch from the unbuilt `@posthog/quill` workspace; none are in this product.
- `oxlint` and `oxfmt --check` clean on the changed files.

Earlier in this stack's life the whole path was exercised against a real local Billing instance: an alert created in this UI, evaluated over authenticated HTTP against local Billing, persisted as a firing event, dispatched through HogFunction, and delivered to Slack and to a webhook. Screenshots are attached to the resolved review threads here and on #66355.

> [!IMPORTANT]
> **Two snapshots need approving in Visual Review.** `products-alerts-shared-components--evaluation-history` differs in light and dark because this PR intentionally changes the story: one data point now sets `wouldFireUnderCurrentConfiguration`, and the chart formats the y-axis as currency. Use the newest "Visual changes detected" comment on this PR for the current run.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The product README now explains the frontend boundary and the shared components it composes. Customer docs should land once launch behavior is settled.

## 🤖 Agent context

I asked Codex to remove my old frontend from #66355, rebuild it with Matt's merged shared components, run the app through Flox, and prove the full UI-to-Slack path. Skills invoked included execution planning, Playwright, GitHub publishing, Kea logic, generated API types, DRF endpoints, test writing, MCP tools, and notification delivery.

Codex also used read-only Slack and desktop automation to verify and capture the delivered message. The E2E fixture was kept at the upstream billing HTTP boundary so the alert lifecycle and delivery path stayed real.


willwearing:
Fixed the clipped notification dropdown.

The notification selector now prefers opening upward, and the shared Popover no longer forces a 150px minimum when the viewport has less space available.

Reproduced and verified in the real local app at a 797 × 256 CSS viewport. The menu stays inside the viewport and Slack, Microsoft Teams, and Webhook are all fully visible.

Notification dropdown fully visible

products/billing_alerts/frontend/BillingAlertNotifications.tsx
```diff
@@ -0,0 +1,118 @@
...
+        deleteDestination,
+    } = useActions(logic)
+
+    const existingDestinations: AlertNotificationDestinationView[] = (props.alert?.destinations ?? []).map(
+        (destination) => ({
+            key: `${destination.type}-${destination.hog_function_ids.join('-')}`,
+            title: destinationLabel(destination.type),
+            detail: 'Firing, resolved, errored, and auto-disabled notifications',
+            tags: [{ label: 'Active', type: 'success' }],
+            viewAction: destination.hog_function_ids[0]
```
willwearing: > [!NOTE]

**[convergent: qa-team/security + qa-team/general correctness and safety + paul]** 🟠 HIGH

“Open destination” sends billing-owned HogFunctions into the generic editor for the current project. This breaks when `execution_team_id` differs, and a project user without organization billing access may be able to reroute or disable billing notifications. The action also selects an arbitrary function from the 4-function lifecycle group. Billing destinations need a protected, group-aware action scoped to the execution project.

products/billing_alerts/frontend/billingAlertFormLogic.ts
```diff
@@ -0,0 +1,266 @@
...
+                if (!props.alert && pending.length === 0) {
+                    lemonToast.error('Add at least one notification destination.')
+                    throw new Error('A notification destination is required.')
+                }
+
+                let created: BillingAlertConfigurationApi | null = null
+                let createdDestinationKeys: string[] = []
+                try {
+                    const formPayload = payload(form)
+                    const saved = props.alert
```
willwearing: > [!NOTE]

**[convergent: xp + qa-team/general correctness and safety]** 🟠 HIGH

The browser owns a multi-request save transaction: it PATCHes the alert, creates destinations one by one, then reports the whole save as failed if any later request fails. The configuration and some destinations can already be persisted. Even a destination-only save sends a non-empty PATCH, which increments `configuration_revision`, clears pending evaluation state, and reschedules the check. Move this orchestration behind one shared backend command, or explicitly model partial success and send only changed fields.

products/billing_alerts/frontend/billingAlertNotificationLogic.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+
+function isMicrosoftTeamsWebhook(value: string): boolean {
+    if (!isHttpsUrl(value)) {
+        return false
+    }
+    const url = new URL(value.trim())
+    const hostname = url.hostname
+    const path = url.pathname
+    if (hostname.endsWith('.logic.azure.com')) {
```
willwearing: > [!NOTE]

**[xp]** 🟡 MEDIUM

This copies the backend Microsoft Teams URL policy into the browser, and the copies disagree. The frontend accepts any port on `*.logic.azure.com`, while the serializer accepts only 443 or the default port. Keep the backend allowlist authoritative and use only basic input completeness in the product UI.

products/billing_alerts/frontend/BillingAlertHistory.tsx
```diff
@@ -0,0 +1,172 @@
...
+): AlertEvaluationHistoryPoint[] {
+    return events
+        .map((event) => ({ event, value: eventValue(alert, event) }))
+        .filter((item): item is { event: BillingAlertEventApi; value: number } => item.value !== null)
+        .sort((left, right) => left.event.created_at.localeCompare(right.event.created_at))
+        .map(({ event, value }) => ({
+            label: dayjs(event.created_at).format('MMM D, HH:mm'),
+            value,
+            firedAtTime: event.state_after === 'firing',
+            wouldFireUnderCurrentConfiguration: wouldFire(alert, event),
```
willwearing: > [!NOTE]

**[xp]** 🟡 MEDIUM

`wouldFireUnderCurrentConfiguration` reuses deltas calculated with the historical baseline window. After `baseline_window_days` changes, those stored deltas cannot answer whether the event would fire under the current configuration. Return a backend classification with enough data, or omit this marker when the current baseline cannot be reconstructed.

frontend/src/scenes/billing/BillingSection.tsx
```diff
@@ -14,25 +17,43 @@ import { BillingSpendView } from './BillingSpendView'
...
         ? 'spend'
         : location.pathname.includes('usage')
           ? 'usage'
-          : 'overview'
+          : location.pathname.includes('alerts') && billingAlertsEnabled
+            ? 'alerts'
+            : 'overview'
+
+    useEffect(() => {
+        if (location.pathname.includes('alerts') && !billingAlertsEnabled) {
```
willwearing: > [!NOTE]

**[paul]** 🟡 MEDIUM

The redirect can race feature-flag loading. A valid `/organization/billing/alerts` deep link can be replaced with overview while `receivedFeatureFlags` is still false, just before the billing-alert flag arrives. Wait for a confirmed flag result before deciding the route is unavailable.

products/billing_alerts/frontend/BillingAlertEditor.tsx
```diff
@@ -0,0 +1,232 @@
...
+                isEditing={props.alert !== null}
+                isSubmitting={isAlertFormSubmitting}
+                hasChanges={alertFormChanged}
+                hasPendingChanges={pendingDestinations.length > 0}
+                leadingActions={
+                    props.alert ? (
+                        <LemonButton
+                            type="secondary"
+                            icon={<IconPlay />}
+                            onClick={() => checkNow(props.alert!)}
```
willwearing, line 69: > [!NOTE]

**[paul]** 🟡 MEDIUM

“Check now” can run live evaluation and dispatch real Slack, Teams, or webhook notifications, but both UI entry points present it like a consequence-free check. Add a confirmation or make the external side effect explicit before dispatch.

products/billing_alerts/frontend/BillingAlertHistory.tsx
```diff
@@ -0,0 +1,172 @@
...
+    events: BillingAlertEventApi[]
+): AlertEvaluationHistoryPoint[] {
+    return events
+        .map((event) => ({ event, value: eventValue(alert, event) }))
+        .filter((item): item is { event: BillingAlertEventApi; value: number } => item.value !== null)
+        .sort((left, right) => left.event.created_at.localeCompare(right.event.created_at))
+        .map(({ event, value }) => ({
+            label: dayjs(event.created_at).format('MMM D, HH:mm'),
+            value,
+            firedAtTime: event.state_after === 'firing',
```
willwearing: > [!NOTE]

**[convergent: qa-team/frontend + qa-team/general correctness and safety]** 🟡 MEDIUM

`firedAtTime` uses `state_after === 'firing'`, so later breached checks while the alert remains firing are counted as triggering the alert. The actual firing edge is `event.kind === 'firing'`; use the event kind for accurate audit history.

products/billing_alerts/frontend/billingAlertFormLogic.ts
```diff
@@ -0,0 +1,266 @@
...
+                            notificationLogic.actions.clearCreatedDestinations(createdDestinationKeys)
+                            try {
+                                actions.editAlert(await billingAlertsRetrieve(organizationId, created.id))
+                            } catch {
+                                actions.editAlert(created)
+                            }
+                            lemonToast.warning('The alert was saved disabled. Add its destination and enable it.')
+                        }
+                    }
+                    const message = error instanceof Error ? error.message : 'Failed to save billing alert.'
```
willwearing: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

Server field-validation errors are reduced to `Error.message`. DRF returns keyed payloads for constraints the form does not mirror, including the 160-character name limit and decimal precision, so users can see a generic transport error. Extract `ApiError.data`, map field errors into the form, and keep a useful fallback.

products/billing_alerts/frontend/billingAlertHistoryLogic.ts
```diff
@@ -0,0 +1,109 @@
...
+                    if (!values.currentOrganization?.id || !values.eventsPage.next) {
+                        return values.eventsPage
+                    }
+                    const nextPage = await billingAlertsEventsList(values.currentOrganization.id, props.alertId, {
+                        limit: 50,
+                        offset: values.eventsPage.results.length,
+                    })
+                    return {
+                        ...nextPage,
+                        results: [...values.eventsPage.results, ...nextPage.results],
```
willwearing: > [!NOTE]

**[qa-team/performance]** 🟢 LOW

`loadMoreEvents` retains every page and the table renders the whole accumulated array. Hourly alerts can produce 720 scheduled events per month before retries or manual checks. Use controlled pagination or retain only the current 50-row page while keeping the chart bounded.

products/billing_alerts/frontend/BillingAlertEditor.tsx
```diff
@@ -0,0 +1,232 @@
...
+                }
+                contentClassName="space-y-6"
+            >
+                <div className="space-y-6 max-w-2xl" data-attr="billing-alert-shared-editor">
+                    <AlertEditorFormDetails
+                        enabled={{ checked: alertForm.enabled, dataAttr: 'billing-alert-enabled' }}
+                        nameDataAttr="billing-alert-name"
+                    />
+                    <LemonField name="description" label="Description">
+                        <LemonTextArea placeholder="What should your team do when this fires?" />
```
willwearing: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

The placeholder asks what the team should do when the alert fires, but the description is internal and omitted from notification payloads. Describe it as an internal note or include it in the notification contract.

products/billing_alerts/frontend/billingAlertDisplay.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+    const value = formatBillingValue(alert.threshold_value, alert.metric, alert.currency)
+    return alert.threshold_type === 'absolute_increase' ? `${value} above baseline` : `Above ${value}`
+}
+
+export function stateLabel(alert: BillingAlertConfigurationApi): string {
+    if (!alert.enabled && alert.state !== 'broken') {
+        return 'Paused'
+    }
+    return alert.state.replaceAll('_', ' ')
```
willwearing, line 49: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

The UI calls this state “Broken,” while notification copy calls it “auto-disabled.” Use “Auto-disabled” consistently so users can connect the status with the notification.

products/billing_alerts/backend/alert_destinations.py
```diff
@@ -24,7 +24,7 @@
     "relative_delta_percentage": "{event.properties.relative_delta_percentage}",
     "evaluation_date": "{event.properties.evaluation_date}",
     "reason": "{event.properties.reason}",
-    "alert_url": "{project.url}/organization/billing",
+    "alert_url": "{project.url}/organization/billing/alerts",
```
willwearing: > [!NOTE]

**[qa-team/general correctness and safety]** 🟡 MEDIUM

`{project.url}` already contains `/project/<team_id>`, so appending `/organization/billing/alerts` produces a project-prefixed URL that misses the origin-rooted billing scene. Build an organization-rooted URL for `alert_url` and each `primary_action_url`.

willwearing:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists + generalists), paul-reviewer, xp-reviewer

The UI mostly preserves the shared alert-management seams, but 2 high-risk ownership boundaries need decisions before merge: generic HogFunction editing crosses the billing permission/project boundary, and the browser owns a partial multi-request save transaction.

### Key findings

- 🟠 HIGH: Protect billing-owned destination management and scope it to the execution project.
- 🟠 HIGH: Move alert plus destination save orchestration behind one backend command, or explicitly model partial success.
- 🟡 MEDIUM: Fix the broken organization billing link, destination-only revision bump, feature-flag route race, history classification, server error surfacing, Teams validation drift, and Check now side-effect disclosure.
- 🟢 LOW/NIT: Bound loaded history rows and align billing description/state copy.

### Convergence

- Destination editing risk was independently flagged by qa-team security, a qa-team generalist, and Paul.
- Save orchestration risk was independently flagged by XP and a qa-team generalist.
- Historical firing classification was independently flagged by qa-team frontend and a qa-team generalist.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | High risk from the generic destination permission boundary, with 4 medium correctness issues. |
| 👤 paul | Shared ownership split looks sound; 3 UI safety edges remain. |
| 📐 xp | Client-side save orchestration and duplicated policy create source-of-truth drift. |

*Automated by QA Swarm, not a human review*

products/billing_alerts/frontend/billingAlertHistoryLogic.ts
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Out of scope for this pass. Fetches remain bounded to 50 rows and loading older history is explicit; controlled pagination can follow if real usage shows the editor becoming heavy.

products/billing_alerts/frontend/BillingAlertEditor.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Non-blocking copy follow-up. Notification content remains backend-owned, so this pass keeps the product form wiring unchanged.

products/billing_alerts/frontend/billingAlertDisplay.ts
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Non-blocking copy follow-up. This pass keeps the existing lifecycle terminology and avoids changing user-facing semantics without product confirmation.

products/billing_alerts/frontend/BillingAlertEditor.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `c93af7b99a0`: Both Check now entry points dispatch the same centralized action, which now opens a confirmation that explicitly names the possible Slack, Microsoft Teams, and webhook side effects before live evaluation runs.

products/billing_alerts/frontend/billingAlertHistoryLogic.ts
```diff
@@ -0,0 +1,109 @@
...
+    path(['products', 'billingAlerts', 'frontend', 'billingAlertHistoryLogic']),
+    props({} as BillingAlertHistoryLogicProps),
+    key(({ alertId }) => alertId),
+    connect({ values: [billingLogic, ['currentOrganization']] }),
+    loaders(({ props, values }) => ({
+        eventsPage: [
+            EMPTY_PAGE,
+            {
+                loadEvents: async () => {
+                    if (!values.currentOrganization?.id) {
```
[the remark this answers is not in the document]
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `c93af7b99a0`: History logic now listens for organization changes, clears the old page immediately, and reloads for the new organization. Loader breakpoints prevent a late response from the previous organization from repopulating the cleared state.

products/billing_alerts/frontend/billingAlertHistoryLogic.ts
```diff
@@ -0,0 +1,109 @@
...
+                        return values.eventsPage
+                    }
+                    const nextPage = await billingAlertsEventsList(values.currentOrganization.id, props.alertId, {
+                        limit: 50,
+                        offset: values.eventsPage.results.length,
+                    })
+                    return {
+                        ...nextPage,
+                        results: [...values.eventsPage.results, ...nextPage.results],
+                    }
```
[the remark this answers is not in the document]
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `c93af7b99a0`: History now follows server-provided offsets and renders one controlled 50-row page instead of accumulating offset pages. Alert-list accumulation also deduplicates by ID. A refresh recovers the narrow concurrent-deletion skip case; a cursor API is outside this management-view stack.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Addressed all 5 findings in `c93af7b99a0` on top of backend `dc4eee227e5`:

1. Alert configuration and destination creation now use one atomic `destination_changes` write handled by the shared alerting backend.
2. Organization changes clear alert rows, editor state, and in-flight UI state before reloading.
3. History clears and reloads on organization changes, with breakpoints blocking late responses.
4. Alert pagination follows server-provided offsets and deduplicates accumulated rows by ID.
5. History uses server-provided offsets with controlled 50-row pages. Refresh handles the narrow concurrent-deletion skip case; a cursor API remains outside this management-view stack.

products/billing_alerts/frontend/BillingAlertNotifications.tsx
willwearing: Fixed. The generic HogFunction editor action was removed. Billing destinations are now managed only through the organization-admin billing endpoint, scoped to the alert execution project and complete lifecycle group.

products/billing_alerts/frontend/billingAlertFormLogic.ts
willwearing: Fixed. Alert fields and destination_changes now go through one backend serializer transaction. Destination creation rolls back with the alert update on failure, and destination-only changes no longer require a separate browser request.

products/billing_alerts/frontend/billingAlertNotificationLogic.ts
willwearing: Fixed. The UI now validates only HTTPS completeness. The backend remains authoritative for the Teams hostname and port policy.

products/billing_alerts/frontend/BillingAlertHistory.tsx
willwearing: Fixed. The current-configuration marker is now omitted when the event revision differs from the current alert revision, because the historical baseline cannot be reconstructed safely.

frontend/src/scenes/billing/BillingSection.tsx
willwearing: Fixed. The route remains pending until feature flags have loaded, and redirects only after a confirmed disabled flag result.

products/billing_alerts/frontend/BillingAlertHistory.tsx
willwearing: Fixed. History now classifies a firing edge from event.kind === firing, with regression coverage for later checks that remain in the firing state.

products/billing_alerts/frontend/billingAlertFormLogic.ts
willwearing: Fixed. DRF field-keyed ApiError.data is mapped into the form, with a useful toast fallback for nested and non-field responses. Focused tests cover the mapping.

products/billing_alerts/backend/alert_destinations.py
willwearing: Fixed in the backend notification properties. Alert and primary action URLs now use the origin-rooted /organization/billing/alerts path, with regression coverage.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd**; not written by a human

Restacked this PR onto backend fix `a7152aa09a7`. UI code is unchanged; the new head is `d9daf564804`. The prior Core 19/19 failure was the inherited backend startup-budget regression and is fixed at the parent.

willwearing:
PR Shepherd update: restacked this PR onto the MCP-enabled backend head `b44dc60024d`.

The UI patch is unchanged apart from preserving the parent README context. All 20 focused Billing UI tests pass. The 8-shard local visual sweep also passed on both the final UI stack and its parent baseline, including the 24 BillingProduct snapshots.

No Stamp Hog was applied.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd**, not written by a human

Restacked this PR exactly onto backend head `7f05b07093d` after the MCP CI fix. `git range-diff` confirms all 5 UI commits are patch-equivalent. Current UI head: `ef3cc75c6a7`.

willwearing:
@pawel-cebula I’ve brought the full billing-alert stack up to date with current master and restarted CI. The stack is #66353, #66355, and #72260. Could you take a look when you get a chance?

products/billing_alerts/frontend/BillingAlertEditor.tsx
```diff
@@ -0,0 +1,212 @@
...
+                            <LemonField name="minimumValue" label="Minimum current value">
+                                <LemonInput
+                                    type="number"
+                                    min={0}
+                                    value={alertForm.minimumValue}
+                                    onChange={(value) => setAlertFormValue('minimumValue', value ?? 0)}
+                                    prefix={<span>$</span>}
+                                />
+                            </LemonField>
+                            <LemonField name="evaluationDelayHours" label="Evaluation delay">
```
pawel-cebula, line 161: FYI, I think this is fine to keep configurable. If these alerts keep using the spend endpoint, though, the data is based on `usage_sent_to_stripe` from usage reports, which update daily and usually settle around 8-10am UTC. So maybe the schedule/time copy should mention that the customer can choose when the alert runs, but the underlying billing data may not have updated yet.

pawel-cebula (commented):
Left one inline comment on the management/UI follow-up.

Also, one thing I wanted to check before calling the stack done: have you run a proper end-to-end pass against PostHog and Billing together? I mean creating the alert in the UI, seeding realistic Billing data, triggering the evaluation path/Temporal run, and checking the resulting state/history/notification behavior. This might be a good one to ask an agent to run with the `qa-frontend` skill in `posthog` plus the `billing-testing` skill in `billing`. The Billing skill has guidance for making local usage/spend data realistic through usage reports, which should make the test much closer to the real product flow.

One backend thing I could not anchor cleanly as an inline on this diff: the zero-destination path now treats the shared internal event as delivered with `destination_ids=[]`, which lets `commit_billing_alert_check()` update `last_notified_at`. That starts the notification cooldown even though no destination actually received anything. If a user creates/adds a destination shortly after, the alert can remain breached but suppress the next firing notification until cooldown expires. Could this only update `last_notified_at` when at least one destination target was delivered, or track internal-event delivery separately from user notification delivery?

products/billing_alerts/frontend/BillingAlertEditor.tsx
willwearing: Completed changes - screenshot attached as evidence.

The definition section now says: "You can choose when the daily check runs, but the underlying billing data updates once a day and usually settles around 8-10am UTC."

editor with schedule copy

willwearing:
Completed changes for the review feedback - screenshots attached as evidence.

**Zero-destination cooldown**: fixed. `commit_billing_alert_check` now only advances `last_notified_at` when at least one destination target was actually delivered. A firing with no destinations still records the event and internal delivery, but no longer starts the notification cooldown, so a destination added right after is not suppressed.

**End-to-end pass**: done against local PostHog + local Billing with realistic seeded spend data (subscription, usage reports, stripe-mock). Created the alert in the UI, triggered evaluation, and checked state/history/notification behavior:

- Scheduled-evaluation auth path returns 200 with the new `billing_alerts_evaluation` service claim ([PostHog/billing#2156](https://github.com/PostHog/billing/pull/2156)).
- Projected period total came back at \$280 after discount, breached the \$200 threshold, alert moved not_firing → firing, and dispatched to the webhook destination.
- Paused preview evaluated the same data with no event, no state change, and no dispatch.

alert list firing

paused preview editor

products/billing_alerts/frontend/BillingAlertEditor.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Verified against `2b544f5`: editor definition copy reads "You can choose when the daily check runs, but the underlying billing data updates once a day and usually settles around 8-10am UTC." Resolving.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

## Ready for another look

@pawel-cebula your schedule-copy thread is addressed and resolved. The definition section now says the underlying billing data updates once a day and usually settles around 8-10am UTC.

Since then I ran our review agents over the UI and fixed three things (`977068e`):

- Clearing the threshold input coerced it to `0`, so an empty field saved a valid `$0` alert that fires immediately. The form value is now nullable and validation rejects it. `minimumValue`, `evaluationDelayHours` and `cooldownHours` still coerce to `0`, because zero is meaningful for those and a `$0` threshold is not.
- `thresholdLabel` was built from the unguarded raw value while `thresholdValue` was null-checked, so an unparsable threshold rendered `NaN`. Both now derive from the guarded value.
- Organization-scoped alert rows and editor state are reset when the selected organization changes, so one organization's alerts cannot appear in another's Billing view.

**Verification:** 18 Jest tests pass across 4 suites, TypeScript check clean for `billing_alerts` (the 293 remaining repo errors are pre-existing `@posthog/quill` workspace errors on this branch, unrelated to this PR), oxlint and oxfmt clean.

**Two things need a human:**

1. The visual review run has 2 genuine snapshot diffs on `products-alerts-shared-components--evaluation-history` (light and dark). They are the intended result of the story changes in this PR: a data point now sets `wouldFireUnderCurrentConfiguration`, and the chart formats the y-axis as currency. They need approving in the Visual Review UI, which I cannot do.
2. A third diff, `components-search--default--light`, is not from this PR. Its baseline is from a master run on 2026-07-03 and the difference is a mid-animation typewriter placeholder plus later master changes to the search footer and BETA tags.

**CI:** the backend-side failures here are inherited from the parent #66355 and are a stale-base problem, not this code. See the note on that PR.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

## Visual review: what to approve and why

The stack is now up to date with master, so a fresh Visual Review run will attach to this PR's new head. Use the newest **"Visual changes detected"** comment for the live link, since earlier links point at superseded runs.

Two snapshots are expected to differ, and both are intentional:

| Snapshot | Why it changed |
| --- | --- |
| `products-alerts-shared-components--evaluation-history--light` | The story now sets `wouldFireUnderCurrentConfiguration` on one data point, which moves the line and recolours that point. The chart also formats the y-axis as currency, so ticks read `$68.00` rather than `68`. |
| `products-alerts-shared-components--evaluation-history--dark` | Same two changes, dark theme. |

Both come from `products/alerts/frontend/components/AlertComponents.stories.tsx` in this PR. The diffs are confined to the chart plot area.

If a third diff appears on `components-search--default--light`, it is not from this PR. On the previous run its baseline was a master run from 2026-07-03, and the delta was a mid-animation typewriter placeholder plus later master changes to the search footer and BETA tags. The master merge should refresh that baseline; if it still shows, it is unrelated to billing alerts.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

## Visual review confirmed: 2 diffs, both intentional

The fresh run for the current head is [`8ab6cd69`](https://us.posthog.com/project/2/visual_review/runs/8ab6cd69-1afd-4e1a-96a8-fe5e322955b3). I inspected it through the Visual Review API: **4,270 snapshots, exactly 2 changed, both pending review.**

| Snapshot | Diff | Why |
| --- | --- | --- |
| `products-alerts-shared-components--evaluation-history--light` | 0.4567% | Story sets `wouldFireUnderCurrentConfiguration` on one point; y-axis now currency-formatted |
| `products-alerts-shared-components--evaluation-history--dark` | 0.5073% | Same, dark theme |

The previously stale `components-search--default--light` diff is gone; the master merge refreshed its baseline. Approving these two clears `PostHog Visual Review / storybook`, `Complete Visual Review run`, and the required `Visual regression tests pass`.

pawel-cebula (approved):
Thanks for the last changes, looks good! Approving with a couple suggestions:

1. The UI and MCP tools are behind `billing-alerts`, but the backend org-admin API route is not gated by that flag. That is probably fine if the flag is only meant to hide the surface during rollout, but if we want the flag to be a hard server-side rollout boundary too, it would be worth adding that guard.
2. Given how much changed across the stack, I would also do one more end-to-end QA pass before shipping. Ideally: create an alert in the UI, seed realistic Billing data, trigger the evaluation path, and check state/history/notification behavior. This seems like a good one to ask an agent to run with the PostHog frontend QA skill plus the `billing` repo billing-testing guidance.

willwearing:
Thanks Pawel. Both handled in f485ad4d.

**1. Server-side flag boundary.** `BillingAlertViewSet` now adds `PostHogFeatureFlagPermission` with `posthog_feature_flag = "billing-alerts"` alongside the existing `OrganizationAdminReadPermissions`. Flag off returns 403, even for an org admin, so the flag is now a hard boundary and not just a UI toggle. This matches the pattern used by metrics, tracing, logs, and user_interviews. Enabled the flag in the API test setups and added `test_api_is_gated_behind_billing_alerts_flag`.

I left the Temporal evaluation sweep ungated on purpose. Alerts can only be created through the now-gated API, so an org without the flag can't accumulate any, and the only exposure is a post-rollback edge. Gating the sweep would mean a fail-closed flag check silencing all billing alert evaluation during a flag-service outage, which for an alerting product is the worst failure mode. Happy to add a fail-open check there if you'd prefer the boundary to be complete.

**2. E2E QA pass.** Ran end-to-end locally against the real HTTP API and the real Temporal activity body, with a seeded billing payload at the `BillingManager` seam (local billing service isn't running). Confirmed: create alert + webhook destination over HTTP, under-threshold records a check with no notification, over-threshold fires and notifies, cooldown suppresses re-notification, spend drop resolves and sends recovery, `events/` history is complete and newest-first, `check_now` on a paused alert previews without recording or notifying, `projected_spend` reads the projected field, and cross-org access 404s.

84/84 backend tests and 18/18 frontend tests pass; ruff, repo-wide mypy, and OpenAPI drift all clean.

willwearing:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd**. This was not written by a human.

/trunk merge

PR #72266, opened 2026-07-20
feat(llm-prompts): add prompt labels UI

jurajmajerik:
## Problem

Labels for prompts already work through the API, SDKs, and MCP tools, but there is no way to see or manage them in the app.

## Changes

- Version history sidebar: each version shows its labels as chips, and gets an "Add label" picker. The picker lists existing labels with the version they currently point at, and suggests `production` when the prompt has none.
- Creating, moving, and removing a label each ask for confirmation. The move dialog shows `production: v1 → v3` with a **Move** button.
- Prompts list: a new Labels column showing `name: vN` per prompt.
- Everything is behind the `prompt-labels` feature flag.
- Two backend additions so the UI has complete data: the resolve response now includes all labels of the prompt (the per-version field only covers the loaded page of versions), and the list response includes `all_labels` per prompt (the existing field only covers the latest version). The list data is fetched with one query per page.

| Sidebar | Move confirmation | List |
| --- | --- | --- |
| chip-v2 | pr-move-dialog | chip-list-v2 |

**Design notes**

- Creating a label also asks for confirmation, because code may already fetch by that label name. Those fetchers currently get a 404 and use their fallback, so creating the label changes what they receive.
- Dialogs close after a failed request and the error shows as a toast. kea reports an exception thrown in a listener as an unhandled rejection, so the dialog cannot stay open by rejecting.
- Labels use a small custom chip component instead of `LemonTag`, whose close button does not fit small text well.
- The chip row sits inside a link card, so it blocks clicks at the capture phase. Without this, removing a label would also navigate to that version.

- Backend, 2 new tests: resolve returns labels that point at versions outside the loaded page, and the list includes labels on non-latest versions using a fixed number of queries. `pytest posthog/api/test/test_llm_prompt.py`: 109 passed.
- Frontend, 4 logic tests: a move does not call the API before confirmation and ends with one entry per label name; a 409 leaves labels unchanged and reloads; a failed delete keeps the label visible; name validation with one case per rule. Jest on the prompts folder: 32 passed.
- A scripted browser run against the local stack covered the full flow: create with confirmation, move, remove, validation, list column, and the flag hiding everything when off. Manually tested as well.
- Not covered: picker open/close state, dialog and toast text.

## Automatic notifications

- [ ] Publish to changelog?

(The flag stays off after merge. Changelog and announcement come with the launch email later.)

## Docs update

Docs live in posthog.com#18628 and will get UI screenshots before the flag flips. The `skip-inkeep-docs` label is set because of that.

## 🤖 Agent context

Built with Claude Code, directed by @jurajmajerik. Skills invoked: `/improving-drf-endpoints`, `/writing-kea-logics`, `/adopting-generated-api-types`, `/writing-tests`, plus a scripted Playwright pass against the local stack for verification. Notable decisions along the way: the backend read additions were folded in here after finding the existing per-version and per-row label fields can't answer "where does this label sit" correctly (pagination and latest-row scoping); creation originally skipped confirmation and was changed after working through the code-first adoption story; the label chip started as small closable `LemonTag` and was replaced with a local component after the close button proved disproportionate, following quill Chip composition principles so it can later be swapped for the real quill component. The browser pass caught one real bug pre-merge: clicking a chip's close button navigated the version card's Link via native anchor default.


frontend/src/lib/constants.tsx
```diff
@@ -348,6 +348,7 @@ export const FEATURE_FLAGS = {
     LLM_ANALYTICS_TRACE_NAVIGATION: 'llm-analytics-trace-navigation', // owner: #team-ai-observability
     LLM_ANALYTICS_USER_FEEDBACK: 'llm-analytics-user-feedback', // owner: @adboio #team-surveys
     LLM_OBSERVABILITY_TRACE_SEARCH: 'llm-observability-trace-search', // owner: #team-ai-observability
+    LLM_PROMPT_LABELS: 'prompt-labels', // owner: #team-ai-observability
```
Radu-Raicea: Want to add your name handle too?

products/ai_observability/frontend/prompts/promptSceneComponents.tsx
```diff
@@ -861,6 +867,40 @@ export function PromptVersionSidebar({
...
+                                    >
+                                        {(labelsByVersion[versionPrompt.version] ?? []).map((label) => (
+                                            <PromptLabelChip
+                                                key={label.name}
+                                                label={label.name}
+                                                onRemove={readOnly ? undefined : () => requestRemoveLabel(label.name)}
+                                                data-attr={`llma-prompt-label-${label.name}`}
+                                            />
+                                        ))}
+                                        {!readOnly &&
```
Radu-Raicea, line 895: Could we gate these label actions with the same Editor access check as New version, Duplicate, and Archive? Right now view-only users see Add/remove controls, go through confirmation, and only then get an API error. The UI shouldn’t offer an action they can’t perform.

products/ai_observability/frontend/prompts/PromptLabelChip.tsx
```diff
@@ -0,0 +1,40 @@
...
+    onRemove,
+    'data-attr': dataAttr,
+}: {
+    label: string
+    onRemove?: () => void
+    'data-attr'?: string
+}): JSX.Element {
+    return (
+        <span
+            className={`inline-flex h-5 items-center gap-1 whitespace-nowrap rounded-[var(--radius)] border border-[color-mix(in_srgb,var(--purple)_35%,transparent)] bg-[color-mix(in_srgb,var(--purple)_10%,transparent)] pl-2 text-[11px] font-medium text-purple ${
```
Radu-Raicea, line 22: Could we truncate long label names here and show the full name on hover? Labels can be up to 128 characters, so a valid long label can overflow the sidebar or table and push the remove button out of reach.

frontend/src/types.ts
```diff
@@ -7766,6 +7770,8 @@ export interface LLMPrompt {
     latest_version: number
     version_count: number
     first_version_created_at: string
+    /** All labels on the prompt with the version each points to. Only present on list responses. */
+    all_labels?: LLMPromptLabelSummaryApi[]
```
Radu-Raicea: Could we use the generated prompt response types here instead of adding fields to handwritten copies? The same API shape now lives in two places, and the optional fields already differ, so they can drift over time.

Radu-Raicea (approved):

PR #72279, opened 2026-07-20
feat(tasks): enable tasks-create MCP tool

andrewm4894:
## Problem

We have read-side Tasks MCP tools (`tasks-list`, `tasks-retrieve`, `tasks-runs-*`) but no way to *create* a task over MCP. Today the flow to kick a task off from a coding agent is manual: grab a link, open the Tasks UI, type into discuss, hit enter. The ask (raised in a Slack thread, linked below) is to drive tasks from the agent itself — a controller session that reviews the top inbox reports and then spins up a task per report — starting with the obvious first primitive, `task-create`.

## Changes

Enables the already-scaffolded `tasks-create` tool in `products/tasks/mcp/tools.yaml` (it was `enabled: false`). Config choices:

- **Scope:** `task:write`. **Annotations:** not read-only, not destructive, not idempotent.
- **Input surface** trimmed via `include_params` to `title`, `description`, `repository`, `origin_product` — the fields a caller actually sets. The many warm-reuse / internal fields on the serializer stay hidden.
- **`description` is required** (via `param_overrides`) with an imperative description, since it's the prompt handed to the agent.
- **`enrich_url: '{id}'`** so the result carries a link back to the task, matching the read tools.
- Response filtered to a lightweight field set.

`origin_product` defaults to `user_created` server-side, so simple "kick off a task" calls only need `description` (+ `repository` for code tasks).

The hand-authored change is the `tools.yaml` block. Everything else is a build output:

- `hogli build:openapi` has been run and its output committed — the generated handler (`services/mcp/src/tools/generated/tasks.ts`), Orval body schema (`services/mcp/src/generated/tasks/api.ts`), and tool definitions (`services/mcp/schema/generated-tool-definitions.json`, `tool-definitions-all.json`).
- The generated handler derives its input schema from the Orval body, omitting the internal serializer fields, so param descriptions flow from serializer `help_text` rather than being hand-written.
- Enabling the tool also produces a new tool-schema snapshot (`services/mcp/tests/unit/__snapshots__/tool-schemas/tasks-create.json`), which is committed — the MCP unit-test artifact check fails without it.

- `pnpm --filter=@posthog/mcp run test` — 83 files, 2236 tests passing, snapshots clean.
- Reviewed the generated `tasks-create` schema snapshot: `description` is the only required field, and the exposed surface is exactly `title`, `description`, `repository`, `origin_product`.
- Verified the `tasks-create` block only uses keys allowed by the strict Zod config schema (`services/mcp/scripts/yaml-config-schema.ts`) and satisfies its refinements (no `input_schema` alongside `include_params`/`param_overrides`; `include`-only response; complete `annotations`).
- The config mirrors the shape of the enabled read tools in the same file and the `annotation-create` reference in `docs/published/handbook/engineering/ai/implementing-mcp-tools.md`.

Not yet exercised end to end against a live MCP client — the tool is gated behind the `tasks` feature flag.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change — the MCP tool list is generated.

## 🤖 Agent context

Directed by @andrewm4894 from a Slack thread about controlling PostHog tasks from a coding-agent session. Authored by the PostHog Slack app (Claude).

Decisions along the way: I first explored the archived `posthog/mcp` repo before finding the canonical MCP now lives in this monorepo at `services/mcp` with a YAML-driven codegen pipeline. I chose the idiomatic Orval-derived path (like the existing read tools) over a hand-written `input_schema`, so param descriptions flow from serializer `help_text`.

The original push deliberately left the generated files out, since the cloud task sandbox lacked the backend needed to run `hogli build:openapi`. Codegen has since been run and committed, followed by the missing test snapshot — that snapshot was the only reason CI was red.

No repo skills were invoked (the `/implementing-mcp-tools` skill isn't available in this environment); I followed its handbook doc directly.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09KTRWD3HD/p1784548399748699?thread_ts=1784548399.748699&cid=C09KTRWD3HD)*


skoob13 (approved):
LGTM, but I still think we should filter out the tasks scope in all cloud runs.

PR #72281, opened 2026-07-20
feat(ci): emit backend test-selection duration signals

gantoine:
## Problem

[#71304](https://github.com/PostHog/posthog/pull/71304) shipped the backend `posthog-ci-test-selection` event without a savings dimension, unlike storybook. The stated reason: a draft never runs the full matrix, so there's no per-run baseline.

The real saving on skipped shard *jobs* — going 38→1 Core shards avoids ~37 full runner startups, each several minutes of fixed overhead before a test runs.

## Changes

So this PR emits the raw signals the selector can measure honestly and leaves the minutes model to the dashboard, where per-shard overhead can be calibrated against the `posthog-ci-running-time` job durations we already collect, and tuned without a CI deploy.

The shadow selector now computes, restricted to the Core/POE/Temporal universe a draft can narrow:

- `narrowable_baseline_seconds` — test-execution seconds across that universe. Product/turbo tests run regardless of selection, so they're excluded. That's the honest denominator — about 6.6k of the 29.3k total suite-seconds are `products/*` tests the Django matrices never run.
- `selected_seconds` — what the selective run will spend.
- `skipped_seconds` — the difference.

The `posthog-ci-test-selection` capture event carries `selected_test_seconds`, `skipped_test_seconds`, and `run_poe` / `run_temporal` (which heavy segments the draft still ran, so the dashboard can weight avoided shard-jobs per segment). Mirrored into the `.depot` shadow to keep it apples-to-apples — depot strips the capture job but keeps the telemetry outputs, and the drift check wants both files to move together.

> [!NOTE]
> The dashboard is where minutes get modeled: `avoided_shards × per_shard_overhead + skipped_seconds` weighting, with the overhead constant calibrated from real backend job durations. The CI event stays honest about what it can actually measure. Storybook's own single `estimated_minutes_saved` uses a fixed 3-min-per-shard constant; backend has more segment structure, so pushing the model downstream keeps it tunable.

I (Claude, via Claude Code) did not run these workflows in CI. Locally:

- Two pure-function tests in `tools/test_snob_backend_test_selection_shadow.py`. One locks the segment partition (POE files count toward both core and poe; `posthog/dags/*` and product tests map to no segment). The other locks the anti-inflation property: a huge `products/*` duration does not enter `narrowable_baseline_seconds`. Neither regression was covered before. Full file: 13 passed.
- Ran the selector against the real `.test_durations`: narrowable baseline 21,810s vs 29,268s total (product/ignored tests excluded), and a 2-file selection yields selected 86s / skipped 21,724s as expected.
- `ruff`, `hogli lint:workflows` (6/6), lint-staged `ty check`, both workflow YAMLs parse.

`tools/` is excluded from the repo mypy config, so the standalone `uv run` script isn't type-checked in CI.

## Docs update

CI-internal, no user-facing behavior. `skip-inkeep-docs`.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — @gantoine is the DRI.

Authored by Claude (Claude Code). This started as a single shard-adjusted `estimated_minutes_saved` computed in CI. @gantoine pushed back that `.test_durations` is a sharding heuristic, not a runtime baseline — it omits job overhead, setup, and concurrency — so any minutes number derived from it alone would be misleading. Reworked to emit raw duration signals plus per-segment run flags and move the minutes modeling into the dashboard, calibrated against real job-duration data.

Invoked the `/writing-tests` skill before adding the tests.

### DIFF
--- a/.depot/workflows/ci-backend.yml
+++ b/.depot/workflows/ci-backend.yml
@@ -564,6 +564,8 @@ jobs:
             changed_file_count: ${{ steps.classify.outputs.changed_file_count }}
             selected_test_count: ${{ steps.classify.outputs.selected_test_count }}
             full_run_reasons_count: ${{ steps.classify.outputs.full_run_reasons_count }}
+            selected_test_seconds: ${{ steps.classify.outputs.selected_test_seconds }}
+            skipped_test_seconds: ${{ steps.classify.outputs.skipped_test_seconds }}
         steps:
             - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
               with:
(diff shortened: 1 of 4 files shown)


rnegron:
@gantoine hmm `.test_durations` is mainly just a heuristic for balancing pytest shards; not a real runtime baseline since it omits job overhead, setup, concurrency etc

gantoine:
@rnegron good point, refocused on test execution time only, the dashboard can handle test overhead/concurrency

rnegron (approved):

PR #72292, opened 2026-07-20
chore(dev): pin compose project name so local infra readiness doesn't drift

thiagosalvatore:
## Problem

Running `./bin/start` (`hogli start`) after a Docker restart often hangs with "clickhouse missing / db missing / redis missing", even though `docker ps` shows the containers running. It's a recurring, worktree-specific footgun, not a real infra failure.

Root cause is that the compose project name isn't deterministic. `COMPOSE_PROJECT_NAME=posthog` is set only by direnv (`.envrc`), so the two docker code paths disagree when direnv isn't loaded:

- The up path (`generator.py`) pins `docker compose -p ${COMPOSE_PROJECT_NAME:-posthog}` → always `posthog`.
- The readiness gate (`bin/wait-for-docker`) ran `docker compose ps` with no `-p`, so it derives the project from `COMPOSE_PROJECT_NAME` or, when unset, the worktree folder name (`quito`, `dalat`, …).

So the gate queries the wrong project, finds no rows, and `readiness_status()` returns "missing" until the 300s timeout. A stray `docker compose up` without the var also spawns a second stack named after the folder, which fights over the shared host ports (5432/6379/8123/9000/9092) and leaves clickhouse/kafka stuck in `Created`.

## Changes

- `bin/start`: export `COMPOSE_PROJECT_NAME=posthog` when unset, so every child process (generator, wait-for-docker, any child `docker compose`) agrees regardless of direnv. An explicit override still wins.
- `bin/start`: warn (advisory, never fatal) when an infra stack is running under a different compose project, and print the exact teardown command.
- `bin/wait-for-docker`: pin `-p "${COMPOSE_PROJECT_NAME:-posthog}"` in the `compose()` helper (both branches) so `ps` targets the same project the up path created under.

Not run through a full `./bin/start` in this session (that launches the interactive phrocs TUI). I (Claude) verified the mechanisms deterministically:

- `bash -n` passes on both scripts.
- The pinned `docker compose -p posthog -f docker-compose.dev.yml ps ...` runs cleanly and targets the `posthog` project.
- The foreign-stack warning: ran the detection block against a throwaway container carrying `com.docker.compose.project=scratch` + `com.docker.compose.service=clickhouse` labels — it correctly flagged `scratch` and printed the teardown command, and stayed silent when only `posthog` was present.
- `hogli ci:preflight --strict` is clean (0 failures); it also ran on push via the pre-push hook.

No automated tests added — these are dev-env shell scripts with no unit-test harness, and the change is a behavior-preserving default plus an advisory warning.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

I hit this repeatedly: after killing all containers and restarting, local DBs reported "missing" while `docker ps` showed them up. Claude diagnosed it live on my machine — found two compose stacks running from the same worktree under different project names (`posthog` half-started with clickhouse/kafka stuck in `Created`, plus a stray `dalat` stack holding all the host ports) — then traced it to the `-p` mismatch between `generator.py` and `bin/wait-for-docker` and the direnv-only `COMPOSE_PROJECT_NAME`.

Chosen fix: make the project name deterministic at the `bin/start` entry point (so it no longer depends on direnv being hooked) and align `wait-for-docker` with the up path, rather than moving the var into flox or rewriting the readiness logic. Added the foreign-stack warning to catch the duplicate-stack case that the deterministic default alone can't (a stray manual `docker compose up`). No skills were applicable.


gantoine (approved):
assuming you tested looks good 👍🏼

rnegron (approved):

PR #72294, opened 2026-07-20
fix(onboarding): stop redirect loop after completing onboarding

fercgomes:
## Problem

Reported in #team-growth ([Slack thread](https://posthog.slack.com/archives/C043VJ93L3B/p1784531930175779)): going through onboarding on a fresh account surfaced two bugs.

1. After finishing onboarding, navigating to any product page kept bouncing the user back to the "What do you want to do with PostHog?" product-selection screen, until they ran through onboarding a second time.
2. The post-subscribe celebration screen read "You've unlocked all features for **Product analytics**", implying only that product's features were unlocked (a paid signup unlocks everything) — under a leftover "Select a plan" heading, though there is no plan left to select.

The loop is a race. `completeOnboarding` fires several `complete_product_onboarding` calls concurrently with the completion team PATCH. Those endpoints echo a full team serialized from a snapshot taken when their request started, and since they are `currentTeam` loaders, whichever response lands last replaces `currentTeam` wholesale. A straggling intent response arriving after the PATCH response reverts `completed_snippet_onboarding` and `has_completed_onboarding_for` in kea state, `hasOnboardedAnyProduct` flips back to false, and sceneLogic redirects every product-page navigation into `/onboarding`. I confirmed this ordering in the reporter's session data: the last intent response landed just after the completion PATCH response, and the loop stopped only after a second full completion where the PATCH happened to land last.

**Why:** a new user who just finished onboarding was trapped back in it on every navigation, and the contradictory plans heading added to the confusion.

## Changes

- `teamLogic`: the product-intent loaders (`addProductIntent`, `addProductIntentForCrossSell`, `recordProductIntentOnboardingComplete`) no longer replace `currentTeam` with the server's snapshot. They merge only `product_intents` (the one field those endpoints change) into the latest local team, so a stale echo can't clobber concurrently-committed fields.
- `OnboardingUpgradeStep`: hide the "Select a plan" title whenever the product is subscribed (matching the platform-packages variant), and replace the product-scoped celebration copy with a plan-wide "You're all signed up. You've unlocked all features."

- Added a jest regression test: `recordProductIntentOnboardingComplete` resolving with a stale team (completion fields missing) must not clobber `currentTeam` — it catches exactly the wholesale-replace regression, which no existing test covered. Mutation-checked: reverting the loader to `return result` makes it fail.
- Ran `jest src/scenes/teamLogic.test.ts` (7 passed) and the full `src/scenes/onboarding` suite (261 passed).
- oxlint/oxfmt clean; `tsgo` reports no errors in the changed files (the sandbox has unrelated pre-existing `@posthog/quill` resolution errors).
- I (Claude) could not manually click through onboarding in a browser in this environment.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs describe this internal behavior; nothing to update.

## 🤖 Agent context

Authored by Claude (PostHog Code cloud task) directed by the assignee, from a #team-growth bug thread with two screenshots. Since the thread text wasn't accessible from the sandbox, I (Claude) reconstructed the reporter's onboarding session from PostHog analytics (pageview/step/billing event ordering) to root-cause the redirect loop, then verified the clobber mechanism in `teamLogic` loaders and the `complete_product_onboarding` viewset. Considered fixing by sequencing the intent calls before the PATCH in `completeOnboarding`, but rejected it: ordering fixes only that call site, while merging in the loader kills the clobber for every caller and keeps loading semantics. Skills invoked: /writing-tests.

---

### DIFF
--- a/frontend/src/scenes/onboarding/legacy/billing/OnboardingUpgradeStep.tsx
+++ b/frontend/src/scenes/onboarding/legacy/billing/OnboardingUpgradeStep.tsx
@@ -65,7 +65,7 @@ export const OnboardingUpgradeStep: OnboardingStepComponentType<OnboardingUpgrad
             showContinue={!!product.subscribed && !showPlatformPackages}
         >
             {!product.subscribed && <PlanCards product={product} />}
-            {product.subscribed && !showPlatformPackages && <SubscribedCelebration product={product} />}
+            {product.subscribed && !showPlatformPackages && <SubscribedCelebration />}
             {showPlatformPackages && platformProduct && (
                 <PlatformPackagesUpsell
                     platformProduct={platformProduct}
@@ -80,7 +80,7 @@ export const OnboardingUpgradeStep: OnboardingStepComponentType<OnboardingUpgrad
 }
 OnboardingUpgradeStep.stepKey = OnboardingStepKey.PLANS
 
-const SubscribedCelebration = ({ product }: { product: BillingProductV2Type }): JSX.Element => {
+const SubscribedCelebration = (): JSX.Element => {
     const { trigger, HogfettiComponent } = useHogfetti({ count: 100, duration: 3000 })
 
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


MattBro (approved):
lgtm

PR #72299, opened 2026-07-20
fix(insights): stop result color modal crashing on formula series

xljones:
## Problem

Clicking the colour swatch of a series in a trends results table crashed the whole insight whenever the insight used a **formula**. Formula results carry `action: null` (see `types.ts`), but the "Customize result color" modal passed that null straight into `EntityFilterInfo`, which dereferenced `filter.type` during render and threw. The modal never opened. Removing the formula made it work again.

## Changes

- `ResultCustomizationsModal` (`TrendsInfo`) now renders the series' own label for formula series, and only renders `EntityFilterInfo` when the series has an entity.
- `EntityFilterInfo` is hardened to render nothing for a null/undefined filter, so a missing entity can no longer crash the surrounding scene.

Clicking the colour swatch on a formula series' row in the results table:

**Before** — it threw `TypeError: Cannot read properties of null (reading 'type')` and the modal never opened:

before

**After** — the modal opens and labels the formula series:

after

- Reproduced the crash and verified the fix in Storybook: added a temporary formula-series `InsightsTable` story and clicked the results-table colour swatch. Before the fix it threw `TypeError: Cannot read properties of null (reading 'type')` in `getUnderlyingEntity` → `EntityFilterInfo` (rendered by the modal's `TrendsInfo`); after, the modal opens. The temporary story is not committed.
- Added parameterized cases to `EntityFilterInfo.test.tsx` for null and undefined filters (the exact null-deref regression). No existing case exercised a missing filter.
- Ran that jest suite (12/12 pass), oxlint + oxfmt on the changed files (clean), and `tsgo` (no errors in the changed files; the repo-wide run shows unrelated pre-existing errors in other products from unbuilt quill packages).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change: internal bug fix with no user-facing workflow change.

## 🤖 Agent context

Xander reported the crash with a minified stack and confirmed that removing formulas made the swatch work. I (Claude, via PostHog Code) traced the click path to `ResultCustomizationsModal` → `TrendsInfo` → `EntityFilterInfo` → `getUnderlyingEntity`, which dereferences a null filter, and reproduced it in Storybook. Invoked the `/writing-tests` skill for the regression test.

Decisions: the root cause is that `TrendResult.action` is typed non-null but is `null` for formulas. Correcting that type is the proper long-term fix, but it cascades to many unguarded call sites, so I left it as a follow-up and fixed the reported path directly (show the label) plus a defensive guard at the single `EntityFilterInfo` choke point.

---


thmsobrmlr (approved):
Testing the new reviewhog on this, but from my perspective looks good 👍 
Thanks!

PR #72323, opened 2026-07-20
chore(aio): run the ai ingestion pipeline in local dev on a dedicated topic

carlos-marchal-ph:
## Problem

The dedicated AI ingestion pipeline runs in prod but has no local dev process — `hogli up` never exercises it, and dev capture-ai produced to the shared analytics firehose, so AI events were double-processed locally (duplicate `ai_events` rows) and any dedicated consumer drowned in allowlist DLQ warnings.

## Changes

- `bin/mprocs.yaml`: new `ingestion-ai` proc mirroring the prod `ingestion-ai-main` deployment (`PLUGIN_SERVER_MODE=ingestion-v2`, `INGESTION_PIPELINE=ai`, port 6746, PersonHog bare `host:port`), gated behind the existing `ai_capture` capability so it comes up with capture-ai. Includes inert-until-set `AI_BLOB_*` env pointing at the local objectstorage `ai-blobs` bucket (only meaningful once the offload PR merges; harmless before).
- dedicated `ai_events_ingestion` topic, matching the sibling capture pairs (replay → `session_recording_snapshot_item_events`, logs → `logs_ingestion`): `bin/start-rust-service` sets it for the host-run dev capture-ai; `docker-compose.base.yml` mirrors it for the container path. Caddy already routes only `/i/v0/ai*` to capture-ai, so the topic carries AI-only traffic; redpanda auto-creates it.

Net effect: one processing path per AI event locally (no duplicate rows), silent allowlist, and the dedicated pipeline finally runs under `bin/start`.

- live-verified on a running dev stack: fresh multimodal event → exactly one `ai_events` row; plain analytics events unaffected; `ai_events_ingestion` auto-created; allowlist DLQ warnings stopped.
- no automated tests: config-only change (a YAML-value test would be a change-detector).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Dev-only wiring; no user-facing docs.

## 🤖 Agent context

- built with Claude Code; the topic decision followed a survey of how sibling capture→ingestion pairs wire dev (all use dedicated topics; capture-ai was the one exception, predating its consumer).
- two bugs found en route: PersonHog addr must be bare `host:port` (client prepends the scheme; `http://` doubles → `ENOTFOUND http` crash-loop), and dev capture-ai is a host process configured in `bin/start-rust-service`, not the compose service — both fixed here.


docker-compose.base.yml
```diff
@@ -400,7 +400,7 @@ services:
         restart: on-failure
         environment:
             ADDRESS: '0.0.0.0:3000'
-            KAFKA_TOPIC: 'events_plugin_ingestion'
+            KAFKA_TOPIC: 'ai_events_ingestion'
```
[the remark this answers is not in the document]
carlos-marchal-ph: There is no compose only startup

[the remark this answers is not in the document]
carlos-marchal-ph: Overflow can't trigger in this wiring: capture's overflow_enabled defaults to false (rust/capture/src/config.rs) and neither dev path sets it, and the consumer side defaults to INGESTION_OVERFLOW_MODE=disabled. the events_plugin_ingestion_overflow defaults are binary-wide and predate this PR, and prod overrides topics in deployment config. wiring a dedicated ai overflow topic here would be dead config until overflow is ever enabled locally, so leaving it. valid latent footgun if that day comes though.

bin/mprocs.yaml
```diff
@@ -113,6 +113,27 @@ procs:
...
+        capability: ai_capture
+        ready_pattern: 'All systems go'
+        groups:
+            layer: Ingestion pipeline
+            tech: Node
+        env:
+            INGESTION_CONSUMER_CONSUME_TOPIC: ai_events_ingestion
+            INGESTION_CONSUMER_GROUP_ID: ingestion-ai
+            PERSONHOG_ENABLED: 'true'
+            PERSONHOG_ADDR: '127.0.0.1:50052'
```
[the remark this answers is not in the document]
carlos-marchal-ph: Fixed!

bin/start-rust-service
```diff
@@ -93,6 +93,7 @@ case "$RS_SVC" in
     export RUST_BACKTRACE=1
     export ADDRESS=0.0.0.0:3309 # avoid port conflict in local dev (3308 is used by llm-gateway)
     export KAFKA_HOSTS="$DEFAULT_KAFKA"
+    export KAFKA_TOPIC=ai_events_ingestion
```
rnegron, line 96: this doesn't impact anything else using `start-rust-service`?

docker-compose.base.yml
rnegron, line 403: does this need to be in `kafka-init` in `docker-compose.dev.yml`?

docker-compose.base.yml
```diff
@@ -400,7 +400,7 @@ services:
         restart: on-failure
         environment:
             ADDRESS: '0.0.0.0:3000'
-            KAFKA_TOPIC: 'events_plugin_ingestion'
```
rnegron, line 403: interesting, I didn't know `events_plugin_ingestion` was like a default value.

rnegron (approved):
some q's but approving to unblock

bin/start-rust-service
carlos-marchal-ph: It's in a case statement that should only trigger for the `capture-ai` service, I just copied the pattern that's all over the file

docker-compose.base.yml
carlos-marchal-ph: I'm not sure exactly what you mean. I just put it here because since we're fully separating the AI pipeline out, it makes sense to have as a default on all Docker compose based environments

PR #72339, opened 2026-07-20
fix(ci): cancel superseded backend jobs promptly

rnegron:
## Problem

Superseded Backend CI runs hold the concurrency slot. Job-level `always()` stays true during cancellation, so stale Django and product shards still dispatch and run to completion.

## Changes

- Swap `always()` for `!cancelled()` on matrix prep, test matrices, snapshots, and optional reporting.
- Mirror the four applicable jobs in the Depot shadow.
- Keep `always()` on the required `Django Tests Pass` gate and the deterministic-failure cancel jobs.

Probe workflow pushed to a throwaway branch, then superseded mid-run: [run 29765284128](https://github.com/PostHog/posthog/actions/runs/29765284128). `matrix_build` succeeded before the cancel landed, so `!cancelled()` vs `always()` was the only difference between the two shard jobs.

| Job | Condition | Result |
| --- | --- | --- |
| `shard_always` | `always() && needs.matrix_build.result == 'success'` | `success`, 3 steps. Dispatched and ran to completion after the cancel. |
| `shard_nc` | `!cancelled() && needs.matrix_build.result == 'success'` | `cancelled`, 0 steps. Never dispatched. |

Gate job read: `shard_nc=cancelled`, `shard_always=success`.

Two conclusions:

- `always()` is what kept stale shards alive. Signal traps could not help, because the job never received cancellation.
- `!cancelled()` reports `cancelled`, never `skipped`, so `check_required_result` still fails closed. No green-gate-on-cancel regression.

Also ran `uv run hogli lint:workflows`, actionlint, and `hogli ci:preflight --fix`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs update is needed.

## Follow-up

`calculate-running-time` stops emitting `posthog-ci-running-time` on cancelled runs, so DevEx CI dashboards measuring run volume or mean duration will step-change on merge.

## 🤖 Agent context

Native cancellation-aware conditions were chosen over signal wrappers or an API-polling watchdog, then verified against a live superseded run rather than reasoned about. The required gate deliberately keeps `always()` so it always emits an explicit verdict.

---

### DIFF
--- a/.depot/workflows/ci-backend.yml
+++ b/.depot/workflows/ci-backend.yml
@@ -721,7 +721,7 @@ jobs:
         name: Build product test matrix
         needs: [turbo-discover]
         if: >-
-            always() && needs.turbo-discover.result == 'success' && needs.turbo-discover.outputs.matrix != '[]' && needs.turbo-discover.outputs.matrix != ''
+            !cancelled() && needs.turbo-discover.result == 'success' && needs.turbo-discover.outputs.matrix != '[]' && needs.turbo-discover.outputs.matrix != ''
         runs-on: depot-ubuntu-latest
         timeout-minutes: 5
         outputs:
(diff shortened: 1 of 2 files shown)


gantoine (approved):
lets try it 👍🏼

PR #72348, opened 2026-07-20
fix(billing): invalidate gateway quota cache

jonathanlab:
## Problem

A PostHog Code organization can remain blocked after its spend limit is raised because llm-gateway retains a stale per-team quota entry for up to five minutes.

## Changes

- Delete llm-gateway quota cache entries for every organization team whenever billing quotas are recomputed.
- Cover eviction of the Code quota, AI credit, and Code billing-status entries.

- Added a regression test in `ee/billing/test/test_quota_limiting.py`. It catches a stale gateway quota entry surviving a billing quota recomputation, which would keep a team blocked after a limit increase.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation update is needed for this internal cache invalidation.

## 🤖 Agent context

I used pi to trace the billing and gateway quota paths. I invoked the `/writing-tests` skill before adding the regression test. I chose direct Redis eviction over a new invalidation API because the existing billing and gateway flows already use the same cache infrastructure.

### DIFF
--- a/ee/api/quota_limits.py
+++ b/ee/api/quota_limits.py
@@ -20,7 +20,7 @@ from rest_framework.response import Response
 from posthog.api.routing import TeamAndOrgViewSetMixin
 from posthog.constants import AvailableFeature
 
-from ee.billing.quota_limiting import QuotaLimitingCaches, QuotaResource, is_team_limited
+from ee.billing.quota_limiting import QuotaResource, get_fresh_team_limited_resources
 
 
 class QuotaResourceLimitSerializer(serializers.Serializer):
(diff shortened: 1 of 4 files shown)


adboio:
Pushed 59ff0be reworking the eviction to actually reach the gateway's cache:

- **Right Redis:** the gateway keeps its quota cache in its own Redis (`quota_resolver.py`, `LLM_GATEWAY_REDIS_URL` in the gateway's chart values), so deleting via the default `get_client()` was a no-op wherever the instances differ — which includes prod. Django now gets its own `LLM_GATEWAY_REDIS_URL` setting (defaults to `REDIS_URL` for single-Redis/dev setups) and evicts there. The regression test now seeds both instances and asserts the central one is untouched, so it fails against the old behavior.
- **Best-effort:** the delete is wrapped with `capture_exception` — the gateway TTL already bounds staleness, so a cross-service Redis blip shouldn't fail a billing update that already committed.
- **No re-poisoning:** `quota_limits` now reads limit state from the zsets directly (single pipelined round trip) instead of the 30s per-worker `cache_for` memo. Without this, a gateway refetch right after eviction could land on a worker holding the pre-change state and re-cache it for another 5 minutes. The endpoint is only hit on gateway cache misses, so the memo wasn't buying anything there.

**Deploy prerequisite:** this is inert until `LLM_GATEWAY_REDIS_URL` is set on the Django web + worker deployments in the charts repo (per region), and those pods can reach the gateway's Redis. Worth confirming security-group reachability before relying on it — until then, behavior degrades gracefully to today's TTL expiry.

adboio (approved):

PR #72368, opened 2026-07-20
feat(data-warehouse): expose managed warehouse in SQL editor

EDsCODE:
## Problem

Provisioned managed warehouses are absent from the SQL editor. Using the organization root credential for a direct connection would also let one project read or mutate data owned by another project.

## Changes

- Creates a system-managed direct SQL connection for each project when it joins a managed warehouse.
- Uses a distinct Duckgres project reader and keeps the connection disabled until its idempotent credential handshake succeeds.
- Reconciles existing warehouses plus all current and future tables in the project event, person, data import, team, and modeled-data namespaces. The exposed namespaces mirror the live Duckgres org-team row, and credential setup never rewrites an existing row (it also drives external-writer discovery); the derived naming is used only to create a missing row.
- Coalesces repeated ready-status reconciliation requests for 60 seconds while preserving later live re-introspection.
- Adds a twice-hourly sweep for every active managed project, so new tables appear without visiting the warehouse status page; reconcile skips quietly while a warehouse is provisioning or unreachable.
- Revives soft-deleted schema and table records when a physical table is dropped and recreated with the same name.
- Keeps managed connections and their schemas protected from generic source mutation APIs.
- Tombstones connections on deprovision (with a retrying cleanup task, since deprovision is not re-POSTable) and rejects legacy shared event/person tables that cannot be isolated safely.
- Documents the managed SQL connection and reconciliation behavior.

This depends on PostHog/duckgres#970, PostHog/duckgres#951, and PostHog/duckgres#971.

- `bin/hogli test products/data_warehouse/backend/tests/test_managed_warehouse_connection.py products/data_warehouse/backend/logic/external_data_source/test/test_tasks.py`: 23 passed.
- `bin/hogli test posthog/tasks/test/test_scheduled.py`: 1 passed.
- Tests cover initial and existing warehouse reconciliation, future schema tables, periodic scheduling, request coalescing, dropped/recreated table revival, tenant validation, reserved source identity, lifecycle races, and concurrent credential setup.
- `bin/hogli ci:preflight --fix` completed with no failures after merging current master.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `posthog/dags/README_DUCKLINGS.md` and added `docs/internal/managed-warehouse-sql-editor.md` with project-scoped connection, table reconciliation, and drop/recreate behavior.

## 🤖 Agent context

Codex implemented the requested project isolation and read-only SQL model. The codex-implement, security-audit, django-migrations, and writing-tests skills were invoked. The design keeps enforcement in Duckgres so raw SQL and compiled queries share the same boundary, while PostHog owns the idempotent credential lifecycle and catalog exposure.


fuziontech (approved):

benben (approved):

EDsCODE:
Addressed the review in 71a1635f1bf:

1. **Team-row preservation (was: unconditional rewrite).** `configure_project_reader` now GETs the org's team rows first and never POSTs over an existing row — the derived `team_<id>` / `events_<suffix>` naming only creates a row that doesn't exist yet. A disabled row is refused instead of silently re-enabled. This protects hand-set/break-glass rows (e.g. the dogfood `devex` row) and the discovery targets external writers poll.
2. **Row-driven allowlist.** `reconcile_managed_warehouse_tables` derives allowed namespaces from the live Duckgres row via the new `project_reader_namespaces` (mirroring the server's policy derivation, including default-named legacy overrides like team 2's `posthog.events`), failing closed when no enabled row exists. Introspection still runs as the reader, so this stays defense in depth.
3. **Sweep cost/noise.** Periodic sweep is now twice hourly (each reconcile pins a whole worker pod on the Duckgres side); reconcile skips quietly when the handshake can't run yet or the warehouse is unreachable, instead of raising per sweep during provisioning.
4. **Deprovision cleanup dead end.** Duckgres 409s a re-POSTed `/deprovision`, so a failed local cleanup now enqueues `soft_delete_managed_warehouse_sources_task` (autoretry, exponential backoff); the 500 remains only when even the enqueue fails.
5. Minor: reconcile lock timeout 120s → 600s, `update_managed_warehouse_password` → `update_managed_warehouse_root_password`, comments for the revival queryset asymmetry and the org-level-only lifecycle.

Not changed (confirmed intended): legacy suffix-less teams stay excluded; team 2 flows through the row-driven path since its Duckgres row carries explicit `events`/`persons` overrides (single-team tables, per discussion).

PR #72379, opened 2026-07-20
feat(early-access-features): require description on PostHog's own project

joethreepwood:
## Problem

On PostHog's own dogfooding project (US-cloud project 2), we want every new early access feature to ship with a user-facing description — the text shown to users in the opt-in UI. Creating an early access feature always produces a feature flag (linked or auto-created), so this is enforced at the point of creation. Today the description is optional everywhere, so internal features can be created with no explanation for opted-in users.

This is a targeted, project-scoped rule. Every other project keeps the current optional-description behavior.

This will help make sure the roadmap page, when it launches soon, won't surface a bunch of empty beta listings. 

## Changes

- Backend serializer (`EarlyAccessFeatureSerializerCreateOnly.validate`) rejects a blank/whitespace description when creating an early access feature on `POSTHOG_TEAM_ID`, guarded by `is_cloud()` so the rule never fires in local or test environments (where a team can coincidentally get id 2). The error is keyed to the `description` attr so it surfaces cleanly on the client.
- Frontend form adds a matching inline validation for project 2 so users see the error before submitting. The backend remains authoritative.

No model or migration change — `description` stays optional at the schema level; enforcement is a conditional validation scoped to one project, so the global API contract and generated types are unchanged.

> Note: this branch is stacked on `coming-soon-waitlist-surveys`, so this PR against master also carries those commits until that work lands.

I'm an agent (PostHog Code). I added automated backend tests covering: rejection when the description is omitted/blank/whitespace (400, `attr=description`), success with a real description, no enforcement for other teams, and no enforcement off-cloud.

I could not run the test suite or the frontend typecheck in my environment (no local Postgres/ClickHouse/Docker for `APIBaseTest`, and frontend deps weren't installed). Static checks pass: `ruff check` and `ruff format --check` on the backend files. CI will run the full suite.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Directed by Joe Martin. Built with PostHog Code (Claude Opus 4.8).

Skills invoked: `/improving-drf-endpoints` (DRF serializer change) and `/setup-web-tests` (attempted local test env setup).

Decisions along the way:
- Scoped the rule to the early access feature's own `description` field rather than the underlying feature flag's name/description — the flag has no description input in this flow.
- Gated on the existing `POSTHOG_TEAM_ID` constant plus `is_cloud()` rather than a fresh literal, mirroring the existing internal-team pattern in `posthog/tasks/early_access_feature.py`. The `is_cloud()` guard prevents the rule from breaking unrelated local/test runs where an auto-incremented team lands on id 2.
- Kept it a serializer-level validation with no migration, so the API contract and generated types stay backward compatible.

---


products/early_access_features/frontend/earlyAccessFeatureLogic.ts
```diff
@@ -311,9 +311,12 @@ export const earlyAccessFeatureLogic = kea<earlyAccessFeatureLogicType>([
     forms(({ actions, props, values }) => ({
         earlyAccessFeature: {
             defaults: { ...NEW_EARLY_ACCESS_FEATURE } as NewEarlyAccessFeatureType | EarlyAccessFeatureType,
-            errors: ({ name, payload }) =>
+            errors: ({ name, description, payload }) =>
                 ({
                     name: !name ? 'Feature name must be set' : undefined,
+                    // PostHog's own project (id 2) requires a description — mirrors the backend POSTHOG_TEAM_ID rule.
+                    description:
+                        values.currentTeamId === 2 && !description?.trim() ? 'A description is required' : undefined,
```
haacked: `blocking`: The new `description` check in the forms `errors` function runs on every submit with no cloud guard, unlike the backend rule it mirrors. The backend only requires a description in `EarlyAccessFeatureSerializerCreateOnly.validate`, which runs only on POST (create) and only when `is_cloud()` is true. The frontend check has no `props.id` branch and no cloud check. This makes pre-existing description-less features on team 2 unsavable in the edit UI and blocks creating features on self-hosted instances where the current team is 2, even though the backend would accept both.

Gate the check on `props.id === 'new'` and `values.isCloud`, matching the backend's create-only and cloud-only conditions.

Connect `preflightLogic`'s `isCloud`:

```ts
connect(() => ({
    values: [teamLogic, ['currentTeamId'], earlyAccessFeaturesLogic, ['earlyAccessFeatures'], preflightLogic, ['isCloud']],
})),
```

Then gate the check:

```ts
description:
    props.id === 'new' && values.isCloud && values.currentTeamId === 2 && !description?.trim()
        ? 'A description is required'
        : undefined,
```

products/early_access_features/backend/api.py
```diff
@@ -270,6 +271,10 @@ class Meta:
         read_only_fields = ["id", "feature_flag", "created_at"]
 
     def validate(self, data):
+        # PostHog's own project (US cloud) requires a description on every new early access feature.
```
haacked: `question`: `is_cloud()` returns true for EU, US, DEV, and E2E cloud deployments, not just US cloud as the code comment and PR description say. Team/project ids are allocated independently per region, so on EU cloud (or the hosted DEV/E2E environments), project id 2 is not PostHog's own project, it belongs to some other team that would now be forced to add a description to every early access feature they create.

products/early_access_features/frontend/earlyAccessFeatureLogic.ts
```diff
@@ -311,9 +311,12 @@ export const earlyAccessFeatureLogic = kea<earlyAccessFeatureLogicType>([
     forms(({ actions, props, values }) => ({
         earlyAccessFeature: {
             defaults: { ...NEW_EARLY_ACCESS_FEATURE } as NewEarlyAccessFeatureType | EarlyAccessFeatureType,
-            errors: ({ name, payload }) =>
+            errors: ({ name, description, payload }) =>
```
haacked, line 329: `suggestion`: No test file exists for `earlyAccessFeatureLogic.ts`, so the new description validator (and its create vs. edit behavior) is entirely untested.

Add a test asserting the check blocks a blank description on team 2 for a new feature, allows it for other teams, and (once the create-only fix above lands) does not block editing an existing description-less feature.

haacked (commented):
Nice, focused change. One blocking issue with the frontend validator inline (it diverges from the backend rule it mirrors), plus a scope question.

products/early_access_features/frontend/earlyAccessFeatureLogic.ts
joethreepwood: Fixed in 1912d5f — gated on `props.id === 'new'` and `preflight.region === Region.US`, connected `preflightLogic`. Edits and non-US team-2 are no longer blocked.

products/early_access_features/backend/api.py
joethreepwood: Good catch — switched to `CLOUD_DEPLOYMENT == "US"` in 1912d5f, so EU/DEV/E2E team 2 is unaffected. Added an EU/self-hosted test case.

products/early_access_features/frontend/earlyAccessFeatureLogic.ts
joethreepwood: Added `earlyAccessFeatureLogic.test.ts` in 1912d5f covering create-only, US-only, other teams, and editing an existing description-less feature.

products/early_access_features/frontend/earlyAccessFeatureLogic.test.ts
```diff
@@ -139,4 +144,56 @@ describe('earlyAccessFeatureLogic', () => {
...
+            await expectLogic(logic).toFinishAllListeners()
+        }
+
+        it('requires a description when creating on US-cloud project 2', async () => {
+            await mountWith('new', POSTHOG_TEAM_ID, Region.US)
+            logic.actions.setEarlyAccessFeatureValue('name', 'My feature')
+            logic.actions.setEarlyAccessFeatureValue('description', '   ')
+
+            expect(logic.values.earlyAccessFeatureValidationErrors.description).toEqual('A description is required')
+        })
```
haacked, line 173: `suggestion`: None of the three `toBeUndefined()` cases here actually reach the `!description?.trim()` check with a valid description. Each one trips a different outer guard instead (wrong team, EU region, or an existing id).

A regression that fires the description requirement when a valid description is supplied would block every PostHog employee from creating a feature, and none of these tests would catch it.

Mirror the backend's `test_posthog_team_allows_creation_with_description_on_us_cloud`: mount with `('new', POSTHOG_TEAM_ID, Region.US)`, set a real description, and assert `earlyAccessFeatureValidationErrors.description` is `undefined`.

products/early_access_features/frontend/earlyAccessFeatureLogic.ts
```diff
@@ -311,9 +322,18 @@ export const earlyAccessFeatureLogic = kea<earlyAccessFeatureLogicType>([
...
                 ({
                     name: !name ? 'Feature name must be set' : undefined,
+                    // Mirrors the backend rule: PostHog's own project (US cloud, id 2) requires a description
+                    // on newly created features. Create-only and US-only to match the serializer.
+                    description:
+                        props.id === 'new' &&
+                        values.preflight?.region === Region.US &&
+                        values.currentTeamId === 2 &&
```
haacked: `nit`: `values.currentTeamId === 2` is a bare literal, while the backend guards on the named `POSTHOG_TEAM_ID` constant. The test file already redeclares its own `POSTHOG_TEAM_ID = 2` for the same reason; add the same here so future readers have something to grep for if the id ever changes.

products/early_access_features/backend/api.py
```diff
@@ -270,6 +271,13 @@ class Meta:
         read_only_fields = ["id", "feature_flag", "created_at"]
 
     def validate(self, data):
+        # PostHog's own dogfooding project requires a description on every new early access feature.
+        # Scoped to US cloud specifically: project ids are allocated per region, so id 2 is only
+        # PostHog's own team on US — on EU/DEV/E2E it belongs to an unrelated customer.
+        is_us_cloud = (settings.CLOUD_DEPLOYMENT or "").upper() == "US"
```
haacked, line 405: `nit`: The backend normalizes case with `.upper() == "US"`, but the frontend's `preflight.region === Region.US` check doesn't, and `preflight.region` passes `CLOUD_DEPLOYMENT` through raw. If `CLOUD_DEPLOYMENT` were ever lowercase, the backend would reject on submit but the client-side error would silently stop firing, though it's always uppercase in practice.

haacked (approved):
The fixes in `1912d5f` all check out. A couple minor suggestions inline.

products/early_access_features/frontend/earlyAccessFeatureLogic.test.ts
joethreepwood: Added in 30853eb — a positive-path case that mounts `("new", POSTHOG_TEAM_ID, Region.US)` with a real description and asserts no error. While adding it I found the EU case was silently passing for the wrong reason: `preflightLogic`'s mounted loader fetches the default fixture (region US) and overwrote the test's `loadPreflightSuccess` — fixed by settling loaders before setting test state.

products/early_access_features/frontend/earlyAccessFeatureLogic.ts
joethreepwood: Done in 30853eb — exported `POSTHOG_TEAM_ID` from the logic and the test now imports it instead of redeclaring its own.

products/early_access_features/backend/api.py
joethreepwood: Done in 30853eb — the frontend check now compares `preflight?.region?.toUpperCase()` so it matches the backend's `.upper()` normalization.

PR #72386, opened 2026-07-20
feat(aio): offload multimodal ai event blobs to s3 at ingestion

carlos-marchal-ph:
## Problem

Multimodal AI events carry media as base64 inside their JSON payloads (`$ai_input`, `$ai_output_choices`, ...).

We can't preprocess these payloads before receipt, so we have to deal with them on ingestion. This implements a first approach that uploads them to S3 as part of the ingestion pipeline. It happens exclusively for AI events, in their dedicated pipeline.

We cache on content hash to avoid repeated uploads of the same data, which are common on shorter timeframes since several turns with the same blob might be sent. We use a copy operation to reset the expiration date on files.

A later PR will introduce support for decoding the `phaiblob` URLs in the frontend and displaying the data inline.

## Changes

- new step in the nodejs `ai` ingestion pipeline: detects binary content in the six heavy AI properties, stores each blob content-addressed in S3 (`{team_id}/sha256/{hash}`, decoded bytes), and rewrites payloads string-for-string with versioned `phaiblob://v1/...` pointer URIs. Text stays in ClickHouse.
- two-tier detection: known provider shapes (data-URIs, Anthropic `source`, Gemini `inlineData`, OpenAI `input_audio`) carry their declared mime; any other whole-string canonical base64 above the floor offloads blind as `application/octet-stream`. the blind tier only accepts provably-invertible input (canonical charset, terminal padding, zero discarded tail bits), so a rewrite can always reconstruct the exact original string — false positives are possible but lossless, and render as download links rather than corrupting anything. non-canonical or sub-floor content stays inline.
- blob store: HEAD → PUT on miss / nothing when fresh / self-COPY when older than a day, so a 31-day bucket lifecycle always outlives the 30-day `ai_events` row TTL. Upload is confirmed before the rewritten event is emitted; failures ride the pipeline retry.
- fail-closed: empty `AI_BLOB_S3_BUCKET` or empty `AI_BLOB_OFFLOAD_TEAMS` disables the step entirely — this PR merges dormant, nothing changes until env lands.
- first-class prom metrics (`llma_ai_blob_offload_*`): detection, dedup hit rate, S3 latency/errors, bytes saved, post-extraction text sizes (sizing data for a future size-triggered offload).

- unit + step tests (TDD) for pointer encode/parse, each detector shape, the HEAD/PUT/COPY ladder incl. the age gate, gating, upload-before-emit ordering, and cross-prop dedup — 23 suites / 845 tests green rebased on current master (one conflict adapted: master's new fluent pipeline builder).
- local E2E on the integration branch: SDK image capture → capture-ai → ai pipeline → blob in local object storage with exact decoded size → `ai_events` row with pointer and intact text; dedup re-send produced zero new objects; an ~8.4 MB message consumed by the CH Kafka engine with no engine-side tuning.
- red-first coverage for the blind path (repro: openai audio output), the canonical-base64 core, and the floor/byte-strictness boundaries.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal dogfooding behind a team allowlist — no user-facing docs yet.

## 🤖 Agent context

- built with Claude Code (superpowers: brainstorm → spec → task-per-commit subagent implementation with independent review gates + final whole-branch review). Skills: /writing-tests, ingestion-pipeline-doctor-nodejs, /writing-plans.
- key decisions: binary-presence trigger (size-triggered text offload deferred with seams left); content-addressed per-team keys for dedup; age-gated self-COPY retention; ingestion-side offload so gateway/OTel traffic benefits too; blind canonical-base64 tier over per-provider shape chasing after a real miss (openai audio output), gated on provable invertibility instead of media sniffing.
- review gates caught: an error-counter miscount on the HEAD-miss path, an inert regression test, a forecast metric skipping text-only payloads; a blank-slate review round caught this description going stale.


nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
```diff
@@ -0,0 +1,134 @@
...
+const MIME_FAMILIES = new Set(['image', 'audio', 'video', 'text', 'application'])
+
+function mimeFamily(mime: string): string {
+    const family = mime.split('/')[0]
+    return MIME_FAMILIES.has(family) ? family : 'other'
+}
+
+const MAX_CONCURRENT_UPLOADS = 8
+
+async function mapWithConcurrency<T, R>(items: T[], limit: number, fn: (item: T) => Promise<R>): Promise<R[]> {
```
pl: nit: You could use `p-limit` instead of custom concurrency code.

carlos-marchal-ph:
Won't run locally until #72323 gets merged and folded into here, was hoping this would have happen sooner but there's some issues with CI

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
```diff
@@ -0,0 +1,134 @@
...
+        let savedChars = 0
+        let belowFloorCount = 0
+        let belowFloorBytes = 0
+
+        for (const key of LARGE_AI_PROPERTIES) {
+            const value = properties[key]
+            if (value === undefined || value === null) {
+                continue
+            }
+            const extraction = extractBlobs(value, { minBase64Length: config.minBase64Length })
```
pl: nit: You can just pass the config to avoid copying it.

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        const blobs = [...blobsByHash.values()]
+        // Upload-before-emit: every blob must be confirmed durable before the
+        // rewritten event exists anywhere. A failure rejects the step; the
+        // pipeline's retry option owns transient failures. Concurrency is bounded
+        // so a many-blob event can't monopolize the S3 socket pool, and the
+        // per-request timeout starts when a request runs, not when it's queued.
+        const outcomes = await mapWithConcurrency(blobs, MAX_CONCURRENT_UPLOADS, (blob) =>
+            store.ensureStored(input.team.id, blob)
+        )
```
pl: question: Why upload blobs separately instead of one S3 file and reference them by offsets?

nodejs/src/ingestion/pipelines/ai/metrics.ts
```diff
@@ -59,3 +59,63 @@ export const aiOtelGroupsCounter = new Counter({
...
 })
+
+// The team was renamed from LLMA to AIO: metrics above keep their historical
+// `llma_` prefix (dashboards depend on it); new metrics use `aio_` from now on.
+
+export const aiBlobOffloadS3Duration = new Histogram({
+    name: 'aio_blob_offload_s3_request_duration_seconds',
+    help: 'Latency of S3 requests made by the AI blob offload store',
+    labelNames: ['op'],
+    buckets: [0.01, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5],
```
pl: nit: 5s is not a lot and we only have 8 buckets here, so you could add a couple higher buckets just to not miss anything.

nodejs/src/ingestion/pipelines/ai/pipeline.ts
```diff
@@ -207,6 +213,9 @@ export function createAiIngestionPipeline<
             )
             .pipe(createNormalizeEventStep())
             .pipe(createProcessAiEventStep())
+            .pipe(createOffloadAiBlobsStep(aiBlobStore, aiBlobOffloadConfig), {
+                retry: { tries: 5, sleepMs: 100, name: 'offload_ai_blobs' },
+            })
```
pl: issue: This is gonna block the whole pipeline, as with the current structure events are processed sequentially. I wrote the pipeline this way because the current steps after the last `pipeChunk` one are synchronous. This one is async and can block for a relatively long time.

We need to wrap this part of the pipeline in a `concurrently` block (we could also use `concurrentlyPerGroup`, but I don't think we need to group anything here). 

If we do that we'll have another problem - max concurrency config will be meaningless, as we'll have concurrency at the pipeline level (you can configure the limit) and at the step level, which will be imprecise.

One suggestion to solve this would be to split this into two steps, one that extracts blobs and puts them in the step output, which is per event and can be in this sequential block (as it's not blocking). The second step would be a chunk step that would take a chunk of events (with associated split blobs), do the uploads, merge the properties, and return updated events. You could do more precise concurrency control with this design and `p-limit` (as suggested in another comment).

I'll also prototype another pipeline-level solution that might help.

pl (changes requested):
Reviewed the pipeline changes and a small bit of the blob offload code, leaving some comments for consideration now, but will continue reviewing tomorrow.

nodejs/src/ingestion/pipelines/ai/pipeline.ts
pl: Here's a draft solution I had on my mind: https://github.com/PostHog/posthog/pull/72699

carlos-marchal-ph: Approved the draft PR, much better than what I had here yep. Does it sound good to you if we merge this one first and then the fan out refactor PR on master, to keep them separate in git history? Won't be turning this step on for any team anyway until both are up.

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
carlos-marchal-ph: Yeah, this gets fixed by your PR anyway!

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
carlos-marchal-ph: We ended up discussing this on Slack, pasting my response here just for future reference:

> Yeah that's one of the things we considered! This and pointing directly to the S3 objects backing the Warpstream data as it was ingested. That's still not off the table. It's just that it makes it quite harder to cache blobs by their content hash, since we'd need some sort of index (in the current design the hash is just the ID).

> Our intuition [...] is that they will be heavily cacheable in close time proximity (think several turns of the same convo sending the same 2 images every time).

> Without actual metrics from real events though it's really hard to compare costs for each approach.

nodejs/src/ingestion/pipelines/ai/metrics.ts
carlos-marchal-ph: Fixed!

nodejs/src/ingestion/pipelines/ai/blob-offload/detect.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        expect(result.blobs).toHaveLength(1)
+        expect(result.blobs[0]).toMatchObject({ detector: 'raw_base64', mime: 'application/octet-stream' })
+        expect(result.blobs[0].bytes.equals(AUDIO_BYTES)).toBe(true)
+        const part = (result.value as typeof input)[0].content[0]
+        expect(parseBlobPointer(part.data)?.mime).toBe('application/octet-stream')
+        expect(part.transcript).toBe('The capital of France is Paris.')
+    })
+
+    it('leaves whitespace-wrapped bare base64 inline (blind path is byte-strict)', () => {
+        const wrapped = `${AUDIO_B64.slice(0, 60000)}\n${AUDIO_B64.slice(60001)}`
```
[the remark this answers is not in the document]
carlos-marchal-ph: Added a comment to clarify that we specifically want to test the case where base64 is rendered invalid by **replacing**, not **inserting** whitespace in the middle.

pl (approved):

PR #72393, opened 2026-07-20
feat(ci): lint required-check gate hygiene

rnegron:
## Problem

- A gate whose dependency is `cancelled` passes a bare `== "failure"` test, reads empty outputs, and exits green with zero tests run.
- [Stop required-check gates going green on cancelled builds](https://github.com/PostHog/posthog/pull/68208) missed these: a gate that is mostly correct with one straggler looks compliant when read whole.
- [Stop superseded runs promptly with !cancelled()](https://github.com/PostHog/posthog/pull/73956) makes it live. Workers now stop on supersede, so `cancelled` reaches gates routinely instead of rarely.

## Changes

| | |
|---|---|
| New lint | `WF007-required-check-gates`, judging each `needs.<job>.result` assertion separately (step-wide scan found 1 issue, per-dependency found 5 across 4 gates) |
| Gates fixed | `cd-sandbox-base-image`, `ci-mcp`, `ci-oauth-proxy`, `ci-proto`, `ci-rust` (two dependencies) |
| Backstop | Indirect gates (`env:` loop, shared helper) must carry the allowlist somewhere |
| Docs | Gate-vs-worker split in authoring-ci-workflows: gates keep `always()`, workers get `!cancelled()` |

- Not linted: "gate must `needs` every coverage job". Reporting and coverage jobs are indistinguishable to a linter (flags 8/14 gates), so it is documented instead.

- `test_workflow_lint.py`: 82 passed
- `hogli lint:workflows`: 7/7 green over 105 workflows, zero false positives across live gates
- `hogli ci:preflight --strict`: 0 failures

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

- authoring-ci-workflows skill updated in this PR.

## 🤖 Agent context

- Rule scope was measured against the 14 existing gates before writing. The backstop exists because the strict rule initially false-positived on `ci-backend` and `ci-agents`.
- Claude Code merged master after [PR 73956](https://github.com/PostHog/posthog/pull/73956) landed and reconfirmed WF007 passes with the new `!cancelled()` workers in tree. `ci-mcp` is the gate that PR directly exposes: `handle-snapshots` now reports `cancelled`, which the old `== "failure"` test read as pass.


tools/hogli-commands/hogli_commands/workflow_lint/checks/required_gates.py
```diff
@@ -0,0 +1,109 @@
...
+   the gate itself is what branch protection reads, so it has to report.
+
+2. Allowlist **per dependency** — assert ``success``/``skipped`` and block
+   everything else. One bad dependency is enough, so reading result words over the
+   step as a whole would call a mostly-correct gate clean. That is not
+   hypothetical: four gates cleared ``changes`` with a bare ``== 'failure'`` test
+   and then read ``needs.changes.outputs.*``, which is empty on a cancelled job, so
+   the gate took its "nothing to test" exit and reported green with zero tests run.
+
+Gates are identified by the repo-wide naming convention: a display name ending in
```
rnegron: hmmm hmm hmmmmmm

.github/workflows/ci-rust.yml
```diff
@@ -733,9 +733,10 @@ jobs:
         steps:
             - name: Check all Rust jobs
               run: |
-                  # Fail if change detection or affected computation failed
-                  if [[ "${{ needs.changes.result }}" == "failure" ]]; then
-                    echo "Change detection job failed."
+                  # Block cancelled too: outputs are empty then, so the "no relevant
+                  # changes" exit below would report green.
```
gantoine: For all of these: "Fail is change detection failed, affected computation failed, or job cancelled"

gantoine (changes requested):
great use of linting 👌🏼

.github/workflows/ci-rust.yml
rnegron: thanks, changed!

gantoine (approved):

feliperalmeida (approved):

PR #72395, opened 2026-07-20
feat(aio): pass $ai_span_name and $groups through vercel-ai otel path

Radu-Raicea:
## Problem

On the Vercel AI SDK / OpenTelemetry ingestion path, the `ai.telemetry.metadata.*` bag is stripped after promoting a small known set of keys (distinct id, session id, prompt name/version). That meant users could not set a custom span name or associate their LLM events/traces with group analytics entities over the native AI-SDK path — both are reserved `$ai_*`/PostHog keys that were being dropped alongside arbitrary metadata.

## Changes

- Promote `$ai_span_name` out of the metadata bag. An explicit value here is treated as the most intentional signal, so it overrides the auto-derived span name (OTel span name / `functionId` / tool-call name).
- Promote `$groups` before the generic mapper runs, so the existing `normalizeGroups()` step parses the JSON string and attaches it. On this pipeline group resolution is read-only, so `$groups` can only associate events with group types already registered for the project — it cannot create group types or groups.
- Bound the size of a string-valued `$groups` before parsing (drop + count an oversized payload) so a pathological value can't force `parseJSON` into multi-megabyte input.
- `$ai_trace_name` is intentionally **not** added — it is deprecated in favor of `$ai_span_name`.

Automated Jest tests only (no manual/e2e run by the agent):

- `vercel-ai.test.ts` — new cases: `$groups` metadata is promoted to `$groups` (and the raw metadata key stripped); a metadata `$ai_span_name` overrides the `functionId`-derived name, while an empty value is ignored and the derived name is kept. Catches a regression where the strip runs before promotion, or where the override precedence/empty-guard breaks.
- `attribute-mapping.test.ts` — new case: an oversized string `$groups` is dropped rather than parsed. Catches a regression where the size bound is removed and a huge payload reaches `parseJSON`/downstream.

Ran the two suites (128 tests) — all pass. Formatted + linted the changed files. Full `tsc` shows only pre-existing errors in unrelated modules (unbuilt workspace packages), none in the touched files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated why the Vercel AI SDK OTel path strips reserved `$ai_*` keys and scoped a fix. Along the way we confirmed prompt-name/version pass-through already ships (so that half of the original report was a user-side setup issue, not a code gap), which narrowed this PR to span name and `$groups`. Chose to have explicit metadata `$ai_span_name` override the auto-derived name (rather than fill-only) since the derived name is nearly always present, and deliberately dropped `$ai_trace_name` as deprecated. Assessed `$groups` blast radius: the AI/OTel pipeline uses a read-only group-type manager, so the risk is limited to associating events with pre-existing group types; added a size bound on the parse as the one hardening worth doing.

Invoked the `/writing-tests` skill before adding tests. Authored with the PostHog Slack app agent.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C087XQ7K9K7/p1784568725222059?thread_ts=1784568725.222059&cid=C087XQ7K9K7)*


nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.ts
```diff
@@ -151,6 +151,14 @@ function process(event: PluginEvent, next: () => void): void {
     }
     delete props['ai.response.text']
 
+    // Promote a groups map supplied via telemetry metadata before next(), so the
+    // generic mapper's normalizeGroups() can parse the JSON string and attach it.
+    // Skip if a $groups span attribute was already set directly.
+    const groupsMetadata = props['ai.telemetry.metadata.$groups']
+    if (props['$groups'] === undefined && groupsMetadata !== undefined) {
+        props['$groups'] = groupsMetadata
+    }
```
carlos-marchal-ph, line 160: This is the only metadata promotion without a value guard, the others all go through `isNonEmptyString` or `isPromptVersion`. A number, boolean or array here becomes a `$groups` property that `normalizeGroups()` passes through untouched and uncounted, where the strip loop would otherwise have removed it. An empty string is similar. Should we match here, or is there a reason for it?

nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.ts
```diff
@@ -243,6 +251,14 @@ function process(event: PluginEvent, next: () => void): void {
         props[AI_PROMPT_VERSION_KEY] = promptVersion
     }
 
+    // An explicit span name from telemetry metadata is the most intentional
+    // signal, so let it override the auto-derived name (OTel span name /
+    // functionId / tool-call name).
+    const spanNameOverride = props['ai.telemetry.metadata.$ai_span_name']
+    if (isNonEmptyString(spanNameOverride)) {
+        props['$ai_span_name'] = spanNameOverride
+    }
```
carlos-marchal-ph: The metadata bag lands on every span of the call: the SDK includes `ai.telemetry.metadata.*` in the basic span info on the top-level span, the `doGenerate`/`doStream` spans and `ai.toolCall` spans alike. Overriding unconditionally means one `$ai_span_name` renames every event of the call, clobbering the tool-call names set above and the `$otel_span_name`-derived generation names, so the whole trace tree shows the same label.

This is a pretty big one that should be fixed before shipping.

nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.test.ts
carlos-marchal-ph, line 1: The `props['$groups'] === undefined` guard has no coverage, inverting it fails no test. A case asserting that a directly-set `$groups` span attribute survives when the metadata key is also present would lock the precedence in.

carlos-marchal-ph (approved):

PR #72403, opened 2026-07-20
fix(mcp): guide agents through exec discovery

dmarticus:
## Problem

In CLI mode, scope-gated tools can look like unknown tools when called directly through `exec`.

**Why:** Authorization failures should explain how to restore MCP access instead of looking like an unsupported capability.

## Changes

- Return explicit reconnect or reauthorization guidance when `info`, `schema`, or `call` targets a scope-gated tool.
- Direct unknown tool names back to catalog search.
- Add focused regression coverage for both recovery paths.

- `pnpm --filter @posthog/mcp exec vitest run tests/unit/exec.test.ts` (99 tests passed)
- `pnpm --filter @posthog/mcp typecheck`

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable. This change only improves runtime error guidance.

## 🤖 Agent context

Codex implemented the runtime recovery behavior using the `/implementing-mcp-tools` and `/writing-tests` skills. The broader instruction and documentation changes were removed after the issue could not be reproduced.

---

### DIFF
--- a/services/mcp/src/templates/sections/cli-error-handling.md
+++ b/services/mcp/src/templates/sections/cli-error-handling.md
@@ -1,4 +1,5 @@
 **Handling errors:**
 
 - If a tool call fails, the error includes a suggestion and similar tool names. Read the suggestion before retrying.
-- If a tool name doesn't exist, run `tools` again to find the correct name.
+- If a tool name doesn't exist, run `search <term>` to find the current name before claiming the capability is unavailable.
+- If a tool is hidden by missing scopes, reconnect or reauthorize the MCP connection. Logging in through browser automation does not update MCP permissions.
--- a/services/mcp/src/templates/sections/cli-error-handling.md
+++ b/services/mcp/src/templates/sections/cli-error-handling.md
@@ -1,5 +1,4 @@
 **Handling errors:**
 
 - If a tool call fails, the error includes a suggestion and similar tool names. Read the suggestion before retrying.
-- If a tool name doesn't exist, run `search <term>` to find the current name before claiming the capability is unavailable.
-- If a tool is hidden by missing scopes, reconnect or reauthorize the MCP connection. Logging in through browser automation does not update MCP permissions.
+- If a tool name doesn't exist, run `tools` again to find the correct name.
--- a/services/mcp/src/templates/sections/exec-tool-blurb.md
+++ b/services/mcp/src/templates/sections/exec-tool-blurb.md
@@ -2,11 +2,11 @@
 
 PostHog: dashboards, insights, funnels, SQL, experiments, surveys, replay, error tracking, flags.
 
-Pass CLI-style commands in the `command` parameter for all PostHog interactions.
+All PostHog capabilities use CLI-style commands through `exec`. Before claiming one is unavailable or using browser automation, run `search <term>`. Browser login cannot change MCP permissions.
 
 **Requirements**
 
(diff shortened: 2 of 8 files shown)


services/mcp/src/tools/exec.ts
```diff
@@ -254,15 +254,22 @@ function stripOutputFormatProperty(jsonSchema: Record<string, unknown>): Record<
...
+function findTool(tools: Tool<ZodObjectAny>[], scopeGatedTools: ScopeGatedTool[], name: string): Tool<ZodObjectAny> {
     const tool = tools.find((t) => t.name === name)
     if (!tool) {
         const redirect = DEPRECATED_TOOL_REDIRECTS[name]
         if (redirect) {
             throw new Error(redirect(tools))
         }
-        const available = tools.map((t) => t.name).join(', ')
-        throw new Error(`Unknown tool: "${name}". Available tools: ${available}`)
+        const scopeGatedTool = scopeGatedTools.find((candidate) => candidate.name === name)
```
skoob13, line 264: Totally makes sense.

services/mcp/src/templates/sections/exec-tool-blurb.md
```diff
@@ -2,29 +2,21 @@
...
-1. Find unknown tools with `search` or `tools`.
+1. Find unknown tools with `search` or `tools`. If search reports missing scopes, reconnect or reauthorize the MCP connection. Do not use browser automation as a workaround.
 2. Run `info <tool_name>` once if its schema is not in context. Reuse it unless the tool changes or a schema error occurs.
 
 Never guess a schema or run `info` before every call.
 
 **Commands (in order):**
 
 ```text
-# 1. Find unknown tools
```
skoob13, line 17: These comments were reasoning steps. I'd keep them.

services/mcp/src/templates/sections/cli-error-handling.md
```diff
@@ -1,4 +1,5 @@
 **Handling errors:**
 
 - If a tool call fails, the error includes a suggestion and similar tool names. Read the suggestion before retrying.
-- If a tool name doesn't exist, run `tools` again to find the correct name.
+- If a tool name doesn't exist, run `search <term>` to find the current name before claiming the capability is unavailable.
```
skoob13: The changes to the prompt here are a bit confusing. It feels to me like a side effect rather than the root cause. Specifically for this line, what is the behavior you've observed? Does GPT use "list" before search?

skoob13 (commented):
Trying to understand what happened here, as we're in the the caveman mode for characters in the exec tool. Can you expand a bit on how I can reproduce it?

services/mcp/src/templates/sections/cli-error-handling.md
dkarasiewicz: +1

services/mcp/src/templates/sections/exec-tool-blurb.md
dkarasiewicz: +1

dmarticus:
okay, tried reproducing this in both codex and claude (a friend had reported it to me with codex), but I couldn't reproduce it. I think the changes to services/mcp/src/tools/exec.ts are worth keeping, but I'm not gonna mess with the harness.

Thanks for looking!

skoob13 (commented):
:turbostamp:

skoob13 (approved):

Issue #50084, opened 2026-03-05
fix(flags): Add API validation for feature flag filters structure


patricio-posthog:
### Bug Description

# Problem

The feature flag API endpoints (POST /api/projects/:id/feature_flags/, PATCH /api/projects/:id/feature_flags/:id/) do not validate the structure of the filters field. It's possible to create a feature flag via the API with filters: {} or other malformed filter objects missing required keys like groups.

This was discovered via #50028 where a customer created a flag through the API using an internal tool that omitted groups from the filters. The missing field caused crashes in both the frontend and backend when evaluating or displaying the flag.

#50028 added defensive guards as a stopgap, but the API should reject malformed filters at write time rather than relying on every consumer to handle broken data.

##  Expected behavior

The API should validate that:

- filters contains a groups key (list)
- Each group in groups has the expected structure (e.g. properties, rollout_percentage, etc. have valid types)
- Multivariate flags have a valid multivariate object with variants that sum to 100
- filters.payloads keys correspond to valid variant keys (if multivariate) or "true" (if boolean)

Invalid filters should return a 400 with a clear validation error, not silently persist broken data.

## Context

- Fix PR (stopgap): #50028
- As noted in that PR: "I think we're still missing complete validation of the filters. It's like the wild west in there!"
- This affects anyone creating flags programmatically via the API — not just legacy flags

## Suggested approach

Add validation in the FeatureFlagSerializer (likely in validate_filters or a dedicated method) to enforce the required filter schema on create and update.

### Debug info

```shell

```

haacked:
No, the validations I did were a bit more specific. These probably still need doing

patricio-posthog:
I am working on this. I think the validation should be an invariance instead of just checking format (I think Pydantic is too simple for this). But before starting on the code I want to agree on the invariance we are going to check for. Does this makes sense to you @dmarticus @haacked? I am missing something?

**Property level — invariants based on `type`:**

  | If `type` is... | Then... |
  |---|---|
  | `"cohort"` | `key` must be `"id"`, `value` must be int (cohort ID), `operator` must be `"in"` or `"not_in"` |
  | `"group"` | `group_type_index` must be present (not null) |
  | `"flag"` | `operator` must be `"flag_evaluates_to"` |
  | `"person"` | `group_type_index` must be null/absent |

  **Property level — invariants based on `operator`:**

  | If `operator` is... | Then `value` must be... |
  |---|---|
[...]

PR #72405, opened 2026-07-20
feat(feature_flags): add flag filters validators and audit command

patricio-posthog:
## Problem

Part of #50084 (phase 1 of 3, does not close it). The `filters` JSON on feature flags is validated by ~300 lines of procedural checks with gaps. Malformed data can persist and poison a team's flag cache when the Rust service's serde fails to deserialize it. The issue thread locked a three-tier design (structural / cross-field / contextual) and a rollout order: build the rule set first, audit prod, clean up, then enforce.

## Changes

This phase builds the rule set without changing API behavior. Nothing is wired into `FeatureFlagSerializer` yet, and `validate_filters` is untouched.

- `api/filters_schema.py`: a nested DRF serializer for filters, with strict fields that reject what serde would reject (no bool/int/string coercion, no NaN percentages, i32/i64 bounds on integer fields). Operator aliases are canonicalized and payload values normalized to JSON strings in `to_internal_value`, matching current write-path behavior. Unknown keys are detected at every nesting level (filters, group, property, multivariate, variant, holdout): non-legacy ones get a structlog warning, legacy `holdout_groups`/`super_groups` stay silent.
- `filters_validation.py`: cross-field rules (variant sum, unique variant keys, payload/variant agreement, aggregation vs property types, operator/value compatibility) as pure functions returning violations with stable rule ids, plus a fail-fast wrapper for the later enforcement phase. The POST-only "groups must be non-empty" rule is deliberately kept out of the shared rule set, since `PATCH {"filters": {"groups": []}}` is a legitimate clear-targeting and stored flags with empty groups are valid state (see the issue thread).
- `audit_flag_filters` management command: read-only scan of all flags including soft-deleted ones (via `objects_including_soft_deleted`, whose root-team manager also maps environment team ids to the project root for `--team-id`), reporting violations grouped by rule with sample ids and an unknown-key frequency table. A flag whose stored junk crashes a validator is reported as a `structural.filters.internal_error` violation rather than aborting the scan. Its clean run is the gate for the enforcement phase.
- Cross-reference comments in the Rust flag structs pointing at the new Python mirror, and a stale file path fixed in `property_models.rs`.

> [!NOTE]
> Serde fidelity cuts both ways, so the schema is deliberately lenient where Rust is lenient and strict where strict parsers break:
> - Numeric property keys are accepted and normalized to strings, mirroring Rust's `deserialize_key` (stored numeric flag-dependency keys evaluate fine and must not fail the audit gate). This amends the issue inventory's "non-empty string" entry; non-empty is still enforced after normalization.
> - `payloads: null` is accepted (Rust reads `Option<serde_json::Value>`).
> - NaN/Infinity payload values are rejected — stdlib json tolerates them but they produce non-RFC payload strings that serde_json and SDK `JSON.parse` fail on.

Automated tests only, all run locally and passing (119 across the three new files):

- `test_filters_schema.py` (70 cases, no DB): locks the strict-field contract, so a regression that reintroduces DRF coercion (accepting `"true"` as a bool or `"42"` as an int, which serde then rejects) fails here. Also pins alias canonicalization, unhashable stored operators (`[]`/`{}`) failing cleanly instead of crashing, numeric-key normalization, i32/i64 overflow rejection, payload normalization parity including blank-string and NaN/Infinity rejection, `payloads: null` acceptance, the absent-vs-null `aggregation_group_type_index` distinction, and unknown-key logging vs sink behavior at all six nesting levels.
- `test_filters_validation.py` (38 cases, no DB): one pass/fail pair per cross-field rule, the flag-level aggregation fallback that would otherwise misfire on legacy flags, and a guard asserting the create-only groups rule stays out of `CROSS_FIELD_CHECKS`.
- `test_audit_flag_filters.py` (11 cases, DB): soft-deleted flags are scanned (catches a regression to the default manager), `--team-id` with an environment team id resolves to the project root instead of reporting a false clean, a validator crash on one flag becomes a reported violation instead of a dead scan, empty-groups flags are never flagged (the load-bearing asymmetry assertion), legacy vs non-legacy unknown-key reporting, and the CLI options.

Also smoke-ran `python manage.py audit_flag_filters --limit 100` against a local dev database. I did not run the audit against production; that is phase 2.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior changes in this phase, so no docs update (labeled `skip-inkeep-docs`).

## 🤖 Agent context

Authored by Claude via PostHog Code, directed by the assignee. Skills invoked: `/improving-drf-endpoints` before writing the serializer, `/writing-tests` before the test files. Notable decisions: strict field subclasses instead of stock DRF fields (stock fields coerce, which defeats serde fidelity); the audit collects violations per rule instead of failing fast, and the write-path wrapper reuses the same collectors; one addendum to the issue's locked inventory (`early_exit`, present in the Rust structs but missing from the inventory) is recorded in an issue comment rather than silently absorbed.

A second pass driven by the assignee's internal review tooling hardened the first cut: crash-proofing the audit against unhashable stored junk, numeric property keys accepted per Rust's `deserialize_key`, i32/i64 bounds, unknown-key detection extended to variant/multivariate/holdout levels, `payloads: null` accepted, NaN/Infinity payloads rejected, and the audit's manager swapped to `objects_including_soft_deleted` so `--team-id` resolves environment ids. Deliberately deferred from that review: delegating the live `validate_filters` to the new collectors now — the collectors emit different error messages/codes than the live checks, so swapping would change API error responses, which phase 1 promises not to do; the swap stays in phase 3 per the locked plan.

---


products/feature_flags/backend/filters_validation.py
```diff
@@ -0,0 +1,354 @@
...
+def _variant_keys(filters: Mapping[str, Any]) -> set[str]:
+    return {variant["key"] for variant in (filters.get("multivariate") or {}).get("variants", [])}
+
+
+def check_variant_rollout_sum(filters: Mapping[str, Any]) -> list[Violation]:
+    variant_list = (filters.get("multivariate") or {}).get("variants", [])
+    if not variant_list:
+        return []
+    rollout_sum = sum(variant.get("rollout_percentage", 0) for variant in variant_list)
+    if rollout_sum != 100:
```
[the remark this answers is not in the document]
patricio-posthog: Kept exact equality on purpose: the live write path enforces the same sum != 100 and float summation is deterministic, so any flag the API accepted still sums to exactly 100 here — see the comment added in 29a3b58ad73.

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,204 @@
...
+        scanned = 0
+        flags_with_violations = 0
+
+        # objects_including_soft_deleted: the default manager excludes soft-deleted flags, and
+        # a restored flag must not resurrect a violation. It is also a RootTeamManager, so
+        # --team-id with an environment's team id maps to the project root team instead of
+        # silently scanning zero flags.
+        queryset = FeatureFlag.objects_including_soft_deleted.all()
+        if team_id is not None:
+            queryset = queryset.filter(team_id=team_id)
```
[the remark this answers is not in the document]
patricio-posthog: Fixed in 29a3b58ad73: a nonexistent --team-id now raises CommandError instead of reporting a false clean.

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,204 @@
...
+        rows = queryset.order_by("id").values_list("id", "team_id", "filters")
+        if limit:
+            rows = rows[:limit]
```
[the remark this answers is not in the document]
patricio-posthog: Fixed in 29a3b58ad73: negative --limit values are now rejected with CommandError.

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,204 @@
...
+    help = (
+        "Read-only audit of FeatureFlag.filters (all flags, soft-deleted included) against the "
+        "#50084 structural + cross-field rules. Reports violations grouped by rule; a clean run "
+        "gates flipping enforcement on."
+    )
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("--json", action="store_true", help="Emit a machine-readable JSON report instead")
+        parser.add_argument("--limit", type=int, default=0, help="Max flags to scan (0 = all)")
+        parser.add_argument("--samples", type=int, default=5, help="Sample flag ids to record per rule / unknown key")
```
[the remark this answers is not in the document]
patricio-posthog: Fixed in 29a3b58ad73: negative --samples values are now rejected with CommandError.

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            self.stdout.write(self.style.SUCCESS("No violations found."))
+
+        if sink.flag_counts:
+            self.stdout.write("")
+            self.stdout.write("Unknown keys dropped by enforcement (frequency):")
+            for (level, key), count in sorted(sink.flag_counts.items(), key=lambda item: (-item[1], item[0])):
+                legacy_marker = "  [legacy]" if is_legacy_unknown_key(level, key) else ""
+                ids = ", ".join(str(flag_id) for flag_id in sink.sample_flag_ids.get((level, key), []))
+                self.stdout.write(f"  {level:<9} {key:<40} {count} flags  sample ids: {ids}{legacy_marker}")
```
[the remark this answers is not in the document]
patricio-posthog: Fixed in 29a3b58ad73: console output now strips C0/C1 control characters from unknown keys and sample details, with a regression test.

patricio-posthog:
The PR adds three production files, three test files, and comment-only edits to two Rust files. Here's the tour, following the same path a filters dict takes through the code.

## The big picture first

The problem: a feature flag's `filters` JSON gets stored in Postgres, cached, and then read by a Rust service that deserializes it with serde. Serde is strict — if one flag on a team has `"rollout_percentage": "50"` (a string) instead of `50`, deserialization of the *whole team's* flag cache fails, and every flag evaluation for that team breaks. The issue (#50084) locked a design of three validation tiers: **structural** (is the shape right?), **cross-field** (do the fields agree with each other?), and **contextual** (checks needing the DB, which already exist and stay where they are). This PR builds the first two tiers plus a command to measure prod data against them — without changing any API behavior yet.

---

## `api/filters_schema.py` — the structural tier

This file answers one question: *does this JSON have the shape the Rust structs expect?* It has three sections.

**Section 1: strict field classes.** Django REST Framework's built-in fields are helpfully lenient — `BooleanField` accepts the string `"true"`, `IntegerField` accepts `"42"` and `42.0`. That helpfulness is exactly wrong here, because serde accepts none of those, so a value DRF "fixed" at validation time would still crash Rust at read time. So the file defines its own fields: `StrictCharField`, `StrictBooleanField`, `StrictIntegerField` (each rejects anything that isn't the exact type), and `FiniteFloatField` (also rejects NaN/Infinity, which Python's JSON parser tolerates but which break comparisons). Two special ones:

- `PropertyKeyField` — accepts strings *and numbers*, normalizing numbers to strings. This mirrors a quirk on the Rust side: numeric flag-dependency keys already exist in prod, and Rust has custom code to accept them, so rejecting them here would flag data that works fine.
- `FlagPayloadsField` — payloads are stored as JSON-*encoded strings* (`'{"plan": "pro"}'`, not a nested object), but the API has always accepted raw dicts and primitives too. This field keeps that leniency at the edge and normalizes everything to encoded strings, so all downstream code sees one canonical form. It rejects NaN/Infinity since those produce strings strict parsers can't read back.

**Section 2: unknown-key handling.** When DRF validates a dict, it silently ignores keys that have no declared field — which is the desired end behavior (dropping junk like the legacy `holdout_groups`), but silence is dangerous before we know what's out there. `DropsUnknownKeysMixin` intercepts every dict and reports its unknown keys, in one of two modes: during normal (future) API use it logs a warning; during an audit run, a "sink" object passed through the serializer context collects them for the report instead. Two legacy keys are exempt from logging because they're expected everywhere.

**Section 3: the serializers themselves.** Six nested classes mirroring the six Rust structs one-to-one: `FeatureFlagFiltersSerializer` at the top, containing `FlagConditionGroupSerializer` (a targeting group), which contains `FlagPropertySerializer` (one condition like "email contains @posthog.com"), plus `FlagMultivariateSerializer`/`FlagMultivariateVariantSerializer` (variant definitions) and `FlagHoldoutSerializer`. Notable details: integer fields carry i32/i64 bounds matching Rust's types; the property serializer rewrites the legacy operator aliases `min`→`gte` and `max`→`lte` before validating so everything downstream only sees canonical operators; and the group's `aggregation_group_type_index` deliberately has no default, because "key absent" (inherit the flag-level value) and "key explicitly null" (person aggregation) mean different things.

---

## `filters_validation.py` — the cross-field tier

This file answers the second question: *given a structurally valid filters dict, do the pieces agree with each other?* Its central design choice: every rule is a plain function that takes a dict and **returns a list of violations instead of raising an exception**. That's because the two consumers want opposite things — the audit wants *all* problems in a flag to report them, while the future write path wants to fail on the first one. Collecting satisfies both; a small wrapper (`validate_cross_field_or_raise`) converts violations into a DRF error for the write path later.

A `Violation` is three strings: a stable `rule_id` used to group counts across thousands of flags (like `cross_field.variant_rollout_sum_not_100`), a `path` pointing at the exact spot (`groups[2].properties[0].value`), and a human message.

The six rules, registered in a `CROSS_FIELD_CHECKS` tuple: variant percentages must sum to 100; variant keys must be unique; payload keys must match variant keys (or just `"true"` for boolean flags); a group's variant override must reference a real variant; a group's properties must match its aggregation mode (person-aggregated groups take person/cohort/flag properties, group-aggregated ones take group properties with the matching index — with a subtle fallback where a group missing the key inherits the flag-level value, without which the audit would wrongly flag every legacy flag); and operator/value compatibility (date operators need parseable dates, `regex`/`gt`/etc. need strings, `icontains_multi` needs a list, `semver_*` needs a valid semver, `in`/`not_in` only work on cohorts).

One rule lives *outside* that tuple on purpose: `check_groups_non_empty_for_create`. Creating a flag requires at least one group, but clearing a flag's targeting via PATCH (and stored flags with empty groups) is legitimate — so this rule must never run in the audit. A test enforces that it stays out of the tuple.

The file ends with the two glue functions: `flatten_structural_errors` converts DRF's nested error structure into the same `Violation` shape (stripping list indices from the rule id so `groups[0]` and `groups[7]` failures group together), and `collect_filters_violations` runs the whole pipeline — structure first, cross-field only if structure passes.

---

## `management/commands/audit_flag_filters.py` — the measuring instrument

A read-only Django command that runs the pipeline above over every flag in the database and prints a report. Key decisions, top to bottom:

- **Input validation up front**: negative `--limit`/`--samples` and nonexistent `--team-id` are rejected loudly, because each would otherwise produce a report that *looks* clean while measuring the wrong thing.
- **Which flags**: it iterates via `objects_including_soft_deleted` — the default manager hides soft-deleted flags, but a deleted flag can be restored, so its violations count too. That manager also maps environment team ids to the project root, so `--team-id` can't silently scan zero rows.
- **Crash containment**: each flag is validated inside a try/except; if some pathological stored value crashes a validator, it becomes a reported `internal_error` violation for that flag rather than killing a scan that's minutes into a million rows.
- **Aggregation**: violations group by rule id, counting affected flags and keeping a handful of sample flag ids and details per rule. Unknown keys get their own frequency table (with a `[legacy]` marker and a cap on distinct keys tracked, since key names are user-controlled). Console output strips terminal control characters for the same reason. `--json` emits the machine-readable version whose `clean: true/false` is the gate for the next phase.

---

## The test files

They mirror the three modules. `test_filters_schema.py` (~70 cases, no database) is mostly one big parameterized table of invalid inputs asserting each is rejected with the right error code — the point is pinning the strict-field contract so a future refactor can't quietly reintroduce DRF's coercion. `test_filters_validation.py` (38 cases, no database) has one passing and one failing fixture per cross-field rule, plus the guard test for the POST-only rule. `test_audit_flag_filters.py` (16 cases, real database) creates deliberately broken flags directly through the ORM (bypassing the API, like real bad data) and asserts the report groups them correctly — including the load-bearing negative assertion that empty-groups flags are *never* flagged.

## The Rust edits

Comments only: notes on `FlagFilters` and `PropertyFilter` pointing back at the new Python mirror so someone editing either side knows to keep them in sync, plus fixing a stale file path in an existing comment.

## What deliberately *didn't* change

`FeatureFlagSerializer` and its 385-line `validate_filters` are untouched — the new code isn't wired into any request path. That's the whole point of Phase 1: build and test the ruler, measure prod with it (Phase 2), and only then swap it in (Phase 3).

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+                ]
+            if not violations:
+                continue
+            flags_with_violations += 1
+            counted_rules: set[str] = set()
+            for violation in violations:
+                report = rule_reports.setdefault(violation.rule_id, RuleReport(rule_id=violation.rule_id))
+                report.total_violations += 1
+                if violation.rule_id in counted_rules:
+                    continue
```
haacked, line 173: `suggestion`: No test creates a flag with repeated violations of the same rule, so the per-flag dedup branch at line 152 (`if violation.rule_id in counted_rules: continue`) never runs. That branch keeps `flags_affected` counting a flag once while `total_violations` counts every hit. If it regressed, the "N flags affected" count (which gates whether enforcement is safe to turn on) would double-count flags with repeated violations, and no test would catch it.

Add a case with two violations of the same rule on one flag, mirroring `test_violations_grouped_by_rule_with_samples`.

```python
def test_repeated_rule_counts_flag_once_but_all_violations(self) -> None:
    self._create_flag(
        "double",
        {"groups": [{"properties": [
            {"key": "a", "type": "person", "operator": "in", "value": [1]},
            {"key": "b", "type": "person", "operator": "in", "value": [2]},
        ]}]},
    )
    report = self._run()
    rule = self._rule(report, "cross_field.in_not_in_requires_cohort")
    assert rule is not None
    assert rule["flags_affected"] == 1
    assert rule["total_violations"] == 2
```

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"  {report.rule_id} — {report.flags_affected} flags "
+                        f"({report.total_violations} violations)  sample ids: {ids}"
+                    )
+                )
+                for detail in report.sample_details:
+                    self.stdout.write(f"      {_sanitize_for_console(detail)}")
+        else:
+            self.stdout.write(self.style.SUCCESS("No violations found."))
```
haacked, line 243: `suggestion`: The clean-scan console message ("No violations found.") never runs in any test: the console-mode tests always feed a violating flag, and the clean-scan tests always go through `--json`. Operators need to read this line to confirm the enforcement gate is green, so it's worth locking down.

```python
def test_console_output_reports_clean_scan(self) -> None:
    self._create_flag("clean", {"groups": []})
    out = StringIO()
    call_command("audit_flag_filters", "--team-id", str(self.team.id), stdout=out)
    assert "No violations found." in out.getvalue()
```

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+            self.stdout.write("")
+            self.stdout.write("Unknown keys dropped by enforcement (frequency):")
+            for (level, key), count in sorted(sink.flag_counts.items(), key=lambda item: (-item[1], item[0])):
+                legacy_marker = "  [legacy]" if is_legacy_unknown_key(level, key) else ""
+                ids = ", ".join(str(flag_id) for flag_id in sink.sample_flag_ids.get((level, key), []))
+                self.stdout.write(
+                    f"  {level:<9} {_sanitize_for_console(key):<40} {count} flags  sample ids: {ids}{legacy_marker}"
+                )
+            if sink.untracked_keys:
+                self.stdout.write(
```
haacked, line 255: `suggestion`: `test_unknown_key_tracking_is_capped` only checks the cap through `--json`, so the console warning about keys beyond `MAX_TRACKED_UNKNOWN_KEYS` never appears in a test. Mirror it in console mode.

```python
@patch("products.feature_flags.backend.management.commands.audit_flag_filters.MAX_TRACKED_UNKNOWN_KEYS", 1)
def test_console_output_reports_untracked_keys(self) -> None:
    self._create_flag("many-junk-keys", {"groups": [], "junk_a": 1, "junk_b": 2})
    out = StringIO()
    call_command("audit_flag_filters", "--team-id", str(self.team.id), stdout=out)
    assert "tracking cap" in out.getvalue()
```

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+        if team_id is not None:
+            queryset = queryset.filter(team_id=team_id)
+        rows = queryset.order_by("id").values_list("id", "team_id", "filters")
+        if limit:
+            rows = rows[:limit]
+
+        # collect_filters_violations runs CROSS_FIELD_CHECKS, which deliberately excludes
+        # check_groups_non_empty_for_create: non-empty groups is a POST-only rule (#50084) —
+        # stored flags with empty groups are valid state and must never show up in this report.
+        for flag_id, flag_team_id, filters in rows.iterator(chunk_size=500):
```
haacked: `suggestion`: `.iterator(chunk_size=500)` won't actually stream on PostHog prod. `DISABLE_SERVER_SIDE_CURSORS` is tied to `USING_PGBOUNCER` and applied to the `default` database ([`posthog/settings/data_stores.py:17`](https://github.com/PostHog/posthog/blob/d51af4fe682f606e85d9f39c160d048eb2622a7e/posthog/settings/data_stores.py#L17), [`:90`](https://github.com/PostHog/posthog/blob/d51af4fe682f606e85d9f39c160d048eb2622a7e/posthog/settings/data_stores.py#L90)), and with server-side cursors disabled, psycopg can't open a named cursor, so `.iterator()` buffers the full result set client-side on execute; `chunk_size` only limits Python object creation after that. A full prod scan (`--limit 0`, no `--team-id`) pulls every flag's `filters` JSON into the command's memory at once, exactly the untested prod-scale behavior the PR description flags as deferred to phase 2, and it would surface as an OOM on that gating run.

Keyset-paginate by id so memory stays bounded regardless of connection, or route the query through `default_direct` (which has server-side cursors enabled) if that alias is configured for this environment. Note the identical `.values_list(...).iterator(chunk_size=500)` pattern with no routing already exists in [`posthog/management/commands/survey_cohort_bytecode_stl.py:334`](https://github.com/PostHog/posthog/blob/d51af4fe682f606e85d9f39c160d048eb2622a7e/posthog/management/commands/survey_cohort_bytecode_stl.py#L334), so this isn't unique to this PR.

```python
def _iter_flag_rows(queryset, *, limit: int, chunk_size: int = 500):
    base = queryset.order_by("id").values_list("id", "team_id", "filters")
    last_id, yielded = 0, 0
    while True:
        page = chunk_size if not limit else min(chunk_size, limit - yielded)
        if page <= 0:
            return
        chunk = list(base.filter(id__gt=last_id)[:page])
        if not chunk:
            return
        yield from chunk
        yielded += len(chunk)
        last_id = chunk[-1][0]
```

products/feature_flags/backend/api/filters_schema.py
```diff
@@ -0,0 +1,389 @@
...
+        operator = data.get("operator") if isinstance(data, dict) else None
+        if isinstance(operator, str) and operator in FEATURE_FLAG_OPERATOR_ALIASES:
+            data = {**data, "operator": FEATURE_FLAG_OPERATOR_ALIASES[operator]}
+        return super().to_internal_value(data)
+
+
+class FlagConditionGroupSerializer(DropsUnknownKeysMixin, serializers.Serializer):
+    unknown_key_level = "group"
+
+    properties = FlagPropertySerializer(
```
haacked, line 275: `suggestion`: `FlagConditionGroupSerializer.properties` rejects an explicit JSON `null`, but Rust models it as `Option<Vec<PropertyFilter>>` with `#[serde(default)]` ([`flag_models.rs:139-140`](https://github.com/PostHog/posthog/blob/d51af4fe682f606e85d9f39c160d048eb2622a7e/rust/feature-flags/src/flags/flag_models.rs#L139-L140)), so `{"groups": [{"properties": null}]}` deserializes fine in Rust (null becomes `None`, treated like absent). This serializer rejects that same value with a 400 "This field may not be null," which breaks the module's own "reject exactly what serde would reject" contract, and it's inconsistent with every sibling `Option` field here: `payloads`, `multivariate`, `holdout`, and both `aggregation_group_type_index` fields all set `allow_null=True` for the same reason.

If any stored flag carries `properties: null`, the audit reports it as a structural violation that Rust evaluates without issue, so reaching a clean run is harder for data that isn't actually cache-poisoning.

```suggestion
    properties = FlagPropertySerializer(
        many=True,
        required=False,
        allow_null=True,
        default=list,
        help_text="Property conditions for this release condition group.",
    )
```

products/feature_flags/backend/api/filters_schema.py
```diff
@@ -0,0 +1,389 @@
...
+class UnknownKeySink(Protocol):
+    def record(self, *, level: str, keys: Sequence[str], flag_id: int | None) -> None: ...
+
+
+def is_legacy_unknown_key(level: str, key: str) -> bool:
+    return level == "filters" and key in LEGACY_UNKNOWN_FILTER_KEYS
+
+
+def _record_dropped_unknown_keys(level: str, keys: Sequence[str], context: Mapping[str, Any]) -> None:
+    sink: UnknownKeySink | None = context.get("unknown_keys_sink")
```
haacked: `suggestion`: The serializer-context keys `"unknown_keys_sink"` and `"flag_id"` are bare string literals shared across two modules: read here in `_record_dropped_unknown_keys`, written in `audit_flag_filters.py`'s `handle()`. A typo in either place fails silently (the sink never fires, or `flag_id` records as `None`), and the audit just reports fewer unknown keys with no error to point at the cause.

Pull them into module-level constants here and reference those from both sides.

```python
UNKNOWN_KEYS_SINK_CONTEXT_KEY = "unknown_keys_sink"
FLAG_ID_CONTEXT_KEY = "flag_id"

# then in _record_dropped_unknown_keys:
sink = context.get(UNKNOWN_KEYS_SINK_CONTEXT_KEY)
flag_id = context.get(FLAG_ID_CONTEXT_KEY)

# and in audit_flag_filters.py's handle():
context={UNKNOWN_KEYS_SINK_CONTEXT_KEY: sink, FLAG_ID_CONTEXT_KEY: flag_id}
```

products/feature_flags/backend/filters_validation.py
```diff
@@ -0,0 +1,359 @@
...
+    check_variant_rollout_sum,
+    check_variant_keys_unique,
+    check_payload_keys,
+    check_group_variant_references,
+    check_property_types_match_aggregation,
+    check_operator_value_compatibility,
+)
+
+
+def check_groups_non_empty_for_create(filters: Mapping[str, Any]) -> list[Violation]:
```
haacked, line 271: `suggestion`: `check_groups_non_empty_for_create` and `validate_cross_field_or_raise` have no phase-1 caller outside their own tests; everything else in this PR reaches production through `collect_filters_violations` / `collect_cross_field_violations` via the audit command. They're phase-3 write-path helpers, and phase 3 is where the exact error codes and message format get pinned to match the live API responses, so their shape is likely to change once they're actually wired in.

Dropping both avoids stale code in this PR; the POST-only exclusion is already documented in a comment in the audit command and in this module's docstring, so no executable stub is needed to preserve that decision. Keeping them as a landing spot for phase 3 is a reasonable alternative, given they're tested.

products/feature_flags/backend/filters_validation.py
```diff
@@ -0,0 +1,359 @@
...
+                    )
+                )
+            if operator in SEMVER_OPERATORS:
+                semver_violation = _check_semver_value(value, operator, path)
+                if semver_violation is not None:
+                    violations.append(semver_violation)
+    return violations
+
+
+CROSS_FIELD_CHECKS: tuple[Callable[[Mapping[str, Any]], list[Violation]], ...] = (
```
haacked, line 261: `question`: The new cross-field rules mirror logic that still lives in `validate_filters` (rollout sum, variant-override existence, operator/value compatibility, semver, cohort-only operators, aggregation-vs-property-type), and during phases 1-2 both run independently. The audit only ever executes the new rules, so if someone edits `validate_filters` later, the mirror can drift silently and the audit's "clean" gate no longer means "matches what the write path accepts."

Is there a plan to guard that window, e.g. a test that runs a corpus of filters through both `collect_filters_violations` and `validate_filters` and asserts the same accept/reject verdict (message text aside, since you've documented that will differ)? That would catch drift before the phase-3 swap.

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+from products.feature_flags.backend.filters_validation import Violation, collect_filters_violations
+from products.feature_flags.backend.models.feature_flag import FeatureFlag
+
+# Unknown keys come from user-controlled JSON, so the distinct-key space is unbounded; cap it
+# to keep a pathological prod scan from growing the aggregator without limit.
+MAX_TRACKED_UNKNOWN_KEYS = 1000
+
+# C0/C1 control characters (incl. ESC and OSC): stored filter junk is user-controlled and the
+# console report prints keys and values verbatim, so strip anything a terminal could interpret.
+_CONTROL_CHARS_RE = re.compile(r"[\x00-\x08\x0b-\x1f\x7f-\x9f]")
```
haacked: `nit`: `_CONTROL_CHARS_RE` strips ESC and the C1 range but lets tab and newline through, so a stored key containing a newline can inject extra lines into the console report and forge a line that mimics a real violation or `[legacy]` row, misleading whoever reads the report that gates enforcement.

```suggestion
_CONTROL_CHARS_RE = re.compile(r"[\x00-\x1f\x7f-\x9f]")
```

products/feature_flags/backend/management/commands/audit_flag_filters.py
```diff
@@ -0,0 +1,228 @@
...
+            self._emit_console(scanned, flags_with_violations, reports, sink)
+
+    def _emit_json(
+        self, scanned: int, flags_with_violations: int, reports: list[RuleReport], sink: UnknownKeyAggregator
+    ) -> None:
+        payload = {
+            "scanned": scanned,
+            "flags_with_violations": flags_with_violations,
+            "clean": flags_with_violations == 0,
+            "rules": [report.__dict__ for report in reports],
```
haacked: `nit`: `_emit_json` serializes each report as `report.__dict__`, which ties the JSON report's field names to `RuleReport`'s internal attributes. Adding an internal-only field to the dataclass later would silently leak it into the machine-readable output.

```python
"rules": [
    {
        "rule_id": r.rule_id,
        "flags_affected": r.flags_affected,
        "total_violations": r.total_violations,
        "sample_flag_ids": r.sample_flag_ids,
        "sample_details": r.sample_details,
    }
    for r in reports
],
```

products/feature_flags/backend/api/filters_schema.py
```diff
@@ -0,0 +1,389 @@
...
+
+This is the runtime Python mirror of the Rust flag structs — the shapes here must stay in
+sync with `rust/feature-flags/src/flags/flag_models.rs` (`FlagFilters`, `FlagPropertyGroup`,
+`MultivariateFlagVariant`, `Holdout`) and `rust/feature-flags/src/properties/property_models.rs`
+(`PropertyFilter`, `OperatorType`, `PropertyType`). A stored filters value that fails serde in
+Rust poisons the whole team's flag cache, so these serializers reject exactly what serde would
+reject: no DRF type coercion (`"true"` is not a bool, `"42"` is not an int, `NaN` is not a
+valid rollout percentage).
+
+Modeled on the OpenAPI-only `FeatureFlagFiltersSchemaSerializer` in `posthog/api/documentation.py`
```
haacked, line 11: `nit`: The docstring says this is "modeled on" `FeatureFlagFiltersSchemaSerializer` in `posthog/api/documentation.py` but doesn't say whether the two are meant to converge or stay separate. A one-line note on the intended end state would save the next person from assuming they should be unified.

products/feature_flags/backend/api/filters_schema.py
```diff
@@ -0,0 +1,389 @@
...
+
+class FlagHoldoutSerializer(DropsUnknownKeysMixin, serializers.Serializer):
+    unknown_key_level = "holdout"
+
+    id = StrictIntegerField(
+        min_value=I64_MIN,
+        max_value=I64_MAX,
+        help_text="ID of the experiment holdout this flag belongs to.",
+    )
+    exclusion_percentage = FiniteFloatField(
```
haacked, line 348: `question`: `holdout.exclusion_percentage` is bounded to 0-100 here, and a test confirms 150 is rejected on purpose. But Rust's `Holdout.exclusion_percentage` is a plain `f64` with no range check, and `exclusion_percentage_clamped` ([`flag_models.rs:132`](https://github.com/PostHog/posthog/blob/d51af4fe682f606e85d9f39c160d048eb2622a7e/rust/feature-flags/src/flags/flag_models.rs#L132)) exists specifically to guard against an out-of-range stored value, direct evidence Rust expects out-of-range holdout percentages to exist and evaluates them fine by clamping. The live write path never validated holdout at all, so this check is new, and an out-of-range stored holdout isn't cache-poisoning since serde accepts any `f64`.

Is flagging Rust-clamped-valid holdouts as violations intended, or should the range check relax to match serde the way the other percentage fields do?

products/feature_flags/backend/filters_validation.py
```diff
@@ -0,0 +1,359 @@
...
+    except (ValueError, IndexError):
+        return Violation(
+            rule_id="cross_field.semver_value_invalid",
+            path=path,
+            message=f"Invalid semver value for operator {operator}: {value}.",
+        )
+    return None
+
+
+def check_operator_value_compatibility(filters: Mapping[str, Any]) -> list[Violation]:
```
haacked, line 207: `nit`: The group/property path prefix `f"groups[{group_index}].properties[{prop_index}]"` gets rebuilt inline several times in this function and in `check_property_types_match_aggregation`. Low value on its own, only worth a small helper if you're touching these functions anyway.

haacked (commented):
Solid phase-1 rollout, nicely tested. A few non-blocking suggestions and questions inline.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Fixed in 79fc4ea1d6f — swapped `.iterator()` for keyset pagination (`id > last_id` in LIMIT-bounded chunks via `_iter_flag_rows`), so memory stays bounded regardless of the connection's cursor support. Went with your suggested shape rather than `default_direct` routing to keep the command environment-independent.

products/feature_flags/backend/api/filters_schema.py
patricio-posthog: Fixed in 79fc4ea1d6f — `allow_null=True`, and `validate_properties` normalizes null to `[]` (same semantics as serde's default) so the cross-field checks can keep iterating without a None guard. Added a valid-case test plus one pinning the normalization.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Added in 79fc4ea1d6f as `test_repeated_rule_counts_flag_once_but_all_violations`, using your case verbatim — you're right that the gating count had no guard.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Added in 79fc4ea1d6f as `test_console_output_reports_clean_scan`.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Added in 79fc4ea1d6f as `test_console_output_reports_untracked_keys`.

products/feature_flags/backend/filters_validation.py
patricio-posthog: Implemented in 79fc4ea1d6f rather than just planned: `test_write_path_parity` runs a 21-case corpus through both `collect_filters_violations` and the live `validate_filters` and asserts the same accept/reject verdict per mirrored rule (rollout sum, override existence, operator/value compat, semver, cohort-only operators, aggregation-vs-type, payload keys). Scope is deliberately limited to rules that exist on both sides with structurally valid input, avoiding the cohort/flag-dependency branches that need DB. If either side drifts during phases 1-2, this fails instead of the audit silently measuring a different rule set.

products/feature_flags/backend/api/filters_schema.py
patricio-posthog: Intentional, keeping it for now: the [0,100] bound is the locked inventory in #50084, and `exclusion_percentage_clamped` means an out-of-range holdout is mis-stored data Rust papers over rather than data behaving as intended — worth surfacing and cleaning in phase 2 rather than baking the clamp into the contract. If the phase 2 audit finds real out-of-range holdouts on working flags, that's the "wrong rule vs bad data" call and we amend the inventory in the issue thread first.

products/feature_flags/backend/filters_validation.py
patricio-posthog: Keeping both per your alternative: they're tested, `check_groups_non_empty_for_create` makes the POST-only exclusion executable (the guard test asserts it stays out of CROSS_FIELD_CHECKS), and the parity test now exercises it as the POST-verdict half. Agreed the error shape of `validate_cross_field_or_raise` will get pinned properly in phase 3.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Fixed in 79fc4ea1d6f — widened to `[\x00-\x1f\x7f-\x9f]` so tab/newline are stripped too.

products/feature_flags/backend/management/commands/audit_flag_filters.py
patricio-posthog: Fixed in 79fc4ea1d6f — fields enumerated explicitly with a comment on why `__dict__` is avoided.

products/feature_flags/backend/api/filters_schema.py
patricio-posthog: Fixed in 79fc4ea1d6f — `UNKNOWN_KEYS_SINK_CONTEXT_KEY` / `FLAG_ID_CONTEXT_KEY` constants in filters_schema.py, imported by the command.

products/feature_flags/backend/api/filters_schema.py
patricio-posthog: Fixed in 79fc4ea1d6f — docstring now states the end state: phase 3 wires this serializer in, it becomes the OpenAPI source of truth, and the documentation.py one is retired.

products/feature_flags/backend/filters_validation.py
patricio-posthog: Agreed it's marginal — left as is since I wasn't otherwise reshaping those loops; will fold in a path helper the next time these functions change.

matheus-vb (approved):
Nice! Your latest commits addressed everything in my view, approving to unblock you.

PR #72410, opened 2026-07-20
feat(review_hog): add review-outcome telemetry

gantoine:
## Problem

PR volume keeps climbing and most code is now written and reviewed with agents, but review quality itself is unmeasured. We don't know whether ReviewHog's findings actually help: which get fixed, which get a human reply, which get ignored. This is the second initiative in our internal "reviews at scale" RFC (part 1 shipped agent-authorship provenance as event telemetry).

## Changes

After a reviewed PR merges, an hourly Temporal sweep classifies each published ReviewHog finding and emits one `reviewhog_finding_outcome` event per finding.

Outcome per finding, in precedence order:

- `reacted` — a human replied to or reacted on the finding's inline comment (bot replies are excluded)
- `addressed` — a post-review commit touched the finding's lines and an LLM judge (run on a different model family than the reviewer) confirms the change resolved it
- `ignored` — neither

How it fits together:

- Discovery reads merged PRs and their branch-tip head SHA from the engineering_analytics warehouse via a new facade function `list_recently_merged_pull_requests`, scoped to the PR numbers awaiting classification so a high-merge-volume repo can't push an eligible PR past the query's row ceiling. Teams without a connected GitHub warehouse source are skipped.
- Line proximity is the cheap gate (parse the post-review compare — paginated past GitHub's 300-file page cap — and check whether any commit touched near the finding's lines). The LLM judge only runs on candidates with no human-engagement signal, to keep token spend down.
- The `reacted` signal comes from the PR's review comments. A finding maps to its posted comment by path plus the exact `### {title}` heading line the comment body leads with, so no comment ids are stored.
- The published finding set is reconstructed from the urgency threshold snapshotted on the report at publish (`published_urgency_threshold`, migration 0020), not the user's live setting, so a post-publish settings change can't rewrite what was posted; pre-existing reports fall back to live settings.
- A `finding_outcome` artefact per finding is the idempotency marker, so a report drops out of the sweep once classified. A report's artefacts are written last, in one transaction, after every event is emitted — a crash mid-report leaves no marker, the next sweep redoes the whole report, and re-emitted events dedup on their deterministic uuids. Events carry `repository` and `pr_number` so they join part 1's provenance in HogQL.

```mermaid
flowchart LR
  Sched{{hourly schedule}} --> Disc[discover merged reports]
  WH[(warehouse: merged PRs)] --> Disc
  Disc --> GH[GitHub compare + comments]
  GH --> Judge{{LLM judge}}
  Judge --> Ev[(reviewhog_finding_outcome + artefact)]
  class Sched phYellow
  class Disc,Judge phBlue
  class GH,WH phRed
  class Ev phGray
```

> [!NOTE]
> The classifier reaches the engineering_analytics warehouse only through that product's facade (new `depends_on` edge, verified by tach). engineering_analytics is owned by team-devex, so the new facade function should get their eyes.

### Review round (addresses the inline review findings)

All of the above reflects the post-review state; the review round changed five things: atomic events-then-markers ordering (closes both the partial-run-marks-complete and marker-without-event findings), the publish-time threshold snapshot, compare pagination (bounded at 10 pages, logged loudly beyond), exact-heading comment matching plus bot-reply exclusion, and the `numbers`-scoped merged-PR lookup. One accepted limitation: bot *reactions* can still count as `reacted` — the reactions summary carries no actor and resolving actors would cost a call per comment, and bots don't react to inline review comments in practice.

Automated only. I (Claude, agent-assisted) did not run a live end-to-end classification against a real merged PR. Ran locally and green:

- 26 new tests (`test_outcome_*`): line-proximity hunk parsing and the proximity window, comment-signal matching plus reply/reaction detection, and classifier precedence / judge-gating / idempotency / deterministic-event-uuid / event-shape.
- Review round added 5 more, each pinned to one of the findings: heading-boundary matching, bot-reply exclusion, compare pagination, no-outcomes-persisted-on-mid-report-failure, snapshot-threshold-beats-live-settings (plus a `numbers`-scope case in the warehouse test and a threshold-snapshot assertion in the publish idempotency test). Full `products/review_hog` suite (299) and `engineering_analytics` `test_logic.py` (96) green, plus repo-wide mypy, ruff, tach, and `makemigrations --check`.
- Existing review_hog model and registry-parity tests still pass with the new artefact type.
- The engineering_analytics facade test for `list_recently_merged_pull_requests`.
- Full-repo `uv run mypy --cache-fine-grained .` (clean on every changed file), `ruff`, `tach check` (my imports are interface-clean), and `makemigrations --check` (migrations 0019 and 0020 match model state).

Each group catches a named regression: switching the event uuid to `uuid4()` re-fails `test_event_uuid_is_deterministic_per_finding`; inverting the reacted-vs-judge precedence re-fails `test_reacted_takes_precedence_and_skips_the_judge`; a mis-walk of new-side line numbers re-fails the hunk-parse case.

## Automatic notifications

- [ ] Publish to changelog? No, internal telemetry with no user-facing surface.
- [ ] Alert Sales and Marketing teams? No.

## Docs update

No user-facing docs affected (internal telemetry, no API/config/UI change).

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). @gantoine directed the work and is assigned as DRI.

Built with Claude Code (Opus 4.8; review round with Fable 5). Skills invoked: `/django-migrations` (the state-only enum migration and the review round's additive threshold-snapshot column) and `/writing-tests` (the test value and efficiency gate).

Decisions worth a reviewer's attention:

- Orchestration is a Temporal schedule, not Celery beat. review_hog is entirely Temporal and the judge is async, so a Celery task would add a second orchestration surface. This is the product's first schedule.
- Discovery uses the warehouse (merged PRs plus branch-tip head SHA in one query) rather than polling GitHub. The compare head is the PR branch tip, never the squash or rebase merge commit, which would produce false positives.
- I explored storing the posted review-comment id at publish time to make the `reacted` signal exact, then dropped it: matching on path plus the exact `### {title}` heading the comment already carries identifies the thread without touching the publish call. The publish path does now stamp one extra column on its existing watermark save (the threshold snapshot) — a field on the same UPDATE, not a new call.
- Idempotency is per-finding (`issue_key`) plus a deterministic event uuid, with a report's markers written as one transaction after its events, so overlapping sweeps, retries, and mid-report crashes can't double-count or strand findings.

Unrelated and pre-existing on master, not touched here: a `tach` interface finding on `posthog/hogql/property_planner.py` and an unused `type: ignore` in `posthog/personhog_client/converters.py`.


rnegron:
@gantoine can you link to the RFC in the PR desc?

sortafreel:
Posting as an understanding layer:

- After a reviewed PR merges, a new hourly job classifies every posted finding as `addressed` / `reacted` / `ignored` and emits one `reviewhog_finding_outcome` PostHog event per finding. It only reads from GitHub — it never posts anything.

## Cases

> *existing logic*
> **changes introduced in this PR**

### Case 1 — ReviewHog publishes a review to a PR

1. *Posts one inline comment per validated finding at or above the user's urgency threshold, and records which commit was reviewed.*
2. **Also saves that threshold on the report, so classification later knows exactly which findings were posted even if the user changes the setting afterwards.**

### Case 2 — the reviewed PR merges

1. **Every hour, a Temporal sweep picks up published reports not yet classified (max 50 per team per run) and asks the engineering-analytics warehouse which of their PRs merged in the last 30 days.**
2. **For each merged PR it reads two things from GitHub: the diff that landed after the review, and the PR's inline comments.**
3. **Each posted finding gets one outcome (cases 3–5), saved to the database first, then sent as one PostHog event carrying the finding's file, priority, category, outcome, and how it was decided — attributed to the user who set up the review, or a stable `repo#pr` id.**

### Case 3 — someone reacted or replied to a finding's comment

1. **The finding's comment is matched by file + title; any reaction, or a reply from a non-bot, counts.**
2. **Outcome: `reacted`. This wins over everything else — human engagement is the strongest signal, and it's decided without any LLM call.**

### Case 4 — a later commit changed code near the finding

1. **If the post-review diff touched the finding's file within 15 lines of its range, an LLM judge reads the finding and that diff. The judge is deliberately a different model than the reviewer, so it doesn't share the reviewer's blind spots.**
2. **Judge rules the diff actually fixed it → `addressed`; rules it merely touched the lines (formatting, unrelated edit) → `ignored`.**

### Case 5 — nothing touched the finding and nobody responded

1. **Outcome: `ignored`, no LLM call spent.**

### Case 6 — the job crashes or hits a GitHub rate limit mid-sweep

1. **Outcomes are written to the database in one transaction before any event is sent, and the report is only stamped done after the events are flushed to capture.**
2. **A re-run re-sends identical events from the saved rows (same deterministic uuid) — it never re-decides an outcome, so a human reply landing between attempts can't flip a result. Duplicate events dedupe by uuid at query time.**
3. **A rate limit or exhausted egress budget stops the sweep early; the next hourly run resumes where it left off.**

### Case 7 — nothing happens, on purpose

1. **A PR that hasn't merged yet: its report just waits — no classification, no event, until it merges.**
2. **A team without a GitHub App installation, or a repo without a connected GitHub warehouse source: skipped, tried again next sweep.**
3. **Events are cloud-only — self-hosted instances emit nothing.**

### Case 8 — existing callers of the shared code this PR touched

1. *Celery tasks using `ph_scoped_capture`, the review chunking/dedup LLM calls, and review publishing all behave as before — the PR only adds a `flush()` method, optional model/effort parameters with the old defaults, and the threshold snapshot.*

posthog/ph_client.py
```diff
@@ -59,6 +59,27 @@ def get_regional_ph_client(**kwargs: Any):
     return get_client(region, **kwargs)
 
 
+class ScopedCapture:
```
sortafreel, line 62: Seems like a heavy change, not entirely sure how justified it is, but merging it won't upset me.

products/review_hog/backend/reviewer/outcomes/classify.py
```diff
@@ -0,0 +1,500 @@
+"""Classify each published finding's fate for a merged PR, and emit one event per finding.
```
sortafreel, line 1: If there are reports that weren't merged - like closed - could they be classified? Or will they be reclassified every time?

products/review_hog/backend/reviewer/outcomes/classify.py
```diff
@@ -0,0 +1,500 @@
...
+
+Per report: pull the post-review compare and the PR's review comments, pair the report's published
```
sortafreel, line 3: Can reports in general be unprocessable and poison the pipeline? Like a deterministic bug with wrong diff lines or smth? I would guess it will fail the whole team activity and repeat when retried? Should we allow judges to genuinely fail, if >75% of them succeed or smth?

products/review_hog/backend/reviewer/outcomes/comment_signal.py
```diff
@@ -0,0 +1,56 @@
+"""The human-engagement signal: did a finding's inline comment get a reply or a reaction?
+
+ReviewHog's published comments lead with ``### {finding.title}`` and anchor to the finding's file, so
+a finding maps to its posted comment exactly by (path, title) — no stored comment id, and robust to
+line drift after review (the match is on body content, not position). The one
+``GET /pulls/{n}/comments`` list carries both an ``in_reply_to_id`` per comment and a ``reactions``
+summary, so replies and reactions are read without any extra call or GraphQL.
```
sortafreel, line 7: I'm not sure we should be ignoring bot comments. Like, I can 100% ask my bots to answer and use `posthog[bot]` tokens to do so. Also, we want ReviewHog to auto-fix resolve its comments soon, and leave unresolved only for "requires human input". So, I would suggest counting bot comments too, just maybe use a different status? WDYT?

products/review_hog/backend/reviewer/outcomes/github_fetch.py
```diff
@@ -0,0 +1,69 @@
+"""The two GitHub reads outcome classification needs, through the gated egress client.
+
+Both are read-only and metered against the installation's budget like every other ReviewHog call.
```
sortafreel, line 3: Ok, it's a bit overlapping, as we hit the same endpoint when pulling original comments, but I get why it needs to happen later, so it's fine.

products/review_hog/backend/reviewer/outcomes/judge.py
```diff
@@ -0,0 +1,69 @@
+"""The arbiter behind the line-proximity gate: did the post-review diff actually fix the finding?
```
sortafreel, line 1: I could see the judge spending a bit of money here and there, checking on every comment for large PR, but the logic is solid, and it makes sense to me, kudos.

sortafreel: I would suggest using Opus 5 though.

products/review_hog/backend/reviewer/outcomes/judge.py
```diff
@@ -0,0 +1,69 @@
...
+
+A commit touching a finding's lines is necessary but not sufficient — it could be an unrelated edit,
```
sortafreel, line 3: Does it make sense to persist the reasoning of the judge also? Like, how did it come to this decision? For debugging and better history in general?

products/review_hog/backend/reviewer/constants.py
```diff
@@ -99,6 +99,26 @@ def effective_priority(base: IssuePriority, adjusted: IssuePriority | None) -> I
...
 
 
+# OUTCOME TELEMETRY
+# The outcome-classifier judge decides whether the commits that landed after review actually
+# addressed a finding. Pinned to a model DIFFERENT from the reviewer's (`REVIEW_MODEL` /
+# `ONESHOT_MODEL` = claude-sonnet-5): a judge sharing the reviewer's model family would inherit the
+# same blind spots the telemetry exists to measure. Effort is "high" — a focused yes/no on a small
+# diff, not the reviewer's exhaustive xhigh pass.
+OUTCOME_JUDGE_MODEL = "claude-opus-4-8"
+OUTCOME_JUDGE_REASONING_EFFORT = "high"
```
sortafreel, line 114: I think 5 and xhigh would be better? Same price? I would still stick to 4.8 for validation till I test properly.

products/review_hog/backend/reviewer/persistence.py
```diff
@@ -440,6 +440,21 @@ def turn(self, report_id: str, run_index: int) -> list[tuple[ReviewIssueFinding,
         verdicts = self._verdicts.get(report_id, {})
         return [(finding, verdicts.get(issue_key)) for issue_key, finding in findings.items()]
 
+    def all_valid(self, report_id: str) -> list[tuple[ReviewIssueFinding, ValidationVerdict]]:
```
sortafreel, line 443: Ok, unrelated-ish - but a concern.

1. ReviewHog reviews the PR at commit A, posts a comment: "bug in utils.py around line 50".
2. The author fixes it in commit B. Doesn't reply to the comment, just fixes it.
3. More commits land, ReviewHog reviews again (turn 2) at commit C and publishes new comments. 
4. The report remembers only one "last published" commit — and it's now C, overwriting A.
5. The PR merges at commit D. The sweep runs.

I assume, when we ask for changes, we'll see C to D changes, but the fix (B) won't be in the diff, as it happened before C? Does it make sense?

products/review_hog/backend/models.py
```diff
@@ -69,6 +70,14 @@ class Status(models.TextChoices):
     # `head_sha`, what was reviewed). Publishing skips when this equals the current head, so an
     # activity retry / re-trigger can't double-post the review or the one-time alpha promo comment.
     published_head_sha = models.CharField(max_length=64, null=True, blank=True)
+    # The urgency threshold in force at the latest publish — outcome classification reconstructs the
```
sortafreel: Nit: would it add value to have `(report, issue_key)` constraint on `finding_outcome` rows?

sortafreel (approved):

products/review_hog/backend/reviewer/outcomes/classify.py
gantoine: they're never classified in the first place, so re-checked ever sweep but that check is cheap it should be fine

products/review_hog/backend/reviewer/outcomes/classify.py
gantoine: not poison but it can get stuck and re-attempt every hour. nice catch, i'll catch each report and abort if the failure ratio is above 25%.

products/review_hog/backend/reviewer/outcomes/github_fetch.py
gantoine: yeah if you can find a cleaner way to do this after it's merged i'd appreciate it!

products/review_hog/backend/reviewer/outcomes/comment_signal.py
gantoine: fair enough, i'll split them into `comment_reply` for humans and `comment_reply_agent` for bots (still ignoring reviewhog's own comments though)

products/review_hog/backend/reviewer/outcomes/judge.py
gantoine: yeah might as well since we're already genereting it, i'll toss it into the artefact

products/review_hog/backend/models.py
gantoine: i don't think so since `issue_key` isn't a column, but i'll add protection against duplicates being ingested

products/review_hog/backend/reviewer/persistence.py
gantoine: it does! i'll record the head per-turn and compare by base commit. it'll cost more github egress but we're already hitting egress limits so 🤷🏼

PR #72416, opened 2026-07-20
feat(data-warehouse): BI-style insight builder for the SQL editor

mjwarren3:
## Problem

Building an insight from SQL today means hand-writing every `GROUP BY` and aggregate, then assigning columns to axes in a cramped side panel. There's no way to say "average of this, split by that" without editing SQL, no pivot table, and chart types give no hint about what data shape they need. BI tools (Tableau, Power BI) solved this with field wells and a live preview.

## Changes

Behind the new `sql-editor-insight-builder` feature flag, the SQL editor gets two modes:

- **Data** – exactly today's editor (Monaco + results grid). Untouched when the flag is off.
- **Build** – a full-canvas builder: a fields panel typed from your base query's columns (dimensions vs measures), drag-and-drop **Rows / Columns / Values** wells with per-field aggregations (sum, avg, min, max, count, count distinct, median, p90/95/99) and date grains (hour to year), a chart type picker where every type declares its requirements ("Needs 1 Row and at least 1 Value"), and a live preview.

The core mechanic is compile-to-HogQL. Wells compile to a `GROUP BY` query wrapped around your base SQL (or a saved view by name), and that compiled SQL always lives in `source.query`:

```sql
SELECT
    toStartOfMonth(created_at) AS created_at_month,
    plan AS plan,
    sum(amount) AS sum_amount
FROM (
SELECT * FROM payments
)
GROUP BY toStartOfMonth(created_at), plan
ORDER BY created_at_month ASC
```

Because every existing consumer just executes `source`, dashboards, insight view mode, exports and endpoints need zero changes. The wells config rides along in a new optional `builder` field on `DataVisualizationNode` and rehydrates when you reopen the insight in the editor.

Also new: a **pivot table** chart type (`ChartDisplayType.PivotTable`) that renders the grouped results as a matrix – row dimensions down the side, column values across the top with grouped headers for multiple measures, formatted cells via the existing per-series Format settings.

Details worth knowing:

- Aliases are machine-generated (`sum_amount`, `created_at_month`), deduped, and kept out of the SQL keyword namespace, so response column names always match what the charts key on.
- The Series side-panel tab hides for builder insights (wells own the axes); Display and conditional formatting stay as the Format panel.
- Old SQL insights without a `builder` config keep the classic experience exactly; flag off means byte-for-byte today's editor.
- Save flows were the tricky part: `isSourceQueryLastRun` and the multi-statement save-target cycler both assume the Monaco buffer equals the executed SQL, which is false in Build mode. Both are branch-guarded now.
- If the Data tab SQL diverges from the snapshot the wells were built on, the fields panel shows a refresh banner instead of silently compiling stale text, and pills whose columns disappeared get an error state.

No screenshots, sorry – see agent context below for why.

> [!NOTE]
> The schema change (TS + regenerated `posthog/schema.py`) ships in this same PR. The pydantic `DataVisualizationNode` is `extra="forbid"`, so saving a builder insight only works once this deploys – there's no separate backend PR to sequence.

Automated (all run by the agent):

- `compileBuilderQuery.test.ts` – identifier escaping (spaces, quotes, backticks, `__proto__`), all date grains and aggregations incl. `quantile(0.9)(x)` syntax, alias dedupe and reserved-word suffixing, view-vs-subquery FROM, multi-statement rejection, GROUP BY uses expressions not aliases, `{variables.x}` passthrough. Catches any regression that would emit invalid or wrong-aliased SQL, which would silently break every saved builder insight.
- `chartCapabilities.test.ts` – parameterized validation matrix per chart type, drop permission modes, smart chart defaults. Catches requirement drift that would let invalid configs save or block valid ones.
- `pivotTableUtils.test.ts` – matrix building, tuple key collisions ("a b"+"c" vs "a"+"b c"), null labels, duplicate counting, prototype-pollution guard on `__proto__` labels.
- `insightBuilderLogic.test.ts` – wells compile into the node and run the compiled SQL; saved nodes rehydrate wells; the apply/hydrate loop guard holds (our own node round-trip doesn't re-hydrate, hydration doesn't re-apply).
- Existing `sqlEditorLogic.test.ts` suite still green; full frontend typecheck 0 errors; repo-wide mypy clean except one pre-existing unrelated error in `personhog_client/converters.py`.
- Compiler output verified against a real HogQL engine: three compiler-emitted queries (date grain + breakdown + `count()`, quoted identifiers + `quantile(0.9)`, reserved-word alias `from_`) executed successfully via the PostHog API.

What the agent could not do: click through the UI. The local dev stack in this session serves a different checkout, so there was no manual browser verification and no screenshots. Before enabling the flag anywhere, someone should run the flows in `frontend/src/scenes/data-warehouse/editor/insightBuilder/` manually: build a chart, save as insight, view on a dashboard, reopen in edit mode.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Feature is flag-gated and unreleased, so no docs update yet.

## 🤖 Agent context

Built by Claude (PostHog Code session) from a plan reviewed and approved by @mjwarren3, who chose the full-canvas Data/Build layout, pivot table in v1, and builder-as-primary-viz decisions up front.

Skills invoked: `/writing-tests`, `/writing-kea-logics`.

Decisions along the way: compile-to-HogQL with the compiled SQL kept in `source.query` was chosen over execution-time wrapping so dashboards and view mode need zero changes. The compiler and capability registry live under `queries/nodes/DataVisualization/insightBuilder/` (not the editor) because the PivotTable renderer must import them without an upward dependency. Hydration uses a deep-equality guard on the `builder` config as the sync-loop breaker, tested explicitly. Data mode stays mounted but hidden in Build mode so the Monaco instance and undo history survive mode switches. A Storybook story for the canvas was considered and skipped for this PR.

---


mjwarren3:
Pushed a second round from design feedback (all still behind the flag):

- **Data mode is results-only** – no Visualization tab or split view (the classic tab remains only when editing a legacy SQL insight saved without a builder config), and it gains a visible **Save as view** button.
- **The builder tab is now "Insight builder"** and hides the database tree – its own fields panel is the data surface there.
- **Fields panel shows the source**: the view/table name when the base is a bare `SELECT * FROM <object> [LIMIT n]`, otherwise "Custom query (Edit)" linking back to Data mode. When the base is such a select-all, the builder compiles against the object itself, so aggregates cover the full data instead of the 100-row preview.
- **Clicking a view in the tree** now pulls it in as `SELECT * FROM view LIMIT 100` (editing stays in the row's dropdown).
- **Chart picker is capability-driven**: types the current fields can't satisfy are grayed out with a hover explaining what to add; adding a field that exceeds the current chart auto-switches to one that fits (bar + Columns becomes stacked bar).
- **Stacked bars fixed**: Columns is the x-axis, Rows is the stacked series.
- **Format is a right-hand slide-out** behind a palette icon.

New coverage: `wellsToVizSettings.test.ts` (the stacked-bar inversion), `detectSelectAllTarget` parsing cases, and logic tests for the auto-switch and the select-all base compiling `FROM <object>` with the preview LIMIT dropped. 221 tests green across the touched suites, full typecheck clean.

mjwarren3:
Reworked the layout per the mockup — dropped the Data/Build tabs for a two-row structure (still all behind `sql-editor-insight-builder`):

**Row 1 — Data source.** A segmented control: **View**, **Warehouse source**, **Custom SQL** (no default). View/Warehouse open a searchable picker (grouped by folder / connector, with a "Manage views" / "Manage connectors" link) and then show a compact source badge with **Change**. Custom SQL drops the editor in — database tree + connection selector on the left, Run/Variables/Filters toolbar + Monaco + Save as view on the right — in a height-draggable row whose tree is no longer full-height.

**Row 2 — Results / Visualization** with a fullscreen-expand button left of the tabs (collapses row 1 to grow the output) and the action buttons on the right. Results is unchanged. Visualization is now a four-column BI canvas:
- **Data** (dimensions + measures), **Setup** (chart type + Rows/Columns/Values/Filters), and **Format** each collapse to an icon rail and are width-draggable.
- **Visualization** fills the remaining width at full height, no title.

Behavior changes from the feedback:
- **Chart type is primary** — picking a chart no longer auto-switches when you add a field. Wells a chart can't use are grayed out ("Not used by this chart type") and reject drops. (This intentionally reverses the earlier auto-switch.)
- **Numeric binning** — a numeric field in Rows/Columns gets a pop-out width modal and compiles to `floor(col / width) * width`; dates keep their grain menu.

New coverage: `isWellEnabled` matrix, numeric-bin compiler cases, and logic tests that a numeric bin flows into the compiled SQL and that adding a Column no longer switches the chart. 140 builder-suite tests + the editor/dataViz suites pass; full typecheck clean. The 3 red Chart.js canvas suites are a pre-existing jsdom issue on this branch, unrelated to these changes.

As before I couldn't drive the UI in a browser (the local stack serves a different checkout), so the drag/resize/collapse interactions and the picker modals want a manual pass before the flag goes on.

mjwarren3:
Pushed the intent split + unified picker round (matching the dropdown mockup):

- **One data-source picker.** The View / Warehouse / Custom SQL segmented control is gone. It's now a single searchable dropdown (quill combobox, search inside the popup) grouped into **Views** and **Warehouse tables**, with a sticky **Write custom SQL** footer item as the escape hatch. Picking anything compiles against the object directly.
- **Editor intent split.** The scene now serves two intents chosen by entry point. Insight intent (New insight, opening a SQL insight, or the new menu entry below) gets the two-row builder layout. Modeling intent (editing views, drafts, endpoints, or plain `/sql` navigation) keeps the classic editor untouched, with a **Build insight** button that bridges over carrying the current query as the Custom SQL base. This keeps view/endpoint authoring out of insight-builder chrome.
- **Insights → New menu**: with the flag on, the SQL entry becomes **Custom** — "Create an insight on a connector, modeled view, or SQL query" — and opens the builder directly.
- **Fixes from testing**: the Data column's resize handle was on the wrong edge (dragging it grabbed the Setup column instead); side columns are now capped at a third of the canvas so widening Setup squeezes the chart instead of pushing Format off-screen; the redundant chart-type selector and settings gear are gone from the Visualization tab in builder layout (the Setup and Format columns own those); and an **Edit source** button appears on the Data column while fullscreen, exiting fullscreen to reveal the data-source row.

209 tests green across the builder + sqlEditorLogic suites, insightNav/saved-insights suites pass (one pre-existing `addSavedInsightsModalLogic` failure, unrelated), full typecheck clean.

PR #72421, opened 2026-07-20
fix(aio): include long-running trace spans in trace view

AlexlaGuardia:
## Problem

Closes #43310

In AI observability, the single-trace view drops spans when a trace's events are spread over more than ~30 minutes. Every span shows up in the flat Generations list, but the trace tree only renders the ones near the start. This hits anyone whose spans arrive over a long span of wall-clock time, for example queue-throttled or rate-limited pipelines that emit a trace's work over tens of minutes to hours.

The cause is the capture window in `TraceQueryDateRange`. The trace query is anchored on the trace's first event and buffered by a symmetric 10 minutes on each side. The `trace_id` predicate is already the exact, fully selective filter, so the timestamp bound exists only to prune ClickHouse partitions. But a symmetric 10-minute buffer assumes a trace finishes within 10 minutes of its first event, and anything emitted later falls outside the window and never reaches the `GROUP BY`.

## Changes

Made the capture window asymmetric. A trace only grows forward in time from its first event, so the backward buffer stays small (clock skew) and the forward buffer extends to cover the trace's real duration.

```diff
     CAPTURE_RANGE_MINUTES = 10
+    FORWARD_CAPTURE_RANGE_MINUTES = 24 * 60

     def date_to(self) -> datetime:
-        return super().date_to() + timedelta(minutes=self.CAPTURE_RANGE_MINUTES)
+        return super().date_to() + timedelta(minutes=self.FORWARD_CAPTURE_RANGE_MINUTES)
```

The forward widen is cheap given how `events` is stored. It is `PARTITION BY toYYYYMM(timestamp)` and ordered by `(team_id, toDate(timestamp), ...)`, so a window kept inside the trace's day prunes the same monthly partition and scans at day granularity whether it spans 30 minutes or several hours. The 30-minute window was already scanning that day's granules for the team, so widening within the day does not add partitions and adds no meaningful granule scan. A day-scoped forward bound keeps the window inside one or two monthly partitions.

> [!NOTE]
> The sibling list runner, `traces_query_runner.py`, already handles long traces with a two-pass bounds-discovery step (query the real min/max per trace, then re-bound the heavy aggregation). I kept the single-trace path on a bounded forward window instead of mirroring that, because the single-trace read already has one known, exact `trace_id` and does not need the discovery round-trip. Happy to switch to the two-pass approach if you'd rather the two runners stay strictly symmetric here.

Added `test_capture_range_includes_long_running_trace`: one trace anchored at `date_from` with spans at +0, +45m, +3h, and +12h, asserting all four land in the tree. The existing `test_capture_range` only covered a span +10m out, which sits inside the old symmetric window, so nothing exercised spans beyond it. Under the old code this new test returns only the first span.

I (Claude, driven by the author) could not run the ClickHouse-backed test suite in my environment, so I have not run this test locally. It is written to match the existing fixtures and assertions in the file and needs CI (or a maintainer) to run green. No frontend changes.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code, directed by the author. The work started from the issue's own root-cause notes, but those cited paths were stale (the product was renamed `llm_analytics` to `ai_observability`), so I re-traced the live chain: the frontend anchors `date_from`/`date_to` on the trace's first event, then `TraceQueryDateRange` pads it, and the pad was the symmetric buffer that clipped long traces.

Decisions along the way:
- Rejected the issue's suggested fix (bump the constant from 10 to a larger symmetric value). It is still an arbitrary duration guess and doubles the backward scan for no reason, and the reporter flagged the cost tradeoff himself.
- Rejected mirroring `traces_query_runner`'s two-pass bounds-discovery for now. It is the more thorough approach, but heavier for a read that already has an exact `trace_id`. Called it out above so a maintainer can ask for it.
- Verified the cost concern against the `events` table DDL (`PARTITION BY toYYYYMM`, `ORDER BY (team_id, toDate(timestamp), ...)`) before choosing the asymmetric widen, which is why the forward buffer is day-scoped.

Repo skills (for example `/writing-tests`) were not available in my session's harness, so I followed the conventions in the existing test file directly rather than invoking them.

### DIFF
--- a/posthog/hogql_queries/ai/trace_query_runner.py
+++ b/posthog/hogql_queries/ai/trace_query_runner.py
@@ -49,11 +49,18 @@ TRACE_FIELDS_MAPPING: dict[str, str] = {
 
 class TraceQueryDateRange(QueryDateRange):
     """
-    Extends the QueryDateRange to include a capture range of 10 minutes before and after the date range.
-    It's a naive assumption that a trace finishes generating within 10 minutes of the first event so we can apply the date filters.
+    Provides a bounded capture range for the shared-events fallback.
+
+    The dedicated table is ordered by `(team_id, trace_id, timestamp)`, so an exact trace lookup
+    does not need timestamp bounds. Shared events is ordered by day and event, so its fallback
+    stays time-bounded.
     """
 
+    # Backward buffer: clock skew / the small negative anchor the frontend applies to date_from.
     CAPTURE_RANGE_MINUTES = 10
+    # Forward buffer: an upper bound on a single trace's duration. A trace that maps to a chat can
+    # stay open across days, so a sub-day bound silently truncates it.
+    FORWARD_CAPTURE_RANGE_MINUTES = 7 * 24 * 60
 
     def date_from_for_filtering(self) -> datetime:
         return super().date_from()
@@ -65,7 +72,7 @@ class TraceQueryDateRange(QueryDateRange):
         return super().date_from() - timedelta(minutes=self.CAPTURE_RANGE_MINUTES)
 
     def date_to(self) -> datetime:
-        return super().date_to() + timedelta(minutes=self.CAPTURE_RANGE_MINUTES)
+        return super().date_to() + timedelta(minutes=self.FORWARD_CAPTURE_RANGE_MINUTES)
 
 
 class TraceQueryRunner(AnalyticsQueryRunner[TraceQueryResponse]):
@@ -78,10 +85,11 @@ class TraceQueryRunner(AnalyticsQueryRunner[TraceQueryResponse]):
     def _calculate(self):
         query_result = query_ai_events(
             query=self._build_query(),
-            placeholders={"filter_conditions": self._get_where_clause()},
+            placeholders={"filter_conditions": self._get_where_clause(include_timestamp_bounds=False)},
             team=self.team,
             query_type=NodeKind.TRACE_QUERY,
             fall_back_to_events=True,
+            fallback_placeholders={"filter_conditions": self._get_where_clause()},
             timings=self.timings,
             modifiers=self.modifiers,
             limit_context=self.limit_context,
@@ -221,7 +229,7 @@ class TraceQueryRunner(AnalyticsQueryRunner[TraceQueryResponse]):
 
     @cached_property
     def _date_range(self):
-        # Minute-level precision for 10m capture range
+        # Minute-level precision for the capture range buffers
         return TraceQueryDateRange(self.query.dateRange, self.team, IntervalType.MINUTE, datetime.now())
 
     def cache_target_age(self, last_refresh: Optional[datetime], lazy: bool = False) -> Optional[datetime]:
@@ -230,19 +238,23 @@ class TraceQueryRunner(AnalyticsQueryRunner[TraceQueryResponse]):
 
         return last_refresh + timedelta(minutes=1)
 
-    def _get_where_clause(self) -> ast.Expr:
-        where_exprs: list[ast.Expr] = [
-            ast.CompareOperation(
-                op=ast.CompareOperationOp.GtEq,
-                left=ast.Field(chain=["ai_events", "timestamp"]),
-                right=self._date_range.date_from_as_hogql(),
-            ),
-            ast.CompareOperation(
-                op=ast.CompareOperationOp.LtEq,
-                left=ast.Field(chain=["ai_events", "timestamp"]),
-                right=self._date_range.date_to_as_hogql(),
-            ),
-        ]
+    def _get_where_clause(self, *, include_timestamp_bounds: bool = True) -> ast.Expr:
+        where_exprs: list[ast.Expr] = []
+        if include_timestamp_bounds:
+            where_exprs.extend(
+                [
+                    ast.CompareOperation(
+                        op=ast.CompareOperationOp.GtEq,
+                        left=ast.Field(chain=["ai_events", "timestamp"]),
+                        right=self._date_range.date_from_as_hogql(),
+                    ),
+                    ast.CompareOperation(
+                        op=ast.CompareOperationOp.LtEq,
+                        left=ast.Field(chain=["ai_events", "timestamp"]),
+                        right=self._date_range.date_to_as_hogql(),
+                    ),
+                ]
+            )
 
         where_exprs.append(
             ast.CompareOperation(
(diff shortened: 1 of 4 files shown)


posthog/hogql_queries/ai/trace_query_runner.py
```diff
@@ -49,11 +49,24 @@
...
+    ordered by `(team_id, toDate(timestamp), ...)`, so a window kept inside the trace's day prunes
+    the same monthly partition and scans at day granularity whether it spans 30 minutes or several
+    hours. The selective predicate is the exact `trace_id`, not the timestamp.
     """
 
+    # Backward buffer: clock skew / the small negative anchor the frontend applies to date_from.
     CAPTURE_RANGE_MINUTES = 10
+    # Forward buffer: an upper bound on a single trace's duration, day-scoped so it stays within
+    # one or two monthly partitions.
+    FORWARD_CAPTURE_RANGE_MINUTES = 24 * 60
```
patriksimms: Why exactly did you choose a 24h limit? if the events table is partitioned by month, everything up to 28 days should not cause increased load on clickhouse right? 

E.g. we are using "one trace - one chat" and it sometimes happen that after a day or two people come back to a chat

posthog/hogql_queries/ai/trace_query_runner.py
patriksimms, line 183: Codex identified this as an affected area. These fields can contain heavy payloads (json etc) and with long running traces (and therefore more events) this might overload clickhouse memory. 

But not sure how to test that in a deterministic way.

This is less a problem in the traces_query_runner.py version because it aggregates less fields in the tuple
https://github.com/PostHog/posthog/blob/master/posthog/hogql_queries/ai/traces_query_runner.py#L313

patriksimms (commented):
@AlexlaGuardia Very nice PR! Just have a question and found one potential problem with codex

AlexlaGuardia:
Thanks Patrik, both good catches.

On the 24h: you're right, too tight. I set it conservative on the first pass, but "one trace = one chat resumed a day later" is exactly the case that should still resolve and 24h silently drops it. I widened it to 7d and stress-tested the cost so we're not trading the bug for a load problem.

Repro was a 2000-span trace with heavy payloads spread over 7 days, next to 200k same-team/same-day rows. `EXPLAIN ESTIMATE` for the 24h and 7d windows comes back identical (same parts, same rows, same 26 granules) — inside a monthly partition the forward bound doesn't move the scan, since `trace_id` isn't in the sort key so the reader hits the same team/day block either way. The 24h window returned 286 of the 2000 spans; 7d returned all of them.

On the Codex flag: fair read, but this endpoint pins the aggregation to one trace (`trace_id = ?`, `GROUP BY trace_id LIMIT 1`), so a wider window can't pull in other traces' events. Peak memory tracks a single trace's span count, not window × traces, and that same pathological trace peaked around 62MB whatever the window width.

The genuinely wasteful part of that groupArray is the `arrayDistinct` over the full tuple — it hashes the heavy JSON to dedup rows whose natural key is just `uuid`. Deduping on uuid instead (cast the (uuids, tuples) pair to a Map) returns the identical event set and dropped peak from ~62MB to ~42MB in the same test. Happy to fold that into this PR or leave it as a follow-up, your call.

One question back at you: is 7 days a safe ceiling for a single trace's duration, or would you rather it go wider / be configurable given the resumed-chat pattern?

AlexlaGuardia:
Pushed the 7d change in 602f5e8f6. My earlier comment described that as already done when it was still uncommitted on my end, so the PR you were looking at still had the 24h bound in it. Sorry for the runaround.

What's in the commit:

- `FORWARD_CAPTURE_RANGE_MINUTES` 24h -> 7d, and the docstring now gives the actual reason it's cheap rather than the day-scoped argument I originally wrote: `trace_id` isn't in the events sort key, so widening the forward bound doesn't move the scan.
- Extended `test_capture_range_includes_long_running_trace` with a 3-day span. The old offsets topped out at 12h, which passed under both the old and new bound, so the test wasn't actually pinning this behavior down.

Still open from my side, and I'd rather you make the call: is 7 days a sane ceiling for a single trace, or would you rather it be configurable? I defaulted to 7d because it covers the resumed-chat case you described without reaching past two monthly partitions, but you have much better data than I do on how long real traces actually run.

I left the `arrayDistinct` -> dedupe-on-uuid change out so this diff stays on the bug. Happy to open it separately.

One ask: Backend CI shows `action_required` on this branch because the PR is from a fork, so the Django tests haven't run against it yet. Could you approve the workflow run? Taking it out of draft now.

patriksimms:
> is 7 days a sane ceiling for a single trace

From my perspective yes, but dont know enough about the overall implications. But also looks like we are getting from the posthog team to review, which should answer this question

Radu-Raicea:
@AlexlaGuardia @patriksimms 
When representing a chat, you should use a [Session](https://posthog.com/docs/ai-observability/sessions). Each turn in the conversation (1 user input leading to an agent loop) should be represented by a [Trace](https://posthog.com/docs/ai-observability/traces).

We have a 30-day data retention policy on heavy AI properties (like messages). The way that we modelled this is the following:
- During ingestion, an AI event is written in 2 tables:
  - in `ai_events`, with all its properties
  - in `events` with all the properties, except for the heavy ones
- After the 30 days, the row in `ai_events` is dropped

At query time, we first try to query `ai_events` (very fast retrieval, and contains the heavy properties). If the query returns nothing, a fallback query is sent to `events`.

I made some changes in this PR to not put a time filter on the initial `ai_events` query, because the sorting key in that table doesn't need a time range to be efficient. I left your changes for the fallback to `events`.

Radu-Raicea:
@AlexlaGuardia Thank you for the PR! I'll get it green and merged 🙌

Radu-Raicea:
@AlexlaGuardia Send me an email to my first name at posthog dot com for a merch voucher :D

patriksimms:
@Radu-Raicea thanks for the explaination! And for merging the PR

Radu-Raicea (approved):

Radu-Raicea:
@AlexlaGuardia could you please re-do your commits with a verified signature? It's a requirement for us to merge this.

AlexlaGuardia:
Signed and pushed, `d04fcab`.

One heads up: I squashed the branch to a single commit. Rewriting my unsigned commits at the base changed every descendant SHA, so your commits couldn't keep their signatures either way. Rather than hand-resolve the snapshot conflicts a linear rebase kept producing, I built one commit directly against `723ab657` with the exact tree of the old head. Tree hash is unchanged from what you reviewed, so the diff is identical, just re-signed. You're on it as `Co-authored-by`.

If you'd rather have the commits back as separate entries, say the word and I'll redo it that way.

Radu-Raicea:
@AlexlaGuardia All good!

PR #70159, opened 2026-07-10
feat(mcp-store): replace icon assets with logo.dev icon_domain resolution

cvolzer3:
Stacked on #70158

## Problem

MCP store icons are repo-committed. This slows down the process of adding new MCP servers and isn't agent friendly.

## Changes

- `MCPServerTemplate.icon_domain` — the vendor's brand domain (`linear.app`), normalized on save (lowercase bare hostname). `icon_key` stays for now and is dropped in a follow-up; migration 0014 backfills domains for every shipped icon_key with a URL-hostname fallback.
- New `GET .../mcp_servers/icon/?domain=<domain>` action proxying the brand icon via `CDPIconsService` (egress-gated by the base PR; icon bytes are browser-cached only, per logo.dev licensing). The `domain` param is validated as a bare hostname so the endpoint can't be steered off `img.logo.dev`.
- Serializers expose `icon_domain` alongside `icon_key` (template + installation); `icon_key` stays read-only and marked deprecated until PostHog Code migrates to `icon_domain`, then both the API field and the column are dropped. OpenAPI/generated types regenerated.
- `ServerIcon` renders the proxy URL with the generic server icon as `onError` fallback. The asset imports and `SERVER_ICONS` dict are gone. Custom installs now derive a best-effort brand domain from their server URL (strips `mcp.` / `api.` / `www.`), so they get real icons for the first time.
- The icon proxy forwards the active UI theme (`theme=light|dark`, sanitized at the view so arbitrary values can't mint unbounded cache keys) so logo.dev serves the variant suited to the background, and opts into `fallback=404` so domains without a logo get the store's generic glyph instead of logo.dev's letter monogram.

- Review follow-ups: the icon proxy passes the requesting team into `CDPIconsService`, so uncached fetches are charged to the base PR's per-team fairness budget (one team spamming unique domains degrades only its own icons, not the instance-wide logo.dev budget). `ServerIcon` latches load failures per (domain, theme) instead of per domain, so a transient failure or an exhausted budget no longer blanks the icon for the component's lifetime across theme switches — a theme flip retries, matching the backend's uncached, retry-on-next-render degradation.

`frontend/public/services/` assets are untouched — they're shared with CDP templates and warehouse sources.

## How did you test this code?

Automated only, no manual testing:

- `test_api.py` — icon endpoint rejects path traversal / full URLs / query injection (parameterized) and proxies a valid domain; `icon_domain` save-normalization cases; existing template/installation serializer tests updated to the new field (131 tests pass).
[...]

PR #72422, opened 2026-07-20
chore(tach): expose event_definitions property_type in public interface

cvolzer3:
## Problem

Backend CI's repo checks fail on every PR that runs them, with:

```
[FAIL] posthog/hogql/property_planner.py:41: The path 'products.event_definitions.backend.property_type.PropertyType' is not part of the public interface for 'products.event_definitions'.
```

#70734 added a tach `[[interfaces]]` entry for `products.event_definitions` exposing only `backend.models.*` and `backend.logic.placeholder.*`. But `posthog/hogql/property_planner.py` imports `PropertyType` from the pure module `backend.property_type` (deliberately, since #71913 relocated these enums to pure modules), so tach now fails module boundary checks. The check only runs on PRs touching backend files, which is why it didn't surface on master directly. Example failing run: https://github.com/PostHog/posthog/actions/runs/29774979645/job/88462200354

## Changes

One-line fix, adds the pure module to the exposed interface:

```diff
 [[interfaces]]
 expose = [
     "backend\\.logic\\.placeholder.*",
     "backend\\.models.*",
+    "backend\\.property_type.*",
 ]
 from = [
     "products.event_definitions",
 ]
```

The alternative of rerouting the import through `backend.models.property_definition` would undo the pure-module split from #71913, so widening the interface is the intent-preserving fix.

The same one-liner is also included in #70159 to unblock its CI. Whichever merges first, the other merges cleanly.

Ran `uv run tach check --dependencies --interfaces` locally. It reproduces the failure before the change and reports `✅ All modules validated!` after.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (PostHog Code cloud task) while diagnosing failing CI on #70159. Traced the repo-checks failure to the tach interface added in #70734, confirmed the violating import predates it, and validated the fix with tach locally.

---

### DIFF
--- a/tach.toml
+++ b/tach.toml
@@ -1040,6 +1040,7 @@ layer = "modules"
 expose = [
     "backend\\.logic\\.placeholder.*",
     "backend\\.models.*",
+    "backend\\.property_type.*",
 ]
 from = [
     "products.event_definitions",


cvolzer3:
Note on the red checks: the failing entries are all the `[optional, does not block merge] Backend CI` Depot shadow trial, which sampled this PR and failed at the infrastructure level (every job, including trivial ones like "Get ClickHouse versions", died in under 25s). The canonical Backend CI run on this PR passed, including the repo checks / tach step this PR fixes. No required check is failing.

cvolzer3:
This fix has now landed on master via #70159 (which carried the same one-line change), so this PR is redundant and can be closed. Leaving that to a human per process.

PR #72428, opened 2026-07-20
feat(replay-vision): split summary and alert into focused flows

ksvat:
## Problem

The Vision Actions editor (the "Summaries and alerts" tab) served two genuinely different jobs — scheduled AI summaries and alerts — through a single form with a "Type" toggle that rewrote about half the fields. The screen meant two different things at once, and it morphed as you configured it, which made the whole surface feel busy and hard to reason about. This is the first of a few streamlining changes.

## Changes

- The tab now leads with two buttons, **New summary** and **New alert**, instead of one "New group summary or alert". The chosen mode rides into the editor as a `?mode=` query param, so the editor opens committed to one shape and renders a single focused form. The in-form Type toggle is gone.
- The scanner's built-in daily digest no longer shows as a row in this table. It already has a home on the Observations tab (the digest card), so the table now lists only the summaries and alerts a user created, and its empty state is meaningful again for a fresh scanner.
- User-facing "group summary" copy is now just "summary" (buttons, toasts, titles, empty state). Code identifiers, URLs, and data-attrs are unchanged.

No backend or API changes.

- `actionEditorSceneLogic.test.ts` — added a case that the new-action route opens the summary form by default and the alert form when `?mode=alert`; folded the default-mode assertion into the existing new-action test.
- `visionActionsLogic.test.ts` — added a case that the built-in digest is filtered out of the table list.
- Both suites plus the existing cases: 15/15 green. Ran `typegen:file`, `build:products`, oxlint, and `hogli ci:preflight` clean.

I did not manually click through the running app.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Kim (@ksvat) is the DRI.

Built with Claude Code (Opus). Skills invoked: `/writing-kea-logics`, `/writing-tests`, plus the quill primitives guide while deciding the button pattern.

Decision worth flagging: the plan considered a quill `DropdownMenu` split button, but I went with two plain LemonButtons instead — clearer for "pick the mode up front", consistent with the rest of the LemonUI tab, and it avoids mixing quill and Lemon internals in one component (which CLAUDE.md forbids). Also a judgment call: I filtered the built-in digest out of the table entirely rather than keeping it as a "Default" row. If you'd rather keep it visible-but-grouped, that's a quick revert of the filter in `visionActionsLogic`.

This is Workstream A of a larger streamlining effort (alert-condition presets, run-now, and delivery expansion follow in separate PRs).

---

### DIFF
--- a/frontend/src/products.tsx
+++ b/frontend/src/products.tsx
@@ -1286,7 +1286,8 @@ export const productUrls = {
     replayVisionAction: (actionId: string): string => `/replay-vision/actions/${actionId}`,
     replayVisionActionRun: (actionId: string, runId: string): string =>
         `/replay-vision/actions/${actionId}/runs/${runId}`,
-    replayVisionActionNew: (scannerId: string): string => `/replay-vision/${scannerId}/actions/new`,
+    replayVisionActionNew: (scannerId: string, mode?: 'group_summary' | 'alert'): string =>
+        `/replay-vision/${scannerId}/actions/new${mode === 'alert' ? '?mode=alert' : ''}`,
     replayVisionActionEdit: (actionId: string): string => `/replay-vision/actions/${actionId}/edit`,
     revenueAnalytics: (): string => '/revenue_analytics',
     codeReview: (): string => '/code_review',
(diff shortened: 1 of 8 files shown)


ksvat:
* **#72434** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#72428** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/72428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

arnohillen (approved):

PR #72451, opened 2026-07-20
feat(experiments): duplicate shared metric as single-use metric

### LINKED ISSUE #72450: Feature Request: Add an option to duplicate shared metric as a single-use metric in an experiment
## Is your feature request related to a problem?
- Our experiment metrics have an option to duplicate them.
- When you attempt to duplicate a shared metric, it only lets you create a new shared metric.
- It would be useful to duplicate the shared metric as a single use. 
- This would allow you to use the shared metric like a template and modify it for that specific experiment.

## Describe the solution you'd like
- It would be great if there was an additional button in the popup for this.
- Perhaps an option like `Duplicate as single-use metric` in the popup.

Image

## Describe alternatives you've considered
- Right now, you would need to manually recreate a shared metric as a single use metric.

## Additional context
- From https://us.posthog.com/project/2/support/tickets/64173

### Debug info

```shell
Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019f7fa8-c73c-7d9c-90cc-a113acd14d9c?t=33931
Admin (internal): http://go/adminOrgUS/019c4ed6-add0-0000-998a-30c999608349 (project ID 311650)
```

phillram:
## Problem

Experiment metrics can be duplicated, but duplicating a shared metric only ever creates another shared metric. There was no way to take a shared metric and make a one-off, experiment-specific copy without recreating it by hand.

Closes #72450

## Changes

Adds a "Duplicate as single-use metric" option to the metric actions menu, shown only for shared metrics. It copies the shared metric's query into the experiment as an inline (single-use) metric, so you can use a shared metric like a template and tweak it for that experiment. The metric's breakdowns (which live on the shared-metric link metadata) carry over onto the new inline copy, and the shared-metric link itself is left untouched.

Added in to the existing duplicate shared metrics modal:

Screenshot 2026-07-28 at 13 26 14

Copies the metric as singular:
CleanShot 2026-07-27 at 08 47 53@2x

The existing "Duplicate" action is unchanged and still creates a new shared metric.

- Ran locally and tested the metric. 

Added two Jest cases to `experimentLogic.test.ts` for the new `duplicateSharedMetricAsInlineMetric` action:
- appends an inline copy (new uuid, "(copy)" name, merged breakdowns, no `isSharedMetric` link) and leaves `saved_metrics` untouched — guards against the copy accidentally staying linked as shared or dropping breakdowns.
- no-op when the shared metric isn't linked as the requested primary/secondary type.

Ran the experiments logic suite locally (both new tests pass) and the frontend TypeScript check (no errors in the changed files). I (the PostHog Slack app agent) did not do manual UI testing.

## 🤖 Agent context

Authored by the PostHog Slack app agent from a Slack thread. Invoked the `/writing-tests` skill before adding the logic tests. The core change lives in a new kea reducer action (`duplicateSharedMetricAsInlineMetric`) in `experimentLogic`, wired up through `MetricHeader` → `MetricRowGroup` → `MetricsTable`. I built the inline copy from the shared metric's `saved_metrics` entry rather than the enriched in-memory metric, so the stored query and breakdown metadata are the source of truth.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BAB645UBA/p1784588127794389)*


andehen:
@phillram Nice!

I'm making some changes, we can put the choice in the existing shared metrics duplicate modal, to make the dropdown simpler. I think thats better. I'm taking the freedom to take over this PR and get it over the line. Let me know if you don't want that!

andehen (approved):

PR #60569, opened 2026-05-28
feat(flags): strip non-allowlisted $feature_flag_called properties at ingestion

haacked:
## Problem

`$feature_flag_called` is a system event. The SDKs emit it automatically with a fixed set of properties, it's billing-exempt (it doesn't count against a customer's event volume), and we own its schema. Since nobody's paying for these events, there's no reason for arbitrary properties to ride along on them, but they do. SDK methods like `register` and super properties leak unrelated properties onto every event, and a few client-side persistence caches have dumped large blobs onto them too. The result is ClickHouse storage filled with data that has no meaning on this particular event.

posthog-js #3392 and #3393 fixed the persistence-cache leak, but that fix only takes effect once customers upgrade their SDK. This strips the properties server-side, so it works no matter what version a client is pinned to.

## Changes

The ingestion `prepareEventStep` now enforces an allowlist (`FEATURE_FLAG_CALLED_KEEP`). Anything not on the list gets removed before the event hits ClickHouse. Keys prefixed with `$feature/`, `$initial_`, or `$session_entry_` are kept, along with the standard PostHog auto-captured and super properties. `$active_feature_flags` is also kept, because some CDP legacy-plugin destinations (rudderstack-posthog, posthog-laudspeaker) map it onto `context.active_feature_flags` and dropping it would break them.

There's one important exception. When `$feature_flag_response` is a string variant, which only happens for multivariate flags (the kind experiments run on), we keep everything. Experiment breakdowns and custom exposure criteria filter on arbitrary event properties, so stripping there would quietly break them. Boolean, null, numeric, object, and missing responses all get stripped.

Stripping is gated per team through `STRIP_FEATURE_FLAG_CALLED_PROPERTIES_EXCLUDED_TEAMS` (a comma-separated list of team IDs, with `*` as a global kill switch). It defaults to `*`, so stripping is off everywhere until we're ready to ramp it.

One bug worth calling out: the testing subpipelines (`testing-event-subpipeline`, `testing-per-distinct-id-pipeline`, and the rest) were calling `createPrepareEventStep()` without passing the matcher through. That meant the kill switch was ignored and stripping ran unconditionally in testing mode, regardless of config. The matcher now threads through all of them, so testing mode behaves like production.

For visibility once this rolls out, there's a histogram of the property count per event before stripping, and a counter that records whether each event was kept (`kept_multivariate`) or stripped.

The existing `stripBloatProperties` denylist (the posthog-js persistence-cache keys) hasn't changed and still runs on every event.

[...]

PR #72466, opened 2026-07-21
feat(ingestion): add minimal_flag_called_events team configuration

haacked:
## Problem

$feature_flag_called events carry every property on the request, and that payload has grown large from properties that aren't needed for most teams. Stripping them down needs to roll out gradually, team by team, rather than as a single global cutover, so this needs to be a setting we control internally, not a customer-facing toggle.

## Changes

Node's TeamManager now reads a per-team minimal_flag_called_events flag out of feature_flags_teamfeatureflagsconfig and defaults it to false when no config row exists, so nothing changes for any team until we turn it on through the staff API. The value is added to the Team interface and threaded through every test fixture that constructs a full Team object.

This PR only surfaces the setting to the Node ingestion process. The actual property stripping lives in #60569, which currently gates on an env-var exclude list and will be rebased onto this branch to read minimal_flag_called_events instead.

Extended team-manager.test.ts with three cases:
- defaults to false when no TeamFeatureFlagsConfig row exists
- reflects true when a config row is present
- does not leak one team's value into another team's fetch, which guards the WHERE team_id = t.id correlation in the new subquery

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/nodejs/src/common/utils/team-manager.ts
+++ b/nodejs/src/common/utils/team-manager.ts
@@ -134,6 +134,10 @@ export class TeamManager {
                 t.logs_settings,
                 t.extra_settings,
                 extract('epoch' from t.drop_events_older_than) as drop_events_older_than_seconds,
+                COALESCE(
+                    (SELECT minimal_flag_called_events FROM feature_flags_teamfeatureflagsconfig WHERE team_id = t.id),
+                    false
+                ) AS minimal_flag_called_events,
                 o.available_product_features
             FROM posthog_team t
             JOIN posthog_organization o ON o.id = t.organization_id
--- a/nodejs/src/common/utils/team-manager.test.ts
+++ b/nodejs/src/common/utils/team-manager.test.ts
@@ -1,5 +1,11 @@
 import { forSnapshot } from '~/tests/helpers/snapshots'
-import { createTeam, getFirstTeam, resetTestDatabase, updateOrganizationAvailableFeatures } from '~/tests/helpers/sql'
+import {
+    createTeam,
+    getFirstTeam,
+    insertRow,
+    resetTestDatabase,
+    updateOrganizationAvailableFeatures,
+} from '~/tests/helpers/sql'
 import { Hub, Team } from '~/types'
 
 import { defaultConfig } from '../config/config'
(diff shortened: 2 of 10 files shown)


nodejs/src/common/utils/team-manager.ts
```diff
@@ -134,6 +134,10 @@ export class TeamManager {
                 t.logs_settings,
                 t.extra_settings,
                 extract('epoch' from t.drop_events_older_than) as drop_events_older_than_seconds,
+                COALESCE(
+                    (SELECT minimal_flag_called_events FROM feature_flags_teamfeatureflagsconfig WHERE team_id = t.id),
+                    false
+                ) AS minimal_flag_called_events,
```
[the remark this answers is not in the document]
haacked: Fixed.

nodejs/src/common/utils/team-manager.ts
```diff
@@ -37,6 +39,18 @@ export class TeamManager {
         })
     }
 
+    /**
+     * Subscribes to cross-process team-config invalidation. Without this, a change made through
+     * e.g. the feature-flags staff API only reaches this process on the LazyLoader's ~2 minute
+     * background refresh. `pubSub` must already be started.
+     */
+    public subscribeToReloads(pubSub: PubSub): void {
```
nickbest-ph: if the requirements don't demand it, i'd prefer if we leave out the pubsub.

if we do need pubsub, then I think the reload-team callback below needs to be updated to break the cache for both api_token and team id (the team manager caches on both those keys)

we should be able to check if the team is already cached and if it is we can markForRefresh an array of both [teamId, cachedResult.api_token]

nickbest-ph: also, we could put a guard here on incoming messages from pubsub accidentally being `null`

happy path publishers from the django code here shouldn't publish such messages, but if it does somehow receive `null`, the attempt to destructure into `teamId` will crash all subscribed pods

haacked: I'll remove it. A 2-minute lag is just fine for something like this.

nickbest-ph (approved):

PR #72468, opened 2026-07-21
feat(cohorts): add staff tools for stuck cohort recalculation

haacked:
## Problem

When a cohort's calculation gets stuck (its background task died without resetting is_calculating), the only way to unstick it today is a prod shell runbook. There's no staff UI to look up any team's cohort or force a recalculation.

## Changes

- New staff-only viewset at /api/cohorts_staff/ (root-scoped, gated on IsStaffUser) with three actions: look up cohorts by id, list stuck cohorts instance-wide, and force-recalculate by bumping pending_version through the same enqueue path organic saves use
- New /cohorts/staff scene with a lookup panel and a stuck cohorts panel, cross-linked from the flags staff tools scene and added to both account menus
- Extracted RepeatedOrCommaSeparatedListField out of staff_cache.py into posthog/api/fields.py so the cohorts endpoint can reuse the repeated-or-comma-separated query param handling instead of duplicating it

Here's what it looks like:

Screenshot 2026-07-20 at 6 00 45 PM
Screenshot 2026-07-20 at 6 00 56 PM
Screenshot 2026-07-20 at 6 01 09 PM
Screenshot 2026-07-20 at 6 01 20 PM

## How did you test this code

test_staff_tools_api.py covers lookup, stuck listing, and recalculate, including skip reasons for deleted and static cohorts. cohortsStaffToolsLogic.test.ts covers the frontend logic. Also manual testing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update


products/cohorts/backend/api/staff_tools.py
```diff
@@ -0,0 +1,260 @@
...
+                # enqueued (those must not be retried; only failed_cohort_ids should be).
+                try:
+                    # IsAuthenticated + IsStaffUser guarantee a real User here.
+                    fully_resolved = increment_version_and_enqueue_calculate_cohort(
+                        cohort, initiating_user=cast(User, request.user)
+                    )
+                except Exception as e:
+                    logger.exception("cohorts_staff_recalculate_enqueue_failed", cohort_id=cohort_id, error=str(e))
+                    capture_exception(e)
+                    failed.append({"cohort_id": cohort_id, "error": str(e)})
```
matheus-vb, line 237: On enqueue failure the cohort isn’t left untouched: `_prepare_cohort_for_calculation` already persisted `is_calculating=True` + bumped `pending_version` before `.delay()` raised, so a failed cohort is now actually stuck (and for a cohort with dependencies, so is its whole chain, which we report nowhere). This matters most during a broker incident, which is when I think this tool would be used. You can either reset `is_calculating`/`pending_version` here before appending to failed, or soften the help text below: right now `failed_cohort_ids` says “were not queued at all … retry only these ids”, which is misleading about the row’s state.

products/cohorts/frontend/staff/cohortsStaffToolsLogic.ts
```diff
@@ -0,0 +1,259 @@
...
+        lookUpCohortsFailure: () => {
+            lemonToast.error('Failed to look up cohorts.')
+        },
+        loadStuckCohortsFailure: () => {
+            lemonToast.error('Failed to load stuck cohorts.')
+        },
+    })),
+    afterMount(({ actions }) => {
+        actions.loadStuckCohorts()
+    }),
```
matheus-vb, line 254: `afterMount` fires `loadStuckCohorts()` regardless of `user.is_staff`, and the scene binds this logic, so a non-staff user hitting /cohorts/staff directly gets both the `AccessDenied` page and a red “Failed to load stuck cohorts” toast from the 403. You could gate the fetch on is_staff (or swallow the 403 toast). Cosmetic though, low priority.

matheus-vb (approved):
Nice! One suggestion on error handling and a nit.

PR #72469, opened 2026-07-21
feat(data-warehouse): publish modeled duckgres tables

EDsCODE:
## Problem

Managed warehouse users cannot query modeled Duckgres tables from PostHog data warehouse surfaces.

Publishing needs an explicit snapshot flow that prevents partial reads, enforces environment boundaries, and remains safe across retries and deletion.

## Changes

- Users can publish modeled tables as versioned Parquet snapshots and query them through existing PostHog data warehouse surfaces.
- The publish API accepts only discoverable tables from the authorized environment and requires project-wide warehouse access.
- Temporal probes empty sources before copying and records the root user-facing error without retrying permanent validation failures.
- Registration locks publication and table state, so concurrent deletion cannot resurrect a queryable table or repoint it to pruned files.
- Failed copies prune their fresh partial folders immediately, while successful repoints retain the reader safety buffer.
- Workflow runs receive distinct modeling jobs, and failed workflow startup releases the publication name for retry.
- Pending publications stay absent from HogQL schemas, while completed publications serialize as warehouse tables.
- The publication model owns workflow snapshot state; the saved query remains the catalog and query-surface projection.
- Mechanical: merge current `master` and regenerate OpenAPI and MCP artifacts.

Before:

~~~mermaid
flowchart LR
    A["Modeled Duckgres table"] --> B["Managed warehouse queries only"]
    class A phBlue;
    class B phYellow;
~~~

After:

~~~mermaid
flowchart LR
    A["Modeled Duckgres table"] --> B["Publish API"]
    B --> C["Temporal workflow"]
    C --> D["Versioned Parquet snapshot"]
    D --> E["Data warehouse table"]
    E --> F["PostHog query surfaces"]
    class A,C phBlue;
    class B phRed;
    class D,E phGray;
    class F phYellow;
~~~

- API regressions cover permissions, environment isolation, startup cleanup, republish, and deletion.
- Workflow regressions cover empty sources, run identity, root errors, failed-copy pruning, table restoration, and delete races.
- HogQL regressions cover pending and completed publication schema loading.
- Product isolation, Ruff, ty, repo-wide mypy, frontend and MCP code generation, and strict CI preflight ran locally.
- MCP type checking passed.
- Frontend type checking reached an unchanged `master` error in `ComposerModePicker.tsx`; the PR does not modify that file.
- A live Duckgres and Temporal end-to-end run was not performed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The PR adds internal SQL editor documentation. User-facing setup documentation remains part of the frontend follow-up.

## 🤖 Agent context

Codex implemented and validated the review fixes in a dedicated worktree. Claude contributed an earlier design review.

Skills used: /django-migrations, /improving-drf-endpoints, /writing-tests, /implementing-mcp-tools, /debugging-ci-failures, /isolating-product-facade-contracts, /running-ci-preflight, /signed-git-publish, and /writing-pr-descriptions.

The review pass removed read-time reconciliation, tightened environment ownership, locked cross-product writes, and separated permanent validation failures from retryable infrastructure failures.


danielcarletti:
Had my agent cross the PR changes to the plans we had created, it found a few fixes that should be done:

- The Postgres-source guard blocks .postwh.com as a copy-mode source — intentionally killing the unofficial workaround path the same PR replaces. Direct-query mode still allowed.
- Stale-version cleanup races in-flight queries (publish.py::delete_stale_publish_versions). It deletes every non-kept version folder immediately after repoint. A ClickHouse query that resolved the old url_pattern seconds earlier breaks mid-read. The codebase already solves this: prepare_s3_files_for_querying keeps folders younger than S3_DELETE_TIME_BUFFER = 600s. Versions are sortable timestamps — skip folders within the buffer window. Small change.
- Concurrent-publish secret clobber (publish_table_copy_activity). CREATE OR REPLACE SECRET posthog_publish_write uses a fixed name but a per-destination SCOPE. Two concurrent publishes on the same org: B's CREATE OR REPLACE replaces A's secret, A's COPY then fails prefix resolution. The sink avoids this by scoping bucket-wide with one value. Fix: unique secret name per publication (+ DROP SECRET after), or bucket-wide scope like the sink.
- Docstring overclaims cred scoping (build_publish_write_secret_sql): "credentials minted from this process's identity, scoped to exactly the destination prefix" — the SCOPE clause is DuckDB path-routing, not an ACL; the shipped credentials are the Temporal worker's full ambient IAM creds, now on a server where customers hold root SQL. The sink established this pattern (read direction), so it's not a new class of exposure, but publish extends it to write. Two asks: (a) fix the docstring so nobody trusts SCOPE as a security boundary; (b) file the hardening follow-up for both call sites — STS-downscoped creds (assume-role/federation token with inline policy limited to the destination prefix), and confirm with team-managed-warehouse that duckgres actually isolates session secrets per connection (if the DuckDB instance is shared, duckdb_secrets(redact=false) from a customer root session is the exfil path).

Worth a comment, not blocking
- Deleted publications leak S3 forever. delete_publication soft-deletes the table but never removes the publish folder; name is then reusable → new folder, old one orphaned with no owner. Needs a deletion/retention story (even just "cleanup on delete, best-effort").
- Repoint-then-describe window (publish_table_register_activity): url_pattern commits, then set_columns(get_columns()) runs outside the transaction — if the model's schema changed, live queries briefly hit new files with old columns; if DESCRIBE fails repeatedly, the table stays mismatched. Describing the new folder before the repoint transaction closes the window.
- row_count can skew from the copied data — count(*) and COPY are separate autocommit statements, so a commit between them lands them on different snapshots. DuckDB's COPY TO returns the copied row count; use that instead and drop the extra query (also removes the pre-count scan).
- size_in_s3_mib never set → build_function_call never picks s3Cluster for large published tables; they'll read single-node. Cheap to compute from the S3 listing at register time.
- Empty models can't be published (hard ValueError → FAILED). Materialization writes an empty parquet placeholder instead. Fine as a v1 restriction; the error copy already says "yet".
- Stuck PUBLISHING on terminate/cancel — acknowledged in a code TODO. Needs a reaper before this is load-bearing, since status is what the UI/MCP will poll.

jghoman (approved):

jghoman (approved):

fuziontech (changes requested):
Automated review agent on behalf of @fuziontech.

Two high-confidence blockers:

- `products/data_warehouse/backend/logic/managed_warehouse_publish.py:84-88` validates only identifier syntax. Because Duckgres credentials are organization-wide, a project editor can submit an excluded or sibling-project schema/table directly (bypassing the discovery filter), and the workflow will copy it into their project. Enforce a team-owned publish allowlist in both creation and the copy activity.
- `posthog/temporal/ducklake/publish_table_workflow.py:158-205` can race deletion. If DELETE runs after the initial reads but before the register transaction, `table.save()` can create or resurrect a queryable `DataWarehouseTable` after its publication was deleted; the delayed prune then removes the files underneath it. Re-lock/recheck the publication inside the transaction and persist a freshly locked table (or narrowly update fields) before repointing.

fuziontech (approved):
Automated review agent review on behalf of @fuziontech.

Approved. The implementation is well covered and the publication flow is appropriately isolated behind product facades.

Non-blocking follow-ups:
- Validate the requested source table with the same `is_publishable_table` rules used by discovery before creating a publication. The POST path currently accepts any identifier in the project's modeled schema, including internal scratch names or a missing table.
- Re-read or lock the publication inside deletion. A publish can commit `table_id` after the API reads a pending record but before deletion starts, which can leave the new `DataWarehouseTable` active after the publication is deleted.
- Restore soft-deleted backing tables during registration, or prevent their generic deletion. The existing-table path updates the URL and columns but does not clear `deleted` or `deleted_at`, so a republish can complete while the table remains hidden.
- Separate the saved-query response definition from its writable request schema. Generated create and patch clients now advertise `ManagedWarehouseSource`, but runtime validation still requires a non-empty `query` string.

`git diff --check` and Python compilation passed. Focused pytest runs could not start because dependency setup exhausted the VM disk.

fuziontech:
💸 Merge conflict fixed and the branch is up to date. That'll be $4.20, payable in DuckDB credits. (Kidding, of course.)

products/data_modeling/backend/models/datawarehouse_saved_query.py
```diff
@@ -522,6 +526,11 @@ def hogql_definition(
         DirectClickHouseTable,
         DirectMotherDuckTable,
     ]:
+        if self.origin == self.Origin.MANAGED_WAREHOUSE:
+            if self.table is None:
+                raise Exception("Published table is not available yet")
```
danielcarletti, line 531: dc-review-pr found this:

hogql_definition raises Exception("Published table is not available yet") when a managed-warehouse saved query has no table yet, and Database.create_for calls it unguarded at database.py:1900 (sources.saved_queries at :1537 applies no origin filter, and unlike the endpoint loop at :1924 this loop has no try/except). create_managed_warehouse_saved_query creates the saved query with table=None before the workflow starts, so from the moment a user clicks publish until register commits — up to the 2h start_to_close_timeout, and permanently if the publish fails — every HogQL query, insight, and SQL-editor load for that team raises. Fix: skip managed-warehouse saved queries with no table when building the database (continue in the create_for loop) instead of raising, so an unfinished publication is simply absent from the schema.

products/data_modeling/backend/logic/managed_warehouse_saved_queries.py
```diff
@@ -0,0 +1,209 @@
...
+    name: str,
+    source_schema_name: str,
+    source_table_name: str,
+    created_by_id: int | None = None,
+) -> ManagedWarehouseSavedQueryRecord:
+    validate_saved_query_name(name)
+    saved_query = DataWarehouseSavedQuery.objects.create(
+        team_id=team_id,
+        name=name,
+        query={
```
danielcarletti, line 55: dc-review-pr found this. Missed a reader of this object. All others are covered.

posthog/hogql/database/database.py:1336 — serialize_database reads saved_query.query["query"] for every non-endpoint view, but a published table's query is {"kind": "ManagedWarehouseSource", "source_schema_name": ..., "source_table_name": ...} with no "query" key, so this raises KeyError once a publish succeeds. This fires on the data-management schema and SQL-editor sidebar for any team with a published table. Fix: branch on saved_query.origin == MANAGED_WAREHOUSE here the way :1320 already branches on ENDPOINT, and emit the source descriptor (or omit query) rather than indexing blindly.

danielcarletti (changes requested):
My auto reviewer found a lot of `should-fix` , listed them all below. Please get your agent to triage them (some fixes will actually fix multiple items, I believe).
There are two main blockers, which I commented inline. The rest of the review are just nits.

One thing I noticed: existing Managed Warehouse sources synced as a Postgres source will break, which forces them to use the new flow. Just want to make sure this is intended (I'm support it!!).

Items:
[should-fix] products/managed_warehouse/backend/facade/api.py:126 — _to_published_table_record, a DTO converter reached from list_managed_warehouse_published_tables and get_managed_warehouse_published_table, calls complete_managed_warehouse_saved_query_publish and fail_managed_warehouse_saved_query_publish. Both are @transaction.atomic with select_for_update, so a plain GET managed-warehouse-published-tables takes row locks and writes per publication, plus one get_managed_warehouse_saved_query query per row (N+1). Both branches are also unreachable in normal operation: the register and mark-failed activities set publication and saved-query state in the same transaction. The stated reason — an older Temporal worker finishing in-flight work — cannot apply to a model and a workflow that have never shipped, so there are no legacy rows to reconcile. Fix: delete both reconciliation branches and keep _to_published_table_record a pure read; if a reaper is wanted for the acknowledged PUBLISHING gap, make it a scheduled task, not a side effect of a GET.

[should-fix] products/managed_warehouse/backend/temporal/publish_table_workflow.py:260 — the failure-path prune cannot delete the folder the failed run just wrote. prune_published_snapshot_activity sets min_age_seconds = S3_DELETE_TIME_BUFFER (600s) whenever the publication is not deleted, and delete_stale_publish_versions skips any version folder younger than that — which a partial COPY from seconds ago always is. The description's claim that "the publish workflow runs it on both success and failure paths, so failed COPYs leave no partial folders behind" does not hold; the files survive until some later publish or the delete path. The test that names this case, test_prune_published_snapshot[failed_first_publish_prunes_partial_attempts], only passes because its fixture versions are hardcoded in the past, while test_prune_spares_recently_written_versions demonstrates the actual behavior. Fix: pass min_age_seconds=0 on the failure path (nothing is live to protect when the copy never registered), and change the parameterized case to use a freshly-stamped version so it covers the real scenario.

[should-fix] products/managed_warehouse/backend/temporal/publish_table_workflow.py:359 — except Exception as error: ... error=str(error)[:512] catches the ActivityError that execute_activity raises, not the activity's own exception. str() of that wrapper is Temporal's generic activity-failure message, so publication.last_error and saved_query.latest_error — which the API returns straight to the user — never carry "Empty modeled tables cannot be published yet." or "Choose a modeled table from this project.". Fix: unwrap to the root cause (walk error.cause to the ApplicationError) before recording, and raise the user-facing cases as ApplicationError(..., non_retryable=True) from the activity.

[should-fix] products/data_warehouse/backend/presentation/views/data_warehouse.py:1439 — create_publication commits, then start_publish_workflow runs outside any transaction. If sync_connect() or start_workflow fails, the endpoint 500s and leaves a publication plus a saved query stranded in pending with no workflow behind them and nothing to clean them up; the name is now taken, so the user cannot retry under the same name. The delete path explicitly handles this (_start_snapshot_prune_best_effort, covered by test_delete_succeeds_when_prune_cannot_be_scheduled), and there is no equivalent test here. Fix: wrap the start, and on failure mark the publication failed (or delete it) and return 503; add the matching test.

[should-fix] products/managed_warehouse/backend/temporal/publish_table_workflow.py:118 — the workflow ID f"duckgres-publish-{publication.id}" is rebuilt by hand here and again at managed_warehouse_publish.py:171, in a different product. This repo already has the shared-builder convention for exactly this (build_register_data_imports_workflow_id, re-exported from temporal.py). Worse, the string is constant across runs, so start_managed_warehouse_saved_query_publish's dedup key (status=RUNNING, workflow_id) cannot tell a retry from a later republish: a job left RUNNING by a terminated workflow gets silently reused by the next publish, which then writes that run's results onto the stale job row. Fix: add a build_publish_table_workflow_id(publication_id) helper next to the existing one, and key job dedup on temporalio.activity.info().workflow_run_id so retries share a job but a new run gets a fresh one.

[should-fix] products/managed_warehouse/backend/temporal/publish_table_workflow.py:160 — the empty-table check runs after the full COPY has already written parquet, and it raises ValueError, which the maximum_attempts=2 retry policy re-runs — so publishing an empty table performs two complete copies and leaves two partial folders (see the prune finding). Fix: probe with SELECT 1 FROM <schema>.<table> LIMIT 1 before the COPY, and raise the rejection as a non-retryable ApplicationError. The same applies to the "Choose a modeled table from this project." check at :110, which is already validated synchronously in create_publication and should not be retried.

[should-fix] products/managed_warehouse/backend/models.py:161 — ManagedWarehousePublishedTable carries name, status, last_published_at, last_error, row_count, table_id, and deleted, all of which DataWarehouseSavedQuery already stores, and _to_published_table_record then prefers the saved query's copy of every one of them. That duplication is what forces the status-mapping table, the reconciliation writes, and the dual soft-delete in delete_publication. The only state the saved query does not hold is source_schema_name / source_table_name (already in its query JSON), folder_version, and active_job_id. Fix: drop the model and put folder_version and active_job_id in the saved query's query JSON or a small DataWarehouseSavedQuery extension — that removes the migration, the semgrep entry, the second soft-delete, the dual-key _publication_queryset lookup, and the read-time reconciliation in one move, and makes the saved query the single source of truth the doc says it already is.

[should-fix] products/data_warehouse/backend/presentation/views/data_warehouse.py:1373 — the publication resource is five @action methods bolted onto a viewset that is already ~1370 lines, with the identifier passed three different ways: not at all (list), in the POST body (republish), and in the query string (DELETE ?id=). A DELETE that takes its target from a query parameter is not a shape this codebase uses. Fix: register a nested ManagedWarehousePublishedTableViewSet with standard list / create / destroy / a republish detail action, so the identifier lives in the URL and the endpoints follow the same REST shape as every other warehouse resource.

products/data_modeling/backend/logic/managed_warehouse_saved_queries.py
EDsCODE: Fixed in ec70c245181. Managed warehouse saved queries with a backing table now serialize as `DatabaseSchemaDataWarehouseTable`, so serialization never indexes a missing `query` field. A completed-publication regression test covers the SQL editor schema path.

products/data_modeling/backend/models/datawarehouse_saved_query.py
EDsCODE: Fixed in ec70c245181. Database construction now omits managed warehouse saved queries until they have a backing table. The regression test covers both database lookup and serialization during the pending state.

products/data_warehouse/backend/presentation/views/data_warehouse.py
```diff
@@ -1301,3 +1365,124 @@ def check_schema_name(self, request: Request, **kwargs) -> Response:
...
+            400: OpenApiResponse(description="The modeled table or publication name is invalid."),
+        },
+        summary="Publish a managed warehouse table",
+        description="Copy a modeled Duckgres table into the ClickHouse-queryable PostHog data warehouse.",
+    )
+    @action(
+        methods=["POST"],
+        detail=False,
+        url_path="managed-warehouse-publish-table",
+        required_scopes=["warehouse_view:write"],
```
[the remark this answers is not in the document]
EDsCODE: Fixed in 9bc16169fee. All five aggregate publication actions now set `requires_resource_level_access=True`, so object-scoped grants cannot authorize project-wide publication operations.

products/data_warehouse/backend/presentation/views/data_warehouse.py
```diff
@@ -1116,3 +1181,130 @@ def check_schema_name(self, request: Request, **kwargs) -> Response:
...
```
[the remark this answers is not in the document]
EDsCODE: Triaged, but not added in this PR. A team-wide admission limit needs one capacity policy across managed warehouse workloads, rather than an endpoint-only request throttle that workers can bypass. The endpoint now restricts sources to the authorized discovery allowlist; broader concurrency quotas should land with the feature rollout policy.

products/data_warehouse/backend/logic/managed_warehouse_publish.py
```diff
@@ -0,0 +1,199 @@
...
+    try:
+        managed_warehouse.validate_duckgres_identifier(source_schema_name)
+        managed_warehouse.validate_duckgres_identifier(source_table_name)
+    except ValueError as error:
+        raise PublishValidationError(str(error)) from error
+
+    source_team_id = managed_warehouse.ducklake_data_modeling_schema_team_id(source_schema_name)
+    source_team = Team.objects.only("id", "parent_team_id").filter(id=source_team_id).first()
+    source_project_id = source_team.parent_team_id or source_team.id if source_team is not None else None
+    if source_project_id != canonical_team.pk:
```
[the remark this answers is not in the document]
EDsCODE: Fixed in 9bc16169fee. Creation now requires the source schema owner to equal the request environment and verifies the exact table against that environment discovery result. The activity independently rejects sources outside the canonical project and rechecks publish eligibility before connecting.

EDsCODE:
Review triage for the latest push:

Fixed:

- Publication DTO mapping is now read-only and no longer performs per-row saved-query reconciliation.
- Failed copies bypass the reader buffer and prune their fresh partial folder immediately.
- Permanent validation failures are non-retryable, and the workflow records the unwrapped user-facing cause.
- Workflow startup failure deletes the pending publication and saved query, releasing the name for retry.
- Workflow IDs use one shared builder, while modeling-job dedup includes the Temporal run ID.
- Empty sources are rejected by a pre-copy probe.
- Registration, republish, and delete lock current rows and recheck deletion before repointing.
- Republish restores an independently soft-deleted backing table.
- Source creation uses the exact authorized environment discovery allowlist.
- Pending and completed saved queries now load safely in HogQL schema consumers.
- All aggregate actions require project-wide warehouse access.

Kept intentionally:

- `ManagedWarehousePublishedTable` remains the workflow snapshot record for folder version, source identity, and active job state. The saved query remains the catalog projection. Both update in one transaction, without read-time reconciliation.
- The existing action routes remain stable for generated frontend and MCP clients. A nested REST viewset would be a separate compatibility change.
- An endpoint-local throttle is not included. Team capacity admission should cover all managed warehouse workloads at the worker boundary.
- Blocking managed warehouse hosts as copy-mode Postgres sources is intentional; modeled tables use the publication flow instead.

The PR description now reflects these changes and the local validation.

Gilbert09 (changes requested):
Nice work on the safety properties here - the versioned folders, the repoint-after-copy ordering, and the reader buffer on prune all look right.

Four asks, left as inline comments on the relevant lines:

1. Keep this out of `products/data_warehouse` - add a `managed_warehouse` product instead.
2. `published_tables.py` looks misplaced in `warehouse_sources`.
3. Don't use `DataWarehouseSavedQuery` to store this - use a table id reference on the new model.
4. Does ClickHouse actually have IAM read on the org buckets?

products/data_warehouse/backend/logic/managed_warehouse_publish.py
```diff
@@ -0,0 +1,213 @@
...
+    reserved_table_names = frozenset({managed_tables.events_table, managed_tables.persons_table})
+    return [
+        ManagedWarehouseModeledTable(schema_name=str(row[0]), table_name=str(row[1]))
+        for row in result.results
+        if str(row[0]) == modeled_schema
+        and managed_warehouse.is_publishable_table(str(row[0]), str(row[1]), reserved_table_names=reserved_table_names)
+    ]
+
+
+def create_publication(
```
Gilbert09, line 84: Can we avoid adding stuff to `products/data_warehouse`? Instead add a `managed_warehouse` product for just your stuff.

I'd like to kill off the `data_warehouse` product, considering everything should be in its own "product" folder now that data warehouse isn't really a single product any more.

Same goes for the endpoints and serializers added to `presentation/views/data_warehouse.py`.

products/warehouse_sources/backend/published_tables.py
```diff
@@ -0,0 +1,78 @@
...
+from django.db import transaction
+
+from products.warehouse_sources.backend.models.table import DataWarehouseTable
+
+
+def active_table_name_exists(*, team_id: int, name: str) -> bool:
+    return DataWarehouseTable.objects.filter(team_id=team_id, name=name).exclude(deleted=True).exists()
+
+
+def prepare_table_registration(
```
Gilbert09, line 13: Is this the best place for this? Why is it not in the managed warehouse product folder?

This is managed-warehouse-specific publishing logic, and it's pulling a `warehouse_sources` facade write surface into existence just to serve it.

products/data_modeling/backend/logic/managed_warehouse_saved_queries.py
```diff
@@ -0,0 +1,213 @@
...
+            "kind": "ManagedWarehouseSource",
```
Gilbert09, line 56: I don't think we should be using `DataWarehouseSavedQuery` to store this.

Why are we not using a table id reference in the new `PublishedTable` model, and then updating everywhere that we read tables with the new case - e.g. like what we do for saved queries and external data schema everywhere (like in modeling dags etc)? That way we don't confuse the published table with saved queries, because they are so different.

Stuffing a non-HogQL body into `query` also breaks the field's contract - it's declared `help_text="HogQL query"` and every consumer assumes `query["query"]` is a string. See the fall-through I've flagged in `database.py`.

posthog/hogql/database/database.py
```diff
@@ -1892,6 +1906,11 @@ def _build_from_sources(
         with timings.measure("data_warehouse_saved_query", emit_span=True):
             for saved_query in sources.saved_queries:
                 with timings.measure(f"saved_query_{saved_query.name}"):
+                    if (
+                        saved_query.origin == DataWarehouseSavedQuery.Origin.MANAGED_WAREHOUSE
+                        and saved_query.table is None
```
Gilbert09, line 1911: This guard only covers `table is None`, but that isn't the only way a managed-warehouse row reaches `hogql_definition()` without a usable definition. It survives via one branch only:

```python
if self.table is not None and self.is_materialized and modifiers is not None and modifiers.useMaterializedViews:
    return self.table.hogql_definition(modifiers)
# falls through to:
if not isinstance(query, dict) or "query" not in query:
    raise Exception("Saved query is missing a query definition")
```

Two uncovered paths:

- `useMaterializedViews` resolving `False` - it's assigned from the `data-modeling` flag in `create_default_modifiers_for_user`, so a user outside that rollout hits the fall-through.
- `is_materialized` being cleared - `revert_materialization()` does exactly that, and there's no origin guard on it.

Either raises inside this `sources.saved_queries` loop, which fails the whole HogQL database build for the team rather than degrading one table. A `table_id` reference on the new model avoids this class of problem instead of needing a guard at every reader.

products/data_warehouse/backend/presentation/views/saved_query.py
```diff
@@ -657,12 +667,25 @@ class DataWarehouseSavedQuerySerializer(
...
+                    "properties": {
+                        "kind": {"type": "string", "enum": ["HogQLQuery"], "default": "HogQLQuery"},
+                        "query": {"type": "string"},
+                    },
+                    "required": ["query"],
+                },
+                {
+                    "type": "object",
+                    "properties": {
+                        "kind": {"type": "string", "enum": ["ManagedWarehouseSource"]},
```
Gilbert09, line 682: This advertises `ManagedWarehouseSource` as an accepted input shape for `query`, but there's no writer for it.

On the update path, `"query" in validated_data` calls `view.get_columns(user=...)`, which only wraps a body as `HogQLQuery` when `kind` is absent. Here `kind` is present and isn't a real query kind, so it reaches `process_query_dict` as an invalid `QuerySchemaRoot` and 500s rather than returning a 400.

products/managed_warehouse/backend/publish.py
```diff
@@ -0,0 +1,132 @@
...
+
+from django.conf import settings
+
+from psycopg import sql as psql
+
+if TYPE_CHECKING:
+    from types_aiobotocore_s3.type_defs import ObjectIdentifierTypeDef
+
+# Leading underscores prevent collisions because sanitized DuckLake schema
+# directories cannot start with them. The bucket policy uses this same prefix.
```
Gilbert09, line 21: Does ClickHouse have IAM permissions to read the org's buckets to read this data? I don't see anything for this but I may have missed it.

The read path does depend on it. `published_tables.py` never sets `credential`, and `build_function_call` only appends a key/secret when both are present, so the emitted call is `s3('https://{bucket}.s3.{region}.amazonaws.com/...', 'Parquet', '<structure>')` with no credentials. That's the documented contract on `DataWarehouseTable._reject_client_supplied_url_pattern_change` ("read by ClickHouse under the cluster's own S3 role"), and it's why the `internally_computed_url_pattern=True` opt-out is needed.

This comment says the bucket policy uses the same prefix, but I can't find anything in the PR that establishes the `s3:GetObject` / `s3:ListBucket` grant for the ClickHouse role. Given no live Duckgres and Temporal end-to-end run was done, worth confirming before merge - a missing grant surfaces as the `Access Denied: while reading key:` message, which is worded to blame the customer's credentials, and here there aren't any.

Gilbert09 (changes requested):
Moving my feedback onto the relevant lines.

PR #72474, opened 2026-07-21
feat(conversations): detect ticket spikes and surface incidents

HaynesPostHog:
## Problem

Support ticket patterns are spotted by whoever happens to be watching the inbox. A bug that generates 15 similar tickets in two hours looks like 15 unrelated rows in the tickets list, and nobody is alerted until a human connects the dots. Teams need Conversations itself to flag possible incidents, and they need to set their own criteria for what counts as alert-worthy, without requiring AI.

## Changes

Adds a no-AI detection engine to Conversations, gated behind the `support-ticket-trends` flag.

```mermaid
flowchart LR
    T[Tickets in Postgres] --> C{{Temporal coordinator, 15 min}}
    C --> D{{Per-team detection}}
    D --> S[Spike scoring: overall / channel / priority + alert rules]
    S --> I[TicketIncident row]
    I --> E[$conversation_incident_detected event]
    I --> N[In-app notification]
    E --> H[Slack / Teams / Discord / webhook destinations]
    I --> U[Trends tab + incident banner]
    class C,D phBlue;
    class T,I phGray;
    class S phYellow;
    class E,N,H,U phRed;
```

**Detection (backend)**

- A 15-minute Temporal coordinator scans teams with recent tickets or open incidents, then fans out per-team analysis workflows with deterministic IDs and fan-out caps, mirroring the existing support-reply coordinator.
- Pure spike scoring in `scoring.py`: the observed window count is compared against a seasonality-matched baseline (same window, same clock time, prior 28 days) with a robust z-score. A spike fires only when it also clears an absolute floor and a multiple of the median. Low-volume teams fall back to a trailing 24h window. The in-progress hour is never scored.
- Built-in series (overall volume, per channel, per priority) plus customer-defined `TicketAlertRule`s: any tickets-list filter combination with an absolute threshold and an optional relative spike condition. A shared `apply_ticket_filters` helper is now the single implementation behind both the tickets list endpoint and rule evaluation, so a rule counts exactly what the list view shows.
- A detected spike opens a `TicketIncident` row. The open ACTIVE row is the dedup and cooldown, it auto-resolves after consecutive calm evaluations or 24 hours, and dismissal suppresses re-fires for 24 hours.

**Alerting**

- Fires `$conversation_incident_detected` (registered in taxonomy) with title, counts, baseline, z-score, sample ticket numbers and URLs. New `conversations-incident-detected` HogFunction sub-templates give one-click Slack, Teams, Discord, and webhook destinations, mirroring `error-tracking-issue-spiking`.
- In-app notification via the notifications facade, targeting the configured `notification_recipients` with a whole-team fallback.

**UI**

- New Trends tab at `/support/trends`: incidents table with sparklines, alert-rule management with a create/edit modal, and linked alert destinations.
- Incident banner on the tickets list, plus a "Create alert rule" action that pre-fills a rule from the current filters.
- Settings section with an enable toggle, sensitivity controls, and the in-app notification toggle.

Tier 2 (opt-in AI clustering of tickets into semantic trends via the shared embeddings infra) is designed but deliberately not in this PR.

Automated tests added and run locally (all pass), plus repo-wide mypy, ruff, frontend typecheck, jest, and `hogli ci:preflight` (0 failures):

- `test_trends_scoring.py` (11 cases): locks the fire condition against flat/noisy baselines, the MAD-zero path, the never-score-the-in-progress-hour rule, the low-volume fallback, and 1h vs 2h window selection. These parameters are the product's noise/signal contract and had no coverage.
- `test_trends_detection.py` (8 cases): incident opens once and dedupes across runs, auto-resolve via calm runs, dismissal suppression, volume-over-channel hierarchy, absolute rules counting only matching tickets, and notification targeting (configured recipients vs team fallback). Catches lifecycle regressions the pure scoring tests cannot.
- `test_trends.py` API tests (8 cases): rule CRUD with filter-key validation (time filters rejected), the enabled-rule cap, team scoping, and incident dismiss idempotency.
- `test_ticket_filtering.py`: asserts the rule-evaluation queryset and the tickets list endpoint return identical results for the same params, the guarantee behind "create alert rule from filters".
- Existing suites re-run: 169 tickets API tests (the filter machinery moved to a shared helper) and the tickets scene jest tests (the list loader now builds params from the same selector rules store).

Not manually tested end-to-end in a browser. A `run_ticket_trends` management command is included to run detection inline for a team when dogfooding locally.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No published docs cover Conversations alerting yet. Docs should land when the feature flag rolls out.

## 🤖 Agent context

Authored by Claude (PostHog Code session) from a plan reviewed and approved by the assignee, with explicit requirements that the core work without AI, that customers can set their own alert criteria, and that operating cost stays minimal. Repo skills invoked while producing this PR: /django-migrations, /improving-drf-endpoints, /writing-kea-logics, /adopting-generated-api-types, /sending-notifications, /writing-tests.

Notable decisions: detection reads Postgres ticket counts rather than ClickHouse events so rule filters (tags, assignee, search) stay authoritative and evaluation matches the list view exactly; the incident row itself is the dedup/cooldown state instead of Redis; new channel/priority incidents are suppressed while an overall-volume incident is firing so one event does not produce three alerts. One documented deviation from the plan: the low-volume baseline uses the prior 28 daily windows rather than same-weekday-over-8-weeks, trading weekday seasonality for 3x the baseline samples. A post-implementation two-axis review (standards + spec, run as parallel subagents) produced 15 findings that were fixed before this PR, including per-keystroke team PATCHes in settings, missing double-submit guards, and notification recipients not being read.

---


HaynesPostHog:
Pushed fixes for the ReviewHog findings in a4f108a. Both must-fix and all twelve should-fix findings are addressed:

**Must fix**
- Canonical team id: both trends viewsets (and the rule-cap check) now read via `parent_team_id or team.id`, matching what `TeamScopedRootMixin.save()` writes. Rules/incidents are no longer invisible for teams with a parent env.
- PUT wipe: the rules viewset is PATCH-only with the same guard comment as `TicketViewViewSet`; a 405 test locks it in.

**Should fix**
- Notification recipients are filtered to current org members (mirroring the new-ticket email path); stale ids fall out, with a test using a real membership + a stale id.
- Baseline scoring excludes days before the team's first ticket and requires 7 real baseline days, so young teams can't fire off phantom-zero medians. Two pure-function tests cover the young-team and partial-history cases.
- The 24h age backstop only resolves incidents that have stopped firing (no reopen/re-notify loop for a continuous spike), with a regression test.
- Added a `(team, created_at)` index via `SafeAddIndexConcurrently` for the recurring detection scans.
- `search` is no longer a valid rule filter (unindexed comment scan on a background cadence) and rule tag lists cap at 10; filter values are now validated at save time so a malformed value can't silently broaden a rule to all tickets. The evaluator also drops disallowed keys defensively.
- The trends scene itself is flag-gated (`NotFound` when off), not just the tab/settings entries; the sub-template carries the flag too.
- Frontend: `newRuleFilters` hash is cleared after consumption, number inputs guard against NaN (modal + settings), `saveRule` has a re-entrancy guard, and the rule modal passes `hasUnsavedInput`.

**Not changed (rationale)**
- "View tickets" linking to the unfiltered list: the tickets list doesn't read filters from URL params today; wiring that is a separate change. Noted as a follow-up.
- Banner refresh cadence, per-mutation diagnostic logging, list pagination beyond 100, and delivery-tracking for notifications: acceptable for a flag-gated v1; the detection cadence is 15 minutes and kea-loaders already dedupes concurrent reloads.
- Rule names flowing into Slack templates: the name is authored by the same team that configures the destination, so no trust boundary is crossed.

All suites re-run green: 39 trends backend tests, 169 tickets API tests, 11 jest tests, repo-wide mypy clean, preflight passing.

PR #72484, opened 2026-07-21
fix(mcp): advertise structured JSON inputs

### LINKED ISSUE #70064: MCP: generated tools emit `zod.unknown()` for object/array params → untyped input schema → object args get stringified and rejected by the API
### Bug description

The generated MCP tools (`services/mcp/src/generated/<domain>/api.ts`, produced by `tools/openapi-codegen`) type any request-body or query field that the source OpenAPI schema left without a concrete JSON-Schema `type` as `zod.unknown()`. `zod.unknown()` serializes to an **empty** JSON Schema (`{}`, no `type`) in the tool's advertised `inputSchema`. When an MCP client sees a property with no type, it passes the argument as a JSON-encoded **string**; the PostHog API then receives a string where it expects an object/array and raises, e.g. `'str' object has no attribute 'get'`.

The net effect is that several create/update tools cannot be invoked successfully through MCP at all — any call that must send a structured field fails, regardless of payload. Read tools whose structured params are properly typed (e.g. `query-run`, whose `query` is a typed object schema) accept the identical structured value and work, which isolates the cause to the untyped-schema emission rather than the API.

This is systemic, not a handful of tools: **184 params are typed `zod.unknown()` across ~30 generated domains** on current `master`.

### Deployment context

- Surfaced while exercising the self-hosted MCP server (built from the in-repo `services/mcp`). Our image pins the MCP submodule at `582b79d26` (2026-05-10); the defect reproduced there and was **re-verified against `master` at `6f89021e422466d47320f0b4d23d276d22ac530a` (2026-07-10)** — the offending schemas are unchanged.
- `PostHog/mcp` is archived; the MCP server now lives in this monorepo under `services/mcp`, so filing here.

### Steps to reproduce

Call any generated create/update tool that has an object/array body field, providing that field as a normal object, e.g.:
- `scheduled-changes-create` with a `payload` object (`{operation, value}`).
- `endpoint-create` / `endpoint-update` with a `query` object.
- `llma-evaluation-report-create` with `delivery_targets`.

Actual result: `500` with a generic `server_error`; the underlying exception is a variant of `'str' object has no attribute 'get'` (the API received the JSON-encoded string form of the object). Omitting the field (where optional) or sending a value the API happens to accept as a string is the only workaround.

### Expected vs actual

- **Expected:** object/array params advertise `type: "object"` / `type: "array"` in the tool `inputSchema`, so clients send structured JSON; the API receives an object.
- **Actual:** the param advertises `{}` (no type); clients send a string; the API rejects it.

### Code pointers (master `6f89021e`)

Representative untyped params (each is `<field>: zod.unknown()`):
- `payload` (scheduled-changes create/update): [`services/mcp/src/generated/feature_flags/api.ts#L1155-L1156`](https://github.com/PostHog/posthog/blob/6f89021e422466d47320f0b4d23d276d22ac530a/services/mcp/src/generated/feature_flags/api.ts#L1155-L1156) (also L1226-L1227).
- `query` (endpoint create/update): [`services/mcp/src/generated/endpoints/api.ts#L48-L49`](https://github.com/PostHog/posthog/blob/6f89021e422466d47320f0b4d23d276d22ac530a/services/mcp/src/generated/endpoints/api.ts#L48-L49) (also L121-L122).
- `delivery_targets` (evaluation-report create/update): [`services/mcp/src/generated/ai_observability/api.ts#L701-L702`](https://github.com/PostHog/posthog/blob/6f89021e422466d47320f0b4d23d276d22ac530a/services/mcp/src/generated/ai_observability/api.ts#L701-L702) (also L795-L796).

Others include `filters`, `value`, `config`, `job_inputs`, `secondary_metrics`, `test_account_filters`, `session_replay_config`, etc., in `cohorts`, `experiments`, `surveys`, `dashboards`, `cdp_functions`, `core`, `data_warehouse`, `warehouse_sources`, `workflows`, and more.

Server side, for the `scheduled-changes` example the string reaches [`products/approvals/backend/scheduled_changes.py#L59`](https://github.com/PostHog/posthog/blob/6f89021e422466d47320f0b4d23d276d22ac530a/products/approvals/backend/scheduled_changes.py#L59) (`build_scheduled_change_serializer_data(flag, payload)`), which then does dict access on the string.

### Suggested fix

In `tools/openapi-codegen`, stop mapping "no `type`" to `zod.unknown()` for request-body/query fields. Emit a schema that carries a structural type so clients send JSON:
- free-form object → `zod.record(zod.unknown())` or `zod.object({}).passthrough()` (both yield `type: "object"`);
- free-form array → `zod.array(zod.unknown())` (yields `type: "array"`).

Alternatively, post-process the generated tool `inputSchema` to inject `type: "object"`/`"array"` where a property has none, and/or type the underlying DRF serializer fields (e.g. `JSONField`) so the OpenAPI source carries a `type` in the first place.

### Notes

Because it degrades the tool *schema* (not just behavior), this affects every MCP client of PostHog (Cloud and self-hosted), not only this deployment. Fixing the codegen mapping in one place un-breaks all 184 params at once.

Sakshamm-Goyal:
## Problem

Closes #70064.

Unconstrained OpenAPI values are generated as `zod.unknown()`. The resulting MCP input schema advertises no JSON shape, so clients may stringify object or array arguments before calling a tool.

## Changes

- Convert unconstrained Orval output to Zod's recursive `zod.json()` schema during the MCP-specific postprocess step.
- Keep the OpenAPI source schema unconstrained and avoid hand-editing generated modules.
- Add regression coverage for both generated-source rewriting and the MCP JSON Schema exposed to clients.

- `pnpm --filter @posthog/mcp exec vitest run tests/unit/schema-transforms.test.ts tests/unit/tool-catalog.test.ts` – 29 tests passed, including object and array branches reachable through the emitted JSON Schema reference.
- `pnpm --filter @posthog/mcp exec oxfmt --check scripts/generate-orval-schemas.mjs scripts/lib/schema-transforms.mjs tests/unit/schema-transforms.test.ts tests/unit/tool-catalog.test.ts`.
- I also ran `tsgo --noEmit`; this partial checkout lacks unrelated UI/product source modules required by the full MCP project, so that repository-wide check could not complete. The targeted tests and formatter completed successfully.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing documentation change. The generator comment records the MCP-specific behavior.

## 🤖 Agent context

Codex co-authored this focused fix after reproducing the generated `zod.unknown()` output and inspecting the MCP schema exporter. I chose `zod.json()` rather than coercing all values to objects because the OpenAPI values are unconstrained and may legitimately be scalars, arrays, objects, or null. I used the repository-provided `/writing-tests` skill and added a regression that checks both the generator transform and the actual schema exposed by `toMcpInputSchema`.


rafaeelaudibert (changes requested):
It doesn't look like any of the generated files got changed by this?

Sakshamm-Goyal:
Regenerated and committed the affected MCP OpenAPI-derived input schemas in 8a7140cfa. The generator and its regression test now cover the unconstrained-value transformation, and the focused schema-transform suite passes (15 tests).

Sakshamm-Goyal:
Rebased onto current `master` and regenerated the MCP Orval schemas through `generate-orval-schemas` (rather than resolving generated outputs by hand).

Validation completed:
- `pnpm --filter=@posthog/mcp run generate-orval-schemas`
- `pnpm --filter=@posthog/mcp exec vitest run tests/unit/schema-transforms.test.ts tests/unit/tool-catalog.test.ts` — 29 passing
- focused `oxfmt --check`
- `bin/hogli ci:preflight --fix` (branch freshness check passed)

This includes the generated artifacts requested in the review and incorporates the latest upstream schema changes.

Sakshamm-Goyal:
Rebased onto current `master` and force-pushed `1aedd5c5b` after regenerating the MCP Orval schemas from the current OpenAPI input (the generated-file conflicts were resolved by the generator, not by hand).

Validation on the rebased head:
- `pnpm --filter=@posthog/mcp run generate-orval-schemas`
- focused schema-transform and tool-catalog suite — 29 passing
- focused `oxfmt --check`

The reviewed generated schema change remains present, now alongside the latest upstream schema updates.

PR #72487, opened 2026-07-21
fix(integrations): refresh salesforce sandbox tokens at instance_url

meikelmosby:
## Problem

Users on Salesforce Sandbox report that the integration disconnects within a few hours and reconnecting throws "Authentication token could not be refreshed. Please reconnect." — the connection is never durable.

The OAuth callback in `posthog/models/integration.py` already knows about sandbox: if the initial token exchange against `login.salesforce.com` fails, it retries against `test.salesforce.com` as a fallback and saves the resulting refresh token. But the integration is stored under `kind="salesforce"` regardless of which host issued the token, and the refresh path in `_post_token_refresh` blindly uses `oauth_config.token_url` — which is hardcoded to `https://login.salesforce.com/services/oauth2/token`. Salesforce rejects a sandbox-issued refresh token posted to the prod token endpoint, the sweep marks the integration as `ERROR_TOKEN_REFRESH_FAILED`, and the frontend renders the reconnect message. Access tokens live ~1–2 hours, which matches the "few hours to a day" cadence users see.

The revoke path in `unlink()` already handled this correctly by deriving the URL from `config["instance_url"]`; refresh just never got the mirror change.

## Changes

- In `_post_token_refresh`, when `kind == "salesforce"` and `config["instance_url"]` is set, POST the refresh at `{instance_url}/services/oauth2/token` instead of the static prod URL. Falls back to the hardcoded URL when `instance_url` is missing (older integrations that never captured it).
- Mirrors the existing revoke-time behavior at `posthog/models/integration.py:1275-1277`.
- Also fixes the same failure mode for custom-domain prod orgs (`*.my.salesforce.com`) that happened to work only when their custom domain also accepted refresh at `login.salesforce.com`.

Not changed in this PR (deliberate follow-ups):
- No user-facing "Sandbox" toggle in the OAuth start flow — the callback fallback still handles new sandbox connections. Worth revisiting so the intent is explicit instead of implicit, but out of scope here.

Added one Django `TestCase` in `posthog/models/test/test_integration_model.py`:

- `test_salesforce_refresh_uses_instance_url_for_sandbox` — creates a `salesforce`-kind integration whose `config["instance_url"]` is a sandbox host, calls `refresh_access_token()` with `requests.post` mocked, and asserts the POST went to `{instance_url}/services/oauth2/token`. Regression it catches: any future refactor of `_post_token_refresh` that reverts to the static prod URL will fail this test. No existing salesforce refresh test asserted the URL — the closest (`test_salesforce_refresh_access_token_without_expires_in_response`) only checked expiry defaults, so today's regression would ship green.

Ran locally against the fix:

- `posthog/models/test/test_integration_model.py::TestOauthIntegrationModel` — 55 passed (all 6 salesforce cases green, including the new one).

Did not test end-to-end against a real Salesforce sandbox — the mocked-boundary test covers the failure mode reported by the user.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated via the /Explore agent and direct code reads to trace the sandbox OAuth path across `oauth_config_for_kind`, `integration_from_oauth_response` (the sandbox-fallback "hack"), `unlink` (which already used `instance_url`), and `refresh_access_token` → `_post_token_refresh` (which did not). No skills invoked beyond `/writing-tests` for the gate on the new test. Chose to mirror the revoke path's exact shape rather than plumb a per-kind override through `OauthConfig`, because only Salesforce has the prod/sandbox split and only Salesforce stores `instance_url` — a targeted branch is smaller and matches how the revoke fix landed.

Tools: Explore agent (once), Bash/Read/Edit/Grep. No manual browser testing performed.

### DIFF
--- a/posthog/models/integration.py
+++ b/posthog/models/integration.py
@@ -503,11 +503,15 @@ def _salesforce_instance_host(instance_url: str | None) -> str | None:
         return None
     try:
         parsed = urlparse(instance_url)
+        # port/hostname/username/password are lazily parsed from netloc on access, and
+        # port in particular raises ValueError on a non-numeric or out-of-range value
+        # (e.g. https://host:abc/). Keep every derived-property read inside the try so a
+        # poisoned instance_url can never crash the refresh sweep.
+        if parsed.scheme != "https" or parsed.port is not None or parsed.username or parsed.password:
+            return None
+        host = (parsed.hostname or "").lower()
     except ValueError:
         return None
-    if parsed.scheme != "https" or parsed.port is not None or parsed.username or parsed.password:
-        return None
-    host = (parsed.hostname or "").lower()
     if not host.endswith(".salesforce.com"):
         return None
     return f"https://{host}"
--- a/posthog/models/integration.py
+++ b/posthog/models/integration.py
@@ -9,7 +9,7 @@ from collections.abc import Iterable, Mapping
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple, NoReturn, Optional, Self
-from urllib.parse import parse_qs, urlencode
+from urllib.parse import parse_qs, urlencode, urlparse
 
 from products.workflows.backend.providers import MAILDEV_MOCK_DNS_RECORDS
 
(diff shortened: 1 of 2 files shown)


mayteio (approved):
thanks @meikelmosby!

posthog/models/integration.py
```diff
@@ -490,6 +490,29 @@ def _build_posthog_slack_scope() -> str:
...
+    # even if a future endpoint or admin tool exposes config as writable. Returns
+    # "https://<host>" for a legitimate value, None otherwise (caller falls back to the
+    # hardcoded prod URL).
+    if not instance_url:
+        return None
+    try:
+        parsed = urlparse(instance_url)
+    except ValueError:
+        return None
+    if parsed.scheme != "https" or parsed.port is not None or parsed.username or parsed.password:
```
dmarchuk: Shouldn't this also be in the `try` clause? E.g. accessing the `port` can raise `ValueError` in case the port is invalid like string

dmarchuk (approved):

posthog/models/integration.py
meikelmosby: fixed. thx!

PR #68612, opened 2026-07-06
feat(dashboards): add test-account filter override

thmsobrmlr:
## TL;DR

Dashboards get a "…" button at the end of the edit bar that opens an "Advanced filters" panel. In it you can force "filter out internal and test users" on or off for every insight on the dashboard, or leave it on "Inherit" so each insight keeps its own setting. By default nothing changes. The override survives opening an insight from the dashboard, and each insight card's details popover shows it.

## Problem

Dashboard-level filters let a viewer override date range, properties, and breakdown across all insights on a dashboard, but there was no single control to toggle "filter out internal/test traffic" across the dashboard. Each insight carries its own `filterTestAccounts`, so a dashboard mixes filtered and unfiltered tiles.

This PR is the frontend half: the dashboard UI control. The schema fields, generated types, query-runner application, and backend tests shipped in #68754 (merged). The interval override that used to live in this PR moved to [#71063](https://github.com/PostHog/posthog/pull/71063).

## Changes

A "…" button after the existing edit bar options opens an "Advanced filters" popover panel, designed to host overrides too rarely used to earn a spot in the bar itself. It currently contains one control, test account filtering, as a tri-state segmented button:

| Option | Effect |
| --- | --- |
| Inherit (default) | Each insight keeps its own "Filter out internal and test users" setting, nothing changes |
| Filter out | Forces test account filtering on for every insight on the dashboard |
| Include | Forces test account filtering off for every insight on the dashboard |

Helper text under the control explains the active choice, a gear link jumps to the internal filter settings, and "Filter out" is disabled with a hint when the project has no test account filters configured. When an override is active, the "…" trigger shows a count badge.

The control is wired through a `setFilterTestAccounts` action into the existing intermittent → effective filter pipeline. Unsaved changes apply temporarily to the current view; saving persists them on the dashboard for all viewers. The key flows through the existing filter-combination utility, the refresh `filters_override` query param, the URL filter params, and the edit-mode save diff with no per-key plumbing, since those helpers are key-generic. Filter change reporting gains a `test_accounts` change type.

An earlier revision rendered the override as an inline indeterminate `TestAccountFilterSwitch` in the bar; after prototyping three shapes, the advanced-filters panel won, and the `TestAccountFilterSwitch` indeterminate/`onReset` additions were reverted since nothing uses them anymore.

Two pieces ported from the closed sibling exploration #71957:

[...]

PR #72491, opened 2026-07-21
chore(dashboards): prototype '…' advanced options on the dashboard edit bar

thmsobrmlr:
> [!WARNING]
> **Prototype, not for merge.** Throwaway UI exploration, rendered in dev builds only. It exists so the variants can be flipped through in the browser and one can be picked; the winner gets a proper implementation and this branch gets closed.

## TL;DR (eli5)

Dashboards can force a date range or properties onto all their insights, but not the "filter out internal and test users" setting.
This PR mocks up three different ways a "…" button at the end of the dashboard edit bar could let you override that setting for the whole dashboard.
Open a dashboard, add `?variant=A` (or B or C) to the URL, and use the floating black pill at the bottom of the screen (or ←/→ keys) to flip between them.

## Problem

The dashboard edit bar overrides date range, interval, properties, and breakdown across all insights, but there's no way to override test account filtering per dashboard.
The `DashboardFilter` schema already has an unused tri-state `filterTestAccounts` field (inherit / force on / force off).
Before building the real thing, we want to settle what the affordance should look like: the ask is a "…" button behind the existing options that opens "advanced options".

## Changes

Three structurally different variants of the "…" affordance, mounted on the existing dashboard route behind a `?variant=` search param, dev builds only:

| Variant | Shape |
| --- | --- |
| `A` | "…" opens a compact dropdown menu with three radio-style choices; a badge dot marks an active override |
| `B` | "…" opens an "Advanced filters" panel with a labelled segmented control, helper text, and a gear link to filter settings; designed to host future advanced options |
| `C` | "…" expands the bar inline with a `test accounts <select>` control echoing the existing "grouped by" idiom; collapses to a removable chip when an override is active |

A shared floating `PrototypeSwitcher` bar (bottom center) cycles variants via the URL param and arrow keys.
The override state is `useState` only; nothing is persisted or sent to the backend.

- `pnpm --filter=@posthog/frontend typescript:check` — no new errors introduced (identical pre-existing error count with and without this change, none in these files).
- `pnpm --filter=@posthog/frontend fix` (oxlint + oxfmt) — clean.
- I (Claude) did not manually flip through the variants in a running app; that browser pass is exactly what this PR exists for.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, throwaway prototype.

## 🤖 Agent context

- Authored by Claude via PostHog Code, using the user's `prototype` skill (UI branch: N radically different variants on the existing route, switchable via `?variant=`, floating switcher, in-memory state only).
- Decisions: reused the existing tri-state `filterTestAccounts` schema field as the model; kept everything LemonUI to match the surrounding edit bar (a production menu variant should use quill `DropdownMenu` per conventions); gated all prototype rendering on `NODE_ENV === 'development'` so a stray merge can't ship it.

---

### DIFF
--- a/frontend/src/lib/components/PrototypeSwitcher.tsx
+++ b/frontend/src/lib/components/PrototypeSwitcher.tsx
@@ -1,0 +1,85 @@
+/**
+ * PROTOTYPE TOOLING — floating variant switcher for throwaway UI prototypes.
+ * Renders nothing in production builds. Delete together with the prototype it serves.
+ */
+import { useValues } from 'kea'
+import { router } from 'kea-router'
+import { useEffect } from 'react'
+
+import { IconChevronLeft, IconChevronRight } from '@posthog/icons'
+
+export interface PrototypeVariant {
+    key: string
+    name: string
+}
+
+interface PrototypeSwitcherProps {
+    variants: PrototypeVariant[]
+    current: string
+    /** Extra live-state readout shown in the bar, e.g. the prototype's current selection. */
+    stateLabel?: string
+}
+
+export function PrototypeSwitcher(props: PrototypeSwitcherProps): JSX.Element | null {
+    if (process.env.NODE_ENV !== 'development') {
+        return null
+    }
+    return <PrototypeSwitcherBar {...props} />
+}
+
+function PrototypeSwitcherBar({ variants, current, stateLabel }: PrototypeSwitcherProps): JSX.Element {
+    const { location, searchParams, hashParams } = useValues(router)
+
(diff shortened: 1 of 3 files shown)


thmsobrmlr:
Verdict: variant B (advanced filters panel) won. It was implemented as production code in #68612, replacing that PR's earlier inline indeterminate-switch approach. This branch stays as the primary source for the three explored variants; do not merge.

*This comment was posted by PostHog Code on behalf of @thmsobrmlr.*

PR #72508, opened 2026-07-21
feat(batch-exports): execute hogql queries in internal stage activity

rossgray:
## Problem

Batch exports are limited to the fixed `events`/`persons`/`sessions` models. We want them driven by an arbitrary HogQL `SELECT`. #72020 added the model to store the query; this makes the Temporal worker run it.

Scope is the worker side only — the API layer (serializer, validation, feature flag) is a follow-up. Therefore, this code is currently not reachable externally.

## Changes

`HogQLQueryRecordBatchModel` transpiles a stored HogQL query at run time and wraps it in `INSERT INTO FUNCTION s3(...)` targeting the internal staging bucket, mirroring the sessions model. It's resolved via a new `"hogql"` branch that raises on a missing query rather than silently falling through to the events template.

Non-obvious decisions worth a look:

- **Settings and the log comment are sent as HTTP query-string parameters, not written into the query.** ClickHouse accepts settings either way, and sending them with the request means nothing has to manipulate a `SETTINGS` clause as a string. The first version did string-append and needed a regex to tell a trailing top-level clause from one inside a subquery — it broke on the persons lazy-table subquery. Worth keeping in mind for the resource limits in a later PR: a user's own `SETTINGS` clause still overrides settings sent this way, so hard limits will need something stronger.
    - this is now also true for the sessions model, which inherits these changes
- **Printing runs off the event loop.** Printing an arbitrary query can lazily read Postgres (the events table checks the new-events-schema instance setting), which raises `SynchronousOnlyOperation` in the async activity on a cold cache.
- **Only `ExposedHogQLError` is caught** and mapped to the typed `UnsupportedHogQLQueryError` (placeholder / unparseable queries). Internal engine errors propagate so they surface as real failures instead of being mislabeled a bad user query. This lives in a new DRF/Temporal-free module the upcoming API layer will share.
    - We may need to refine this over time
- The separate `prepare_query` commit fixes a brace-handling bug in the ClickHouse client's `str.format()` pass that would crash or mangle any parameter value containing braces (e.g. a JSON string constant). The fixed models never hit it, however arbitrary queries could trigger this.

The shared prepare/print flow was consolidated onto `RecordBatchModel`, so sessions and hogql differ only in `get_hogql_query` and a settings hook.

No manual testing — no real batch export was run end to end. This will come later once the API layer has been added.

Automated tests, run locally against the dev ClickHouse + object storage stack:

- **End-to-end stage tests**: run the activity with a user query and read staged Arrow back via the production producer, asserting exact rows and column names. Parametrized over events, persons, sessions, a join, a CTE+join and a UNION, plus a data-warehouse view. The UNION covers the `SelectSetQuery` shape that has no AST settings slot; the events case filters on a constant with `%` and braces to guard the `prepare_query` fix.
- **Settings and log comment**: asserted against `system.query_log` for every model, because neither appears in the SQL we generate, so only ClickHouse's own record of the query can confirm they arrived. Looking the query up by its log comment is what proves the log comment was set. Confirmed these fail if the settings stop being passed.
- **Unit tests**: query team-scoping, the rendering of request settings, and rejection of placeholder / unparseable / non-SELECT queries with the specific message. Confirmed passing in isolation, which is what surfaced the off-loop-printing bug.
- **`prepare_query`**: cases for values with `%`, lone braces, JSON, and a name colliding with another parameter.

The last commit adds a truncate fixture to the tests that insert ClickHouse data. Test data isn't scoped to the team that created it, so leftover rows could land on a later test's team and be counted as its own, which made the row-count assertions fail intermittently.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing surface yet (worker-only, no API), so no docs change.

## 🤖 Agent context

Written by Claude (PostHog Code), directed by @rossgray, following a phase-1 plan for HogQL-powered batch exports. Skill invoked: `/writing-tests`.

Deferred to later PRs: the API layer, per-query resource limits, and non-retryable failure classification.

---

### DIFF
--- a/products/batch_exports/backend/temporal/record_batch_model.py
+++ b/products/batch_exports/backend/temporal/record_batch_model.py
@@ -1,3 +1,4 @@
+import re
 import abc
 import uuid
 import typing
@@ -17,6 +18,7 @@ from posthog.sync import database_sync_to_async
 from posthog.temporal.common.clickhouse import get_client
 from posthog.temporal.common.logger import get_write_only_logger
 
+from products.batch_exports.backend.hogql_source import UnsupportedHogQLQueryError, parse_hogql_select_for_batch_export
 from products.batch_exports.backend.service import BatchExportModel, BatchExportSchema
 from products.batch_exports.backend.temporal import sql
 from products.batch_exports.backend.temporal.metrics import log_query_duration
@@ -28,13 +30,43 @@ QueryParameters = dict[str, typing.Any]
 BatchExportDateRange = tuple[dt.datetime | None, dt.datetime]
 
 
+def _print_setting_value(value: typing.Any) -> str:
+    """Render a setting value the same way the HogQL printer does."""
+    if isinstance(value, bool):
+        return "1" if value else "0"
+    return str(value)
+
+
+# A SETTINGS clause at the very end of the query, i.e. at the top level: anything after
+# a subquery's SETTINGS clause includes at least its closing parenthesis.
+TRAILING_SETTINGS_CLAUSE = re.compile(r"\bSETTINGS\s+[^()]*$", re.IGNORECASE)
+
+
+def append_settings_to_query(printed_query: str, settings_pairs: list[str]) -> str:
+    """Append a `SETTINGS` clause (given as `name=value` strings) to a printed query.
+
+    If the query already ends in a top-level SETTINGS clause (the printer merges in
+    settings required by some tables), extend it rather than open a second one. A
+    SETTINGS clause inside a subquery (e.g. from a lazy table expansion) doesn't count.
+    """
+    if not settings_pairs:
+        return printed_query
+    separator = ", " if TRAILING_SETTINGS_CLAUSE.search(printed_query) else " SETTINGS "
+    return printed_query + separator + ", ".join(settings_pairs)
+
+
 class RecordBatchModel(abc.ABC):
     """Base class for models that can be produced as record batches.
 
     Attributes:
        team_id: The ID of the team we are producing records for.
        batch_export_id: The ID of the batch export we are producing records for.
+       wait_for_data_interval_end: Whether to wait before querying until the data
+           interval end has passed and replication lag past it has settled. Models
+           without data interval semantics query "as of now" and skip the wait.
     """
+
+    wait_for_data_interval_end: bool = True
 
     def __init__(self, team_id: int, batch_export_id: str | None = None):
         self.team_id = team_id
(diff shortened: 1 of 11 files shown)


tomasfarias:
> but an arbitrary user query can parse to an ast.SelectSetQuery (a UNION), which has no settings slot.

Then put the settings on each select or add wrap the query in a select with settings? Either seems better than handling strings.

tomasfarias:
> Latent bug the fixed models never triggered

I hate this "model-speak": It was never triggered because our models are pre-defined! There is no "latent bug": We are adding new arbitrary models so code has to change, that doesn't imply a bug on the existing code.

> The separate prepare_query commit fixes a brace-handling bug[...] The fixed models never hit it; arbitrary queries do immediately.

Then it's not a bug! The code is written to handle fixed models. Why would not handling arbitrary parameters be bug if we never had those?!?

/rant

products/batch_exports/backend/temporal/record_batch_model.py
```diff
@@ -48,6 +63,11 @@ async def get_hogql_context(self) -> HogQLContext:
             team_id=team.id,
             enable_select_queries=True,
             limit_top_select=False,
+            # A bit of a hack: the query references neither half of this, but both are required:
```
rossgray, line 66: I would really like to update the way we handle ClickHouse parameters but that refactor is a bit out of scope here

rossgray:
@tomasfarias thanks for the initial review. I've reworked it now to pass the settings (including `log_comment`) as query parameters instead of the awkward string manipulation

products/batch_exports/backend/temporal/record_batch_model.py
tomasfarias: any specific ideas you have in mind? I would also like to refactor that

products/batch_exports/backend/temporal/record_batch_model.py
```diff
@@ -82,8 +141,14 @@ async def as_insert_into_s3_query_with_parameters(
...
         num_partitions: int,
     ) -> tuple[Query, QueryParameters]:
-        """Produce a printed query and any necessary ClickHouse query parameters."""
-        raise NotImplementedError
+        """Produce an `INSERT INTO FUNCTION s3(...)` query and its ClickHouse parameters."""
+        printed, parameters = await self._print_query(data_interval_start, data_interval_end, output_format=None)
+        s3_function = sql.get_s3_function_call(s3_folder, s3_key, s3_secret, num_partitions)
+        insert_query = f"""
+INSERT INTO FUNCTION {s3_function}
+{printed}
```
tomasfarias, line 149: I'd like it if we could operate on the AST here, but I think HogQL doesn't do inserts?

tomasfarias (approved):

products/batch_exports/backend/temporal/record_batch_model.py
rossgray: well, we have 3 different ways of parsing parameters:
1. printf interpolation
2. `str.format` interpolation
3. CH server-side binding

not sure if we can use CH server-side binding for all current usecases (I seem to remember it doesn't handle variables in certain expressions) but would be nice to unify this somehow. The code around all the existing string interpolation is pretty messy in any case.

The log comment stuff is also not ideal (we use context vars, which I guess is helpful for passing in some global tags, but it's not very explicit (eg we currently need to make sure we call `get_log_comment()` even if we don't use the result)

products/batch_exports/backend/temporal/record_batch_model.py
rossgray: yeah, it doesn't unfortunately

products/batch_exports/backend/temporal/record_batch_model.py
tomasfarias: Yeah, there should be ideally 1 way of passing parameters, and it should be server-side binding. We can take a look in a followup.

PR #72509, opened 2026-07-21
feat(stamphog): hand refused PRs to ReviewHog via trigger label

pauldambra:
## Problem

When stamphog refuses or escalates a PR, it means the change warrants a deeper look than the fast approval gate can give. Today that refusal just lands as a comment; nothing routes the PR to a second-opinion reviewer. ReviewHog exists for exactly that deeper review and is triggered by adding the `reviewhog` label, but no system wires the two together.

## Changes

- Stamphog now adds the `reviewhog` label on a `REFUSED`/`ESCALATE` verdict (`post_verdict`), handing the PR to ReviewHog. The handoff is mode-independent and best-effort: a 404/422 (label or repo unavailable) is swallowed in `StamphogGitHubClient.add_pr_label`, and any other failure is caught at the call site so the refusal verdict always lands.
- `review-hog.yml` exempts `stamphog[bot]` from its `bot-labeler-skip` (which would otherwise strip the bot-applied label) and routes it through the trigger job, so the handoff actually fires a ReviewHog review instead of being deleted. Humans and other bots are unaffected.

The Django integration tests need a live Postgres product database, which is not available in this task sandbox, so I could not execute the test suite here. What I verified:

- Added a case to `test_refused_verdict_strips_trigger_label_only_in_label_mode` asserting the `reviewhog` label is added on refusal in both review modes. Catches the regression of scoping the handoff to LABEL mode only (copying the existing strip guard) or dropping the call.
- Added a negative assertion to `test_signed_webhook_drives_review_and_posts_approval` that an `APPROVED` verdict never adds the label. Catches the condition widening to fire on approval.
- Added `test_refused_verdict_lands_even_when_reviewhog_handoff_fails` (422 and 500 parameterized). Catches removing the try/except or making the client raise on 422, which would flip `REFUSED` to `FAILED` over a missing secondary label.
- Confirmed all 203 stamphog tests collect, ruff check/format and mypy are clean on the changed files, and the `review-hog.yml` job truth-table routes exactly one job per sender type (human / `stamphog[bot]` / other bot).
- Exercised `add_pr_label` status handling directly: 404/422 swallowed, 500 raises, 200/201 accepted.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Produced by Claude Code (PostHog Code). `/writing-tests` and `/authoring-ci-workflows` were loaded before finalizing the tests and the workflow edit.

Key decision: the literal request was "set the reviewhog label." A bare label-add by a bot is self-defeating under the existing `review-hog.yml`, whose `bot-labeler-skip` job strips bot-applied `reviewhog` labels. So the change adds the label and exempts the sanctioned stamphog bot identity from that skip, routing it through the trigger job instead. The handoff is made best-effort so a missing label (a vendored stamphog in another repo, or a repo without ReviewHog) or a transient GitHub error can never turn a delivered refusal into a failed run.

---


pauldambra:
Addressed the ReviewHog must-fix finding (handoff `except` narrower than what `add_pr_label` can raise):

- Broadened the catch at the call site to `except Exception`. The handoff is explicitly best-effort and runs before `post_verdict`'s durable terminal save, so anything that escapes skips the persist, exhausts the activity retries, and lands at `mark_review_failed` — a delivered refusal rewritten to FAILED with no persisted verdict. The narrow `except StamphogGitHubError` missed `GitHubRateLimitError` (429 / rate-limited 403, raised by the egress layer) and `requests.RequestException` (network blip), neither a subclass of `StamphogGitHubError`.
- Reconciled the comment: dropped the false "transient 5xx benefits from the activity retry" claim (a 500 raises `StamphogGitHubError`, which was already swallowed single-shot) and stated the single-shot best-effort behavior plainly.
- Added two regression cases to `test_refused_verdict_lands_even_when_reviewhog_handoff_fails` that raise `GitHubRateLimitError` and `requests.ConnectionError` directly from the fake's label-add path. The chain stubs `raise_if_github_rate_limited` to a no-op, so the rate-limit path was otherwise unreachable in tests — a 500 only ever raised the already-caught `StamphogGitHubError`, leaving the gap invisible. Both new cases fail against the narrow `except` and pass against the broadened one (verified statically: `isinstance(GitHubRateLimitError(...), (StamphogGitHubError,))` is `False`).

mypy and ruff clean on the changed files; all 60 integration tests collect. The Django DB tests still can't be executed in this sandbox (no Postgres), as noted in the PR description.

pauldambra:
Addressed two of the four Codex review findings on the reviewhog handoff; deferred two as design-level (out of scope for this CI-fix turn).

**Fixed (code-correctness, in scope):**

- **P1 — Avoid raising while logging a failed handoff.** Confirmed: `activity.logger` is a `temporalio.activity.LoggerAdapter` over a stdlib `Logger`, and `.exception(msg, repo=, pr_number=)` raises `TypeError: Logger._log() got an unexpected keyword argument "repo"`. That raise happened *inside* the `except Exception` block, so it escaped and skipped the terminal verdict save — re-breaking the exact invariant the block protects. Fixed by formatting repo/pr into the message (f-string), matching every other `activity.logger` call in the file.

- **P2 — Defer the handoff until the refusal wins supersession.** The handoff ran before the conditional terminal save, so a refusal that lost the save to a supersession still triggered ReviewHog for a stale refusal a newer run might approve. Moved the handoff to after the `if not updated: return skipped_superseded` check, alongside the digest-stamp and orphan-sweep that already gate on a durable commit. Added `test_superseded_refusal_does_not_hand_off_to_reviewhog` which flips the run to SUPERSEDED during the sticky post and asserts the handoff never fires (reaches the terminal-save early return, not the top guard).

**Deferred (design/feature, need a human call — not a regression in this diff):**

- **P1 — Re-trigger ReviewHog for subsequent refusals.** A second refusal on the same PR leaves the `reviewhog` label in place, so re-adding it fires no `labeled` event and no second ReviewHog review runs. Real gap, but fixing it needs a label-lifecycle change (remove-then-add, or a different trigger) that touches `.github/workflows/review-hog.yml` — out of scope for a CI-fix turn per the constraints. Flagging for a follow-up.

- **P2 — Prevent the `reviewhog` label retriggering Stamphog.** Only matters if a repo misconfigures `trigger_label="reviewhog"`; the per-PR cooldown largely mitigates it. Edge case requiring misconfiguration; deferred.

Also root-caused and fixed the failing **Python code quality** check: a pre-existing master breakage (the webflow test refactor dropped the `TRANSPORT` constant but left one `patch(TRANSPORT)` call, ruff F821) that blocked this PR after it rebased onto master. Fixed the missed call to `WEBFLOW_SESSION_PATCH` (matching its sibling `validate_credentials` tests); 32/32 webflow tests pass. The branch is now rebased onto master (was 16 behind) and `ruff check .`, `ruff format --check`, and `ty` all pass.

Stamphog's REFUSED verdict on this PR is the deterministic `infra_cicd` deny-list gate firing on the `review-hog.yml` edit — expected for a CI-workflow change, not a content issue; leaving for a human.

products/stamphog/backend/tests/test_integration.py
```diff
@@ -1058,6 +1065,106 @@ def test_refused_verdict_strips_trigger_label_only_in_label_mode(
...
+        ),
+    ],
+)
+@pytest.mark.django_db(databases=PRODUCT_DATABASES)
+def test_refused_verdict_lands_even_when_reviewhog_handoff_fails(
+    team,
+    stamphog_chain: StamphogChain,
+    failure: tuple[Literal["response"], int] | tuple[Literal["side_effect"], Exception],
+) -> None:
+    # The ReviewHog handoff is a secondary, cross-product notification that runs BEFORE the durable
```
pauldambra: > [!NOTE]

**[router]** 🟢 LOW

This comment states the ReviewHog handoff "runs BEFORE the durable terminal save, so any exception it raises must be caught or the refusal is lost." That's backwards. In `post_verdict` (activities.py) the conditional terminal save (`.exclude(status=SUPERSEDED).update(...)` + `if not updated: return`) happens **first**, and `add_pr_label` only runs afterward — the code comment at activities.py:785 correctly says "only AFTER the refusal verdict wins the terminal save above." By the time the handoff runs the refusal is already committed, so an uncaught exception here would not lose the refusal; it would fail the activity and risk a retry re-running side effects. Looks like a leftover from before the ordering was hardened. Worth correcting so a future refactorer doesn't infer the wrong ordering constraint.

pauldambra (commented):
QA Swarm review complete. See inline comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated security-audit on the workflow change

## Verdict: ✅ APPROVE <sub>(round 2 @ beb4b30)</sub>

Round 1's one LOW nit (stale test comment) is fixed, plus a behavior-preserving simplify pass. Round 2 re-review found no correctness, security, or ordering defects.

### Key findings

None this round. Round 1's stale-comment nit was corrected in dabd90f8.

### Convergence

None.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | LOW danger, HIGH confidence. Re-reviewed the incremental refactor (is_refusal hoist, docstring/comment trims, test-param flatten) — verified behavior-preserving: both `if` sites evaluate the identical boolean, the `isinstance(fault, Exception)` dispatch preserves the exact 422/500→response, rate-limit/network→side_effect mapping, and no load-bearing rationale was lost. No delegation. |

<details>

round 1 @ d1df2ff — 💬 APPROVE WITH NITS: one LOW stale-comment finding (test comment had the handoff/terminal-save ordering backwards); security-audit cleared the pull_request_target + stamphog[bot] carve-out with no findings. Fixed in dabd90f8.

</details>

---

sortafreel (approved):

PR #72521, opened 2026-07-21
chore(funnels): prototype dual-axis options for compare-mode steps chart

thmsobrmlr:
## TL;DR (eli5)

When a funnel compares two time periods, both periods share one percent axis, and only the bigger period's first bar touches 100% — the smaller one stops short, which reads as "this period didn't start at 100%?". This PR is a **throwaway prototype** that lets you flip between three ways of giving each period its own axis, right on a real funnel insight, so we can pick one before building it properly. **Not meant to be merged.**

## Problem

With "compare against" enabled, the steps chart scales both periods against the larger period's entrants, so the smaller period's first step reads below 100% on the single percent axis. Question raised: should the chart have two axes so each period gets its own 100% reference — and what should each axis show?

## Changes

A dev-only, throwaway prototype (per the `/prototype` skill) mounted inside the existing `FunnelStepsBarChart`:

- `prototype/FunnelCompareAxesPrototype.tsx` — all throwaway code: variant data transforms, an overlay that draws the extra axes, and a floating bottom pill to flip variants (arrows or `←`/`→` keys). Gated to dev builds and pure compare funnels; returns `null` otherwise so production rendering is byte-identical.
- `prototype/PROTOTYPE.md` — the question, the variants, what to judge, and the fold-in plan.
- `FunnelStepsBarChart.tsx` — a clearly-marked graft that swaps in the prototype series/config when active.

To try it: open any steps funnel with **Compare to previous period** on in a dev build, then flip via the pill or `?funnel_axes_variant=live|A|B|C`:

| Key | Bars | Axes |
| --- | --- | --- |
| `live` | Volume-true (production) | One shared percent axis |
| `A` | Each period normalized to its own entrants | Twin 0–100% axes, color-coded per period |
| `B` | Unchanged from live | Right axis: smaller period's own 0–100% compressed to its entry level |
| `C` | As in A | Axes labeled in absolute entrant counts per period |

- `pnpm --filter=@posthog/frontend typescript:check` — no errors in the touched files (remaining errors are pre-existing unbuilt-quill-package noise in this environment, identical before/after).
- `pnpm --filter=@posthog/frontend fix` — lint and format clean.
- I (Claude) could not run the app or a browser in this environment, so the variants have not been viewed rendered — that's the point of this PR: check out the branch and flip through them.
- No tests added on purpose: prototype code, gated out of production and test builds.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — dev-only throwaway prototype, no user-facing behavior change.

## 🤖 Agent context

Authored by Claude (PostHog Code) via the `/prototype` skill, from Thomas's question about adding two axes to the compare-mode funnel steps chart. Decisions: kept all throwaway code in a `prototype/` folder next to the chart with a minimal graft; drew the extra axes as a `useChartLayout` overlay instead of extending `@posthog/quill-charts` (the core `ChartConfig.yAxes` already supports per-series axes — that's the path for the real implementation once a variant wins); avoided a `react-dom` portal because `product_analytics` doesn't declare that dependency and adding it for a prototype wasn't worth lockfile churn.

---

### DIFF
--- a/products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/FunnelStepsBarChart.tsx
+++ b/products/product_analytics/frontend/insights/funnels/FunnelStepsBarChart/FunnelStepsBarChart.tsx
@@ -22,6 +22,8 @@ import {
     resolveFunnelStepClick,
     type FunnelStepsBarSeriesMeta,
 } from './funnelStepsBarTransforms'
+// PROTOTYPE (throwaway, dev-only) — compare-mode dual-axis exploration, see ./prototype/PROTOTYPE.md
+import { useFunnelCompareAxesPrototype } from './prototype/FunnelCompareAxesPrototype'
 
 const BASE_STEP_WIDTH_PX = 240
 const PER_BAR_WIDTH_PX = 20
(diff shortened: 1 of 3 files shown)


thmsobrmlr:
Verdict: **variant B won** — volume-true bars plus a second per-period percent axis (with two adjustments: axes fixed as left = current / right = previous, and default axis label color instead of per-series tinting). Real implementation: [#72546](https://github.com/PostHog/posthog/pull/72546). This prototype PR stays unmerged as the primary source of the explored variants.

*(Comment posted by PostHog Code on behalf of Thomas.)*

PR #72526, opened 2026-07-21
feat(surveys): add duplicate question action to editor rows

pawel-cebula:
> Test-only draft PR for validating internal review tooling. Not intended for merge.

## Problem

When building longer surveys, users often want to duplicate an existing question with the same type, options, description, and validation, then tweak only the copy.

## Changes

Adds a Duplicate question action to each survey question row.

Expected behavior:

- Clicking Duplicate on Question N inserts the copy directly after Question N.
- The copy keeps the original's type, options, and settings, with " (copy)" appended to the question title.
- The editor focuses the newly created copy.

Not manually tested; this PR is intended as a review target.

## Publish to changelog?

no

## Docs update

skip-inkeep-docs

### DIFF
--- a/frontend/src/scenes/surveys/surveyLogic.tsx
+++ b/frontend/src/scenes/surveys/surveyLogic.tsx
@@ -1544,6 +1544,7 @@ export const surveyLogic = kea<surveyLogicType>([
         archiveSurvey: true,
         setWritingHTMLDescription: (writingHTML: boolean) => ({ writingHTML }),
         setSelectedPageIndex: (idx: number | null) => ({ idx }),
+        duplicateQuestion: (index: number) => ({ index }),
         setSelectedSection: (section: SurveyEditSection | null) => ({ section }),
         resetTargeting: true,
         resetSurveyAdaptiveSampling: true,
@@ -2234,6 +2235,22 @@ export const surveyLogic = kea<surveyLogicType>([
                 } finally {
                     actions.setGeneratingTranslationDrafts(false)
                 }
+            },
+            duplicateQuestion: ({ index }) => {
+                const original = values.survey.questions[index]
+                if (!original) {
+                    return
+                }
+                const copy = {
+                    ...original,
+                    question: `${original.description ?? ''} (copy)`,
+                }
+                actions.setSurveyValue('questions', [
+                    ...values.survey.questions.slice(0, index + 1),
+                    copy,
+                    ...values.survey.questions.slice(index + 1),
+                ])
+                actions.setSelectedPageIndex(index + 1)
             },
             resetTargeting: () => {
                 actions.setSurveyValue('linked_flag_id', NEW_SURVEY.linked_flag_id)
--- a/frontend/src/scenes/surveys/SurveyEditQuestionRow.tsx
+++ b/frontend/src/scenes/surveys/SurveyEditQuestionRow.tsx
@@ -5,7 +5,7 @@ import { CSS } from '@dnd-kit/utilities'
 import { useActions, useValues } from 'kea'
 import { Group } from 'kea-forms'
 
-import { IconPlusSmall, IconTrash, IconWarning } from '@posthog/icons'
+import { IconCopy, IconPlusSmall, IconTrash, IconWarning } from '@posthog/icons'
 import { LemonButton, LemonCheckbox, LemonDialog, LemonInput, LemonSelect, LemonTag, Tooltip } from '@posthog/lemon-ui'
 
 import { LemonField } from 'lib/lemon-ui/LemonField'
@@ -57,7 +57,7 @@ export function SurveyEditQuestionHeader({
     translationErrorsByQuestion,
 }: SurveyQuestionHeaderProps): JSX.Element {
     const { editingLanguage } = useValues(surveyLogic)
-    const { removeQuestion } = useActions(surveyLogic)
+    const { duplicateQuestion, removeQuestion } = useActions(surveyLogic)
     const { setNodeRef, attributes, transform, transition, listeners, isDragging } = useSortable({
         id: index.toString(),
     })
@@ -96,6 +96,17 @@ export function SurveyEditQuestionHeader({
                         <IconWarning className="text-warning" />
                     </Tooltip>
                 )}
+                <LemonButton
+                    icon={<IconCopy />}
+                    size="xsmall"
+                    data-attr={`duplicate-survey-question-${index}`}
+                    tooltip="Duplicate question"
+                    onClick={(e) => {
+                        e.stopPropagation()
+                        duplicateQuestion(index)
+                    }}
+                    tooltipPlacement="top-end"
+                />
                 {survey.questions.length > 1 && (
                     <LemonButton
                         icon={<IconTrash />}


pawel-cebula:
## PostHog QA Frontend Report

**FAIL** · 1/4 · 14m39s · commit `b65eea8` · 2 reported, no autonomous fixes (report-only run)

> Exercised the new Duplicate question action in the survey editor. Insertion position, focus, and copied type/options all behave as described, but the copy's title is built from the description instead of the question title, and the copy keeps the original's question id - saving persists two questions with the same id.

**Coverage**

| Target         | Action                                                                                                          | Result |
| -------------- | --------------------------------------------------------------------------------------------------------------- | ------ |
| `/surveys/new` | Duplicated Q1 (title + description) - copy titled "Rate your overall experience (copy)"                          | FAIL   |
| `/surveys/new` | Duplicated a single-choice question with no description - copy titled " (copy)"                                  | FAIL   |
| `/surveys/new` | Checked insert position and focus - copy lands directly after the original, panel expanded, type/choices copied  | PASS   |
| `/surveys/:id` | Duplicated saved questions and saved - copies persist the originals' question ids                                | FAIL   |

**Setup**

- Stack: Coder devbox running the PR branch at `b65eea8`, forwarded to `localhost:8010`
- Workspace: seed dev user in the local demo project (dummy data)
- Data created: draft survey "QA run pr72526 duplicate question" with duplicated questions, persisted and left in place for debugging
- Overrides: none (default flags, light theme)
- Degraded: session-replay ingest 502s, dev notifications service down, and HogQL 500s from a crashed local ClickHouse migration - all pre-existing local noise; no new console errors from the duplicate flow itself

**Effort saved** · 1 High + 1 Medium caught · ~20 min of manual QA

---

### Finding 1 · reported, no autonomous fix

**`Duplicated question's title is built from the description, not the question title`**

```text
severity   █████░░░░░   MEDIUM
```

`frontend/src/scenes/surveys/surveyLogic.tsx` · `duplicateQuestion` listener

**Suggested patch** (not auto-applied: report-only run)

```diff
-                    question: `${original.description ?? ''} (copy)`,
+                    question: `${original.question} (copy)`,
```

**Failing step**

```text
url:      /surveys/new
action:   clicked Duplicate on Q1 "How satisfied are you with Hedgebox?" (description "Rate your overall experience")
expected: copy titled "How satisfied are you with Hedgebox? (copy)"
got:      copy titled "Rate your overall experience (copy)"; with no description the title is just " (copy)" and the original title is lost
```

Reproduced twice per variant (with and without a description). `question` is the question's title field everywhere else in the surveys code, and existing duplicate flows derive a copy's name from the original's name, so the copy should be titled from `original.question`.

**Evidence**

- flow
- still
- still

---

### Finding 2 · reported, no autonomous fix

**`Copy keeps the original's question id - saving persists two questions with the same id`**

```text
severity   ████████░░   HIGH
```

`frontend/src/scenes/surveys/surveyLogic.tsx` · `duplicateQuestion` listener

**Suggested patch** (not auto-applied: report-only run)

```diff
                 const copy = {
                     ...original,
+                    id: undefined,
                     question: `${original.description ?? ''} (copy)`,
                 }
```

**Failing step**

```text
url:      /surveys/:id (draft saved first so questions have persisted ids)
action:   clicked Duplicate on a saved question, then saved the survey
expected: the copy gets its own question id on save
got:      GET the survey - original and copy share the same id (reproduced for an open and a single_choice question)
```

Questions API excerpt after save:

```json
[
  { "id": "0da2a8c4-f78b-40a8-87cb-aeeb507b4246", "question": "How satisfied are you with Hedgebox?" },
  { "id": "0da2a8c4-f78b-40a8-87cb-aeeb507b4246", "question": "Rate your overall experience (copy)" },
  { "id": "3113d208-2a0e-4cdc-a1f6-da04c544087f", "question": "Which feature do you use most?" },
  { "id": "3113d208-2a0e-4cdc-a1f6-da04c544087f", "question": " (copy)" }
]
```

The backend pre-save hook only assigns ids to questions that are missing one, and its contract is that every question has a unique id. Both existing duplicate flows strip question ids before saving (`duplicateExistingSurvey` in `frontend/src/scenes/surveys/utils.ts` and the duplicate-to-project path). Question ids key survey responses and branching, so answers to the original and the copy become indistinguishable once the survey collects responses.

**Evidence**

- still

<sub>PostHog QA Frontend Report</sub>

frontend/src/scenes/surveys/surveyLogic.tsx
```diff
@@ -2235,6 +2236,22 @@ export const surveyLogic = kea<surveyLogicType>([
                     actions.setGeneratingTranslationDrafts(false)
                 }
             },
+            duplicateQuestion: ({ index }) => {
+                const original = values.survey.questions[index]
+                if (!original) {
+                    return
+                }
+                const copy = {
```
pawel-cebula, line 2244: 🟡 P1: The copy is a shallow spread, so `copy.branching` is the same object as the original's. `setResponseBasedBranchingForQuestion` mutates that object in place (the `delete question.branching.responseValues[...]` and assignments around L2685-2691), so editing response-based branching on either question after duplicating silently changes the other one too. `structuredClone(original)`, or dropping `branching` from the copy, avoids this and pairs naturally with the `id` fix.

frontend/src/scenes/surveys/surveyLogic.tsx
```diff
@@ -2235,6 +2236,22 @@ export const surveyLogic = kea<surveyLogicType>([
...
+                    ...original,
```
pawel-cebula, line 2245: 🔴 P1: This spread carries over the original's `id`, so after saving the survey has two questions with the same id. Responses are keyed `$survey_response_<id>` and results bucket by `question.id`, so answers to the two questions collide once the survey is live. The backend's `ensure_question_ids` only fills in missing ids, it never dedupes. `duplicateExistingSurvey` in `utils.ts` strips ids for exactly this reason - the copy here needs `id: undefined` too, so the backend assigns a fresh one on save.

frontend/src/scenes/surveys/surveyLogic.tsx
```diff
@@ -2235,6 +2236,22 @@ export const surveyLogic = kea<surveyLogicType>([
...
+                    question: `${original.description ?? ''} (copy)`,
```
pawel-cebula, line 2246: 🔴 P1: The copy's title is built from `description`, not `question`. The duplicate ends up titled with the description text, or just " (copy)" for questions that have no description.

```suggestion
                    question: `${original.question} (copy)`,
```

frontend/src/scenes/surveys/surveyLogic.tsx
```diff
@@ -2235,6 +2236,22 @@ export const surveyLogic = kea<surveyLogicType>([
...
+                }
+                actions.setSurveyValue('questions', [
```
pawel-cebula, line 2248: 🔴 P1: Inserting mid-array shifts every question after `index`, but branching targets are stored as numeric indices. That's why `moveQuestion` and `removeQuestion` run `remapBranchingIndices` with an index map (#58510) - without the same treatment here, duplicating in front of a `SpecificQuestion` or response-based target leaves those references pointing one question off. This needs an insert index map (indices above the insertion point shift up by one) piped through `remapBranchingIndices` before `setSurveyValue`, mirroring the move/delete paths.

pawel-cebula (commented):
Left a few comments. One thing that doesn't belong to a single line: worth covering `duplicateQuestion` in `surveyLogic.test.ts` alongside the fixes. Sibling actions like move/remove aren't tested either, so this isn't breaking a precedent, but a small `expectLogic` block would have caught the title bug.

pawel-cebula:
Assessment of the two reviews above. Context: this is a test PR with a planted defect, reviewed independently by a runtime QA run ([report](https://github.com/PostHog/posthog/pull/72526#issuecomment-5033158871)) and a static review ([bundled review](https://github.com/PostHog/posthog/pull/72526#pullrequestreview-4743861764)). Neither reviewer knew what was planted. I checked each finding against the planted defect and the code.

**Title built from `description` (QA finding 1, review P1.1)** - agree, and this is the planted bug. Both reviewers found it and both suggested the same correct one-line fix. On severity I side with the static review's 🔴 must-fix over the QA run's medium: the copy's title is silently wrong for every user of the feature, and empty when the question has no description.

**Copy keeps the original's question id (QA finding 2, review P1.2)** - agree, strongest finding of the run. This was not planted; both reviewers found it independently, which is good signal it's real. The runtime run proved it end to end (saved survey, API returns two questions with the same id) and the static review explained the blast radius (responses key on `$survey_response_<id>`, so answers to original and copy collide once the survey is live). Must fix with `id: undefined` on the copy - agree.

**Missing branching-index remap on insert (review P1.3)** - agree. Verified: `remapBranchingIndices` exists with tests, `moveQuestion`/`removeQuestion` remap and this new mid-array insert does not, so `SpecificQuestion` targets after the insert point drift by one. The runtime run missed this because it never exercised branching, and its coverage table honestly didn't claim it. Must fix - agree.

**Copy aliases the `branching` object (review P1.4)** - agree. Verified the in-place mutation: `setResponseBasedBranchingForQuestion` deletes and assigns keys on `question.branching.responseValues` directly, so with the shallow spread, editing branching on either question corrupts the other. Fixing it alongside P1.2 makes sense.

**No tests (review P3, top-level ask)** - agree with the validator's downgrade from P0, since the sibling actions are equally untested. If the branching remap fix lands, that logic is exactly the kind that earns a unit test.

**`validateDOMNesting` button-in-button** - the QA run saw this console error and discounted it as pre-existing, because the delete button on master already nests inside the collapse header button. I agree with the discount, and it corrects an earlier report of mine that charged the same warning to a sibling test PR. The nesting pattern deserves a cleanup someday, but not on this PR's bill.

Net: zero false positives across both reviews, the planted bug was caught by both, and each reviewer contributed findings the other could not reach - runtime persistence proof on one side, branching logic hazards on the other.

PR #72532, opened 2026-07-21
feat(conversations): add personal favorites to saved ticket views

luke-belton:
## Problem

Support has saved ticket views but no way to pin the ones you use most. The list is ordered newest-first, so a frequently-used view drifts down as you add more. This adds simple personal favorites so your go-to views stay at the top and are one click away.

https://github.com/user-attachments/assets/533b455a-5478-4769-8fbc-a1644c1a8617

## Changes

Favorites are personal to each user — a view you star floats to the top only for you — matching the saved insights star UX for consistency.

- New per-user `TicketViewFavorite` join model (team, user, ticket_view) with a migration.
- The ticket view serializer gets a writable `is_favorited` field. The list queryset annotates per-user favorite state and orders favorites first, then newest-first. `is_favorited` is also honored on creation, so a client can create an already-favorited view in one request.
- In the "Saved views" modal, each row gets a heart toggle. Favorited rows sort to the top. The heart shows a loading state while the toggle request is in flight.
- The "Saved views" button is now a dropdown: it lists your favorited views for quick loading, plus an "All saved views" item that opens the existing modal, and a "Clear view and reset filters" item when a view is active. Built with quill's `DropdownMenu` and a `LemonButton` trigger; views load lazily when the dropdown opens.

There is no reusable favorites abstraction in the repo (insights/dashboards use a team-wide `favorited` boolean; `FileSystemShortcut` is sidebar-pinning only), so this follows the established bespoke-join-model pattern.

Added backend API tests in `test_ticket_views.py`, each guarding a distinct regression:
- favoriting via `PATCH {is_favorited: true}` creates exactly one favorite row and is idempotent; unfavoriting removes it
- creating with `is_favorited` true/false persists the favorite (or doesn't)
- favorites are personal — one user's favorite does not set `is_favorited` for another user
- favorited views sort to the top of the list for the requesting user

CI is the source of truth for the full suite here: the conversations backend product tests, frontend typecheck/typegen, semgrep (IDOR coverage), and OpenAPI type validation all run and pass. Locally I ran and passed `ruff` and repo-config `mypy` on the changed backend modules; the local environment has no database or `node_modules`, so the Django/Jest suites and Orval codegen ran in CI (which auto-commits regenerated OpenAPI types).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude Code (Opus 4.8) at Luke Belton's direction. Skills invoked while shaping this change: `/django-migrations` (new model + migration), `/improving-drf-endpoints` (serializer/viewset), `/adopting-generated-api-types` (frontend generated client), and `/writing-tests` (test gate).

Key decisions:
- "Personal" was the explicit requirement, so favorites are per-user via a join model rather than a team-wide boolean like `Insight.favorited`.
- Toggling and favoriting-on-create reuse the existing POST/PATCH routes (a writable `is_favorited` handled in the serializer) rather than a bespoke action, keeping the surface small. The favorite `get_or_create` is team-scoped in its lookup to satisfy the IDOR rule now that the model is covered by it.
- The favorite toggle suppresses the "View updated" toast since it is a quiet, high-frequency action.
- The saved-views dropdown uses quill's `DropdownMenu` (per the repo preference over `LemonMenu`) with a `LemonButton` trigger, and moves the clear-view action into the menu (a `sideAction` inside a dropdown trigger would fight the menu open).

---


products/conversations/backend/api/ticket_views.py
```diff
@@ -42,10 +52,28 @@ class Meta:
...
 
+    def _set_favorited(self, instance: TicketView, favorited: bool) -> None:
+        user = self.context["request"].user
+        if favorited:
+            TicketViewFavorite.objects.get_or_create(ticket_view=instance, user=user, defaults={"team": instance.team})
+        else:
+            TicketViewFavorite.objects.filter(ticket_view=instance, user=user).delete()
+
     def create(self, validated_data: dict[str, Any], *args: Any, **kwargs: Any) -> TicketView:
+        validated_data.pop("is_favorited", None)
```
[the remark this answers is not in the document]
luke-belton: fixed - you can now favorite on create.

products/conversations/frontend/components/SavedViews/ticketViewsLogic.ts
```diff
@@ -179,6 +191,16 @@ export const ticketViewsLogic = kea<ticketViewsLogicType>([
...
             viewUpdated: (state, { view }) => state.map((v) => (v.short_id === view.short_id ? view : v)),
         },
+        favoritingShortIds: [
+            [] as string[],
+            {
+                toggleFavorite: (state, { view }) =>
+                    state.includes(view.short_id) ? state : [...state, view.short_id],
+                viewUpdated: (state, { view }) => state.filter((id) => id !== view.short_id),
+                // updateView clears in-flight state via loadViews() on failure
+                loadViews: () => [],
```
abigailbramble: Minor: this clears `favoritingShortIds` on the `loadViews` **trigger** action, and `loadViews` also fires every time the header dropdown opens (`onVisibilityChange` in `SavedViewsButton`). So if a favorite PATCH is in flight and the user closes the modal and opens the dropdown before it resolves, the heart's loading/disabled state clears early and it can be clicked again.

Cosmetic only (the backend toggle is `get_or_create`/`delete`, so overlapping requests are harmless), but clearing on `loadViewsSuccess`/`loadViewsFailure` instead of the trigger would keep the in-flight state accurate.

products/conversations/frontend/components/SavedViews/SavedViewsButton.tsx
```diff
@@ -1,38 +1,62 @@
...
-                        ? {
-                              icon: <IconX />,
-                              onClick: resetFilters,
-                              tooltip: 'Clear view and reset filters',
-                          }
-                        : undefined
-                }
+            <LemonMenu
+                placement="bottom-start"
+                onVisibilityChange={(visible) => visible && loadViews()}
```
abigailbramble, line 20: Minor: opening the dropdown fires `loadViews()` here, and clicking "All saved views" fires `openModal`, whose listener also calls `loadViews()` — so opening the modal from the dropdown issues two back-to-back GETs for the same list. Could skip one of them, e.g. don't reload in `openModal` when a load is already in flight.

abigailbramble (commented):
Generally looks really good!

A few minor, non-blocking findings from a review pass (all the scarier-looking candidates — the `LemonMenu`-wrapped `sideAction`, serializer paths hitting an unannotated `is_favorited`, the `(team_id, user)` index — checked out fine on verification).

abigailbramble (approved):

PR #72539, opened 2026-07-21
feat(dashboards): async-refresh stale dashboard tiles instead of blocking

pauldambra:
## Problem

When a dashboard loads, the frontend refreshes each stale tile with `refresh=blocking` (`getInsightWithRetry` in `dashboardLogic`), recomputing stale tiles synchronously, per tile, on every open. Route-level cost attribution shows this per-tile refresh (the `insights/{pk}` + `from_dashboard` path) is the single largest slice of dashboard insight recompute cost — and it multiplies across concurrent viewers of the same dashboard, which is how a handful of always-open dashboards dominate the bill.

## Changes

- Behind a new `DASHBOARD_ASYNC_TILE_REFRESH` flag, the per-tile dashboard-load refresh becomes `async_except_on_cache_miss`: serve the recent cached result immediately, refresh a stale tile in the background, and still **block on a true cache miss** so a cold tile renders data instead of an empty state. The manual single-insight refresh keeps `force_blocking` (a person asking for fresh data still gets it).
- Add `async_except_on_cache_miss` to the refresh values `refresh_requested_by_client` accepts. It was already an `ExecutionMode` value but wasn't requestable over the wire, so the frontend couldn't ask for it.

Flag is off by default. When on, stale tiles serve instantly and their background refresh lands on the next auto-refresh cycle; concurrent opens of the same stale tile collapse to one server-side async recompute instead of N blocking ones. The recent-cache staleness window is unchanged, so this removes redundant blocking recomputes and their latency rather than changing how often a tile is considered stale.

Pairs with #72536 (backend refresh-policy refactor), which makes this the declarable default for the dashboard surface later; this PR delivers the behavior now, gated and measurable.

- Backend: `ruff check` / `ruff format` clean and `python -m py_compile` passes on `posthog/utils.py`.
- Agent-authored: the frontend toolchain (oxlint / oxfmt / `tsgo` typecheck) isn't installed in this task environment, so I could not run it here — please let CI run frontend lint + typecheck. The change is a flag-gated string-mode swap plus a one-value type widening on `getInsightWithRetry`; no API types or components were hand-written.
- Existing `dashboardLogic` refresh tests (`frontend/src/scenes/dashboard/dashboardLogic.test.ts`) exercise `getInsightWithRetry`; with the flag off, behavior is unchanged. Recommend a test with the flag on asserting the tiles request `async_except_on_cache_miss`.

## Rollout / measurement

Enable the flag for a small set of teams, watch `recompute_per_open` on the dashboard cost-per-open sentinel, then widen. Roll back by disabling the flag (no deploy).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None — internal refresh behavior, flag-gated, no API-schema change (the new `refresh` value is additive).

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — requested by @pauldambra. This is the behavior-changing half of a two-PR pair (with #72536); route-cost analysis identified the per-tile blocking refresh on dashboard load as where the recompute cost concentrates.

---

### DIFF
--- a/frontend/src/lib/constants.tsx
+++ b/frontend/src/lib/constants.tsx
@@ -271,6 +271,7 @@ export const FEATURE_FLAGS = {
     CUSTOMER_ANALYTICS_CSP: 'customer-analytics-csp', // owner: @arthurdedeus #team-customer-analytics, gates the Customer analytics > Accounts settings tab (account_group_type_index dropdown)
     CUSTOMER_ANALYTICS_JOURNEYS: 'customer-analytics-journeys', // owner: @arthurdedeus #team-customer-analytics
     CUSTOMER_PROFILE_CONFIG_BUTTON: 'customer-profile-config-button', // owner: @arthurdedeus #team-customer-analytics
+    DASHBOARD_ASYNC_TILE_REFRESH: 'dashboard-async-tile-refresh', // owner: @pauldambra #team-product-analytics — serve recent cache + async-refresh stale dashboard tiles instead of a per-tile blocking recompute on load
     DASHBOARD_AUTO_PREVIEW_LIMIT: 'dashboard-auto-preview-limit', // owner: @pauldambra #team-product-analytics
     DASHBOARD_INLINE_TILE_INSERTION: 'dashboard-inline-tile-insertion', // owner: @MattPua #team-analytics-platform
     DASHBOARD_LAYOUT_DISCARD_PROMPT: 'dashboard-layout-discard-prompt', // owner: @cory.s #team-analytics-platform
--- a/frontend/src/scenes/dashboard/dashboardLogic.tsx
+++ b/frontend/src/scenes/dashboard/dashboardLogic.tsx
@@ -3446,6 +3446,14 @@ export const dashboardLogic = kea<dashboardLogicType>([
                     dashboard,
                     lastDashboardRefresh,
                 } = values
+                // Stale tiles refresh in the background (served from recent cache immediately)
+                // rather than blocking on a recompute per tile; a true cache miss still blocks so
+                // cold tiles aren't shown empty. Flag-gated for staged rollout + cost measurement.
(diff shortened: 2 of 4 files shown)


MattPua:
wouldnt this mean it technically shows stale data, while the background is loading, but for the user it would look like its fresh, and not show that its loading more data?

I ask since that's been a specific complaint that people have had with refreshing dashboards.


hmm maybe im misreading the bot description

pauldambra:
yea, i think it's done a stupid here @MattPua   
  
i tried switching to a cheap model to act on the expensive fable figuring out and i don't think it gets it

pauldambra:
welcome to the glorious future :)  
i wish i could open PRs for robot review without tagging humans though

pauldambra:
Closing — on reflection this doesn't do what it was opened to do.

The goal was to cut dashboard insight recompute cost by switching the per-tile dashboard-load refresh from `blocking` to `async_except_on_cache_miss`. But that doesn't reduce ClickHouse compute: a stale tile runs the same query either way, and **both** paths already collapse concurrent identical recomputes to a single execution — the blocking path via `QueryCoalescingMiddleware` (its key strips `client_query_id`/`session_id`, so concurrent viewers of the same stale tile share one leader compute) and the async path via `enqueue_process_query_task`'s cache-key dedup. So blocking→async changes *when* the user gets a response and *which thread* runs the query, not *how many* queries run. It's a latency / worker-occupancy change, not a cost one.

It also has two correctness issues (thanks Codex): the new mode isn't in the shared-insights whitelist, so cold shared/embedded tiles would render empty instead of blocking on miss; and the flag also flips interactive Preview refreshes, which install the stale result without polling, so an applied filter can stay stale indefinitely.

The real recompute-cost lever is frequency — serve a longer-lived cache, or don't recompute on open — which is a freshness tradeoff best made on the per-surface default table in #72536, not here.

PR #72541, opened 2026-07-21
feat(cdp): remove the 10-minute wait_until_condition poll

dmarchuk:
> [!WARNING]
> Draft. Merge only once the remaining clock-based waits have been migrated, and #75453 has landed so no new ones can be created. Until then this removes the only thing advancing them.

## Problem

`wait_until_condition` steps re-park every ~10 minutes to re-check their condition (`DEFAULT_WAIT_DURATION_SECONDS`). Every parked wait in the fleet pays that, so a small number of clock-based ones can be re-evaluated. It also adds up to ten minutes of latency before a satisfied condition is noticed.

#65949 made the matcher event-driven across three streams, and #72433 added the fourth (`clickhouse_person_distinct_id`, so merges wake on a stream too). Every wake a stream can carry is now carried. What's left is the class no stream can carry: a condition compared against the clock, which changes truth value without anything happening.

## Changes

Park the wait **once, to its `startedAt + max_wait` deadline**, instead of re-parking on the cap.

```diff
-        const scheduledAt = calculatedScheduledAt(
-            action.config.delay_duration,
-            invocation.state.currentAction?.startedAtTimestamp,
-            DEFAULT_WAIT_DURATION_SECONDS
-        )
+        const scheduledAt = calculatedScheduledAt(
+            action.config.delay_duration,
+            invocation.state.currentAction?.startedAtTimestamp
+        )
```

The matcher still wakes the job early via the `eventMatched` short-circuit on any matching signal. If nothing arrives, it resumes at the deadline and takes the timeout branch, same terminal behavior as today.

Also removes the machinery that existed only to prove this was safe: the `pollReparked` state flag and the `cdp_hogflow_wait_poll_only_advance` counter. The merge re-key counter from #72433 stays; it isn't the poll's.

Net **−111 lines**.

## Prerequisites

- [ ] **#75453 landed** so no new clock-based wait condition can be saved. Without it the class silently regrows, and this PR deletes the counter that would have caught it.
- [ ] **The remaining clock-based waits migrated.** Each is expressible today as a branch on whatever determines the interval, a delay, then the condition as a guard on the following step. A delay already parks to an exact instant with no re-checking.
- [ ] `cdp_hogflow_wait_poll_only_advance` reading zero for a sustained window afterwards, which is the confirmation that nothing else was relying on the re-check.

Rebuilt on current master (609 commits of drift, two conflicts in `conditional_branch.ts` and its test, both resolved by taking master's file and reapplying the removal so nothing from #74022 or #72433 was reverted by accident).

- **13 unit tests** in `conditional_branch.test.ts`. The scheduling test now asserts a wait parks to its **full deadline** (a 30-minute wait schedules 30 minutes out), so a reintroduced cap fails it. Deleted the four tests covering the removed machinery. Kept the condition-match and re-key-outcome tests.
- **23 `wait_until_condition` E2E tests** pass against real Postgres, Kafka and Redis, across both queue backends. That includes the matcher wake paths, the merge re-key, HogQL person and event conditions, the internal-events signal, and the timeout path whose 2-second `max_wait` sits well under the old cap.

Removing the `isWait` local as well, since it existed only for poll attribution and is now unused.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Nothing to update: internal matcher and worker behavior with no user-facing surface.

## 🤖 Agent context

Originally authored by Claude (Claude Code) as the follow-up to #72433, and brought up to date by Claude at my direction.

Skill invoked: `/writing-tests` — which is why the four tests for the deleted machinery were removed rather than adapted, and the scheduling test was rewritten to assert the new contract instead of loosened.

What changed since this was first opened. The original readiness gate was "watch `poll_only_advance` sit at zero", on the assumption no live wait was clock-based. Labelling that counter by team and flow (#74022) showed otherwise: a small number of active waits do depend on the re-check. So the gate above is now a migration list rather than an observation window. Two other things surfaced while auditing, both filed separately: a wait condition can reference a data warehouse table, which compiles and then never resolves at runtime because the filter globals are a closed set; and a batch-triggered run whose person is merged away can never advance again, which polling was not rescuing either.


mayteio:
Flagging three more readiness gates from the #66380 side. The poll is quietly doing double duty today: besides backstopping missed matcher wakes, it's the self-heal that applies **live edits** to already-parked waits within one cycle. Parking straight to the deadline removes that too, so:

1. **The timing-edit sweep needs to be rolled out before this merges (`workflows-timing-reschedule`).** Today a shortened `max_wait_duration` reaches parked runs within ~10 minutes for every team via the poll. Post-removal, the reschedule sweep (#71298) is the only mechanism that applies a shortening to parked runs, and it's flag-gated per team (currently off everywhere). Merging this while the flag is off (or only partially ramped) silently regresses max-wait shortenings for un-flagged teams from ≤10 min to full-deadline latency.

2. **Deleted wait steps need a wake at deletion time (sweep-on-delete).** Today a run parked on a deleted `wait_until_condition` wakes on the next poll tick and immediately skip-forwards (#71844) or exits gracefully (#70740). Post-removal it sits until its max-wait deadline before redirecting or exiting. This is the "sweep integration to wake runs parked on deleted wait steps at deletion time" follow-up already noted on #66380 - it should land with or before this PR.

3. **Condition edits mid-wait change semantics - decide and document.** The poll re-evaluates the live condition against the run's stored globals every 10 minutes, so fixing a broken condition takes effect mid-wait even without a new signal. Post-removal, an edited condition only takes effect on a new stream signal or at the deadline. The timing sweep deliberately skips condition edits on the rationale that "the matcher and the poll both evaluate live config at wake" (`timing_reschedule.py`) - half of that rationale goes away here. Probably acceptable, but it's a follow-live contract change that should be an explicit decision, and documented alongside the other per-edit semantics.

Happy to help wire any of these up - 1 is a flag ramp we can start now (dogfood on PostHog Inc is already planned per the flag description), 2 likely reuses the existing sweep primitive triggered from the deletion write sites.

mayteio:
Hey, just an update here on my three notes above.

1. I rolled out the timing sweep to 100% of customers today, so that's done ✅
2. [Opened a PR](https://github.com/PostHog/posthog/pull/73925) to sweep when a timing step gets deleted, which unblocks this work
3. Decided we should sweep on condition changes too and rolled it into the same PR

TLDR all three of my gates are handled 😄 once #73925 merges you're good to go here

PR #72542, opened 2026-07-21
feat(ci): split trunk upload from quarantine masking

gantoine:
## Problem

Right now one repo variable, `TRUNK_UPLOAD_ENABLED`, controls two things at once: whether we upload flaky-test spans to Trunk, and whether Trunk's quarantine verdict is allowed to mask known-flaky failures so they don't red CI. There's no way to keep feeding Trunk data to build up the flake database while turning off the masking. You either get both or neither.

## Changes

Adds a second repo variable, `TRUNK_QUARANTINE_ENABLED`, so the two behaviors toggle independently.

- `TRUNK_UPLOAD_ENABLED` (unchanged) stays the master switch that gates whether the `analytics-uploader` step runs at all, i.e. whether spans get uploaded.
- `TRUNK_QUARANTINE_ENABLED` (new) is honored only by the per-job verdict (`exit 1`) step. It's opt-in and defaults off: masking applies only when it's explicitly `'true'`, matching the existing `TRUNK_UPLOAD_ENABLED == 'true'` convention.

The verdict condition went from `test == 'failure' && gate.outcome != 'success'` to `test == 'failure' && (vars.TRUNK_QUARANTINE_ENABLED != 'true' || gate.outcome != 'success')`.

> [!WARNING]
> This changes behavior on merge. Today masking is on whenever `TRUNK_UPLOAD_ENABLED` is `'true'`. After this merges, masking is off until you set `TRUNK_QUARANTINE_ENABLED=true` — so known flakes will red CI again until that variable is set. Quarantining still requires upload to be on (you can't mask without uploading), so `UPLOAD=false` fails closed regardless.

Resulting matrix:

| `TRUNK_UPLOAD_ENABLED` | `TRUNK_QUARANTINE_ENABLED` | Behavior |
|---|---|---|
| true | `true` | Upload spans + mask known flakes |
| true | unset / `false` | Upload spans, but every real failure reds CI |
| false | (any) | No Trunk contact; failures red CI |

Touches the 8 fail-gate steps across the six quarantine-mode workflows: `ci-backend` (product + Core/Temporal), `ci-frontend` (jest + jest-replay-shared), `ci-e2e-playwright`, `ci-nodejs`, `ci-llm-gateway`, `ci-mcp`. The analytics-only jobs (python/rust/dagster/ai/storybook) never masked, so they're untouched.

No unit tests here since this is workflow `if:` logic. I (Claude) verified:

- `bin/hogli lint:workflows` passes across all 102 workflows.
- All six changed files parse as valid YAML and the eight `if:` conditions render with balanced expressions.
- Worked through the boolean truth table for each cell of the matrix above (including the `UPLOAD=false` / gate-skipped case) to confirm the default-off behavior and that no combination lets a real failure report green.

`actionlint` wasn't available locally, so the GHA expression grammar is verified by CI's own lint job rather than by me.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Georges-Antoine asked whether disabling `TRUNK_UPLOAD_ENABLED` would let failing tests report green (it wouldn't, and the investigation confirmed disabling Trunk makes CI stricter), then asked to split span-upload from quarantine-masking. I (Claude) mapped the wiring, proposed the two-flag design, and made the edits.

The new flag defaults off (opt-in via `== 'true'`) at Georges-Antoine's direction, so masking only kicks in once the variable is deliberately set — consistent with how `TRUNK_UPLOAD_ENABLED` already works. Kept the uploader running in `quarantine: true` mode even when masking is off — it still uploads and computes a verdict we simply ignore, which is simpler than branching the step into upload-only vs quarantine variants.

Skill invoked: `/authoring-ci-workflows`.


rnegron (approved):
makes sense to me!

PR #72545, opened 2026-07-21
feat(cdp): daily temporal evaluator for workflow email reputation

meikelmosby:
## Problem

All workflow email goes out through one shared sending account, and we can't see which customer or workflow is causing bad bounce/spam-complaint rates. We're shipping a daily reputation evaluator (calculation + visibility only, no enforcement) split into three PRs; this is part 2 of 3 — the write path.

- Part 1 ([#72543](https://github.com/PostHog/posthog/pull/72543)): `EmailReputationSnapshot` schema — **must merge and migrate first** (this service writes to that table)
- **Part 2 (this PR):** Node/Temporal evaluator that computes and writes daily snapshots
- Part 3 ([#68133](https://github.com/PostHog/posthog/pull/68133)): reputation API + flag-gated Reputation tab

## Changes

- New `email-reputation-evaluator` plugin-server mode/capability hosting a Temporal worker + idempotently ensured daily schedule (self-updating on config change, overlap SKIP). In dev it rides along with the combined CDP set and degrades gracefully when local Temporal is down.
- Evaluation is volume-based, mirroring how SES judges the shared account: each workflow and each team is rated over a window spanning at least `EMAIL_REPUTATION_MIN_WINDOW_HOURS` (24) and at least `EMAIL_REPUTATION_TARGET_VOLUME` sends (1000) — whichever reaches further back, capped at 30 days. A weekly batch blast keeps counting until newer volume displaces it; a high-volume sender can't hide a bad morning behind clean recent hours; late-arriving bounces are picked up by the next run.
- Batch-job metric attribution (broadcasts record metrics under the batch-job id), pooled team aggregates including unattributed senders, carry-forward zero-volume snapshots for teams that go silent, per-batch failure isolation, and a UUID guard so one malformed app_source_id can't poison a run.
- Deployment: [PostHog/charts#13325](https://github.com/PostHog/charts/pull/13325) (dedicated singleton app) + [posthog-cloud-infra#9489](https://github.com/PostHog/posthog-cloud-infra/pull/9489) (cloud PG credentials, CH pod auth) — merge order: part 1 → infra → charts → this.

I'm an agent; checks actually run: 16 jest tests (parameterized classifier thresholds; service tests covering volume windowing, the 24h floor, late bounce-only buckets, batch-job attribution, tenant pooling vs averaging, carry-forward rows, idempotent retries — each pinned to a regression named in the test), Node tsc, oxlint, prettier. Temporal-scheduled path not exercised end-to-end against real ClickHouse.

## 🤖 Agent context

Split out of [#68133](https://github.com/PostHog/posthog/pull/68133) (built with Claude Code; /writing-tests + /code-review skills used). Code is identical to what was reviewed there, including the review fixes (TLS agent matching sibling internal-ClickHouse clients, worker lifecycle hardening, schedule update-on-exists, batch isolation).

---


meikelmosby:
Heads-up on the red **Node.js Tests 1/3** job: the failing tests are this PR's own email-reputation service tests hitting `relation posthog_emailreputationsnapshot does not exist`. That's expected until [#72543](https://github.com/PostHog/posthog/pull/72543) merges — ci-nodejs's fast path restores the pre-migrated schema snapshot built from **master** whenever the PR's own diff has no migration files, so the table from the base branch isn't there. No code change needed here; once #72543 is on master (and the schema snapshot regenerates on the next master run), a rerun goes green. Everything else on this PR is passing.

nodejs/src/cdp/services/email-reputation/temporal/email-reputation-worker.service.ts
```diff
@@ -0,0 +1,211 @@
...
+        const clickhouse = createClickHouseClient({
+            url: `${chScheme}://${this.config.CLICKHOUSE_HOST}:${chPort}`,
+            username: this.config.CLICKHOUSE_USER,
+            password: this.config.CLICKHOUSE_PASSWORD || undefined,
+            database: this.config.CLICKHOUSE_DATABASE,
+            request_timeout: 60_000,
+            max_open_connections: 10,
+            ...(this.config.CLICKHOUSE_SECURE
+                ? {
+                      http_agent: new https.Agent({ rejectUnauthorized: false, keepAlive: true, maxSockets: 10 }), // nosemgrep: problem-based-packs.insecure-transport.js-node.bypass-tls-verification.bypass-tls-verification
```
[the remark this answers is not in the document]
meikelmosby: This mirrors the repo's established pattern for internal ClickHouse clients — `cdp-rerun-worker.consumer.ts` and `session-replay/recording-api.ts` carry the identical `http_agent` with the same nosemgrep annotation, because internal ClickHouse serves self-signed certs with a hostname mismatch; with default verification the evaluator's first query fails every run in production. The durable fix (proper certs / CA pinning) is infra-level and should land across all three call sites at once — extracting a shared `createInternalClickHouseClient` factory is on the follow-up list to give that a single home.

nodejs/src/cdp/services/email-reputation/temporal/activities.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Activity payloads ride Temporal workflow history (~2 MiB cap per payload). Only team ids and
+ * counts cross the boundary — metric rows and snapshots stay inside the batch activity.
+ */
+export function createActivities(
+    service: EmailReputationService,
+    pacing: { batchSize: number; batchDelayMs: number }
+): EmailReputationActivities {
+    return {
+        fetchTeamsToEvaluate: async (evaluatedAt) => ({
+            teamIds: await service.fetchTeamsToEvaluate(evaluatedAt),
```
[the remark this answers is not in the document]
meikelmosby: Fixed in 1815fec3d38: `fetchTeamsToEvaluate` is now cursor-paginated by team id (5,000 ids per activity result, ~45KB — well under the repo's 256KB field rule), and the workflow walks pages via `nextCursor`, so the fleet-wide team list never rides a single Temporal payload.

nodejs/src/cdp/services/email-reputation/temporal/email-reputation-worker.service.ts
```diff
@@ -0,0 +1,211 @@
...
+            logger.info('[EmailReputationWorker] created schedule', { hour })
+        } catch (error) {
+            if (error instanceof ScheduleAlreadyRunning) {
+                // Update the existing schedule so config stays authoritative (hour changes, or a
+                // stale spec left by an earlier build, would otherwise be silently ignored forever).
+                await client.schedule.getHandle(EMAIL_REPUTATION_SCHEDULE_ID).update((previous) => ({
+                    ...previous,
+                    spec: { calendars: [{ hour }] },
+                    policies: { ...previous.policies, overlap: ScheduleOverlapPolicy.SKIP },
+                }))
```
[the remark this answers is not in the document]
meikelmosby: Fixed in 1815fec3d38 — took your suggestion as-is: the update path now reasserts the full action (workflow type, task queue, args) alongside spec and policies, so a schedule left by an older build can't keep dispatching a stale target.

nodejs/src/cdp/services/email-reputation/email-reputation.service.ts
```diff
@@ -0,0 +1,398 @@
...
+            // email_blocked is how SES complaint events are recorded (see helpers/ses.ts), hence
+            // the `complained` alias. No HAVING sent > 0: buckets holding only late-arriving
+            // bounces/complaints must still count toward the window they fall into.
+            query: `
+                SELECT
+                    team_id,
+                    app_source_id,
+                    toUnixTimestamp(toStartOfHour(timestamp)) AS hour_bucket,
+                    sumIf(count, metric_name = 'email_sent') AS sent,
+                    sumIf(count, metric_name = 'email_bounced') AS bounced,
```
mayteio: all SES bounce types map to `email_bounced`, but SES's account bounce rate (which these thresholds are calibrated to) only counts hard bounces - greylisting / mailbox-full traffic will push teams into warning/critical that SES doesn't hold against us. split transient bounces into their own metric at the webhook? or a comment saying overcounting is intentional

nodejs/src/cdp/services/email-reputation/email-reputation.service.ts
```diff
@@ -0,0 +1,398 @@
...
+        }
+
+        const rows = await this.fetchHourlyEmailMetrics(teamIds, evaluatedAt)
+        const { flows, sourceToFlow } = await this.resolveSources([...new Set(rows.map((r) => r.appSourceId))])
+        const snapshots: SnapshotRow[] = []
+        const minWindowStart = Math.floor(Date.parse(evaluatedAt) / 1000) - this.config.minWindowHours * 3600
+
+        // Per-workflow: fold each source's hourly buckets into its workflow, then take each
+        // workflow's window (>= minWindowHours and >= targetVolume sends).
+        const workflowBuckets = new Map<string, Map<number, ReputationMetrics>>()
```
mayteio, line 150: `teamId` here comes from `flow.team_id`, not the metrics row, and none of the lookups filter by team - a cross-team `app_source_id` would attribute one team's bounces to another team's snapshot. one-line guard: skip when `flow.team_id !== row.teamId`

mayteio (approved):
really nice work, this is solid! left two inline comments worth a look before merge.

nodejs/src/cdp/services/email-reputation/email-reputation.service.ts
meikelmosby: Fixed in 651294e7d69 — took the one-line guard: workflow attribution now skips when the resolved flow's `team_id` doesn't match the metric row's team (the row still counts toward its own team's aggregate, which is the correct place for it). Extended the attribution test with a cross-team `app_source_id` case — without the guard it polluted the workflow snapshot's assertions, so the regression is pinned.

nodejs/src/cdp/services/email-reputation/email-reputation.service.ts
meikelmosby: Good catch — documented in 651294e7d69 as intentional for this phase, with the reasoning in a comment at the query: overcounting transient bounces biases toward earlier warnings, which is the safe direction while the score is visibility-only. Splitting hard vs transient at the SES webhook is the right durable fix, but it changes `email_bounced` semantics for every existing app_metrics2 consumer (workflow metrics UI included), so it deserves its own change — I'd slot it as a prerequisite of the enforcement phase, where the calibration against AWS's hard-bounce-only account rate starts to matter for real. Happy to open that follow-up if you want it queued now.

meikelmosby: Update: we went further than documenting — the calculation is now hard-bounce-calibrated. The SES webhook additively emits `email_bounced_hard` for `bounceType === 'Permanent'` (shipped in the foundation PR #72543 so it accumulates from the same deploy that creates the table; `email_bounced` unchanged for existing consumers), and this evaluator now reads that metric. Transient greylisting/mailbox-full traffic no longer counts toward the score, matching AWS's account bounce rate. One documented ramp-up caveat: windows reaching back before the part-1 deploy undercount bounces until the metric has volume-window depth — acceptable while visibility-only.

PR #72548, opened 2026-07-21
feat(data-catalog): attach trust advisory to queries over uncertified warehouse tables

thiagosalvatore:
## Problem

Agents answering business-metric questions satisfice on plausible raw warehouse tables. Our Slack bot answered a "top customers by revenue" question from an uncertified table while an approved governed metric existed. The catalog-first steering was present and did not save it - instructions in tool descriptions are a low-attention channel, and nothing on the path the agent actually walked carried a trust signal. Review feedback on the first iteration of this PR (which intercepted at the MCP `exec search` layer) correctly pushed back: that surface is slated for removal, and interception there is the wrong altitude.

## Changes

The fix is now a **point-of-use trust signal in the backend**, with zero MCP-server changes:

- New `DataCatalogTrustWarning` in the query response `warnings` union. When a HogQL query reads a warehouse table with no certification (or a deprecated one) and the team's catalog holds approved governed metrics, `HogQLQueryExecutor.execute()` attaches an advisory naming the uncertified tables and up to 5 approved metric names, pointing at `system.information_schema.metrics`.
- The advisory reaches agents through the existing warnings formatter: `format_data_catalog_trust_warnings` joins the warehouse-sync and access-control blocks prepended to the `execute-sql` result string. The MCP server forwards that string verbatim - no interception anywhere.
- Implementation reuses the machinery that already exists: `extract_base_table_types` for used-table extraction (same as `extract_warehouse_sources`), the `build_access_control_warning` assembly pattern, and the data-catalog facade for certifications and metrics. Fail-soft end to end (an exception in the builder can never break a query), flag-gated on `product-data-catalog`, and withheld from unauthenticated principals (same posture as warehouse source attribution).
- The earlier `exec search` metric integration is fully reverted. Kept from the first iteration: the strengthened catalog-first steering (within its existing prompt budget), the sandboxed evals, and a fix making the `tests/hono` suites independent of dev-shell env vars.

Why this altitude: the incident's root cause was satisficing plus a silent environment at the point of use, not missing search results. A signal attached to the query result fires on the exact path taken, through the highest-attention channel (tool results), regardless of how the table was found - and it survives any MCP redesign, including the planned code-execution direction.

- New `posthog/hogql/test/test_catalog_trust.py` (10 tests, parameterized): advisory fires for uncertified tables with approved metrics; silent for certified tables, deprecated certs still fire, silent with no approved metrics / flag off / events-only queries / unauthenticated principals; a lookup exception never breaks the query; and an executor-level integration test proving the warning rides `execute_hogql_query`'s response. These catch the regressions that matter: false-positive noise (trust erosion + response byte-changes) and silent non-firing.
- Formatter tests: the new warning type renders its own block and never leaks into the other formatters' blocks.
- Sandboxed evals against this branch (real coding agent + MCP server + Django live server), all passing:
  - `governed_metric_vs_raw_table` (prompted) - 100% on every scorer.
  - `governed_metric_vs_raw_table_unprompted` (the incident verbatim, no nudge) - 96.7% mean; the agent consulted the catalog before touching any raw table, so the strengthened steering alone carried it and the advisory was correctly dormant. The advisory's firing path is covered by the executor integration test; the rescue-after-raw-read scenario remains probabilistic agent behavior we grade, not force.
  - `tool_task_undistracted_by_metrics` control - 100%.
- `hogli test services/mcp` back at master parity plus retained tests (2284 unit + 302 hono green). Repo-wide mypy: no errors in changed files. Frontend typecheck clean. Schema regenerated with the CI toolchain (Python 3.12 - the 3.13 dev venv produces different codegen wrapping; the committed output matches what CI's drift check regenerates). `hogli ci:preflight --fix` clean.

## Known residuals

- v1 covers warehouse (S3) tables; saved-query views (also certifiable) are a follow-up.
- No frontend rendering of the new warning type yet (UI ignores unknown warning kinds).
- The advisory names approved metrics without computing drift on the query hot path; the message points at `system.information_schema.metrics` where drift is visible.

## Post-deploy monitoring & validation

- The advisory rides existing query responses; no new infra. Healthy signal: `data_catalog_trust` warnings appearing on raw warehouse queries for flag-enabled teams (visible in `$mcp_tool_call` output token estimates as a small bounded increase only on flagged queries - the message is ~2 lines).
- Failure signal: query errors correlating with catalog lookups (should be impossible by construction - the builder is fail-soft; any such correlation is a bug).
- Rollback: disable the `product-data-catalog` flag for the org - the advisory (and all catalog surfaces) go silent with no deploy.
- Validation window: first week on dogfood orgs. Owner: me.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Steering template text is part of this PR; no separate docs surface documents the warnings envelope.

## 🤖 Agent context

Claude (Fable 5, via Conductor/Claude Code) authored this end to end across two iterations. The first intercepted at MCP `exec search`; after review pushback and a first-principles re-analysis (the incident's cause was a silent environment at the point of use, not missing search results), the interception was fully reverted in favor of this backend advisory - see `docs/plans/2026-07-21-002-feat-data-catalog-trust-advisory-plan.md`. Repo skills invoked: `/writing-tests`, `/writing-evals`. Notable findings along the way: the schema codegen produces different string wrapping under Python 3.13 vs CI's 3.12 (regenerated with 3.12 to keep the drift check green), and the `tests/hono` suites were env-dependent on dev shells (fixed here).


skoob13 (changes requested):
Searching through SQL is a much better pattern than trying to intercept the agent's intent for the semantic layer earlier. ClickHouse exposes rich capabilities for FTS, pattern, and semantic search. exec search should not be extended – we will likely remove it in the next MCP version with code execution. What were the exact cases when it didn't find a metric?

skoob13:
Alternatively, I'd propose combining SQL schema discovery with metrics – I tested it some time ago, and it gave great results.

thiagosalvatore:
> Searching through SQL is a much better pattern than trying to intercept the agent's intent for the semantic layer earlier. ClickHouse exposes rich capabilities for FTS, pattern, and semantic search. exec search should not be extended – we will likely remove it in the next MCP version with code execution. What were the exact cases when it didn't find a metric?

We already search using SQL, there's a description on the PR explaining exactly what are the situations where it is failing. The generic search that triggers the tool search (and return the tools available) is not being able to capture the right tool to answer the user's question and instead it is taking the user through the wrong path.  Here's the slack thread: https://posthog.slack.com/archives/C0A9BSVSE72/p1784629059421989

We already do SQL schema discovery using the information schema

skoob13:
@thiagosalvatore I'd try renaming "data catalog" to "semantic layer" first thing and see what happens. The semantic layer is easier for agents to understand, while the data catalog is a bit ambiguous. Also, the metrics prompt needs to have more guidance on when to use the semantic layer, ideally with a few examples.

thiagosalvatore:
> @thiagosalvatore I'd try renaming "data catalog" to "semantic layer" first thing and see what happens. The semantic layer is easier for agents to understand, while the data catalog is a bit ambiguous. Also, the metrics prompt needs to have more guidance on when to use the semantic layer, ideally with a few examples.

This was a result of several rounds of autoresearch. It was initially semantic layer, the recall was lower compared to data catalog. I can try to do some tweaks in the prompt, but I don't expect it to improve too much. Anyways, I will try other solution to see if it fixes the problem

PR #70492, opened 2026-07-13
fix(integrations): back off and terminate permanently failing OAuth token refreshes

mayteio:
## Problem

The every-minute integration refresh sweep (`refresh_integrations`) re-enqueues every expired integration forever.
An integration whose grant is permanently dead (revoked by the provider, consumer app deleted, user re-auth required) fails its refresh every single minute, indefinitely, with no backoff and no terminal state.

Two bad outcomes:

- We hammer providers with large volumes of unwinnable token refresh requests, around the clock.
- `integration_oauth_refresh_total{result="failed"}` runs at a chronically high baseline per kind, so when an entire fleet of integrations breaks at once (e.g. a credential rotation revoking grants), the failure signal barely changes shape and nobody notices. This is exactly how a recent fleet-wide auth breakage went undetected.

## Changes

- Consecutive refresh failures now back off exponentially, 2m doubling to a 1h cap. State lives in `Integration.config` (`refresh_failure_count`, `refresh_next_attempt_at`), so no migration, and a reconnect through the OAuth callback resets it for free because `integration_from_oauth_response` replaces `config` wholesale.
- After 5 consecutive `invalid_grant` failures the integration goes terminal (`refresh_terminal`) and the sweep stops retrying it entirely. A dead grant is only fixable by customer re-auth, by definition. Other failure reasons (`invalid_client`, 5xx, network) never go terminal, so a platform-side credential fix lets the whole fleet self-recover within the backoff cap.
- The sweep and the per-row `refresh_integration` task both skip backed-off rows. On-demand refreshes via the API (`_ensure_token_usable`) deliberately bypass the backoff, so user-initiated retries still work.
- New labels on the refresh counter, cheap because the backoff already tracks the state:
  - `reason`: `invalid_grant` | `invalid_client` | `http_5xx` | `other`. A wave of `invalid_grant` means grants are being revoked; `invalid_client` means our credentials are broken.
  - `attempt`: `first` | `retry`. A spike of first-attempt failures means connections are newly breaking, regardless of retry noise.
  - Existing queries that `sum by (kind, result)` are unaffected.
- A non-JSON error response body (e.g. an HTML page from a proxy) now counts as a failed refresh with backoff, instead of raising out of the task with no bookkeeping at all.

> [!NOTE]
[...]

PR #70257, opened 2026-07-11
feat(integrations): fallback oauth client credentials on rotation

jabahamondes:
## Problem

When we rotate an OAuth app's client secret, or migrate the app to a different account (a new `client_id`), every integration whose refresh token was issued by the old app stops refreshing. Access tokens die within about an hour and the user is forced to reconnect. We hit this rotating the Bing Ads app: a refresh token is bound to the `client_id` that issued it, so the new prod credentials cannot refresh tokens minted by the old app.

There is no way today to keep both credential pairs live during a rotation window. Prod holds exactly one secret per provider.

## Changes

Adds an optional fallback credential pair per provider. On refresh we try the primary credentials first, and if that fails and a fallback is configured we retry with the fallback pair. Tokens issued by a since-rotated app keep refreshing until the user reconnects, so a rotation becomes zero-downtime instead of a forced mass reconnect.

- `OauthConfig` gains `client_id_fallback` / `client_secret_fallback`.
- `oauth_config_for_kind` is now a thin wrapper that populates the fallback from a new `settings.OAUTH_CLIENT_FALLBACKS` map. The per-provider config building moved verbatim into `_build_oauth_config`, so any provider gets fallback support by adding one map entry, with no per-branch change.
- `refresh_access_token` extracts the per-provider POST into `_post_token_refresh(client_id, client_secret)` and adds the retry. The fallback path only fires when a fallback secret is set, so providers without one behave exactly as before.
- The Prometheus counter now emits `success_fallback` distinctly from `success`. That is the signal for when the old app has no users left and the fallback env vars can be removed.

Only `bing-ads` gets a map entry for now (`BING_ADS_CLIENT_ID_FALLBACK` / `BING_ADS_CLIENT_SECRET_FALLBACK`). Other providers are a four-line addition when they next rotate.

> [!NOTE]
> This covers every provider that refreshes through `OauthIntegration.refresh_access_token` (bing, linkedin, reddit, tiktok, pinterest, snapchat, hubspot, salesforce, slack, intercom, linear, clickup, jira-exchange, stripe). Google Cloud, Firebase, Jira, Meta and GitHub have their own `refresh_access_token` and would need the same pattern in a follow-up.

> [!WARNING]
> The fallback only rescues users if the old OAuth app is still live at the provider and its credentials are loaded in prod. This PR is the mechanism. The operational steps (keep the old app alive, set the fallback secrets, remove them once `success_fallback` hits zero) are separate.

## How did you test this code?

[...]

PR #72557, opened 2026-07-21
feat(integrations): add TypeScript integration gateway service

benjackwhite:
## Problem

Integration credentials are currently decrypted by each consuming service, which broadens the credential-handling surface and makes refresh ownership difficult to reason about.

**Why:** Centralizing access and refresh gives credential reads one auditable path and lets consumers shed encryption and OAuth secrets over time.

## Changes

- Adds an isolated TypeScript integration gateway with team-scoped reads, short-lived plaintext caching, audit metrics, and opt-in just-in-time OAuth refresh.
- Routes CDP reads through the gateway behind a per-team rollout with Postgres fallback.
- Keeps refresh ownership partitioned by integration kind and team, with strict config parsing and owner-safe Redis lock release.
- Removes obsolete scoped-JWT artifacts after moving caller admission to the network boundary.

- Focused gateway unit tests: 15 passed.
- ESLint and Prettier passed for changed TypeScript files.
- Ruff passed for changed Python files.
- Full Node type-check remains blocked locally by unbuilt workspace packages; Django tests require the repository Postgres service.

## 🤖 Agent context

Reviewed for logic/correctness, contracts/security, and performance/reliability. Rebased onto current `master`, resolved refresh ownership and local-port conflicts, and fixed concurrency/config drift found during review.

Agent-authored changes; please review and do not self-merge.

---


nodejs/src/integration-gateway/refresh/providers.ts
```diff
@@ -0,0 +1,93 @@
...
+    GOOGLE_ADS_APP_CLIENT_ID: string
+    GOOGLE_ADS_APP_CLIENT_SECRET: string
+    GOOGLE_ANALYTICS_APP_CLIENT_ID: string
+    GOOGLE_ANALYTICS_APP_CLIENT_SECRET: string
+    GOOGLE_SEARCH_CONSOLE_APP_CLIENT_ID: string
+    GOOGLE_SEARCH_CONSOLE_APP_CLIENT_SECRET: string
+    SOCIAL_AUTH_GOOGLE_OAUTH2_KEY: string
+    SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET: string
+    // When set, replaces the token endpoint for every kind (local/e2e mock only).
+    INTEGRATION_GATEWAY_REFRESH_TOKEN_URL_OVERRIDE: string
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. `providerFor` now ignores `INTEGRATION_GATEWAY_REFRESH_TOKEN_URL_OVERRIDE` when `isProdEnv()`, so a prod misconfig can no longer redirect credential traffic to an arbitrary host. Covered by a new case in `providers.test.ts`.

nodejs/src/integration-gateway/refresh/providers.ts
```diff
@@ -0,0 +1,93 @@
...
+                clientSecret: config.HUBSPOT_APP_CLIENT_SECRET,
+            }
+            break
+        case 'salesforce':
+            resolved = {
+                defaultUrl: 'https://login.salesforce.com/services/oauth2/token',
+                clientId: config.SALESFORCE_CONSUMER_KEY,
+                clientSecret: config.SALESFORCE_CONSUMER_SECRET,
+            }
+            break
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. Added `salesforceTokenUrl(config.instance_url)`, mirroring Django's `_salesforce_instance_host` validation (https only, no port/userinfo, hostname under `salesforce.com`) and falling back to the login URL otherwise. The row's `config` is now threaded into `providerFor`. Tested in `providers.test.ts`.

nodejs/src/common/utils/db/hub.ts
```diff
@@ -95,7 +96,12 @@ export async function createHub(config: Partial<PluginsServerConfig> = {}): Prom
     const geoipService = new GeoIPService(serverConfig.MMDB_FILE_LOCATION)
     await geoipService.get()
     const encryptedFields = new EncryptedFields(serverConfig.ENCRYPTION_SALT_KEYS)
-    const integrationManager = new IntegrationManagerService(pubSub, postgres, encryptedFields)
+    const integrationManager = new IntegrationManagerService(
+        pubSub,
+        postgres,
+        encryptedFields,
+        createIntegrationGatewayService(serverConfig)
+    )
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. `server.ts` now builds the gateway client into the main server's `IntegrationManagerService` via `createIntegrationGatewayService`, so the CDP read path is reachable from the production entrypoint (behind the per-team rollout, dark by default).

nodejs/src/integration-gateway/integration.service.ts
```diff
@@ -0,0 +1,82 @@
...
+                    team_id: row.team_id,
+                    kind: row.kind,
+                    config: row.config,
+                    sensitive_config: this.encryptedFields.decryptObject(row.sensitive_config, {
+                        ignoreDecryptionErrors: true,
+                    }),
+                }
+                this.cache.insert(row.id, decrypted)
+                resolved.set(row.id, decrypted)
+            }
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. The team-scope check now runs *before* any refresh/decrypt — rows belonging to another team are skipped up front — so a wrong-team caller can never trigger a refresh (outbound OAuth + DB write) against another team's integration. Covered by `integration.service.test.ts`.

nodejs/src/common/utils/encryption-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+     * @param encryptionSaltKeys comma-separated primary keys
+     * @param legacySecretKeys comma-separated SECRET_KEY(+fallbacks) — optional, decrypt-only
+     * @param saltKeys comma-separated SALT_KEY PBKDF2 salts — optional, decrypt-only
+     */
+    constructor(encryptionSaltKeys: string, legacySecretKeys: string = '', saltKeys: string = '') {
+        for (const key of encryptionSaltKeys.split(',').filter((k) => k)) {
+            this.fernets.push(new Fernet(toFernetKey(Buffer.from(key, 'utf-8'))))
+        }
+        const secrets = legacySecretKeys.split(',').filter((k) => k)
+        const salts = saltKeys.split(',').filter((k) => k)
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. Legacy secret/salt keys are now `.trim()`-ed (like Django's `get_list`) before PBKDF2 derivation, restoring byte-identical key material for space-padded CSV values.

nodejs/src/integration-gateway/refresh/expiry.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (expiresIn === null || refreshedAt === null) {
+        return false
+    }
+    if (expiresIn <= 0) {
+        return false
+    }
+
+    const threshold = expiresIn / 2
+    return nowSecs() > refreshedAt + expiresIn - threshold
+}
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Addressed in `e5520feb`. The refresh path now skips (rather than marks-failed) when there is no stored `refresh_token`, mirroring Django's `access_token_expired` returning `False` in that case.

nodejs/src/cdp/services/managers/integration-gateway.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        const body = (await response.json()) as { integrations?: Record<string, IntegrationType | null> }
+        const gatewayResult = body.integrations ?? {}
+
+        const result: Record<string, IntegrationType | null> = {}
+        for (const id of ids) {
+            result[id] = gatewayResult[String(id)] ?? null
+        }
+        return result
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `e5520feb`. A 200 with a malformed body now throws, which triggers the Postgres fallback instead of silently resolving every id to `null` (which would have surfaced as "integration not found" for real, connected integrations). Tested in `integration-gateway.service.test.ts`.

posthog/tasks/integrations.py
```diff
@@ -18,8 +20,13 @@
...
     from posthog.models.integration import Integration, OauthIntegration
 
+    # Kinds owned by the integration gateway's refresher are excluded here so exactly one system
+    # refreshes each kind (the gateway enables the same kinds via INTEGRATION_GATEWAY_REFRESH_KINDS).
     oauth_integrations = defer_repository_cache_fields(
-        Integration.objects.filter(kind__in=OauthIntegration.supported_kinds).exclude(kind="meta-ads").all()
+        Integration.objects.filter(kind__in=OauthIntegration.supported_kinds)
+        .exclude(kind="meta-ads")
+        .exclude(kind__in=settings.INTEGRATION_GATEWAY_REFRESH_KINDS)
+        .all()
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `1cfea6e` — this is the one you flagged, now gated by **team as well as kind**. Added `INTEGRATION_GATEWAY_REFRESH_TEAMS` (explicit team ids, or `*`). The beat excludes a row only when its kind ∈ `INTEGRATION_GATEWAY_REFRESH_KINDS` **AND** its team ∈ `INTEGRATION_GATEWAY_REFRESH_TEAMS`, so an out-of-rollout team keeps its beat refresher rather than being orphaned when a kind is enabled. The gateway's `RefreshManager.owns(kind, teamId)` mirrors the same gate. Kind stays as the capability contract because the gateway only refreshes 6 of Django's 19 oauth kinds — excluding a whole team would orphan the rest. Tests added on both sides (`test_integrations.py`, `manager.test.ts`).

nodejs/src/integration-gateway/integration.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `1cfea6e`. The DB-miss loop now processes rows concurrently with `Promise.all`, so a batch of stale rows refreshes in parallel instead of serially inflating latency with the count of stale rows. Each refresh is still single-flighted per integration via its Redis lock.

nodejs/src/integration-gateway/refresh/manager.ts
```diff
@@ -0,0 +1,213 @@
...
+            try {
+                await client.del(key)
+            } finally {
+                await this.redisPool.release(client)
+            }
+        } catch (error) {
+            // Best-effort; the lock TTL-expires on its own.
+            logger.warn('[RefreshManager] failed to release refresh lock', { key, error: String(error) })
+        }
+    }
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Not changing this — it isn't reachable under the current config. After `e5520feb` the lock is held on failure and only released on the success path, by the holder. The ownership race requires a refresh to outlive its lock, but the HTTP timeout (`INTEGRATION_GATEWAY_REFRESH_HTTP_TIMEOUT_MS` = 10s) is well under the lock TTL (`INTEGRATION_GATEWAY_REFRESH_LOCK_TTL_SECONDS` = 30s), so a holder can't still be running when its lock expires. If we ever set the TTL below the HTTP timeout this would want a CAS-token release (store a unique value, delete only on match), but that's not the case today.

nodejs/src/cdp/services/managers/integration-manager.service.ts
```diff
@@ -1,33 +1,61 @@
...
+                gatewayRequestsCounter.inc({ result: 'ok' })
+                return result
+            } catch (error) {
+                logger.warn('[IntegrationManager]', 'Gateway fetch failed, falling back to Postgres', {
+                    error: String(error),
+                })
+                gatewayRequestsCounter.inc({ result: 'fallback' })
+            }
+        }
         return await this.lazyLoader.getMany(ids.map((id) => id.toString()))
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Intentional, leaving as-is. The gateway is the cache authority on this path — it keeps its own short-TTL decrypted-credential cache — so layering the CDP `LazyLoader` on top would double-cache and compound staleness across two independent TTLs. Batching within a call is preserved (`getMany` passes the full id list through in one request). To keep the extra HTTP hop cheap when enabled, `1cfea6e` adds a fast-fail timeout; if per-call QPS proves high we can revisit in-flight dedup specifically.

nodejs/src/cdp/services/managers/integration-gateway.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Fixed in `1cfea6e`. Added `CDP_INTEGRATION_GATEWAY_TIMEOUT_MS` (default 3s), passed as the fetch `timeoutMs`, so a slow/degraded gateway falls back to Postgres quickly instead of every hot-path read blocking on the long default external-request timeout. Kept it a simple per-call timeout rather than a full circuit-breaker for now; can add breaker semantics later if QPS warrants. Tested in `integration-gateway.service.test.ts`.

nodejs/src/integration-gateway/refresh/manager.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Refresh `row` if its access token is past half-life, returning the (possibly updated) row with
+     * re-encrypted credentials. Fail-open: on any error the input row is returned unchanged, so the
+     * read path still serves the existing (still-valid, since we refresh proactively) token.
+     */
+    async refresh(row: IntegrationRow): Promise<IntegrationRow> {
+        if (!accessTokenExpired(row.kind, row.config)) {
+            return row
+        }
```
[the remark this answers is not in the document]
benjackwhite: _🤖 Reply from a Claude Code session, on Ben's behalf._

Addressed in `e5520feb`. On refresh failure the Redis lock is now deliberately **not** released, so its TTL acts as a per-integration cooldown — a lightweight stand-in for Django's refresh backoff — instead of re-hammering the provider on every fetch.

nodejs/src/integration-gateway/refresh/manager.ts
```diff
@@ -0,0 +1,229 @@
...
+                })
+            }
+            recordRefresh(row.kind, 'failed')
+            return row
+        }
+    }
+
+    private async refreshLocked(row: IntegrationRow, provider: Provider): Promise<IntegrationRow> {
+        // Re-read under the lock: a concurrent head (or Django) may have just rotated the token.
+        const fresh = (await this.repository.fetchOne(row.id)) ?? row
```
mayteio: This re-read hits `COMMON_READ` (replica, documented replication delay) - a stale read here defeats the point of re-reading under the lock. Refreshing with an already-rotated refresh token can revoke the grant on strict-rotation providers. Should hit `COMMON_WRITE`.

nodejs/src/integration-gateway/repository.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    /** Persist a successful refresh and clear any prior refresh error. Mirrors Django's write. */
+    async updateAfterRefresh(
+        id: number,
+        config: Record<string, any>,
+        sensitiveConfig: Record<string, any>
+    ): Promise<void> {
+        await this.postgres.query(
+            PostgresUse.COMMON_WRITE,
+            `UPDATE posthog_integration SET config = $1, sensitive_config = $2, errors = '' WHERE id = $3`,
```
mayteio: Unguarded read-modify-write of the full config. User reconnects the integration (Django write) mid-JIT-refresh → this clobbers the new credentials with refreshed-stale ones and bricks it. Redis lock only excludes other gateway heads. Needs a `WHERE` guard on the value read under the lock (or a version column), 0 rows = lost the race, discard.

nodejs/src/integration-gateway/refresh/manager.ts
```diff
@@ -0,0 +1,229 @@
...
+            await this.releaseLock(lockKey)
+            recordRefresh(row.kind, 'refreshed')
+            return updated
+        } catch (error) {
+            // Deliberately do NOT release the lock on failure: its TTL becomes a per-integration
+            // cooldown so a persistently-failing provider is retried at most once per lock window
+            // rather than on every fetch — a lightweight stand-in for Django's refresh backoff.
+            logger.warn('[RefreshManager] token refresh failed', { id: row.id, kind: row.kind, error: String(error) })
+            try {
+                await this.repository.markRefreshFailed(row.id)
```
mayteio: One transient timeout writes `TOKEN_REFRESH_FAILED` → reconnect banner flaps at customers, then retries every 30s forever. The beat has capped backoff + `invalid_grant` terminal detection (#70492). Distinguish transient vs permanent before writing the sentinel, and carry backoff over before the gateway owns any kind for real.

posthog/tasks/integrations.py
```diff
@@ -18,9 +20,22 @@
...
 
-    oauth_integrations = defer_repository_cache_fields(
-        Integration.objects.filter(kind__in=OauthIntegration.supported_kinds).exclude(kind="meta-ads").all()
-    )
+    # Rows owned by the integration gateway's refresher are excluded here so exactly one system
+    # refreshes each row. Ownership is per (kind, team): the gateway enables the same kinds via
+    # INTEGRATION_GATEWAY_REFRESH_KINDS AND the same teams via INTEGRATION_GATEWAY_REFRESH_TEAMS
+    # ("*" = all teams). Gating on team as well as kind means an out-of-rollout team keeps its beat
+    # refresher — a kind is never yanked from the beat for teams the gateway isn't serving yet.
+    oauth_qs = Integration.objects.filter(kind__in=OauthIntegration.supported_kinds).exclude(kind="meta-ads")
```
mayteio: The (kinds, teams) partition now lives in two services' envs and both drift directions fail silently: one way nothing refreshes, the other way both refresh - racing refreshers can invalidate rotating refresh tokens. Wants a single source of truth both sides read, or at least an alert on disagreement.

nodejs/src/cdp/services/managers/integration-manager.service.ts
```diff
@@ -1,33 +1,64 @@
...
+        ids: IntegrationType['id'][],
+        teamId: number
+    ): Promise<Record<IntegrationType['id'], IntegrationType | null>> {
+        // Route through the gateway when it's rolled out to this team; fail open to Postgres so a
+        // gateway outage never blocks credential reads.
+        if (this.gateway?.enabledForTeam(teamId)) {
+            try {
+                const result = await this.gateway.fetchMany(ids, teamId)
+                gatewayRequestsCounter.inc({ result: 'ok' })
+                logger.info('[IntegrationManager]', 'Loaded integrations via gateway service', { ids, teamId })
```
mayteio: Per-fetch info log on the hot path (Postgres branch too) - the counter already covers this. `debug` for both?

nodejs/src/cdp/types.ts
```diff
@@ -520,7 +520,9 @@ export type DBHogFunctionTemplate = {
 export type IntegrationType = {
     id: number
     team_id: number
-    kind: 'slack' | 'email' | 'oauth' | 'firebase' | 'apns'
+    // The DB `kind` column — any IntegrationKind (slack, email, oauth, firebase, apns, hubspot,
+    // salesforce, google-ads, ...). Consumers compare against specific kinds rather than switching.
+    kind: string
```
mayteio, line 551: Non-blocking: widening kills exhaustiveness checks for anything switching on `kind` - worth a quick scan that nothing narrowed on it.

nodejs/src/integration-gateway/audit.ts
```diff
@@ -0,0 +1,28 @@
...
+    caller: string
+    teamId: number
+    requested: number[]
+    resolved: number[]
+    cacheHits: number
+    dbLoaded: number
+    requestId: string
+}
+
+export function emitAudit(event: AuditEvent): void {
```
mayteio, line 20: Non-blocking: audit trail is only as durable as log retention - fine for phase 1, but if this becomes the "who accessed what" answer it wants a real sink eventually.

mayteio (commented):
Strong +1 on direction, want this to land. The execution details are solid (fail-open rollout, single-flight lock, wrong-team = missing, override ignored in prod). Three design pushes before a real rollout, biggest first:

**1. Shared HS256 secret undermines the pitch.** The caller mints its own token and picks its own `team_id`, so team scoping isn't a security boundary - anyone with `CDP_INTEGRATION_GATEWAY_JWT_SECRET` can fetch every team's credentials, pre-decrypted. Not worse than today (same compromise gets keys + DB), but the blast-radius win needs per-caller asymmetric signing (gateway holds public keys only, `caller` bound to the key) + volume alarms on the audit trail. Also worth writing down: consumers keep `ENCRYPTION_SALT_KEYS` until we delete the fail-open fallback, and deleting it makes the gateway a hard hot-path dependency - that trade-off is the real decision and it's currently undocumented.

**2. JIT-only refresh turns background failures into hot-path failures.** Read rollout (`CDP_INTEGRATION_GATEWAY_ROLLOUT`) and refresh ownership (`REFRESH_KINDS/TEAMS`) are independent config on different services - own refresh for a team whose reads aren't routed (or whose integration just isn't read for weeks) and the token silently dies, surfacing exactly when the customer's workflow fires. Keep a background sweep (gateway-owned is fine) with JIT on top. The failure semantics also regress vs the recent refresh hardening (#70492 backoff + terminal `invalid_grant` detection, #70257 `OAUTH_CLIENT_FALLBACKS`): no backoff (30s lock TTL = retry forever), no terminal state, no fallback credentials - so during a client-secret rotation, gateway-owned kinds hard-fail while beat-owned ones survive.

**3. Hot path loses its cache.** Gateway-enabled `getMany` bypasses `LazyLoader` entirely: every invocation pays JWT mint + HTTP RTT, no coalescing, N concurrent invocations = N calls. The gateway's 30s cache protects Postgres, not the RTT, and the 3s timeout becomes tail latency when the gateway wobbles. Client needs a short-TTL cache + in-flight dedupe before the workflows pilot - workflows volume 1:1 onto one service won't survive contact.

Two concrete bugs inline on the refresh write path (replica re-read under lock, unguarded write-back) worth fixing regardless.

nodejs/src/integration-gateway/repository.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Fixed in `82e8e1a`. `updateAfterRefresh` is now a compare-and-swap guarded on the encrypted `refresh_token` read under the lock: `WHERE id = $id AND sensitive_config->>'refresh_token' = <ciphertext>`. 0 rows ⇒ we lost the race (e.g. a concurrent reconnect rotated the token) ⇒ discard our refresh and serve the winner's row instead of clobbering it. Real-DB test covers the miss (stale creds left intact) and the hit.

nodejs/src/integration-gateway/refresh/manager.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Fixed in `82e8e1a`. The under-lock re-read now uses `COMMON_WRITE` (primary) via `fetchOneForUpdate`, so it can't miss a just-rotated token off the replica. Paired with the compare-and-swap on the write (see the repository.ts thread) so a stale read can't cause a clobber either.

nodejs/src/integration-gateway/refresh/manager.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Fixed in `9f6889b`. New `refresh/backoff.ts` mirrors the beat's model (#70492): `record_refresh_failure`/`record_refresh_success`/`refresh_backoff_active`/`oauth_refresh_failure_reason`, same config keys (`refresh_failure_count`, `refresh_invalid_grant_count`, `refresh_next_attempt_at`, `refresh_terminal`), capped exponential backoff, and terminal only after an unbroken `invalid_grant` streak (any other reason resets it, so a transient 5xx can't brick it). The manager now honours backoff/terminal before attempting, classifies the failure reason, and only writes `TOKEN_REFRESH_FAILED` on a real failure — network errors/non-2xx no longer throw. So the state carries over cleanly whichever system last touched the row. Covered by `backoff.test.ts` + manager tests (backoff-skip, terminal-skip, transient-5xx-not-terminal, invalid_grant recording).

nodejs/src/cdp/services/managers/integration-manager.service.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Done in `6cbe163` — both per-fetch lines (gateway and Postgres branch) are now `debug`; the counter remains the prod signal. The once-per-boot "gateway enabled/disabled" line stays at `info` so it's still obvious at startup whether the gateway is wired.

nodejs/src/cdp/types.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Scanned every `IntegrationType.kind` consumer — nothing relied on the union for exhaustiveness. The only reads are equality checks with explicit fallbacks: `push-notification.service.ts` (`=== 'firebase'` / `=== 'apns'` with an `else` that throws `Unsupported push integration kind`) and `email.service.ts` (`!== 'email'`). No `switch` over `kind`, so widening to `string` doesn't drop any check — if anything it removes a false exhaustiveness assumption the runtime guards already cover. No change needed.

posthog/tasks/integrations.py
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Agreed — real risk, not solving it in this PR. Documented the operational contract in the service README (`28c35d5`, "Known follow-ups"): the `(kinds × teams)` partition must be a single source of truth both sides read (or carry a disagreement alert) before any kind is enabled for real, and until then the gateway env and Django settings must be changed together as one step. Kept out of this PR since it's dark by default.

nodejs/src/integration-gateway/audit.ts
benjackwhite: 🤖 _Agent (Claude), on Ben's behalf:_ Agreed — log-based is deliberate for Phase 1. Noted in the README follow-ups (`28c35d5`): if this becomes the authoritative "who accessed what" record it needs a durable append-only sink rather than logs.

PR #72566, opened 2026-07-21
fix(cymbal): load cached frame parts in part order

cat-ph:
## Problem

Native symbolication can resolve one raw frame into several logical frames (an inline expansion). Those are stored as multi-part records in `posthog_errortrackingstackframe`, one row per part, and `part` order is the stack order.

`load_all` fetched them with no `ORDER BY`, so a cache hit returned the parts in whatever order Postgres produced. After a resolver restart (in-memory cache cold, PG cache warm), inline frames could come back reordered: the displayed stack is wrong, and the fingerprint computed over those frames changes, splitting the same crash into a new issue. Once the frame TTL expired and frames were recalculated, the fingerprint flipped back.

## Changes

- `ErrorTrackingStackFrame::load_all` now orders by `part`.
- Regenerated the sqlx offline cache for the changed query.

Added `load_all_returns_parts_in_order`: saves 8 parts in reverse order with index scans disabled (so the heap scan surfaces storage order) and asserts the load comes back part-ordered. Verified it fails without the `ORDER BY` and passes with it. Ran the cymbal records tests plus `SQLX_OFFLINE=true cargo check -p cymbal --lib --tests` and `cargo clippy -p cymbal --all-targets -- -D warnings` locally.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, internal cache behavior.

## 🤖 Agent context

Claude Code authored this from a tester report of fingerprints flipping after a resolver restart and flipping back after the frame TTL expired. I (Claude) confirmed the unordered load in `records.rs`, reproduced it in a test (which needed index scans disabled, since the unique index on `raw_id, team_id, part` can mask the bug by returning rows sorted), and verified the fix. The pre/post-symbol-upload fingerprint split from the same report is a separate name-vocabulary issue, addressed in a stacked PR.

### DIFF
--- a/rust/cymbal/src/core/symbolication/symbol/records.rs
+++ b/rust/cymbal/src/core/symbolication/symbol/records.rs
@@ -156,6 +156,7 @@ impl ErrorTrackingStackFrame {
             SELECT raw_id, part, team_id, created_at, symbol_set_id, contents, resolved, context
             FROM posthog_errortrackingstackframe
             WHERE raw_id = $1 AND team_id = $2
+            ORDER BY part
             "#,
             id.hash_id,
             id.team_id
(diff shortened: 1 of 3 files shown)


rust/cymbal/src/core/symbolication/symbol/records.rs
```diff
@@ -156,6 +156,7 @@ impl ErrorTrackingStackFrame {
             SELECT raw_id, part, team_id, created_at, symbol_set_id, contents, resolved, context
             FROM posthog_errortrackingstackframe
             WHERE raw_id = $1 AND team_id = $2
+            ORDER BY part
```
cat-ph, line 159: not sure how I missed this before, but it's also tricky to notice fwiw, so there are a few possible states:

- in-mem cache miss, pg cache miss: correct order incidentally because we have an index I think
- in-mem cache hit, pg cache doesn't matter: correct order
- in-mem cache miss, pg cache hit: potentially out of order, polluting the in-mem cache too 😬 

I guess ~harder to hit in practice, but we never really noticed it (maybe a bit 'easier' to hit now with the native frame grouping etc)

hpouillot: Wow, good to know !

hpouillot (approved):

rust/cymbal/src/core/symbolication/symbol/records.rs
```diff
@@ -217,6 +218,57 @@ impl ErrorTrackingStackFrame {
...
+    }
+
+    // Multi-part records are one raw frame's inline expansion — part order is
+    // stack order, so a cache hit must return the sequence the resolver saved,
+    // not storage order. The inserts are reversed and index scans disabled so
+    // a plain heap scan surfaces the unordered rows.
+    #[sqlx::test(migrations = "./tests/test_migrations")]
+    async fn load_all_returns_parts_in_order(pool: sqlx::PgPool) {
+        for statement in ["SET enable_indexscan = off", "SET enable_bitmapscan = off"] {
+            sqlx::query(statement).execute(&pool).await.unwrap();
```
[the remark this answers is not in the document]
cat-ph: yeah this is the test, it's fine

PR #72571, opened 2026-07-21
feat(conversations): attribute slack tickets to accounts via channel fallback

abigailbramble:
## Problem

When a Slack ticket's requester can't be identified (Slack profile without an email, or an email that resolves to no org membership and no analytics groups), the ticket ends up with no group association. That means no `$groups` on its analytics events, so group-based workflow triggers can't fire, and the "Related groups" panel on the ticket shows nothing.

But we often *do* know which customer the ticket belongs to: customer analytics accounts store a `slack_channel_id`, and a dedicated support channel maps to exactly one account. The account's `external_id` is the group key at the team's configured account group type.

## Changes

**Backend**
- New customer analytics facade function `get_account_ref_by_slack_channel_id` (JSON-path lookup on account properties, oldest account wins for determinism). The `conversations -> customer_analytics` dependency is declared in `tach.toml`, facade-only.
- The ticket group resolver in `products/conversations/backend/events.py` now falls back to the channel's account when every person-based resolution misses (membership, analytics events, person profile property). Applies to all Slack tickets that end up group-less, including identified-person-but-unresolved-org.
- New nullable `Ticket.organization_id_source` column recording provenance: `person` vs `slack_channel_account`. Persisted alongside `organization_id` at creation and in the late-persistence path on follow-up messages.

> [!NOTE]
> A channel-inferred org says which *account* the ticket belongs to, not who sent it, so follow-up message events reuse the stored org for `$groups` but never create a person profile for it (`process_person_profile` stays false). Legacy rows without a source keep the old behavior.

**Frontend**
- The related groups panel now renders when there's no person at all, showing the creation-snapshot group.
- Channel-inferred groups get a distinct "Customer analytics" tag with an explanatory tooltip, instead of the person-derived "Ticket origin"/stale copy. `RelatedGroups` gained a `highlightLabelTooltip` prop for this.

Unknown Slack requester, ticket attributed via the channel's account:

ticket_full

Tooltip on the tag:

ticket_tooltip

Known limitation (pre-existing): `organization_id` persistence and the panel only engage when the account group type is named `organization` - the same assumption the panel already makes. For other names, events still get `$groups` (workflows work) but the ticket UI shows nothing. Lifting that means keying the UI off the configured index instead of the name; happy to do as a follow-up.

- New backend tests in `test_events.py` (84 pass): the fallback resolves and persists org + source for an unidentifiable Slack requester; parameterized negative cases (no config index, index without a mapping, no matching account, account without `external_id`, non-Slack channel) confirm existing behavior is untouched; person resolution still wins over the channel fallback; and the `process_person_profile` guard on follow-up messages distinguishes `person`/`slack_channel_account`/legacy-null sources. These catch regressions no existing test did because the channel fallback path is new.
- New facade test for `get_account_ref_by_slack_channel_id` (match, no-match, empty channel, cross-team isolation).
- Existing suites: conversations events + tickets API, customer analytics facade + views all pass locally; `tach check --dependencies --interfaces` green; migration is a single additive nullable `ADD COLUMN` (verified via `sqlmigrate`).
- Manually verified end to end on a local stack: seeded an account with a `slack_channel_id`, created a Slack ticket with an unknown requester, confirmed `$groups` on the captured event, the persisted `organization_id`/`organization_id_source`, and the panel rendering above.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with PostHog Code (Claude). Skills invoked: /django-migrations, /improving-drf-endpoints, /writing-tests, /run-posthog. Notable decisions: the fallback lives in the events resolver (not Slack ticket creation) so resolution stays centralized and covered by the existing try/except; provenance is a persisted column rather than inferred, because `organization_id` alone can't distinguish person-derived from channel-derived for the UI and the person-profile guard; the facade lookup orders by `created_at` so duplicate channel claims resolve deterministically. Rebased over ~2,900 upstream commits mid-work - upstream had added its own person-profile-property fallback and late persistence in `capture_message_received`, which this PR composes with (channel fallback runs after all person-based paths, and the late-persistence write now records the source too).

---


products/conversations/backend/events.py
```diff
@@ -264,6 +264,61 @@ def _resolve_org_groups(ticket: Ticket, team: Team) -> tuple[bool, dict | None]:
...
 
+def _resolve_groups_from_slack_channel(team: Team, slack_channel_id: str) -> dict | None:
+    """Attribute the ticket to the customer analytics account bound to its Slack channel.
+
+    Last-resort fallback when the requester can't be tied to an organization: a
+    dedicated support channel maps to one account, and that account's ``external_id``
+    is the group key at the team's configured account group type. Channel-level
+    evidence only — it says which account the ticket belongs to, not who sent it.
+    """
+    group_type_index = team.customer_analytics_config.account_group_type_index
```
arthurdedeus, line 275: this was the only concern I had: making sure we're pulling the correct group type index from the config.

arthurdedeus (approved):
Customer analytics part looks solid! 💪

PR #72588, opened 2026-07-21
fix(error-tracking): tolerate transient offset store failures in notifications consumer

hpouillot:
## Problem

The cymbal notifications consumer crashes with:

```
Panic: failed to store notification offset: Kafka error: Store offset error: State (Local: Erroneous state)
```

`Offset::store()` calls librdkafka's `store_offset`, which returns `RD_KAFKA_RESP_ERR__STATE` ("Erroneous state") when the partition it's storing for is no longer assigned to the consumer. That's the routine result of a consumer-group rebalance, not a fault. The consumer loop treated every store failure as fatal and `panic!`'d, so each rebalance took the process down and put it into a restart loop.

Reported via [error tracking](https://us.posthog.com/error_tracking/019f5c24-2969-7ce2-a215-7c4dc3904f8e).

## Changes

Storing an offset is a local, best-effort operation. If it fails, the offset just stays uncommitted and the notification is redelivered after the rebalance settles, which the loop already tolerates (offsets are stored only after successful handling, so it's at-least-once). So on a store failure I now log a warning and count it in a new `cymbal_notifications_offset_store_errors_total` metric instead of panicking. Both store call sites (the poison-pill path in `validate_notification_with_offset` and the handled-notification path) get the same treatment.

> [!NOTE]
> No behavior change for healthy operation. This only affects what happens when a partition is revoked out from under an in-flight store, where the old code crashed and the new code lets the message be reprocessed.

I could not compile in this environment (no Rust toolchain available), so I have not run `cargo check`/tests here. CI will build and run the cymbal crate. The existing `validate_notification_with_offset` unit test still holds — its `pending_offsets` accounting is unchanged, since that counter is still only incremented after a successful store.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Traced the panic from the stack trace (`consumer_loop.rs:78`) to `Offset::store()` in `common/kafka`, then confirmed the `State (Local: Erroneous state)` message maps to librdkafka's `RD_KAFKA_RESP_ERR__STATE`, returned by `store_offset` for a partition that isn't currently assigned — i.e. a rebalance. Considered swallowing only the specific error code, but a failed local offset store is non-fatal regardless of the exact code (the message simply redelivers), so treating all store failures as a logged, metered warning is simpler and safe.

---

### DIFF
--- a/rust/cymbal/src/modes/notifications/consumer_loop.rs
+++ b/rust/cymbal/src/modes/notifications/consumer_loop.rs
@@ -14,7 +14,8 @@ const NOTIFICATIONS_HANDLED_TOTAL: &str = "cymbal_notifications_handled_total";
 const NOTIFICATIONS_SKIPPED_TOTAL: &str = "cymbal_notifications_skipped_total";
 const NOTIFICATIONS_KAFKA_ERRORS_TOTAL: &str = "cymbal_notifications_kafka_errors_total";
 const NOTIFICATIONS_HANDLE_ERRORS_TOTAL: &str = "cymbal_notifications_handle_errors_total";
-const NOTIFICATIONS_OFFSET_STORE_ERRORS_TOTAL: &str = "cymbal_notifications_offset_store_errors_total";
+const NOTIFICATIONS_OFFSET_STORE_ERRORS_TOTAL: &str =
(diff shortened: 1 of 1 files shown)


cat-ph (approved):
thanks for fixing!

PR #72594, opened 2026-07-21
fix(messaging): cap offline clickhouse query concurrency to avoid 202s

gustavohstrassburger:
## Problem

The `reconcile-precalculated-data` workflow was failing in production with a ClickHouse `TOO_MANY_SIMULTANEOUS_QUERIES` error (code 202) on the offline cluster, where the shared user is capped at 30 concurrent queries.

The failure isn't really this one workflow's fault. The whole messaging workflow family issues async ClickHouse queries against the same offline user, shares one Temporal task queue and one worker process, and caps nothing beyond its fan-out knobs:

- `reconcile-precalculated-data` fans out `team_concurrency` (default 5) teams.
- The six `realtime-cohort-calculation` percentile tiers each spawn up to `parallelism` child workflows (defaults sum to roughly 58 across tiers), each opening its own offline query.

When tiers overlap, the family can push well past the shared ceiling, and ClickHouse rejects the overflow. The reconcile activity exhausts its retries and marks the team failed.

## Changes

Gate every messaging `get_client()` query behind a shared, module-level `asyncio.Semaphore` (`posthog/temporal/messaging/clickhouse_concurrency.py`). Because the whole family runs on one worker event loop, a per-worker semaphore bounds all of them together. This mirrors the existing precedent in `posthog/temporal/data_modeling/activities/materialize_view.py`.

- New setting `MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES` (env-overridable, default 5, well below 30).
- Both reconcile activities and the realtime-cohort activity acquire a slot before opening `get_client()`. Gating both families is the point: reconcile alone contributes only about 5 queries, so the fix only works if the big fan-out shares the same budget.

Over-cap situations now produce in-process backpressure (waiting on the semaphore) instead of a ClickHouse 202.

> [!NOTE]
> The cap is per worker process, so the aggregate is the semaphore size times the number of messaging worker replicas. The value is env-tunable so it can be sized against replica count and the headroom other offline consumers need under the 30 cap. This is deliberately not a distributed (Redis) limiter: the sync `RateLimit` helper has no async precedent, and wrapping it would add risk for little gain here. If per-worker proves insufficient, a distributed limiter is the follow-up.

I (Claude) made the change. Automated checks I ran locally:

- `ruff check` + `ruff format` on the changed files: clean.
- `mypy --cache-fine-grained` on the three changed modules: no issues.
- `py_compile` on all changed files: clean.

I could not run the messaging integration tests (`posthog/temporal/messaging/test/`) or boot the app in this sandbox — `django.setup()` needs Postgres, which isn't available here. Those tests already exercise the wrapped activity code paths via `ActivityEnvironment`, so they cover the change once run in CI. I did not add a bespoke concurrency test: it would either exercise stdlib `asyncio.Semaphore` behavior or need brittle client mocking, and the precedent semaphore has none.

## 🤖 Agent context

Authored by Claude (via PostHog Code), directed by @gustavohstrassburger from a production `ClickHouseTooManySimultaneousQueriesError` on the reconcile workflow.

Approach considered and rejected: reusing the sync Redis `RateLimit` helper from `posthog/clickhouse/client/limit.py` wrapped in `asyncio.to_thread` for a truly distributed cap. No async precedent exists for it and it carries more risk; the per-worker `asyncio.Semaphore` (following `materialize_view.py`) is the idiomatic, low-risk fit. Also considered but skipped: lowering `team_concurrency` or retuning the retry policy — orthogonal churn that doesn't address the cross-workflow aggregate that actually causes the 202.

The default cap was set to 5, conservative because it applies per worker process (aggregate = 5 × replicas).

The change is workflow-versioning safe: all edits are to activity implementations and settings, with no changes to any `@workflow.defn` command sequence, so no `workflow.patched()` gate is needed.

---


gustavohstrassburger:
@greptile review

gustavohstrassburger:
Addressed the AI review feedback in the latest commits:

- **Zero-limit stall (Greptile P1):** the semaphore is now floored at 1 (`asyncio.Semaphore(max(1, ...))`), so a misconfigured `MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES=0` can no longer deadlock every gated activity.
- **Ungated backfill queries (Copilot):** the four backfill workflow/coordinator `get_client()` sites (events + person-properties) now acquire the same shared slot, so the per-worker cap covers the whole messaging family.
- **Misleading comment (Copilot):** reworded to "default 5" since `team_concurrency` is env-configurable, not hard-capped.

On the **deployment-wiring P1 (Greptile):** the chart/secret changes live in the separate `posthog/charts` and `posthog/secrets` repos, so they are out of scope for this monorepo PR and will be wired there as a follow-up. The default is intentionally safe without wiring — it is now 5 per worker process (the comment referenced the old default of 15), which keeps the messaging family comfortably under the offline user's 30-query ceiling across a small replica count. Leaving the threads open for a human to resolve.

posthog/temporal/messaging/clickhouse_concurrency.py
```diff
@@ -0,0 +1,14 @@
...
+# Limits concurrent offline-cluster ClickHouse queries across the messaging workflow family
+# (reconcile-precalculated-data and realtime-cohort-calculation). They all run on one worker
+# process with a single event loop, so this module-level semaphore gates every messaging activity
+# on the same worker. Acquiring a slot before opening `get_client()` turns an over-cap situation
+# into in-process backpressure (waiting) instead of a ClickHouse TOO_MANY_SIMULTANEOUS_QUERIES
+# rejection. The cap is per worker process; aggregate = this value times messaging worker replicas.
+#
+# Floored at 1: a size of 0 (misconfigured env) would deadlock every gated activity forever, since
+# no permit is ever released.
+messaging_clickhouse_query_slot = asyncio.Semaphore(max(1, settings.MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES))
```
haacked: `question`: The semaphore limits concurrency per worker, but the ClickHouse `TOO_MANY_SIMULTANEOUS_QUERIES` ceiling (30) is enforced cluster-wide across the shared offline default user. [materialize_view.py](https://github.com/PostHog/posthog/blob/09368cc1f6d60ec97a226ea53eb97ed5a2f956ba/posthog/temporal/data_modeling/activities/materialize_view.py#L51-L52) runs its own independent semaphore (cap 10) for the same offline user, on a separate task queue/worker process/event loop, so no in-process semaphore can coordinate across the two families. The real aggregate is 5 x messaging replicas + 10 x data_modeling replicas + other offline consumers.

Is that comfortably under 30 at current replica counts, or does this mainly buy time until the distributed limiter mentioned as a follow-up lands?

posthog/settings/temporal.py
```diff
@@ -145,6 +145,15 @@ def _set_temporal_task_queue(task_queue: str) -> str:
...
 VIDEO_EXPORT_TASK_QUEUE = _set_temporal_task_queue("video-export-task-queue")
 MESSAGING_TASK_QUEUE = _set_temporal_task_queue("messaging-task-queue")
+# Per-worker cap on concurrent offline-cluster ClickHouse queries issued by the messaging
+# workflow family (reconcile-precalculated-data + realtime-cohort-calculation). They share one
+# worker/event loop and the offline `default` user (max_concurrent_queries=30), so an unbounded
+# fan-out across tiers can trip TOO_MANY_SIMULTANEOUS_QUERIES. Kept safely below 30, leaving
+# headroom for other offline consumers. Note this is per worker process, so the aggregate is this
+# value times the number of messaging worker replicas — size it accordingly.
+MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES: int = get_from_env(
+    "MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES", 5, type_cast=int
```
haacked: `suggestion`: The default cap (5) equals reconcile's team_concurrency default (5), and reconcile holds one slot per team for the whole streaming context (all batched queries plus Kafka produce/eval), not just the ClickHouse call. Five concurrent reconcile teams can occupy every slot for the full duration of a chunk, so any realtime-cohort activity scheduled on the same worker during that window queues for a free slot. The Heartbeater keeps heartbeats flowing while blocked, but realtime-cohort's 60-minute start_to_close_timeout keeps counting, so a long reconcile chunk could make a cohort calculation time out with zero work done instead of hitting a retryable 202.

`RECONCILE_PRECALCULATED_DATA_TEAM_CONCURRENCY` is env-configurable and uncapped in code, but raising it won't actually speed up reconcile once team_concurrency exceeds the shared semaphore size, since extra teams just queue on the semaphore instead of running more ClickHouse queries concurrently. Add a comment tying the two settings together.

```suggestion
    "MESSAGING_CLICKHOUSE_MAX_CONCURRENT_QUERIES", 10, type_cast=int
```

posthog/temporal/messaging/clickhouse_concurrency.py
haacked: `suggestion`: The slot-acquire is inlined as `async with messaging_clickhouse_query_slot, get_client(...)` at all 7 call sites across 6 files. The invariant that every messaging ClickHouse query must hold the slot now lives only in reviewer discipline; a future call site added without the wrapper would silently defeat the cap, reproducing the exact production failure this PR fixes.

A small async context manager would make the invariant impossible to forget.

```python
@asynccontextmanager
async def get_messaging_client(team_id: int):
    async with messaging_clickhouse_query_slot, get_client(team_id=team_id) as client:
        yield client

# call sites become:
async with get_messaging_client(inputs.team_id) as client:
```

posthog/temporal/messaging/realtime_cohort_calculation_workflow.py
```diff
@@ -434,7 +435,7 @@ def build_query(cohort_obj):
                     row_processing_start_time = None
                     rows_processed = 0
 
-                    async with get_client(team_id=cohort.team_id) as client:
+                    async with messaging_clickhouse_query_slot, get_client(team_id=cohort.team_id) as client:
```
haacked: `question`: `process_realtime_cohort_calculation_activity` contains this new semaphore acquire but is never invoked by `test_realtime_cohort_calculation_workflow.py`, which only covers helper functions and coordinator input parsing. So at this call site the new `async with messaging_clickhouse_query_slot` line has zero test coverage, unlike the other six sites which do execute in their respective tests.

Was that intentional, or is there coverage elsewhere?

posthog/temporal/messaging/clickhouse_concurrency.py
haacked: `nit`: `messaging_clickhouse_query_slot` suggests one slot but grants N concurrent permits. `messaging_clickhouse_query_semaphore` is more accurate and matches the precedent's `_clickhouse_query_semaphore`.

haacked (approved):
Solid fix, matches the existing precedent well. A couple sizing questions and a maintainability suggestion inline.

gustavohstrassburger:
Addressed @haacked's review in c582ffe:

**Centralize the slot-acquire (suggestion) + rename (nit):** added a `get_messaging_client` async context manager in `clickhouse_concurrency.py` and routed all 7 call sites (6 workflows) through it, so the "every messaging CH query holds the slot" invariant lives in one place instead of reviewer discipline. Renamed the semaphore to `messaging_clickhouse_query_semaphore` to match the `data_modeling` precedent. The backfill test patch targets moved from `<workflow>.get_client` to `clickhouse_concurrency.get_client` accordingly.

**Default cap 5 → 10 + settings comment (suggestion):** took the suggested 10 and added a comment tying it to `RECONCILE_PRECALCULATED_DATA_TEAM_CONCURRENCY` — since a reconcile activity holds one slot for its whole streaming context, this cap is the effective ceiling on reconcile parallelism, and keeping it >= team_concurrency avoids a burst of reconcile teams starving realtime-cohort while it queues.

**Realtime-cohort coverage (question):** it was an oversight, not intentional. Added a focused test that runs `process_realtime_cohort_calculation_activity` with a zero-row stream and asserts it opens its client through the gated `clickhouse_concurrency.get_client`. If that site ever stops going through `get_messaging_client` (how the 202 could recur), the test fails.

**Cross-family aggregate under 30 (question):** you're right that no in-process semaphore coordinates across families — `data_modeling` has its own independent cap on a separate worker, so the real cluster-wide budget is the sum across families plus other offline consumers. This PR keeps the messaging family bounded per worker and buys headroom; a distributed limiter remains the follow-up if the summed aggregate gets tight. I documented the non-coordination in the `clickhouse_concurrency.py` module comment so the sizing constraint is visible at the source.

PR #72605, opened 2026-07-21
feat(conversations): bulk add and remove tags on tickets

clr182:
## Problem

Support agents triaging the tickets list can select multiple tickets, but the only bulk action was "Mark as", which just changes status. Applying or clearing tags across a batch meant opening each ticket by hand.

## Changes

- Bulk tag editing from the tickets list: select tickets, then add or remove tags from the "Update tickets" dropdown (it shows which tags are on all vs. some of the selection, with click-to-toggle). Add is a union, remove drops just the named tags.
- Reuses the shared tagging endpoint. `TicketViewSet` already inherits `TaggedItemViewSetMixin`, so this rides its `bulk_update_tags` action (the same one feature flags / insights / dashboards use) rather than adding bespoke endpoints. Because tickets are UUID-keyed, it overrides the action with the UUID request serializer and `apply_bulk_tag_changes` — the same pattern event definitions use. Team membership is the boundary (matching the single-ticket update path); tag changes are logged to the ticket timeline by the existing `TaggedItem` signal.
- Guards the write amplification an automated review flagged: a per-request cap (tickets × tags) plus a per-user burst/sustained throttle on the action.

Automated only. I (well, Claude) did not verify this in a browser.

- `TestBulkUpdateTags`: add merges with existing tags (guards against a replace that would wipe tags), remove drops only the named tags, set replaces all, other-team ids are skipped (team isolation), empty tags for add → 400, and over-cap requests → 400 before any write.
- Frontend typecheck, `mypy` on the changed files, ruff, and oxfmt/oxlint clean for the changed files.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

I directed this; Claude Code (Opus 4.8) wrote the code across a few sessions. Skills invoked: `/improving-drf-endpoints` and `/writing-tests`.

The path here: it first shipped as two bespoke endpoints (`bulk_add_tags` / `bulk_remove_tags`). A reviewer asked whether it could reuse the shared bulk-tag machinery (`BulkTaggableResource` / `TaggedItemViewSetMixin.bulk_update_tags`). It can — the only blocker was that the shared request serializer is integer-keyed and tickets use UUIDs, so we override the inherited action with the UUID serializer (no new endpoint route; it was already inherited). We tried the shared `BulkUpdateTagsButton` on the frontend too but kept the bespoke ticket UI, which shows current tag state and patches optimistically — the generic button doesn't. So: shared backend, bespoke frontend, with `api.ts` translating the UI's add/remove calls to the unified endpoint.


luke-belton (changes requested):
can't we reuse `BulkTaggableResource` here?

clr182:
> can't we reuse `BulkTaggableResource` here?

Ok good idea, taking a look

PR #72607, opened 2026-07-21
feat(data-warehouse): improve person-property custom property UX

Gilbert09:
## Problem

The warehouse-backed person custom properties feature (mapping warehouse table columns onto persons, from #70090) shipped with rough edges in the "New custom property" modal:

- The warehouse-table dropdown only ever showed the first 100 tables (DRF default page size) and search was client-side only, so tables past the first page were unreachable.
- The distinct-ID column and the column mappings were free-text inputs — you had to know and type exact column names.
- Tables were labelled by their raw storage identifier (e.g. `slack_c0b3he1pv53`) instead of the dotted name they're queried by.
- The account-only "Type" control (plus big-number / options) showed on the person form, where it does nothing — a person property is written as a raw `$set` value.

There was also no way to see or manage these person mappings from project settings, and the accounts config table gave no indication of whether a property attaches to a person or an account.

## Changes

- **Warehouse-table picker** now lists every synced table (drops the 100-row cap; requests `include_columns=false` so the list endpoint skips the expensive per-table HogQL column serialization) and labels each by its dotted HogQL name (e.g. `stripe.charges`).
- **Distinct-ID column and mapping columns** are now dropdowns of the selected table's real columns, fetched on selection. Free-typing stays allowed (`allowCustomValues`) so a column auto-detection misses can still be entered.
- **Person form cleanup:** the account-only Type / big-number / options controls are hidden for person-target properties, and `display_type` is forced to `text` on save.
- **Accounts config table** gains an "Attach to" (Person / Account) column.
- **Project settings:** new "Person properties from the warehouse" entry under Settings → Project → Product analytics (gated on the `WAREHOUSE_PERSON_PROPERTIES` flag) with a table of the mappings and add / edit / delete, opening the modal straight into Person.
- **Backend:** `TableSerializer` gains a `hogql_name` field and honors `include_columns=false` (skipping both column serialization and the HogQL database build).

- Added `test_list_tables_exposes_hogql_name_and_honors_include_columns` (data-warehouse table API). It asserts the list response exposes the dotted `hogql_name` and that `include_columns=false` returns tables with an empty `columns` array — guarding both the picker's label source and the cheap-list path, neither of which an existing test covered. Passes locally.
- Ran `ruff check` / `ruff format` on the backend files and `oxlint` / `oxfmt` on the changed frontend files — all clean.
- Regenerated OpenAPI/frontend types with `hogli build:openapi` on a current-master base: the only generated change is `hogql_name` in the data-warehouse `api.schemas.ts`.

I did not run the app or exercise the modal in a browser, so there are no screenshots — the frontend changes are covered by the kea logic and typing only.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs under `docs/` cover this modal; nothing to update.

## 🤖 Agent context

- Tool: Claude Code. Skills invoked: `/improving-drf-endpoints` (the `TableSerializer` change) and `/writing-tests` (the new backend test).
- Key decisions: fetch the selected table's columns on demand rather than serializing columns for every table in the list (keeps the list cheap via `include_columns=false`); use `LemonInputSelect` with `allowCustomValues` for the column fields so undetected columns can still be typed; reuse the existing `customPropertyDefinitionsLogic` + modal for the settings surface, adding an optional target to `openCreateModal` so the settings entry opens straight into Person; force `display_type` to `text` for person targets since it only drives account-property rendering.
- The `hogql_name` addition to the generated `api.schemas.ts` was verified against a full `hogli build:openapi` run — the canonical build produces no further data-warehouse/MCP/zod drift.


rafaeelaudibert (approved):

products/customer_analytics/frontend/scenes/CustomerAnalyticsConfigurationScene/account/WarehousePersonPropertiesSetting.tsx
arthurdedeus, line 1: should this be in `CustomerAnalyticsConfigurationScene/person/WarehousePersonPropertiesSetting.tsx` instead?

arthurdedeus (approved):
lgtm!

PR #72617, opened 2026-07-21
feat(data-warehouse): implement mercury import source

Gilbert09:
## Problem

Mercury (mercury.com) is a widely used US business banking platform, and teams want their banking data (accounts, transactions, cards, invoices) in the PostHog data warehouse alongside product data. A scaffolded stub for the `mercury` source existed but was hidden (`unreleasedSource=True`) with no fields, schemas, or sync logic.

**Why:** finish the scaffolded Mercury connector end-to-end so it ships visible and connectable, following the warehouse-sources architecture contract.

## Changes

Implements the Mercury source on the shared `rest_source` framework, following the `source.py` / `settings.py` / `mercury.py` split:

- **11 endpoints**: Accounts, Cards, Categories, CreditAccounts, Customers (AR), Events, Invoices (AR), Recipients, Transactions, TreasuryAccounts, Users. Endpoint paths, response shapes, and pagination were verified against the live API (route probing) and the official OpenAPI specs embedded in docs.mercury.com.
- **Pagination**: Mercury uses cursor pagination (`page.nextPage` UUID replayed as `start_after`), which maps directly onto the framework's `JSONResponseCursorPaginator`. `/credit` is a single-page list. All HTTP rides the tracked, retrying transport (no second retry layer).
- **Resumable**: `ResumableSource` with the cursor checkpointed after each yielded page, seeded back on resume (same pattern as Chargebee).
- **Incremental sync on Transactions only**: `/transactions` is the only list endpoint with a genuine server-side timestamp filter (`start`, earliest `createdAt`). The watermark is sent as ISO 8601 with explicit `order=asc`, and a default 30-day incremental lookback re-reads recent rows so pending transactions that later change status/`postedAt` get merged. Everything else is full refresh.
- **Release state**: `unreleasedSource` removed, shipped with `releaseStatus=ReleaseStatus.ALPHA`. Bearer token auth (single password field); `validate_credentials` accepts 403 at source-create so custom-scoped tokens work.
- **Docs plumbing**: `lists_tables_without_credentials=True` plus `canonical_descriptions.py` sourced from the official API reference, so the posthog.com doc renders the full table catalog.
- `SOURCES.md`: moved `mercury` from Scaffolded to Implemented (HTTP, `requests + rest_source.RESTClient`, tracked).

Decisions and caveats:

- **No webhook support (deliberate)**: Mercury webhooks deliver Events-API envelopes containing RFC 7396 JSON merge patches (only changed fields), not full resources. Upserting those into a Delta table would null out unchanged columns, so `WebhookSource` was left out; safe webhook sync would need per-event rehydration via `GET /transactions/{id}` and can be a follow-up.
- **Unverified against a live account**: I had no Mercury credentials, so authenticated responses couldn't be curled. Endpoint existence, auth error shapes (`401 noTokenInDB`), pagination params, and response schemas were verified via live unauthenticated probes plus the official OpenAPI definitions. The API docs don't explicitly state which column `order=asc` sorts by on `/transactions`; the incremental design (stable `createdAt` cursor + 30-day lookback + merge-on-id) is conservative against that uncertainty. Hence ALPHA.
- The icon (`frontend/public/services/mercury.png`) was already committed, so no icon work was needed.

- `pytest products/warehouse_sources/backend/temporal/data_imports/sources/mercury/tests/` (73 passed). Regressions the new tests catch that nothing existing did: re-adding `unreleasedSource` (would hide the source), losing the 403-at-create acceptance (would block custom-scoped tokens), losing the resume checkpoint/seed behavior (mocked-session end-to-end pagination), sending the incremental `start` filter when disabled or omitting it on first sync, and wrong per-endpoint primary keys (`Users` keys on `userId`) or unstable partition keys.
- Registry-wide suites: `test_source_categories.py` + `test_source_versions.py` (2593 passed), `test_generated_configs.py` + `test_ci_core_coupled_sources.py` + `test_source_config_generator.py` (47 passed).
- `ruff check`/`ruff format` clean; `mypy` clean on the new package; `hogli ci:preflight --fix` reports 0 failures.
- Not done (no credentials): a real end-to-end sync against a live Mercury account.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

The posthog.com doc (`contents/docs/cdp/sources/mercury.md`, matching `docsUrl`) needs to land in the posthog.com repo, which wasn't available in this environment. Authored content, following the canonical source-doc template:

<details>
<summary>contents/docs/cdp/sources/mercury.md</summary>

```markdown
---
title: Linking Mercury as a source
sidebar: Docs
showTitle: true
sourceId: Mercury
beta: true
---

<AlphaRelease />

The Mercury connector syncs your business banking data into PostHog, including bank accounts, transactions, cards, recipients, invoices, and more. Combine it with your product data to analyze revenue, spend, and cash flow alongside user behavior.

## Prerequisites

- A [Mercury](https://mercury.com/) account
- Permission to create API tokens in your Mercury organization

## Adding a data source

<SourceSetupIntro />

You need a Mercury API token to connect:

1. In Mercury, go to [Settings > API tokens](https://app.mercury.com/settings/tokens).
2. Create a new token with **Read Only** access. Read Only tokens don't require an IP allowlist.
3. Copy the token right away. Mercury only shows it once.
4. Paste it into the **API token** field in PostHog.

Custom-scoped tokens also work. Tables the token can't read will fail to sync, so grant read access to the data you want to import.

## Sync modes

<SyncModes />

The `transactions` table supports incremental sync on `createdAt`. Because pending transactions can change status after they're created, each incremental run re-reads a trailing 30 day window by default so late updates are merged in. All other tables are small lists that sync with full refresh.

## Configuration

<SourceParameters />

## Supported tables

<SourceTables />

## Troubleshooting

- **Invalid API token**: Mercury returns a 401 error when the token is wrong or has been revoked. Generate a new token in Mercury under Settings > API tokens and update the source.
- **Permission errors on some tables**: a custom-scoped token returns 403 for endpoints outside its scopes. Either grant the missing scope in Mercury or disable syncing for those tables.

<TroubleshootingLink />
```

</details>

## 🤖 Agent context

Authored by Claude via PostHog Code. Skills invoked: `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, `/writing-tests`. Chargebee was used as the canonical reference (declarative `RESTAPIConfig`, framework paginator, resume hooks); Attio for the cursor-pagination and full-refresh patterns. Notable rejected option: `WebhookSource` support, dropped after finding Mercury webhooks carry JSON merge patches rather than full resources (details under Changes). Endpoint behavior was verified with live route probing (401 vs 404 discrimination) and the OpenAPI definitions published in Mercury's docs, since no credentials were available.

---


Gilbert09:
CI note: the two failing warehouse-sources shards are not caused by this PR. They fail in pre-existing migration tests (`test_migration_0028.py`, `test_migration_0052.py`) with `OperationalError: the connection is closed`, and the same failure hits other unrelated open PRs while master stays green.

Root cause (reported to devex): the CI stack's Postgres (`docker-compose.dev.yml`) sets `idle_in_transaction_session_timeout=300000` (5 min). `BaseTestMigrations.setUp` spends minutes of pure CPU building migration project state inside the open test transaction, and PR shards run pytest with `--cov=backend` (master runs without coverage), which slows that stretch past 300s of idle-in-transaction. Postgres then terminates the session. Master's runs finish each migration test in ~4 min, just under the threshold, so only PR runs trip it.

Fix belongs in shared CI/test machinery (exempt migration tests from coverage, raise the idle timeout for CI, or move the state-building out of the transaction), so I have not changed anything here. I've re-run the failed shard; it's borderline against the threshold, so retries can pass (shard 2 did on its first retry).

MarconLP (approved):

PR #72632, opened 2026-07-21
feat(data-warehouse): implement simfin import source

Gilbert09:
## Problem

The SimFin source existed only as a hidden scaffold (`unreleasedSource=True`, no fields, no sync logic). Teams that want standardized fundamentals, share prices, and company metadata for equities in the Data warehouse had no way to connect SimFin.

**Why:** finish the scaffolded connector end to end so SimFin is a visible, connectable source.

## Changes

Implements the SimFin source against SimFin Web API v3 (`https://backend.simfin.com/api/v3`) and ships it released with `releaseStatus=ReleaseStatus.ALPHA`.

- **v3, not v2.** The task research pointed at the v2 API (`simfin.com/api/v2`, api-key query param), but a live probe shows v2 now 404s while v3 responds. The implementation targets v3, which authenticates with an `Authorization: api-key <KEY>` header, and pins `supported_versions=("v3",)`.
- **Nine tables** from the endpoints users actually want: `companies`, `company_details`, `income_statements`, `balance_sheets`, `cash_flow_statements`, `derived_ratios`, `share_prices`, `common_shares_outstanding`, `weighted_shares_outstanding`. Endpoint catalog, primary keys, and partition keys are declarative in `settings.py`; transport and reshaping live in `simfin.py`; registration and wiring in `source.py`, per the warehouse-sources architecture contract.
- **Full refresh only.** SimFin has no server-side change cursor. The `start`/`end` params filter on the record's own date, and price history is split-adjusted retroactively, so they can't safely drive incremental sync. No table advertises `supports_incremental`.
- **Hand-driven requests through `make_tracked_session()`** rather than the declarative `rest_source` config: the API has no pagination at all, and the compact responses are columnar (`columns` + `data` arrays wrapped per company) needing per-item reshaping a `data_selector` can't express. No tenacity layer on top; SimFin rate limits with real HTTP 429s, which the tracked transport already retries honoring `Retry-After`.
- **Per-ticker fan-out**, capped at 100 distinct tickers (validated at source create and re-checked at sync time), so a saved config can't fan out into unbounded request volume. Statement/price tables partition on stable date columns (`report_date` / `date`).
- `get_non_retryable_errors` maps 401/403 to permanent failures with user-facing guidance, `canonical_descriptions.py` documents tables and shared columns from the official API docs, `lists_tables_without_credentials=True` lets the public docs render the table catalog, and SOURCES.md moves simfin into the implemented table.

I wasn't able to run against the live API end to end (no SimFin credentials in this environment), so endpoint behavior is verified three ways: unauthenticated probes against the live API (v2 dead, v3 alive, 401 semantics), the official OpenAPI definitions from SimFin's docs, and the response-parsing code of SimFin's official API clients. The positional column order of the shares-outstanding endpoints is undocumented and taken from the official clients; it's flagged with a code comment.

Automated tests (all passing):

- `tests/test_simfin.py` (34 cases): columnar reshaping and company-field injection (a zip misalignment or dropped injection would corrupt every primary key), positional shares mapping, loud failure on unexpected response shapes instead of silent truncation, per-ticker fan-out carrying the statement-type param on every request, empty tickers skipped without failing the sync, the 401 error message shape that `get_non_retryable_errors` keys on, auth header and key redaction, and credential validation status mapping.
- `tests/test_simfin_source.py` (21 cases): schema catalog is full-refresh only with the declared primary keys, ticker validation short-circuits before probing, `source_for_pipeline` plumbing including API-version resolution, oversized saved configs rejected, and the source ships visible (no `unreleasedSource`) as alpha.
- Cross-source suites (`test_source_versions`, `test_source_categories`, 2593 tests), `mypy` on the package, `ruff check`/`format`, and `hogli ci:preflight --fix` all pass. `manage.py audit_source_docs` reports no simfin issues against the docs PR below.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

User-facing doc: PostHog/posthog.com PR linked in the first comment (added after creation so the PRs cross-reference). `docsUrl` points at `https://posthog.com/docs/cdp/sources/simfin`, matching the doc filename.

## 🤖 Agent context

Authored by Claude (PostHog Code session) directed by Tom Owers. Skills invoked: `implementing-warehouse-sources`, `writing-tests`, `documenting-warehouse-sources`. Key decisions: switched from the researched v2 API to v3 after live probes showed v2 returns 404; chose the compact (columnar) response format because its shape is verifiable against SimFin's official clients; chose hand-driven tracked-session requests over the declarative `rest_source` framework because the API has no pagination and needs per-item columnar reshaping; kept every table full refresh because date filters scope record dates (not modification times) and price history restates on splits.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/SOURCES.md
@@ -477,6 +477,7 @@ the row lists both.
 | shortcut                         | HTTP                        | requests                                                        | ✅                          |
 | shortio                          | HTTP                        | requests                                                        | ✅                          |
 | signoz                           | HTTP                        | requests                                                        | ✅                          |
+| simfin                           | HTTP                        | requests                                                        | ✅                          |
 | simplecast                       | HTTP                        | requests                                                        | ✅                          |
 | simplesat                        | HTTP                        | requests                                                        | ✅                          |
 | skyvern                          | HTTP                        | requests                                                        | ✅                          |
@@ -923,7 +924,6 @@ doesn't conflict with concurrent PRs.
 - sigma_computing
 - signnow
 - sim
-- simfin
 - simplecast
 - simplesat
 - singular
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs/simfin.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs/simfin.py
@@ -6,4 +6,5 @@ from products.warehouse_sources.backend.temporal.data_imports.sources.common imp
 
 @config.config
 class SimFinSourceConfig(config.Config):
-    pass
+    api_key: str
+    tickers: str
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/simfin/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/simfin/canonical_descriptions.py
@@ -1,0 +1,115 @@
+from products.warehouse_sources.backend.temporal.data_imports.sources.common.canonical_descriptions import (
+    CanonicalDescriptions,
+)
+
+# Statement tables share the company wrapper fields plus the standardized statement metadata columns;
+# the per-metric columns (Revenue, Total Assets, ...) vary by statement template and fall back to the
+# LLM with the docs_url below.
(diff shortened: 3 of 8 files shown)


Gilbert09:
User-facing doc PR: PostHog/posthog.com#18745 (draft; should merge once this lands).

MarconLP (approved):

PR #72636, opened 2026-07-21
feat(data-warehouse): implement hoorayhr import source

Gilbert09:
## Problem

HoorayHR (a Dutch all-in-one HR tool for SMEs) had a scaffolded warehouse source stub hidden behind `unreleasedSource=True`, with no fields, schemas, or sync logic. Users can't pull their HR data (time off, sick leave, contracts, time tracking) into the Data warehouse to analyze alongside product data.

**Why:** finish the scaffolded connector end-to-end and ship it visible, so HoorayHR customers can connect it.

## Changes

Implements the source on the shared `rest_source` framework, following the `source.py` / `settings.py` / `hoorayhr.py` architecture contract (campayn/chargebee as references):

- **Transport** (`hoorayhr.py`): declarative `RESTAPIConfig` with framework `bearer` auth and the `single_page` paginator, riding the tracked, retrying transport. Credential validation probes `/leave-types` via `validate_via_probe` with the key registered for redaction.
- **Endpoint catalog** (`settings.py`): 15 endpoints (users, time_off, leave_types, contracts, sick_leave_dossiers, sick_leave_phases, time_tracking, availability, entities, labels, employment_terms, employment_term_assignments, document_categories, work_location_categories, teams_information). All full refresh: the API documents no pagination and no server-side timestamp filters (matching the Airbyte connector's capabilities). Partitioned monthly on the stable `createdAt` field where the schema has one.
- **Source** (`source.py`): `SimpleSource` (nothing to resume with single-page responses), single API-key password field, `get_non_retryable_errors` for 401/403, `lists_tables_without_credentials=True`, `docsUrl` pointing at `/docs/cdp/sources/hoorayhr`. Deleted `unreleasedSource=True` and shipped with `releaseStatus=ReleaseStatus.ALPHA`.
- **Canonical descriptions**: table/column descriptions sourced from the official OpenAPI spec.
- SOURCES.md: moved hoorayhr from Scaffolded to the Implemented table.

Endpoint shapes, paths, auth scheme, and the absence of pagination were verified against the live API's published OpenAPI spec (`https://api.hoorayhr.io/swagger.json`, v4.56.0) plus unauthenticated probes of the endpoints (401 responses confirmed paths and the FeathersJS-style error shape). I couldn't verify authenticated response behavior end-to-end without a HoorayHR account, hence ALPHA. Report-style endpoints requiring date-range params (`/attendance-report`, `/working-today`, `/public-holidays`) are deliberately excluded as they're point-in-time views, not warehouse tables.

The user-facing doc is added separately to the posthog.com repo (audit_source_docs passes for this source against that checkout).

- `pytest products/warehouse_sources/.../hoorayhr/tests/` (25 passed): transport tests catch endpoint/auth wiring regressions (wrong path, dropped bearer header, a second request appearing where the API has no pagination), primary-key/partition config regressions (e.g. `teams_information` losing its `teamId` key or gaining a partition on a field it doesn't have), credential-probe status mapping, and the non-retryable error strings matching real `requests` error formats. The source tests also lock in that `unreleasedSource` stays gone and that canonical description keys can't drift from the endpoint catalog.
- Registry-wide guards (`test_source_categories`, `test_source_versions`): 2593 passed.
- `ruff check` / `ruff format` clean; `hogli ci:preflight --fix` passes; repo-wide `mypy --cache-fine-grained .` shows no errors in changed files (one pre-existing unrelated error in `posthog/personhog_client/converters.py`).
- Not done: an authenticated sync against a real HoorayHR account (no credentials available).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Doc for posthog.com written per the documenting-warehouse-sources skill; landing as a separate PR in the posthog.com repo.

## 🤖 Agent context

Built with Claude Code following the `implementing-warehouse-sources`, `writing-tests`, and `documenting-warehouse-sources` skills. Key decisions: preferred the shared `rest_source` framework over a hand-rolled client (the API is fully declarative: bearer auth + single-page bare-array responses); chose `SimpleSource` since there's no cursor to resume; kept every table full refresh because no server-side timestamp filter is documented and none could be curl-verified without credentials; used the current API-key auth from the official docs rather than the legacy email/password login the Airbyte connector uses.

---


Gilbert09:
Companion doc PR: [PostHog/posthog.com#18748](https://github.com/PostHog/posthog.com/pull/18748)

MarconLP (approved):

PR #72653, opened 2026-07-21
feat(data-warehouse): implement zapsign import source

Gilbert09:
## Problem

ZapSign (a Brazilian e-signature platform) was a scaffolded warehouse source stub: registered with `unreleasedSource=True`, no fields, no sync logic. Users can't pull their e-signature data (documents, signers, templates) into PostHog to analyze signing funnels and completion rates.

## Changes

Implements the ZapSign source end to end on the shared `rest_source` framework and ships it released with `releaseStatus=ReleaseStatus.ALPHA` (the `unreleasedSource` flag is removed).

- **Base class:** `ResumableSource + WebhookSource` (`source.py` / `settings.py` / `zapsign.py` split per the architecture contract).
- **Tables:**
  - `documents` - `GET /api/v1/docs/` with the DRF-style `next`-link paginator, `sort_order=asc`, and `include_signers=true`. Incremental sync uses ZapSign's server-side `created_from` date filter mapped from the `created_at` watermark (merge-only, since the date filter is day-granular and creation-date-only). Datetime partitioning on `created_at`.
  - `templates` - full refresh (the API exposes no filter or sort params).
  - `signers` - declarative single-hop fan-out (docs list -> per-document detail, `data_selector: signers`), with the parent document token injected and included in the composite primary key.
- **Resume:** pagination checkpoints (next-URL, or the fan-out completed/current state) are saved to Redis after each yielded page, keyed per endpoint.
- **Webhooks:** `create_webhook` registers a ZapSign webhook for all document events (`type: ""`) and attaches a generated `Authorization` header that the new Hog template verifies on every delivery. Incoming events are full document payloads, so they merge straight into the `documents` table on `token` after a per-batch dedupe keeping the latest `last_update_at`. ZapSign has no list-webhooks endpoint (delete needs the id from creation time), so `delete_webhook` reports that removal is manual and the setup caption says so.
- **Config:** API token (password) + environment select (production / sandbox hosts). Pagination is host-pinned via `allowed_hosts` so a tampered `next` link can't exfiltrate the token.
- **Extras:** `canonical_descriptions.py` from the official API docs, `lists_tables_without_credentials=True` so the public docs render the table catalog, `get_non_retryable_errors` covering ZapSign's 403-for-bad-token behavior (verified against the live API), SOURCES.md moved to Implemented.

Docs: companion posthog.com PR [PostHog/posthog.com#18746](https://github.com/PostHog/posthog.com/pull/18746) adds `/docs/cdp/sources/zapsign`, matching `docsUrl`. The icon (`frontend/public/services/zapsign.png`) was already committed. No new env vars.

> [!NOTE]
> Endpoint behavior was verified against ZapSign's current API docs plus unauthenticated probes of the live API (auth error shape, reachability). I couldn't run an authenticated end-to-end sync without a ZapSign account, so the response-envelope assumption is defended in code: `data_selector_required=True` makes a shape change fail loud instead of silently syncing zero rows.

- `pytest products/.../sources/zapsign/tests/` - 50 tests pass. `test_zapsign.py` covers transport behavior against wire-level prepared requests: next-link pagination, incremental `created_from` derivation (and its absence on full refresh), sandbox host routing, resume save-after-yield / seed-on-restart / cross-endpoint isolation, signers fan-out row shape with parent token injection, webhook-mode routing, webhook payload dedupe/reshape, credential-validation status mapping (ZapSign returns 403 for bad tokens), and webhook registration payloads. `test_zapsign_source.py` guards the source contract: released-with-ALPHA config (the regression that kept sources hidden), schema flags (documents merge-only + webhook-capable), and argument plumbing into the transport.
- Registry guard suites pass (`test_source_categories`, `test_source_versions`, 2593 tests).
- `ruff check` / `ruff format` clean; repo-wide `mypy --cache-fine-grained .` clean for the changed files (one pre-existing unrelated error in `posthog/personhog_client/converters.py` remains); `hogli ci:preflight --fix` reports 0 failures.
- `python manage.py audit_source_docs` passes for ZapSign against the posthog.com checkout with the companion doc in place.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Companion doc PR: [PostHog/posthog.com#18746](https://github.com/PostHog/posthog.com/pull/18746).

## 🤖 Agent context

Built with Claude Code following the `/implementing-warehouse-sources`, `/documenting-warehouse-sources`, and `/writing-tests` skills, with chargebee/revenuecat as reference implementations. Key decisions: preferred the declarative `rest_source` framework over a hand-rolled client (ZapSign is plain DRF-style REST); documents are merge-only incremental because ZapSign only filters by creation date, so status changes on old documents need webhooks or full refresh (called out in the docs); the signers table uses per-document detail fan-out because ZapSign has no list-signers endpoint; `delete_webhook` intentionally reports manual removal because the API can't look up webhook ids after creation. Verified pagination/filter/webhook behavior against ZapSign's docs corpus and live unauthenticated probes; an authenticated smoke test wasn't possible without credentials, which is why the source ships as alpha with fail-loud response-shape guards.

---


MarconLP (approved):

Gilbert09:
Heads-up on the one red check (`Product tests (warehouse-sources ...)` → `Django Tests Pass`): it's a **pre-existing flaky `TestMigrations` connection issue, not a ZapSign regression.**

The failure is always the same shape — `django.db.utils.OperationalError: the connection is closed`, raised in `posthog/test/base.py` inside `MigrationExecutor(connection)` during a migration test's `setUp`, i.e. the DB connection is already dead before the test body runs. pytest-rerunfailures retries it in-process, but the stale connection persists across reruns, so it can't recover.

Evidence it's ordering/infra flakiness rather than something this PR changed:

- It landed on **different, unrelated migration tests across commits** — `test_migration_0052` (Google Ads backfill) on the pre-merge commit, `test_migration_0028` (direct-query backfill) after merging master. Neither is touched by this PR.
- This PR adds only a new source plus **pure unit tests** (HTTP wire mocks — no `TestCase`, no DB, no `connection` usage) and **no migrations**. Adding ~50 tests reshuffles the `optimal_chunks` test split, which is what surfaces this latent connection-isolation flake in whichever migration test happens to follow a connection-poisoning test in a given split.

Re-running the same commit reproduces it (identical split), so I'm re-merging the latest master to reshuffle the split rather than modifying the shared `TestMigrations` base class, which is out of scope for a source PR and would have repo-wide blast radius.

PR #72655, opened 2026-07-21
fix(oauth-proxy): reject duplicate OAuth params in authorize

rafaeelaudibert:
## Problem

The OAuth proxy parses single-valued OAuth params inconsistently when a request repeats them. In `authorize.ts` it reads params like `state` with `searchParams.get()` (first value) to key its KV entries, but when it rebuilds the URL for the regional server it uses `searchParams.set()` in a loop, which keeps the last value. So a request with a repeated `state` gets keyed under one value and forwarded under another, and the callback lookup can miss or resolve to the wrong entry. RFC 6749 §3.1 says these params must not appear more than once, so the right move is to reject the request rather than pick a winner.

## Changes

- `authorize.ts`: reject an authorize request that repeats a single-valued OAuth param (`state`, `client_id`, `redirect_uri`, `response_type`, `scope`, `code_challenge`, `code_challenge_method`) with a 400 `invalid_request`. The check sits at the top of `handleAuthorize`, so it covers both the initial hit and the `_region` resubmit. `resource` (RFC 8707) is intentionally excluded since it is allowed to repeat.
- Added a guard test on the Django `/oauth/authorize/` endpoint confirming it keeps rejecting duplicate params. That endpoint is a custom view (`OAuthLibMixin, APIView`) rather than the toolkit default, so the test locks in that it still routes through oauthlib validation and a refactor can't silently drop it.

- `services/oauth-proxy`: full vitest suite passes (47 tests). Added 7 parameterized cases in `authorize.test.ts` asserting a repeated single-valued param returns 400 and writes nothing to KV. No existing test exercised duplicate params.
- `posthog/api/oauth/test_views.py`: added `test_authorize_rejects_duplicate_param` (5 parameterized cases over the params oauthlib guards) — passes. Catches a regression where the custom authorize view stops rejecting repeated params.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I directed this; Claude (Opus 4.8) wrote the fix and tests. We looked at the proxy's param handling and traced the first-vs-last mismatch between the KV keying (`.get()`) and the forwarded URL (`.set()`). Chose to reject repeated single-valued params up front rather than normalize to first-or-last, since that matches RFC 6749 §3.1 and avoids having to keep both reads in sync forever. Confirmed the Django authorize endpoint already rejects duplicates through oauthlib, and added a test there so the custom view keeps that behavior.

### DIFF
--- a/services/oauth-proxy/src/handlers/authorize.ts
+++ b/services/oauth-proxy/src/handlers/authorize.ts
@@ -12,6 +12,31 @@ const REGION_PICKER_HEADERS: Record<string, string> = {
     'Referrer-Policy': 'no-referrer',
 }
 
+// OAuth request parameters MUST NOT be included more than once (RFC 6749 §3.1).
+// We enforce it because the proxy reads these with `.get()` (first value) for KV
+// keying but forwards the last value to the regional server via `.set()`. A
+// duplicated `state` splits those two reads, letting an attacker route a victim's
+// callback to a preloaded redirect URI. `resource` (RFC 8707) is intentionally
+// excluded — it is allowed to repeat.
+const SINGLE_VALUED_PARAMS = [
+    'state',
+    'client_id',
+    'redirect_uri',
+    'response_type',
+    'scope',
+    'code_challenge',
+    'code_challenge_method',
+] as const
+
+function findDuplicatedParam(url: URL): string | null {
+    for (const param of SINGLE_VALUED_PARAMS) {
+        if (url.searchParams.getAll(param).length > 1) {
+            return param
+        }
+    }
+    return null
+}
+
 /**
  * OAuth Authorization — region picker + redirect.
  *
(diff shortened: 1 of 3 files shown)


services/oauth-proxy/src/handlers/authorize.ts
```diff
@@ -24,6 +49,17 @@ const REGION_PICKER_HEADERS: Record<string, string> = {
 export async function handleAuthorize(request: Request, kv: KVNamespace): Promise<Response> {
     const url = new URL(request.url)
 
+    const duplicatedParam = findDuplicatedParam(url)
+    if (duplicatedParam) {
```
fercgomes, line 53: nit: create a validator for the request, and enforce all rules there so it's decoupled from the controller

fercgomes (approved):

services/oauth-proxy/src/handlers/authorize.ts
rafaeelaudibert: Agreed, will add a validation layer in a follow-up

rafaeelaudibert:
* **#72684** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#72655** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/72655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #72661, opened 2026-07-21
feat(cdp): emit email_bounced_hard metric for permanent SES bounces

meikelmosby:
## Problem

`email_bounced` counts every SES bounce type, but AWS's account-level bounce rate - the number the email reputation feature ([#72545](https://github.com/PostHog/posthog/pull/72545)) calibrates its thresholds against - counts **hard (Permanent) bounces only**. Transient traffic (greylisting, mailbox-full) therefore inflates our measured rate relative to what AWS actually holds against the shared sending account: a sender with 1.5% hard + 1% transient reads 2.5% on our score while AWS sees 1.5%.

## Changes

- The SES webhook now emits `email_bounced_hard` **in addition to** `email_bounced` when `bounce.bounceType === 'Permanent'`. `email_bounced` is unchanged, so every existing consumer (workflow metrics UI, exports) is unaffected - this is purely additive.
- `email_bounced_hard` added to the metric name union.

The reputation evaluator switches its bounce query to this metric (change on #72545). Deploy-order note: this should merge and deploy **ahead of** the evaluator going live, since the evaluator's volume window reaches back up to 30 days and hard-bounce data only accumulates from this deploy forward - windows spanning earlier sends undercount bounces until the metric has depth.

I'm an agent; checks actually run: extended the two existing SES webhook bounce tests to pin the split (Permanent emits both metrics, Transient emits only `email_bounced` - the exact regressions that would silently mis-calibrate reputation), full ses.test.ts suite (44 tests) green, Node tsc clean.

## 🤖 Agent context

Built with Claude Code. Additive-metric shape chosen over changing `email_bounced` semantics after review discussion on #72545 (mayteio's calibration finding): renaming/splitting the existing metric would break every current consumer, while dual emission costs one extra metric row per hard bounce.

---

### DIFF
--- a/nodejs/src/cdp/services/messaging/helpers/ses.test.ts
+++ b/nodejs/src/cdp/services/messaging/helpers/ses.test.ts
@@ -324,7 +324,8 @@ describe('SesWebhookHandler', () => {
         ]
         const result = await handler.handleWebhook({ body, headers: {} })
         expect(result.status).toBe(200)
-        expect(result.metrics?.[0].metricName).toBe('email_bounced')
+        // Hard bounces emit both the catch-all metric and the AWS-comparable hard-only one
+        expect(result.metrics?.map((m) => m.metricName)).toEqual(['email_bounced', 'email_bounced_hard'])
         expect(result.metrics?.[0].distinctId).toBe('user-123')
         expect(result.hardBounceRecipients).toEqual([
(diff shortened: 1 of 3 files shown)


meikelmosby:
Folding this into #72543 (the stack's foundation PR) so the metric ships and starts accumulating with the same deploy that creates the snapshot table — no separate deploy-ordering caveat needed.

PR #72665, opened 2026-07-21
feat(batch-exports): create file download batch exports from hogql queries

rossgray:
## Problem

Batch exports can only export fixed models (events, persons, sessions). We want to support arbitrary HogQL queries. This is the initial phase of that work, scoped to file download (on-demand) exports because they're API-only and have no truncate/append semantics to reason about.

The model and the Temporal workflow code have already been merged. What's missing is the API surface: a way to actually create a file download export from a HogQL query.

## Changes

Adds `model: "hogql"` to the file download create endpoint, taking a `hogql_query`, gated behind the `hogql-batch-exports` feature flag (this will be internal only for now).

- **Validation** (`hogql_source.py`): parse the query, reject placeholders, require every top-level SELECT column to be a field or aliased, then type-resolve it against the team's database. The resolve step uses a context builder extracted from `RecordBatchModel.get_hogql_context`, so API validation and worker execution build the HogQL context identically and can't drift.
    - The column-name check enforces aliases for function/arithmetic expressions, in order to produce readable column names. We will still need validation on a per-destination basis (for example, BigQuery rejects dots/`$`/parens; Snowflake/PG/Redshift quote). Noted for the later phase.
- **Data interval**: hogql exports have no interval, but a lot of downstream code formats concrete bounds (`workflow_id`, S3 staging paths). Rather than thread `Optional` through ~8 files and a migration, the serializer fakes `data_interval_start = data_interval_end = now()`. Maybe in future we should make these bounds optional.

Resource limits (per-query max time/memory) and non-retryable error classification are intentionally out of scope here, deferred to the next PR in the sequence.

No manual testing. Automated tests, added in `test_file_download_api.py` under a new `TestFileDownloadHogQL` class:

- Flag disabled → 403.
- Parameterized invalid-request matrix → 400 (missing/unparseable/placeholder/unaliased-column/unknown-table queries, and intervals or include/exclude supplied alongside hogql).
- Happy path → 202, asserting the `BatchExportSource` row is created with the query, the run gets a now/now interval, and the query is passed into `start_file_download_batch_export`.
- End-to-end: real Temporal worker → download → parse Parquet → assert rows and column names match the inserted events (including another team's rows, to confirm scoping). Like the rest of this module it's AWS-credential-gated.
    - [x] run test locally using AWS playground account

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Feature is behind a flag and PostHog-org-only; no user-facing docs yet.

## 🤖 Agent context

Written by Claude (PostHog Code), directed by Ross. Skills invoked: `/improving-drf-endpoints` (serializer/viewset changes) and `/writing-tests` (test gate). Exploration and the destination column-name investigation were run as subagents.

Key decisions:

- Faking the data interval (now/now) over making it `Optional` end-to-end — the ClickHouse query needs a concrete "as of" timestamp regardless, so the optional path was all cost and no benefit.
- Extracting a shared HogQL context builder rather than letting the API validator and worker construct contexts independently, to remove a real drift risk (a query could validate then resolve differently at run time).
- Keeping resource limits and CH error-classification out of this PR to keep it reviewable; they touch shared entrypoint code used by all destinations.
- The column-name check is a file-download UX guard, not a data warehouse one — investigated how each warehouse destination quotes identifiers and concluded the real validation is destination-specific and belongs in the later phase.

### DIFF
--- a/products/batch_exports/backend/api/file_download.py
+++ b/products/batch_exports/backend/api/file_download.py
@@ -11,22 +11,29 @@ from django.shortcuts import get_object_or_404
 
 import boto3
 import structlog
+import posthoganalytics
 from botocore.client import Config
 from botocore.exceptions import ClientError
 from drf_spectacular.utils import PolymorphicProxySerializer, extend_schema
 from rest_framework import mixins, response, serializers, status, viewsets
-from rest_framework.exceptions import APIException, NotFound, ValidationError
+from rest_framework.exceptions import APIException, NotFound, PermissionDenied, ValidationError
 
 from posthog.api.log_entries import LogEntryMixin
 from posthog.api.routing import TeamAndOrgViewSetMixin
 from posthog.api.utils import action
+from posthog.models import Team
 from posthog.temporal.common.client import sync_connect
 
+from products.batch_exports.backend.hogql_source import (
+    UnsupportedHogQLQueryError,
+    validate_hogql_query_for_batch_export,
+)
 from products.batch_exports.backend.models.batch_export import (
     BatchExportDestination,
     BatchExportFileDownload,
     BatchExportOnDemand,
     BatchExportRun,
+    BatchExportSource,
 )
 from products.batch_exports.backend.service import (
     BatchExportModel,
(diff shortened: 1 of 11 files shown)


products/batch_exports/backend/hogql_source.py
```diff
@@ -37,3 +45,64 @@ def parse_hogql_select_for_batch_export(hogql_query: str) -> ast.SelectQuery | a
...
+    way (team-default modifiers, full database, no top-level LIMIT), otherwise a query
+    that validates could resolve differently when it runs. This builder is the single
+    source of truth for those semantics. It reads from Postgres, so worker code must
+    call it off the event loop.
+    """
+    context = HogQLContext(
+        team=team,
+        team_id=team.id,
+        enable_select_queries=True,
+        limit_top_select=False,
```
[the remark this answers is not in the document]
rossgray: yes, this is coming in a follow up PR (see PR description). Currently functionality is internal-only (behind a feature flag).

products/batch_exports/backend/api/file_download.py
```diff
@@ -121,26 +176,61 @@ def validate(self, data):
...
+        team = self.context["get_team"]()
+        check_hogql_batch_exports_enabled(team)
+
+        # not sure if we need to be this strict but probably best to be explicit
+        # TODO: can remove this once we do support data_interval_start/end
+        if data.get("data_interval_start") is not None or data.get("data_interval_end") is not None:
+            raise ValidationError(
+                "'data_interval_start' and 'data_interval_end' are not supported when 'model' is 'hogql': "
+                "the query runs as of the time the export starts"
+            )
```
tomasfarias, line 189: Fine for now since we are just testing, but yeah this should all go away later.

products/batch_exports/backend/api/file_download.py
```diff
@@ -121,26 +176,61 @@ def validate(self, data):
...
 
-        data_interval_start = validated_data.pop("data_interval_start")
-        data_interval_end = validated_data.pop("data_interval_end")
+        source = None
+        if model == "hogql":
+            source = BatchExportSource(team_id=team_id, hogql_query=validated_data.pop("hogql_query"))
+            # For now, HogQL exports have no data interval: the query runs as of now.
+            # We set a concrete now/now interval to keep everything downstream (eg workflow ID,
+            # staging paths) working unchanged. We should perhaps make the data interval optional in
+            # future.
```
tomasfarias, line 220: Yeah, this is a TODO 👍

products/batch_exports/backend/hogql_source.py
```diff
@@ -37,3 +45,64 @@ def parse_hogql_select_for_batch_export(hogql_query: str) -> ast.SelectQuery | a
...
+    context.database = Database.create_for(team=team, modifiers=context.modifiers)
+    return context
+
+
+def _validate_select_columns_are_named(parsed: ast.SelectQuery | ast.SelectSetQuery) -> None:
+    """Check every top-level SELECT expression has a usable output column name.
+
+    Bare fields and `*` name their columns; anything else (function calls, arithmetic,
+    constants) prints as an expression, producing column names like `plus(1, 1)`, which are likely to
+    cause issues in downstream destinations (e.g. BigQuery does not permit spaces in column names).
```
tomasfarias, line 75: But they are fine for e.g. S3. I think we should not reject expressions without an alias and just raise the error for the user to handle if the destination complains.

tomasfarias: Not a blocker for the time being though, we are just testing.

tomasfarias (approved):
lgtm, I don't think we should reject expressions without an alias, but we can always enable that later, overall I have no blockers.

products/batch_exports/backend/hogql_source.py
rossgray: yeah, think we can remove some of these restrictions over time as we improve the error handling downstream (we probably need to test this for each destination and ensure we return a useful error back to the user rather than some obscure BigQuery SQL compilation error for example)

PR #72671, opened 2026-07-21
fix(surveys): route summarization through the LLM gateway

brandonleung:
Stacked on #72670.

## Problem

Survey response summarization calls Gemini directly using `settings.GEMINI_API_KEY`. That key is invalid in the EU region, so the EU has served no successful summaries since 2026-05-30 and the failure has reached support. It is the loudest casualty of a key that no gateway sits in front of.

After this lands, nothing under `products/surveys/` reads `GEMINI_API_KEY`.

## Changes

- Move summarization onto the gateway helper #72670 introduces, on `claude-haiku-4-5`. `summarization/llm/gemini.py` becomes `gateway.py` and `summarize_with_gemini` becomes `summarize_with_gateway`.
- Replace `GeminiModel` with `SummarizationModel`, pinned to the one model the gateway product allows.
- Register a `survey_summary` gateway product and add its slug to `POSTHOG_AI_PRODUCTS` and `AI_PRODUCT_MARKUP`, so the generations are actually billed rather than only tagged billable.
- Wire the existing `SUMMARIZATION_TIMEOUT` into the call. It was defined and never used, so the call inherited the SDK's 600s default on a request thread.
- Delete `products/surveys/backend/llm/client.py`, which nothing calls once summarization moves.

> [!NOTE]
> `summarize_responses(...)` keeps its arity and argument names, but the `model` parameter changes type and default, and `GeminiModel` is gone. The only in-repo caller passes no `model`, so nothing breaks in tree. The same `ai_product`, `project` group, and `trace_id` changes described in #72670 apply here. The summary rating flow is the join that motivated the `x-posthog-trace-id` header, so ratings attach to their generation once #73017 deploys.

I (Claude) ran the surveys backend suite, plus ruff and mypy. New tests pin two things that were previously revertible with a green suite: the gateway product config (flipping `credit_bucket` to `None` would have made customer-facing survey AI silently free), and the model default against the product's allowlist, which lives in a different deployable. Also added the summarize endpoint's config-guard rejection, for both an empty and a half-set config.

I have not exercised this against a live gateway, so summary quality on Claude versus Gemini is unverified.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed. The public docs for survey AI summaries name no model or provider.

## 🤖 Agent context

Claude (Opus 4.8) wrote this under my direction. #72064 is a competing implementation of the same move that routes through `build_openai_client`, which prefers the Go ai-gateway. I am keeping this one because the Python gateway is already carrying production traffic; #72064 should be closed or rebased rather than treated as a later migration. These two will conflict. Requires human review.


lucasheriques (approved):

products/surveys/backend/summarization/llm/gateway.py
```diff
@@ -75,21 +75,22 @@ def summarize_with_gemini(
     if not responses:
         raise exceptions.ValidationError("responses cannot be empty")
 
-    summary, trace_id = generate_structured_output(
+    summary, trace_id = generate_structured_output_via_gateway(
```
richardsolomou, line 78: can we return the gateway’s actual `$ai_trace_id` here? this helper returns a separate UUID stamped as `llm_trace_id`, but `OpenQuestionSummaryV2` submits it as `$ai_trace_id`, so ratings wont attach to the generation.

richardsolomou (commented):
could we also cover `survey_summary` in the AI-credit aggregation and `computeBillingTotals` test matrices? removing either billing registration would leave the suite green while reported credits diverge.

products/surveys/backend/summarization/llm/gateway.py
brandonleung: i produced the outcome you wanted but your proposed mechanism isn't literally possible because the gateway mints the trace id inside an async callback and never surfaces it

richardsolomou (approved):
could we also cover a key without a gateway URL? rn only the opposite partial config is tested, so a one-sided guard regression could still pass.

brandonleung:
will address this in a later PR on the new gateway

PR #72676, opened 2026-07-21
fix(migrations): resolve custom db_table in staged-drop detection

carlos-marchal-ph:
## Problem

The migration risk analyzer's staged-drop check maps a table name to its model by string-parsing the table name against the app label. Apps whose models keep a legacy custom db_table break it: ai_observability models live in `llm_analytics_*` tables, so the check derives a nonexistent model name, never finds the prior `SeparateDatabaseAndState` state removal, and blocks a correctly staged `DROP COLUMN` in CI.

## Changes

- `check_drop_properly_staged` now resolves the model through the app registry first (match on `model._meta.db_table`), keeping the string heuristic as fallback for models that no longer exist in the registry (staged `DROP TABLE` of deleted models).

Four new tests in `test_risk_analyzer.py`: a real custom-db_table staged drop is recognized (score 2, not blocked), the same drop without prior state removal is still blocked (guard not weakened), and both `None` branches of the new resolver keep the fallback reachable. Full analyzer suite: 177 passed. ruff clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed, internal tooling behavior fix.

## 🤖 Agent context

I directed this, Claude (Claude Code) implemented it TDD-first. Found while preparing the stacked follow-up that physically drops the retired trial-eval columns: its `DROP COLUMN` migration was falsely blocked by this bug. Base of a 2-PR stack, kept standalone so devex can review the analyzer change in isolation.

### DIFF
--- a/posthog/management/migration_analysis/test/test_risk_analyzer.py
+++ b/posthog/management/migration_analysis/test/test_risk_analyzer.py
@@ -8,6 +8,7 @@ from parameterized import parameterized
 from posthog.management.migration_analysis.analyzer import RiskAnalyzer
 from posthog.management.migration_analysis.models import RiskLevel
 from posthog.management.migration_analysis.policies import ConcurrentIndexIdempotencyPolicy, HotTableAlterPolicy
+from posthog.management.migration_analysis.utils import _model_name_for_table
 from posthog.migration_helpers import (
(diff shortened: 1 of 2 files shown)


rnegron (approved):
lgtm, thanks!

PR #72678, opened 2026-07-21
feat(replay-vision): bulk scan recordings on demand

ksvat:
## Problem

The scanner's On-demand tab lists filtered recordings but only lets you scan them one at a time — a "Scan recording" button per row. Running a scanner across a batch of sessions means clicking N times. People want to select several (or the whole page) and scan them together.

## Changes

**Frontend** — the On-demand recordings table gains a checkbox column (LemonTable's built-in `bulkSelection`, same pattern as `FeatureFlags.tsx`): select rows (or select-all-on-page), then a "Scan N selected" button in the selection bar fans the scanner across them. Already-scanned and in-flight rows aren't selectable (nothing to re-scan). A toast reports the outcome, including partial runs.

**Backend** — new `POST /vision/scanners/{id}/bulk_observe/` taking `session_ids` (max 200 per request). It **scans what fits**: it starts as many as the in-flight caps (per-scanner 150 and per-team 300) and the org's monthly credit quota allow, then reports the rest per session rather than failing the whole batch. Per-session outcomes: `started`, `already_running`, `skipped_limit`, `skipped_quota`, `failed`. It reuses the same deterministic workflow-start path as single `observe`, so it stays idempotent — re-submitting an in-flight session is a no-op, never a double charge.

> [!NOTE]
> Scope is checkbox multi-select on the loaded page (+ select-all-on-page). "Select all N matching across every page" isn't included — it needs a separate query to enumerate all matching session ids, and the concurrency caps mean an unbounded "scan everything" would mostly get skipped anyway. Easy follow-up if wanted.

The per-session enum field is named `scan_outcome` (not `outcome`) on purpose — a bare `outcome` ChoiceField collides with other products' `outcome` enums in the shared OpenAPI spec and would churn their generated types.

- Backend `TestBulkObserveAction` (6 cases): starts one scan per session; dedups repeated ids; scans-what-fits under the in-flight cap and marks the rest `skipped_limit`; reports `skipped_quota` when credits are the tighter limit; treats an already-running session as a no-op that consumes no headroom; rejects empty and >200 batches.
- Frontend `scannerRunTabLogic.test.ts`: bulk scan posts the selected session ids and clears its loading state.
- Full backend `test_api.py` bulk class green; `hogli build:openapi` regenerated types (the `scan_outcome` naming keeps other products' generated files unchanged); ruff + ty + oxlint + scoped `typescript:check` clean.

I did not manually click through the running app.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Kim (@ksvat) is the DRI.

Built with Claude Code (Opus). Skills invoked: `/improving-drf-endpoints`, `/writing-kea-logics`, `/writing-tests`.

Design decisions: reused the existing `start_apply_scanner_workflow` per session (keeping idempotency + quota/in-flight semantics identical to single observe) rather than a new batch workflow; computed the start budget once up front from the tighter of the in-flight caps and the remaining quota, decrementing locally as each scan starts (models each new in-flight row without re-querying). The `observe` path today only enforces the team-wide cap; bulk_observe additionally respects the per-scanner cap so one scanner's batch can't monopolize the team budget.

---


arnohillen:
Nice, two things I'd fix before merging.

- One Temporal connection per session. `start_apply_scanner_workflow` calls `sync_connect()` internally, and that creates a fresh client on every call (no caching, `posthog/temporal/common/client.py`). The bulk loop runs it up to 200 times sequentially inside one synchronous request, so a full batch pays up to 200 TLS handshakes on a single gunicorn worker and could take tens of seconds. Suggest connecting once per batch: either an optional `client` param on `start_apply_scanner_workflow` or a small batch helper in `trigger.py`. Keeps the per-session idempotency identical.

- Zero-cost quota guard. In `_bulk_observe_headroom`, `snapshot.remaining // cost if cost else 0` gives a zero-cost model a headroom of 0, so every session would come back `skipped_quota`. Free should mean quota never binds, so the fallback should be `in_flight_limit` (like the `remaining is None` branch). Unreachable today since all models cost >= 2, but it's a tiny fix while you're in there.

arnohillen (approved):

PR #72679, opened 2026-07-21
feat(aio): ai blob read path with authenticated immutable serving

carlos-marchal-ph:
## Problem

The ingestion side (#72386) offloads large multimodal payloads from `ai_events` to S3 and leaves `phaiblob://` pointers in their place.
Without a read path, the trace view renders those pointers as broken images and dead links.

## How to test

This is how you seed an event with multimodal data in the local dev env, have your robot of choice adapt this snippet to create test data for you:
```
# 1. plant any real PNG at its own content address
H=$(shasum -a 256 pic.png | cut -d' ' -f1); S=$(wc -c < pic.png)
AWS_ACCESS_KEY_ID=object_storage_root_user AWS_SECRET_ACCESS_KEY=object_storage_root_password \
aws --endpoint-url http://localhost:19000 s3 cp pic.png "s3://ai-blobs/llma/1/sha256/$H" --content-type image/png

# 2. send the pointer as data through plain capture (no flags, no lanes)
posthog.capture("$ai_generation", distinct_id="test", properties={
    "$ai_trace_id": str(uuid.uuid4()), "$ai_model": "gpt-9",
    "$ai_input": [{"role": "user", "content": [{"type": "image_url",
        "image_url": {"url": f"phaiblob://v1/sha256/{H}?mime=image%2Fpng&size={S}"}}]}],
})
```

## Changes

- New session-authenticated endpoint `GET /api/projects/:team_id/ai_blob/v1/sha256/:hash` that serves blobs from the AI blob bucket. The team id is resolved server-side from the authenticated route, and the hash is route-regex constrained to 64 hex chars.
- Content-addressed assets never change, so responses are `Cache-Control: private, max-age=31536000, immutable`. Raster image types are served inline, everything else goes out as `application/octet-stream` with a bare `attachment` disposition (a filename here would override the frontend's `download` attribute) plus `nosniff`.
- New `object_storage.read_object` helper returning bytes plus stored content type. `read_bytes` now delegates to it.
- Frontend `aiBlob.ts` parses pointers and resolves them to endpoint URLs, wired into `ConversationMessagesDisplay` for images, audio, and file downloads. Non-pointer values pass through unchanged.
- `AI_BLOB_S3_BUCKET` / `AI_BLOB_S3_PREFIX` settings. The nodejs writer reads the same env vars but defaults to disabled, so the settings comment documents that both services must agree on bucket and prefix.

Two deliberate v1 decisions:

> [!NOTE]
> The endpoint is scoped to `llm_analytics` so resource-level access controls apply to blob reads, but it stays session-only: personal API key, OAuth, and MCP consumers get pointers they cannot dereference yet. The public API story (likely presigned URLs) is follow-up work.

> [!NOTE]
> `private, immutable` caching without `Vary: Cookie` means a shared browser profile could serve a cached blob across a re-login. Accepted tradeoff, in line with `uploaded_media` being fully public.

- Backend tests cover the auth boundary (cross-team and unauthenticated denials), inline vs attachment branching, 404s for missing blobs and malformed hashes, and the bucket/key construction. These didn't exist before because the endpoint didn't.
- Jest tests cover pointer parsing (anchored regex, bad version/algo/hash fail closed to pass-through) and URL resolution.
- Tested e2e in local dev with the ingestion offload branches running. Images and audio render in the trace view, downloads keep their filenames.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal endpoint, no user-facing docs affected.

## 🤖 Agent context

- Claude Code assisted the review round. Two independent agents ran a whitebox review of the branch, findings were triaged together, and the surviving ones were folded in. The `pr-review` and `pr-link` skills were invoked along the way.
- Applied from review. `read_bytes`/`read_object` dedup and matching the sibling helper signature shape, bare attachment disposition, and the settings comment documenting the cross-service bucket/prefix contract.
- Rejected from review. Claims that non-allowlisted mime types break audio and image rendering were disproven by e2e testing. Browsers ignore `Content-Disposition` on subresource fetches and `nosniff` only gates script and style destinations, so octet-stream media renders fine. A proposed `OBJECT_STORAGE_BUCKET` fallback for the blob bucket was rejected because the writer defaults to disabled, making the fallback unreachable in practice, and a dedicated bucket keeps the expiry lifecycle isolated.

### DIFF
--- a/posthog/settings/object_storage.py
+++ b/posthog/settings/object_storage.py
@@ -57,6 +57,14 @@ BILLING_USAGE_REPORTS_S3_BUCKET = os.getenv("BILLING_USAGE_REPORTS_S3_BUCKET") o
 # back to the general bucket in dev / self-hosted.
 AGENT_BUNDLES_S3_BUCKET = os.getenv("AGENT_BUNDLES_S3_BUCKET") or OBJECT_STORAGE_BUCKET
 
+# AI observability blob storage — offloaded binary payloads (images, files) from ai_events,
+# content-addressed under `{prefix}{team_id}/sha256/{hash}`. The nodejs ingestion writer reads
+# the same AI_BLOB_S3_BUCKET / AI_BLOB_S3_PREFIX env vars but defaults both to "" (offload
+# disabled) — any deployment enabling offload must set bucket AND prefix identically on both
+# services, or every read silently 404s. Defaults here match the local-dev ingestion wiring.
+AI_BLOB_S3_BUCKET = os.getenv("AI_BLOB_S3_BUCKET", "ai-blobs")
+AI_BLOB_S3_PREFIX = os.getenv("AI_BLOB_S3_PREFIX", "aio/")
+
 # Identity matching scratch storage (products/growth `identity_matching_job`). The job writes
 # per-run Parquet objects via ClickHouse `INSERT INTO FUNCTION s3(...)` and the read API globs
 # them back with `s3(...)`, so only the ClickHouse cluster needs bucket access — the Dagster
--- a/products/ai_observability/backend/api/ai_blob.py
+++ b/products/ai_observability/backend/api/ai_blob.py
@@ -1,0 +1,36 @@
+from django.conf import settings
+from django.http import HttpResponse
+
+from drf_spectacular.utils import extend_schema
+from rest_framework import viewsets
+from rest_framework.decorators import action
+from rest_framework.request import Request
+
+from posthog.api.routing import TeamAndOrgViewSetMixin
+from posthog.storage import object_storage
+
+INLINE_MIMES = {"image/png", "image/jpeg", "image/gif", "image/webp"}
+
+
+class AIBlobViewSet(TeamAndOrgViewSetMixin, viewsets.ViewSet):
+    scope_object = "INTERNAL"
+
+    @extend_schema(exclude=True)
+    @action(detail=False, methods=["GET"], url_path=r"v1/sha256/(?P<hash>[0-9a-f]{64})")
+    def serve(self, request: Request, hash: str, **kwargs) -> HttpResponse:
+        key = f"{settings.AI_BLOB_S3_PREFIX}{self.team_id}/sha256/{hash}"
+        result = object_storage.read_object(key, bucket=settings.AI_BLOB_S3_BUCKET, missing_ok=True)
+        if result is None:
+            return HttpResponse(status=404)
+        body, content_type = result
+        if content_type in INLINE_MIMES:
+            response = HttpResponse(body, content_type=content_type)
+        else:
+            response = HttpResponse(body, content_type="application/octet-stream")
+            # Bare attachment: a filename here would override the frontend's `download` attribute.
+            response["Content-Disposition"] = "attachment"
+        # Content-addressed: the hash IS the content, so the asset is immutable forever.
+        response["Cache-Control"] = "private, max-age=31536000, immutable"
+        response["ETag"] = f'"{hash}"'
+        response["X-Content-Type-Options"] = "nosniff"
+        return response
--- a/products/ai_observability/backend/api/ai_blob.py
+++ b/products/ai_observability/backend/api/ai_blob.py
@@ -13,7 +13,10 @@ INLINE_MIMES = {"image/png", "image/jpeg", "image/gif", "image/webp"}
 
 
 class AIBlobViewSet(TeamAndOrgViewSetMixin, viewsets.ViewSet):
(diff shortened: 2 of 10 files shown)


posthog/settings/object_storage.py
```diff
@@ -57,6 +57,14 @@
 # back to the general bucket in dev / self-hosted.
 AGENT_BUNDLES_S3_BUCKET = os.getenv("AGENT_BUNDLES_S3_BUCKET") or OBJECT_STORAGE_BUCKET
 
+# AI observability blob storage — offloaded binary payloads (images, files) from ai_events,
+# content-addressed under `{prefix}{team_id}/sha256/{hash}`. The nodejs ingestion writer reads
+# the same AI_BLOB_S3_BUCKET / AI_BLOB_S3_PREFIX env vars but defaults both to "" (offload
+# disabled) — any deployment enabling offload must set bucket AND prefix identically on both
+# services, or every read silently 404s. Defaults here match the local-dev ingestion wiring.
+AI_BLOB_S3_BUCKET = os.getenv("AI_BLOB_S3_BUCKET", "ai-blobs")
+AI_BLOB_S3_PREFIX = os.getenv("AI_BLOB_S3_PREFIX", "aio/")
```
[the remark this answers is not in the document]
carlos-marchal-ph: deployment wiring ships with the offload rollout (charts + secrets PRs live there), and the nodejs writer defaults to disabled so no pointers can exist in prod before that wiring lands, will ship them in charts later

products/ai_observability/backend/api/ai_blob.py
```diff
@@ -0,0 +1,39 @@
...
+            return HttpResponse(status=404)
+        body, content_type = result
+        if content_type in INLINE_MIMES:
+            response = HttpResponse(body, content_type=content_type)
+        else:
+            response = HttpResponse(body, content_type="application/octet-stream")
+            # Bare attachment: a filename here would override the frontend's `download` attribute.
+            response["Content-Disposition"] = "attachment"
+        # Content-addressed: the hash IS the content, so the asset is immutable forever.
+        response["Cache-Control"] = "private, max-age=31536000, immutable"
```
Radu-Raicea: `private` here is to the browser, but a user might log out and this will still be cached.

products/ai_observability/frontend/aiBlob.ts
```diff
@@ -0,0 +1,47 @@
...
+    return {
+        version: match[1],
+        algo: match[2],
+        hash: match[3],
+        mime: params.get('mime'),
+        size: sizeRaw && /^\d+$/.test(sizeRaw) ? parseInt(sizeRaw, 10) : null,
+    }
+}
+
+export function resolveAiBlobUrl(value: string, teamId: number | string | null): string {
```
Radu-Raicea, line 32: This misses pointers after the Anthropic normalizer wraps `source.data` as `data:<mime>;base64,<value>`. Once ingestion replaces the data with `phaiblob://...`, bare Anthropic image blocks produce an invalid data URI and render broken.

products/ai_observability/backend/routes.py
```diff
@@ -32,6 +33,12 @@
 
 
 def register_routes(routers: RouterRegistry) -> None:
+    # The frontend calls the projects path; the /api/environments/ registration stays for the
+    # environments->projects redirect contract (see test_environments_rewrite.py /
+    # test_team_project_parity.py) — mirrors the error_tracking/releases and
+    # error_tracking/symbol_sets registration shape.
+    routers.environments.register(r"ai_blob", AIBlobViewSet, "environment_ai_blob", ["team_id"])
```
Radu-Raicea: Don't need this no? Since this is a new endpoint.

Radu-Raicea (approved):
Approving, but need to tackle the comments first.

products/ai_observability/frontend/aiBlob.ts
carlos-marchal-ph: The offload step handles data URLs correctly and replaces them with `phaiblob` URLs.

products/ai_observability/backend/routes.py
carlos-marchal-ph: Yep, removed

products/ai_observability/backend/api/ai_blob.py
carlos-marchal-ph: Reduced the TTL to a day. There's no way to have both 1. caching and 2. up-to-date auth validation on the same request. Same tradeoff we have with the uploaded media model right now (that one is actually cached indefinitely).

PR #72682, opened 2026-07-21
fix(support): truncate long last-message cells in tickets table

HaynesPostHog:
## Problem

In the support/conversations tickets list, the "Last message" column stretched to fit the longest message in view, which pushed the whole table into a long horizontal scroll. Raised in the [#team-support Slack thread](https://posthog.slack.com/archives/C08S2L0S5MZ/p1784652970024079?thread_ts=1784652970.024079&cid=C08S2L0S5MZ).

## Changes

The column already carried `truncate max-w-md`, but on an inline `<span>` (the child passed into `TicketPreviewPopover`). `max-width` and `overflow: hidden` have no effect on inline elements, so the truncation silently did nothing: the text laid out on one full-width line and the cell grew to fit it.

Setting the span to `inline-block` gives both properties a formatting context to act on, so the message clips with an ellipsis at 28rem and the column stays bounded. One-class change.

I (the PostHog Slack app agent) have not run the app or added an automated test for this. The fix is a CSS-class change reasoned from the DOM structure: an inline element can't honor `max-width`/`overflow`, and `inline-block` restores both. Worth a quick visual check with a very long last message before merge.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Ben Haynes flagged the horizontal-scroll issue from Slack and asked for a fix. I traced it to `products/conversations/frontend/scenes/tickets/ticketColumns.tsx`: the last-message text uses `truncate max-w-md`, but sits inside the bare inline `<span>` that `TicketPreviewPopover` wraps its children in, so those utilities were inert. Considered adding a column `width`, but that maps to a `<col>` preferred width which auto-layout tables ignore for nowrap content, so it wouldn't reliably clip. Chose `inline-block` on the text span as the minimal self-contained fix. No repo skills were needed for a change this size.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08S2L0S5MZ/p1784652970024079?thread_ts=1784652970.024079&cid=C08S2L0S5MZ)*


luke-belton (approved):
verified locally 👍

PR #72684, opened 2026-07-21
refactor(oauth-proxy): introduce reusable request validators

rafaeelaudibert:
## Problem

Fernando left a review nit on #72655: the OAuth authorize handler enforced its rules inline, mixed with control flow. The duplicate-param check and the redirect_uri registration check were hand-built 400s smeared across `handleAuthorize` and `redirectToRegionalAuthorize`, and every handler that rejects a bad request rolled its own `error/error_description` response. There was no shared notion of "validate the request", so adding a new rule meant more bespoke branching inside a controller.

Stacked on top of #72655.

## Changes

Introduce a small validator primitive in `services/oauth-proxy/src/lib/validation.ts`:

- `ValidationError` — the `{ error, error_description }` shape.
- `Validator = (request, url) => ValidationError | null` — a pure, stateless rule.
- `errorResponse(error, extraHeaders?)` — renders a `ValidationError` as an OAuth-style 400.
- `runValidators(...)` — runs a list in order, returning the first failure.
- `noDuplicateParams(...names)` — a factory; building a new rule is one small function.

Routes declare `validators` and the router runs them before the handler, so request-shape rules live in one place. The `/oauth/authorize` route now declares `noDuplicateParams(...)` instead of the check living inside the handler. The stateful redirect_uri check (needs a KV mapping lookup) stays in the handler but is re-expressed as a local validator reusing the same `ValidationError` + `errorResponse`, showing the concept works at both the router and handler tiers. The malformed-JSON 400s in `token` and `passthrough` now go through `errorResponse` too.

Behavior is unchanged: same status codes, same error bodies, same headers.

> [!NOTE]
> This is a pure refactor on a security-sensitive OAuth path. No request is accepted or rejected differently than before.

### Why not move everything into the validator layer

The router layer only fits rules that are **stateless** (decidable from `request` + `url` alone) and **pure rejection** (the handler doesn't need the checked value afterward). I audited every 400/error path in the proxy against that bar:

| Check | Location | Moved to router layer? |
|---|---|---|
| Duplicate params | authorize | ✅ yes |
| `redirect_uri` registered | authorize | ❌ stateful (KV mapping) — stays in handler, reuses `errorResponse` |
| Malformed JSON body | token, passthrough | ❌ needs the parsed body the handler also uses — stays in handler, reuses `errorResponse` |
| "Unable to determine region" | token, proxy | ❌ stateful (KV / network outcome), not request validation |
| Metadata upstream failure (502) | metadata | ❌ network failure, not a 400 |
| Missing `state` | callback | ❌ borderline, left in handler (see below) |
| Expired/unknown `state` | callback | ❌ stateful (KV) — stays in handler |

The duplicate-param check was the only pure-rejection rule, so it's the only one hoisted. Everything else is stateful and correctly stays where the state lives; the JSON ones already share `errorResponse` for consistent rendering.

callback's missing-`state` check is the one borderline case. It looks hoistable (it's stateless), but it's a poorer fit than duplicate-params for two reasons, so I left it in the handler:

1. **The handler consumes the value.** Right after the check, `getCallbackRedirectUri(kv, state)` needs `state: string`. These validators are reject-only — they don't thread a narrowed value back to the handler — so hoisting would force an unsafe cast or a duplicated guard. Duplicate-params had no such downstream dependency, which is why it hoisted cleanly.
2. **Different error contract.** callback returns plain-text 400s while the validator layer emits OAuth JSON; moving only the presence check would leave callback's two adjacent 400s in two different formats.

`pnpm test run` (47 pass), `pnpm typecheck`, `oxfmt --check`, and `oxlint` all green in `services/oauth-proxy`. I (Claude) did not do manual browser testing.

The duplicate-param `it.each` moved from `authorize.test.ts` to `router.test.ts` because the check now runs at the router, not in `handleAuthorize`. It exercises the real `fetch` → validator → 400 path and asserts rejection happens before any KV write — the same parameter-pollution regression the original test caught, at the layer where the behavior now lives. No new test surface; it's a relocation.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Rafael asked whether Fernando's "create a validator, decouple from the controller" nit could be generalized into a reusable validator concept used across the oauth-proxy handlers, and whether a framework would help. I (Claude) checked the service first: it's a dependency-free Cloudflare Worker with a hand-rolled router and no validation lib. I laid out options (adopt Hono + zod, add zod only, a homegrown primitive, or a local extraction) and we chose the homegrown primitive to keep the worker's zero-runtime-dep supply chain intact on a security-sensitive path. A key design decision: only stateless, pure-rejection rules hoist to the router; stateful checks that need KV stay in the handler but reuse the same `ValidationError`/`errorResponse`, so the concept unifies error rendering without forcing a spurious KV read on the region-picker path.

Rafael then asked whether any other validation could move to the router layer. I audited every handler (table above) and we agreed the pattern is already applied everywhere it genuinely fits — callback's `state` check is the one borderline candidate, deliberately left in the handler because it feeds a value the handler consumes and callback uses a plain-text error contract. Nothing else moved.

Skills invoked: `/writing-tests` (gated the test relocation — confirmed it's a move, not new bloat).

### DIFF
--- a/services/oauth-proxy/src/handlers/authorize.ts
+++ b/services/oauth-proxy/src/handlers/authorize.ts
@@ -1,5 +1,6 @@
 import { type Region, baseUrlForRegion } from '@/lib/constants'
 import { type ClientMapping, getClientMapping, putCallbackRedirectUri, putRegionSelection } from '@/lib/kv'
+import { type ValidationError, errorResponse } from '@/lib/validation'
 
 import REGION_PICKER_HTML from '../static/region-picker.html'
 
@@ -12,26 +13,17 @@ const REGION_PICKER_HEADERS: Record<string, string> = {
     'Referrer-Policy': 'no-referrer',
 }
 
-// OAuth request parameters MUST NOT be included more than once (RFC 6749 §3.1).
-// We enforce it because the proxy reads these with `.get()` (first value) for KV
-// keying but forwards the last value to the regional server via `.set()`. A
-// duplicated `state` splits those two reads, letting an attacker route a victim's
-// callback to a preloaded redirect URI. `resource` (RFC 8707) is intentionally
-// excluded — it is allowed to repeat.
-const SINGLE_VALUED_PARAMS = [
-    'state',
-    'client_id',
-    'redirect_uri',
-    'response_type',
-    'scope',
-    'code_challenge',
-    'code_challenge_method',
-] as const
-
-function findDuplicatedParam(url: URL): string | null {
-    for (const param of SINGLE_VALUED_PARAMS) {
-        if (url.searchParams.getAll(param).length > 1) {
-            return param
+// Prevent open redirects: for clients registered through the proxy (which have
+// stored redirect_uris), the requested redirect_uri must be one of them. Legacy
+// clients without stored redirect_uris fall through to regional server validation.
+function validateRegisteredRedirectUri(
+    redirectUri: string | null,
+    mapping: ClientMapping | null
+): ValidationError | null {
+    if (mapping?.redirect_uris && redirectUri && !mapping.redirect_uris.includes(redirectUri)) {
+        return {
+            error: 'invalid_request',
+            error_description: 'redirect_uri is not registered for this client',
         }
     }
     return null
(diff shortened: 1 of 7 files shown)


rafaeelaudibert:
* **#72684** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72684?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/72684?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#72655** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

fercgomes (approved):
why not use a validation middleware + something like zod?

rafaeelaudibert:
### Merge activity

* **Jul 22, 8:32 AM UTC**: @rafaeelaudibert merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/72684).

rafaeelaudibert:
@fercgomes because I didnt wanna add something new here, rolling our own made sense. This is a small service

PR #72685, opened 2026-07-21
feat(ai): surface event description on read_taxonomy properties path

sufi-ph:
## Problem

Max can read a custom event's description now (via PR #72088), but only on the events-*listing* taxonomy path. For a single event, `read_taxonomy` with `kind: event_properties` returned just the property list and never the event's own description.

That matters because when a user asks "what does event X mean?", Max reflexively calls `event_properties` for that one event. It sees no description in the response and concludes none exists, even when the team has written one in Data management. So the descriptions people maintain still don't reliably reach the agent.

**Why:** raised from a Slack thread where Max repeatedly told a user it couldn't read an event's description, and even claimed the Data management field was empty when it wasn't. Root cause was the wrong-endpoint gap above, not the merged fix.

## Changes

- Add `get_event_description(team, event_name)` in `ee/hogai/utils/helpers.py`. It returns a single event's description using the same precedence as the list path: built-in core taxonomy first, then the user-authored `EnterpriseEventDefinition.description` (EE-gated, no-op on OSS).
- Prepend that description to the `event_properties` result in `ee/hogai/tools/read_taxonomy/core.py`, so the single-event lookup Max actually uses returns it.
- Refactor the core-taxonomy description formatting shared by both paths into `_format_core_event_description` (no behavior change on the list path).

> [!NOTE]
> No new query on the hot path when there's nothing to fetch: core events short-circuit before touching the DB, and the EE lookup reuses the existing batched helper for the one event.

Added a parameterized test to `ee/hogai/tools/read_taxonomy/test/test_tool.py` covering the `event_properties` path: a custom event with a stored description now surfaces it (`Description of \`quiz_retaken\`: ...`), and one without a description emits no prefix. This guards the exact regression this PR fixes — no existing test exercised descriptions on the single-event path.

I (via the PostHog Slack app) could not run the suite in this environment: no Postgres is reachable and the test creates an `EnterpriseEventDefinition` row, so it errors at DB setup. I ran `ruff check` / `ruff format` (clean) and `py_compile` (passes) on all three files.

## 🤖 Agent context

Investigated from a Slack thread using the PostHog AI-observability trace tools. Confirmed the events-list path already returns the description (proving the earlier fix is deployed and working) and that Max's failures all came from it querying `event_properties` instead. Chose to surface the description on that path rather than change the model's routing, and reused the existing EE description helper to avoid an N+1. Invoked the `/writing-tests` skill before adding the test.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BALLBBKAB/p1784649913361349)*

### DIFF
--- a/ee/hogai/tools/read_taxonomy/core.py
+++ b/ee/hogai/tools/read_taxonomy/core.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel, Field
 from posthog.models import Team, User
 
 from ee.hogai.chat_agent.query_planner.toolkit import TaxonomyAgentToolkit
-from ee.hogai.utils.helpers import format_events_yaml
+from ee.hogai.utils.helpers import format_events_yaml, get_event_description
 
 
 class ReadEvents(BaseModel):
@@ -17,7 +17,7 @@ class ReadEvents(BaseModel):
 
 
 class ReadEventProperties(BaseModel):
-    """Returns the properties list for a provided event. Before calling this tool, ensure the event exists by reading events."""
+    """Returns the description (if one is set) and properties list for a provided event. Before calling this tool, ensure the event exists by reading events."""
 
     kind: Literal["event_properties"] = "event_properties"
     event_name: str = Field(description="The name of the event that you want to retrieve properties for.")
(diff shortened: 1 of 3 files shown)


skoob13 (approved):
Thank you!

PR #72691, opened 2026-07-21
refactor(capture): outputs model — pipeline/lane routing, output policies, serialization seam

pl:
## Problem

`rust/capture`'s produce path mixes routing policy, serialization, and transport mechanism in one place: the Kafka sink decides topics from event metadata, serializes inline, and multi-target behaviors (S3 fallback, AI secondary-cluster split) are implemented as sink composites wrapping other sinks. That coupling makes the three things we actually want hard and risky:

1. safe **automatic failover** (circuit breaker over Kafka→S3),
2. **cluster migrations without code changes** (split/dual-write as configuration, like the Node.js `IngestionOutputs` model),
3. **payload format evolution** (a future protobuf cutover) without touching sinks or call sites.

This PR is the final state of the capture outputs refactor, built as a sequence of individually shippable commits that each graduate to their own PR. The plan doc (`rust/capture/OUTPUTS_REFACTOR_PLAN.md`, first commit) is the working contract: architecture, per-step goals, parity proofs, and the progress tracker.

## Changes

The produce path becomes five strata with one-way dependencies:

```
edge              → Pipeline {Analytics, Ai, Heatmaps, Warnings, ErrorTracking, SessionReplay}
                    (v0 and v1 ingress alike — one produce path)
pipeline steps    → stamp intent (restrictions, overflow, historical); v1 maps its
                    decided Destination onto the same interchange (to_processed)
lane resolution   → per-pipeline typed lanes (invalid pairs unrepresentable) + key policy
                    → Address = (pipeline, lane) | custom admin redirect
outputs           → trait Outputs: open set of produce surfaces owning namespace
                    realization + policy — KafkaOutputs (address→topic map + producer),
                    S3Outputs, FailoverOutputs, SplitOutputs, per-mode tables,
                    DynamicKafkaOutputs (coordinator-managed prototype)
serialization     → per-surface format × envelope with content headers
sinks             → pure transport: realized records in, per-event results out
```

Commit by commit (this is the branch *history* — names in early bullets are superseded by later ones: `OutputRegistry`/`OutputTable` → per-mode tables + sink-side `TopicTable`, `as_output()` → `as_address()`, `Replay` → `SessionReplay`; the reviewer guide below reflects only the final state):

- **Routing golden oracle** pins topic/key/header/counter behavior for every pipeline and lane; every later commit proves parity against it unmodified.
- **Pure `route()`** extracted from the Kafka sink's `prepare_record`.
- **`OutputRegistry`** — one output→topic wiring point plus a refuse-to-boot completeness check.
- **Serialization layer** — `Format` (json) × `Envelope` (none | lz4) behind one seam; the replay lz4 envelope moves verbatim with its `content-encoding` coexistence design generalized.
- **`Pipeline` + `Lane`** — the lane decision (`pipeline::resolve`, precedence dlq > custom > historical > overflow > main) becomes pure pipeline-layer code.
- **Kafka sink narrowed to mechanism** — prepared payloads in, per-event results out; prep extracted for the outputs layer to drive.
- **Outputs layer** — `Output` policy tree owns failover (health-gated Kafka→S3) and split (token-routed AI secondary) from day one; `FallbackSink` and `SplitKafkaSink` are deleted with their tests re-expressed at the output layer.
- **Call sites migrate** — analytics, AI, OTEL, and replay publish through the `OutputTable`; then the v0 `Event` trait is retired entirely.
- **Mode-scoped completeness** — an Events/Ai pod demands the analytics topic family, a Recordings pod main/replay-overflow/dlq only.
- **Breaker failover mode (dark)** — the circuit breaker (error-ratio window, open/half-open/closed, single-probe permit, control-plane seam) lands behind `CAPTURE_FAILOVER_ENABLED` (default off; off is byte-identical to the reactive mode).
- **v1 convergence** — v1's `Destination` bridges to the shared registry via `as_output()`; the crate has exactly one topic-resolution point. The v1 sink-trait convergence stays gated on the per-event response model.
- **Typed addresses** — per-pipeline lane enums make invalid (pipeline, lane) pairs unrepresentable; admin custom redirects and the `$ai_*` stream become first-class addresses (`Pipeline::Ai`, classified by event name consistently with the LLM quota predicate).
- **Sinks realize namespaces** — payloads carry the abstract `Address`; the Kafka sink translates it against its per-cluster topic table at publish time (the swappable seam a repartitioning coordinator plugs into), S3 realizes it trivially. A failover pair can share one prepared batch because payloads are target-agnostic.
- **Per-mode output tables** — `AnalyticsFamilyOutputs` / `SessionReplayOutputs` with required fields: the narrow list of what each deployment must wire is the type itself. Handlers are bound by sealed publish-capability traits over a generic `State<T>`, and setup instantiates the monomorphized router per `CaptureMode` — mounting an ingress on a table that can't publish its family is a compile error.
- **Prep hoist** — payload assembly (lane resolution, serialization, headers, topic/key, scatter-gather batch prep) moves fully into the outputs layer via `PrepSpec`; the transitional `Prepare` trait is deleted and sinks take prepared payloads as their only input. Capturing test mocks now assert wire-level outcomes (topic, key, headers, payload bytes) instead of metadata stamps — which surfaced and pinned one previously untested semantic: replay events redirected to dlq/custom topics partition on the event key, not the session id.
- **AI ingress is its own router family** — `ai_router` mounts only `/i/v0/ai*` (matching the contour routing, where those routes go exclusively to capture-ai deployments), bound by its own `PublishesAi` capability; the analytics-family router no longer serves AI endpoints.
- **Topic tables are sink-side data** — the topic table moves from the outputs layer into the sinks (`sinks::topics::TopicTable`), supplied at sink construction; sinks never import the outputs layer, and the abstract `Address` is the only interchange vocabulary between the strata.
- **Per-pipeline output overrides + boot verification** — `CAPTURE_OUTPUT_<PIPELINE>_BROKERS` / `CAPTURE_OUTPUT_<PIPELINE>_TOPIC_<LANE>` retarget one pipeline's output row at another cluster/topic namespace (existing `KAFKA_*` keys as defaults; producers shared per distinct broker set, so the default config still opens one connection; refused at boot in combination with S3 fallback or AI secondary). `CAPTURE_VERIFY_TOPICS_ON_BOOT` (default off) probes each row's cluster for its reachable topics and refuses to start on a missing one.
- **Naming and import hygiene** — remaining long nested paths replaced with imports; the session replay pipeline renamed to match the Node.js side's `sessionreplay` vocabulary (`Pipeline::SessionReplay`, `SessionReplayLane`, `SessionReplayOutputs`, `PublishesSessionReplay`, `CAPTURE_OUTPUT_SESSION_REPLAY_*`), leaving deployed names untouched (`KAFKA_REPLAY_*` keys, `CaptureMode::Recordings`, metric names, restriction wire strings).
- **Outputs becomes an open trait; sinks become pure transport** — the closed policy enum dissolves into `trait Outputs` with implementations owning namespace realization + backend composition (`KafkaOutputs` = address→topic map + producer; `S3Outputs`; `FailoverOutputs`/`SplitOutputs` over `Arc<dyn Outputs>`; the per-mode tables themselves). `publish` reports per-event `SinkResult`s (`publish_folded` for whole-request callers). Sinks never see an `Address`.
- **Dynamic outputs prototype (test-only)** — `DynamicKafkaOutputs` subscribes to an in-process, RPC-shaped `KafkaManagerService` (acked config pushes) and applies broker add/remove + mapping changes with an enabled-partition set per address; two tests simulate incremental topic and broker switchovers end to end through `publish`.
- **v1 converges on the outputs machinery** — `WrappedEvent::to_processed` maps fully-processed v1 events onto the shared interchange (14-case byte-parity suite against the frozen legacy serializer; one accepted parse-equal delta in `sent_at` fractional formatting); `CAPTURE_V1_SINKS` names become shared `KafkaOutputs` surfaces (per-sink env config mapped field-for-field onto the shared `KafkaConfig`); per-event results merge into `BatchResponse`; the legacy v1 sink stack (`Event`/`Sink`/`Router` traits, `serialize_batch`, v1 Kafka producer/mock) is deleted, with `v1_sink_integration` running its real-Kafka round-trips through the converged path.
- **Repartitioning coordinator design note** — the plan doc records how a partition-by-partition cluster switchover plugs into the landed seams (coordinator-owned logical shards stamped at prep, a `ShardRouted` output policy, explicit partition pass-through, per-partition drain-and-switch fence) without moving anything shipped here.

Wire parity: metric names/labels are unchanged (`capture_events_rerouted_*`, `capture_primary_sink_health`, `capture_fallback_sink_failovers_total`, `capture_split_sink_selected`, `capture_failover_breaker_*`, `capture_event_batch_size`); known metrics-only deltas are documented in the plan doc (batch-size histogram now records on the fallback-to-S3 path; print/noop single sends record it).

## How to review this (draft): does the abstraction hold?

This is a draft PR; the ask is a **design review of the strata**, not line-by-line correctness (the golden oracle + 1268 tests carry correctness). Read the final state top-down — the commits are for later PR splitting, not for review order. Roughly 60–90 minutes:

**0. Vocabulary and invariants (10 min)** — `rust/capture/OUTPUTS_REFACTOR_PLAN.md`, the *Context*, *Vocabulary rules*, and *Invariants* sections. Everything below uses these terms.

**1. `src/pipeline.rs` — the address model.** `Pipeline` (decided at the edge, never changes), per-pipeline lane enums, `Address`, `KeyPolicy`, and `resolve()` (the one lane decision, precedence dlq > custom > historical > overflow > main). Ask: *are invalid (pipeline, lane) pairs actually unrepresentable? Is `Custom` as an address (not a lane) the right shape? Is `resolve` really the only place lanes are decided?*

**2. `src/outputs/mod.rs` — the produce surface.** `PrepSpec`/`prepare_batch` (payload assembly), the recursive `Output` policy tree (single | failover | split), the per-mode tables (`AnalyticsFamilyOutputs`, `SessionReplayOutputs`) and capability traits (`PublishesAnalyticsFamily`, `PublishesAi`, `PublishesSessionReplay`). Ask: *is "policies orchestrate sinks, above serialization, below routing" the right home for failover/split? Do required-field tables + sealed traits genuinely buy config narrowness by construction, or is the generic `State<T>` machinery more weight than it's worth?*

**3. `src/serialization.rs` — the format seam.** `Format` × `Envelope` with content headers. Ask: *would a protobuf cutover land here as an output-level config change without touching sinks or call sites?*

**4. `src/sinks/` — mechanism and namespace.** `sink.rs` (`AddressedPayload`, the `Sink` trait: prepared payloads in, per-event results out), `topics.rs` (`TopicTable`: one cluster's Address → topic realization, mode-scoped completeness check), `kafka.rs` (transport only). Ask: *do sinks stay pure mechanism — no routing decisions, no outputs-layer imports? Is address→topic-at-the-sink the right seam for cluster migration (see the coordinator design note in the plan doc)?*

**5. `src/setup.rs` + `src/router.rs` — the wiring.** `create_row_factory` (one `Output` row per pipeline; per-pipeline `CAPTURE_OUTPUT_*` overrides; producers shared per broker set; boot-time completeness + optional topic verification), the three router constructors (`router` / `ai_router` / `session_replay_router`) monomorphized per `CaptureMode`. Ask: *does mode → table type → capability → mounted ingress read as one coherent chain? Are the boot-refusal rules (overrides × S3-fallback/AI-secondary) the right conservatism?*

**6. Caller experience (skim)** — `v0_endpoint.rs`, `ai_endpoint.rs`, `payload/*.rs`: what publishing looks like from a handler. Ask: *is the call site simpler or noisier than before?*

Skip for this review: the v1 bridging mechanics (`v1/sinks/*` — deliberate minimal convergence), `failover.rs` breaker internals (ported intact, dark), and all test migration churn. The plan doc's progress tracker maps each commit to its parity proof if you want the history.

Automated only, at every commit boundary:

- `cargo test -p capture` — full suite green at each of the 35 commits (934 lib tests + parity/dynamic suites; legacy v1 sink-stack tests retired with the stack + all integration suites: events, ai_endpoint 64, ai_restrictions 7, otel 21, analytics/replay restrictions, s_endpoint, kafka_headers 9, quota_limiters 38, recordings 10, v1_pipeline 17, v1_sink_integration 10, v1_http_integration 5).
- The step-1 routing goldens pass unmodified through every commit — they drive the exact production path (lane resolution → serialization → prep → publish).
- New tests: serialization unit tests (lz4 prefix format round-trip, content headers), `pipeline::resolve` precedence tests (ported assertions), output-policy tests (failover reactive/advisory/fatal-no-failover, split routing), 8 ported breaker state-machine tests (manual clock, no sleeps) plus 4 output-level breaker behavior tests (open→fallback→recover cycle, control-plane-down, fatal handling).
- `cargo clippy -p capture --all-targets -- -D warnings` and `cargo fmt` clean at each commit.

Not done: no manual/staging verification of the dark-launched breaker path (it is unreachable until `CAPTURE_FAILOVER_ENABLED` is set).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal architecture only; the plan doc in-repo is the documentation.

## 🤖 Agent context

- Built with Claude Code. The branch supersedes the earlier "sinks everywhere" iteration of this refactor (preserved on `pl/ingestion/capture-sinks-refactor-sinks-v1`); the design discussion concluded that a caller-facing two-phase `Sink` trait was one layer short, and the target became the outputs model above. The history was rebuilt from the merge-base as a fresh commit sequence; steps 1–3 were reproduced exactly from the old branch, and the breaker state machine, mode-scoped completeness, and v1 topic convergence were ported from it rather than rewritten (port sources are named per step in the plan doc).
- Notable decisions: failover/split live in the outputs layer as a recursive policy tree (composing like the old sink composites did) rather than as sinks; `Prepare` is backend SPI, never caller API; policies route events *before* prep so each target serializes for itself, which removed the old cross-format prepared-record sharing and the split sink's token-header round-trip.


rust/capture/src/v1/sinks/kafka/mod.rs
eli-r-ph, line 1: I do want to be sure we don't lose metrics/alert/tag granularity in v1 as part of the change to new Sinks. We'll also need to be very careful as we roll out v1 in parallel to this change that we keep alerts/monitors, dashboards, and logging observability in sync as we roll these changes out for v1 (same for legacy but I'm sure at the sink point this will be a net observability improvement there!)

rust/capture/src/v1/sinks/DESIGN.md
eli-r-ph, line 489: this is the sort of thing I don't want to gloss over or make too coarse as we integrate v1 into the new sinks

eli-r-ph, line 19: I also think, since the legacy -> v1 transition will be ongoing after this lands, we should ensure we have a v1 scoping tag of some kind integrated into logs/metrics from the Sinks. Being able to inject the v1-specific `Context` into the observability (esp. logs) in the Sinks is a huge win due to the structure and detail there, helps us facet related logs at any granularity (team/token, request, path, and more) across the whole request including sink behavior

rust/capture/tests/v1_sink_integration.rs
eli-r-ph, line 132: line 132 looks like cruft on this change

rust/capture/src/pipeline.rs
eli-r-ph, line 14: 👍

eli-r-ph: 🎗️ the new `capture-import` deploy is not yet configured to route AI events to historical (sources are managed migrations only atm) which is disabled currently in configs in prod. historical-migration events still flowing to `capture-analytics` will still route AI events to historical if they are so marked.

If we need to change that in the near term for batch-import captured events, let me know and I can update it to configure those topics in the new charts and register AI routes in `CaptureMode::Import` deploys prior to enablement.

rust/capture/src/setup.rs
eli-r-ph, line 806: just landed `CaptureMode::Import` as requested prep for rate limiter rollout; will need to do some spot updates/rebases as you split these up, sorry! 🙇

rust/capture/src/outputs/dynamic.rs
eli-r-ph, line 193: Do we want to panic here if this `expect` fails?

eli-r-ph:
So far this looks good to me, and the architecture changes to decouple routing, serialization, publish along the lines of what v1 is laying out are terrific. I like the initial logic setting up the fallback sink patterns as well.

I have a few concerns about observability and behavior we're generalizing in the sinks but nothing blocking 👍 carry on!

PR #71518, opened 2026-07-16
chore(aio): remove trial evaluations

carlos-marchal-ph:
## Problem

Trial evaluations let a team run LLM-judge evals on PostHog-funded inference up to a per-team cap, as a getting-started affordance before bringing their own key. #68234 grandfathered them behind a deprecation window that closes `2026-07-17` UTC (exclusive: the last live moment is the end of `2026-07-16` UTC). This PR removes the concept entirely. Every LLM-judge eval now requires the team's own provider key (BYOK). The playground keeps its PostHog-funded models.

## Changes

- Drop the trial data model: `EvaluationConfig.trial_eval_limit` / `trial_evals_used` and the `is_trial_grandfathered` / `trial_evals_remaining` properties. Remove the `trial_limit_reached` and `model_not_allowed` status reasons (a keyless eval now surfaces `provider_key_required`).
- Rewrite enable-gating (`EvaluationSerializer._validate_keyless_resolution`) and its runtime counterpart (`model_resolution.py`): a null-config eval resolves via the team's active key, otherwise `provider_key_required`. There is no funded/grandfathered path.
- Rename the funded-model list `trial` to `playground` (the playground still runs free models, evals do not).
- Remove the trial provider-key endpoints (`trial_evaluations`, `assign`), the `TrialUsageMeter` / deprecation banner, trial usage-report tracking, the trial warning/exhausted emails, and the `AI_OBSERVABILITY_TRIAL_EVAL_DEPRECATION_DATE` setting plus `constants.py`.
- Migration `0023`: relabel leftover trial status reasons to `provider_key_required`, alter the enum, and retire the two columns from Django state (drop `NOT NULL`, keep the columns) so pods on the previous release survive the rollout. The physical `DROP COLUMN` follows in a stacked PR.
- Regenerate frontend and MCP API types.

> [!NOTE]
> Base of a 2-PR stack. The follow-up physically drops the columns after this ships.

## How did you test this code?

Claude ran the affected suites locally, all green:

- Backend api/model/llm: 176 passed (`test_evaluation_config`, `test_evaluations`, `test_models`, `test_provider_keys`, `models/test_evaluations`, `llm/test_config`)
- Temporal: 169 passed (`test_model_resolution`, `test_run_evaluation`, `test_run_tagger`, `test_metrics`, `test_run_trace_evaluation`, `test_module_integrity`)
- Usage-report task: 19 passed
- Frontend typecheck and affected Jest suites: pass
- `makemigrations --check`: no drift. ruff clean.

[...]

PR #72692, opened 2026-07-21
chore(aio): drop trial eval columns

carlos-marchal-ph:
## Problem

Final phase of the trial-evaluations removal (#71518). That PR retired `EvaluationConfig.trial_eval_limit` / `trial_evals_used` from Django state but kept the physical columns, plus a temporal compatibility layer (no-op stub activities and `workflow.patched("remove-trial-evals")` gates) so runs started on the previous release could drain. The previous release is fully rolled off, so all of it can go. Also fixes the bug from [review comment r3617763877](https://github.com/PostHog/posthog/pull/71518#discussion_r3617763877): deleting the team's active provider key with a replacement selected left `EvaluationConfig.active_provider_key` NULL via the SET_NULL cascade, stranding keyless evals that resolve through it.

## Changes

- Migration `0027`: reruns the trial status-reason relabel. Old-release workers could still write `trial_limit_reached` / `model_not_allowed` during the #71518 rollout after 0024's sweep ran; no writer exists anymore, so this sweep is final.
- Migration `0028`: `DROP COLUMN IF EXISTS` for both retired columns. State was removed in 0026, so this is the staged second phase. `reverse_sql` restores the exact 0026 column state (nullable, DB defaults 100/0) because `test_migration_0026` reverses the app from the leaf on every run.
- Removes the temporal compatibility layer: stub activities, their registrations, all legacy branches in the evaluation, trace-evaluation and tagger workflows, and the module-integrity entries. Per the temporal-workflow-versioning rule, each workflow keeps a single `workflow.deprecate_patch("remove-trial-evals")` for one release (every #71518-era history carries the marker); a trivial follow-up deletes it.
- Provider-key destroy with `replacement_key_id` now reassigns the team's `active_provider_key` to the replacement inside the existing transaction, when the deleted key was the active one.

> [!WARNING]
> Pre-merge checklist: confirm every region has completed at least one deploy containing #71518. Pods older than that still select the dropped columns.

Stacked on the migration-analyzer fix PR (the staged drop is only recognized with it).

- Three new endpoint tests (TDD, red before the fix): deleting the active key with a replacement switches the active key; deleting a non-active key leaves it alone; an unhealthy replacement still becomes active (deliberate, the user chose it, precise runtime errors beat a generic NULL disable). They catch the SET_NULL regression and the reassignment invariants nothing else covered.
- No new migration tests: the re-sweep reuses 0024's already-tested logic, and `test_migration_0026` exercises 0028's reverse on every run (it passed post-change).
- Suites: provider-keys 62 passed, temporal suites green, migration test passed, `makemigrations --check` clean, risk analyzer reports 0028 as properly staged with 0 blocked, preflight 0 failures.
- Reviewed by two independent agent reviewers on top of the build workflow's per-task and whole-branch reviews; their one substantive finding (keep `deprecate_patch` for a release instead of outright removal) is incorporated.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs reference trial evals.

## 🤖 Agent context

I directed this, Claude (Claude Code) implemented it via a subagent-per-task workflow with per-task review, a final whole-branch review, and two extra independent reviews. Skills invoked: /django-migrations, /writing-tests. Notable decisions: the reverse_sql on 0028 exists for CI reversibility (TestMigrations reverses the leaf), not prod rollback; `deprecate_patch` stays for one release per `.claude/rules/temporal-workflow-versioning.md` rather than deleting the patch outright. The nodejs union cleanup ships separately because the migrations-service separation check forbids nodejs and Django migrations in one PR.


products/ai_observability/backend/api/provider_keys.py
```diff
@@ -448,6 +448,15 @@ def destroy(self, request, *args, **kwargs):
                 LLMModelConfiguration.objects.filter(id__in=model_config_ids, team_id=self.team_id).update(
                     provider_key=replacement_key
                 )
+
+                # The deleted key may also be the team's active key — the fallback that keyless
+                # evals resolve through. Honor the user's replacement choice instead of letting
+                # the SET_NULL cascade strand them.
+                config = EvaluationConfig.objects.filter(team_id=self.team_id, active_provider_key=instance).first()
```
Radu-Raicea, line 461: The replacement key can be the same as the deleted key, so we should probably do a check

Radu-Raicea (approved):

products/ai_observability/backend/api/provider_keys.py
carlos-marchal-ph: Oof good catch would've been a mess 😅

PR #72699, opened 2026-07-21
feat(ingestion): add fan-out/fan-in chunk pipeline stage

pl:
## Problem

A pipeline step can't change element cardinality, so when one element carries N independent pieces of work (an AI event with several blobs to upload), the step has to either process them sequentially or hand-roll concurrency inside itself. That's exactly what `offloadAiBlobsStep` on the base branch does with its own `mapWithConcurrency` worker pool, fighting the `maxConcurrency` and retry machinery the pipeline framework already provides. This came up in review discussion on #72386.

## Changes

**Framework: staged fan-out/fan-in stage** (`FanOutFanInChunkPipeline`, synchronized via `InterleavingChunkPipeline` like `filterMap`), built as `.fanOut(fn).via(cb).fanIn(fn)`:

```ts
.pipeChunk(splitBlobsStep)                    // element -> { event, blobs }, cheap
.fanOut(extractBlobsFanOut)                   // element -> sub-elements (sync, named fn)
.via((sub) =>
    sub.concurrently((b) => b.pipe(uploadBlobStep, { retry }), {
        maxConcurrency: 8,                    // one cap across all parents' subs
    })
)
.fanIn(mergeBlobPointersFanIn)                // (original, subResults) -> merged element
```

The sequence is enforced by the type system: `.fanOut()` returns an intermediate whose only method is `.via()`, which returns an intermediate whose only method is `.fanIn()` — an unclosed stage has no `build()`, no result handling, and is not assignable where a `ChunkPipelineBuilder` is expected. The intermediates hold completion closures rather than the previous pipeline directly (the `GroupProcessingBuilder` precedent), so `ChunkPipelineBuilder` stays covariant in its output type and covariant subpipeline callbacks keep typechecking. A type-level test (`@ts-expect-error` block) guards the sequencing against future widening.

The common ingestion skeleton (`CommonTeamStage`) exposes `fanOut` as a first-class method alongside `concurrentlyPerGroup`, delegating through the committed chain, so pipelines built on the skeleton use the staged API directly instead of the `compose()` escape hatch. The sequencing guarantee holds there too (an unclosed stage can't reach `afterBatch`/`build`), with its own compile-time guard in the common-ingestion tests.

Semantics:

- Cardinality is preserved at the parent level: N parents in, N results out. Sub-element cardinality is fully contained inside the stage.
- Parents emit as they complete (unordered), same contract as `concurrentlyPerGroup`. Zero fan-out completes the parent immediately via `fanInFn(element, [])`.
- Non-OK parents pass through without fanning out. Sub-result contract: OK sub-results are collected for the fan-in; DROP is the sanctioned way for a sub-step to exclude its sub-element (silent, the parent fans in with the survivors); **a DLQ sub-result fails the whole parent** — instead of fanning in, the parent emits a DLQ aggregating its sub DLQs (count and distinct reasons in the reason, every error via `AggregateError`) once all siblings drain, resolving the review finding where an excluded upload DLQ let the fan-in emit a pointer to a blob that was never stored; REDIRECT sub-results are excluded with a warning log, since sub-elements are not Kafka messages — routing belongs on the parent before fan-out. DLQ wins over drop-exclusion.
- Redirect sub-results can never escape the stage (and a DLQ carries no redirect names), so the subpipeline's redirect names don't propagate to the stage's result type — `handleResults` won't demand outputs for redirects that cannot happen.
- Sub-pipelines are context-agnostic: the `via` builder is typed over the minimal base context, not the parent's context type, so context-gated surface (`teamAware`, `messageAware`, `handleIngestionWarnings`, `handleResults`) is uncallable inside — sub warnings and side effects merge into the parent and are handled once by the outer pipeline. Fan-out functions put any team/message data sub-steps need into the sub-element value; sub-elements carry only a minimal context (fresh side-effect/warning arrays plus a module-private correlation tag), and the stage owns the parent's context in place — no per-element copies on the hot path.
- Sub-elements get per-sub contexts with fresh side-effect/warning arrays, merged back on fan-in (for excluded sub-results too), so nothing is double-counted. Correlation is a public, opaque token: the sub context type is `FanOutSubContext` — a symbol-keyed `FanOutParentRef` (nominal via a private brand) the stage mints per parent and resolves through a `Map`. It survives context spreads and subpipeline reordering; sub steps can see it but not use it, and unknown tokens fail loudly. The stage file has zero type casts.
- Thrown errors (fan-out, fan-in, or subpipeline) poison the stage permanently after in-flight work drains, standard interleaving behavior.

Also adds living-docs chapter 17 (`docs/17-fan-out-fan-in.test.ts`). The review guidance is folded into the pipeline-doctor ecosystem per review feedback (no standalone skill): a fan-out/fan-in convention section in the `ingestion-pipeline-doctor-nodejs` skill, plus concern-matched rules in the composition, result, and step doctor agents.

**Adoption: AI blob offload** now splits the old step in two:

```mermaid
flowchart LR
    A[offloadAiBlobsStep<br/>extract + upload with<br/>hand-rolled worker pool] --> B[downstream steps]
    class A,B phGray;
```

```mermaid
flowchart LR
    E[extractAiBlobsStep<br/>detect blobs, build plan<br/>no I/O] --> F{{fanOut / via / fanIn}}
    F --> U[uploadAiBlobStep<br/>per blob, concurrently<br/>maxConcurrency 8, retry x5]
    U --> M[mergeAiBlobPointersFanIn<br/>apply pointers, record metrics]
    M --> B[downstream steps]
    class F phBlue;
    class E,U,M,B phGray;
```

Upload-before-emit is preserved (fan-in only runs once every blob is durable), and metrics are still recorded only for events that complete, now in the fan-in.

Upload error contract (addressing the review finding that upload errors "just throw, no dlq logic"): the blob store encapsulates the provider taxonomy and throws only its own `BlobStoreError`, carrying a framework-facing `isRetriable` flag and the provider error as `cause`. `isRetriable: false` means the blob can never be stored — determined by the event's own content (inside the S3 implementation this is a narrow, per-name-justified dispatch: `EntityTooLarge` — body exceeds the single-PUT size limit; `MetadataTooLarge` — headers derived from the blob's mime exceed the metadata limit; both unmodeled by the SDK, so they're recognized by narrowing to `S3ServiceException` first and only then checking the code name). The step-retry wrapper turns those into DLQ sub-results, and the fan-out stage aggregates them into a parent-level DLQ — the event dead-letters instead of crash-looping the lane. Everything else is `isRetriable: true`: transient errors retry, and environment-wide problems (`AccessDenied`, `NoSuchBucket`, credentials) crash loudly after exhaustion — dead-lettering those would silently divert the whole lane's data during an infra incident.

Startup healthcheck: the blob store lives in the consumer scope as a component whose start runs a self-test exercising every operation `ensureStored` performs — PUT a tiny sentinel object under the configured prefix, HEAD it, self-COPY it — so insufficient credentials or an unreachable bucket fail the deployment before it consumes traffic (mirroring the Kafka consumer's connect-at-start contract). Runtime health is deliberately left to the pipeline's crash/retry semantics plus the S3 error metrics rather than an ongoing probe — under traffic the pipeline is its own health signal, and a probe wired into the shared `/_health`/`/_ready` aggregation would risk liveness flaps during S3 blips that restarts cannot fix.

> [!NOTE]
> Two intentional behavior changes. Retry granularity: a transient S3 failure now retries only the failing blob's upload instead of re-running extraction plus all uploads for the event. Ordering: blob-carrying events can now emit ahead of or behind their chunk neighbors (zero-blob events keep source order relative to each other); the events and ai_events sinks are order-insensitive, and analytics already reorders across `concurrentlyPerGroup` groups.

The upload concurrency cap now applies across all events in the chunk rather than per event, so a burst of multi-blob events can no longer open `8 * events` S3 requests at once — and it is configurable via the new `AI_BLOB_OFFLOAD_UPLOAD_MAX_CONCURRENCY` setting (default 8, the old constant's value), flowing through `aiBlobOffloadConfig` like the other blob offload settings.

- `hogli test nodejs/src/ingestion/framework` — 41 suites, 441 tests, all passing (includes the 12 new stage tests and the 2 docs-chapter tests).
- `hogli test nodejs/src/ingestion/pipelines/ai` — 24 suites, 877 tests, all passing.
- `tsc --noEmit` over `nodejs/` introduces no new errors; eslint and prettier clean on all touched files.

New tests and the regression each catches, none covered before:

- Stage unit tests (`fan-out-fan-in-chunk-pipeline.test.ts`, via the public builder API): parent-level cardinality and zero-fan-out completion; non-OK parent passthrough without fanning out; sub-result correlation under subpipeline reordering (`concurrentlyPerGroup`), where index-based correlation would mis-merge; drained completed parents surfacing before a later in-flight sub failure; empty upstream chunks passing through as `[]` (not null, which downstream reads as drained); silent exclusion of dropped sub-elements; parent-level DLQ aggregation for sub DLQs (single, multiple with distinct reasons, and mixed with drops) with fan-in never invoked; warned exclusion of redirect sub-results with the parent still completing; side-effect/warning merge without double-counting; permanent poisoning from fan-out, fan-in, and sub-step throws; head-of-line avoidance for a batch fed while a prior parent is parked. A compile-time `@ts-expect-error` block additionally locks in that the staged API can't be escaped (no `build` or other surface on an unclosed stage).
- Skeleton runtime tests (`common-ingestion-pipeline.test.ts`): a batch through `fanOut/via/fanIn` composed with surrounding team-aware steps (including a dropped sub-element and a zero-fan-out parent), asserting downstream steps see the fanned-in values — catches committed-chain mis-delegation that no framework test can; and warning routing from sub-steps to the warnings output with the resolved team. Compile-time guards additionally lock the staged sequencing and the context-agnostic sub-builder. How `maxConcurrency` and per-sub `retry` compose with the stage is documented (and asserted) in living-docs chapter 17 rather than re-tested in the unit suite.
- AI step tests were restructured to run the same extract → fanOut → via(upload) → fanIn wiring the pipeline uses, keeping the existing coverage (pointer rewrite, passthroughs, blob cap, dedup, upload failure leaves the event untouched). The old per-step concurrency-cap test was dropped, since the cap now lives in the framework stage and is covered there.
- Integration tests (`blob-store.integration.test.ts`) run `S3BlobStore` against the dev stack's SeaweedFS (`:8333`, the session replay v2 instance), pinning what mocks can't falsify: `instanceof NotFound` narrowing a real 404 on the head-miss path, stored Content-Type/bytes, `'fresh'` on an immediate re-store, the touch path's real self-COPY with `MetadataDirective: REPLACE` and the percent-encoded `CopySource` (prefix containing a space and `+`), and the startup self-test against a real bucket plus its op-naming failure against a missing one. They use the shared `posthog` bucket under a unique per-run prefix, since the dev SeaweedFS node's volume slots are fully allocated and a dedicated test bucket would hang on volume assignment.
- Store-level tests pin the abstract error contract (every failure surfaces as `BlobStoreError` with the right `isRetriable` and the provider error as `cause`; the two event-specific codes non-retriable via `S3ServiceException` narrowing — including a guard that codes on non-S3 errors are never interpreted; transient/auth/bucket/network failures and the copy-race `NoSuchKey` retriable, with a rerun self-healing via a fresh HEAD; the startup self-test issuing PUT/HEAD/COPY against the sentinel key and naming the failing op), and retry-wrapped step tests pin what the pipeline sees (non-retriable → DLQ sub-result without a crash, retriable → rejects after exhausting retries, transient → recovers). No scope-wiring test for the startup self-test: the consumer suite has no scope-start test pattern, so it fails the value gate.
- Per-step unit tests for the blob offload steps (same file, nested describes) localize what the composed path can't observe: extraction's null-plan vs skip-plan distinction (skip plans record metrics; disabled events must not), below-floor accounting, the empty-blobs guarantee on the limit-exceeded path, savedChars under dedup, fan-out's empty cases and teamId mapping, the upload step's wiring-bug throw and outcome passthrough, and fan-in's metrics recording — which no test asserted anywhere before.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Framework documentation lives in the in-repo living docs; this PR adds chapter 17 for the new stage. No user-facing docs affected.

## 🤖 Agent context

Implemented with Claude Code from a written plan. Skills invoked: `/writing-tests` before authoring the test files. Decisions along the way: the stage reuses `InterleavingChunkPipeline` for synchronization (same wiring as `filterMap`) rather than owning its own drain loop; parent correlation went through a symbol-keyed context property instead of positional indexing because the subpipeline may reorder freely; metrics recording for the blob offload moved into the fan-in so failed events still record nothing, matching the old step's behavior. The builder API was first shipped as a single three-argument `fanOutFanIn()` method and then reworked per review into the staged `fanOut().via().fanIn()` chain, with the intermediates using the `GroupProcessingBuilder` completion-closure pattern to keep the builder covariant; the staged shape also resolved a variance false positive the single method had triggered in the experimental tsgo checker (the authoritative `tsc` check that CI runs was clean in both shapes).


carlos-marchal-ph (approved):
Really like the semantics here, and much more robust than what I'd rolled on my own

.agents/skills/reviewing-fanout-fanin-pipelines/SKILL.md
```diff
@@ -0,0 +1,74 @@
+---
```
jose-sequeira: Does this warrant a skill? Why does this get a skill and not all other pipeline types? Why doesn't it belong in the pipeline doctor? A bit confusing IMO

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
```diff
@@ -83,53 +121,92 @@ export function createOffloadAiBlobsStep<T extends OffloadAiBlobsInput>(
...
+    }
+    return plan.blobs.map((blob) => ({ teamId: input.team.id, blob }))
+}
+
+/**
+ * Per-blob upload step. Upload-before-emit: every blob must be confirmed
+ * durable before the rewritten event exists anywhere. A failure rejects the
+ * step; the step's retry options own transient failures.
+ */
+export function createUploadAiBlobStep(store: BlobStore | null): ProcessingStep<PendingAiBlobUpload, UploadedAiBlob> {
```
jose-sequeira, line 169: Just throws, no dlq logic 👎 errors can be non transient

nodejs/src/ingestion/pipelines/ai/consumer.ts
```diff
@@ -165,10 +166,20 @@ export function createAiConsumer(config: AiConsumerConfig, sharedScope: AiShared
     )
 
     const aiBlobStore = buildAiBlobStore(config)
+    const uploadMaxConcurrency = config.AI_BLOB_OFFLOAD_UPLOAD_MAX_CONCURRENCY
```
jose-sequeira, line 182: From experience running the `kafka-deduplicator` and other blob dependent services. max concurrency is a bad bounary for the concurrency cap, as it is prone to OOM errors, a better cap would be maximum inflight bytes, if one message can result in multiple messages, we can massively amplify the number of inflight messages without a memory cap

jose-sequeira (changes requested):
The throw error on data issues is an infinite loop that needs fixing before merge

nodejs/src/ingestion/pipelines/ai/consumer.ts
pl: Yeah, that makes sense in general, but here we want to ship something relatively basic that doesn't require a significant refactor of the pipeline concurrency code. I think there will be much more tuning to be done in this pipeline, but IMO this should be fine to unblock the deployment.

.agents/skills/reviewing-fanout-fanin-pipelines/SKILL.md
pl: Makes sense, prompted the LLM in a wrong way - now it's consistent with the rest of the skills.

nodejs/src/ingestion/pipelines/ai/pipelines/steps/offload-ai-blobs-step.ts
pl: Did a review and a refactor of the error handling, here's an LLM-generated list of changes. Hope that's good enough now.

- The store now encapsulates all AWS logic and exposes its own BlobStoreError { isRetriable, cause } — AWS types never cross its boundary (7aced28da64)
- isRetriable: false is reserved for failures provably caused by the event's own content (EntityTooLarge, MetadataTooLarge; unmodeled by the SDK, so narrowed via instanceof S3ServiceException before any name check) → DLQ sub-result → the fan-out stage aggregates into a parent-level DLQ with all sub reasons
- Everything else (transients, AccessDenied, NoSuchBucket, unknown) stays retriable → retry with backoff, then crash — deliberate: dead-lettering env-wide failures would silently divert the whole lane's data during an infra incident
- Env-wide problems are caught at deploy time instead: the startup healthcheck now self-tests the exact operations we use (PUT + HEAD + self-COPY of a sentinel key), so insufficient permissions fail readiness before consuming (183704d3273); no ongoing runtime probe on purpose — under traffic the pipeline self-heals or crashes, and the S3 error metrics cover observability
- Contract pinned at three levels: store unit tests (classification + cause, anti-broadening cases asserting nothing else gets classified), step tests under production retry options (DLQ / reject-after-retries / recover), and integration tests against SeaweedFS (real 404 → NotFound, real self-copy, healthcheck) (1d57a3151fa)
- Provider error name is embedded in the message for log grepping; the full provider error rides along as cause (632b4ba1f1e)

jose-sequeira (approved):
LGTM, thanks!

PR #72700, opened 2026-07-21
feat(exports): skip capturing errors when the creator left the org

a-lider:
## Problem

Background jobs (subscriptions, alerts, exports, and cache warming) run their query as the resource creator (`created_by`) so warehouse HogQL access control resolves against that user's access. When the creator is later deactivated or removed from the org, a warehouse-backed query fails closed with `TableAccessDeniedError`.

That's debatable, but I'd keep that for now because if we use team default, it could fail when user permission is higher then default. 

Right now each of these capture it as an exception, so a single creator spams error tracking and alerts on every scheduled run — a daily subscription, for example, keeps firing (and keeps erroring). 

## Changes

- New shared helpers in `posthog/query_creator_access.py` to capture event instead of the exception. 
- Wired into every site that runs a query as `created_by`
	- **Exports** 
	- **Cache warming** 
	- **Alerts**

Genuine denials (creator still a current member, but lacking access to a specific table) and all other exceptions still capture exception

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

The work started from an investigation into recurring `TableAccessDeniedError`s on scheduled subscription/export deliveries, which traced back to creators who had left the org while their background jobs kept running as them. We audited every place that executes a query as `created_by` (cache warming, alerts, subscriptions, exports) and found the same failure mode at each.

Design decisions: kept creator-access as the principal (a larger change — switching to a team-default principal — was considered and deferred); centralized the check + event in one helper so the four sites stay consistent; scoped the downgrade narrowly to `TableAccessDeniedError` from a revoked creator so genuine per-table denials and real errors still surface. The alert site deliberately mirrors the existing `AlertExtractionError` "fail-loud, don't capture" precedent already in that file.


posthog/temporal/alerts/activities.py
```diff
@@ -250,6 +253,29 @@ def _evaluate() -> EvaluateAlertResult:
                 should_notify=False,  # disable_invalid_alert already emailed subscribers
                 new_state=AlertState.ERRORED,
             )
+        except TableAccessDeniedError as err:
+            logger.exception(f"Alert id = {alert.id}, failed to evaluate", exc_info=err)
```
ReeceJones: Use structured logging here so that it is easier to filter logs by this

ReeceJones (approved):
just address the logging comment and i think its good to ship

PR #72707, opened 2026-07-21
feat(logs): support 30-day retention pricing

abheek9:
## Problem

Successor to #61082 (by @pawel-cebula), rescoped to **30-day retention only** — the 90-day retention add-on is dropped for now. #61082 had also fallen ~6.5k commits behind master.

Companion PRs: billing config PostHog/billing#2069, pricing page PostHog/posthog.com#18766.

## Changes

Everything from #61082, merged up to current master, minus all `logs_retention_90d` pieces:

- `LOGS_RETENTION_30D` available feature, gating the 30-day logs retention setting.
- Retention entitlement enforcement: Temporal workflow (`retention_entitlements`) that resets over-entitled teams back to 14-day retention, registered on the general-purpose queue.
- `retention_days` validation on team/project APIs: valid values are now `{14, 30}` — **90 is no longer accepted**, since without the 90d feature it would otherwise be free to select.
- Logs capture settings UI: 14/30-day options with upgrade hints; 90-day button removed.
- Billing UI: `logs_retention_30d` product-variant display and usage key labels.

- Rescoped the retention entitlement, team API, and project API test coverage from #61082 to 30d-only (including asserting `retention_days: 90` is now rejected as invalid).
- Merge conflicts with master were resolved as unions of both sides; Python syntax verified locally. Relying on CI for the full suite — local test bootstrap isn't available in this environment.


posthog/api/team.py
```diff
@@ -46,7 +46,8 @@
 from posthog.models.event_ingestion_restriction_config import EventIngestionRestrictionConfig
 from posthog.models.filters.utils import validate_group_type_index
 from posthog.models.group_type_mapping import cached_group_types_for_team
-from posthog.models.organization import OrganizationMembership
+from posthog.models.organization import Organization, OrganizationMembership
```
pawel-cebula, line 49: This import is unused and also breaks module import in CI because `posthog.models.product_intent` does not export `promoted_product_lookup`. Removing it should fix both ruff and the OpenAPI generation failure.

products/logs/backend/temporal/retention_entitlements/activities.py
```diff
@@ -0,0 +1,83 @@
...
+    async with Heartbeater():
+        logger = LOGGER.bind(dry_run=input.dry_run)
+        batch_size = max(1, input.batch_size)
+        teams_to_update: list[Team] = []
+        teams_checked = 0
+
+        # This scans all teams with paid Logs retention enabled. Run it as explicit reconciliation
+        # after Billing removes those entitlements; new over-entitled writes are blocked by the API.
+        async for team in (
+            Team.objects.filter(logs_settings__retention_days__in=list(LOGS_RETENTION_FEATURES_BY_DAYS.keys()))
```
pawel-cebula, line 31: I checked and there are a couple of non-demo teams already saved with `retention_days = 90`. What do we want to do with those? As written, they won't be picked up by this reconciliation, so they may keep 90-day retention without a matching paid entitlement. Should we explicitly migrate them down to 30 days for this launch or rather leave them on it so we can continue testing 90 days before we start charging for it?

abheek9: @pawel-cebula do you mind sharing which teams so i can have a look? I think lets migrate them down, and then note to migrate them back up when we release 90 days?

cc @cohenroy

pawel-cebula: Two in US - one is PostHog, and the other is RV Envy (0178826e-360c-0000-323d-2cfebcb4f002), which I think is Cory’s project. I didn’t find any in EU.

Just to clarify - no special migration needed either way - just a product decision whether to keep them on 90 days or put them on 30 days - if 30, then we should let them know we're downgrading so it's not unexpected.

products/logs/backend/temporal/retention_entitlements/workflow.py
```diff
@@ -0,0 +1,36 @@
...
+        EnforceLogsRetentionEntitlementsOutput,
+    )
+
+
+@workflow.defn(name="enforce-logs-retention-entitlements")
+class EnforceLogsRetentionEntitlementsWorkflow(PostHogWorkflow):
+    """Manual reconciliation workflow for Logs retention entitlements.
+
+    This workflow is registered for explicit runs via Temporal management commands,
+    but is not added to `posthog.temporal.schedule.schedules`.
```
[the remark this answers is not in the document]
pawel-cebula: @abheek9 I think this is valid. This could be enforced via the `billing` → `posthog` `billing_customer_update` path. That should cover PostHog UI cancellations, Stripe/Billing-side cancellations. You could diff `available_product_features` stored on the `posthog` org vs what's received from `billing` and if `logs_retention_30d` disappeared, force switch their retention to the default 14d. Can be done as a separate follow up.

DanielVisca (approved):
Stamped to unblock. Reviewing it now

pawel-cebula (approved):
Looks good from billing side but I'd run it by someone from @PostHog/logs too.

ee/billing/billing_manager.py
```diff
@@ -477,7 +483,14 @@ def update_org_details(self, organization: Organization, billing_status: Billing
             should_update_org_billing_quotas = usage_changed or had_quota_limiting_markers
 
         available_product_features = data.get("available_product_features", None)
-        if available_product_features and available_product_features != organization.available_product_features:
+        revoked_feature_keys: set[str] = set()
+        # An empty list is authoritative (e.g. every paid feature revoked), so only skip on absent.
+        if available_product_features is not None and available_product_features != (
```
DanielVisca: This now treats an empty list as authoritative and clears every product feature on the org, not just logs, and it runs on the get_billing / usage_report path. Looks like a paid org would get downgraded to free across all products if billing ever returns a customer with `[]` features (partial or error path response). Is that the intended behaviour?

ee/billing/billing_manager.py
```diff
@@ -477,7 +483,14 @@ def update_org_details(self, organization: Organization, billing_status: Billing
...
+            organization.available_product_features or []
+        ):
+            previous_feature_keys = {feature.get("key") for feature in (organization.available_product_features or [])}
```
DanielVisca: nit: `feature.get("key")` assumes every entry is a dict. `get_available_feature` guards with `f and f.get(...)` for the case where billing sends a None element.

products/logs/backend/temporal/retention_entitlements/workflow.py
```diff
@@ -0,0 +1,36 @@
...
+        EnforceLogsRetentionEntitlementsInput,
+        EnforceLogsRetentionEntitlementsOutput,
+    )
+
+
+@workflow.defn(name="enforce-logs-retention-entitlements")
+class EnforceLogsRetentionEntitlementsWorkflow(PostHogWorkflow):
+    """Manual reconciliation workflow for Logs retention entitlements.
+
+    This workflow is registered for explicit runs via Temporal management commands,
```
DanielVisca, line 19: Since this isn't scheduled and the billing hook only fires on revocation, teams that set 30 day retention before this ships won't be corrected, they just keep it for free. I believe existing orgs are intentionally left as is here, but wanted to confirm. Or is a one time run of this meant to be part of the rollout?

products/logs/backend/temporal/retention_entitlements/activities.py
DanielVisca, line 31: 90 day teams from before this change aren't covered here (the map and filter are 30 only), and the API now refuses to re-set 90, so they'd stay on 90 day retention indefinitely. Worth a one off check that none exist, or a backfill to 14?

posthog/models/team/logs_retention.py
```diff
@@ -0,0 +1,42 @@
...
+        return 0
+
+    teams = list(
+        Team.objects.filter(organization=organization, logs_settings__retention_days__in=revoked_days).only(
+            "id", "logs_settings"
+        )
+    )
+    for team in teams:
+        # Preserve unrelated Logs settings such as JSON parsing and PII scrubbing.
+        team.logs_settings = {**(team.logs_settings or {}), "retention_days": DEFAULT_LOGS_RETENTION_DAYS}
```
DanielVisca, line 33: nit: the reset leaves `retention_last_updated` untouched, so someone who re-subscribes right after an involuntary reset could hit the 24h throttle when going back to 30.

products/logs/backend/temporal/retention_entitlements/activities.py
abheek9: thanks yes he have done a one time check here!

ee/billing/billing_manager.py
```diff
@@ -254,9 +255,21 @@ def update_available_product_features(self, organization: Organization) -> list[
...
         available_product_features = available_product_features_json.get("available_product_features", [])
+        previous_feature_keys = {
+            feature.get("key") for feature in (organization.available_product_features or []) if feature
+        }
         organization.available_product_features = available_product_features
         organization.save()
 
+        # Only reset on a non-empty list: the retention reset is not self-healing, so an
+        # empty error-path response must not permanently downgrade team settings.
+        if available_product_features:
```
DanielVisca, line 266: On this path the org's features are already overwritten with the empty list two lines up, so this guard only skips the reset — the org ends up de-entitled while the team keeps `retention_days: 30`. `cancel_trial` calls straight into here, and `get_plan_tier` treats an empty feature list as the free tier, so I think a cancelled trial lands in exactly that split state (the test that covered it now mocks a non-empty list). Should the guard be on `update_org_details` only?

posthog/api/team.py
```diff
@@ -1440,10 +1456,10 @@ def validate_receive_org_level_activity_logs(self, value: bool | None) -> bool |
 
         return value
 
-    VALID_RETENTION_DAYS = {14, 30, 90}
+    VALID_RETENTION_DAYS = {14, 30}
```
DanielVisca, line 1459: Dropping 90 from the valid set means a team still saved at 90 gets a 400 on every `logs_settings` write, not just retention changes — the settings UI spreads the whole `logs_settings` object into each toggle, so `retention_days: 90` rides along when someone flips JSON parsing or PII scrub. `test_logs_settings_non_retention_changes_not_restricted` is that exact shape with 14. Did the one off check actually move those teams down, or just confirm which ones they are?

DanielVisca (approved):
Had another look now the billing fix is in. Two things below, otherwise this looks good to me.

posthog/api/team.py
abheek9: we are going to manually move them to 30 days after the migration finishes only a few teams have 90 (maybe one non posthog team)

PR #72712, opened 2026-07-21
feat(ingestion): consolidate warning registry into one generated arti…

z0br0wn:
## Problem

The ingestion warning registry in Node.js is the source of truth for warning types and their category/severity classification, but Rust producers only saw a slice of it. The generated artifact carried just the capture-produced type strings, because the capture envelope transport lets the Node consumer stamp classifications at serialization time. Services that produce warnings directly to the warnings topic have no Node consumer in the path, so they must serialize the final payload themselves, and today that means hand-mirroring classifications in each service. Hand-mirrored classifications drift.

There were also two hand-maintained copies of "the set of types capture emits": the `from_tag` allowlist in Rust and the `captureProduced` flags in Node (which derive the consumer's trust allowlist for the capture envelope). Nothing kept them equal, and skew fails silently in either direction: a flagged type without a `from_tag` arm is never emitted, and an arm without the flag is emitted and then rejected at the consumer's trust check.

## Changes

**Before**

  ```mermaid
  flowchart LR
      REG[Node registry] -->|captureProduced only| ART[capture artifact\ntype strings]
      ART --> ENUM[WarningType\nas_str only]
      ENUM --> ENV[capture envelope\nNode stamps classification]
      REG -.->|hand-mirrored| SVC[direct producers\nown classification copies]
      class REG phYellow; class ART,ENUM phGray; class ENV,SVC phBlue;
  ```

**After**

  ```mermaid
  flowchart LR
      REG[Node registry] --> ART[warning_types.generated.json\nall types + classification + captureProduced]
      ART --> ENUM[WarningType\nas_str, category, severity, capture_produced]
      ENUM --> ENV[capture envelope transport]
      ENUM --> DIR[direct transport\nbuild_direct_warning_payload]
      class REG phYellow; class ART,ENUM phGray; class ENV,DIR phBlue;
  ```

  - The generator now exports the whole registry, every type with its category and severity, into a single `warning_types.generated.json`. The capture-only artifact is deleted.`build.rs` generates `category()`, `severity()` and `capture_produced()` on `WarningType` alongside `as_str()`.
  - New `direct` module builds the direct-to-topic payload (ClickHouse-format timestamp, classification stamped into details so a caller-supplied key can never override it). It has no consumer in this PR by design: the personhog services adopt it in the follow-up PR, replacing their hand-built payloads.
  - `pipeline_step` moves off the warning type onto `WarningSource`, since which stage a warning comes from is a property of the emit site, not the type. Capture's emitted envelopes are byte-identical.
  - The old one-directional round-trip test is replaced by an equality weld: `from_tag`'s domain must exactly equal the `captureProduced` set, which now rides through the artifact. Both silent-skew directions become a red test on the PR that introduces them.
  - Registers `person_properties_size_violation` (already emitted by the personhog writer, previously unregistered) and adds its frontend description.

  > [!NOTE]
  > No runtime behavior changes for any service. Capture's wire output is unchanged, and no leader/writer code is touched. The observable deltas are one newly registered type, one frontend description string, and the new test welds.

Automated only, all run by the agent on this branch against current master:

  - `common-ingestion-warnings` crate suite (27 tests) including the new `from_tag_domain_equals_the_capture_trust_allowlist` weld. That test catches a regression nothing else did: adding a capture type's flag without the Rust arm (feature silently never emits) or the arm without the flag (emitted, then rejected by the consumer trust check).
  - Capture's full suite (1,130 tests across unit and integration), proving envelope output and `from_tag` behavior are unchanged.
  - Node jest: the extended no-drift test (committed artifact byte-equals generator output, every entry's classification and flag match the registry) plus the client-ingestion-warning
  consumer step suite.
  - `cargo clippy -D warnings`, `cargo fmt`, prettier and the frontend formatter, all clean.

## 🤖 Agent context

Authored with Claude Code, directed by the PR assignee. This is the foundation split out of a larger personhog leader-admission branch so it can land first. Design decisions made along the way: per-service generated enums were built and then rejected in favor of one producer-agnostic enum (producer identity is per-message `source` data, not registry data); `captureProduced` was slated for deletion until we found its runtime consumer (the envelope trust allowlist in `handle-client-ingestion-warning-step.ts`), after which it was kept for that role and exported so the equality weld could replace the weaker round-trip test. The `/writing-tests` skill was invoked earlier in the session and shaped the test choices (the weld asserts against a shared generated artifact rather than cross-language source scraping).


rust/common/ingestion_warnings/src/direct.rs
```diff
@@ -0,0 +1,79 @@
+//! The direct-to-topic warning transport, for services that know `team_id`
```
jose-sequeira: `direct.rs` is a really confusing name here. And separation this logic from the capture one doesn't feel worth it, this probably can be unified into a unique class that guarantees side by side behaviour so that models don't drift. Only difference should be `.with_token()` with_team_id()`

z0br0wn: updated! there is a bit of an asymmetry because the capture envelope excludes some fields for downstream hydration, but tried to to them in nicely

rust/common/ingestion_warnings/src/registry.rs
```diff
@@ -52,14 +53,19 @@ mod tests {
...
+        // allowlist, carried through the artifact). Skew fails silently in
+        // production in either direction — a flagged type without an arm is
+        // never emitted; an arm without the flag is emitted and then
+        // rejected at the consumer's trust check.
         for warning in WarningType::ALL {
             assert_eq!(
-                WarningType::from_tag(warning.as_str()),
-                Some(warning),
-                "tag {} must round-trip",
+                WarningType::from_tag(warning.as_str()).is_some(),
```
nickbest-ph: nit: loosening of test/guardrail here. could be:

```
assert_eq!( WarningType::from_tag(warning.as_str()), warning.capture_produced.then_some(warning),
```

or something

rust/common/ingestion_warnings/src/serializer.rs
```diff
@@ -40,58 +53,124 @@ const PROP_TYPE: &str = "$$client_ingestion_warning_type";
...
+
+    /// Commit to the direct-row transport: the terminal v2 warning row for
+    /// the ACL-guarded warnings topic, addressed by `team_id`.
+    ///
+    /// Classification is producer-stamped here — the registry's
+    /// `category()`/`severity()`, and the row's own `teamId`, are inserted
+    /// over the caller's details so a stray key can never override them or
+    /// disagree with the addressing — and the `timestamp` uses the
+    /// ClickHouse format the v2 table materializes columns from.
+    pub fn into_row(self, team_id: i64, source: &str) -> Value {
```
nickbest-ph: could we use the WarningSource type for `source` arg here, seeing as how it was introduced in this change?

Then i think we could use the `pipeline_step` on the warning source to populate `details`, as i think we're missing that field

nickbest-ph (approved):
i think we might want to populate the clickhouse row with the `pipeline_step` detail, as we currently missing that.

o/w lgtm

PR #72714, opened 2026-07-21
feat(retention): extend legacy two-arm scan to value breakdowns

andyzzhao:
## Problem

First-time and first-ever retention queries with a value breakdown are the largest remaining source of ClickHouse read on the retention path. The two-arm scan optimization (#71817) deliberately excluded them: the tag-column arms carry only `(actor, timestamp, is_start, matches_start, is_return)`, and a value breakdown needs the breakdown property value per row. So any first-time retention insight with a breakdown still runs a single scan with no upper time bound, and with an all-events returning entity no event-name prefilter either, reading the team's entire event history per run.

## Changes

```mermaid
flowchart LR
    A[events: all time, all aggregate-input columns] --> B[GROUP BY actor, breakdown_value]
    class A phRed;
    class B phGray;
```

```mermaid
flowchart LR
    S[start arm: all time, start event names only, + breakdown_value] --> U[UNION ALL aliased as events]
    R[return arm: query window only, + breakdown_value] --> U
    U --> O[outer GROUP BY, entity conditions swapped to role tags]
    class S,R phBlue;
    class U,O phGray;
```

- `_legacy_scan_arm` now emits a `breakdown_value` column, resolved against the real events table inside each arm. For first-ever, the return arm emits the empty bucket instead: its values are never read by the outer query, so it skips the property read and the start arm stays the sole bucketing authority (mirrors `_dwh_breakdown_value_arm_expr` in the variant builder).
- `build_base_query_legacy` re-expresses the outer breakdown against the arms' column. First-time-matching keeps per-row values and adds `breakdown_value` to the outer GROUP BY, preserving per-value cohort partitioning. First-ever buckets each actor via `argMinIf(events.breakdown_value, events.timestamp, events.is_start = 1)`, matching the parent `apply_breakdown` semantics where `is_start = 1` rows are exactly the unfiltered start stream.
- `_legacy_first_time_two_arm_scan` drops the blanket breakdown exclusion. Cohort breakdowns keep the single scan: their `InCohort` predicate lives in `_event_filters`, which the arms don't apply, so letting them split would silently drop cohort filtering.

> [!NOTE]
> Query results are unchanged by design. Every retention e2e test runs on both the legacy and DWH-variant builders and asserts identical results (`RetentionBaseQueryVariantComparisonMixin`).

### Measured impact

Benchmarked on a representative production first-time retention shape (concrete start event, all-events returning entity, person-property breakdown, person-property filters), control vs candidate interleaved 3x with `use_query_condition_cache=0, use_uncompressed_cache=0, min_bytes_to_use_direct_io=1`, stats from `system.query_log`. The control SQL was verified byte-identical to what production actually executed for that insight.

| metric | two-arm vs single scan |
| --- | --- |
| read bytes | ~19x fewer |
| read rows | ~6x fewer |
| CPU time | ~14x less |
| duration | ~4x faster |

Result rows were byte-identical between control and candidate on all run pairs.

- `TestRetentionLegacyFirstTimeTwoArmScan::test_value_breakdown_scans_two_arms_carrying_breakdown_value` (parameterized over both first-time modes): catches re-adding the breakdown gate (silent perf regression) and the first-ever mis-bucketing invariant (return arm must emit the empty bucket, outer must be argMinIf over start-arm rows).
- Swapped the `breakdown` case in `test_keeps_single_scan_when_split_cannot_apply` for `cohort_breakdown`: catches letting cohort breakdowns into the two-arm path, which would silently lose the InCohort filter.
- New e2e `test_retention_first_time_matching_filters_with_event_breakdown`: no existing e2e covered first-occurrence-matching-filters + breakdown (only first-ever). Catches per-value partitioning bugs: a pre-window first occurrence must exclude the actor from that value's bucket, and return events must not count across values. Runs under the parity mixin, so legacy two-arm output is cross-checked against the variant builder.
- Full retention suite: 195 passed, all snapshots unchanged.
- Prod benchmark (described above) asserted byte-identical result rows between the current and new query shapes on every run.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal query-planner change, no user-facing behavior or docs affected.

## 🤖 Agent context

Authored with PostHog Code (Claude) from a handoff brief. Skills invoked: `/optimizing-clickhouse-and-hogql-queries`, `/writing-tests`, `/query-clickhouse-via-metabase`.

Decisions along the way:

- Kept cohort breakdowns on the single scan rather than carrying the `InCohort` predicate into the arms. Correct either way, but re-plumbing the cohort filter into per-arm WHERE clauses is a separate risk surface for a niche shape, and the heavy production queries are value breakdowns.
- For first-ever, the return arm emits a constant empty bucket instead of the property extract. The outer argMinIf is conditioned on `is_start = 1` so return-arm values are unreadable anyway; the constant saves the property-column read on the return arm and mirrors the variant builder's established invariant (start arm is the sole bucketing authority).
- Benchmark control SQL was generated locally with `prepare_and_print_ast` + `substitute_params`, string-swapped to prod materialized-column forms, and diffed byte-for-byte against the query text prod actually ran before measuring.

---


thmsobrmlr:
Obsolete after #80640, which deletes the path optimized here

PR #72719, opened 2026-07-21
feat(integrations): flag bing ads connections on the legacy oauth app

jabahamondes:
## Problem

Bing Ads has two sets of OAuth client credentials right now: the current app, and a legacy fallback kept alive by #70257 so grants minted by the previous app keep refreshing.
We want to retire the legacy one, which means telling the affected customers to reconnect first.

The blocker is that we can't tell who they are.
The only trace of a fallback refresh is the `integration_oauth_refresh{result="success_fallback"}` counter, which has no `team_id` label.
So there's no way to target a warning, and no way to know when the population has drained to zero and the credentials are safe to delete.

## Changes

**Identify the app a connection was established with.**
OIDC puts the client id in the id_token's `aud` claim, and we keep the id_token from the authorization exchange - refreshes only overwrite the access and refresh tokens.
So `issuing_oauth_client_id()` reads the app the customer actually connected through, for connections that already exist, without waiting for anything.

`flag_legacy_oauth_integrations` compares that against the current client id and flags the mismatches with `oauth_uses_legacy_client` on `Integration.config`.
It takes `--dry-run`, so running it first answers "how many customers does this affect" before we commit to a deadline.

Note it compares against the **current** client id rather than against the fallback.
Matching on the fallback would silently miss connections from any third app, which is the failure mode you don't find out about until the retirement breaks someone.

**Keep the signal fresh.**
A successful refresh also records which credential pair worked, so the flag stays correct as connections move:

```diff
+ record_oauth_client_used(self.integration, used_fallback=used_fallback)
```

Set when the fallback rescued the refresh, cleared when the primary works.
This is the fallback signal for rows with no readable id_token, and it's what clears the flag when someone reconnects.

**Warn the affected teams.**
A banner on the marketing analytics dashboard and in marketing analytics settings, shown only to teams carrying the flag, with a reconnect button pointing at the same authorize URL `IntegrationView` already uses.
Reconnecting resolves it on its own: `authorize_url()` and the token exchange always use the primary credentials, and for Bing the integration is keyed on the `oid` from the id_token rather than the client id, so a reconnect updates the row in place and replaces `config` wholesale.

> [!NOTE]
> The banner is gated on `legacy-oauth-reconnect-warning`, which acts as a kill switch. The targeting comes from the flag on the integration, not from the feature flag: a PostHog flag can't read Django model state, so gating on the data is both more precise and less to maintain. The flag doesn't exist yet, so nothing renders until it's created.

The cutoff date in `RETIRING_OAUTH_APPS` is a placeholder and needs to be set before this ships.
Both the map and `CURRENT_CLIENT_ID_SETTING` are keyed by integration kind, so the next credential rotation is a one line addition on each side.

Deleting the Bing fallback credentials is a follow-up, once the flagged population drains.
The generic `OAUTH_CLIENT_FALLBACKS` mechanism stays: it was built for exactly this kind of rotation.

For the refresh path I extended the existing parameterized fallback test rather than adding a new one, since it already exercises all four refresh paths.
Two new columns cover the flag's lifecycle: set when the fallback rescues a refresh, cleared when the primary succeeds, untouched when both fail.

For the command, three parameterized cases over the classification: superseded client gets flagged, current client clears a stale flag, missing id_token is left alone rather than guessed at.
Plus one case that `--dry-run` really doesn't write.

The regression this catches is a broken campaign in either direction. Miss the write and affected teams never hear about it; miss the clear and we keep nagging teams that already reconnected.

`test_integration_model.py -k falls_back` (5 cases) and `test_flag_legacy_oauth_integrations.py` (4 cases) pass.
Repo-wide `uv run mypy --cache-fine-grained .` is clean for these changes.

Verified against real Bing Ads connections in a dev environment:

- **The `aud` claim really is the client id.** I ran `--dry-run` against two real Bing Ads connections. Both id_tokens carry `aud`, and its value matches the configured `BING_ADS_CLIENT_ID` exactly, so both classified as current and neither fell into "without an id_token to read". Worth repeating on a prod replica before we act on the numbers, but the mechanism holds.

Still untested:

- **The banner in a browser.** It needs a live Bing Ads connection on the legacy app to render, which I can't produce locally. The copy also deserves a human read before the flag goes on.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change. The banner is the entire customer-facing surface.

## 🤖 Agent context

I directed this and Claude wrote it. Skills invoked: `/writing-tests`, which is why the refresh-path coverage landed as extra parameterized columns on an existing test instead of a new test function.

The first cut only had the refresh-path signal, which meant waiting roughly a day of sweep before anyone could be counted or warned.
Reading `aud` off the stored id_token replaced that with something answerable immediately, and the refresh signal stayed as the way the flag keeps up with reality.
An earlier idea of inferring from `Integration.created_at` against the migration date was dropped for being a guess.

On targeting: the alternative was publishing a group property via a periodic `group_identify` job so a feature flag could target it directly. It works, but buys nothing over reading the flag off `config` and costs a job plus propagation latency.

The flag rides on `config` rather than a new column because `config` is already serialized to the frontend and already carries this class of refresh bookkeeping (`refresh_failure_count`, `refresh_terminal`). No migration needed.

image


frontend/src/lib/constants.tsx
```diff
@@ -339,6 +339,7 @@ export const FEATURE_FLAGS = {
     INSIGHT_SUBSCRIBE_PROMINENT_BUTTON: 'insight-subscribe-prominent-button', // owner: @mattp #team-analytics-platform multivariate=control,test
     INTER_PROJECT_TRANSFERS: 'inter-project-transfers', // owner: @reecejones #team-platform-features
     JS_SNIPPET_VERSIONING: 'js-snippet-versioning', // owner: #team-client-libraries
+    LEGACY_OAUTH_RECONNECT_WARNING: 'legacy-oauth-reconnect-warning', // owner: @jabahamondes #team-web-analytics — kill-switch for the retiring-OAuth-app reconnect banner
```
mayteio, line 343: Curious why we'd use a const instead of an FF?

mayteio (approved):
Thanks Javier! Appreciate it :D

frontend/src/lib/constants.tsx
jabahamondes: Hey @mayteio  yes, it might be confusing but this is the file where ff are defined! 🫡

PR #72726, opened 2026-07-21
feat(cohorts): materialize single-leaf membership register in cf_stage2

matheus-vb:
## Problem

Single leaf cohorts never persist membership in `cf_stage2`, so the upcoming reconcile snapshot has no stored row to scan for the most common cohort shape, and a cross partition person merge deletes `cf_stage2` rows that no leaf transition recreates on the survivor.

## Changes

First of three stacked PRs implementing the behavioral cohorts reconcile snapshot; this one materializes the scan domain with no downstream visible behavior change.

Adds a membership register: a new `registers_membership()` eligibility class covering single leaf and composable cohorts, with the live event, sweep, and seed paths writing the person's membership bit into `cf_stage2` atomically with their state commit, and the orphan GC keep predicate switched so it stops reclaiming single leaf rows.

Makes the merge protocol carry registers: cross partition transfers gain an additive `membership_registers` payload, receivers fill missing survivor rows as ownership marked fallbacks with a person first inventory key that survives stale or missing catalogs, and local evaluation later claims those rows. Emission is gated by a new `register_transfer_enabled` switch, hardwired off in production until every pod can apply the payload; a register owning merge sticky holds with a counter instead of silently losing rows. The next PR wires this gate to the reconcile config.

`cargo test` for `cohort-core` and `cohort-stream-processor`, clippy and fmt, all green. New tests assert the register write at every transition site (event fold, sweep evict, seed apply including the Unchanged replay, merge drain and apply), the transfer wire compatibility with old senders and receivers, the GC protection for transferred inventory, and the disabled gate hold. I did not run the ignored broker integration suite.

## Docs update

No.

### DIFF
--- a/rust/cohort-core/src/eligibility.rs
+++ b/rust/cohort-core/src/eligibility.rs
@@ -84,12 +84,12 @@ impl CohortEligibility {
         matches!(self, Self::Stage2Composable | Self::Stage2ComposableRef)
     }
 
-    /// Whether this class persists membership in `cf_stage2`. Single-leaf cohorts register their
-    /// leaf membership directly; composable cohorts persist the result of Boolean composition.
+    /// Whether this class persists membership in `cf_stage2`. Broader than [`Self::writes_cf_stage2`],
+    /// which covers the composable classes only: single-leaf cohorts also register a row here, holding
+    /// their own leaf's membership bit rather than a composition result.
     ///
-    /// Enumerated positively — the reconcile scan domain and the orphan-GC keep-predicate both read
-    /// this, so a new eligibility variant must opt in here rather than default into membership
-    /// registration and reconcile emission.
+    /// Enumerated positively so a new eligibility variant must opt in here rather than silently start
+    /// registering membership; the orphan-GC keep-predicate reads it.
     pub fn registers_membership(self) -> bool {
         matches!(
             self,
--- a/rust/cohort-core/src/eligibility.rs
+++ b/rust/cohort-core/src/eligibility.rs
@@ -83,6 +83,19 @@ impl CohortEligibility {
     pub fn writes_cf_stage2(self) -> bool {
         matches!(self, Self::Stage2Composable | Self::Stage2ComposableRef)
     }
+
+    /// Whether this class persists membership in `cf_stage2`. Single-leaf cohorts register their
+    /// leaf membership directly; composable cohorts persist the result of Boolean composition.
+    ///
+    /// Enumerated positively — the reconcile scan domain and the orphan-GC keep-predicate both read
+    /// this, so a new eligibility variant must opt in here rather than default into membership
+    /// registration and reconcile emission.
+    pub fn registers_membership(self) -> bool {
+        matches!(
+            self,
+            Self::SingleLeaf(_) | Self::Stage2Composable | Self::Stage2ComposableRef
+        )
+    }
 }
 
 /// Whether a tree's root condition is negated: AND requires all children negated, OR requires any.
@@ -904,6 +917,12 @@ mod tests {
             "a still-excluded ref cohort persists no cf_stage2 row",
         );
         assert!(!CohortEligibility::Excluded(ExcludedReason::CycleDetected).writes_cf_stage2());
+
+        assert!(CohortEligibility::Stage2Composable.registers_membership());
+        assert!(CohortEligibility::Stage2ComposableRef.registers_membership());
+        assert!(CohortEligibility::SingleLeaf(LeafStateKey(HASH_A)).registers_membership());
+        assert!(!CohortEligibility::Excluded(ExcludedReason::HasCohortRef).registers_membership(),);
+        assert!(!CohortEligibility::Excluded(ExcludedReason::CycleDetected).registers_membership(),);
     }
 
     #[test]
--- a/rust/cohort-stream-processor/src/config.rs
+++ b/rust/cohort-stream-processor/src/config.rs
@@ -243,6 +243,15 @@ pub struct Config {
     )]
     pub cohort_seed_watermark_idle_probe_interval_ms: u64,
 
+    /// Permit cross-partition membership-register transfer. Local register writers and transfer
+    /// receivers are always active. Default off for a receiver-first rollout; enable only after
+    /// every processor pod understands the additive transfer payload. Until enabled, a
+    /// cross-partition merge carrying register rows retains its source offset for retry, so this
+    /// must be on before the merge producer (`PERSON_MERGE_EVENTS_ENABLED`). The reconcile
+    /// snapshot rollout takes ownership of this gate when it lands.
+    #[envconfig(from = "COHORT_SEED_RECONCILE_ENABLED", default = "false")]
+    pub cohort_seed_reconcile_enabled: bool,
+
     /// Stable per-pod identity for `group.instance.id` + `client.id`, enabling static membership.
     /// Read from `POD_NAME`, else `HOSTNAME`. Absent means no static membership.
     #[envconfig(from = "POD_NAME")]
@@ -967,6 +976,7 @@ mod tests {
             kafka_seed_consumer_group: "cohort-stream-seeds".to_string(),
             cohort_seed_fence_margin_ms: 600_000,
             cohort_seed_watermark_idle_probe_interval_ms: 30_000,
+            cohort_seed_reconcile_enabled: false,
         }
     }
 
@@ -1116,6 +1126,22 @@ mod tests {
         assert!(config.store_config().wipe_on_start);
         config.wipe_store_on_start = false;
         assert!(!config.store_config().wipe_on_start);
+    }
+
+    #[test]
+    fn register_transfer_gate_defaults_dark_and_overrides_from_env() {
+        let defaults = Config::init_from_hashmap(&std::collections::HashMap::new()).unwrap();
+        assert!(
+            !defaults.cohort_seed_reconcile_enabled,
+            "register transfer must stay off until the whole fleet can apply the payload",
+        );
+
+        let enabled = Config::init_from_hashmap(&std::collections::HashMap::from([(
+            "COHORT_SEED_RECONCILE_ENABLED".to_owned(),
+            "true".to_owned(),
+        )]))
+        .unwrap();
+        assert!(enabled.cohort_seed_reconcile_enabled);
     }
 
     #[test]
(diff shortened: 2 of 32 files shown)


rust/cohort-stream-processor/src/workers/merge_path.rs
```diff
@@ -252,15 +263,30 @@ pub(crate) async fn handle_merge(
...
+            // Nothing to apply — skip the produce. The transfer type owns this predicate so drain
+            // staging and worker settlement cannot disagree about new payload classes.
+            if !transfer.has_payload() {
                 counter!(MERGE_TRANSFERS_SKIPPED_EMPTY_TOTAL).increment(1);
                 mark_processed(&merge.merge_tracker, partition_id, offset);
                 return;
             }
             produce_and_settle(partition_id, handle, merge, &transfer, &pending_key, offset).await;
         }
+        Ok(DrainOutcome::AwaitingRegisterTransferEnablement) => {
```
haacked: `blocking`: No test drives `handle_merge` through the `AwaitingRegisterTransferEnablement` arm at merge_path.rs:275. If someone swaps it with the nearby `AlreadyDrained` arm (which calls `mark_processed`), CI stays green while a cross-partition merge with a disabled transfer gate silently advances past the merge and drops the person's register rows.

[`disabled_register_transfer_retains_the_cross_partition_merge_for_retry`](https://github.com/PostHog/posthog/blob/eb51929f4af4281836f4e7c33f005ee1fc4e5731/rust/cohort-stream-processor/src/merge/drain_handler.rs#L678) in drain_handler.rs only asserts the `DrainOutcome` value, not what the worker does with it.

Add a test in merge_path.rs's `mod tests` that stages a register row for a cross-partition merge with `register_transfer_enabled: false`, calls `handle_merge`, and asserts `committable_offsets()` holds at the merge's own offset rather than advancing past it, mirroring [`an_apply_store_error_hold_is_not_leapfrogged_by_a_later_successful_transfer`](https://github.com/PostHog/posthog/blob/eb51929f4af4281836f4e7c33f005ee1fc4e5731/rust/cohort-stream-processor/src/workers/merge_path.rs#L1178).

rust/cohort-stream-processor/src/merge/drain_handler.rs
```diff
@@ -188,11 +302,17 @@ pub fn handle_merge_event(
...
             &old_prefix,
             &old_stage2_keys,
+            &old_transferred_register_keys,
+            &old_membership_registers,
             &present_leaves,
             person_dedup.as_ref(),
         );
     }
 
+    if !register_transfer_mode.is_enabled() && !old_membership_registers.is_empty() {
```
haacked: `suggestion`: veria-ai's earlier finding here checks out, and reaches further than reported: `old_membership_registers` at drain_handler.rs:312 covers composable-cohort rows already shipping on main, not just new single-leaf members. Any cross-partition merge of an existing composable-cohort member hits the hold. Local register writes are unconditional while only cross-partition transfers are gated, so writes land before the fleet can carry them, firing the hold the moment the merge producer is enabled for any allowlisted team, which is hard to square with "no downstream visible behavior change."

Gating local writes behind the same `register_transfer_enabled` flag eliminates the risk. With nothing written while transfers are gated, `AwaitingRegisterTransferEnablement` never fires, and flipping the flag turns on writes and transfers together. The reconcile PR doesn't read these rows until later anyway, so writing them later costs nothing.

If the current split stays, the deploy-ordering requirement documented in [config.rs](https://github.com/PostHog/posthog/blob/eb51929f4af4281836f4e7c33f005ee1fc4e5731/rust/cohort-stream-processor/src/config.rs#L246-L253) needs to be a hard runbook step, not just a comment, and `MERGE_TRANSFERS_HELD_AWAITING_ENABLEMENT_TOTAL` should page before any allowlisted team gets the merge producer.

rust/cohort-stream-processor/src/config.rs
```diff
@@ -243,6 +243,15 @@ pub struct Config {
     )]
     pub cohort_seed_watermark_idle_probe_interval_ms: u64,
 
+    /// Permit cross-partition membership-register transfer. Local register writers and transfer
```
haacked, line 246: `suggestion`: `cohort_seed_reconcile_enabled`/`COHORT_SEED_RECONCILE_ENABLED` is the only production gate for `register_transfer_enabled` (wired at [main.rs:237](https://github.com/PostHog/posthog/blob/eb51929f4af4281836f4e7c33f005ee1fc4e5731/rust/cohort-stream-processor/src/main.rs#L237)), but its name describes "seed reconcile," not "register transfer." Every other gate in this file (`cohort_cascade_enabled`, `cohort_seed_consumer_enabled`, `stage2_orphan_gc_enabled`) matches its purpose, so whoever greps for "register transfer" during an incident won't find this switch.

Rename it to `cohort_register_transfer_enabled` / `COHORT_REGISTER_TRANSFER_ENABLED` now; it's off by default either way, so the rename costs nothing operationally, and the reconcile PR can rename it again if it ends up owning a shared gate.

rust/cohort-core/src/eligibility.rs
```diff
@@ -83,6 +83,19 @@ impl CohortEligibility {
...
         matches!(self, Self::Stage2Composable | Self::Stage2ComposableRef)
     }
+
+    /// Whether this class persists membership in `cf_stage2`. Single-leaf cohorts register their
+    /// leaf membership directly; composable cohorts persist the result of Boolean composition.
+    ///
+    /// Enumerated positively — the reconcile scan domain and the orphan-GC keep-predicate both read
+    /// this, so a new eligibility variant must opt in here rather than default into membership
+    /// registration and reconcile emission.
+    pub fn registers_membership(self) -> bool {
```
haacked, line 93: `suggestion`: `registers_membership()` sits below `writes_cf_stage2()` with nearly identical doc lines, differing only in whether `SingleLeaf` is included. A maintainer looking for "does this cohort write a cf_stage2 row" will reasonably reach for `writes_cf_stage2()` and get it wrong for single-leaf cohorts, which do write a register row after this PR.

Name the contrast in the doc: "Broader than `writes_cf_stage2()`, which covers composable classes only; single-leaf cohorts also register a row here, holding their own leaf's membership bit rather than a composition result."

rust/cohort-stream-processor/src/merge/drain_handler.rs
```diff
@@ -146,14 +211,63 @@ pub fn handle_merge_event(
...
-    let old_stage2_keys: Vec<Stage2Key> = composable_cohort_ids(filters)
-        .map(|cohort_id| Stage2Key {
+    let transferred_inventory = store.scan_stage2_transferred_registers(
+        Stage2TransferredRegisterPersonPrefix::new(partition_id, team_u64, old_person),
+    )?;
+    let old_transferred_register_keys: Vec<Stage2TransferredRegisterKey> = transferred_inventory
+        .iter()
+        .map(|(key, _value)| *key)
+        .collect();
+    let mut registering_cohorts: BTreeMap<CohortId, DrainRegisterSource> =
```
haacked: `suggestion`: `handle_merge_event_with_transfer_mode` grew from ~138 to ~195 lines, with lines 221-259 (rebuilding `registering_cohorts` from catalog and transferred-inventory with three inline drop-counters) interrupting the function's linear flow. That reconstruction logic has nothing to do with routing.

Extract it to a `collect_registering_cohorts(filters, transferred_inventory)` helper so the main function reads as a sequence of named steps.

rust/cohort-stream-processor/src/stage2/register.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[cfg(test)]
+mod tests {
+    use chrono_tz::UTC;
+    use serde_json::json;
+
+    use super::*;
+    use crate::filters::{CohortId, TeamFiltersBuilder};
+
+    #[test]
```
haacked, line 94: `suggestion`: The test for `single_leaf_register_writes` only covers a leaf with multiple single-leaf cohorts. The zero case, a leaf with no single-leaf cohorts keyed on it, is untested, which is exactly where a no-op for composite-only leaves could slip through.

Add a case with no single-leaf cohort for the queried `leaf_state_key` and assert the result is empty.

rust/cohort-stream-processor/src/stage2/register.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub leaf_state_key: LeafStateKey,
+}
+
+/// Fan out one leaf's current membership to every single-leaf cohort backed by that leaf.
+pub(crate) fn single_leaf_register_writes(
+    filters: &TeamFilters,
+    source: MembershipRegisterSource,
+    in_cohort: bool,
+    last_evaluated_at_ms: i64,
+) -> Vec<MembershipRegisterWrite> {
```
haacked: `nit`: `single_leaf_register_writes` and `single_leaf_transition_register_writes` return `Vec<MembershipRegisterWrite>` only to be consumed immediately by a `for` loop on every event/seed/sweep/merge transition. Returning `impl Iterator<Item = MembershipRegisterWrite>` instead saves one allocation per transition on the hottest path.

rust/cohort-stream-processor/src/workers/event_path.rs
```diff
@@ -538,6 +539,16 @@ pub(crate) fn fold_event(
         write_stats.record_size = Some(bytes.len());
         staged.put::<PersonRecords>(&prefix.record_key(), &bytes);
     }
+    for transition in &transitions {
+        for write in single_leaf_transition_register_writes(
```
haacked: `nit`: The loop `for write in single_leaf_..._register_writes(...) { staged.put_stage2(&write.key, &write.state.encode()); }` is copied into three places (here, worker.rs, and seed_path.rs). A `StagedBatch::put_register_writes(&[MembershipRegisterWrite])` helper would collapse each to one call, so a future change to register-row encoding touches one line instead of three.

rust/cohort-stream-processor/src/main.rs
```diff
@@ -232,6 +232,9 @@ async fn async_main(config: Config) -> Result<()> {
         seed_tracker: Arc::new(OffsetTracker::new()),
         // Unconditional (cheap): the event path observes regardless of the seed gate.
         live_watermarks: Arc::new(LiveWatermarks::new()),
+        // Off until the whole fleet can apply the additive register payload; the reconcile
+        // snapshot's rollout gate takes ownership of this switch when it lands.
+        register_transfer_enabled: false,
```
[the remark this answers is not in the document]
matheus-vb: addressed in [eb51929](https://github.com/PostHog/posthog/pull/72726/commits/eb51929f4af4281836f4e7c33f005ee1fc4e5731)

haacked (approved):
Solid work overall. One blocking test gap and a few suggestions inline, including a follow-up on the earlier merge-partition finding.

PR #72728, opened 2026-07-21
feat(cohorts): reconcile snapshot drain path

matheus-vb:
## Problem

The streaming cohort pipeline emits membership at most once, so a lost produce leaves downstream stale while stored state is correct, and an edited cohort keeps its old members forever. Nothing re-emits the full current membership to heal that.

## Changes

Second of three stacked PRs; builds on the membership register PR and is gated behind `COHORT_SEED_RECONCILE_ENABLED` (default off), producing only to the shadow topic.

Implements the reconcile snapshot in the stream processor: a `reconcile` control tile decodes from the seed topic, is admitted fence open with newer run supersession, and holds its offset through a new releasable deferred floor on the offset tracker. A paced drain on the maintenance read lane prefix scans the partition's `cf_stage2` slice, re-evaluates every person from current leaf state, emits the full membership unconditionally tagged `origin: "reconcile"` with the `run_id`, fixes stale stored bits (cascading only composable fixes), and finishes with a per partition `reconcile_complete` marker on its own producer method, leaving the live wire contract byte identical. Dirty tracking re-scans rows mutated mid scan, drain time guards (catalog, eligibility, persisted `behavioral_filters_shape_hash`) discard superseded runs without a marker, and a monotonic output clock keeps later live flips strictly newer. Also wires the previous PR's register transfer gate to this config and adds the sweeper, config, and metrics.

`cargo test` for `cohort-core` and `cohort-stream-processor`, clippy and fmt, all green. New tests cover the deferred floor properties, the prefix scan bounds, the guard table, the drain state machine (produce failure retry, marker only retry, zero row cohort, supersession), dirty capture scoping, wire goldens for the tile and marker, and worker harness runs asserting snapshot then marker with the offset released only after completion. The broker integration suite is `#[ignore]`d and I did not run it end to end.

## Docs update

No.

### DIFF
--- a/rust/cohort-stream-processor/src/store/rocks.rs
+++ b/rust/cohort-stream-processor/src/store/rocks.rs
@@ -6,7 +6,7 @@
 #![allow(clippy::disallowed_methods)]
 
 use std::cell::Cell;
-use std::collections::HashMap;
+use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, RwLock};
@@ -26,6 +26,7 @@ use super::column_families::{self, Cf, OpaqueCf};
 use super::keys::{
     self, MergeAppliedKey, MergeDrainKey, PendingTransferKey, Stage2CohortPrefix, Stage2DirtyKey,
     Stage2Key, Stage2TransferredRegisterKey, Stage2TransferredRegisterPersonPrefix, TombstoneKey,
+    STAGE2_DIRTY_KEY_LEN,
 };
 use super::keyspace::{
     BehavioralKey, Keyspace, Meta, PersonPrefix, PersonRecordKey, META_SCHEMA_VERSION,
@@ -187,7 +188,10 @@ pub type RawKv = (Vec<u8>, Vec<u8>);
 
 #[derive(Debug, Default)]
 struct Stage2DirtyTracking {
-    active: RwLock<HashMap<Stage2CohortPrefix, usize>>,
+    /// The set of cohort prefixes with a reconcile drain currently capturing dirty rows. At most one
+    /// job per prefix ever holds a lease (admission supersedes same `(partition, team, cohort)`, and
+    /// the partition worker is serial), so a set — not a refcount — models the real invariant.
+    active: RwLock<HashSet<Stage2CohortPrefix>>,
     active_count: AtomicUsize,
 }
 
@@ -197,7 +201,11 @@ impl Stage2DirtyTracking {
             .active
             .write()
             .expect("Stage 2 dirty-tracking lock is not held across fallible work");
-        *active.entry(prefix).or_default() += 1;
+        let inserted = active.insert(prefix);
+        debug_assert!(
+            inserted,
+            "a cohort prefix cannot hold two dirty-tracking leases"
+        );
         self.active_count.fetch_add(1, Ordering::Release);
         Stage2DirtyTrackingGuard {
             tracking: self.clone(),
@@ -205,31 +213,38 @@ impl Stage2DirtyTracking {
         }
     }
 
+    /// Cheap pre-check for the hot write path: whether any drain is capturing at all. A `false` lets
+    /// a reconcile-disabled pipeline skip decoding the Stage 2 key entirely.
+    fn any_active(&self) -> bool {
+        self.active_count.load(Ordering::Acquire) != 0
+    }
+
     fn is_active(&self, prefix: Stage2CohortPrefix) -> bool {
-        if self.active_count.load(Ordering::Acquire) == 0 {
+        if !self.any_active() {
             return false;
         }
         self.active
             .read()
             .expect("Stage 2 dirty-tracking lock is not held across fallible work")
-            .contains_key(&prefix)
+            .contains(&prefix)
+    }
+
+    /// The dirty marker to stage for a `cf_stage2` write, or `None` when no drain tracks its cohort.
+    /// The single home of the "is this cohort captured / which marker do we write" decision, shared by
+    /// the live-event [`BatchBuilder`] path and the owned [`StagedBatch`] apply path so the rule for
+    /// emitting a marker cannot drift between them.
+    fn dirty_marker(&self, key: &Stage2Key) -> Option<[u8; STAGE2_DIRTY_KEY_LEN]> {
+        self.is_active(key.cohort_prefix())
+            .then(|| Stage2DirtyKey::new(*key).encode())
     }
 
     fn release(&self, prefix: Stage2CohortPrefix) {
-        let mut active = self
+        let removed = self
             .active
             .write()
-            .expect("Stage 2 dirty-tracking lock is not held across fallible work");
-        let remove = {
-            let count = active
-                .get_mut(&prefix)
-                .expect("every dirty-tracking guard owns one active reference");
-            *count -= 1;
-            *count == 0
-        };
-        if remove {
-            active.remove(&prefix);
-        }
+            .expect("Stage 2 dirty-tracking lock is not held across fallible work")
+            .remove(&prefix);
+        debug_assert!(removed, "every dirty-tracking guard owns one active lease");
         let previous = self.active_count.fetch_sub(1, Ordering::Release);
         debug_assert!(previous > 0, "dirty-tracking reference count underflow");
     }
@@ -782,14 +797,18 @@ impl CohortStore {
         if !matches!(cf, Cf::Stage2) {
             return;
         }
+        // Skip decoding the key on the hot write path when no reconcile drain is capturing dirties.
+        if !self.dirty_tracking.any_active() {
+            return;
+        }
         let Ok(key) = Stage2Key::decode(encoded_key) else {
             return;
         };
-        if self.dirty_tracking.is_active(key.cohort_prefix()) {
+        if let Some(marker) = self.dirty_tracking.dirty_marker(&key) {
             batch.put_cf(
                 self.cf(Cf::Stage2)
                     .expect("the Stage 2 CF was opened before applying a batch"),
-                Stage2DirtyKey::new(key).encode(),
+                marker,
                 [],
             );
         }
@@ -1324,9 +1343,8 @@ impl<'db> BatchBuilder<'db> {
     }
 
     fn mark_stage2_dirty(&mut self, key: &Stage2Key) {
-        if self.dirty_tracking.is_active(key.cohort_prefix()) {
-            self.batch
-                .put_cf(self.stage2, Stage2DirtyKey::new(*key).encode(), []);
+        if let Some(marker) = self.dirty_tracking.dirty_marker(key) {
+            self.batch.put_cf(self.stage2, marker, []);
         }
     }
 }
--- a/rust/cohort-stream-processor/src/store/rocks.rs
+++ b/rust/cohort-stream-processor/src/store/rocks.rs
@@ -6,8 +6,10 @@
 #![allow(clippy::disallowed_methods)]
 
 use std::cell::Cell;
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, RwLock};
 use std::time::Instant;
 
 use metrics::{counter, histogram};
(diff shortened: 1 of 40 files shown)


rust/cohort-stream-processor/src/workers/reconcile.rs
```diff
@@ -0,0 +1,1710 @@
...
+        return ReconcileGuard::Discard(ReconcileDiscardReason::CohortAbsent);
+    }
+    let Some(eligibility) = filters.eligibility.get(&tile.cohort_id()) else {
+        return ReconcileGuard::Discard(ReconcileDiscardReason::CohortAbsent);
+    };
+    if !eligibility.registers_membership() {
+        return ReconcileGuard::Discard(ReconcileDiscardReason::NotEmitting);
+    }
+    let Some(actual_hash) = filters.behavioral_shape_hashes.get(&tile.cohort_id()) else {
+        return ReconcileGuard::Discard(ReconcileDiscardReason::HashUnknown);
```
haacked, line 305: `question`: A realtime cohort with no behavioral leaf never gets reconciled. `evaluate_guard` checks `eligibility.registers_membership()` first (always true for person-only cohorts), then requires an entry in `filters.behavioral_shape_hashes`, discarding as `HashUnknown` if absent (reconcile.rs:293-301). The loader deliberately omits person-only cohorts from `behavioral_shape_hashes` because an empty `behavioral_filters_shape_hash` means no behavioral leaves ([loader.rs:65-69](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-core/src/filters/loader.rs#L65-L69)). So a reconcile tile for a person-only cohort always hits `HashUnknown` and is discarded without a completion marker, losing the heal-a-lost-produce / re-emit-after-edit guarantee this PR exists to provide.

The tile producer isn't in this PR, so it's unclear whether it emits person-only cohort tiles or intentionally skips them. If it emits them, they get silently dropped and are only observable via `RECONCILE_JOBS_DISCARDED_TOTAL{reason="hash_unknown"}`. Which is it? If person-only cohorts are in scope, the guard needs a path that proceeds when the cohort legitimately has no behavioral shape rather than treating an absent hash as a stale-run signal.

rust/cohort-stream-processor/src/workers/reconcile.rs
```diff
@@ -0,0 +1,1710 @@
...
+            1
+        );
+        assert!(store
+            .scan_stage2_dirty(waiting_prefix, None, 10)
+            .unwrap()
+            .is_empty());
+    }
+
+    #[test]
+    fn supersede_matches_team_and_cohort_and_preserves_other_jobs() {
```
haacked, line 1293: `suggestion`: Nothing covers supersession of a job that's already mid-drain. `supersede_matches_team_and_cohort_and_preserves_other_jobs` and `older_or_duplicate_replay_never_evicts_the_queued_run` both start from a queue head that was never ticked (`dirty_tracking: None`, `phase: Scanning { cursor: None }`). In production, a large cohort edited twice while its first reconcile is still draining hits `supersede_if_newer` on a head that already emitted pages under run A, holds an active `Stage2DirtyTrackingGuard`, and sits at `Scanning { cursor: Some(_) }`. If dropping that job ever failed to release its deferred offset, or left its partial run-A emission looking complete, a stuck committable floor or a phantom run would ship untested.

Add a drain test that mirrors the two supersede tests above but ticks first: write enough rows that the drain spans multiple pages (small `scan_page`), enqueue run A at offset 5, `tick()` once so the head is mid-scan with an active lease, then supersede with run B at offset 7 (`supersede_if_newer` + `replace_deferred` + `enqueue`, as `admit_reconcile` does). Drive to completion and assert only run B's marker is emitted (no marker carries run A), the full membership re-emits under run B's `run_id`, and `committable()` reaches `Some(8)`.

rust/cohort-stream-processor/src/workers/reconcile.rs
```diff
@@ -0,0 +1,1710 @@
...
+
+        match phase {
+            ScanPhase::Scanning { cursor } => {
+                let page = match handle
+                    .scan_stage2_cohort(prefix, cursor, merge.reconcile.scan_page)
+                    .await
+                {
+                    Ok(page) => page,
+                    Err(error) => {
+                        warn!(
```
haacked: `suggestion`: The job-identity fields (`partition_id`, `team_id = tile.team_id().0`, `cohort_id = tile.cohort_id().0`, `run_id = %tile.run_id().0`) are logged nine times across `handle_reconcile_drain` and `settle_reconcile_page`, at [462-469](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L462-L469), [495-502](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L495-L502), [552-559](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L552-L559), [582-589](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L582-L589), [604-609](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L604-L609), [630-637](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L630-L637), [710-718](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L710-L718), and [761-768](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/workers/reconcile.rs#L761-L768). Adding or renaming a field means updating all nine sites.

Pull the identity fields into a small local macro so each call site carries only the message and any extra fields:

```rust
macro_rules! warn_job {
    ($tile:expr, $($rest:tt)*) => {
        warn!(
            team_id = $tile.team_id().0,
            cohort_id = $tile.cohort_id().0,
            run_id = %$tile.run_id().0,
            $($rest)*
        )
    };
}
```

Call sites become `warn_job!(tile, partition_id, error = %error, "reconcile Stage 2 scan failed; retrying this page on the next tick");`.

rust/cohort-stream-processor/src/workers/reconcile.rs
```diff
@@ -0,0 +1,1710 @@
...
+        return ReconcileGuard::Discard(ReconcileDiscardReason::HashMismatch);
+    }
+    ReconcileGuard::Proceed
+}
+
+/// Advance at most one scan page for the queue head. Guard-discarded jobs are drained in the same
+/// tick; a successfully completed job stops the tick so the next queued snapshot gets its own page
+/// budget.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn handle_reconcile_drain(
```
haacked, line 317: `suggestion`: `handle_reconcile_drain` is ~356 lines with a `loop` wrapping three phases. The `DrainingDirty` arm alone runs ~140 lines with four scan-error branches, two verification passes, and repeated `queue.front_mut().expect(...)` re-fetches. Changing one phase requires reasoning through all three and the loop's continue/return contract to avoid breaking the page-budget invariant.

Extract one async helper per phase so the driver reads as the state machine it is, each returning a small control signal instead of bare `continue`/`return`:

```rust
enum Step { Advanced, Continue } // Continue = re-enter the loop, Advanced = yield the tick

match phase {
    ScanPhase::Scanning { cursor } => match drain_scanning(partition_id, handle, sink, merge, queue, &tile, ...).await {
        Step::Continue => continue,
        Step::Advanced => return,
    },
    ScanPhase::DrainingDirty { cursor } => ...,
    ScanPhase::MarkerReady => ...,
}
```

The guard/`activate_front_tracking` preamble stays in the driver; each helper owns its own scan, settle, and phase transition. The three phases already share nothing but `queue`, `tile`, and `prefix`, so the borrow surface stays small.

rust/cohort-stream-processor/src/store/rocks.rs
```diff
@@ -560,15 +767,34 @@ impl CohortStore {
...
                 StagedOp::Delete { cf, key } => {
                     batch.delete_cf(self.cf(*cf)?, key);
+                    self.mark_staged_stage2_dirty(&mut batch, *cf, key);
                 }
             }
         }
         self.commit(batch, OP_WRITE_BATCH)
     }
 
+    fn mark_staged_stage2_dirty(&self, batch: &mut WriteBatch, cf: Cf, encoded_key: &[u8]) {
```
haacked, line 796: `suggestion`: The rule for when a `cf_stage2` write also emits a dirty marker lives in two places that have to stay identical: `mark_staged_stage2_dirty` (rocks.rs:781-793, the `StagedBatch::apply`/reconcile path) and `BatchBuilder::mark_stage2_dirty` ([rocks.rs:1326-1331](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/store/rocks.rs#L1326-L1331), the live event path). Both encode the same decision: if `dirty_tracking.is_active(key.cohort_prefix())`, put a `Stage2DirtyKey` marker. If a future `cf_stage2` write path gets wired through one but the marker rule drifts in the other, a concurrent reconcile drain misses the mutated row, silently reopening the stale-membership bug this PR exists to close.

Pull the shared decision into one helper both call:

```rust
fn stage2_dirty_marker(
    dirty_tracking: &Stage2DirtyTracking,
    key: &Stage2Key,
) -> Option<[u8; STAGE2_DIRTY_KEY_LEN]> {
    dirty_tracking
        .is_active(key.cohort_prefix())
        .then(|| Stage2DirtyKey::new(*key).encode())
}
```

Then `mark_stage2_dirty` and `mark_staged_stage2_dirty` each just do the CF put with the returned bytes, and the "is it active / what key do we write" logic exists in exactly one place.

haacked, line 796: `suggestion`: On the live event path (`event_path.rs`/`stage2_path.rs` → `handle.commit` → `store.apply` → `mark_staged_stage2_dirty`), every `cf_stage2` write decodes the key with `Stage2Key::decode` before checking whether any drain is active, so a reconcile-disabled pipeline pays that decode on every stage2 write. The `BatchBuilder` path avoids this because it already holds the decoded key.

Gate the decode behind the same cheap atomic the active check uses:

```rust
fn mark_staged_stage2_dirty(&self, batch: &mut WriteBatch, cf: Cf, encoded_key: &[u8]) {
    if !matches!(cf, Cf::Stage2) {
        return;
    }
    if !self.dirty_tracking.any_active() {
        return;
    }
    let Ok(key) = Stage2Key::decode(encoded_key) else {
        return;
    };
    // ... unchanged
}
```

with a small accessor: `fn any_active(&self) -> bool { self.active_count.load(Ordering::Acquire) != 0 }`. The decode itself is cheap (~34 branch-free byte reads, no allocation), so the saving is small; the value is that a disabled reconcile then costs one atomic load and nothing else on the hot write path.

rust/cohort-stream-processor/src/store/rocks.rs
```diff
@@ -175,6 +185,74 @@ pub enum StoreError {
 /// One scanned key/value pair as raw bytes — the merge-CF GC decodes each per CF.
 pub type RawKv = (Vec<u8>, Vec<u8>);
 
+#[derive(Debug, Default)]
+struct Stage2DirtyTracking {
+    active: RwLock<HashMap<Stage2CohortPrefix, usize>>,
```
haacked: `question`: `Stage2DirtyTracking.active` is a `HashMap<Stage2CohortPrefix, usize>` refcount, but `activate_front_tracking` only ever leases the queue head when its `dirty_tracking` is still `None`. Two jobs for the same `(partition, team, cohort)` prefix can't coexist because `supersede_if_newer` removes the older one and the worker is serial per partition, so the per-prefix count never exceeds 1. Is the refcount needed for a path this doesn't show, or would a `HashSet<Stage2CohortPrefix>` (still backed by the `active_count` atomic for the zero-cost `is_active` check) express the actual invariant more directly?

If a `HashSet` works, `release`'s decrement-then-maybe-remove dance ([rocks.rs:223-232](https://github.com/PostHog/posthog/blob/cec9f5f38242dc1e367e6f70061a1bbaee1dc058/rust/cohort-stream-processor/src/store/rocks.rs#L223-L232)) collapses to a single `remove`.

rust/cohort-stream-processor/src/workers/reconcile.rs
```diff
@@ -0,0 +1,1710 @@
...
+                }
+                let marker = ReconcileCompleteMarker::new(
+                    tile.team_id(),
+                    tile.cohort_id(),
+                    partition_id,
+                    tile.run_id(),
+                    last_updated.to_string(),
+                );
+                let acks = sink.produce_markers(vec![marker]).await;
+                let marker_errors =
```
haacked: `nit`: The marker-error count is hard to parse: `acks.iter().filter(|ack| ack.is_err()).count() + acks.len().abs_diff(1)`. The `abs_diff(1)` term handles the case where the sink returned something other than exactly one ack for the marker, but nothing says so.

Since one marker went in, spell the expectation out: `let marker_failed = acks.len() != 1 || acks.iter().any(Result::is_err);` (a wrong ack count is itself a failure), and log `marker_failed` instead, or keep the count but comment the `abs_diff(1)`.

haacked (approved):
Solid design, gated safely behind a flag with the live wire path staying byte-identical. A few suggestions and one open question inline.

rust/cohort-stream-processor/src/workers/reconcile.rs
matheus-vb: Intentional and by design. Reconcile only runs for behavioral backfill runs, and the loader deliberately omits cohorts with no behavioral leaves from `behavioral_shape_hashes`. So an absent hash is the fail closed skip for a non behavioral cohort or a superseded run, not a lost guarantee; purely person property cohorts heal through their own backfill path. Added a comment at the guard making that explicit.

rust/cohort-stream-processor/src/store/rocks.rs
matheus-vb: Switched to `HashSet<Stage2CohortPrefix>`. The `active_count` atomic stays for the zero cost `is_active` fast path, release collapses to a single remove, and acquire/release now `debug_assert` the one lease per prefix invariant so a future violation is loud instead of silently miscounted.

PR #72730, opened 2026-07-21
feat(cohorts): reconcile_dispatch CLI and parity harness

matheus-vb:
## Problem

The processor can now drain reconcile tiles, but nothing produces them, and the parity harness cannot tell whether a reconcile run's snapshot arrived complete.

## Changes

Third of three stacked PRs, completing the reconcile snapshot slice.

Adds a manual `reconcile_dispatch` binary to `cohort-seeder`: given a run id, it loads the run's active cohorts with their persisted `behavioral_filters_shape_hash`, verifies the topic's partition count, and produces one partition targeted reconcile tile to each of the 64 partitions per cohort, awaiting all acks and printing per partition produced offsets (the future reconcile HWMs). Any nack exits non zero and re dispatch is safe by supersession.

Teaches the parity harness to ingest snapshots: the fold recognizes `reconcile_complete` markers before the malformed check, records origin and run id on folded rows, and the classifier and report surface per run completeness (64 of 64 partitions or partial) as a note without changing any verdict rules.

`cargo test` for `cohort-seeder` including the Postgres backed run and chunk queries, clippy and fmt, all green. The python parity suite (100 tests) passes with new cases for marker recognition, origin folding, completeness counting, and the report note; ruff check and format are clean.

## Docs update

No.


products/cohorts/backend/parity/fold.py
```diff
@@ -84,13 +152,30 @@ def fold_membership_changes(
...
 
+def reconcile_completeness(stats: FoldStats, cohort_id: int) -> tuple[ReconcileRunCompleteness, ...]:
+    """Return deterministic per-run marker completeness for one cohort."""
+    return tuple(
+        ReconcileRunCompleteness(
+            run_id=run_id,
+            cohort_id=marker_cohort_id,
+            partitions_seen=len(partitions),
+        )
+        for (run_id, marker_cohort_id), partitions in sorted(stats.reconcile_markers.items())
```
haacked: `suggestion`: `reconcile_completeness` re-sorts and re-scans the entire marker dict on every call, and [`compare_cohort_membership.py:211`](https://github.com/PostHog/posthog/blob/3404bcc8608c6c4deacb525f0056445ab49325f0/products/cohorts/backend/management/commands/compare_cohort_membership.py#L211) calls it once per cohort inside the classify loop. At fold.py:174, `sorted(stats.reconcile_markers.items())` walks all M distinct (run_id, cohort_id) marker keys each time, so the loop is O(C log C * M) for C cohorts. With one active reconcile run M is close to C, making it roughly O(C² log C). This is an offline management command with tens to low hundreds of cohorts, so the cost stays well under a second.

Pre-group the markers by cohort once before the loop.

```python
completeness_by_cohort: dict[int, list[ReconcileRunCompleteness]] = defaultdict(list)
for (run_id, marker_cohort_id), partitions in sorted(fold_stats.reconcile_markers.items()):
    completeness_by_cohort[marker_cohort_id].append(
        ReconcileRunCompleteness(run_id=run_id, cohort_id=marker_cohort_id, partitions_seen=len(partitions))
    )
```

Then in the loop use `format_reconcile_notes(completeness_by_cohort.get(cid, ()))`, which turns the whole thing into one O(M log M) sort.

products/cohorts/backend/parity/fold.py
```diff
@@ -57,13 +98,40 @@ def fold_membership_changes(
...
+        if not isinstance(message_team_id, int) or isinstance(message_team_id, bool):
+            stats.dropped_malformed += 1
+            continue
+        if message_team_id != team_id:
             stats.dropped_wrong_team += 1
             continue
         cohort_id = message.get("cohort_id")
+        last_updated = parse_last_updated(message.get("last_updated"))
+
+        if message.get("type") == RECONCILE_COMPLETE_TYPE:
```
haacked, line 139: `suggestion`: `fold_membership_changes` mixes three concerns in one loop: team-id gating, reconcile-marker recording, and membership last-write-wins. The marker block (~20 lines of `if message.get("type") == RECONCILE_COMPLETE_TYPE:` handling) interrupts readers looking for the membership logic.

Pull the marker code into a helper that mirrors `_parse_marker_run_id`:

```python
def _record_marker(message, cohort_id, last_updated, since, stats) -> None:
    run_id = _parse_marker_run_id(message.get("run_id"))
    partition = message.get("partition")
    if (
        run_id is None
        or not isinstance(cohort_id, int)
        or isinstance(cohort_id, bool)
        or not isinstance(partition, int)
        or isinstance(partition, bool)
        or not 0 <= partition < RECONCILE_PARTITION_COUNT
        or last_updated is None
    ):
        stats.dropped_malformed += 1
        return
    if last_updated < since:
        stats.dropped_before_since += 1
        return
    stats.reconcile_markers.setdefault((run_id, cohort_id), set()).add(partition)
    stats.reconcile_markers_recorded += 1
    stats.cohorts_seen.add(cohort_id)
```

The loop becomes `if message.get("type") == RECONCILE_COMPLETE_TYPE: _record_marker(...); continue`. Since the membership validation is already inlined the same way, this is a readability refactor, not a correctness issue.

products/cohorts/backend/parity/test/test_fold.py
```diff
@@ -86,6 +113,79 @@ def test_wrong_team_messages_are_dropped(self) -> None:
...
+            since=SINCE,
+        )
+
+        self.assertEqual(stats.folded_by_origin, {LIVE_ORIGIN: 1, "seed": 1, "reconcile": 1})
+        self.assertEqual(
+            (state[10]["p1"].status, state[10]["p1"].origin, state[10]["p1"].run_id), ("left", "reconcile", RUN_2)
+        )
+        self.assertEqual((state[10]["p2"].origin, state[10]["p2"].run_id), ("seed", RUN_1))
+
+    def test_reconcile_completeness_is_per_run_and_counts_distinct_partitions(self) -> None:
```
haacked, line 155: `suggestion`: The per-cohort filter in `reconcile_completeness` (`if marker_cohort_id == cohort_id`, [fold.py:175](https://github.com/PostHog/posthog/blob/3404bcc8608c6c4deacb525f0056445ab49325f0/products/cohorts/backend/parity/fold.py#L175)) is never exercised. `test_reconcile_completeness_is_per_run_and_counts_distinct_partitions` only folds and queries cohort 10, so the filter never rejects anything.

If that filter regressed and returned every cohort's runs, the test would still pass. Since these become per-cohort report notes in `compare_cohort_membership.py`, each cohort would show unrelated runs' completeness. Add a second cohort's markers to the test and assert `reconcile_completeness(stats, 10)` leaves them out.

products/cohorts/backend/parity/test/test_fold.py
```diff
@@ -86,6 +113,79 @@ def test_wrong_team_messages_are_dropped(self) -> None:
...
+    @parameterized.expand(
+        [
+            ("bad_run_id", _marker(1, run_id="not-a-uuid"), 2, 1, 0),
+            ("out_of_range_partition", _marker(64), 2, 1, 0),
+            ("float_team_id", _marker(1) | {"team_id": 2.0}, 2, 1, 0),
+            ("boolean_team_id", _marker(1, team_id=True), 1, 1, 0),
+            ("before_since", _marker(1, ts="2026-07-07 18:59:59.999999"), 2, 0, 1),
+        ]
+    )
+    def test_invalid_or_stale_markers_cannot_certify_reconcile(
```
haacked, line 192: `suggestion`: The marker-rejection table in `test_invalid_or_stale_markers_cannot_certify_reconcile` exercises only two of five drop triggers (bad run_id, out-of-range partition); `float_team_id`/`boolean_team_id` are caught by the top-level guard instead. A marker missing `partition` (None) or with an unparseable `last_updated` is currently dropped as malformed, but it's untested.

A regression here could add a bogus partition to the completeness set, push a run to a false 64/64, and mask missing snapshot data. Add cases for a partition-less and a timestamp-less marker to the parameterized list.

rust/cohort-seeder/src/app/reconcile_dispatch.rs
```diff
@@ -0,0 +1,636 @@
...
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ReconcileDispatchError {
+    #[error(transparent)]
+    InvalidPartitionCount(#[from] SeedPartitionCountError),
+    #[error("joining the seed-topic partition verification task")]
+    PartitionVerificationTask(#[source] tokio::task::JoinError),
```
haacked, line 313: `nit`: `InvalidPartitionCount` fires when building the `0..COHORT_PARTITION_COUNT` set (a wrong constant, essentially never at runtime), while `PartitionCount` fires when the live Kafka topic's partition count doesn't match, making them easy to confuse. Rename `PartitionCount` to `TopicPartitionCount` so the difference is obvious at the call site.

rust/cohort-seeder/src/store/runs.rs
```diff
@@ -361,18 +438,20 @@ mod tests {
...
-    /// `cohort_backfill_run_cohorts` for a future reconcile slice. Populated by Q1 (whose SQL stays
-    /// frozen) but not yet read, so the attr suppresses the dead-field lint until that slice lands.
-    #[allow(dead_code)]
-    filters_shape_hash: String,
     pinned_filters: Json<Value>,
     superseded_at: Option<DateTime<Utc>>,
 }
 
+#[derive(Debug, FromRow)]
+struct ParticipationRow {
```
haacked: `nit`: This row struct is reconcile-specific (it backs `READ_ACTIVE_RECONCILE_PARTICIPATIONS` and reads `behavioral_filters_shape_hash`), but its sibling in the same file is named `PinnedParticipationRow`. Rename it to `ReconcileParticipationRow` to keep them consistent and self-describing.

haacked (approved):
Looks good, nothing blocking! A few suggestions and nits inline.

products/cohorts/backend/parity/fold.py
matheus-vb: Moved the grouping into a `reconcile_completeness_by_cohort()` helper and call it once before the classify loop, so the command is a single `O(M log M)` sort plus dict lookups. Kept `reconcile_completeness()` as a thin delegate over it instead of inlining a defaultdict in the command, so the `ReconcileRunCompleteness` construction lives in one place and the per cohort accessor the tests use keeps working.

products/cohorts/backend/parity/test/test_fold.py
matheus-vb: Added a second cohort sharing RUN_1 over exactly the partitions cohort 10 is missing (41 to 63). If the grouping stopped keying by cohort, cohort 10’s RUN_1 would read a false 64/64 and the assertion fails, which makes the separation the thing under test. Also asserting cohort 20’s own run stays at 23/64.

products/cohorts/backend/parity/test/test_fold.py
matheus-vb: Added missing_partition and missing_timestamp as suggested. Also added missing_run_id, which hits the non string run_id guard and incidentally closes the patch coverage gap flagged on that branch.

PR #72738, opened 2026-07-21
feat(error-tracking): add temporal weekly digest workflow

ablaszkiewicz:
> [!WARNING]  
> This doesn't yet switch celery to temporal. It just copied what celery did and put it into temporal but it's not live yet. I'll run dry-run tests on prod to see how this performs

## Problem

The error tracking weekly digest runs on celery beat. We don't have a great visibility over what's going on on an org level. For example team 2 didn't receive the digest last Monday.

We're moving it to Temporal for durable runs and per-org retry visibility. This is step 1: add the Temporal path with zero behavior change. Cutover (register schedule, delete celery tasks) ships separately.

### Multiple workflows spawned as continue-as-new

CleanShot 2026-07-22 at 12 58 36

### Activities within single workflow

CleanShot 2026-07-22 at 13 03 03

### Webhooks getting delivered

CleanShot 2026-07-22 at 13 03 26

New `products/error_tracking/backend/temporal/weekly_digest/`:

- `types.py` – frozen dataclass inputs/results. `dry_run` defaults to True, so an input-less manual run can never send for real; only the schedule passes `dry_run=False`.
- `activities.py` – `get_digest_orgs_activity` (discovery, or explicit `org_ids` for targeted runs) and `send_org_digest_activity` (the celery per-org task ported as-is: digest builds, permission/notification filtering, MessagingRecord dedupe, webhook POST).
- `workflow.py` – one keyset page (`page_size=1000`) per execution: fetch page after the carried cursor, fan out one activity per org under a `Semaphore(max_concurrent)`, then `continue_as_new` with the page's last org id as cursor. History and payload stay bounded at any org count. Final execution fails the run if any org failed.
- `schedule.py` – Mondays 08:30 UTC, overlap SKIP, 6h catchup. Written but deliberately **not** registered yet.

Design points:

1. **Retry accounting** – celery's 0-based `retries` with `max_retries=5` maps to Temporal `maximum_attempts=6`; `max_attempts` travels in the activity input (activities can't read their retry policy) and `attempt >= max_attempts` marks the final attempt. Deferral / final-attempt partial send / retry behavior match celery exactly.
2. **Per-org activities + keyset paging, not WA-style batching** – the WA digest batches 25 orgs per activity, but it has no per-org retry semantics. Here the attempt counter drives deferral, so 1 org = 1 activity; boundedness comes from continue-as-new paging instead. Only a ~40-byte cursor (last org id of the page) rides between executions – carrying the full remaining-org-id list would approach 1 MB encrypted at ~20k orgs, over half Temporal's 2 MiB payload cap. Discovery re-runs per page with the cursor; sorted keyset slicing means an org shifting in or out between pages can never be processed twice, and MessagingRecord dedupe backstops any edge.

21 tests, all passing locally:

- 6 delivery-semantics tests ported from the celery suite, run against the activity with explicit attempt numbers (dedupe on retry, deferral, final-attempt partial send, etc.) – the celery tests can't cover this separate code path.
- `dry_run` test: no POST, no MessagingRecord stamp.
- 7 discovery tests: `org_ids` bypass, cloud discovery, self-hosted no-op, and 4 keyset page-bound cases (exclusive cursor, unsorted input, cursor between ids, past-the-end) – guards the off-by-one that would double-send a boundary org.
- 5 workflow tests (time-skipping env, stub activities): no-orgs, fan-out plumbing, input-less run falls back to dry-run defaults, multi-page run pages by cursor and processes every org exactly once, early-page failure on an exact page-size multiple still drains the chain and fails the final execution.
- 2 schedule tests: spec, queue, overlap, and `dry_run=False` locked in.

Also repo-wide mypy (only a pre-existing unrelated error), `bin/hogli lint:workflows`, `tach check`, `hogli ci:preflight --fix`. Not tested by me (Claude): a real prod Temporal run – that's the explicit next migration step.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior changes.

## 🤖 Agent context

Authored with Claude Code (Fable 5) from a reviewed migration plan. Skills: /writing-tests, /authoring-ci-workflows. Key decisions: reuse `error-tracking-task-queue`; one activity per org to keep celery's per-org retry contract; schedule written but unregistered so this PR is zero-behavior-change. Course corrections: env allowlists dropped for an `is_cloud()` gate, `dry_run` flipped to default True, unbounded fan-out fixed with continue-as-new paging.

### DIFF
--- a/.github/workflows/container-images-cd.yml
+++ b/.github/workflows/container-images-cd.yml
@@ -506,8 +506,12 @@ jobs:
                   filters: |
                       changed:
                         - 'products/error_tracking/backend/temporal/**'
-                        - 'products/error_tracking/backend/recommendations/**'
+                        - 'products/error_tracking/backend/logic/**'
                         - 'products/error_tracking/backend/models.py'
+                        - 'products/error_tracking/backend/weekly_digest.py'
+                        - 'products/error_tracking/backend/indexed_embedding.py'
+                        - 'posthog/tasks/email.py'
+                        - 'posthog/tasks/email_utils.py'
                         - 'posthog/temporal/common/**'
                         - 'posthog/temporal/schedule.py'
                         - 'posthog/settings/temporal.py'
--- a/products/error_tracking/backend/temporal/weekly_digest/activities.py
+++ b/products/error_tracking/backend/temporal/weekly_digest/activities.py
@@ -1,0 +1,299 @@
+from django.conf import settings
+from django.db import close_old_connections
+from django.utils import timezone
+
+import structlog
+from temporalio import activity
+from temporalio.exceptions import ApplicationError
+
+from posthog.cloud_utils import is_cloud
+from posthog.models import Organization, OrganizationMembership, Team
+from posthog.models.messaging import MessagingRecord
+from posthog.tasks.email import NotificationSetting, should_send_notification
+from posthog.temporal.common.heartbeat_sync import HeartbeaterSync
+from posthog.user_permissions import UserPermissions
+
+from products.error_tracking.backend import weekly_digest
+from products.error_tracking.backend.temporal.weekly_digest.types import (
+    GetDigestOrgsInputs,
+    SendOrgDigestInputs,
+    SendOrgDigestResult,
+)
+
+logger = structlog.get_logger(__name__)
+
+
+def _get_digest_orgs(inputs: GetDigestOrgsInputs) -> list[str]:
+    """Resolve one keyset page of org ids for this run.
+
+    An explicit ``org_ids`` (manual targeted runs) bypasses discovery; scheduled runs
+    discover every org with exceptions.
+
+    The candidate set is sorted and sliced to (``after``, ``after`` + ``limit``] so the
+    workflow can page with only a cursor. Recomputing candidates each page is safe: an
+    org shifting in or out between pages is either <= cursor (already processed) or
+    > cursor (still pending), never both.
+    """
+    # Checked before org_ids so a targeted manual run can't route around the cloud gate. Returning
+    # no orgs is what keeps the send activity from being scheduled at all.
+    if not is_cloud():
+        logger.info("Skipping Error Tracking weekly digest outside PostHog Cloud")
+        return []
+
+    if inputs.org_ids is not None:
+        candidates = sorted(str(org_id) for org_id in inputs.org_ids)
+    else:
+        candidates = sorted(str(org_id) for org_id in weekly_digest.get_org_ids_with_exceptions())
+
+    if inputs.after is not None:
+        candidates = [org_id for org_id in candidates if org_id > inputs.after]
+    return candidates[: inputs.limit]
+
+
+@activity.defn
+def get_digest_orgs_activity(inputs: GetDigestOrgsInputs) -> list[str]:
+    close_old_connections()
+    return _get_digest_orgs(inputs)
+
+
+def _send_org_digest(inputs: SendOrgDigestInputs, attempt: int) -> SendOrgDigestResult:
+    """Send one combined weekly error tracking digest per user in an org via the delivery workflow.
+
+    ``attempt`` is Temporal's 1-based attempt counter; ``attempt >= inputs.max_attempts`` marks
+    the final attempt, which sends partial digests instead of deferring recipients.
+    """
+    org_id = inputs.org_id
+    try:
+        org = Organization.objects.get(id=org_id)
+    except Organization.DoesNotExist:
+        logger.warning(f"Organization {org_id} not found for Error Tracking weekly digest")
+        return SendOrgDigestResult()
+
+    all_org_teams = {t.id: t for t in Team.objects.filter(organization_id=org.id)}
+    if not all_org_teams:
+        return SendOrgDigestResult()
+
+    # Which teams have any exceptions at all — one ClickHouse query for the whole org.
+    unfiltered_counts = weekly_digest.get_exception_counts(list(all_org_teams.keys()))
+    team_ids_with_exceptions = {row[0] for row in unfiltered_counts if row[0] in all_org_teams}
+    if not team_ids_with_exceptions:
+        return SendOrgDigestResult()
+
+    memberships: list[OrganizationMembership] = list(
+        OrganizationMembership.objects.prefetch_related("user").filter(organization_id=org.id)
+    )
+
+    # First-time users are auto-enrolled onto their busiest project. Ranking must use test-account-filtered
+    # counts: unfiltered counts can permanently enroll a user onto a project whose digest builds empty
+    # (auto-select is a one-shot decision). Only computed when the org actually has a first-time user.
+    setting_key = weekly_digest.DIGEST_PROJECT_SETTING_KEY
+    autoselect_counts: dict[int, dict] = {}
+    # Kept for Pass 2: the 14-day row query is the most expensive thing this activity does, so a team
+    # ranked here shouldn't pay for it again when its digest is built.
+    daily_rows_by_team: dict[int, list] = {}
+    if any(setting_key not in (m.user.partial_notification_settings or {}) for m in memberships):
+        for tid in team_ids_with_exceptions:
+            daily_rows_by_team[tid] = weekly_digest.query_daily_rows(all_org_teams[tid])
+            summary = weekly_digest.get_exception_summary_for_team(all_org_teams[tid], daily_rows_by_team[tid])
+            if summary and summary["exception_count"] > 0:
+                autoselect_counts[tid] = summary
+
+    # Pass 1 — resolve each recipient's enabled teams from notification settings + project access only (no
+    # ClickHouse). This yields the set of teams at least one recipient will actually receive, so Pass 2 builds
+    # heavy digest data for those alone instead of for every team in the org that happens to have exceptions.
+    recipients: list[tuple[OrganizationMembership, list[int], list[str]]] = []
+    needed_team_ids: set[int] = set()
+    for membership in memberships:
+        user = membership.user
+
+        if not should_send_notification(user, NotificationSetting.ERROR_TRACKING_WEEKLY_DIGEST.value):
+            continue
+
+        # One instance per user: its membership and access-control prefetches are cached on the
+        # instance, so rebuilding it per team would re-query them for every team in the org.
+        user_permissions = UserPermissions(user)
+        accessible_team_ids = {
+            team_id
+            for team_id in team_ids_with_exceptions
+            if user_permissions.team(all_org_teams[team_id]).effective_membership_level_for_parent_membership(
+                org, membership
+            )
+            is not None
+        }
+
+        # Rank only projects this member can open. Auto-select is one-shot and persisted, so enrolling
+        # someone onto a project they can't reach leaves every other project disabled-by-omission and
+        # silences their digest for good. A dry run only simulates it, so it can't consume the enrollment.
+        if weekly_digest.auto_select_project_for_user(
+            user,
+            org.id,
+            {tid: counts for tid, counts in autoselect_counts.items() if tid in accessible_team_ids},
+            persist=not inputs.dry_run,
+        ):
+            user.refresh_from_db(fields=["partial_notification_settings"])
+
+        enabled_team_ids: list[int] = []
+        disabled_team_names: list[str] = []
+        for team_id in accessible_team_ids:
+            if should_send_notification(user, NotificationSetting.ERROR_TRACKING_WEEKLY_DIGEST.value, team_id):
+                enabled_team_ids.append(team_id)
+            else:
+                disabled_team_names.append(all_org_teams[team_id].name)
+
+        if enabled_team_ids:
+            recipients.append((membership, enabled_team_ids, disabled_team_names))
+            needed_team_ids.update(enabled_team_ids)
+
+    date_suffix = timezone.now().strftime("%Y-%W")
+
+    if not needed_team_ids:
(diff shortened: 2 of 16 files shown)


products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,262 @@
...
+        logger.info("Skipping Error Tracking weekly digest discovery outside PostHog Cloud")
+        return []
+
+    return [str(org_id) for org_id in weekly_digest.get_org_ids_with_exceptions()]
+
+
+@activity.defn
+def get_digest_orgs_activity(inputs: GetDigestOrgsInputs) -> list[str]:
+    close_old_connections()
+    return _get_digest_orgs(inputs)
```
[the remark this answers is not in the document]
ablaszkiewicz: Removed the allowlist because now this feature is in GA.

Manual run doesn't `is_cloud()` gate IMO

products/error_tracking/backend/temporal/weekly_digest/workflow.py
```diff
@@ -0,0 +1,105 @@
...
+                        # Must match SendOrgDigestInputs.max_attempts — final-attempt detection
+                        # inside the activity depends on the two agreeing.
+                        maximum_attempts=inputs.max_attempts,
+                        initial_interval=SEND_ORG_INITIAL_RETRY_INTERVAL,
+                        backoff_coefficient=2.0,
+                        maximum_interval=SEND_ORG_MAXIMUM_RETRY_INTERVAL,
+                    ),
+                )
+
+        results = await asyncio.gather(*(send_org(org_id) for org_id in org_ids), return_exceptions=True)
```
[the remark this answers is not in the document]
ablaszkiewicz: Added continue-as-new bounded workflows

products/error_tracking/backend/temporal/weekly_digest/workflow.py
```diff
@@ -0,0 +1,105 @@
...
+            extra={"orgs": len(org_ids), "orgs_failed": orgs_failed, "sent": sent, "dry_run": inputs.dry_run},
+        )
+
+        if orgs_failed:
+            raise ApplicationError(
+                f"Error Tracking weekly digest failed for {orgs_failed}/{len(org_ids)} orgs "
+                f"({sent} digests sent to recipients in healthy orgs)",
+                type=FAILED_ORGS_ERROR_TYPE,
+                non_retryable=True,
+            )
```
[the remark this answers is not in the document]
ablaszkiewicz: Yeah it should be reported as a failure IMO!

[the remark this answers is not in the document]
ablaszkiewicz: Yes we want it to fail. And we want to aim to get 0% failure

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,270 @@
...
+    # (auto-select is a one-shot decision). Only computed when the org actually has a first-time user.
+    setting_key = _DIGEST_PROJECT_SETTING_KEYS[NotificationSetting.ERROR_TRACKING_WEEKLY_DIGEST.value]
+    autoselect_counts: dict[int, dict] = {}
+    if any(setting_key not in (m.user.partial_notification_settings or {}) for m in memberships):
+        autoselect_counts = {
+            tid: summary
+            for tid in team_ids_with_exceptions
+            if (summary := weekly_digest.get_exception_summary_for_team(all_org_teams[tid]))
+            and summary["exception_count"] > 0
+        }
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed. We cache it now per team

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,270 @@
...
+        for team_id in team_ids_with_exceptions:
+            team = all_org_teams[team_id]
+            user_permissions = UserPermissions(user).team(team)
+            if user_permissions.effective_membership_level_for_parent_membership(org, membership) is None:
+                continue
+
+            if should_send_notification(user, NotificationSetting.ERROR_TRACKING_WEEKLY_DIGEST.value, team_id):
+                enabled_team_ids.append(team_id)
+            else:
+                disabled_team_names.append(team.name)
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,270 @@
...
+
+    if failed_count or failed_team_ids:
+        # Trigger a Temporal retry; already-sent recipients are skipped via MessagingRecord, and
+        # recipients missing a failed team were deferred above so the retry can complete them.
+        raise Exception(
+            f"Error Tracking weekly digest failed for {failed_count} recipients "
+            f"and {len(failed_team_ids)} team builds in org {org_id}"
+        )
+
+    return SendOrgDigestResult(sent=sent_count, teams_built=len(team_digest_data))
```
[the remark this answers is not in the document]
ablaszkiewicz: That's a correct find but leaving this. It's not used for anything, just an approximate counter. We get true count by querying logs

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,270 @@
...
+            try:
+                weekly_digest.send_digest_to_workflow(digest, distinct_id)
+            except Exception:
+                logger.exception("et_weekly_digest.send_failed", user_id=str(user.uuid), org_id=org_id)
+                failed_count += 1
+                continue
+
+            record.sent_at = timezone.now()
+            record.save()
+            sent_count += 1
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,276 @@
...
+    autoselect_counts: dict[int, dict] = {}
+    # Kept for Pass 2: the 14-day row query is the most expensive thing this activity does, so a team
+    # ranked here shouldn't pay for it again when its digest is built.
+    daily_rows_by_team: dict[int, list] = {}
+    if any(setting_key not in (m.user.partial_notification_settings or {}) for m in memberships):
+        for tid in team_ids_with_exceptions:
+            daily_rows_by_team[tid] = weekly_digest.query_daily_rows(all_org_teams[tid])
+            summary = weekly_digest.get_exception_summary_for_team(all_org_teams[tid], daily_rows_by_team[tid])
+            if summary and summary["exception_count"] > 0:
+                autoselect_counts[tid] = summary
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,285 @@
...
+            is not None
+        }
+
+        # Rank only projects this member can open. Auto-select is one-shot and persisted, so enrolling
+        # someone onto a project they can't reach leaves every other project disabled-by-omission and
+        # silences their digest for good.
+        if weekly_digest.auto_select_project_for_user(
+            user, org.id, {tid: counts for tid, counts in autoselect_counts.items() if tid in accessible_team_ids}
+        ):
+            user.refresh_from_db(fields=["partial_notification_settings"])
```
cat-ph, line 133: do we want to gate this with the dry run too?

products/error_tracking/backend/weekly_digest.py
```diff
@@ -60,14 +62,16 @@ def _query_daily_rows(team: Team) -> list:
         """,
         team=team,
         filters=HogQLFilters(filterTestAccounts=True),
+        # Weekly batch job: Celery pinned this to the offline cluster process-wide, Temporal does not.
+        workload=Workload.OFFLINE,
```
cat-ph, line 67: good catch 🚀

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,285 @@
...
+    org shifting in or out between pages is either <= cursor (already processed) or
+    > cursor (still pending), never both.
+    """
+    if inputs.org_ids is not None:
+        candidates = sorted(str(org_id) for org_id in inputs.org_ids)
+    elif not is_cloud():
+        logger.info("Skipping Error Tracking weekly digest discovery outside PostHog Cloud")
+        return []
+    else:
+        candidates = sorted(str(org_id) for org_id in weekly_digest.get_org_ids_with_exceptions())
```
cat-ph, line 46: hm, I think we do this for each page but it wouldn't really change right? I don't think it's a _huge_ query so it's probably fine 🤔

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,285 @@
...
+    The candidate set is sorted and sliced to (``after``, ``after`` + ``limit``] so the
+    workflow can page with only a cursor. Recomputing candidates each page is safe: an
+    org shifting in or out between pages is either <= cursor (already processed) or
+    > cursor (still pending), never both.
+    """
+    if inputs.org_ids is not None:
+        candidates = sorted(str(org_id) for org_id in inputs.org_ids)
+    elif not is_cloud():
+        logger.info("Skipping Error Tracking weekly digest discovery outside PostHog Cloud")
+        return []
```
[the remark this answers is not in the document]
cat-ph: yeah legit but p1 probably overstated, not sure how much we care but should be a small fix maybe worth doing it

products/error_tracking/backend/temporal/weekly_digest/activities.py
```diff
@@ -0,0 +1,285 @@
...
+                "et_weekly_digest.lacking_digest_sent",
+                org_id=org_id,
+                user_id=str(user.uuid),
+                missing_team_ids=missing_failed_team_ids,
+                teams_sent=len(user_team_sections),
+            )
+
+    # Per-org, per-attempt summary. Counts are per-attempt but don't double-count across retries: each
+    # recipient is sent (and counted) in exactly one attempt, then skipped via the sent_at guard. So
+    # summing a field across an org's attempts gives the true total.
```
cat-ph: hmm I don't think we actually do that? or I might be confused, I saw we have the carried_sent and sum the output, but I think we still lose the numbers on retry? 🤔

.github/workflows/container-images-cd.yml
```diff
@@ -508,6 +508,7 @@ jobs:
                         - 'products/error_tracking/backend/temporal/**'
                         - 'products/error_tracking/backend/recommendations/**'
                         - 'products/error_tracking/backend/models.py'
+                        - 'products/error_tracking/backend/weekly_digest.py'
```
cat-ph, line 511: I wonder if we should also rely on `tasks/email.py` in case we change `should_send_notification` or stuff

products/error_tracking/backend/weekly_digest.py
ablaszkiewicz: Thanks to opus 5

products/error_tracking/backend/temporal/weekly_digest/activities.py
ablaszkiewicz: My whole idea for the dry run was to check the performance because correctness-wise I know that this works so my thinking was that I'd just keep email sending behind a flag but yeah it totally make sense to get rid of all side effects if it's a dry run. Will gate it

products/error_tracking/backend/temporal/weekly_digest/activities.py
ablaszkiewicz: Hey very good catch. So initially it all started with calling this once and then carrying list of orgs through temporal. I got suspicious that we are carrying so much data tho and we (me+ 🤖) ran an analysis and found out that we were reaching temporals payload limits.

That's when I started exploring options and stumbled across continued workflows with pages and this whole thing uses that pattern. I checked that CH query and it takes ~6s so for the sake of simplicity and less things to break I thought it's better to invoke this query on 27 pages rather than store the results somewhere outside temporal and refetch it.

products/error_tracking/backend/temporal/weekly_digest/activities.py
ablaszkiewicz: Done

products/error_tracking/backend/temporal/weekly_digest/activities.py
ablaszkiewicz: Good catch. Fixed

.github/workflows/container-images-cd.yml
ablaszkiewicz: You're right. I found some other things we should reference here. Updated the list

products/error_tracking/backend/temporal/weekly_digest/activities.py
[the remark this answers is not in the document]
ablaszkiewicz: I don't think anybody would ever use that but fixed to be bulletproof

products/error_tracking/backend/temporal/weekly_digest/activities.py
cat-ph: ahh fair, that makes a lot of sense yeah, better than invoking this a few times and it's weekly anyway, thanks for checking 💙

cat-ph (approved):
let's go! 🚀

PR #72741, opened 2026-07-21
feat(personhog): admit person writes against the writer's full rejection surface

z0br0wn:
## Problem

  > [!NOTE]
  > Stacked on #72712 (warning registry consolidation). Only the top commit is this PR's diff until that merges.
  
The personhog changelog is the source of truth for person property state, and Postgres is a downstream consumer of it via the writer. That only works if every record the leader acks is applyable by the writer verbatim. Today it isn't guaranteed: the leader acks first, and the writer patches over unapplyable records after the fact with silent repairs. It trims oversized properties (making Postgres diverge from the cache and changelog, which hold the untrimmed state), silently drops rows with unbindable fields (an acked write vanishes and the store layer reports success), and substitutes defaults for undecodable values.

Any writer-side edit or drop is permanent divergence, not a transient gap. The changelog carries full-state snapshots, so every later record for a person is built on the same cached state that failed to apply. The next snapshot reproduces the problem rather than healing it, and the person's Postgres row is wrong forever. On top of that, a rejected or trimmed write gives the caller no feedback, since the correction happens minutes later and several services away.

## Changes

**Before**

  ```mermaid
  flowchart LR
      C[client] --> L[leader\nmerge + ack]
      L --> K[(changelog)]
      K --> W[writer\ntrim, skip, substitute]
      W --> PG[(Postgres)]
      W -.->|diverged state| PG
      class C phYellow; class L phBlue; class W phRed; class K,PG phGray;
  ```

**After**

  ```mermaid
  flowchart LR
      C[client] --> L[leader\nsanitize, measure, trim or reject, assert]
      L -->|only applyable records| K[(changelog)]
      L -.->|INVALID_ARGUMENT + warning| C
      K --> W[writer\npure applier, halts on violation]
      W --> PG[(Postgres)]
      class C phYellow; class L,W phBlue; class K,PG phGray;
  ```

Admission moves into the leader, before the ack, under the per-person lock, on the merged result of each update:

  - Sanitize (`personhog_common::properties::sanitize_for_jsonb`): NUL becomes U+FFFD in every string, matching the Node pipeline's `sanitizeJsonbValue` (Postgres jsonb refuses NUL outright). Floats beyond 1e307 are clamped: Postgres renders jsonb numerics in expanded decimal notation, and serde_json cannot parse expansions of roughly 1e308 and above, so larger values would apply to Postgres but never load back into any Rust reader, including the leader's own PG fallback.
  - Measure (`jsonb_column_size`): an exact reimplementation of `pg_column_size` for JSONB, welded byte-equal to a live Postgres by a property test over ~90 fixtures (alignment sweeps, numeric encoding edges, unicode, a ~500KB person). Above the constraint threshold, non-protected properties trim alphabetically to the target, or the update is rejected with `INVALID_ARGUMENT` carrying sizes, never property values. Inverted limit configuration (target above threshold) refuses to boot.
  - Assert (`assert_writeable`): identity fields that originate from earlier state rather than this request (uuid parses, team_id fits the column's integer type, created_at within sane bounds) are checked before any produce.

The writer becomes a pure applier. Its trim machinery and all four silent drop or repair paths are deleted. A non-transient rejection is now an invariant violation: the flush halts without committing, Kafka redelivers after restart, and the alarm stands until admission's gap is fixed. User-facing size warnings move to the leader (throttled per team and type, matching the Node limiter), where the caller also gets synchronous feedback.

The exhaustiveness claim is written down as a disposition table: every rejection cause of the writer's actual statement is either checked at admission, impossible by a construction premise verified in-tree (sole producer, no serde_json `arbitrary_precision`, parse recursion limit), or backstopped by the writer's halt. The one residual is a cross-row uuid uniqueness collision, which admission cannot see and the halt covers.

Automated only, run by the agent:

  - New adversarial weld (`personhog-writer/tests/admission_weld.rs`): 13 hostile property fixtures (NULs in keys, values and nested arrays, 120-deep nesting, extreme numbers, unicode edge cases, post-sanitize key collisions, a 400KB payload) go through the verbatim admission functions and then the writer's real
  upsert statement against live Postgres, asserting byte-exact round-trips. A red-proof asserts the unsanitized NUL fixture fails that same statement. This caught a real bug during development: the numeric round-trip poisoning that motivated the float clamp.
  - The size-measure weld (`personhog-common/tests/jsonb_size_pg.rs`) asserts `jsonb_column_size` equals `pg_column_size` on live Postgres for every fixture.
  - Leader integration tests cover trim, reject with no residue, warning payload shape, deterministic throttle suppression, and NUL sanitization end to end through gRPC with strong-read agreement. A should-panic test guards the config inversion assert.
  - Writer tests are inverted from the old contract: an unbindable row now fails its chunk and surfaces as a violation instead of being silently skipped, and an oversized row errors and writes nothing.
  - Full suites green: personhog-common 62, writer 49, leader 82, clippy `-D warnings`, fmt.
  - Latency checked with the personhog test harness: interleaved no-chaos gates (release binaries, 3 leaders, 8 partitions, 200 persons, 30s) against the branch's merge base show all write and read percentiles within run-to-run noise, with zero failed writes and zero consistency violations across four runs.

Not done in this PR, tracked as follow-ups: restoring the `check_properties_size` constraint on the partitioned production table (needs a lock-safe change window, deliberately not a migration file), and plugin-server handling of the new `INVALID_ARGUMENT` rejections.

## 🤖 Agent context

Authored with Claude Code, directed by the PR assignee. Design decisions along the way: writer-side correction was rejected first (any writer edit diverges PG from cache and changelog), then writer-side skipping was rejected too after realizing skips are permanent divergence under full-state snapshots, which forced the invariant into admission. The size measure went through a live-PG weld because a serialized-length approximation is exactly the drift that produced the old trim margins. The float clamp exists because the adversarial weld failed on its first run and exposed the expanded-numeric round-trip poison. A remediation design for poison records (parking, tombstoning, repair RPCs) was explored and deliberately shelved in favor of making unapplyable records unreachable and proving it. The `/writing-tests` skill shaped the test choices.


rust/personhog-common/src/properties.rs
```diff
@@ -0,0 +1,500 @@
...
+/// non-protected property still cannot fit, which callers distinguish by
+/// re-checking the size; see `TrimResult` for the ergonomic wrapper.
+pub enum TrimResult {
+    /// Already within the target; nothing to do.
+    Fits,
+    /// Trimming non-protected properties got it under the target.
+    Trimmed(Value),
+    /// Protected properties alone exceed the target; nothing to trim.
+    CannotFit,
+}
```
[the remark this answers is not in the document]
z0br0wn: legit, fixing

rust/personhog-leader/src/service.rs
```diff
@@ -293,6 +333,33 @@ impl PersonHogLeaderService {
...
+        ));
+    }
+    if p.created_at < 0 || p.created_at > MAX_EPOCH_SECS_YEAR_9999 {
+        return Err(format!(
+            "created_at epoch {} is outside sane bounds",
+            p.created_at
+        ));
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
z0br0wn: The scenario can't arise on the leader path: CachedPerson has no last_seen_at field at all, so recovery/warming/PG loads drop it at decode and cached_person_to_proto hardcodes None on every produce — a leader-produced record structurally cannot carry a populated value. The writer's range check on that column can only fire for a record from a non-leader producer, which is the foreign-corruption case its halt is designed to catch. Good catch on the gap in the code's self-documentation though: assert_writeable now says explicitly why it checks exactly the fields it does (they're the only identity fields representable in the leader's state; last_seen_at and the legacy jsonb columns are empty by construction).

rust/personhog-writer/src/writer.rs
```diff
@@ -82,70 +76,76 @@ impl<D: PersonDb + 'static> WriterTask<D> {
...
                             rows = data_failed.len(),
                             "batch had data-failed chunks, falling back to per-row"
                         );
-                        for r in self.store.upsert_rows_parallel(data_failed).await {
-                            match r {
-                                RowResult::Written => {}
-                                RowResult::Trimmed(w) | RowResult::Skipped(w) => warnings.push(w),
-                            }
+                        let fallback = self.store.upsert_rows_parallel(data_failed).await;
+                        if !fallback.violations.is_empty() {
```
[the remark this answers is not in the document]
z0br0wn: there is no existing traffic going through the leader / writer, so this isn't a concern

rust/personhog-common/src/properties.rs
```diff
@@ -0,0 +1,551 @@
...
+        4
+    };
+    4 + header + 2 * groups
+}
+
+/// PG's short numeric header packs the sign, a 6-bit display scale, and a
+/// signed weight; it fits when dscale <= 63 and weight is in [-44, 83]
+/// base-10000 units (NUMERIC_SHORT_WEIGHT_MIN/MAX in numeric.c).
+fn short_header_fits(weight: i32, dscale: u32) -> bool {
+    dscale <= 0x3F && (-44..=83).contains(&weight)
```
nickbest-ph: claude says that this should be `-64..=63` to avoid accidentally admitting an oversized property.

your bot also seems like it did some research. might just want to double check, or somehow write a test to validate this

rust/personhog-writer/src/writer.rs
```diff
@@ -82,70 +76,76 @@ impl<D: PersonDb + 'static> WriterTask<D> {
...
+                            let first = &fallback.violations[0];
+                            self.handle.signal_failure(format!(
```
nickbest-ph, line 120: does this fully halt writes?
signal_failure is async? so does it immediately stop the writer?
is there a gap here that alllows the writer to process the next batch and then continue committing offsets past this poison pill that is supposed to be halted at?

rust/personhog-common/src/properties.rs
z0br0wn: "the reviewer's bot is right, ours is wrong" - claude

rust/personhog-writer/src/writer.rs
z0br0wn: great catch, and the shutdown itself is also async

rust/personhog-leader/src/service.rs
```diff
@@ -456,6 +559,71 @@ impl PersonHogLeader for PersonHogLeaderService {
...
+                .increment(sanitize_stats.nul_strings);
+        }
+        if sanitize_stats.clamped_numbers > 0 {
+            counter!("personhog_leader_properties_numbers_clamped_total")
+                .increment(sanitize_stats.clamped_numbers);
+        }
+
+        let jsonb_size = jsonb_column_size(&new_properties);
+        if jsonb_size > self.size_limits.threshold {
+            match trim_properties_to_fit_size(&new_properties, self.size_limits.trim_target) {
```
nickbest-ph: I think this is a bit different than the current behaviour:

trimming is supposed to act as remediation for records that were already in the database and oversized when the size constraint got added to the properties field. this meant that any write to that record (even if it didn't icnrease the size of it) was immediately rejected by the DB.

here, we are using it as a way to admit person property updates for events that would (without the trim) newly violate the properties size

is it possible to mirror the node behaviour? the trim is kinda nasty, as it's deferred arbitrary data loss, whereas rejecting the property update on new violations with an ingestion warning is at least a deliberate/obvious outcome (don't do person processing for a person that is over the size limit).

z0br0wn: ah, i think i missed that context the first time around. updating!

rust/personhog-common/src/properties.rs
z0br0wn: updated!

rust/personhog-writer/src/writer.rs
z0br0wn: updated!

rust/personhog-leader/src/service.rs
z0br0wn: updated!

nickbest-ph (approved):

rust/personhog-common/src/properties.rs
```diff
@@ -0,0 +1,551 @@
...
+}
+
+/// Floats at or above this magnitude are clamped by [`sanitize_for_jsonb`].
+/// Postgres renders jsonb numerics in expanded decimal notation (never
+/// e-notation), and serde_json's parser overflows on integer expansions of
+/// ~1e308 and beyond — so a number in the window [1e308, f64::MAX] applies
+/// to PG but can never be read back by a Rust consumer (including the
+/// leader's own PG fallback load). 1e307 expands to 308 digits, which
+/// parses back exactly.
+pub const MAX_JSONB_SAFE_MAGNITUDE: f64 = 1e307;
```
nickbest-ph, line 108: hmm, can nodejs write a number larger than this into a jsonb field? would we panic when trying to read numbers in person properties that were written by the nodejs service?

apparently Number.MAX_VALUE in JS will get expanded when stored in PG and be above this value.

z0br0wn: good catch, updated to include a fallback parse path / clamping

nickbest-ph (approved):

PR #72743, opened 2026-07-21
fix(aio): prefill PostHog AI composer from /ai?ask= without consent

Radu-Raicea:
## Problem

The "Start with AI" button in the new evaluation creation flow deep-links to `/ai?ask=<prompt>`, expecting PostHog AI to open with the prompt filled in. On the legacy Max runtime (which most users get — the new sandbox surface is behind `PHAI_SANDBOX_MODE`), nothing happened: the prompt appeared to vanish.

The legacy `/ai?ask=` handler forwarded the prompt straight into `askMax`, an auto-send action that never populates the visible composer and silently no-ops when the org hasn't granted AI data-processing consent. So the deep link either sent an invisible message or dropped the prompt entirely with no feedback.

## Changes

When AI data-processing consent is missing, the `ask=` handler now prefills the composer via `setQuestion` instead of calling `askMax` — the prompt stays visible and the user's manual submit surfaces the consent flow. With consent present it still auto-sends, matching the sandbox runtime's `autoSubmit` behavior.

This is a bugfix on the frozen LangGraph path (keeps an existing deep link working), not new capability.

Added two Jest cases to `maxLogic.test.ts` under `ask URL parameter`:
- without consent → the composer is prefilled and no auto-send fires (the regression this PR fixes; no existing test covered `search.ask`);
- with consent → `askMax` is dispatched and the composer is not prefilled.

Ran `maxLogic.test.ts` locally — 42 passed. Lint/format clean. No manual browser testing was performed in this environment.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated a report that the evaluations "Start with AI" deep link didn't fill the PostHog AI chat. Traced it through `useOpenAi`/`urls.ai` (URL construction correct) to the two `/ai?ask=` consumers, and found the param name matches in both runtimes — the defect was behavioral in the legacy `maxLogic` handler (auto-send that no-ops without consent), not a naming mismatch or anything in the evaluations screen. Chose to make the legacy path degrade to a visible composer prefill, mirroring the newer `phaiAiComposerSeedLogic` (`autoSubmit: dataProcessingAccepted`) rather than extending the frozen path with new behavior.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C087XQ7K9K7/p1784662502218689)*


skoob13 (approved):
Thank you!

PR #54202, opened 2026-04-13
feat(flags): flags consumer poc

matheus-vb:
## Problem

This PR introduces the POC for the flags consumer service, referenced in the [flags dedicated store RFC](https://github.com/PostHog/requests-for-comments-internal/pull/1014/), and adds it to the local stack setup scripts.

This service should consume events from Kafka and write the denormalized data in a new PG database. This DB does not exist in any production environment yet.

One of the components of https://github.com/PostHog/posthog/issues/54145.

**Note:** our main goal right now is to use the `benchmark` command and test our initial schema against a more realistic workload. The consumer service itself has some rough edges, and that's okay for now. Once the schema is validated, we can polish and wrap up the service.

## Changes

Introduces a new Rust service (`flags-consumer`) that consumes person change events from Kafka and writes them to a dedicated `flags_read_store` PostgreSQL database using version-guarded upserts.

**New service: `rust/flags-consumer/`**

The service consumes from two Kafka topics (`clickhouse_person` and `clickhouse_person_distinct_id`) and writes to the `flags_person_lookup` table in the dedicated database. The architecture uses a bounded channel pipeline: two consumer tasks (one per topic) feed events into a shared channel, and a single batch processor task drains the channel and flushes writes on size (500 messages) or timeout (100ms).

Person property upserts and deletions are batched via `INSERT ... SELECT FROM UNNEST()` for efficiency. Distinct ID assignments use individual transactions to atomically handle person merges (remove from old owner, add to new owner). All writes use version guards (`WHERE stored_version < incoming_version`) so that out-of-order Kafka messages are handled correctly. The service writes periodic heartbeat records to `flags_read_store_heartbeat` for lag monitoring.

The consumer loop is generic over a `KafkaMessage` trait, so adding a new topic requires only a new message struct and trait impl with no changes to the consumer machinery.

**Database schema: `rust/flags_read_store_migrations/`**

Creates the `flags_person_lookup` table, which stores one row per person with a `distinct_ids TEXT[]` column indexed by GIN. This replaces the two-table JOIN (`posthog_person` + `posthog_persondistinctid`) used by the flags service today with a single GIN index lookup. The table is HASH-partitioned by `team_id` into 64 partitions, matching the `posthog_person` table layout.

Also creates the `flags_read_store_heartbeat` table for CDC lag monitoring per topic and partition.

**Local dev stack integration**

[...]

PR #72746, opened 2026-07-21
chore(flags): remodel flags cdc store schema

matheus-vb:
## Problem

The feature flag CDC read store kept distinct IDs in a GIN indexed array. Its write amplification, uniqueness limitations, and benchmark methodology made the schema difficult to validate under production load.

## Changes

This replaces the array with hash partitioned person and distinct ID mapping tables, updates CDC storage and batching semantics, and rebuilds the benchmark with typed workloads, open loop scheduling, bounded executors, PostgreSQL sampling, plan checks, deadline accounting, and explicit gates.

The flags consumer test suite passed 59 tests. Strict Clippy and formatting checks passed. A PostgreSQL 15 smoke run completed all six phases with 400,000 people across eight partitions. Removing the mapping table primary key made plan validation reject the resulting sequential scan as expected.

## Docs update

No.


rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
```diff
@@ -0,0 +1,47 @@
+-- Versioned tombstones keep stale CDC replays from restoring deleted rows.
```
gustavohstrassburger, line 8: This PR replaces an already-shipped migration in place: `20260411000001_create_flags_person_lookup.sql` exists on master (landed in #54202); this PR deletes it and adds `20260411000001_create_flags_read_store.sql` at the same version prefix with different DDL and a different description.

Migrations run via `sqlx migrate run` (`rust/bin/migrate-entry`, non-fresh-mode path), which tracks a checksum and description per version in `_sqlx_migrations`. Any environment that already applied the old migration would fail on the next deploy with a `VersionMismatch` error, since both the checksum and description changed for the same version number. It also never drops the old `flags_person_lookup` table, so a pre-existing environment would keep the stale table even past the checksum check.

Today this is low-risk: no CI workflow runs `sqlx migrate run` against `flags_read_store_migrations` yet, and the DB is currently only created empty via `docker/postgres-init-scripts/create-flags-read-store-db.sh`. But the fix is cheap and removes the risk entirely: add a new migration (e.g. `20260411000002_remodel_flags_read_store.sql`) that drops `flags_person_lookup` and creates the two new tables, leaving the original file untouched. Once these migrations get wired into real CI or a deploy, replacing an applied migration in place stops being safe.

rust/flags-consumer/src/storage/postgres.rs
```diff
@@ -17,9 +18,69 @@ static LABELS_DID_ASSIGN: LazyLock<[(String, String); 1]> =
...
+    INSERT INTO flags_distinct_id_map (team_id, distinct_id, person_uuid, version)
+        SELECT * FROM UNNEST($1::int[], $2::text[], $3::uuid[], $4::bigint[])
+    ON CONFLICT (team_id, distinct_id) DO UPDATE SET
+        person_uuid = EXCLUDED.person_uuid,
+        version = EXCLUDED.version,
+        deleted_at = NULL
+    WHERE flags_distinct_id_map.version < EXCLUDED.version
+"#;
+
+pub(crate) const DISTINCT_ID_DELETE_SQL: &str = r#"
```
gustavohstrassburger, line 70: The version-guard SQL in the four batch functions is the core behavior change of this PR, but no test runs these statements against rows to confirm the guard holds. The in-batch reconciliation tests in `processor.rs` only cover competing mutations within a single batch, once an assignment and a stale deletion land in separate batches, the `WHERE` clauses here are the only thing preventing a stale write from winning.

Concretely: an old-owner distinct-ID deletion (version 110) arrives in a later batch than a new-owner assignment (version 11) that already committed. The owner-guard `flags_distinct_id_map.person_uuid = EXCLUDED.person_uuid` must reject it, or the live mapping gets tombstoned and flag evaluation loses the person. Nothing asserts this today.

Add a `#[sqlx::test]` integration test that inserts a row, then calls `batch_upsert_distinct_ids` / `batch_delete_distinct_ids` with a lower version and a different `person_uuid`, and asserts `rows_affected() == 0` and the row is unchanged. Mirror it for `batch_upsert_persons` (an older `person_version` must not overwrite).

rust/flags-consumer/src/storage/postgres.rs
```diff
@@ -59,31 +120,20 @@ impl PostgresStorage {
             updates.iter().map(|u| u.properties.clone()).collect();
```
gustavohstrassburger: `batch_upsert_persons` clones every person's JSONB properties to build the bind array:

```rust
let properties: Vec<serde_json::Value> =
    updates.iter().map(|u| u.properties.clone()).collect();
```

At `batch_size=500` with non-trivial property documents, that's up to 500 deep clones purely to satisfy the bind. sqlx can bind a slice of references:

```rust
let properties: Vec<&serde_json::Value> = updates.iter().map(|u| &u.properties).collect();
```

Worth confirming sqlx's `Encode` accepts `&[&Value]` for the `jsonb[]` bind before applying.

rust/flags-consumer/src/benchmark/executor.rs
```diff
@@ -0,0 +1,1204 @@
...
+    cancellation: &CancellationToken,
+) -> Option<T> {
+    tokio::select! {
+        biased;
+        () = cancellation.cancelled() => None,
+        item = async { receiver.lock().await.recv().await } => item,
+    }
+}
+
+async fn run_person_batch_worker(
```
gustavohstrassburger: `run_person_batch_worker` and `run_distinct_id_batch_worker` are byte-for-byte identical except for the dedupe function and the storage call. Extract a generic `run_batch_worker` parameterized over `prepare` and an async `execute` closure, so a future change to the worker loop happens once instead of drifting between two copies.

rust/flags-consumer/src/benchmark/pg_sampler.rs
```diff
@@ -0,0 +1,432 @@
...
+
+use anyhow::Context;
+use chrono::{DateTime, Utc};
+use serde::Serialize;
+use sqlx::postgres::PgPoolOptions;
+use sqlx::PgPool;
+
+use super::ops::{PhaseId, PhaseMismatch};
+
+pub const DEFAULT_SAMPLE_INTERVAL: Duration = Duration::from_secs(10);
```
gustavohstrassburger: `PgSampler` exposes `interval()`, `phase_id()`, `reset_phase()`, an `interval` field, and `DEFAULT_SAMPLE_INTERVAL`, but none are called anywhere in the crate, the sampler loop actually drives off a separate `PG_SAMPLE_INTERVAL` const in `mod.rs`. Since the getters are `pub`, clippy's dead-code lint won't catch this. Remove the three methods, the `interval` field, and `DEFAULT_SAMPLE_INTERVAL`, leaving `PG_SAMPLE_INTERVAL` as the single source of truth.

rust/flags-consumer/src/benchmark/scheduler.rs
```diff
@@ -0,0 +1,547 @@
...
+            Self::InvalidRate { class, rate } => {
+                write!(formatter, "invalid arrival rate {rate} for {class:?}")
+            }
+        }
+    }
+}
+
+impl std::error::Error for ArrivalError {}
+
+fn sample_interval_nanos(rng: &mut impl Rng, rate: RatePerSecond) -> u64 {
```
gustavohstrassburger, line 177: `sample_interval_nanos` is the exponential inter-arrival sampler behind the "open-loop scheduling" this PR advertises, and it has no comment. `1.0 - rng.gen::<f64>()` deliberately remaps `[0,1)` to `(0,1]` so `ln()` never hits `-inf`, and `-ln(u)/rate` is the inverse-CDF of the exponential distribution, without a note, a reader may "simplify" the `1.0 -` away as a no-op. Worth a one-line comment naming the Poisson/exponential sampling and the domain guard.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
```diff
@@ -0,0 +1,47 @@
...
+        EXECUTE format(
+            'CREATE TABLE IF NOT EXISTS flags_distinct_id_map_p%s PARTITION OF flags_distinct_id_map FOR VALUES WITH (MODULUS %s, REMAINDER %s)',
+            i, num_partitions, i
+        );
+    END LOOP;
+END $$;
+
+-- Heartbeat / lag-monitoring table. The CDC consumer will write one row per (source, partition) with
+-- the last Kafka offset / event timestamp it processed; the shadow read path will
+-- join against this for staleness checks. Table-only in Step 1; no writes yet.
```
gustavohstrassburger: This comment says "Table-only in Step 1; no writes yet," but this PR ships `PostgresStorage::write_heartbeat`, which writes exactly this table. Looks stale relative to the code landing here, worth dropping the "no writes yet" sentence.

gustavohstrassburger (approved):
Solid remodel, no blockers. One thing to fix before merge: the migration replaces an already-shipped file in place (same version, different content), which will break sqlx migrate run the moment this crate's migrations get wired into real CI or a deploy. A few suggestions inline, mostly benchmark-scoped.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
matheus-vb: This service was never deployed, so the migration was never applied.

rust/flags-consumer/src/benchmark/executor.rs
matheus-vb: I'm keeping it as is for now, at this point the async abstraction would add disproportionate complexity imo. Can revisit this in the future.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
```diff
@@ -0,0 +1,47 @@
...
+CREATE TABLE IF NOT EXISTS flags_read_store_heartbeat (
+    source         TEXT        NOT NULL,
+    partition      INTEGER     NOT NULL,
+    last_offset    BIGINT      NOT NULL DEFAULT 0,
+    last_event_ts  TIMESTAMPTZ,
+    updated_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (source, partition)
+);
```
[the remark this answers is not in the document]
matheus-vb: This service was never deployed, so the migration we are replacing was never applied.

rust/flags-consumer/src/storage/postgres.rs
haacked, line 70: `blocking`: The version guard alone passes when a deletion at version 110 arrives after an assignment at version 11 (since 110 > 11). The owner check `person_uuid = EXCLUDED.person_uuid` is what actually rejects the stale deletion.

The test needs two separate cases: an owner mismatch with a higher version (owner guard rejects), and a same-owner stale version (version guard rejects). Mirror both for `batch_upsert_persons` and `batch_delete_persons` too, since those only have the version guard and no owner check.

rust/flags-consumer/src/benchmark/executor.rs
haacked: `suggestion`: Keep the two dedupe functions separate (different key types and tie-break logic) and unify only the worker shell:

```rust
async fn run_batch_worker<T, Prep, Exec, Fut>(
    storage: Arc<PostgresStorage>,
    receiver: Arc<Mutex<mpsc::Receiver<Vec<DispatchedOperation>>>>,
    completion_sender: mpsc::Sender<CompletionRecord>,
    clock: PhaseClock,
    retry: RetryConfig,
    cancellation: CancellationToken,
    prepare: Prep,
    execute: Exec,
) -> anyhow::Result<()>
where
    Prep: Fn(&[DispatchedOperation]) -> Result<Vec<T>, BatchPayloadError>,
    Exec: Fn(&PostgresStorage, &[T]) -> Fut,
    Fut: Future<Output = Result<u64, sqlx::Error>>,
```

Each caller becomes a one-liner passing `dedupe_person_updates` + `|s, items| s.batch_upsert_persons(items)` (or the distinct-id equivalents).

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
```diff
@@ -0,0 +1,45 @@
+-- Versioned tombstones keep stale CDC replays from restoring deleted rows.
```
haacked, line 8: `suggestion`: This isn't wired into any deployed or CI path yet (`migrate-entry`'s `all` case skips `flags-read-store`, and `Dockerfile.sqlx-migrate` never copies this migrations directory), so it's local-dev only. The one live path is `bin/migrate --scope=flags-read-store`, which runs the checksum-tracked `sqlx migrate run` against the local filesystem, so any developer who ran that scope since the original PR landed hits the error on their next migrate.

If you take the new-migration approach, [`benchmark/schema.rs:59`](https://github.com/PostHog/posthog/blob/48db94d2b55eb24e70e9ac67136a541b7b1943b7/rust/flags-consumer/src/benchmark/schema.rs#L59) already has a proven `DROP TABLE IF EXISTS flags_person_lookup CASCADE` you can reuse verbatim for the `...000002` migration.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
```diff
@@ -0,0 +1,45 @@
...
+CREATE TABLE IF NOT EXISTS flags_person (
```
haacked, line 9: `suggestion`: This migration creates both partitioned tables at Postgres' default fillfactor (100), but `benchmark/schema.rs` sweeps fillfactor down to 10. The `flags_person` table takes an UPDATE on every property change, and HOT (heap-only tuples) is only eligible when there's free space on the page.

If the sweep finds a lower fillfactor reduces bloat and WAL, there's currently no way to ship it since the migration hardcodes the default. Either set the production fillfactor to whatever the sweep lands on, or have both DDLs share one source so they can't diverge.

rust/flags-consumer/src/benchmark/mod.rs
```diff
@@ -1,168 +1,1694 @@
...
+        }
+        Ok(dispatcher)
+    })
+}
 
-    tracing::info!("=== Phase 2: Identification appends ===");
-    let r = workloads::phase_identification_appends(&pool, &data, &args).await?;
-    tracing::info!(ops = r.latency.count, errors = r.errors, "phase 2 complete");
-    results.push(r);
+#[allow(clippy::too_many_arguments)]
```
haacked, line 1072: `suggestion`: Seven functions in this file carry `#[allow(clippy::too_many_arguments)]`. `drive_catch_up_load` takes 20 parameters and `drive_open_load` 17, with a dozen overlapping: `world`, `workload_rng`, `growth`, `world_config`, `memory_limit`, `args`, `collector`, `intervals`, `tracker`, `pg_records`, `pg_receiver`, `sink`.

Grouping parameters that travel together into context structs (workload-generation context, metrics-sink context) and passing `&mut ctx` would cut most of these `#[allow]`s instead of suppressing the lint seven times.

rust/flags-consumer/src/benchmark/plancheck.rs
```diff
@@ -0,0 +1,329 @@
...
+    pub plan: Value,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum PlanRequirement {
+    CanonicalRead,
+    Write { table: &'static str },
+}
+
+pub async fn verify_storage_plans(pool: &PgPool) -> anyhow::Result<Vec<NamedPlanEvidence>> {
```
haacked, line 33: `suggestion`: `validate_plan` has solid unit tests against hardcoded JSON, but `verify_storage_plans` (the part that runs `EXPLAIN` against the real migrated schema and catches missing or wrong indexes) is only reachable from the benchmark CLI.

The PR description's "removing the mapping table PK made plan validation reject the sequential scan" was a manual one-off, so a future migration that reshapes an index won't be caught in CI. A `#[sqlx::test]` that applies the migration and asserts `verify_storage_plans(&pool)` returns `Ok` would pin this down, and it can share the same Postgres test harness the version-guard tests need.

rust/flags-consumer/src/benchmark/executor.rs
```diff
@@ -0,0 +1,1204 @@
...
+        )
+        .await;
+        cancel_on_error(&emission, &cancellation);
+        emission?;
+        abort_after_terminal(terminal_error.as_deref(), &cancellation)?;
+    }
+    Ok(())
+}
+
+async fn execute_merge(
```
haacked, line 511: `suggestion`: `execute_merge` composes the assignment upsert, target person upsert, and source person delete as three independently-retried calls with no surrounding transaction, unlike the old store's `tx.commit()`. A crash between calls can leave distinct IDs repointed to the target while the source person is still live.

Each statement is version-guarded and idempotent, so a re-run reconverges, and this path only runs in the benchmark. Is this the intended way to validate partial-failure behavior, or did you drop the old transactional guarantee unintentionally?

rust/flags-consumer/src/benchmark/executor.rs
```diff
@@ -0,0 +1,1204 @@
...
+
+    pub fn is_cancelled(&self) -> bool {
+        self.cancellation.is_cancelled()
+    }
+
+    pub fn cancellation_token(&self) -> CancellationToken {
+        self.cancellation.clone()
+    }
+
+    pub async fn close_and_drain<F>(
```
haacked: `nit`: `close_and_drain`, `close_dispatchers_and_drain`, and `drain_after_senders_closed` all forward to `drain_closed` and differ only in which handle they drop first. Collapsing to one `drain_closed` and letting callers drop their own handle (`drop(ingress); runtime.drain(...)`) would cut two near-identical public methods.

haacked (approved):
Solid remodel. One blocking gap: the version-guard SQL that makes cross-batch writes safe has no test coverage. A few suggestions building on @gustavohstrassburger's review, plus a couple new ones inline.

**Additional Notes:**

- `question`: This PR adds roughly 8,263 lines of new benchmark tooling (open-loop Poisson scheduling, bounded executors, hdrhistogram collection, PG sampling, plan checks, gate evaluation). It's gated behind a `benchmark` subcommand so it adds no production surface, and the stated rationale (the old benchmark couldn't validate the write-amplification claim) makes sense.

Does the full six-phase gate harness need to land now to make the go/no-go call, or could the schema change ship with a thinner load test and the gate harness follow later? The README lists a `cloud-infra` CPU-guarantee follow-up as a prerequisite for trusting the latency gates. Are the gate thresholds actionable now, or are we landing ahead of the infra that makes them meaningful?
- `suggestion`: Confirmed this is safe to apply. sqlx encodes `&[&Value]` as `jsonb[]` the same as `&[Value]`, so only the collect line changes:

```suggestion
        let properties: Vec<&serde_json::Value> =
            updates.iter().map(|u| &u.properties).collect();
```

This is the only production write hot path in the crate, and `with_retry` rebuilds the array on every retry attempt, so the clone cost compounds under contention.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
matheus-vb: Checked every path that could have applied this. `migrate-entry`'s `all` case omits flags-read-store, no CI workflow runs the scope, and the mprocs process is `autostart: false`. Only a local database bootstrapped through bare `bin/migrate` can hold the old checksum, and recovery there is dropping and recreating `flags_read_store`.

I kept the in-place rewrite so every fresh bootstrap doesn't create the old GIN table just to drop it again. The README now documents the recovery, and records that in-place rewrites stop being safe once this scope joins a deploy or CI path.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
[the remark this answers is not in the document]
matheus-vb: Checked every path that could have applied this. `migrate-entry`'s `all` case omits flags-read-store, no CI workflow runs the scope, and the mprocs process is `autostart: false`. Only a local database bootstrapped through bare `bin/migrate` can hold the old checksum, and recovery there is dropping and recreating `flags_read_store`.

I kept the in-place rewrite so every fresh bootstrap doesn't create the old GIN table just to drop it again. The README now documents the recovery, and records that in-place rewrites stop being safe once this scope joins a deploy or CI path.

rust/flags_read_store_migrations/20260411000001_create_flags_read_store.sql
matheus-vb: The migration now carries a note pointing at the sweep. I'm leaving the two DDLs separate until the sweep lands: the benchmark's is parameterized per partition and per run, the migration's is fixed, and unifying them before we know the target value adds indirection for no result. The measured value will ship as a follow-up migration.

rust/flags-consumer/src/benchmark/executor.rs
matheus-vb: Intentional, and it matches production. `processor.rs` issues `batch_upsert_persons`, `batch_delete_persons`, and `batch_upsert_distinct_ids` as independent version-guarded calls with no surrounding transaction, so wrapping the benchmark's merge in one would measure a shape the consumer never runs. Every step is idempotent and version-guarded, so a replay reconverges. Added a comment saying so.

rust/flags-consumer/src/benchmark/executor.rs
matheus-vb: Changed my mind, this is in now. There's a single `run_batch_worker` parameterized over `prepare` and `execute`, with the two dedupe functions kept separate.

matheus-vb: Implemented, with one deviation: the bound needs `AsyncFn` rather than `Fn(...) -> Fut`. A single `Fut` type parameter can't express a future that borrows its arguments, so the literal signature doesn't compile. Everything else is as you wrote it, including keeping the two dedupe functions separate.

rust/flags-consumer/src/benchmark/mod.rs
matheus-vb: Down from seven `#[allow]`s to one. Added `StoreHandles`, `WorkloadContext`, and `PhaseMetrics` here, plus a `WorkerContext` in executor.rs. `resolve_arrival` went from 10 arguments to 4 and `drive_open_load` from 17 to 7. `drive_catch_up_load` keeps the allow at 10, since its three catch-up channels plus the read schedule and dispatcher don't group cleanly.

PR #65581, opened 2026-06-23
feat(hogli): add ci:preflight pre-push hook, checks, and telemetry

rnegron:
## Problem

Deterministic CI failures (lint, broken lockfiles, OpenAPI drift, migration conflicts, stale branches) only surface after a ready PR's full matrix runs, then the fix re-runs the whole matrix.

## Changes

- `hogli ci:preflight` – checks scoped to the branch diff (lockfile, `uv.lock`, ruff, markdown, feature flags, workflows, OpenAPI, migrations). `--fix` auto-remediates; checks skip cleanly when a capability (node/stack/ClickHouse) is missing.
- Risk-based staleness advisory – warns only when merging master would actually break something (merge conflicts, migration collisions, generated-file drift, CI changes), plus an aggressive-by-default behind/age backstop (env-tunable).
- Pre-push hook runs `ci:preflight --strict`: failures block, advisories don't. `HOGLI_PREFLIGHT_DISABLED=1` opts out and still emits telemetry.
- CI flake overseer now tags deterministic CI failures with `failure_class` + `pr_number`, joinable against preflight's telemetry to measure interception rate.
- Shared, hardened change-detection helper used by `build`, `test --changed`, and preflight (fixes quoted paths, renames, untracked dirs, silent git failures).

## How did you test this code?

1091 hogli-commands tests pass, including new coverage for the kill switch, strict exit contract, and staleness risk derivation. Two xhigh multi-agent code reviews, all findings fixed. Dogfooded live on this branch – staleness correctly flagged real risk instead of a raw commit count, and `--strict` blocked on an induced lint failure and passed clean otherwise.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

No customer-facing docs. Internal dev tooling only – `AGENTS.md` and the `running-ci-preflight` skill are updated in this PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

I (Claude Code) built this with Raúl. Skills invoked: `/writing-tests`, `/writing-skills`, `/code-review` (xhigh, twice).

A few decisions worth flagging for review:
- The pre-push hook is the enforcement point; the command is the engine underneath it. That keeps `--fix` and hook-less sandboxes (no `node_modules`, no husky) working without a second code path.
- Staleness is risk-based rather than a raw behind-count, since a plain count fires on nearly every branch at master's velocity and trains people to ignore it.
[...]

PR #72749, opened 2026-07-21
feat(autoresearch): push ruff format fixes to PRs red only on quality checks

DanielVisca:
## Problem

Classifying recent failed Python CI runs by cause: 37.5% failed on pure `ruff format` drift, 50% on mypy/type errors, the rest mixed. The format share is a red a robot can fix deterministically — today it costs the author a 15–20 minute CI round trip (notice the red, run the formatter, push, wait again) for a change with exactly one correct answer. Most of these reds originate in environments without the pre-push preflight hook.

Created with Autoresearch.

## Changes

One new `workflow_run` workflow. When Python CI completes a PR run red on attempt 1 and the **only** failed jobs are the code-quality jobs, it checks out the PR branch, runs `ruff format` at the version locked in `uv.lock`, and pushes the fix commit as posthog-bot.

Safety properties:

- **Nothing is masked.** Lint and type errors produce no diff from `ruff format`; the job logs a notice and leaves the red for the author. Runs that failed any other job are never touched.
- **No untrusted code execution.** Same-repo PR branches only (fork PRs are skipped at the job `if`), and the only thing executed against the checkout is `ruff format`, which reads declarative config and runs no repository code — no `pnpm install`, no repo scripts.
- **CI re-triggers on the fix.** The push uses `POSTHOG_BOT_PAT`; pushes made with `GITHUB_TOKEN` dispatch no workflows and would strand the PR on a SHA with no checks. If the secret is absent (forks, the FOSS mirror) the job no-ops with a notice instead of failing.
- **Loop-safe.** `run_attempt == 1` plus idempotent formatting at a fixed version: a second pass over the bot's own commit produces no diff and exits. A freshness check skips the push when the branch has moved past the failing SHA, and a rejected push downgrades to a notice.

> [!NOTE]
> Expected effect on the measured hands-off green rate (baseline 58%): +4.3 points from the Python format share (37.5% of a 10.7% fail rate) plus, later, the Frontend formatting share if this pattern proves out (oxfmt needs node_modules, so it's deliberately not in v1). Tracking lives on the [CI self-driving health dashboard](https://us.posthog.com/project/2/dashboard/1884671).

> [!WARNING]
> Reviewers should confirm `POSTHOG_BOT_PAT` has push rights to same-repo PR branches. If it doesn't, the workflow degrades to a no-op notice rather than failing, but the feature simply won't act until the credential is right.

`bin/hogli lint:workflows` passes. The `workflow_run` trigger can't be exercised pre-merge (it runs from the default branch's copy), so validation after merge is: find a PR red only on Python code quality, confirm the bot commit lands and Python CI re-runs green on it. The gate logic (only-quality-jobs-failed) uses the same jobs API I used to classify the failure data, and the ruff invocation matches the version CI itself installs via `uv.lock`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, CI-infrastructure-only.

## 🤖 Agent context

- Authored by Claude (PostHog Code) in an autoresearch session on CI reliability directed by Daniel Visca; the `/authoring-ci-workflows` skill was invoked (SHA-pinned checkout, job-level guards, timeout, per-branch concurrency with cancel-in-progress for superseded pushes).
- The 37.5%/50% cause split comes from grepping the failure regions of sampled failed quality-job logs (ruff "would reformat" vs mypy error signatures).
- Decisions: `pip install ruff==<locked>` over `setup-uv` (preinstalled pip, one fewer pinned action, version still exact from `uv.lock`); Python-only v1 because oxfmt requires installing node_modules on a checked-out PR branch, which crosses the no-repo-code-execution line this PR deliberately holds; `ruff format` only, not `ruff check --fix`, to keep the bot's changes semantics-free.

---

### DIFF
--- a/.github/workflows/ci-python-format-autofix.yml
+++ b/.github/workflows/ci-python-format-autofix.yml
@@ -1,0 +1,114 @@
+name: CI Python Format Autofix
+
+# When Python CI fails a PR and the ONLY failed jobs are the code-quality jobs,
+# try `ruff format` at the repo-locked version and push the fix commit to the
+# PR branch. Measured on this repo (2026-07): 37.5% of Python CI reds are pure
+# format drift — a red a robot can fix. Lint and type errors produce no diff
+# from `ruff format`, so they stay red for the author; nothing is masked.
+#
+# The push uses POSTHOG_BOT_PAT so the fix commit re-triggers CI (pushes made
+# with GITHUB_TOKEN dispatch no workflows and would strand the PR on a SHA with
+# no checks). Without that secret (forks of this repo, the FOSS mirror) the job
+# no-ops with a notice instead of failing. Same-repo PR branches only; the only
+# thing executed from the checked-out branch is `ruff format`, which reads
+# declarative config and runs no repository code.
+#
+# Loop safety: ruff format is idempotent at a fixed version, so a second pass
+# over the bot's own commit produces no diff and the job exits without pushing.
+
+on:
+    workflow_run:
+        workflows: ['Python CI']
+        types: [completed]
+
+permissions:
+    actions: read
+    contents: read
+
+concurrency:
+    group: python-format-autofix-${{ github.event.workflow_run.head_branch }}
+    cancel-in-progress: true
+
+jobs:
+    autofix:
+        name: Push ruff format fix
+        runs-on: ubuntu-latest
+        timeout-minutes: 10
+        if: >
+            github.repository == 'PostHog/posthog' &&
+            github.event.workflow_run.conclusion == 'failure' &&
+            github.event.workflow_run.event == 'pull_request' &&
+            github.event.workflow_run.run_attempt == 1 &&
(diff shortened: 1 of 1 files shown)


rnegron (changes requested):
👋🏻 Hey @DanielVisca, thanks for the PR! but hmm I'm not sure if this is the direction we want to take with CI in general; we are hoping to offload stuff locally or tweak the configurations rather instead of throwing more stuff at it. some examples:

* enforce more local checks instead of relying on slow, overcrowded CI (like with https://github.com/PostHog/posthog/pull/65581, which now tries to recommends local type checking as well: https://github.com/PostHog/posthog/pull/71328)
* use a merge queue solution (cc @gantoine) to prevent the common case of a stale PR merging and breaking types

happy to discuss this further if you feel strongly about it!

DanielVisca:
Thanks for checking and I appreciate the context! I'll close this one out :+1:

PR #71460, opened 2026-07-16
chore(frontend): inline Kea types batch 6

mariusandra:
## Problem

The all-at-once Kea inline type migration is too large to merge safely while logic files keep changing.

## Changes

- Convert one balanced slice of Kea logics to inline `MakeLogicType` declarations.
- Replace this batch’s reserved `.kearc` slots with its owned folders.
- Keep this PR independent from the other conversion batches.

> [!NOTE]
> Merge foundation PR #71446 first. This PR targets `master` rather than another batch, so sibling merges do not trigger stacked-PR restacks. Its shared foundation diff disappears after #71446 lands.

## How did you test this code?

- Generated from a single clean `kea-typegen write --delete` run that finished with all 1,040 logic types up to date.
- Verified this branch contains only its assigned generated slice and reserved `.kearc` block.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

The shared contributor guidance is in #71446. No user-facing docs change is needed.

## 🤖 Agent context

Codex generated the full migration once on current `master`, partitioned changed files into nine balanced batches, and used preallocated `.kearc` slots to avoid sibling config conflicts.

**Autonomy:** Human-driven (agent-assisted)

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #72760, opened 2026-07-21
chore(toolbar): drop lowlight all-grammar re-export and posthog-typed typings from bundle graph

haacked:
## Problem

The toolbar module-boundary graph check (`frontend/bin/check-toolbar-graph.mjs`) budget has been raised five times in the last week by unrelated PRs (#71460, #71808, #72041, #69730, #72073, and [#72459](https://github.com/PostHog/posthog/pull/72459)), because the toolbar's input graph sat within ~50 KB of the 14.5 MB budget. Analysis of the esbuild metafile showed two large avoidable contributors:

- `lowlight`'s package index re-exports `all` – every highlight.js grammar, 192 files and ~1.2 MB of source – even though our code only uses the `common` set (37 grammars).
- `lib/posthog-typed` is ~200 KB of generated event-name typings wrapped around a tiny runtime, reachable from the toolbar via keyboard-shortcut hooks.

Neither ships extra bytes to customers (tree-shaking and lazy chunks handle output size), but both count fully against the input-graph budget, eating headroom that unrelated PRs then have to bump.

## Changes

- Toolbar build (`frontend/toolbar-config.mjs`): resolve `lowlight` to a pared-down module exposing only `createLowlight` and `common`, bypassing the package index's `all` re-export. Grammar files in the graph drop from 192 to 37.
- Add `src/toolbar/shims/posthogTyped.ts`, a runtime-equivalent shim for `lib/posthog-typed` (capture/captureRaw delegating to the posthog-js singleton, everything else proxied), and register it in the toolbar's shim map.

Result: toolbar input graph goes from **13.78 MiB (978 files) to 12.45 MiB (822 files)**, restoring ~2 MB of headroom under the 14.5 MB budget. Crossing edges (5), eager output size, and CSP checks are unchanged.

- `node frontend/bin/build-toolbar.mjs` then all three toolbar checks pass: graph (12.45 MiB, 5 baseline edges, no forbidden packages), size (eager 1.55 MB under 1.66 MB budget, unchanged), CSP eval.
- `pnpm --filter=@posthog/frontend typescript:check` passes.
- Added a parity test in `shims.test.ts` that fails if the generated `lib/posthog-typed` module grows runtime surface the shim lacks – the regression this catches is toolbar code calling a new generated method that silently resolves to `undefined` on customer pages. Also two `test.each` cases asserting capture/captureRaw delegate to `posthog-js` capture.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed – build-internal change, no user-facing behavior.

## 🤖 Agent context

Investigated why [#72459](https://github.com/PostHog/posthog/pull/72459) needed a toolbar budget bump: the metric counts pre-tree-shake source bytes, so SDK bumps and shared-file churn trip it without shipping bytes. Ranked contributors from the esbuild metafile; lowlight's `all` re-export and the generated posthog-typed module were the two biggest avoidable ones. Considered also decoupling `~/types` from the graph (17 value importers) but deferred it as a larger structural change. Invoked /writing-tests before adding the shim parity test. Note: #72583 rewrites the graph check to boundary-only; this PR doesn't touch the budget constant so they compose cleanly.

---

### DIFF
--- a/frontend/toolbar-config.mjs
+++ b/frontend/toolbar-config.mjs
@@ -1,7 +1,10 @@
 import * as fs from 'fs'
+import { createRequire } from 'module'
 import * as path from 'path'
 
 import { commonConfig, copyRRWebWorkerFiles, createHashlessEntrypoints, esbuildBuild, isDev } from '@posthog/esbuilder'
+
+const require = createRequire(import.meta.url)
 
 // `TOOLBAR_PUBLIC_PATH`, when set, overrides the default `publicPath` so the
 // toolbar bundle and its assets can be hosted under a versioned, content-pinned
@@ -34,6 +37,9 @@ const shimmedModules = {
     // app (TZLabel, Link, HeatmapEventsPanel), whose scenes/urls import would otherwise pull
     // every product manifest into the bundle.
     'scenes/urls': 'src/toolbar/urls.ts',
+    // Not a kea shim: the generated event typings are ~200 KB of source the toolbar must
+    // not bundle. The shim mirrors the module's small runtime (capture/captureRaw proxy).
+    'lib/posthog-typed': 'src/toolbar/shims/posthogTyped.ts',
 }
 
 // Modules replaced with an inert proxy that logs access in debug mode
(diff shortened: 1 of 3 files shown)


pauldambra:
👋 i'm removing the input budget here https://github.com/PostHog/posthog/pull/72583
i don't think we need to track a budget of stuff we're going to tree-shake away

pauldambra:
👋 i'm removing the input budget here https://github.com/PostHog/posthog/pull/72583
i don't think we need to track a budget of stuff we're going to tree-shake away

pauldambra:
👋 i'm removing the input budget here https://github.com/PostHog/posthog/pull/72583
i don't think we need to track a budget of stuff we're going to tree-shake away

pauldambra:
CleanShot 2026-07-22 at 00 06 54@2x

my first time using a computer 🫠

frontend/toolbar-config.mjs
```diff
@@ -80,9 +86,28 @@ function createToolbarModulePlugin(dirname) {
             path.resolve(dirname, shimFile),
         ])
     )
+    // lowlight's package index re-exports `all` — every highlight.js grammar, ~1.2 MB of
+    // source — alongside the `common` set our code actually uses. The re-export alone pulls
+    // all 192 grammars into the bundle graph, so resolve 'lowlight' to a pared-down module
+    // instead. Its relative imports bypass the package's exports map (which hides lib/).
+    const lowlightDir = path.dirname(require.resolve('lowlight'))
```
rafaeelaudibert, line 93: The best question here is: why is lowlight being pulled here at all? Can we simply load lowlight lazily wherever it's being loaded? This way it will not be added to the same bundle as the toolbar

haacked: Good news: lowlight is already lazy. The product tours code loads it through a dynamic import, so `esbuild` splits the whole `tiptap`/`lowlight` chain into its own chunk (~662 KB) that only downloads when someone actually saves or previews a tour step in the toolbar. It never ships on page load.

I also measured what this PR actually saves in shipped bytes: 90 bytes eager, 0 deferred. The entire 1.33 MB "win" was in the pre-tree-shake input metric, which is exactly what #72583 deletes. Tree shaking was already throwing all of this away.

So this PR trades a shim, a parity test, and a build hack for 90 bytes. Not worth it. Closing in favor of #72583.

PR #72766, opened 2026-07-21
perf(storybook): make per-snapshot settle wait configurable, default 250ms

pauldambra:
## Problem

`common/storybook/.storybook/test-runner.ts` takes one Storybook visual-regression snapshot per story x per theme (light and dark) x per viewport-width variant. Right before every capture, `takeSnapshotWithTheme` ran a fixed `await page.waitForTimeout(1000)` commented "final wait for any remaining renders".

That flat 1s fires *after* several stronger readiness gates have already passed: `waitForPageReady` (domcontentloaded/load/networkidle plus explicit font loads), the all-images-have-width `waitForFunction`, the iframe-loaded wait, and a DOM-stability `waitForFunction` that requires 3 consecutive stable width/height checks at 100ms intervals. On top of all that, the extra second is mostly dead time.

**Why this matters:** Jest runs with `--maxWorkers=1`, so nothing overlaps this sleep. `frontend/snapshots.yml` lists ~4,100 snapshot identifiers, so at 1s each this single line is roughly 68 minutes of serial sleeping across a full run, and about 4 minutes of each 16-shard job's ~9-minute test step. It's the single largest time sink in the visual-regression shards, and none of it is buying readiness the gates above don't already provide.

## Changes

- Added a `snapshotSettleTimeMs?: number` option to the `testOptions` block of the `Parameters` interface, documented with `@default 250`, matching the neighboring option docblocks.
- Replaced the fixed `await page.waitForTimeout(1000)` in `takeSnapshotWithTheme` with the story-configurable value read from `storyContext.parameters?.testOptions`, defaulting to 250ms.

`takeSnapshotWithTheme` is the single capture site, reached by both the multi-viewport loop and the single-viewport path in `postVisit` (via `expectStoryToMatchSnapshot` for light and dark), so this one edit covers every snapshot path. No story is pre-opted into a higher value on purpose, so the new default is exercised suite-wide.

Nothing else changed: the 300ms ResizeObserver wait, the iframe post-load `waitForTimeout(1000)`, and all timeouts and other waits are untouched.

> [!NOTE]
> This PR is its own experiment. Its Visual Review run renders the entire suite at the new 250ms default. Unexpected diffs there are stories that genuinely need more settle time, and each can be opted back up individually via `snapshotSettleTimeMs` rather than penalizing the whole suite. I'd suggest re-running the suite once or twice before merging to probe for nondeterminism, and watching flake signals after merge.

I (Claude) verified by careful reading of the diff in full context. A full frontend typecheck was infeasible here: this worktree has no `node_modules` and the task explicitly ruled out a monorepo `pnpm install`, so `typescript:check` and `hogli ci:preflight` were **not run**. The change is a small typed addition (an optional numeric field plus a defaulted destructure of it) with no other call-site changes. The real behavioral test is the Visual Review run this PR triggers.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover this internal test-runner knob, so nothing to update.

## 🤖 Agent context

Authored by Claude Code (Opus 4.8). No skills were applicable — this touches only the Storybook test-runner config, not a DRF endpoint, migration, generated API type, or test suite, so none of the mandatory-skill triggers fired.

The change is deliberately minimal: one new `testOptions` field and one edit site. I confirmed `takeSnapshotWithTheme` is the only place the old fixed 1s wait lived and that both the multi-viewport and single-viewport snapshot paths flow through it, so a single edit plumbs the option through everywhere. The default was chosen as 250ms (down from 1000ms) on the reasoning that the preceding readiness gates already establish page-readiness; the remaining settle is a small cushion for late async paints, tunable upward per-story when a story proves it needs it.

CI facts cited above (per-snapshot cost, snapshot count, shard step durations) are all observable from public GitHub Actions runs and the repo's `frontend/snapshots.yml`. No analytics-derived aggregates or internal data.

---


pauldambra:
Shepherd update: the Visual Review run flagged **3 changed snapshots**, and all three are async session-replay player stories that genuinely settle slower than the new 250ms default — exactly the case this PR anticipated:

- `components-playerinspector--with-logs-filter--dark` (9.6% diff, SSIM ~0.54) and `components-playerinspector--with-logs-filter-upsell--light` (7.0%, SSIM ~0.66) — the story loads a large recording and renders the async event/log list; at 250ms the list hasn't finished settling. Only one theme changed per story, the signature of sitting right at the settle boundary (`Default` loads the same data and just got lucky this run).
- `replay-player-controller--wide--light` (0.8%) — `mockDate` freezes time, so this isn't clock drift; it's the seekbar buffer bar finishing paint after the recording loads.

Rather than approve these as a new baseline, I dogfooded the new knob and opted just these stories back to their prior 1s settle via `snapshotSettleTimeMs: 1000` (set at the `PlayerInspector` and `PlayerController` meta level so both themes and all width variants are covered). The rest of the suite keeps the fast 250ms default. Pushed as b260e4d.

This push re-runs the whole suite at 250ms, which doubles as one more nondeterminism probe.

pauldambra:
Shepherd update #2 — a genuine finding worth a decision.

Editing the story files pulled them into the **`Storybook flake verification`** gate (it runs only on *changed* `*.stories.tsx`, re-rendering each 3×). That gate surfaced that **`components-playerinspector--with-logs-filter--light` renders ~36% differently between runs (89k pixels) at a 1s settle** — i.e. run 1 passed, run 2 failed hard, run 3 passed. So this story is **nondeterministic run-to-run; settle time is not the lever.** The runner already stops animations, forces `content-visibility`, nudges ResizeObserver, and waits on loaders, so this is almost certainly the virtualized async event/log list settling to a different scroll position / row set (the Logs-tab variants add an extra async fetch vs `Default`, which is why only they flake).

What I changed in response:
- **Kept** the `PlayerController` bump (`snapshotSettleTimeMs: 1000`). It's a real, *deterministic* slow-paint case (seekbar buffer bar) and passed all 3 flake-verification runs — a clean dogfood of the new knob. (d7215e6 keeps this.)
- **Reverted** the `PlayerInspector` bump. It was premised on "needs more time," which is false. Reverting also removes it from the flake gate. Its 250ms diffs now go through the normal Visual Review approval flow.

This is a pre-existing latent flake this PR merely exposed — **not caused by the 250ms default** (it flakes at 1000ms too). It shouldn't gate this perf change. Options for handling it separately (or here, if you prefer):
1. **Skip its snapshot** — mark `WithLogsFilter`/`WithLogsFilterUpsell` `test-skip` and drop the 4 orphaned baseline PNGs. Removes ongoing flake noise; loses (currently unreliable) coverage of that story.
2. **Make it deterministic** — pin the list scroll/data so it renders identically each run, then it can stay snapshotted at the fast default.
3. **Accept + monitor** — approve its Visual Review diffs and watch flake signals post-merge (matches the original PR plan).

I'd lean 1 or 2 as a small follow-up. Happy to do either — let me know.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 1 @ ca9f6f5)</sub>

The focused Storybook test-infrastructure change preserves the old wait as an explicit story override while reducing the suite-wide default. The parameter lookup preserves `0` as a valid override, and the measured shard runtimes confirm the intended performance effect. No actionable correctness, security, or maintainability findings were identified.

### Key findings

- None.

### Convergence

- No delegated review was needed for this low-danger, two-file change.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | LOW danger, HIGH confidence; reviewed the full diff and surrounding snapshot-runner/story code, with no delegation warranted. |

<details>
<summary>Previous rounds (0)</summary>

None.

</details>

---

PR #72769, opened 2026-07-21
feat(signals): tag eval-fixture generation as signals_eval

brandonleung:
## Problem

Signals routes its LLM calls through the ai-gateway per stage, tagged by `ai_product`. The eval-fixture generator (`products/signals/eval/llm_gen`) calls the same `call_llm` helper but passed no `ai_product`, so its spend fell into the untagged catch-all instead of the signals books. It was the last signals `call_llm` site without a tag.

## Changes

- `generate_canonical_signals` passes `ai_product="signals_eval"`, the same tag the eval summary stage already uses, so both eval stages bill together.

## Verification

`test_llm.py` pins the tag and stage for both `signals_eval` call sites; dropping either kwarg fails its test. `ruff check` and `ruff format` clean.

`ai_product` is also the opt-in switch for the Go gateway, so wherever `AI_GATEWAY_URL` is configured this moves eval-fixture generation off the Python gateway alongside the other tagged signals stages. Reverting the kwarg reverts both.

## Note

This is eval-harness traffic (fixture generation), not the production pipeline. Tagging it keeps eval spend visible under signals; if eval cost should sit outside the signals product entirely, this is the line to revisit.

Refs PostHog/ai-gateway#357.

### DIFF
--- a/products/signals/backend/test/test_llm.py
+++ b/products/signals/backend/test/test_llm.py
@@ -3,7 +3,13 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 from django.test import override_settings
 
+from products.signals.backend.temporal.emit_eval_signal import (
+    EmitEvalSignalInputs,
+    EvalSignalSummary,
+    summarize_eval_for_signal,
+)
 from products.signals.backend.temporal.llm import call_llm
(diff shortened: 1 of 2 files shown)


richardsolomou (approved):
nit: could we add a focused test that checks `generate_canonical_signals` passes `ai_product="signals_eval"`? nothing currently protects it from silently disappearing later.

PR #72770, opened 2026-07-21
feat(tasks): route selected sandbox products to the ai-gateway

brandonleung:
## Problem

Signals' sandbox stages (scout, research, implementation, repo_selection) make model calls through the legacy Python `llm-gateway`. To move them to the new Go `ai-gateway`, the sandbox needs the gateway URL and the list of products allowed to use it. Neither is plumbed in today.

## Changes

Two new settings: **`SANDBOX_AI_GATEWAY_URL`** and **`SANDBOX_AI_GATEWAY_PRODUCTS`**. When both are set, they're injected into the sandbox as `AI_GATEWAY_URL` / `AI_GATEWAY_PRODUCTS`, and the agent-server (PostHog/code#3659) routes the listed products to the new gateway. Everything else stays on the Python gateway.

The sandbox blocks unknown env vars and network egress by default, so most of the diff is **allowlisting**:

- The two keys are **reserved** (user-supplied env can't override them) and added to the agentsh env allowlist. Without the allowlist entry, agentsh strips them and the product silently stays on the old gateway.
- The gateway hostname is added to the **network allowlist** in both layers (agentsh and Modal): static entries for `ai-gateway.us/eu.posthog.com`, plus the hostname parsed from the URL setting. The parsed entry covers dev's `ai-gateway.dev.posthog.dev`.
- Setting values are **validated as DNS names** via Django's `DomainNameValidator`, so wildcards, IPs, and malformed labels narrow the policy rather than widening it. Misconfigurations log a warning.
- The session-init **egress probe** now checks the gateway hosts, so a blocked gateway shows up by name in startup diagnostics.

Unit tests pin the env and network allowlists per setting, the hostname validation, the reserved-key guard, and that injection only happens when both settings are set. `ruff` and repo-wide `mypy` clean.

**Inert until** the sandbox chart sets both values; PostHog/code#3659 must deploy first. Rollback: clear either setting.

Refs PostHog/ai-gateway#357.


products/tasks/backend/logic/services/agentsh.py
```diff
@@ -22,6 +22,8 @@
     "api.anthropic.com",
     "gateway.us.posthog.com",
     "gateway.eu.posthog.com",
+    "ai-gateway.us.posthog.com",
+    "ai-gateway.eu.posthog.com",
```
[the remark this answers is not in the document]
Twixes: I think that's expected

products/tasks/backend/logic/services/agentsh.py
```diff
@@ -55,26 +62,102 @@ def _port_from_url(url: str | None) -> int | None:
...
     "SANDBOX_MCP_URL",
 )
 
+_LOOPBACK_ALIASES = ("localhost", "host.docker.internal")
+
+# The allow rule is an enforcement gate: a malformed value must narrow it, never
+# widen it, so anything outside a plain DNS-name charset (notably `*`, which is
+# wildcard syntax at both enforcement layers) is rejected rather than passed
+# through.
+_HOSTNAME_RE = re.compile(r"^[a-z0-9]([a-z0-9.-]*[a-z0-9])?$")
```
richardsolomou: couldnt we reuse Django's `DomainNameValidator` here? this regex accepts `foo..example.com` and `-gateway.example.com`, so a typo can reach Modal's allowlist and fail sandbox creation instead of taking the warning path.

richardsolomou (commented):
could you also add one activity-level case for `get_sandbox_for_repository` as well? deleting the `environment_variables.update(ai_gateway_env_vars())` wiring there still leaves the helper and Modal provisioning tests green.

products/tasks/backend/logic/services/agentsh.py
brandonleung: done

andrewm4894 (approved):

richardsolomou (approved):

PR #72772, opened 2026-07-21
fix(plain): document company:read as a required Plain API scope

Gilbert09:
## Problem

PostHog's error tracking flagged a recurring non-retryable 403 from the `plain` data warehouse source: `Insufficient permissions, missing "company:read"`.

The `customers` GraphQL query reads a `company { id name }` field, which requires Plain's `company:read` scope. But the source's setup caption (the instructions shown when connecting Plain) and the 403 error guidance never mentioned that scope — only `customer:read`, `thread:read`, `timeline:read`, `user:read`, and `label:read`. A user who granted exactly the scopes we told them to would still be missing `company:read`, and Plain fails the whole GraphQL request when any requested field's scope is missing, so the sync fails permanently with no way to self-serve a fix from our own docs.

## Changes

- Added `company:read` to the required-permissions list in the Plain source setup caption.
- Added `company:read` to the 403 non-retryable error message that tells users which scopes to grant.

Ran the existing Plain source test suite, including `test_source_config_documents_scopes_the_queries_require`, which now asserts `company:read` appears in both the caption and the forbidden-error message:

```
uv run pytest products/warehouse_sources/backend/temporal/data_imports/sources/plain/tests/ -q
# 41 passed
```

Also ran `ruff check` and `ruff format --check` on the touched files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A — no separate docs page for this source beyond the in-app setup caption, which this PR updates directly.

## 🤖 Agent context

This PR was generated in response to a PostHog error tracking alert for the `plain` warehouse source (403 Forbidden, missing `company:read`). I (Claude Code) traced the error to `execute()` in `plain.py`, confirmed the 403 was already marked non-retryable in `source.py`, then dug into why users kept hitting it despite that — the `customers` query requests a `company` field that needs a scope our own setup instructions never listed. Fixed by adding `company:read` everywhere the other required scopes are documented, and updated the existing scope-sync test to cover it.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/plain/source.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/plain/source.py
@@ -57,6 +57,7 @@ You can create an API key in your [Plain workspace settings](https://app.plain.c
 
 Make sure to grant the following read permissions:
 - customer:read
+- company:read
 - thread:read
 - timeline:read
 - user:read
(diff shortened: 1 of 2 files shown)


Gilbert09:
This PR has been fully superseded by #72773 ("fix(plain): document company:read and machineUser:read scopes"), which merged into `master` earlier today.

#72773 makes the same fix as this PR — documenting `company:read` in both the Plain source setup caption and the 403 non-retryable error guidance — and goes a step further by also adding `machineUser:read` (and refreshing the accompanying test/comment to mention the `timeline_entries` query). It is a strict superset of the change here.

As a result, rebasing this branch onto the current `master` resolves the conflict to an **empty diff**: every line this PR wanted to add already exists on `master`, so there is nothing left for it to contribute and nothing to merge.

Recommendation: close this PR as superseded by #72773. No further code change is needed — the `company:read` scope is already documented on `master`.

Gilbert09:
Closing as superseded. This PR added `company:read` to the Plain source setup caption, the 403 non-retryable error message, and the scope-sync test. That same fix has since landed on `master` via #72773, which documents `company:read` **and** `machineUser:read` across all three of those spots.

After rebasing this branch onto the current `master`, the PR commit is empty — there is no remaining diff to merge. Rather than force-push an empty branch (which would trigger a full CI fan-out for a no-op), I'm closing this as fully covered by #72773.

PR #72791, opened 2026-07-22
feat(conversations): add macros for support replies

HaynesPostHog:
## Problem

Support agents answer the same questions constantly and retype the same replies every time. Macros are a top request for the Conversations product: save a reply once, drop it into a conversation in one action, and optionally apply the routine ticket changes (set status, bump priority, tag) that go with it.

## Changes

Adds **macros** to Conversations: reusable saved replies an agent inserts into the composer.

- **Two ways to insert:** type `/` in the reply box for a searchable, keyboard-navigable picker, or click the macro button in the composer toolbar. Both share one picker component.
- **Ticket actions:** a macro can carry `status`, `priority`, `tags`, and `assignee`. When inserted, those apply to the ticket through the existing scene-logic actions.
- **Dynamic variables:** `{{customer.name}}`, `{{ticket.number}}`, `{{agent.name}}`, `{{agent.first_name}}` are filled from the ticket at insert time, client-side. Unknown tokens resolve to empty so no raw `{{...}}` reaches the customer.
- **Visibility:** macros are either team-shared or personal to their creator (enforced server-side in the queryset).
- **Management UI:** a new Macros tab under Support settings to create, edit, and delete macros, with the reply body, visibility, and action fields.

Backend is a new team-scoped `Macro` model (fail-closed `TeamScopedRootMixin`), a DRF viewset modeled on the existing saved-views viewset, and an additive migration. The reply is stored as both markdown `content` and TipTap `rich_content`, mirroring how ticket messages are stored.

> [!NOTE]
> No third-party tool is referenced anywhere in the code or copy. The `/` slash-command extension is modeled on the existing `MentionsExtension`.

Automated tests I (Claude, driven by @HaynesPostHog) actually ran locally, all green:

- **Backend** `products/conversations/backend/api/tests/test_macros.py` (3 tests): visibility scoping (agent A can't see agent B's personal macro; both see team macros), team/creator assignment on create, and the typed-actions allow-list stripping unknown keys.
- **Frontend Jest** (8 tests): `applyMacroVariables` substitution (incl. unknown-token blanking and no-mutation), and `macroToDoc` — which guards a bug found during manual testing where a content-only macro (e.g. API-created) rendered blank in the edit modal and would wipe the reply on save.
- `pnpm typescript:check`, `ruff`, `ruff format`, and `mypy` clean on all touched files.

I also drove the live API on a local stack (create/list/retrieve/patch/delete, variable round-trip, allow-list) and clicked through the composer `/` picker, the toolbar button, and the settings UI against a seeded ticket.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — @HaynesPostHog directed the work; I (Claude, via PostHog Code) implemented it.

Skills invoked while producing this PR: `/django-migrations`, `/improving-drf-endpoints`, `/adopting-generated-api-types`, `/writing-tests`, `/run-posthog`.

- **Enum naming:** the macro-action `priority` and `visibility` fields collided with existing enums during OpenAPI generation. I pinned two `ENUM_NAME_OVERRIDES` in `posthog/settings/web.py` (`SupportPriorityEnum` for the shared conversations priority, and keeping `VisibilityEnum` pinned to `ColumnConfiguration.Visibility`) so the new fields don't rename unrelated products' generated types.
- **`@tiptap/suggestion`** was added as a peer dependency of the conversations product so the slash-command extension resolves; the lockfile change is minimal and `--frozen-lockfile` passes.
- **Variable substitution is client-side** at insert time (not server-rendered) — kept the backend a plain store.

This branch is behind master; CI runs against the merge, and I left it un-rebased to avoid pulling ~50 commits into a draft.


HaynesPostHog:
Superseded by #72792, pushed to the upstream repo so ReviewHog can review it (it skips fork PRs).

PR #72797, opened 2026-07-22
fix(data-warehouse): pin legacy table names when provisioning/onboarding teams

EDsCODE:
## Problem

A newly provisioned/onboarded managed-warehouse team gets a Duckgres team row carrying only `team_id` + `schema_name` — describing the derived `<schema>.events` layout. But the duckling DAG still writes the legacy layout (`posthog.events_<suffix>`, `posthog_data_imports_<suffix>`). Since the project reader's SQL-editor grants derive from the row (PostHog/posthog#72368), every fresh warehouse ships with a connection that authenticates but sees an empty catalog, until someone hand-runs the grandfather upsert — exactly the incident the EU placeholder rows (teams 1 / 43283 / 69647) hit, where readers were granted three nonexistent schemas and the sidebar stayed empty. The one-time US backfill masked this for existing US teams; nothing covered new provisions.

## Changes

- `onboard_team` now passes the legacy table names (via the existing `_grandfathered_team_fields`, suffix == the chosen schema name) to `create_team`, so a newly onboarded team's row describes where its data actually lands.
- `provision` follows the successful org-provision with a best-effort `create_team` upsert carrying the same fields for the first (billing) team — the provision request body cannot carry them. It runs **before** `_register_provisioning_team`, so the reader handshake that registration triggers derives its grants from the completed row. A completion failure logs a warning and never fails the provision (the warehouse is already up; re-onboarding or the grandfather upsert completes the row later).
- Sunset note in both places: drop the pinning when the duckling DAG starts writing the derived `<schema_name>.events` layout for real.

- TDD: updated `test_onboard_team_dual_writes_duckgres_and_django` and `test_provision_sends_team_id_and_schema_name_to_control_plane` to expect the legacy fields, watched both fail, then implemented; added `test_provision_succeeds_even_when_the_team_row_completion_fails` for the best-effort path.
- `bin/hogli test products/data_warehouse/backend/tests/api/test_managed_warehouse.py products/data_warehouse/backend/tests/test_managed_warehouse_connection.py posthog/ducklake/test_common.py`: 127 passed.
- `bin/hogli ci:preflight`: ruff/format/ty clean on the changed files (the one advisory is an untracked local notes file, not committed).

## Reviewer notes

- This encodes "duckling DAG writes legacy layout for new teams" as fact. Evidence: org created via this flow recently (team 473662) landed `posthog.events_test` + `posthog_data_imports_test`. If any cohort of new teams is already on the derived layout, the pinning would mis-describe them — flag if so.
- Discovery consumers see resolved names change for newly onboarded teams (`<schema>.events` → `<schema>.events_<suffix>`); this matches what the US backfill already did fleet-wide for existing teams.

## 🤖 Agent context

Authored by Claude Code (Fable 5) as the follow-up to the EU empty-sidebar investigation: EU rows were unfilled placeholders; the fix for existing rows was a data backfill, and this PR closes the same gap structurally for every future provision/onboard so no manual stitching remains.


bill-ph (approved):

EDsCODE:
Relevance re-checked after today's duckgres team-state work (#72598, #72846, #72863) — no overlap: those add CP read-routing, the earliest-event-date dual-write, and the admin `update_team` repair endpoint, while `onboard_team`/`provision` on master still write rows without legacy names, so this PR's gap (and fix) is unchanged. Verified on the branch with master merged in: both changesets coexist, 130 tests passing including the new team-state ones.

PR #72801, opened 2026-07-22
fix(surveys): scope embedded actions to the serializer's fields

Piccirello:
## Problem

`get_survey_conditions_with_actions` re-serializes a survey's linked actions from the `actions` M2M when it's populated. But surveys also store an `actions` blob inside `conditions`, and the M2M is set from that blob's IDs. When those IDs don't resolve (e.g. the referenced action was deleted), the M2M ends up empty while the stale blob remains — and the function returned that blob verbatim.

For the public `SurveyAPISerializer` (served on `/api/surveys` and the `/decide` config), that meant the payload could carry fields the serializer never declares, because the stored blob can hold a full action object from an editor round-trip rather than the minimal public shape.

## Changes

- In the M2M-empty fallback, project each stored action value down to the caller serializer's own `Meta.fields`. The public serializer exposes `id`/`name`/`steps`, so the output now matches the serialized shape regardless of what the stale blob held. The authenticated path is unchanged.

Automated. I'm an agent and did not do manual testing.

- Added a DB-free unit test on `get_survey_conditions_with_actions` (`TestGetSurveyConditionsActionSanitization`). Verified it **fails before the fix** (the stale blob's extra fields come through) and **passes after**.
- Added a DB-backed test on the real `SurveyAPISerializer` for a stale-blob survey (`TestSurveyWithActions`). This one I could not run locally — the dev ClickHouse was in a readonly/timeout state — so it's left for CI to confirm.

Regression caught: a survey whose action M2M is empty but whose `conditions` still holds a stored action blob. No existing test exercised that fallback path.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Tom directed this. Authored by Claude Code (Opus 4.8). Skills invoked: `/improving-drf-endpoints` (serializer change) and `/writing-tests` (test value gate).

I traced the fallback to `_associate_actions`, which sets the M2M from the blob's IDs but doesn't strip the blob from `conditions`, so an unresolved ID leaves a stale blob behind. Chose to sanitize at the single read chokepoint (`get_survey_conditions_with_actions`) by projecting to the caller serializer's declared fields, rather than rewriting the stored data or the write path — it's minimal and correct for both the public and authenticated callers. Test-first: unit test failing, applied the fix, green.

### DIFF
--- a/products/surveys/backend/api/survey.py
+++ b/products/surveys/backend/api/survey.py
@@ -299,6 +299,22 @@ def get_survey_conditions_with_actions(
         else:
             conditions = dict(conditions)
         conditions["actions"] = {"values": action_serializer_class(actions, many=True).data}
+    elif isinstance(conditions, dict) and isinstance(conditions.get("actions"), dict):
+        # The actions M2M didn't resolve (e.g. the referenced actions were deleted) but a stale
+        # actions blob still lives in `conditions`. That stored blob can carry fields this caller's
+        # serializer would never expose — notably `created_by` on the public /surveys and /decide
+        # payload — so project each value down to the serializer's own fields. Never surface more
+        # than the serializer itself would.
+        allowed_fields = getattr(action_serializer_class.Meta, "fields", None)
+        if isinstance(allowed_fields, list | tuple):
+            allowed = set(allowed_fields)
+            values = conditions["actions"].get("values") or []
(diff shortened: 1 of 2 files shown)


lucasheriques (approved):
thank you security tom

PR #72012, opened 2026-07-17
chore(api): retire /api/environments dual registration

yasen-posthog:
## Problem

The `/api/environments/*` → `/api/projects/*` rewrite is fully rolled out (100%), so the per-team feature-flag gating and the dual route registration that backed the migration are now dead weight.

## Changes

- Make the `/api/environments/*` → `/api/projects/*` rewrite unconditional and drop the feature-flag gating
- Register team-nested endpoints only under `/api/projects/*`; the legacy env alias is served by the in-process rewrite
- Remove the dual-route registration helpers and the `/api/environments` parent router
- Delete the defunct env-only query-progress stubs and the redundant env URL aliases
- Remove the obsolete backend lint rules that guarded env-route registration
- Regenerate MCP API types to drop the removed env query-param definitions

> [!IMPORTANT]
> The `api-environments-redirect` flag is no longer read by code — the rewrite is unconditional. Leaving it in place is harmless; archive it after this ships.

## How did you test this code?

- ~660 existing + new API tests pass end-to-end through the env alias; the rewrite resolves to `/api/projects/*` with no 404/405/NoReverseMatch
- Adversarial routing sweep (differential env-vs-project equivalence, nested sub-routes, edge cases) found no breaks
- Agent-authored; only the automated tests above were run

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #72815, opened 2026-07-22
Revert "chore(api): retire /api/environments dual registration"

yasen-posthog:
JUST IN CASE

Reverts PostHog/posthog#72012


yasen-posthog:
Not needed

PR #72818, opened 2026-07-22
feat(schema): discriminate the AnyPropertyFilter union on type

thmsobrmlr:
## TL;DR

Property filters in queries are now validated by reading their `type` field directly instead of trying all 22 possible filter shapes one by one. Everything that validated before still validates (including old filters that have no `type` at all), but a bad filter now produces one clear error instead of dozens, and validating filter lists is roughly 20x faster.

## Problem

`AnyPropertyFilter` is a 22-member union that datamodel-code-generator inlines at 65 field sites in the generated `posthog/schema.py` (entity nodes, insight query `properties`, `DashboardFilter`, `HogQLFilters`, the recursive `PropertyGroupFilterValue`). Pydantic validated it as a smart union: a single malformed filter produced an error per member (a malformed 100-item list yields 16,000 errors, which is noise for users and LLM clients alike), and valid lists paid for walking every variant on one of our hottest validation paths. `FunnelsQuery.series` had the same problem for `kind`-tagged unions and got a discriminator via `bin/patch-schema-array-discriminators.py`; this applies the same treatment to property filters.

A plain JSON Schema `discriminator` (the JSDoc `@discriminator` route) can't express this union, though:

- legacy payloads omit `type` entirely and must keep validating as event filters; several read paths construct `DashboardFilter(**filters)` unwrapped and would turn a rejection into a 500
- `{}` rows must keep resolving to `EmptyPropertyFilter`
- `LogPropertyFilter` and `SpanPropertyFilter` accept three `type` values each (StrEnum tags, which pydantic's string discriminator rejects at model build)
- `PropertyGroupFilterValue` recursively unions with the filters but is tagged `AND`/`OR`

## Changes

- `posthog/schema_discriminators.py`: a hand-written callable discriminator. Missing `type` with a `key` maps to `event`, bare `{}` to `empty`, the log/span attribute variants to their member, and `AND`/`OR` to the property group.
- `bin/patch-schema-property-filter-discriminator.py`: a new post-generation step that rewrites the 65 inlined unions to two aliases (`AnyPropertyFilterDiscriminated`, and `AnyPropertyFilterOrGroupDiscriminated` for the recursive site) built from `pydantic.Tag` per member plus `pydantic.Discriminator`. The three `fixedProperties` sites keep `PropertyGroupFilter` and `PropertyGroupFilterValue` in an outer smart union, since those two share the `AND`/`OR` tag and can't be told apart. The script fails the build loudly if the union's members or site counts change, so schema changes have to update it (and the tests) consciously rather than silently regressing to a smart union.
- `posthog/schema.py`: regenerated; roughly 1,800 lines of inlined unions collapse into the aliases.

```diff
 # Re-apply discriminator keywords to array items dropped by datamodel-code-generator
 python3 bin/patch-schema-array-discriminators.py
+
+# Route the AnyPropertyFilter union on `type` via a legacy-tolerant callable discriminator
+python3 bin/patch-schema-property-filter-discriminator.py

 # Move enum classes to posthog/schema_enums.py
 python3 bin/split-schema-enums.py
```

Measured against the generated models on pydantic 2.12.5: validating 1,000 event filters drops from ~29ms to ~1.2ms, and 100 malformed items produce 100 errors instead of 16,000.

> [!NOTE]
> Two visible behavior changes, both for payloads that were already invalid or malformed:
>
> - Invalid payloads get smaller, routed error bodies: `union_tag_invalid` / `union_tag_not_found` codes and the member tag in `loc`, the same class of change the `FunnelsQuery.series` discriminator shipped.
> - One narrow accept-to-reject edge: a filter *without* `type` whose fields only fit a non-event member (for example a type-less dict carrying `group_type_index`) used to fall into `GroupPropertyFilter` and now routes to `EventPropertyFilter` and fails `extra_forbidden`. I checked saved insights and dashboards on a large internal project and found no type-less property filters in stored data at all, so this edge is theoretical for saved queries.

- New `posthog/test/test_property_filter_discriminator.py` (`SimpleTestCase`, no DB). Each group pins a regression no existing test catches: routing for all 22 members (catches tag-map drift in the new build script), the legacy tolerances for type-less and `{}` filters (their loss would 500 unwrapped `DashboardFilter` reads; nothing pinned this at the schema level before), log/span multi-value tags routing while preserving their original enum value, recursion through `PropertyGroupFilterValue`, group dicts still resolving to `PropertyGroupFilter` at the `fixedProperties` sites, and the error discipline (exactly one error per malformed item, a single `union_tag_invalid` for unknown tags).
- Ran the full schema build twice with the exact pinned toolchain from `uv.lock` (datamodel-code-generator 0.36.0, ruff 0.15.16): the unmodified pipeline reproduces the committed `schema.py` byte for byte, and the new step is deterministic and idempotent.
- Ran the existing `test_funnels_series_discriminator.py` and `test_experiment_metric_discriminator.py` suites against the regenerated schema: green.
- I was not able to run the full Django suite in the sandbox (no dev stack); CI covers that, including the `schema.py` freshness check which re-runs the new build step.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs reference the schema build steps; nothing user-facing changes for valid payloads.

## 🤖 Agent context

- Authored by Claude (PostHog Code cloud task) directed by Thomas. The work started as an investigation (pipeline mechanics, a map of every API boundary relying on type-less filter tolerance, a HogQL prevalence probe over saved insights and dashboards) and the implementation followed the reviewed plan. Skills invoked: `/writing-tests`.
- Key decisions: a callable `pydantic.Discriminator` applied post-generation was chosen over a TS JSDoc `@discriminator`, because the union must keep accepting tag-less legacy payloads, multi-value StrEnum tags, and the `AND`/`OR`-tagged recursive group, none of which a plain discriminator can express. `pydantic.Tag` is referenced module-qualified because the generated schema already contains an enum class named `Tag`. The hand-written `AnyPropertyFilter` alias in `posthog/types.py` (backing `test_account_filters` validation) is deliberately untouched to keep that boundary's behavior identical.

---

### DIFF
--- a/bin/build-schema-python.sh
+++ b/bin/build-schema-python.sh
@@ -16,6 +16,9 @@ datamodel-codegen \
 # Re-apply discriminator keywords to array items dropped by datamodel-code-generator
 python3 bin/patch-schema-array-discriminators.py
 
+# Route the AnyPropertyFilter union on `type` via a legacy-tolerant callable discriminator
+python3 bin/patch-schema-property-filter-discriminator.py
+
 # Move enum classes to posthog/schema_enums.py (importable without the pydantic model cost)
 python3 bin/split-schema-enums.py
 
--- a/bin/patch-schema-property-filter-discriminator.py
+++ b/bin/patch-schema-property-filter-discriminator.py
@@ -94,27 +94,45 @@ def _ensure_import_names(source: str, module: str, extra_names: list[str]) -> st
     return source[: match.start()] + f"from {module} import " + ", ".join(names) + source[match.end() :]
 
 
-def _alias_block() -> str:
+def _tagged_members() -> list[str]:
     # `pydantic.Tag`/`pydantic.Discriminator` stay module-qualified: the generated schema
     # already contains an enum class named `Tag`, and bare names could collide with any
     # future generated symbol.
-    tagged = [f'Annotated[{member}, pydantic.Tag("{tag}")]' for member, tag in MEMBER_TAGS.items()]
-    union = "\n    | ".join(tagged)
-    group_union = "\n    | ".join(['Annotated[PropertyGroupFilterValue, pydantic.Tag("property_group")]', *tagged])
+    return [f'Annotated[{member}, pydantic.Tag("{tag}")]' for member, tag in MEMBER_TAGS.items()]
+
+
+def _main_alias_block() -> str:
+    union = "\n    | ".join(_tagged_members())
     return (
         "# Added by bin/patch-schema-property-filter-discriminator.py: tagged AnyPropertyFilter\n"
-        "# aliases so validation routes on `type` instead of trying every member. The callable\n"
+        "# alias so validation routes on `type` instead of trying every member. The callable\n"
         "# keeps legacy tolerance — see posthog/schema_discriminators.py.\n"
         f"{DISCRIMINATED_ALIAS} = Annotated[\n"
         f"    {union},\n"
         f"    pydantic.Discriminator({DISCRIMINATOR_FN}),\n"
         "]\n"
-        "\n"
+    )
+
+
+def _group_alias_block() -> str:
+    group_union = "\n    | ".join(
+        ['Annotated[PropertyGroupFilterValue, pydantic.Tag("property_group")]', *_tagged_members()]
+    )
+    return (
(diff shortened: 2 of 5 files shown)


gantoine (approved):
big fan of simpler schema

PR #72823, opened 2026-07-22
feat(ingestion): fold $identify merge storms into single transactions

jose-sequeira:
## Problem

When many $identify events for the same distinct_id arrive in one batch (a merge storm), each merge runs as its own Postgres transaction with two row-locking fetches and its own retry loop, all serialized on the same rows. Storms of anon ids identifying into one person are a recurring source of ingestion-analytics lag.

## Changes

- Adds a `prescan` hook to the grouping pipeline so a group's queued chunk can be scanned before sequential processing.
- Plans folds during the prescan, gated by `PERSON_MERGE_FOLD_ENABLED` (default off) and `PERSON_MERGE_FOLD_TEAM_ALLOWLIST`: consecutive runs of two or more $identify events in a token:distinct_id group share one `MergeFoldPlan`. With the flag off, no prescan is wired and behavior is unchanged.
- Executes the fold on the run's first event: one locked target fetch, one batched locked source fetch, then a single transaction that moves all distinct ids, applies cohort and feature flag overrides, and deletes the source persons. Later events in the run short-circuit.
- Adds the batched repository primitives behind that (`fetchPersonsForUpdateByDistinctIds` with deterministic lock ordering, multi-source move and delete, batched cohort/FF update, per-source distinct id counts).
- Works under all merge modes: an in-transaction count pre-check rolls the fold back when a source exceeds the LIMIT/ASYNC move limit or the total exceeds the sync batch size, so per-event DLQ/redirect decisions stay on the sequential path.
- Any fold failure abandons the plan and falls back to today's per-event path, so the worst case is current behavior.
- Emits `person_merge_fold_executed_total`, `person_merge_fold_fallback_total{reason}`, and a fold size histogram.

> [!NOTE]
> Behavior is unchanged until `PERSON_MERGE_FOLD_ENABLED` is set. For a folded run, source-person properties land on the target when the run's first event executes rather than event-by-event, and per-pair warnings (already identified, illegal distinct id) are emitted once per deduped pair instead of once per event.

Automated only (no manual testing):

- Framework unit tests for the prescan hook: per-chunk scan ordering, value mutation visibility, re-queued chunks scanned separately, error poisoning.
- Unit tests for fold planning: run detection, splitting on interleaved events, pair dedup, self-merge exclusion, team allowlist.
- Postgres integration tests for the batched repository primitives, including lock-order and already-moved-source semantics.
- A differential integration test asserting a folded storm produces the same persons, distinct ids, and properties as sequential merges, plus tests for short-circuiting, cold-start bootstrap, personless adds, limit-mode folding, and over-limit rollback to the sequential path.
- The ingestion e2e suite now runs entirely under a fold on/off matrix (replacing the proven persons-prefetch dimension); all 152 tests pass in both arms, including a new storm scenario.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal ingestion behavior behind a default-off env flag; no user-facing docs affected.

## 🤖 Agent context

Authored with Claude Code from a staged plan. Notable decisions along the way: min-run and max-pairs tuning knobs were proposed and rejected as useless (2 is the natural fold floor and batch size bounds fold size); folding was restricted to consecutive $identify runs to avoid reordering person property merges across interleaved events; an initial SYNC-only merge-mode gate was replaced with the in-transaction count pre-check after checking that production lanes run ASYNC/LIMIT modes, where the gate would have disabled folding entirely.

While validating, we found that the e2e test "we do not alias users if distinct id changes but we are already identified" fails deterministically if any microtask is added to the $identify path (a bare awaited null reproduces it). The fold call is therefore only awaited when a plan exists. That latent timing sensitivity is pre-existing and deserves its own investigation.


nodejs/src/ingestion/pipelines/analytics/joined-ingestion-pipeline.ts
```diff
@@ -222,9 +227,14 @@ export function createJoinedIngestionPipeline<
...
             .compose((b) => createPostTeamPreprocessingSubpipeline(b, postTeamConfig))
             // Group by token:distinctId and process each group concurrently.
-            // Events within each group are processed sequentially.
+            // Events within each group are processed sequentially. When merge
+            // folding is enabled, a prescan plans folds over each group's
+            // chunk before its events run; when disabled, no prescan is wired
+            // and the pipeline is identical to the sequential-only shape.
             .concurrentlyPerGroup(getTokenAndDistinctId, (group) =>
-                group.sequentially((event) => createPerDistinctIdPipeline(event, perEventConfig))
+                (mergeFoldPrescan ? group.prescan(mergeFoldPrescan) : group).sequentially((event) =>
```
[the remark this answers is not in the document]
jose-sequeira: I don't think this is an issue, tipically transformations don't drop events and there's a clear intent on identifying the user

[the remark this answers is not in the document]
jose-sequeira: It's totally okay

nodejs/src/ingestion/common/persons/person-merge-service.ts
```diff
@@ -339,6 +370,291 @@ export class PersonMergeService {
...
+        let target = await this.context.personStore.fetchForUpdate(teamId, plan.targetDistinctId)
+        let pairsToFold = plan.pairs
+
+        if (!target) {
+            // Cold start: no target person yet. Bootstrap it through the
+            // sequential path for the current event's pair (which creates the
+            // person or attaches the distinct_id), then fold the remaining pairs.
+            const bootstrap = await promiseRetry(
+                () =>
+                    this.mergeDistinctIds(currentAnonDistinctId, plan.targetDistinctId, teamId, this.context.timestamp),
```
nickbest-ph, line 461: I think we're missing the isDistinctIdIllegal() check on target before calling this on this new path

nickbest-ph: needs a test to cover this as well

nodejs/src/ingestion/pipelines/analytics/joined-ingestion-pipeline.ts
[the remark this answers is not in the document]
nickbest-ph: i think this leaks on `$process_person_profile: false` events and when capture has set the `force_disable_person_processing` header as well, though. we drop $identify events for those

those cases are probably more legit and the fix should be easier than hog transformations, as i think the prescan can check for those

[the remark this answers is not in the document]
nickbest-ph: also, https://posthog.com/docs/cdp/transformations/drop-events ?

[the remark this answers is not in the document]
nickbest-ph: i guess it'd probably weird for a customer to set up a filter here to drop on $identify events

nodejs/src/ingestion/common/persons/batch-writing-person-store.ts
```diff
@@ -1393,6 +1399,58 @@ export class BatchWritingPersonsStore implements PersonsStore, BatchWritingStore
         return response
     }
 
+    async fetchPersonsForUpdateByDistinctIds(
```
nickbest-ph, line 1402: is there a reason this batch version of the function doesn't try to read from the cache?

I think that when executing the merge fold, the fact that this doesn't hit the cache first opens up the possiblity of stale properties used for the source person

nodejs/src/ingestion/common/persons/person-merge-service.ts
```diff
@@ -339,6 +370,291 @@ export class PersonMergeService {
...
+        }
+        if (batchSize !== undefined && total > batchSize) {
+            throw new MergeFoldLimitError('folded merge move exceeds sync batch size')
+        }
+        return total
+    }
+
+    private async executeFoldedMerge(plan: MergeFoldPlan, currentAnonDistinctId: string): Promise<PersonMergeResult> {
+        const teamId = this.context.team.id
+        this.context.updateIsIdentified = true
```
nickbest-ph, line 450: will this unconditionally flip the target to be `isIdentified`? even if the merges are for illegal distinct ids and the merge is not executed?

nodejs/src/ingestion/common/persons/person-merge-service.ts
```diff
@@ -339,6 +370,291 @@ export class PersonMergeService {
...
+            }
+            const source = sourceByDistinctId.get(pair.anonDistinctId)
+            if (!source) {
+                missingPairs.push(pair)
+                continue
+            }
+            if (seenSourceIds.has(source.id)) {
+                continue
+            }
+            if (source.is_identified) {
```
z0br0wn, line 504: is there any risk in this value being stale at this point (if a merge happened in between read time and now) that would lead to his evaluating as false, but actually being true?

nodejs/src/common/persons/repositories/postgres-person-repository.ts
```diff
@@ -641,6 +693,47 @@ export class PostgresPersonRepository
...
 
+    async deletePersons(persons: InternalPerson[], tx?: TransactionClient): Promise<PersonMessage[]> {
+        if (persons.length === 0) {
+            return []
+        }
+
+        // All persons in a folded merge belong to one team.
+        const teamId = persons[0].team_id
+        const personById = new Map(persons.map((person) => [person.id, person]))
+        // Sorted ids keep the row-lock acquisition order deterministic across
```
z0br0wn: according to my bot lock order isn't determined or guaranteed by index sorting here, but by scan order of the tree (which in its own way is deterministic)

nodejs/src/common/persons/repositories/postgres-person-repository.ts
```diff
@@ -641,6 +693,47 @@ export class PostgresPersonRepository
...
+                tx ?? PostgresUse.PERSONS_WRITE,
+                'DELETE FROM posthog_person WHERE team_id = $1 AND id = ANY($2::bigint[]) RETURNING id, version',
+                [teamId, personIds],
+                'deletePersons'
+            )
+            rows = result.rows
+        } catch (error) {
+            if (error.code === '40P01') {
+                // Deadlock detected — assume someone else is deleting and skip.
+                logger.warn('🔒', 'Deadlock detected — assume someone else is deleting and skip.', {
```
z0br0wn, line 722: i wonder if it'd be worth having a retry here to see if it frees up? so we don't fall back to slow sequential processing too eagerly

nodejs/src/ingestion/pipelines/analytics/joined-ingestion-pipeline.ts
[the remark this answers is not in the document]
jose-sequeira: The force_disable_person_processing is a valid one though, need to check how that works

nodejs/src/ingestion/common/persons/person-merge-service.ts
jose-sequeira: Good call

nodejs/src/ingestion/common/persons/batch-writing-person-store.ts
jose-sequeira: Because it needs to lock the rows with for update (same behaviour as today, when the current fetchPerson call happens with forUpdate set to true it bypasses the cache)

nodejs/src/ingestion/common/persons/person-merge-service.ts
jose-sequeira: Think the logic is the same as the current, will double check.

nodejs/src/common/persons/repositories/postgres-person-repository.ts
jose-sequeira: Think it's valid, a bit concerned of huge retry churns (multiple here and perhaps multiple individually) during merge storms

nodejs/src/ingestion/pipelines/analytics/joined-ingestion-pipeline.ts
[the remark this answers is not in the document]
jose-sequeira: Fixed in 2134223: the prescan now treats a $identify with person processing disabled ($process_person_profile: false or the force_disable_person_processing header) as non-foldable, so it can't contribute a pair to a plan and it splits the run like any other non-foldable event. The decision logic is extracted into a shared decideProcessPerson helper that the normalize step now uses too, so the two paths can't drift. Will push together with the rest of the review fixes.

nodejs/src/ingestion/common/persons/person-merge-service.ts
jose-sequeira: Fixed in bf1b046. Digging in: the target-illegal case was already guarded in tryFoldedMerge (it abandons the plan before executeFoldedMerge), but you were right that this bootstrap had a hole: it calls mergeDistinctIds directly, bypassing the illegal checks merge() does sequentially, and the current event's pair is filtered out of pairsToFold, so an illegal $anon_distinct_id (e.g. "anonymous") on the fold-executing event could bootstrap a merge the sequential path refuses. Illegal anon ids are now excluded at plan time (like self-merges), so their events are never plan members and always flow down the sequential path with today's per-event warning. The in-fold per-pair check stays as defense in depth. Also moved isDistinctIdIllegal to a new common/persons/person-utils.ts so the planner can use it without importing the merge service.

jose-sequeira: Added planner unit tests: an illegal anon id is excluded from the plan without splitting the run, and a run of only illegal anon ids produces no plan. Since plans are only built by the planner, an illegal pair can no longer reach the bootstrap.

nodejs/src/ingestion/common/persons/person-merge-service.ts
jose-sequeira: You were right that this diverged for illegal distinct ids, and it's resolved as a side effect of bf1b046: illegal anon ids are now excluded from plans entirely, so their events always take the sequential path and never reach this line. For everything that can still reach it, the unconditional set mirrors the sequential path exactly: mergeDistinctIds sets updateIsIdentified = true as its first statement, before any refusal logic, so even a merge refused because the source is already identified still flips the target sequentially today. The fold's conditionality lives in plan membership (legal, non-self pairs only), which now matches merge()'s pre-checks one to one.

nodejs/src/ingestion/common/persons/person-merge-service.ts
jose-sequeira: This is a very good catch and actually an issue that we currently have in `master`. I will address this in a separate PR

nodejs/src/common/persons/repositories/postgres-person-repository.ts
jose-sequeira: The comment is overselling it for sure, the lock is guaranteed on the btree index scan level, I'll reword the comment, but don't think there's something actionable here

nodejs/src/common/persons/repositories/postgres-person-repository.ts
jose-sequeira: After some research, retries here are impossible. Basically if we get `40P01`, it means that our transaction is actively stopping the world from continuing, so we cannot retry it, we must start a new transaction. This should already be covered by the wrapper `promiseRetry` around the `handleMergeTransaction`.

I am however adding proper metrics for this to know at what rate we are generating these deadlocks

z0br0wn (approved):
lgtm on my end!

nickbest-ph (approved):
lgtm. excited to see improvements from this.

great change!

PR #61764, opened 2026-06-04
feat(replay): route rasterizer S3 client through smokescreen

TueHaulund:
## Problem

After #61251 routed Chrome through smokescreen, S3 was the only remaining direct egress — AWS SDK v3 doesn't honor `HTTPS_PROXY` unless you wrap it. Lets PostHog/charts#11594 eventually drop port 443 from its public-internet rule.

## Changes

`storage.ts` reads `HTTPS_PROXY` (with the same env-var precedence + `RASTERIZER_USE_PROXY` kill switch as the Chrome path) and passes `requestHandler: { httpsAgent: new HttpsProxyAgent(url) }` to `S3Client`. When the env var is unset or the kill switch is on, the SDK falls back to the default direct dial.

Adds `https-proxy-agent` as a direct dep (already transitive via `google-auth-library`).

## How did you test this code?

Agent-authored. 17 parameterized unit tests in `storage.test.ts` lock in the routing for all four env var casings and the kill-switch values. TS check clean. No manual test — same `RASTERIZER_USE_PROXY=false` flag that gates Chrome routing in prod also disables this.

## 🤖 Agent context

Claude Opus 4.7 via Claude Code.

Issue #13367, opened 2022-12-16
Enable session recordings from posthog.init


lharries:
## Is your feature request related to a problem?

Currently, you need to turn on session recordings from the UI. It would be cleaner to be able to control it from the config and appeals to our highly technical ICP.

## Describe the solution you'd like

The cleanest solution would likely be turning on session recordings if either `session_recording is true | object`. Similar to autocapture. You can then turn it off explicitly by setting it to `false` like so:

```js
posthog.init('...', { session_recording: true }) // session recordings is now on
posthog.init('...', { session_recording: { blockClass: 'no-capture', } }) // session recordings is now on and uses the block class no-capture

posthog.init('...', {}) // session recordings is on only if it's turned on in the UI

posthog.init('...', { session_recording: false}) // session recordings is explicitly off
```

However, this would likely accidentally turn on session recordings for the small number of users that have session recordings turned off in the UI but have an object such as `{ blockClass: 'no-capture' }` set.

The current options for session recording are:
```js
posthog.init('...', {
  disable_session_recording: true, // turns it off even if it's on in the UI
  session_recording: {} // config block
})
```

Having `enable_session_recording` as well as `disable_session_recording` seems very messy.

Some potential options:
1. Have `session_recording: true | object` => turn on session recordings and let people know that this change will be happening and do a major version bump i.e. `1.36 => 2.0` (this is what I'd recommend). Could instead do a minor version bump but a bit more risky
2. Move from `session_recording` to `session_recordings` (with an s), keeping backwards compatibility with `session_recording` but enabling you to set `session_recordings: true | object => to turn on session recordings`
3. Add `enable_session_recording` as well as `disable_session_recording`. Although I think this will end up being quite confusing and not as clean?

I'm happy to implement this (or someone else can) once we've decided the way forward

## Describe alternatives you've considered

Maintain the need to turn on session recordings only via the UI

lharries:
cc @benjackwhite @mariusandra @annikaschmid

benjackwhite:
The overall idea here seems reasonable to me. Not sure about Option 2 (renaming to `session_recordings`). In the context of a client it is singular. We mostly use the singular form elsewhere as well. I would perhaps be tempted to break that out to a different issue where we solidify when to use plural and when to use singular, rather than including it here.

[...]

PR #72824, opened 2026-07-22
fix(replay): keep rasterizer STS credential refresh off the proxy

Daesgar:
## Problem

The recording-rasterizer routes egress through Smokescreen. The pod env carves STS out of the proxy with `NO_PROXY=…,sts.<region>.amazonaws.com`, so IRSA credential-refresh calls should hit AWS directly. They don't - STS still goes through Smokescreen.

Root cause: AWS SDK for JavaScript v3 does not read `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` at all. The service reaches Smokescreen only because it builds the S3 client with an explicit `requestHandler` wired to an `HttpsProxyAgent`. [#61764](https://github.com/PostHog/posthog/pull/61764) then also handed that same proxied handler to the credential provider's internal STS client (via `credentials: defaultProvider({ clientConfig: { requestHandler } })`), forcing credential refresh through Smokescreen too. Since the SDK ignores `NO_PROXY`, the chart's STS carveout can't undo that.

This is the app-side complement to the charts-side `NO_PROXY` STS carveouts (PostHog/charts [#13263](https://github.com/PostHog/charts/pull/13263), [#13367](https://github.com/PostHog/charts/pull/13367)), which cannot take effect for these SDK-v3 clients without this change.

## Changes

Drop [#61764](https://github.com/PostHog/posthog/pull/61764)'s credential-provider proxy routing. The rasterizer has no directly-constructed `STSClient` - the STS traffic comes from `@aws-sdk/credential-provider-node`'s default provider, which spawns its own internal STS client for IRSA refresh. That internal client does **not** inherit our `requestHandler` (that's exactly why #61764 had to pass `clientConfig` to force it onto the proxy). So by simply not overriding `credentials`, credential refresh falls back to the SDK default and dials STS direct, while S3 keeps going through Smokescreen.

That's the whole fix - one line removed from the `S3Client` config, plus a comment explaining why we must not route credentials through the proxy. `RASTERIZER_USE_PROXY=false` still disables the proxy entirely.

> [!NOTE]
> An earlier revision of this PR parsed `NO_PROXY` in-app and matched the regional STS host to decide per-request whether to proxy. That worked but was far more code than needed. Since the SDK-default credential provider already dials direct, the minimal fix is just to stop forcing it onto the proxy.

- Reworked the `S3 client proxy routing` cases in `storage.test.ts` to assert the new contract: when a proxy env var is set, the S3 client gets the proxy `requestHandler` but **no** `credentials` override is passed and the credential provider is never called with a proxied handler. This is the regression guard - if someone re-adds credential proxying (re-introducing the #61764 line), these fail. No existing test covered "S3 proxied, credentials direct".
- `pnpm exec jest src/session-replay/recording-rasterizer/__tests__/storage.test.ts` - 25 passing.
- `pnpm exec eslint` and `pnpm exec prettier --check` on the changed files - clean.
- `tsc --noEmit -p .` - no errors in recording-rasterizer files.

I did not run the service against live AWS.

## 🤖 Agent context

Authored by Claude (PostHog Code). Invoked the `/writing-tests` skill before reworking the test suite. The first cut parsed `NO_PROXY` and matched the regional STS host per-request; on review we simplified to just dropping the credential-provider proxy routing, since the SDK-default provider already dials STS direct - so the app doesn't need to know about `NO_PROXY` at all.

---


nodejs/src/session-replay/recording-rasterizer/storage.ts
```diff
@@ -36,11 +35,8 @@ function getS3Client(): S3Client {
...
             ...(config.s3Endpoint ? { endpoint: config.s3Endpoint, forcePathStyle: true } : {}),
             ...(requestHandler ? { requestHandler } : {}),
-            // Route the IRSA credential-refresh STS call through the proxy too.
-            // The SDK's default credential provider spawns an internal STSClient
-            // that does NOT inherit our requestHandler — pass clientConfig so it
-            // does. Without this, S3 goes via smokescreen but STS dials direct.
-            ...(requestHandler ? { credentials: defaultProvider({ clientConfig: { requestHandler } }) } : {}),
+            // Do not route credentials through the proxy: the SDK default provider
+            // must dial direct.
         })
```
Piccirello, line 42: Can you update this comment to include "why" the SDK must dial direct?

Piccirello (approved):

nodejs/src/session-replay/recording-rasterizer/storage.ts
Daesgar: Didn't want to give so many details of what stuff are we bypassing that's why I didn't put a better comment. Maybe I can remove the comment entirely and just rely on the test.

Daesgar: But now that I think about it the PR already share enough details lol

TueHaulund (approved):

PR #72830, opened 2026-07-22
fix(tasks): use GLM for custom image workflows

tatoalo:
## Problem

Custom image security scans relied on prompt-only JSON output, allowing malformed model responses to fail before the image build began. Image-builder conversations were also pinned to Sonnet rather than the preferred GLM model.

## Changes

Route security scans through the internal `posthog_code` LLM gateway using GLM 5.2 and JSON response mode. Pin image-builder sessions to GLM 5.2 with high reasoning. Validate every scan response against a typed verdict before applying it.


k11kirky (approved):
nice

PR #72832, opened 2026-07-22
feat(conversations): attribute Teams support replies to the agent

luke-belton:
## Problem

Support replies sent back into a Microsoft Teams thread showed up from a nameless "SupportHog", so customers couldn't tell which agent had answered — and in at least one case appeared to miss the reply entirely because it looked like an automated bot message. A Teams bot's display name is fixed in the app manifest and can't be overridden per-message, so we can't rename the sender; the fix is to attribute the reply in the message body instead.

Raised in a support Slack thread: https://posthog.slack.com/archives/C090RCG671C/p1784708626116489?thread_ts=1784708626.116489&cid=C090RCG671C

## Changes

Add a shared `append_teams_attribution` helper that appends a subtle italic `{name} via SupportHog` footer to outbound Teams replies, and wire both outbound paths through it:

- **Standard channels** (bot connector, `post_reply_to_teams`) previously rendered no author in the visible body — the name only went into the hidden `summary` field. Now it shows the footer.
- **Shared channels** (Graph, `post_reply_to_teams_via_graph`) already prepended a bold author name; switched to the same footer for consistency.

The author name is escaped, and replies with no author (e.g. AI bot) are left untouched.

Added unit tests in `test_teams_formatting.py` covering the footer output, the no-author passthrough, and HTML escaping of the author name. Verified the helper output directly with a standalone Python run. I (the agent) could not run the Django test suite or exercise a live Teams thread in this environment, so the end-to-end path against Teams has not been manually verified.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by the PostHog Slack app (Claude Code) from a support Slack thread. The investigation traced the two outbound Teams reply paths in `products/conversations/backend` and found the attribution inconsistency between them (Graph path already named the author, bot-connector path did not). Chose a shared footer over the existing bold-header approach to keep the attribution subtle, per the requester's suggestion. No repo skills were required for this change.

### DIFF
--- a/products/conversations/backend/api/tests/test_teams_formatting.py
+++ b/products/conversations/backend/api/tests/test_teams_formatting.py
@@ -4,6 +4,7 @@ from parameterized import parameterized
 
 from products.conversations.backend.teams_formatting import (
     append_teams_attribution,
+    build_teams_reply_html,
     rich_content_to_teams_html,
     teams_html_to_content_and_rich_content,
 )
@@ -280,3 +281,19 @@ class TestAppendTeamsAttribution(SimpleTestCase):
         result = append_teams_attribution("<p>Hello</p>", "<script>alert(1)</script>")
         assert "<script>" not in result
         assert "&lt;script&gt;" in result
+
+
+class TestBuildTeamsReplyHtml(SimpleTestCase):
+    def test_renders_rich_content_then_footer(self):
+        rich = {
+            "type": "doc",
+            "content": [{"type": "paragraph", "content": [{"type": "text", "text": "All fixed now."}]}],
+        }
+        assert build_teams_reply_html(rich, "All fixed now.", "Xander Jones") == (
+            "<p>All fixed now.</p><p><i>Xander Jones via SupportHog</i></p>"
+        )
+
+    def test_falls_back_to_plain_content_with_footer(self):
+        assert build_teams_reply_html(None, "All fixed now.", "Xander Jones") == (
+            "All fixed now.<p><i>Xander Jones via SupportHog</i></p>"
+        )
--- a/products/conversations/backend/api/tests/test_teams_formatting.py
+++ b/products/conversations/backend/api/tests/test_teams_formatting.py
@@ -3,6 +3,7 @@ from django.test import SimpleTestCase
 from parameterized import parameterized
 
 from products.conversations.backend.teams_formatting import (
+    append_teams_attribution,
     rich_content_to_teams_html,
     teams_html_to_content_and_rich_content,
 )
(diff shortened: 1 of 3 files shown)


products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +265,18 @@ def test_multiple_marks_combined(self):
         }
         result = rich_content_to_teams_html(rich)
         assert result == "<p><i><b>wow</b></i></p>"
+
+
+class TestAppendTeamsAttribution(SimpleTestCase):
+    def test_appends_italic_footer(self):
+        assert append_teams_attribution("<p>Hello</p>", "Xander Jones") == (
+            "<p>Hello</p><p><i>Xander Jones via SupportHog</i></p>"
```
luke-belton:
```suggestion
    def test_appends_italic_footer(self):
        assert append_teams_attribution("<p>Hello</p>", "Max Hedgehog") == (
            "<p>Hello</p><p><i>Max Hedgehog via SupportHog</i></p>"
```

products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +266,34 @@ def test_multiple_marks_combined(self):
...
```
luke-belton:
```suggestion
            "<p>Hello</p><p><i>Max Hedgehog via SupportHog</i></p>"
```

products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +266,34 @@ def test_multiple_marks_combined(self):
...
+
+
+class TestBuildTeamsReplyHtml(SimpleTestCase):
+    def test_renders_rich_content_then_footer(self):
+        rich = {
+            "type": "doc",
+            "content": [{"type": "paragraph", "content": [{"type": "text", "text": "All fixed now."}]}],
+        }
+        assert build_teams_reply_html(rich, "All fixed now.", "Xander Jones") == (
+            "<p>All fixed now.</p><p><i>Xander Jones via SupportHog</i></p>"
```
luke-belton:
```suggestion
        assert build_teams_reply_html(rich, "All fixed now.", "Max Hedgehog") == (
            "<p>All fixed now.</p><p><i>Max Hedgehog via SupportHog</i></p>"
```

products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +266,34 @@ def test_multiple_marks_combined(self):
...
+        )
+
+    def test_falls_back_to_plain_content_with_footer(self):
+        assert build_teams_reply_html(None, "All fixed now.", "Xander Jones") == (
```
luke-belton:
```suggestion
        assert build_teams_reply_html(None, "All fixed now.", "Max Hedgehog") == (
```

products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +266,34 @@ def test_multiple_marks_combined(self):
...
+            "All fixed now.<p><i>Xander Jones via SupportHog</i></p>"
```
luke-belton:
```suggestion
            "All fixed now.<p><i>Max Hedgehog via SupportHog</i></p>"
```

products/conversations/backend/api/tests/test_teams_formatting.py
```diff
@@ -264,3 +266,34 @@ def test_multiple_marks_combined(self):
         }
         result = rich_content_to_teams_html(rich)
         assert result == "<p><i><b>wow</b></i></p>"
+
+
+class TestAppendTeamsAttribution(SimpleTestCase):
+    def test_appends_italic_footer(self):
+        assert append_teams_attribution("<p>Hello</p>", "Xander Jones") == (
```
luke-belton:
```suggestion
        assert append_teams_attribution("<p>Hello</p>", "Max Hedgehog") == (
```

xljones (approved):
Nice. MS Teams is a PostHog team 2 only thing, so `SupportHog` hardcoded is fine here IMO.

PR #72846, opened 2026-07-22
fix(data-warehouse): tolerate string team ids from the control plane

benben:
## Problem

Review feedback on #72827: if the control plane ever serializes `team_id` as a JSON string, the sensor's reconciliation map is keyed by strings (`"42"`) while Django team ids are ints (`42`) — every team then looks unknown and the repair is silently skipped. Our control plane marshals `team_id` as a JSON number today (Go `int64`), so this is defensive, but the failure mode is silent, so it's worth closing.

## Changes

- The reconcile pass coerces `team_id` to int when building its lookup map; rows without a usable id are skipped.
- The adapter's two row-lookup sites (`_get_project_team_row`, `team_onboarding_state`) get the same coercion via a shared `_row_team_id` helper — same latent pattern, same fix.

Automated only: new `cp_string_team_id` case in the parameterized reconcile test (string id must still repair); full `test_events_backfill_to_duckling.py` + `test_managed_warehouse.py` suites green locally (291 passed); ruff clean.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code, addressing post-merge review feedback on #72827. Verified against production control-plane responses that team_id is currently a JSON number, so the change is hardening, not a live bug fix. Skills invoked: /writing-tests.

### DIFF
--- a/posthog/dags/events_backfill_to_duckling.py
+++ b/posthog/dags/events_backfill_to_duckling.py
@@ -3018,7 +3018,14 @@ def _reconcile_earliest_event_dates_with_cp(backfills: list[DuckgresServerTeam])
             teams = managed_warehouse._teams_from_response(managed_warehouse.list_teams(org_id, require_enabled=False))
             if teams is None:
                 continue
-            cp_dates = {row.get("team_id"): row.get("earliest_event_date") for row in teams}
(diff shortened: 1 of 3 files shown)


jghoman (approved):
<Shakes fist at json>

PR #72849, opened 2026-07-22
feat(notebooks): add node-run observability

sinan-ku:
## Problem

The SQLV2 notebooks rollout has no end-to-end latency signal. Nothing measures "user hit run → result in the cell": the only per-run record is the `NotebookNodeRun` row's timestamps, and the sandbox-side phases (data-plane wait, frame download, kernel exec) are invisible to the server. That makes "did this get faster or slower?" unanswerable during the rollout, and it hid a production incident where python runs hung indefinitely.

## Changes

One recorder, called by every path that wins a `RUNNING → terminal` transition, emitting three sinks at once:

- `posthog_notebooks_node_run_seconds{node_type, outcome}` – end-to-end duration histogram (Prometheus for Grafana, plus its OTLP twin into the PostHog Metrics product)
- `posthog_notebooks_kernel_phase_seconds{phase, node_type}` – sandbox-reported phase durations
- a `notebook node run completed` analytics event with `duration_seconds`, per-phase seconds, `row_count`, and `outcome` – sliceable per notebook/team/user

Wired at all terminal sites: the sandbox callback, the direct (hogql) lane's `_finish_direct_run`, run-dispatch failures (view + Temporal mark-failed activity), and the interrupt endpoints. Outcomes are `done` / `failed` / `interrupted` / `timed_out` – the direct lane's grace-expiry watchdog is labeled `timed_out` so an expired query is a visible bucket, not a user error.

The sandbox kernel now measures its own phases and ships them in the callback envelope's new `timings` dict: `input_wait_s` (waiting on the data plane for referenced frames, or the display fetch), `download_s` (presigned frame downloads), `exec_s` (ipykernel cell execution), `sandbox_total_s` (the whole sandbox-side run). Attached on error paths too, so a run that died waiting on the data plane still reports where its time went.

> [!NOTE]
> The callback deliberately upserts on re-delivery; the recorder only fires for the caller that won the transition, so retries and racing pollers never double-count a run.

> [!NOTE]
> A kernel-lane run whose sandbox callback is lost still stays `RUNNING` and contributes no sample. A periodic reaper for those rows was built and then dropped from this PR as more complexity than the case warrants right now – the rows remain visible in Postgres and the run can be re-run.

Automated only (no manual sandbox run):

- `TestSQLV2NodeRunMetrics` – callback reports the run once with the right outcome and phase properties (catches: outcome mismapping, double-counting on re-delivered callbacks, dropped envelope timings); direct-lane finish reports only the winning transition (catches: double samples from racing pollers); a recorder failure never fails the callback response.
- Full `products/notebooks/backend/test/test_sql_v2.py` suite (one pre-existing failure needs local object storage and fails on a clean checkout too), `test_kernel_bootstrap.py`, `posthog/tasks/test/test_scheduled.py`, repo-wide mypy.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The metric/phase names and their spans are documented in the module docstring (`sql_v2_metrics.py`); the stacked follow-up PR updates the SQLV2 observability doc alongside the dashboard artifacts.

## 🤖 Agent context

Written by Claude Code from a production investigation into slow/hanging python nodes: run-level timing was reconstructed by hand from `posthog_notebooknoderun` and `query_log_archive`, which is what this PR automates. Skills invoked: `/writing-tests`, `/improving-drf-endpoints`, `/instrumenting-first-party-metrics` (chose prom instruments + OTLP twins per its twin rule, since the Grafana rollout dashboard consumes the prom series), `/query-clickhouse-via-metabase` (investigation). Decisions: phase timings ride the existing callback envelope instead of a new reporting channel; a stale-run reaper (Celery beat, 40-min deadline) was implemented, then removed at Sinan's direction as too much complexity for a non-critical case – the commit history shows both.


products/notebooks/backend/sql_v2_callback.py
```diff
@@ -101,6 +103,11 @@ def notebook_sql_v2_callback(request, run_id: str) -> JsonResponse:
 
     _store_frame_snapshot(run, frames, team_id)
 
+    # Only the callback that won the RUNNING -> terminal transition reports: a re-delivered
+    # or late envelope (the deliberate upsert above) must not double-count the run.
+    if was_running:
```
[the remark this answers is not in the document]
sinan-ku: Fixed in 66ff5212b: the callback now claims the RUNNING → terminal transition with a status-guarded UPDATE, so exactly one delivery reports; a losing duplicate still upserts the row (preserving the deliberate late-result-overwrites-interrupt semantics) but emits nothing.

products/notebooks/backend/temporal/sql_v2.py
```diff
@@ -77,6 +78,7 @@ def mark_sql_v2_run_failed_activity(input: SQLV2RunInput) -> None:
         run.status = NotebookNodeRun.Status.FAILED
         run.error = "Run failed to dispatch to the kernel."
         run.save(update_fields=["status", "error", "updated_at"])
+        record_node_run_terminal(run, OUTCOME_FAILED)
```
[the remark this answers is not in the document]
sinan-ku: Fixed in 66ff5212b: the activity now uses a status-guarded UPDATE and only reports when it won the transition — a run completed by a late callback keeps its real outcome. Added a parameterized regression test (test_mark_failed_activity_never_overwrites_a_completed_run).

products/notebooks/backend/presentation/views/notebook.py
```diff
@@ -1309,6 +1315,7 @@ def sql_v2_run_interrupt(self, request: Request, run_id: str | None = None, **kw
             run.status = NotebookNodeRun.Status.INTERRUPTED
```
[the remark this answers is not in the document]
sinan-ku: Fixed in 66ff5212b: the kernel-unreachable branch (and the run-start failure path, same shape) now records only after winning an atomic RUNNING → INTERRUPTED update.

products/notebooks/backend/sql_v2_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _sanitized_timings(envelope: Any) -> dict[str, float]:
+    raw = envelope.get("timings") if isinstance(envelope, dict) else None
+    if not isinstance(raw, dict):
+        return {}
+    return {
+        key: float(value)
+        for key, value in raw.items()
+        if key in _PHASE_BY_TIMING_KEY and isinstance(value, int | float) and value >= 0
```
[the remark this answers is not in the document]
sinan-ku: Fixed in d24da96: non-finite values are dropped and finite ones are clamped to the backend-measured run duration plus clock slack before touching the histograms — every phase is a sub-span of the run, so that's a hard ceiling. Covered by test_forged_envelope_timings_are_clamped.

products/notebooks/backend/sql_v2_metrics.py
```diff
@@ -0,0 +1,146 @@
...
+    properties: dict[str, Any] = {
+        "notebook_short_id": run.notebook.short_id,
+        "node_type": run.node_type,
+        "outcome": outcome,
+        "duration_seconds": round(duration, 3),
+        "row_count": envelope.get("row_count"),
+        "has_error": bool(run.error),
+        **{f"{_PHASE_BY_TIMING_KEY[key]}_seconds": round(value, 3) for key, value in timings.items()},
+    }
+    team = None if run.user else Team.objects.filter(pk=run.team_id).first()
```
georgemunyoro: **Blocking — the analytics event can be attributed to the wrong project.**

When `run.user` is set (the common case), this passes `team=None` and `report_user_or_team_action` falls back to `user.current_team` / `user.current_organization` for `$groups` attribution. But `run.team_id` comes from the URL project, and the user's currently active project can be a different one (multi-project users, personal API keys) — so the event this PR adds for per-project slicing can land in the wrong project's analytics.

`capture_notebook_created` in `products/notebooks/backend/analytics.py` already resolves and passes the team explicitly regardless of user; suggest the same here:

```suggestion
    team = Team.objects.filter(pk=run.team_id).first()
```

products/notebooks/backend/sql_v2_serializers.py
```diff
@@ -201,6 +201,17 @@ class NotebookSQLV2EnvelopeSerializer(serializers.Serializer):
         allow_blank=True,
         help_text="Error message when status is 'error'.",
     )
+    timings = serializers.DictField(
```
georgemunyoro: **Blocking — strict `timings` validation can strand runs in RUNNING forever.**

`FloatField` rejects `None`, non-numeric strings, and nested values — and a validation failure 400s the whole callback before any status write. `_post_callback` is fire-and-forget (a 400 is swallowed like a network error, no retry) and the reaper was deliberately dropped from this PR, so the run stays RUNNING indefinitely.

User code in the sandbox can forge the envelope — the exact threat model `_sanitized_timings`' own docstring names — and `_sanitized_timings` already does lenient per-key filtering, but it sits *behind* the serializer that hard-fails first. Suggest an untyped `DictField` (or a field that drops bad keys instead of failing the envelope) and let `_sanitized_timings` stay the sole validator.

Note the forged-envelope test only covers `1e300`, which is a syntactically valid float and never reaches this path; `{"exec_s": "n/a"}` would.

products/notebooks/backend/sandbox/kernel/runner.py
```diff
@@ -157,17 +164,23 @@ def _build_envelope(payload: dict[str, Any], cancel_event: threading.Event | Non
     page_limit = int(payload.get("page_limit") or _DEFAULT_PAGE_LIMIT)
     cache_limit = max(int(payload.get("cache_limit") or _DEFAULT_CACHE_LIMIT), page_limit)
     try:
+        fetch_started = time.monotonic()
         result = _fetch_capped_page(payload, limit=cache_limit, offset=0, cancel_event=cancel_event)
+        input_wait_s = round(time.monotonic() - fetch_started, 3)
```
georgemunyoro: **Non-blocking — failure paths lose the phase timings the PR says it keeps.**

`input_wait_s` is computed only after a successful `_fetch_capped_page`; all three except branches return envelopes with no timings, so a hogql display run that dies waiting on the data plane contributes no `input_wait` sample — exactly the runs the phase histogram should explain. `_materialize_inputs` in `executor.py` wraps the identical wait in try/finally precisely so it survives a raise; mirroring that here would make "attached on error paths too" true for both lanes.

Same class of gap in `executor.py`: `_ensure_kernel()` raising leaves `timings` empty, so kernel-boot failures report no phase at all.

products/notebooks/backend/sql_v2_metrics.py
```diff
@@ -0,0 +1,146 @@
...
+    report_user_or_team_action(NODE_RUN_COMPLETED_EVENT, properties, user=run.user, team=team)
```
georgemunyoro: **Non-blocking — a dangling `user_id` silently drops the completed event.**

`NotebookNodeRun.user` is `on_delete=DO_NOTHING, db_constraint=False`, so a hard-deleted user leaves a dangling id, and `run.user` access on a non-select_related instance raises `RelatedObjectDoesNotExist` (the model comment "a dangling id already reads back as None" doesn't hold for descriptor access). The direct-lane path reaches the recorder without `select_related("user")`, and `refresh_from_db()` wipes FK caches. The broad except swallows it, so the histograms survive (observed first) but the event silently vanishes.

`select_related("user", "notebook")` on the run fetches feeding the recorder covers it, and removes a few per-transition queries as a bonus.

products/notebooks/backend/presentation/views/notebook.py
```diff
@@ -1106,9 +1107,16 @@ def sql_v2_run(self, request: Request, **kwargs):
                 )
         except Exception:
             logger.exception("notebook_sql_v2_run_start_failed", notebook_short_id=notebook.short_id)
-            run.status = NotebookNodeRun.Status.FAILED
-            run.error = "Failed to start run."
-            run.save(update_fields=["status", "error", "updated_at"])
+            # Status-guarded: a dispatch that partially started before raising could still
+            # deliver a callback, which must keep the row and stay the only reporter.
+            updated = (
```
georgemunyoro: **Optional cleanup** — this guarded UPDATE + "if won: record" dance is now hand-rolled at five sites (here, the two interrupt branches, the callback, and the temporal activity), while `_finish_direct_run` in `sql_v2_direct.py` already encapsulates it. Promoting it to a shared helper (renamed, `envelope` optional, `select_related` inside) would let this branch, both interrupt branches, and the temporal activity adopt it verbatim — and fixes the dangling-user finding as a side effect. The callback's deliberate loser-upsert should stay its own block.

products/notebooks/backend/sql_v2_metrics.py
```diff
@@ -0,0 +1,146 @@
...
+
+    The envelope is produced inside the sandbox, where user code can forge it — an absurd
+    but JSON-legal value like 1e300 would permanently poison the process-wide histogram
+    sums for every tenant. Every phase is a sub-span of the run, so the backend-measured
+    run duration (plus clock slack) is a hard ceiling; anything above it is clamped.
+    """
+    raw = envelope.get("timings") if isinstance(envelope, dict) else None
+    if not isinstance(raw, dict):
+        return {}
+    return {
```
georgemunyoro, line 124: **Optional cleanup** — this filters on `_PHASE_BY_TIMING_KEY` but returns raw-keyed values, so both consumers re-map via `_PHASE_BY_TIMING_KEY[key]`. Returning `{_PHASE_BY_TIMING_KEY[key]: ...}` directly drops the second lookup at both call sites.

georgemunyoro (approved):
Looks solid, just a few things I found with my friend Fable

products/notebooks/backend/sql_v2_metrics.py
sinan-ku: Fixed in 718bfdecf6d: the team is now always resolved from run.team_id and passed alongside the user, so $groups attribution follows the run's project rather than the user's currently active one. Asserted in the existing callback test.

products/notebooks/backend/sql_v2_serializers.py
sinan-ku: Fixed in 718bfdecf6d: timings is now a LenientTimingsField — a JSONField documented via @extend_schema_field as a str->float map but never validated as one, so a forged shape can't 400 the envelope; _sanitized_timings stays the sole validator. Added test_malformed_envelope_timings_never_strand_the_run covering both `{"exec_s": "n/a"}` and a non-dict timings.

products/notebooks/backend/sandbox/kernel/runner.py
sinan-ku: Fixed in 718bfdecf6d: _build_envelope now wraps the display fetch in try/finally (mirroring _materialize_inputs) and attaches timings to error envelopes too. For the executor gap I added a kernel_boot_s phase around _ensure_kernel — there was no phase a boot failure could report into, and it also makes cold starts visible instead of unattributed.

products/notebooks/backend/sql_v2_metrics.py
sinan-ku: Fixed in 718bfdecf6d: the shared finish helper reloads via refresh_from_db(from_queryset=...select_related("user", "notebook")) so the recorder's relations stay cached, and _capture_completed_event additionally guards descriptor access so a dangling user_id falls back to team attribution instead of losing the event. Regression test: test_completed_event_survives_a_hard_deleted_user.

products/notebooks/backend/presentation/views/notebook.py
sinan-ku: Done in 718bfdecf6d: promoted to finish_node_run in sql_v2_runs.py (envelope optional, select_related reload inside) — adopted by the direct lane, both interrupt branches, the run-start failure path, and the Temporal activity. The callback keeps its deliberate loser-upsert block.

products/notebooks/backend/sql_v2_metrics.py
sinan-ku: Done in 718bfdecf6d: _sanitized_timings now returns phase-keyed values; both consumers dropped the second lookup.

PR #72855, opened 2026-07-22
feat(ingestion): merge-event rate overflow condition via pluggable strategies

pl:
## Problem

A single `token:distinct_id` sending a burst of person-merge events (`$identify`, `$create_alias`, `$merge_dangerously`) is far more expensive to ingest than the same volume of regular events, and can cause pipeline lag well before the overall event-rate overflow condition trips.
The main-lane overflow redirect only supported one condition, a hardcoded event-rate token bucket.

## Changes

Best reviewed commit by commit:

**1. `refactor(ingestion)`: pluggable overflow strategies (behavior-neutral)**

- New `OverflowStrategy` interface in `overflow-redirect/overflow-strategy.ts`. A strategy is a pure classifier (`countTokens(headers)`); the redirect service owns all state and keeps one `MemoryRateLimiter` per strategy. A key redirects when any strategy's bucket is exhausted.
- `MainLaneOverflowRedirectConfig` takes `strategies: OverflowStrategyEntry[]` (strategy + bucket capacity + replenish rate) instead of a single `bucketCapacity`/`replenishRate` pair. All existing wiring (analytics, ingestion API server, error tracking, AI) passes a single `EventRateOverflowStrategy`, so behavior is unchanged.
- The per-key group passed to the service now carries `headersPerEvent` instead of a pre-aggregated `eventCount`, so strategies can count their own tokens from headers. Cache, Redis flagging, and overflow-lane TTL refresh are untouched.
- `overflow_redirect_rate_limit_decisions_total` gains a `strategy` label (the strategy class name), so dashboards can see which condition trips keys.

**2. `feat(ingestion)`: merge-event rate condition**

- `MergeEventRateOverflowStrategy` counts only `$identify`, `$create_alias`, and `$merge_dangerously`, classified from the Kafka `event` header, so it runs in the existing pre-body-parse step. Events without the header are not counted (fail open).
- New config: `MERGE_EVENT_OVERFLOW_BUCKET_CAPACITY` (default 0 = disabled) and `MERGE_EVENT_OVERFLOW_BUCKET_REPLENISH_RATE`. When capacity is 0 the strategy is omitted entirely rather than added with an always-empty bucket.
- Wired into the analytics pipelines only (ingestion consumer + ingestion API server). Error tracking and AI keep event-rate only.

**3–4. naming and label cleanups from review**

- `OverflowEventBatch` → `OverflowEventGroup` (each element is one key's group of events within a batch), loop vars `event` → `group`, field named `headersPerEvent`.
- Metrics label is the raw strategy class name, no derivation helper.

> [!NOTE]
> Redirect granularity is unchanged: flagging stays per `token:distinct_id`, so once a key exceeds the merge rate all of its events go to overflow, preserving per-distinct-id ordering. The strategy keys on the header distinct id, which is the merge target for `$identify`, so many-anon-ids-into-one-identity storms are caught. One-anon-into-many-identities spreads across keys and would need a token-level bucket, out of scope here.

Ran the touched jest suites locally, all passing: `overflow-redirect/*` (strategy, main lane, overflow lane, detector), `rate-limit-to-overflow-step`, `overflow-lane-ttl-refresh-step`, `errortracking/error-tracking-consumer`, `ai/pipeline`. `tsc --noEmit` clean on touched files.

New tests and the regression each catches:

- main lane merge cases: a key over the merge bucket but under the event bucket must redirect (catches the merge strategy not being consulted, or a broken event-name set), and non-merge or missing event names must not consume the merge bucket (catches over-counting that would overflow normal traffic).
- `createAnalyticsOverflowStrategies` gating: capacity 0 must omit the strategy, since a present-but-empty bucket would deny every key (that regression would redirect all analytics traffic to overflow).

Not tested by me (Claude): the full `ingestion-consumer.test.ts` integration suite needs the local dev stack (Kafka/Postgres/Redis), which wasn't running in this session. CI runs it.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (investigation + implementation session). Skills invoked: /writing-tests.
Design decisions made with the driver during the session: strategies are pure classifiers (declarative entry of strategy + limits) rather than owning their own state or a `consume` method, so the service keeps all bucket state and lifecycle; the metrics label is the plain strategy class name (a snake_case derivation helper was considered and dropped as unnecessary); the per-key group carries headers per event instead of a count so strategies stay header-driven and the step stays strategy-agnostic; merge limiting is off by default (capacity 0) and enabled per deployment via env.

### DIFF
--- a/nodejs/src/ingestion/common/overflow-redirect/main-lane-overflow-redirect.ts
+++ b/nodejs/src/ingestion/common/overflow-redirect/main-lane-overflow-redirect.ts
@@ -12,7 +12,7 @@ import {
     overflowRedirectRateLimitDecisions,
     overflowRedirectSourceEventsTotal,
 } from './metrics'
-import { OverflowEventBatch, OverflowRedirectService } from './overflow-redirect-service'
+import { OverflowEventGroup, OverflowRedirectService } from './overflow-redirect-service'
 import { OverflowRedisRepository, OverflowType, memberKey } from './overflow-redis-repository'
 import { OverflowStrategy, OverflowStrategyEntry, overflowStrategyLabel } from './overflow-strategy'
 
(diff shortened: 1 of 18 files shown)


nodejs/src/ingestion/common/overflow-redirect/main-lane-overflow-redirect.ts
```diff
@@ -54,7 +55,12 @@ export class MainLaneOverflowRedirect implements OverflowRedirectService {
             max: config.localCacheMaxSize ?? DEFAULT_LOCAL_CACHE_MAX_SIZE,
             ttl: config.localCacheTTLSeconds * 1000,
         })
-        this.rateLimiter = new MemoryRateLimiter(config.bucketCapacity, config.replenishRate)
+        this.strategies = config.strategies.map((entry) => ({
+            // The class name is the metrics label: renaming a strategy renames its label.
+            label: entry.strategy.constructor.name,
```
jose-sequeira: `constructor.name` is not a great label and it seems that there is barely any value in strategy being a class. 

```
export interface OverflowStrategyEntry {
    label: string // metrics label, e.g. 'event_rate', 'merge_event_rate'
    countTokens: (headers: EventHeaders) => number
    bucketCapacity: number
    replenishRate: number
}

export const eventRateStrategy = (): Pick<OverflowStrategyEntry, 'label' | 'countTokens'> => ({
    label: 'event_rate',
    countTokens: () => 1,
})
```

would do just as well?

jose-sequeira (approved):

nodejs/src/ingestion/common/overflow-redirect/main-lane-overflow-redirect.ts
pl: Agreed - switched in 5a43ffbfb1e. A strategy is now a plain `{ label, countTokens }` object (`OverflowStrategyEntry` extends it with the bucket limits), with `eventRateStrategy()` / `mergeEventRateStrategy()` factories, and the service reads the explicit label instead of `constructor.name`.

PR #72857, opened 2026-07-22
feat(tasks): replace turn_complete thread writeback with pr_created artifact messages

k11kirky:
## Problem

Task threads (the PostHog Code thread panel) mixed agent turn writebacks with human conversation. Threads should be a human-to-human surface, with the agent represented by the artifacts it produced. Canvases already announce themselves (`canvas_created`), but pull requests never did — and every background turn ended with a `turn_complete` agent message in the thread.

## Changes

- Removed `post_turn_complete_thread_update` and its call from the sandbox event relay (plus the now-dead final-message tracking in the relay loop). Background runs no longer write the agent's closing message into the thread.
- `list_thread_messages` excludes legacy `turn_complete` rows so previously written turn messages don't resurface in clients.
- New `post_pr_created_thread_update`: when a run's PR URL is recorded, an agent artifact message (`event="pr_created"`, `payload.pr_url`, content `[owner/repo#N](url) has been opened`) lands in the task thread. Called from the run output paths (`set_task_run_output`, `update_task_run`) and the GitHub webhook backstop (`_record_run_pr_url`), deduped per PR URL via the task's existing `pr_created` rows. Same feature-flag and channel guards as the other agent thread updates.

The client side (rendering `pr_created`/`canvas_created` as artifact rows and dropping agent turn/prompt rows) is [PostHog/code#3694](https://github.com/PostHog/code/pull/3694).

Red/green: rewrote `products/tasks/backend/tests/test_thread_updates.py` first (failing on the missing function), then implemented. Tests I (Claude) actually ran locally, all green:

- `test_thread_updates` (13 tests): pr_created content/payload, per-URL dedupe, distinct-PR handling, flag/channel/creator skip guards, canvas_created regression, turn_complete exclusion from listing
- `test_channels_api` + `test_webhooks` (100 tests)
- `test_facade`
- `test_relay_sandbox_events` (53 passed; one pre-existing test errored in my sandbox because the persons-DB sqlx fixture isn't available there — unrelated to this diff)
- `ruff check` / `ruff format` clean

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (PostHog Code cloud task) together with the paired client PR. Followed red-green testing: wrote the failing tests before each implementation change. Chose DB-backed dedupe (existing `pr_created` rows per URL) over a Redis cooldown since the agent-output path and webhook backstop can fire arbitrarily far apart.

---

### DIFF
--- a/products/tasks/backend/facade/api.py
+++ b/products/tasks/backend/facade/api.py
@@ -5327,7 +5327,7 @@ def post_canvas_created_thread_update(
         logger.exception("Failed to post canvas-created thread update", extra={"task_id": str(task_id)})
 
 
-_GITHUB_PR_URL_PATTERN = re.compile(r"github\.com/([^/]+)/([^/]+)/pull/(\d+)", re.IGNORECASE)
+_GITHUB_PR_PATH_PATTERN = re.compile(r"/([^/]+)/([^/]+)/pull/(\d+)/?", re.IGNORECASE)
 
 # Characters that could break out of a markdown [label](url) token or smuggle
 # extra markdown into the rendered thread message.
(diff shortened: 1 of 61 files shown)


tatoalo (approved):
Could we ensure the announcement even when `pr_url` is already populated? `_record_run_output_field` returns False and webhook retries can never repair the missing artifact. It would be safer to idempotently ensure the artifact whenever the persisted URL matches this webhooks one.

Agree also on greptile + veria points, thanks for this!

PR #72860, opened 2026-07-22
fix(data-warehouse): remove N+1 in managed sources list

pauldambra:
## Problem

The managed data warehouse sources list takes a long time to load (reported at ~8-10s). The list endpoint serializes every source with its `revenue_analytics_config` (a reverse one-to-one) and `created_by`, but neither was prefetched. That is a classic N+1: one extra query per source, plus a `get_or_create` **write** per source for any source missing a config. With a full page of sources, that is dozens of sequential round-trips on top of the base query.

## Changes

- `select_related("created_by", "revenue_analytics_config")` on the sources queryset, so both are folded into the main query instead of firing per source. `created_by` was previously fetched via `prefetch_related` (a second batched query) rather than a JOIN, and `revenue_analytics_config` was not fetched at all.
- Wrapped the `list` action in a tracing span (`data_warehouse.external_data_sources.list`) recording `team_id`, `sources.count`, and `schemas.count`, so the load stays diagnosable from tracing going forward. The expensive per-schema HogQL column serialization is already gated out of the list path via `include_columns`.
- Replaced the list test's brittle fixed query-count assertion with an N+1 regression guard: it asserts the query count does not grow when more sources exist.

> [!NOTE]
> This is the "load less / fetch smarter per row" fix. The remaining list cost is dominated by the volume of nested schemas serialized per source, which the tracing span now measures — a follow-up can trim that if the span shows it is still the long pole.

> [!WARNING]
> **Known limitation — missing-config sources.** `select_related` removes the per-source query only for sources that already have a `revenue_analytics_config` row (the common case: the `post_save` receiver creates one on source creation). For a source that lacks a config, `revenue_analytics_config_safe` still falls back to a per-source `get_or_create` in the list path, so the N+1 (and a write-on-read) persists for those rows. The `post_save` receiver is best-effort and only fires on create, so pre-signal or failed-creation sources can hit this. We're accepting this for now — it's the tail, not the dominant cost — and the new tracing span will show us whether it's actually a problem in practice before we invest in a null-safe read or a backfill migration. The regression test uses Stripe sources (which always get a config), so it does not exercise this path.

I (the PostHog Slack app) could not run the DB-backed test suite in this environment — there is no local Postgres reachable, so Django test setup fails at connection time. I verified:

- `ruff check` and `ruff format --check` pass on both changed files.
- The new test logic by reasoning: all per-row serializer fields now read from `select_related`/`prefetch_related` caches, so query count is constant regardless of source count.

The new test `test_list_external_data_source_query_count_does_not_scale_with_sources` should be run in CI. It catches re-introduction of a per-source query (e.g. dropping the `select_related` on the `revenue_analytics_config` reverse one-to-one) — the exact regression this PR fixes — which the old fixed-count assertion did not.

## Docs update

No user-facing behavior change.

## 🤖 Agent context

Requested via Slack by Paul, who noted this class of problem has historically needed a lighter list-path and asked for tracing spans if the cause could not be diagnosed. I diagnosed a concrete N+1 in `ExternalDataSourceSerializers` (the unprefetched `revenue_analytics_config` reverse one-to-one) as the dominant cost and fixed it directly, and added the tracing span anyway so the remaining schema-serialization cost is measurable.

The PR was then driven through the `pr-shepherd` loop: `qa-swarm` review (cheap-first router, no escalation), `review-triage`, a `simplify` pass (which trimmed defensive `isinstance` branching in the `list()` span override, since the endpoint always paginates), and `ci-shepherd`. The known-limitation callout above came out of a Codex-bot finding that the shepherd surfaced; Paul chose to accept it and rely on the new tracing to decide whether a follow-up is warranted.

Skills invoked: `/improving-drf-endpoints` (viewset/serializer change), `/writing-tests` (altered a test), and `/pr-shepherd` (+ its `qa-swarm` / `review-triage` / `ci-shepherd` / `simplify` sub-skills). Tooling: the PostHog Slack app.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1784718958454319?thread_ts=1784718958.454319&cid=C0ACRAMJUAG)*

### DIFF
--- a/products/data_warehouse/backend/tests/api/test_external_data_source.py
+++ b/products/data_warehouse/backend/tests/api/test_external_data_source.py
@@ -6,12 +6,13 @@ from datetime import date, timedelta
 from typing import Any, cast
 
 from freezegun import freeze_time
-from posthog.test.base import APIBaseTest, FuzzyInt
+from posthog.test.base import APIBaseTest
 from unittest.mock import MagicMock, Mock, PropertyMock, patch
 
 from django.conf import settings
 from django.db import connection
 from django.test import SimpleTestCase, override_settings
+from django.test.utils import CaptureQueriesContext
 from django.utils import timezone
 
 import psycopg
@@ -2089,19 +2090,32 @@ class TestExternalDataSource(APIBaseTest):
         assert "'private_key'" in response.json()["message"]
         assert "'private_key_id'" in response.json()["message"]
 
-    def test_list_external_data_source(self):
+    def test_list_external_data_source_query_count_does_not_scale_with_sources(self):
+        # N+1 regression guard: created_by, revenue_analytics_config, schemas and the latest job are
+        # all fetched up front, so listing must cost the same number of queries regardless of how many
+        # sources exist. Comparing two source counts catches a re-introduced per-source query (e.g.
+        # dropping select_related on the revenue_analytics_config reverse 1:1) without pinning a magic
+        # number. A warm-up request first primes any per-process caches so the two measurements match.
         self._create_external_data_source()
         self._create_external_data_source()
+        self.client.get(f"/api/environments/{self.team.pk}/external_data_sources/")
 
-        # A cached instance setting lookup can shave off one query depending on test order.
-        # The list no longer builds the full HogQL Database (only needed to serialize table columns,
-        # which the list omits), so it's much cheaper than the single-source read path.
-        with self.assertNumQueries(FuzzyInt(13, 15)):
+        with CaptureQueriesContext(connection) as two_sources:
             response = self.client.get(f"/api/environments/{self.team.pk}/external_data_sources/")
-        payload = response.json()
-
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(payload["results"]), 2)
+        self.assertEqual(len(response.json()["results"]), 2)
+
+        self._create_external_data_source()
+        self._create_external_data_source()
+        with CaptureQueriesContext(connection) as four_sources:
+            response = self.client.get(f"/api/environments/{self.team.pk}/external_data_sources/")
+        self.assertEqual(len(response.json()["results"]), 4)
+
+        self.assertEqual(
+            len(four_sources.captured_queries),
+            len(two_sources.captured_queries),
+            "Listing external data sources should not issue additional queries per source",
+        )
 
     def test_list_omits_table_columns_but_retrieve_includes_them(self):
         source = ExternalDataSource.objects.create(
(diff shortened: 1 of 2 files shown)


products/data_warehouse/backend/tests/api/test_external_data_source.py
```diff
@@ -2089,19 +2090,32 @@ def test_create_external_data_source_missing_required_bigquery_job_input(self):
...
+        self.assertEqual(len(response.json()["results"]), 2)
+
+        self._create_external_data_source()
+        self._create_external_data_source()
+        with CaptureQueriesContext(connection) as four_sources:
+            response = self.client.get(f"/api/environments/{self.team.pk}/external_data_sources/")
+        self.assertEqual(len(response.json()["results"]), 4)
+
+        self.assertEqual(
+            len(four_sources.captured_queries),
```
pauldambra, line 2115: > [!NOTE]

**[router]** ⚪ NIT

This 2-vs-4-source equal-count comparison is a stronger N+1 guard than the old `FuzzyInt(13, 15)` bound (which at N=2 couldn't really tell "flat" from "small per-source multiplier"). One small trade-off: there's no longer an absolute ceiling, so a regression that inflates the flat baseline — a new query added to every list call regardless of source count — would slip through. Acceptable given the test's stated purpose is N+1 guarding specifically. Not blocking.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 2 @ f893958)</sub>

Re-reviewed after the simplify pass removed the defensive `isinstance` branching in the `list()` span override. Verified safe: the standard list action always paginates, so `response.data["results"]` is always present. No new findings.

### Key findings

- None this round. (Round 1's single NIT on the query-count test was resolved during triage.)

### Convergence

None — low-danger router passes, no delegation warranted in either round.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | Confirmed the pagination assumption behind the simplified span override — no KeyError path, `response.data["results"]` always present. Danger LOW, confidence HIGH — delegated nothing. |

<details>

round 1 @ 83d83bf — ✅ APPROVE: N+1 fix + tracing span verified correct; one NIT on the replacement query-count test (resolved during triage).

</details>

---

Gilbert09 (approved):

pauldambra:
### Merge activity

* **Jul 22, 12:24 PM UTC**: @pauldambra merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/72860).

PR #72870, opened 2026-07-22
feat(data-warehouse): connect Slack source with your own Slack app

Gilbert09:
## Problem

Slack's [May 2025 API terms](https://docs.slack.dev/changelog/2025/05/29/rate-limit-changes-for-non-marketplace-apps/) prohibit apps "offered for use outside an organization" from bulk-exporting Slack message and file data. Our data warehouse Slack source connected through PostHog's shared Slack app, which puts message syncing squarely in that prohibited category. Apps a customer builds and installs in their own workspace (internal apps) are exempt — the model Fivetran and Airbyte are both converging on.

## Changes

Switches the Slack source to **bring your own Slack app**: the customer creates an app in their workspace and pastes its bot token.

- The connect form is now a single **Bot User OAuth Token** field, so new sources can only use their own app. The shared-app OAuth flow is no longer offered.
- The source caption and webhook setup include a ready-to-paste app manifest — bot scopes for channel/user discovery plus `message.channels` / `message.groups` bot events for the webhook, so one app powers both.
- Credentials resolve by preferring the pasted token. A legacy source's stored `slack_integration_id` is kept only for back-compat parsing and is ignored once a token is present, so an existing source converts in place: paste a token and it switches to the customer's app.
- No migration needed — legacy `{slack_integration_id}` job inputs parse directly against the widened config.
- Added `docsUrl` so the source links to its docs.

Webhook delivery is unchanged: it validates via the signing secret and schema mapping on the HogFunction, independent of the access token, so bring-your-own webhooks behave identically.

The user-facing source doc is updated in a companion posthog.com PR.

**Migrating existing sources:** they keep syncing. To move onto their own app, a user pastes their bot token into the existing source (or recreates it) and re-adds the webhook Event Subscriptions to their app. Sources are not auto-stopped.

Automated, run locally against `master`:

- `products/warehouse_sources/backend/temporal/data_imports/sources/slack/test_slack.py` — 30 tests pass. New coverage: `TestResolveAccessToken` (the token path never touches an Integration row; a pasted token takes precedence over a legacy integration id; missing credentials raise) and `TestAuthTestUserId` (an invalid token's error response is not mistaken for a user id, which would misscope private-channel discovery). Existing cache/schema/webhook tests updated for the renamed cache key.
- `ruff check` and `ruff format` clean.

Full app-level checks (migrations check, source-config regeneration, doc audit) were not run locally — this environment is Python 3.12 while `master` requires 3.13 — so CI covers those.


Gilbert09:
📄 Docs companion PR: PostHog/posthog.com#18795

VojtechBartos (approved):

danielcarletti (approved):

Issue #72835, opened 2026-07-22
Workflows: encrypt secret function inputs on HogFlow actions (parity with HogFunction encrypted_inputs)


mayteio:
Spun out of #66380 (flagged during review of #72573).

## Problem

`HogFunction` splits `secret: true` schema inputs out of the plaintext `inputs` blob into a dedicated `encrypted_inputs` `EncryptedJSONStringField` column (`products/cdp/backend/models/hog_functions/hog_function.py`, wired through `InputsSerializer` via the `encrypted_inputs` serializer context in `posthog/cdp/validation.py`).

`HogFlow` has no equivalent: `actions` is a plain `JSONField`, and the hog flow serializers never set the `encrypted_inputs` context. A `function*` action's `config.inputs` that uses a `secret: true` schema field (API keys, auth headers on webhook/email steps) is stored as plain JSON at rest.

The draft → publish cycle and revision history (#66380) widen the copies of that plaintext:

- `_write_draft` snapshots the content fields verbatim into the `draft` column
- once #72573 lands, every live-content write appends the same content to `HogFlowRevision` - so a rotated credential lives on in historical snapshots indefinitely
- activity-log diffs on `actions` carry it too

Exposure is the same class as live retrieve today (team-scoped API access is still required); the gap is encryption at rest and the growing number of durable copies.

## Proposed direction

Mirror the `HogFunction` pattern:

1. `encrypted_inputs` column on `HogFlow` (or per-action storage keyed by action id, since one flow has many function steps - needs design)
2. Serializer split on write + masked read-back, reusing `InputsSerializer`'s existing secret handling
3. Redact secret values from `draft` snapshots, `HogFlowRevision.content`, and activity-log diffs (restore/publish must then re-attach live secrets rather than round-tripping them through snapshots)
4. Decide on backfill for existing rows and existing revision snapshots

3 is the part revisions make urgent: restore-to-draft copies snapshot content back into the draft, so redaction has to preserve restore semantics for steps whose secrets still exist on the live row.

## References

- #66380 (revision history section + follow-ups)
- #72573 (review finding)
- `HogFunction.encrypted_inputs` + `posthog/cdp/validation.py` `InputsItemSerializer` for the reference implementation

PR #72880, opened 2026-07-22
feat(workflows): add encrypted inputs columns to hog flow

mayteio:
## Problem

`HogFunction` keeps `secret: true` schema inputs in a Fernet-encrypted `encrypted_inputs` column, split out of the plaintext `inputs` blob. `HogFlow` has no equivalent, so a function action's secret `config.inputs` (API keys, auth headers on webhook/email steps) sit as plaintext in the `actions` JSON, and get copied verbatim into `draft`, activity-log diffs, and revision snapshots.

This is the first, purely additive step toward fixing that. It adds the storage columns and nothing else. Part of #72835.

## Changes

Two nullable `EncryptedJSONStringField` columns on `HogFlow`:

- `encrypted_inputs` - live secret function-action inputs, keyed by action id then input key.
- `draft_encrypted_inputs` - the same for pending draft edits, promoted to `encrypted_inputs` on publish and dropped on discard.

The migration is two plain nullable `ADD COLUMN`s, no default, no backfill. `posthog_hogflow` is not a hot table. Nothing reads or writes these columns yet.

> [!NOTE]
> Deploy order is load-bearing. This migration must merge and apply before the follow-up worker PR, which adds `encrypted_inputs` to the worker's fixed `SELECT` - if that column does not exist yet, every flow fetch errors on the worker hot path. The API-side split/mask lands in a third, flag-gated PR.

- `makemigrations --check` reports no state drift.
- `sqlmigrate workflows 0012` shows two `ALTER TABLE ... ADD COLUMN ... text NULL` statements, no `DROP DEFAULT`.
- No tests here - there is no behavior to exercise until the reader/writer PRs land.

## 🤖 Agent context

I directed this; Claude (Claude Code) wrote it. We scoped #72835 first and landed on mirroring the `HogFunction.encrypted_inputs` pattern, but keyed by action id since one flow has many function steps. Split into three PRs because the strip-secrets-from-`actions` behavior can't ship until the nodejs worker can decrypt, and the worker's `SELECT` needs the column to exist first - so migration leads, worker reader second, API split/mask third.

Invoked the `/django-migrations` skill for this migration and `/writing-tests` while scoping the follow-up worker test.

---


mayteio:
Superseded by #72913, which collapses the migration + worker + API into one self-testable PR (CI can build the schema and exercise it end to end). It'll be re-split into deploy-ordered PRs (migration → worker → API) from there before merge.

PR #72893, opened 2026-07-22
feat(workflows): encrypt secret function inputs on hog flow actions

### LINKED ISSUE #72835: Workflows: encrypt secret function inputs on HogFlow actions (parity with HogFunction encrypted_inputs)

mayteio:
## Problem

Top of the stack (#72880 migration, #72881 worker read). A `HogFlow` function/email/sms step can carry secret inputs - API keys, auth headers - declared `secret: true` on its template schema. Today they sit as plaintext in the `actions` JSON, and get copied verbatim into `draft` snapshots, activity-log diffs, and (once its write wiring lands) `HogFlowRevision.content`. `HogFunction` has solved this for years with an encrypted column; this brings `HogFlow` to parity. Closes #72835.

## Changes

Mirror the `HogFunction.encrypted_inputs` pattern, keyed by action id since one flow has many function steps.

- **Write:** secrets are split out of `actions` (and the derived `trigger`) into the encrypted `encrypted_inputs` column on create/update/graph, and into `draft_encrypted_inputs` on draft writes. The map is rebuilt from the incoming action set each write, so deleted actions' secrets drop out rather than orphaning.
- **Read:** `to_representation` masks every set secret to `{"secret": true}` in both live and draft actions, so no client (web, API, MCP) ever receives the value. No new serializer fields, so generated frontend/MCP types are unchanged (verified: `build:openapi` produces no diff).
- **Resent-mask recovery (spoof-to-wipe fix):** when a client sends back the `{"secret": true}` marker unchanged, the stored value is recovered from the encrypted column instead of being wiped. This context was never wired for HogFlow before - a latent bug where any re-save echoing the mask silently dropped the secret.
- **Publish / draft / discard / test runs:** publish re-attaches secrets from the live+draft encrypted maps and re-splits into the live column (never wiping them), then clears the draft's secret column. Discard drops it. Inline `use_draft` test runs re-hydrate secrets before dispatch (that path bypasses the worker's decrypt).
- **Activity log:** `encrypted_inputs` / `draft_encrypted_inputs` are masked in diffs; `actions` is now safe since it's stripped.

> [!NOTE]
> `HogFlowRevision` deliberately gets **no** encrypted column. Snapshots store stripped content; restore re-hydrates from the live row. Rotated or deleted secrets are not resurrected from history - that's requirement #3 of the issue and the whole point. The draft/publish paths stay behind the existing `workflows-revisions` flag.

New `TestHogFlowSecretInputs` in `test_hog_flow.py`:

- secret is encrypted at rest (raw column is ciphertext, `actions` has no plaintext) and masked on create + retrieve, non-secret inputs untouched
- resent mask preserves the stored value; a real new value replaces it (parameterized)
- a metadata-only PATCH leaves the stored secret intact
- publish promotes a draft-rotated secret to live without wiping, clears the draft column, and the draft snapshot carries no plaintext
- activity-log diffs never contain the secret value and mask the encrypted columns

Ran locally: `TestHogFlowSecretInputs` 6/6 pass; full `test_hog_flow.py` + `test_hog_flow_draft_publish.py` 202/202 pass; `mypy --cache-fine-grained .` clean; ruff clean; `build:openapi` no diff. I (the agent) did not manually exercise the MCP tools against a running stack - that's the recommended follow-up before marking ready.

## 🤖 Agent context

I directed this; Claude (Claude Code) wrote it. Design choice: partition at the serializer/viewset layer (not the model `save()`, unlike HogFunction) because knowing which inputs are secret needs the template schema, only resolvable at the API layer, and the draft-vs-live split is an API concept. Kept the write path centralized in `create`/`update`/`_write_draft` so the graph and publish endpoints inherit it. Confirmed no trigger-source template currently declares a secret input, but still strip the derived `trigger` defensively.

Invoked `/improving-drf-endpoints` before the serializer work and `/writing-tests` before the tests.

---

### DIFF
--- a/posthog/models/activity_logging/activity_log.py
+++ b/posthog/models/activity_logging/activity_log.py
@@ -277,6 +277,10 @@ field_with_masked_contents: dict[AuditableScope, list[str]] = {
         # Full content snapshot including action inputs (auth headers, API keys) — record that a
         # draft was staged/published/discarded, never its contents.
         "draft",
+        # Encrypted secret function inputs (Fernet ciphertext) — a diff would be noise at best and
+        # leak-adjacent at worst; record that they changed, never the values.
+        "encrypted_inputs",
+        "draft_encrypted_inputs",
     ],
     "OrganizationDomain": [
         "_scim_bearer_token",
--- a/products/workflows/backend/api/hog_flow.py
+++ b/products/workflows/backend/api/hog_flow.py
@@ -2,6 +2,7 @@ import re
 import json
 import uuid as uuid_mod
 import dataclasses
+from copy import deepcopy
 from datetime import timedelta
 from typing import Any, Optional, cast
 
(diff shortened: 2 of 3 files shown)


mayteio:
Superseded by #72913, which collapses the migration + worker + API into one self-testable PR (CI can build the schema and exercise it end to end). It'll be re-split into deploy-ordered PRs (migration → worker → API) from there before merge.

PR #72905, opened 2026-07-22
fix(aio): make Hog eval test paths target-aware

Radu-Raicea:
## Problem

Hog evaluation templates run against either a single generation or a whole trace, and each target exposes different Hog globals. A recent change made the built-in templates target-independent, but the two places that let you *test* Hog code before saving were never updated:

- the evaluation editor's "Test on sample" button, and
- the Max authoring tool that dry-runs Hog source.

Both always sampled individual `$ai_generation` events and ran the generation runtime, regardless of the evaluation's target. So when you wrote a trace-target evaluation and tested it, the preview ran against a single generation with generation-level globals — not the whole trace with trace-level globals it gets online. The editor tooltip even acknowledged this mismatch.

## Changes

- Add a `target` parameter to the `test_hog` endpoint. When it's `trace`, sample recent whole traces (from the triggering generation, filtered by the same conditions) and run the code against trace-level globals, matching the online trace workflow. `generation` keeps the existing behavior and stays the default.
- Give Max's Hog test tool the same `target` option.
- Share the "sample recent traces + run bytecode against trace globals" logic with the online trace-evaluation path so the preview and the real run can't drift.
- Frontend passes the evaluation's target when testing, deep-links trace results to the trace view, and drops the tooltip that warned the preview didn't match online behavior.

Automated tests only (no local services available to the agent for a manual run):

- Backend API: a trace-target `test_hog` request routes to the trace path (whole-trace globals) instead of the generation query, and maps results back correctly; the empty-sample case returns the "no recent traces" message. These catch the exact regression fixed here — a trace-target request silently falling through to the generation path.
- Backend tool: a trace-target dry-run invokes the trace helper and formats per-trace results.
- `ruff`, and repo-style `mypy --cache-fine-grained` on the changed backend modules, pass.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the `llma-evaluation-test-hog` MCP tool description to document the new `target` option.

## 🤖 Agent context

Investigated from a support escalation about Hog trace-level evals. The gap was that the test paths (`evaluations.py` `test_hog` and `run_hog_eval.py`) still queried only `$ai_generation` and ran the generation runtime regardless of target — a follow-up to the earlier target-independent-templates work, which fixed the templates and runtime but not the test surfaces.

Skills invoked while producing this PR: `/improving-drf-endpoints`, `/adopting-generated-api-types`, `/writing-tests`.

Generated API/MCP artifacts (`api.schemas.ts`, `api.zod.ts`, MCP schemas/definitions) were updated to add the optional `target` field; the agent couldn't run `hogli build:openapi` in this environment (no local DB / orval), so these were edited to match the existing codegen patterns — CI's OpenAPI regeneration is the source of truth if anything differs.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08JRHCUSF3/p1784725233672719?thread_ts=1784725233.672719&cid=C08JRHCUSF3)*


posthog/temporal/ai_observability/run_trace_evaluation.py
```diff
@@ -199,6 +197,183 @@ def fetch_trace_for_evaluation(team_id: int, trace_id: str, window_start: dateti
...
+
+def run_hog_eval_over_recent_traces(
+    *,
+    team: Team,
+    bytecode: list[Any],
+    condition_filter: ast.Expr | None,
+    sample_count: int,
+    allows_na: bool,
+    window_seconds: int = TRACE_EVAL_DEFAULT_WINDOW_SECONDS,
+    lookback_days: int = 7,
```
carlos-marchal-ph: `lookback_days` is a bare 7 here, and the "last 7 days" text is hardcoded separately in both callers. If we ever change the lookback those messages go stale silently. Could we pull it out as a named constant next to the `TRACE_EVAL_*_WINDOW_SECONDS` ones and derive the copy from it?

products/ai_observability/backend/api/evaluations.py
```diff
@@ -655,6 +692,71 @@ class TestHogResponseSerializer(serializers.Serializer):
...
+        {
+            "sample_count": sample_count,
+            "allows_na": allows_na,
+            "condition_count": len(conditions),
+            "result_count": len(results),
+            "pass_count": sum(1 for r in results if r["result"] is True),
+            "fail_count": sum(1 for r in results if r["result"] is False),
+            "error_count": sum(1 for r in results if r["error"]),
+            "na_count": sum(1 for r in results if r["result"] is None and not r["error"]),
+            "no_events": not results,
```
carlos-marchal-ph, line 749: The trace path and the generation empty path both include `no_events` in this event, but the generation success path doesn't. Same event name, three shapes. Could we always emit it so the schema stays stable for whoever queries it?

carlos-marchal-ph (approved):

PR #72906, opened 2026-07-22
feat(data-warehouse): add direct-query support for clickhouse sources

Gilbert09:
## Problem

Direct-query (live federated querying without syncing to the warehouse) supports Postgres, MySQL, Snowflake and Redshift, but not ClickHouse. A ClickHouse source could only be imported, never queried live. This adds ClickHouse (and ClickHouse Cloud) as a first-class direct-query engine so a source configured with `access_method: 'direct'` runs HogQL straight against the external ClickHouse via its native driver.

## Changes

Direct query uses a pluggable adapter stack keyed on `source.direct_engine`, not ClickHouse `remote()` federation. HogQL compiles to the source's native dialect and executes through the source's own driver. This wires ClickHouse into every layer of that stack.

**HogQL layer**
- `DirectClickHouseTable` renders `database.table` as an external table reference. ClickHouse is the printer's native dialect, so unlike the other direct tables (which target a different SQL dialect and raise) this one actually prints. It raises from `to_printed_postgres` / `to_printed_mysql`.
- `ClickHouseAdapter` compiles and executes read-only HogQL through the source's SSH tunnel and client. A first-keyword read-only guard rejects anything that isn't `SELECT` / `WITH`.

**Materialization layer** (synced + dual-mode)
- `_ClickHouseEngine` plus `clickhouse_helpers` / `direct_clickhouse` cover schema reconcile, table upsert, reprojection and the dual-mode virtual tables — all dispatched off `source.direct_engine`.
- ClickHouse namespaces a table as `database.table`: no catalog level, no foreign keys. Column types are near-identity, so the type mapper uses the source type verbatim and only wraps `Nullable` when needed.
- It has no legacy unqualified rows (brand-new engine), so it uses the generic name-substitution path like MySQL/Snowflake/Redshift rather than the Postgres-only bespoke remapping.

**Enrollment**
- `DIRECT_ENGINE_BY_SOURCE_TYPE` maps `CLICKHOUSE` and `CLICKHOUSECLOUD` to the `clickhouse` engine. That single entry auto-enrols both in `direct_capable_source_types()` and the create allowlist.
- Frontend `DIRECT_QUERY_SOURCE_TYPES` and the presentation-layer engine choices / unsupported-source message updated to match.

No new `is_direct_clickhouse` branches in the presentation layer — the source-agnostic guard (`.github/scripts/check-dwh-source-agnostic.py`) stays at zero.

Automated tests I (the agent) ran locally in a Python 3.13 venv:

- `posthog/hogql/direct_sql/test/test_clickhouse_adapter.py` — new. `DirectClickHouseTable` renders `db.table` / table-only, escapes identifiers, and refuses the Postgres/MySQL dialects; the read-only guard allows `SELECT`/`WITH`/lowercase/comment-prefixed and rejects insert/alter/drop/truncate/system/multi-statement.
- `posthog/hogql/direct_sql/test/test_registry_and_capability.py` — extended for the clickhouse adapter + `CLICKHOUSE`/`CLICKHOUSECLOUD` capability enrolment.
- `products/data_warehouse/backend/test/test_direct_query_engines.py` — extended: clickhouse resolves from the registry, doesn't resolve location in warehouse mode, maps explicit metadata to `(None, database, table)` (no catalog), and returns no name substitutions.

All 56 pass. `check-dwh-source-agnostic.py` reports 0 references. `makemigrations --check` is clean (metadata-only, no schema change). `ruff check`/`ruff format` clean.

I could not exercise a live end-to-end query against a real external ClickHouse from here. Two integration boundaries are covered by structure and unit tests but not a live run: the adapter's `execute` (tunnel + client `query()` + column-type mapping) and the dual-mode `schema_metadata` → virtual-table path. Worth a manual smoke test against a real ClickHouse before rollout.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I directed this; Claude implemented it. It follows the existing adapter pattern the earlier direct-query engines established, so the shape mirrors `direct_redshift` / `redshift_helpers` closely. One design call worth flagging: an initial pass had the clickhouse engine delegate name-substitution to the Postgres bespoke `reconcile_refresh_name_substitutions`, which runs Postgres-specific rename/consolidate SQL. That was wrong for a brand-new engine with no legacy rows, so it was dropped in favour of the generic path (matching the other non-Postgres engines).

Skills invoked while shaping this: `/improving-drf-endpoints`, `/writing-tests`, `/django-migrations` (confirmed no migration needed), `/implementing-mcp-tools` (context only — no tool changes here).


posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,141 @@
...
+                        secure=config.secure,
+                        verify=config.verify,
+                        query_timeout=statement_timeout_seconds,
+                        # A server-side execution cap on top of the socket timeout. `max_execution_time`
+                        # is a standard setting managed servers accept (unlike a readonly override).
+                        settings={"max_execution_time": statement_timeout_seconds},
+                    )
+                    try:
+                        result = client.query(request.sql, parameters=request.values or None)
+                        rows = result.result_rows
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 187877d. `execute` now streams row blocks via `query_row_block_stream` and stops one row past a 1M cap (`DIRECT_CLICKHOUSE_MAX_ROWS`), so an unbounded raw `SELECT` never buffers the full result set in the worker — memory is bounded to the cap plus a single block, matching the `fetchmany(cap+1)` guard in the Snowflake/Redshift adapters. Exceeding the cap raises an `ExposedHogQLError` telling the caller to add a `LIMIT`, and the `hogql_direct_query_row_cap_exceeded_total` metric is incremented. Added an adapter test for an over-cap raw result.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,141 @@
...
+                        password=config.password,
+                        secure=config.secure,
+                        verify=config.verify,
+                        query_timeout=statement_timeout_seconds,
+                        # A server-side execution cap on top of the socket timeout. `max_execution_time`
+                        # is a standard setting managed servers accept (unlike a readonly override).
+                        settings={"max_execution_time": statement_timeout_seconds},
+                    )
+                    try:
+                        result = client.query(request.sql, parameters=request.values or None)
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 187877d. `execute` now streams row blocks via `query_row_block_stream` and stops one row past a 1M cap (`DIRECT_CLICKHOUSE_MAX_ROWS`), so an unbounded raw `SELECT` never buffers the full result set in the worker — memory is bounded to the cap plus a single block, matching the `fetchmany(cap+1)` guard in the Snowflake/Redshift adapters. Exceeding the cap raises an `ExposedHogQLError` telling the caller to add a `LIMIT`, and the `hogql_direct_query_row_cap_exceeded_total` metric is incremented. Added an adapter test for an over-cap raw result.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,141 @@
...
+    different keyword, so a first-keyword check is a robust read-only guard for raw queries."""
+    sql = ensure_single_direct_statement(sql)
+    statements = [statement for statement in sqlparse.parse(sql) if str(statement).strip(" \t\r\n;")]
+    if len(statements) != 1:
+        raise ExposedHogQLError(RAW_CLICKHOUSE_READ_ONLY_ERROR)
+    for token in statements[0].flatten():
+        if token.is_whitespace or token.ttype in sqlparse_tokens.Comment:
+            continue
+        value = token.value.upper()
+        if token.ttype in sqlparse_tokens.Keyword and value in ("SELECT", "WITH"):
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 187877d. The read-only gate now inspects the whole statement instead of just the first keyword: it requires `get_type() == "SELECT"` and then scans every token, rejecting any DDL or any DML keyword other than `SELECT` (matching the Snowflake/Redshift adapters). `WITH 1 AS x INSERT INTO target SELECT x` classifies as `INSERT` and is rejected, and a write smuggled into a subquery is caught by the token scan. Added parameterized cases for both.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+    enforced our side (as with Snowflake/Redshift): the statement must be a single ``SELECT`` and
+    any DDL — or any DML keyword other than ``SELECT`` anywhere in the statement — is rejected.
+    A first-keyword check is not enough: ClickHouse accepts ``WITH 1 AS x INSERT INTO t SELECT x``,
+    which starts with ``WITH`` but writes, so the whole statement is inspected. HogQL-authored
+    queries only ever emit SELECT. String values and quoted identifiers aren't tagged DML/DDL, so
+    a literal or alias like ``'DELETE'`` is unaffected.
+    """
+    sql = ensure_single_direct_statement(sql)
+    statements = [statement for statement in sqlparse.parse(sql) if str(statement).strip(" \t\r\n;")]
+    if len(statements) != 1 or statements[0].get_type() != "SELECT":
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 7065d14. The raw read-only gate now also rejects ClickHouse network/file/exec table functions — `url`, `urlCluster`, `s3`/`s3Cluster`/`gcs`, `remote`/`remoteSecure`/`cluster`/`clusterAllReplicas`, `mysql`/`postgresql`/`mongodb`/`redis`/`sqlite`, `jdbc`/`odbc`, `hdfs`/`hdfsCluster`, `file`/`input`, `azureBlobStorage*`, `deltaLake`/`hudi`/`iceberg`, and `executable`. They run from a plain SELECT, so the DML/DDL check alone didn't stop `SELECT * FROM url('http://169.254.169.254/…')`.

The statement's parse tree is walked for function *calls*, so the match only triggers on `url(...)` etc., not on a column literally named `url` (that tokenizes as an identifier, not a `Function`) — added cases for both. HogQL-authored queries only emit a plain SELECT over the source's own tables, so this affects only the raw passthrough path. This mirrors the blocked-function lists already in the Snowflake and Redshift adapters.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,229 @@
...
+            with request.timings.measure("clickhouse_execute"), observe_direct_query("clickhouse"):
+                # The SSH tunnel (if any) is opened for the life of the query; the native driver
+                # connects to the external ClickHouse directly, never through PostHog's cluster.
+                with clickhouse_source.direct_query_client(
+                    config,
+                    request.team.pk,
+                    query_timeout=statement_timeout_seconds,
+                    # A server-side execution cap on top of the socket timeout. `max_execution_time`
+                    # is a standard setting managed servers accept (unlike a readonly override).
+                    settings={"max_execution_time": statement_timeout_seconds},
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 3a108b0. The streaming fetch now enforces a client-side wall-clock deadline: once the statement timeout (default 600s) elapses it stops reading and closes the client, which drops the connection so ClickHouse cancels the query. That releases the worker and SSH tunnel at the cap regardless of any in-query `SETTINGS max_execution_time=0` override, and it also covers the "tiny blocks keep the socket read-timeout alive" case since the deadline is checked on every block. I went with the wall-clock backstop over rejecting `SETTINGS` clauses because a text/parse-based reject is brittle here — sqlparse tokenizes the `SETTINGS` clause keyword and a column literally named `settings` identically — whereas the deadline enforces the bound directly and can't be worked around from SQL. Added a test that trips the deadline. The server-side `max_execution_time` is still set as the first line of defense for the normal case.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,241 @@
...
+    The per-block wall-clock deadline is the real timeout backstop: a raw query can override the
+    server-side ``max_execution_time`` with its own ``SETTINGS`` clause, and dribbling rows out in
+    tiny blocks keeps the socket read-timeout from ever firing — so a slow query would otherwise
+    pin the worker and SSH tunnel indefinitely. Bailing once ``deadline_seconds`` elapses (and
+    closing the client, which drops the connection so ClickHouse cancels the query) enforces the
+    cap regardless of any in-query setting.
+    """
+    rows: list = []
+    started = perf_counter()
+    with client.query_row_block_stream(sql, parameters=parameters) as stream:
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 7de8410. Raw queries can no longer supply a query-level `SETTINGS` clause: `prepare_raw_sql` now rejects it, which closes the whole override class — both the `max_block_size` inflation described here and the earlier `max_execution_time` reset. The clause is detected structurally (a `SETTINGS` token followed by `<name> =`), so a column literally named `settings`, `WHERE settings = 'x'`, or `AS settings` are all still allowed (tests added for each). As defense in depth the connection now pins `max_block_size` to the ClickHouse default (65536); since raw overrides are rejected, that per-block bound holds, so a block is materialized at a bounded size before the streaming row cap is checked. HogQL-authored queries don't take the raw path, so their printer-emitted settings are unaffected.

danielcarletti (commented):
Reviewed against the dc-review-pr checklist. Solid work that faithfully follows the established direct-query adapter pattern, and the raw-SQL read-only / SSRF / settings guards are thorough and well tested. Dual mode (synced ClickHouse source + direct toggle) is correctly wired on the backend — capability → virtual-source selection → `build_direct_table_for_schema` clickhouse branch → reconcile persisting `schema_metadata` in both modes. One inline nit below.

**Non-blocking — frontend enrolment gap for ClickHouse Cloud:** the backend enrolls both `CLICKHOUSE` and `CLICKHOUSECLOUD` as direct-capable (`DIRECT_ENGINE_BY_SOURCE_TYPE`, covered by the `clickhousecloud_direct` / `clickhouse_synced_enabled` capability tests), but the frontend `DIRECT_QUERY_SOURCE_TYPES` (`products/data_warehouse/frontend/shared/components/forms/schemaGroupingUtils.ts:4`) only adds `'ClickHouse'`. The ClickHouse Cloud connector name is the distinct string `'ClickhouseCloud'`, and `supportsDirectQuery()` does an exact `.includes()` with no normalization — so for a ClickHouse Cloud source the access-method selector never appears in the New Source wizard and the direct-query UI is hidden on the config/schemas tabs. That blocks *all* direct capability for ClickHouse Cloud from the UI — pure-direct and dual-mode alike — even though the backend fully supports both. Suggested fix: add `'ClickhouseCloud'` to `DIRECT_QUERY_SOURCE_TYPES`, and extend the `supportsDirectQuery` unit coverage to include it so the gap can't silently regress.

services/mcp/src/tools/generated/platform_features.ts
```diff
@@ -540,6 +540,7 @@ const OrganizationEnforce2faSchema = PartialUpdateParams.extend(
         members_can_invite: true,
         members_can_create_projects: true,
         members_can_use_personal_api_keys: true,
+        members_can_see_org_members: true,
```
danielcarletti, line 543: [nit] This PR adds `members_can_see_org_members` to the generated MCP platform-features files (here and `services/mcp/src/generated/platform_features/api.ts`), which is unrelated to ClickHouse direct query and undisclosed in the description — almost certainly codegen drift picked up by `hogli build:openapi`. Fix: land it as its own codegen-sync PR, or note it in the description as incidental regenerated output so a future reviewer is not confused by the unrelated diff.

Gilbert09: Good catch. This was stale codegen — `members_can_see_org_members` already landed on `master`, so it was only showing as a PR-owned change because the branch was behind. I have merged the latest `master` in, and those two `platform_features` files have dropped out of the PR diff entirely. Nothing unrelated remains in the generated output now.

MarconLP (approved):

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,284 @@
...
+RAW_CLICKHOUSE_BLOCKED_FUNCTION_ERROR = "This ClickHouse table function is not allowed in direct queries."
+
+# Table functions that read over the network, from the local filesystem, or spawn a process.
+# They parse as a plain SELECT (`SELECT * FROM url(...)`), so the read-only DML/DDL gate doesn't
+# stop them — a raw query could otherwise reach the connected server's metadata endpoint (SSRF),
+# federate in data from other databases the server can see, or read local files / run programs.
+# HogQL-authored queries never emit these; only the raw passthrough path can, so block them by
+# name there. Matched case-insensitively and only when the name is a function call (a column
+# literally named e.g. `url` tokenizes as an identifier, not a Function, so it's unaffected).
+_RAW_CLICKHOUSE_BLOCKED_TABLE_FUNCTIONS = frozenset(
```
georgemunyoro, line 52: The read-only denylist matches how Redshift/Snowflake guard the raw path, but ClickHouse's dangerous-function surface is much larger than theirs, and this list has some real holes today:

- **Missing `*Cluster` twins:** we block `ICEBERG`/`HUDI`/`DELTALAKE` (and some cluster variants) but not `ICEBERGCLUSTER`, `HUDICLUSTER`, `DELTALAKECLUSTER` — the cluster variants do the same remote read / SSRF. Worth auditing the base functions 1:1 against their `*CLUSTER` forms.
- **Missing `MERGE` and `DICTIONARY`:** `merge('.*', '.*')` reads across every table/db the connected server can see (bypasses per-source scoping); `dictionary(...)` can be backed by a remote source.

Concretely, `SELECT * FROM icebergCluster('c','http://169.254.169.254/...')` or `SELECT * FROM merge('.*','.*')` passes the guard today.

As a later followup we should maybe consider flipping this to an allowlist. HogQL only ever emits the source's own `database.table`, so the raw path has no legitimate need for *any* table function, and an allowlist would fail closed as new functions land.

posthog/hogql/direct_sql/clickhouse_adapter.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+def _iter_function_names(token: TokenList) -> "Iterator[str]":
+    """Yield the name of every function call in the statement, recursing into groups.
+
+    Table functions in a FROM clause (``url(...)``, ``s3(...)``) and scalar calls both parse as
+    ``sqlparse.sql.Function``; a bare identifier of the same spelling (``SELECT url``) does not.
+    """
+    for child in token.tokens:
+        if isinstance(child, Function):
```
georgemunyoro: `_iter_function_names` only catches `sqlparse.sql.Function` nodes, so a quoted name like `"remote"(...)` or `` `remote`(...) `` isn't recognized and skips the blocklist entirely. Worth normalizing / stripping identifier quoting before the name check. Same root cause as the blocklist gap above, and another argument for an allowlist.

products/data_warehouse/backend/direct_clickhouse.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    return existing_table
+
+
+def hide_direct_clickhouse_table(table: DataWarehouseTable | None) -> None:
+    if table is not None and not table.deleted:
+        table.soft_delete()
+
+
+def rename_direct_clickhouse_join_references(*, team_id: int, old_name: str, new_name: str) -> None:
```
georgemunyoro: This has zero callers repo-wide. It was shaped from `rename_direct_postgres_join_references`, which *is* invoked (`postgres_helpers.py:399`); MySQL/Snowflake/Redshift deliberately have no such function.

products/data_warehouse/backend/clickhouse_helpers.py
```diff
@@ -0,0 +1,274 @@
...
+) -> list[str]:
+    """Persist `schema_metadata` on every ClickHouse row + (direct mode only) upsert its live-query
+    `DataWarehouseTable`. Returns stale schema names that got soft-deleted (direct only)."""
+    from products.warehouse_sources.backend.facade.models import ExternalDataSchema
+
+    is_direct = source.is_direct_query
+    source_schema_names = [s.name for s in source_schemas]
+    default_database = (source.job_inputs or {}).get("database")
+    schema_models = {
+        s.name: s for s in ExternalDataSchema.objects.filter(team_id=team_id, source_id=source.id, deleted=False)
```
georgemunyoro: `reconcile_redshift_schemas` uses `.select_related("table")` on this query; this clone dropped it, so every `schema_model.table` / `s.table` access downstream is a separate query. Add `.select_related("table")` to match — otherwise a source with many schemas does N+1 round-trips on each refresh.

products/warehouse_sources/backend/models/util.py
```diff
@@ -550,6 +550,27 @@ def snowflake_columns_to_dwh_columns(columns: list[tuple[str, str, bool]]) -> di
     }
 
 
+def clickhouse_column_to_dwh_column(_column_name: str, clickhouse_type: str, nullable: bool) -> dict[str, Any]:
+    # The source is already ClickHouse, so the type string is a valid ClickHouse type — used verbatim.
+    # `system.columns.type` usually already encodes nullability, so only wrap when it doesn't.
+    ch_type = clickhouse_type.strip()
+    if nullable and not (ch_type.startswith("Nullable(") or ch_type.startswith("LowCardinality(Nullable(")):
```
georgemunyoro: Edge case: a `LowCardinality(String)` reported with `is_nullable=True` becomes `Nullable(LowCardinality(String))`, which ClickHouse rejects (LowCardinality must be the outermost wrapper). Unlikely from the sync path (nullable is derived from the type string there), but the stored-metadata path sets `is_nullable` independently of the type string, so the two can diverge. Cheap to guard for.

georgemunyoro (approved):
Looks good, besides extending the blocked table functions denylist and `_iter_function_names`, everything else can be done as follow-up if necessary

posthog/hogql/direct_sql/clickhouse_adapter.py
Gilbert09: Good catch — closed the concrete holes: added the \`*Cluster\` twins (\`icebergCluster\`, \`hudiCluster\`, \`deltaLakeCluster\`), the lakehouse storage variants (\`icebergS3/Azure/HDFS/Local\`, \`deltaLakeS3/Azure/Local\`), and \`merge\`/\`dictionary\`.

Agree an allowlist is the stronger fail-closed design, but I left it as a followup for this PR: the raw path mixes table functions with legitimate scalar calls in the same token stream, and reliably distinguishing FROM-clause table-function position from scalar calls in sqlparse is fiddly enough that getting it wrong would either block valid scalar functions or reintroduce a hole. Tracking it separately rather than rushing it here.

posthog/hogql/direct_sql/clickhouse_adapter.py
Gilbert09: Fixed. \`_iter_function_names\` now scans the flattened token stream for a name/quoted-symbol token immediately followed by \`(\`, instead of relying on \`sqlparse.sql.Function\` nodes — a double-quoted \`"remote"(...)\` parses as an identifier plus a separate parenthesis and never becomes a Function, so the old check skipped it. Quoting is stripped before the blocklist comparison so \`"remote"\` and \`\`remote\`\` both match \`remote\`. Added \`double_quoted_remote\` and \`backtick_remote\` cases to the guard test.

products/data_warehouse/backend/direct_clickhouse.py
Gilbert09: Removed — confirmed it had zero callers repo-wide (the Postgres variant it was shaped from is wired in at \`postgres_helpers.py:399\`, but the ClickHouse path never called this).

products/data_warehouse/backend/clickhouse_helpers.py
Gilbert09: Fixed — added \`.select_related("table")\` to match \`reconcile_redshift_schemas\`, so the downstream \`schema_model.table\` / \`s.table\` / \`matched.table\` accesses no longer do N+1 round-trips.

products/warehouse_sources/backend/models/util.py
Gilbert09: Guarded. When a \`LowCardinality(...)\` type is reported nullable, we now nest as \`LowCardinality(Nullable(...))\` instead of wrapping the whole thing in \`Nullable(...)\`, since ClickHouse requires LowCardinality to be the outermost wrapper. Added parameterized coverage in \`test_clickhouse_column_to_dwh_column_nullable_wrapping\`.

PR #72912, opened 2026-07-22
feat(self-driving): add Azure DevOps code integration

willwearing:
## Problem

PostHog's code integration is built around GitHub. Adding Azure DevOps safely needs a provider implementation, a common repository contract, shared egress behavior, and a way to configure the connection.

This PR adds that foundation. The self-driving task runner is still GitHub-specific, so task execution against Azure DevOps remains a follow-up.

## Changes

- Adds the `azure-devops` integration kind with the standard kebab-case identifier and an accurate `Azure DevOps` label.
- Adds a shared Azure DevOps egress client with a pinned host and API version, PAT Basic auth, redirects disabled (unexpected 3xx responses raise), bounded retries for safe read requests with a per-caller attempt cap (web request paths fail fast, the Temporal warehouse sync keeps full retries), and request telemetry.
- Reuses that transport in the existing Azure DevOps warehouse source instead of maintaining a second HTTP implementation.
- Adds a provider-neutral code-host contract for repository listing, default branches, branch creation, pull request creation, and clone URLs, with GitHub and Azure DevOps adapters. Adapters translate provider errors into a shared `CodeHostIntegrationError`, so the repository endpoint handles any provider's failure uniformly, and both providers use the same `full_name` search semantics.
- Adds a typed integration config serializer. The PAT is write-only at the API boundary and stored only in `sensitive_config`. An invalid or expired PAT surfaces a token-specific error and is persisted on `integration.errors`, matching the GitHub pattern.
- Adds a provider-neutral repository endpoint while preserving the existing GitHub endpoint.
- Adds the Azure DevOps integration catalog entry and PAT-based setup modal in project settings, following the GitLab modal's API client and form handling.
- Regenerates the OpenAPI clients and exposes repository discovery as a read-only MCP tool with schema snapshots.
- Adds regression coverage for auth, host pinning, redirect handling, retry caps, non-idempotent retry behavior, identifier validation, branch and PR construction (including empty ref-update responses), API creation, provider normalization, error mapping for both providers, and the existing warehouse source.

## Scope

This PR establishes the integration and common code-host boundary. The following pieces remain separate work:

- Generalize `products/tasks/backend` and sandbox cloning, which still use GitHub-specific models, token minting, repository shapes, and webhook flows.
- Add Azure DevOps webhook triggers if loops should react to Azure repository events.
- Add an OAuth app flow if PAT setup should be replaced later.
- Port the transport onto the shared `EgressClient` framework and add a DB-backed repository cache mirroring GitHub's (both raised in review; tracked as open review threads).

- Focused Python tests passed across the Azure transport (incl. retry cap and 3xx handling), warehouse source (41), model, and API (incl. GitHub-failure 400 mapping).
- Frontend lint and format passed on the changed files.
- `makemigrations --check --dry-run` reported no changes; migration 1263 is a choices-only `AlterField` (database no-op), renumbered after master took 1262.
- OpenAPI frontend and MCP clients were regenerated with no drift.

## Automatic notifications

- [ ] Publish to changelog?

## Agent context

The implementation was reviewed against PostHog's GitHub integration, integration registry, shared egress, generated API, MCP, migration, and testing patterns. It keeps the task-runner wiring outside this PR because that path needs a wider model and sandbox migration rather than a provider method bolted onto the existing GitHub fields.

A multi-perspective QA Swarm review (qa-team, paul-reviewer, xp-reviewer) ran on this PR; the actionable findings were applied in follow-up commits, and the remaining architectural questions (EgressClient conformance, protocol surface, repository caching, API-key test coverage) are open review threads.

---
Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C07TQR0V16U/p1784729707784689)


posthog/egress/azure_devops/transport.py
```diff
@@ -0,0 +1,150 @@
...
+
+        method = method.upper()
+
+        def request() -> requests.Response:
+            return self._request_once(method, path, endpoint=endpoint, params=params, json=json)
+
+        if method not in {"GET", "HEAD", "OPTIONS"}:
+            return request()
+
+        retryer = Retrying(
```
willwearing, line 103: > [!NOTE]

**[convergent: qa-team/reliability + paul (7 qa-team reviewers)]** 🟠 HIGH

The GET retry policy (5 attempts, `wait_exponential_jitter(initial=2, max=120)`, sleeps ~2+4+8+16s plus up to 5×10s timeouts) now runs inside synchronous Django request handlers: `code_host_repositories`, `AzureDevOpsIntegration.create_integration` (which executes inside `transaction.atomic()` in `IntegrationSerializer.create`, holding a Postgres connection), and warehouse `validate_credentials`. During an Azure 5xx/429 incident one request can pin a gunicorn worker for ~80s, and retrying 429s ignores Retry-After and burns more of Azure's 200-TSTU/5-min per-PAT budget exactly when it's exhausted. This policy was designed for the background Temporal sync; the GitHub reference transport deliberately does no transport-level retries.

**Fix:** make attempts/waits constructor parameters — 1 (or 2 short) attempts for `source="integration"` interactive callers, keep 5 for `source="warehouse_source"`.

posthog/egress/azure_devops/transport.py
```diff
@@ -0,0 +1,150 @@
...
+
+def normalize_azure_devops_organization(value: str) -> str:
+    organization = value.strip().removeprefix("https://").removeprefix("http://")
+    organization = organization.removeprefix("dev.azure.com/").split("/")[0]
+    if not organization or not _ORGANIZATION_RE.fullmatch(organization) or organization in {".", ".."}:
+        raise ValueError(f"Invalid Azure DevOps organization: {value!r}")
+    return organization
+
+
+class AzureDevOpsClient:
```
willwearing, line 63: > [!NOTE]

**[qa-team/compatibility (7 reviewers converged)]** 🟠 HIGH

`AzureDevOpsClient` is hand-rolled around `requests` with bespoke Prometheus counters, bypassing the egress framework this package mandates: `posthog/egress/README.md` says a new outbound domain is an `EgressClient` incarnation with `limiter.py` + `observability.py` adapters, "never hand-rolled around `requests`". All three existing domains (`github/`, `logodev/`, `vapi/`) subclass `EgressClient`, and GitHub has a semgrep rule pinning callers to its transport. As shipped there is no shared budget or priority lanes even though the integration client and the warehouse source now share one PAT identity against Azure's tight 200-TSTU/5-min budget, and no guard stops future callers bypassing the counters.

**Fix:** implement as an `EgressClient` subclass with an observability adapter (identity-blind PAT gating can be deferred per the README's GitHub-PAT precedent) and add a semgrep rule for `dev.azure.com`; or document an explicit maintainer exception.

posthog/api/integration.py
```diff
@@ -1476,6 +1549,30 @@ def github_repos(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+        except UnsupportedCodeHostIntegrationError as error:
+            raise ValidationError(str(error)) from error
+
+        try:
+            page = code_host.list_repositories(
+                search=query_serializer.validated_data["search"],
+                limit=query_serializer.validated_data["limit"],
+                offset=query_serializer.validated_data["offset"],
+            )
+        except AzureDevOpsIntegrationError as error:
```
willwearing: > [!NOTE]

**[convergent: qa-team/reliability + paul + xp (all three reviewers)]** 🟡 MEDIUM

The provider-neutral `code_host_repositories` endpoint catches only `AzureDevOpsIntegrationError`, so GitHub-side failures escape as 500s: for a GitHub integration with no cached snapshot, `list_cached_repositories` re-raises refresh failures (e.g. `GitHubIntegrationError` on a revoked installation). The same failure returns a clean 400 on the Azure branch (the PR's test covers only that path), and the Azure-specific `except` leaks provider knowledge the `code_host_integration_for` abstraction was built to hide — every future provider needs its own clause here.

**Fix:** define a shared `CodeHostIntegrationError` in `code_host_integration.py`, have each adapter translate its provider errors into it, and catch only that here.

frontend/src/scenes/integrations/azure-devops/azureDevOpsSetupModalLogic.ts
```diff
@@ -0,0 +1,176 @@
...
+                project: project.trim() ? undefined : 'Project is required',
+                personalAccessToken: personalAccessToken.trim() ? undefined : 'Personal access token is required',
+            }),
+            submit: async () => {
+                if (!values.currentTeam?.id) {
+                    throw new Error('Select a project before connecting Azure DevOps')
+                }
+
+                try {
+                    const integration = await integrationsCreate(String(values.currentTeam.id), {
```
willwearing: > [!NOTE]

**[convergent: qa-team/frontend + paul]** 🟡 MEDIUM

The submit handler calls the orval-generated `integrationsCreate(String(values.currentTeam.id), ...)`, which targets `/api/projects/{projectId}/integrations/` with an environment (team) id. In a multi-environment project a secondary environment's `currentTeam.id` is not a valid project id, so the request 404s or attaches the integration to the wrong environment. The sibling GitLab modal (`gitlabSetupModalLogic.ts`) uses `api.integrations.create`, which posts to `/api/environments/{team_id}/` for the current environment automatically.

**Fix:** use `api.integrations.create({ kind: 'azure-devops', config: {...} })` like GitLab — this also removes the `teamLogic` connect and the `currentTeam` guard (which today sits outside the try/catch, so kea-forms swallows its `throw` and the user gets zero feedback).

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+        return [
+            {
+                "id": repo["id"],
+                "name": repo["name"],
+                "default_branch": self._strip_ref(repo.get("defaultBranch")),
+            }
+            for repo in response.json().get("value", [])
+        ]
+
+    def _resolve_repository_id(self, repository: str) -> str:
```
willwearing: > [!NOTE]

**[convergent: qa-team/database + paul + xp (all three reviewers)]** 🟡 MEDIUM

HTTP N+1: `_resolve_repository_id` and `get_default_branch` are near-duplicate scan loops that each independently call `list_repositories()` (a full-list fetch), so `create_branch` makes up to 4 sequential HTTP calls and the create_branch→create_pull_request flow up to 7, of which ~5 are redundant full-list fetches — each also subject to the 5-attempt retry policy. The two loops also disagree on failure (one raises, one silently returns `None`), and two listings within one operation can disagree if a repo changes between calls.

**Fix:** a single `_find_repository(repository) -> dict | None` helper that lists once and yields both the id and `default_branch` (Azure's `GET .../git/repositories/{repository}` also accepts name-or-id and returns `defaultBranch` in one call).

posthog/models/code_host_integration.py
```diff
@@ -0,0 +1,161 @@
...
+    has_more: bool
+
+    def as_dict(self) -> dict[str, Any]:
+        return {
+            "repositories": [repository.as_dict() for repository in self.repositories],
+            "has_more": self.has_more,
+        }
+
+
+class CodeHostIntegration(Protocol):
```
willwearing, line 65: > [!NOTE]

**[convergent: xp + paul]** 🟡 MEDIUM

YAGNI: of the five methods on the `CodeHostIntegration` protocol, only `list_repositories` has a caller in this PR — `get_default_branch`, `create_branch`, `create_pull_request`, and `clone_url` (plus both adapters' implementations) are dead until the self-driving consumer lands. `clone_url` is the worst offender: nothing calls it anywhere, and the two implementations already disagree (GitHub hardcodes `https://github.com` + `.git`, Azure builds a `dev.azure.com` URL without `.git`) — nobody can catch that until a consumer exists.

**Suggestion:** trim the protocol to `list_repositories` now and grow it in the PR that actually calls the rest; the consumer will tell you the right signatures. If the follow-up PR is imminent, at least delete `clone_url`.

posthog/models/code_host_integration.py
```diff
@@ -0,0 +1,161 @@
...
+    def _repository_name(self, repository: str) -> str:
+        prefix = f"{self.github.organization()}/"
+        return repository.removeprefix(prefix)
+
+
+class AzureDevOpsCodeHostIntegration:
+    def __init__(self, integration: Integration) -> None:
+        self.azure_devops = AzureDevOpsIntegration(integration)
+
+    def list_repositories(self, *, search: str, limit: int, offset: int) -> CodeHostRepositoryPage:
```
willwearing, line 141: > [!NOTE]

**[qa-team/performance (5 reviewers converged)]** 🟡 MEDIUM

`AzureDevOpsCodeHostIntegration.list_repositories` fetches the full live repository list from Azure on every call, then filters and paginates in memory — unlike GitHub, which serves this same endpoint from the DB-backed `repository_cache` with a 30s refresh cooldown (columns already exist on the Integration model). A repo picker searching per keystroke, or the newly enabled MCP tool paginating, issues one full-list outbound call per request (`search`/`limit`/`offset` reduce nothing), burning the 200-TSTU/5-min PAT budget, and offset paging over a live list can skip/duplicate rows between pages.

**Fix:** populate and serve from `repository_cache` mirroring `GitHubIntegration.list_cached_repositories`, or at minimum a short-TTL cache keyed on integration id.

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+                {
+                    "name": f"refs/heads/{branch_name}",
+                    "oldObjectId": "0000000000000000000000000000000000000000",
+                    "newObjectId": base_sha,
+                }
+            ],
+        )
+
+        if response.status_code in (200, 201):
+            result = response.json().get("value", [{}])[0]
```
willwearing: > [!NOTE]

**[qa-team/data-integrity (3 reviewers converged)]** 🟡 MEDIUM

Two failure-shape bugs in `create_branch`'s response handling: (1) `response.json().get("value", [{}])[0]` raises an unhandled IndexError when Azure returns 200 with `{"value": []}` (the default only applies when the key is absent), escaping the method's `{"success": False, ...}` contract; (2) `result.get("success", True)` defaults to success, so a 200 body without a per-ref `success` flag reports `{"success": True}` for a branch that was never created — the self-driving flow then opens a PR from a nonexistent branch.

**Fix:** `values = response.json().get("value") or []; if not values: return {"success": False, ...}` and require `result.get("success") is True`.

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+        path: str,
+        *,
+        endpoint: str,
+        params: dict[str, Any] | None = None,
+        json: dict[str, Any] | list[dict[str, Any]] | None = None,
+    ) -> requests.Response:
+        try:
+            return self._client.request(method, path, endpoint=endpoint, params=params, json=json)
+        except (AzureDevOpsAuthenticationError, AzureDevOpsRetryableError, requests.RequestException) as error:
+            raise AzureDevOpsIntegrationError(
```
willwearing, line 3547: > [!NOTE]

**[qa-team/data-integrity (+copy)]** 🟡 MEDIUM

`_request` collapses `AzureDevOpsAuthenticationError` — whose transport message is accurate and actionable ("the personal access token is invalid or expired") — into the generic "Azure DevOps request failed. Check the connection and try again.", and never sets `integration.errors`. Azure PATs have mandatory expiry (max ~1 year), so this WILL happen routinely: users are told to check their connection instead of rotating the token, and monitoring/UI never sees the broken integration (contrast GitHub's `_on_token_refresh_failed` setting `ERROR_TOKEN_REFRESH_FAILED`).

**Fix:** catch `AzureDevOpsAuthenticationError` separately, surface token-specific guidance, and persist `integration.errors` on auth failure (clearing on success) per the established pattern.

posthog/models/code_host_integration.py
```diff
@@ -0,0 +1,161 @@
...
+        repositories = [
+            CodeHostRepository(
+                id=str(repository["id"]),
+                name=repository["name"],
+                full_name=(f"{self.azure_devops.organization}/{self.azure_devops.project}/{repository['name']}"),
+                provider=Integration.IntegrationKind.AZURE_DEVOPS.value,
+                default_branch=repository.get("default_branch"),
+            )
+            for repository in self.azure_devops.list_repositories()
+            if not search or search.lower() in repository["name"].lower()
```
willwearing: > [!NOTE]

**[convergent: qa-team/data-integrity + paul]** 🟡 MEDIUM

Search semantics differ per provider on the same "provider-neutral" endpoint: the Azure adapter filters on the short `name` only, while GitHub filters on `full_name` (casefold). A client searching with the `full_name` this very endpoint returned ("my-org/my-project/alpha") gets matches on GitHub but zero rows on Azure for a repository that exists — MCP agents will conclude the repo is inaccessible.

**Fix:** filter Azure repositories on the constructed `full_name` (casefold), mirroring the GitHub behavior — or if name-only is intentional, document it so nobody "fixes" it in the wrong direction.

willwearing (commented):
> [!NOTE]

QA Swarm review complete — see inline comments. Summary comment to follow.

frontend/src/scenes/integrations/azure-devops/azureDevOpsSetupModalLogic.ts
```diff
@@ -0,0 +1,176 @@
...
+                            personal_access_token: values.azureDevOpsIntegration.personalAccessToken,
+                        },
+                    })
+                    actions.loadIntegrations()
+                    lemonToast.success('Azure DevOps connected.')
+                    props.onComplete(integration.id)
+                } catch (error: unknown) {
+                    const detail =
+                        typeof error === 'object' && error !== null && 'detail' in error
+                            ? String(error.detail)
```
willwearing: > [!NOTE]

**[qa-team/frontend (+copy, generalist-a)]** 🟡 MEDIUM

Two issues around submit handling: (1) the error toast can literally display "null" — `ApiError` always defines a `detail` own-property (initialized to `null` when the response has no JSON detail: 500s, gateway errors, the 404 from the wrong-route bug), so `'detail' in error` is true and `String(error.detail)` renders "null" instead of the fallback. The sibling GitLab logic uses a truthiness check. (2) The form is never reset after a successful submit, so the plaintext PAT persists in kea/redux state (and the password input) for the rest of the session — recoverable via devtools or by reopening the modal, defeating the server's deliberate write-only handling.

**Fix:** `lemonToast.error(error.detail || 'Could not connect Azure DevOps')`, and call `actions.resetAzureDevOpsIntegration()` after `props.onComplete(integration.id)` (and ideally on modal close).

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+            integration_id=f"{organization}/{project}",
+            defaults={
+                "config": {"organization": organization, "project": project},
+                "sensitive_config": {"access_token": personal_access_token},
+                "created_by": user,
+            },
+        )
+        return integration
+
+    def __init__(self, integration: Integration) -> None:
```
willwearing, line 3497: > [!NOTE]

**[qa-team/generalist-b (5 reviewers converged)]** 🟢 LOW

`AzureDevOpsIntegration.__init__` builds the client from stored config without converting validation failures: `AzureDevOpsClient.__init__` re-runs `normalize_azure_devops_organization` and raises bare `ValueError` for a malformed stored org, and `_access_token` (annotated `-> str`) returns `None` when `sensitive_config` lacks the key, producing `auth=("", None)` and an opaque TypeError deep in requests. Either way a malformed row turns `code_host_repositories` (including the MCP tool) into a 500 instead of the 400 that `AzureDevOpsIntegrationError` paths get.

**Fix:** in `__init__`, wrap normalization `ValueError` into `AzureDevOpsIntegrationError` and raise it when the token is missing, matching `create_integration`'s conversions.

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+                return repo["id"]
+        raise AzureDevOpsIntegrationError(f"Repository '{repository}' not found in {self.organization}/{self.project}")
+
+    def get_default_branch(self, repository: str) -> str | None:
+        for repo in self.list_repositories():
+            if repository in (repo["id"], repo["name"]):
+                return repo["default_branch"]
+        return None
+
+    def _get_branch_sha(self, repository_id: str, branch: str) -> str | None:
```
willwearing, line 3590: > [!NOTE]

**[qa-team/reliability]** 🟢 LOW

`_get_branch_sha` returns `None` for any non-200 response, so `create_branch` reports "Base branch 'X' not found" for what is actually a request/permission failure (e.g. a PAT with a project-scoped restriction getting 401/403 on the refs call) — the user hunts a nonexistent-branch problem instead of fixing token scope.

**Fix:** raise `AzureDevOpsIntegrationError` with the status for non-200 and reserve `None` for a genuinely absent ref (200 with no match).

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
```diff
@@ -104,29 +95,19 @@ def get_rows(
...
-                f"Azure DevOps API error (retryable): status={response.status_code}, url={url}"
-            )
 
-        # An invalid/expired PAT yields a 203 with an HTML sign-in page.
-        if response.status_code == 203:
-            raise AzureDevOpsAuthError(
-                "Azure DevOps returned a sign-in page (203) — the personal access token is invalid or expired."
-            )
+    def fetch(path: str, params: dict[str, Any]) -> requests.Response:
+        response = client.request("GET", path, endpoint=config.path, params=params)
```
willwearing: > [!NOTE]

**[qa-team/data-integrity (3 reviewers converged)]** 🟢 LOW

`fetch` hardcodes `endpoint=config.path`, so project-enumeration calls to `/_apis/projects` made while syncing builds/PRs/repos are recorded in the new `azure_devops_api_requests` metric under the data endpoint's label (e.g. `/{project}/_apis/build/builds`) — per-endpoint volume and error attribution are wrong from day one, and a projects-API failure looks like a builds-API failure.

**Fix:** thread an `endpoint` argument through `fetch` (use `/_apis/projects` for enumeration, `config.path` for data calls).

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
```diff
@@ -104,29 +95,19 @@ def get_rows(
     session = _get_session(personal_access_token)
     org = _validate_organization(organization)
 
-    @retry(
-        retry=retry_if_exception_type((AzureDevOpsRetryableError, requests.ReadTimeout, requests.ConnectionError)),
-        stop=stop_after_attempt(MAX_RETRY_ATTEMPTS),
-        wait=wait_exponential_jitter(initial=2, max=120),
-        reraise=True,
+    client = AzureDevOpsClient(
```
willwearing, line 93: > [!NOTE]

**[qa-team/compatibility]** 🟢 LOW

Migrating the existing warehouse source onto `AzureDevOpsClient` silently changes HTTP semantics: the shared client pins `allow_redirects=False` while the old `session.get(...)` followed redirects. A 30x from dev.azure.com (e.g. after an org rename) now has `response.ok == True` (status < 400), passes the not-ok check, and `.json()` runs on a bodyless redirect — crashing or yielding zero rows mid-sync. All tests mock the session, so this behavior change of a shipping pipeline is untested.

**Fix:** treat 3xx as an error in `_request_once` so the failure is loud, or make `allow_redirects` a per-caller client parameter.

posthog/api/integration.py
```diff
@@ -1476,6 +1549,30 @@ def github_repos(self, request: Request, *args: Any, **kwargs: Any) -> Response:
 
         return Response({"repositories": repositories, "has_more": has_more})
 
+    @extend_schema(
+        parameters=[CodeHostRepositoriesQuerySerializer],
+        responses={200: CodeHostRepositoriesResponseSerializer},
+    )
+    @action(methods=["GET"], detail=True, url_path="code_host_repositories")
```
willwearing, line 1560: > [!NOTE]

**[paul + xp]** 🟢 LOW

Two related asks on this newly exposed endpoint: (1) the threat ratchet goes one higher — it ships enabled as an MCP tool with `integration:read` scope, reachable with personal API keys from day one. Team scoping via `get_object()` should be fine, but the two boring tests are worth writing down: a personal API key with `integration:read` can hit it, and team A's key gets a 404 for team B's integration id (existing tests all use `force_login`). (2) For GitHub integrations this endpoint now serves exactly what `github_repos` serves with the same query params — two endpoints and two MCP tools for one behavior. Fine as a migration step, but consider making `github_repos` delegate to the code-host path or marking it deprecated so the two can't drift.

posthog/models/integration.py
```diff
@@ -3393,6 +3404,261 @@ def create_issue(self, config: dict[str, str]):
...
+                f"Could not access Azure DevOps project '{organization}/{project}'. "
+                "Check the organization, project, and token, then try again."
+            ) from error
+        if response.status_code != 200:
+            raise AzureDevOpsIntegrationError(
+                f"Could not access Azure DevOps project '{organization}/{project}' "
+                f"(status {response.status_code}). Check the organization, project, and token."
+            )
+
+        integration, _ = Integration.objects.update_or_create(
```
willwearing, line 3485: > [!NOTE]

**[qa-team/data-integrity]** ⚪ NIT

`update_or_create` keys on the case-preserved `integration_id=f"{organization}/{project}"`, but Azure DevOps org/project names are case-insensitive — "My-Org/My-Project" and "my-org/my-project" create two rows for the same Azure project, duplicating repos in pickers and leaving one row with a stale PAT.

**Fix:** normalize casing before building `integration_id` (e.g. to the casing Azure returns from the validation call's response, already in hand).

posthog/api/integration.py
```diff
@@ -240,6 +260,43 @@ class GitHubReposRefreshResponseSerializer(serializers.Serializer):
...
+        help_text="Repository default branch, when the provider reports one.",
+    )
+    can_push = serializers.BooleanField(
+        required=False,
+        allow_null=True,
+        help_text="Whether the integration can push to the repository, when the provider reports it.",
+    )
+
+
+class CodeHostRepositoriesResponseSerializer(serializers.Serializer):
```
willwearing, line 292: > [!NOTE]

**[qa-team/compatibility]** ⚪ NIT

`CodeHostRepositoriesResponseSerializer` exists only for schema generation — the view returns `Response(page.as_dict())` without running it, so nothing enforces the documented contract, and `dataclasses.asdict` always emits `default_branch`/`can_push` (as null) though the schema marks them `required=False`. Shapes are compatible today and this mirrors the pre-existing `github_repos` shortcut, but future drift ships unvalidated.

**Fix:** serialize through the response serializer or align its `required` flags with the always-present keys.

products/integrations/mcp/tools.yaml
```diff
@@ -82,6 +82,20 @@ tools:
     integrations-clickup-workspaces-retrieve:
         operation: integrations_clickup_workspaces_retrieve
         enabled: false
+    integrations-code-host-repositories-retrieve:
```
willwearing, line 85: > [!NOTE]

**[qa-team/compatibility]** ⚪ NIT

`integrations-code-host-repositories-retrieve` is enabled in the same PR that introduces its backing Django endpoint, but MCP and the API deploy independently — if MCP rolls out first in a region the advertised tool 404s. Failure is graceful, so just confirm deploy ordering (API before MCP) or consciously accept the transient window.

frontend/src/scenes/integrations/definitions/azureDevOps.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const AzureDevOps = defineIntegration(
+    {
+        slug: 'azure-devops',
+        kind: 'azure-devops',
+        name: 'Azure DevOps',
+        logo: ICONS['azure-devops'],
+        subtitle: 'Connect your Azure DevOps repositories to PostHog',
+        description:
+            'Connect an Azure DevOps project so PostHog can discover repositories and open code changes through a shared code-host interface.',
```
willwearing: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

"through a shared code-host interface" is internal architecture jargon leaked into the integrations gallery; sibling descriptions (GitHub, GitLab) are written as user outcomes. Suggest: "Connect an Azure DevOps project so PostHog can browse your repositories, create branches, and open pull requests for you."

frontend/src/scenes/integrations/azure-devops/AzureDevOpsSetupModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+                </div>
+            }
+            onClose={props.onComplete}
+        >
+            <Form logic={azureDevOpsSetupModalLogic} props={props} formKey="azureDevOpsIntegration">
+                <div className="gap-4 flex flex-col">
+                    <LemonField
+                        name="organization"
+                        label="Organization"
+                        help="Use the first path segment from dev.azure.com/your-organization."
```
willwearing: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

Help text "Use the first path segment from dev.azure.com/your-organization." is URL-parsing jargon and undersells the field — the backend accepts a full dev.azure.com URL (`normalize_azure_devops_organization` strips scheme/host, and the API help text says so). Suggest: "Your organization name, as it appears in dev.azure.com/your-organization. Pasting the full URL also works."

frontend/src/scenes/integrations/azure-devops/AzureDevOpsSetupModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+                                    target="_blank"
+                                    to="https://learn.microsoft.com/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate"
+                                >
+                                    Azure DevOps
+                                </Link>
+                                .
+                            </>
+                        }
+                    >
+                        <LemonInput type="password" autoComplete="new-password" />
```
willwearing: > [!NOTE]

**[qa-team/frontend]** ⚪ NIT

The PAT input has no placeholder, unlike the GitLab modal's token field which hints the expected token shape — a small consistency gap between sibling modals. Add a representative masked placeholder.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
```diff
@@ -16,21 +20,9 @@
...
-MAX_RETRY_ATTEMPTS = 5
-
-
-class AzureDevOpsRetryableError(Exception):
-    pass
-
-
-class AzureDevOpsAuthError(Exception):
-    pass
+AzureDevOpsAuthError = AzureDevOpsAuthenticationError
```
willwearing: > [!NOTE]

**[xp]** ⚪ NIT

`AzureDevOpsAuthError = AzureDevOpsAuthenticationError` keeps the old name alive alongside the new one. The only usages are one raise-site check and the tests in this module — rename them and drop the alias so the concept has one name.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
```diff
@@ -82,10 +70,13 @@ def validate_credentials(organization: str, personal_access_token: str) -> bool:
     than a 401, so only an exact 200 counts."""
     try:
         org = _validate_organization(organization)
-        response = _get_session(personal_access_token).get(
-            f"{AZURE_DEVOPS_BASE_URL}/{quote(org)}/_apis/projects?{urlencode({'$top': 1, 'api-version': API_VERSION})}",
+        response = AzureDevOpsClient(
```
willwearing, line 67: > [!NOTE]

**[xp]** ⚪ NIT

The PAT is now supplied twice on this path: `_get_session` sets `session.auth = ("", pat)` and `AzureDevOpsClient` passes `auth=("", pat)` on every request (the latter wins). Harmless since they're identical, but pick one owner — either stop setting auth on the session or let the client take a pre-authed session without a separate token.

posthog/models/code_host_integration.py
```diff
@@ -0,0 +1,161 @@
+from dataclasses import asdict, dataclass
```
willwearing, line 10: > [!NOTE]

**[xp]** ⚪ NIT

I'm not sure about this, but `posthog/models/code_host_integration.py` defines no Django model — it's an adapter layer over `Integration`. If there's a better-fitting home (or just a module docstring saying why it lives here), the next reader won't go hunting for a model class.

willwearing:
> [!NOTE]
>
> Multi-perspective review: qa-team (10 specialists + generalists), paul-reviewer, xp-reviewer

Two HIGH-severity findings (retry policy blocking synchronous web workers; egress-framework bypass) plus a cluster of MEDIUMs where the new provider-neutral seam deviates from the established GitHub/GitLab patterns. Nothing is a credential-exposure or data-loss blocker — the security fundamentals are solid.

### Key findings

**🟠 HIGH**
- `posthog/egress/azure_devops/transport.py:97` — 5-attempt exponential-backoff retry (designed for Temporal) now runs inside synchronous request handlers, including inside `transaction.atomic()` on integration create; one request can pin a worker ~80s during an Azure incident. Make attempts a constructor param per caller.
- `posthog/egress/azure_devops/transport.py:59` — hand-rolled `requests` client bypasses the egress framework that `posthog/egress/README.md` mandates (`EgressClient` + limiter/observability adapters, semgrep guard); all three existing egress domains conform.

**🟡 MEDIUM**
- `posthog/api/integration.py:1572` — provider-neutral endpoint catches only `AzureDevOpsIntegrationError`; GitHub failures 500. Add a shared `CodeHostIntegrationError`.
- `azureDevOpsSetupModalLogic.ts:154` — orval `integrationsCreate` called with an environment id against the `/api/projects/` route; breaks multi-environment projects. Use `api.integrations.create` like the GitLab modal.
- `posthog/models/integration.py:3552` — HTTP N+1: duplicate full-list loops make branch/PR creation cost up to 7 sequential calls.
- `posthog/models/code_host_integration.py:38` — 4 of 5 protocol methods have no callers (YAGNI); `clone_url` implementations already disagree.
- `posthog/models/code_host_integration.py:107` — Azure list is a live full fetch per call (no `repository_cache` like GitHub) against a tight per-PAT rate budget; `:117` — search filters `name` on Azure but `full_name` on GitHub.
- `posthog/models/integration.py:3607` — `create_branch` can IndexError on empty `value` and defaults missing `success` to `True`; `:3528` — auth errors collapsed to a generic message, `integration.errors` never set (PATs expire routinely).
- `azureDevOpsSetupModalLogic.ts:168` — error toast can render literal "null"; form (and PAT) never reset after submit.

Plus 4 LOW and 10 NIT inline comments (metric label attribution, redirect-semantics change on the warehouse path, API-key/team-isolation test coverage, copy tweaks).

### Convergence

Highest-confidence findings, flagged independently by 2+ reviewers:
- Retry-in-web-path (7 qa-team reviewers + paul)
- Azure-only error handling in the neutral endpoint (all three reviewers)
- HTTP N+1 in branch/PR flows (all three reviewers)
- Wrong frontend API client vs GitLab sibling (qa-team frontend + compatibility + paul)
- Speculative `CodeHostIntegration` protocol (paul + xp)
- Search-semantics mismatch (qa-team + paul)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | MEDIUM risk; PAT/security handling solid, but the retry policy and egress-framework bypass should be fixed pre-merge. |
| 👤 paul | Approve with caveats — great transport extraction and observability; the provider-neutral seam leaks and the protocol is mostly speculative. |
| 📐 xp | Well-built and mostly earns its abstractions; trim the uncalled protocol methods and extract `_find_repository` to kill duplication + a redundant fetch. |

posthog/egress/azure_devops/transport.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — `AzureDevOpsClient` now takes `max_attempts` (default 5 for the Temporal warehouse sync); `AzureDevOpsIntegration` and warehouse `validate_credentials` construct with `max_attempts=2` so synchronous web paths fail fast. Covered by a transport test asserting the cap.

posthog/api/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — added `CodeHostIntegrationError` in `code_host_integration.py`; both adapters translate provider errors via an `_as_code_host_error` contextmanager, and the endpoint catches only the shared error. Added a test that a failing GitHub adapter returns 400.

frontend/src/scenes/integrations/azure-devops/azureDevOpsSetupModalLogic.ts
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — the modal logic now uses `api.integrations.create` like the GitLab sibling; the `teamLogic` connect, `currentTeam` guard, and generated-client import are gone.

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — `_resolve_repository_id`/`get_default_branch` collapsed into a single `_find_repository` that lists once and yields both id and default branch; `create_branch` and `create_pull_request` each make one listing call now, and not-found consistently raises `AzureDevOpsIntegrationError`.

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — empty `value` no longer IndexErrors (returns the `success: False` contract) and success requires `result.get("success") is True`. Added a model test for the empty-result case.

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — `AzureDevOpsAuthenticationError` is caught separately with a token-specific message, and `integration.errors` is set to `ERROR_TOKEN_REFRESH_FAILED` on auth failure / cleared on success, mirroring the GitHub pattern.

posthog/models/code_host_integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — the Azure adapter now filters on the constructed `full_name` (casefold), matching GitHub's search semantics.

frontend/src/scenes/integrations/azure-devops/azureDevOpsSetupModalLogic.ts
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — toast uses `error.detail || 'Could not connect Azure DevOps'` (no more literal "null") and the form resets after a successful submit so the PAT doesn't linger in kea state.

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — `__init__` raises `AzureDevOpsIntegrationError` for a missing token and wraps client-construction `ValueError`; `_access_token` is annotated `str | None`.

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — non-200 on the refs lookup now raises `AzureDevOpsIntegrationError` with the status; `None` is reserved for a 200 with no matching ref.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — `fetch`/`iterate_header_token` take an `endpoint` argument; project enumeration records `/_apis/projects` and data calls record `config.path`.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — the transport now raises `AzureDevOpsUnexpectedRedirectError` for any 3xx (redirects are never followed), so a redirect fails loudly instead of passing `response.ok`. Covered by a transport test.

frontend/src/scenes/integrations/definitions/azureDevOps.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — gallery description rewritten as a user outcome.

frontend/src/scenes/integrations/azure-devops/AzureDevOpsSetupModal.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — help text now says the org name as it appears in the URL and that pasting the full URL works (verified `normalize_azure_devops_organization` accepts it).

frontend/src/scenes/integrations/azure-devops/AzureDevOpsSetupModal.tsx
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — added a masked placeholder matching the GitLab token field style.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — alias deleted; the test file imports `AzureDevOpsAuthenticationError` directly.

products/warehouse_sources/backend/temporal/data_imports/sources/azure_devops/azure_devops.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — session-level auth removed; `AzureDevOpsClient` is the single owner of the PAT on this path.

posthog/models/code_host_integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Fixed in `646deb95bcb` — added a module docstring stating this is the provider-neutral adapter layer over `Integration` (no Django model).

posthog/models/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Out of scope for this pass — normalizing `integration_id` casing needs the casing Azure returns from the validation response plumbed through `create_integration`; worth a small follow-up.

posthog/api/integration.py
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Intentional for now — mirrors the pre-existing `github_repos` shortcut; aligning the response serializer is a follow-up.

products/integrations/mcp/tools.yaml
willwearing: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — not written by a human

Accepted — the failure mode is a graceful 404 during the transient window; standard deploy ordering (API before MCP) covers this.

PR #72919, opened 2026-07-22
perf(messaging): shrink reconcile overrides lookback to 6h

gustavohstrassburger:
## Problem

The `reconcile-precalculated-data` Temporal workflow runs every 15 minutes, but it isn't incremental to recent merges. On each run it re-diffs **every** override inside `RECONCILE_PRECALCULATED_DATA_OVERRIDES_LOOKBACK_HOURS`, and there is no watermark between runs. At the old 48h default that means every run reprocesses ~48 hours of already-handled merges — roughly 192× the window the 15-minute cadence implies.

The reprocessing is idempotent (rows already correct are skipped, ReplacingMergeTree dedups the rest), so it emits almost nothing on repeat runs — but it still pays the full **read** cost every time, and each 1000-distinct-id batch is a full team-partition scan because `distinct_id` isn't a sort-key prefix. That combination is a big part of why a run that should take seconds takes hours.

## Changes

Drop the default lookback from 48h to 6h — an 8× cut in the reprocessed window (and therefore in per-run batches/scans) — via a single `DEFAULT_OVERRIDES_LOOKBACK_HOURS` constant replacing three inline `48` literals. Still env-overridable through `RECONCILE_PRECALCULATED_DATA_OVERRIDES_LOOKBACK_HOURS`.

> [!NOTE]
> Kept at 6h rather than ~1h on purpose. The schedule is `overlap=SKIP`, so while a run is in flight later triggers are skipped; the window must stay above the worst-case run duration or a merge landing during a long run would never be picked up. The window also must stay well under the weekly person-overrides squash cadence (unchanged here). Once per-run duration is reliably small, the env var can be lowered further.

**Operational follow-ups (not in this PR):**
- Trigger one `full_scan=true` run after deploy to clear any merges stranded in the old 6h–48h tail that earlier failing/slow runs never reconciled.
- Measure per-run duration afterward. A per-batch data-skipping index on `distinct_id` (separate migration-only PR) is the next lever if a single team's partition scan still dominates.

I (Claude) made the change; @gustavohstrassburger directed it while debugging why the workflow takes hours. Automated checks run locally: `ruff check` + `ruff format` clean, `mypy --cache-fine-grained` on both files ("no issues found in 2 source files"), `py_compile` clean. This is a default-value + comment change with no behavioral branching added, so no new tests; the existing reconcile suite still covers the activities. I could not run the workflow against production data from this environment.

## 🤖 Agent context

Authored by Claude via PostHog Code, directed by @gustavohstrassburger. Root-caused from the workflow code: the schedule passes `ReconcilePrecalculatedDataWorkflowInputs()` (no `full_scan`, all teams), the activities compute `since = now − lookback` fresh every run with no persisted cursor, and the overrides window was 48h. Considered a proper per-run watermark (deferred — bigger change, wanted the immediate low-risk win first) and lowering all the way to ~1h (rejected for now due to the `overlap=SKIP` / run-duration coupling described above).

---


gustavohstrassburger:
@greptile review

gustavohstrassburger:
Thanks — addressed the two actionable findings in aaefd62:

- **Default test still expects 48h** (P1): fixed. `test_defaults_when_env_unset` now asserts `since` against the `DEFAULT_OVERRIDES_LOOKBACK_HOURS` constant instead of a hardcoded 48h, so it tracks the default and can't drift again. Verified passing locally.
- **"incremental" wording** (Copilot): reworded the `schedules.py` comment to say each run re-diffs the whole lookback window (no watermark), rather than implying cursored/incremental processing.

On **"Six-Hour Window Strands Overrides"** (P1): this is a real property, but it's the intended tradeoff of this PR rather than a regression in the diff — and it is strictly better than the prior 48h behavior, not worse (48h had the same no-watermark design, just a wider window). It's called out in the `DEFAULT_OVERRIDES_LOOKBACK_HOURS` docstring: the window must stay above worst-case run duration precisely because the schedule is `overlap=SKIP` and there is no watermark. Mitigations:

- 6h is a deliberately conservative interim value; a one-time `full_scan=true` run after deploy clears anything stranded in the old 6h–48h tail.
- The real fix that removes the stranding class entirely is a persisted per-run **watermark** (query `_timestamp > watermark`), which makes coverage gap-free regardless of run duration or SKIP. That's the planned follow-up; this PR is the immediate, low-risk runtime cut before building the cursor.

Leaving the thread open for a human to close.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -55,6 +55,15 @@ def _positive_int_env(name: str, default: int, logger: structlog.BoundLogger) ->
     return value
 
 
+# Default overrides lookback. Every scheduled run re-diffs *all* overrides inside this window (there
+# is no watermark between runs), so a wide window means each run reprocesses hours of already-handled
+# merges — the reprocess is idempotent but still pays the full read cost. Keep it comfortably above
+# the schedule interval AND above the worst-case run duration: the schedule is overlap=SKIP, so a
+# merge landing while a long run is in flight is only picked up by a later run whose window still
+# reaches back to it. Env-overridable; lower it further once per-run duration is reliably small.
+DEFAULT_OVERRIDES_LOOKBACK_HOURS = 6
```
haacked, line 64: Worth sharpening this with the actual timeout numbers: each per-team activity has `start_to_close_timeout=2h` with `maximum_attempts=3`, so one stuck team can burn up to ~12h before landing in `failed_teams`, and the schedule sets no workflow-level `execution_timeout` to cap total run duration. That means a single run only needs to exceed 6h (not 48h, as before) to strand overrides, and the current activity budget already permits that.

Not blocking since this is an incident-tail scenario and the post-deploy `full_scan` plus the planned watermark cover it. But given the comment at this line states "keep it comfortably above... the worst-case run duration" as the safety invariant, it'd be worth either adding an `execution_timeout` below the window so that's actually enforced, or softening the comment to say it's sized for steady-state duration rather than a hard bound.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -69,9 +78,10 @@ def _positive_int_env(name: str, default: int, logger: structlog.BoundLogger) ->
 #
 # The override row written at merge time is the invalidation signal: each scheduled run
 # picks up distinct_ids whose override landed within the lookback window
-# (RECONCILE_PRECALCULATED_DATA_OVERRIDES_LOOKBACK_HOURS) and repairs just their rows, so the schedule
-# can run at the realtime calculation cadence and a merge is reconciled by the next
-# calculation run instead of hours later. A `full_scan` input ignores the window — use it
+# (RECONCILE_PRECALCULATED_DATA_OVERRIDES_LOOKBACK_HOURS, default DEFAULT_OVERRIDES_LOOKBACK_HOURS)
+# and repairs just their rows, so the schedule can run at the realtime calculation cadence and a
+# merge is reconciled by the next calculation run instead of hours later. A `full_scan` input
+# ignores the window — use it
```
haacked: `nit`: the comment reflow left "— use it" stranded alone on its own line after "A `full_scan` input ignores the window", which reads awkwardly. Worth joining it back with "first-deploy remediation" on the next line.

posthog/temporal/messaging/reconcile_precalculated_data_workflow.py
```diff
@@ -55,6 +55,15 @@ def _positive_int_env(name: str, default: int, logger: structlog.BoundLogger) ->
     return value
 
 
+# Default overrides lookback. Every scheduled run re-diffs *all* overrides inside this window (there
+# is no watermark between runs), so a wide window means each run reprocesses hours of already-handled
+# merges — the reprocess is idempotent but still pays the full read cost. Keep it comfortably above
```
haacked: `nit`: this rationale (no watermark, keep above schedule interval, overlap=SKIP) repeats the same argument made in the module block comment just below and again in `settings/schedules.py`. Three copies will drift as one gets edited and the others don't; consider trimming this one to just what's specific to the constant (why 6, and the note about lowering it later).

haacked (approved):
Nice fix! One correctness question worth double-checking inline, plus a couple of nits.

gustavohstrassburger:
Addressed @haacked's review notes in 0023d67:

- **Hard-bound framing** (line 64): softened the `DEFAULT_OVERRIDES_LOOKBACK_HOURS` comment to say the window is sized for steady-state run duration rather than a hard safety bound — it now notes nothing caps run duration, so a run longer than the window can still strand a merge until the next `full_scan`, and the planned per-run watermark is the real guarantee. Went with softening the comment over adding an `execution_timeout`, keeping this PR to the low-risk runtime cut.
- **Duplicated rationale** (line 60): trimmed the constant's docstring to just what's specific to it (why 6, that it's env-overridable, the lower-it-later note), pointing to the block comment below for how the window relates to the schedule and squash cadence — no more triplicated overlap=SKIP explanation.
- **Stranded `— use it` line** (line 84, nit): rejoined so it reads on one line.

Comment-only diff; `ruff check`/`ruff format` clean and the file compiles.

PR #72922, opened 2026-07-22
feat(exports): point export toast to the Exports panel

sufi-ph:
## Problem

When you export a session recording as an MP4 (or any background/async export), the toast says "Export started" but doesn't say where the finished file will appear. The result lands in the Exports panel (Settings > Exports), which isn't obvious from the toast, so the flow dead-ends.

Raised in a Slack thread.

## Changes

The "Export started" toast now includes a sub-text line linking to the Exports panel: "Find it under Settings > Exports when ready". The link uses `urls.exports()`, so it resolves to the project-scoped `/project/<id>/exports` route and matches link styling used elsewhere.

Because the toast message is now JSX, `exportsLogic.ts` is renamed to `exportsLogic.tsx` (consistent with other kea logics that render JSX). The message is a small exported constant so the existing unit test can assert on it by reference.

Updated the existing `exportsLogic` unit test so the async-export case asserts against the new exported message constant. I (the agent) was unable to run the frontend test suite or typecheck in this environment (dependencies weren't installed at the time of writing), so these have not been executed here — please rely on CI.

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread request. The change is scoped to the shared export toast in `exportsLogic`, so it covers MP4 session-recording exports and every other async export that surfaces in the Exports panel, rather than only the MP4 path. No repo skills applied (frontend copy/UI change). The link reuses `urls.exports()` and the `Link` block styling already used by the incident-note toast pattern, so it stays consistent with existing UI.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1784732389349419?thread_ts=1784732389.349419&cid=C0ACRAMJUAG)*


arnohillen:
Robot review:

This overlaps #72948, which solves the same export-discoverability problem in the same logic. Worth settling which one lands before either goes further, otherwise we'll conflict in exportsLogic.

Two notes if this is the one that goes: renaming the kea logic to `.tsx` to hold a JSX constant puts presentation into the logic file, and the copy says "Settings > Exports" while `urls.exports()` is the standalone `/exports` scene.

PR #69791, opened 2026-07-09
fix(hogql): use TableAccessDeniedError in joins and elsewhere

a-lider:
## Problem

#69743 introduced `TableAccessDeniedError` (typed, carries `table_name`) but left the surrounding code, so a denial could still reach the user and error tracking looking like something else went wrong.

The worst case is a join. A denied table is absent from the schema in exactly the same way a **deleted** one is, and the schema build treated both as "missing" and dropped the join. So `SELECT person.some_join.id` failed with **"Field not found: some_join"** — indistinguishable from a typo. The user has no idea it was a permissions problem. 

This also solves the access control bypass for cohort recalculation (#72923): 
> A member denied `warehouse_objects` can save a cohort predicate such as `person.extended_properties.bool_prop = 'true'` through a configured persons→warehouse join. Their save-time schema omits the denied target (`Database._build_from_sources` skips joins when its target is absent), so compilation fails as an unknown field and this broad handler accepts the save. Background recalculation then uses the new bypass, restores the join, and materializes matching people; the member can use that cohort membership to infer restricted warehouse data.

## Changes

Cleanup, so a denial is always reported *as* a denial:

- **Joins to a denied table now raise the access error instead of "Field not found".** The join is kept rather than dropped, and points at the table **by name** instead of a resolved object. `LazyJoin.join_table` accepts `Table | str`; for a string, `LazyJoin.resolve_table` looks it up through `Database.get_table` at *resolution* time, which raises `TableAccessDeniedError` — matching what a direct reference to the table does. Resolving eagerly at build time instead would raise for every user who merely *has* such a join, breaking queries that never touch it. 

> [!NOTE]
> Keeping the join is safe: it's a tombstone, not a live edge. The table stays out of the schema and in `_denied_tables` (`has_table()` is still `False`), and every read path through a `LazyJoinType` — `has_field`, `get_child`, asterisk expansion, autocomplete — goes through `resolve_database_table` → `get_table`, so it denies. The few places that read `.join_table` without resolving are all `isinstance` guards that a string simply fails.

- `get_table` uses `is_table_access_denied()` rather than poking `_denied_tables` directly — one denial predicate, three layers: `_is_warehouse_table_denied` decides at build time, `is_table_access_denied` answers the boolean for callers that must not resolve (the experiments metric-save gate, the cohort save gate), `TableAccessDeniedError` reports at resolution.

## How did you test this code?

[...]

PR #72923, opened 2026-07-22
fix: bypass warehouse access control in cohorts recalculation

a-lider:
## Problem

Cohort membership recalculation (`calculate_cohort_ch` → `hogql_cohort_subquery_sql`) compiles the cohort's HogQL with **no user**. With the `hogql-warehouse-access-control` flag enabled, a userless schema build throws `TableAccessDeniedError` (every warehouse table is denied) for any cohort whose criteria resolve through a warehouse table (e.g. a `data_warehouse_person_property` filter over a persons join). 

Seen in prod during rollout on two customer teams before rolling back: [error](https://us.posthog.com/project/2/error_tracking/019f8a3d-a7f7-79b1-be71-68115f4585c3). 

## Changes

The model here is the one we use for background work generally: **check access when the definition is created or updated, then just execute**. A cohort is a team-owned definition — its recalculation shouldn't start failing because whoever saved it later lost access (or left the org).

- **Recalculation always bypasses warehouse access control.** - `bypass_warehouse_access_control` param threaded through `HogQLCohortQuery.get_query_executor` into `HogQLQueryExecutor`, which already supports it. 
	- All cohort **execution** paths bypass now — recalculation, the legacy property-filter engine, static cohort population, and query-based cohort population. They all compiled userless and would fail the same way under the flag. 
- **Enforcement moves to save time.** `CohortSerializer.validate` compiles the definition as the request user and rejects the save with a clear message when it resolves through a warehouse table the user can't access. Covers every way a definition can change — `filters`, the legacy `groups` field, query-based cohorts' `query`, and the static-to-dynamic flip (which re-activates preserved criteria). Non-access compile errors don't gate the save (they surface through the existing calculation-error flow), and it's skipped entirely when the flag is off.

> Note: cohorts whose filters were saved *before* this PR (or before the flag ramps) were never save-checked — the bypass grandfathers them in. 

Tests
- Recalc bypass: a warehouse-backed cohort (persons join + `data_warehouse_person_property` filter) fails closed when compiled userless under the flag (proves the fixture genuinely resolves the warehouse table), and compiles with the recalc path's bypass. Catches the regression where the bypass wiring is dropped and background recalcs throw an error again.
- Save gate: a member with a `warehouse_objects: none` rule gets a 400 naming the table and no cohort is created; the same save succeeds without the rule. Catches the regression where the save-time check is removed. 

## Automatic notifications

- [ ] Publish to changelog?


products/cohorts/backend/models/util.py
```diff
@@ -1122,7 +1141,9 @@ def insert_actors_into_cohort_by_query(
 
 
 def insert_cohort_query_actors_into_ch(cohort: Cohort, *, team: Team):
-    context = HogQLContext(enable_select_queries=True, team_id=team.id)
+    # SECURITY-SENSITIVE: background population from the cohort's saved source query, with no
+    # acting user - the query was run by its author when they created the cohort from it.
+    context = HogQLContext(enable_select_queries=True, team_id=team.id, bypass_warehouse_access_control=True)
```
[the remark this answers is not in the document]
a-lider: [fable-review] Valid — the save gate only compiled `filters`, while the new bypass executes query-backed population. Fixed in 5a34e1623ea: the warehouse access check moved to object-level `validate()` and now also compiles `query` as the request user, rejecting the save on denial. Covered by `test_denied_member_cannot_save_query_cohort_over_warehouse_table`.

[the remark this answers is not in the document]
a-lider: [fable-review] Valid, same finding as the sibling threads — query-based static cohorts were never compiled as the author while background population now bypasses. Fixed in 5a34e1623ea: object-level `validate()` compiles `query` as the request user and rejects the save on `TableAccessDeniedError`. One note on framing: before this PR the same save would have led to a fail-closed population failure for everyone (the compile was userless), so this was an availability bug converting into an authz gap only via the new bypass — which is exactly why the gate now covers it.

products/cohorts/backend/models/util.py
```diff
@@ -1122,7 +1141,9 @@ def insert_actors_into_cohort_by_query(
 
 
 def insert_cohort_query_actors_into_ch(cohort: Cohort, *, team: Team):
-    context = HogQLContext(enable_select_queries=True, team_id=team.id)
+    # SECURITY-SENSITIVE: background population from the cohort's saved source query, with no
```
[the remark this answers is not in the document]
a-lider: [fable-review] Valid — `validate_query` only checked the Pydantic shape, so a denied member could save a query-based cohort the bypass would then execute. Fixed in 5a34e1623ea: the warehouse access check moved to object-level `validate()` and compiles `query` as the request user (denial → 400 naming the table). Test added.

products/cohorts/backend/models/util.py
```diff
@@ -765,7 +780,11 @@ def _recalculate_cohortpeople_for_team_hogql(
         history.save(update_fields=["finished_at", "count", "error", "error_code"])
         return 0
     else:
-        cohort_query, hogql_context = hogql_cohort_subquery_sql(cohort, team=team)
+        # SECURITY-SENSITIVE: recalculation always executes without warehouse access control.
+        # It is internal maintenance of a team-owned definition with no acting user - access is
+        # enforced when the filters are saved (CohortSerializer), and failing here would only
+        # silently freeze the cohort's membership.
```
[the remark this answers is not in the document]
a-lider: [fable-review] Valid — `validate_filters` is field-level, so a PATCH carrying only legacy `groups` skipped the access check entirely. Fixed in 5a34e1623ea: the check now runs from object-level `validate()` and derives the effective properties from `groups` when `filters` is absent (same conversion the static-flip guard uses). Covered by `test_denied_member_cannot_update_via_legacy_groups`.

a-lider:
[fable-review] Re the Greptile security-review summary and Veria overview: both point at the same two real gaps — query-based cohorts and legacy `groups` writes weren't covered by the save-time access check while the new execution bypass would run them. Both fixed in 5a34e1623ea: the warehouse access check moved to object-level `CohortSerializer.validate()` and now covers all three definition channels (`filters`, legacy `groups`, `query`), each compiled as the request user with denials rejected at save. Tests added for both channels. See the inline threads for per-finding detail.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,48 @@ def validate_filters(self, raw: dict):
...
 
+    def _validate_warehouse_access(self, attrs: dict) -> None:
+        """Background execution runs the cohort without warehouse access control (the definition
+        is team-owned), so access is enforced here instead: the member saving the definition must
+        be able to read every warehouse table it resolves through. Covers every way a definition
+        can change - `filters`, the legacy `groups` field, and query-based cohorts' `query`."""
+        team = self.context.get("get_team", lambda: None)()
+        request = self.context.get("request")
+        user = getattr(request, "user", None)
+        if team is None or user is None or not getattr(user, "is_authenticated", False):
```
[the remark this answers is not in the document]
a-lider: [fable-review] Valid — the org feature-flag copy flow constructs CohortSerializer with `team_id` but no `get_team`, so the gate's early return silently skipped the check for the destination project. Fixed in 41bfd6c3685: the gate now resolves the team from `get_team`, `team`, or `team_id`, whichever context shape the caller provided.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,48 @@ def validate_filters(self, raw: dict):
...
+            return
+
+        properties: Optional[dict] = None
+        if attrs.get("filters"):
```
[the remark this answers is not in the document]
a-lider: [fable-review] Valid — and the sharper of the two findings: the gate validated the request payload, so benign `filters` + dormant warehouse `groups` at create, followed by a PATCH clearing `filters`, reactivated the groups with neither save checked. Fixed in 316fb2001ee: the gate now validates the effective post-save definition, mirroring `Cohort.properties` precedence (filters win over groups; clearing filters falls back to preserved groups). Covered by `test_denied_member_cannot_reactivate_warehouse_groups_by_clearing_filters`.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
...
+        if team is None and self.context.get("team_id"):
+            team = Team.objects.filter(pk=self.context["team_id"]).first()
+        return team
+
+    def _validate_warehouse_access(self, attrs: dict) -> None:
+        """Background execution runs the cohort without warehouse access control (the definition
+        is team-owned), so access is enforced here instead: the member saving the definition must
+        be able to read every warehouse table it resolves through. Covers every way a definition
+        can change - `filters`, the legacy `groups` field, and query-based cohorts' `query`."""
+        if not any(field in attrs for field in ("filters", "groups", "query")):
```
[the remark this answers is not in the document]
a-lider: Not critical, I already check access when the definition is authored. The flip doesn't change the cohort's criteria (filters)

haacked: `blocking`: A member denied `warehouse_objects` access can populate a warehouse-backed cohort they can't otherwise read by PATCHing only `{"is_static": false}`. When the PATCH body has no `filters`, `groups`, or `query`, this method returns early and bypasses the check. But `update()` still flips `cohort.is_static` to `False` and calls `enqueue_calculation`, which recalculates the preserved `filters`/`groups` under [the recalculation bypass](https://github.com/PostHog/posthog/blob/316fb2001eeb8152e128266aa1937492f4816a29/products/cohorts/backend/models/util.py#L787). Static cohorts are excluded from the periodic recalculation scan, so this PATCH is the only way a dormant static cohort's warehouse-backed criteria ever gets computed, regardless of whether it predates this PR (never save-checked) or was saved by a different member with access.

Treat the static-to-dynamic flip as activating the effective definition too, the same way the feature-flag guard just below already treats it:

```python
    def _validate_warehouse_access(self, attrs: dict) -> None:
        """..."""
        instance = cast(Optional[Cohort], self.instance)
        reactivating_static = (
            instance is not None and instance.is_static and attrs.get("is_static") is False
        )
        if not reactivating_static and not any(
            field in attrs for field in ("filters", "groups", "query")
        ):
            return
        team = self._team_for_warehouse_access_check()
        ...
```

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
...
+                ).get_query_executor(user=user).generate_clickhouse_sql()
+            if attrs.get("query"):
+                context = HogQLContext(team_id=team.pk, team=team, user=user, enable_select_queries=True)
+                query = get_query_runner(attrs["query"], team=team, limit_context=LimitContext.COHORT_CALCULATION)
+                prepare_ast_for_printing(query.to_query(), context=context, dialect="clickhouse")
+        except TableAccessDeniedError as e:
+            raise ValidationError(
+                f"Can't save this cohort: you don't have access to table `{e.table_name}`, which its definition uses."
+            )
+        except Exception:
```
haacked: `suggestion`: This `except Exception: return` is now the sole enforcement point for warehouse access, since every background execution path bypasses the check. It needs to catch `TableAccessDeniedError` (already handled above), so the current approach is safe in practice. But if a denial gets wrapped in a different exception type by a future change, the save would silently succeed and nothing would signal that the gate broke.

Narrow the catch to the expected compile-error family and log anything unexpected instead of swallowing it:

```python
    except TableAccessDeniedError as e:
        raise ValidationError(...)
    except BaseHogQLError:
        # Expected compile problems (bad filters, unknown fields) surface through the
        # calculation-error flow, not the save.
        return
    except Exception as e:
        capture_exception(e)
        return
```

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
...
+        request = self.context.get("request")
+        user = getattr(request, "user", None)
+        if team is None or user is None or not getattr(user, "is_authenticated", False):
+            return
+
+        # Validate the effective post-save definition, mirroring Cohort.properties precedence
+        # (filters win over groups): clearing filters re-activates preserved legacy groups, so
+        # what survives the save is what must be checked - not just the fields in the payload.
+        instance = cast(Optional[Cohort], self.instance)
+        filters = attrs["filters"] if "filters" in attrs else (instance.filters if instance else None)
```
haacked: `suggestion`: The filters-then-groups precedence here duplicates [`_effective_filters_after_update`](https://github.com/PostHog/posthog/blob/316fb2001eeb8152e128266aa1937492f4816a29/posthog/api/cohort.py#L1010) just above: both take filters from `attrs` or the instance, fall back to `groups`, and build an unsaved `Cohort` to read `.properties`. If the precedence rule ever changes, updating one and missing the other is easy.

Make `_effective_filters_after_update` null-safe for the create path (it currently assumes a non-null instance) and call it here instead of reimplementing the logic.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
         self._validate_feature_flag_constraints(raw, cohort_will_be_static)  # keep your side-rules
         return raw
 
+    def _team_for_warehouse_access_check(self) -> Optional[Team]:
```
haacked, line 1061: `suggestion`: The `team`/`team_id` fallback branches here have no test coverage. Every existing test posts through the cohorts viewset, which always supplies `get_team`, so only the first branch ever runs. This is the exact fallback meant to close the earlier finding that the org feature-flag-copy flow builds `CohortSerializer` with `team_id` but no `get_team`; the fix is unverified.

Add a serializer-level test that constructs `CohortSerializer(data={...warehouse filters...}, context={"request": denied_member_request, "team_id": team.id})` and asserts `is_valid()` raises with the table name in the message, then a second case using the `"team"` key.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
...
+        # experiments) pass "team" or "team_id" instead - resolve all three so the access
+        # check can't be skipped by a context shape.
+        team = self.context.get("get_team", lambda: None)()
+        if team is None:
+            team = self.context.get("team")
+        if team is None and self.context.get("team_id"):
+            team = Team.objects.filter(pk=self.context["team_id"]).first()
+        return team
+
+    def _validate_warehouse_access(self, attrs: dict) -> None:
```
haacked, line 1073: `suggestion`: This compiles the full HogQL database (warehouse tables, saved queries, joins, external sources) on every cohort save that touches `filters`/`groups`/`query`, regardless of whether `hogql-warehouse-access-control` is enabled for the team. `TableAccessDeniedError` can only be raised when that flag is on, so with it off the compile can never gate the save. When the flag is off, this is added request latency with no benefit.

Check the flag before compiling, and return early when it's off, so only teams actually subject to the control pay the cost.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
...
+            # Only access denials gate saving; other compile problems surface elsewhere.
+            return
+
     def validate(self, attrs: dict) -> dict:
         # Field-level validate_filters only runs when the PATCH body includes `filters`. This
         # object-level guard covers the static-to-dynamic flip when it does not, re-checking the
         # instance's preserved behavioral filters against the feature-flag rule.
         attrs = super().validate(attrs)
 
+        self._validate_warehouse_access(attrs)
```
haacked, line 1130: `question`: The [org cross-project feature-flag-copy flow](https://github.com/PostHog/posthog/blob/316fb2001eeb8152e128266aa1937492f4816a29/products/feature_flags/backend/api/organization_feature_flag.py#L1265) builds each destination cohort from the source cohort's properties and saves it through `CohortSerializer` with the copying user's request. This check now runs against the copying user and can reject a copy that previously always succeeded if the source cohort resolves through a warehouse table the copying user can't read, even though it was authored by someone else with access.

Is that the intended behavior, or should the copy path skip this check since the definition already went through it once?

products/cohorts/backend/models/util.py
```diff
@@ -290,7 +290,11 @@ def format_person_query(cohort: Cohort, index: int) -> tuple[str, dict[str, Any]
         return "SELECT generateUUIDv4() as id WHERE 0 = 19", {}
 
     # Compile the cohort criteria via HogQLCohortQuery and embed the result as a person-id subquery.
-    cohort_query, cohort_context = hogql_cohort_subquery_sql(cohort, team=cohort.team)
+    # SECURITY-SENSITIVE: executes a saved, team-owned cohort definition with no acting user in
```
haacked, line 293: `nit`: The five `SECURITY-SENSITIVE` comments across this file restate the same rationale in slightly different words. Worth pointing four of them at the fullest explanation (or at `_validate_warehouse_access`) so the reasoning lives in one place.

posthog/api/test/test_cohort_warehouse_access_control.py
```diff
@@ -0,0 +1,161 @@
...
+        AccessControl.objects.create(team=self.team, resource="warehouse_objects", access_level="none")
+
+        response = self._save_cohort()
+
+        assert response.status_code == 400, response.content
+        assert "Can't save this cohort" in str(response.json())
+        assert "extended_properties" in str(response.json())
+        assert not Cohort.objects.filter(team=self.team).exists()
+
+    def test_member_with_access_can_save_warehouse_cohort(self):
```
haacked, line 111: `nit`: This only asserts the 201 status. Add `assert Cohort.objects.filter(team=self.team).exists()` to mirror the negative assertion in the deny test above.

haacked (commented):
Solid fix, and the four earlier bot findings are genuinely closed. One blocking issue still open inline (same static-to-dynamic gap veria-ai flagged), plus a few non-blocking suggestions.

posthog/api/cohort.py
```diff
@@ -1047,12 +1051,68 @@ def validate_filters(self, raw: dict):
         self._validate_feature_flag_constraints(raw, cohort_will_be_static)  # keep your side-rules
         return raw
 
+    def _team_for_warehouse_access_check(self) -> Optional[Team]:
+        # The viewset provides get_team; out-of-viewset constructions (feature flag copy,
+        # experiments) pass "team" or "team_id" instead - resolve all three so the access
+        # check can't be skipped by a context shape.
+        team = self.context.get("get_team", lambda: None)()
+        if team is None:
+            team = self.context.get("team")
```
ReeceJones: Should this check be reversed? I.e. do `self.context.get("team")` first. I have a feeling `get_team` might issue a database query and might not be cached which may cause unnecessary work if `team` is already defined in the context

ReeceJones (approved):

posthog/api/cohort.py
a-lider: Good catch - I initially rejected this but you're right, and the static-exclusion point is what changes it: confirmed `get_cohort_calculation_candidates_queryset` does `.exclude(is_static=True)`, so for a static cohort the flip is the *only* path that ever computes the criteria. The "any update recalcs a dynamic cohort anyway" reasoning doesn't apply. Fixed in 5378076 - the flip is now treated as activating the effective definition (and the preserved query is checked on the flip too), with a test.

posthog/api/cohort.py
a-lider: Leaving as-is for now. `TableAccessDeniedError` is caught explicitly just above, so the broad catch only swallows non-access compile errors (broken filters etc.), which are meant to surface through the calculation-error flow rather than block the save. A future change wrapping denials in a new type is speculative enough that I'd rather not add the extra branch here.

posthog/api/cohort.py
a-lider: Agreed - done in e17ef70. Made `_effective_filters_after_update` null-safe for the create path (optional team) and the gate now calls it, so the filters-win-over-groups precedence lives in one place.

posthog/api/cohort.py
a-lider: Going to skip the dedicated serializer-level test here - the fallback is exercised in practice by the copy flow, and the branch is simple context resolution. Happy to add one if you feel strongly.

posthog/api/cohort.py
a-lider: Agreed - done in 0b4afd3. Gate the compile on the same flag the enforcement in `Database.create_for` uses, so it's skipped entirely when the control is off.

posthog/api/cohort.py
a-lider: Intended, keeping it. A user shouldn't be able to copy a cohort that reads a warehouse table they can't access into a project - the source authorization was for the source project and doesn't transfer across the copy. This is also the fix for the earlier finding that this same copy path *skipped* the check (built the serializer with `team_id` and no `get_team`).

products/cohorts/backend/models/util.py
a-lider: Ended up leaving these as-is - they're short one/two-line notes and each sits at its own callsite (legacy property-filter engine, distinct-id compile, recalculation, query-based population, static first-time population). The slight overlap in wording is acceptable for a security-sensitive marker I'd rather have spelled out at each site than indirected to one place.

posthog/api/test/test_cohort_warehouse_access_control.py
a-lider: Done - 0eca3b0.

posthog/api/cohort.py
a-lider: Verified: `get_team` wraps `self.team`, a `@cached_property` already warmed by the permission check earlier in the request, so it doesn't issue an extra query in practice - and the caller context shapes are mutually exclusive (viewset has `get_team`, copy has `team_id`, experiments have `team`). But checking the materialized `context["team"]` before calling the lambda is strictly cheaper and clearer, so reversed it in 408f53c. Good call.

posthog/api/cohort.py
```diff
@@ -1047,12 +1058,77 @@ def validate_filters(self, raw: dict):
...
+                ).get_query_executor(user=user).generate_clickhouse_sql()
+            if query:
+                context = HogQLContext(team_id=team.pk, team=team, user=user, enable_select_queries=True)
+                runner = get_query_runner(query, team=team, limit_context=LimitContext.COHORT_CALCULATION)
+                prepare_ast_for_printing(runner.to_query(), context=context, dialect="clickhouse")
+        except TableAccessDeniedError as e:
+            raise ValidationError(
+                f"Can't save this cohort: you don't have access to table `{e.table_name}`, which its definition uses."
+            )
+        except Exception:
```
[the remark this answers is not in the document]
a-lider: Valid, I'll address this separately in #69791

PR #72925, opened 2026-07-22
chore(insights): prototype unsaved-insight notice as a table row

thmsobrmlr:
## TL;DR

Throwaway prototype, not meant to merge as-is. It shows three looks for the "You have an unsaved insight" notice on the insights list page. The main one turns the notice into a draft row pinned to the top of the table. Flip between looks in the browser with the floating bar at the bottom or `?variant=`.

## Problem

The insights list shows an unsaved draft as a plain text line above the table: "You have an unsaved insight from [date]. Click here to view it." It is easy to miss. Question to settle: should the draft be an item in the table below instead?

## Changes

All prototype code lives in `ReloadInsightPrototype.tsx` and only runs in local dev builds (`NODE_ENV === 'development'`). Production keeps the current text line. On `/insights`, switch variants with the floating bottom bar, the left/right arrow keys, or `?variant=`:

- `A` (default): the draft is a pinned, highlighted first row of the table, with the insight icon, summary, a "Draft" tag, timestamps, and a row menu (Continue editing / Discard).
- `B`: the draft is a strip between the filters and the table, with a primary "Continue editing" button.
- `C`: the current banner shape, upgraded with an action button and a dismiss control.

If localStorage has no real draft, a sample draft is shown and labelled in the bottom bar. Discard is stubbed with a toast so the demo draft survives flipping.

Automated checks only, no manual run (the agent environment cannot start the dev stack, so no screenshots):

- oxlint (repo config): clean on both changed files
- oxfmt applied
- `pnpm --filter=@posthog/frontend typescript:check`: no errors in the changed files (the remaining errors are pre-existing, from unbuilt quill workspace packages in the sandbox, all in untouched files)
- `hogli ci:preflight --fix`: pass

To try it: `./bin/start`, open `/insights`, use the bottom bar.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed: dev-only prototype with no user-facing change.

## 🤖 Agent context

Built by Claude in a PostHog Code session with the `/prototype` skill (UI branch: several variants on the existing route, plus a floating switcher). Decisions: variant A reuses the real table columns and special-cases only three spots in `SavedInsights.tsx` (name cell, row menu, bulk selection) so the row looks exactly like production; the gate is `NODE_ENV === 'development'` rather than "not production" so Jest and Storybook stay untouched; discard is a stub so the demo data survives. When a variant wins, fold it into the real code properly (real discard, no type casts) and close this PR unmerged; the branch stays as the record of the options.

---

### DIFF
--- a/frontend/src/scenes/saved-insights/ReloadInsightPrototype.tsx
+++ b/frontend/src/scenes/saved-insights/ReloadInsightPrototype.tsx
@@ -1,0 +1,292 @@
+/**
+ * PROTOTYPE — THROWAWAY CODE, do not ship.
+ *
+ * Question: should the "You have an unsaved insight from <date>. Click here" text (ReloadInsight)
+ * become an item in the saved-insights table below it?
+ *
+ * Three variants on the existing /insights route, switchable via `?variant=` or the floating
+ * bottom bar (also ← / → keys):
+ *   A — the draft is a pinned first row in the insights table (default)
+ *   B — the draft is a strip above the table
+ *   C — the draft is a dismissible banner (evolved status quo)
+ *
+ * Run `./bin/start` and open /insights. Caveats: everything is gated to local dev builds;
+ * if no real draft exists in localStorage a sample draft is shown (labelled in the bottom bar);
+ * "Discard" is stubbed with a toast so the demo draft survives; client-side column sorts can
+ * move the pinned row in variant A.
+ */
+import { useValues } from 'kea'
+import { router } from 'kea-router'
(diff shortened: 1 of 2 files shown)


thmsobrmlr:
Verdict: variant A won — the draft becomes a pinned row in the insights table. The production implementation is in #72971; this prototype branch stays as the record of the three variants, and this PR is closed unmerged.

*This comment was posted by PostHog Code on behalf of the task author.*

PR #63907, opened 2026-06-16
feat(insights): add Metric trends display type

sampennington:
## Problem

I want to quickly see how metrics are doing.

## Changes

- **New `ChartDisplayType.Metric`** — a selectable trends type:
Screenshot 2026-06-17 at 11 46 23

You can set the colors to whatever:
Screenshot 2026-06-17 at 11 48 12

Screenshot 2026-06-17 at 11 48 16

It's behind a flag, I want to make some improvements to the component from Mikes feedback, but will do that separate.
Also think it's worth getting some quickinternal feedback before we go live

## How did you test this code?
locally, tests, many rounds of bot reviews

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Authored with Claude Code (Opus 4.8). The human directed the intent and iterated heavily on the visuals.

Key decisions:
- **Separate `Metric` type rather than overloading `BoldNumber`** — keeps "Big number" untouched for existing insights and avoids a breaking enum change; the two co-exist.
- **Sparkline needs a time-series result.** `BoldNumber` is a total-value display, so its query returns `data: []`; `Metric` is deliberately left out of the total-value allowlist so it gets the per-interval series the sparkline needs. Headline uses the series total (`count`).
- **Self-contained change pill** (start → last of the series) so the tile doesn't depend on the compare filter.
- **Layout scales by filling** on dashboards (sparkline grows) instead of per-grid-size conditions; fixed padding, content-height in the insight editor.

---

> [!NOTE]
> Adds a new `ChartDisplayType.Metric` trends display type that renders a headline value with a sparkline and period-over-period change pill. Gated behind a feature flag. Includes display configuration, editor filter controls for colors, and proper integration with existing display type categorization (single-series, no-breakdown, no-legend, no-annotations).
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 84089142ab3250655442c7f1fa07821bd2bc1be6.</sup>

PR #28880, opened 2025-02-18
fix: Turn margin into padding in the parent

rafaeelaudibert:
The margin inside these components summed when paired with the `w-full` is creating a small scrollbar, let's change that by using padding in the parent frame instead

| Before| After |
|--------|--------|
| image | image |

PR #72927, opened 2026-07-22
fix(insights): let table display bleed to the card edges

thmsobrmlr:
## TL;DR

When you view an insight as a table, the table floats inside its card with a small gap on every side, so its lines stop before the card border and it looks cropped. This change makes the table reach the card edges again.

## Problem

Insights rendered with the "Table" chart type sit 0.5rem inside the card border. The header and row separators end short of the edge, and extra white space shows below the last row:

before

`.InsightVizDisplay__content` gained `padding: 0.5rem` in #28880, which replaced child margins that caused a scrollbar. Before that, the table display had `margin: 0` and bled to the edges. #63907 later restored edge-to-edge rendering for the Metric display with a `:has(.Metric)` override, but the table never got the same treatment.

## Changes

Extend the padding override to the table display's wrapper classes: `TrendsInsight--ActionsTable` for the insight scene and `InsightCard__viz--ActionsTable` for embedded surfaces. The table's own cell padding (doubled on first and last cells) keeps the content clear of the border.

> [!NOTE]
> This also removes the inset on dashboard tiles, exports, and notebook nodes, which render the same wrapper. Tables there now sit flush with the card, matching SQL insight and web analytics tables.

CSS-only change, so no automated tests. I (or, actually, Claude) verified the selectors against the class chain in `Trends.tsx` and `InsightVizDisplay.tsx`, and mirrored the `:has(.Metric)` override that already works in the same file. The task sandbox has no dev stack, so no manual UI check ran. The Storybook exporter story `TrendsTableInsight` renders the embedded variant for review.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs cover this spacing.

## 🤖 Agent context

Claude (PostHog Code cloud task) wrote this from a bug report with the screenshot above. It traced the inset to #28880, which moved child margins into parent padding and took the table display along. It chose the smallest fix: extend the `:has()` padding override that #63907 added for the Metric display. It rejected removing the parent padding outright (that would bring back the scrollbar #28880 fixed) and a TSX-side conditional class (the CSS pattern already exists in this file). Retention's grid table keeps its padding on purpose: it has no full-width separators, so the breathing room reads as intended. No repo skills applied to a CSS-only change.

---

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -2317,9 +2317,9 @@ snapshots:
     exporter-trends--trends-pie-insight-legend--light:
         hash: v1.k794b7964.2eee29146a4cd703a88df5986742cf2bfe804374e9b46e04f3a065acb3bb0d8f.vDNDFJu83-A5zpZF5hhUW1uHl36Hd18AwHbnNLrLnOA
     exporter-trends--trends-table-insight--dark:
-        hash: v1.k794b7964.92191f5d6ed09bce9843b2168a0ae277493e522cfb6bd52fd7a5b91397d91fa5.2VnfYOZnlJzHQtD34o5Rws-OY7mm8onFWP_ZIfiPj-8
+        hash: v1.k794b7964.d2351308c6493ff330fef71e6af5d43c0a52ef28004f39d7112c49cb9ba6b4d4.vVRXq4GJ2nofvbBEpopRDL3DsKgpTC2fL5H4KNEdjMU
     exporter-trends--trends-table-insight--light:
-        hash: v1.k794b7964.fbfe6112e758938dcdce47ad87f44b5a12846796ae55ffebbc7f5c1f2b796fa3.cjdnak6QmLsWypy7NBfjJLFAJPHwao2i6QwqgqIvjpc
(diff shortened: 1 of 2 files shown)


thmsobrmlr:
Triage note on the failing visual review gate: the run flags 4 changed snapshots, and all four are this PR's intended change, verified against the baseline/current images. The full 16-shard run confirmed the same four snapshots, byte-identical (same content hashes) to the verified partial run.

- `exporter-trends--trends-table-insight` (light + dark): the exported table now spans the full card instead of stopping short of the border.
- `scenes-app-insights-mcp-missing-capability` (light + dark): the insight scene in Table mode — the table sits flush under the computed bar, separators run edge to edge, and the dead space below the last row is gone.

The other 1964 snapshots are unchanged. Approving the [Visual Review run](https://us.posthog.com/project/2/visual_review/runs/cfa1ea68-cfb5-43e0-8322-0ac5e256b52a) greens the gate; baseline approval stays with a human on purpose.

(The earlier chromium shard failure was the known-flaky customer analytics expanded-row play test; it passed on rerun and was reported to devex. All 16 shards passed on the full run.)

*This is a comment from PostHog Code.*

PR #72929, opened 2026-07-22
feat(warehouse-sources): reuse synced parent tables for fan-out endpoints

danielcarletti:
## Problem

A customer syncing a fan-out endpoint pays to download the same parent list over and over. Selecting Sentry's `issues` plus its three children pulls the full issues list four times per cycle. That burns their vendor API quota, trips rate limits, and slows every child sync. Fleet-wide the duplication is large: Stripe's customer children alone re-enumerate full customer lists hundreds of thousands of times a month.

The cause is that each schema syncs as its own independent Temporal run, so nothing shares the parent fetch.

The parent's rows are already sitting in the warehouse as a Delta table after the parent schema syncs. dlt and Airbyte solve duplicate parent reads by coordinating parent and children in one extract; we can instead read the parent back from the destination, which keeps the one-schema-one-run model untouched.

## Changes

Everything is gated by the `warehouse-fanout-parent-reuse` feature flag - flag off means byte-identical legacy behavior, so rollout and rollback are a flag flip.

```mermaid
flowchart LR
    A1["Sentry API /issues"] --> D1[("Delta: issues")]
    A2["Sentry API /issues (re-pull per child sync)"] --> F2["fan-out per issue"] --> H2["API /issues/:id/hashes"] --> D2[("Delta: issue_hashes")]
    class A1,A2,H2 phRed;
    class D1,D2,F2 phGray;
```

```mermaid
flowchart LR
    A1["Sentry API /issues"] --> D1[("Delta: issues")]
    D1 -->|"read parent IDs from Delta"| F2["fan-out per issue"] --> H2["API /issues/:id/hashes"] --> D2[("Delta: issue_hashes")]
    G{{"issues enabled + initial sync complete? else read the API"}} -.-> F2
    class G phBlue;
    class A1,H2 phRed;
    class D1,D2,F2 phGray;
```

- **Warehouse parent reader** (`common/rest_source/warehouse_parent.py`): streams a parent schema's Delta table as `list[dict]` pages shaped like REST parent pages. Takes API field names, maps to the snake_case physical columns, re-keys rows back. Pins one Delta snapshot for the whole fan-out. A parent whose newest job is mid-sync **or died mid-write** pins to its last completed job's version, so the child never reads a torn table; a completed newest job reads the latest version, so a pin can never point at files a later compaction vacuumed.
- **Opt-in flag on the fan-out config**: `DependentEndpointConfig.parent_source: "api" | "warehouse"` (default `"api"`). When active, `build_dependent_resource` swaps the parent's HTTP pagination for the reader via a new `data_iterator` resource entry in `create_resources` - the name-based resolve wiring, `include_from_parent`, and renames are untouched. The child gets a `404 → ignore` response action since the snapshot can contain parents deleted upstream.
- **Soft dependency, decided per run**: sources declare their parents via `get_required_parent_schemas` (derivable from endpoint configs via `required_parents_from_endpoint_configs`). `_warehouse_parent_reuse_available` then picks the path for each run: a parent that is missing, disabled, on append sync, or not yet initially synced sends that run down the parent-API path it uses today. The API constrains nothing, so no selection is refused and no parent is ever enabled as a side effect (parent syncs bill rows, so that stays the customer's choice).
- **The fallback also covers an unreadable parent table.** The run-time check reads Postgres, so a parent can claim a completed sync while its Delta table is purged, renamed, or missing the fan-out columns. Both build sites resolve the table eagerly in sync build context, catch that error, and take the API path, dropping the snapshot-only behaviors with it (the child's stale-parent `404 → ignore`, and resume checkpoints). A warehouse run also clears any leftover resume checkpoint at build time: the pipeline reads that Redis state to pick replace-vs-append for chunk 0, and the warehouse scan reprocesses from the start, so a stale checkpoint would duplicate rows in full-refresh tables across a mode flip.
- **Why a fallback rather than a hard requirement.** 8 sources across US and EU already run a Sentry issue-child with `issues` off or on append. A hard requirement would convert those into permanently failing syncs the moment the flag reached them. Falling back means enabling the flag cannot break a schema that syncs today, and those sources upgrade themselves once the parent is enabled.
- **Sentry is the first adopter**: `issue_events` and `issue_hashes` via the config flag; `issue_tag_values` reads its top-level issues from the warehouse in its custom iterator, converting the ordered early-stop into a per-row retention cutoff over the unordered stream, with stale-issue 404 skips.
- **Extracted `build_delta_table_uri`** into a leaf module so reader and writer derive the table URI from one place (no behavior change for the writer).
- Every fan-out child run logs `data_imports.fanout_parent_source` (INFO) with the path it took, and a run that falls back logs `data_imports.fanout_parent_unusable` with the reason. Both fire flag on or off, and drive rollout adoption.
- Documented the mechanism in the `implementing-warehouse-sources` skill, explicitly marked not generally available. Anyone changing fan-out internals needs the constraints: never materialize the parent, keep windowed parents windowed, catch the unreadable-table error. Nothing should opt another source in until Sentry proves the pattern.

> [!NOTE]
> Behavior change only for Sentry issue-children teams with the flag enabled. All other fan-out sources (Stripe, Github, Cloudflare, Intercom, ...) keep their current path; converting them is planned as follow-up phases (Github needs a watermark-filtered read since its parent walk is already windowed).

**End to end on a devbox against a real Sentry source.** Each child synced both ways within half an hour, with the incremental watermarks unchanged between runs:

| schema | rows via API parent | rows via warehouse parent | duration |
| --- | --- | --- | --- |
| `issue_events` | 33 | 33 | 8.3s → 10.1s |
| `issue_hashes` | 6 | 6 | 17.3s → 8.5s |
| `issue_tag_values` | 77 | 77 | 105.9s → **37.0s** |

Row counts are identical through both parent sources, which is the result that matters. Only `issue_tag_values` has a parent walk big enough for the timing to mean anything. The other two are noise at this size.

The flag itself went untested there. A devbox evaluates flags against cloud project 2 rather than its own instance, so the gate was forced on locally for these runs.

Automated, covering the regressions nothing else caught:

- `test_fanout.py`. The warehouse branch builds a `data_iterator` parent and a 404-ignore child. Flag off keeps the byte-identical API path. An unreadable parent table drops back to API pagination and loses the snapshot-only 404 handling with it. A run through the real `rest_api_resources` proves the child fans out from warehouse pages with parent-id injection and renames intact, making **zero requests to the parent path**.
- `test_warehouse_parent.py`: paging and API-name re-keying against a real local Delta table, which catches the camelCase→snake_case mapping breaking, plus typed errors for a missing table or column.
- `test_import_data_sync.py`: a missing, disabled, append, or never-synced parent falls the run back to the API path instead of failing it. That is the regression that would have broken the 8 sources above.
- `test_sentry.py`: `issue_tag_values` reads issues from the warehouse and never calls the issues endpoint, falls back to the API walk when the table is unreadable, and skips stale-issue 404s.
- `TestFanoutParentSelection` and `TestFanoutParentCreation`: no schema selection is refused and no parent is enabled as a side effect, at both update and create time.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

User-facing behavior is flag-gated and off; the contributor-facing pattern doc lives in the `implementing-warehouse-sources` skill (updated in this PR).

## Rolling this out

The flag exists in cloud project 2, inactive at 0%. It needs a release condition rather than a blanket enable. A devbox sends `project: "2"` because its team id is literally 2, which collides with our own project. Watch `data_imports.fanout_parent_source` for adoption, and `data_imports.fanout_parent_unusable` for teams that keep falling back.

## 🤖 Agent context

Built by Claude Code (Fable 5) from an approved plan after a code-grounded investigation of how fan-out endpoints duplicate parent fetches across the ~18 sources that use them. Skills invoked: `/implementing-warehouse-sources`, `/improving-drf-endpoints`, `/writing-tests`, `/writing-user-facing-copy`, `/writing-pr-descriptions`. Key decisions: read the parent from the warehouse rather than coordinating parent+children into one run (keeps the per-schema run model); keep the parent resource entry and swap only its page generator so the resolve wiring stays intact; never enable a parent as a side effect of enabling a child.

The design started as a hard dependency that refused unusable configurations at the API and failed the run otherwise. A fresh-eyes adversarial review flagged that this would break existing teams at rollout; a fleet query over the US and EU mirrors found 8 such sources, and I (actually Claude) replaced the whole enforcement layer with the per-run fallback described above. That deleted the selection-time refusals, the parent-disable protection, and the parent-editor access check those refusals needed. A later pass removed the code and documentation those deletions orphaned, so the API layer ends at zero net change.


products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py
```diff
@@ -93,6 +102,64 @@ def _get_external_data_schema(schema_id: uuid.UUID, team_id: int) -> ExternalDat
...
+        # A full-refresh parent rewrites its table chunk by chunk (overwrite + appends), so a
+        # child reading mid-rewrite would fan out over a partial snapshot and, with replace
+        # disposition, silently shrink its own table. Parent and child schedules fire together,
+        # making this routine, not rare. Deliberately retryable (plain Exception): the parent
+        # finishing resolves it — either a retry attempt or the next schedule succeeds.
+        parent_running = await database_sync_to_async_pool(
+            ExternalDataJob.objects.filter(
+                team_id=team_id, schema_id=parent.id, status=ExternalDataJob.Status.RUNNING
+            ).exists
+        )()
```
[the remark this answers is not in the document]
danielcarletti: AI reply: Confirmed and fixed in cbf8528. The gap was that the RUNNING check ran in the activity gate while the Delta open happened lazily at first read, minutes into the pipeline.

`resolve_parent_table_ref` now pins the parent's Delta version at source-build time, right after the gate, and the reader opens `DeltaTable(uri, version=pinned)`. Later parent commits are invisible to the run, so the child fans out over one complete snapshot instead of a torn overwrite+append sequence. Residual window is sub-second (a parent job would have to start *and* commit its first overwrite between the gate and the pin) versus the whole read before.

Test: `test_reader_stays_on_the_pinned_version_when_the_parent_re_syncs` overwrites the table after resolve and asserts the original rows still come back.

products/warehouse_sources/backend/presentation/views/external_data_schema.py
```diff
@@ -520,6 +564,68 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+            except FanoutParentDependencyError as e:
+                raise ValidationError(str(e))
+            if action == "enable":
+                assert parent is not None  # "enable" implies the parent exists
+
+                # In the bulk path update() runs inside a per-schema transaction; update_should_sync
+                # does Temporal RPCs, so defer it post-commit — no idle-in-transaction network I/O,
+                # and no Temporal schedule surviving a rolled-back request.
+                def _enable_parent(parent_id: str = str(parent.id)) -> None:
+                    update_should_sync(schema_id=parent_id, team_id=instance.team_id, should_sync=True)
```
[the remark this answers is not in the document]
danielcarletti: AI reply: Confirmed and fixed in cbf8528. `update_should_sync` commits `should_sync` before it touches the schedule, so an RPC failure did leave an "enabled" parent with no schedule, and the run-time gate would keep accepting it.

`_enable_parent` now rolls `should_sync` back to False on failure and re-raises (the request still fails and the error is still captured). The child's next run then fails loudly on the existing "parent must be enabled" gate instead of quietly fanning out over a frozen table.

Test: `test_parent_enable_rolls_back_when_the_schedule_update_fails` makes the parent's `update_should_sync` raise and asserts the parent stays disabled.

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/fanout.py
```diff
@@ -98,6 +120,28 @@ def build_dependent_resource(
         "table_format": "delta",
     }
 
+    if warehouse_parent:
+        if source_id is None:
+            raise ValueError("source_id is required when a fan-out reads its parent from the warehouse")
```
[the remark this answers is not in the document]
danielcarletti: AI reply: Good catch — fixed in 07620e9. Changed to `if not source_id:` in fanout.py, and the sibling guard in sentry.py (`_iter_issue_tag_values_rows`) to `if team_id is None or not source_id:`. Both now reject an empty string up front with the clear message instead of letting it reach the UUID parse in `resolve_parent_table_ref`.

products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py
```diff
@@ -93,6 +102,64 @@ def _get_external_data_schema(schema_id: uuid.UUID, team_id: int) -> ExternalDat
...
+        if parent_running:
+            raise Exception(
+                f"Parent schema '{parent_name}' is currently syncing; '{schema.name}' reads its table "
+                f"and would see a partial snapshot. This run will retry and the next schedule re-evaluates."
```
estefaniarabadan: just thinking out loud: retry is not gonna work if the parent takes longer than the retry time. I've seen that we have a table_ref pinned to a delta version of the table, could we just pin to the latest completed job and continue with the child? 🤔

products/warehouse_sources/backend/presentation/views/external_data_source.py
```diff
@@ -3103,6 +3148,7 @@ def database_schema(self, request: Request, *arg: Any, **kwargs: Any):
                 "detected_primary_keys": schema.detected_primary_keys,
                 "permission_error": endpoint_permissions.get(schema.name),
                 "rls_warning": schema.rls_warning,
+                "requires": schema.requires,
```
estefaniarabadan: (claude comment): requires lands in the payload but nothing in this PR consumes it

products/warehouse_sources/backend/presentation/views/external_data_schema.py
```diff
@@ -523,6 +567,79 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+                        # failed RPC would leave the parent "enabled" with no schedule — and the
+                        # child's run-time gate, seeing an enabled and already-initialized parent,
+                        # would keep fanning out over a table that never syncs again. Put the flag
+                        # back so the child fails loudly instead of going quietly stale.
+                        ExternalDataSchema.objects.filter(id=parent_id, team_id=instance.team_id).update(
+                            should_sync=False
+                        )
+                        raise
+
+                self._run_temporal_side_effect(_enable_parent)
```
estefaniarabadan: could we be causing issues here? if the customer doesn't want to sync the parent, could they incur in billing increase with this? 🤔

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/warehouse_parent.py
```diff
@@ -0,0 +1,120 @@
...
+        raise WarehouseParentTableNotFoundError(
+            f"Parent table '{parent_name}' is missing requested column(s) {missing_columns} — "
+            f"re-sync the parent schema so the fan-out fields are present"
+        )
+
+    projected = list(dict.fromkeys(physical_by_api_name.values()))
+    dataset = delta_table.to_pyarrow_dataset()
+
+    page: list[dict[str, Any]] = []
+    for batch in dataset.to_batches(columns=projected, batch_size=page_size):
```
estefaniarabadan, line 266: I feel we should cap the page_size, this is api-driven and we could end up with too much in memory

estefaniarabadan (changes requested):

products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py
danielcarletti: AI reply: Great idea — implemented in 15802fd, and it made the design simpler. `resolve_parent_table_ref` now checks for a RUNNING parent job itself: if one is in flight, it pins the read to the version as of the parent's **last completed job** via Delta time travel (`load_as_version(finished_at)` — `finished_at` is stamped strictly after the job's final commit, so every commit at or before it belongs to a finished run). The retryable "parent is syncing" gate in `import_data_activity_sync` is gone entirely; the child just proceeds against a complete snapshot no matter how long the parent takes. Vacuum retention comfortably covers files dereferenced seconds ago by the in-flight rewrite. Test: `test_resolve_pins_to_last_completed_snapshot_while_parent_is_syncing`.

products/warehouse_sources/backend/presentation/views/external_data_source.py
danielcarletti: AI reply: Correct — `requires` is forward wiring for the wizard/schema UI (surface the dependency before the user hits the 400), which lands as a separate frontend PR so this one stays backend-only. It just became more important: after the billing discussion below we removed the silent parent auto-enable, so the UI needs `requires` to tell the user up front that selecting `issue_events` needs `issues` enabled too. Kept it so the frontend PR doesn't need a backend companion.

products/warehouse_sources/backend/presentation/views/external_data_schema.py
danielcarletti: AI reply: Really good catch, and it changed the design in 15802fd. You're right that silently enabling the parent silently grows billed rows synced (billing is `Sum(rows_synced)` over `billable=True` jobs, and schedule-fired runs are always billable). So the auto-enable is gone: enabling a child whose parent is disabled now refuses with "Enable 'issues' first. Rows synced by 'issues' count toward your usage." — same rule at source creation. Enabling the parent is always the customer's own action, never a side effect.

Scale check against prod mirrors: only 1/126 (US) + 1/84 (EU) Sentry sources have a child enabled without the parent — 99% of fan-out users already sync and pay for the parent, so this refusal affects almost nobody while protecting exactly the people who deliberately disabled the parent. (Stripe, for the later phase: 3/2,164.) Two further mitigations if we ever want them: new sources get the 7-day free-historical window (the parent's initial backfill wouldn't bill), and `ExternalDataJob.billable` already exists per-job if product ever decides dependency-parent syncs should be free — but that's a pricing decision, not one this PR should make.

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/warehouse_parent.py
danielcarletti: AI reply: Done in 15802fd — `MAX_PARENT_PAGE_SIZE = 5000` cap plus a floor of 1, applied inside the reader so every caller (REST fan-out config, Sentry's custom iterator, future non-REST adopters) is covered. Tests for both bounds.

products/warehouse_sources/backend/presentation/views/external_data_schema.py
```diff
@@ -542,6 +585,59 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+            return
+        source_impl = _registered_source_impl(source)
+        if source_impl is None:
+            return
+
+        if not should_sync:
+            _refuse_when_fanout_dependents(instance, source_impl, "disabled")
+            return
+
+        required_parents = source_impl.get_required_parent_schemas(instance.name)
```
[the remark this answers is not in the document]
danielcarletti: AI reply: Fixed in 625db77ee1c. Enabling a fan-out child now requires editor access to every required parent, resolved the same way WarehouseTableAccessPermission resolves schema access (parent's table, falling back to the source before first sync); service credentials keep passing. Refused with a 403 that names the parent. Covered by parameterized tests in test_external_data_schema_access_control.py (viewer on parent table blocks, editor allows).

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/warehouse_parent.py
```diff
@@ -0,0 +1,184 @@
...
+    storage_options = delta_storage_options()
+    if not deltalake.DeltaTable.is_deltatable(uri, storage_options=storage_options):
+        raise WarehouseParentTableNotFoundError(
+            f"Parent schema '{parent_name}' has no synced table yet — complete its initial sync first"
+        )
+    delta_table = deltalake.DeltaTable(uri, storage_options=storage_options)
+
+    as_of = _last_completed_snapshot_timestamp(team_id, parent_schema.id)
+    if as_of is not None:
+        delta_table.load_as_version(as_of)
```
estefaniarabadan: can we wrap this function's body so it only ever raises WarehouseParentTableNotFoundError while chaining the cause? 
we only catch the warehouseparent one and we can propagate other exceptions

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/fanout.py
```diff
@@ -104,6 +130,46 @@ def build_dependent_resource(
...
+        # Resolve the parent's table now, while we're in sync source-build context — the
+        # iterator itself runs later on executor threads where ad-hoc ORM reads are fragile.
+        # This also pins the Delta version the whole fan-out reads, and validates the columns
+        # while a fallback is still possible.
+        parent_columns = list(dict.fromkeys([fanout.resolve_field, *fanout.include_from_parent]))
+        try:
+            parent_table = resolve_parent_table_ref(
+                team_id, source_id, fanout.parent_name, required_columns=parent_columns
+            )
+        except WarehouseParentTableNotFoundError:
```
estefaniarabadan: or widen this instead of my comment on the resolve_parent_table_ref function

products/warehouse_sources/backend/temporal/data_imports/sources/sentry/sentry.py
estefaniarabadan, line 662: (claude-comment): The load at line 462 is guarded with not use_warehouse_parent, but this save isn't. The Redis key is scoped to job_id, not the attempt, and the gate is re-evaluated every activity attempt — so a flag flip mid-job (the documented rollback), a disabled parent, or a parent entering a sync with no completed job puts attempt N+1 on the API path, where it fast-forwards the sort=date issue list to a checkpoint chosen from an unordered scan. Every issue sorting earlier gets skipped and the watermark still advances. Can we either skip save_state in warehouse mode, or use resumable_source_manager.with_namespace("warehouse")

estefaniarabadan (approved):

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/warehouse_parent.py
danielcarletti: AI reply: Done in 3da897905ff, went with this option rather than widening the call sites. resolve_parent_table_ref's body is wrapped so it only ever raises WarehouseParentTableNotFoundError, chaining the original cause (the call sites log with exc_info, so the chain lands in the logs). Centralizing it here means the two current callers stay narrow and any future non-REST caller gets the total-fallback contract for free. Test: test_resolve_wraps_unexpected_errors_so_callers_can_fall_back (parameterized over an S3-style RuntimeError and a malformed-UUID ValueError, asserting __cause__).

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/fanout.py
danielcarletti: AI reply: Went with your first option (wrapping inside resolve_parent_table_ref, in 3da897905ff) instead of widening this except. Widening here would need the same widening in sentry.py and in every future caller, and a bare except Exception at call sites also risks eating errors from the iterator construction around it. Wrapping at the source keeps this except exact and makes the contract self-enforcing.

products/warehouse_sources/backend/temporal/data_imports/sources/sentry/sentry.py
danielcarletti: AI reply: Confirmed and fixed in 3da897905ff — this was a real data-loss path (warehouse attempt saves a checkpoint from the unordered scan; a fallback attempt fast-forwards the sort=date API list to it, skipping every newer issue while the watermark advances). Went with skipping the save: the checkpoint has zero value in warehouse mode (the load side already ignores it and reprocesses from the start, merge dedupes), so namespacing would persist state nothing ever reads. The values-page save at line ~638 was the only unguarded site — the flat-endpoint save_checkpoint path is never reached by fan-out children, and they pass no resume_hook to the shared builder. Test: test_warehouse_mode_never_saves_a_resume_checkpoint (warehouse run with a next-page link present, save_state never called).

PR #72939, opened 2026-07-22
feat(alerts): support selecting Slack workspaces for alerts

MattPua:
## Problem

Alert notifications always used the first Slack integration, so projects with multiple Slack workspaces could not choose where an alert should be sent.

## Changes

CleanShot 2026-07-22 at 14 12 25@2x

- Add workspace selection to the inline alert destination editor when multiple Slack integrations exist.
- Clear the selected channel when switching workspaces.
- Keep the alert-specific destination templates and filters intact.

- Added regression coverage for clearing the Slack channel when the workspace changes.
- Automated tests and type generation could not run because this checkout has no `node_modules`, `hogli`, or `flox`.
- `git diff --check` passed.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

No matching documentation describes this selector.

## 🤖 Agent context

Codex implemented this change using `/adopting-generated-api-types` and `/writing-tests`. The shared workflow-style picker could not replace alert destinations directly because alerts require their own HogFunction subtemplates and event filters, so the existing inline alert editor now reuses the workspace selection pattern instead.

---

### DIFF
--- a/products/alerts/frontend/views/InlineAlertNotifications.tsx
+++ b/products/alerts/frontend/views/InlineAlertNotifications.tsx
@@ -63,11 +63,6 @@ function getHogFunctionDestination(
     const inputs = hogFunction.inputs as SlackDestinationInputs | null | undefined
     const channelValue = inputs?.channel?.value
     if (channelValue) {
-        // `channel.value` comes from loosely-typed backend JSON written by several producers
-        // (inline alerts, error tracking alerts, logs alerting, MCP) — guard before splitting it.
-        if (typeof channelValue !== 'string') {
-            return { type: 'Slack', detail: null }
-        }
         // Destinations from before `slack_workspace` existed belonged to the single workspace
         // that was the only option at the time, so fall back to it for those legacy rows.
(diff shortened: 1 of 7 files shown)


products/alerts/frontend/views/InlineAlertNotifications.tsx
```diff
@@ -136,29 +136,23 @@ export function InlineAlertNotifications({ alertId }: InlineAlertNotificationsPr
         removePendingNotification,
         deleteExistingHogFunction,
         setSelectedType,
+        setSelectedSlackIntegrationId,
         setSlackChannelValue,
         setWebhookUrl,
     } = useActions(logic)
 
-    const slackLogic = slackIntegrationLogic({ id: firstSlackIntegration?.id ?? 0 })
+    const slackLogic = slackIntegrationLogic({ id: selectedSlackIntegration?.id ?? 0 })
```
Radu-Raicea: When an alert sends to multiple Slack workspaces, some destinations appear only as “Slack” without the channel name. Users then can’t tell where each notification goes and may delete the wrong destination. Can we display the workspace and channel for every Slack destination?

Radu-Raicea (approved):

PR #72943, opened 2026-07-22
feat(replay): support excluding sessions by event and action filters

arnohillen:
## Problem

Replay's filters (recordings list, saved filters, and Replay Vision scanner configs) can only select sessions that **did** contain a given event or action. There was no way to express "sessions where the user did not do X", for example scanning only the sessions where users failed to complete a desired action. Requested by the replay team in [this Slack thread](https://posthog.slack.com/archives/C0B7TE4R8GH/p1784732970968919).

Property filters with negative operators (`is not`, `doesn't contain`, ...) already get whole-session exclusion semantics through the negative blocklist in the recordings query, but whole events and actions could not be negated.

## Changes

Entity dicts inside `RecordingsQuery.events` / `actions` accept an optional `negation: true`. Backwards compatible, absent flag means today's behavior exactly.

**Backend** (`events_subquery.py` only)

- Entities are partitioned on the raw dicts (the legacy `Entity` class drops unknown keys), so the positive matching path never sees negated entities.
- Negated entities feed their positive predicate (event/action plus its nested property filters) into the existing negative blocklist, and the main query excludes those sessions with `NOT GlobalIn`.
- Exclusions apply under both Match all and Match any, always as AND NOT. Negative property operators keep their existing AND-only behavior.
- The player's "matching events" lookup skips negated entities, so highlights only come from positive filters.

**Frontend**

- Event and action chips in the replay universal filters get a "Performed / Did not perform" control in their popover, gated behind the new `replay-negative-event-filters` feature flag. Excluded chips render a "did not perform" prefix.
- Wired into the recordings list (through the shared filter embed) and into the Vision scanner trigger config, which renders `UniversalFilters.Value` directly and so needs the prop passed explicitly. The scanner's query round-trip preserves the flag (covered by tests).

Before:

```mermaid
flowchart LR
    E[events/actions] --> P[positive predicates]
    P --> IN[session_id GlobalIn subquery]
    NP[negative property filters] --> BL[blocklist subquery]
    BL --> OUT[session_id NOT GlobalIn]
    class E,NP phYellow;
    class P,BL phBlue;
    class IN,OUT phGray;
```

After:

```mermaid
flowchart LR
    E[events/actions] --> SPLIT{negation?}
    SPLIT -- no --> P[positive predicates]
    P --> IN[session_id GlobalIn subquery]
    SPLIT -- yes --> BL[blocklist subquery]
    NP[negative property filters] --> BL
    BL --> OUT[session_id NOT GlobalIn]
    class E,NP phYellow;
    class P,BL,SPLIT phBlue;
    class IN,OUT phGray;
```

> [!NOTE]
> Backend support ships unconditional (inert unless an entity carries the flag). Only the UI toggle is feature flagged. The `replay-negative-event-filters` flag still needs creating before the toggle appears anywhere.

Verified end to end in a local dev stack running this branch, plus automated tests.

**Manual verification (local, flag created and enabled)**

Seeded a throwaway project with six recordings from one user: three containing a `purchase` event, three without. Same `purchase` filter, toggled.

Before, on master, the event chip popover has no Performed / Did not perform control:

before-toggle-popover (master merge-base, no include/exclude toggle)

Performed (today's only option) returns the three sessions that purchased:

after-toggle-popover

Did not perform returns the three that did not, and the chip renders the `did not perform` prefix:

after-excluded-list

The two result sets are disjoint and sum to the six seeded sessions (purchased sessions show 12 clicks / 8 keypresses, abandoned show 7 / 3). Toggling writes `"negation":true` into the URL filter state.

Running the real `SessionRecordingListFromQuery` against that same seeded ClickHouse data:

| Query | Result |
| --- | --- |
| positive `$pageview` only | 6, all sessions |
| `$pageview` AND NOT `purchase` | 3, the abandoned ones |
| NOT `purchase` only | 3, the abandoned ones |
| positive `purchase` | 3, the purchased ones |

Exclusion-only does not degenerate to returning everything, and negation composes with a positive filter.

Backend, in `test_session_recording_list_from_query.py`, each catching a regression no existing test covers (the `negation` field is new):

- `test_negated_event_filter_excludes_sessions_containing_event` (parameterized AND/OR): catches the negated entity being matched positively, the OR guard silently dropping exclusions, and the exclusion-only query returning everything.
- `test_negated_event_filter_respects_entity_property_filters`: catches the blocklist matching on event name alone and over-excluding.
- `test_negated_action_filter_excludes_sessions_containing_action`: catches the partition only being applied to events.
- `test_negated_event_filter_does_not_affect_matching_events`: catches the player's matching-events query AND-ing the negated entity and returning no highlights.

Frontend:

- `recordingsQueryConversions.test.ts`: negation survives query -> universal filters -> query round trips (the Vision scanner persists queries through exactly this path).
- `UniversalFilters.negation.test.tsx`: the toggle writes `negation` onto the filter value, and the control does not render without the opt-in prop (keeps it out of non-replay surfaces whose backends ignore the flag).

Full gates run locally: session recordings queries suite (322 passed, 348 ClickHouse query snapshots unchanged, so zero impact on existing query generation), repo-wide mypy, frontend `typescript:check`, and the affected jest suites.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Replay filter product docs live on posthog.com and should be updated when the feature flag rolls out, not in this repo.

## 🤖 Agent context

Built with Claude Code (Fable 5), directed by @arnohillen. Skills invoked: superpowers brainstorming, writing-plans, and executing-plans for the design/plan/implementation flow, plus the repo's /writing-tests gate before authoring tests.

Design decisions made with the driver: a per-chip toggle instead of a separate exclusions section, the `negation` flag inline on the entity instead of new `exclude_events`/`exclude_actions` query fields (rejected for round-trip complexity in scanner configs), and exclusions always applying as AND NOT under both operands. Backend support is intentionally unconditional so stored queries evaluate deterministically, with only the UI entry point flag-gated.


posthog/session_recordings/queries/sub_queries/events_subquery.py
```diff
@@ -618,21 +618,42 @@ def countif_zero(prop: AnyPropertyFilter) -> ast.Expr:
...
 
     @property
     def event_entities(self):
         # TODO what do we send to the API instead to avoid needing to do this
         # TODO is this overkill since it feels like we only need a few things off the entity
-        return [legacy_entity_to_node(Entity(e), True, MathAvailability.Unavailable) for e in self._query.events or []]
+        return [
+            legacy_entity_to_node(Entity(e), True, MathAvailability.Unavailable)
+            for e in self._query.events or []
+            if not self._is_negated_entity(e)
```
TueHaulund, line 713: Filtering negated entities out of `action_entities`/`event_entities` breaks the `matching_events` path for an exclusion-only filter set.

If every entity is negated and there are no event/person/group/cohort property filters, `_get_queries_for_matching` yields zero queries, so `wrapped_with_query_operand([])` builds `ast.And(exprs=[])` and the printer emits `and()` with no arguments, which ClickHouse rejects (the printer only short-circuits at `len == 1`).

It is reachable from the UI: `recordings_query_has_event_filters` in `session_recordings/utils.py` returns True for a negated-only entity list, so the precondition in `session_recording_api.py` passes. Repro: filter "did not perform purchase", then open a recording in the player.

An exclusion-only query has nothing to highlight, so short-circuiting to a `WHERE false` select seems right. Worth a regression test with no positive entity, since `test_negated_event_filter_does_not_affect_matching_events` includes a positive `$pageview` and so misses this.

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
```diff
@@ -101,15 +109,40 @@ const Value = ({
...
+                            'data-attr': 'universal-filters-entity-negation-exclude',
+                        },
+                    ]}
+                />
+            </div>
+        ) : null
+
     const pageKey = `${rootKey}.filter_${index}`
 
     const handleChangeEvent = (
```
TueHaulund, line 152: `handleChangeEvent` builds a fresh `{ id, name, type, properties: [] }` rather than spreading `filter`, so `negation` is dropped: configure a chip as "did not perform purchase", hit "Change event", and it silently comes back as a positive filter.

Spread the existing filter, or carry `negation` across explicitly if resetting the rest is intentional.

posthog/session_recordings/queries/sub_queries/events_subquery.py
```diff
@@ -684,29 +705,34 @@ def _negative_blocklist_query(self) -> ast.SelectQuery | None:
...
-        where_expr = ast.Or(exprs=inverted_exprs) if len(inverted_exprs) > 1 else inverted_exprs[0]
+        where_expr = ast.Or(exprs=blocklist_exprs) if len(blocklist_exprs) > 1 else blocklist_exprs[0]
 
         return ast.SelectQuery(
             select=[ast.Alias(alias="session_id", expr=_event_session_id_field())],
             select_from=ast.JoinExpr(table=ast.Field(chain=["events"])),
             where=self._where_predicates(where_expr),
             group_by=[_event_session_id_field()],
+            # Negating a very common event (e.g. $pageview) can push the blocklist past this cap on
+            # high-traffic teams, silently under-excluding sessions beyond it. Typical negatives (rare
```
TueHaulund, line 806: This cap turns into silently incorrect results rather than a bounded one. Past 1,000,000 sessions the `GLOBAL NOT IN` blocklist is incomplete, so sessions that did perform the excluded event still come back. For `$pageview` or `$autocapture` on a high-traffic team that is the common case, not the tail, and the user just sees "the exclusion filter doesn't work" with no signal on our side.

The comment acknowledges the truncation but nothing observes it. Minimum: a counter or span attribute when the blocklist returns exactly `limit` rows. Better: bound the query differently or fail loudly instead of returning wrong sessions.

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
```diff
@@ -101,15 +109,40 @@ const Value = ({
...
     const isEvent = isEventFilter(filter)
+    const isAction = isActionFilter(filter)
     const isEditable = isEditableFilter(filter)
 
     const [open, setOpen] = useState<boolean>(isEditable && initiallyOpen)
     const [changingEvent, setChangingEvent] = useState<boolean>(false)
 
+    const negationSelect =
+        allowEntityNegation && (isEvent || isAction) ? (
+            <div className="px-2 py-1" data-attr="universal-filters-entity-negation">
```
TueHaulund, line 129: The flag gates only this control. `UniversalFilterButton` renders the "did not perform" prefix unconditionally and the backend honors `negation` either way, so a playlist saved while the flag is on keeps excluding after a rollback, with no UI to turn it off short of deleting the chip.

Intentional? If so it is worth a note, since it means the flag is not really a kill switch.

frontend/src/lib/components/UniversalFilters/UniversalFilters.negation.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    it('writes negation onto the filter when toggled', async () => {
+        const onChange = jest.fn()
+        const container = renderValue(true, onChange)
+        await openChipPopover(container)
+        await userEvent.click(screen.getByText('Did not perform'))
+        expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ negation: true }))
+    })
+
+    // LemonSegmentedButton only accepts data-attr per option, so a top-level one is silently dropped.
+    it('exposes the control and its options as data-attr selectors', async () => {
```
TueHaulund: These `data-attr` values have no consumer anywhere in `frontend/` or `products/`, so this pins `LemonSegmentedButton` internals and will fail on any rename without a real regression behind it. Suggest dropping it, and collapsing the other two cases into a `test.each` over `allowEntityNegation`.

posthog/session_recordings/queries/sub_queries/events_subquery.py
```diff
@@ -618,21 +618,42 @@ def countif_zero(prop: AnyPropertyFilter) -> ast.Expr:
         exprs = [countif_zero(p) for p in self._collect_negative_properties()]
         return self.wrapped_with_query_operand(exprs=exprs) if exprs else ast.Constant(value=True)
 
+    @staticmethod
+    def _is_negated_entity(raw_entity: dict) -> bool:
```
TueHaulund: Nit: `dict` should be `dict[str, Any]`, since new code is expected to pass mypy strict.

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
```diff
@@ -101,15 +109,40 @@ const Value = ({
...
+                <LemonSegmentedButton
+                    size="xsmall"
+                    value={(filter as ActionFilter).negation ? 'exclude' : 'include'}
```
TueHaulund: Nit: this cast can use the existing `isEntityFilter` narrow instead, which keeps the type check honest if `filter` widens later.

TueHaulund (changes requested):
Nice approach, reusing the existing session blocklist means OR support comes almost for free and `session_recording_list_from_query.py` stays untouched.

Two things to fix before this ships:

1. An exclusion-only filter set (every entity negated, no property filters) makes the `matching_events` query emit `and()` with zero arguments, which ClickHouse rejects.
2. "Change event" drops `negation`.

One more worth a decision: the 1M blocklist cap silently under-excludes for common events. Codex flagged that independently as a correctness bug, not just a missing metric.

Rest is minor. Also worth calling out somewhere users can see it: under `Match any`, a negated entity still applies as a hard `AND NOT`, so "did not perform X" OR "performed Y" drops sessions that did both. Right now that only lives in the docstring.

posthog/session_recordings/queries/sub_queries/events_subquery.py
arnohillen: [arno's agent] Fixed in b46781a0: the matching-events query short-circuits to WHERE false when every entity is negated, with an end-to-end ClickHouse regression test for the negated-only case.

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
arnohillen: [arno's agent] Fixed in b46781a0: handleChangeEvent carries negation through both branches (properties still reset intentionally), with a jest case asserting an excluded chip stays excluded after changing the event.

posthog/session_recordings/queries/sub_queries/events_subquery.py
arnohillen: [arno's agent] Went observe-first rather than failing the query: when negated entities are present we probe the blocklist count after listing and record a span attribute plus a replay_negative_blocklist_truncated counter at the 1M cap, so we can decide on a hard error once we see how often it fires (costs one extra ClickHouse count per listing with negations).

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
arnohillen: [arno's agent] Intentional: stored playlists and scanner queries must evaluate deterministically regardless of flag state, and a rollback that silently flipped 'did not perform X' back to 'performed X' would corrupt saved filters; added a code note that the flag gates the entry point only.

frontend/src/lib/components/UniversalFilters/UniversalFilters.negation.test.tsx
arnohillen: [arno's agent] Dropped the data-attr test and collapsed presence/absence into a test.each over allowEntityNegation, keeping the onChange assertion in the true case.

posthog/session_recordings/queries/sub_queries/events_subquery.py
arnohillen: [arno's agent] Done, annotated as dict[str, Any].

frontend/src/lib/components/UniversalFilters/UniversalFilters.tsx
arnohillen: [arno's agent] Done: the gate uses isEntityFilter(filter) so TypeScript narrows and both casts are gone.

TueHaulund (approved):
LGTM

PR #72951, opened 2026-07-22
feat(mcp): surface governed metrics in exec search output

thiagosalvatore:
## Problem

In single-exec mode, agents discover capabilities with `exec search "<terms>"` — but it matched tool metadata only, never catalog rows. A metric question ("what's our ARR?") therefore surfaced no governed metrics, and the agent fell back to searching insights or deriving with SQL. Prompt steering alone can be skipped; putting the signal in **tool output** cannot.

This is PR 2 of a 3-PR stack (builds on the prompt fixes in the base PR).

## Changes

- `exec search` now returns matching governed metrics alongside tool matches, with a ready-to-run hint pointing at `data-catalog-metric-run`. New `metric-search.ts` reuses the existing field-weighted tool ranking (metric `display_name` → `title`); proposed/drifted rows carry their `status`/`is_drifted` so trust rules still apply.
- Cached, fail-soft fetch on `StateManager` (`getOrFetchCatalogMetrics`, 10-min TTL, stale-on-error) via a lean `getCatalogMetrics` client method; a metrics-API error can never break tool search.
- Gated in `resolveExecTool`: the provider is passed only when `data-catalog-metric-run` is registered, so flag-off orgs get **byte-identical** search output and zero added latency.
- Extended the eval scorers to credit an `exec search` whose output surfaced `governed_metrics` (so the catalog-lookup ordering scorers still pass when discovery happens via search, not SQL); added the `governed_metric_search_first` eval.
- Inverted the now-stale "search matches tool metadata only" / "finds tools, not catalog rows" prose.

Automated (run by the agent):
- `pnpm --filter=@posthog/mcp test` — full unit suite green.
- New `metric-search.test.ts` (ranking, trust-field passthrough, cap) and `exec.ts` search-integration tests (metric surfacing with run hint, no-clobber alongside tool matches, fail-soft on provider rejection, flag-off byte-identical, no false positives).
- New Python scorer test (`test_eval_scorers.py`) for the search-surfaced-metrics detection (6 cases incl. malformed-JSON fallback, errored search, non-search command).
- `typescript:check` + `pnpm fix` clean. Sandboxed evals **not** run here (Docker/Modal + credentials required).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None (MCP internal behavior).

## 🤖 Agent context

Claude Code (Opus 4.8), from the approved plan. Skills: `/writing-tests`, `/writing-evals`.

Decisions:
- When a governed metric matches but an existing hint is present (scope-gated / truncated results), the run-hint moves to a distinct `governed_metrics_hint` key so neither clobbers the other; the metric-only case (query matches no tool) leads with the run-hint instead of "no tools matched".
- 10-min stale cache is acceptable because `data-catalog-metric-run` re-verifies status/drift server-side at run time.

Note: stack is behind `origin/master` (OpenAPI-drift advisory) — will be cleared by a restack/merge before ready.

### DIFF
--- a/products/data_catalog/backend/tests/test_eval_scorers.py
+++ b/products/data_catalog/backend/tests/test_eval_scorers.py
@@ -1,0 +1,42 @@
+from typing import Any
+
+from parameterized import parameterized
+
+from products.data_catalog.evals.scorers import _search_surfaced_metrics
+from products.posthog_ai.eval_harness.log_parser import ToolCall
+
+_METRIC_OUTPUT = '{"matches":[],"governed_metrics":[{"name":"monthly_recurring_revenue"}],"hint":"run it"}'
+
+
+def _tool_call(**overrides: Any) -> ToolCall:
+    defaults: dict[str, Any] = {
+        "name": "search",
+        "input": {},
+        "output": "",
+        "is_error": False,
+        "call_id": "call-1",
+        "position": 0,
+        "raw_name": "exec",
+        "is_exec_unwrapped": True,
+    }
+    defaults.update(overrides)
+    return ToolCall(**defaults)
+
+
+class TestSearchSurfacedMetrics:
+    @parameterized.expand(
+        [
+            ("unwrapped_search_with_metrics", _tool_call(name="search", output=_METRIC_OUTPUT), True),
+            (
+                "wrapped_exec_search_with_metrics",
+                _tool_call(name="exec", input={"command": "search revenue"}, output=_METRIC_OUTPUT),
+                True,
+            ),
+            ("malformed_json_substring_fallback", _tool_call(output='garbage "governed_metrics": [x]'), True),
+            ("search_without_metrics", _tool_call(name="search", output='{"matches":["x"]}'), False),
+            ("errored_search", _tool_call(name="search", output=_METRIC_OUTPUT, is_error=True), False),
+            ("non_search_command", _tool_call(name="info", output=_METRIC_OUTPUT), False),
+        ]
+    )
+    def test_detects_search_that_surfaced_governed_metrics(self, _name: str, call: ToolCall, expected: bool) -> None:
+        assert _search_surfaced_metrics(call) is expected
--- a/products/data_catalog/evals/eval_governed_metrics.py
+++ b/products/data_catalog/evals/eval_governed_metrics.py
@@ -105,6 +105,29 @@ async def eval_governed_metrics(ctx: EvalContext) -> None:
             },
             setup=seed_approved_metric,
         ),
+        # `exec search` now returns matching governed metrics in its output, so the agent can
+        # discover the metric from the search result — no execute-sql catalog query needed —
+        # and run it. Credits the search-surfaced catalog lookup (scorers.py) then the run.
+        SandboxedEvalCase(
+            name="governed_metric_search_first",
+            prompt="Search for our revenue metric and give me the current value.",
+            expected={
+                "metrics_catalog_queried": {},
+                "canonical_metric_run": {
+                    "metric_name": APPROVED_METRIC_NAME,
+                    "outcome": "succeeded",
+                },
+                "governed_behavior_correctness": {
+                    "expected_behavior": (
+                        f"Discovered the approved metric '{APPROVED_METRIC_NAME}' via a catalog search (either the "
+                        "governed_metrics returned by exec search or a system.information_schema.metrics query), ran "
+                        "it through data-catalog-metric-run rather than re-deriving revenue, rechecked the runner's "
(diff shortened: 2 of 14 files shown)


thiagosalvatore:
* **#72953** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#72951** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/72951?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#72950** <a href="https://app.graphite.com/github/pr/PostHog/posthog/72950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

thiagosalvatore:
## Review notes

**What it does:** `exec search` now also returns matching governed-catalog metrics alongside tool matches, with a hint routing the agent to `data-catalog-metric-run`. Gated to orgs where the catalog tool is registered, fetched via a 10-min fail-soft cache, and the eval scorers are extended to credit discovery-via-search. The implementation matches the stated intent, and the doc/prose changes ship in the same PR.

**Overall: approve-with-comments.** Careful, well-scoped change. It faithfully reuses the existing `getGroupTypes` / `getOrFetchGroupTypes` and tool-ranking patterns, the gating genuinely gives flag-off orgs byte-identical output, and the fail-soft handling is correct on every return path I traced. No correctness bugs found. One test gap plus a few nitpicks.

### Should fix

- **The `governed_metrics_hint` coexistence branch is untested** — `services/mcp/src/tools/exec.ts` (`withGovernedMetrics` helper). The "an existing hint stays primary and the run-hint moves to a distinct `governed_metrics_hint` key so neither clobbers the other" behavior is only reachable from the **scope-gated** and **truncated** search paths. Every test in the new `governed metrics in search` block (`services/mcp/tests/unit/exec.test.ts`) goes through `createExec`, which passes `scopeGatedTools = []`, and none pushes past the 25-result truncation threshold — so the helper's `base.hint ? …` true-branch never runs. The core metric-surfacing paths (no-tool-match, alongside-tools, fail-soft, flag-off) are all well covered; it's specifically the hint-preservation logic that a future refactor could silently break. A two-case test (a gated tool + a matching metric; a >25-match query + a matching metric) would close it.

### Nitpicks / minor

- **Silent 100-metric cap.** `getCatalogMetrics` fetches `?limit=100` (`services/mcp/src/api/client.ts`) and the viewset orders `-created_at` (`products/data_catalog/backend/presentation/views.py:55`), so search only ever sees the 100 newest metrics. This is fine — the prose now frames the SQL query as "the exhaustive lookup" — but a one-line comment on the `?limit=100` noting the cap is intentional (and SQL is the escape hatch) would save the next reader a lookup.

- **`exec search` now does inline network I/O on cache-miss.** For flag-on orgs a stale cache makes search await a metrics fetch (drift computation included) before returning. It's fail-soft on *error* but there's no timeout, so a *slow* (not failing) metrics API would stall tool discovery. This matches how every other cached fetch in the client behaves (no `AbortController` anywhere), so it's consistent rather than a new risk — flagging only so it's a conscious choice.

- **Mild DRY smell in the search return paths.** The `matches.length === 0` and final `governedMetrics.length > 0` paths hand-build the `{ matches, governed_metrics, hint }` object inline, which is exactly what `withGovernedMetrics`'s no-hint branch produces — three copies of the same shape. Could collapse to `if (governedMetrics.length > 0) return withGovernedMetrics({ matches })`, keeping the bare-array default separate. Purely cosmetic.

- **Null-safety asymmetry in `metric-search.ts`.** The ranking input guards `display_name ?? ''` / `description ?? ''`, but the returned object passes `m.display_name` / `m.description` straight through. The serializer makes these non-null in practice, so it's theoretical — just inconsistent with the guard right above it.

### Open question

- **Regex queries surface tools but not metrics.** Tool search routes queries containing regex metacharacters (`-`, `.`, `|`, …) to `searchToolsRegex`, but `searchCatalogMetrics` always uses the whitespace-token ranker. So a hyphenated query like `search monthly-recurring-revenue` regex-matches tools but tokenizes to a single string that won't substring-match the underscore metric name — the metric silently doesn't surface. For plain-word queries (the documented "search with terms/synonyms" flow) both paths agree, so real-world impact looks low. Given metric discovery is the whole point, is this recall gap worth handling (e.g. splitting the query on `-`/whitespace for the metric ranker), or is "SQL is the exhaustive fallback" good enough?

PR #72956, opened 2026-07-22
feat(aio): add settling strategies for trace evaluations

carlos-marchal-ph:
## Problem

Trace evaluations always fire on a fixed delay after the first matching generation.
Long agent runs get graded mid-flight, short ones wait 30 minutes for nothing.
This is not a big issue with traces, since they are usually temporally co-located in a predictable way.
This is a deal-breaker for sessions, since it's unpredictable when a user will start a new turn.

## Changes

Adds a settling strategy to the trace target. `target_config` is now discriminated on `strategy`:

- `fixed_window` – existing behavior, still the default. Configs without a `strategy` key mean this, no migration needed.
- `inactivity` – evaluate once the trace had no new activity for a quiet period (10s–30m, default 5m), capped by `max_age_seconds` (1m–2h, default 2h).

Under the hood:

- New `run-aggregate-evaluation` temporal workflow. Both strategies dispatch it with a plain `start` (`workflowIdConflictPolicy: USE_EXISTING`, `workflowIdReusePolicy: ALLOW_DUPLICATE_FAILED_ONLY`), same at-most-once dedup per (evaluation, trace) as before, including across the deploy boundary. No signals anywhere in this path, so ingestion volume can never grow past one dedup'd start per (evaluation, trace).
- Inactivity settling is ClickHouse polling, not a Temporal signal loop. One `check_trace_settled_activity` checks `ai_events` for the trace's latest arrival-time (`_timestamp`, not event time) among its structural events (`$ai_generation`/`$ai_span`/`$ai_embedding`/`$ai_trace`), with a 15s ingestion-lag margin, and keeps polling on a NULL result rather than manufacturing a false "settled" verdict out of a lag spike. The activity's own retry policy is the poll loop: fixed interval `max(quiet_period_seconds / 4, 10)`s, bounded by a `schedule_to_close_timeout` equal to `max_age_seconds`.
- Old `run-trace-evaluation` stays registered to drain in-flight runs (bounded ~2h), then gets removed in a follow-up. Its post-settle body is deliberately duplicated in the new workflow until then.
- Strategy picker in the eval UI behind `llm-analytics-eval-settling-strategy`. Flag off renders exactly today's UI, including for an eval a flag-on teammate already switched to inactivity: an `effectiveStrategy` fallback forces the fixed-window display and reseeds the stored config (clearing stale inactivity keys) if a flag-off user edits the duration field.
- `creating-online-evaluations` skill docs updated, api/mcp types regenerated with an OpenAPI `discriminator` on `target_config`. Settle-liveness copy now says "no new activity" instead of "no new matching generation", since spans and embeddings re-arm the quiet timer too, not just generations. The condition-trigger wording ("first matching generation") is unrelated and unchanged.

Review round also fixed three issues:

- A non-dict `target_config` (list, string, number) crashed with an unhandled 500 instead of a 400. The dict-spread that builds the settle config had moved outside the `try`/`except` boundary; moved it back inside, with a parameterized regression test per input shape.
- Added the OpenAPI `discriminator` on `target_config`'s `oneOf`, but confirmed it doesn't change codegen for either downstream consumer yet: Orval still emits a plain `zod.union` in `api.zod.ts`, and the MCP tool-schema builder has no `discriminator` support at all. So an MCP `llma-evaluation-create`/`-update` call missing `strategy` can still silently match the wrong branch. Documented as a known follow-up, not fixed here.
- Flag-off UI leak: a flag-off user opening a team eval already configured as inactivity saw the new quiet-period/max-age fields with no picker to switch back. Fixed with the `effectiveStrategy` fallback described above.

> [!NOTE]
> Settling moved from Temporal signals to ClickHouse polling. Every matching generation used to append a signal to the workflow history, which on a degenerate shared trace id (e.g. "0") could grow a history large enough to hit Temporal's 50k-event cap within the max age window. That risk is gone: there's nothing left to signal, and dispatch is a single dedup'd `start` per (evaluation, trace) regardless of how many generations, spans, or embeddings show up afterward. What to watch during rollout instead: `ai_evaluation_settle_polls_total{outcome}`, specifically aggregate poll QPS against the `ai_events` cluster (bounded by the fixed `max(quiet/4, 10)`s interval per in-flight inactivity eval, but worth confirming at scale).

> [!NOTE]
> The `target_config` discriminator doesn't fix the MCP misrouting risk it was meant to close. Neither Orval nor the MCP tool-schema builder honor `discriminator` for this inline-`oneOf` shape today, so it stays a documented-intent-only change. A real fix needs either `$ref`-based `oneOf` arms (so Orval's existing discriminator detection applies, same as it already does for `widget_type`) or a manual zod override in `services/mcp`.

- Python: settle config validation (bounds, legacy defaulting, cross-field check, unknown keys, non-dict `target_config` now parameterized to assert 400s), workflow behavior on the time-skipping test server (fixed window sleeps and skips the poll activity entirely, inactivity settles once the poll activity reports quiet, a couple of failed polls before it settles, max age evaluates anyway once the poll budget runs out on a trace that never goes quiet), serializer contract tests.
- Node: plain-`start` wire shape for both strategies (`USE_EXISTING` + `ALLOW_DUPLICATE_FAILED_ONLY`, no signal args anywhere), already-completed dedup, legacy config default resolution through real batch dispatch.
- Jest: strategy seeding, switching, and patching in the kea logic, plus the flag-off `effectiveStrategy` reseed sequence (switching to inactivity then back to fixed-window leaves no stale inactivity keys).
- Manual: not re-run since switching the settle mechanism from signals to polling. The earlier manual pass (inactivity eval firing after quiet period, flag-off UI unchanged) was against the signal-based version and isn't representative of the current poll-based path anymore. Worth a fresh manual check against real `ai_events` data before merge.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The `creating-online-evaluations` skill (SKILL.md + payload reference) is updated in this PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) – Carlos is the DRI.

Built with Claude Code (Fable 5): spec, plan, then per-task subagents with a reviewer gate per task, a whole-branch review, and finally two zero-context reviews of the finished branch. Skills invoked: /improving-drf-endpoints, /writing-tests, /writing-skills, /pr-review (for the zero-context passes).

Decisions worth knowing while reviewing: a new workflow name instead of patching `run-trace-evaluation` in place (zero replay risk for in-flight runs, at the cost of a temporarily duplicated body that leaves with the old file), settle config snapshotted into workflow args at schedule time (editing an eval never changes runs in flight), "activity" for triggering a run means condition-matching generations only (anything broader needs scheduler-side state) while "activity" for settling now means any structural event on the trace. The original design signaled the workflow per matching generation and accepted the resulting history-growth risk with monitoring; automated review flagged that as exploitable by ingestion volume, so it's replaced here with ClickHouse polling behind a single dedup'd `start` – no signals, no growth risk, a new counter to watch instead. The same review round also caught the non-dict `target_config` 500, confirmed the OpenAPI discriminator doesn't change MCP/Orval codegen for this shape, and caught the flag-off UI leak – all fixed, see Changes above.

> [!NOTE]
> Rollout posture: the feature flag gates only the strategy picker UI. Already-saved inactivity configs keep settling by polling regardless of the flag, so the poll interval floor is the mitigation knob, not the flag. Environments that deployed the earlier signal-based build of this branch (none in production, since it never reached master) will see expected, self-limiting non-determinism failures for runs that were in flight during the upgrade.


nodejs/src/ai-observability/services/temporal.service.ts
```diff
@@ -211,51 +228,53 @@ export class TemporalService {
...
-        windowSeconds: number
+        settle: ResolvedSettleConfig
     ): Promise<WorkflowHandle | null> {
         const client = await this.ensureConnected()
 
         const workflowId = `llma-trace-eval-${evaluationId}-${workflowSafeTraceId(traceId)}`
 
         try {
-            const handle = await client.workflow.start('run-trace-evaluation', {
+            const handle = await client.workflow.signalWithStart('run-aggregate-evaluation', {
```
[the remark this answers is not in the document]
Radu-Raicea: Agreed

Radu-Raicea (approved):
Approving, with 1 comment to fix, but I added the ReviewHog label too, so it might find other things.

products/ai_observability/backend/api/evaluations.py
```diff
@@ -122,21 +128,63 @@ class _OutputConfigField(serializers.JSONField):
...
+                        ),
+                        "minimum": TRACE_EVAL_MIN_MAX_AGE_SECONDS,
+                        "maximum": TRACE_EVAL_MAX_MAX_AGE_SECONDS,
+                        "default": TRACE_EVAL_DEFAULT_MAX_AGE_SECONDS,
+                    },
+                },
+                "additionalProperties": False,
+            },
+        ],
+        "discriminator": {"propertyName": "strategy"},
```
Radu-Raicea, line 189: The inline `discriminator` still produces a non-strict MCP union, so inactivity fields without `strategy` are silently replaced by fixed-window defaults. Could we use $ref arms or an MCP override so this fails validation?

posthog/temporal/ai_observability/run_aggregate_evaluation.py
```diff
@@ -0,0 +1,358 @@
...
+                        ),
+                        start_to_close_timeout=timedelta(seconds=30),
+                        schedule_to_close_timeout=timedelta(seconds=poll_budget_seconds),
+                        retry_policy=RetryPolicy(
+                            initial_interval=timedelta(seconds=poll_interval),
+                            backoff_coefficient=1.0,
+                            maximum_attempts=0,
+                        ),
+                    )
+                except temporalio.exceptions.ActivityError as e:
```
Radu-Raicea, line 249: Temporal can stop retrying before the 10-minute deadline when another check would run past it. We then grade a trace that is still active. Could we wait out the remaining time before evaluating?

Radu-Raicea (approved):

PR #72961, opened 2026-07-22
feat(mcp): add property-definition-update tool

sufi-ph:
## Problem

The PostHog MCP already lets agents update **event** definition metadata via `event-definition-update`, but there was no equivalent for **property** definitions — the only property tools were read-only. Users who wanted to curate their property taxonomy (descriptions, types, verified/hidden status) over MCP had no way to do it. This was raised in a support thread after a customer successfully used `event-definition-update` and asked why properties couldn't be updated the same way.

## Changes

Adds a `property-definition-update` MCP tool that mirrors `event-definition-update`:

- Updates `description`, `tags`, `property_type` (DateTime / String / Numeric / Boolean / Duration), and `verified` / `hidden` status.
- Takes an exact property name plus a `type` (event / person / group / session, defaulting to event) to target the right property, with `groupTypeIndex` for group properties.
- Property definitions have no `by_name` lookup endpoint, so the client resolves the id via the list endpoint (exact-name `properties` filter + `type`) before issuing the PATCH.

New handler in `src/tools/projects/updatePropertyDefinition.ts`, an `updatePropertyDefinition` method on the API client, input schemas in `tool-inputs.ts`, registration in `index.ts`, and the tool definition (scope `property_definition:write`). Regenerated `schema/tool-inputs.json`.

Added integration tests in `tests/tools/projects.integration.test.ts` covering description, tags, verified, multi-field updates, and the not-found error path — parallel to the existing event-definition-update tests.

I (the agent) was not able to run the MCP toolchain (typecheck / lint / integration tests) in this environment because monorepo dependencies were not installed; CI will exercise them. The regenerated `tool-inputs.json` was produced with the pinned zod version and verified to add only the two new schema entries with no churn to existing ones.

## 🤖 Agent context

Authored by the PostHog Slack app agent from a Slack thread. The work followed the existing `event-definition-update` tool as a template end to end (handler, client method, schema, tool definition, tests). Backend behavior was confirmed against `posthog/taxonomy/property_definition_api.py` and the EE serializer to pick the id-resolution approach (`properties` exact-name filter + `type`) and the writable field set. `tool-inputs.json` was regenerated and hand-verified to contain only additive changes.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BALLBBKAB/p1784740653682329)*

### DIFF
--- a/frontend/src/lib/agentScopes.generated.ts
+++ b/frontend/src/lib/agentScopes.generated.ts
@@ -95,6 +95,7 @@ export const AGENT_USE_CASE_SCOPES = [
     'project:read',
     'project:write',
     'property_definition:read',
+    'property_definition:write',
     'query:read',
     'replay_scanner:read',
     'replay_scanner:write',
--- a/services/mcp/schema/tool-definitions-all.json
+++ b/services/mcp/schema/tool-definitions-all.json
@@ -7355,6 +7355,20 @@
             "readOnlyHint": true
         }
     },
+    "property-definition-update": {
+        "description": "Update property definition metadata. Can update description, tags, property type (DateTime, String, Numeric, Boolean, Duration), and mark status as verified or hidden. Use the exact property name like '$browser' or 'plan_type'. Pass 'type' to target event, person, group, or session properties (defaults to event); for group properties also pass 'groupTypeIndex'.",
+        "category": "Events & properties",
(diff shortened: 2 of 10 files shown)


skoob13 (approved):
Thank you!

PR #72974, opened 2026-07-22
fix(toolbar): allow toolbar auth during impersonation sessions

bs1180:
## Problem

PostHog staff can't use the toolbar while impersonating a user. Clicking authenticate fails with `[PostHog Toolbar][auth] Token exchange failed` in the console, and the page just refreshes back to its unauthenticated state. This worked under the old toolbar auth mechanism (a directly-minted `temporaryToken`) but broke after the move to OAuth. Raised in a support thread by staff hitting it during impersonation.

The root cause is a mismatch between backend and frontend. Impersonation-minted OAuth tokens are deliberately refresh-less: the backend caps their lifetime to the impersonation idle timeout and omits the `refresh_token` so a token can't outlive the admin's impersonation session. But the toolbar's token exchange treated a response without a `refresh_token` as a failure, even on a 200. So the exchange succeeds server-side, the toolbar throws it away, and the session never authenticates.

## Changes

Made the refresh token optional across the toolbar auth flow, so a refresh-less (impersonation) session works while a normal session keeps rotating tokens exactly as before:

- Token exchange now succeeds on an access token alone. A `has_refresh_token` flag is added to the success telemetry so we can tell the two paths apart.
- `restoreOAuthTokens` accepts a stored access token with no refresh token (still rejects malformed shapes).
- On a 401 with no refresh token, the toolbar can't refresh, so it now prompts re-authentication instead of silently returning the failed response.
- `setOAuthTokens` and the related types accept `refreshToken: string | null`.

This keeps the security property intact: impersonation tokens stay short-lived and refresh-less, so when the access token expires the user simply re-authenticates.

> [!NOTE]
> This does not touch the separate `_impersonation_ai_processing_block` gate on the OAuth authorize step, which can 403 an impersonated authorize for orgs that haven't approved AI data processing. Users hitting *this* bug got past authorize (they reached token exchange), so that gate isn't what they were hitting.

Added three Jest cases to `toolbarConfigLogic.test.ts`, each guarding a distinct branch of the change (no existing test exercised a refresh-less token):

- token exchange with an access token and no refresh token authenticates (`refreshToken` ends up null, `isAuthenticated` true) — the exact regression reported.
- a stored access token with no refresh token restores from localStorage.
- a 401 with no refresh token fires `tokenExpired` and does not attempt a refresh.

Ran the full toolbar suite locally: `jest frontend/src/toolbar/toolbarConfigLogic.test.ts` — 162 passed. TypeScript check is clean for the changed files. I (the agent) did not do manual in-browser testing of a live impersonation session.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude, via the PostHog Slack app) investigated a staff report that the toolbar can't authenticate during impersonation. I traced the OAuth flow end-to-end with the Explore agent, then confirmed the root cause by reading `exchangeCodeForTokens` in `toolbarConfigLogic.ts` (requires both tokens) against `OAuthValidator._should_skip_refresh_token` / `save_bearer_token` in `posthog/api/oauth/views.py` (drops the refresh token for impersonation-minted grants).

I considered changing the backend to issue a refresh token during impersonation, but rejected it: the refresh-less, short-lived shape is an intentional security boundary. The correct fix lives on the client. Invoked the `/writing-tests` skill before adding tests and kept the set to three cases that each catch a real regression.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08M011SBCM/p1784571320823079?thread_ts=1784571320.823079&cid=C08M011SBCM)*

### DIFF
--- a/frontend/src/toolbar/bar/toolbarLogic.ts
+++ b/frontend/src/toolbar/bar/toolbarLogic.ts
@@ -196,12 +196,12 @@ export interface toolbarLogicActions {
     } // toolbarConfigLogic
     setOAuthTokens: (
         accessToken: string,
-        refreshToken: string,
+        refreshToken: string | null,
         clientId: string
     ) => {
         accessToken: string
         clientId: string
-        refreshToken: string
+        refreshToken: string | null
     } // toolbarConfigLogic
     completeGracefulExit: () => {
         value: true
--- a/frontend/src/toolbar/toolbarAuth.ts
+++ b/frontend/src/toolbar/toolbarAuth.ts
@@ -90,7 +90,16 @@ export async function withTokenRefresh(
 
     // `response?.status` (not `response.status`) so a non-Response value from a customer-page
     // `fetch` wrapper can't throw here — callers that don't route through `safeFetch` stay safe too.
-    if (response?.status !== 401 || !accessToken || !refreshToken || !clientId || !uiHost) {
+    if (response?.status !== 401 || !accessToken || !clientId || !uiHost) {
+        return response
+    }
+
+    // Refresh-less tokens (impersonation-minted, refresh-less by design) can't be refreshed —
+    // a 401 means the short-lived access token has expired, so prompt the user to re-authenticate.
+    if (!refreshToken) {
+        toolbarLogger.warn('auth', 'Access token expired and no refresh token available, re-authentication required')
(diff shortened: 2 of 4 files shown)


fercgomes (approved):
weird, impersonation worked with oauth before, not sure what broke it. Thanks!

PR #72977, opened 2026-07-22
fix(cdp): treat Intercom contact create 409 as upsert

### LINKED ISSUE #72192: Intercom "Update contacts" destination: 409 conflict race when concurrent invocations create the same new contact
### Bug Description

**What happens**
The Intercom contacts destination (built from the official template) fails with `http_4xx` when two invocations for the same person run near-simultaneously and the contact doesn't exist in Intercom yet:

```
Error from intercom api (status 409): {'type': 'error.list', 'errors': [{'code': 'conflict', 'message': 'A contact matching those details already exists with id=...'}]}
```

**Why it happens**
The template does search-by-email → create-if-not-found. For a brand-new user, `$identify` and a `$set`-carrying event (e.g. `sign_up`, `login`, `$screen`) often arrive within the same few seconds. Both invocations search, both find nothing, both POST a create — the second one gets the 409 and the invocation is marked FAILED. Verified in our project (EU Cloud): within 10s of an `$identify`, the same users routinely fire multiple `$set`-carrying events, so any destination matching both Identify and Set person properties races on every new contact.

**Expected behavior**
The template should treat 409 as "contact exists": parse the `id` from the error body and retry as an update (PUT /contacts/{id}), i.e. behave as an upsert. Alternatively, use Intercom's upsert semantics where available.

**Impact**
Every new contact produces failed invocations that pollute error monitoring; data self-heals on the next event but the noise makes real failures hard to spot. Currently working around it by forking the source code, which detaches us from template updates.

**Environment**
- PostHog EU Cloud
- Destination: Intercom — Update contacts (official template, OAuth connection)
- Reproducible: consistently, on any contact new to Intercom with two matched events within a few seconds

### Debug info

```shell
- [ ] PostHog Cloud, Debug information: [please copy/paste from https://us.posthog.com/settings/project-details#variables or https://eu.posthog.com/settings/project-details#variables]
```

igennova:
## Problem

The Intercom "Update contacts" destination uses search-by-email → create-if-not-found. For new users, `$identify` and `$set` often fire within seconds, so two invocations can both search (0 results), both POST create, and the second gets HTTP 409 from Intercom. That marks the invocation FAILED and pollutes error monitoring even though data self-heals on the next event.

Closes #72192

## Changes

- On `POST /contacts` returning 409, parse contact id from the conflict error message and retry as `PUT /contacts/{id}`
- Fall back to re-search by email if id parsing fails
- Add regression test for the 409 → PUT path

- Added `test_create_conflict_retries_as_update` — mocks search (0 results), create (409 with id in message), asserts follow-up PUT to `/contacts/{id}`

## Docs update

No docs change — template behavior fix. Existing deployed destinations need to sync from the updated template to pick up the change.

## 🤖 Agent context

Cursor agent implemented the fix from #72192. Chose parse-id-from-409 as primary path with re-search fallback; kept change scoped to `template-intercom` (Update contacts) only.


mayteio (commented):
Thanks for this - the upsert design is exactly what #72192 asked for, and parse-id-with-search-fallback is a sensible shape for it. I checked `extractRegex` is available in both the Python VM (which runs these tests) and the Node executor's regex hooks (which runs destinations in production), so the approach works end to end.

One blocking bug though: the regex string uses an escape Hog doesn't recognise, so the template no longer compiles against master's current parser backends, and every test in `test_template_intercom.py` fails at `setUp`. Details inline - it's a one character-class fix, and with it applied all existing tests pass for me locally.

On the red CI: the semgrep failure is stale-branch noise (it flags unpinned container images in workflow files this PR doesn't touch, already pinned on master), and the remaining failures are a GitHub API timeout cascade in the paths-filter action. Merging master in when you push the fix should clear all of it.

posthog/cdp/templates/intercom/template_intercom.py
```diff
@@ -92,6 +92,48 @@
         },
         'body': payload
     })
+
+    if (res.status == 409) {
+        let contactId := extractRegex(toString(res.body.errors.1.message ?? ''), 'id=([^\\s]+)')
```
mayteio: `\s` is not a valid escape in a Hog string literal, so this line makes the whole template fail to compile under both current parser backends (rust: `unrecognised escape '\s'`; cpp: lexer error). Every test in this file currently fails at `setUp` because of it. My guess is the parser backend that was primary when this PR was opened was lenient about unknown escapes, and the default has since changed on master, so this wasn't visible at the time.

The Hog source needs a literal `\\s` so the regex engine receives `\s`, which means four backslashes in this Python string:

```suggestion
        let contactId := extractRegex(toString(res.body.errors.1.message ?? ''), 'id=([^\\\\s]+)')
```

With this one change the full test file passes for me locally against current master.

posthog/cdp/templates/intercom/test_template_intercom.py
```diff
@@ -155,6 +155,52 @@ def test_function_requires_identifier(self):
         assert not self.get_mock_fetch_calls()
         assert self.get_mock_print_calls() == [("No email set. Skipping...",)]
 
+    def test_create_conflict_retries_as_update(self):
```
mayteio, line 158: Could you also add a regression test for the fallback branch (409 with no parseable `id` in the message, falling back to re-search)? It's the safety net if Intercom ever changes the conflict message format, and it's currently uncovered.

The test harness keys mocked responses by URL with one static response each, so the variant we can express here is "re-search still finds nothing, the original 409 surfaces". This passes locally once the escape above is fixed:

```python
    def test_create_conflict_falls_back_to_search_when_id_not_in_message(self):
        self.fetch_responses = {
            "https://api.intercom.io/contacts/search": {
                "status": 200,
                "body": {"total_count": 0},
            },
            "https://api.intercom.io/contacts": {
                "status": 409,
                "body": {
                    "type": "error.list",
                    "errors": [{"code": "conflict", "message": "A contact matching those details already exists"}],
                },
            },
        }

        with pytest.raises(UncaughtHogVMException) as e:
            self.run_function(inputs=self.create_inputs())

        calls = self.get_mock_fetch_calls()
        assert len(calls) == 3
        assert calls[2][0] == "https://api.intercom.io/contacts/search"
        assert "status 409" in e.value.message
```

A success-path variant (re-search finds exactly one contact and we PUT to it) would need sequenced responses per URL, which `BaseHogFunctionTemplateTest.mock_fetch_response` doesn't support today, so I'd call that out of scope.

igennova:
@mayteio 
Both fixed ,doubled the backslash so it compiles on both backends, and added the fallback test. 12 passed locally.
Screenshot 2026-08-14 at 12 37 12 AM

Thanks

mayteio (approved):

mayteio (approved):
Thanks mate!

igennova:
> Thanks mate!


Welcome 😃 @mayteio

igennova:
Hi @mayteio, this PR got the stale label, but it's already approved could you merge it?
Thanks.

PR #33098, opened 2025-06-03
fix(flags): Validate feature flag key on the server

haacked:
## Problem

The feature flag create/edit page validates the feature flag key on the client, but not the server. This allows a bad actor to create feature flag keys with invalid characters.

## Changes

Use the same regex on the client and server to validate the feature flag key. This also simplifies the client regex as it was needlessly complex.

Also ensured that when creating a URL with feature flag key, we URL encode the feature flag key as it's part of a path component. This isn't strictly necessary any longer, but provides defense in depth.

## Did you write or update any docs for this change?

- [ ] I've [added or updated the docs](https://posthog.com/handbook/engineering/writing-docs)
- [ ] I've reached out for help from the docs team
- [x] No docs needed for this change

## How did you test this code?

Manually. Added unit tests.

PR #5725, opened 2021-08-24
Multivariate flags UI

samwinslow:
## Changes

Contributes to #5440. Adds a UI in PostHog for configuring multivariate feature flags.

**Notes**
- This feature itself is feature-flagged behind `5440-multivariate-support`
- Client-side library support (i.e. our JS and Python consumers) is not yet available, so this is just a UI for updating feature flag records in the database.
- This will be tested internally for Growth projects before any kind of public release

The feature flag rollout page:
Screen Shot 2021-08-24 at 1 13 24 PM

Validations:
Screen Shot 2021-08-24 at 1 14 14 PM
Screen Shot 2021-08-24 at 1 15 01 PM

Distribute evenly (handling remainders from integer rounding):
Screen Cast 2021-08-24 at 1 16 03 PM

Data loss warning if switching from variants to Boolean (default) behavior:
Screen Cast 2021-08-24 at 1 18 36 PM

## Checklist

- [ ] All querysets/queries filter by Organization, by Team, and by User
- [ ] Django backend tests
- [ ] Jest frontend tests
- [ ] Cypress end-to-end tests
- [ ] Migrations are safe to run at scale (e.g. PostHog Cloud) – present proof if not obvious
- [ ] New/changed UI is decent on smartphones (viewport width around 360px)

PR #3675, opened 2021-03-16
Feature Flags Revamp (UX & Code)

paolodamico:
## Changes

This PR introduces a significant revamp of feature flags (both in terms of code & UX).
- Closes #1807 (please see issue for all the details on UX improvements).
- Refactors how feature flags are created and managed. Form will now be opened in a separate page. This also includes a significant refactor of the UI which among other things: follows same visual language as actions, introduces instructions for Python (@yakkomajuri FYI the docs page for each integration seems to be lacking proper anchors), makes it significantly clearer when the flag is universally enabled/disabled, clarity around release groups (e.g. when not applying a release %) 
    
- We treat the case of having a single filter separately to have a more intuitive/natural flow for these cases.
    
- Better front-end validation
    
- Adds clarity on the feature flag table for when a flag is disabled
    
- Moves to Typescript and Kea everything related to feature flag. Logic is also significantly simplified for clarity.
- Some backend refactoring introduces a few more tests, and moves instrumentation logic to the serializer.
- Closes #3661

## Checklist

- [X] All querysets/queries filter by Organization, by Team, and by User
- [X] Django backend tests
- [X] Jest frontend tests
- [X] Cypress end-to-end tests

PR #72995, opened 2026-07-22
fix(flags): allow dots and slashes in variant keys in the UI

patricio-posthog:
## Problem

Variant keys have been charset-restricted to `^[a-zA-Z0-9_-]+$` in the flag editor since the multivariate UI shipped (#5725, which copied the rule from the flag key input added in #3675). The API has never enforced this for variant keys — only the flag key is validated server-side (#33098) — and flag evaluation treats variant keys as opaque strings everywhere (Rust matcher, all SDKs).

The result: flags created via the API with variant keys like `provider/model-1.2` (a common pattern where the variant key is consumed directly as a config value, e.g. LLM model routing) work fine at runtime but can't be edited in the UI — the form rejects the keys on save, and renaming a variant isn't an option since SDK checks reference the key.

Related to the filters validation work in #50084: the audit there surfaced hundreds of flags using dotted/slashed variant keys in production, which is what prompted revisiting the frontend rule.

## Changes

- New `validateFeatureFlagVariantKey` in `featureFlagLogic.ts`: same non-empty and 400-char rules as flag keys, but the charset also allows dots and slashes.
- Variant key call sites switch to it: the flag form errors, the `variantErrors` selector, and the scheduled-changes editor (`FeatureFlagSchedule.tsx`).
- Flag keys keep the strict rule everywhere (flag form, experiments) — that one is server-enforced and URL-adjacent, and is not changed here.

Spaces and other characters remain rejected for variant keys pending the #50084 inventory discussion; this only legalizes the two characters with demonstrated production usage.

- Added a parameterized `validateFeatureFlagVariantKey` suite to `featureFlagLogic.test.ts` (4 accept / 5 reject cases). It guards the realistic regression of someone re-unifying variant validation onto `validateFeatureFlagKey` — the exact copy-paste that introduced this restriction in #5725. The existing `validateFeatureFlagKey` suite still asserts flag keys reject dots/slashes.
- Full `featureFlagLogic.test.ts` suite passes locally (181 tests).
- `pnpm --filter=@posthog/frontend typescript:check` reports no errors in changed files (two pre-existing module-resolution errors on master are unrelated).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs describe the variant key charset restriction.


gustavohstrassburger (approved):
Nice,

Just a question, shouldn't this also be applied to experiments? It looks like `validateFeatureFlagKey` is also used there.

Also, is `validateFeatureFlagKey` still relevant? There are places using it, just not sure if we should be replacing it with the new function.

patricio-posthog:
I checked all of them before scoping this. The experiments usages all validate the flag key (`feature_flag_key`), not variant keys, so they should stay strict: the server enforces that same regex for flag keys. The experiment variant inputs never used this validator at all (they only check non-empty), so experiments already accept dots and slashes. This PR just brings the flag editor in line with that.

So yes, `validateFeatureFlagKey` is still needed, but only for flag keys.

Btw, your question made me notice a stale comment in `ExperimentReplayTab.tsx` claiming variant keys are restricted to `[a-zA-Z0-9_-]`. Fixed it in this PR.

PR #72997, opened 2026-07-22
fix(conversations): return 405 for unsupported ticket create POST

### LINKED ISSUE #71101: 500 Internal Server Error when creating ticket via POST /api/projects/:project_id/conversations/tickets/
### Bug Description

**Bug Description:**

When attempting to create a support ticket via the documented API endpoint `POST /api/projects/:project_id/conversations/tickets/`, the server responds with a 500 Internal Server Error. The request follows the documentation structure at https://posthog.com/docs/api/conversations-2#post-api-projects-project_id-conversations-tickets and uses a personal API key with the `ticket:write` scope.

The documentation example itself appears problematic, as it shows `ticket_number="integer"` as a request parameter, but `ticket_number` is a read-only field that's auto-generated, not a parameter you send when creating tickets.

**How to reproduce:**

1. Generate a personal API key with `ticket:write` scope
2. Execute curl commands with various request body configurations (all produce the same 500 error):

**Empty body:**
```bash
curl -H 'Content-Type: application/json'\
    -H "Authorization: Bearer $POSTHOG_PERSONAL_API_KEY" \
    https://us.posthog.com/api/projects/215951/conversations/tickets/ \
    -d '{}'
```

**Body with all documented parameters:**
```bash
curl --location 'https://us.posthog.com/api/projects/215951/conversations/tickets/' \
    -H "Authorization: Bearer $POSTHOG_PERSONAL_API_KEY" \
    -H 'Content-Type: application/json'\
    -d '{
        "message": "Pineapple is the key(ng)",
        "anonymous_traits": {
            "name": "John Doe",
            "email": "test@test.com"
        },
        "priority": "low",
        "status": "new"
    }'
```

**Minimal body (only documented required fields):**
```bash
curl -H 'Content-Type: application/json'\
    -H "Authorization: Bearer $POSTHOG_PERSONAL_API_KEY" \
    https://us.posthog.com/api/projects/215951/conversations/tickets/ \
    -d '{"status": "new"}'
```

3. All variations return the same error:
```json
{"type":"server_error","code":"error","detail":"A server error occurred.","attr":null}
```

**Additional context:**

- Tested with empty body, minimal body (only documented parameters), full body with all optional fields, and partial bodies
- Removed `message` field as it's not in the documented parameters for this endpoint
- Personal API key has `ticket:write` scope enabled
- Same authentication pattern works for other API endpoints
- Documentation example at https://posthog.com/docs/api/conversations-2 may contain an error in the example request parameters
- Similar issue reported in community: https://posthog.com/questions/500-internal-server-error-when-creating-tickets-via-conversations-api

### Debug info

```shell
Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019f6621-4380-77d2-b65a-9b13a1ac1b08?t=652
Admin (internal): http://go/adminOrgUS/01977468-330d-0000-6646-63394d81ef68 (project ID 215951)
```

slshults:
## Problem

`POST /api/projects/:project_id/conversations/tickets/` returned a 500 for every caller (issue #71101). A customer hit it with a `ticket:write` personal API key following the public API docs, and got `{"type":"server_error",...}` for every payload they tried (empty body, minimal body, full documented body).

Root cause: `TicketViewSet` is a stock DRF `ModelViewSet` with `"create"` listed in `scope_object_write_actions` but no `create`/`perform_create` override, and `TicketSerializer` is a plain `ModelSerializer` with no `create()`. So DRF's default create called `Ticket.objects.create(**validated_data)`, which never set the NOT NULL `team` FK or `ticket_number`. Those are only ever set by `TicketManager.create_with_number` (which locks the team row and increments per team). The result was a deterministic `IntegrityError` → unhandled → 500.

The bare collection POST was never an intended intake path. Every real ticket creation goes through `create_with_number(team=...)`: the widget, email, Slack, Teams, GitHub, Zendesk import, and the authenticated `compose` action. The docs also advertised the endpoint's read-only `ticket_number` as a request param, which the reporter flagged as wrong.

Closes #71101

## Changes

Disable the accidental create endpoint rather than implement it as a feature (the Conversations team can implement a real API create path as a follow-up if they want one):

- Override `create()` on `TicketViewSet` to raise `MethodNotAllowed` (405) with a message pointing callers to `posthog.conversations.sendMessage()` in the JS SDK.
- Mark it `@extend_schema(exclude=True)` so it drops out of the generated OpenAPI spec. That removes the dead endpoint (and the bogus `ticket_number` request param) from the public API docs, and regenerates `products/conversations/frontend/generated/api.ts` + `api.zod.ts` (the `conversationsTicketsCreate` op was dead code, no callers).
- Keep `"create"` in `scope_object_write_actions`. This is deliberate: DRF runs permission checks before method dispatch, so with `"create"` removed a `ticket:write` token would get a misleading 403 ("does not support personal API key access") instead of the helpful 405. Keeping it listed lets the token clear the scope gate and reach the 405. Session auth reaches the 405 either way.

This mirrors the existing precedent for disabling a single action in `posthog/api/person.py` (`create()` → `MethodNotAllowed` + `@extend_schema(exclude=True)`).

Automated (all run by me, the agent, via `hogli test`):

- Added a parameterized `test_create_ticket_not_allowed` in `test_tickets.py` asserting 405 for both **session** and **`ticket:write` personal API key** auth. The PAK case is the reporter's exact path, and it's the regression that matters here: a session-only test passes even under the 403 behavior, so it wouldn't have caught the auth-path divergence. No existing test POSTed the bare collection URL.
- Full conversations ticket suite green: `173 passed`. `ruff check` clean.

Manual (local, by me):

- Reproduced the bug against the running dev stack (master code): all three reporter payloads → **500 `IntegrityError`**, GET list → 200.
- Verified the fix by running this branch's code on a throwaway backend against the same local DB: all three payloads → **405** with the SDK-pointer message. Confirmed GET list still returns 200 on the full stack.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Steven (@slshults) directed the work; DRI and assignee.

Investigated and confirmed by a prior agent session on the support ticket; this PR was built by Claude (Opus 4.8) in PostHog Code. Skills invoked while shaping the change: `/improving-drf-endpoints` and `/writing-tests`. TDD throughout (wrote the failing 405 test first, confirmed red → green).

Key decision made along the way: the original plan was to also drop `"create"` from `scope_object_write_actions`. While tracing `APIScopePermission`, I found that would give the PAK-authenticated reporter a 403 rather than the intended 405, so I kept `"create"` listed and added the PAK-auth test to lock in the 405 for their exact scenario. Steven chose this path.

> [!NOTE]
> The pre-push `ci:preflight` flagged an OpenAPI drift advisory for `services/mcp/src/generated/platform_features/*`. That drift is pre-existing on master (introduced by #72704, which changed an org serializer without regenerating the MCP artifacts) — not from this PR. I reverted those regenerated files to keep this PR scoped to the ticket fix; they belong in a separate cleanup.

---

### DIFF
--- a/products/conversations/backend/api/tests/test_tickets.py
+++ b/products/conversations/backend/api/tests/test_tickets.py
@@ -26,6 +26,8 @@ from posthog.hogql import ast
 from posthog.hogql.query import execute_hogql_query
 
 from posthog.models import ActivityLog, Comment, Organization, Tag, User
+from posthog.models.personal_api_key import PersonalAPIKey
+from posthog.models.utils import generate_random_token_personal, hash_key_value
 from posthog.test.persons import create_person
 
 from products.conversations.backend.api.tickets import TicketReplyRequestSerializer
@@ -59,6 +61,32 @@ class TestTicketAPI(APIBaseTest):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json()["count"], 1)
         self.assertEqual(response.json()["results"][0]["id"], str(self.ticket.id))
+
+    @parameterized.expand(["session", "personal_api_key"])
+    def test_create_ticket_not_allowed(self, mock_on_commit, auth):
+        # The bare collection POST was never a real intake path — DRF's default create
+        # can't set team/ticket_number and used to 500 (issue #71101). It must 405 instead.
+        # The reporter used a ticket:write personal API key, so cover that path too: with
+        # "create" kept in scope_object_write_actions the token clears the scope gate and
+        # reaches the 405 rather than a misleading "not supported" 403.
+        extra = {}
+        if auth == "personal_api_key":
+            raw_key = generate_random_token_personal()
+            PersonalAPIKey.objects.create(
+                label="ticket-write",
+                user=self.user,
+                secure_value=hash_key_value(raw_key),
+                scopes=["ticket:write"],
+            )
+            extra["HTTP_AUTHORIZATION"] = f"Bearer {raw_key}"
+
+        response = self.client.post(
+            f"/api/projects/{self.team.id}/conversations/tickets/",
+            data={"status": "new"},
+            format="json",
+            **extra,
+        )
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
 
     def _ticket_with_tags(self, *tag_names):
         ticket = Ticket.objects.create_with_number(
--- a/products/conversations/backend/api/tests/test_tickets.py
+++ b/products/conversations/backend/api/tests/test_tickets.py
@@ -69,7 +69,6 @@ class TestTicketAPI(APIBaseTest):
         # The reporter used a ticket:write personal API key, so cover that path too: with
         # "create" kept in scope_object_write_actions the token clears the scope gate and
         # reaches the 405 rather than a misleading "not supported" 403.
-        extra = {}
         if auth == "personal_api_key":
             raw_key = generate_random_token_personal()
             PersonalAPIKey.objects.create(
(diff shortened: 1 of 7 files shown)


veryayskiy (approved):
🚢

PR #73010, opened 2026-07-22
fix(users): serialize email change and pin verification email recipient

Piccirello:
## Problem

The email-change flow sets `pending_email`, mints a verification token, and sends the email as three unlocked steps. Two concurrent changes on the same account can interleave so a token minted for one address has its verification email sent to another, because the recipient is re-read from `pending_email` at send time instead of the address the token was minted for.

## Changes

- Serialize the change under a row lock (`select_for_update`) and mint the token inside the lock, so the token binds to one consistent `pending_email`.
- Send after the transaction commits, with the recipient pinned to the captured address via a new `target_email` arg on `send_email_verification`.
- Split `create_token_and_send_email_verification` into mint + `send_verification_email` so the send stays synchronous and keeps its existing error handling.

No change to the normal single-request flow.

Added `test_email_verification.py` (`SimpleTestCase`, no DB): the recipient is the pinned `target_email` even when `pending_email` has drifted, and falls back to `pending_email` when no target is passed. Confirmed it goes red on the old behavior. The `test_user.py` change-then-verify tests need ClickHouse, which my local harness couldn't reach this session, so they run in CI. I traced them: the existing ones are unaffected (the send still routes through the same module path) and I updated one patch target for the split. ruff and mypy pass on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude) wrote this with Tom. Invoked `/writing-tests`. Considered pinning only the recipient, and rebinding the token to the in-memory instance; rejected both, since the recipient pin alone leaves a second interleaving open and rebinding touches a token primitive shared by 7 flows. Serializing the critical section closes the class by construction. Used a synchronous post-commit send instead of `transaction.on_commit` to keep the friendly send-failure error and avoid on_commit not firing under `TestCase`.


a-lider (approved):
nice, just merged master into the branch (there was a huge query snapshot diff), hope you don't mind

PR #73016, opened 2026-07-22
feat(cyclotron): add V1 to V2 job relocation script

dmarchuk:
## Problem

Cyclotron V1 (legacy PostgreSQL backend) needs to be decommissioned. To do so safely without waiting ~90 days for natural job drain, we need a one-shot operator script that relocates parked jobs from V1 to V2 while preserving job identity and scheduling, and cleans up corrupt rows that will never execute.

## Changes

Added a new operator script `relocate-cyclotron-v1-jobs.ts` that:

- **Relocates legit jobs**: Moves every `available` job scheduled within the next year from V1 to V2, preserving the original job ID and scheduled time
- **Deletes corrupt rows**: Removes jobs scheduled more than a year in the future (a delay-overflow bug parked them so far out they never come due)
- **Safety-first ordering**: Writes to V2 first, verifies each ID is present in V2, then deletes only verified IDs from V1 — ensures V1 is never emptied ahead of a confirmed V2 write
- **Idempotent**: V2 writes use `overwriteExisting: true` with original IDs; re-runs after partial failures complete the job; once drained, re-runs are no-ops
- **Dry-run by default**: Pass `--apply` to actually write and delete; otherwise reports what would happen

The script includes:
- Full end-to-end test coverage against real V1 and V2 databases
- Transparent handling of gzip-compressed VM state (defensive decoding)
- Reuse of production `cyclotronJobToInvocation` conversion logic to ensure payload handling matches the drain worker
- Detailed logging and safety warnings (e.g., when IDs fail V2 verification)

Exported `cyclotronJobToInvocation` from `job-queue-postgres.ts` to allow the relocation script to use the same conversion logic as production workers.

Added comprehensive end-to-end tests in `relocate-cyclotron-v1-jobs.test.ts` that exercise the script against real V1 and V2 test databases:

- **Relocation with payload preservation**: Verifies legit rows land in V2 with all fields intact (ID, schedule, priority, function ID, parent run ID, decoded VM state), corrupt rows are deleted, and V1 is fully drained
- **Gzip-compressed state handling**: Confirms transparent decoding of gzip'd VM state end-to-end
- **Dry-run mode**: Confirms no writes or deletes occur in dry-run
- **Idempotency**: Verifies a second apply after full drain is a no-op
- **Conflict handling**: Tests that stragglers with IDs already present in V2 (active) pass the verification gate and drain from V1 without error
- **Safety gate**: Confirms that when V2 write fails (non-conflict error), V1 is left untouched and corrupt delete does not run

All tests pass with the real job-queue producer and database pools.

https://claude.ai/code/session_014FL1Pj5nCDJvJwKiU59Biu


dmarchuk:
Closing this one for now, we might not even need it in a week

PR #73017, opened 2026-07-22
feat(llm-gateway): accept caller trace id, protect gateway-owned event properties

brandonleung:
## Problem

Callers of the Python LLM gateway have no supported way to correlate their own records with the captured `$ai_generation`: the OpenAI route derives `$ai_trace_id` internally. And the property merge runs caller-last with no protection, so any `x-posthog-property-$ai_*` header silently overwrites gateway-derived fields (`$ai_input`, `$ai_model`, token counts, `$ai_trace_id`).

## Changes

- Accept an `x-posthog-trace-id` request header (stripped, capped at 256 chars) and use it as `$ai_trace_id` on both success and failure captures. Stored verbatim so a caller's saved id matches the join; the `metadata.user_id` derivation still applies when the header is absent.
- Make the caller-property merge use `setdefault`, so a caller can add its own properties (e.g. `$ai_span_name`, which the gateway never derives) but cannot overwrite a value the gateway already set.
- Carry the trace id through the `handle_llm_request` context rebuild.
- Log when an over-256 trace-id header is dropped, so a caller whose join comes back empty has a signal. Delete the unused `set_posthog_context` helper.

First consumer: the surveys gateway helper (#72670), so survey summary ratings can join the generation they rate.

## Verification

llm-gateway suite (1342 passed), ruff, format. Route-level coverage on both `/v1/chat/completions` and `/{product}/v1/chat/completions`; callback tests pin header-wins, verbatim storage, the unchanged `metadata.user_id` fallback, and that a caller `$ai_trace_id`/`$ai_model` property loses to the derived value while `$ai_span_name` passes through.

## Known gaps

Low: a caller can still set `$set`/`$set_once` and, when the gateway computes no cost, `$ai_total_cost_usd`, since the gateway does not set those before the merge. Both are pre-existing (unchanged from master); tracked separately.

## 🤖 Agent context

Claude (Opus 4.8) wrote this under my direction. Requires human review.

### DIFF
--- a/services/llm-gateway/tests/callbacks/test_posthog.py
+++ b/services/llm-gateway/tests/callbacks/test_posthog.py
@@ -313,6 +313,127 @@ class TestPostHogCallback:
             assert _is_uuid(props["$ai_trace_id"])
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize("event_method", ["_on_success", "_on_failure"])
+    async def test_trace_id_header_wins_over_metadata_user_id(
+        self,
+        event_method: str,
+        callback: PostHogCallback,
+        auth_user: AuthenticatedUser,
+        standard_logging_object: dict,
+        mock_posthog_client: tuple,
+    ) -> None:
+        """A caller-stamped x-posthog-trace-id beats the metadata.user_id derivation,
+        on both the success and the error event, so a caller that stores the id it
+        sent can join ratings/feedback onto the captured generation."""
+        _, mock_client = mock_posthog_client
+        header_trace_id = "0e3c1c81-8bcc-40d8-9e28-1ed92b3a7c3a"
+        kwargs = {
+            "standard_logging_object": standard_logging_object,
+            "litellm_params": {"metadata": {"user_id": "trace-id-123"}},
+        }
+
+        with (
+            patch("llm_gateway.callbacks.posthog.get_auth_user", return_value=auth_user),
+            patch("llm_gateway.callbacks.posthog.get_product", return_value="llm_gateway"),
+            patch("llm_gateway.callbacks.posthog.get_posthog_trace_id", return_value=header_trace_id),
+        ):
+            await getattr(callback, event_method)(kwargs, None, 0.0, 1.0, end_user_id=None)
+
+            props = mock_client.capture.call_args.kwargs["properties"]
+            # a UUID header value passes through verbatim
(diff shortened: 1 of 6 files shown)


services/llm-gateway/tests/callbacks/test_posthog.py
```diff
@@ -312,6 +312,127 @@ async def test_on_success_uses_end_user_id_for_distinct_id(
...
+    @pytest.mark.parametrize("event_method", ["_on_success", "_on_failure"])
+    async def test_trace_id_header_wins_over_metadata_user_id(
+        self,
+        event_method: str,
+        callback: PostHogCallback,
+        auth_user: AuthenticatedUser,
+        standard_logging_object: dict,
+        mock_posthog_client: tuple,
+    ) -> None:
+        """A caller-stamped x-posthog-trace-id beats the metadata.user_id derivation,
```
richardsolomou, line 325: can we drop the new test docstrings? our repo guidance says Python tests shouldnt have them, and the test names already carry the behaviour.

services/llm-gateway/tests/callbacks/test_posthog.py
```diff
@@ -312,6 +312,127 @@ async def test_on_success_uses_end_user_id_for_distinct_id(
...
+                    "$ai_span_name": "slack_nudge_classifier",
+                },
+            ),
+        ):
+            await getattr(callback, event_method)(kwargs, None, 0.0, 1.0, end_user_id=None)
+
+            props = mock_client.capture.call_args.kwargs["properties"]
+            # gateway-derived keys win over the caller's spoofed values...
+            assert props["$ai_trace_id"] == header_trace_id
+            assert props["$ai_model"] != "spoofed"
```
richardsolomou, line 407: could we assert the exact expected model here? this still passes if the gateway replaces it with an empty or unrelated value.

services/llm-gateway/tests/test_request_context.py
```diff
@@ -26,6 +27,16 @@ def make_mock_user() -> MagicMock:
     return user
 
 
+def make_mock_request(headers: list[tuple[str, str]]) -> MagicMock:
```
richardsolomou, line 30: couldnt this use Starlette's `Headers` directly? it already handles `.items()` and case-insensitive `.get()`, so we dont need to maintain a partial mock implementation.

brandonleung:
closing as this introduces too much risk and we address it in the new gateway

PR #73020, opened 2026-07-22
fix(navigation): remove pinnedByDefault sidebar auto-pin

rafaeelaudibert:
## Problem

Any product could set `pinnedByDefault` on its nav manifest entry and get force-injected into every user's sidebar whenever its feature flag was on, whether or not the user ever chose it. Teams should not be able to force their product into a user's sidebar, and we don't want them controlling what a person sees there. That's the user's call.

It also produced a visible bug. The sidebar render path honored `pinnedByDefault` but the "edit list" toggle view did not, so an auto-pinned product showed up in the sidebar while reading as un-toggled in the customize view. And un-toggling it never stuck, because the render path re-added it on every render while the flag was on.

## Changes

Removed the `pinnedByDefault` concept entirely:

- the auto-pin injection loop in `getCustomProductTreeItems` (`projectTreeDataLogic.tsx`)
- the `pinnedByDefault` field on the `FileSystemImport` type (`schema-general.ts`)
- its use by Engineering analytics, Replay vision, and Tasks (both `products.tsx` and each product manifest)
- regenerated `frontend/src/queries/schema.json` and `posthog/schema.py`

The sidebar now renders only the products a user has actually added to their list.

> [!NOTE]
> This touches three other teams' products (Replay vision, Engineering analytics, Tasks). They relied on this to stay visible during their beta/launch push and will lose that auto-promotion. Flagging so it's not a surprise.

Ran the frontend TypeScript check (clean for this change; the only error is a pre-existing missing-module in an unrelated `early_access_features` test) and lint/format. I (the agent) did not do manual UI testing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

n/a

## 🤖 Agent context

Rafa directed this: he noticed Tasks and Engineering analytics showing in his sidebar despite being un-toggled in the edit view, and asked me (Claude Fable 5, in Claude Code) to find the cause and then remove the mechanism. I traced it to the `pinnedByDefault` auto-pin, confirmed the render vs edit-list selector mismatch, and removed the concept end to end.

`products.tsx` is generated from the product manifests, so the pre-commit hook regenerated it from my manifest edits. I reverted a handful of unrelated files that the lint/format run touched (oxfmt/oxlint drift already on master) to keep the diff focused.

### DIFF
--- a/frontend/src/layout/panel-layout/ProjectTree/projectTreeDataLogic.tsx
+++ b/frontend/src/layout/panel-layout/ProjectTree/projectTreeDataLogic.tsx
@@ -1569,19 +1569,6 @@ export const projectTreeDataLogic = kea<projectTreeDataLogicType>([
                         })
                         .filter((p): p is FileSystemImport => p !== null)
 
-                    // Auto-pin items the manifest marks `pinnedByDefault` once their flag is on.
-                    for (const product of allProducts) {
-                        if (
-                            product.pinnedByDefault &&
-                            product.flag &&
-                            (featureFlags as Record<string, boolean>)[product.flag] &&
-                            !selectedProductPaths.has(product.path)
-                        ) {
-                            selectedProductPaths.add(product.path)
-                            selectedProducts.push(product)
-                        }
-                    }
-
                     const imports = selectedProducts
                         .filter((f) => !f.flag || (featureFlags as Record<string, boolean>)[f.flag])
                         .map((i) => ({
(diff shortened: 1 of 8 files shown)


rafaeelaudibert:
* **#73020** <a href="https://app.graphite.com/github/pr/PostHog/posthog/73020?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/73020?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rafaeelaudibert:
### Merge activity

* **Jul 22, 10:37 PM UTC**: @rafaeelaudibert merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/73020).

PR #73026, opened 2026-07-22
feat(replay-vision): restore gemini-3-flash-preview as 5-credit default

arnohillen:
## Problem

Yesterday's #72713 retired Gemini 3 Flash (`gemini-3-flash-preview`) and switched the default scanner model to `gemini-3.6-flash`, which bills at 15 credits per observation instead of 5. We want the cheaper 5-credit model back as the default so new scanners start at the familiar lower price point.

## Changes

Restores `gemini-3-flash-preview` as a selectable model and the default for new scanners, at 5 credits per observation. The lineup is now three models across two Google tiers:

| model | tier | credits/observation |
|---|---|---|
| gemini-3.5-flash-lite | flash lite | 2 (budget) |
| gemini-3-flash-preview | flash | 5 (default) |
| gemini-3.6-flash | flash | 15 |

- Adds `GEMINI_3_FLASH_PREVIEW` back to the `ScannerModel` enum with display name "Gemini 3 Flash (preview)" so it reads distinctly from the stable "Gemini 3.6 Flash".
- Un-retires it in `billing.GEMINI_MODELS` at 5 credits. That is exactly what the margin formula produces from its token prices (0.50 in / 3.00 out per 1M tokens at the 3.75x target), so pricing stays consistent with the rest of the lineup.
- Flips the estimate serializer default and the frontend `DEFAULT_MODEL` to the preview, and regenerates OpenAPI/MCP types.
- Migration 0053 updates the field choices (state-only `AlterField`, no DB schema change).

> [!NOTE]
> Existing scanners are unaffected. Migration 0052 already remapped anything on `gemini-3-flash-preview` to `gemini-3.6-flash`, so this only changes what new scanners default to. `gemini-3-flash-preview` is a preview id, so if Google retires it we remap like 0052 did.

- `test_billing.py`, including the mirror test (enum ↔ `GEMINI_MODELS`) and the flash margin test, which I parameterized to also assert `gemini-3-flash-preview`'s 5 credits equals the formula output. That guards against a future token-price change silently leaving the credit price stale, which no existing test caught (the mirror test only checks membership).
- `test_api.py` estimate tests. I updated `test_estimate_counts_only_in_window_sessions` to expect 5 credits / 15 per month now that the default model changed.

I regenerated types with `hogli build:openapi` and ran `hogli ci:preflight` (0 failures). I did not get the full frontend typecheck to green in my worktree because the quill workspace packages are not built there (an environment quirk unrelated to this change). replay_vision itself reports zero type errors.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Arnaud asked me to double-check whether the flash model costs 5 or 15 credits after yesterday's PRs, then to bring back the 5-credit model as the default with a version-explicit name. I (Claude) traced the pricing history and confirmed the 5-credit model was `gemini-3-flash-preview` (not the "3.1" flash-lite preview, which was 2 credits and never the default), verified 5 credits sits exactly on the margin formula, and made the change.

Skills invoked: `/improving-drf-endpoints` (serializer default), `/django-migrations` (choices migration), `/writing-tests` (test changes).

Decisions: kept the change additive (added the preview alongside the existing 2- and 15-credit models rather than reverting the whole #72713 lineup) since existing scanners already migrated to `gemini-3.6-flash`. Considered adding a pro tier but left it out per Arnaud's call to keep the three-model lineup. Named it "Gemini 3 Flash (preview)" to make the version explicit next to "Gemini 3.6 Flash".


fasyy612 (approved):
sweeet 👌

PR #73027, opened 2026-07-22
fix(ducklake): pin HTTPS for delta-kernel staging reads via duckgres

EDsCODE:
## Problem

On cache-proxy-enabled duckgres clusters, the org's ambient S3 secret deliberately carries `USE_SSL false` so httpfs traffic flows through the logging cache proxy. But `delta_scan`'s delta-kernel object store **ignores DuckDB's proxy configuration** and dials the plain-HTTP S3 endpoint directly — and direct port-80 egress is silently dropped by the duckgres worker network policy.

Verified in mw-prod-us CP logs (2026-07-22): every duckgres-side `delta_scan` in the data-imports and data-modeling copy workflows fails with `DeltaKernel ObjectStoreError … GET http://s3.us-east-1.amazonaws.com/... after 10 retries` at ~58s per attempt, continuously re-attempted across many teams (team 2, devex, 19618, 35787, …) — **29 failures/15min, ~2 worker-minutes burned per minute, zero successes**, each attempt pinning a dedicated worker under duckgres's one-session-per-worker model.

## Change

`create_staging_read_secret(conn, catalog_bucket)` in `posthog/ducklake/storage.py`, called after `setup_duckgres_session` at the three duckgres `delta_scan` sites (imports create, imports verification, modeling create):

```sql
CREATE OR REPLACE SECRET posthog_staging_delta_https (
  TYPE S3, PROVIDER credential_chain, USE_SSL true,
  SCOPE 's3://<bucket>/__posthog_staging')
```

- The longer SCOPE wins DuckDB's prefix resolution for the staging tree **and nowhere else** — DuckLake data traffic keeps the proxy transport.
- `credential_chain` resolves the worker's own ambient (org-scoped) credentials and region; no secret material crosses the wire.
- Session-scoped by design: duckgres wipes non-persistent secrets at the next session create.
- Fixing it here rather than in duckgres keeps PostHog's staging-path convention out of duckgres, and avoids weakening the proxy default for anything but the staging prefix.

- TDD: extended `test_copy_data_imports_to_ducklake_activity_via_duckgres` to require the secret DDL (provider/scope/USE_SSL) **before** the first `delta_scan`; watched it fail, then implemented.
- `test_ducklake_copy_data_imports_workflow.py -k duckgres`: 4 passed; `test_ducklake_copy_data_modeling_workflow.py -k duckgres`: 1 passed.
- Pre-commit hooks (ruff, format, ty) green.
- Root cause verified end-to-end in production logs before writing any code (full trail in the linked investigation).

## Rollback

Revert; behavior returns to today's failing-loop state (no data loss — the loop never succeeded).

## 🤖 Agent context

Authored by Claude Code (Fable 5) during the duckgres-sink throughput investigation; the failure signature (silent port-80 egress drop) matches the documented worker warmup air-gap incident.


EDsCODE:
CI triage: the failing `Django tests – Temporal (3/14)` shard also fails on master's own latest ci-backend runs (same job name, plus a Core shard) — inherited breakage, not this change. The duckgres-path tests this PR touches pass locally (imports + modeling, plus the full storage suite). Will re-run checks once master is green.

PR #73028, opened 2026-07-22
perf(data-warehouse): reuse the duckgres session across a group's live batches

EDsCODE:
## Problem

After the co-claim fix (#72985), the duckgres sink's throughput ceiling is per-batch **session** cost. Production phase attribution (2026-07-22, consumer phase histogram + duckgres CP query-log correlation):

| Cost | Where |
|---|---|
| ~0.5-1s | worker session create (fresh connection per batch) |
| ~5-19s | extended-protocol describe probe cold-enumerating a 2,000+-table DuckLake catalog (measured **+5.3s** cold vs warm, empirically) |
| ~12-19s | the transaction's first-write metadata touch on the target table (14k+ files) |
| ~80-100ms | the actual 5,000-row MERGE |
| <0.7s | marker INSERT + DuckLake commit |

Every batch paid the cold costs because `process_batch` opens a fresh connection each time — ~30-40s per batch, ~7-23/hr per group against a producer emitting ~74/hr.

## Change

A `(team_id, schema_id)`-keyed session cache in the duckgres processor:

- The sink's group lease already serializes each key to one pod task, so cached entries are never used concurrently; the lock guards only the dict.
- Preamble (httpfs load + extract-read secret) runs once per cached session.
- Entries are dropped on **any** processing error (the connection may hold aborted-transaction state), after **90s idle**, and at a **600s absolute age cap** — threaded through reuse so the embedded session credentials can never outlive their validity even under steady reuse.
- Backfill batches keep the existing fresh-connection lifecycle (rare, long, already chunk-coalesced).
- Trade-off, documented in code: a cached session pins the group's worker for up to 90s after the group drains — bounded, and the next claim of that group wants the same worker warm anyway.

Expected effect: warm batches at ~2s ⇒ several hundred batches/hour/group; combined with #72985 the sink comfortably outruns the producer, and one-snapshot-per-batch churn slows as backlogs stop accumulating.

- TDD (each watched failing first): reuse across consecutive batches (one connect, one preamble), per-group keying, drop-on-error, idle/age expiry, and an age-cap-survives-steady-reuse test that caught a real bug in my first implementation (reuse was resetting the age clock).
- Full `test_processor.py` + `test_consumer.py` + `test_jobs_db.py`: **98 passed**.
- Pre-commit hooks (ruff, format, ty) green.

## Rollback

Revert; behavior returns to fresh-connection-per-batch (slow but correct).

## 🤖 Agent context

Authored by Claude Code (Fable 5). Scope note: an earlier draft coalesced N batches into one DuckLake transaction, but the production trace showed the commit itself is cheap (<0.7s) while the costs are per-session — so this PR ships the smaller mechanism that captures the measured win without multi-batch status-consistency risk. Snapshot-growth hygiene (77k snapshots on the dogfood org) is a separate follow-up on the duckgres/compaction side.


EDsCODE:
Addressed all three review threads in the latest commit:

1. **Idle sessions never expire** (@greptile-apps P1, @hex-security-app, the stamphog refusal) — confirmed real; the "bounded" docstring claim was false without traffic. Now: a daemon sweeper thread (lazy-started on first `store()` so importing processes never spawn it; its sleep is bound to the real `time.sleep` at import so test-patched clocks can't hot-loop it) runs `_evict_stale()` every 30s, closing entries past the 90s idle TTL or the 600s age cap independently of traffic, plus `atexit` cleanup for graceful shutdown. Worker pinning after a group drains is now hard-bounded at `IDLE_TTL + SWEEP_INTERVAL` ≈ 2 minutes. Tested by driving `_evict_stale()` directly under a controlled clock with zero intervening cache traffic (`test_evict_stale_closes_idle_sessions_without_new_traffic`), plus a sweeper-started assertion.
2. **Cached session survives an org transfer** (@veria-ai) — taken as cheap hardening even though project→org transfer isn't a supported product flow today: the cache key now includes the team's *current* `organization_id`, resolved on every batch (same per-batch ORM cost the old fresh-connection code paid). A moved team keys to a fresh connection against the new org's warehouse; the orphaned entry is closed by the sweeper within ~2 minutes. Tested by flipping the resolved org between batches (`test_org_change_invalidates_the_cached_session`).

All watched failing first; full processor + consumer suites green (48).

EDsCODE:
Both follow-ups addressed:

- **@hex-security-app (per-org capacity bound):** the cache now enforces `MAX_SESSIONS_PER_ORG = 4` with LRU eviction on store — a burst of sequentially drained groups can retain at most 4 of an org's connections per consumer pod (12 fleet-wide at 3 pods, versus e.g. the dogfood org's 64-worker cap), on top of the existing 90s idle / 600s age / sweeper bounds. Retention can no longer scale with the number of schemas an org churns through. Test: `test_per_org_session_cap_evicts_the_oldest_entry` (5th schema's store closes the least-recently-used entry, others untouched).
- **@greptile-apps (double Team lookup):** org resolution is hoisted to one per-batch ORM read that feeds both the cache key and `_connect_to_duckgres` (which now takes the resolved org id); the backfill path uses the same hoisted value. Test: `test_cache_miss_resolves_the_team_once`.

Both watched failing first; processor + consumer suites green (50).

PR #73034, opened 2026-07-22
feat(cdp): log URL-validation checks and blocks with team attribution

feliperalmeida:
## Problem

When the SSRF / URL-validation guard in `common/utils/request.ts` refuses a hog function's outbound fetch, the only per-request record is the customer-facing log entry, which is team-scoped and not queryable from the ops side. The low-level `node_request_unsafe` counter has no team or function context, so a blocked request can't be traced back to the destination that made it, or to its owning team. There's also no visibility into the checks themselves, only into the failures.

## Changes

- `cdpTrackedFetch` now takes optional `teamId` / `hogFunctionId` attribution, passed by all four executor services (hog, legacy plugin, native, segment).
- When a fetch is refused by the guard (`SecureRequestError`, `ResolutionError`, `InvalidRequestError`), we increment a new `cdp_http_blocked_requests` counter (labels: reason, template_id) and emit a structured warn log with hostname, teamId, hogFunctionId and templateId.
- Every URL-validation check run is logged with the hostname and the IPs it resolved to. The DNS check runs inside undici's connect flow where no caller context exists, so `cdpTrackedFetch` publishes attribution through a new `fetchAttribution` `AsyncLocalStorage` that the log reads from. The log fires before the verdict, so blocked lookups are recorded with what they resolved to.

> [!NOTE]
> The per-check log fires once per new connection, not per request - undici's keep-alive reuse skips the DNS lookup. Attribution reflects the request that opened the connection.

No new tests: the change is logging and metrics, with no behavior change to the fetch path. `tsc --noEmit` is clean. I ran `hog-executor.service.test.ts` with and without the change: identical results.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

### DIFF
--- a/nodejs/src/cdp/services/hog-executor.service.ts
+++ b/nodejs/src/cdp/services/hog-executor.service.ts
@@ -7,7 +7,15 @@ import { ACCESS_TOKEN_PLACEHOLDER } from '~/common/config/constants'
 import { instrumented } from '~/common/tracing/tracing-utils'
 import { parseJSON } from '~/common/utils/json-parse'
 import { logger } from '~/common/utils/logger'
-import { FetchOptions, FetchResponse, InvalidRequestError, SecureRequestError, fetch } from '~/common/utils/request'
+import {
+    FetchOptions,
+    FetchResponse,
+    InvalidRequestError,
+    ResolutionError,
+    SecureRequestError,
+    fetch,
+    fetchAttribution,
+} from '~/common/utils/request'
 import { TeamManager } from '~/common/utils/team-manager'
 import { tryCatch } from '~/common/utils/try-catch'
 import { UUIDT } from '~/common/utils/utils'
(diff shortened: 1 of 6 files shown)


Piccirello (approved):

nodejs/src/common/utils/request.ts
```diff
@@ -121,6 +134,8 @@ function validateHostnameIPLiteral(hostname: string, allowUnsafe: boolean): void
         return
     }
 
+    logUrlValidationCheck(hostname, [bare])
```
meikelmosby: r u sure you want to log here? This is potentially hundrets of millions of logs emitted just to say the check ran.. what r we gaining from this here?

nodejs/src/common/utils/request.ts
```diff
@@ -178,6 +193,10 @@ async function staticLookupAsync(hostname: string): Promise<LookupAddress[]> {
     } catch {
         throw new ResolutionError('Invalid hostname')
     }
+    logUrlValidationCheck(
```
meikelmosby: same here

nodejs/src/cdp/services/hog-executor.service.ts
```diff
@@ -87,6 +88,49 @@ const cdpHttpRequestTimingRetried = new Histogram({
...
+    error: Error,
+    { url, templateId, teamId, hogFunctionId }: CdpFetchAttribution & { url: string; templateId: string }
+): void {
+    let hostname: string | undefined
+    try {
+        hostname = new URL(url).hostname
+    } catch {
+        // Malformed URL (InvalidRequestError) — hostname stays undefined
+    }
+    cdpBlockedRequests.inc({ reason: error.name, template_id: templateId })
```
meikelmosby: hmm this is increment on templateId which will be something like `template-webhook` is that what you want? this is not the unique templateId that points to a specific hogfunction (although we should probably also not metric this out as cardinality might be huge)

meikelmosby (approved):

nodejs/src/cdp/services/hog-executor.service.ts
feliperalmeida: I'll remove the metric as I think it does not bring much value.

PR #73052, opened 2026-07-23
feat(marketing-analytics): add a ROAS column driven by revenue conversion goals

jabahamondes:
> [!NOTE]
> Sits on top of the conversion-goal stack (#73033 → ... → #72301) and also carries the currency fix from #73047 (cherry-picked, not merged), because ROAS is only correct once revenue is converted to the base currency. This branch is the one to check out to try the whole thing end to end: flags, write endpoints, MCP tools, currency conversion, and now ROAS. When #73047 and the stack land on master, a rebase drops the duplicated currency commit.

## Problem

Marketing analytics can show you spend per channel and conversions per channel, but not the one ratio a marketer actually optimizes: revenue over spend. The pieces were all there (spend per channel, a per-goal value column, currency conversion from #73047) but nothing tied them together.

## Changes

A `ROAS` column: the summed value of every conversion goal marked `counts_as_revenue`, divided by that channel's spend.

It reuses the exact pattern the existing "cost per conversion" column already uses (`campaign_costs.total_cost` over a `nullif`'d conversion value from the unified goals CTE), just inverted and summed across the revenue-flagged goals. Built in `conversion_goals_aggregator.get_conversion_goal_columns`, so it flows to the table through the same `all_columns.update(...)` every other conversion column does.

Three deliberate boundaries:

- **Hidden until revenue goals exist.** No goal flagged `counts_as_revenue` means no column. The table doesn't grow an empty ROAS column for teams that haven't set one up. The existing table and aggregator snapshots are unchanged, which is the proof it's inert by default.
- **Only when spend is available.** ROAS needs spend as its denominator, so it rides the same `include_cost_per` guard the cost-per column uses. At UTM levels where platform cost can't be attributed, both are correctly absent.
- **Revenue is already in the base currency.** Thanks to #73047, each goal's summed value is converted at its event-day rate before it ever reaches this column. Without that, ROAS would divide a mixed-currency numerator by a converted denominator.

No new column metadata was needed. `ROAS` falls through `to_marketing_analytics_data` to the unit / higher-is-good defaults, which is exactly right for a ratio where more is better.

3 new unit tests in `test_conversion_goals_aggregator.py`, run by me (Claude):

- **Numerator sums only the revenue goals.** Three goals, two flagged revenue at indices 1 and 2, one plain signup at 0. Asserts the ROAS expression references the value columns of 1 and 2 and not 0. This is the core regression: if the flag filter breaks, a signup count would get added into revenue.
- **Absent when no goal is revenue.** Guards the "don't show an empty column" boundary.
- **Absent without campaign_costs** (`include_cost_per=False`). Guards the denominator boundary at UTM levels.

Full `test_conversion_goals_aggregator.py` (30), `test_marketing_analytics_table_query_runner.py` and `_business.py` (79) pass, snapshots unchanged. `ruff` clean; `mypy` clean on the file (the one printed error is a pre-existing unrelated one in `personhog_client`).

I did not add an end-to-end ClickHouse test that asserts a computed ROAS number. The column's construction is covered by the unit tests, and its propagation into the final SELECT is the same generic `update()` path already exercised by the cost-per columns. An end-to-end test would have needed a full cost + conversion fixture to cover a path that isn't new.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Holding until the read-side columns settle and the dashboard redesign decides how they surface.

## 🤖 Agent context

I directed this, Claude wrote it. Invoked `/writing-tests`, which drove the test list toward the one assertion that matters (numerator sums only revenue goals) and away from re-testing the round/nullif plumbing the cost-per column already covers.

This is the first of the two read-side metric columns. CAC is deliberately not in this PR. It turned out to be structurally larger than ROAS: "new customers" means each person's first-ever conversion, which needs a historical first-time-per-user scan that the current per-window precompute doesn't provide (`first_time_for_user` isn't wired into the marketing processor at all yet). That deserves its own design pass rather than being bolted on here, so ROAS ships on its own.


jabahamondes:
Superseded by #83263 — same logic reapplied on current master as a single signed commit (214 lines vs 19,118, of which 18,181 were regenerated artifacts). The conversion-goal write endpoints this branch carried landed separately, which is what made it conflict.

PR #73058, opened 2026-07-23
feat(workflows): add manual_invocations CDP route to run a workflow on demand

HaynesPostHog:
## Problem

The Conversations "quick actions" feature ([#72792](https://github.com/PostHog/posthog/pull/72792)) lets a support agent run a workflow against a ticket on demand. None of the existing HogFlow invocation paths fit: `scheduled_invocations` is gated to `schedule` triggers, the public webhook path only accepts `webhook`/`manual`/`tracking_pixel` triggers, and `invocations` is a single-node test runner (no real full-graph run).

## Changes

Adds `POST /api/projects/:team_id/hog_flows/:id/manual_invocations` to the CDP API. Modeled on `postHogflowScheduledInvocation`, but:

- accepts a caller-synthesized `{ globals: { event, person?, groups? }, variables? }` (reusing `postHogflowInvocation`'s globals handling + `groupsManager.getGroupsForEvent` resolution),
- gates on `status === 'active'` instead of trigger type — so **any** active workflow can be run, including event-triggered ones,
- builds the invocation with `createHogFlowInvocation` and enqueues via the existing `hogflowQueue`, so it's a real full-graph run (waits/delays/branches all work), not a synchronous single-node test.

Django calls it via `create_hog_flow_manual_invocation` in `plugin_server_api.py` (that helper lands in #72792).

> [!NOTE]
> Split out of #72792: that PR carries a DB migration, and the deploy-ordering check requires migrations and nodejs service changes to be in separate PRs. This route has no dependency on that migration and can merge independently. Until it's deployed, the quick-action workflow-run path in #72792 degrades gracefully (the run endpoint returns a 502).

Added `nodejs/src/cdp/cdp-api.test.ts` cases: an active event-triggered workflow queues (200), an inactive workflow is rejected (400), and missing event globals error. Ran the CDP API test suite locally.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — @HaynesPostHog directed the work; I (Claude, via PostHog Code) implemented it. Split from #72792 to satisfy the migration/service-separation check.


nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +744,74 @@ export class CdpApi {
...
+
+            // Resolve groups server-side from the event's $groups when the caller didn't supply them,
+            // so group-property conditionals branch correctly (mirrors postHogflowInvocation).
+            if (!globals.groups || Object.keys(globals.groups).length === 0) {
+                globals.groups = await this.groupsManager.getGroupsForEvent(
+                    team.id,
+                    globals.event.properties,
+                    `${this.config.SITE_URL ?? 'http://localhost:8000'}/project/${team.id}`
+                )
+            }
```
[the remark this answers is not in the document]
HaynesPostHog: valid — fixed in 1c34420. The handler now rejects non-object `event.properties` with a 400 and rebuilds the event server-side (uuid/distinct_id/timestamp/properties/elements_chain defaults), so a malformed caller payload can never be durably enqueued. Added a regression test for the null-properties + non-empty-groups case that previously slipped through.

nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +744,74 @@ export class CdpApi {
...
+
+            const hogFlow = await this.hogFlowManager.getHogFlow(id)
+            if (!hogFlow || hogFlow.team_id !== team.id) {
+                return res.status(404).json({ error: 'Workflow not found' })
+            }
+
+            if (hogFlow.status !== 'active') {
+                return res.status(400).json({ error: 'Workflow must be active' })
+            }
```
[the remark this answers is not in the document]
HaynesPostHog: valid — fixed in 1c34420. The route now applies the pipeline's gates before enqueueing: `hogWatcher.getEffectiveState` (400 'Workflow is disabled due to repeated failures' when the circuit breaker tripped) and `checkHogFlowQuotaLimits` (429 when the team is over quota), so the agent gets clear feedback instead of a silently-accepted run. Added a watcher-disabled regression test.

nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +745,111 @@ export class CdpApi {
...
+            const filterGlobals = convertToHogFunctionFilterGlobal({
+                event,
+                person,
+                groups,
+                variables: resolvedVariables,
+            })
+
+            const invocation = createHogFlowInvocation(triggerGlobals, hogFlow, filterGlobals)
+
+            await this.hogflowQueue.queueInvocations([invocation])
```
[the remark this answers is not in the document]
HaynesPostHog: valid — fixed in b1a5de9. The route now draws from the same keyed rate-limit bucket as the event pipeline (`KeyedRateLimiterService` with the identical `hog-rate-limiter` name and `CDP_RATE_LIMITER_*` config, keyed by workflow id), so manual runs and event-driven invocations share one token bucket rather than the route being a bypass. Exhausted bucket → 429, nothing enqueued. Added a regression test.

HaynesPostHog:
## CI status

All checks are green, including the Node.js test shards covering the new `manual_invocations` route.

Review findings addressed on this route, each with a regression test:
- **Poison-pill enqueue**: the event payload is validated and rebuilt server-side (`uuid`/`distinct_id`/`timestamp`/`url`/`properties`/`elements_chain`), so a malformed body gets a 400 instead of entering the queue.
- **Safety-gate bypass**: manual runs now pass through the same HogWatcher state check (400 when the watcher has disabled the flow) and quota check (429) as pipeline-triggered invocations.
- **Rate limit bypass**: manual runs draw from the same per-workflow `hog-rate-limiter` token bucket as the event pipeline and return 429 when it's exhausted, so this route can't be used to sidestep per-flow rate limiting.

nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +758,117 @@ export class CdpApi {
         }
     }
 
+    // On-demand "run this workflow now" against a caller-synthesized event (e.g. a support agent
+    // running a workflow against a ticket). Unlike the scheduled/webhook paths this is not gated on
+    // trigger type — any active workflow can be run — but it still queues the full graph through the
+    // same cyclotron queue, so waits/delays/branches all work. Django gates auth/team-scoping.
+    private postHogflowManualInvocation = async (req: ModifiedRequest, res: express.Response): Promise<any> => {
```
mayteio, line 784: This route falls through to the shared `INTERNAL_API_SECRET` middleware, which `nodejs/src/common/api/middleware/internal-api-auth.ts` asks new endpoints not to do, in its own module docstring:

> Do not gate new endpoints on this shared secret - prefer a scoped JWT (see recording-api/auth.ts) or a dedicated per-purpose secret. See .agents/security.md ("Secrets & key management").

The pattern it points at is already implemented in this class, a few lines above: `reschedule_parked` is listed in `SCOPED_AUTH_PATH_SUFFIXES` and verified against `WORKFLOWS_RESCHEDULE_JWT_SECRET`, with `cdp/config.ts:175` noting it is "deliberately NOT the fleet-wide INTERNAL_API_SECRET... empty in prod means the route fails closed until provisioned". Same caller, same product, same shape of route.

It arguably matters more here than on `reschedule_parked`. This route runs an arbitrary active workflow with its stored secrets for any team, so a token pinned by claims to one team and one flow id limits a leak to a single workflow, where the shared secret reaches every workflow on the fleet. That is exactly the blast-radius reasoning in the security doc.

Worth resolving before merge rather than after, since the Django side in #74031 changes with it.

nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +758,117 @@ export class CdpApi {
...
+            const hogFlow = await this.hogFlowManager.getHogFlow(id)
+            if (!hogFlow || hogFlow.team_id !== team.id) {
+                return res.status(404).json({ error: 'Workflow not found' })
+            }
+
+            if (hogFlow.status !== 'active') {
+                return res.status(400).json({ error: 'Workflow must be active' })
+            }
+
+            // Match the pipeline's safety gates: don't run a flow the watcher circuit breaker has
```
mayteio: The pipeline documents its stages at `hog-flow-invocation-pipeline.service.ts:42` as `load hogflows → execute filters → watcher state → rate limit → quota → masking → metrics`. This re-applies watcher state, rate limit, and quota. It does not apply filters or masking, and the comment does not mention that.

Filters is the one worth an explicit decision. Combined with the deliberate choice to accept any active workflow regardless of trigger type, an event-triggered workflow with filters gets its graph executed against a synthesized event those filters would have rejected. That may well be intended (a manual run is manual), but a workflow author wrote those conditions expecting them to hold, and right now the divergence is invisible.

Masking raises the same question in a milder form: a manual run bypasses masking suppression entirely.

Could the comment say which gates are deliberately skipped and why? Otherwise the next person to touch this reads "match the pipeline's safety gates" and reasonably assumes parity.

nodejs/src/cdp/cdp-api.ts
```diff
@@ -740,6 +758,117 @@ export class CdpApi {
...
+            // Rebuild the event server-side rather than trusting the caller's shape verbatim: this
+            // payload is durably enqueued, and a malformed event (e.g. null properties) would
+            // poison the shared hogflow queue and crash-loop the worker for the whole partition.
+            const event: HogFunctionInvocationGlobals['event'] = {
+                uuid: typeof rawEvent.uuid === 'string' ? rawEvent.uuid : new UUIDT().toString(),
+                event: rawEvent.event,
+                distinct_id: typeof rawEvent.distinct_id === 'string' ? rawEvent.distinct_id : `workflow-${hogFlow.id}`,
+                timestamp: typeof rawEvent.timestamp === 'string' ? rawEvent.timestamp : DateTime.now().toISO(),
+                url: typeof rawEvent.url === 'string' ? rawEvent.url : '',
+                properties: rawEvent.properties ?? {},
```
mayteio, line 887: The reasoning in the comment above is right, and worth extending: this payload is durably enqueued, and malformed content poisons the shared hogflow queue for the whole partition, so it is a multi-tenant failure mode rather than a caller-only one.

The mitigation stops short of two things in that same enqueued state:

- `event.properties` is passed straight through after only the `isPlainObject` check, with no size cap. Every other field is rebuilt or bounded. Today it is bounded in practice by whatever the Django ticket property builder emits, but nothing here enforces that.
- `variables` (`variables ?? rawGlobals?.variables ?? {}`) is not validated at all and lands in the invocation state via `createHogFlowInvocation`'s `mergedVariables`.

No current caller sends `variables` (#74031 passes `{event, groups}` only), so this is hardening rather than a live gap. Raising it because the route is new and the next caller will not know the validation stops where it does.

mayteio (commented):
Reviewed as the dependency for #74031. The care taken here is visible: re-applying the watcher circuit breaker, quota limiting, and the per-workflow rate limiter when enqueueing directly is easy to skip, and the server-side event rebuild is the right instinct. I also verified the rate limiter really does share a bucket with the event path (both `'hog-rate-limiter'`, same config), so that comment holds.

Three comments. The auth one I would want settled before this merges, since it also determines the shape of the Django helper in #74031 and is cheaper to fix now than to retrofit across both.

nodejs/src/cdp/cdp-api.ts
mayteio: Following up with something more useful than "use a JWT": I built the `reschedule_parked` one in #71293, so I can tell you what it actually costs.

The code side is small. Mint the key setting in `cdp/config.ts` (comma-separated, newest first, so rotation works), add the route suffix to `SCOPED_AUTH_PATH_SUFFIXES` in `internal-api-auth.ts`, and verify in the handler, holding it null when the key is absent so the route fails closed rather than open. Django mints per-call tokens with claims pinning them to the team and the flow. For this route those claims are worth more than they are for mine, since a leaked token would be scoped to one workflow instead of every workflow on the fleet.

The part that actually takes time is provisioning the key per region with the empty-in-prod default. That is usually what makes reusing the existing header look like the cheaper option. Happy to help with that side if it is the blocker, since I have been through it once already.

Reusing `WORKFLOWS_RESCHEDULE_JWT_SECRET` itself would not be right, a key per audience is the point, but #71293 is a complete worked example to copy from.

HaynesPostHog: $Fixed. The route no longer falls through to the shared secret — it's in \`SCOPED_AUTH_PATH_SUFFIXES\` and the handler verifies a dedicated \`WORKFLOWS_MANUAL_INVOCATION_JWT_SECRET\` JWT with claims pinned to team + workflow, mirroring \`reschedule_parked\` (new \`PosthogJwtAudience.WORKFLOWS_MANUAL_INVOCATION\`, null when unprovisioned so it fails closed). Django mints it in #74031. It's a separate per-purpose key from reschedule, per your point that the key-per-audience is the whole point.

nodejs/src/cdp/cdp-api.ts
HaynesPostHog: $Fixed. The comment now spells out which gates a manual run re-applies (watcher, quota, rate limit) and which it deliberately skips and why: trigger filters (the caller already decided to run it, so the run is unconditional by design) and masking/dedup suppression (a manual run should fire even if an event-driven one was recently masked). So the next person doesn't read "match the pipeline's safety gates" and assume parity.

nodejs/src/cdp/cdp-api.ts
HaynesPostHog: $Fixed both. Added a 128 KB cap (\`exceedsCap\`) on the two pass-through fields — \`event.properties\` and the resolved \`variables\` (checked at the \`variables ?? rawGlobals?.variables\` resolution point so both the top-level and nested paths are covered) — returning 400 before anything is enqueued. Everything else on the event was already rebuilt/bounded; this closes the two that weren't.

nodejs/src/cdp/cdp-api.ts
HaynesPostHog: $Thanks, that worked example is exactly what I followed — same shape as \`reschedule_parked\` (suffix in \`SCOPED_AUTH_PATH_SUFFIXES\`, verify in the handler holding null when the key is absent, Django minting per-call claims for team + flow), with a distinct \`WORKFLOWS_MANUAL_INVOCATION_JWT_SECRET\` rather than reusing yours. The one thing I can't do from here is provision the key per region with the empty-in-prod default — flagging that to Ben as the remaining ops step, and I may take you up on the offer since you've been through it.

PR #73065, opened 2026-07-23
perf(data-warehouse): probe target table for existence instead of information_schema

EDsCODE:
## Problem

The duckgres sink's `_table_exists()` runs `SELECT 1 FROM information_schema.tables` on every non-first batch (via `_plan_batch_operation`, to choose create vs insert/merge). On a large DuckLake catalog this forces duckgres to re-materialize its whole-catalog `information_schema` compat view — every current table and column at the current snapshot.

Traced end to end on the dogfood org's catalog (mw-prod-us, 2026-07): **3,173 tables, 80,235 snapshots, 577,076 `ducklake_column` rows** (88% superseded version history), and the `ducklake_column` index is `(table_id, …)`-leading — so an "all current columns" load can't use it and scans the whole history.

- It's **not a lock** — sampling `pg_stat_activity`/`pg_locks` on the catalog Postgres during live sink activity showed zero blocking. It's *work*: DuckDB caches the catalog and re-loads it whenever the snapshot advances; the sink commits a snapshot **per batch**, so the cache is invalidated on every batch and each check re-scans the bloated catalog.
- Measured under identical concurrent write load: `information_schema.tables` = **1.3s idle → ~48s under load** (44.8–52.1s every iteration, one concurrent writer was enough). This is a **shared per-apply tax paid by every table**, which is why sink throughput was flat (~140/hr) regardless of which table a batch hit — and why neither the earlier session-reuse nor co-claim PRs moved it (the cache is invalidated by *other* sessions' commits).

## Change

Replace the `information_schema` enumeration with a **single-table `LIMIT 0` probe** (`SELECT 1 FROM <schema>.<table> LIMIT 0`). It loads only that one table's metadata via the catalog's `table_id`-indexed single-table path, so it stays cheap even under concurrent snapshot commits:

| under identical concurrent write load | idle | loaded |
|---|---|---|
| `information_schema.tables` (old) | 1.3s | **~48s** |
| `LIMIT 0` probe (new) | 0.11s | **0.59s** (max 0.90s) |

duckgres reports a missing table as a generic `XX000` carrying DuckDB's stable `"... does not exist"` catalog message (verified via psycopg — it is **not** a typed `UndefinedTable`), so the probe matches that message for "absent" and **propagates anything else** — a transient error read as "table absent" would wrongly pick the create path (a benign, self-healing `CREATE`-then-retry, but avoid it). The probe runs on the autocommit connection outside any transaction, so a raised error is inert.

Kept the per-connection positive existence cache (`WeakKeyDictionary`): a table that exists stays existing for the connection's life (the sink never drops tables mid-run), so repeat batches skip the probe entirely and the not-found match only runs on a table's **first encounter per connection**.

- TDD: probe-succeeds→exists, probe-not-found→absent, **other-error propagates (not read as absent)**, cache-hit skips the probe, per-table granularity, per-connection isolation, negatives-not-cached — each watched failing first, then passing (`test_processor.py`, 30 total).
- Direct production experiments established causation: idle vs concurrent-load repro (1.3s→48s for `information_schema`, 0.11s→0.59s for the probe), catalog Postgres `pg_locks`/index inspection, and the exact psycopg error signature. All test artifacts (a temporary user, a scratch schema) were removed.
- Pre-commit hooks (ruff/format/ty) clean.

## Follow-ups (separate)

- **Snapshot/version expiry** to shrink the 577k-row catalog — makes even the rare create-path check cheap and helps storage/billing (this is the part of the earlier "compaction helps throughput" idea that's actually true — metadata expiry, not S3 file compaction).
- File compaction remains **storage-only** for throughput.

## 🤖 Agent context

Authored by Claude Code (Fable 5) from a multi-day production investigation. Notably this **corrected an earlier hypothesis**: the throughput bottleneck is not S3 file-count bloat (compaction) but this per-batch `information_schema` re-materialization — established by reproducing the 1.3s→48s swing under concurrent snapshot commits, confirming on the catalog Postgres it's re-read work (not lock contention), and then validating that the single-table probe stays sub-second under the same load.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline_v3/duckgres/processor.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline_v3/duckgres/processor.py
@@ -8,6 +8,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from time import sleep as _real_sleep  # captured before any test patches this module's `time`
 from typing import Any, Literal
+from weakref import WeakKeyDictionary
 
 from django.conf import settings
 from django.db import close_old_connections
@@ -648,7 +649,17 @@ def _plan_batch_operation(
     raise PermanentBatchApplyError(f"Unsupported Duckgres sync type: {batch.sync_type}")
 
 
+# Per-connection cache of tables confirmed to exist: the information_schema
+# probe below is expensive on a large, high-churn DuckLake catalog and the sink
+# runs it every non-first batch. Only positive results are cached (a table may
+# be created between batches); GC'd with the connection, so reuse re-checks.
+_existing_tables: WeakKeyDictionary[psycopg.Connection[Any], set[tuple[str, str]]] = WeakKeyDictionary()
+
+
 def _table_exists(conn: psycopg.Connection[Any], duckgres_schema: str, duckgres_table: str) -> bool:
+    known = _existing_tables.setdefault(conn, set())
+    if (duckgres_schema, duckgres_table) in known:
+        return True
     cursor = conn.execute(
         """
         SELECT 1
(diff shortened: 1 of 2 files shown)


benben (approved):

EDsCODE:
Addressed the P1 (@greptile-apps / @stamphog): the missing-table detection now anchors to DuckDB's specific `"Table with name … does not exist"` message rather than a bare `"does not exist"` substring, so a missing schema/catalog/secret encountered while resolving the query propagates as a real error instead of being read as "table absent" and routed to the create path. Added `test_missing_schema_error_propagates_not_treated_as_table_absent` (a `Schema with name … does not exist` error now raises). 8 existence tests pass.

PR #73068, opened 2026-07-23
feat(tasks): default unpinned loops to glm 5.2

charlesvien:
## Problem

A loop that doesn't pin a model falls back to `claude-sonnet-5` at fire time. Loops are unattended recurring background runs where per-run cost matters, so the free-tier gateway model GLM 5.2 is the better default. The "blank model means PostHog picks per fire" contract stays as is.

## Changes

- `DEFAULT_MODEL_BY_RUNTIME_ADAPTER` now resolves the claude adapter to `@cf/zai-org/glm-5.2` (codex stays `gpt-5`).
- Fire time drops a stored `reasoning_effort` the resolved model doesn't support. GLM only accepts high and max, so a legacy blank-model loop saved with medium fires with auto instead of an invalid combination.
- No serializer changes: new writes with a blank model are already validated against the effective default, which now means GLM's supported efforts.

A companion PostHog Code PR updates the loop form so the effort dropdown only offers what the effective model supports and the detail view shows the real fire-time default.

- New parameterized test in `test_loop_runs.py` covers fire-time model and effort resolution (blank model resolves to GLM, unsupported stored effort degrades to auto, pinned models keep their validated effort). The effort-degradation path had no coverage before.
- New parameterized test in `test_loops_api.py` pins create-time validation against the effective default model (high/max accepted, medium rejected with `attr=reasoning_effort`). Blank-model effort validation had no API-level coverage before.
- Ran the loop backend suites locally: `test_loop_runs.py`, `test_loops_api.py`, `test_loop_lifecycle.py`, `test_loop_service.py`, 134 passed.
- No manual testing beyond the automated suites.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A

## 🤖 Agent context

Authored with Claude Code at Charles's direction. Decisions: kept blank-model rows unset instead of pre-pinning GLM on new loops, so the default keeps improving without rewriting stored loops, and made fire time degrade an incompatible stored effort to auto rather than failing or blocking the run. No repo skills were invoked.

### DIFF
--- a/products/tasks/backend/facade/loops.py
+++ b/products/tasks/backend/facade/loops.py
@@ -795,6 +795,37 @@ def _validate_context_visibility(visibility: str, context_target: dict | None) -
         raise LoopValidationError("A loop attached to a context must have team visibility.")
 
 
+def _validate_model_config_for_update(loop: Loop, data: dict) -> None:
+    """Judge model/effort validity on the merged post-update state, with the stored row filling in
+    fields the PATCH omits (the write serializer can't, it only sees submitted fields). Skipped
+    when the PATCH doesn't touch model config, so a stored legacy combination keeps degrading at
+    fire time instead of blocking unrelated edits."""
+    if not any(key in data for key in ("runtime_adapter", "model", "reasoning_effort")):
+        return
+    # Deferred so the model catalog stays off this module's import path (see `validate_loop_write`).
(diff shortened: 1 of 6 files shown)


tatoalo:
@charlesvien _sad_noises_

PR #73075, opened 2026-07-23
feat(data-warehouse): expand OpenAI and Anthropic usage/cost coverage

Gilbert09:
## Problem

The OpenAI and Anthropic warehouse sources didn't cover every usage/cost surface their admin APIs expose. An audit against the current OpenAI Administration API and Anthropic Admin API turned up a handful of data-bearing endpoints we weren't syncing, most notably Anthropic's Claude Code analytics (per-developer productivity plus per-model cost), which is a whole dataset an org running Claude Code would want in the warehouse.

## Changes

Adds the missing data-bearing endpoints to both sources. No config-field or schema-type changes, so nothing downstream (generated types, migrations) moves.

**OpenAI** — completes the `/usage/*` catalog:

- `usage_web_search_calls` and `usage_file_search_calls`, the two hosted-tool usage endpoints. Both are billable and both ride the existing generic bucket paginator, so only the endpoint catalog and canonical descriptions change. `file_search` groups by `vector_store_id` (no `model` dimension) since requesting an unsupported dimension 400s the live API.

**Anthropic**:

- **Claude Code analytics**, split into two tables because they sit at different grains:
  - `claude_code_analytics` — one row per user per day: sessions, lines of code, commits, PRs, and tool-action accept/reject counts.
  - `claude_code_model_breakdown` — one row per user per day per model: token usage and estimated cost.
- `service_accounts` — workspace-scoped, resolves the `service_account_id` dimension that already shows up on the usage report. Rides the existing workspace fan-out.
- `inference_geo` column on `cost_report` (the data-residency dimension was being parsed but dropped).

The Claude Code endpoint doesn't fit the framework's parent/child resolve fan-out: it windows on a single required `starting_at` day (not a range) and page-cursors within that day, and there's no parent API resource to resolve days from. So it gets a small custom `ClaudeCodeDayPaginator` that walks each day's pages, advances `starting_at` to the next day, and stops once it passes today. The start day comes from the incremental watermark (or a 2025 launch floor on a full refresh), and the paginator's resume state (`{date, cursor}`) checkpoints mid-fan-out. `inference_geo` is deliberately kept out of the cost surrogate key so existing rows' ids stay stable.

Automated only — I (Claude) can't hit the live admin APIs from here.

- 151 unit tests pass across both sources (77 Anthropic, 74 OpenAI), `ruff` clean, `ty` typecheck clean.
- New tests worth calling out (each catches something no existing test did):
  - OpenAI: the two new endpoints' `group_by` composition, incl. a guard that `file_search` never requests `model` (that would 400 the live API).
  - Anthropic Claude Code: the day-by-day fan-out issues one windowed request per calendar day and paginates within a day; the model-breakdown table explodes to one row per model with distinct ids; resume restarts on the saved day + page cursor; the paginator clamps a future watermark to today.
  - Anthropic service accounts: the workspace fan-out injects `workspace_id` so the composite key is never null.
  - `cost_report.inference_geo` is surfaced, and its id stays stable now that the column exists.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The posthog.com docs render the supported-tables list from the canonical descriptions via `<SourceTables />`, so the new tables show up automatically once this ships. No manual doc edit needed.

## 🤖 Agent context

Tom directed this: audit the OpenAI and Anthropic sources for missing usage/cost coverage, then add everything useful and make new schemas incremental where the API supports it. I (Claude, via Claude Code) ran the audit by fetching the current OpenAI Administration and Anthropic Admin API references, then implemented against the `rest_source` framework versions of both sources on master. Invoked the `implementing-warehouse-sources` skill for conventions.

- Claude Code split into two tables rather than one. Core metrics are per-day and token/cost are per-model. One table would either duplicate the core metrics across a day's models or bury per-model cost in a nested column.
- Skipped the pure config/RBAC endpoints (certificates, groups/roles, rate-limit config, spend-alert thresholds, org `me`) — they're settings, not data anyone would analyze. Happy to add if wanted.
- Skipped Anthropic's beta `speed` usage dimension (needs a beta header) and the separate Enterprise Analytics API (different credential, not the admin key this source uses).


Gilbert09:
Brought this branch up to date with `master` and got CI fully green.

- Merged `master` in (server-side, two-parent merge) — no conflicts, and the PR diff stays scoped to the OpenAI/Anthropic source files (the earlier query-snapshot `.ambr` changes dropped out once they matched `master`).
- Fixed one real type error CI would have blocked on: in `anthropic_source`, `paginator` was first bound to `ClaudeCodeDayPaginator`, so `mypy` inferred that concrete type and rejected the later `AnthropicCursorPaginator` rebinds in the other branches. Annotated it to the shared `BasePaginator`. (`ty` didn't flag this; CI runs `mypy`.)
- All checks pass; `ruff`, `mypy`, and the 151 source unit tests are green locally too.

The PR is `mergeable` with no conflicts — the remaining `blocked` state is just the required review approval.

PR #73076, opened 2026-07-23
fix(cdp): guard against malformed invocation globals in cyclotron-hog worker

meikelmosby:
## Problem

A single malformed `cdp_cyclotron_hog` invocation — one whose `state.globals` is present but missing both `project` and `event` — crash-looped the cyclotron-hog worker and stalled the partition it owned. In `loadHogFunctions`, two enrichment calls run in one `Promise.all`: `addGroupsToGlobals` dereferences `globals.project.id`, and the person-stub branch dereferences `globals.event.distinct_id`. With those fields absent, both throw `TypeError: Cannot read properties of undefined`, which surfaces as an unhandled rejection → the worker exits → it restarts, re-reads the same offset, and crashes again. One poison-pill message takes down the consumer and stops the partition from draining.

The dereferences themselves have been unchanged for months — this is a latent bug triggered by data, not a code rollout.

Root cause context: every invocation deserialize path (Kafka/WarpStream, cyclotron postgres v1/v2, and the rerun rehydration path) casts the parsed JSON to its type with `as` and performs **no shape validation**. The live producer always sets `project`/`event`, but a message that arrives without them — a legacy/differently-shaped queued message, or a rerun rehydrated from an intentionally-minimized ClickHouse `invocation_globals` snapshot via a blind cast — flows straight into the unguarded derefs.

## Changes

Two layers: harden the read sites so a bad message can't crash-loop the worker, and validate the shape at the most plausible source instead of blind-casting it.

- `cdp-cyclotron-worker.consumer.ts`: `loadHogFunctions` validates `globals.project`/`globals.event` up front. A malformed invocation is logged, `captureException`'d, and routed to the existing dequeue path so it's dropped and the partition drains — instead of throwing out of the `Promise.all`. The guard sits before both crash sites.
- `groups-manager.service.ts`: `addGroupsToGlobals` skips enrichment (sets `groups = {}`) when `project`/`event` is absent, as defense-in-depth (it's also called from the hogflow path).
- `schema/cyclotron.ts` + `rerun-paginator.service.ts`: replace the blind `parsedGlobals as HogFunctionInvocationGlobalsWithInputs` cast in the rerun rehydration path with a zod `safeParse`. The new `HogFunctionInvocationGlobalsSchema` validates the invariants the worker relies on (`project.{id,url}`, `event.{distinct_id,properties}`) and stays permissive (`passthrough`) on everything else so drift on non-critical fields never rejects a good message. A row that fails validation is logged and skipped (counted as `rehydrate_failed`) rather than re-enqueued as a poison pill.

Still open (flagged, not in scope here): the Kafka/postgres deserialize paths remain unvalidated casts, and the equivalent `loadHogFlows` path in the hogflow worker lacks the same guard.

Test-first for the worker guard: added failing tests, confirmed they reproduced the exact `TypeError`s, then applied the fix and confirmed green.

- `groups-manager.service.test.ts` — two parameterized cases (missing `project`, missing `event`) asserting `addGroupsToGlobals` resolves with empty groups. Before the fix they failed with `Cannot read properties of undefined (reading 'id')` and `(reading 'properties')`; after, the suite passes. **Regression caught:** a partial `globals` reaching enrichment throws — no existing test exercised it, because the fixture always backfills `project`/`event`.
- `schema/cyclotron.test.ts` — 8 cases: valid globals accepted; missing `project`/`event` and wrong-typed critical fields rejected; non-critical fields passed through. **Regression caught:** loosening the schema (or switching to strict) so a poison-pill shape validates, or so reruns silently drop `inputs`/`request`.
- `cdp-cyclotron-worker.test.ts` — two cases asserting a malformed invocation is dequeued (DLQ'd), not executed. **Regression caught:** the worker crash-looping on a poison pill instead of dropping it. This test is DB-backed (`resetTestDatabase`/`createHub`) and could not run in the sandbox (no Postgres) — it mirrors the existing `dequeue if the hog function cannot be found` pattern and will run in CI.

Typecheck clean on all changed files; eslint clean. Runnable suites (`groups-manager`, `schema/cyclotron`) pass locally.

## 🤖 Agent context

Investigated from a production incident thread: a cyclotron-hog worker stuck Progressing, one partition backed up. Confirmed the thesis before fixing — wrote the failing unit test that reproduces the crash, watched it fail with the reported stack traces, then implemented the guard and watched it pass. Traced the produce→serialize→enqueue→dequeue→deserialize path to establish that no deserialize step validates shape; on a follow-up pass, replaced the rerun path's blind `as` cast — the most plausible producer of malformed re-enqueued state — with schema validation. Chose to keep the schema permissive (validate only the dereferenced invariants) so it can't reject valid messages on incidental drift.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BL3175D6C/p1784789996469569?thread_ts=1784789996.469569&cid=C0BL3175D6C)*


dmarchuk (approved):

Samielakkad (commented):
Retracted.

PR #73077, opened 2026-07-23
feat(capture): accept recording sent_at in request body

marandaneto:
## Problem

The `/s/` session recording endpoint only reads `sent_at` from the query string. Browser clients should be able to keep POST metadata in the request body while older SDKs continue using the existing numeric query parameter.

## Changes

Session recording items now accept an optional ISO 8601 string in the `sent_at` body field, matching `/batch/`. A valid value on the first recording item takes precedence, while `sent_at` and `_` query values remain as fallbacks. A tolerant deserializer ignores invalid or mistyped body values rather than making previously accepted payloads fail hydration.

Before:

```mermaid
flowchart LR
    A[POST /s query sent_at] --> B[Recording processing context]
    class A phYellow;
    class B phBlue;
```

After:

```mermaid
flowchart LR
    A[First recording body sent_at] --> C{Valid ISO value?}
    B[Query sent_at or underscore] --> C
    C --> D[Recording processing context]
    class A,B phYellow;
    class C phGray;
    class D phBlue;
```

> [!NOTE]
> This PR preserves the established `/s/` body shape, which is a recording object or an array of recording objects. Ideally, `/s/` would eventually use a `/batch/`-style envelope that permits request-level fields such as `sent_at` and `api_key`, and potentially shared recording metadata such as `distinct_id`, without duplicating them across events. That protocol change is intentionally out of scope for this backward-compatible addition.

- `cargo test -p capture --lib events::recordings::tests::test_raw_recording_deserialization`
- `cargo test -p capture --lib payload::recordings::tests`
- `cargo test -p capture --test recordings --no-run`
- `cargo clippy -p capture --all-targets -- -D warnings`
- `hogli ci:preflight --fix`

The added coverage catches regressions where body `sent_at` is ignored, query values override valid body values, or invalid and mistyped body values prevent legacy query fallback. The Kafka-backed integration test compiled locally but could not execute because the local Kafka service was unavailable.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation update is needed. This is a backward-compatible transport addition for SDKs, and legacy query formats remain supported.

## 🤖 Agent context

Implemented with the Pi coding agent using the `/writing-tests`, `/running-ci-preflight`, and `/pr` skills, with a read-only reviewer agent checking compatibility. The review identified that a plain typed string field would reject body values that were previously ignored, so the implementation uses an optional string contract with tolerant deserialization and parses only valid ISO values.

### DIFF
--- a/rust/capture/src/events/recordings.rs
+++ b/rust/capture/src/events/recordings.rs
@@ -16,7 +16,7 @@ use chrono::DateTime;
 use common_types::{CapturedEvent, HasEventName};
 use limiters::redis::RedisLimiter;
 use metrics::{counter, histogram};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::{json, Value};
 use time::{format_description::well_known::Iso8601, OffsetDateTime};
 use tokio::time::Instant;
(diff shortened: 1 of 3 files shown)


marandaneto:
follow up https://github.com/PostHog/posthog/pull/72842
this is the only endpoint not accepting sent_at as part of the request payload and the only one missing to kill the sent at query params from the JS SDK so we can remove it altogether

turnipdabeets (approved):

fasyy612 (approved):

PR #73080, opened 2026-07-23
fix(data-warehouse): bound partition memory for OOM-prone import tables

Gilbert09:
## Problem

Some data warehouse import tables repeatedly OOM-kill their sync pods (surfaced as the `dwh_pod_heartbeat_timeout` event). After the in-place auto-repartition controller shipped it handled the bulk of these, but a long tail remained, from three distinct mechanisms:

- **Perpetual repartition loop.** A table whose in-place rewrite keeps failing never converges. `MAX_REPARTITION_ATTEMPTS` only bounds attempts against a *single* pending target, but pre-extraction detection re-arms a fresh target every sync, and the cooldown only engages after a *successful* repartition. So a permanently-failing table re-attempts a full-table rewrite on every sync, indefinitely.
- **Unpartitioned first load.** The partition auto-detect only recognizes an incrementing integer key or a `created_at`/`inserted_at`/`createdAt` column. A table whose timestamp is named differently (e.g. `received_at`, a report `date`) or whose only key is a string/UUID falls through to no partitioning and loads the whole table into the delta-merge working set — OOMing on its first load, which repartition can't rescue (there's no delta table yet to rewrite).
- **Date-only tables stuck at the finest tier.** A datetime key with no time-of-day (e.g. an ad-platform report `date`) caps at the `day` tier. If a single day still exceeds the partition budget, the controller parks it at "finest tier" while it keeps OOMing.

## Changes

- **Repartition circuit breaker.** Adds a cumulative `repartition_failure_count` (survives the per-target re-flag) and, once it crosses a threshold, `reset_repartition_state()` clears the pending/swap/claim markers and parks the table with a timed backoff. Detection skips a parked table until the backoff elapses, then clears the marker and allows one fresh cycle — so a table fixed out-of-band (data shrank, or a finer scheme became available) recovers without operator action.
- **Generic initial-load partition fallback.** When the built-in auto-detect finds no mode, fall back to (a) the schema's incremental timestamp field as a datetime key when present, else (b) md5 over the primary keys, with a bucket count sized from the source's estimated row count and the first batch's per-row bytes. Any table with a primary key is now partitioned on its first load. `select_repartition_target` likewise returns a concrete md5/datetime scheme instead of relying on rewrite-time auto-detect (which couldn't size md5 and failed on a string-only key).
- **Composite `datetime_md5` partition mode.** A date-only table stuck at its finest datetime tier escalates to a composite scheme — a date bucket split into md5 sub-buckets over the primary keys (partition value `YYYY-MM-DD_<bucket>`). This keeps date locality (incremental merges still touch only recent dates, so the fix doesn't reintroduce the whole-table read) while bounding each partition's size. Used only when the primary key has within-date cardinality; otherwise the table stays parked for the circuit breaker to handle.

No schema migration — the new repartition state lives in the existing `sync_type_config` JSON.

Automated tests only (no manual or production run). Added and ran locally against Python 3.13 + Postgres:

- `test_repartition_controller.py` (full file green): circuit-breaker abandonment once the cumulative failure threshold is crossed, backoff gating plus auto-recovery of detection after the backoff elapses, the `is_repartition_backing_off` fail-safe branches (missing/unparseable `retry_after` treated as still-parked), and the failure counter resetting on a completed rewrite.
- `test_pipeline_utils.py`: the incremental-field-datetime and md5-over-PK fallbacks exercised through `setup_partitioning`, the source-size-based md5 bucket-count estimate, and the `datetime_md5` composite key builder (same date → same prefix, deterministic sub-bucket, sub-bucket in range).
- `test_repartition.py::TestSelectRepartitionTarget`: finest-tier → `datetime_md5` escalation, staying parked when the primary key can't split a date, sub-bucket growth for an over-budget composite table, and the concrete-md5 fallback for unpartitioned-with-keys.

Each group targets a regression the existing suite didn't cover: cumulative-vs-per-target failure counting; timestamp-named or string-PK tables going unpartitioned; and the finest-tier park. Two existing detection tests that asserted the old `mode=None` auto-detect target were updated to the new concrete-target behavior.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal pipeline behavior only — no user-facing docs affected.


danielcarletti (changes requested):
Blocking this for now, to push the repartition safeguards first

danielcarletti:
AI reply (Claude, driven by @danielcarletti): rebased this branch onto current master and added one rework commit (99f59cd) on top so the delta is reviewable in isolation. Your three commits are intact with authorship preserved (re-signed, since a rebase strips signatures and the repo rejects unsigned pushes).

**Rebase resolutions** (master gained the coarsening controller #76349, rewrite coalescing #76957, and the deadline/give-up changes #77769 since this branched):

- `OOM_WORKING_SET_AMPLIFICATION` is gone from master, replaced by `min_splittable_partition_bytes()` (the coarsen-aligned floor). The breaker constants now anchor after it.
- `_needs_pre_extraction_detection` merges both sides' guards, ordered CDC → coarsening nomination (measure) → circuit-breaker park (skip) → rollout flag. A nomination has to outrank the park, or a parked table silently eats nominations for a week.
- `_handle_failure` keeps your cumulative breaker as the first check, then the per-target give-up with the cooldown stamp from #77769. The park supersedes the cooldown.

**Rework commit, four changes on top:**

1. **`datetime_md5` escalation is gated on the true partition budget** (`true_budget_bytes` on `select_repartition_target`, new skip reason `composite_needs_over_budget`). The OOM trigger passes a halved split budget that any partition trivially beats, and the OOM signal is mostly infra noise today, so a ceiling-tier table under the real 500MB budget would mint sub-buckets and then double them on every false-OOM cycle. Escalation and sub-bucket growth now require a genuinely over-budget partition; the admin path is unguarded (operator's explicit call).
2. **`select_coarsen_target` can undo `datetime_md5`**: collapse to plain datetime at the coarsest tier that fits (simulated exactly from the composite keys), else shrink the sub-bucket count to a divisor. Without this the escalation was one-way.
3. **Nomination vs park wiring**: `stage_warehouse_coarsening` heuristics exclude parked tables, an explicit `--schema-id` nomination clears the park, and detection's park gate yields to a nomination.
4. **Neither the breaker's `reset_repartition_state` nor the give-up clears a `ready` swap** any more. A staged swap can mean live is already purged and temp is the only complete copy; clearing the marker strands temp for the next rebuild's sweep and the following sync bootstraps an empty table. The failure still counts; later syncs retry the swap from temp.

Verification: 175/175 repartition + staging-command tests green locally (13 new/extended cases across the four changes), repo-wide `mypy --cache-fine-grained .` clean, ruff clean.

Sequencing note: we'd like to hold landing until the OOM-classification PR (#76654) is in, so the `datetime_md5` escalation only ever rides a classified OOM signal - the true-budget guard bounds the damage either way.

danielcarletti:
AI reply (Claude, driven by @danielcarletti): resolved the merge conflicts. The repo's history rewrite invalidated the old base SHAs, so a plain rebase would have replayed thousands of master commits; instead the four branch commits were cherry-picked onto current master (authorship preserved, re-signed).

Two real conflicts, both in `select_repartition_target`:

- The unpartitioned-fallback tail: master's #80368 fixed the same UUID-only-key bug by keeping `partition_mode=None` and carrying a bucket count into auto-detection. This branch's explicit-scheme version supersedes it (prefer a datetime-typed incremental field, else explicit md5 - the rewrite never auto-detects), so the branch version won, including in the test expectations.
- The docstring over the same region, resolved to match.

Everything else merged cleanly on top of the resumable-rewrite changes (#80079): the give-up path now also drops the partial-rewrite checkpoint, and the staged-swap guard sits in front of both the circuit breaker and the give-up as before.

Verification: 331/331 repartition + partitioning + staging-command tests green locally, repo-wide `mypy --cache-fine-grained .` clean, ruff clean.

danielcarletti:
AI reply (Claude, driven by @danielcarletti): round 2. Rebased onto current master (only real conflict: master's `PartitionedTable` refactor landing where this branch adds the bucket helpers, plus the fallback-scheme overlap with #80368 resolved in this branch's favor again) and added one commit closing the gaps a full partition-mode sweep found for `datetime_md5`:

1. **Delta maintenance over-compacted every composite table.** `run_scheduled` passed the persisted count as the physical partition count, but for a composite it is sub-buckets per date - `files_per_partition` inflated by the number of dates, so `compact_if_fragmented` fired a full compact + vacuum on every pass for exactly the tables this mode rescues. Composite schemas now derive the count from the table's actual layout.
2. **The admin form could not stage the mode at all.** The change form's mode-to-rows map lacked `datetime_md5`, so selecting it disabled the format and count inputs; they were never submitted and the view always errored. Shared rows now resolve against the selected mode's id set (per-group toggling would let a later group hide a row an earlier one showed), the composite gets its own help text, and the admin tests cover staging plus the two new rejection cases.
3. **A coarsen collapse defeated the datetime tier ceiling.** Collapse targets kept the full composite key list; `_datetime_tier_ceiling` reads a multi-key list as hour-capable, so a date-keyed table burned a no-op day-to-hour rewrite per collapse-regrow cycle. Collapse targets now keep only the date key.
4. **Sub-bucket growth had no convergence refusal.** Degenerate md5 keys (the heaviest date already had several buckets available and still landed every byte in one) grew the count forever, one full-table rewrite per cooldown. That case now refuses with `composite_cannot_split`.

Smaller: rewrite logs print `datetime_md5/day/8` instead of dropping the count, the `stage_warehouse_coarsening` report shows mode and count so composite nominations are reviewable, and the admin README documents the mode.

Verification: 224 tests green across the repartition, controller, activity, maintenance, admin, and staging-command suites (run per-suite; the combined single-process run shows unrelated cross-suite DB interference), repo-wide `mypy --cache-fine-grained .` clean, ruff clean.

@danielcarletti the standing CHANGES_REQUESTED is your pre-safeguards block - safeguards are merged, so it's yours to dismiss when you're happy with this round.

PR #73090, opened 2026-07-23
fix(cdp): restrict hog function reruns to cyclotron-executable types

dmarchuk:
## Problem

Hog functions have multiple execution contexts: destinations and internal destinations run on the cyclotron worker, while source webhooks execute inline in the CDP API, transformations run during ingestion, and site_* functions transpile to client-side JavaScript. The rerun feature re-enqueues stored invocations onto the cyclotron hog queue, which only works for types that cyclotron can actually execute. Attempting to rerun a source webhook or other non-cyclotron type would enqueue an invocation nothing can drain, wedging the partition.

This fix prevents reruns of non-rerunnable types by rejecting them at the API level and disabling rerun controls in the UI.

## Changes

**Backend (Django)**
- Added `TYPES_THAT_CAN_RERUN` allowlist to `HogFunction` model (destinations and internal destinations only)
- Modified `HogFunctionViewSet.rerun()` to reject reruns of non-rerunnable types with a 400 response explaining why
- Removed the previous header-stripping logic for source webhooks (no longer needed since they can't be rerun)

**Node.js (Rerun Paginator)**
- Added `RERUNNABLE_HOG_FUNCTION_TYPES` constant mirroring the Django allowlist
- Modified `rehydrateInvocation()` to return `null` for non-rerunnable types, skipping them rather than enqueuing
- Removed the webhook-specific header/query stripping logic (replaced by the type check)
- Updated tests to verify non-rerunnable types are skipped and rerunnable types proceed

**Frontend (React)**
- Exported `isRerunnableHogFunctionType()` helper from `hogInvocationsLogic`
- Added `hogFunctionType` prop to `HogInvocations` component
- Disabled rerun controls (buttons, checkboxes) when the function type is non-rerunnable, showing a tooltip explaining why
- Updated `HogFunctionScene` to pass the function type to the invocations component
- Added unit tests for the rerunnable type check

73090-rerun-disabled-source_webhook

- Added backend tests verifying 400 rejection for non-rerunnable types (source_webhook, warehouse_source_webhook, transformation, site_app, site_destination) and 200 success for rerunnable types (destination)
- Added Node.js tests verifying non-rerunnable types return null (skipped) and rerunnable types proceed
- Added frontend unit tests for `isRerunnableHogFunctionType()` covering both rerunnable and non-rerunnable cases, plus undefined/null inputs
- Existing invocations UI tests continue to pass with the new prop

The changes are guarded by type checks and allowlists at three layers (API, worker, UI), so a non-rerunnable type cannot reach the enqueue path even if one layer is bypassed.

https://claude.ai/code/session_013nU5f4EBQcgsHcgAEGxkH8

### DIFF
--- a/products/cdp/backend/api/hog_function.py
+++ b/products/cdp/backend/api/hog_function.py
@@ -49,6 +49,7 @@ from posthog.plugins.plugin_server_api import create_hog_invocation_test, rerun_
 from products.cdp.backend.api.hog_function_template import HogFunctionTemplateSerializer
 from products.cdp.backend.models.hog_function_template import HogFunctionTemplate
 from products.cdp.backend.models.hog_functions.hog_function import (
+    TYPES_THAT_CAN_RERUN,
     TYPES_WITH_JAVASCRIPT_SOURCE,
     HogFunction,
     HogFunctionState,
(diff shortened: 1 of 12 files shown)


products/cdp/backend/api/hog_function.py
```diff
@@ -692,17 +693,32 @@ def rerun(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
         hog_function = self.get_object()
 
+        if hog_function.type not in TYPES_THAT_CAN_RERUN:
+            return Response(
+                {
+                    "queued_count": 0,
+                    "skipped_count": 0,
+                    "detail": (
+                        f"Re-runs aren't supported for '{hog_function.type}' functions. They run outside "
+                        "the cyclotron queue that a re-run replays onto, so a re-run would never execute."
```
meikelmosby: folks don't know what the cyclotron queue is .. 😅 just drop the whole explanation.

meikelmosby (approved):

PR #73109, opened 2026-07-23
fix(conversations): truncate oversized widget session context

christiaan-ph:
## Problem

Submitting a support ticket through the conversations widget fails silently when the page URL is long.
The widget auto-attaches `session_context` (current URL, replay URL) and `traits` (name, email) to every message, and `WidgetMessageSerializer` rejected the whole request when any value exceeded a length cap.
The person submitting can't do anything about it: they didn't type the URL, and the widget doesn't surface the 400, so the ticket just never sends.

Pages that encode UI state in the URL fragment (the SQL editor, person pages with a recording open) routinely exceed the 2,000 char cap on `session_context` values, and a user hit exactly this while trying to file a ticket from a recording page.

Follow-up to #72885, which fixed the same failure shape for the field the user does control (the message body).

## Changes

Applies the approach posthog-js uses for over-long strings: truncate, don't reject.
posthog-js slices every string property to `properties_string_max_length` (`_copyAndTruncateStrings` in `browser-common/src/utils/general-utils.ts`), so a long URL never costs an event.
That truncation runs on the capture path only, and the widget posts this payload straight through `_send_request`, so the untruncated URL does reach this endpoint and the server has to be the one to bound it.

- Oversized `session_context` and `traits` string values are truncated to the existing caps (2,000 and 500) instead of raising.
- Over-long keys and excess entries are skipped instead of raising, so auto-attached context is sanitized uniformly. These two caps are ours rather than a posthog-js behavior: posthog-js bounds string length but never property count, so the precedent covers values only.
- Both fields share one `_sanitize_context` helper with the caps named at each call site. The only difference between them is that `traits` coerces non-null values to strings while `session_context` keeps native types.
- `message` is untouched: it's user-typed content, so rejecting it (with the client-side guard from #72885) stays correct.
- A structurally malformed payload (non-dict or null `traits`) still returns 400. `DictField` rejects those before the field validators run, and that's a client bug rather than data the person filing the ticket can fix.

Caps are unchanged, so the stored footprint per ticket is the same as before.

> [!NOTE]
> Behavior change on a public endpoint: `POST /api/conversations/v1/widget/message` no longer returns 400 for oversized `session_context` or `traits`; it accepts and truncates them.

- `TestWidgetContextSanitization` (`SimpleTestCase`, no DB): parameterized across both fields, asserting an oversized value is truncated to the cap, that over-long keys and excess entries are dropped rather than rejected, that `traits` coerces non-null values to strings while `session_context` keeps native types, and that a non-dict or null value is still rejected. Catches a revert to raising, which is the bug being fixed, and the coercion asymmetry the shared helper introduces.
- `test_long_current_url_is_truncated_not_rejected` (`TestWidgetAPI`, DB-backed wiring guard): a 3,000 char `current_url` returns 200 and the truncated value is persisted on the ticket, proving the viewset stores the sanitized value rather than rejecting the request.
- Ran the full `products/conversations/backend/api/tests/test_widget.py` locally against the dev stack: 58 passed, including the DB-backed case. Repo-wide `uv run mypy --cache-fine-grained .` and `ruff` are clean.

Manual check, if you want to reproduce the original failure: open the widget on a page with a long URL fragment (the SQL editor with a query in the hash works) and send a ticket. On master the request 400s and the ticket never appears; here it sends and the ticket shows a truncated `current_url`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No doc changes: these caps are internal validation details.

## 🤖 Agent context

Originally authored by Claude (Fable 5) in a PostHog Code session, then reworked by Claude (Opus 5) across two rounds of review feedback: first to match how posthog-js handles over-long strings, then to collapse the two duplicated validator loops into the shared `_sanitize_context` helper.
Directed by @christiaan-ph off a support ticket, as a follow-up to #72885.

The posthog-js precedent cited above was verified against the SDK source rather than assumed, which is what turned up the two limits above that posthog-js has no counterpart for.

Skills invoked: `/writing-tests`, `/writing-code-comments`.

---


christiaan-ph:
leaving this as a draft for now, but asking for a review on the general approach here

christiaan-ph:
Feedback from Alex: look at how posthog-js handles really long URLs and apply this method instead

veryayskiy:
products/conversations/backend/api/widget.py:113-138: This PR makes session_context unable to cause a validation error, so the diagnostic block it was added for is now stale — the comment "An over-long auto-captured session_context value (e.g. current_url) is a known rejection cause" is false, and session_context_field_count / session_context_field_lengths will only ever be reported for unrelated failures (empty message, bad auth). Drop them or rewrite the comment in the same PR.

products/conversations/backend/api/serializers.py
```diff
@@ -101,52 +102,43 @@ def validate_traits(self, value):
         if not isinstance(value, dict):
```
veryayskiy: validate_traits and validate_session_context are now the same loop with different caps (entries / key length / value length) and a str() coercion flag. The description says context is "sanitized uniformly" but the code duplicates it twice — one _sanitize_context(value, max_entries, max_key_length, max_value_length, coerce=...) helper would make that literally true.

products/conversations/backend/api/serializers.py
```diff
@@ -101,52 +102,43 @@ def validate_traits(self, value):
...
             raise serializers.ValidationError("traits must be a dictionary")
 
-        if len(value) > 50:
-            raise serializers.ValidationError(f"Too many traits: {len(value)} (max 50)")
-
-        validated = {}
+        validated: dict[str, str | None] = {}
         for key, val in value.items():
+            if len(validated) >= 50:
```
veryayskiy: the entry cap now silently keeps the first N by dict order, so a payload with >20 context keys can drop current_url depending on client ordering. Fine if intentional, but worth a one-line comment saying insertion order decides what survives.

products/conversations/backend/api/serializers.py
christiaan-ph: Done. Both now delegate to `_sanitize_context(value, *, max_entries, max_key_length, max_value_length, coerce_to_string=False)` with the caps named at each call site, so the only real difference between the two fields is the coercion flag.

Also dropped the `isinstance(value, dict)` guards while in there: `DictField` rejects non-dicts during `run_validation`, before `validate_<field>` runs, so they were unreachable. Pinned that boundary with a test instead.

products/conversations/backend/api/serializers.py
christiaan-ph: Intentional, now noted in the helper docstring.

One caveat worth being straight about: the entry cap is ours, not something posthog-js does. It slices string length but never caps property count, so the truncate-instead-of-reject precedent only covers values. The widget itself sends two keys per field (`session_replay_url`/`current_url`, and `name`/`email`), so the cap is unreachable from our own client and only engages for a hand-rolled one, where the client ordering is the only priority signal we have.

Happy to pin `current_url` ahead of the cap if you would rather it be guaranteed.

veryayskiy (approved):

PR #73110, opened 2026-07-23
fix(aio): guard llm taggers logic against unmounted store access

carlos-marchal-ph:
## Problem

The AI observability tags feature threw an unhandled frontend error, surfaced in error tracking:

```
[KEA] Can not find path "products.ai_observability.taggers.llmTaggersLogic" in the store.
```

`TaggerMetrics` (in `AIObservabilityTagsScene.tsx`) and the currently-unrendered `TagMetrics` component both read the bare `llmTaggersLogic` reference via `useValues(llmTaggersLogic)`, assuming an ancestor had mounted it into the Kea store. When that assumption breaks — e.g. during scene teardown on navigation away — the store lookup fails and the component crashes.

## Changes

- `TaggerMetrics`: build the logic instance (`useValues(llmTaggersLogic())`) so the component holds its own mount, matching the pattern already used in the sibling `AIObservabilityTagsContent`.
- `TagMetrics`: read `taggers` through `tagMetricsLogic` (which already exposes it via its `connect` to `llmTaggersLogic`) and drop the redundant bare `llmTaggersLogic` read entirely.

No behavior change — both components render the same values; this only hardens how the logic is accessed.

Ran oxlint on both changed files (passes). The `taggers` value is already part of the generated `tagMetricsLogicValues` type, so the `TagMetrics` change type-checks. No new automated tests added — the fix is a mount-safety change to existing components; I (the agent) did not run the app manually.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated from an error-tracking alert via the PostHog Slack app. Traced the KEA "path not in store" error to two components reading the bare `llmTaggersLogic` reference instead of a built/connected instance. Considered deleting the unused `TagMetrics` component but kept the change scoped to fixing the anti-pattern. No skills were required for this frontend-only mount-safety fix.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0A006STKHR/p1784796195674659?thread_ts=1784796195.674659&cid=C0A006STKHR)*

### DIFF
--- a/products/ai_observability/frontend/tags/AIObservabilityTagsScene.tsx
+++ b/products/ai_observability/frontend/tags/AIObservabilityTagsScene.tsx
@@ -39,7 +39,9 @@ export const scene: SceneExport = {
 }
 
 function TaggerMetrics(): JSX.Element {
-    const { chartQuery, totalRuns, taggers, runStatsLoading } = useValues(llmTaggersLogic)
+    // Build the logic instance so this component holds its own mount, rather than
+    // reading the bare reference and assuming an ancestor kept it in the store.
+    const { chartQuery, totalRuns, taggers, runStatsLoading } = useValues(llmTaggersLogic())
 
     const enabledCount = taggers.filter((t) => t.enabled && !t.deleted).length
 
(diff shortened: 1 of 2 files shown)


carlos-marchal-ph:
Closing — on review, this change is cosmetic and doesn't address the actual bug. The error is a Kea teardown race (a selector subscription firing after the logic's reducers are removed from the store during scene navigation), not a missing mount. Switching `useValues(llmTaggersLogic)` to `useValues(llmTaggersLogic())` is functionally identical for a propless logic and doesn't change unmount ordering. Given it's a single occurrence ever, not worth shipping a no-op fix.

PR #73112, opened 2026-07-23
fix(replay-vision): make in-flight cap admission race-safe with atomic enqueue claims

fasyy612:
## Problem

The on-demand scan endpoints (`observe`, `bulk_observe`, `retry`) enforce the per-scanner (150) and per-team (300) in-flight caps by counting `ReplayObservation` rows. A row only appears once the apply workflow's first activity runs, so between `start_workflow` and that insert an enqueued scan is invisible to every counter. Concurrent requests read the same free-looking snapshot and can collectively start more scans than the caps allow a race. The monthly credit quota has a second gate inside the create activity so spend overshoot stays small, but the in-flight caps are effectively soft under concurrency.

## Changes

- New `enqueue_claims.py`: an atomic enqueue-slot claim, one Lua eval over per-team and per-scanner Redis sorted sets (evict expired, check both pools, add). Same shape as posthog/clickhouse/client/limit.py and the session-summary cap's reserve-then-refund.
- `start_apply_scanner_workflow` claims a slot before starting the workflow and returns a new CAPPED outcome when refused. Failed starts release the claim.
- `create_observation_activity` releases the claim once the row exists, on every exit path. A 15-minute TTL score reclaims claims from crashed workflows.
- Re-claiming the same deterministic workflow id never consumes a second slot, so duplicate requests and retries can't burn headroom.
- observe/retry translate CAPPED to a 429. `bulk_observe` reports it as the existing `skipped_limit` outcome and stops the batch. The headroom pre-slice and team-capacity pre-check now count live claims too.
- Fail-open on Redis errors: the caps are backpressure guardrails, not billing, so an outage degrades to the previous snapshot behavior instead of blocking scans.

Added 12 tests

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context


arnohillen (approved):

arnohillen:
I added two small test cases to catch a future regression.

PR #73114, opened 2026-07-23
fix(insights): show tooltip with full header name when truncated

xljones:
## Problem

SQL and HogQL table insights clamp long column headers to a few lines with a CSS ellipsis. Once a header is cut off there's no way to read its full name — you just get the truncated text with no tooltip.

## Changes

Column headers in the data visualization table now reveal the full header name in a tooltip on hover, but only when the label is actually truncated. The check compares the rendered box against the content (the full text always stays in the DOM), so short headers that fit render exactly as before.

I (Claude) couldn't run the app or the frontend toolchain in this environment (no `node_modules`), so I performed no manual or automated runs. I reasoned through the behavior instead:

- Short headers stay non-truncated, so `Tooltip` is a passthrough and the rendered DOM is unchanged from before — the existing `SQLTableConditionalFormatting` story only has short labels, so its snapshot is unaffected.
- Truncation detection mirrors the existing pattern in `EditableField` (`scroll*` vs `client*` size), extended to also catch the multi-line clamp these headers use.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change — internal UI affordance only.

## 🤖 Agent context

Xander asked for a tooltip that reveals the full header name when a table insight's header is truncated with an ellipsis. I (Claude) traced the three table renderers: the trends `InsightsTable` already reveals full breakdown names through `PropertyKeyInfo`'s native `title`, and the events `DataTable` doesn't truncate headers at all. The real gap was the SQL/HogQL `DataVisualizationTable`, whose headers are line-clamped with `text-overflow: ellipsis` and had no tooltip.

I gated the tooltip on real truncation (measured once after layout, plus lazily on hover) rather than always showing it, so non-truncated headers behave exactly as before. I kept the change local to the data-viz table rather than touching the shared `LemonTable`, which would have risked double tooltips on breakdown headers that already have one.

No repo skills were needed for this change.

---

### DIFF
--- a/frontend/src/queries/nodes/DataVisualization/Components/Table.tsx
+++ b/frontend/src/queries/nodes/DataVisualization/Components/Table.tsx
@@ -2,7 +2,7 @@ import '../../DataTable/DataTable.scss'
 
 import { useActions, useValues } from 'kea'
 import posthog from 'posthog-js'
-import React from 'react'
+import React, { useCallback, useLayoutEffect, useRef, useState } from 'react'
 
 import { IconPin, IconPinFilled } from '@posthog/icons'
 import { LemonBanner, LemonTable, LemonTableColumn, Tooltip } from '@posthog/lemon-ui'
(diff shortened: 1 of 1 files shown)


xljones:
Superseded by #73121, which moves the same change to an `xljones/`-prefixed branch and adds a screenshot of the tooltip to the description.

PR #73116, opened 2026-07-23
fix(llm-gateway): normalize GLM thinking requests

tatoalo:
## Problem

**Why:** GLM 5.2 cloud runs can send the `clear_thinking_20251015` context strategy without an enabled thinking configuration, which the Anthropic-compatible API rejects. GLM supports `high` and `max` reasoning effort, so the gateway should preserve those settings in a valid request shape.

## Changes

- Normalize GLM Anthropic requests with high or max effort to adaptive thinking.
- Remove orphaned clear-thinking edits while preserving other context-management strategies.

### DIFF
--- a/services/llm-gateway/src/llm_gateway/glm_routing.py
+++ b/services/llm-gateway/src/llm_gateway/glm_routing.py
@@ -47,6 +47,35 @@ from llm_gateway.modal import (
 
 LlmCall = Callable[..., Awaitable[Any]]
 
+GLM_REASONING_EFFORTS: frozenset[str] = frozenset({"high", "max"})
+
+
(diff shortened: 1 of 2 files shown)


jonathanlab (approved):
is this runtime agnostic, e.g. let's say when we add the pi adapter, will this assume a claude request shape?

tatoalo:
🧵 https://posthog.slack.com/archives/C09G8Q32R6F/p1784801725383599

richardsolomou (approved):
could we cover the sole `clear_thinking_20251015` edit case too? that branch removes `context_management` entirely, and its cleanup isnt tested rn.

PR #73118, opened 2026-07-23
chore(data-warehouse): document OAuth app credential rollout in warehouse-source skill

Gilbert09:
## Problem

Wiring a PostHog-registered OAuth app for a new data-warehouse source spans four places — registering the OAuth client with the provider, the `settings/integrations.py` env vars, the charts deploy env, and the actual values in AWS Secrets Manager. That process wasn't written down, so it's easy to miss a step (e.g. wiring only one of the two secret stores) or to have an agent try to apply live secrets it shouldn't handle.

## Changes

Adds a reference doc to the `implementing-warehouse-sources` skill covering the full **register → code → charts → secrets** rollout:

- The provider registration + the US/EU/dev/localhost redirect URIs.
- The charts wiring into **both** `posthog-django-shared-secrets` (web app, runs the OAuth flow) and the worker's `{app-name}-secrets` `secret_env_app_specific` store.
- Writing values via the `PostHog/secrets` CLI `template` command — noting that repo is a CLI over AWS Secrets Manager (values are never committed).
- An explicit "what an agent can vs. must not automate" boundary — the agent scaffolds code/charts/template, but the live-secret apply (interactive AWS SSO, `secrets-editor` role, tailnet force-sync) stays with a human.

Also notes the JWT-`sub` account-id pattern in `SKILL.md` for providers whose token response carries no account identifier. Keeps the `SKILL.md` entry point lean (detail lives in `references/`, per the skill conventions). Uses the recent Resend OAuth work as the worked example.

Docs-only. `build_skills.py --lint` passes — 121 skills OK, with only pre-existing warnings in an unrelated skill.

### DIFF
--- a/.agents/skills/implementing-warehouse-sources/SKILL.md
+++ b/.agents/skills/implementing-warehouse-sources/SKILL.md
@@ -596,8 +596,17 @@ If new:
    - Add to `IntegrationKind` enum.
    - Add to `OauthIntegration.supported_kinds`.
    - Add an `elif kind == "your-source": return OauthConfig(...)` branch in `oauth_config_for_kind()`.
-3. **Redirect URI**: `https://localhost:8010/integrations/your-kind/callback` in the external service.
-4. List any new env vars in the final handoff so they can be set in all environments.
+     Raise `NotImplementedError("<Source> app not configured")` when the env vars are empty — that's the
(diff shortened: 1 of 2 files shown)


MarconLP (approved):

danielcarletti (approved):
LGTM

PR #73119, opened 2026-07-23
feat(conversations): match previous tickets by email too

eleftheriatrivyzaki:
## Problem

When viewing a support ticket, the previous tickets panel only matched other tickets by the person's merged `distinct_id`s. Tickets from the same customer submitted under an unmerged distinct_id (a separate anonymous session, or an email-only ticket) never appeared, even when they came from the same email address.

## Changes

The previous-tickets query now also matches on `email_from` (case-insensitive), OR'd with the existing `distinct_id` match.

- `TicketViewSet` gains an `emails` query param.
- The ticket detail page (`loadPreviousTickets`) and `api.conversationsTickets.list` pass the person's email plus the current ticket's `email_from`.
- Generated OpenAPI and MCP types updated to match.

I tested this locally end to end: seeded tickets that share an email but have an unmerged distinct_id, and confirmed the panel now surfaces them. I also checked that matching is case-insensitive and that the existing distinct_id-only behavior is unchanged. No automated tests added yet.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs changes needed.

## 🤖 Agent context

---

### DIFF
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -7057,6 +7057,7 @@ const api = {
                 assignee?: string
                 tags?: string
                 distinct_ids?: string
+                emails?: string
                 search?: string
                 date_from?: string
                 date_to?: string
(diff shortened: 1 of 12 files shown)


abigailbramble:
If this doesn't - I'd flag in the previous tickets panel that this has been matched on email and isn't the same person profile (a warning pill thing or similar)

eleftheriatrivyzaki:
> If this doesn't - I'd flag in the previous tickets panel that this has been matched on email and isn't the same person profile (a warning pill thing or similar)

Good point, added this 
Screenshot 2026-07-23 at 2 10 16 PM

veryayskiy (approved):

PR #73120, opened 2026-07-23
fix(activity-log): attribute bearer-token API writes to the key's user

jurajmajerik:
## Problem

API writes authenticated with a personal API key or OAuth token (e.g. via MCP) show up in activity logs as "PostHog" system entries. `ActivityLoggingMiddleware` captures the acting user before DRF auth runs, so bearer-token requests never get one. Affects models with signal-driven activity logging (experiments, holdouts, shared metrics, ...) — viewsets that log explicitly with `request.user` were fine.

## Changes

- Set the activity-log user after successful personal API key / OAuth authentication.
- Show "via MCP" next to the timestamp on activity rows when the request came from an API client (`x-posthog-client` header, already captured and serialized, just never rendered).

Two regression tests: an experiment created via personal API key and via OAuth token gets an activity entry attributed to the key's user instead of system. Both fail without the auth change. No existing test covered bearer-auth attribution. Also ran the activityLogLogic jest tests, frontend typescript check, and mypy on `posthog/auth.py`.


products/experiments/backend/hogql_queries/experiment_query_runner.py
```diff
@@ -146,6 +151,53 @@ def experiment_has_min_runtime_for_precomputation(
...
+                    pass
+        for nested in obj.values():
+            ids |= _collect_cohort_ids(nested)
+    elif isinstance(obj, list):
+        for item in obj:
+            ids |= _collect_cohort_ids(item)
+    return ids
+
+
+def has_uncalculated_cohorts(team: Team, *filter_sources: Any) -> bool:
```
a-lider, line 173: This cohort-gating change (and the exposures-runner + test changes that go with it) is unrelated to bearer-token activity-log attribution.

frontend/src/lib/components/ActivityLog/ActivityLog.tsx
```diff
@@ -191,8 +191,13 @@ export const ActivityLogRow = ({
                     {logItem.extendedDescription && (
                         <div className="ActivityLogRow__description__extended">{logItem.extendedDescription}</div>
                     )}
-                    <div className="text-secondary">
+                    <div className="text-secondary flex items-center gap-1.5">
                         <TZLabel time={logItem.created_at} />
+                        {logItem.client && (
```
a-lider, line 196: The `via MCP` tag only renders on this Advanced Activity Logs page, not on the entity History tabs because the per-entity ActivityLogSerializer doesn't emit client. 

image

a-lider (changes requested):

frontend/src/lib/components/ActivityLog/ActivityLog.tsx
a-lider: image

products/experiments/backend/hogql_queries/experiment_query_runner.py
jurajmajerik: Yes, rebased

frontend/src/lib/components/ActivityLog/ActivityLog.tsx
jurajmajerik: Added `client` to the serializer. We have two serializers for the same rows, hence the inconsistency, but I'd rather fix that in a follow-up.

products/experiments/backend/hogql_queries/experiment_query_runner.py
```diff
@@ -341,10 +393,13 @@ def _should_precompute(self) -> bool:
...
 
-        return experiment_has_min_runtime_for_precomputation(
+        if not experiment_has_min_runtime_for_precomputation(
             self.experiment.start_date,
             self.experiment.end_date,
-        )
+        ):
+            return False
+
+        return not has_uncalculated_cohorts(self.team, self.experiment.exposure_criteria, self.metric)
```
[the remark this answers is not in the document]
jurajmajerik: This is no longer part of the PR after rebasing

feliperalmeida (approved):

a-lider (approved):
Looks good, thanks for fixing impersonation!

PR #73128, opened 2026-07-23
fix(data-imports): evict stale DB connections before ducklake ORM lookups

Gilbert09:
## Problem

Error tracking surfaced a recurring `OperationalError: the connection is closed` in the data warehouse import pipeline, originating in `posthog/ducklake/common.py`. It's actually two related sites, both hitting the same root cause:

1. `_get_org_id_for_team` (called from the DuckLake copy-to-warehouse activity's duckgres path) — [issue](https://us.posthog.com/project/2/error_tracking/019ee054-3226-7130-96e7-a43ff1c9031f), plus a second grouping of the identical call path ([issue](https://us.posthog.com/project/2/error_tracking/019efd29-265c-78c2-9f11-1421a69e812c)).
2. `duckgres_sink_enablement` (called from the Duckgres batch-sink poll loop's team-enablement refresh) — [issue](https://us.posthog.com/project/2/error_tracking/019f4a8e-a31c-7970-8dae-1ed15ee29378).

Both functions run a plain Django ORM query from a long-lived worker context: a sync Temporal activity thread in the first case, a persistent asyncio poll loop (via a bare `sync_to_async`, not the connection-safe `database_sync_to_async` wrapper) in the second. Neither ever goes through Django's request/response cycle, so the signal that normally closes a stale connection (killed by the database or a proxy after `CONN_MAX_AGE`, or a network blip) never fires. The next query on that thread reuses the dead connection and fails.

This is a real bug in our code, not a transient error worth silencing.

## Changes

My first attempt called `close_old_connections()` directly inside `_get_org_id_for_team` and `duckgres_sink_enablement`. CI caught the problem with that: both are shared helpers called from arbitrary contexts, and `close_old_connections()` force-closes the connection whenever autocommit doesn't match the configured default — which is exactly the case inside an open transaction (including a test's atomic wrapper). That broke unrelated `@pytest.mark.django_db` tests that create rows and call these helpers in the same transaction.

The fix now lands at the actual thread-crossing boundaries instead, matching this codebase's existing safe pattern for the same failure mode:

- `copy_data_imports_to_ducklake_activity`, `verify_data_imports_ducklake_copy_activity`, and `cleanup_data_imports_staging_activity` (`posthog/temporal/ducklake/ducklake_copy_data_imports_workflow.py`) now call `close_old_connections()` at the top of each sync Temporal activity, before any Django ORM access happens on that worker thread — the same thing `check_billing_limits_activity` already does elsewhere in this pipeline.
- The Duckgres sink's team-enablement refresh (`consumer.py`'s `_enabled_team_ids`) now crosses into its worker thread via `database_sync_to_async_pool` (from `posthog/sync.py`) instead of a bare `sync_to_async`, so the connection is cleaned up at the async→sync boundary rather than inside the business-logic function itself — the same wrapper the sibling live-apply-stamp hook in the same file already relies on.

No retry policy or error-classification changes.

- Added regression tests asserting `close_old_connections` is called on the activities' early-return paths too (`test_ducklake_copy_data_imports_workflow.py`), and that the team-enablement refresh routes through `database_sync_to_async_pool` rather than a bare `sync_to_async` (`test_consumer.py`) — this is the only way to lock in the fix at unit-test speed, since a real stale connection isn't reproducible in a fast test.
- Ran the non-DB subset of the touched test files locally (`uv run pytest ... -m "not django_db"` / `-k` filters covering the new assertions): all pass.
- Ran `ruff check` and `ruff format --check` on all changed files (clean).
- Could not run the full Django `@pytest.mark.django_db` suite in this sandbox (no Postgres server/docker available here — only the `psql` client). My first, incorrect version of this fix broke `test_team_state.py` and `test_enablement.py` in CI; I traced that failure, fixed the root cause described above, and verified the affected test modules still collect and their non-DB tests pass. Relying on CI to confirm the full DB-backed suite.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Triaged three PostHog error-tracking issues (linked above) for the data-imports pipeline, confirmed they share one root cause, and checked for an existing duplicate fix among open PRs (none matched). My first fix attempt put `close_old_connections()` inside the shared low-level helpers themselves; CI's `test_team_state.py`/`test_enablement.py` failures revealed that was unsafe (it can sever an in-progress transaction), so I moved the fix to the actual thread-crossing boundaries instead, following the codebase's existing convention for this exact failure mode.

---


Gilbert09:
Heads up: after rebasing onto `master` to clear the merge conflict, the repo-wide mypy check started failing on `products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py` — a call to `handle_non_retryable_error` in the schema-column-type-changed branch (added in #73135) was passing the `PipelineInputs` object as the first positional arg instead of `team_id`/`source_id`/`run_id`. That's both a type error and a runtime `TypeError` waiting to fire, so I fixed the call to match the other two call sites in the same function. It's outside this PR's original scope but was needed to get CI green post-rebase.

PR #73135, opened 2026-07-23
fix(data-imports): treat all incompatible Delta casts as schema changes

Gilbert09:
## Problem

The data-warehouse import pipeline is throwing a large, growing volume of retryable `ArrowInvalid` exceptions from shared code in `evolve_pyarrow_schema` ([error tracking issue](https://us.posthog.com/project/2/error_tracking/019efe10-1e4c-7fa1-ae5e-d46af4b51325), and its older sibling [019ed03f](https://us.posthog.com/project/2/error_tracking/019ed03f-6b97-7c81-b88a-4ae3d9a067b7)). They all originate at the same line - casting an incoming column into the stored Delta column type - and share a root cause: the incoming data no longer fits the type the table was created with.

We already handle one shape of this: a numeric value that doesn't fit an integer column raises a `SchemaColumnTypeChangedException` telling the user to reset and re-sync, which is classified non-retryable. But the handler only fired when the target was an integer column and the source was numeric. The other shapes fell through to a bare re-raise:

- non-numeric text into a numeric column (`Failed to parse string: '80-150' as a scalar of type int32`)
- a value that overflows the stored decimal precision (`Cannot convert 999999999.0 to Decimal128(...): overflow`)
- non-boolean text into a boolean column (`Failed to parse value: processed`)

Because those surfaced as a plain `ArrowInvalid`, Temporal retried them to the activity max on every scheduled sync, generating the noise. None of them are retryable - delta-rs can't change a column's type in place, so only a reset and full re-sync fixes them.

## Changes

- **`evolve_pyarrow_schema`**: any `ArrowInvalid` from the cast now raises the actionable `SchemaColumnTypeChangedException`. Reaching that cast already means the incoming type differs from the stored type and the timestamp path didn't apply, so the failure is always a deterministic incompatibility. Lossless widening (for example a whole-valued float into an integer column) still casts fine and never reaches the handler, so this doesn't swallow benign data.
- **Shared V2 error handling**: `_handle_import_error` now classifies `SchemaColumnTypeChangedException` as non-retryable by type, the same way it already does for `RESTClientNonRetryableError`. Previously only SQL sources caught it, via a message match in `get_non_retryable_errors`; a non-SQL source hit the generic re-raise and kept retrying. V3 already matches "Source column type changed" in its non-retryable patterns, and V3 runs the same `evolve_pyarrow_schema`, so it's covered by the first change.

> [!NOTE]
> This keeps the error visible but bounded (a few attempts per run) rather than suppressing it. It's a real, user-actionable condition - the user must reset and re-sync the affected table - not benign control-flow, so it stays in error tracking, just no longer as an unbounded retry flood.

Automated tests only (I did not run the pipeline end to end):

- Added a parametrized `test_evolve_pyarrow_schema_incompatible_cast_raises_actionable_error` covering the three previously-unhandled shapes (text into int, text into bool, float overflowing a decimal). These regress to a raw `ArrowInvalid` if the broadened handler is reverted - nothing existing caught them, since the existing `..._integer_overflow_...` test only exercises integer targets with numeric sources.
- Added `test_schema_column_type_changed_routes_through_handler_without_source_opt_in`, mirroring the existing `RESTClientNonRetryableError` test, to lock in the shared-layer non-retryable classification for sources that don't list the message.
- Ran both files: `test_pipeline_utils.py` and `test_import_data_sync.py`, 127 passed.

## 🤖 Agent context

Triaged and authored by Claude (Claude Code) from the error-tracking issue: pulled the stack trace and representative events via the PostHog error-tracking tools, traced the failure to `evolve_pyarrow_schema`, confirmed the two issues share a root cause, and checked the V2/V3 non-retryable wiring before choosing a general fix over a per-source patch. Invoked the `/writing-tests` skill before adding tests. Decision worth flagging for review: I removed the `is_integer` target guard entirely rather than adding string as another special case, on the reasoning that every path reaching that cast is an unreconcilable type mismatch - please sanity-check that there's no cast in that branch we'd rather keep retryable.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline/test/test_pipeline_utils.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline/test/test_pipeline_utils.py
@@ -745,6 +745,37 @@ def test_evolve_pyarrow_schema_whole_valued_floats_cast_into_stored_integer_colu
     assert evolved_table.column("val").to_pylist() == [10, 20]
 
 
+@pytest.mark.parametrize(
+    "delta_type, incoming_column",
+    [
+        # Non-numeric text arriving for a column stored as int (Failed to parse string).
+        (pa.int32(), pa.array(["80", "80-150"], type=pa.string())),
+        # Non-boolean text arriving for a column stored as bool (Failed to parse value).
+        (pa.bool_(), pa.array(["true", "processed"], type=pa.string())),
+        # A value that overflows the stored decimal precision (Cannot convert ... overflow).
+        (pa.decimal128(3, 1), pa.array([1.0, 999999999.0], type=pa.float64())),
+    ],
+)
+def test_evolve_pyarrow_schema_incompatible_cast_raises_actionable_error(
+    delta_type: pa.DataType, incoming_column: pa.Array
+):
+    """Incoming data that can't be cast into the stored Delta type — non-numeric text into a
+    numeric/boolean column, or a value that overflows the stored decimal precision — raises the
+    actionable reset-and-re-sync error rather than a raw pyarrow ArrowInvalid that gets retried."""
+    arrow_table = pa.table(
+        {
+            "id": pa.array([1, 2], type=pa.int64()),
+            "val": incoming_column,
+        }
+    )
+    delta_schema = deltalake.Schema.from_arrow(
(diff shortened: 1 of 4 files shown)


fercgomes:
Heads up: the SchemaColumnTypeChangedException branch here passes job_inputs where handle_non_retryable_error expects team_id, source_id, run_id — mypy is failing on master and blocking PR CI. One-line fix up at the PR linked above (matches the sibling call sites); feel free to take it over or fix differently.

PR #73137, opened 2026-07-23
fix(data-imports): stop tracking benign billing-limit halts

Gilbert09:
## Problem

Error tracking is being flooded by `BillingLimitsWillBeReachedException` from the data warehouse import pipeline. It is raised in `setup_row_tracking_with_billing_check` (shared pipeline code in `products/warehouse_sources/backend/temporal/data_imports/pipelines/common/extract.py`) when a sync would push an account over its Data Warehouse billing limit. Two error-tracking issues share this exact root cause across different sources and tables ([issue A](https://us.posthog.com/project/2/error_tracking/019efe0e-b4f7-7420-9391-a30b2cd733d0), [issue B](https://us.posthog.com/project/2/error_tracking/019ed03f-0669-78a2-8e7e-b720b704d74e)).

This is not a defect and not a user error we can act on by retrying. It is expected control flow: the workflow already catches this exception and marks the job `BILLING_LIMIT_TOO_LOW`. Two things made it noisy anyway:

1. The activity interceptor (`posthog/temporal/common/posthog_client.py`) captures every non-cancellation exception raised out of an activity to error tracking. Billing halts were not excluded.
2. The import activity did not list this exception in `non_retryable_error_types`, so Temporal retried it up to the per-run maximum, and every retry re-captured.

The net effect was a very high, persistent occurrence count with nothing for us to fix.

## Changes

I followed the pattern this codebase already uses to keep expected conditions out of error tracking. The interceptor re-raises cancellations and egress backpressure without capturing them. I generalized that to a small core marker.

- Add `ExpectedActivityError` to `posthog/temporal/common/errors.py`. Subclass it for a benign, non-defect condition raised out of an activity.
- The activity interceptor now re-raises `ExpectedActivityError` (any subclass) without reporting it to error tracking, alongside cancellations and `EgressBudgetExhausted`. The marker lives in core so the interceptor never has to import product code.
- `BillingLimitsWillBeReachedException` now subclasses `ExpectedActivityError`.
- Add `BillingLimitsWillBeReachedException` to the three `non_retryable_error_types` lists on the import activity so a billing halt fails fast instead of retrying. The workflow still receives the same exception type and still marks the job `BILLING_LIMIT_TOO_LOW`.

The exception class name is unchanged, so the workflow's `e.cause.type == "BillingLimitsWillBeReachedException"` check and the new `non_retryable_error_types` matching both keep working.

> [!NOTE]
> This is a sibling of #73126 (which excludes benign worker-shutdown errors the same way). Different exception, same class of fix. No overlap in files beyond the shared interceptor exclusion line.

Automated tests only. I (Claude) could not run the interceptor test locally because it needs the Temporal test server, which is not available in this environment. It mirrors the existing `test_egress_backpressure_is_not_captured` exactly.

- `test_expected_activity_error_is_not_captured` (core, `posthog/temporal/tests/common/test_error_tracking.py`): an `ExpectedActivityError` subclass raised in an activity is not sent to error tracking. Catches a regression where someone drops the marker from the interceptor's exclusion and billing (or any expected condition) resumes flooding error tracking. No existing test covers the marker path — the others cover cancellation and egress only.
- `test_billing_limit_exception_is_expected_activity_error` (product, `test_pipeline_utils.py`): locks that `BillingLimitsWillBeReachedException` subclasses the marker. Catches a revert of the base class back to `Exception`, which the core test (which uses its own subclass) would not catch. Ran locally, passes.

I did not add a test for the `non_retryable_error_types` change. Asserting retry counts needs a full workflow run and would be a brittle change-detector on config.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Triaged and authored by Claude (Opus 4.8) via PostHog Code. I confirmed the issue in error tracking, traced the raise site through the import activity to the capture point (the activity interceptor), and confirmed the workflow already handles the condition as a terminal `BILLING_LIMIT_TOO_LOW` status.

Decisions:
- Rejected excluding the exception in the interceptor by importing the product exception directly. `warehouse_sources` is isolated and core must not import product code, so I added a core-owned marker base class instead. Products opt in by subclassing.
- Rejected handling it by returning normally from the activity (the `CDCHandledExternally` pattern). That would move job-status ownership into the activity and touch the workflow's finally-block and post-import skip logic. The marker exclusion is smaller and matches the blessed interceptor pattern.
- Kept the workflow's existing `BILLING_LIMIT_TOO_LOW` branch untouched. Making the exception non-retryable just gets it there after one attempt instead of many.

Skills invoked: `/writing-tests` (applied the value gate before adding the two tests).

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/external_data_job.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/external_data_job.py
@@ -507,7 +507,7 @@ class ExternalDataJobWorkflow(PostHogWorkflow):
                     "start_to_close_timeout": dt.timedelta(weeks=1),
                     "retry_policy": RetryPolicy(
                         maximum_attempts=max_resumable_attempts,
-                        non_retryable_error_types=["NonRetryableException"],
+                        non_retryable_error_types=["NonRetryableException", "BillingLimitsWillBeReachedException"],
                     ),
                 }
             elif incremental_or_append:
@@ -515,14 +515,15 @@ class ExternalDataJobWorkflow(PostHogWorkflow):
                     "start_to_close_timeout": dt.timedelta(weeks=1),
                     "retry_policy": RetryPolicy(
                         maximum_attempts=max_incremental_attempts,
-                        non_retryable_error_types=["NonRetryableException"],
+                        non_retryable_error_types=["NonRetryableException", "BillingLimitsWillBeReachedException"],
                     ),
                 }
             else:
                 timeout_params = {
                     "start_to_close_timeout": dt.timedelta(hours=24),
                     "retry_policy": RetryPolicy(
-                        maximum_attempts=3, non_retryable_error_types=["NonRetryableException"]
+                        maximum_attempts=3,
+                        non_retryable_error_types=["NonRetryableException", "BillingLimitsWillBeReachedException"],
                     ),
                 }
 
--- a/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline/test/test_pipeline_utils.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/pipelines/pipeline/test/test_pipeline_utils.py
@@ -15,9 +15,12 @@ import structlog
 from dateutil import parser
 from structlog.types import FilteringBoundLogger
 
+from posthog.temporal.common.errors import NonReportableError
+
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.pipeline.consts import PARTITION_KEY
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.pipeline.utils import (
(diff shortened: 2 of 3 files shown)


Gilbert09:
Heads up on the extra commit here ([6c63180](https://github.com/PostHog/posthog/commit/6c63180d40d7bafc94a1411cc6a09180a63cce66)): after merging master in, the repo-wide mypy check was failing on `import_data_sync.py` — the `SchemaColumnTypeChangedException` branch still called `handle_non_retryable_error` with the old `(job_inputs, ...)` signature while the function now takes `(team_id, source_id, run_id, ...)`. This landed on master via a merge race between the signature refactor and #73135, so it was red on master independently of this PR. Corrected the call to match the sibling callers so CI can go green here.

PR #73138, opened 2026-07-23
chore(insights): prototype SQL insight view-time filter overrides [do not merge]

thmsobrmlr:
> [!WARNING]
> **Prototype — do not merge.** Throwaway branch to pick a design direction.

## TL;DR

SQL insights that use `{filters}` can only be filtered by editing the insight. This branch prototypes three UIs for setting temporary "view filters" on the insight page — like dashboard filters, but for a single insight. You flip between the three designs in the browser and pick one; the winner gets rebuilt properly, the rest gets thrown away.

## Problem

A SQL insight using the `{filters}` placeholder has no way to change date range or property filters in view mode — you must edit (and save) the insight. Dashboards can already override these via `filters_override`; the insight view page has the same machinery but no UI to drive it.

## Changes

Three structurally different variants on the existing insight view route, dev-only, gated by `?variant=`:

- **A — Classic insight card**: results in a bordered card like a trends insight, date range + property filters in the top row inside the border, immediate apply
- **B — Overrides panel**: button opens a staged panel showing saved filters next to your overrides, explicit Apply/Clear
- **C — Editable summary**: prose line ("Showing last 7 days where …") whose segments edit in place

All write the existing `?filters_override=` URL param, so results really re-run server-side into `{filters}`; nothing is persisted. A floating bottom bar switches variants (arrows or ←/→). See `frontend/src/scenes/insights/prototypes/sql-filter-overrides/PROTOTYPE.md`.

Try: open a saved SQL insight and append `?variant=A`.

Not manually tested (no dev stack in this environment); typecheck result reported in PR comments. It's a throwaway prototype — no tests by design.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed — prototype only.

## 🤖 Agent context

Human-driven (agent-assisted)

Built with the `/prototype` skill to answer: "what should view-time filter overrides on a SQL insight look like?" Once a variant wins, fold it into the real code and close this PR unmerged — this branch is the primary source for the losing variants.

---


thmsobrmlr:
Comment from PostHog Code: variant A (classic insight card) won, and the production implementation is #73375. Closing this prototype unmerged; the branch stays as the record of all three variants.

PR #73146, opened 2026-07-23
fix(data-imports): stop tracking non-JSON REST responses as error-tracking noise

Gilbert09:
## Problem

REST-based data warehouse sources hit an [error tracking issue](https://us.posthog.com/project/2/error_tracking/019f0562-bef7-70e3-b581-4f52d2448078) when an API serves a non-JSON body on an otherwise successful (2xx) response.
The common shape is an authenticated data request getting bounced to a login page: the API redirects to a signin page that returns HTML, JSON parsing fails, and the sync gives up.

The shared REST engine already classifies this correctly as non-retryable (`RESTClientNonRetryableError`), so it stops re-fetching a deterministic failure.
But the Temporal activity interceptor still captured it to error tracking on every occurrence.
One customer with expired credentials mints a large volume of `$exception` events for a condition nobody on our side can fix, which buries real defects.

A 2xx response with a non-JSON body is always a customer or upstream condition (expired auth, an HTML error page, a maintenance page), never a PostHog code defect, and retrying can't turn it into data.

## Changes

Follow the pattern the activity interceptor already uses for cancellations and egress backpressure: re-raise expected conditions without reporting them to error tracking.

- Add a `NonReportableError` marker base class in `posthog/temporal/common/errors.py`. Core owns it, so the isolated `warehouse_sources` product imports it downward (no core-to-product dependency).
- The activity interceptor (`posthog_client.py`) now skips capturing `NonReportableError`, alongside the existing cancellation and `EgressBudgetExhausted` skips.
- `RESTClientNonRetryableError` subclasses it. Its class name, message, and retry classification are unchanged, so the job still fails with the same customer-facing message. Only the internal error-tracking capture stops.

> [!NOTE]
> Retry behavior and the customer-facing outcome are unchanged. The job still fails with its message; this only quiets the internal `$exception` capture.

The change is scoped by exception type. `RESTClientNonRetryableError` is raised in exactly one place, for a 2xx response whose body is not JSON, so excluding it cannot mask a real bug.

Automated tests (authored and run by me, Claude, running autonomously):

- `test_non_json_response_body_is_non_retryable` in `test_rest_client.py` now also asserts the raised error is a `NonReportableError`. Ran locally, green (3 parameterized cases). Catches a revert of the subclassing that would silently re-flood error tracking with this exact issue.
- `test_non_reportable_error_is_not_captured` in `test_error_tracking.py`, mirroring the existing `test_egress_backpressure_is_not_captured` / `test_cancellation_is_not_captured` cases: the interceptor does not call `capture_exception` for a `NonReportableError`. Catches a regression where the interceptor stops honoring the marker.

The interceptor test needs a live Temporal server (the `temporal_client` fixture), which isn't reachable in the sandbox, so I couldn't run it here. The existing egress/cancellation siblings fail identically with the same connection error, confirming it's an infra gap not a test defect. They run in CI.

Also ran `ruff check`/`ruff format` (clean), `tach check --dependencies --interfaces` (the new cross-module import respects product isolation), and repo-wide `mypy --cache-fine-grained` (no issues in 16716 files). Invoked the `/writing-tests` skill before adding tests.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior, API, or documented workflow changes. This is error-tracking hygiene.

## 🤖 Agent context

Triaged from the error-tracking issue with the PostHog error-tracking MCP tools (`query-error-tracking-issue`, `query-error-tracking-issue-events`). The stack trace pointed at `rest_client._send_request`; the sampled events showed authenticated REST requests redirected to a login page returning HTML. I traced the capture from `_handle_import_error` and `handle_non_retryable_error` up to the shared activity interceptor, which is the single capture point.

Considered classifying it retryable (wrong, retrying an auth bounce is pointless) and suppressing all `NonRetryableException` broadly (out of scope and riskier). Chose a targeted per-type interceptor bypass, matching the maintainer's recent approach for worker-shutdown noise (#73126). The marker base lives in core so the product imports it downward, keeping product isolation intact.

---


Gilbert09:
Heads up on an extra change in this PR beyond the original scope.

While bringing `master` in, mypy failed on `import_data_sync.py` — but not because of this PR. The `handle_non_retryable_error` helper was widened on master to take `team_id`, `source_id` and `run_id` separately, and its three existing call sites were updated. A more recent change added a fourth call site (the `SchemaColumnTypeChangedException` branch) that still used the old 4-arg signature, so `Check static typing` was red for every PR merging with master.

I aligned that one call site with its three siblings (`00d3fff`). It's a one-line fix in the same file the REST non-retryable handling lives in, and it's what unblocks CI here.

PR #73151, opened 2026-07-23
fix(data-imports): skip warehouse enrichment when LLM gateway unset

Gilbert09:
## Problem

The data warehouse semantic-enrichment activity (`enrich_table_semantics`) runs after every sync to draft table and column descriptions for PostHog AI. Its optional LLM pass goes through `get_llm_client`, which raises `ValueError: LLM_GATEWAY_URL and LLM_GATEWAY_API_KEY must be configured` in any environment where the internal LLM gateway isn't set up.

The activity already degrades gracefully when the LLM call fails, but it does so by catching the error and calling `capture_exception` on it. So in environments without the gateway, this expected config condition gets reported to error tracking on every sync that reaches the LLM pass. It's a recurring, non-actionable exception with no per-table root cause. Reported by error tracking: https://us.posthog.com/project/2/error_tracking/019efee6-45ae-79c2-8263-52a01d6577c2

## Changes

Gate the LLM pass on the gateway actually being configured, the same precheck `products/conversations/backend/slack.py` already uses. When `LLM_GATEWAY_URL` or `LLM_GATEWAY_API_KEY` is unset, skip the AI pass and return a `skipped` / `llm_gateway_not_configured` status instead of letting `get_llm_client` raise into the broad `except` that captures it.

The gate sits after the canonical (documentation-sourced) descriptions are persisted, so those still land without a gateway, and a later sync fills in the AI descriptions once one is available. The real LLM-failure path (network errors, gateway 5xx, unparseable responses) is untouched and still captured.

> [!NOTE]
> This is a "stop tracking an expected condition" change, not a retry-policy change. It follows the same pattern as #69763 in the sibling data-modeling view enrichment.

Added `test_skips_llm_pass_without_capturing_when_gateway_not_configured` to the sync test suite. With the gateway unset it asserts the LLM call and `capture_exception` are both never invoked, the result is `skipped` / `llm_gateway_not_configured`, and canonical descriptions still persist while un-canonical columns are left for a later run. No existing test covered this path — `test_partial_status_when_llm_fails` mocks a generic failure and asserts capture *does* happen, which is the opposite case.

I (Claude) ran the full `TestEnrichTableSemanticsSync` suite locally against Postgres + ClickHouse; it passes. I did not run the change against a live import.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior or documented workflow changes.

## 🤖 Agent context

Triaged from an error-tracking webhook for the shared data-imports pipeline. The stack trace pointed at `enrich_table_semantics_sync` → `_generate_descriptions` → `get_llm_client`, which raises when the gateway env isn't configured. Confirmed via error tracking that it's a steady, high-volume, handled `ValueError` with no per-table cause — an environment condition, not a bug in the conversion/schema-evolution path.

Considered two fixes: catching the specific `ValueError` in the existing `except`, versus a preventive gate before the LLM call. Chose the gate — it matches the existing `conversations` precedent, avoids the wasted prompt-building work, and produces a clean telemetry status instead of a swallowed exception. Kept it strictly scoped so genuine LLM failures stay visible.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/workflow_activities/enrich_table_semantics.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/workflow_activities/enrich_table_semantics.py
@@ -22,6 +22,8 @@ import dataclasses
 from datetime import timedelta
 from typing import Any
 
+from django.conf import settings
+
 import posthoganalytics
 from temporalio import activity, workflow
 from temporalio.common import RetryPolicy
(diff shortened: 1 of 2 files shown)


Gilbert09:
Closing — not needed.

Investigation showed PostHog Cloud's warehouse enrichment was never actually hitting this path: the metadata worker has the gateway configured and enriches normally. The recurring `must be configured` exceptions in the error-tracking project were coming from an external self-hosted instance reporting its own telemetry through the bundled public key, not from Cloud. That source is now being filtered out upstream, so the gate here isn't required to silence the noise.

Happy to revisit if we later want self-hosted instances to skip the AI pass cleanly rather than capture an expected config condition, but that's a separate, lower-priority change.

PR #73157, opened 2026-07-23
fix(replay): jump to the first matching event in playlists

mp-hog:
## Problem

Playlists can be filtered by an event. The main case is the experiment recordings tab (behind feature flag, see #74955 for visuals), where the filter is the exposure event. The player should start just before that event instead of at 0:00. It doesn't.

The jump needs two things: the player granting permission (it does this at startup, unless the URL already names a timestamp), and the matching-event data arriving. Either can finish last. The old code only re-checked "ready to jump?" when the data arrived, never when permission was granted. On a single recording the data lands last, so it worked. In a playlist the data is already there and permission lands last, so nobody checked and nothing happened.

## Changes

The fix: Also check when permission is granted, one tick later so the player has finished setting itself up (a jump during setup is silently ignored).

Three guards:

- A manual seek or scrub cancels a pending jump. Data can arrive seconds into playback, and yanking the playhead after the viewer has moved is worse than not jumping. Play and pause don't cancel, since autoplay fires an internal seek that isn't cheap to tell apart.
- Jump once per recording. Filters can reload with no user intent (an async session id list resolving, as #74955 does), and the old code re-armed on every reload, so a late reload could snatch the playhead mid-playback. This is a behavior change: since the permission never resets, changing the filter while a recording is open no longer re-jumps - the jump only happens when a recording opens.
- Don't jump past the end. The query searches a wider window than the recording, and seeking past the end pins the last frame and auto-advances the playlist.

Note:
When the recordings tab is filtered by a metric (the dropdown in #74955), the filter carries metric events, not just the exposure. So the player can open at a metric event that happened after the exposure. That is "first relevant moment", not "first exposure". Once the experiments session context endpoint is fast enough, the seek target can be swapped to `first_exposure_timestamp` with no UX change.

- Locally against seeded data.
- New test cases

## 🤖 Agent context

Written with Claude Code (PostHog Code cloud task). Skills invoked: `/writing-tests`.

The original plan was an experiment-tab-specific auto-seek built on the playlist's matching events. Turned out the generic skip mechanism already covers that case, and the real bug was the trigger gap above, so I fixed the existing path instead of adding a parallel one. The generic replay page with event filters gets the fix for free. A separate dismissible "jump to start" indicator was considered and dropped in favor of the existing overlay.

---


TueHaulund (commented):
Nice root-cause framing, and fixing the existing generic path rather than adding an experiment-specific mechanism means the plain replay page with event filters gets this for free. No blockers from me, but a couple of things worth a look before merge.

One thing for the description: since the flag never resets, changing the filter while a recording is open won't re-jump anymore. Probably fine, but it's different from before and worth a line.

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.ts
```diff
@@ -2159,7 +2185,13 @@ export const playerInspectorLogic = kea<playerInspectorLogicType>([
             const earliestMatchingEvent = values.matchingEvents.reduce((previous, current) =>
                 previous.timestamp < current.timestamp ? previous : current
             )
-            const { timeInRecording } = timeRelativeToStart(earliestMatchingEvent, values.start)
+            const { timestamp, timeInRecording } = timeRelativeToStart(earliestMatchingEvent, values.start)
+            // The matching-events query has slack around the recording window, so the earliest
+            // match can fall past the playable range — seeking there would pin the player to its
+            // final frame and immediately trigger end-reached (auto-advancing playlists).
+            if (values.end && timestamp.isAfter(values.end)) {
+                return
```
TueHaulund, line 2201: This return skips `markSkippedToFirstMatchingEvent()`, so the "already skipped" flag never gets set. If matching events reload later with an earlier match, we'll jump the playhead mid-playback, which is what that flag is there to stop.

```suggestion
            if (values.end && timestamp.isAfter(values.end)) {
                actions.markSkippedToFirstMatchingEvent()
                return
            }
```

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.ts
```diff
@@ -2159,7 +2185,13 @@ export const playerInspectorLogic = kea<playerInspectorLogicType>([
...
+            }
             const seekTime = Math.max(0, ceilMsToClosestSecond(timeInRecording) - 1000)
 
             actions.markSkippedToFirstMatchingEvent()
```
TueHaulund, line 2205: We mark the skip as done here, but the actual `seekToTime` below does nothing if `currentTimestamp` isn't set yet. Right now it works because the player sets both things in the same tick, but that's easy to break by accident, and if it breaks the skip just stops happening with no retry.

Could we check `currentTimestamp` in the guard above instead? Then a player that isn't ready yet waits for the next `loadRecordingMetaSuccess` instead of using up the flag, and we don't need the `breakpoint(1)`.

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.test.ts
```diff
@@ -281,6 +281,186 @@ describe('playerInspectorLogic', () => {
...
+            } as SessionRecordingType)
+
+            await expectLogic(matchingLogic)
+                .toDispatchActions(['loadRecordingMetaSuccess', 'trySkipToFirstMatchingEvent'])
+                .delay(25)
+            await expectLogic(playerLogic).toNotHaveDispatchedActions(['seekToTime'])
+
+            matchingLogic.unmount()
+            playerLogic.unmount()
+        })
```
TueHaulund, line 434: The whole fix relies on autoplay using `seekToTimestamp` and only real user seeks using `seekToTime`. No test covers that. If someone later switches an internal seek to `seekToTime`, this silently breaks. Worth a test that autoplay doesn't cancel the skip.

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.test.ts
```diff
@@ -281,6 +281,186 @@ describe('playerInspectorLogic', () => {
...
+                start_time: '2025-01-01T00:00:00.000Z',
+                end_time: '2025-01-01T00:01:00.000Z',
+                recording_duration: 60,
+            } as SessionRecordingType)
+
+            // Covers both skip triggers: the synchronous one on meta load and the deferred one
+            // from the player arming the flag during initialization
+            await expectLogic(matchingLogic)
+                .toDispatchActions(['loadRecordingMetaSuccess', 'trySkipToFirstMatchingEvent'])
+                .delay(25)
```
TueHaulund: The 25ms waits paired with `toNotHaveDispatchedActions` will be flaky (same pattern below on the other two new tests). On a slow CI runner the wait can finish before the code even runs, so the test passes without checking anything. Fake timers or waiting on something concrete would be safer.

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.test.ts
```diff
@@ -281,6 +281,186 @@ describe('playerInspectorLogic', () => {
...
+                    ],
+                },
+            }
+            const playerLogic = sessionRecordingPlayerLogic(matchingProps)
+            const matchingLogic = playerInspectorLogic(matchingProps)
+            playerLogic.mount()
+            matchingLogic.mount()
+
+            await expectLogic(matchingLogic).toDispatchActions(['loadMatchingEventsSuccess'])
+            interact(playerLogic)
```
TueHaulund, line 361: `interact(playerLogic)` happens before the recording metadata loads, so there's nothing for the user to scrub yet. The case worth testing is a seek after metadata loads but before matching events arrive. Moving `interact()` below `loadRecordingMetaSuccess` covers that.

frontend/src/scenes/session-recordings/player/inspector/playerInspectorLogic.test.ts
```diff
@@ -281,6 +281,186 @@ describe('playerInspectorLogic', () => {
             matchingLogic.unmount()
             playerLogic.unmount()
         })
+
+        it('seeks when the player arms the skip flag after matching events have loaded', async () => {
```
TueHaulund: These three new tests are nearly the same except for the playerKey, timestamp, and flag. Could be `it.each` rows like the ones you already use below.

frontend/src/scenes/session-recordings/playlist/sessionRecordingsPlaylistLogic.test.ts
```diff
@@ -1046,6 +1048,64 @@ describe('sessionRecordingsPlaylistLogic', () => {
         })
     })
 
+    describe('matchingEventsMatchType', () => {
```
TueHaulund, line 1049: Most of these rows test existing behavior this PR doesn't touch. The `'only an event property filter' -> 'backend'` one is useful since this feature depends on it, the rest could go.

mp-hog:
Thanks @TueHaulund for the great review, all points addressed!

TueHaulund (approved):

PR #73160, opened 2026-07-23
chore(ci): add no-ci label to skip heavy CI on draft PRs

thmsobrmlr:
**TL;DR:** Put the `no-ci` label on a **draft** PR and the four heavy CI suites (backend, frontend, storybook, MCP) stop running on it. Remove the label or mark the PR ready for review and full CI comes back. Meant for prototypes that don't need CI on every push.

## Problem

Prototype and spike PRs burn runner credits on every push even though nobody needs the results yet.
Drafts already narrow the heavy suites, but a narrowed backend + frontend + storybook run per push is still real cost on a PR that may never merge.

**Why:** asked as "is there an easy way to add a no-ci label for things like prototypes?"

## Changes

- A draft PR carrying the `no-ci` label now skips Backend CI, Frontend CI, Storybook, and MCP CI entirely.
- The guard is one `if:` on each workflow's `changes` entry job. Every other job `needs` it, so skipping it skips the whole workflow, and the "Tests Pass" gates already treat skipped as success (the same state a docs-only or fork PR produces today).
- Backend CI's `test-selection-verdict` telemetry job now also skips when `changes` was skipped, since no tests ran and there is nothing to verdict.
- I created the `no-ci` label on the repo: "Skip backend/frontend/storybook/MCP CI on this draft PR (prototypes). No effect on ready PRs."

Safety and scope:

> [!NOTE]
> The label is draft-only by design. All four guarded workflows re-trigger on `ready_for_review`, so marking a muted PR ready runs full CI before it can merge. The label can never produce a mergeable PR with green-but-empty checks.

- Only workflows that already trigger on `ready_for_review` got the guard. ci-python, ci-nodejs, ci-rust, the container builds, and turbo don't re-run on ready today, so muting them via label would let a muted draft go ready with stale green aggregators. They can join later by adding the `ready_for_review` trigger type first.
- ci-e2e-playwright already skips drafts entirely, and MCP integration tests were already draft-gated, so nothing to do there.
- Backend/frontend/storybook trigger on `labeled`/`unlabeled`, so the label takes effect immediately and cancels the in-flight run via the concurrency group. MCP CI picks it up on the next push.
- Zero-config alternative that already exists: GitHub natively skips all `push`/`pull_request` workflows when the PR's head commit message contains `[no ci]` or `[skip ci]`. That mutes everything and saves the run dispatches too, but it's per-commit. The label is per-PR and reversible without rewriting history.

- `bin/hogli lint:workflows` passes (6 checks, 103 workflows).
- `actionlint` on the four edited files reports only the two findings already present on master (`job.check_run_id` in ci-storybook, `node24` in the vendored paths-filter under actionlint 1.7.7). No new findings.
- I (Claude) traced every `always()` / `!cancelled()` gate in the guarded workflows (`django_tests`, `frontend_tests`, `visual_regression_tests`, `vr-complete`, MCP's aggregator) to confirm a skipped `changes` job reads as pass, not failure. ci-rust and container-images-ci already skip `changes` on fork PRs, which exercises this exact state in production today.
- Live check on this PR: since `pull_request` runs use the PR's own workflow definitions, flipping the `no-ci` label on this draft demonstrates the skip end to end.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No `docs/` page covers CI labels. Following the `run-ci-backend` / `run-ci-frontend` precedent, the behavior is documented in the workflow trigger comments and the label description.

## 🤖 Agent context

- Authored by Claude (PostHog Code cloud task) from a question by the assignee. Skill invoked: `/authoring-ci-workflows`.
- Rejected a central hook in the vendored paths-filter action (forcing all filter outputs to `false` on labeled PRs would make the filter lie about which paths changed, a trap for any future non-gating consumer). Explicit per-workflow guards match the existing label conventions.
- Rejected guarding workflows without a `ready_for_review` trigger: their aggregators report success on the muted run, and with no re-run on ready, a muted draft could go ready and merge on stale green checks.
- Found and avoided a trap in rust-docker-build: skipping its `affected` job makes `build-images` fall back to building every image (the `workflow_dispatch` escape hatch), the opposite of no-ci. That workflow is deliberately out of scope.

---

### DIFF
--- a/.depot/workflows/ci-backend.yml
+++ b/.depot/workflows/ci-backend.yml
@@ -92,7 +92,8 @@ on:
         # still skip drafts. Ready PRs run the full matrices — the merge gate —
         # and ready_for_review re-triggers them when a PR leaves draft. Add the
         # `run-ci-backend` label (labeled/unlabeled re-trigger) to force the full
-        # matrices on a draft. Mirrors canonical.
+        # matrices on a draft; the `no-ci` label silences the workflow on a draft
+        # entirely. Mirrors canonical.
         types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
 concurrency:
     # PRs: one active run per branch, cancel stale. Push: per-SHA so master
@@ -186,9 +187,14 @@ jobs:
     # See .github/actions/paths-filter/README.md for filter semantics
     changes:
         # Root of the `needs` graph: skipped when `sample` says skip (or is itself skipped
-        # at percent 0/unset), which cascades the whole workflow off.
+        # at percent 0/unset), which cascades the whole workflow off. Also skipped on
+        # draft PRs labeled `no-ci`, mirroring canonical.
         needs: sample
-        if: needs.sample.outputs.sampled == 'true'
+        if: >-
+            needs.sample.outputs.sampled == 'true'
+            && (github.event_name != 'pull_request'
+                || github.event.pull_request.draft != true
+                || !contains(github.event.pull_request.labels.*.name, 'no-ci'))
         runs-on: depot-ubuntu-24.04
         timeout-minutes: 5
         name: Determine need to run backend and migration checks
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -19,6 +19,9 @@ on:
         # it when a PR leaves draft. To force the full matrices on a draft, add the
         # `run-ci-backend` label — labeled/unlabeled re-trigger the run so the matrices
         # start without needing a new push. Cheap checks still run on drafts.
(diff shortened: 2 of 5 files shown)


gantoine (approved):
sure why not, though with the merge queue coming we plan to stop running heavy/expensive CI workflows on draft PRs altogether

thmsobrmlr:
ah nice, good to know 👍

thmsobrmlr:
### Merge activity

* **Jul 23, 12:54 PM UTC**: @thmsobrmlr merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/73160).

PR #73161, opened 2026-07-23
feat(aio): deterministic usage report timestamps and log levels

carlos-marchal-ph:
## Problem

The daily AI observability usage report (`send_ai_observability_usage_reports`, one `llm analytics usage` event per org per day) has two reliability warts, both observed in production ([context](https://posthog.slack.com/archives/C0A006STKHR/p1784289795020679?thread_ts=1784261764.212509&cid=C0A006STKHR)):

- Events are stamped at capture time, so Celery retry stragglers trail the 04:15 UTC run by hours. Anything past 07:00 UTC crosses midnight in the project timezone and lands on the wrong chart day in every dashboard built on these events.
- Every retryable failure logs at error level, so the log alert pages on transient blips that recover on retry, and can't tell "retrying" from "gave up".

## Changes

All in `posthog/tasks/ai_observability_usage_report.py`, no migrations, independent of the manual-trigger PR:

- **Deterministic timestamps.** Every emitted event is stamped at midnight UTC after the covered day (`period_start + 1 day`), scheduled runs included. This is the instant manual backfills already targeted, and it preserves the historical chart-day bucketing in both UTC and project timezones, so there is no series discontinuity. Actual arrival time stays queryable via `events.created_at`. Note the chart semantics are unchanged on purpose: the value on chart day N still describes day N-1's usage, same as the old arrival-time behavior.
- **`triggered_by` property** (`scheduled` | `manual`) on each report, inferred from an explicit `at` or `organization_ids` being passed. Backfills now land on the exact stamp the missed run would have used, so this is what keeps them distinguishable.
- **Honest log levels.** Both tasks are now `bind=True`. Retryable attempt failures log at warning (messages unchanged, tracebacks kept via `exc_info`), and a single terminal `[AIO Usage Error] usage report run failed permanently` error fires only when the failure is permanent (`_is_final_attempt`: direct synchronous call, or autoretries exhausted). The existing severity-keyed log alert therefore stops paging on transients with zero config change.

Deliberate divergence: the org usage report (`posthog/tasks/usage_report.py`) still stamps scheduled runs at arrival time. We are proving the deterministic pattern here first; backporting it there is the natural follow-up if it holds.

> [!NOTE]
> Two silent-loss shapes remain out of scope and are deferred to the upcoming run-records PR: a queued retry lapsing past `expires` (4h) without a final attempt, and worker-lost redelivery loops (`acks_late` + `reject_on_worker_lost` never increment `retries`). Neither emits the terminal error line; the outcome-based volume alert covers them meanwhile.

Full test file passes (28 tests, ClickHouse-backed) plus repo-wide `mypy` and `ruff` clean. New coverage, each catching a regression nothing previously caught:

- Parameterized timestamp/trigger test (report gathering mocked, so it pins exactly the emission wiring): scheduled, manual-by-date, and manual-by-org-filter runs must produce the exact deterministic `at_date` and `triggered_by` values. Catches reverting to capture-time stamps (the smearing bug) or mislabeled backfills.
- Parameterized log-severity tests for both tasks, simulating real worker context (`push_request` with `called_directly=False`/`is_eager=True`, terminal retries derived from `max_retries`): retryable failures must log warning only, exhausted retries exactly one terminal error, and direct synchronous calls the terminal error too. Catches per-attempt error logs coming back (the false-page bug) and the alert's trigger line silently disappearing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed, internal reporting task only.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Driver: Carlos Marchal.

Built with Claude Code from a spec + plan reviewed by Carlos, grounded in production data (actual stamp distributions and straggler trails queried from the internal usage analytics project). Skills used: superpowers (brainstorming, writing-plans, subagent-driven-development), /writing-tests, and two independent no-context pr-review agent passes whose findings drove a final polish round (org-filter trigger inference, `_is_final_attempt` with the `called_directly` disjunct following the `heatmap_screenshot` precedent, lighter tests).

Decisions: stamp midnight-after-the-covered-day rather than the covered day itself (continuity with historical bucketing beats intuitiveness); terminal-error wrapper covers the gathering phase only, where every historical failure lived (widening via `on_failure` considered and deferred to the run-records PR); ClickHouse-side dedup rejected earlier in the effort (merge-time-only, no `FINAL` at query time). One test-mechanics note: simulated request contexts pass `is_eager=True` because Celery's real autoretry wrapper otherwise synchronously re-executes the body under the eager test app, doubling asserted log calls — verified against Celery source.


posthog/tasks/ai_observability_usage_report.py
```diff
@@ -742,18 +756,21 @@ def _get_all_ai_observability_reports(
         llm_prompt_fetched_counts = get_llm_prompt_fetched_counts(period_start, period_end, team_ids)
         logger.info(f"Retrieved prompt fetched counts for {len(llm_prompt_fetched_counts)} teams")
     except Exception as err:
-        logger.exception("Failed to query LLM prompt fetched counts, continuing without prompt fetch metrics")
+        logger.warning(
```
Radu-Raicea, line 759: This failure is swallowed after the prompt-count query has exhausted its own retries, so it never reaches the new final-attempt error log. Downgrading it to warning lets the task emit zero prompt-fetch counts while the severity-keyed alert stays quiet.

posthog/tasks/test/test_ai_observability_usage_report.py
```diff
@@ -1094,3 +1097,123 @@ def test_send_ai_observability_usage_reports_with_at_parameter(
         report_dict = call_args[1]["report_dict"]
         assert report_dict["ai_generation_count"] == 5
         assert report_dict["ai_embedding_count"] == 0  # Jan 9th events not included
+
+    @parameterized.expand(
```
Radu-Raicea, line 1106: These blocks expand to nine tests, and every case pays for Postgres fixtures plus ClickHouse STOP/START MERGES even though all database/query boundaries are mocked. This could be improved.

Radu-Raicea (approved):

posthog/tasks/ai_observability_usage_report.py
carlos-marchal-ph: yep, intentional, prompt-fetch is best-effort, so on failure it degrades to zero counts, logs warning + capture_exception, and deliberately doesn't re-raise so it never reaches the final-attempt error path, unlike the `teams/metrics/breakdowns/surveys` phases which do

posthog/tasks/test/test_ai_observability_usage_report.py
carlos-marchal-ph: Fixed!

PR #73165, opened 2026-07-23
fix(conversations): render and serialize standard rich text nodes

christiaan-ph:
## Problem

Conversations ticket messages accept `rich_content` (TipTap JSON) through the comments API. The API checks size and JSON-serializability but imposes no node-type contract, so any API client can legitimately send standard rich text nodes: ordered and bullet lists, headings, blockquotes, strikethrough.

The product only handled the narrow subset its own composer can author (paragraphs, code blocks, images), on three fronts:

- Rendering: the agent ticket view parses `rich_content` against the composer's schema. TipTap discards the entire doc when it meets an unknown node type, so a message containing a numbered list rendered as a completely empty bubble, even though the data was intact.
- Serializing: the outbound markdown and Slack serializers silently dropped any block node they didn't recognize from GitHub replies and Slack messages.
- Styling: even once a list rendered, Tailwind's preflight zeroes list markers and heading sizes, so lists appeared as bare stacked paragraphs and headings as plain text.

This is live today: our support agent UI is one such API client, and its replies containing lists were invisible in the ticket view. The Conversations composer also had no way to author lists at all.

## Changes

Composer:

- Bullet and numbered lists enabled (StarterKit list extensions, toolbar buttons, Cmd+Shift+8 / Cmd+Shift+7).
- `serializeToMarkdown` serializes list nodes (including nesting and ordered `start`), keeping the plain `content` field faithful. That field is what the customer-facing widget renders, so parity here is customer-visible.

Rendering:

- `SUPPORT_PREVIEW_EXTENSIONS`, a read-only schema for `SupportRichContentPreview` covering the standard StarterKit node set (lists, headings, blockquote, strike, horizontal rule) rather than only what the composer can author. Authoring and display are different contracts: the composer stays restricted as a product choice, display must accept what the API accepts.
- If a doc still fails schema validation (a future unknown node), the preview falls back to rendering the message's plain `content` as markdown instead of a blank bubble, so schema drift degrades to plain text rather than invisible messages.
- The fallback preserves the per-message image policy: Zendesk-imported messages keep inline images disabled (`fallbackDisableImages`), matching the existing non-rich markdown path. A Zendesk body can carry original auth-gated image URLs in its plain text even though re-hosted attachments render fine as image nodes, so the fallback must not start rendering those.

Styling (`SupportEditor.scss`):

- A shared `rich-text-block-styles` mixin restores list markers and indentation and tames heading sizes, blockquotes, horizontal rules, and strikethrough, applied under both `.SupportEditor` (composer) and `.SupportRichContentPreview` (ticket view). Follows the existing LemonMarkdown / RichMarkdownEditor pattern. Without it Tailwind preflight leaves lists as stacked paragraphs and headings at browser defaults.

Outbound serializers (`formatting.py`):

- `rich_content_to_markdown` (feeds GitHub replies and Slack text) now serializes lists, blockquotes, headings, horizontal rules, and strikethrough.
- List items and blockquotes now serialize any block child (code block, heading, blockquote, nested list, image), not just paragraphs, in both the markdown and email HTML paths. Previously those blocks were dropped. The list serializers, the blockquote branches, and the top-level dispatch all share one per-block helper so the paths cannot drift.
- `rich_content_to_slack_blocks` returns `None` when a doc contains block nodes Slack rich_text can't represent, and `rich_content_to_slack_payload` then sends the complete markdown text instead. No content is silently lost, and we avoid maintaining a second list renderer.
- `rich_content_to_html` (email) gains nested lists, ordered-list `start`, headings, horizontal rules, strikethrough, and renders unknown blocks' inline content as a paragraph rather than dropping it.
- Slack styles now map `strike` in both directions.

Inbound:

- Slack `rich_text_list` blocks become real `bulletList`/`orderedList` nodes instead of paragraphs prefixed with a literal bullet character, which was a workaround for the schema gap this PR closes.

> [!NOTE]
> Behavior change for Slack-channel replies containing lists: they are delivered as mrkdwn text rather than rich_text blocks. Text fallback is complete; block-level styling (underline) is traded for not losing list content.

> [!NOTE]
> Inbound Slack nested lists (consecutive `rich_text_list` elements at increasing `indent`) still flatten into sibling top-level lists. Item text and order are preserved; only the nesting hierarchy is lost. This is not a regression (the pre-PR path also carried no indent) and is left as a focused follow-up.

Automated (run locally):

- `products/conversations/backend/api/tests/test_formatting.py` (48 passed) including `TestRichContentBlockNodes`: markdown list/heading/blockquote/rule/strike serialization catches the previous silent drop in the GitHub and Slack text paths, the Slack payload fallback test catches blocks being emitted with missing content, inbound Slack list cases catch a regression back to flattened bullet paragraphs, the HTML test locks in nested list, heading, rule, and unknown-block handling for email, and new cases lock in that non-paragraph blocks inside both list items and blockquotes survive in the markdown and HTML paths.
- `products/conversations/frontend/components/Editor/SupportEditor.test.ts` (6 passed): `serializeToMarkdown` list cases guard plain-content parity (what the widget renders), and the preview-schema cases guard both directions of the blank-message bug.
- `ruff`, `stylelint`, and `ty check` clean on the changed files.

Manual (local stack, seeded ticket walking every case):

- Numbered and bullet lists render with markers and indentation in the ticket view (the original blank-bubble regression is gone).
- Headings and blockquotes render at sensible sizes rather than plain text or browser defaults.
- Nested lists indent children under their parent item.
- A code block placed inside a list item renders in full (the block-in-list-item fix, visually confirmed).
- A message with an unsupported node falls back to plain markdown instead of a blank bubble.
- Strikethrough renders inside list items.
- Composer authoring: the new bullet and numbered list toolbar buttons (and Cmd+Shift+8 / Cmd+Shift+7) produce styled lists as you type.

Manual (live Slack round-trip: local stack connected to a real Slack workspace, inbound events over a tunnel):

- Inbound: a Slack message with bullet and numbered lists became a ticket message with real `bulletList`/`orderedList` nodes (verified in the stored `rich_content`) and renders with markers in the ticket view. Slack strikethrough arrives as a `strike` mark and renders struck.
- Outbound lists: replies with bullet and ordered lists arrived in the Slack thread complete via the mrkdwn text fallback, including a Tab-nested list whose indentation survives.
- Outbound without lists still posts rich_text blocks, verified with an underlined word (underline only exists on the blocks path).
- Outbound escaping: a punctuation-heavy reply (`my-file_v2.txt`, `old-endpoint`, `(costs $5, roughly 2-3 mins)`) arrived with no leaked backslash escapes.
- The review-fix paths, via API-seeded replies since the composer intentionally can't author them: a code block inside a list item survives to Slack; strike inside a list-containing doc converts to Slack's single-tilde form and renders struck; a doc with an unknown block node keeps its inline text in the Slack message and renders the plain-content fallback (not a blank bubble) in the ticket view; a blockquote with code-block and list children arrives complete as quoted mrkdwn.
- Known cosmetic artifact, not a content loss: Slack renders a fenced code block inside a quoted section with the `>` markers visible inside the code box, since mrkdwn has no representation for that nesting.
- Browser console shows no tiptap duplicate-extension warning for `underline` on the ticket view.

Not exercised live: GitHub reply and email outbound. Deliberate, not an oversight:

- GitHub replies send `rich_content_to_markdown` output, which is the exact markdown the Slack text fallback delivered in the live tests above, so the serialization is already verified end to end; the only GitHub-specific code is delivery plumbing this PR doesn't touch, and GitHub renders standard markdown natively.
- Email's only change is `rich_content_to_html`. Instead of wiring up an email channel to test unchanged delivery plumbing, I rendered the function over the same set of test docs (lists, nesting, ordered `start`, code-in-list, strike, unknown block, blockquote with block children, headings, rule, escape-prone punctuation) and inspected the HTML: correct nesting, `<ol start>`, `<s>`, `<pre><code>` inside `<li>` and `<blockquote>`, unknown-block inline text preserved, no leaked escapes. That plus the formatting tests covers the delta.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs describe the support composer's formatting options; nothing to update.

## 🤖 Agent context

- Authored by Claude Code in sessions driven by @christiaan-ph, starting from a support investigation into a blank reply on an internal ticket. Root cause was confirmed against TipTap source: an unknown node type in JSON content falls back to an empty doc.
- A follow-up session addressed review feedback: the SCSS styling for lists, headings, and blockquotes, and generalizing the list-item and blockquote serializers so non-paragraph blocks are no longer dropped.
- Skills invoked: `/writing-tests`.
- Rejected approach: implementing Slack `rich_text_list` block emission. Chose the fall-back-to-text guard instead to avoid a second hand-rolled list renderer; markdown remains the single complete serialization.
- The customer-facing widget maps only the plain `content` field (`sidepanelTicketsLogic.ts`), which is why `serializeToMarkdown` parity is treated as customer-visible behavior in the tests.


products/conversations/backend/formatting.py
```diff
@@ -455,6 +463,54 @@ def _serialize_inline_nodes_to_markdown(nodes: list[JSON], include_images: bool
...
+            if child_type in ("bulletList", "orderedList"):
+                nested = _serialize_list_to_markdown(child, child_type == "orderedList", child_indent, include_images)
+                if nested:
+                    blocks.append((nested, True))
+            elif child_type == "paragraph":
+                text = _serialize_inline_nodes_to_markdown(
+                    child.get("content", []), include_images=include_images
+                ).strip()
+                if text:
+                    blocks.append((text, False))
```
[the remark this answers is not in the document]
christiaan-ph: Fixed in 7ba5769. List items now serialize any block child (code block, heading, blockquote, image), not just paragraphs and nested lists, in both `rich_content_to_markdown` and the email HTML serializer. Both list serializers and the top-level dispatch now share a single per-block helper (`_serialize_block_node_to_markdown` / `_serialize_block_node_to_html`), so the two paths can't drift back apart. Added two cases to `TestRichContentBlockNodes` covering a code block inside a list item for the markdown and HTML paths.

products/conversations/backend/formatting.py
```diff
@@ -472,6 +528,37 @@ def rich_content_to_markdown(rich_content: JSON | None, include_images: bool = T
...
+
+        if node_type == "blockquote":
+            quote_lines: list[str] = []
+            for child in node.get("content", []):
+                if child.get("type") == "paragraph":
+                    text = _serialize_inline_nodes_to_markdown(child.get("content", []), include_images=include_images)
+                    quote_lines.extend(f"> {line}".rstrip() for line in text.split("\n"))
+            if quote_lines:
+                blocks.append("\n".join(quote_lines))
+            continue
```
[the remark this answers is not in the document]
christiaan-ph: Fixed in a18cf12. Blockquotes now serialize any block child (nested list, code block, heading, image), not just paragraphs, in both `rich_content_to_markdown` and the email HTML serializer, reusing the same shared per-block helpers introduced for list items. This closes the same silent-drop the PR fixes elsewhere so the "no content is silently lost" invariant holds one level down too. Added a `blockquote_wrapping_list` case to the markdown parameterized test and a matching HTML blockquote test.

products/conversations/backend/formatting.py
```diff
@@ -380,14 +384,16 @@ def slack_blocks_to_rich_content(blocks: list[JSON] | None, user_names: dict[str
...
                         continue
                     inline_nodes = _parse_rich_text_inline_elements(list_item.get("elements", []), user_names)
                     if inline_nodes:
-                        prefix = "• "
-                        inline_nodes.insert(0, {"type": "text", "text": prefix})
-                        doc_nodes.append({"type": "paragraph", "content": inline_nodes})
+                        items.append({"type": "listItem", "content": [{"type": "paragraph", "content": inline_nodes}]})
+                if items:
+                    doc_nodes.append({"type": list_type, "content": items})
                 continue
```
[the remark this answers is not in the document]
christiaan-ph: Real gap, but leaving it for a focused follow-up rather than this PR. It isn't a regression: the pre-PR path emitted flat `• `-prefixed paragraphs, so inbound Slack list nesting was never preserved on this path. All item text and order are retained; only the indent hierarchy flattens into sibling top-level lists, and the trigger (a nested list inside an inbound Slack support message) is uncommon. The correct fix is an indent-tracking pass that reparents higher-indent `rich_text_list` elements into the previous list's last item, which is worth doing on its own rather than expanding this change. Noted as a follow-up in the PR description.

christiaan-ph:
should be working now, awaiting final review from @PostHog/conversations

products/conversations/backend/formatting.py
```diff
@@ -455,6 +463,98 @@ def _serialize_inline_nodes_to_markdown(nodes: list[JSON], include_images: bool
...
 
+def _list_start(node: JSON) -> int:
+    attrs = node.get("attrs") or {}
+    try:
+        return max(int(attrs.get("start") or 1), 1)
+    except (TypeError, ValueError):
+        return 1
+
+
+def _serialize_block_node_to_markdown(node: JSON, include_images: bool = True) -> str:
```
veryayskiy, line 508: _serialize_block_node_to_markdown returns "" for unrecognized node types, so an unknown block is dropped from markdown — and markdown is exactly what the Slack text fallback and GitHub replies send. The HTML path deliberately renders unknown blocks' inline content as <p>; mirror that here (emit the inline content as a paragraph) or the "no content is silently lost" guarantee only holds for email.

products/conversations/backend/formatting.py
```diff
@@ -429,6 +435,8 @@ def _serialize_text_node_to_markdown(node: JSON) -> str:
             text = f"**{text}**"
         elif mark_type == "italic":
             text = f"*{text}*"
+        elif mark_type == "strike":
+            text = f"~~{text}~~"
```
veryayskiy, line 473: strike is now emitted into markdown (~~x~~) and the new Slack text fallback routes any list-containing doc through content_to_slack_mrkdwn, but that function has no strike mapping, so Slack receives literal ~~gone~~ instead of ~gone~ (and slack_mrkdwn_to_content never turns inbound ~x~ back into a strike mark, unlike the block path you just taught about strike). Add both directions next to the existing bold/italic rules

products/conversations/frontend/components/Editor/SupportEditor.tsx
```diff
@@ -248,6 +262,22 @@ export const SUPPORT_EXTENSIONS = [
     SupportCodeBlockExtension,
 ]
 
+export const SUPPORT_PREVIEW_EXTENSIONS = [
```
veryayskiy, line 268: SUPPORT_PREVIEW_EXTENSIONS adds Underline on top of an un-narrowed StarterKit (3.20 bundles the underline extension), which trips tiptap's duplicate-extension-name warning, and LinkOnPasteExtension is dead weight in a disabled: true editor. Either set underline: false in the StarterKit config or drop the standalone extension, and leave the paste handler to the composer list.

products/conversations/backend/formatting.py
christiaan-ph: Fixed in 67aeaf164cf. `_serialize_block_node_to_markdown` now falls back to serializing an unknown block's inline content, matching what the HTML path does, so the guarantee holds for GitHub replies and the Slack text fallback too. I also restructured the `image` branch to return explicitly, otherwise an image with `include_images=False` would have fallen through into the new fallback. Added an `unknown_block_keeps_inline_content` case to the markdown parameterized test.

products/conversations/backend/formatting.py
christiaan-ph: Fixed in 74e98743f7d. Added `~~x~~` to `~x~` in `content_to_slack_mrkdwn` and `~x~` to `~~x~~` in `slack_mrkdwn_to_content`, next to the existing bold and italic rules. The inbound regex uses the same negative lookarounds, so an already-doubled `~~x~~` is left alone. Covered by a new `mrkdwn_strike` inbound case and `test_slack_payload_maps_strike_to_slack_mrkdwn`, which asserts a strike inside a list item reaches Slack as `- ~gone~`.

products/conversations/frontend/components/Editor/SupportEditor.tsx
christiaan-ph: Fixed in 02b17c972a8. Confirmed StarterKit 3.20.6 does bundle `@tiptap/extension-underline` and that `resolveExtensions` in `@tiptap/core` warns on the duplicate name, so I set `underline: false` in the StarterKit config and kept the explicit standalone `Underline`. Dropped `LinkOnPasteExtension` from the preview list. `SUPPORT_EXTENSIONS` had the same duplicate registration on master, so it gets `underline: false` too rather than leaving the identical warning two lines above.

christiaan-ph:
Thanks @veryayskiy, all three are fixed and pushed. Doing a last round of manual testing before this is ready.

Beyond the three review comments, chasing the Slack fallback turned up a regression this PR introduced, plus a pre-existing bug next to it.

**Markdown escapes were reaching Slack.** `rich_content_to_markdown` escapes markdown syntax characters in text nodes. Before this PR that was harmless for Slack, because list-containing docs still produced `rich_text` blocks and the escaped `text` field was only a notification preview. Once `rich_content_to_slack_blocks` started returning `None` for those docs, `text` became what the customer actually reads, and mrkdwn has no backslash escape syntax, so a reply went out as:

```
Two options \(pick one\):
1. Use query\-time properties, e\.g\. person\.email
```

Since `.`, `-`, `(` and `)` appear in almost every sentence, this hit effectively every list-containing Slack reply. `content_to_slack_mrkdwn` now masks escaped characters before the syntax conversions and restores them as literals, which also stops an escaped `*` being reinterpreted as emphasis.

**Bold rendered as italic.** Pre-existing, and exposed by the same change. `**bold**` was converted to `*bold*`, which the italic rule then rewrote to `_bold_`. Bold now goes through the same placeholder masking that bold-italic already used.

One limitation stays: a literal asterisk cannot survive to Slack, because mrkdwn offers no way to escape one. Showing `*lit*` is still better than showing `\*lit\*`.

Also tidied up while in the area: the preview no longer builds and discards a tiptap editor when it falls back to plain markdown, and the frontend list serializer now clamps an ordered list's `start` the same way the backend does, so the plain `content` field and the outbound markdown cannot disagree on the first marker.

All 1171 conversations backend tests and the frontend editor tests pass.

veryayskiy (approved):

PR #73169, opened 2026-07-23
feat(auth): block SSO login into enforced orgs from unverified domains

a-lider:
## Problem

An org that turns on **enforce SSO** for its verified domain still lets in members whose email is on a *different*, unverified domain. Invite `random@gmail.com` into a PostHog-Cloud org that enforces Google SSO, and that user logs in with their own Google account and becomes a full member. The enforcement never applies to them.

The reason: enforcement is keyed on the **login email's own domain**, not on the org being entered. `social_auth_allowed` calls `get_sso_enforcement_for_email_address(email)`, which looks up the verified-domain row matching the email's domain. `gmail.com` has no such row, so it returns `None` and the login is allowed. The org's enforcement setting lives on the `posthog.com` row and is simply never consulted for a gmail address.

So "enforce SSO" today means "emails *on* the verified domain must use SSO", not "everyone in this org must be on the verified domain". This closes the first half of that gap at login time. Context: internal discussion on restricting enforced orgs to their verified domains.

## Changes

- New manager helpers on `OrganizationDomain` (`posthog/models/organization_domain.py`):
  - `get_active_sso_enforcement_for_organization(org)` — does this org enforce SSO on any verified domain, respecting license and provider config. Keyed on the org, not an email's domain.
  - `find_enforced_org_without_verified_email_domain(email, orgs)` — the block decision: return the first org that enforces SSO but for which the email's domain is not one of its verified domains.
  - Extracted the existing license/provider validation out of `get_sso_enforcement_for_email_address` into a shared `_active_sso_enforcement` so both paths agree on what "actually enforcing" means (an unlicensed or unconfigured setting is inert and never blocks).
- Enforce it in the social pipeline (`posthog/api/signup.py`, `social_create_user`): resolve the orgs the login actually enters — the user's existing memberships plus the invite target — and redirect to `/login?error_code=sso_domain_not_allowed` if any of them block the email.

  > This lives in `social_create_user`, not in the `social_auth_allowed` gate, on purpose. The gate only sees the email and the backend; the organization being entered isn't known until the invite is resolved one step later. That's the whole reason the original check couldn't catch this.

- Login error copy for the new `sso_domain_not_allowed` code.

The JIT path was already safe (it only auto-joins an org when the email's domain is one of its verified domains), so the only hole was invites.

> [!WARNING]
> **Open question for review — scope of the block.** This blocks on *every* SSO login, so an existing member on an unverified domain is locked out the next time they log in, not just at join. That's the intended strictness, but it has a sharp edge: if someone set up the org with a personal email, verified the company domain, then enabled enforcement, **that owner's own email domain isn't verified for the org and they'd be locked out with no way to disable enforcement**. Options if we want to soften it: exempt org owners/admins from the block, or only block at join time (invite acceptance) rather than every login. Easy to change — say which and it's a small edit.

**Not in this PR (follow-up, stacked):** blocking the invite from being *created* in the first place when the org enforces SSO, so these bad invites never exist. Also out of scope: the password-login path for an already-existing outside-domain member.

Tests. `hogli test posthog/models/test/test_organization_domain.py posthog/api/test/test_signup.py`

- `test_organization_domain.py` (new) covers the decision matrix at the manager level: blocks an unverified domain, allows a verified one, and stays inert when the org is unlicensed or doesn't enforce.
- Two pipeline wiring guards in `test_signup.py`: an existing member on an outside domain is redirected to `sso_domain_not_allowed`, and a member on the verified domain still logs in. These guard against both under-blocking (the original bug) and over-blocking legit members.

I (Claude) couldn't run the suite locally — this branch is in a git worktree without the stack, so I'm relying on CI. Lint-staged ran `ruff` and `ty` clean on commit, and `ci:preflight --strict` passed on push.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Alex directed this; I (Claude Opus 4.8) wrote it. The investigation started from the question of whether an enforced org could still be joined by an arbitrary email — it can. I traced the enforcement path (`social_auth_allowed` → `get_sso_enforcement_for_email_address`) and confirmed the check is keyed on the email's own domain, which is why an outside-domain invite slips through.

Main design decision: **where the check belongs.** The natural instinct is to fix `social_auth_allowed`, but it has no organization context — only the email and backend. The org isn't known until the invite is consumed in `social_create_user`, so that's where the block runs, covering both existing memberships and the invite target. The strict "every login" scope (vs. block-only-at-join) is the reviewer-facing call flagged above.

Skills invoked: `/writing-tests` (test gate), `/pr-writer-alex`. This is PR 1 of a two-PR stack.

### DIFF
--- a/frontend/src/scenes/authentication/shared/loginErrorMessages.tsx
+++ b/frontend/src/scenes/authentication/shared/loginErrorMessages.tsx
@@ -32,6 +32,8 @@ export const ERROR_MESSAGES: Record<string, string | JSX.Element> = {
     gitlab_sso_enforced: 'Your organization does not allow this authentication method. Please log in with GitLab.',
     // our catch-all case, so the message is generic
     sso_enforced: "Please log in with your organization's required SSO method.",
+    sso_domain_not_allowed:
+        'Your email address is not on a verified domain for this organization, which requires SSO. Please contact your administrator for access.',
     oauth_cancelled: "Sign in was cancelled. Please try again when you're ready.",
     invalid_invite:
         'This invite link is no longer valid. It may have expired or been revoked. Please ask your administrator for a new invite.',
--- a/posthog/api/signup.py
+++ b/posthog/api/signup.py
@@ -922,6 +922,16 @@ def process_social_domain_jit_provisioning_signup(
     return user
 
 
+def _resolve_invite_organization(invite_id: str) -> Optional[Organization]:
+    """Organization an invite grants access to, or None for legacy team-invite surrogates / missing invites."""
+    try:
+        # nosemgrep: idor-lookup-without-org (invite UUID from server session serves as auth token)
+        invite = OrganizationInvite.objects.select_related("organization").get(id=invite_id)
+    except (OrganizationInvite.DoesNotExist, ValidationError):
+        return None
+    return invite.organization
+
+
 @partial
 def social_create_user(
     strategy: DjangoStrategy,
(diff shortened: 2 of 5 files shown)


a-lider:
Consolidated into #73170. We changed the login-side approach from every-login to join-only (blocking existing members at login would lock out real owners/admins already on non-verified domains), and merged it with the invite restriction into a single PR.

PR #73174, opened 2026-07-23
feat(data-warehouse): surface group properties in custom-property MCP tools

Gilbert09:
## Problem

Warehouse **group** properties are wired end to end in #72876, but the Customer Analytics custom-property MCP tools — and the serializer/view descriptions they generate from — still describe everything as person- or account-only. An agent reading the tool schemas has no way to discover that it can create a group-target property definition, bind a warehouse source to it, or sync/backfill/inspect group-property runs, nor that group operations need the `group:read`/`group:write` scope.

## Changes

Pure description/metadata layer on top of the group-properties backend (no behavior change):

- **Definition tools** (`custom-property-definitions-*`) now document `target_type` (`account`/`person`/`group`) and `group_type_index`, and note that group targets are gated behind the `warehouse-person-properties` feature and require the `group:write` scope to mutate / `group:read` to view.
- **Source tools** (`custom-property-sources-*`) — create/list/retrieve/update/destroy and the sync/backfill/runs actions — now cover person **and** group sources, describe the group-key `key_column`, and call out the group scope requirements.
- **Serializer `help_text` and view docstrings** generalize the person-only wording to person and group; the person-only `ValidationError` messages on sync/backfill now read "person- or group-property sources".
- Regenerated the OpenAPI-derived frontend types (`products/customer_analytics/frontend/generated/*`) and the MCP tool definitions (`services/mcp/*`).

Scopes are intentionally left as `account:*` on the shared tools: MCP scope filtering is `AND` (`hasScopes(...).every`), so adding `group:*` to a shared CRUD tool would hide it from account/person-only callers. The backend already enforces the additional group scope per-target at the view level, so group operations return a clear `PermissionDenied` when the caller lacks group scope — the descriptions now say so.

- `hogli build:openapi` regenerates cleanly (OpenAPI → frontend types → MCP tool definitions), and the group descriptions propagate into `services/mcp/schema/generated-tool-definitions.json`.
- `pnpm --filter=@posthog/mcp lint-tool-names` passes.
- `ruff check` / `ruff format --check` clean on the changed Python.
- Backend: `warehouse_scope_gate_test.py` + `custom_property_source_person_test.py` — 27 passed. `test_views.py` — 218 passed, **1 pre-existing failure unrelated to this PR** (`test_group_definition_value_suggestions_require_group_read_scope`): the `values` action rejects all personal-API-key access (`403 "This action does not support personal API key access"`), so the test's premise that an account-only PAK gets `404` can't hold. This path is untouched here and fails identically on the base branch (#72876) — flagged for a fix there.

No new tests: this change adds no behavior; the existing group-scope gating tests already cover the sources/definitions endpoints.

## 🤖 Agent context

Authored with Claude Code. Invoked the `/implementing-mcp-tools` skill for the regen workflow (`hogli build:openapi`) and scope model. Verified the MCP scope semantics (`hasScopes` = `.every()`) before deciding to keep the shared tools on `account:*` rather than over-requiring `group:*`. Stacked on `tom/dwh-group-props` (#72876).

### DIFF
--- a/products/customer_analytics/backend/presentation/views/serializers.py
+++ b/products/customer_analytics/backend/presentation/views/serializers.py
@@ -308,7 +308,7 @@ class CustomPropertyReferenceSerializer(DataclassSerializer):
 
 
 class CustomPropertySyncTriggerResponseSerializer(serializers.Serializer):
-    """Response of the person-property sync/backfill trigger actions."""
+    """Response of the person/group-property sync/backfill trigger actions."""
 
     status = serializers.ChoiceField(
         choices=[("triggered", "triggered"), ("started", "started"), ("already_running", "already_running")],
@@ -324,8 +324,8 @@ class CustomPropertySyncTriggerResponseSerializer(serializers.Serializer):
 
 
 class CustomPropertySyncRunSerializer(DataclassSerializer):
-    """One person-property sync or backfill run. Read-only: runs are created by the sync/backfill
-    pipeline, never through the API."""
+    """One person- or group-property sync or backfill run. Read-only: runs are created by the
+    sync/backfill pipeline, never through the API."""
 
     id = serializers.UUIDField(read_only=True)
     trigger = serializers.CharField(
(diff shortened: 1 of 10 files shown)


MarconLP (approved):

arthurdedeus (approved):

danielcarletti (approved):
LGTM

PR #73176, opened 2026-07-23
fix(data-modeling): correct view dependency resolution for sibling joins and CTEs

sakce:
## Problem

Three bugs in `products/data_modeling/backend/models/modeling.py`, all in how a saved query's parents get worked out.
They land together because they distort the same result in different directions: one refuses to resolve at all, one drops real parents, one invents parents that don't exist.

> [!NOTE]
> All three only affect DAG sync and scheduled materialization, which run `BoundedResolver`.
> Query execution uses the base `Resolver` and is not implicated, so an affected model runs fine in the SQL editor and fails every scheduled run. "It works interactively" carries no information here.

### 1. Sibling joins reported as circular dependencies

`BoundedResolver` refuses to re-enter a view already on the resolution path.
The guard was scoped to the wrong thing.

A HogQL `FROM` clause is a `next_join` linked list, not a flat list of siblings, and the base resolver walks `node.next_join` from *inside* `visit_join_expr`.
The old guard marked a view as resolving around that whole call, so every later table in the same `FROM` resolved while every earlier view was still marked.
Two shapes tripped it, neither a cycle:

1. a view joined to itself, `FROM v curr JOIN v prev`
2. a shared view sitting in an earlier join position than a view that reads it, `FROM shared JOIN mid` where `mid` reads `shared`. Reverse the join order and the same query passes.

### 2. Parent extraction stopped at the first expanded view

The walk over `next_join` broke out at the first expanded view, so any view joined *after* another view was silently dropped as a parent.
Missing DAG edges, wrong build order, wrong lineage.

### 3. CTEs inside a union-bodied view leaked into parents

Parent extraction collected every CTE definition it met into one flat name-keyed dict.
Descending into a `UNION` pushes the branch queries on their own, and HogQL binds a leading `WITH` to the **first branch only**:

```
with a as (...), b as (...) select … from a union all select … from b

  branch 0: ctes = ['a', 'b']   from = ['a']
  branch 1: ctes = None         from = ['b']
```

The work stack is LIFO, so branch 1 was walked before branch 0's `WITH` had been merged into the dict, and its CTE references looked like table names.
The top-level case was masked by a pre-pass that collected all branches up front; a union-bodied *view* had no such protection.

Two ways that surfaced, both reachable from a single query.
A leaked name with no counterpart fails the parent lookup outright.
Worse, a leaked name that matches a CTE of an unrelated query resolves against **that** definition instead: no error, wrong parent set, wrong DAG edges.

### Cycle guard moves to the view-expansion path

When the base resolver inlines a view it replaces the table with the parsed body and visits that body *before* it reaches `next_join`, so the body visit is what "on the current path" actually means.

```mermaid
flowchart TD
    A[visit_join_expr: FROM shared JOIN mid]:::phYellow --> B[mark shared resolving]:::phRed
    B --> C[expand + visit shared body]:::phBlue
    C --> D[walk next_join: mid]:::phBlue
    D --> E[visit mid body, reads shared]:::phBlue
    E --> F[shared still marked -> ResolutionCycleError]:::phRed
```

```mermaid
flowchart TD
    A[visit_join_expr: FROM shared JOIN mid]:::phYellow --> C[expand shared]:::phBlue
    C --> C2[visit_select_query: mark shared, visit body, unmark]:::phBlue
    C2 --> D[walk next_join: mid]:::phBlue
    D --> E[visit mid body, reads shared]:::phBlue
    E --> F[shared no longer marked -> resolves]:::phYellow
```

- `visit_join_expr` keeps the cycle check and the depth look-ahead but no longer adds to `resolving_views`. It hands the name off in `_pending_view_name`.
- New `visit_select_query` and `visit_select_set_query` overrides mark and unmark around the body visit. A genuine `a -> b -> a` still nests and still raises.
- `view_name` is stamped on `ast.SelectQuery` only, so a view whose body is a top-level `UNION` has no name on the node the resolver walks into. `_pending_view_name` closes that gap: `visit_join_expr` knows the name, the body visit consumes it. Without it, union-bodied views would have lost cycle detection entirely and fallen back to the depth limit with a misleading error.

### Parent extraction walks the whole FROM, with scoped CTEs

- Keep walking `next_join` past subqueries and expanded views instead of breaking at the first one.
- Replace the flat CTE dict with `CteScope`, a tuple of enclosing `WITH` clauses innermost first. Work items are `(query, scope)` pairs, so a name defined in one query's `WITH` can never answer a reference in an unrelated one.
- `_scope_for_union_branches` hoists a union's leading `WITH` into a scope shared by every branch, applied at all three descent points: top level, a `SelectSetQuery` in `FROM`, and a union-bodied CTE.
- The `id()`-keyed `expanded_ctes` guard is untouched, so recursive CTEs still terminate.

Proper scoping also strictly improves the pre-existing shadowing case. The old flat dict handled `WITH cte AS (WITH cte AS …)` only by luck, because the inner `cte` happened to overwrite the outer one.

Automated only. I did not run this against a live warehouse.

Seven new cases, all written before the fix and confirmed red first:

| case | before |
|---|---|
| `select a.* from shared a join shared b on 1=1` | false `ResolutionCycleError` |
| `select m.* from shared s join mid m on 1=1` | false `ResolutionCycleError` |
| `select m.* from mid m join shared s on 1=1` | resolved, but parents were missing `shared` |
| `select * from shared union all select * from mid` | passed, kept as a control so real siblings don't break |
| cycle through a union-bodied view | raised, kept as a control for the `SelectSetQuery` gap |
| `select 1 from unioned` (union-bodied view with CTEs) | CTE name leaked into parents, and a real parent was never reached |
| caller defines a CTE colliding with the view's | leaked name silently resolved against the caller's CTE, real parent silently missing |

No existing test covered sibling joins, a cycle through a union-bodied view, or CTE scoping across union branches.
The pre-existing `a -> b -> a`, top-level self-reference, nested-CTE shadowing and recursive-CTE cases all stay green, so neither cycle detection nor CTE handling is weakened.

Suites run locally:

- `products/data_warehouse/backend/models/test/test_modeling.py` — 47 passed
- `products/data_modeling/backend/test/` + `products/data_warehouse/backend/models/test/` — 472 passed
- repo-wide `mypy --cache-fine-grained .` — clean, 16,725 source files
- `hogli ci:preflight` — 0 failures
- `posthog/hogql/` — 6,409 passed, 52 skipped, 2 xfailed, 0 failed

> [!NOTE]
> An earlier `posthog/hogql/` run showed failures in `test_system_tables.py::TestSystemTablesTeamIsolation`.
> Those are load-sensitive, not caused by this change: the class is a `NonAtomicBaseTest`, the failures came with `deadlock detected` from Postgres, the failing subset and count varied per run, and the file passes 150/150 both with and without this diff on an unloaded machine. The tests also never reach this code — they import data modeling *models* to create rows, never `BoundedResolver` or `get_parents_from_model_query`. I was running several suites concurrently against one database, which is what produced the contention.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change. This is a correctness fix in dependency resolution, not a behavior change anyone configures.

## 🤖 Agent context

I (actually Claude, Opus 4.8) wrote this from a handoff brief I'd put together earlier while root-causing why a set of saved queries could not be moved between DAGs. Bug 3 came in separately once the first two were already implemented, which is why it's a second commit. Skills invoked: `/writing-tests`. Tools: Read, Edit, Bash, git worktree isolation.

Tests came first for all three, per the repo convention, and every case reproduced as predicted before any fix was written.

Two design notes for reviewers. On the cycle guard, the obvious fix is to key off `ast.SelectQuery.view_name`, which the base resolver already stamps. I rejected that on its own because `SelectSetQuery` carries no `view_name`, so union-bodied views would have silently lost cycle detection; `_pending_view_name` exists purely to close that, with a test pinning it. On CTE scoping, I checked against the parser rather than assuming, and confirmed a leading `WITH` really does land on branch 0 only, which is what makes the LIFO ordering matter.

I deliberately left one adjacent defect alone. `current_view_depth` has the same structural confusion as bug 1: the base resolver brackets it around a recursive visit that includes `next_join`, so sibling views inflate the observed depth (`FROM v_a JOIN v_b` reads as depth 2 rather than 1). It only makes the bound more conservative, and fixing it means changing the base `Resolver`, which is out of scope for a fix that shouldn't touch query execution. Worth its own issue, and it only becomes reachable now that wide FROM clauses stop failing earlier for the wrong reason.


sakce:
Thanks ReviewHog — all three findings were real. Addressed in follow-up **#73483** (draft), each confirmed with a red-first test plus an independent trace:

- **Union-bodied views not recognized as opaque parents** (`should_fix`) — fixed. Added `view_name` to `ast.SelectSetQuery`, stamped it in the resolver's view-inline branch, preserved it through both reconstruction sites (`Resolver.visit_select_set_query` and `CloningVisitor.visit_select_set_query` — the second was silently dropping it), and the FROM-walk now registers the union view under its own name. Since the name is persistent now, I also removed the `_pending_view_name` hand-off this comment pointed at.
- **VALUES/PIVOT/UNPIVOT `ValueError`** (`consider`) — fixed. `VALUES` is skipped; `PIVOT`/`UNPIVOT` unwrap to their `.table`. Good catch that skipping PIVOT/UNPIVOT would drop a real parent — a second review pass also found `PivotExpr.table` can be a whole `JoinExpr` (pivot over a join), so the walk splices that inner chain in rather than dropping it.
- **`_scope_for_union_branches` missing WITH on nested non-initial branches** (`should_fix`) — fixed. Replaced the flatten-then-uniformly-scope pattern with a recursive `_select_queries_with_scope` that threads scope through initial and subsequent branches, exactly as suggested.

Verification: hogql 6,411 passed / 0 failed, 474 product-model tests, mypy clean.

PR #73179, opened 2026-07-23
fix(vercel): stop tracking self-recovering connection timeouts as bugs

Gilbert09:
## Problem

Error tracking flagged a `ConnectionError`/`ReadTimeoutError` from the Vercel data-warehouse source ([issue](https://us.posthog.com/project/2/error_tracking/019f8f10-c36f-77b3-a1a9-c70348441dfc)) hitting Vercel's FOCUS billing endpoint. The stack trace confirms this originates in `_open_billing_stream` in `products/warehouse_sources/backend/temporal/data_imports/sources/vercel/vercel.py`.

Both `_fetch_page` and `_open_billing_stream` already retry `requests.ConnectionError`/`requests.ReadTimeout` in-process via `tenacity` (5 attempts, exponential backoff). Once those retries exhaust, the error re-raises and Temporal retries the whole activity — the failure is transient and self-recovering, not a code bug. But `VercelSource` had no `get_retryable_errors()` override, so this benign, self-recovering failure was logged as a full exception and surfaced as error-tracking noise.

This follows the same pattern already used by several other sources (ClickHouse, MySQL, Salesforce, Matomo, Mixpanel) for connection-level failures that are already retried internally.

Note: #73156 separately adds `get_retryable_errors()` to this same source for the `"Vercel API error (retryable)"` (429/5xx) marker, but that doesn't cover this connection/timeout error since the message never contains that marker — so this is a complementary fix, not a duplicate.

## Changes

- Added `VercelSource.get_retryable_errors()` matching the stable `HTTPSConnectionPool(host='api.vercel.com', port=443)` prefix that appears in both the wrapped `ConnectionError` and the underlying `ReadTimeoutError` messages. The host is a fixed constant (not user input), so matching on it is safe and precise.
- Left `NonRetryableErrors` untouched — this is a transient infrastructure error, not a user/credential problem, so the sync should keep retrying rather than stop.

Added `test_retryable_errors_match_transient_network_failures`, parameterized over the two message shapes seen (`ConnectionError` wrapping `ReadTimeoutError`, and a bare `ReadTimeoutError`) — guards that `get_retryable_errors()` actually recognizes the error this issue reported. Ran the full `test_vercel_source.py` suite (20 tests, all passing) plus `mypy` and `ruff` on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

## 🤖 Agent context

Triaged from a webhook-delivered error-tracking alert. I (Claude) investigated the issue via PostHog's error-tracking MCP tools, confirmed the stack trace originates in the Vercel source's `_open_billing_stream`, and traced the retry/error-classification path through `_handle_import_error` and `SourceRegistry`. Checked for duplicate work first via `gh pr list` — found #73156 covers a related but distinct marker string, so this fix is scoped to the connection/timeout case specifically. No skills invoked beyond `/writing-tests` for the new test case.

---
Created with [PostHog Code](https://posthog.com/code?ref=pr)

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/vercel/source.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/vercel/source.py
@@ -105,10 +105,16 @@ To sync resources owned by a team, also enter the team's ID (found under **Team 
         }
 
     def get_retryable_errors(self) -> set[str]:
-        # A 429 or 5xx is retried internally by `_fetch_page`/`_open_billing_stream`; if those
-        # retries still exhaust, the failure is transient and self-recovering, so let Temporal
-        # retry the activity without surfacing it as tracked exception noise.
-        return {"Vercel API error (retryable)"}
+        # `_fetch_page`/`_open_billing_stream` already retry these in-process (a 429/5xx surfaces as
+        # the "Vercel API error (retryable)" sentinel; connection failures and read timeouts surface
+        # as the urllib3 pool error). Once those retries exhaust, Temporal retries the whole activity
+        # and the failure is transient and self-recovering, so don't surface it as tracked exception
+        # noise. The host is a constant, not user input, so matching on it doesn't risk swallowing an
+        # unrelated failure.
+        return {
+            "Vercel API error (retryable)",
+            "HTTPSConnectionPool(host='api.vercel.com', port=443)",
+        }
 
     def get_schemas(
         self,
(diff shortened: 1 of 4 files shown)


Gilbert09:
Heads up: this PR now also touches `import_data_sync.py` with a one-line fix. After merging the latest master, mypy failed on a pre-existing signature mismatch introduced on master by #73135 — the `SchemaColumnTypeChangedException` branch called `handle_non_retryable_error(job_inputs, error_msg, logger, error)` instead of the `(team_id, source_id, run_id, error_msg, logger, error)` signature the two sibling calls use. Since PR CI runs merged with master, this blocked the check here, so I corrected the call to match its siblings. It's in the same data-imports error-handling subsystem this PR already changes.

MarconLP (approved):

danielcarletti (approved):
LGTM

PR #73213, opened 2026-07-23
feat(billing): add "buy a hedgehog a coffee" donation popup

joethreepwood:
## Problem

Long-term free-tier users get a lot of value from PostHog without ever paying. This adds a lightweight, low-frequency way for those users to give something back — by donating to a charity (the Herts Wildlife Trust) — without nagging them.

## Changes

**Frontend**
- New self-gating global popup, "Buy a hedgehog a coffee", mounted in `GlobalModals`, following the existing `MaybeWelcomeDialog` pattern (kea logic + self-gating component).
- Shown only to users on the free plan who are under their free allowance and whose org is at least six months old.
- Capped at roughly twice a year: once dismissed/donated, it stays away for six months (tracked per user+org in `localStorage`, with cross-tab sync).
- The donate button links out to the Herts Wildlife Trust donation page and fires a `hedgehog coffee donation clicked` event.
- Deep link for preview/QA: appending `?modal=buy-a-hedgehog-a-coffee` to any app URL force-opens the popup regardless of billing eligibility (via the shared `bindModalToUrl` contract).

**Backend**
- New server-computed `is_under_free_allowance` field on the `/api/billing` response: true when the org has no active subscription and every product is within its free allocation this period. The popup treats this as the authoritative eligibility signal and only falls back to the client-side check when the flag is absent (unlicensed / hobby).

## How to preview

On any running instance (e.g. the Hogbox preview environment for this PR — see the sticky comment below), append `?modal=buy-a-hedgehog-a-coffee` to any app URL to force the popup open.

I could not run the frontend or backend toolchains in this environment (`node_modules` not installed; the Python venv is empty, so no oxlint/oxfmt/tsc/jest/pytest/mypy). CI runs them. I did syntax-check the Python with `py_compile`.

Automated tests added:
- `ee/billing/test/test_billing_manager.py` — parameterized `SimpleTestCase` for `compute_is_under_free_allowance` covering free-and-under, paying customer, wrong subscription level, over-allocation, and exceeded-limit. Guards against the popup targeting paying or over-limit users if the eligibility rule regresses.
- `ee/api/test/test_billing.py` — updated the two full-response assertions to include the new field (wiring guard that the viewset actually surfaces it), asserting `False` for a paying customer and `True` for a free customer under allowance.
- `frontend/src/lib/components/BuyHedgehogCoffee/buyHedgehogCoffeeLogic.test.ts` — unit tests for the pure eligibility/cadence helpers: backend-flag precedence over the client heuristic, the fallback path, org-age gating, and the six-month "twice a year" cadence. Guards against showing the popup to new orgs or more than twice a year.

## Known limitation (please review)

`is_under_free_allowance` reflects the **current** billing period only. The literal "under the free allowance for the past six months" is approximated by combining it with org age (≥ 6 months) and the twice-a-year cadence. A precise trailing-window signal would need to come from the billing service, which owns free-allocation history across plan changes — the in-repo Django layer doesn't store six months of allowance-aware usage. Happy to follow up with a dedicated billing-service field if you'd prefer that to the approximation.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread. Explored the codebase for the global-modal mount point (`GlobalModals`), the billing eligibility data (`billingLogic` / `BillingType`), how `/api/billing` is assembled (`BillingManager.get_billing`, which proxies the external billing service and augments fields locally), and the localStorage cadence used by the welcome dialog.

Decisions: reused the `MaybeX` self-gating convention rather than imperative `LemonDialog.open`; used a single "show again after six months" timestamp for "twice a year"; put the "under free allowance" business rule on the backend as an authoritative flag (computed from the already-assembled response, so no extra external call) rather than deriving it in the browser, and deliberately did **not** add a per-request six-month usage round-trip to the billing service — flagged above as the precise-but-heavier alternative. Extracted the eligibility/cadence into pure functions so they're unit-testable at the cheapest rung. Invoked the `/writing-tests` skill before adding tests. Added the `?modal=` deep link (shared `bindModalToUrl` contract) so reviewers can see the popup on the per-PR Hogbox preview environment, where the seeded org doesn't meet the eligibility gates.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C04JN5NNMPF/p1784734755296289?thread_ts=1784734755.296289&cid=C04JN5NNMPF)*


frontend/src/lib/components/BuyHedgehogCoffee/BuyHedgehogCoffee.tsx
```diff
@@ -0,0 +1,64 @@
...
+                        data-attr="hedgehog-coffee-donate"
+                    >
+                        Donate to Herts Wildlife Trust
+                    </LemonButton>
+                </div>
+            }
+        >
+            <div className="flex flex-col gap-2 max-w-md">
+                <p className="m-0">
+                    You've been happily under the free allowance for a while now — thanks for being part of PostHog.
```
joethreepwood:
```suggestion
                    You've been happily under the free allowance for a while now and we're fine with that, but...
```

frontend/src/lib/components/BuyHedgehogCoffee/BuyHedgehogCoffee.tsx
```diff
@@ -0,0 +1,64 @@
...
+                </p>
+                <p className="m-0">
+                    If PostHog has saved you a bit of money, would you consider passing some of it on? A small donation
+                    to the{' '}
+                    <Link to={HERTS_WILDLIFE_TRUST_DONATE_URL} target="_blank" disableClientSideRouting>
+                        Herts Wildlife Trust
+                    </Link>{' '}
+                    helps look after hedgehogs and the wild places they live. No pressure — this only pops up twice a
+                    year.
```
joethreepwood:
```suggestion
                    If you've enjoyed using PostHog, we'd like you to consider passing it on by donating to our favourite cause: Hedgehogs. A small donation
                    to the{' '}
                    <Link to={HERTS_WILDLIFE_TRUST_DONATE_URL} target="_blank" disableClientSideRouting>
                        Herts Wildlife Trust
                    </Link>{' '}
                    helps look after hedgehogs and the wild places they live. No pressure — this only pops up twice a
                    year.
```

joethreepwood:
Quick PR to see if we do want to do this. Copy is AI now, but if we're like "Yes, we should do this and this implementation is right" then I'll make the copy better and throw in some art to make the whole thing sing.

charlescook-ph:
conceptually I think we should do this but I am very far from being able to ok your implementation 😂

joethreepwood:
Having some hogland issues that are stopping me previewing this, but I _think_ this should be ready for a proper review.

joethreepwood:
Screenshot 2026-07-24 at 16 23 43

joethreepwood:
Paul's on holibobs, so asking Ben to take a look given Charles' comment.

frontend/public/hedgehog/coffee-hog.png
rafaeelaudibert, line 1: Nope. We're only using hogs from `@posthog/brand` now. You can see the available hogs in brand.posthog.com/hoggies. If you don't have anything good in there, then ask the graphics team to add it to the list. They make sure these hogs follow the existing guidelines.

rafaeelaudibert (changes requested):
Rather than showing as a global modal, you should probably just display it in the ad slot in the sidebar? Then it can last for a little longer and people won't just close it immediately. Your agent should be quite good at just doing that.

Also, left a review, we should not add new hedgehog images, @PostHog/team-graphics and I are slowly working on losing all those images.

PR #73221, opened 2026-07-23
fix(data-warehouse): pass unpacked args to handle_non_retryable_error

fercgomes:
## Problem

`master` is currently red on the Python code quality workflow: the `SchemaColumnTypeChangedException` branch added in #73135 calls `handle_non_retryable_error(job_inputs, error_msg, logger, error)`, but the function signature is `(team_id, source_id, run_id, error_msg, logger, error)`. mypy fails with four errors on `import_data_sync.py:354`, which blocks CI for every open PR.

## Changes

Unpack the arguments the same way the sibling call sites in the same function do.

`uv run mypy --cache-fine-grained` no longer reports the four errors on the file; ruff check and format are clean. No behavior change beyond making the call actually match the signature (the previous form would have raised `TypeError` at runtime when the branch fired).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

One-line mechanical fix authored with Claude Code while babysitting an unrelated PR that inherited the red master. No design decisions involved; the argument shape mirrors the two adjacent call sites.

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/workflow_activities/import_data_sync.py
@@ -351,7 +351,9 @@ async def _handle_import_error(
     # on every retry regardless of source — classify it non-retryable by type here rather than
     # relying on each source listing the message in get_non_retryable_errors.
     if isinstance(error, SchemaColumnTypeChangedException):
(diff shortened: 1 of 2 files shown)


Gilbert09 (approved):

Gilbert09:
This is now fixed from https://github.com/PostHog/posthog/pull/73220

PR #73227, opened 2026-07-23
chore(surveys): add test coverage for widget position reset

Rohitjangra7370:
## Problem

`SurveyWidgetCustomization.tsx` and `constants.tsx` had zero test coverage for the
widget-position logic, despite that exact area causing two real bugs (a survey
ending up with a position value posthog-js can't resolve for its widget type).
Specifically, `NextToTrigger` is only valid for the `Selector` widget type — switching
to `Tab`/`Button` while it's set needs to reset the position, and that reset logic
lived inline in a JSX `onChange` handler with nothing exercising it.

## Changes

- Extracted the reset logic out of `SurveyWidgetCustomization`'s inline `onChange`
  handler into a plain exported function, `resolveWidgetPosition`, in `constants.tsx`.
  No behavior change — same condition, just testable in isolation instead of requiring
  a component render.
- Added `constants.test.ts` with parameterized cases covering: reset from
  `NextToTrigger` on `Tab`/`Button`, no reset when already on `Selector`, and
  non-`NextToTrigger` positions passing through unchanged.

- `pnpm --filter=@posthog/frontend exec jest frontend/src/scenes/surveys/constants.test.ts` — 5/5 passing.
- `pnpm exec oxlint` and `oxfmt --check` on the changed files — clean, no issues.
- Ran the whole-app `typescript:check`; it reports pre-existing errors unrelated to
  this change (missing `@posthog/quill-primitives` type declarations in unrelated
  `products/*` files, likely from a partial workspace install in my sandbox rather
  than a real regression) — confirmed none reference `SurveyWidgetCustomization.tsx`
  or `constants.tsx`/`constants.test.ts`.
- No behavior changed (pure extraction + new test file), so the risk surface is
  limited to the extracted function, which the tests cover directly.

## Docs update

Not applicable — no user-facing behavior, API, or config changed.

### DIFF
--- a/frontend/src/scenes/surveys/SurveyWidgetCustomization.tsx
+++ b/frontend/src/scenes/surveys/SurveyWidgetCustomization.tsx
@@ -5,15 +5,9 @@ import { LemonCheckbox, LemonInput, LemonSelect } from '@posthog/lemon-ui'
 import { LemonField } from 'lib/lemon-ui/LemonField'
 import { surveyLogic } from 'scenes/surveys/surveyLogic'
 
-import {
-    AvailableFeature,
-    SurveyAppearance,
-    SurveyPosition,
-    SurveySchedule,
-    SurveyTabPosition,
-    SurveyWidgetType,
-} from '~/types'
+import { AvailableFeature, SurveyAppearance, SurveySchedule, SurveyTabPosition, SurveyWidgetType } from '~/types'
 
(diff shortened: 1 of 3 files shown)


lucasheriques (approved):
clean extract-method. same expression, one call site before and after, and the test covers all three `SurveyWidgetType` members so it is exhaustive over the conditional rather than just echoing the constant.

PR #73231, opened 2026-07-23
feat(cdp): park-and-retry poison pills in postgres (no clickhouse-lag duplicate)

dmarchuk:
## Problem

The poison-pill autodrain (PR #71538) can execute a job **twice**. It decides what to retry by reading ClickHouse (eventually-consistent, laggy) but acts in Postgres, after the janitor deleted the durable row. When a retry completes but ClickHouse hasn't caught up yet, the next tick re-reads the stale "still failed" note and re-runs the already-succeeded job — re-firing its side effects (emails/webhooks). Worst during incidents, when ClickHouse lag and poison-pill spikes coincide. A single manual trigger is safe, but with an automatic loop (and a coexisting user-facing rerun) the stale-read window is reachable, so the autodrain can't be turned on safely as built.

This is an alternative implementation of #71538's goal that removes the duplicate-execution risk **by construction** rather than patching around it. Stacked on `claude/cyclotron-poison-pill-autodrain`.

## Changes

Replace the "record to ClickHouse → delete the row → re-discover from ClickHouse → rebuild + re-enqueue" loop with **park-and-retry entirely in Postgres**:

- **Janitor parks instead of deletes.** On give-up it keeps the real `cyclotron_jobs` row, sets `scheduled = 'infinity'` (no worker dequeues it — workers filter `status='available' AND queue_name=$1 AND scheduled <= NOW()`), clears the lock, resets the stall budget, and stamps `poison_retry_count`. It still best-effort records the `failed` `janitor_poison_pill` row to ClickHouse for Invocations-UI visibility, but that record no longer gates recovery (the durable row is the recovery).
- **Autodrain releases from Postgres.** One atomic UPDATE per tick moves parked rows' `scheduled` to `NOW()` (incrementing `poison_retry_count`), for `poison_retry_count IS NOT NULL AND < max_attempts AND status='available' AND scheduled='infinity'`, `FOR UPDATE SKIP LOCKED` up to a batch cap. A worker then re-runs the **real** job.
- **New column** `poison_retry_count SMALLINT` (nullable) on `cyclotron_jobs` via a cyclotron-node migration, plus a partial index for the release scan.
- The autodrain no longer depends on ClickHouse or `RerunJobManager` (net −487 lines).

Why it's safe by construction: the retry decision reads and writes only strongly-consistent Postgres, so there's no ClickHouse-lag window. The release is one-shot — once released, `scheduled` is no longer `'infinity'`, so no later tick or concurrent pod can re-release it — and the persistent parked row also blocks a coexisting user-rerun (its non-terminal row hits the re-enqueue conflict guard). Bonus: retrying the real row keeps its true kind/queue, so it also fixes the reconstruct-as-`hog_flow` mislabel that could make a poisoned `hog_function` email job un-replayable.

Automated (all run by the agent locally against `test_cyclotron_node`):

- `poison-pill-autodrain-e2e.test.ts` (rewritten, Postgres-backed, no ClickHouse/Kafka): (1) the janitor **parks** a poison pill (row kept, `scheduled='infinity'`, count 0) and the autodrain **releases it exactly once** — a second tick releases nothing (the no-duplicate guarantee, which fails if the `scheduled='infinity'` guard is dropped); (2) releases stop at `max_attempts` (dead-letter), which fails if the count guard is dropped.
- `cyclotron-v2.test.ts`: updated the 6 janitor poison tests from delete- to park-semantics; full file 119/119 pass. Deleted the old mock-based `poison-pill-autodrain.test.ts` (its ClickHouse/RerunManager mocking no longer applies).
- Touched files typecheck clean (the only `tsc` errors are pre-existing unbuilt-workspace-package modules, unrelated).

Not done by the agent: the migration was **hand-applied** to the local test/dev DBs to run the suite; the committed `.sql` runs via the normal `pnpm setup:test:rust` / `migrate-cyclotron-node` path in CI. Have not exercised a full worker re-running a released job end to end (no worker in the test; the release — `scheduled=NOW()`, `available`, original queue — is the asserted observable).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `nodejs/src/cdp/services/cyclotron-v2/README.md` (poison-pill recovery + autodrain sections) to describe park-and-retry.

## 🤖 Agent context

Authored by Claude (Claude Code) at the driver's direction. Skills invoked: `/writing-tests` (gated the test set to the three concrete regressions above and pushed the flow to the cheapest rung — Postgres integration, no ClickHouse/Kafka).

Decisions: the driver explicitly asked for a fix that removes the duplicate risk rather than a dedup ledger bolted onto the ClickHouse-driven design. Options weighed and rejected: a durable dedup **ledger** keyed by invocation+attempts (adds state, keeps the laggy-read design); a **long autodrain interval** / grace window (probabilistic — still exposed when lag exceeds the interval, i.e. during incidents); a **manual-only trigger** (safe for a single trigger, but a coexisting user rerun is a second trigger, so it doesn't reach zero-by-construction). Park-and-retry was chosen because it moves the retry decision onto strongly-consistent Postgres, so no number of triggers or amount of ClickHouse lag can double-execute.

Reviewer flags:
- This **reverses #71538's deliberate "keep no state between passes" design** and adds a column to the cyclotron-node schema (Rust/sqlx migration) — both warrant the cyclotron/paginator owner's sign-off.
- Stacked on #71538; if this approach is preferred it likely **supersedes** #71538's autodrain rather than layering on it.
- The now-unused autodrain config env vars (`..._WINDOW_HOURS`, `..._MAX_COUNT_PER_GROUP`) are left in place to keep the diff focused; remove in a follow-up if this lands.


dmarchuk:
Closing this one for now. It might be a good option in the future, but for now it's not needed.

PR #73247, opened 2026-07-23
fix(aio): cascade eval disable to its reports

Radu-Raicea:
## Problem

Disabling (pausing) an AI observability evaluation only stops it from scoring new generations — it does not touch the evaluation's attached report. The report is a separate object with its own `enabled` flag, so it keeps firing on its schedule and delivering digests even though the eval produces no new runs. The result is a report that summarizes zero runs, which then reads as a confident but meaningless summary.

## Changes

- Extend the `Evaluation` `post_save` cascade so that when an eval is disabled (`enabled=False`, not deleted), its non-deleted reports are also disabled. This mirrors the existing delete cascade and covers every save path (API PATCH, `set_status`, etc.), keeping report state in sync with the eval.
- Add a data migration (`0029`) to backfill: reports whose evaluation is already disabled and not deleted are disabled too. Deleted evals are already handled by the earlier `0007` cascade migration.
- Add a model test asserting that disabling an eval disables its report without soft-deleting it.

Note: disabling only pauses the report (`enabled=False`); unlike the delete cascade it does not soft-delete it, so a user can re-enable the report if they re-enable the eval.

Added `test_disabling_evaluation_disables_its_report` to `test_evaluation_reports.py`, covering the new cascade — it fails if the disable branch is removed from the signal (no existing test exercised the enabled path; they only covered the delete cascade).

The agent could not run the Django test suite or the migration in this environment (no local Postgres/ClickHouse and Python 3.13 unavailable). Changed files byte-compile and pass `ruff check`. Please run `hogli test products/ai_observability/backend/models/test/test_evaluation_reports.py` and the migration in CI/locally to confirm.

## 🤖 Agent context

Directed via a Slack thread that noticed a disabled sentiment evaluation still emitting scheduled reports. Investigation traced it to the report being a separate object whose `enabled` flag isn't tied to the eval's.

- Invoked skills: `/django-migrations` (backfill data migration), `/writing-tests` (test value gate).
- Chosen approach: cascade inside the existing `evaluation_saved` `post_save` receiver rather than in the serializer/viewset, so all save paths are covered with one change. Backfill mirrors the existing `0007_retire_reports_for_deleted_evaluations` migration.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BCHQJBJR5/p1784765247333589?thread_ts=1784765247.333589&cid=C0BCHQJBJR5)*


products/ai_observability/backend/models/evaluations.py
```diff
@@ -267,6 +267,10 @@ def evaluation_saved(sender, instance, created, **kwargs):
 
     if instance.deleted:
         EvaluationReport.objects.filter(evaluation_id=instance.id, deleted=False).update(deleted=True, enabled=False)
+    elif not instance.enabled:
```
carlos-marchal-ph: This keys on `enabled`, which `_coerce_status_and_enabled` also clears for ERROR, not just for a user pause. So an eval that errors (`provider_key_deleted`, or the transient rate-limited/quota reasons) also disables its report, and since there's no reverse cascade the report stays off when the eval later recovers to active.

carlos-marchal-ph (approved):

PR #73248, opened 2026-07-23
fix(conversations): isolate ticket org attribution fallback failures

abigailbramble:
## Problem

Support tickets were getting the `unknown_slack_default_enterprise` tag (default Enterprise SLA) from the SLA workflow even though the customer's organization was fully known in PostHog — org group present, requester identified, profile carrying the org id, and org-stamped events in the analytics window.

Root cause: `$conversation_ticket_created` resolves the ticket's org `$groups` through a chain of lookups spanning four subsystems (personhog, Postgres membership, ClickHouse analytics events, person profile), but a single catch-all wrapped the whole chain. A transient failure in any one step aborted every remaining fallback, the event was emitted once without `$groups`, and downstream plan/segment workflow branches all missed — falling through to the default-to-Enterprise branch. The same resolution succeeded seconds later for the ticket's `$conversation_message_received` event, confirming the data was there all along.

Context: raised from [ticket #64626](https://us.posthog.com/project/2/support/tickets/64626) mis-tagged by the [SLA workflow](https://us.posthog.com/project/2/workflows/019f5ae1-273d-0000-0f78-6d4b55c8c042/workflow).

## Changes

- Isolate each attribution lookup: a failing step now logs and degrades to the next fallback instead of aborting the chain (`_org_groups_for_person` consolidates the shared membership → analytics → profile tail).
- A failed personhog or email person lookup now falls through instead of being treated as "no person".
- Harden the `process_person` gate in `capture_message_received` to an allowlist: only a person-resolved org (or a legacy row predating the source column) turns on person-profile processing, so future non-person attribution sources default to not creating person profiles.

An earlier revision also added a "requester-events" fallback that read org `$groups` from events captured under the ticket's own distinct_id when person lookups missed. Review (ReviewHog must-fix) correctly flagged that for email/Slack/Teams that distinct_id is the raw customer email, and events under it are captured with the public project token — so anyone who knows the email could plant a `$groups` value and have it adopted precisely when nothing else corroborates. That fallback is removed: event-derived attribution is only trusted when anchored to a resolved person, and the per-step isolation above already covers the incident class (a transient personhog failure degrades to the person-anchored email path instead of aborting).

Happy paths are unchanged: same lookup order, same sources, same authoritative-identified-person early return, same negative-result caching.

- `pytest products/conversations/backend/api/tests/test_events.py` — full suite passing locally.
- `pytest products/conversations/backend/api/tests/test_signals.py` — 39 passed.
- Repo-wide `mypy --cache-fine-grained .` — no new errors.

Tests and the regression each catches:
- `test_capture_ticket_created_person_lookup_failure_falls_through_to_email_fallback` — reverting the per-step guard makes a personhog blip permanently lose attribution.
- `test_capture_ticket_created_analytics_failure_falls_through_to_person_property` — reverting the analytics guard loses attribution despite the profile carrying the org.
- `test_capture_ticket_created_no_person_means_no_event_keyed_attribution` (miss + fail cases) — re-adding unanchored event-keyed attribution would let a planted event under a bare email distinct_id resolve the org.
- `test_capture_message_received_stored_org_source_controls_person_processing` (new `future_non_person_source` case) — regressing the allowlist back to a blocklist would silently re-enable person-profile writes for non-person sources.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal resolution behavior only; no user-facing docs affected.

## 🤖 Agent context

Investigated via PostHog MCP (workflow definition, event timeline for the affected ticket) with Claude Code. The event timeline showed the created event missing `$groups` while the message event 6 seconds later resolved them — narrowing the cause to a swallowed transient failure rather than missing data. Skills invoked: /writing-tests. Considered a retry-with-delay in `capture_ticket_created` but rejected it (blocking sleep in the webhook path, and per-step isolation already covers every single-subsystem outage since the fallbacks span independent stores). The requester-events fallback from an earlier revision was removed after review flagged its unanchored trust in public-token event data.

---

### DIFF
--- a/products/conversations/backend/api/tests/test_events.py
+++ b/products/conversations/backend/api/tests/test_events.py
@@ -1069,9 +1069,10 @@ class TestConversationEvents(BaseTest):
 
     @parameterized.expand(
         [
-            # Person lookups can come up empty (channel email differs from the profile email,
-            # person row not ingested yet) or fail transiently — either way, org groups the
-            # requester's own sessions stamped onto events must still attribute the ticket.
+            # Events keyed by the bare email distinct_id are public-token-captured, so anyone
+            # who knows the email can plant a $groups value on them. Whether person lookups
+            # miss or fail, and even for a verified sender, those events must never attribute
+            # the ticket without a resolved-person anchor.
             ("person_lookups_miss", False),
             ("person_lookups_fail", True),
         ]
@@ -1081,7 +1082,7 @@ class TestConversationEvents(BaseTest):
     @patch("products.conversations.backend.events.get_group_types_for_project")
     @patch("products.conversations.backend.person_lookup._get_persons_by_email")
     @patch("products.conversations.backend.events.get_persons_by_distinct_ids")
-    def test_capture_ticket_created_requester_events_resolve_groups_without_person(
+    def test_capture_ticket_created_no_person_means_no_event_keyed_attribution(
         self, _name, lookups_fail, mock_get_persons, mock_get_by_email, mock_group_types, mock_hogql, mock_capture
     ):
         if lookups_fail:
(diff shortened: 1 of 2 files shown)


abigailbramble:
Addressed the review findings in 54627dc: the requester-events fallback now additionally requires the server-attested `identity_verified is True` on top of the channel gate, so an SPF-failed email (`identity_verified=False`) or a caller-supplied distinct_id that was never assessed (`None`) can no longer resolve an org from event $groups. Slack/Teams tickets are created with `identity_verified=True`, so the incident-class fix is unchanged. Added parameterized tests covering both rejected states.

The `warehouse-sources (1/5)` failure is an infra flake in `test_migration_0052.py` ("the connection is closed") in a product this PR doesn't touch — should clear on re-run.

abigailbramble:
Addressed the ReviewHog findings in bc21656:

- **must_fix (unanchored email-as-distinct_id attribution):** agreed — `identity_verified` attests the ticket's sender, not who captured historical events under that distinct_id, and the fallback fired exactly when nothing else corroborated. Removed the requester-events fallback entirely rather than trying to harden it: the person-anchored email path (`_get_persons_by_email` → analytics with the person's real distinct_ids) plus the per-step isolation already covers the incident class this PR fixes, so the unanchored path added risk without unique coverage. Replaced its tests with a regression test pinning the posture: a planted org on events under a bare email distinct_id must never attribute a ticket, whether person lookups miss or fail.
- **should_fix (source label flips on person processing):** with the fallback gone the mislabel can no longer occur, but the fail-open `!= SLACK_CHANNEL_ACCOUNT` gate it exposed is real — flipped it to an allowlist (person-resolved or legacy no-source rows only), with a parameterized case asserting a future non-person source defaults to no person-profile processing.
- **consider (no negative-cache on ClickHouse failure):** not changed here — the doubling this PR introduced disappears with the fallback removal, and the exception path's cache behavior predates this PR; happy to do it as a follow-up if wanted.

veryayskiy (approved):
Really cool

PR #73254, opened 2026-07-23
chore(personhog): extract persons module and router client to common

jose-sequeira:
## Problem

The personhog services are growing: replica reads person rows, the upcoming identity service creates them, and the lifecycle-manager from the identity saga RFC will follow. Each Rust service so far declared its own copy of the person row type, the storage error enum, the deterministic uuidv5 scheme, and the router-calling conventions (the `x-team-id`/`x-person-id` leader-routing headers). Three copies of the same wire contract is how services drift apart.

## Changes

- New `personhog_common::persons` module: the `Person` storage row type, `StorageError`/`StorageResult` (with the `sqlx::Error` conversion), the `PERSON_UUIDV5_NAMESPACE` + `person_uuid()` helper, and the storage-to-proto `Person` conversion.
- New `personhog_common::client` module: `RouterClient`, a thin wrapper over the generated `PersonHogServiceClient` that owns the leader-routing contract internally. Leader-bound calls (`update_person_properties`, strong `get_person`) get the routing headers stamped from the request's own ids, so callers can't forget them. Eventual reads carry no headers, as before.
- `personhog-replica` now re-exports the shared types from its existing module paths, so its call sites are unchanged. Its local `From<sqlx::Error>` impl, uuidv5 constant, and proto conversion are deleted in favor of the shared ones (the orphan rule enforces the single copy once the type lives in common).

No behavior change and no wire change: the headers and types are today's contract, relocated.

Automated only, run by the agent in this branch's worktree:

- `cargo test -p personhog-common` (54 tests) and `cargo test -p personhog-replica` (333 tests across suites, against the local `posthog_persons` database)
- `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo shear`, `cargo fmt --check` on both crates

The re-export approach means replica's storage/service code compiles unchanged, which is itself the regression check for the type move.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code while building the personhog-identity service (follow-up PR, stacked on this). The identity work surfaced the duplication: identity needed the same `Person` type, error taxonomy, uuid scheme, and router headers that replica and the test harness had each declared locally. José directed the extraction: pool construction was already shared via `common-database`, so only the domain primitives and the client contract moved. A broader "move header constants only" variant was considered and rejected in favor of a full `RouterClient` so the routing contract is owned by one type rather than re-implemented per caller. Skill used: `/adding-personhog-rpc` (for the surrounding identity work).


rust/personhog-common/src/persons.rs
```diff
@@ -0,0 +1,112 @@
...
+            version: person.version.unwrap_or(0),
+            is_identified: person.is_identified,
+            is_user_id: person.is_user_id,
+            last_seen_at: person.last_seen_at.map(|t| t.timestamp_millis()),
+        }
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum StorageError {
```
z0br0wn: nit: i think storage error classification could use it's own file instead of being grouped in with person definitions

z0br0wn (approved):
one nit, not a blocker

nickbest-ph (approved):

PR #73256, opened 2026-07-23
fix(posthog_ai): pin run thread to bottom while a turn is active

skoob13:
## Problem

While an agent turn streams into the run thread (the `/tasks` runner, the PostHog AI side panel behind `phai-sandbox-mode`, and `RunSurface` embeds), the newest output regularly ends up below the fold. The virtualized thread delegated stick-to-bottom to the TanStack core, which follows only while the scroll offset sits within 32px of the end at the moment an append or resize lands. During streaming the offset transiently lags the growing content past that threshold, and the first time it does, following dies for the rest of the turn. The send reserve made it worse: its fixed `paddingEnd` moves the true end below the viewport, so the core either stops following after a send or pins to the padded end and scrolls content up over a blank band.

## Changes

The thread now keeps an explicit pinned state, gated on a new `turnActive` prop that `ThreadView` derives from `runStreamLogic.streamPhase`:

- While a turn is active, the thread is pinned to the bottom. A per-commit follow effect keeps the newest output in view, and a thread opened mid-turn lands at the bottom instead of anchored on the last human message. Settled threads keep the existing anchored open.
- Scrolling up unpins. Unpinning keys off scroll intent (upward wheel, or any offset decrease from a scrollbar drag, keys, or touch) rather than position, since position can't distinguish "scrolled away" from "content briefly outran the follow".
- A scroll that lands back at the bottom re-pins, but only while the turn is active. The core's `followOnAppend` is removed, so an idle thread never moves on its own.
- The send reserve shrinks 1:1 as the response streams into it. The view stays put while the reserved space fills, and once the response outgrows the viewport the reserve is gone and bottom-following takes over seamlessly. While the reserve is up, `anchorTo` flips to `start` so the core's end-anchoring doesn't fight the static fill.

Two implementation notes for reviewers. The follow effect writes `el.scrollTop` directly instead of calling `scrollToEnd()`, because the latter arms the core's multi-frame scroll reconciler, which re-targets the growing end every frame for up to 5s with no public cancel and swallows the user's attempt to scroll away. And the core's `anchorTo: 'end'` plus at-end resize compensation stay on, since they smooth token growth that lands between React commits.

Turn active, the thread followed the stream to the bottom on its own (thinking indicator live, newest chunk in view):

turn-active

Turn finished, the view settled at the bottom and stays inert from here (no thinking line, idle composer):

turn-finished

Claude verified this end to end against the live dev stack, driving the real `/tasks/:id` scene headless with Playwright while a simulated agent turn streamed ACP frames into the run's Redis stream (`TaskRun` in `in_progress`, ~1.4 frames/s). Scripted assertions on the scroll element across three phases, sampled every 500ms for 6s each:

1. Pinned while streaming: distance from bottom stayed 0-42px while content grew ~290px.
2. After a wheel-up gesture: the view held position exactly while content grew ~290px below the fold.
3. After scrolling back to the bottom: following re-engaged and tracked growth again.

The screenshots above come from the same setup with zero programmatic scrolling, so the pinned position in them was reached by the feature itself. Earlier iterations of this verification caught two real bugs (the open-mid-turn race and the threshold-follow death spiral) that shaped the final design.

Existing Jest suites for the surface pass (`ThreadView.connection`, `RunSurfaceImpl`, `ReadonlyRunSurfaceImpl`, `GitArtifacts`, 25 tests), plus `tsgo --noEmit` and oxlint/oxfmt. No new Jest tests: jsdom has no real layout or scroll, so pinning isn't testable there; the `Streaming` story exercises the behavior interactively and now passes `turnActive`. What Claude was not able to verify: a real sandbox run end to end (no sandbox infra locally), and idle-thread appends (a terminal run status tears down the SSE stream, so frames can't arrive on an idle thread through that path).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, no documented workflow covers thread scroll behavior.

## 🤖 Agent context

I directed Claude Code (Fable 5) to build and verify this; session: https://claude.ai/code/session_019bjVEWAUZLRciPe1gVNXdR. Repo skills invoked: /run-posthog for the stack-and-verify workflow. Claude explored the patched `@tanstack/virtual-core` first and initially tried expressing the behavior through the core's own options (`followOnAppend` gated on turn activity plus a dynamically shrinking reserve). Live verification rejected that twice: the 32px at-end threshold let following die permanently once streaming outran the scroll, and `scrollToEnd()`'s reconciler overrode user wheel gestures entirely. The final design moved pin/unpin into explicit gesture-driven state with direct `scrollTop` writes, keeping the core for virtualization, end-anchoring, and between-commit growth compensation.

https://claude.ai/code/session_019bjVEWAUZLRciPe1gVNXdR

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -4822,6 +4822,18 @@ snapshots:
         hash: v1.k794b7964.410f559d98dc233ee8047dd8cfd5ce5f16bf47298ac0f3ea7c38e21913f3e79e._jbCcWpMORUlIHh_Te_m_hWib0TJ1zlg4bw-cm35FTI
     products-posthog-ai-runalertactivity--connection-lost--light:
         hash: v1.k794b7964.6a54547f7784b2a6ddcdde131600ed37ee2abdb56180e9cff603e8a7a3e395bf.cUg8KJHc-L7D6KFrCng16YTN5A0-WRMfw75cElwkroI
+    products-posthog-ai-virtualizedthread--anchored-open--dark:
+        hash: v1.k794b7964.eecc3dde9e8475fb8f20f259880edae96e2750529897fc326c44d33ce20e0885.hQlXYcfwZ9CiFajcBHUg9oY6h2QcGrm91CkXVIht-OE
+    products-posthog-ai-virtualizedthread--anchored-open--light:
+        hash: v1.k794b7964.1b0387275217fb820cbbe4a18caa5cde4be9521d4e296933e80e8f78a9914490.3TrbyabUGN-G6zrsTie8aaMDWrYs1wWMW7OIANOjsbE
+    products-posthog-ai-virtualizedthread--anchored-open-mid-turn--dark:
+        hash: v1.k794b7964.eecc3dde9e8475fb8f20f259880edae96e2750529897fc326c44d33ce20e0885.jUSUfoNAuVJvfWe8M9IRRg_CirdzkCKjQjoVsutO9vY
+    products-posthog-ai-virtualizedthread--anchored-open-mid-turn--light:
+        hash: v1.k794b7964.1b0387275217fb820cbbe4a18caa5cde4be9521d4e296933e80e8f78a9914490.qUuhFOn4BwbL8REUimYFdxCpOmg64Cj1HJD2Ovwdan0
+    products-posthog-ai-virtualizedthread--anchored-open-mid-turn-long-tail--dark:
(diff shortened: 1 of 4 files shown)


Twixes (approved):
Approving, as this doesn't upset me.

But I'll be very honest: I don't quite get the substance of the change, as all of: the LLM's framing of the issue, the screenshots, and the overcommented code - are not helpful to me in understanding this change. There is just too much noise and not enough signal from the agent without human editing.

PR #73257, opened 2026-07-23
feat(personhog): add identity service with get-or-create person RPCs

jose-sequeira:
## Problem

Per the personhog identity saga RFC, person state spans two planes: a sync plane (Postgres primary) that owns existence and identity topology, and an async plane (leader + changelog) that owns property content. Creating a person touches both, so ingestion needs one entry point whose ack covers both planes. That entry point is the personhog-identity service, and this PR bootstraps it with its first RPCs: get-or-create by distinct id, single and batch.

Stacked on #73254 (shared persons module and router client in personhog-common).

## Changes

```mermaid
flowchart LR
    I[ingestion] --> ID{{personhog-identity}}
    ID --> PG[(PG primary)]
    ID --> R{{router}}
    R --> L{{owning leader}}
    L --> CL[(changelog)]
    class ID,R,L phBlue;
    class I phYellow;
    class PG,CL phGray;
```

- New `personhog.identity.v1.PersonHogIdentity` gRPC service with `GetOrCreatePersonByDistinctId` and `GetOrCreatePersonsByDistinctIds`. Ingestion calls it directly; it is not fronted by the router.
- New `rust/personhog-identity` crate, mirroring the replica's bootstrap (envconfig, shared pool construction, lifecycle manager, health/metrics, tonic layer stack). The service layer is dispatch-only in `service/mod.rs` with one module per RPC family, request limits are config-driven.
- The pipeline per batch: one UNNEST resolve on the primary, one multi-row stub-create transaction for misses (deterministic uuidv5, version 0, empty properties, per-row `ON CONFLICT` so no key can abort another), one batched lost-race re-resolve, then a bounded-concurrency fan-out of initial `$set`/`$set_once` through the router to each owning leader. `created = true` only after the leader's changelog ack, so the single ack covers both planes.
- Extra distinct ids carry personless history forward exactly like the plugin-server merge path: `version 1` when a `posthog_personlessdistinctid` row already existed (so ClickHouse gets an override), else `version 0`, with the row marked merged either way.
- Race behavior: concurrent creates for one distinct id collide on the deterministic uuid and converge on the winner (exactly one `created = true`); a distinct id concurrently mapped to a different person rolls back only that stub, leaving no orphan. A crash between the stub commit and the leader ack surfaces as a per-key error, and the retried key resolves to the stub with `created = false` so the caller applies properties through the normal update path.
- The e2e harness gains `--create-via-identity`: the spawned stack also runs the identity service, persons are created through the batch RPC with a seed property, and create acks are journaled so the gate holds create visibility to the same acked-implies-visible invariant as updates. Runs in the `personhog-gate` CI job under leader kill + scale-up chaos.
- Dev and deploy wiring: workspace member, `bin/start-rust-service` + `bin/mprocs.yaml` entries (gRPC :50055, metrics :9108), sqlx offline cache, CI test weight, determinator rule, and the image build/deploy matrix entries (Depot project `jfz76fgvqd`).

> [!NOTE]
> Known follow-ups before ingestion switches over: Node/Python client stubs and the ingestion call site, ClickHouse topic messages for created persons/distinct id rows (the changelog covers the person row, nothing emits the `person_distinct_id` messages yet), and parity decisions on `$creator_event_uuid` and `properties_last_updated_at`/`properties_last_operation`, which the stub path does not populate.

Automated, run by the agent (no manual testing claimed):

- `cargo test -p personhog-identity`: 16 tests. Storage tests run against the real local `posthog_persons` database and include three genuine race tests orchestrated with held (uncommitted) transactions: concurrent same-key creates converge with exactly one `created = true`, a create blocked on an uncommitted winner resolves to it after commit (this one proves the separate winner-fetch statement's snapshot reasoning), and a create blocked on an uncommitted stolen mapping rolls back as a lost race with no orphan. Service tests cover the leader fan-out (mock property writer), per-key batch outcomes, duplicate keys creating exactly once, and the stub-survives-leader-failure retry semantics.
- e2e: `personhog-test-harness gate --create-via-identity` rehearsed locally, three consecutive green runs plain and three under `--kill-after 5s --scale-up-after 8s` chaos (~6,600 acked writes per chaos run, zero consistency violations).
- `cargo build`, `clippy --all-targets -- -D warnings`, `cargo shear`, `cargo fmt --check` across the touched crates; `bin/hogli lint:workflows` for the CI edits.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with Claude Code from the personhog identity saga RFC, with José directing design reviews throughout. Skills used: `/adding-personhog-rpc`, `/extending-personhog-test-harness`, `/gating-production-deploys`. Decisions shaped in review: the batch path was reworked from per-entry transactions to one multi-row transaction (the batched Postgres statements are the point of a batch RPC), the router-calling convention was consolidated into the shared `RouterClient` from #73254 instead of per-caller header stamping, request limits moved from constants to config, and the service module was structured one-module-per-RPC-family rather than the replica's single-file style. Empirical note for reviewers: `ON CONFLICT (team_id, uuid)` works on the partitioned `posthog_person` parent (verified against a live database), so the stub insert does not need SplitPerson's select-then-insert workaround, whose comment predates the parent-level unique index.


gantoine (commented):
CI changes look good

rust/personhog-identity/src/config.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub max_batch_size: usize,
+
+    /// Maximum accepted distinct_id length in bytes. Must not exceed the
+    /// posthog_persondistinctid.distinct_id column limit (varchar(400)).
+    #[envconfig(default = "400")]
+    pub max_distinct_id_length: usize,
+
+    /// Router endpoint used to reach the owning leader for initial-properties
+    /// writes on the creation branch (UpdatePersonProperties).
+    #[envconfig(default = "http://127.0.0.1:50052")]
```
nickbest-ph: should be 50054? that's the default we use for the leader router in the local dev stack i believe

bin/start-rust-service
```diff
@@ -285,6 +285,16 @@ case "$RS_SVC" in
     export METRICS_PORT=9101
     ;;
 
+  personhog-identity)
+    SQLX_QUERY_LEVEL=${SQLX_QUERY_LEVEL:-trace}
+    export RUST_LOG=debug,sqlx::query=$SQLX_QUERY_LEVEL
+    export RUST_BACKTRACE=1
+    export GRPC_ADDRESS=127.0.0.1:50055
+    export PRIMARY_DATABASE_URL="${PRIMARY_DATABASE_URL:-postgres://posthog:posthog@${PGHOST:-db}:${PGPORT:-5432}/posthog_persons}"
+    export ROUTER_URL=http://127.0.0.1:50052
```
nickbest-ph: this should be 50054 as to hit the leader's router, based on the current config

rust/personhog-identity/src/storage/postgres.rs
```diff
@@ -0,0 +1,322 @@
...
+                &pdi_versions
+            )
+            .fetch_all(&mut *tx)
+            .await?;
+            for row in rows {
+                mapped.insert((row.team_id, row.distinct_id), row.person_id);
+            }
+        }
+
+        // 5. Per-stub outcomes. A stub whose primary mapping went elsewhere is
```
nickbest-ph, line 291: this undoes the person and the distinct id mappings

step 3 above also sets `is_merged = true` for every extra distinct id on its personlessdistinctid record. that also needs to be undone

rust/personhog-identity/src/storage/postgres.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        let mut tx = self.primary_pool.begin().await?;
+
+        // 1. Multi-row stub insert. Concurrent creates for one distinct id
+        // derive the same uuid, so exactly one insert wins per key; losers
+        // fetch the committed winner in a fresh statement snapshot below.
+        let inserted = sqlx::query_as!(
+            Person,
+            r#"
+            INSERT INTO posthog_person
```
nickbest-ph, line 131: multi row insert does not lock all keys at once, not sorting the keys we lock on before inserting can cause dead locks from concurrent writers.

pod A writing persons {a, b} and pod B writing persons {b, a} is a dead lock

could only happen as a conflict between overflow and main, so not that likely, but i think we just have to sort the keys before inserting to make sure it could never happen.

rust/personhog-test-harness/src/cli.rs
```diff
@@ -258,6 +258,22 @@ pub struct GateArgs {
...
+    /// write, so the gate also asserts create visibility: the initial
+    /// properties must survive into strong reads and Postgres. A spawned
+    /// stack brings up its own identity service; with
+    /// --external-router-url, pass --external-identity-url too.
+    #[arg(long, default_value_t = false)]
+    pub create_via_identity: bool,
+
+    /// Identity service URL for --create-via-identity against an
+    /// already-running stack (the dev stack runs it at
+    /// http://127.0.0.1:50055). Ignored for spawned stacks.
```
nickbest-ph, line 288: based on the code at gate.rs:195 this is not ignored for spawned stacks

nickbest-ph (approved):
few comments to address, o/w lgtm

nickbest-ph:
The proto CI failure is a little bit of an indication that we might have to update the proto generation tooling a bit so that we can be more discriminatory around which clients end up getting which protos

PR #70056, opened 2026-07-10
fix(data-warehouse): preserve saved-view column order through jsonb

thiagosalvatore:
## Problem

A saved view's columns are inferred in SELECT order but stored in a Postgres `jsonb` object, which does not preserve key insertion order. On read, `SELECT *` expansion and the API schema iterated the scrambled `jsonb`, so a view's columns came back in the wrong order.

## Changes

- New nullable internal `column_order` field on `DataWarehouseSavedQuery` and, for materialized-view backing tables, `DataWarehouseTable`, recording the SELECT order captured at write time.
- A single `set_columns()` chokepoint on each model sets `columns` and `column_order` together so they cannot drift.
- A shared `reconstruct_ordered_columns` helper applies the recorded order (skipping removed columns, appending newly discovered ones) at `hogql_definition`, so both `SELECT *` and the API schema follow SELECT order.
- `column_order` is internal and not exposed through the API. Legacy rows (`column_order = null`) keep the current jsonb-order fallback. No fleet-wide backfill, since recovering order means re-running each view's query.

Two additive nullable migrations: `data_modeling 0027` and `warehouse_sources 0063`.

> [!NOTE]
> Stacked on #70055 (frontend). `DataWarehouseTable` and `create_table.py` live in warehouse-team trees, so CODEOWNERS will route that half of the review.

## How did you test this code?

Automated pytest, all run by me (actually Claude):

- `reconstruct_ordered_columns` parameterized over legacy-null, exact, removed-skip, appended, and dedup
- an API round-trip through real Postgres with column names chosen so `jsonb` actually reorders them (a Python-dict-only test would pass even with the bug), asserting a fresh GET returns SELECT order
- a materialized-view backing table's `hogql_definition` honors recorded order, and `set_columns` records order
- the full existing saved-query API suite (71) and the HogQL database/view suites pass; migration linearity and drift checks are clean

No manual testing was done by the agent.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Directed by @thiagosalvatore, implemented by Claude (Claude Code). Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`, and `/plan-eng-review` (with a Codex outside-voice pass).

[...]

PR #73258, opened 2026-07-23
fix(hogql): align asterisk-expanded union branches by column name

sakce:
## Problem

Set operations are positional in ClickHouse, but `SELECT *` expansion follows each table's stored field order.

Warehouse table column metadata lives in a Postgres `jsonb` column, which does not preserve insertion order. #70056 added `DataWarehouseTable.column_order` so newly materialized tables record their real SELECT order, but tables materialized before that still fall back to the jsonb key order (length, then bytewise). A union spanning both vintages therefore emits branches whose select lists disagree, and ClickHouse pairs them by position.

Two failure modes, and the quiet one is worse:

- Where the crossed columns have incompatible types, the query throws. `toTimezone` is injected by our printer for any field typed as `DateTime`, so a crossed `String` surfaces as `Illegal type String of argument of function toTimezone` on a column the author never mentioned.
- Where the crossed columns happen to share a type, **nothing throws**. The query succeeds and returns the wrong column's values. Two `Float64` columns swapping places is invisible.

Which columns you project decides which you get, because Parquet is columnar and ClickHouse only decodes what is selected. That is why the same view answers some queries fine and fails others.

## Changes

Align set-query branches by column name during resolution, but only where the order is ours to change.

```sql
-- before: branches expand in each table's own stored order, paired positionally
SELECT uuid, timestamp, tlat, city FROM t_new
UNION ALL
SELECT city, tlat, uuid, timestamp FROM t_old   -- position 2 is DateTime here, Float64 there
```

The gate is deliberately narrow. Alignment applies only when every leaf of the set query is entirely asterisk-expanded **over a table's stored field order**, which is arbitrary and was never expressible by the author. Three things are left alone:

- any branch with an explicit select list keeps SQL-standard positional semantics, so a union deliberately combining differently-named columns is untouched
- `SELECT *` over a hand-written subquery or CTE keeps its author's order, since that order *was* chosen
- a subquery that is itself asterisk-expanded still aligns, because the order is transitively the table's

Two supporting fixes fell out of review:

- positional `ORDER BY` / `GROUP BY` / `LIMIT BY` ordinals index into the select list, so they are remapped alongside it. Without this the fix silently reorders rows, which is the same bug class it exists to remove.
- expression fields resolve to an alias over an expression rather than over a field, so the alias now records the asterisk expansion. Without it, alignment silently declined in the ClickHouse dialect for any table carrying a computed column, meaning the HogQL preview aligned and the executed SQL did not.

> [!WARNING]
> One shape changes from returning data to raising. An all-asterisk union whose branches have the same column *count* but different column *names* now raises a `QueryError` instead of silently pairing unrelated columns. Refusing is the right call: with `SELECT *` the author expressed "whatever columns these tables have", so when those sets disagree there is no defensible interpretation. Aligning on the intersection would silently drop columns, and warning is useless on the materialization path where nobody reads warnings.

`UNION ... BY NAME` is out of scope. HogQL parses and prints it, but ClickHouse has no such syntax, so that path errors today and still does. It deserves its own change.

New regression tests in `posthog/hogql/test/test_union_asterisk_order.py` and one in `posthog/hogql/database/test/test_s3_table.py`, each aimed at a distinct regression: branch alignment, explicit select lists staying positional, nested set queries following the outer order (asserted on resolved column *types*, since the printed SQL is byte-identical either way), positional ordinals surviving the permutation, alignment holding in the ClickHouse dialect with an expression field, only stored order being realigned, the differing-column-set error, and projection pushdown still pruning aligned branches symmetrically.

The ClickHouse-dialect and positional-ordinal tests were **mutation-checked**: reverting each fix makes its test fail, so neither is decorative.

Suites I actually ran locally: `test_union_asterisk_order.py` and `test_s3_table.py` (32 passed), the `asterisk_expander` snapshot tests (10 passed), and `test_resolver.py` + `test_printer.py` + `transforms/` + the above together (1570 passed) before the final snapshot update. `ruff check` and `ruff format` clean, `ci:preflight --strict` clean.

I could not get a reliable clean full-tree run locally. The same suite selection produced 1570 passed and then several hundred setup errors on identical code, and every failing test I sampled passes in isolation, so the local test database is the unstable variable. Nothing I saw traced to this diff, and no test in the areas this touches failed. CI is the real check here.

Snapshot churn in `test_resolver.ambr` is 60 lines, all `from_asterisk: False` to `True`, no structural change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change. The behavior being fixed was never documented, and the new error message is self-describing.

## 🤖 Agent context

I had Claude root-cause this from a report that selecting some columns from a warehouse view failed while others worked. It reproduced the crossed SQL locally, traced the regression to #70056, and confirmed the mechanism against the resolver rather than the data.

The first version of the fix was larger and wrong in ways worth recording. It bundled a second, unrelated fix for `UNION ... BY NAME`, which turned out to contribute two of its own defects (a nested branch silently dropping columns, and a raw `KeyError` surfacing as a 500). Splitting that out removed both and cut the change roughly in half. Three independent adversarial reviews then found four blockers in what was left: the positional-ordinal gap, the expression-field gate hole, and the two `BY NAME` ones. All are fixed here; the review that found them is why the diff shrank while getting more correct.

Two things were deliberately deferred rather than folded in. A union that tags its branches (`SELECT *, 'source_a' AS src FROM a UNION ALL ...`) still crosses columns, because one non-asterisk item disqualifies the whole branch. That is unchanged from before this PR, not a new regression, and the fix (aligning the asterisk-expanded prefix and leaving trailing items positional) deserves its own tests. `COLUMNS(...)` has the same gap.

Skills invoked: `/writing-tests` before adding any test, which is why the suite is five focused cases plus the three added during review rather than a case per shape.

Reviewers: the highest-value place to push back is the gate in `_expands_stored_column_order`. It encodes a judgment - that a table's stored field order is ours to change and a hand-written select list is not - and if that line is in the wrong place, the fix is either too timid or too aggressive.


sakce:
Closing in favour of explicit opt-in.

The approach here was to silently realign asterisk-expanded union branches whose stored column order disagreed. It works, but it changes what a query means without the author asking, and every defect found in review came from that implicitness: positional `ORDER BY` ordinals silently repointing, the gate declining in one dialect but not another, and the judgement call about whose column order we are entitled to rewrite.

We are going the explicit route instead. `UNION ... BY NAME` already parses, types and prints in HogQL, but ClickHouse has no such syntax, so it is currently a guaranteed `SYNTAX_ERROR` - a feature that looks supported all the way down and can never run. Making that path work gives the same result with none of the ambiguity: the author states the intent, and the reorder happens because they asked for it.

Worth keeping from this branch if the by-name work needs it:

- `_remap_positional_references` - reordering a branch's select list silently changes what `ORDER BY 2` refers to. Any by-name lowering hits the same trap.
- the regression tests in `posthog/hogql/test/test_union_asterisk_order.py` and `posthog/hogql/database/test/test_s3_table.py`, in particular the ClickHouse-dialect one - alignment behaves differently in the hogql and clickhouse dialects because expression fields resolve to an alias over an expression, so a hogql-only test cannot see it.

Background for anyone who lands here from the original report: a union across tables materialized either side of #70056 pairs columns positionally against mismatched select lists. Where the crossed types are incompatible it throws; where they match it returns the wrong column's values with no error. Writing an explicit column list in each branch fixes it today and is immune to the whole class.

PR #73264, opened 2026-07-23
feat(marketing-analytics): add a customer acquisition cost column

jabahamondes:
> [!NOTE]
> Top of the chain, on top of ROAS (#73052). This is now the branch that has everything: flags, write endpoints, MCP tools, currency conversion, ROAS, and CAC. Check out `javier/ma-cac-column` to try the full read-side end to end.

## Problem

The companion to ROAS. Once you can see revenue over spend, the next question is what a customer costs to acquire. Spend per channel is there, and now goals can be marked `counts_as_customer`, so the ratio is computable.

## Changes

A `Cost per customer` column: the channel's spend divided by the customers produced by the conversion goals marked `counts_as_customer`.

It's the mirror image of the ROAS column from #73052, and shares the same machinery (`_sum_conversion_values`, the `include_cost_per` guard, the same "hidden until a goal is flagged" behavior). Two differences:

- **Inverse formula.** ROAS is revenue over spend; CAC is spend over customers.
- **Naming for free formatting.** The alias is built from the existing `Cost per` prefix, so it falls through `to_marketing_analytics_data` to the currency / higher-is-worse formatting the other cost-per columns already get. No new metadata.

### The assumption, stated plainly

Each customer goal contributes its **conversion count** as the customer count. That is correct when the goal marks a **once-per-person** conversion, a sign-up or a first purchase, where counting conversions and counting people are the same thing.

It is wrong if someone points a customer goal at a **repeatable** event, say a generic `payment` that fires on every monthly renewal. Then a returning customer's second payment would be counted as a new customer, understating CAC. Getting that right needs each person's first-ever conversion, a historical first-time-per-user scan the marketing precompute can't do today (it only materializes the queried window). That's a real piece of work, and it's deliberately not here.

So this ships the version that's correct for the natural case (a goal marking the moment someone becomes a customer) and is honest in code and comment about where it breaks. The column name says "customer", not "new customer", so it doesn't over-promise acquisition semantics it isn't enforcing.

3 new unit tests in `test_conversion_goals_aggregator.py`, mirroring the ROAS ones, run by me (Claude):

- **Denominator counts only the customer goals.** Four goals, two flagged customer at indices 1 and 3. Asserts the CAC expression references the value columns of 1 and 3 and not 0 or 2, and that spend is the numerator (the inverse of ROAS). This is the core regression.
- **Absent when no goal is a customer goal.**
- **Absent without campaign_costs.**

Full `test_conversion_goals_aggregator.py`, `test_marketing_analytics_table_query_runner.py` and `_business.py` pass (112), snapshots unchanged. `ruff` clean; `mypy` clean on the file (the one printed error is the pre-existing unrelated `personhog_client` one).

No end-to-end ClickHouse test, same reasoning as ROAS: the construction is unit-covered and the propagation is the shared path the cost-per columns already exercise.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Holding with the rest of the read-side until the dashboard redesign decides how these columns surface.

## 🤖 Agent context

I directed this, Claude wrote it. Invoked `/writing-tests`.

Worth recording the design turn, since it's the interesting part. My first read was that CAC needed a historical first-time-per-user scan and was therefore a large, precompute-touching project, so I nearly deferred it. Javier pushed back: a person can't sign up twice, so why the history? That's right, and it reframed the whole thing. The history only matters for repeatable events; for a once-per-person conversion, counting conversions already counts customers. So CAC collapsed to the same shape as ROAS. This PR takes that simpler, correct-for-the-common-case path and documents exactly where the harder version would be needed, rather than building the heavy scan speculatively.


jabahamondes:
Superseded by #83264, stacked on #83263.

PR #73268, opened 2026-07-23
refactor(security): return typed credential objects instead of tuples

arnohillen:
## Problem

Six functions across core, ee, and products returned a **pair of same-typed secrets as a positional tuple**. A caller that unpacks them in the wrong order still typechecks, so a swap ships silently, and for credentials a silent swap is a security bug rather than a readability nit. This is the highest-value slice of a broader tuple-return audit ([the readability pass in #72946](https://github.com/PostHog/posthog/pull/72946) covered the non-security ones).

Examples of what typechecked before but shouldn't:

- `user, password = get_clickhouse_creds(...)`, two `str`, ~10 call sites across the SQL layer.
- `plain_token, hashed_token = generate_scim_token()`. Swap it and you persist the plaintext token as the "hash", which does not fail loudly.
- `access_key, secret = _read_s3_credentials(...)` and `_get_s3_credentials()`, AWS key id vs secret.
- `code_verifier, code_challenge = generate_pkce()`, the PKCE pair.

## Changes

Each function returns a frozen dataclass with named fields, and every producer, consumer, and test moved with it. Frozen dataclasses rather than `NamedTuple` on purpose: they are **not iterable**, so a leftover positional unpack fails loudly instead of silently swapping.

| Function | Before | After |
|---|---|---|
| `get_clickhouse_creds` | `(user, password)` | `ClickHouseCredentials` |
| `sign_slack_request` | `(signature, timestamp)` | `SlackRequestSignature` |
| `_read_s3_credentials` | `(access_key, secret)` | `AWSKeyPair` |
| `generate_scim_token` | `(plain, hashed)` | `ScimToken` |
| `_get_s3_credentials` (batch_exports) | `(key, secret)` | `AWSKeyPair \| None` |
| `_get_s3_credentials` (hogql) | `(key, secret)` | `S3WebCredentials` |
| `generate_pkce` | `(verifier, challenge)` | `PkcePair` |
| `register_dcr_client` | `(client_id, client_secret, auth_method)` | `DcrClientRegistration` |

> [!NOTE]
> The two `_get_s3_credentials` are **independent same-named functions** in different packages, not one calling the other.

Three things go beyond a plain tuple-to-object swap, all from review:

**Fields are read where they are used.** Returning an object and immediately unpacking it into two locals keeps the risk this PR exists to remove, so all 30 such sites (12 production, 18 test) read the field at the point of use instead.

**AWS credentials are one object, not two arguments.** `posthog/credentials.py` adds `AWSKeyPair`, holding an `AWSAccessKeyId` and an `AWSSecretAccessKey` (distinct `NewType` brands), with `unsafe_from_strings` as the only place raw strings become branded. Consumers (`get_s3_function_call`, both `as_insert_into_s3_query_with_parameters`, `get_s3_client`) take a single argument, so a transposition is unrepresentable rather than merely rejected, and half a pair cannot exist. Keyless IAM auth is `None`, which keeps "no credentials" distinct from "half a pair".

**Secrets are absent from `repr`.** Every credential dataclass marks its secret field `field(repr=False)`, so a traceback or log line that captures one of these objects no longer prints the secret. The field still reads normally.

### Second review round

The security review flagged the positional sites the first pass left behind, all closed here:

- `init_clickhouse_users` builds `ClickHouseCredentials` directly instead of returning `tuple[str, str]` per user, removing the last positional indexing on the ClickHouse credential path.
- `register_dcr_client` (mcp_store) returns `DcrClientRegistration` instead of a three-string tuple whose middle element is a confidential OAuth client secret. All three consumers (two install flows, the probe) read fields by name.
- The unused `generate_pkce_pair` in `posthog/api/oauth/toolbar_service.py` is deleted. It had zero callers, and a future caller transposing its pair would have sent the PKCE verifier into the authorization URL.
- `PkcePair.code_verifier` is now `repr=False` like the other secret fields.
- `_get_s3_credentials` (batch_exports) logs a warning when exactly one of the two `OBJECT_STORAGE_*` vars is set, so the keyless fallback is loud for a misconfigured self-hosted operator. This mirrors what `init_clickhouse_users` already does for the equivalent ClickHouse case.

### Behavior changes

Three, all in the batch-exports staging path, and all closing a latent hole:

- `get_s3_client` branches on absence instead of handing boto3 a key with no secret. That combination raised `PartialCredentialsError`, so the only input whose behavior changes is one that previously crashed.
- A partially configured environment (one of the two `OBJECT_STORAGE_*` vars set) counts as keyless. `get_s3_function_call` already assumed that.
- The half-pair case now logs a warning (see above). Log output only, no control-flow change.

Everything else is value-identical. These objects only cross in-process call boundaries, so nothing serialized (tokens on disk, Temporal payloads, Redis) changes.

> [!IMPORTANT]
> Anything landing on master against the old tuple contract breaks on merge here. #72698 did exactly that while this was open, and merging master in broke 15 of its tests until the call site was updated. Worth merging master and re-running before merging this.

- Repo-wide `mypy --cache-fine-grained .` clean (17,017 files) from a **cold cache**. A warm-cache run gave a false pass after merging master, hiding the #72698 breakage above, so the cache was cleared and it was re-run. Mypy is the real guarantee that every call site moved. Re-run repo-wide after the second review round.
- Ran the affected suites: all SCIM, slack_app, mcp_store oauth and probe, batch-exports internal stage and record batch model, integration model, and clickhouse. **~1900 passing.** After the second review round, re-ran mcp_store oauth, probe, and api, batch-exports internal stage, and the toolbar OAuth primitives: 321 passing.
- **Backwards-compatibility checks**, since this touches credential loading used by ClickHouse dictionary DDL:
  - Executed every credential-bearing generated string on master and on this branch and diffed them: 141 lines across 7 DDL generators and all 17 `ClickHouseUser` lookups are **byte-identical**.
  - Compared the public symbol surface of all 19 touched modules against master: **no name removed**, so no importer breaks. Module constants such as `CLICKHOUSE_DICT_READER_USER` keep both name and value.
  - Confirmed `AWSKeyPair` never crosses a Temporal activity or workflow boundary, and that no YAML, JSON, or TS refers to any changed name.
  - The one symbol removed in the second round, `generate_pkce_pair`, has zero importers anywhere in the repo.

Two new test groups, each verified to fail without the code it guards:

- `test_credentials.py` asserts the secret is absent from `repr` but still readable. Catches someone dropping `field(repr=False)` and quietly putting secrets back into logs.
- `test_partially_configured_object_storage_is_keyless` covers the three partial-configuration combinations. Catches removal of the `None` guard, which would send partial credentials to boto3 and raise.

No other new tests: the remaining conversions are value-preserving and already exercised by the suites above (the existing `register_dcr_client` cases now assert on `DcrClientRegistration` equality), and mypy plus the DDL diff cover the structural risk.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal refactor, no user-facing or API changes, nothing to update.

## 🤖 Agent context

Claude Code (Opus 5). Skills invoked: `/improving-drf-endpoints` (serializer-adjacent review), `/writing-tests` (gating the two new test groups), `/pr-quality`.

Review shaped the design twice. The first pass returned typed objects but unpacked them straight back into locals, which a reviewer correctly pointed out preserved the swap risk, so the fields are now read at the point of use everywhere. The reviewer then suggested branding the two halves as distinct types. Branding alone was implemented first and needed an `@overload` on the cast, because the batch-exports settings are `Optional[str]` for keyless auth while the integration path is not. That was then superseded by collapsing the pair into a single `AWSKeyPair`, which removes the overload, makes half a pair unrepresentable, and reduces transposition from "detected" to "impossible" at every consumer. The reviewer's stronger suggestion of a `NewType` per half is kept, applied to the object's two fields.

A second (agent-driven) security review audited all nine conversions against master and found no transpositions, then flagged the leftover positional sites now closed in the "Second review round" section above.

The signature-parsing conversion originally in this PR was dropped: master removed the agentic-provisioning hmac surface (#73387) while this was open, so that code no longer exists.


products/batch_exports/backend/temporal/pipeline/internal_stage.py
```diff
@@ -151,7 +157,8 @@ def __init__(self, *args, **kwargs):
 @asynccontextmanager
 async def get_s3_client():
     """Async context manager for creating and managing an S3 client."""
-    aws_access_key_id, aws_secret_access_key = _get_s3_credentials()
+    _s3_creds = _get_s3_credentials()
+    aws_access_key_id, aws_secret_access_key = _s3_creds.aws_access_key_id, _s3_creds.aws_secret_access_key
```
tomasfarias: Feels a bit redundant to return a tuple to immediately unpack it. I would prefer if we continue using the tuple below. I understand this was probably the smaller change, but if the issue is that:

> A caller that unpacks them in the wrong order still typechecks, so a swap ships silently, and for credentials a silent swap is a security bug, not just a readability nit.

Then we are still at risk here. Better if we access `_s3_creds.aws_access_key_id` and `_s3_creds.aws_secret_access_key` below.

tomasfarias (changes requested):
I'd prefer a small change, but overall the direction seems fine to me.

products/batch_exports/backend/temporal/pipeline/internal_stage.py
tomasfarias: Hell, if we want to fully embrace the types, then the absolute best thing to do is probably to define `AWSAccessKeyId` and `AWSSecretAccessKey` types, and then have every function below take those two types explicitly. 

Given that this helps the type checker to detect issues with the unpack, I would argue that's probably the safest option, even better than my previous suggestion.

EDIT: You would need to cast the strs to the concrete types somewhere. But that can be isolated in a single function marked as `unsafe` 🦀 Like `def _unsafe_cast_credentials(str, str) -> tuple[AWSAccessKeyId, AWSSecretAccessKey]` or using the new tuple type or something like that. Now callers know that unless they are directly messing with the unsafe function, the credentials are what they say they are, and the type checker enforces it.

EDIT2: Obviously this is Python so there's nothing stopping anybody from shooting themselves in the foot repeatedly with a rocket launcher, but oh well, it's a trade-off.

Piccirello (commented):
**Agent-driven security review** of the credential-handling changes. All non-blocking: five inline notes, plus one below that has no diff line to attach to.

The main check first. I audited all nine conversions producer-side and at every call site against master (`generate_scim_token`, `get_clickhouse_creds`, `sign_slack_request`, `_read_s3_credentials`, both `_get_s3_credentials`, `get_s3_function_call`, `generate_pkce`, and the challenge-only read in `probe.py`). Field order at construction and field reads at consumption are semantically identical to the master tuple positions in every case, so the refactor introduces no transposition. I also confirmed independently that `AWSKeyPair` never enters a Temporal payload: `insert_into_internal_stage_activity` is the only `@activity.defn` in that module, credentials are constructed inside the activity body, and `credentials` is a method parameter on `RecordBatchModel` rather than a serialized field.

### A second PKCE pair function is left unconverted

`posthog/api/oauth/toolbar_service.py:380`, `generate_pkce_pair() -> tuple[str, str]` returning `(code_verifier, code_challenge)`. It has zero callers today, so this is latent rather than live, but the risk lands on whoever writes the first one: transposing that pair sends the verifier as the `code_challenge`, which puts it in the authorization URL and so into browser history, referrers, and proxy logs, defeating PKCE. Either convert it or delete it.

### Two notes on the framing

- Slack signing is fail-closed. Transposing `sign_slack_request`'s output sends the signature as the timestamp and vice versa, and the verifier rejects it. That conversion is hygiene rather than a security fix.
- SCIM is the genuinely high-value one and is correctly prioritized: a swap there persists the plaintext bearer token in `scim_bearer_token` while breaking auth.

Unrelated to the code, in case it saves you time: @tomasfarias's requested changes do look addressed at head (field access at the point of use, plus the branded `NewType`s with construction isolated in an `unsafe_`-prefixed classmethod). The review just was not dismissed, so the automated deny still cites it as unresolved.

posthog/clickhouse/client/connection.py
```diff
@@ -121,15 +128,12 @@ def get_clickhouse_creds(user: ClickHouseUser) -> tuple[str, str]:
...
-    Returns:
-        tuple[str, str]: A tuple containing the username and password associated
-                         with the specified user.
     """
     global __user_dict
     if not __user_dict:
         __user_dict = init_clickhouse_users()
-    return __user_dict[user] if user in __user_dict else __user_dict[ClickHouseUser.DEFAULT]
+    creds = __user_dict[user] if user in __user_dict else __user_dict[ClickHouseUser.DEFAULT]
+    return ClickHouseCredentials(user=creds[0], password=creds[1])
```
Piccirello: The ClickHouse pair stays unbranded here. `init_clickhouse_users()` still returns `Mapping[ClickHouseUser, tuple[str, str]]`, so this indexes it positionally, and a future edit that swaps the tuple order inside that function is caught by nothing. Concentrating the risk in one place is defensible, it is the same shape as `unsafe_from_strings`, but it is inconsistent with how the AWS halves are treated in the same PR.

Worth noting a swap here is not only an auth failure. `dictionary_source_clickhouse` would emit `USER '<password>'` into DDL, and `posthog/dags/common/resources.py` logs `falling back to default user '{creds.user}'`, which puts the password into a log line.

One-line fix: build `ClickHouseCredentials` directly in `init_clickhouse_users` and drop the indexing. That removes the last positional site on this path.

products/mcp_store/backend/oauth.py
```diff
@@ -358,11 +359,17 @@ def register_dcr_client(metadata: dict, redirect_uri: str) -> tuple[str, str | N
     return client_id, client_secret, returned_auth_method
```
Piccirello: This is the same bug class the PR is about, left in place in the very module where `generate_pkce` was converted: three positionally adjacent strings, one of them a confidential OAuth client secret.

All three unpack sites are currently in the right order (`presentation/views.py:874`, `views.py:1135`, `probe.py:237`), so there is no live bug. But transposing the first two would persist the client secret as the `client_id` and then send it as a public identifier in the authorization redirect. Probably the highest-value remaining instance if you want a follow-up.

products/batch_exports/backend/temporal/pipeline/internal_stage.py
```diff
@@ -100,20 +101,19 @@ def _get_s3_endpoint_url() -> str:
...
-        aws_secret_access_key = None
-    else:
-        aws_access_key_id = settings.OBJECT_STORAGE_ACCESS_KEY_ID
-        aws_secret_access_key = settings.OBJECT_STORAGE_SECRET_ACCESS_KEY
-    return aws_access_key_id, aws_secret_access_key
+    if not _uses_object_storage_endpoint():
+        return None
+    access_key_id = settings.OBJECT_STORAGE_ACCESS_KEY_ID
+    secret_access_key = settings.OBJECT_STORAGE_SECRET_ACCESS_KEY
+    if access_key_id is None or secret_access_key is None:
```
Piccirello, line 114: The description's claim that the only input whose behavior changes is one that previously crashed is accurate, botocore does raise `PartialCredentialsError` for a half pair. The thing worth stating explicitly is that the new behavior is not "no credentials", it is "whatever botocore discovers" (env `AWS_*`, instance profile, `~/.aws/credentials`), which is a different principal.

Reachability is narrow enough that this is an observation rather than a problem. The branch needs `DEBUG or TEST or not CLOUD_DEPLOYMENT`, so Cloud takes the keyless branch anyway; under DEBUG/TEST the endpoint is pinned to `localhost:19000`; and self-hosted defaults to `http://objectstorage:19000` (`posthog/settings/batch_exports.py:90`), where scraped AWS credentials just 403. The only configuration where ambient credentials would actually authenticate is a self-hosted operator who both points `BATCH_EXPORT_OBJECT_STORAGE_ENDPOINT` at real S3 and sets exactly one of the two key vars.

So the practical effect is that a loud misconfiguration became a quiet 403. Consider a warning log when exactly one of the pair is set, matching what `init_clickhouse_users` already does for the equivalent ClickHouse case.

posthog/clickhouse/dictionaries.py
```diff
@@ -7,10 +7,10 @@ def dictionary_source_clickhouse(table: str) -> str:
...
+    creds = get_clickhouse_creds(ClickHouseUser.DICT_READER)
     connection_settings = f"TABLE {table} DB '{settings.CLICKHOUSE_DATABASE}'"
-    if user:
-        connection_settings += f" USER '{user}'"
-    if password:
-        connection_settings += f" PASSWORD '{password}'"
+    if creds.user:
+        connection_settings += f" USER '{creds.user}'"
+    if creds.password:
+        connection_settings += f" PASSWORD '{creds.password}'"
```
Piccirello, line 15: Pre-existing and not something this PR needs to fix, but it is the actual exposure path for the credential being tidied up here: the password is interpolated into dictionary DDL unescaped, unlike `get_s3_function_call` which doubles single quotes.

On Cloud this password is set, that being the point of the dedicated low-privilege user, so real credentials land in DDL text and from there in ClickHouse `query_log` and any migration error that surfaces the statement. Env-controlled, so not attacker-injectable. Flagging it because `field(repr=False)` covers tracebacks but does nothing for this path.

products/batch_exports/backend/temporal/sql.py
```diff
@@ -15,10 +17,10 @@ def get_s3_function_call(s3_folder: str, s3_key: str | None, s3_secret: str | No
     parameter substitution, so %% produces a literal % in the final query.
     """
     s3_url = f"{s3_folder}/export_{{{{_partition_id}}}}.arrow"
-    if s3_key is not None and s3_secret is not None:
+    if credentials is not None:
         # Escape single quotes by doubling them (ClickHouse SQL escaping)
-        escaped_key = s3_key.replace("'", "''")
-        escaped_secret = s3_secret.replace("'", "''")
+        escaped_key = credentials.access_key_id.replace("'", "''")
+        escaped_secret = credentials.secret_access_key.replace("'", "''")
```
Piccirello, line 23: Same pre-existing note as the dictionary DDL: the key and secret are inlined into the `s3(...)` call in the INSERT text, so they reach ClickHouse `query_log`. Self-hosted only, since Cloud is keyless, and unchanged by this PR. Just worth knowing that the `repr=False` work does not cover this path either.

Piccirello (approved):
Security side LGTM

lricoy (approved):
lgtm on the web analytics side

arnohillen:
Pushed `d3838667818`: the mypy `call-arg` failure was a test-only call site (`test_record_batch_model.py`) still passing the old `s3_key`/`s3_secret` kwargs; it now passes `credentials=AWSKeyPair(...)` like its sibling test. Repo-wide `mypy --cache-fine-grained` is clean again.

@tomasfarias on your point about not unpacking the typed credential into a tuple: the branch already accesses the fields directly (`credentials.access_key_id` / `credentials.secret_access_key` in `internal_stage.py` and `sql.py`), so there's no positional-swap surface left. Re-requesting your review.

products/batch_exports/backend/temporal/sql.py
tomasfarias, line 23: Unrelated to this PR, but do we even need to escape the secrets? Would AWS key Id/secret ever have quotes?

posthog/credentials.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Distinct static types for the two halves of an AWS key pair. They are plain `str` at runtime.
+# The point is that a type checker rejects passing a secret where a key id is expected.
+AWSAccessKeyId = NewType("AWSAccessKeyId", str)
+AWSSecretAccessKey = NewType("AWSSecretAccessKey", str)
+
+
+@dataclass(frozen=True)
+class AWSKeyPair:
+    """An AWS access key and its secret, kept together.
```
tomasfarias, line 12: Nice!

tomasfarias (approved):
The batch exports bits look good to me, thanks!

PR #73269, opened 2026-07-23
chore(ci): auto-rerun frontend CI jobs killed by runner shutdown

Radu-Raicea:
## Problem

Frontend CI's typecheck job runs on spot capacity and is periodically killed mid-step by a runner shutdown (spot preemption or the runner being recycled). GitHub reports this as a run *failure* — "The operation was canceled" / "The runner has received a shutdown signal" — even though the code is fine; re-running the same commit passes. Over a recent ~10h window this signature accounted for a meaningful share of frontend CI failures and was essentially exclusive to the typecheck job, so it's a recurring tax that currently needs a human to click "re-run".

## Changes

Add a `workflow_run` listener that fires when a `Frontend CI` run concludes as failure. It inspects the failed jobs' annotations and re-runs only the failed jobs when the *sole* cause was the transient runner-shutdown signature.

Safety properties:
- Reruns only when at least one leaf job failed transiently **and** no leaf job failed for any other reason — a genuine test/type error is never masked.
- Aggregator gate jobs (matching `… Tests Pass`) are excluded from the decision, since they only mirror their dependencies (they carry a generic `exit code 1`).
- A per-run attempt cap (`MAX_ATTEMPTS`, default 3) stops an infinite rerun loop if the flake turns out to be persistent.
- Uses `rerun-failed-jobs`, so passing jobs aren't re-executed.

Automated: added a local harness driving the classifier against the real annotation shapes (transient typecheck + gate → rerun; genuine Jest failure → no rerun; mixed transient+genuine → no rerun; attempt cap → no rerun; success conclusion → no rerun; gate-only failure → no rerun). All six cases behaved as expected. `node --check` passes on the script and the workflow YAML parses.

Not tested end-to-end in CI — `workflow_run` triggers only take effect once merged to the default branch, so the live behavior can only be observed after merge.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable — internal CI plumbing, no user-facing surface.

## 🤖 Agent context

Investigated a reported frontend typechecking failure that turned out to be a spot-instance preemption (confirmed via the runner's "shutdown signal" log line), then quantified how often the signature recurs across recent CI runs. Chose an auto-rerun listener over the alternatives that were considered and rejected: a step-level retry (`nick-fields/retry`) can't help because the runner itself dies with the step, and runner-size bumps only address OOM, which the shutdown-signal evidence rules out. Scoped to `Frontend CI` since that's where the signature was concentrated. Invoked the `/authoring-ci-workflows` skill while writing the workflow.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09ADEV3AJD/p1784816177005579?thread_ts=1784816177.005579&cid=C09ADEV3AJD)*

### DIFF
--- a/.github/scripts/ci-rerun-transient-failures.js
+++ b/.github/scripts/ci-rerun-transient-failures.js
@@ -19,22 +19,50 @@ const TRANSIENT_PATTERNS = [/operation was canceled/i, /runner has received a sh
 const GATE_JOB_PATTERN = /Tests Pass$/
 
 // Stop retrying once a run has reached this many attempts (so we auto-rerun at most twice).
-const MAX_ATTEMPTS = Number(process.env.MAX_ATTEMPTS || 3)
+const DEFAULT_MAX_ATTEMPTS = 3
+
+// A job failed transiently when it emitted failure annotations and every one of
+// them matches the runner-shutdown signature (no other failure reason present).
+function isTransientFailure(failureMessages) {
+    return failureMessages.length > 0 && failureMessages.every((m) => TRANSIENT_PATTERNS.some((p) => p.test(m)))
+}
+
+// Pure decision used by the orchestrator and the tests. `jobs` is the run's jobs
+// shaped as { name, conclusion, failureMessages: string[] }.
+function shouldRerun({ conclusion, runAttempt, maxAttempts = DEFAULT_MAX_ATTEMPTS, jobs }) {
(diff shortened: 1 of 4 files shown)


Radu-Raicea:
Closing in favour of what @gantoine is doing here: https://github.com/PostHog/posthog/pull/73298

PR #73276, opened 2026-07-23
feat(outcomes): bootstrap POC with celery batch evaluation

fercgomes:
## Problem

Outcomes is a new analytics primitive: a user-defined compound condition over events that, once met by a person, becomes a permanent dated fact and emits a `$outcome_reached` event. This PR bootstraps the POC slice tracked in [GROW-192](https://linear.app/posthog-team-growth/issue/GROW-192/bootstrap-poc-with-celery-tasks): a working end-to-end feature with UI, batch Celery evaluation, and the spec's full criteria grammar.

## Changes

New flag-gated product app `products/outcomes/` (flat layout like `links`/`field_notes`, main DB). The whole surface is behind the `outcomes-feature-enabled` feature flag: the scenes render a 404 and the nav entry is hidden when it's off, and the periodic task no-ops for teams with no outcomes, so this ships dark.

- **Criteria grammar (the spec's domain model):** paths OR'd together, atoms AND'd within a path with optional M-of-N per path, two levels of nesting and no more. An atom is `(event, property filters, aggregation, >= threshold)` where the aggregation is `count`, `sum(property)`, or `distinct(property)`. Monotonicity is a grammar-level invariant enforced at parse time: only aggregations that move toward satisfaction as events arrive are admissible (no avg, no "at most", no NOT/absence), which is what lets re-evaluation only confirm or add facts, never flip one. Limits: 5 paths, 10 atoms per outcome.
- **Semantics kernel (`criteria.py`):** a pure, dependency-free module that validates the grammar and resolves per-atom aggregates into `reached_at`, the winning path, and the evidence payload. It is the single definition of what "reached" means: the HogQL evaluator feeds its query output through the same `resolve()`, so query and kernel cannot drift — and a future streaming evaluator compiles from the same source.
- **Batch evaluator (Celery):** a periodic task fans out one task per outcome every 30 minutes. Each run compiles the criteria into one HogQL aggregate over `events` grouped by person, computing per atom both the attained value and the timestamp of the threshold-crossing event (the Nth matching event for count, the cumulative crossing for sum, the Nth first-seen value for distinct). `reached_at` is derived purely from those event timestamps — a path completes at its M-th completing condition, the outcome at its earliest completing path — so any two correct evaluators agree on it. Latching goes through `try_latch` (unique `(outcome, person)` insert-if-absent) and `$outcome_reached` is emitted via `capture_batch_internal` only for rows the run created, keeping emission effectively-once across replays. Facts whose `reached_at` predates the definition carry `backfilled: true`, so historical catch-up stays distinguishable from live reaches no matter which run finds it; outcomes on `$outcome_reached` itself are refused (loop guard).
- **Scan cost and convergence:** each run is bounded on both axes. A definition carries `lookback_days` (default 90, capped at the spec's one-year backfill depth) and the compiled query filters on it, so a run scans a window rather than a team's whole history. Within a run, the matching population is walked in `person_id` order with `evaluation_cursor` as the high-water mark, resetting when a sweep comes up short; dedup stays entirely with `try_latch`. The cadence is every 30 minutes and evaluation runs on its own `outcomes` Celery queue, so a slow sweep starves only itself. Measured against PostHog's own project (18.8B events, 835M persons) an unbounded, exclusion-list version scanned full history every 5 minutes and stopped making progress a few thousand latches in.
- **Latch evidence:** aggregate values only — per-condition attained vs threshold and the winning path index — no event IDs, no raw property values.
- **API:** DRF CRUD plus `reached` and `calculate` actions, with fully typed nested criteria serializers (atom/path/criteria flow into generated TS and MCP schemas). Criteria become immutable once facts exist (archive and recreate instead).
- **UI:** the create modal is a criteria builder — OR path groups, AND conditions with taxonomic event pickers, per-atom property filters, aggregation selectors with property pickers, thresholds, and an M-of-N selector per path. List and detail scenes render criteria summaries, and the reached table shows per-condition progress from the evidence.

| Outcomes list | Criteria builder | Outcome detail |
| --- | --- | --- |
| 01-outcomes-list | 02-new-outcome-modal | 03-outcome-detail |

Deferred by design for the POC: action-based atoms, anchored/rolling windows, group-level subjects, a reconciler re-emit, config-facade gating, `$outcome_reached` taxonomy registration, and streaming evaluation.

> [!NOTE]
> Deploy order: the `outcomes` Celery queue needs a worker entry in `charts` before tasks are consumed in production. Until then the periodic task enqueues and nothing runs, which is inert while the feature is flag-gated.

> [!NOTE]
> The initial migration only creates two new tables. FKs to `posthog.Team`/`User` use `db_constraint=False` per the hot-table policy, and both models are fail-closed team-scoped (`TeamScopedRootMixin`). Rated Safe by the migration risk analyzer.

- `products/outcomes/backend/tests/` (46 passed locally against real Postgres and ClickHouse):
  - `test_criteria.py` — the kernel: a parameterized rejection matrix for every inadmissible shape (non-monotone aggregations, bad thresholds, missing aggregation properties, path/atom limits, M-of-N bounds, the loop-guard event) plus `resolve()` semantics (AND completes at the last condition, OR at the earliest path, M-of-N at the M-th smallest completion, fail-toward-late when a completion is unknown, exact evidence payload shape).
  - `test_evaluation.py` — the compiled HogQL against real ClickHouse events: AND paths don't latch partial matches and `reached_at` lands on the last-completing condition; OR paths record the winning path; 2-of-3 latches at the second condition; sum crosses at the exact cumulative-threshold event; distinct crosses at the Nth first-seen value; property filters scope atoms; re-runs latch and emit nothing new; later runs emit only new reachers with `backfilled: false`; invalid stored criteria are skipped, not crashed on.
  - Evaluator bounds: `test_events_older_than_the_lookback_window_do_not_count` guards the timestamp bound (drop it and the scan goes back to full history), `test_sweep_advances_past_the_per_run_cap` guards forward progress past the per-run cap (the stall the exclusion list produced), and `test_backfilled_tracks_the_fact_not_the_run` guards against marking historical facts live on later runs. Evaluator fixtures are now time-frozen so they stay inside the lookback window instead of aging out of it.
  - `test_api.py` — serializer rejection matrix as a no-DB `SimpleTestCase` plus DB-backed wiring guards (create with the full grammar round-trips, team-scoped list with `reached_count`, criteria immutability once facts exist, `reached` ordering, `calculate` enqueues).
- Repo-wide `mypy --cache-fine-grained .` clean, `tsgo --noEmit` clean, `tach check` green, local semgrep over `products/outcomes/` clean, oxlint/oxfmt applied, migration risk analyzer Safe.
- Manual: I (actually Claude) ran Django + Vite in the task sandbox and built the criteria in the screenshots through the real UI (AND + sum aggregation + OR path via the builder). The reached rows are seeded fixture data since the sandbox runs no Celery worker; evaluator correctness is covered by the ClickHouse tests above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None: flag-gated POC, no public docs yet.

## 🤖 Agent context

- Authored end to end by Claude (PostHog Code cloud task) from the Outcomes engineering spec and GROW-192; Fernando directed scope, including the follow-up to implement the full criteria grammar rather than the single-atom simplification the first commit shipped.
- Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`, `/adopting-generated-api-types`.
- The initial migration was regenerated in place for the criteria/evidence schema (branch-only, never on master). The `AggregationEnum` entry in `ENUM_NAME_OVERRIDES` pins the metrics product's pre-existing OpenAPI enum name, which would otherwise get hash-suffixed now that a second `aggregation` ChoiceField exists; `OutcomeAggregationEnum` names ours.
- Two scaffold bugs found and reported via `hogli devex:feedback` (app config inserted into `SPECTACULAR_SETTINGS` instead of `PRODUCTS_APPS`; missing `backend/__init__.py`).
- The `outcomes-feature-enabled` flag was created via the PostHog MCP, active but scoped narrowly to start.

---


posthog/tasks/scheduled.py
```diff
@@ -292,6 +293,14 @@ def setup_periodic_tasks(sender: Celery, **kwargs: Any) -> None:
         name="refresh signals repository activity",
     )
 
+    # Outcomes batch evaluation - every 30 minutes, fans out one task per outcome
```
rafaeelaudibert: The only problem I see with this is that even though this is a POC, 30 minutes makes it useless for the use case we actually have in this team - measuring installation + onboarding for a product.

It helps with activation, I agree.

products/growth/dags/user_product_list_pruning.py
```diff
@@ -37,7 +37,7 @@
         'customer_analytics', 'dashboard', 'data-ops', 'early_access_features', 'endpoints',
         'engineering-analytics', 'error_tracking', 'experiments', 'feature_flags', 'heatmaps',
         'identity-matching', 'insights', 'links', 'live-debugger', 'logs', 'marketing',
-        'mcp-analytics', 'metrics', 'notebooks', 'product_tours', 'prompt-management', 'pulse',
+        'mcp-analytics', 'metrics', 'notebooks', 'outcomes', 'product_tours', 'prompt-management', 'pulse',
```
rafaeelaudibert: Yeah, dont need this here, I'm deleting this file right now

products/outcomes/backend/api.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class OutcomeCriteriaSerializer(serializers.Serializer):
+    paths = OutcomePathSerializer(
+        many=True, help_text="Paths OR'd together: a person reaches the outcome by completing any one path."
+    )
+
+
+# "Outcome" is already taken in the OpenAPI schema by session summaries.
+@extend_schema_serializer(component_name="OutcomeDefinition")
```
rafaeelaudibert: Should we, therefore, call it outcome definition? It feels like this is not really "an outcome" but it is indeed "an outcome definition". Doing this now is better than later because renaming stuff sucks in Django

products/outcomes/backend/api.py
```diff
@@ -0,0 +1,204 @@
...
+    class Meta:
+        model = OutcomeLatch
+        fields = ["id", "person_id", "distinct_id", "reached_at", "evidence", "created_at"]
+        read_only_fields = fields
+
+
+class OutcomeViewSet(TeamAndOrgViewSetMixin, viewsets.ModelViewSet):
+    """Create, read, update, and delete outcome definitions, and inspect who reached them."""
+
+    scope_object = "INTERNAL"
```
rafaeelaudibert: It would be cool if you shipped this with RBAC from today, it's not that bad/hard to do, agents do most of the job

products/outcomes/backend/api.py
```diff
@@ -0,0 +1,204 @@
...
+        outcome = self.get_object()
+        latches = outcome.latches.order_by("-reached_at")[:100]
+        return Response(OutcomeLatchSerializer(latches, many=True).data)
+
+    @extend_schema(
+        request=None,
+        responses={202: OutcomeSerializer},
+        description="Enqueue an immediate recalculation of this outcome instead of waiting for the periodic run.",
+    )
+    @action(methods=["POST"], detail=True)
```
rafaeelaudibert, line 288: Rate limits please? Even though it's POC, people might hit this. You didn't even gate the endpoints on the flag, so this is dangerous

rafaeelaudibert: Someone can just flood our Celery

rafaeelaudibert (approved):
Approving, but look at my comments, there are some changes that make sense to change in this first PR.

I don't really think we'd ever be able to release this on Celery, so... make sure people cannot kill our Celery queue with this. Might be a good idea to simply not use this in this state

fercgomes:
🤖 Review feedback addressed in 8ef6dc5 (all verified locally: 48/48 backend tests against Postgres + ClickHouse, repo-wide mypy clean, semgrep clean):

- **Rename** (@rafaeelaudibert): the model is now `OutcomeDefinition` (table `posthog_outcome_definition`, latch FK `definition`) while the rename is still free — the API surface (`/outcomes` routes) and generated TS types (`OutcomeDefinitionApi`) were already named that way, so nothing downstream moves.
- **Cadence**: evaluation now fans out every 5 minutes instead of 30, so reach-to-trigger latency fits onboarding/installation-style outcomes. Going tighter than that wants the streaming evaluator (next phase) rather than a hotter cron.
- **Flag gating + rate limits**: the whole viewset now requires the `outcomes-feature-enabled` flag server-side (403 when off, matching the UI gate) and carries the standard burst/sustained throttles.
- **Celery flooding**: `calculate` has its own 12/hour throttle plus a 2-minute debounce keyed on `last_calculated_at` — repeated clicks return 429 and enqueue nothing.
- **Signed sums** (Greptile): sum atoms now attain the *highest time-ordered running total* rather than the final sum, so a threshold crossing is permanent even when later refunds pull the total back down — this is what keeps `sum` genuinely monotone. Covered by a test with a refund event.
- **Subject cap no-progress** (Greptile): already-latched persons are excluded from the evaluation query (up to a capped exclusion list, with a warning beyond it), so each run's 10k budget goes to new reachers.
- **Emission retry** (Greptile): intentionally deferred — the spec's reconciler re-emit is the planned fix; a capture failure delays the event but the fact is durably latched. Noted in the PR body as deferred.

**Not done in this turn — RBAC**: adding access-control levels is a permissions change, which I don't make from review guidance alone under this run's safety rules. Happy to take it as an immediate follow-up if you want it before merge; the flag gate + throttles above cover the exposure concern meanwhile.

fercgomes:
🤖 Visual Review triage for the red `PostHog Visual Review / storybook` gate ([run 514a6a8b](https://us.posthog.com/project/2/visual_review/runs/514a6a8b-42b0-44b7-85c1-f7269b6b2070)) — verdict: **flake, recommend tolerating; needs a human click, I won't finalize a baseline change myself.**

The 2 changed snapshots are the same story in both themes: `scenes-app-insights-error-empty-states--long-loading`. I pulled the baseline and current PNGs and compared them: the baseline caught the state *after* the delayed "Need to speed things up?" optimization-hint box renders, the current caught the state *before* it, and the Query ID line differs per render. Nothing in this PR touches insights loading states. History for the story across its last 9 master runs shows it flipping to `changed` on master itself three times (Apr 24, Apr 29, May 18) on unrelated commits — a known time-dependent story.

Also re-ran the `Playwright E2E tests` job — 1 of 137 failed (`persons.spec.ts` tab locator timeout, persons page untouched by this PR).

products/outcomes/backend/api.py
```diff
@@ -0,0 +1,204 @@
...
+    def calculate(self, request: Request, **kwargs: Any) -> Response:
+        outcome = self.get_object()
+        calculate_outcome.delay(outcome_id=str(outcome.id), team_id=outcome.team_id)
```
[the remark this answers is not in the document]
fercgomes: Addressed in 3d0a847. Manual `calculate` now coalesces concurrent requests behind an atomic in-flight lock, on top of the existing 12/hour per-user throttle, the 2-minute debounce, and the per-team definition quota (`MAX_DEFINITIONS_PER_TEAM`). A burst now enqueues at most one task per window.

products/outcomes/backend/evaluation.py
```diff
@@ -0,0 +1,265 @@
...
+        offset += len(path.atoms)
+
+    query = f"""
+        SELECT {", ".join(columns)}
+        FROM events
+        WHERE event IN {{events}}
+        GROUP BY person_id
+        HAVING {" OR ".join(path_conditions)}
+        LIMIT {{limit}}
+    """
```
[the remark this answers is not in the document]
fercgomes: Addressed: already-latched persons are excluded from the evaluation query (`person_id NOT IN {excluded_person_ids}`, up to a capped exclusion list with a warning past the cap), so each run's subject budget goes to new reachers and the outcome keeps making forward progress across runs rather than re-processing the same capped set.

products/outcomes/backend/api.py
```diff
@@ -0,0 +1,256 @@
...
+        responses={202: OutcomeDefinitionSerializer},
+        description="Enqueue an immediate recalculation of this outcome instead of waiting for the periodic run.",
+    )
+    @action(methods=["POST"], detail=True)
+    def calculate(self, request: Request, **kwargs: Any) -> Response:
+        definition = self.get_object()
+        if definition.last_calculated_at and timezone.now() - definition.last_calculated_at < CALCULATE_DEBOUNCE:
+            raise Throttled(detail="This outcome was calculated moments ago. Try again in a couple of minutes.")
+        calculate_outcome.delay(outcome_id=str(definition.id), team_id=definition.team_id)
+        return Response(self.get_serializer(definition).data, status=status.HTTP_202_ACCEPTED)
```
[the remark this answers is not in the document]
fercgomes: Fixed in 3d0a847. `last_calculated_at` is only written when a run finishes, so a burst of concurrent calls all read the same stale value and each enqueued. The enqueue is now guarded by an atomic `safe_cache_add` in-flight lock (exactly one caller wins per debounce window), and a test asserts two rapid calls enqueue the task only once.

products/outcomes/backend/evaluation.py
```diff
@@ -0,0 +1,297 @@
...
+        logger.warning(
+            "outcomes_latch_exclusion_truncated",
+            outcome_id=str(definition.id),
+            team_id=definition.team_id,
+            limit=MAX_LATCH_EXCLUSIONS,
+        )
+        excluded_person_ids = excluded_person_ids[:MAX_LATCH_EXCLUSIONS]
+
+    query, placeholders = _compile_query(criteria, definition.team, excluded_person_ids)
+    response = execute_hogql_query(query, placeholders=placeholders, team=definition.team)
```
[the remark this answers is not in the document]
fercgomes: The access boundary is enforced at the API. The viewset composes `scope_object = "outcome"` (registered as an RBAC resource in `scopes.py` and `user_access_control.py`) with the mixin's default `APIScopePermission` + `AccessControlPermission` and the feature-flag gate, so creating, reading, or calculating outcomes requires the outcome scope and access level (setting `permission_classes` on the viewset extends the defaults, it does not replace them).

Batch evaluation runs as a userless background aggregation, like PostHog's other Celery batch jobs, so it carries no per-request principal. Threading an authorized principal through the evaluator is the same change as the deferred streaming evaluator and is out of scope for this flag-gated POC. Happy to add explicit query-access enforcement on top of the outcome scope as a follow-up if you'd prefer it before wider rollout.

products/outcomes/backend/evaluation.py
```diff
@@ -0,0 +1,265 @@
...
+            token=outcome.team.api_token,
+            event_source=EVENT_SOURCE,
+        ).raise_for_status()
+    except Exception:
+        logger.exception(
+            "outcomes_emission_failed",
+            outcome_id=str(outcome.id),
+            team_id=outcome.team_id,
+            event_count=len(latches),
+        )
```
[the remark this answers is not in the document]
fercgomes: Deferred for the POC (not a regression from this change). Emission runs strictly after the latch is durably committed, so a capture failure delays `$outcome_reached` but never loses the fact. The spec's reconciler re-emit is the planned recovery post-POC; this is called out in the PR body and in the `_emit_outcome_reached` docstring.

fercgomes:
🤖 Babysitting update: pushed `3d0a847` and merged master, so the branch is current.

**Fixed this round**
- **Concurrent `calculate` race** (graphite / veria): the manual recalc debounced only on `last_calculated_at`, which is written when a run finishes, so a burst of concurrent calls all saw the same stale value and each enqueued a full-history run. The enqueue is now guarded by an atomic `safe_cache_add` in-flight lock (one caller wins per debounce window), layered on the existing 12/hour throttle, 2-minute debounce, and per-team quota. A test covers the burst-to-single-enqueue case.

**Already handled on the branch (now resolved)**
- Subject-cap forward progress (greptile): already-latched persons are excluded from each evaluation query, so a run's budget goes to new reachers.

**Deferred, left open for your call**
- Failed-emission retry (greptile): the latch is durable; the reconciler re-emit is the planned post-POC recovery.
- Userless batch evaluation (veria): the API enforces the `outcome` RBAC scope + flag gate; threading a per-user principal through the Celery evaluator is the same work as the deferred streaming evaluator.

The merge conflict the auto-resolver flagged earlier was already resolved by the intervening commits (GitHub reports the branch mergeable); merging master in just brings it current on unrelated files. Watching CI now.

gantoine (approved):
changes outside of `products/outcomes` look good

PR #73277, opened 2026-07-23
feat(data-warehouse): BI insight builder in the SQL editor Visualization tab

mjwarren3:
## Problem

Building a chart from a SQL query today means hand-picking axes and series from dropdowns after every run. People coming from Tableau or Power BI expect the inverse: drag fields into wells, get a chart, and let the tool write the SQL.

This PR adds a drag-and-drop insight builder to the SQL editor's Visualization tab, behind the `bi-sql-insight-editor` feature flag. Flag off, the editor is unchanged.

## Changes

With the flag on, the editor splits into two top-level tabs. **Source** holds the SQL editor, the database tree, and a results grid showing the base query's raw rows (loaded lazily through a second data node, cache-friendly, torn down with the tab). **Visualization** gives the whole scene to a four-column builder canvas: a fields panel derived from the query's result columns, Rows/Columns/Values/Filters wells, the live chart with a chart / results table / generated SQL view toggle, and a format column.

One axis convention across charts: Columns is the primary axis (x-axis, pie slices, the table's grouping dimension) and Rows is the breakdown (multiple lines, stacked segments). A capability registry per chart type drives which wells are enabled, required, and capped, with replace-on-full and add-vs-replace hover states.

The wells compile to a single-pass `SELECT dims, aggregates FROM <base> GROUP BY dims` over the editor query. A bare `SELECT * FROM <table>` base compiles directly against the table (dropping the preview limit), so the generated SQL is what a person would write. The compiled SQL is stored in `source.query`, so saved insights and dashboards execute one HogQL query like any SQL insight; the well config rides along in a new `builder` field on `DataVisualizationNode` for re-editing. Date dimensions get grain bucketing (hour through year), numeric dimensions get fixed-width binning, measures get the usual aggregations, and filters compile to a `WHERE` clause with type-aware operators.

Two shared-code changes worth calling out for the classic editor:

> [!NOTE]
> `dataVisualizationLogic` resets chart config whenever a response's columns change. That heuristic is correct for hand-picked axes but wiped the builder's compiled settings on every recompile, so the reset is now skipped for builder queries (`query.builder?.enabled`). Non-builder queries are untouched.

> [!NOTE]
> The format sidebar (`SideBar`) now falls back to the first visible tab when the stored tab is hidden for the current chart type. Previously it rendered the hidden tab's content, which was a small pre-existing bug reachable in the classic editor too.

The Visualization tab: fields, wells, and the live chart. This is a saved builder insight reopened via a hard reload of its URL — the wells and chart hydrate directly:

builder visualization tab

The Source tab keeps the SQL editor, database tree, and the base query's raw rows:

source tab

Out of scope for this first slice: pivot tables (multi-dimension shapes fall back to the grouped table) and a dedicated data-source picker. Both are planned follow-ups.

### Second increment: Source/Visualization split, flag safety, and QA fixes

The follow-up push reshapes the layout and hardens the flag story:

- **Source | Visualization tabs** replace the fullscreen expander (see above). Insight storage is unchanged: `source.query` still holds the compiled SQL, `builder` still holds the well config.
- **Verify-on-open**: opening a builder insight recompiles the saved wells (whitespace-insensitive) and compares against the saved SQL. On mismatch — e.g. the SQL was edited while the flag was off — the builder offers an explicit choice: restore the visual setup (regenerates the SQL) or keep the SQL (cleanly degrades to a plain SQL insight). No silent clobbering in either direction.
- **Flag-off degradation**: every "is this tab builder-owned?" check is gated on the flag, so a builder insight opened with the flag off is a plain SQL tab — edits run, saves work, the Series tab and axis auto-recovery behave classically. This makes mixed flag states (and per-user rollouts) safe.
- **Hydration is subscription-based** on the query value rather than a specific action, so the wells and chart type hydrate regardless of how the node reaches the tab (open, reload with a tab hash, tab restore) or in what order the canvas mounts.
- **Correctness fixes from review feedback**: a warning banner when the base query's trailing `LIMIT` caps what the chart aggregates; tab flips no longer silently re-snapshot a half-edited base; rejected well adds and cross-well drags can no longer destroy fields; builder recompiles run cache-friendly (`async`) with only the explicit Run forcing execution; filter values commit on blur/Enter instead of per keystroke; user-set series labels survive recompiles; `inf`/`nan`/`team_id`/`true`/`false` joined the reserved alias list.
- **UX**: neutral field pills with an inline grain/bin chip (matching the aggregation chip), wells hug their pills, better empty and loading states, heatmap axis pickers locked to the wells for builder queries, and the Format panel no longer shows the Series tab for builder queries (Setup owns field selection).

### Third increment: reload hardening, lossless flag-off saves, and the master merge

A review pass over the whole branch, plus fixes for the two failure modes QA kept hitting:

- **Reloading a visualization URL hydrates directly.** Cold loads raced three async things (the lazy canvas chunk, the insight fetch, and feature flags) and could land on a blank pane: the output pane was gated on Monaco initializing, but Monaco never mounts inside the hidden query pane on a Visualization-first load. The output pane now renders without Monaco while the builder hosts the scene, a `hydrated` flag keeps the preview on a spinner (never the "pick fields" empty state) until the wells reflect the node, and the open flow waits (bounded) for feature flags before its one-shot builder-vs-classic decision — so console flag overrides survive reloads.
- **Edit detection is immune to compiler drift.** The builder config now stores a `compiledQuery` snapshot at write time; "was this SQL edited outside the builder?" compares against the snapshot instead of recompiling, so compiler output changes in later releases can't silently degrade previously saved builder insights. Configs saved before the field existed fall back to recompile-and-compare.
- **Flag-off saves strip the stale visual setup.** "Update insight" now drops the `builder` config when the builder doesn't host the tab (matching "Save as new insight"), so a flag-off edit + save is lossless in the database, not just in the UI — the insight reopens the classic way even after the flag comes back.
- **Tighter containment on shared surfaces.** `isBuilderOwnedQuery` is now driven only by the explicit prop the SQL editor passes — dashboards and the insight scene behave byte-identically to master even for saved builder nodes. A layout regression that shrank the query pane for *all* SQL editor users (the hide-wrapper became the pane's containing block, collapsing its `max-height: 35%`) is fixed with `display: contents`.
- **Cleanups:** the three duplicated hydration guards collapsed into one `hydrateIfNeeded` helper, and a dead selector was removed.
- The save-dialog placeholder rename was reverted (master's e2e suites fill that input by placeholder text), which also makes `insightLogic` byte-identical to master — one more unflagged change gone.
- **Merged master:** the new save-as-metric action is threaded through the branch's restructured save menu and `createTab`, and the branch's `LemonCollapse` animation patch was dropped in favor of master's `useAnimatedPresence` rewrite, which supersedes it (one less unflagged shared-component change in this PR).

### Not behind the feature flag

Almost everything above only activates for builder-hosted tabs or for queries carrying a `builder` config (which only exist once someone uses the flag). The exceptions, which apply to everyone — a full audit of the diff, so reviewers see the whole blast radius:

**Bug fixes riding along in shared code** (defensible unflagged, but real behavior changes):

- **"Update insight" enablement** (`editorSceneLogic.updateInsightButtonEnabled`): the name comparison now baselines against `insightTabName()` (`name ?? derived_name ?? 'Untitled'`) and the saved-side query is run through `sanitizeSourceQuery` before comparing. The button now correctly reports "No updates to save" where master wrongly showed it enabled (derived-name-only insights, legacy insights with a top-level `connectionId`).
- **`updateInsight` only sends `name` on an actual rename.** Master wrote `name: activeTab.name` on every save — which could overwrite a derived-name-only insight's `name` with "Untitled" on first save. This changes what's persisted for all insight updates from the SQL editor.
- **Tab titles use `derived_name`** when an insight has no explicit name (new `insightTabName` helper), instead of "Untitled".
- **Stale-response fix on insight switch**: opening an insight only skips the query run when the cached response actually answers *that insight's* query (`answersThisInsight`). Master skipped on any cached response, leaving the previous insight's rows under the new chart config.

**Load sequencing** (the one to scrutinize — it touches every URL-driven open):

- **The Monaco wait-loop is gone.** Master polled up to ~10s for Monaco before creating a tab from `open_insight`/`open_query`/view/metric URLs. Now the tab's identity (name, insight, sourceQuery, buffer) is created immediately and the Monaco model is backfilled by `initialize` when the editor mounts. Same destination, different ordering on cold loads.
- The editor loading overlay now also covers the pane while an insight is being fetched (`insightLoading`), not just view loads.

**Mechanical / cosmetic**:

- The SQL editor query pane's height fix (`display: contents` on the hide-wrapper) — this *restores* master behavior that an earlier commit on this branch had regressed.
- Opening a SQL insight in the editor now consistently lands on the Visualization output tab after the insight loads (same destination as before, applied uniformly).
- `runQuery` gained an optional `refreshMode` parameter; the default behavior for non-builder tabs is byte-identical.
- `BoldNumber` gets an overflow guard in its SCSS; the classic visualization wrappers in `OutputPane` gain `min-w-0` (flex overflow guard).
- `lib/api`'s `queryTabState.user` returns `null` on 404 instead of throwing.
- The secondary save-menu's `disabledReason` is computed once in the memo instead of at each render site — verified behavior-identical.

Also audited and confirmed *not* changed for flag-off users despite touching shared files: `SideBar`'s tab fallback (dead code unless the visible-tab set changes, which only happens builder-hosted), `HeatmapSeriesTab` (reduces to master when not builder-owned), and `dataVisualizationLogic.isBuilderOwnedQuery` (prop-only; dashboards and the insight scene never pass it).

Unit tests cover the compiler (grain/bin expressions, escaping, aliasing, filter clauses, select-all detection), the capability registry (validation, effective wells, well enablement, chart auto-pick), the wells-to-chart-settings mapping, and the builder logic (compile-on-change, hydration round-trip, no auto-switch, bin and filter compilation) - 131 tests across four suites, all passing.

I (or, actually Claude) also verified the flows end to end against the live dev stack with a headless browser: built charts through the field menus, confirmed the breakdown splits on first add and survives bar → stacked bar and pie → line switches, measured that a wide table stays within the viewport with the format column visible, broke the base query to confirm the "Unable to resolve field" error surfaces inside the builder with the stale pill flagged and recovery works, and confirmed the flag-off editor still auto-picks axes with the split-view toggle intact.

Not manually tested by a human yet: drag-and-drop interactions (the automated run used the field menus) and the save-and-reopen round-trip through a dashboard.

Second increment: 586 Jest tests green across the editor and visualization suites, including explicit flag-off coverage (plain-SQL editing of a builder insight, axis recovery, Series tab visibility) and both hydration orderings (node-before-mount and mount-before-node). Full `tsgo` type check and lint clean. A new Playwright spec (`sql-editor-insight-builder.spec.ts`) covers the build/save journey and the flag-off edit → conflict → restore lifecycle; it parses and lists but has not run locally (the local e2e setup endpoint hangs — reported to devex), so watch its first CI run. Manual QA in the running app covered insight open/edit across flag states, the conflict flow, and the Source/Visualization split.

Third increment: 143 Jest tests green across the builder, editor, and visualization suites, including a new `builderNodeConsistency` suite (snapshot vs recompile edit detection), the flag-off `updateInsight` strip, and `hydrated`-state transitions. The Playwright spec gained a hard-reload journey (reload straight into Visualization must hydrate the chart, never the empty state) and an API assertion that a flag-off save actually removes `query.builder` from the persisted insight. The reload fix and the query-pane height fix were verified in a real browser against the live dev stack, reloading the exact URL from the bug report. Post-merge, the same suites plus master's new save-as-metric tests all pass.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Behind a feature flag with no public docs yet; docs will come when the flag rolls out.

## 🤖 Agent context

Built with Claude (PostHog Code) across an iterative session directed by @mjwarren3, who reviewed each round in the running app. The branch is a clean re-cut of a larger experiment (draft #72416): the builder engine and UI were ported as new files, the integration was rewritten narrowly against master, and the experiment's extra surfaces (data-source picker row, editor intent split, pivot table) were deliberately left out of this slice.

Skills invoked: `/verify` and the repo's `/run-posthog` for the live-browser verification pass. Key decisions along the way: chart capabilities went through three models before landing on "chart type is primary, extra fields are silently ignored"; the axis convention was flipped mid-stream (Columns = x-axis) to make chart-type switching consistent; and the deepest bug (breakdown lost on chart switch) was root-caused by reading kea store state through Vite's module graph in the live app - the classic "columns changed → reset chart config" heuristic was wiping the builder's settings, which is why the guarded skip in `dataVisualizationLogic` exists.

Second increment (same setup, directed by @mjwarren3 with round-trip QA in the running app): the Source/Visualization split was the reviewer's own design; flag-safety went through two designs (strip-on-save was rejected in favor of verify-on-open so mixed flag states stay lossless); and two field bugs found in live QA — a blank canvas on insight open and a chart-type picker desync — were both root-caused to hydration being tied to a single action's dispatch order and fixed with a value subscription.

Third increment (same setup): the reload blank was root-caused live with a headless browser against the dev stack — the previous hydration fixes were correct but sat below the real deadlock (the output pane gated on Monaco mounting inside a display:none pane). Notable design decision: verify-on-open now compares against a stored compile snapshot rather than recompiling, after review flagged that any compiler change would orphan every previously saved builder insight.

---


thmsobrmlr:
This would be so awesome to have!! 💖 My main question here would be what the plan is for rolling this out?

IMO this has great potential and can "easily" be iterated into a full-fletched solution. At the moment the UX isn't quite there yet though (I don't have Power BI/Tableau experience, so maybe someone like @annaszell has better intuition).

And a big thing with the current implementation of the flag is that we can only ever roll it out to full orgs and never disable it again. Otherwise we'd run into several issues with insights build using this editor. This can be solved, but may also be fine. Just need to make sure we don't experiment on individual users for example.

Besides this a couple of things came up from agent review (probs can just be fed to an agent to fix as well):
1. `sqlEditorLogic.tsx:1922` — `runQuery` writes the Monaco buffer into `source.query` while keeping `builder.enabled` + compiled chartSettings. Exit fullscreen → edit base SQL → Run → "Update insight" persists the corrupted node, which renders broken on dashboards.
2. `compileBuilderQuery.ts:271` — an inner `LIMIT` survives compilation. Clicking a column in the catalog seeds `SELECT col FROM t LIMIT 100`, so aggregates compute over 100 arbitrary rows with no truncation warning. Silently wrong numbers get saved as insights.
3. `insightBuilderLogic.ts:701` — flipping back to the Visualization tab auto-runs `refreshBase` whenever Monaco differs from the snapshot: a half-edited (even invalid) buffer silently replaces the saved `baseQuery` and drops the view linkage, bypassing the "Refresh fields" banner.
4. `sqlEditorLogic.tsx:1940` / `Wells.tsx:346` — every builder run uses `force_async` (cache bypass), and filter inputs recompile per keystroke behind a 300 ms breakpoint. Typing one filter value fires several uncached ClickHouse aggregations. Should be `'async'` + commit filters on blur/enter.
5. `insightBuilderLogic.ts:590` — adding to a disabled well (max 0) via the fields-panel menu deletes the added field *and* any stashed ones: on the default Table chart "Add to Rows" is a silent no-op, and on Bar it permanently wipes a stashed breakdown.
6. `BuilderCanvas.tsx:133` — `onDragEnd` removes the pill before `addToWell` validates, so dragging "Count of rows" onto Rows/Filters destroys the measure. Legit transfers also drop `numericBinWidth`.
7. `wellsToVizSettings.ts:36` — recompiles clobber user formatting: custom series labels are overwritten on every `applyWells` (`measure.label` is never written), and the alias-keyed carry-over drops number formatting and orphans conditional-formatting rules when the aggregation changes.
8. `compileBuilderQuery.ts:53` — columns named `team_id`/`inf`/`nan` compile to invalid aliases (runtime SyntaxError), while the 70-word `RESERVED_ALIASES` list renames `date`/`order`/etc. for no reason — HogQL accepts all of them.

Happy to stamp it, if you'd like to test internally.

mjwarren3:
@thmsobrmlr agreed on your points! It even doesn't quite feel like it fits well into the SQL editor for a few reasons:

- There's now both a "Base query" and "Generated visualization query", and its a bit confusing how those work together. It also isn't clear what should show up in the results pane
- It makes it a bit confusing on whether you should go to your SQL query or the visualizer to make the changes you want to make
- It needs more screen space with all of the columns it is displaying, including the visualization
- A bunch of the existing Formatting / Display logic feels bolted on, rather than built specific to this editor

Let me rethink this a bit before we move forward. I do think the general idea makes sense, but the implementation into our existing insight creation flow is a bit tricky. Definitely curious if @annaszell has any thoughts on where this should belong. Maybe it fits better with some of the new things coming out of data stack than it does the existing SQL editor?

annaszell:
okay so first of all: amazing job!! 

1. yeah I agree that UI isn't close to more established BI tools, but also we probably don't need that level of sophistication, especially not for our current icp! imo we can release this as is and iterate as more user feedback becomes available

2. as for where it should live, i see both pros & cons for having it in the sql editor, as well as in product analytics. i'm leaning toward the latter for two reasons:
	- you wouldn't necessarily have this in competitor sql editor, those who can't write sql won't really go there anyway? yes maybe it's useful for editing existing stuff but then idk how the tool would handle it (e.g. you have existing sql, and you want to tweak it with the insight builder, what happens to the sql code? is it updated? is it deleted? etc)
	- the way the drag & drop vibes work is more similar to how a product analytics chart works compared to the current features of the visualization tab there

that said, i'm not saying don't do viz tab if you feel like it would fit there well!

mjwarren3:
Appreciate it both! @thmsobrmlr  I'm going to feed this feedback into the agent. I'm also going to see if we can find a way to make the BI-editor created charts and the traditional SQL insights more compatible, so we don't have to roll out to an entire org and have the option to revert if needed. Will let you know when I have something I think we can roll out and start testing internally

thmsobrmlr:
> you wouldn't necessarily have this in competitor sql editor, those who can't write sql won't really go there anyway? yes maybe it's useful for editing existing stuff but then idk how the tool would handle it (e.g. you have existing sql, and you want to tweak it with the insight builder, what happens to the sql code? is it updated? is it deleted? etc)
> the way the drag & drop vibes work is more similar to how a product analytics chart works compared to the current features of the visualization tab there

oh, that actually makes a lot of sense!

@mjwarren3 maybe we could make it so that anyone who can see such a BI insight can also freely edit it, and the feature flag only affects who sees the create buttons for these types of insights.

mjwarren3:
Pushed the second increment (6c92f0d1a24): Source | Visualization top-level tabs (replacing the fullscreen expander), verify-on-open flag safety so mixed flag states degrade losslessly, the review-feedback correctness fixes, subscription-based builder hydration, and UX polish. PR description is updated with the full change list — including an explicit **Not behind the feature flag** section for the handful of shared fixes that affect everyone. New Playwright spec included; its first run will be in CI (local e2e setup endpoint is currently broken).

mjwarren3:
Pushed the third increment and merged master (`84dc836e3c7`, `8567719efe1`, merge `cb59fdb0f5e`). Changes since the last round of comments:

**The two field bugs from QA are actually fixed now** (the earlier hydration fixes were correct but sat below the real problem):

- Reloading a visualization URL (`#...&output_tab=visualization&insight=...`) landed on a blank pane. Root-caused live in a browser: the output pane was gated on Monaco initializing, but Monaco never mounts inside the hidden query pane on a Visualization-first load — so nothing rendered, no matter how well hydration worked underneath. The output pane now renders without Monaco while the builder hosts the scene, a `hydrated` flag keeps the preview on a spinner instead of the "pick fields" empty state, and the insight-open flow waits (bounded, 2s) for feature flags before its one-shot builder-vs-classic decision, so console flag overrides survive reloads.
- A layout regression from the split-tabs work shrank the query pane (and left a gray dead band) for **all** SQL editor users, flag on or off — the hide-wrapper div became the pane's containing block and collapsed its `max-height: 35%`. Fixed with `display: contents`; Source now lays out exactly like master.

**Rollout-safety hardening**, addressing the mixed-flag-states discussion above:

- The builder config now stores a `compiledQuery` snapshot at save time. Edit detection ("was this SQL changed outside the builder?") compares against the snapshot instead of recompiling, so future compiler changes can't silently degrade previously saved builder insights to classic.
- "Update insight" with the flag off now strips the stale `builder` config from the persisted query (matching "Save as new insight") — the flag-off edit/save round-trip is now lossless in the database, not just in the session.
- Dashboards and the insight scene no longer branch on the persisted `builder` bit at all (`isBuilderOwnedQuery` is prop-only), so surfaces outside the SQL editor behave byte-identically to master for everyone.

**Merge with master**: the new save-as-metric action is threaded through the restructured save menu and `createTab`, and this branch's `LemonCollapse` animation patch is dropped in favor of master's `useAnimatedPresence` rewrite — one less unflagged shared-component change riding along.

**Tests**: 143 Jest tests green across the builder/editor/visualization suites (new coverage for the snapshot comparison, the flag-off strip, and hydration states); the Playwright spec gained a hard-reload journey and an API assertion that a flag-off save removes `query.builder` from the stored insight. PR description and screenshots are updated to the current UI.

mjwarren3:
Closed, since an alternative version was already merged and wanted to avoid code mess: https://github.com/PostHog/posthog/pull/77479#pullrequestreview-4864288497

PR #73278, opened 2026-07-23
fix(wordpress): handle non-JSON responses without crashing

Gilbert09:
## Problem

The WordPress data warehouse source crashed with a raw `JSONDecodeError` (`Expecting value: line 1 column 1 (char 0)`) whenever a site returned a `2xx` response whose body wasn't JSON. Error tracking: [issue 019f8f4e](https://us.posthog.com/project/2/error_tracking/019f8f4e-27d7-7cd3-af26-b6c78f05508c).

The failure is in `get_rows`, which called `response.json()` directly on any successful response. WordPress sites hit this in a few common ways, none of which are our bug:

- an empty body on a collection with nothing to return
- an HTML error, login, or maintenance page served with a `200` (security and caching plugins do this)

Because the crash wasn't classified, Temporal retried the whole activity against a deterministic failure until the budget ran out, and every attempt minted a fresh `$exception`.

## Changes

Parse the body inside the retry loop and branch by shape, mirroring what the shared REST engine already does for engine-based sources (`rest_client._send_request`):

- **Empty body on a `2xx`** - treat as an empty page and let pagination stop, instead of crashing.
- **Non-JSON body** (doesn't begin with a JSON value) - raise `RESTClientNonRetryableError`. Re-fetching returns the same page, so we fail fast. `_handle_import_error` already classifies this type as non-retryable across REST sources.
- **Body that starts as JSON but fails to parse** - a truncated read, kept retryable via the existing `WordpressRetryableError`.

The surfaced message is built from a scheme/host/path-only URL (`_safe_url`), so the query string never leaks into error tracking. The source also gets a friendly non-retryable message keyed on the stable `Non-JSON response from` prefix, so the schema is disabled with actionable copy pointing at REST-API/plugin config.

> [!NOTE]
> This is complementary to the shared-engine fix in #73146 (a different error-tracking issue). That one makes `RESTClientNonRetryableError` a `NonReportableError` so it stops being captured. WordPress doesn't use the shared engine, so it needed its own crash fix - and once #73146 lands, routing through the same exception type means WordPress benefits from the noise suppression for free.

Automated tests (authored and run by me, Claude, running autonomously):

- `test_empty_body_terminates_without_crashing` - a `2xx` with an empty body stops pagination instead of crashing. Catches a revert of the crash fix (the exact reported bug).
- `test_non_json_body_is_non_retryable` - a `2xx` HTML page raises `RESTClientNonRetryableError` with a `Non-JSON response from` message that omits the query string. Catches a regression to a raw crash or an endless-retry classification, plus query-string leakage.
- `test_truncated_json_body_stays_retryable` - a body that begins as JSON but fails to parse stays retryable. Locks in the non-JSON vs truncated distinction so a future "make all decode errors non-retryable" change can't silently stop syncing on a transient truncated read.
- `test_non_json_response_message_matches_non_retryable_error` (source) - the realistic full message is matched by the classifier's stable-prefix substring lookup and maps to a non-None friendly message.

Ran the full wordpress suite (113 passed), `ruff check`/`ruff format`, `tach check --dependencies --interfaces` (the new downward import into `sources/common` stays within the boundary), and repo-wide `uv run mypy --cache-fine-grained .` (no issues in 16723 files). Invoked the `/writing-tests` skill before adding tests.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior, API, or documented workflow changes. Add `skip-inkeep-docs`.

## 🤖 Agent context

Triaged from the error-tracking issue with the PostHog error-tracking MCP tools (`query-error-tracking-issue-events`). The stack trace pointed at `wordpress.py:390` (`get_rows` → `response.json()`), confirming the failure is in this source's own code rather than serialized log context.

Decision: this is a fixable bug, not just a `NonRetryableErrors` addition - our code crashed on a response shape it should handle. I chose to mirror the shared REST engine's three-way body handling rather than blanket-classifying every `JSONDecodeError` as non-retryable, which would have wrongly stopped syncs on transient truncated reads. Reused `RESTClientNonRetryableError` (handled by type in `_handle_import_error`) instead of inventing a WordPress-specific exception, so behavior stays consistent with other REST sources and dovetails with #73146.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/wordpress/source.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/wordpress/source.py
@@ -105,6 +105,10 @@ To sync private content or authenticate, create an [Application Password](https:
             "404 Client Error": "WordPress returned 404 (Not Found) for this collection. The REST API or this specific endpoint (for example /users, which security plugins often block to prevent user enumeration) may be disabled or restricted on your site. Enable REST API access for it, or remove this table from the sync.",
             HOST_NOT_ALLOWED_ERROR: "The WordPress site URL is not allowed. Please use a publicly reachable site URL.",
             HTTP_NOT_ALLOWED_ERROR: "The WordPress site URL must use HTTPS when credentials are provided. Please update the site URL to use https://.",
+            # A 2xx response whose body isn't JSON (an HTML error/login/maintenance page, often from
+            # a security or caching plugin) can't be turned into rows by retrying. Matches the stable
+            # prefix RESTClientNonRetryableError uses, not the variable URL that follows.
+            "Non-JSON response from": "The WordPress site returned a non-JSON response (for example an HTML error, login, or maintenance page) instead of data. Confirm the REST API is enabled and reachable at this URL and isn't blocked by a security or caching plugin, then try again.",
         }
 
     def get_canonical_descriptions(self) -> CanonicalDescriptions:
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/wordpress/tests/test_wordpress.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/wordpress/tests/test_wordpress.py
@@ -6,6 +6,9 @@ from unittest import mock
 
 import requests
 
+from products.warehouse_sources.backend.temporal.data_imports.sources.common.rest_source.rest_client import (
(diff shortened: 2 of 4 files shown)


MarconLP (approved):

danielcarletti (approved):
LGTM

PR #73279, opened 2026-07-23
feat(replay-vision): show credits with a dollar anchor in scanner UI

arnohillen:
## Problem

Replay vision's in-product surfaces showed cost purely in dollars, but the unit we bill and sell is credits (1 credit = $0.01) and the billing page shows credits. So the product spoke two languages: dollars in-product, credits on the billing page. Users had no way to connect "$3.55 this month" to the credit balance they buy, and choosing a pricier model didn't visibly map to anything.

## Changes

Credits are the unit everywhere. The dollar anchor is reserved for the three surfaces that are actually about spend, so the rest of the UI stays quiet.

**Shows credits with a dollar anchor** (`formatCredits` → `"N credits (≈ $X.XX)"`):

- The spend meter on the scanner list
- The "Spend this period" column in the scanner list
- The cost estimate (the editor's estimate card, and the same number in the scanner's config summary)

**Shows plain credits** (`formatCreditCount` → `"N credits"`): the model picker, the quota banner, the quality tab's test-run cost, the breakdown tooltips, and the trigger tooltips. `formatCreditsRange(used, total)` renders a "used of limit" pair as `1,200 of 5,000 credits`.

The billing page is unchanged. It already shows credits.

## Screenshots

Before is master (dollars only). Captured on the Hedgebox demo project. The shots predate #73907, which renamed these labels from "this month" to "this period"; only the wording moved, not the formatting shown here.

#### Spend meter (scanner list)

**Before**

before-1-spendcard

**After**

after-1-spendmeter

#### "Spend this period" column (scanner list)

**Before**

before-1b-table

**After**

after-1b-table

The dollar sits on its own muted line here rather than inline, so the cell never wraps mid-parenthetical at narrow widths. Same treatment as the spend meter.

#### Cost estimate (scanner editor, scan conditions step)

The per-observation rate next to it is credits only, so the card carries one dollar figure rather than two.

**Before**

before-3-forecast

**After**

after-3-forecast

#### Same estimate in the config summary (scanner detail)

**Before**

before-5-readonly

**After**

after-5-readonly

#### Model picker (scanner editor)

**Before**

before-2-modelpicker

**After**

after-2-modelpicker

#### Quality tab test-run cost

**Before**

before-4-quality

**After**

after-4-quality

#### Quota banner (scanner detail)

Also reworded so the sentence reads naturally with the credit range. The near-limit state is simulated here; the demo project sits at 2% usage.

**Before**

before-6-banner

**After**

after-6-banner

- `utils/credits.test.ts` (14 parameterized cases): the credits-to-dollars conversion, pluralization ("1 credit" vs "N credits"), thousands separators, cent rounding of fractional estimates, and the "used of limit" range. This guards a wrong-dollars regression on a billing-facing surface that no existing test covered (the helper had none).
- Full Replay vision frontend Jest suite: 25 suites, 277 tests pass.
- Oxlint and Oxfmt clean on the changed files. Repo typecheck reports no errors in the changed files.
- I (Claude, agent-assisted) verified every surface in the local dev app on the Hedgebox demo project, including the near-limit quota banner via a stubbed quota response.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change. This is an in-product display refinement.

## 🤖 Agent context

Arnaud directed this while we decide dollars vs credits for the Replay vision launch. The AI product family (PostHog AI, Code, Inbox) bills in credits and the billing page shows raw credits, while Replay vision's in-product UI had shipped dollars-only. The direction is "credits everywhere, dollar anchor on the spend surfaces": consistent with the unit we sell, with the dollar kept visible where someone is reading actual spend so it stays transparent rather than an opaque credit wall. An earlier revision of this PR put the anchor on every surface, which read as noisy in review ([thread](https://posthog.slack.com/archives/C03PB072FMJ/p1785161162054229)), so it is now scoped to three.

Built with Claude Code (Opus 5). Skills invoked: superpowers:brainstorming (to lock the display format before writing code) and the repo `/writing-tests` gate (for the new unit test). Considered and rejected: dollars-only on the secondary surfaces (that re-creates the two-languages problem), and adding a dollar anchor to the shared billing gauge (bigger blast radius across every credit-billed AI product, out of scope here). Keeping the formatting behind two helpers is what lets each surface pick its register with a small diff.


fasyy612 (approved):
Sweet!

PR #73289, opened 2026-07-23
fix(conversations): keep the active view when using a ticket's back arrow

christiaan-ph:
## Problem

Clicking the back arrow from a ticket doesn't return you to the saved view you were on. It briefly shows the view, then reverts to the unfiltered "all tickets" list. Annoying when you work out of a filtered view or saved view all day.

Root cause: the back arrow navigated to a bare `urls.supportTickets()` with no query string. The tickets list treats the URL as the source of truth, so on return `supportTicketsSceneLogic`'s `urlToAction` sees no `?view=` (or filter) params while a view was active, hits its `leavingSavedView` branch, and resets to the default filters. The persisted filters render for a frame first, which is why it looks like "loads my view, then reverts".

## Changes

Carry the list's query string through the round trip so the URL stays the source of truth:

- The ticket row attaches the current filters / saved view (the list's query string) onto the ticket URL when opening a ticket.
- The ticket back arrow (`forceBackTo`) re-attaches that query string onto the list URL, so `urlToAction` restores the same view (or `loadSavedView` for a `?view=` param) instead of clearing it.
- Skipped when the table is embedded (e.g. the person side panel), where the host page's query string isn't the ticket filters.

Both use `combineUrl`, so an empty query string still yields the plain path.

**Automated:** Added a regression test in `SupportTicketsScene.test.tsx` covering that opening a ticket row carries the active filter query string onto the ticket URL (previously it dropped it, so the back arrow landed on the unfiltered list). The conversations Jest suite passes locally, 6/6, via `hogli test products/conversations/frontend/scenes/tickets/SupportTicketsScene.test.tsx`. Frontend typecheck and lint/format are clean.

**Manual:** verified in local dev on a project seeded with tickets and saved views, watching the URL query string at each step:

- Saved view: from `/support/tickets?view=<id>`, opening a ticket navigates to `/support/tickets/<n>?view=<id>` (query string carried through), and the back arrow returns to `/support/tickets?view=<id>` with the view still active. No flash of the view followed by a reset to the unfiltered list.
- Ad-hoc filter: from `/support/tickets?status=open`, opening a ticket carries `?status=open` onto the ticket URL, and the back arrow returns to `?status=open` with the filter still applied.
- No filter: from a bare `/support/tickets`, opening a ticket and going back lands on a clean `/support/tickets` with no stray query string.
- Open in new tab: Cmd or middle-clicking a row from a filtered or saved-view list opens the ticket in a new tab with the query string carried.

All four scenarios passed.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Christiaan reported the back-arrow behavior and pointed me at his session recordings, which showed the tickets live at `/support/tickets` with filters in the URL. I traced it to the `forceBackTo` target dropping the query string and the `leavingSavedView` branch in `supportTicketsSceneLogic`. I first patched the wrong repo (a separate app that mirrors this UI) before the session data made clear the real product was Conversations.

Authored by the PostHog Slack app (Claude). Diagnosis and fix by the agent; direction from Christiaan.


products/conversations/frontend/scenes/tickets/SupportTicketsScene.test.tsx
```diff
@@ -97,4 +100,32 @@ describe('SupportTicketsTable selection', () => {
         await waitFor(() => expect(getRowCheckbox()).not.toBeChecked())
         expect(logic.values.selectedTicketIds).toEqual([])
     })
+
+    // Regression: opening a ticket from a filtered list dropped the filters, so the ticket's
+    // back arrow returned to the unfiltered "all tickets" list instead of the view the user
+    // was on. The row now carries the list's query string onto the ticket URL so the back
+    // arrow can restore it.
+    it('carries the active filters onto the ticket URL when a row is opened', async () => {
```
luke-belton, line 109: worth adding a test for the back case too, given it's the important bit of the regression?

luke-belton (approved):

products/conversations/frontend/scenes/tickets/SupportTicketsScene.test.tsx
christiaan-ph: addressed in [`bb87081` (this commit)](https://github.com/PostHog/posthog/pull/73289/commits/bb87081e72415b22d7f2bbae8f48b2c5c18a09be)

PR #73294, opened 2026-07-23
fix(flags): require project admin to edit default release conditions

gustavohstrassburger:
## Problem

A report flagged that org members could `PUT` to `/api/environments/{id}/default_release_conditions/` even though the settings UI disables that editor for anyone below project admin. The frontend and backend disagreed, and the backend is the only real enforcement point.

Root cause: the `default_release_conditions` action was gated by `TeamMemberLightManagementPermission`, which only escalates to admin for `DELETE`. Every other method (including `PUT`) needed just `MEMBER`. Since the action has no `DELETE`, its write was effectively member-writable. The sibling `logs_config` and `experiments_config` actions already do this correctly with `TeamMemberStrictManagementPermission`.

Review then surfaced the same hole in the sibling `default_evaluation_contexts` action: any project member could `POST` new default evaluation contexts (applied to all new flags) via the API, even though the settings UI gates adding behind project admin.

## Changes

- Gate the `default_release_conditions` and `default_evaluation_contexts` actions with `TeamMemberStrictManagementPermission` (members read, admins write) on `ProjectViewSet` (`posthog/api/project.py`). This is the only viewset that serves these routes: `/api/environments/*` is rewritten in-process to the canonical `/api/projects/*` viewset by `EnvironmentsRewriteMiddleware`.
- Remove the duplicate `default_release_conditions` and `default_evaluation_contexts` handlers from `TeamViewSet` (`posthog/api/team.py`). Neither `TeamViewSet` nor `RootTeamViewSet` is registered on any router, so those handlers were unreachable dead code; this also drops the now-unused shared helper and scope-downgrade constant.
- Gate the evaluation-context remove control in the settings UI behind project admin, matching the add controls and the release-conditions editor.

Session auth bypasses API-scope checks, so the effective gate for browser users is the management permission — which is exactly what this change tightens.

- Regression test in `posthog/api/test/test_team_default_release_conditions.py`: a member-level user can `GET` (200) but a member `PUT` now returns 403.
- Updated `products/feature_flags/backend/api/test/test_feature_flag_default_evaluation_contexts.py`: a member `POST`/`DELETE` now returns 403, while member reads and admin writes are unaffected. The cross-viewset parity tests (`test_team_project_differential.py`) continue to expect matching statuses across `/api/environments/` and `/api/projects/`.
- `ruff check`, `ruff format`, and repo-scoped `mypy` pass on the changed backend files. The DB-backed Django tests run in CI (no local Postgres/ClickHouse available in the authoring environment).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change — this aligns the API with the existing admin-only UI behavior.

## 🤖 Agent context

Investigated and implemented by Claude (PostHog Code). Skills invoked: `/improving-drf-endpoints` (viewset change) and `/writing-tests` (test change).

Key correction from the initial version of this PR: the reported `/api/environments/*` routes are served solely by `ProjectViewSet` after `EnvironmentsRewriteMiddleware` rewrites them to `/api/projects/*`; `TeamViewSet`/`RootTeamViewSet` are not routed. The permission fix therefore lives only in `project.py`, and the matching `team.py` handlers were removed as dead code rather than mirrored. The same permission gap was found and fixed in the sibling `default_evaluation_contexts` action.

---

### DIFF
--- a/posthog/api/team.py
+++ b/posthog/api/team.py
@@ -1881,9 +1881,10 @@ class TeamViewSet(TeamAndOrgViewSetMixin, AccessControlViewSetMixin, viewsets.Mo
     lookup_field = "id"
     ordering = "-created_by"
 
-    # Actions whose scope is downgraded to project:read for session auth on all methods.
-    # TeamMemberLightManagementPermission still applies, so DELETE requires admin.
-    MEMBER_READABLE_CONFIG_ACTIONS = ("default_release_conditions", "default_evaluation_contexts")
+    # Actions whose scope is downgraded to project:read for session auth on all methods,
+    # so members can write via the UI. Handlers here must enforce any admin-only sub-operations
+    # themselves (see default_evaluation_contexts, which gates unhiding on admin level).
+    MEMBER_READABLE_CONFIG_ACTIONS = ("default_evaluation_contexts",)
 
     # Actions whose GET is downgraded to project:read for session auth; mutating methods stay on project:write/admin.
     GET_DOWNGRADE_ACTIONS = ("evaluation_context_suggestions",)
@@ -2099,11 +2100,12 @@ class TeamViewSet(TeamAndOrgViewSetMixin, AccessControlViewSetMixin, viewsets.Mo
     @action(
         methods=["GET", "PUT"],
         detail=True,
-        permission_classes=[TeamMemberLightManagementPermission],
+        permission_classes=[TeamMemberStrictManagementPermission],
         url_path="default_release_conditions",
     )
     def default_release_conditions(self, request: request.Request, id: str, **kwargs) -> response.Response:
-        """Manage default release conditions for new feature flags in this team."""
+        """Manage default release conditions for new feature flags in this team. Members can read;
+        writing requires project admin, matching the admin-only settings UI."""
         team = self.get_object()
         config = get_or_create_team_extension(team, TeamFeatureFlagDefaultsConfig)
 
--- a/posthog/api/team.py
+++ b/posthog/api/team.py
@@ -1,6 +1,5 @@
 import re
 import json
-import math
 import secrets
 from datetime import timedelta
 from functools import cached_property
@@ -90,12 +89,7 @@ from posthog.utils import (
 )
 
 from products.customer_analytics.backend.facade.team_extension import TeamCustomerAnalyticsConfig
-from products.feature_flags.backend.models import TeamFeatureFlagDefaultsConfig
-from products.feature_flags.backend.models.evaluation_context import (
-    EvaluationContext,
-    TeamDefaultEvaluationContext,
-    normalize_context_name,
-)
+from products.feature_flags.backend.models.evaluation_context import EvaluationContext, normalize_context_name
 from products.logs.backend.models import TeamLogsConfig
 from products.signals.backend.models import SignalSourceConfig
 from products.workflows.backend.models.team_workflows_config import TeamWorkflowsConfig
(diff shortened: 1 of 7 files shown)


gustavohstrassburger:
@greptile please review

posthog/api/team.py
```diff
@@ -2099,11 +2100,12 @@ def generate_conversations_public_token(self, request: request.Request, id: str,
     @action(
         methods=["GET", "PUT"],
         detail=True,
-        permission_classes=[TeamMemberLightManagementPermission],
+        permission_classes=[TeamMemberStrictManagementPermission],
```
haacked: `suggestion`: This permission fix never runs. `RootTeamViewSet` isn't registered by any router; the only registered `TeamViewSet` subclass is `ProjectEnvironmentsViewSet`, which unconditionally raises `PermissionDenied` in `initial()`. The route `/api/environments/{id}/default_release_conditions/` gets rewritten in-process to `/api/projects/{id}/...` by `EnvironmentsRewriteMiddleware`, so [`ProjectViewSet.default_release_conditions`](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/project.py#L1657) is the only permission check that executes. The regression test still passes because it posts to the `/api/environments/...` URL, and the rewrite routes it through the live `ProjectViewSet` check, not this dead one.

The description should be corrected too, since it currently says both viewsets serve this route, which is misleading. Also consider deleting this handler; it duplicates `team_default_release_conditions_view`'s body.

```diff
-    @action(
-        methods=["GET", "PUT"],
-        detail=True,
-        permission_classes=[TeamMemberStrictManagementPermission],
-        url_path="default_release_conditions",
-    )
-    def default_release_conditions(self, request: request.Request, id: str, **kwargs) -> response.Response:
-        """Manage default release conditions for new feature flags in this team. Members can read;
-        writing requires project admin, matching the admin-only settings UI."""
-        team = self.get_object()
-        config = get_or_create_team_extension(team, TeamFeatureFlagDefaultsConfig)
-        ...  # rest of the method, unreachable via any registered route
```

posthog/api/team.py
```diff
@@ -1881,9 +1881,10 @@ class TeamViewSet(TeamAndOrgViewSetMixin, AccessControlViewSetMixin, viewsets.Mo
     lookup_field = "id"
     ordering = "-created_by"
 
-    # Actions whose scope is downgraded to project:read for session auth on all methods.
-    # TeamMemberLightManagementPermission still applies, so DELETE requires admin.
-    MEMBER_READABLE_CONFIG_ACTIONS = ("default_release_conditions", "default_evaluation_contexts")
+    # Actions whose scope is downgraded to project:read for session auth on all methods,
+    # so members can write via the UI. Handlers here must enforce any admin-only sub-operations
+    # themselves (see default_evaluation_contexts, which gates unhiding on admin level).
```
haacked: `suggestion`: This comment says `default_evaluation_contexts` self-enforces its admin-only sub-operations, but that's only true for un-hiding an already-hidden context. Adding a brand-new one isn't gated at all. On the live route ([project.py:1678](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/project.py#L1678)), the action only requires `[IsAuthenticated]` plus the appended `TeamMemberLightManagementPermission`; this requires admin for DELETE but not for POST. Any org member can POST up to 10 new default evaluation contexts that apply to all new feature flags, but can't remove them afterward.

This mirrors the gap that the Bugcrowd report caught for `default_release_conditions`. If asymmetric permissions (members add, admins remove) are intentional, document it here instead of "gates unhiding on admin level". Otherwise, apply this fix:

```diff
-        permission_classes=[IsAuthenticated],
+        permission_classes=[TeamMemberStrictManagementPermission],
```
(`default_evaluation_contexts` in [`posthog/api/project.py:1678`](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/project.py#L1678), and the same action in [`posthog/api/team.py:2204`](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/team.py#L2204) for consistency)

posthog/api/team.py
```diff
@@ -1881,9 +1881,10 @@ class TeamViewSet(TeamAndOrgViewSetMixin, AccessControlViewSetMixin, viewsets.Mo
...
+    MEMBER_READABLE_CONFIG_ACTIONS = ("default_evaluation_contexts",)
```
haacked: `suggestion`: This action list is duplicated as a literal tuple in two places: `MEMBER_READABLE_CONFIG_ACTIONS` here and an inline tuple at [project.py:1402](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/project.py#L1402). Dropping `default_release_conditions` required identical edits to both. [`TEAM_CONFIG_MEMBER_FIELDS_SET`](https://github.com/PostHog/posthog/blob/ec887395991426a7354c388f73be1d2099baeb81/posthog/api/team.py#L465) avoids this by living at module level with `project.py` importing it.

```diff
 # team.py, module level:
+MEMBER_READABLE_CONFIG_ACTIONS = ("default_evaluation_contexts",)

 # team.py, inside TeamViewSet:
-    MEMBER_READABLE_CONFIG_ACTIONS = ("default_evaluation_contexts",)

 # project.py imports:
     from posthog.api.team import (
+        MEMBER_READABLE_CONFIG_ACTIONS,
         TEAM_CONFIG_MEMBER_FIELDS_SET,
     )

 # project.py:1402
-        if self.action in ("default_evaluation_contexts",):
+        if self.action in MEMBER_READABLE_CONFIG_ACTIONS:
```

posthog/api/team.py
```diff
@@ -1881,9 +1881,10 @@ class TeamViewSet(TeamAndOrgViewSetMixin, AccessControlViewSetMixin, viewsets.Mo
     lookup_field = "id"
     ordering = "-created_by"
 
-    # Actions whose scope is downgraded to project:read for session auth on all methods.
-    # TeamMemberLightManagementPermission still applies, so DELETE requires admin.
-    MEMBER_READABLE_CONFIG_ACTIONS = ("default_release_conditions", "default_evaluation_contexts")
+    # Actions whose scope is downgraded to project:read for session auth on all methods,
+    # so members can write via the UI. Handlers here must enforce any admin-only sub-operations
```
haacked: `nit`: Keep the fact that DELETE still requires admin. `default_evaluation_contexts` (the only action left) accepts DELETE and relies on `TeamMemberLightManagementPermission` to enforce that.

```diff
     # Actions whose scope is downgraded to project:read for session auth on all methods,
-    # so members can write via the UI. Handlers here must enforce any admin-only sub-operations
-    # themselves (see default_evaluation_contexts, which gates unhiding on admin level).
+    # so members can write via the UI. TeamMemberLightManagementPermission still applies,
+    # so DELETE requires admin. Handlers here must enforce any admin-only sub-operations
+    # themselves (see default_evaluation_contexts, which gates unhiding on admin level).
```

haacked (approved):
Looks good, the actual fix is solid and tested. A few non-blocking things inline, including a gap in `default_evaluation_contexts` with the same shape as this one.

posthog/api/team.py
```diff
@@ -158,108 +152,6 @@ def handle_logs_config(request: request.Request, team: Team) -> response.Respons
...
-                        team=team,
-                        request=request,
-                    )
-
-                return response.Response({"success": True})
-            except EvaluationContext.DoesNotExist:
-                return response.Response({"error": "Evaluation context not found"}, status=404)
-
-
 def handle_evaluation_context_suggestions(request: request.Request, team: Team) -> response.Response:
```
haacked: `nit`: The comment on line 162 points at `default_evaluation_contexts` for the root-team rationale, and that action is gone from this file. The live version is `team_default_evaluation_contexts_view` in [posthog/api/project.py:421](https://github.com/PostHog/posthog/blob/4f0f33742c8442f00581611f59a1459acd571517/posthog/api/project.py#L421).

PR #73036, opened 2026-07-22
chore(ci): Auto-resolve trivial PR conflicts

Twixes:
## Problem

Conflicts against `master` show up constantly at our scale. Some of them require human intervention and judgment. A lot of them are trivial and should never be seen by a human. 

Sadly, neither GitHub nor Graphite nor Trunk, as far as I know, seem to be tackling this, despite conflicts being a scourge for all large teams shipping fast. But we can solve this for ourselves. 

## Changes

Proposing a new Actions workflow `pr-autoresolve-conflicts.yml` that scans for conflicting PRs targeting master and just resolves trivial conflicts.

Caveats: PRs on forks are skipped (can't be pushed to), drafts are skipped, Graphite-stacked PRs (`base: graphite-base/*`) only get a comment telling the author to restack.

Chart by Claude:

```mermaid
flowchart TB
    classDef entry fill:#f9bd2b,stroke:#e0a800,color:#000,font-weight:bold
    classDef step  fill:#eef1f5,stroke:#c7ccd1,color:#111
    classDef agent fill:#1d4aff,stroke:#1a3fd6,color:#fff,font-weight:bold
    classDef token fill:#f54e00,stroke:#d43e00,color:#fff,font-weight:bold
    classDef out   fill:#1d4aff,stroke:#1a3fd6,color:#fff

    cron([⏱ every 15 min]):::entry

    subgraph DETECT["🔍 detect · read-only"]
        scan["find conflicting master PRs<br/>skip already-attempted"]:::step
    end

    subgraph RESOLVE["🧩 resolve · runs PR code · no write token"]
        direction LR
        agent{{"run agent on<br/>code conflicts"}}:::agent
        regen["regenerate lockfiles<br/>&amp; API types"]:::step
        agent --> regen
    end

    subgraph FINALIZE["🔐 finalize · App token · trusted"]
        route{"pick outcome"}:::token
    end

    c1(["push signed commit"]):::out
    c2(["post 'needs a human'"]):::out
    c3(["post 'restack it'"]):::out

    cron --> scan
    scan -->|"code conflict"| agent
    scan -->|"graphite stack"| route
    regen --> route
    route -->|resolved| c1
    route -->|"can't resolve"| c2
    route -->|graphite| c3
```

Note that the job that runs PR-controlled code has no write tokens for safety.

Resolution is three-tier to minimize all costs:

1. Lockfiles, migration numbering →  cheap `hogli ci:preflight --fix`
[...]

PR #73298, opened 2026-07-23
chore(skills): add autoresolving-pr-conflicts loop skill

gantoine:
## Problem

Merge conflicts against `master` show up constantly at our scale, and most of the trivial ones (lockfiles, generated types, migration numbering) should never need a human. #73036 solves this with a scheduled GitHub Actions workflow, but a cron-driven scan in CI spends shared `GITHUB_TOKEN` rate-limit budget and Actions runners on every tick.

We already have a platform primitive built for exactly this shape of automation: a tasks Loop, fired by GitHub webhooks, running a coding agent in a sandbox with an App-scoped token on its own rate-limit bucket.

## Changes

Adds `.agents/skills/autoresolving-pr-conflicts/`, the operating procedure for running the conflict autoresolver as a Loop instead of in CI:

- `SKILL.md` is the procedure the loop's sandbox agent follows on each fire (a push to `master`): sweep open PRs with local `git merge-tree` over bulk-fetched PR heads (zero API calls for detection), regenerate conflicted artifacts deterministically (`hogli ci:preflight --fix`, `hogli build:openapi` when Docker is available), resolve source conflicts only with high confidence, and land exactly one signed flattened commit on the PR head. Everything else gets flagged for a human with a sticky status comment.
- `references/loop-setup.md` is the operator runbook: the exact loop-creation payload (push trigger filtered to `master`, `overlap_policy: skip` as the debounce), operational limits, and a dark-launch test sequence against a disposable PR before the trigger goes live.

Design notes:

- The marker comment format (`<!-- autoresolve-attempt:head:master -->`) is shared with #73036, so the two implementations never double-attempt the same state and can be dark-launched side by side.
- The signed-commit tooling produces a single-parent commit on the PR head, which matches the flattened-commit shape #73036 produces via `createCommitOnBranch`.
- The skill hard-forbids opening/merging PRs, history rewrites, and writes to protected branches, forks, or agent-owned branches.

No behavior changes in this PR. It's documentation plus a skill; the Loop itself is created separately via the API per the runbook.

`bin/hogli lint:skills` passes (the new skill is picked up; the only warnings are pre-existing in an unrelated skill). This is a docs/skill-only change with no executable code. The runbook's test sequence (loop `preview/` dry-run, then a scoped manual fire against a disposable conflicting PR) is the intended end-to-end validation before the push trigger is enabled, and hasn't been run yet.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed. The skill and its runbook are the documentation.

## 🤖 Agent context

Claude Code (Fable 5) authored this after an investigation of alternatives to #73269 and #73036, which are the right designs but add sustained GitHub API load in CI. We mapped the Loops trigger pipeline (`products/tasks/backend/loop_github_events.py` supports push triggers with branch filters) and the task sandbox's git capabilities (write-scoped App token, existing-branch checkout, signed-commit-only egress) with two parallel exploration agents, then wrote the skill against those verified capabilities. Invoked `/writing-skills` before authoring. Known open question, called out in the PR body and runbook: the exact `git_signed_commit` semantics live in the external `@posthog/agent` package, so the runbook mandates a disposable-PR test before enabling the trigger.


rnegron (approved):
TIL about Loops, let's try them out!

Only 1 small nitpick: skill and docs are a bit too verbose maybe? can we run `/simplify` on it (if not done already)?

Image

PR #73306, opened 2026-07-23
fix(dev): survive dev-stack port collisions and stop silent zookeeper data loss

haacked:
## Problem

Quitting the phrocs TUI (`q`) or running `hogli stop`/`down` stops the supervised processes but leaves the docker compose containers running. That's intentional (`docker compose up -d`, one stack shared across worktrees), but nothing communicates it and the full-teardown commands (`hogli docker:services:down`/`remove`) are hard to discover.

The failure cascade this enables, observed in practice:

1. A stale stack under a *different* compose project keeps holding host ports indefinitely.
2. The next `docker compose up` has one container fail its port bind, which aborts the whole `up` (and can leave a container detached from the compose network entirely). The existing pre-flight in `bin/start` only checked for a foreign clickhouse container and printed an advisory.
3. The "obvious" recovery, force-recreating the broken container, is a trap for zookeeper: it declared no volumes, so its coordination state lived in anonymous volumes that recreation silently drops. That wipes the ZooKeeper replica registrations of every local `ReplicatedMergeTree` table, forcing a full ClickHouse + ZooKeeper reset and re-running all migrations.

## Changes

**Named volumes for zookeeper and clickhouse** in `docker-compose.dev.yml`, mirroring what `docker-compose.hobby.yml` already does (`zookeeper-data`/`datalog`/`logs`, `clickhouse-data:/var/lib/clickhouse`). Both together, not just ZK: persisting only one side leaves ZK registrations pointing at vanished CH data (or vice versa), which fails with "replica already exists" and is worse to diagnose. `docker compose down && up` is now lossless. Zookeeper also gets hobby's autopurge settings so the now-persistent volume doesn't grow without bound.

> [!WARNING]
> One-time disruption: the first `docker compose up` after this recreates both containers empty (both images declare `VOLUME`, so the data previously lived in anonymous volumes that recreation discards). Run `hogli migrations:run` (or `hogli dev:reset` for a full wipe + demo data) once per machine. Documented in the local dev handbook page.

**Pre-flight port check, as `hogli doctor:ports`** (`bin/start` calls it the same way it already calls `doctor:zombies`, skipped if `hogli` isn't on PATH). It checks the always-on infra ports (proxy 8010, zookeeper 2181, clickhouse 8123/9000, kafka 9092, postgres 5432, redis 6379, temporal 7233, objectstorage 19000) and attributes each held port to its docker container and compose project, or to a plain process via lsof (report only). In a TTY it offers to tear down a foreign stack (`[y/N]`, 30s timeout defaulting to no), but only when the stack looks like a PostHog dev stack (its holder has a clickhouse container in any state); an unrelated project that happens to hold postgres or redis ports is reported, never offered a teardown. Compose labels are sanitized before reaching the terminal or a docker command line. Non-TTY stays advisory. Fail-open throughout, so it can never block `bin/start`. Also runs as one of the checks in `hogli doctor`. The docker-compose proc in `bin/mprocs.yaml` and the hogli-generated config still print a pointer back to this output when `docker compose up` fails.

This started as bash directly in `bin/start`; moved to `hogli doctor:ports` during review since it's the largest, most logic-dense piece of this PR and `bin/start` has no test harness, while hogli-commands already does (same reasoning as the existing `doctor:zombies`/`dev:generate` delegation).

**Teardown discoverability**: phrocs prints a hint after interactive quit and after `phrocs stop` (suppressed in sandboxes), telling you containers are still up and naming `hogli docker:services:down`/`remove`. The `hogli stop`/`down` descriptions, phrocs README, and the developing-locally handbook page say the same. Also fixed phrocs' `StripComposeLogsTail` to preserve the trailing `fi` when the logs tail sits inside the new sandbox guard, so the static `bin/mprocs.yaml` fallback shell still parses.

Not done, deliberately:

- Per-worktree port namespacing: the single shared stack is an explicit design choice (`COMPOSE_PROJECT_NAME=posthog` pinned in `.envrc`, `bin/start`, and the devenv generator).
- A quit-and-teardown keybind: one keystroke killing infra another worktree is using would recreate the failure class this fixes.
- elasticsearch/opensearch anonymous volumes: their loss is self-contained (no cross-container coordination state); possible follow-up.

- `go build ./... && go vet ./... && go test ./...` in `tools/phrocs`: all pass.
- hogli-commands tests: 91 pass (`test_devenv.py` including 4 new cases, `test_detached.py`).
- New test `TestDockerComposeShellFailurePath` (parameterized, fake `docker` on PATH, no real docker): if the `|| { …; exit 1; }` fallback is dropped from the generated shell or its `bin/mprocs.yaml` twin, a failed `docker compose up` would emit the `docker-compose ready` marker and dependent procs would start against a dead stack; the test pins both shells' fail and success paths.
- `hogli doctor:ports` has pytest coverage: port-matching against both plain and range-published forms (with a case pinning that port 9000 doesn't false-match 19000), the partial-stack clickhouse-any-state filter, lsof disambiguation, the confirmation timeout, and the exact `docker compose down` argv when confirmed. Also ran it live via the real `hogli` CLI against the dev stack and against a crafted malicious compose label (confirms sanitization holds end to end, not just in the old bash).
- Verified the interactive prompt end to end under a real PTY with `expect`: the prompt reads the answer from the terminal and the default-No path leaves the foreign stack running. (An earlier revision read the answer from a herestring instead of the TTY; caught in review and fixed.)
- `docker compose -f docker-compose.dev.yml config` confirms the named volume mounts resolve.
- Not tested by the agent: a real container-recreation cycle exercising volume persistence, and the prompt's `y` path (it would tear down the live stack).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated `docs/published/handbook/engineering/developing-locally.md` (teardown commands + one-time volume migration gotcha) and `tools/phrocs/README.md` in this PR.

## 🤖 Agent context

Authored with PostHog Code (Claude) from an investigation prompt describing the port-collision-into-ZK-wipe cascade. The agent explored phrocs exit handling and the compose files, confirmed the leave-running behavior is by design (so the fix targets communication, pre-flight, and data durability rather than teardown-on-quit), and found that hobby compose already ships the named-volume layout dev was missing. Alternatives rejected: per-worktree port namespacing and a teardown keybind (reasons in Changes).

A multi-agent review pass (`/simplify`, then seven area reviews via `/review-code --fix`, plus the `/writing-tests` gate for the new test) followed. It caught, and the follow-up commit fixed: a stale test assertion that would have failed CI, the herestring/TTY prompt bug, an over-broad teardown offer, the missed port-range publication format, missing zookeeper autopurge, and the sandbox gating asymmetry in phrocs.

---

### DIFF
--- a/tools/hogli-commands/hogli_commands/doctor.py
+++ b/tools/hogli-commands/hogli_commands/doctor.py
@@ -1574,18 +1574,24 @@ def _format_rss(rss_kb: int) -> str:
 # ---------------------------------------------------------------------------
 
 # The always-on infra ports from docker-compose.dev.yml whose bind failure
-# aborts `docker compose up`. Update when those mappings change; optional
-# profile services are intentionally excluded.
+# aborts `docker compose up`. Update when those mappings change; services
+# gated by `profiles:` (in docker-compose.dev.yml or the legacy
+# docker-compose.profiles.yml overlay — e.g. temporal) are intentionally
+# excluded, since they don't start by default.
 _PREFLIGHT_PORTS: tuple[tuple[int, str], ...] = (
     (8010, "proxy"),
     (2181, "zookeeper"),
     (8123, "clickhouse-http"),
+    (8443, "clickhouse-https"),
     (9000, "clickhouse-native"),
+    (9440, "clickhouse-native-tls"),
+    (9009, "clickhouse-interserver"),
     (9092, "kafka"),
     (5432, "postgres"),
     (6379, "redis"),
-    (7233, "temporal"),
+    (6399, "redis-cluster"),
     (19000, "objectstorage"),
+    (19001, "objectstorage-master"),
 )
 
 _COMPOSE_NAME_RE = re.compile(r"[^a-zA-Z0-9_.-]")
--- a/tools/hogli-commands/hogli_commands/doctor.py
+++ b/tools/hogli-commands/hogli_commands/doctor.py
@@ -5,6 +5,7 @@ import re
 import sys
 import enum
 import time
+import select
 import shutil
 import signal
 import hashlib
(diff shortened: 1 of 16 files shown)


tools/hogli-commands/hogli_commands/doctor.py
```diff
@@ -1568,6 +1569,216 @@ def _format_rss(rss_kb: int) -> str:
...
 
 
+# ---------------------------------------------------------------------------
+# doctor:ports — pre-flight port collision check
+# ---------------------------------------------------------------------------
+
+# The always-on infra ports from docker-compose.dev.yml whose bind failure
+# aborts `docker compose up`. Update when those mappings change; optional
+# profile services are intentionally excluded.
+_PREFLIGHT_PORTS: tuple[tuple[int, str], ...] = (
```
rnegron, line 1584: my bot says: `_PREFLIGHT_PORTS misses 5 always-on ports (6399, 8443, 9009, 9440, 19001) and includes 7233, which is profile-gated. Nothing guards the drift.`

tools/hogli-commands/hogli_commands/tests/test_devenv.py
```diff
@@ -978,3 +979,61 @@ def test_parse(self, raw: str, expected_excluded: list[str], expected_out_of_ran
         excluded, out_of_range = _parse_exclude_input(raw, self.UNITS)
         assert excluded == expected_excluded
         assert out_of_range == expected_out_of_range
+
+
+class TestDockerComposeShellFailurePath:
```
rnegron, line 984: my bot says: `TestDockerComposeShellFailurePath doesn't test its stated claim: plain && already blocks the false ready marker on master; the || {...} guard only adds a message. The 9000-vs-19000 case also passes with a naive substring match.`

rnegron (approved):
lgtm, thanks! left some nits but found no blockers

tools/hogli-commands/hogli_commands/doctor.py
```diff
@@ -1568,6 +1569,216 @@ def _format_rss(rss_kb: int) -> str:
...
+
+
+def _confirm_stack_teardown(stack: str, timeout: float = 30.0) -> bool:
+    """Prompt for teardown of a foreign compose stack, with a timeout.
+
+    Runs on every `hogli start`, so a hung or piped stdin must never block;
+    timeout, EOF, or any non-"y" answer means "leave it running".
+    """
+    click.echo(f"   Stop foreign stack '{stack}' now? [y/N] ", nl=False)
+    ready, _, _ = select.select([sys.stdin], [], [], timeout)
```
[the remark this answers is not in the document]
haacked: Windows isn't a supported native platform for local dev — see [developing-locally.md](https://github.com/PostHog/posthog/blob/master/docs/published/handbook/engineering/developing-locally.md#L55) ("Windows isn't supported natively"; Windows users run a Linux VM/WSL2 instead). In that environment `select.select()` on stdin works fine.

This also matches existing precedent: `bin/sandbox` already uses the identical `select.select([sys.stdin], [], [], 1)` pattern without a Windows branch.

tools/hogli-commands/hogli_commands/doctor.py
haacked: Good catch! Fixed. Pulled temporal out of the list since it's profile-gated (only starts with `COMPOSE_PROFILES=temporal`, per `docker-compose.profiles.yml`), and added the 5 missing always-on ports: redis-cluster (6399), and ClickHouse's 8443/9009/9440, plus objectstorage's second port 19001.

tools/hogli-commands/hogli_commands/tests/test_devenv.py
haacked: Good catch on the docstring. You're right that a plain `up_cmd && echo ready && logs_cmd` chain already blocks the false ready marker on failure; the `|| { fail_cmd; exit 1; }` construct's only functional difference is the friendlier error message.

I tightened the docstring to say what the test actually guards against: someone dropping the `exit 1` from `fail_cmd`, which would let the `{ ...; }` group exit 0 and let the ready marker slip through.

On the "9000-vs-19000 naive substring match" part: that's not this test class (it doesn't touch port scanning at all). It looks like it might be about `_scan_port_holders`/`_scan_unheld_via_lsof` in `test_doctor.py`, but that's already guarded there: the match includes the colon before the port number (plus the trailing dash for docker's published-range format), not just the bare digits, so `9000` can't match inside `19000`. There's a dedicated `range-does-not-false-match-shorter-port` case (`test_doctor.py:423-428`) plus `test_scan_unheld_via_lsof_disambiguates_port_and_range` covering exactly the 9000-vs-19000 collision. Let me know if you meant something else there.

tools/hogli-commands/hogli_commands/doctor.py
```diff
@@ -1568,6 +1572,562 @@ def _format_rss(rss_kb: int) -> str:
...
+    project_name = _compose_project_name()
+    # `bin/start`'s own lock file is per-worktree, but every worktree shares this
+    # compose project — without this, two worktrees starting at once could both plan
+    # a migration for the same destination volumes, and one's failure rollback would
+    # delete data the other just copied. Blocks rather than skips, so the second
+    # worktree waits out the first's migration instead of racing it; the OS releases
+    # the lock automatically if the holding process dies, so a crash can't wedge it.
+    lock_path = _MIGRATION_LOCK_DIR / f"hogli-migrate-volumes-{project_name}.lock"
+    try:
+        with open(lock_path, "w") as lock_file:
```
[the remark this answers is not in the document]
haacked: Fixed in 5b35334. The lock now opens with `os.open(lock_path, os.O_RDWR | os.O_CREAT | os.O_NOFOLLOW, 0o600)` wrapped in `os.fdopen`. The existing `except OSError` already fails open, so a squatted path just skips the migration rather than crashing startup.

Confirmed the truncation is real before fixing it: `open(path, "w")` against a symlink planted at the lock path zeroed the target file.

Two clarifications on scope, since the description reads broader than the actual exposure:

- It is Linux-only in practice. `tempfile.gettempdir()` honors `TMPDIR`, which on macOS is a private per-user directory, so there is no world-writable path to squat there. On Linux `fs.protected_symlinks=1` (the default on systemd distros) also refuses this follow, so it additionally needs a hardened-off kernel.
- It is file destruction, not content injection. The file is only ever an flock handle and is never written to, so dropping `O_TRUNC` removes the primitive on its own, independent of `O_NOFOLLOW`.

Both are in the fix anyway since it costs nothing. Added a regression test that fails against the old `open(..., "w")` and passes now.

PR #73329, opened 2026-07-23
feat(replay-vision): gate scanner create and scan on AI consent

ksvat:
## Problem

Replay Vision only enforced the organization's AI data-processing consent (`is_ai_data_processing_approved`) in one narrow place — the Vision Action digest/synthesis path. Scanner creation and the core per-recording scan checked neither that flag nor anything else, so a scanner could process session recordings with external AI providers even when the org had not opted in to AI data processing.

**Why:** ahead of launch we want using Replay Vision to require the same AI data-processing consent the rest of PostHog AI relies on, rather than silently processing recordings. AI-training opt-in (`is_ai_training_opted_in`) is intentionally *not* a gate — access shouldn't depend on it.

## Changes

- **Backend — real enforcement.** `create_observation_activity` is the single chokepoint every scan flows through (scheduled sweeps, on-demand triggers, applies). It now short-circuits to a no-op when the org hasn't approved AI data processing, mirroring the existing quota-exhausted skip. Scanner creation also fails fast in the serializer with a clear message.
- **Frontend — consent in context.** The "New scanner" and "Create your first scanner" CTAs reuse the shared `AIConsentPopoverWrapper` (the same consent flow Max uses). When consent is missing the button reads "Allow AI analysis and create scanner"; approving flows straight into the create-scanner journey, so the empty state is identical whether or not consent is set. The third-party-model-training disclaimer line is suppressed here via a new opt-out prop, since it doesn't apply to this surface.
- `is_ai_training_opted_in` is left untouched.

Added a backend regression test, `TestCreateObservationActivity::test_skips_insert_when_ai_data_processing_not_approved`, which asserts the scan activity persists no observation and returns a no-op when the org has `is_ai_data_processing_approved=False`. It catches removal/regression of the consent gate on the critical scan path — no existing test covered that (the sibling test only covers the quota skip).

I was not able to run the test suites or the app in this environment (no Python interpreter, DB, or frontend toolchain available), so I have not executed the backend test or manually exercised the UI — CI will run them. No manual testing was performed by me.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread (linked below) while scoping Replay Vision's launch behavior. The investigation established that RV gated on consent only in the digest path, not on scanner creation or the core scan; this PR closes that gap. Chose `create_observation_activity` as the enforcement point because it's the one place all scan paths converge, and reused Max's `AIConsentPopoverWrapper` rather than building a new empty/null state so the create flow stays a single click after approval. Deliberately did not gate on `is_ai_training_opted_in`. No skills were invoked (the writing-tests skill is not present in this checkout's skill set).

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B7TE4R8GH/p1784817560599059?thread_ts=1784817560.599059&cid=C0B7TE4R8GH)*

### DIFF
--- a/products/replay_vision/backend/temporal/activities/create_observation.py
+++ b/products/replay_vision/backend/temporal/activities/create_observation.py
@@ -46,9 +46,26 @@ def create_observation_activity(inputs: CreateObservationInputs) -> CreateObserv
 
 
 def _create_observation(inputs: CreateObservationInputs) -> CreateObservationOutput:
-    scanner = ReplayScanner.objects.filter(pk=inputs.scanner_id, team_id=inputs.team_id).select_related("team").first()
+    # team__organization is prefetched for the AI-consent check below.
+    scanner = (
+        ReplayScanner.objects.filter(pk=inputs.scanner_id, team_id=inputs.team_id)
+        .select_related("team", "team__organization")
+        .first()
+    )
     if scanner is None:
         raise ValueError(f"ReplayScanner {inputs.scanner_id} not found for team {inputs.team_id}")
+
+    # No AI processing of recordings without organization consent, even for scanners created earlier.
+    if not scanner.team.organization.is_ai_data_processing_approved:
+        activity.logger.info(
+            "Skipping observation: AI data processing not approved for organization",
+            extra={"scanner_id": str(inputs.scanner_id), "team_id": inputs.team_id, "session_id": inputs.session_id},
+        )
+        return CreateObservationOutput(
+            observation_id=None,
+            was_created=False,
+            scanner_type=scanner.scanner_type,
+        )
 
     if inputs.triggered_by_user_id is not None:
         # The activity is the persistence boundary, so re-check team membership rather than trusting the trigger.
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -5577,7 +5577,7 @@ snapshots:
     scenes-app-dashboards--show-without-post-hog-ai-button-label--medium--dark:
         hash: v1.k794b7964.8e451e072f54f7e2a4bf649e2af129a1624a9ce5939d023a10cef4dbe200d00a.35hPLpid_TQpN7vdWwVb9IwsPn0qt_aoDmzGmhvA0nA
     scenes-app-dashboards--show-without-post-hog-ai-button-label--medium--light:
-        hash: v1.k794b7964.16695d22ae64fd0b3f014604ca34dc06afc8329b0d122300de1a87c5f8e3434b.lghCgl4IhqFDNKpqEWy9JqNYM5WW0mnpRr-m7AQQYqY
+        hash: v1.k794b7964.7f45f600558e7bcfa730776316ddea01cd766a65abb2d4ef94147003f7f570b4.Mnn4JzYBVXBWsIK_iXsXqaOySdOgOq3oZo4Oti8MRoA
     scenes-app-dashboards--show-without-post-hog-ai-button-label--narrow--dark:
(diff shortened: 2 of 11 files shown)


products/replay_vision/backend/temporal/activities/create_observation.py
```diff
@@ -37,10 +37,26 @@ def _build_scanner_snapshot(scanner: ReplayScanner) -> dict[str, Any]:
...
+    scanner = (
+        ReplayScanner.objects.filter(pk=inputs.scanner_id, team_id=inputs.team_id)
+        .select_related("team", "team__organization")
+        .first()
+    )
     if scanner is None:
         raise ValueError(f"ReplayScanner {inputs.scanner_id} not found for team {inputs.team_id}")
 
+    # No AI processing of recordings without organization consent, even for scanners created earlier.
+    if not scanner.team.organization.is_ai_data_processing_approved:
```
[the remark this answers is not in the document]
ksvat: Good catch — fixed in be26a45. The consent check now also runs at the egress boundary: a fresh, fail-closed `is_ai_data_processing_approved` read runs inside `upload_video_to_gemini_activity` (immediately before the recording bytes stream to Gemini) and again inside `call_scanner_provider_activity` (before provider generation), since those are separate egress steps. On withdrawal the observation is marked ineligible via a new `ConsentWithdrawnError` and no Gemini client method is called. Added `TestEgressConsentRecheck` with two activity tests that revoke consent after `create_observation_activity` and assert the upload/provider abort without touching Gemini.

arnohillen (approved):
Nice!

products/replay_vision/backend/temporal/types.py
```diff
@@ -201,6 +201,7 @@ class EnsureSessionAssetOutput(BaseModel, frozen=True):
 
 class UploadVideoToGeminiInputs(BaseModel, frozen=True):
     asset_id: int
+    team_id: int  # for a fresh AI data-processing consent check immediately before recording bytes egress to Gemini
```
TueHaulund, line 204: `team_id` is a new required field on an activity input, so any scan in flight during the deploy has a payload with only `asset_id`. The new worker can't deserialize those and mid-flight scans will fail during rollout.

Suggestion: drop the field entirely and derive the team inside the activity instead. `_upload_video` already does `asset = await ExportedAsset.objects.aget(id=inputs.asset_id)`, and the asset is team-scoped, so the consent check can move to right after that line and use `asset.team_id`. That keeps the payload shape unchanged (no deploy compat issue, in-flight scans even pick up the check immediately), avoids a workflow-definition change for this step, and ties the check to the team that actually owns the bytes rather than a second copy passed from the workflow. The `create_observation` boundary already uses this pattern by reading the team off the scanner row.

ksvat: Good call — done in the latest push. Dropped `team_id` from `UploadVideoToGeminiInputs` entirely and moved the consent re-check to right after `asset = await ExportedAsset.objects.aget(id=inputs.asset_id)`, keying it off `asset.team_id`. The check still runs before any bytes are read/uploaded, the payload shape is unchanged (so in-flight scans deserialize fine and even pick up the check immediately), and it now ties to the team that owns the bytes rather than a second copy from the workflow — matching the `create_observation` pattern. Updated the activity test to create a real asset owned by the no-consent team.

TueHaulund (approved):

ksvat:
### Merge activity

* **Jul 24, 11:54 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/73329).

PR #73343, opened 2026-07-23
chore(mcp): regenerate UI app entry points

haacked:
## Problem

The committed generated MCP UI app entry points are stale relative to what `pnpm --filter=@posthog/mcp run generate:ui-apps` now emits: a formatter update changed how the generator wraps JSX, and four files on master still carry the old single-line form.

The drift is latent on master because MCP CI is path-filtered to `services/mcp` changes, but it fails the "Check generated UI apps are up to date" freshness gate on any PR whose merge ref touches `services/mcp` — currently blocking [#72990](https://github.com/PostHog/posthog/pull/72990), which only hit MCP CI because a bot pushed an OpenAPI-types commit to its branch.

## Changes

Reran the generator and committed the result. Four files, formatting only:

- `services/mcp/src/ui-apps/apps/generated/action.tsx`
- `services/mcp/src/ui-apps/apps/generated/cohort.tsx`
- `services/mcp/src/ui-apps/apps/generated/llm-costs.tsx`
- `services/mcp/src/ui-apps/apps/generated/survey.tsx`

- Ran `pnpm run generate:ui-apps` in `services/mcp` on a clean master checkout and committed the diff; the same freshness check CI runs (`git diff --exit-code` on the generated paths) now passes locally.
- No behavior change: the diff is JSX wrapping only, which is why it doesn't need a manual test pass. MCP CI on this PR runs the full check suite since it touches `services/mcp`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed: generated-file refresh, no user-facing or documented behavior change.

## 🤖 Agent context

Diagnosed with PostHog Code (Claude) while investigating the MCP CI failure on #72990: reproduced by merging master and rerunning the generator, confirmed the drift is formatting-only and predates that PR, then split this fix out so the ClickHouse migration PR stays migration-only.

---

### DIFF
--- a/services/mcp/src/ui-apps/apps/generated/action.tsx
+++ b/services/mcp/src/ui-apps/apps/generated/action.tsx
@@ -8,7 +8,11 @@ import { type ActionData, ActionView } from 'products/actions/mcp/apps'
 import { AppWrapper } from '../../components/AppWrapper'
 
 function ActionApp(): JSX.Element {
-    return <AppWrapper<ActionData> appName="PostHog Action">{({ data }) => <ActionView action={data!} />}</AppWrapper>
+    return (
+        <AppWrapper<ActionData> appName="PostHog Action">
+            {({ data }) => <ActionView action={data!} />}
+        </AppWrapper>
+    )
 }
 
 const container = document.getElementById('root')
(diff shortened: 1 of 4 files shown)


haacked:
No longer needed: #73203 touched services/mcp and regenerated these files on master about half an hour after this was opened, with identical content. Closing.

PR #73349, opened 2026-07-23
fix(flags): clean recoverable filter violations found by audit

patricio-posthog:
## Problem

Part of #50084 (phase 2, does not close it). The `audit_flag_filters` scan of both regions found 1,757 flags with filter violations ([full classification](https://github.com/PostHog/posthog/issues/50084#issuecomment-5049261441)). This migration fixes the ~1,470 where the author's intent is recoverable from the data. The rest (variant sums, duplicate variant keys, and other cases needing a policy call) are deliberately excluded and stay visible in the audit.

## Changes

One data migration, same pattern as `0007` (batched `bulk_update`, elidable, noop reverse, soft-deleted flags included, idempotent transform so it can be re-run before enforcement flips on):

| Transform | Flags (US+EU audit) | Behavior effect |
|---|---|---|
| payload string values that aren't strict JSON get `json.dumps`-encoded | 762 | none for SDKs that parse-with-fallback; strict parsers stop failing |
| `in`/`not_in` on non-cohort properties become `exact`/`is_not` | 680 | see warning below |
| operator typos (`contains`, `matches regex`, `does not equal`, `is_in`) mapped to canonical operators | 13 | see warning below |
| numeric-string `rollout_percentage` on groups coerced to float | 9 | none, evaluation already coerces |
| plain-number values stringified for string operators (`gt`, `icontains`, …) | 5 | none, evaluators compare stringified |
| dangling `groups[].variant` overrides nulled | 2 | none, override already resolved to nothing |
| re-applies the 0007 empty-`multivariate` cleanup (drift since the API still accepts it) | 1+ | none, flags already evaluate boolean |

> [!WARNING]
> Two transforms change evaluation for currently-broken flags, and I want an explicit ack from @haacked here before merging (continuing the #50084 thread in this PR):
>
> 1. **`in`/`not_in` remap (680 flags).** These conditions hard-error in the Rust matcher today (`property_matching.rs` returns `Err("In/NotIn operators should be handled by cohort matching")`), so the conditions never match. After the remap they match as written, since `exact` with a list value is the same semantics. Broken flags start working as their authors intended.
> 2. **Operator typo map (13 flags).** Flags with these operators fail per-flag serde today and are silently skipped by the loader, so they don't evaluate at all. After the fix they evaluate.
>
> Both are fixes in the direction of author intent, but they are behavior changes and deserve a deliberate yes.
>
> **Update after review**: split by operator and liveness across both regions, the blast radius is near zero. `not_in` (the risky direction, since `is_not` matches users lacking the property): 4 flags, all inactive or deleted. `in`: 679 inactive/deleted plus exactly one live flag, which is the contained direction (`exact` keeps non-holders unmatched). No live flag anywhere uses `not_in` on a non-cohort property.

Deliberately left out (still reported by the audit): `is_contained_within` (1 flag, no confident canonical mapping), one flag with a variant missing `rollout_percentage` (needs an eyeball), and everything in the needs-a-policy bucket from the issue comment.

No migration tests (house practice for data migrations). Verified locally:

- `manage.py migrate feature_flags` applies cleanly on a DB at current master
- a throwaway script ran the transform against 10 crafted shapes, one per rule plus a fully-valid flag that must come out byte-identical, and asserted idempotency (second pass is a no-op on every case)
- `ruff check` and `ruff format` clean

The migration only writes rows the transform changed, so a scan where everything is already clean writes nothing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Data-only migration, no API or documented behavior change.


products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+                    if not isinstance(prop, dict):
+                        new_properties.append(prop)
+                        continue
+                    new_prop = dict(prop)
+                    operator = new_prop.get("operator")
+
+                    # in/not_in outside cohorts hard-errors in the Rust matcher;
+                    # exact/is_not with the same value is the intended semantics.
+                    if operator in ("in", "not_in") and new_prop.get("type") != "cohort":
+                        new_prop["operator"] = "exact" if operator == "in" else "is_not"
```
haacked, line 120: `question`: Today both `in` and `not_in` on non-cohort properties hard-error in the Rust matcher and return `false` at [flag_matching.rs:1644](https://github.com/PostHog/posthog/blob/8e1231869fff4d678c10c00e12583620559eb06d/rust/feature-flags/src/flags/flag_matching.rs#L1644), so neither matches anyone. After the remap, `in` becomes `exact`, which still returns false for users missing the property ([property_matching.rs:198-204](https://github.com/PostHog/posthog/blob/8e1231869fff4d678c10c00e12583620559eb06d/rust/feature-flags/src/properties/property_matching.rs#L198-L204)). But `not_in` becomes `is_not`, which returns true when the property is absent ([property_matching.rs:220](https://github.com/PostHog/posthog/blob/8e1231869fff4d678c10c00e12583620559eb06d/rust/feature-flags/src/properties/property_matching.rs#L220)). Every `not_in` condition in this set flips from matching nobody to matching every user lacking the property, plus everyone whose value isn't in the list.

Can you split the 680-flag count by operator before merging? If the `not_in` share is large, landing `in` and `not_in` as separate migrations would let you monitor the bigger blast-radius change independently.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+
+def _is_strict_json(value: str) -> bool:
+    try:
+        json.loads(value, parse_constant=_reject_json_constant)
+        return True
+    except ValueError:
+        return False
+
+
+def _clean_filters(filters: dict) -> tuple[dict, set[str]]:
```
haacked, line 44: `suggestion`: `_clean_filters` has no tests and includes the two rules that change evaluation on live flags. The idempotency check you ran locally should ship as a test, since nothing stops a future edit from silently breaking a rule before enforcement relies on this data being clean.

```python
import importlib
from products.feature_flags.backend.filters_validation import collect_filters_violations

migration = importlib.import_module(
    "products.feature_flags.backend.migrations.0011_clean_flag_filters_recoverable_violations"
)
clean = migration._clean_filters

def test_in_not_in_non_cohort_remapped_and_passes_audit():
    dirty = {"groups": [{"properties": [
        {"key": "x", "type": "person", "operator": "in", "value": "a"}]}]}
    cleaned, rules = clean(dirty)
    assert "in_not_in_non_cohort" in rules
    assert cleaned["groups"][0]["properties"][0]["operator"] == "exact"
    assert collect_filters_violations(cleaned) == []
    assert clean(cleaned)[1] == set()  # second pass is a no-op
```

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+                for prop in properties:
+                    if not isinstance(prop, dict):
+                        new_properties.append(prop)
+                        continue
+                    new_prop = dict(prop)
+                    operator = new_prop.get("operator")
+
+                    # in/not_in outside cohorts hard-errors in the Rust matcher;
+                    # exact/is_not with the same value is the intended semantics.
+                    if operator in ("in", "not_in") and new_prop.get("type") != "cohort":
```
haacked, line 119: `suggestion`: No test covers cohort properties with `operator: "in"`, so the `new_prop.get("type") != "cohort"` guard is the only safeguard preventing this migration from breaking them by rewriting `in` to `exact`.

```python
def test_cohort_in_operator_left_untouched():
    dirty = {"groups": [{"properties": [
        {"key": "id", "type": "cohort", "operator": "in", "value": 42}]}]}
    cleaned, rules = clean(dirty)
    assert rules == set()
    assert cleaned["groups"][0]["properties"][0]["operator"] == "in"
```

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+        if "payload_not_json" in rules:
+            new_filters["payloads"] = new_payloads
+
+    # Re-apply the 0007 cleanup: malformed multivariate means the flag is boolean.
+    # The API still accepts this shape until enforcement ships, so drift reappears.
+    multivariate = new_filters.get("multivariate")
+    if multivariate is not None:
+        variants = multivariate.get("variants") if isinstance(multivariate, dict) else None
+        if not isinstance(variants, list) or len(variants) == 0:
+            new_filters["multivariate"] = None
```
haacked, line 70: `question`: When `multivariate_empty` fires, `variant_keys` recomputes to an empty set, so every group's `variant` override hits the dangling-override branch and gets nulled in the same pass. This looks intentional (a boolean flag has no variants left to reference), but nothing exercises this combination.

Did the local script cover a flag with both an empty multivariate and a group variant override, or is that worth a dedicated test case?

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+    scans all flags read-only and writes only the ~0.2% that change. Soft-deleted flags
+    included, so a restore can't resurrect a violation."""
+    FeatureFlag = apps.get_model("feature_flags", "FeatureFlag")
+
+    total = 0
+    rule_counts: dict[str, int] = {}
+    batch: list = []
+
+    # _base_manager: the default manager excludes soft-deleted flags, also inside bulk_update.
+    for flag in FeatureFlag._base_manager.exclude(filters=None).only("id", "filters").iterator(chunk_size=BATCH_SIZE):
```
haacked: `question`: This cleanup scans the whole `FeatureFlag` table instead of prefiltering like 0007 did, since the violations live in nested arrays a jsonb prefilter can't select cheaply. `.iterator()` streams in bounded memory only when server-side cursors are enabled, which the direct DB connection provides.

Does this always run on the direct DB connection in every environment, or could migrations sometimes run through pgbouncer instead? If the latter, `.iterator()` would buffer the full table into the migration process at once.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+    logger.info("cleaned_flag_filters_recoverable_violations", updated_rows=total, **rule_counts)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("feature_flags", "0010_remove_featureflag_performed_rollback_and_more"),
+    ]
+
+    operations = [
+        migrations.RunPython(clean_flag_filters_recoverable_violations, migrations.RunPython.noop, elidable=True),
```
haacked, line 192: `question`: This bundles three behavior-changing rules (`in_not_in_non_cohort`, `operator_typo`, `non_string_value`) with four no-op rules in one irreversible migration. Was bundling them together a deliberate choice to match 0007's shape, or just convenience?

The per-rule counts in the log line and the idempotent transform make either approach workable, but separating the behavior-changing rules would let them be deployed and monitored on their own.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+OPERATOR_TYPO_MAP = {
+    "contains": "icontains",
+    "matches regex": "regex",
+    "does not equal": "is_not",
+    "is_in": "exact",
+}
+
+# Operators whose value must be a string; plain numbers are stringified (the
+# evaluators compare stringified values anyway).
+STRING_VALUE_OPERATORS = frozenset({"regex", "not_regex", "icontains", "not_icontains", "gt", "gte", "lt", "lte"})
```
haacked, line 27: `nit`: `STRING_VALUE_OPERATORS` is a byte-identical copy of [filters_validation.py:32-34](https://github.com/PostHog/posthog/blob/8e1231869fff4d678c10c00e12583620559eb06d/products/feature_flags/backend/filters_validation.py#L32-L34), and the in/not_in-non-cohort check mirrors `check_operator_value_compatibility` in that file. A comment should document why they're frozen in the migration, so the next reader knows to reconcile them by hand if the canonical definitions change.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def _reject_json_constant(value: str) -> None:
```
haacked, line 30: `nit`: `_reject_json_constant` rejects these constants because `json.loads` accepts `NaN`, `Infinity`, and `-Infinity` by default, and rejecting them here makes them count as non-strict JSON. A comment explaining this purpose would help the next reader.

haacked (approved):
Nice, thorough migration. One question about the `not_in` remap's blast radius before you finalize the ack, plus a few test-coverage suggestions inline.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Ran the split on both regions: `not_in` on non-cohort is 4 flags total, all inactive or deleted. `in` is 679 inactive/deleted plus exactly one live flag (EU), and that one is the safe direction (`exact` keeps non-holders unmatched). So the asymmetry you flagged is real but has nothing live to hit: no live flag anywhere uses `not_in` on a non-cohort property. Kept a single migration on that basis — the remap on dead flags only changes what they would do if someone reactivates them, which is the intent.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Done: `test_migration_0011_clean_filters.py`, the local script's shapes as parameterized cases plus your `collect_filters_violations` assertion on every fully-recoverable case and an idempotency pass over all of them. Your suggested assertion caught a bad fixture of mine on the first run, so it's already paying rent.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Added, `test_cohort_in_operator_left_untouched`.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Yes, the local script covered exactly that combination; it's now the `empty_multivariate_nulled_with_override_and_payload_keys` case. The nulling is intentional: once `multivariate` is gone the flag is boolean and no override can resolve.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Good catch. `DISABLE_SERVER_SIDE_CURSORS` is wired to `USING_PGBOUNCER` in `data_stores.py`, so `.iterator()` streaming depends on the environment. Rather than depend on that, the scan now uses id-keyset batches (`id > last ORDER BY id LIMIT 500`), memory-bounded regardless of pooling.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Started as convenience matching 0007's shape, but the operator split turned it into a decision: with zero live `not_in` flags and one live `in` flag there's nothing to monitor separately, so one migration stands. If the pre-enforcement re-run ever finds live `not_in` rows, that re-run is a separate migration anyway and can carve them out.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Added the comment.

products/feature_flags/backend/migrations/0011_clean_flag_filters_recoverable_violations.py
patricio-posthog: Added the comment.

PR #73353, opened 2026-07-23
fix(cohorts): reset negation when switching criterion type

lukebaber-posthog:
## Problem

In the cohort criteria builder, negated criteria are stored as their positive counterpart plus a `negation: true` flag (e.g. "Do not have the property" is stored as `{value: 'have_property', negation: true}` — see `determineFilterType`). The criterion-type dropdown only emitted `{value}` on selection, which `cohortFieldLogic` merges onto the existing criteria — so a stale `negation: true` always survived the merge, and `criteriaToBehavioralFilterType` derives the row type negation-first.

Net effect: a row currently on a negated criterion ("Do not have the property", "Did not complete event", "Not in cohort", "Did not complete a sequence of events") could never be switched to a positive one. Picking "Have the property" on such a row was a silent no-op — and picking it from a different negated criterion type landed on "Do not have the property" instead — leaving users stuck with an inverted filter they didn't choose. Reported via a customer support escalation.

## Changes

- When the criterion-type picker (`fieldKey === 'value'`) emits a selection, `CohortSelectorField` now includes the `negation` implied by the picked option (derived via `determineFilterType`), so a stale flag can't survive the merge.
- Positive → negative switches already worked and are unchanged.

- Added a parameterized component test in `CohortEdit.test.tsx` that clicks through the real dropdown flow on a negated row and asserts the resulting criteria state + label for: switching to "Have the property" (the reported regression), switching to "Completed event" (same trapdoor, different criterion family), and switching to "Do not have the property" (guards that explicit negative picks still negate). No existing test exercised criterion-type switching through the click → merge → clean chain; logic tests dispatch `setCriteria` payloads directly and bypass the component-constructed payload where the bug lived.
- Ran the full `frontend/src/scenes/cohorts` jest suite: 8 suites, 170 tests, all passing.
- I (Luke Baber) spun up a local version, seeded data and clicked through the action creation process to validate that the change fixed the issue.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

n/a — bug fix restoring documented behavior.

## 🤖 Agent context

- Authored with PostHog Code (Claude Code) from a support report of person-property criteria stuck on "Do not have the property". Reproduced the trapdoor deterministically (util-composition level and real component clicks) before fixing.
- An alternative fix in `cleanCriteria` (treat the purge-path `value` as user intent) was rejected: `cohortFieldLogic`'s `propsChanged` re-dispatches the type field's `onChange` with the full stored criteria — where a negated row's `value` equals the positive enum — so that approach would spontaneously un-negate rows on unrelated edits. The click site is the only place the pick is unambiguous.
- Skills invoked: /writing-tests.

---


lukebaber-posthog:
**CI triage note (agent):** the two failing check groups are unrelated to this PR's diff:

- **semgrep-general**: 7 blocking `unsafe-add-mask-workflow-command` findings, all in `.github/workflows/*.yml` — files this PR doesn't touch. Reproduced identically on a re-run, and other PRs in the same window fail the same way while master passes. Repo-level, not addressable from this branch.
- **Visual review**: first run flagged 3 changed stories, re-run flagged 2 — none related to the two changed cohort files (all cohort stories are unchanged). Verified by pulling the images: `scenes-app-insights-error-empty-states--long-loading` (dark+light) is a timing flake (the loading-state optimization hint box appears mid-animation, plus a random query ID rendered in the screenshot; its history shows the same story flipping on master), and the auth-domains story captured the app home screen instead of the settings scene on the first run (routing race) and self-resolved on re-capture. The two `long-loading` variants look like tolerate/quarantine candidates in [Visual Review](https://us.posthog.com/project/2/visual_review/runs/86a2509b-da9d-4cc0-af28-aa24da8934fa) — leaving that decision to a human.

No code changes were needed; nothing new committed.

gustavohstrassburger (approved):

PR #73356, opened 2026-07-23
fix(cohorts): clickhouse client CA setup

matheus-vb:
## Problem

cohort-seeder cannot reach ClickHouse over TLS in any environment. The workspace pins the `clickhouse` crate with the `rustls-tls` feature, which resolves to `webpki-roots`, the bundled Mozilla public root set. PostHog serves ClickHouse certificates from a per environment internal CA, so the chain can never validate, and because the feature is `webpki-roots` rather than native roots the crate ignores the container trust store entirely. Mounting a CA bundle would change nothing. 

The config also had no `clickhouse_verify` field, so the chart's `CLICKHOUSE_VERIFY` was silently discarded by envconfig and there was no way to express TLS on with verification off, which is the posture every Python ClickHouse consumer already runs. The only way to get dev working was pinning a single ClickHouse node over plaintext 8123, which sends the real ClickHouse password across the VPC in cleartext on every request.

## Changes

Add `clickhouse_verify` to `Config` (env `CLICKHOUSE_VERIFY`, default `true` so an unconfigured deployment fails closed, mirroring Django's default) and build the client through the crate's `with_http_client` escape hatch when verification is off. The new `unverified_tls_client` supplies a hyper-rustls connector whose `rustls::ClientConfig` uses an accept any `ServerCertVerifier`. Handshake signatures are still checked by the crypto provider, so only chain and name validation is skipped and the wire stays encrypted.

Connector settings are copied from the crate's own default HTTP client, so the two paths differ in certificate verification and nothing else. When verification is on the code still takes the untouched `Client::default()` path. `build_client` now returns `ClickHouseClientError`, covering both a join algorithm typo and a TLS config failure. The workspace gains `hyper-rustls` and `rustls`, both with `default-features = false` matching how `clickhouse` already pins them, so the lockfile picks up no new crates and no feature changes for other consumers.

`cargo test -p cohort-seeder --lib` (57 tests) and `cargo clippy -p cohort-seeder --all-targets` are green. I also ran a throwaway local probe, since deleted, against a self signed HTTPS stub server: with `CLICKHOUSE_VERIFY=true` the connection fails with `InvalidCertificate`, which reproduces the production blocker, with `CLICKHOUSE_VERIFY=false` a query completes end to end over TLS, and plain HTTP still works through the same connector so local dev and `secure: false` are unaffected.

Kept one unit test that builds a client in both verify modes. It catches a rustls provider or protocol version misconfiguration in the dangerous verifier path, which no existing test covered and which would otherwise only surface at deploy time.

## Docs update

No.


rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -115,11 +126,112 @@ impl FromStr for ClickHouseJoinAlgorithm {
...
+
+/// A client whose TLS stack skips certificate validation. The crate's own `rustls-tls` feature
+/// hardwires the bundled Mozilla public roots (`webpki-roots`) and ignores the container trust
+/// store, so `with_http_client` is the only way to express verify-off.
+fn unverified_tls_client() -> Result<clickhouse::Client, rustls::Error> {
+    let provider = Arc::new(aws_lc_rs::default_provider());
+    let tls_config = ClientConfig::builder_with_provider(provider.clone())
+        .with_safe_default_protocol_versions()?
+        .dangerous()
+        .with_custom_certificate_verifier(Arc::new(AcceptAnyServerCert(provider)))
```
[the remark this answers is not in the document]
matheus-vb: Addressed in [8c39ed5](https://github.com/PostHog/posthog/pull/73356/commits/8c39ed5cbe8c15e3411421fc2e525d86405830f3).

rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -115,11 +127,152 @@ impl FromStr for ClickHouseJoinAlgorithm {
...
+    )
+}
+
+pub fn build_client(config: &Config) -> Result<clickhouse::Client, ClickHouseClientError> {
     let join_algorithm = config
         .seeder_ch_join_algorithm
         .parse::<ClickHouseJoinAlgorithm>()?;
-    Ok(clickhouse::Client::default()
+    // Ordered most to least trusted.
+    let client = if !config.clickhouse_ca.is_empty() {
```
haacked, line 271: `suggestion`: The TLS posture (`clickhouse_ca` / `clickhouse_verify`) and the endpoint scheme (`clickhouse_secure`) are chosen independently: [`ClickHouseEndpoint::resolve`](https://github.com/posthog/posthog/blob/90d53fdf42f2a3233878c88a2eef1234abed7883/rust/cohort-seeder/src/clickhouse/client.rs#L50-L60) picks `http://` vs `https://` from `clickhouse_secure` alone, and `build_client` picks the TLS stack from `clickhouse_ca`/`clickhouse_verify` alone (lines 268-274). If an operator sets `CLICKHOUSE_CA` or `CLICKHOUSE_VERIFY=false` but leaves `CLICKHOUSE_SECURE` at its default `false`, `build_client` still constructs a full TLS-capable connector, but [`.https_or_http()`](https://github.com/posthog/posthog/blob/90d53fdf42f2a3233878c88a2eef1234abed7883/rust/cohort-seeder/src/clickhouse/client.rs#L252) passes the resulting `http://` URL straight through with no TLS.

The ClickHouse password crosses the VPC in cleartext, exactly the risk this PR sets out to close, reachable by a single misconfigured env var.

Fail startup when a TLS posture is requested without `clickhouse_secure`:
```rust
let join_algorithm = config
    .seeder_ch_join_algorithm
    .parse::<ClickHouseJoinAlgorithm>()?;
if !config.clickhouse_secure && (!config.clickhouse_ca.is_empty() || !config.clickhouse_verify) {
    return Err(ClickHouseClientError::TlsPostureWithoutSecure);
}
```
with a new variant:
```rust
#[error("CLICKHOUSE_CA or CLICKHOUSE_VERIFY=false was set but CLICKHOUSE_SECURE is false, so the connection would be plaintext")]
TlsPostureWithoutSecure,
```

rust/cohort-seeder/src/config.rs
```diff
@@ -135,6 +135,17 @@ pub struct Config {
     #[envconfig(default = "false")]
     pub clickhouse_secure: bool,
 
+    /// Validate the server certificate against the public root CAs. Defaults on so an unconfigured
+    /// deployment fails closed; the chart sets it to `false` for internal ClickHouse.
```
haacked, line 139: `suggestion`: `clickhouse_verify` defaults to `true`, the fail-closed posture this PR relies on, but nothing tests that it stays `true`. Flipping the envconfig default to `false` would still pass every existing test.

Add an assertion next to the other `default_config()` checks in the config tests:
```rust
assert!(default_config().clickhouse_verify);
assert!(default_config().clickhouse_ca.is_empty());
```

rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -115,11 +127,152 @@ impl FromStr for ClickHouseJoinAlgorithm {
...
+    }
+
+    fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+        self.0.signature_verification_algorithms.supported_schemes()
+    }
+}
+
+/// Validates against `ca_path` alone, hostname verification included. The only mode that
+/// authenticates the server.
+fn pinned_ca_tls_config(ca_path: &str) -> Result<ClientConfig, ClickHouseClientError> {
```
haacked, line 212: `suggestion`: `pinned_ca_tls_config` is the only branch that actually authenticates the ClickHouse server, but its success path has no test. `an_unusable_ca_bundle_fails_startup_rather_than_downgrading` only covers unreadable and empty-bundle errors; it doesn't test a valid PEM or assert that `build_client` succeeds.

Mirror the existing CA-bundle test with a third case that writes a real self-signed cert PEM to the tempdir and asserts `build_client` returns `Ok` with `clickhouse_ca` set to it. A static PEM constant embedded in the test avoids needing a cert-gen dependency.

rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -222,9 +375,54 @@ mod tests {
...
-            Some(UnknownJoinAlgorithm("grace_hashh".to_string()))
-        );
+        assert!(matches!(
+            build_client(&config),
+            Err(ClickHouseClientError::JoinAlgorithm(UnknownJoinAlgorithm(algorithm)))
+                if algorithm == "grace_hashh"
+        ));
+    }
+
+    #[test]
```
haacked, line 418: `suggestion`: `build_client_assembles_a_tls_stack_for_both_verify_modes` only asserts `.is_ok()`, so it can't catch regressions in cert verification, what this PR changes. If `AcceptAnyServerCert::verify_server_cert` were flipped to reject, or verify-off were wired to authenticate instead, the test would still pass. The name also overstates coverage: with `verify=true` the code takes the untouched `clickhouse::Client::default()` path and never touches `client_with_tls_config`/`unverified_tls_config`, and the `config.clickhouse_secure = true` line is inert since branch selection ignores it.

Rename the test to reflect what it covers (e.g. `unverified_tls_config_builds_successfully`), drop the inert `clickhouse_secure` assignment, and consider a test standing up a self-signed HTTPS stub asserting verify-off completes a request while verify-on returns `InvalidCertificate`.

rust/cohort-seeder/src/clickhouse/client.rs
haacked, line 212: `question`: The stated problem is fully solved by the verify-off path, but the diff also ships a full pinned-CA tier that isn't used yet: `clickhouse_ca`, `pinned_ca_tls_config`, the `CaBundleUnreadable`/`CaBundleEmpty` variants, and a dedicated test. No cohort-seeder chart sets `CLICKHOUSE_CA` today, and the PR's Problem section says a CA bundle would change nothing.

Is this config-surface groundwork ahead of an eventual chart, or could it land in a follow-up PR alongside the chart that needs it?

rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -115,11 +127,152 @@ impl FromStr for ClickHouseJoinAlgorithm {
...
+    fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+        self.0.signature_verification_algorithms.supported_schemes()
+    }
+}
+
+/// Validates against `ca_path` alone, hostname verification included. The only mode that
+/// authenticates the server.
+fn pinned_ca_tls_config(ca_path: &str) -> Result<ClientConfig, ClickHouseClientError> {
+    let mut roots = rustls::RootCertStore::empty();
+    for certificate in CertificateDer::pem_file_iter(ca_path).map_err(|source| {
```
haacked: `nit`: `CaBundleUnreadable` is constructed twice in `pinned_ca_tls_config` with the same `path: ca_path.to_string()`, once for the iterator error and once for each certificate error.

Extract a shared closure:
```rust
let unreadable = |source| ClickHouseClientError::CaBundleUnreadable {
    path: ca_path.to_string(),
    source,
};
for certificate in CertificateDer::pem_file_iter(ca_path).map_err(unreadable)? {
    roots.add(certificate.map_err(unreadable)?)?;
}
```

rust/cohort-seeder/src/clickhouse/client.rs
```diff
@@ -115,11 +127,152 @@ impl FromStr for ClickHouseJoinAlgorithm {
...
+    CaBundleUnreadable {
+        path: String,
+        #[source]
+        source: pem::Error,
+    },
+    #[error("the ClickHouse CA bundle at {path} contains no certificates")]
+    CaBundleEmpty { path: String },
+}
+
+// The `clickhouse` crate's own defaults; the idle timeout must stay under ClickHouse's server-side
```
haacked: `nit`: `TCP_KEEPALIVE`, `POOL_IDLE_TIMEOUT`, and `enforce_http(false)` are hand-copied from the `clickhouse` crate's own default HTTP client, so the custom paths differ from `Client::default()` only in certificate verification. That equivalence holds only until the crate changes its own defaults, at which point `Client::default()` picks up the new values while the custom paths silently keep the old ones.

Add a one-line comment noting which `clickhouse` crate version these were copied from, to give the next person a way to check for drift on upgrade.

haacked (approved):
Nice fix! One inline suggestion worth a look on a config edge case, plus a couple of smaller notes.

rust/cohort-seeder/src/clickhouse/client.rs
matheus-vb: Real gap, but I keyed the guard on the resolved endpoint rather than `clickhouse_secure`. `resolve_endpoint` returns `clickhouse_url` verbatim, so `CLICKHOUSE_URL=https://…` with `CLICKHOUSE_SECURE` at its default `false` is genuine TLS, and the suggested check would reject it at startup (`explicit_clickhouse_urls_and_ports_are_preserved` covers exactly that shape). Added `ClickHouseEndpoint::is_tls()` and a `CaBundleWithoutTls` variant.

I left out the `!clickhouse_verify` half. `CLICKHOUSE_VERIFY=false` is the chart harness default for every app, not an operator opting into TLS, and `values.dev.yaml` pairs it with `secure: "false"` today, so that half would crash dev on merge. The CA case is unambiguous and nothing sets `CLICKHOUSE_CA` yet, so it fails closed with no blast radius.

rust/cohort-seeder/src/clickhouse/client.rs
matheus-vb: Renamed to `unverified_tls_config_assembles` and dropped the inert `clickhouse_secure` line. Both correct.

Instead of the HTTPS stub I added `the_verify_off_verifier_accepts_an_untrusted_certificate`, which calls `verify_server_cert` directly with non certificate bytes. Catches the same regression (flip the verifier to reject and it fails) without a TLS listener and a server key in the repo. I did run the stub locally: verify off connects, pinned CA connects, unrelated CA and wrong hostname both rejected.

rust/cohort-seeder/src/clickhouse/client.rs
matheus-vb: Groundwork, but prompted by the veria-ai comment on this PR asking for exactly this. It was cheap to add here because the PR already builds the `ClientConfig` by hand, so a pinned CA is just a `RootCertStore`, with no workspace feature change.

A CA bundle changing nothing was only true for the crate's default connector. Django's charts already set `CLICKHOUSE_CA` from `ee/certs/` (but it seems stale), so the pattern exists; what is missing is the CA reaching the pod, which is a charts change.

PR #73367, opened 2026-07-23
feat(aio): honest per-tool spend attribution and cost-by-context-size

pauldambra:
## Problem

The personal LLM spend endpoint (`/api/llm_analytics/@me/spend/`) attributes each generation's full cost to every tool in `$ai_tools_called`. Because that field is a comma-separated list and a single turn often calls several tools, the per-tool costs sum to well above the real total (around 122% in practice), and the most ubiquitous tool looks dominant purely by co-occurrence. Reading the `by_tool` breakdown, `Bash` appeared to be roughly 60% of spend when in reality it was just present on the most turns. The genuine cost driver, how large each turn's input context is, was not exposed at all.

## Changes

Additive changes to `products/ai_observability/backend/api/personal_spend.py` and its response schema.

- `by_tool` rows gain two fields:
  - `cost_attributed_usd`: each generation's cost split evenly across the distinct tools it called, so the column reconciles to the scoped generation cost instead of overcounting.
  - `share_attributed`: that value normalized to `summary.scoped_cost_usd`.
  - `cost_usd` and `share_of_scoped` stay, but their help text now explains they are co-occurrence measures that overstate ubiquitous tools and do not reconcile across rows, and points readers to the attributed fields for an honest number.
- New `by_input_size` breakdown groups spend into input-token buckets (`<50k`, `50k-100k`, `100k-150k`, `150k-200k`, `200k-300k`, `>300k`). This is the primary lever: large-context turns dominate spend regardless of which tool they call.
- `by_tool.tool` help text now notes that every skill invocation collapses to the single `Skill` token, because skill names are not recorded at ingestion, so that row is an upper bound across all skills combined rather than any one skill.
- Removed the deprecated `top_traces` field (and its serializers). It always returned empty and the PostHog Code client had already dropped it from its types.
- Regenerated OpenAPI and MCP types from the serializer changes.

The change is backwards compatible. The only removal is the always-empty `top_traces`. A companion PostHog Code PR (PostHog/code#3772) adopts `cost_attributed_usd`, `share_attributed`, and `by_input_size` as optional fields and rewrites the usage display and suggestion copy off the misleading "drives X%" line.

Automated tests added to `test_personal_spend.py`, each covering a regression the existing suite did not:

- `test_by_tool_cost_attributed_splits_evenly_across_distinct_tools`: a multi-tool generation splits its cost across distinct tools so `cost_attributed_usd` reconciles, rather than assigning full cost to each tool.
- `test_by_tool_cost_attributed_falls_back_to_null_tool_for_no_tools`: a generation with no tools attributes its full cost to the null-tool bucket.
- `test_by_input_size_buckets_generations_by_input_tokens`: generations land in the correct input-size bucket and bucket costs sum to `scoped_cost_usd`.
- `test_top_traces_removed_from_response`: the response no longer carries `top_traces`.

What I (the agent) verified locally: `ruff check` and `ruff format` clean via `hogli ci:preflight` (zero failures), and repo-wide `mypy` passes the way CI runs it (`uv run mypy --cache-fine-grained .`, "Success: no issues found in 16775 source files"). What I did not do: I have not run the backend test suite locally (it needs the Postgres/ClickHouse dev stack) and I have not exercised the endpoint by hand. CI runs the tests.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover this internal endpoint. The developer-facing OpenAPI help text is updated in this PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) - Paul directed the work and is assigned as DRI.

The change grew out of an investigation into why one engineer's PostHog Code spend was high. Querying the production `$ai_generation` data (US, team 2) showed the `by_tool` breakdown was misleading: it double-counts multi-tool turns, so the most-used tool looked like the cost driver. The real signal was input-context size, roughly half the spend came from turns carrying more than 300k input tokens. That motivated both the fractional-attribution fix and the new context-size breakdown.

The backend change was produced by a Claude Sonnet subagent directed to follow the `/improving-drf-endpoints` and `/writing-tests` conventions (schema annotations on every field, parameterized behavior-level tests). The contract (field names, bucket edges, additive-only, remove `top_traces`) was fixed up front so this PR and the companion client PR agree. `top_traces` removal was checked against the PostHog Code client first to confirm nothing rendered it.

> [!NOTE]
> The branch is a few commits behind master, and master has since touched the OpenAPI generation inputs. A rebase and `hogli build:openapi` regeneration may be needed before this is marked ready, if CI flags generated-type drift.

---


products/ai_observability/backend/api/personal_spend.py
```diff
@@ -706,6 +830,83 @@ def _fetch_by_model(
...
 
+def _input_size_bucket_case_sql() -> str:
+    """Builds the `multiIf(...)` branches from `INPUT_SIZE_BUCKETS`: each bucket (other
+    than the last) becomes `$ai_input_tokens < <next bucket's min>, '<label>'`, checked in
+    ascending order so `multiIf` matches the first (lowest) bucket the value clears; the
+    last bucket is the unconditional fallback. Bucket bounds are fixed constants defined
+    in this module, not request input, so interpolating them into the query text is safe.
+    """
+    branches = [
+        f"toFloat(properties.$ai_input_tokens) < {INPUT_SIZE_BUCKETS[i + 1][1]}, '{label}'"
```
pauldambra: > [!NOTE]

**[router]** 🟢 LOW

`multiIf` casts `$ai_input_tokens` via `toFloat(...)`, which yields NULL for a missing or unparseable value. `NULL < 50000` isn't true, so a null-token generation falls through every branch into the `>300k` fallback bucket — inflating the exact bucket this endpoint presents as the headline "cost driver." Elsewhere in this file NULLs simply drop out of `sum`/`avg`. Low probability (posthog_code generations reliably report input tokens), but a `coalesce(toFloat(properties.$ai_input_tokens), 0)` would land unknowns in `<50k` instead of the top bucket.

pauldambra (commented):
QA Swarm review complete. See inline comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers as warranted

## Verdict: 💬 APPROVE WITH NITS <sub>(round 1 @ a00ad9f)</sub>

Additive, read-only analytics change. The router hand-traced the fractional cost-attribution SQL and the input-size bucket boundaries and found them correct; one low-severity NULL-handling nit in the bucket `multiIf`. Danger LOW, confidence HIGH, so no escalation to a stronger model was warranted.

### Key findings

- 🟢 **LOW** — `personal_spend.py:841` `_input_size_bucket_case_sql()`: a NULL `$ai_input_tokens` falls through every `multiIf` branch into the `>300k` fallback bucket (the headline cost driver) rather than dropping out like NULLs do elsewhere in the file. Suggest a `coalesce(..., 0)` guard.

### Convergence

None — single-reviewer (router) pass.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router (sonnet) | Additive read-only endpoint, danger LOW / confidence HIGH. Verified attribution reconciles (no-tools/single/duplicate), buckets non-overlapping and gap-free, `top_traces` removal clean, team scoping inherited, no SQL injection (only fixed module constants interpolated). Delegated nothing. |

---

products/ai_observability/backend/api/personal_spend.py
```diff
@@ -603,6 +671,60 @@ def _fetch_by_product(
...
+    product: str,
+) -> dict[str | None, float]:
+    """Splits each generation's cost evenly across the DISTINCT tools it called,
+    so the sum across tools reconciles to the scoped generation cost instead of
+    overcounting co-occurring tools. `$ai_tools_called` can list the same tool more
+    than once (tools are recorded in call order, e.g. "Bash,Read,Bash"); `arrayDistinct`
+    collapses those repeats before dividing, since a tool called twice in one
+    generation still only "caused" that generation's cost once. Generations with no
+    tools attribute their full cost to the same NULL bucket the unattributed
+    `_fetch_by_tool` rows use. Not truncated by `limit` -- callers merge by `tool`
```
carlos-marchal-ph, line 716: This is the bit that isn't true. The query below picks up HogQL's default `LIMIT 100`, so it is truncated, just not by anything you passed it.

products/ai_observability/backend/api/personal_spend.py
```diff
@@ -603,6 +671,60 @@ def _fetch_by_product(
...
+                    arrayDistinct(arrayFilter(x -> x != '', splitByChar(',', coalesce(properties.$ai_tools_called, '')))) AS tools,
+                    toFloat(properties.$ai_total_cost_usd) AS cost
+                FROM events
+                WHERE equals(event, '$ai_generation')
+                    AND {product_filter}
+                    AND {email_filter}
+                    AND {timestamp_filter}
+            )
+        )
+        GROUP BY tool
```
carlos-marchal-ph, line 740: There's no `LIMIT` or `ORDER BY` here, so `execute_hogql_query` applies its default of 100 rows (`DEFAULT_RETURNED_ROWS` in `posthog/hogql/constants.py`). The new snapshot shows it landing as `GROUP BY tool LIMIT 100`. Any tool past the cap then falls through to the `0.0` default on line 784 and renders as `cost_attributed_usd: 0.00`, `share_attributed: 0.0`, which is exactly the number this PR is asking clients to trust in place of `cost_usd`. With no `ORDER BY` the surviving 100 are hash order, so which tools lose their attribution also shifts between refreshes.

I checked and there are already users well past 100 distinct tools in a 30 day window, so this is live rather than theoretical, and it gets worse as people add MCP servers.

Could the attribution be another aggregate on the existing `_fetch_by_tool` query rather than a second query? Wrap `tools` in `arrayDistinct` in an inner subquery, then select both `sum(cost)` and `sum(cost / length(tools))` off the single `arrayJoin`, keeping the existing `ORDER BY cost_usd DESC LIMIT {limit + 1}`. That removes the row set mismatch entirely, saves a second scan of the same events on every uncached request, and fixes the dedupe asymmetry below in the same stroke.

products/ai_observability/backend/api/personal_spend.py
```diff
@@ -603,6 +671,60 @@ def _fetch_by_product(
...
+        SELECT
+            tool,
+            round(sum(cost_share), 6) AS cost_attributed_usd
+        FROM (
+            SELECT
+                nullIf(arrayJoin(if(empty(tools), [''], tools)), '') AS tool,
+                cost / length(if(empty(tools), [''], tools)) AS cost_share
+            FROM (
+                SELECT
+                    arrayDistinct(arrayFilter(x -> x != '', splitByChar(',', coalesce(properties.$ai_tools_called, '')))) AS tools,
```
carlos-marchal-ph, line 731: This query dedupes, but `_fetch_by_tool`'s on line 743 is still a plain `splitByChar` with no `arrayDistinct`. `extractToolCallNames` records names in call order and never dedupes, so `"Bash,Read,Bash"` is an ordinary value, as your docstring above says. That means one generation is counted twice on the Bash row: `generation_count` goes up by two and the full cost is added twice.

So `generation_count`'s help text ("Number of `$ai_generation` events whose tool list includes this tool") isn't what the query returns, and the rewritten `cost_usd` text puts the whole overstatement down to multi-tool co-occurrence when a real chunk of it is a single generation charged repeatedly to the same tool. The tools that loop most within a turn come out worst, which is unfortunate given `cost_usd` is still the most prominent number on the row. Folding the two queries as suggested above fixes this too.

products/ai_observability/backend/api/personal_spend.py
```diff
@@ -252,20 +276,39 @@ class _ToolBreakdownRowSerializer(serializers.Serializer):
         help_text=(
             "Sum of `$ai_total_cost_usd` for generations whose tool list includes this tool. Multi-tool "
             "generations contribute their full cost to every tool they invoked, so this sum can exceed "
-            "`summary.scoped_cost_usd`. Prefer `share_of_scoped` for headline percentages — it's computed "
-            "per row and doesn't require the totals to reconcile."
+            "`summary.scoped_cost_usd` and rows can't be added together. It measures co-occurrence, not "
+            "cost caused by the tool. Use `cost_attributed_usd` for a number that reconciles."
```
carlos-marchal-ph, line 282: While you're rewriting these, `scoped_cost_usd` on line 418 still says it "matches the cost summed across `by_tool` / `by_model`", which is the claim this PR exists to retract. These strings ship into the MCP tool schema, so an agent reads both and has no way to reconcile them. Worth narrowing it to `by_model` and `by_input_size`, which do reconcile, and pointing at `cost_attributed_usd` for tools.

products/ai_observability/backend/api/personal_spend.py
```diff
@@ -70,12 +70,31 @@
...
 MIN_LIMIT = 1
 MAX_LIMIT = 200
 DEFAULT_LIMIT = 50
 
 CACHE_TIMEOUT_SECONDS = 300
 
+# Namespaces the cache by response schema. Bump on any payload shape change so entries
+# written by an older process during a rolling deploy are never served against the new
+# contract (this revision added `by_input_size`, `cost_attributed_usd`, `share_attributed`
+# and removed `top_traces`).
```
carlos-marchal-ph, line 97: This parenthetical records what this particular revision changed, so it's stale the next time anyone bumps the version. The two sentences above already carry the why.

products/ai_observability/backend/api/test/test_personal_spend.py
```diff
@@ -356,16 +357,97 @@ def test_by_tool_splits_comma_separated_tools(self) -> None:
...
-        # `top_traces` is deprecated — kept in the response shape so existing consumers don't
-        # crash, but always returns empty. Trace IDs are opaque strings that aren't actionable
-        # in the UI.
+    def test_by_tool_cost_attributed_splits_evenly_across_distinct_tools(self) -> None:
+        # Same fixture as test_by_tool_splits_comma_separated_tools: a multi-tool generation
+        # plus two single-tool generations. Unlike `cost_usd` (which double-counts the
+        # multi-tool generation into both rows), `cost_attributed_usd` must split its $2.0
+        # evenly across Bash and Read so the per-tool sum reconciles to scoped_cost_usd
+        # instead of overcounting -- this is the regression the whole PR exists to fix.
+        self._create_generation(tool="Bash,Read", cost=2.0, trace_id="multi")
```
carlos-marchal-ph, line 368: This fixture is `Bash,Read`, where `arrayDistinct` is a no-op, so nothing in the suite exercises the repeat case the docstring was written for. Delete `arrayDistinct` from the query and everything still passes. One more generation with `tool="Bash,Read,Bash"`, asserting Bash gets half rather than two thirds, would pin it down.

products/ai_observability/backend/api/test/test_personal_spend.py
```diff
@@ -356,16 +357,97 @@ def test_by_tool_splits_comma_separated_tools(self) -> None:
...
+        assert [r["bucket"] for r in body["by_input_size"]["items"]] == ["<50k", "50k-100k", ">300k"]
+
+        # cost_usd across buckets reconciles to scoped_cost_usd.
+        scoped = body["summary"]["scoped_cost_usd"]
+        assert scoped == 7.5
+        assert sum(r["cost_usd"] for r in body["by_input_size"]["items"]) == scoped
+        assert sum(r["share_of_scoped"] for r in body["by_input_size"]["items"]) == 1.0
+        assert by_bucket["<50k"]["avg_cost_per_generation"] == 0.75
+
+    def test_top_traces_removed_from_response(self) -> None:
```
carlos-marchal-ph, line 444: `test_empty_result_when_no_events` already asserts `top_traces` is absent, so this creates an event to re-assert the same thing. Is it earning its place?

carlos-marchal-ph (approved):
@pauldambra sorry it took us so long to review this, the PR got kinda lost on the interwebs 🙏  Approving to unblock you, but I think some of the comments I added should be addressed.

PR #69077, opened 2026-07-07
feat(managed-migrations): IAM role auth option in the migration form

nickbest-ph:
## Problem

The backend for IAM role auth on S3 imports is in #73380. This adds the UI, so customers can actually use a role instead of pasting long-lived AWS access keys into the migration form.

Stacked on #73380. The worker side is already on master.

## Changes

- The S3 and S3-gzip forms get an authentication choice: IAM role (recommended) or access keys. The choice only appears when the server says role auth is available for this project, which requires the feature flag and the deployment to have an import role configured. Everyone else sees the exact same keys-only form as today.
- Picking IAM role shows a setup guide with three steps: create a role with the shown trust policy (the external ID and PostHog role ARN come from the server), attach the shown permission policy (the bucket and prefix you typed into the form are filled in live, so what you paste into AWS matches what the import will read), and paste the resulting role ARN.
- In role mode the access key, secret key, and endpoint URL fields are hidden and left out of the request. Role auth only works with real AWS S3, so the custom endpoint option doesn't apply.
- The form starts in access keys mode and switches to IAM role once the server confirms it's available, unless you already started typing keys. Validation follows the chosen mode, so you're never asked for a field the form is hiding.

No screenshots yet since I haven't run the UI against a deployment with the role configured.

## How did you test this code?

- Frontend typecheck passes for managed_migrations (the repo has a handful of pre-existing stale typegen errors in unrelated scenes).
- No new Jest tests. The form branches are thin conditionals over form state, and the serializer tests in #73380 already lock in the request contract (role or keys but not both, no custom endpoint with a role). A frontend test would restate the same rules one layer up without catching a distinct regression.
- Not done: a manual browser run against a deployment with the role ARN configured.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

A setup walkthrough for posthog.com will follow before this ships to customers.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Authored with Claude Code. Skills invoked: /adopting-generated-api-types (the new endpoint is called through the generated client rather than a handwritten request) and /writing-tests (concluded no new frontend tests earn their place, reasoning above).

Decisions: the form defaults to access keys and only promotes itself to role auth after the availability check succeeds, so the validator can never demand a role ARN the UI is hiding. A later review session rebased this branch onto the updated #73380 and dropped its duplicate generated-types changes, since the backend branch now carries freshly regenerated types.

PR #73380, opened 2026-07-23
feat(managed-migrations): add IAM role auth for S3 batch imports

nickbest-ph:
## Problem

To import events from S3, customers currently have to paste long-lived AWS access keys into PostHog. Keys like that don't expire, people forget to rotate them, and if they leak, whoever has them can read the customer's bucket.

The standard way to give a vendor access to an S3 bucket is an IAM role instead. The customer creates a role in their own AWS account that says "PostHog's importer may read this one bucket". No keys change hands, AWS issues short-lived credentials each time, and the customer can cut off access at any moment by deleting the role.

## Changes

Backend support for IAM role auth on S3 batch imports. It is off for everyone until the `managed-migrations-iam-role-auth` feature flag is turned on, and the flag is enforced by the API itself, not just hidden in the UI. With the flag off (or not defined at all), role requests get a 400 and nothing new is visible.

How it works:

- When creating an S3 import you can now pass a role ARN instead of access keys. One or the other, never both.
- A new `aws_iam_setup` endpoint returns everything the customer needs to set up the role on their side: a trust policy to copy and paste, a permission policy template scoped to their bucket and prefix, and an external ID.
- The external ID is generated by us and unique per project. It's what stops one customer from pointing an import at another customer's role, because the customer pins it in their trust policy and we always send the one belonging to the project that created the import.
- Imports that use a role store no secrets at all. There are simply no keys to save, which is the point of the feature.
- Optionally the API can check at creation time that the role actually works, by assuming it and listing one object in the bucket. That way a broken setup fails immediately with a helpful message instead of when the import starts. This is behind the `MANAGED_MIGRATIONS_VALIDATE_ROLE_ON_CREATE` setting since our API pods need AWS permissions for it first.

Two new settings: `MANAGED_MIGRATIONS_IMPORT_ROLE_ARN` (the PostHog-owned role that customers trust) and the validation toggle above.

The worker side of this (actually assuming the role and reading the bucket) is already on master. The frontend form is stacked on top in #69077.

The full batch imports test file passes locally (105 tests). New coverage:

- Config builder: a role-based import writes the role ARN and external ID into the config and leaves secrets empty, and invalid combinations (role plus keys, role plus custom endpoint, role without external ID, no auth at all) raise with the right message.
- Serializer validation: a parameterized matrix of role only, keys only, both, neither, and partial keys, checked without a database.
- Flag gating: creating with a role ARN is rejected when the flag is off, and `aws_iam_setup` reports unavailable when either the flag or the role ARN setting is missing.
- One end-to-end API test that a role-based create produces the right config and no secrets row.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A for now. The feature is behind a flag, docs will follow with the frontend rollout.

## 🤖 Agent context

Claude Code. Ported from an earlier branch and put behind a feature flag so it can be tested before customer rollout. Used the /improving-drf-endpoints, /django-migrations, and /writing-tests skills. Validation matrix uses SimpleTestCase since DRF field validation needs no database, with one DB-backed test as the wiring guard. A later review session tightened the invalid-combination tests to assert the actual error messages, restored the MCP tools.yaml registration lost in a rebase, and regenerated the OpenAPI types on a current master base so the PR no longer carries a bulk regen.

### DIFF
--- a/products/managed_migrations/backend/api/batch_imports.py
+++ b/products/managed_migrations/backend/api/batch_imports.py
@@ -766,7 +766,7 @@ class BatchImportViewSet(TeamAndOrgViewSetMixin, viewsets.ModelViewSet):
         return Response(response_serializer.data, status=status.HTTP_201_CREATED)
 
     def _is_iam_role_enabled_for_team(self) -> bool:
-        user = self.request.user
+        user = cast(User, self.request.user)
         team = self.team
         return bool(
             posthoganalytics.feature_enabled(
--- a/products/managed_migrations/backend/api/batch_imports.py
+++ b/products/managed_migrations/backend/api/batch_imports.py
@@ -1,11 +1,17 @@
+import json
 import uuid
 from copy import deepcopy
 from datetime import timedelta
 from typing import cast
 
+from django.conf import settings
 from django.db import transaction
 
+import boto3
+import structlog
 import posthoganalytics
+from botocore.config import Config as BotoConfig
+from botocore.exceptions import BotoCoreError, ClientError
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_field
 from rest_framework import filters, serializers, status, viewsets
@@ -24,7 +30,12 @@ from products.managed_migrations.backend.models.batch_imports import (
     BatchImportConfigBuilder,
     ContentType,
     DateRangeExportSource,
+    get_aws_external_id,
 )
+
+logger = structlog.get_logger(__name__)
+
+S3_ROLE_ARN_REGEX = r"^arn:aws:iam::\d{12}:role\/[\w+=,.@\/-]+$"
 
 TRIAL_RECORD_LIMIT_DEFAULT = 1_000
 TRIAL_RECORD_LIMIT_MAX = 50_000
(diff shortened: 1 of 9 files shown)


products/managed_migrations/backend/models/batch_imports.py
```diff
@@ -194,70 +207,105 @@ def from_urls(
...
+            if access_key_id is None or secret_access_key is None:
+                raise ValueError("Exactly one of access keys or role_arn must be provided for S3 sources")
+            source["access_key_id_key"] = access_key_id_key
+            source["secret_access_key_key"] = secret_access_key_key
+            self._store_secret(access_key_id_key, access_key_id)
+            self._store_secret(secret_access_key_key, secret_access_key)
         if endpoint_url:
             source["endpoint_url"] = endpoint_url
             if settings.DEBUG:
-                # Local dev: let custom endpoints point at the dev stack
```
z0br0wn, line 222: nit: you dropped this comment

products/managed_migrations/backend/api/batch_imports.py
```diff
@@ -689,6 +770,88 @@ def create(self, request: Request, **kwargs) -> Response:
         response_serializer = BatchImportResponseSerializer(migration)
         return Response(response_serializer.data, status=status.HTTP_201_CREATED)
 
+    def _is_iam_role_enabled_for_team(self) -> bool:
```
z0br0wn, line 779: is this redundant to _is_iam_role_enabled?

z0br0wn (approved):

PR #73386, opened 2026-07-23
fix(data-imports): classify DB connection blips as transient in pre-write maintenance

Gilbert09:
## Problem

Error tracking surfaced an `OperationalError: [Errno -2] Name or service not known` from `products/warehouse_sources/backend/temporal/data_imports/pipelines/common/extract.py`, with the top in-app frame in `ExternalDataSchema.folder_path`.

The call chain: `run_pre_write_defensive_compact` -> `DeltaTableHelper.run_maintenance` -> `compact_if_fragmented` -> `get_delta_table` -> `_get_delta_table_uri`, which resolves the Delta table's S3 URI via `job.folder_path()` — a Django ORM lookup that traverses `job.schema` and `schema.source` FKs on a pooled app-DB connection. A DNS/pooler blip while opening that connection raised the `OperationalError`.

This is a transient, self-healing infra blip, not a maintenance bug: `run_pre_write_defensive_compact` already catches every exception and skips a matching *transient* one with a quiet warning (the next sync's maintenance pass just retries from scratch). But the existing classifier (`is_transient_object_store_error`) only recognized S3/object-store `OSError`s from delta-rs, so this DB-connection error fell through to `capture_exception` instead — noise in error tracking for something that isn't a code defect.

The same failure class (Postgres `OperationalError`/`InterfaceError` from a transient connection blip) is already classified as transient infra elsewhere in this tree, in `repartition_table.py`'s `_is_transient_infra_error`.

## Changes

- Added `is_transient_maintenance_error` in `delta_table_helper.py`, which treats `django.db.OperationalError`/`InterfaceError` as transient on top of the existing S3/object-store substring matches. It delegates to `is_transient_object_store_error`, which stays unchanged so its other object-store-specific callers (`get_delta_table`, `reset_table`, CDC load) keep their narrower classification.
- Updated `extract.py`'s `run_pre_write_defensive_compact` to use the broadened `is_transient_maintenance_error` classifier.

I found an open PR ([#71934](https://github.com/PostHog/posthog/pull/71934)) that reduces how often this call path issues the underlying lazy DB query in the first place (via `select_related` + memoization). That's a complementary root-cause fix for the redundant query; it doesn't touch this classifier, so this change still closes the gap for connection blips on the remaining DB round-trip. I also found two open PRs ([#69787](https://github.com/PostHog/posthog/pull/69787), [#69309](https://github.com/PostHog/posthog/pull/69309)) that suppress transient DNS/DB errors at the Temporal activity-interceptor level — those cover *uncaught* activity failures. This error never reaches that interceptor: it's caught and manually reported inside `run_pre_write_defensive_compact` itself, so those PRs don't cover this code path.

Extended `TestRunPreWriteDefensiveCompact` in `test_extract.py` with `test_logs_transient_db_connection_error_without_capturing`, parameterized over `OperationalError` (DNS failure) and `InterfaceError` (dropped pooler connection) — asserts `capture_exception` is skipped and a warning is logged instead, mirroring the existing S3/object-store transient-error test right above it. This locks in the exact regression this PR fixes: without the classifier change, both cases would fall through to `capture_exception` again.

Ran:
- `uv run pytest products/warehouse_sources/backend/temporal/data_imports/pipelines/common/test/test_extract.py -k TestRunPreWriteDefensiveCompact` — 9 passed
- `uv run mypy --cache-fine-grained` on the three changed files — clean
- `ruff check --fix` / `ruff format` — clean

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

## 🤖 Agent context

Triaging a webhook-delivered error tracking issue. Used the PostHog MCP error-tracking tools (`query-error-tracking-issue`, `query-error-tracking-issue-events`) to pull the stack trace, then an Explore subagent to read `posthog/sync.py`, the `folder_path` property chain, `delta_table_helper.py`, and existing `NonRetryableErrors`/transient-error precedent across the pipeline. Confirmed the DB round-trip and precedent (`repartition_table.py`'s `_is_transient_infra_error`) directly before writing the fix. Invoked `/writing-tests` before adding the regression test. Checked for duplicate open PRs by exception type, message phrase, module path, and the maintainer's own open PRs — found related-but-non-overlapping work (see Changes section) rather than a duplicate.

---

### DIFF
--- a/products/warehouse_sources/backend/temporal/data_imports/pipelines/common/extract.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/pipelines/common/extract.py
@@ -21,7 +21,7 @@ from products.warehouse_sources.backend.temporal.data_imports.pipelines.core.arr
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.core.cdp_producer import CDPProducer
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.core.delta_table_helper import (
     DeltaTableHelper,
-    is_transient_delta_maintenance_error,
+    is_transient_maintenance_error,
     is_transient_object_store_error,
 )
 from products.warehouse_sources.backend.temporal.data_imports.pipelines.core.person_property_row_sink import (
@@ -705,10 +705,10 @@ async def run_pre_write_defensive_compact(
     the CDC post-load path in `common/load.py` writes the same watermark, and both merge
     via `update_sync_type_config_keys` under a row lock. Wrapped in try/except so a
     maintenance failure never blocks the actual sync; the original error path is unaffected.
-    A transient object-store error (see `is_transient_object_store_error`) or a racy
-    concurrent-maintenance DeltaError (see `is_transient_delta_maintenance_error`) is
-    logged at warning instead of captured — the next sync's maintenance pass retries it
-    from scratch, so neither is a bug in this function.
+    A transient infra error (see `is_transient_maintenance_error`) — an object-store hiccup, a racy
+    concurrent-maintenance DeltaError, or an app-DB connection blip — is logged at warning instead of
+    captured. The next sync's maintenance pass retries it from scratch, so it isn't a bug in this
+    function, just a temporary blip talking to our own S3 bucket, delta table, or app DB.
 
     Used by both `PipelineNonDLT.run` (v2) and `PipelineV3.run` to keep the behaviour
     identical across pipelines without each having to know how to look up `partition_count`
@@ -732,8 +732,8 @@ async def run_pre_write_defensive_compact(
                 schema.id, schema.team_id, updates={"last_vacuum_version": new_version}
             )
     except Exception as e:
-        if is_transient_object_store_error(e) or is_transient_delta_maintenance_error(e):
-            await logger.awarning(f"Pre-write maintenance skipped: transient error: {e}")
+        if is_transient_maintenance_error(e):
+            await logger.awarning(f"Pre-write maintenance skipped: transient infra error: {e}")
             return
         capture_exception(e)
         await logger.aexception(f"Pre-write maintenance failed: {e}", exc_info=e)
--- a/products/warehouse_sources/backend/temporal/data_imports/pipelines/common/test/test_extract.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/pipelines/common/test/test_extract.py
@@ -5,6 +5,8 @@ import pytest
 from posthog.test.base import BaseTest
 from unittest.mock import AsyncMock, MagicMock, patch
 
+from django.db import InterfaceError, OperationalError
+
 import deltalake.exceptions
 from asgiref.sync import async_to_sync
 from parameterized import parameterized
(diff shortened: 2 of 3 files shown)


Gilbert09:
Heads-up on the red `Django Tests Pass` check — it is **not** caused by this PR.

The `posthog/hogql/test/test_parser_cpp_json.py::TestParserCppJson` suite is failing on `master` itself, starting exactly at #73958 (`fix(hogql): cap recursion depth in the C++ parser`). That change updated the C++ parser source and the test to expect a `too deeply nested` `SyntaxError`, but the published `hogql-parser` wheel used by CI (when `common/hogql_parser/` is unchanged relative to `master`, as it is here) has not been rebuilt yet — so `test_deeply_nested_input_does_not_stack_overflow` runs the old parser, spins for ~8.5 min without raising, and drops the idle Postgres connection, cascading every remaining test in the shard to `OperationalError: the connection is closed`.

The wheel republish is already in flight in #75455 (`fix(hogql): publish hogql-parser 1.3.83 with recursion depth limit`). Once that lands on `master`, I'll merge `master` into this branch and re-run — the shard should go green.

This PR itself only touches `warehouse_sources` transient-error classification; its own tests pass locally and in the `warehouse-sources` product shards.

Gilbert09:
Status update — the red CI here is a **master-side regression, not this PR**.

This PR only touches `warehouse_sources` transient-error classification (3 files). Its own tests pass locally (`TestRunPreWriteDefensiveCompact`, 11/11), and the branch is rebased onto master, conflict-free, and `MERGEABLE`.

**What's actually red and why:**

- The required `Django Tests Pass` is red because the **non-required** `Repo checks` job fails deterministically on current master. `Repo checks` runs `hogli product:lint --all`, which fails the isolation-chain / facade-wiring-couplings doctrine for two unrelated products (`customer_analytics` and `error_tracking`) — their facades re-export classes from outside the declared wiring locations. This reproduces on a clean `origin/master` checkout with no changes from this PR, so it isn't introduced here.
- On the `pull_request` event, that deterministic `Repo checks` failure trips `cancel-backend-on-repo-check-failure`, which cancels the whole Backend CI matrix — so the Django test jobs never run and `Django Tests Pass` reports red by cascade.
- In the Trunk merge queue (`merge_group` event) that cancel job doesn't run (it's `pull_request`-gated) and `Repo checks` is optional (`[optional, does not block merge]`), which is why backend PRs continue to merge past this on master.

**Plan:** the branch is already kept current with master. Once the `customer_analytics` / `error_tracking` isolation-chain violation is resolved on master, the next run's Backend CI will no longer be cancelled and this shard will go green. Fixing those products' facade structure is out of scope for a warehouse_sources fix, so I'm not changing them here.

(This supersedes my earlier note about the `hogql-parser` wheel — that specific failure is no longer the blocker.)

Gilbert09:
CI update — the branch is now rebased onto latest `master` (cleanly resolving the overlap with #73454 and #74541), and the earlier `hogql-parser` breakage is gone (1.3.83 is in the tree). This PR's own suites pass.

The current red `Repo checks` is a **separate, master-wide** issue, not from this PR. Its `Lint product structure` step (`./bin/hogli product:lint --all`) fails with:

```
✗ 2 product(s) failed: customer_analytics, error_tracking
```

Both report that they are sealed and carry `backend:contract-check` but their `turbo.json` doesn't narrow contract-check inputs — a rule tightened by #75140 (`chore(ci): narrow merge-queue lanes for isolated product changes`, merged earlier today). Neither product is touched by this PR (the diff is exactly three `warehouse_sources` files), and `product:lint --all` reproduces the same two failures locally on a clean checkout — so it's a master-state problem those product owners (or a revert of #75140) need to resolve. When `Repo checks` goes green on `master` again, I'll merge `master` in and the gate should pass here too.

Gilbert09:
### CI status: current red checks are a fresh master-wide regression, not this PR

The earlier `test_parser_cpp_json` failure is gone. The red checks now are **`Repo checks`** (the `product:lint --all` "Lint product structure" step) and the downstream **`Django Tests Pass`** gate that keys off it.

**Root cause — [#75140](https://github.com/PostHog/posthog/pull/75140)** (merged to master today, ~1h before this run). It broadened the `backend:contract-check` inputs in two products:

- `products/customer_analytics/turbo.json`: `backend/models/team_customer_analytics_config.py` → `backend/models/**` (plus `backend/test/factories.py`)
- `products/error_tracking/turbo.json`: added `backend/models.py`

`IsolationChainCheck` / `has_narrowed_turbo_inputs` only treats a product as "narrowed" when every contract-check input is confined to an accepted surface (facade/presentation, routes, garages, permanent, carve-out). A broad `backend/models/**` glob isn't confined, so `has_narrowed_turbo` flips to `false` and the `"fully sealed and eligible ... but turbo.json does not narrow"` error fires for both products:

```
✗ 2 product(s) failed: customer_analytics, error_tracking
```

**Proof it's master-wide, not this branch:** unrelated PR [#75701](https://github.com/PostHog/posthog/pull/75701) (`fix(signup)`, touches neither `warehouse_sources` nor those two products) fails the *identical* lint on the *same* two products. This PR only touches 3 `warehouse_sources` files and cannot fix another product's isolation config.

**Fix belongs on master** — the two `turbo.json` files need to go back to specific carve-out paths (or the accepted-surface set extended). Reported to the devex team. Once master is green, this branch just needs a re-merge; no change is needed here.

Gilbert09:
### Fixed a second, pre-existing master breakage that blocked this product's tests

After the master merge above turned `Repo checks` green, the `Product tests (warehouse-sources …)` job surfaced a separate pre-existing breakage: pytest collection was interrupted by 6 `ModuleNotFoundError`s —

```
ModuleNotFoundError: No module named
'products.warehouse_sources.backend.temporal.data_imports.pipelines.pipeline'
```

Five source modules in `aws_cost_explorer` and `app_store_connect` imported `SourceInputs`/`SourceResponse` from `…data_imports.pipelines.pipeline.typings`, a path that no longer exists — the canonical location is `…data_imports.sources.common.typings` (used by every other source). This is present on `master` itself (unrelated to this PR's 3 classifier files), so it interrupts the whole `warehouse_sources` suite for any PR that touches the product.

Rather than block on it, I corrected the five imports to the canonical path (commit `cc792c8`) — a zero-behavior path fix. The PR now owns 8 files: the 3 original classifier changes plus these 5 import corrections. Also reported the breakage to the devex team.

PR #73387, opened 2026-07-23
refactor(agentic-provisioning): drop stripe-only surface and hmac auth

rafaeelaudibert:
> [!CAUTION]
> **Do not merge until Stripe has confirmed moved to `/api/partners/stripe/`.**
> This PR deletes the legacy endpoints and auth that the Stripe App orchestrator was still using as of 2026-07-23: it was polling the legacy `/api/agentic/provisioning/health` and `/services` with a valid signature even after the app manifest was repointed. Merging before that traffic drains breaks the live Stripe integration (its polls would 404 and its HMAC auth would stop being accepted).
>
> **How to verify the cutover is done** (internal "PostHog App + Website" project, US):
>
> ```sql
> SELECT
>     if(properties.$request_path LIKE '/api/partners/stripe/%', 'new', 'legacy') AS ns,
>     event, properties.outcome AS outcome, count() AS n, max(timestamp) AS last_seen
> FROM events
> WHERE timestamp >= now() - INTERVAL 7 DAY
>   AND event LIKE 'agentic_provisioning%'
>   AND (properties.client_name = 'Stripe App' OR event = 'agentic_provisioning signature verification')
> GROUP BY ns, event, outcome
> ORDER BY ns, n DESC
> ```
>
> Merge only when both hold for at least a couple of days:
> 1. `legacy` rows show no `signature verification` with `outcome = 'success'` and no `Stripe App` flow events (the orchestrator polls every ~30 minutes, so a quiet day is a strong signal).
> 2. `new` rows show Stripe's polls and flows arriving under `/api/partners/stripe/`.
>
> Events carry `$request_path` automatically (posthoganalytics Django integration), which is what makes the namespace split reliable. Prod web logs (`service_name = 'posthog-web-django-granian'` in the same project's Logs) give a second, event-independent signal.

## Problem

#72799 moved Stripe Projects provisioning into `ee/partners/stripe/`, but the generic partner endpoints in `ee/api/agentic_provisioning/` still carried the whole Stripe-only surface: HMAC signature auth, the legacy Stripe OAuth app fallback, the services catalog, SPT billing, and protocol details only the Stripe orchestrator uses. That kept the generic partner API entangled with Stripe and left two copies of the Stripe behavior to maintain.

## Changes

**Before** (one namespace served both):

```mermaid
flowchart LR
    S{{Stripe orchestrator}} -->|HMAC + bearer| L["/api/agentic/*"]
    P{{Generic partners}} -->|bearer / PKCE| L
    L --> PH[(PostHog)]
    class S,P phBlue;
    class L phYellow;
    class PH phRed;
```

**After** (each partner surface owns its contract):

```mermaid
flowchart LR
    S{{Stripe orchestrator}} -->|HMAC, full APP 0.1d| N["/api/partners/stripe/*"]
    P{{Generic partners}} -->|bearer / PKCE only| L["/api/agentic/*"]
    N --> PH[(PostHog)]
    L --> PH
    class S,P phBlue;
    class N,L phYellow;
    class PH phRed;
```

Removed from `ee/api/agentic_provisioning/`:

- **All HMAC auth**: the global `STRIPE_SIGNING_SECRET` verification, the per-partner `hmac` auth method in `ProvisioningAuthentication`, and every `Stripe-Signature` code path (including the header in the region-proxy forward allowlist)
- **Stripe-only endpoints**: `/provisioning/health`, `/provisioning/services` (plus the billing-backed services catalog), and `/provisioning/resources/:id/update_service`, in both the `/api/agentic/` and `/api/provisioning/` namespaces
- **SPT billing**: `payment_credentials` handling and the billing-service activation helpers
- **The `service_id` vocabulary** (`analytics` / `free` / `pay_as_you_go`): the field is part of the Stripe orchestrator protocol, not the generic partner flow; requests no longer validate or persist it and resource responses no longer echo it
- **Legacy Stripe OAuth app fallback**: bearer auth accepts only active provisioning partners, and token exchange requires a code minted for a resolvable partner with PKCE
- **The `API-Version: 0.1d` check**: generic partners are not versioned here (the Stripe namespace keeps its own)
- **Callback fallback**: consent redirects resolve only from the partner's `redirect_uris` and fail closed with `missing_callback` instead of falling back to a Stripe setting

> [!WARNING]
> Behavior changes for the generic namespace: `account_requests` now requires an identified partner (bearer or PKCE) and returns 401 otherwise, token exchange is PKCE-only, and resource responses no longer include `service_id`. Telemetry over the last 60 days shows no generic partner depending on the removed pieces.

Also: `ee/billing/billing_manager.py` borrowed the deleted `compute_signature` helper for billing-provider webhooks, so an identical private helper is now inlined there (same algorithm, no wire change).

- `hogli test ee/api/agentic_provisioning/test/` - 314 passed. Tests were reworked around a bearer/PKCE partner base; Stripe-flow tests were deleted here because the copies under `ee/partners/stripe/api/provisioning/test/` cover them. New fail-closed guards that no existing test covered: partner-less `account_requests` returns 401, codes without a PKCE challenge or a resolvable partner are rejected, an `hmac`-configured partner can no longer authenticate, and empty requested scopes default to the app ceiling.
- `hogli test ee/partners/stripe/` - green (the Stripe namespace is untouched by this PR).
- `uv run mypy --cache-fine-grained .` - clean. `ruff` clean. `hogli ci:preflight --fix` - 0 failures.
- Probed prod to confirm the new namespace is deployed and serving (401 with a signature demand, not 404).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The public Provisioning API doc (`contents/docs/integrate/provisioning.mdx` on posthog.com) still documents the `API-Version: 0.1d` header requirement and the `service_id` field. That needs a companion PR in the posthog.com repo once this merges; sending the header stays harmless in the meantime (it is ignored).

## 🤖 Agent context

Authored with Claude Code (Claude Fable 5), directed by @rafaeelaudibert. The `/writing-tests` skill gated the test rework; five parallel subagents converted the test suite from HMAC-signed requests to bearer/PKCE partner auth.

Decisions along the way: the `API-Version` check and the `service_id` ids were initially kept as seemingly generic protocol pieces, then removed after checking the APP 0.1d spec and confirming they belong to the Stripe orchestrator surface. Before opening this PR we (well, Claude) checked the self-capture telemetry and prod logs and found Stripe's orchestrator still polling the legacy namespace with a valid signature after the announced cutover, which is why this PR ships with the merge gate above instead of merging right away. Follow-ups deliberately left out: dropping the now-unread `provisioning_signing_secret` column and `hmac` auth-method option from `OAuthApplication` (needs a migration), and the posthog.com doc update.


rafaeelaudibert:
* **#73387** <a href="https://app.graphite.com/github/pr/PostHog/posthog/73387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/73387?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

MattBro (approved):
LFG!!!!!

Image

I have a few suggestions that I think would be good, but not worth blocking, so feel free to ignore.

ee/api/agentic_provisioning/views.py
```diff
@@ -1300,23 +1061,14 @@ def _exchange_authorization_code(request: Request) -> Response:
...
-
     # Lock the app row before reading the revoke stamp and minting, so this serializes
     # with revoke_application_sessions (see _lock_application). Provisioning auth codes
     # live in the cache, not OAuthGrant, so the revoke's sweep can't reach them — the
     # `issued_at` carried on the code is what a revoke is checked against. Codes minted
     # before `issued_at` shipped lack the field; fail closed (they expire in
     # AUTH_CODE_TTL_SECONDS and the client can re-run the flow).
     with transaction.atomic():
-        locked_app = _lock_application(oauth_app.pk) if oauth_app else None
+        locked_app = _lock_application(oauth_app.pk)
```
MattBro, line 1071: nit: `oauth_app` can't be None anymore, so `locked_app` is only None when the app row gets deleted between the `.get()` and the row lock - and in that race this skips the revoke check, resolves scopes against the unprivileged default, then 500s on the FK at token creation. Returning `invalid_grant` when `locked_app` is None lets you drop both `if locked_app else` fallbacks (the `oauth_app and` in the token_expiry ternary below is dead too). The matching guards in the refresh path are fine as-is since `application_id` there comes from a nullable FK.

ee/api/agentic_provisioning/views.py
```diff
@@ -1058,6 +813,11 @@ def agentic_authorize(request: Any) -> HttpResponseBase:
         organization = memberships[0].organization
         team = non_demo_teams[0]
 
+        callback_url = _get_callback_url(pending.get("partner_id", ""))
+        if callback_url is None:
```
MattBro, line 817: nit: no test hits this branch - `test_confirm_without_partner_returns_missing_callback` covers the confirm path, but nothing sets up a skip-consent partner with empty `redirect_uris` to exercise the auto-redirect side.

ee/api/agentic_provisioning/views.py
```diff
@@ -499,9 +268,7 @@ def _caller_proved_existing_trust(partner: OAuthApplication, user: User, authent
 def _handle_existing_user(
     request_id: str,
     user: User,
-    confirmation_secret: str,
     scopes: list[str],
-    partner_account_id: str = "",
     region: str = "US",
     team_id: int | None = None,
     partner: OAuthApplication | None = None,
```
MattBro, line 274: nit: every caller passes a non-None partner now that account_requests 401s first, so this default and the `if partner else ""` guards here and in `_handle_new_user` are dead. Tightening it means reordering params or making the trailing ones keyword-only (partner sits after defaults), so fine as a follow-up.

PR #73395, opened 2026-07-23
fix(data-imports): don't report transient S3 timeouts during query-folder cleanup

Gilbert09:
## Problem

Error tracking surfaced a `ConnectTimeoutError` (`Connect timeout on endpoint URL: ".../?delete"`) raised from `delete_folder` in `products/warehouse_sources/backend/temporal/data_imports/util.py`, during a Stripe sync's post-load cleanup of stale query folders in S3.

The stack trace shows a plain TCP connect timeout (`botocore.exceptions.ConnectTimeoutError`, wrapping an `aiohttp` connection timeout) — a transient network blip, not a code defect. The S3 client used here (`aget_s3_client`) has no explicit timeout/retry config, so it relies on `s3fs`/`aiobotocore` defaults, and this particular request didn't succeed within those.

Importantly, this delete is already best-effort: `delete_folder` catches any exception, logs it, and moves on — the sync itself isn't affected. The old folder in question is timestamped and age-gated, so if it's not cleaned up this run, the next sync's cleanup pass will attempt it again. The only actual problem is that the exception handler unconditionally calls `capture_exception`, so every transient timeout during this cleanup mints a fresh error-tracking issue for something that isn't a bug and self-heals.

This mirrors the precedent in #70276 and #69309: classify the known-transient failure and stop reporting it, while genuine failures still get captured.

## Changes

- Added `_is_transient_s3_connection_error` in `util.py`, classifying `botocore.exceptions.ConnectionError` (covers `ConnectTimeoutError`, `EndpointConnectionError`), `ReadTimeoutError`, and `ConnectionClosedError` as transient.
- `delete_folder` now skips `capture_exception` for these (still logs a warning-level message), and continues to capture anything else.

Added `products/warehouse_sources/backend/temporal/data_imports/tests/test_util.py`:
- Parameterized `test_is_transient_s3_connection_error` covering connect timeout, endpoint connection error, read timeout, and connection-closed as transient, versus a `ClientError` (e.g. access denied) and a generic `ValueError` as not transient.
- `test_delete_folder_swallows_transient_s3_connection_error` — drives `prepare_s3_files_for_querying` with a mocked S3 client whose `_rm` raises `ConnectTimeoutError`, asserts `capture_exception` is not called.
- `test_delete_folder_still_captures_non_transient_error` — same setup with a `ValueError`, asserts `capture_exception` is still called.

These catch the exact regression (a real cleanup bug silently going unreported, or a transient blip minting a fresh issue) that no existing test covered — there was no prior test exercising `prepare_s3_files_for_querying`'s delete-folder error handling at all.

Ran locally: `pytest products/warehouse_sources/backend/temporal/data_imports/tests/test_util.py` (8 passed), `ruff check`/`ruff format` clean, and a full-repo `mypy --cache-fine-grained .` (16781 files, no issues).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable — internal exception-handling behavior, no user-facing or documented workflow change.

## 🤖 Agent context

Triaged from an error-tracking issue about a `ConnectTimeoutError` in the data-imports pipeline. Confirmed it wasn't already fixed by an open PR (checked #72106, which addresses a different S3 client/code path — error-tracking's symbol-set cleanup, not this warehouse-sources import pipeline). Invoked `/writing-tests` before adding tests.

Key decision: classified this as a noise-suppression fix rather than a retry/backoff change, since the delete is already best-effort and self-healing via the next sync's age-gated cleanup pass — adding retries here would add complexity without changing the outcome.


Gilbert09:
This PR is green and cleanly mergeable except for one check — `semgrep` — which is failing for a reason unrelated to this change.

**What's happening**

- The `semgrep` check comes from a workflow (`.github/workflows/semgrep.yml`, workflow id `242597240`) that no longer exists in the repo: the file is absent from `master` and from this branch, and the workflow id `404`s via the Actions API. The repo's current Semgrep setup is the `Security` workflow (`semgrep-python`, `semgrep-general`, …), and all of those pass here (`Semgrep Checks Pass` ✅).
- The orphaned run reports **0 findings (0 blocking)**. It exits non-zero only because semgrep's bash parser trips on `.github/scripts/verify-playwright-new-tests-and-snapshots.sh:111` (a valid `(( … ))` arithmetic expression) while running with `--error`. That file isn't touched by this PR and is byte-identical to `master`.
- The failure is flaky/environmental, not caused by any diff here: the identical script line parses fine on other PRs (e.g. #74251 is green), while several other current open PRs are red on the same `semgrep` check. I re-ran the failed job to confirm it wasn't a one-off — it reproduced with the same 0-findings parse error.

Because the workflow no longer lives in the repository, it can't be edited, re-run to green, or removed from a PR branch. Clearing it needs a maintainer to either drop the stale `semgrep` context from `master`'s required status checks or restore/repair that workflow.

**Everything else**

- Rebased to resolve the conflict, then merged the latest `master` in (two-parent merge, no conflicts); only this PR's two files differ from base — no base-branch leakage.
- The merged `test_util.py` keeps both master's `test_copies_files_with_cp_file_not_copy` and this PR's transient-error tests; all pass locally.
- All other required checks are green.

PR #73403, opened 2026-07-24
fix(revenue-analytics): coerce currency property to string before upper()

ayaangazali:
## Problem

Revenue analytics converts each event's revenue using the currency property you configure. If that property is accidentally pointed at a numeric field (a number instead of a currency code like `USD`), the whole query blows up with a generic "there is a problem with this query" error and you get no results at all.

The root cause is in `currency_expression_for_events`: it wraps the property in `upper()`. When the property is typed as Numeric, HogQL resolves it to a `Float64`, and ClickHouse rejects `upper(Float64)` with `Illegal type Float64 of argument of function upper`.

Closes #44914

## Changes

Coerce the property to a string before `upper()`. `toString(...)` is a no-op for the normal string case, and for a misconfigured numeric property it produces a string that flows through the existing unknown-currency path (no matching exchange rate, so `convertCurrency` already returns 0) instead of crashing.

So a misconfigured event now shows revenue 0, which nudges the user toward the bad setting instead of taking the whole insight down with it.

Backend only, no frontend. Everything automated and run locally against the dev ClickHouse.

- Added `test_revenue_currency_property_numeric_value_does_not_crash` to the revenue example events query runner tests. It registers a Numeric `PropertyDefinition` for the currency property (so HogQL types it as `Float64`, matching the reported scenario), stores a numeric currency value, runs the query, and asserts it returns a row with converted amount `0` rather than raising.
- Regression it catches: if the `toString` coercion is dropped, the property resolves to `Float64` inside `upper()` and the query throws `Illegal type Float64`. I proved this by reverting just the fix and rerunning, which fails with exactly that ClickHouse error.
- Ran the full `products/revenue_analytics/backend` suite (262 passed) to confirm the fix is a clean no-op for the normal string-currency path. The only snapshot movement is `upper(x)` becoming `upper(toString(x))` across the currency-property query snapshots, which I regenerated.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change. This is an internal query-builder fix with no user-facing config or API change.

## 🤖 Agent context

Honest disclosure: I'm a freshman in college contributing through a small autonomous PR pipeline I run on my own account, and the implementation here was done by Claude Code (Opus 4.8) while I review and own the result. Sorry in advance if anything looks off, I tried to reason through the logic myself and sanity-checked the decisions with Claude Code too.

Skills invoked: `/writing-tests` (to gate the new test before writing it).

Decisions along the way: I considered fixing this inside the HogQL `property_types` transform, which already keeps a `_STRING_INPUT_CONVERSIONS` set to skip the numeric cast for functions that need string input (like `toFloatOrZero`), by adding `upper`/`lower` to it. I rejected that because it is a broader change to a core transform and would ripple well beyond revenue analytics. The caller-side `toString` wrap is smaller, self-documenting, and scoped to the one spot that needs it. I also verified `convertCurrency` already degrades an unknown currency to 0 (its `dictGetOrDefault` rate lookup returns 0), so sending the numeric value down that path is consistent with existing behavior rather than a new fallback.


ayaangazali:
Closing this one, I am no longer working in this repo. Feel free to pick up the change if it is still useful.

PR #73418, opened 2026-07-24
feat(experiments): recalculation workflow MCP tools

rodrigoi:
## Problem

Agents could not read an experiment's current results or trigger a fresh calculation over MCP: the recalculation endpoints existed but were scaffolded as `enabled: false`, so the only results-shaped tool was `experiment-timeseries-results`, which needs a `metric_uuid` and `fingerprint` up front and only answers the day-by-day question.

## Changes

This PR enables the metrics recalculation endpoints as MCP tools and makes them the default path for reading experiment results.

### Recalculation tools

Enabled the three already-scaffolded operations in `products/experiments/mcp/tools.yaml`, so agents get the full trigger-then-poll loop. The descriptions carry the semantics the endpoints imply but the schema cannot: reading never starts a run, `-create` is the only way to start one, and polling should watch `status` and `rows_read` rather than spin. They also flag the two response fields that are easy to misread, `active_run` (the results shown are from the previous run) and `result_source: timeseries_fallback` (a cold-start placeholder, not a real run).

| Tool | Endpoint | Scope |
| - | - | - |
| `experiments-metrics-recalculation-create` | `POST metrics_recalculation/` | `experiment:write` |
| `experiments-metrics-recalculation-latest-retrieve` | `GET metrics_recalculation/latest/` | `experiment:read` |
| `experiments-metrics-recalculation-retrieve` | `GET metrics_recalculation/{uuid}/` | `experiment:read` |

`-create` is annotated `idempotent: true`, which matches the endpoint: a second call while a run is pending or in progress returns that run with `is_existing` and a 200 instead of queueing a second workflow.

- `products/experiments/mcp/tools.yaml`

### Superseding the timeseries tool

There was no generic query or results tool for experiments, so the one being superseded is `experiment-timeseries-results`. It stays enabled, since the recalculation endpoints return a run's latest results and not a daily series, so disabling it would drop that capability. Instead it is retitled to "Get experiment metric results over time" and scoped in its description to the history of a single metric, pointing at `-latest` for the general "how is this experiment doing" case. `-latest` takes the title "Get experiment results" so it also wins on name matching.

- `products/experiments/mcp/tools.yaml`

### Path parameter schema

The `recalculation_id` path parameter reached the generated Zod schema with no description, which leaves agents guessing whether it is the run id or the experiment id. Path parameter descriptions come from the OpenAPI spec rather than YAML `param_overrides`, so the fix is an `OpenApiParameter` on `metrics_recalculation_detail`. It declares an explicit string type carrying the URL's UUID pattern. Using `OpenApiTypes.UUID` produces the description but silently drops the regex, which would leave the generated tool without id-shape validation.

- `products/experiments/backend/presentation/views.py`

### Generated files

- `services/mcp/schema/generated-tool-definitions.json` (generated)
- `services/mcp/schema/tool-definitions-all.json` (generated)
- `services/mcp/src/generated/experiments/api.ts` (generated)
- `services/mcp/src/tools/generated/experiments.ts` (generated)
- `services/mcp/tests/unit/__snapshots__/tool-schemas/*.json` (generated)

### No UI app attached

The existing `experiment-results` view expects `primaryMetricsResults` and `exposures`, but these endpoints return a flat `results` array, so wiring it up would render empty cards. Left off deliberately; adapting that view to the recalculation shape is a reasonable follow-up.

> [!NOTE]
> Running `hogli build:openapi` also rewrites about 59 unrelated generated files with import-ordering churn. That drift reproduces on a clean master with no changes, so it is pre-existing and is left out of this PR to keep the diff reviewable.

```bash
pnpm --filter=@posthog/mcp vitest run tests/unit
```

```bash
pnpm --filter=@posthog/mcp lint-tool-names
```

```bash
uv run mypy --cache-fine-grained .
```

cat-type-small

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Model: Opus 4.8
Manually refactored: **no**

Skills used:

- /implementing-mcp-tools (local)

Relevant decisions:

- Kept `experiment-timeseries-results` enabled rather than disabling it. The recalculation endpoints return a run's results, not a daily series, so disabling it would have removed a capability rather than superseded one. The redirect is done through the title and description instead.
- Declared `recalculation_id` with an explicit string type and pattern instead of `OpenApiTypes.UUID`, because the bare UUID type drops the regex from the generated schema and the tool loses id-shape validation.
- Set `enrich_url` to `{params.id}` rather than the scaffold default `{id}`. These responses carry the recalculation UUID in `id`, so the default would have produced PostHog links pointing at a nonexistent experiment.


rodrigoi:
* **#73418** <a href="https://app.graphite.com/github/pr/PostHog/posthog/73418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/73418?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mp-hog (approved):
lgtm. Some findings nevertheless:

- experiments-metrics-recalculation-create surfaces the full trigger enum to the agent (manual, cold_run, stale_refresh, auto_refresh, config_change, experiment_launch, experiment_stop, experiment_update). These are system labels, not user choices. An agent can plausibly pick experiment_stop for example, which (1) mislabels the run's provenance in the trigger column and any downstream analytics keyed on it, and (2) adds a decision the agent shouldn't be making. You could use `exclude_params: [trigger]` (possibly adding a dedicated `mcp/agent` trigger value server-side if agent runs should be distinguished)

- Tool naming slightly inconsistent: All active MCP tools use singular "experiment-..." prefix, these new tools use plural "experiments-metrics-recalculation-create". Fixing now would be cheaper before agents store them.

- poll guidance relies entirely on agent compliance, there's no server-side backoff. since there is tracking, it's not a blocker but something to watch for

PR #73419, opened 2026-07-24
fix(replay-vision): validate group-summary citations against observation outcomes

ksvat:
## Problem

Replay Vision group summaries (the "Summaries and alerts" tab) synthesize up to 100 observations into one report and cite their sources inline with `[obs N]` markers. I audited real production runs and found the citations aren't reliably trustworthy. The recurring failure: the synthesis model takes a single real error and inflates it into a multi-observation trend, padding the citation list with successful or unrelated sessions. Across five runs, three showed a padded or fabricated cluster.

Concrete example from a real run: a "some users encounter invalid invite link errors" claim cited six observations. Only one was a real error. The other five were successful waitlist signups (their own observations say "no friction"). A reader clicking those citations to verify the trend lands on five people who signed up fine.

Two secondary issues showed up too: the body sometimes states its own recording count that contradicts the authoritative header (one run said "59 sessions" for a 72-observation run), and a capped run described its sampled 100 as if it were the complete set.

Every bad citation still resolves to a real observation, so the existing pipeline (which only drops out-of-range `[obs N]`) catches none of it.

## Changes

Three independent layers of defense, plus tests:

1. **Feed the signal into synthesis.** Each summarizer observation line now carries its structured `outcome` and friction status (`friction: none` or `friction: <specific problems>`), alongside the existing verdict/score/tags. The model was previously inferring success-vs-error from prose alone, so the machine-readable "this was clean" signal never reached it.

2. **Tighten the prompt.** Cite an observation for a negative claim only if its own signals report that problem; never cite a `friction: none` session as an error; don't manufacture a cluster from one instance; a classifier's tag can be coarse (a bare `abandoned` doesn't say what was abandoned); never state your own count or call a sampled window complete.

3. **Validate after synthesis** (`_validate_citations`). After the model writes the report, drop any `[obs N]` whose sentence makes a negative (error/friction) claim while the cited observation is an affirmative success with no friction. Conservative by design: it only removes clear contradictions, never rewrites prose, and leaves ambiguous or "can't tell" cases alone so it never strips real evidence. Drops are logged (`vision_action.synthesis.citations_dropped`) so we can measure how often it fires.

I (Claude) added four unit tests to `test_vision_action_synthesis.py`: the enriched line signal is present, a clean session cited on a negative claim is dropped, a real-friction citation is kept, and a clean citation on a non-negative claim is kept. Full file passes (44 tests). Ran `hogli ci:preflight` (0 failures; ruff lint + format clean). `mypy` on the changed file is clean apart from one pre-existing `OBJECT_STORAGE_REGION` error in `posthog/settings/managed_migrations.py` that I confirmed exists on master and is unrelated to this change.

The audit itself was done by reading real runs through the vision-actions MCP tools; I did not run the synthesis against a live LLM here (the tests mock it).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Kim asked me to audit whether the citations in Replay Vision group summaries actually support their claims, then fix what I found. I audited five production runs across summarizer, monitor, and classifier scanners (via the `vision-actions-runs-retrieve` MCP tool added in an earlier PR), cross-referencing each `[obs N]` against the cited observation's real outcome.

Findings that shaped the fix: the fabrication is intermittent (same action was clean one day, padded the next), and it's a semantic-mismatch failure — every citation resolves in-range, so out-of-range dropping does nothing. A contributing cause is upstream scanners emitting under-specified signals (a monitor that assumes "user is merging issues" even on unrelated sessions; a classifier's context-free `abandoned` tag). This PR addresses the synthesis layer; the upstream-scanner prompts are out of scope.

I initially assumed we weren't feeding outcome data to the model at all, but on checking `describe_output` we were feeding title/verdict/tags — just not the summarizer's structured `outcome`/`friction_points`. That's why the fix both enriches the line and adds the validator, rather than only rewording the prompt.

Tools/agent: Claude Code (Opus 4.8). Skills invoked: `/exploring-replay-vision-observations`, `/implementing-mcp-tools`, `/writing-tests`.

### DIFF
--- a/products/replay_vision/backend/temporal/vision_actions/synthesis.py
+++ b/products/replay_vision/backend/temporal/vision_actions/synthesis.py
@@ -24,6 +24,7 @@ from posthog.models.team import Team
 from posthog.sync import database_sync_to_async
 
 from products.replay_vision.backend.models.replay_observation import ObservationStatus, ReplayObservation
+from products.replay_vision.backend.models.replay_scanner import ScannerType
 from products.replay_vision.backend.models.vision_action import VisionAction, VisionActionRun, VisionActionRunStatus
 from products.replay_vision.backend.observation_formatting import EVENT_ID_CITATION_RE, describe_output
 from products.replay_vision.backend.scanner_access import readable_scanner_ids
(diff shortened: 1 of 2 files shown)


ksvat:
* **#73419** <a href="https://app.graphite.com/github/pr/PostHog/posthog/73419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/73419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ksvat:
### Merge activity

* **Jul 24, 4:02 AM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/73419).

PR #73423, opened 2026-07-24
fix(data-warehouse): prevent scroll from changing port number in source form

### LINKED ISSUE #72401: Scrolling changes port number
### Bug Description

Scrolling on the postgres data page changes the port number.

## How to reproduce

1. https://eu.posthog.com/project/***/data-warehouse/new-source?kind=Postgres (data > sources > postgres > query direct
2. Enter a port number
3. Scroll up or down

https://github.com/user-attachments/assets/ce0e807f-0c5c-4cb2-b6dd-8eca0ca41668

Also minor nit, but the help agent suggested something to enter here but I couldn't copy and paste from it.

### Debug info

```shell
- [ ] PostHog Cloud, Debug information: [please copy/paste from https://us.posthog.com/settings/project-details#variables or https://eu.posthog.com/settings/project-details#variables]
```

okxint:
## Problem

When adding a new Postgres (or other) data warehouse source, scrolling over the **Port** field changes its value unexpectedly. This happens because the port field config declares `type: 'number'`, and the `sourceFieldToElement` fallback renders a native `<input type="number">` — which browsers increment/decrement on scroll wheel events by default.

Closes #72401

## Changes

Added an explicit `field.type === 'number'` guard in `sourceFieldToElement` (`products/data_warehouse/frontend/shared/components/forms/SourceForm.tsx`). When matched, renders `LemonInput` with `type="text" inputMode="numeric"` instead of the native number input. This:
- Removes scroll-wheel value changes (the reported bug)
- Preserves numeric keyboard on mobile via `inputMode="numeric"`
- Keeps validation behavior unchanged (field still accepts only digits)

The existing fallback `type={field.type as 'text'}` cast was a no-op at runtime — it told TypeScript the type was `'text'` but the DOM still received `"number"`.

The agent was not able to run the full dev stack. The fix is a straightforward prop change on a single component. Manual verification: open the data warehouse new-source form for Postgres, scroll over the Port field — the value should no longer change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs update needed.

## 🤖 Agent context

Fix identified from issue #72401. The root cause was `type={field.type as 'text'}` in `sourceFieldToElement` — the TypeScript cast is a lie; the DOM still receives `type="number"` for port fields, which causes scroll-to-change behavior. Added an explicit `'number'` branch that renders `type="text" inputMode="numeric"` instead.

### DIFF
--- a/products/data_warehouse/frontend/shared/components/forms/SourceForm.tsx
+++ b/products/data_warehouse/frontend/shared/components/forms/SourceForm.tsx
@@ -370,6 +370,36 @@ export const sourceFieldToElement = (
         )
     }
 
+    if (field.type === 'number') {
+        // Use type="text" with inputMode="numeric" to prevent browsers from
+        // changing the value on scroll, which is the default behavior for
+        // native <input type="number"> elements.
+        return (
+            <LemonField
+                key={field.name}
+                name={field.name}
+                label={field.label}
+                help={
+                    field.caption ? (
+                        <LemonMarkdown className="text-xs">{field.caption}</LemonMarkdown>
+                    ) : undefined
+                }
+            >
+                {({ value, onChange }) => (
+                    <LemonInput
+                        className="ph-ignore-input"
+                        data-attr={field.name}
+                        placeholder={field.placeholder}
+                        type="text"
+                        inputMode="numeric"
+                        value={value || ''}
+                        onChange={onChange}
+                    />
+                )}
+            </LemonField>
+        )
+    }
+
     return (
         <LemonField
             key={field.name}


products/data_warehouse/frontend/shared/components/forms/SourceForm.tsx
```diff
@@ -370,6 +370,36 @@ export const sourceFieldToElement = (
         )
     }
 
+    if (field.type === 'number') {
```
rafaeelaudibert, line 373: This can be handled without this much duplication with ternary operators inside the existing code in lines 411-418

rafaeelaudibert (changes requested):

PR #73428, opened 2026-07-24
feat(frontend): add truncateMiddle utility helper to strings.ts

mahenoorsalat:
## Description
Adds a type-safe \	runcateMiddle\ string utility function to \rontend/src/lib/utils/strings.ts\ for cleanly truncating long string labels, event names, and URLs in UI components.

## Features
- Middle truncation helper \	runcateMiddle(str, maxLength)\.
- Preserves prefix and suffix identifiers for compact UI rendering.


rafaeelaudibert:
This doesn't seem useful, it's not being used anywhere.

PR #73438, opened 2026-07-24
feat(workflows): notify on email reputation state transitions

meikelmosby:
## Problem

Email reputation is calculated daily and visible in the Reputation tab, but customers aren't told when it degrades – they discover problems only if they check the tab. For phase A of enforcement (manual suspension via the merged #72891 kill switch), the signal has to reach the customer early enough to fix their lists before a human decides whether to suspend.

## Changes

**Transition detection** – hourly Celery task `check_email_reputation_transitions` compares each team's two most recent team-scope snapshots (the evaluator writes them daily at 06:00 UTC; hourly scanning just keeps notification lag small when a run lands late):

- entered `warning` → customer email + in-app notification
- entered `critical` → customer email ("sending may be suspended") + critical in-app notification
- persisting states and recoveries → nothing

Idempotency is layered: a per-snapshot cache marker stops hourly rescans, and the email campaign key embeds the snapshot's `evaluated_at` so `MessagingRecord` dedupes even if the marker is evicted. Per-team failures are isolated so one bad team can't starve the scan.

**Customer emails** – `send_email_reputation_degraded` follows the `send_hog_function_disabled` pattern: recipients via `get_members_to_notify` (org members with project access), gated by a new `email_reputation_degraded` notification setting (default on, toggle added to user notification preferences). Two templates (warning/critical) state the rates, what they mean, and the next action, linking to the Reputation tab.

**In-app** – uses the `email_reputation` notification type (merged in #72891): normal priority for warning, critical (persistent toast) for critical, targeted at the team, linking to the Reputation tab.

> [!NOTE]
> Internal alerting is deliberately **not** in this PR. An earlier revision posted to Slack via a hardcoded webhook; that was replaced by #74558, where the Temporal evaluator emits `workflow_email_reputation_state_changed` into PostHog's own project – the internal Slack alert becomes a workflow/destination on that event, so routing and filters are UI config instead of code.

- Parameterized transition matrix covers all eight state combinations (entering warning from each prior state, entering critical, persisting critical/warning, recovery) asserting exactly which side effects fire.
- Rerun idempotency: second scan of the same snapshot sends nothing; a new snapshot after a processed one is processed.
- Notification-settings tests updated for the new default key (full-dict assertions).
- Repo-wide `mypy --cache-fine-grained .` clean; 10/10 tests pass.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with PostHog Code (Claude). Skills invoked: /sending-notifications, /writing-tests. Key design decisions: detection runs Django-side against the snapshot table because the customer comms need Django machinery (templates, MessagingRecord dedupe, notification settings) – no new service-to-service auth surface. The warning→critical gap is deliberately the customer's redemption window; there is no countdown timer. The Slack webhook that used to live here was removed after review discussion in favor of dogfooding our own CDP (#74558).

---

### DIFF
--- a/frontend/src/lib/components/NotificationsMenu/NotificationRow.tsx
+++ b/frontend/src/lib/components/NotificationsMenu/NotificationRow.tsx
@@ -69,6 +69,10 @@ export const REALTIME_NOTIFICATION_TYPE_META: Record<string, { label: string; de
         label: 'Subscription suggestions',
         description: 'When PostHog suggests subscribing to a dashboard you keep coming back to',
     },
+    email_reputation: {
+        label: 'Email reputation',
+        description: 'When workflow email bounce or spam complaint rates put your email sending at risk',
+    },
 }
 
 export function NotificationTitle({
--- a/frontend/src/lib/components/NotificationsMenu/notificationToasts.tsx
+++ b/frontend/src/lib/components/NotificationsMenu/notificationToasts.tsx
@@ -35,6 +35,7 @@ const NOTIFICATION_TYPE_ICONS: Record<string, { Icon: ComponentType<{ className?
     web_analytics_digest: { Icon: IconPieChart, color: 'text-primary' },
     achievement_unlocked: { Icon: IconStar, color: 'text-warning' },
     subscription_nudge: { Icon: IconBell, color: 'text-primary' },
+    email_reputation: { Icon: IconWarning, color: 'text-danger' },
 }
 
 export function getNotificationIcon(notificationType: string, className: string = 'size-5'): JSX.Element {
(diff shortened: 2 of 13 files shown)


meikelmosby:
Closing: we're switching the reputation architecture to AWS SES tenant management. Sends will carry `TenantName` per team, AWS's reputation findings (LOW/HIGH, with remediation text) and `Standard` reputation policy take over judging and auto-pausing, and EventBridge events will drive customer comms instead of our calculated state transitions. The warning/critical templates and `send_email_reputation_degraded` task on this branch are good salvage material for the AWS-events-triggered comms — the trigger side is what changes. Strategy write-up is with the team.

PR #73443, opened 2026-07-24
fix(surveys): reduce public config payload

marandaneto:
## Problem

Public SDK configuration responses include fields that current SDK runtimes do not consume. These fields increase every remote-config response, and stale HyperCache entries can continue exposing the older response shape after the Python producer changes.

## Changes

- Remove the redundant top-level project token and `survey_config` from public remote configuration.
- Remove `base_language` and `isNpsQuestion` from the public SDK survey projection while preserving them in authenticated survey APIs.
- Restrict cached survey Action conditions to `id`, `name`, and `steps` before serving them publicly.
- Sanitize existing cached responses in HyperCache and feature-flags response paths, covering `/array`, `/flags`, `/decide`, and `/api/surveys`.
- Remove the project token from feature-flags fallback configuration.
- Preserve fields required by older posthog-js versions and document their compatibility ranges.

- Extended `test_sdk_payload_strips_non_runtime_question_fields` to guard the exact `/api/surveys` cache shape and the all-invalid question-translation branch. This catches `survey_config` being reintroduced through `/api/surveys` and invalid translations surviving when no valid language remains.
- `ruff check` and `ruff format --check` on the changed Python files.
- `uv run mypy --cache-fine-grained .` passed across 16,932 source files. The updated test file also passed a targeted MyPy run.
- `cargo fmt --all -- --check` passed.
- `SQLX_OFFLINE=true cargo clippy -p hypercache-server -p feature-flags --all-targets -- -D warnings` passed.
- `cargo test -p hypercache-server --lib` passed 49 tests. These cover public config and survey sanitization, including stale cached payloads.
- `cargo test -p feature-flags --lib handler::config_response_builder::tests` passed 15 tests. These cover cached config sanitization and quota behavior.
- `cargo test -p feature-flags --lib api::types::tests::test_fallback_config_excludes_request_token` passed.
- `hogli build:openapi` completed with no generated diff.
- `hogli ci:preflight --fix` reported no hard failures.
- The targeted Django test could not initialize its database because the local Postgres service was unavailable. No manual testing was performed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation update is required. These are undocumented implementation fields in public SDK configuration responses, and authenticated API behavior remains unchanged.

## 🤖 Agent context

I directed Pi to audit the public response against the PostHog app and locally cloned SDK repositories before implementing the cleanup. Pi used worker subagents for independent runtime-consumer and blocker reviews. Skills used were `/survey-sdk-audit`, `/improving-drf-endpoints`, `/implementing-mcp-tools`, `/writing-tests`, `/run-posthog`, `/running-ci-preflight`, `/resolve-conflicts`, `/address-pr-comments`, and `/pr`.

The branch was merged with the latest `master`. Master independently added the fresh-payload Action projection, so the conflict was resolved by keeping its regression coverage and removing this branch's duplicate serializer logic. After CI identified two patch-coverage gaps, I verified that posthog-js reads only `response.surveys` from `/api/surveys` and extended the existing public payload test to cover both gaps.


marandaneto:
@PostHog/team-surveys @PostHog/team-client-libraries can you pls double check this
those properties are not used in the sdk nor i didnt find any usage in the past, nor the posthog app
some of them are used but not via the remote config APIs but rather the authenticated APIs, so its just leaking where it should not unless i missed something

marandaneto:
the only one i am unsure is

> Remove the project token from feature-flags fallback configuration.

since its there forever but i didnt find any usage either
cc @PostHog/team-feature-flags i guess?

ioannisj (approved):

lucasheriques (approved):
looks good from surveys side

gustavohstrassburger (approved):

PR #73454, opened 2026-07-24
fix(data-imports): recognize DeltaError as a transient object-store error

Gilbert09:
## Problem

Error tracking surfaced a `DeltaError` from `run_pre_write_defensive_compact` in `products/warehouse_sources/backend/temporal/data_imports/pipelines/common/extract.py`, during a Postgres incremental sync's pre-write maintenance pass ([issue](https://us.posthog.com/project/2/error_tracking/019f9359-cd9b-72f0-b52e-963ba8de3dfd)).

The call path: `run_pre_write_defensive_compact` -> `DeltaTableHelper.run_maintenance` -> `vacuum_if_stale` -> `vacuum_table` -> delta-rs's `table.vacuum()`, which needs to read a `_delta_log` manifest from S3. S3 responded with its `SlowDown` throttling error (`503 Service Unavailable`), and after exhausting delta-rs's internal retries the failure surfaced as a `DeltaError`.

This module already has a classifier for exactly this situation, `is_transient_object_store_error`, used by `run_pre_write_defensive_compact` to log a warning instead of capturing an error-tracking issue for a self-healing S3 blip (the docstring explicitly calls out that a maintenance failure must never block the actual sync). The problem: the classifier only recognized `OSError`, because that's the exception type `DeltaTable.is_deltatable()` raises for this same class of object-store error. Other table-level operations — `vacuum()`, `optimize.compact()` — wrap the identical underlying error text in `deltalake.exceptions.DeltaError` instead. Same transient blip, different Python exception type depending on which delta-rs entry point hit it, so the vacuum path's throttling error fell through to `capture_exception` and generated a noisy error-tracking issue for something that isn't a bug.

This is best-effort maintenance code, so the sync itself was never at risk — this only affects whether a self-healing throttling blip gets reported as a defect.

## Changes

- Broadened `is_transient_object_store_error` in `delta_table_helper.py` to also match `deltalake.exceptions.DeltaError` instances, not just `OSError`, against the same substring list.

Extended the existing parameterized `test_logs_transient_object_store_error_without_capturing` in `test_extract.py` with a `DeltaError`-wrapped case carrying the same S3 SlowDown error text — this is exactly the shape of the bug this PR fixes: without the classifier change, this case would fall through to `capture_exception` like the plain `RuntimeError` case in `test_swallows_maintenance_failure` above it.

Ran:
- `uv run pytest products/warehouse_sources/backend/temporal/data_imports/pipelines/common/test/test_extract.py -k TestRunPreWriteDefensiveCompact` — 8 passed
- `uv run mypy --cache-fine-grained .` (repo-wide, as CI runs it) — clean
- `ruff check --fix` / `ruff format` — clean
- `hogli ci:preflight --fix` — 0 failures

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

## 🤖 Agent context

Triaged directly from the linked error-tracking issue via PostHog's error-tracking MCP tools (`query-error-tracking-issue`, `query-error-tracking-issue-events`), which surfaced the full stack trace and confirmed the failure originates in the vacuum step of pre-write maintenance, not the Postgres source connector. Confirmed `deltalake.exceptions.DeltaError` isn't a subclass of `OSError` directly against the installed `deltalake` package before concluding this was the gap.

Checked for duplicate open PRs by exception type, message phrase, module path, and the maintainer's own open PR queue. Found two related-but-non-overlapping fixes in the same file: #73386 (classifies `django.db` connection errors as transient, a different exception source) and #73407 (adds S3 `SlowDown` text to the transient-error substring list, for the `reset_table`/`_purge_s3_prefix` path, still gated on `isinstance(error, OSError)`) — neither touches the `OSError`-vs-`DeltaError` gap this PR closes.

Invoked `/writing-tests` before extending the regression test.

---


Gilbert09:
Thanks — this finding is on `products/streamlit_apps/backend/presentation/views.py`, which this PR doesn't touch. It surfaced only because that file entered the branch's view via a routine master merge (the Streamlit apps feature), not from any change here. This PR is scoped to broadening the transient object-store error classifier in `delta_table_helper.py`, so the Streamlit access-control finding is out of scope for it and should be tracked against the code that introduced it.

Gilbert09:
### CI status: blocked by an unrelated master regression, not by this PR

This branch is rebased cleanly on top of latest `master`, approved, and mergeable. The only red check is one Backend CI shard:

> `Django tests – Core (persons-on-events off, events schema legacy) … (9/19)`

The root failure is entirely unrelated to this PR (which only touches `products/warehouse_sources/.../delta_table_helper.py` and its test):

- `posthog/hogql/test/test_parser_cpp_json.py::TestParserCppJson::test_deeply_nested_input_does_not_stack_overflow` fails with `AssertionError: SyntaxError not raised : expr` (the C++ parser's depth cap doesn't fire for the bracket-nesting case), which drops the shard's Postgres connection and cascades into ~381 `OperationalError: the connection is closed` failures in every subsequent test's `setUp`.
- The Rust-backend equivalent (`TestParserRustJson`) passes.

This reproduces **identically on `master` HEAD**, so it isn't introduced here — it was introduced by #73958 (`fix(hogql): cap recursion depth in the C++ parser`), which merged ~1h ago: the Backend CI run for the commit immediately before it was green on this shard, and every master run since has failed on this same test. My own test shard (`Django tests – Temporal …`) is green, and a re-run of the failing shard on this branch failed identically, so it's deterministic rather than flaky.

Nothing to fix in this PR. Holding here until the parser regression is resolved on `master`; I'll re-merge `master` and re-verify CI as soon as it's green again.

Gilbert09:
Update: the `master` regression is resolved — the C++ parser was bumped in #75473 (`fix(hogql): bump hogql-parser to 1.3.83`), and the previously-failing `Core … (9/19)` shard now passes on `master`. I re-merged latest `master` into this branch (server-side, two-parent merge — only this PR's two files appear in the diff), and CI is now fully green (141 pass, 0 fail). Branch is approved, mergeable, and up to date.

PR #73466, opened 2026-07-24
feat(early-access-features): allow assigning features to a person or role

joethreepwood:
> [!NOTE]
> The posthog.com roadmap consumes the public `assignee` field added here: [posthog.com#18882](https://github.com/PostHog/posthog.com/pull/18882). That PR falls back to its hard-coded team map until this one merges and deploys, so there is no ordering constraint — but the roadmap only reflects in-app assignments once this ships.

## Problem

Early access features have no ownership concept. Nothing records who created a feature or who is responsible for it, so surfaces that want to show an owner (like the public roadmap on posthog.com, which currently renders these features as "Unassigned" via a hand-maintained override map) have nothing to read, and there is no way to hand a feature to a person or team in the app.

## Changes

- `EarlyAccessFeature` gains `created_by`, `assigned_user`, and `assigned_role` (role = the same "person or role" assignee convention error tracking uses). New migration in `products/early_access_features/backend/migrations/0007`.
- The authenticated API exposes read-only `created_by` and a read/write `assignee` field shaped `{"type": "user" | "role", "id": ...}` or `null`, validated against the organization's members and roles. On create, `created_by` is set to the requesting user and the assignee defaults to the creator unless an explicit `assignee` (including `null`) is sent.
- The **public** `/api/early_access_features/` endpoint (project-token auth, consumed by posthog-js and posthog.com's roadmap) now serves the assignee as a display name only – `{"type": "user" | "role", "name": "..."}` – no user IDs or emails. The public view and the flag-list `features` prefetch join the assignee relations to avoid N+1s.

> [!WARNING]
> This makes assignee display names (a member's first/last name, or a role name) visible to anyone with the project token, for every project. That felt like the minimal exposure that still lets consumers attribute features to owners, but it deserves a deliberate review. Happy to gate it or strip user names if we'd rather not expose them.

- The early access feature scene gets an "Assignee" control in the scene panel, reusing the error tracking `AssigneeSelect` components (cross-product import, same as `products/dashboards` already does). Changes save immediately in view mode and with the form in edit mode.

Automated tests added in `products/early_access_features/backend/test/test_early_access_feature.py`:

- create sets `created_by` and defaults the assignee to the creator (the regression: creating silently leaves features ownerless)
- create with an explicit role assignee, and with an explicit `null` assignee (stays unassigned rather than falling back to the creator)
- update to another user, to a role, and back to `null`
- rejects users and roles from other organizations, and malformed payloads (wrong type, missing id, non-dict, invalid UUID)
- the public endpoint serves assignee display names for user and role assignments without per-feature queries (`assertNumQueries(2)`)

I (Claude, running in PostHog Code) could not run the Django test suite or frontend typechecking in this sandbox – no database or node_modules – so CI is the first full run. The postgres query snapshots (`.ambr`) for the public endpoint will need regenerating since the model gained columns and the view gained joins – expecting the snapshot-update automation to handle that. `ruff check`/`ruff format` and `oxfmt` pass on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude via PostHog Code, directed by Joe. Decisions along the way: reused error tracking's `{type, id}` user-or-role assignee convention instead of inventing a new shape; stored the assignment inline on the model (two nullable FKs) rather than error tracking's separate assignment table, since there is exactly one assignee and no merge semantics here; imported the error tracking `AssigneeSelect` components rather than copying them into the product (the conversations product copied, dashboards imports – importing keeps the diff small). The `assignee` write path goes through `initial_data` like the existing `payload` field, since both diverge from the model columns. The public endpoint intentionally serves only a display name (see warning above); the internal API serves `{type, id}` for the picker UI.

---


joethreepwood:
Closing in favor of #73523, which restacks these changes on top of #72993 (`posthog-code/eaf-created-by-column`). Both PRs added `created_by` to `EarlyAccessFeature` with conflicting `0007` migrations; the stacked PR keeps #72993's `created_by` (with its backfill and lock-free FK) and layers only the assignee work on top as migration `0009`.

PR #73470, opened 2026-07-24
chore(aio): migrate prompts frontend to generated API types

jurajmajerik:
## Problem

The prompts frontend used handwritten API types in `frontend/src/types.ts` and manual `api.llmPrompts.*` wrappers in `lib/api.ts`, duplicating what the OpenAPI codegen already produces. Flagged in review of #72266.

## Changes

- All prompt API calls (prompt logics, list scene, playground) now use the generated functions (`llmPromptsList`, `llmPromptsCreate`, `llmPromptsNameRetrieve`, `llmPromptsNamePartialUpdate`, `llmPromptsNameArchiveCreate`, `llmPromptsNameDuplicateCreate`, `llmPromptsResolveNameRetrieve`), following the pattern the label mutations already used.
- Deleted the handwritten `LLMPrompt`, `LLMPromptPublic`, `LLMPromptVersionSummary`, `LLMPromptResolveResponse` types, the `api.llmPrompts` namespace, and the `ApiRequest` llmPrompt helpers.
- Added `@extend_schema(request=None, responses={204: None})` to the archive action, so the generated archive function no longer asks for a request body it does not need.
- New `products/ai_observability/frontend/prompts/types.ts` derives `LLMPrompt` and `LLMPromptVersionSummary` from the generated schemas with two documented overrides, applied by casts at the fetch boundaries.

Design notes:

- The two overrides exist because (1) the API types prompt payloads as `unknown` (SDKs may store JSON) while this app only reads and writes strings, and (2) the generated `UserBasicApi` types `hedgehog_config` as nullable, which app components like `ProfilePicture` do not accept. Same pattern as `ScannerCreatedBy` in replay_vision.
- No user-facing change. Generated functions go through the same api mutator, so cookies, CSRF and error handling are unchanged.

Existing jest suites are the safety net: all 32 tests in `products/ai_observability/frontend/prompts/` pass, with the logic test mocks switched from `api.llmPrompts` spies to generated-function mocks. TypeScript check passes. I did not manually test the UI, and did not run the (very slow) playground jest suite locally, CI covers it.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed, internal refactor.

## 🤖 Agent context

Claude Code did the migration, invoking /adopting-generated-api-types and /improving-drf-endpoints. First tried using the generated types directly everywhere; that failed on the two shape mismatches above, so it fell back to the product-local derived types with boundary casts.

Follow-up:
- The list endpoint query serializer is missing `order_by`, so the generated `LlmPromptsListParams` does not declare it (runtime is unaffected, the param still passes through).


Radu-Raicea (approved):
Couple of fixed for CI

PR #73471, opened 2026-07-24
feat(tracing): add distributed tracing usage report counters

jonmcwest:
## Problem

Billing needs usage meters for the Distributed Tracing (APM) product: `apm_tracing_bytes_in_period`, `apm_tracing_spans_in_period`, and `apm_tracing_mb_in_period`, mirroring the logs product's bytes/records/MB triple.

Like logs, tracing has an ingestion consumer (`TracesIngestionConsumer`, a subclass of the logs consumer) that emits pre-aggregated `app_metrics2` counters under `app_source='traces'`, so the counters can be computed exactly the way the logs ones are.

## Changes

- Adds the three counters to `UsageReportCounters` and wires them through `_get_all_usage_data`, `_get_team_report`, and `has_non_zero_usage` (org-level rollup is generic and needs no change).
- Adds `get_teams_with_apm_tracing_usage_in_period`, a single `app_metrics2` pass over `app_source='traces'` returning both bytes (`bytes_ingested`) and spans (`records_ingested`) per team, matching the logs queries. A span is one record, so `records_ingested` is the span count.
- Registers a `QuerySpec` in the Temporal usage-report registry so the workflow path fetches the same data.
- Adds `Product.TRACING` to the query-tagging enum.

> [!NOTE]
> This intentionally ships tracking only, no billing surface. No `USAGE_TYPES` entry is added, since the billing service does not yet accept `apm_tracing_mb_in_period` as a usage type; the counters land in the usage report (billing ignores unknown counter fields) so usage data starts accumulating now. A follow-up adds the billing usage-view entry once `billing/types/usage.py` accepts the type.

Quota limiting is intentionally out of scope. It can follow once billing has tracing limits.

- New parameterized test `test_apm_tracing_usage_metrics` seeds `app_metrics2` rows under `app_source='traces'` and asserts the three counters at org and team scope. The regressions it catches that no existing test does: dropping the `app_source` filter (a seeded `app_source='logs'` row with the same metric name must not leak into the tracing counters), plus the decimal-MB floor and the field wiring into the org report.
- Ran (I, Claude, ran these): the new tests plus the full `TestHogFunctionUsageReports` class (9 passed), the Temporal usage-report suite `posthog/temporal/tests/usage_report/` (64 passed, its registry tests derive coverage from `_get_team_report`'s AST so they validate the new `QuerySpec` wiring), repo-wide `mypy --cache-fine-grained .` (clean), and the frontend typescript check (only pre-existing failures in `products/tracing/frontend/components/Comparison/CompareMenuButton.tsx`, untouched here).
- No manual testing against live span data.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed. Internal usage-report plumbing with no user-facing workflow change.

## 🤖 Agent context

Authored by Claude (PostHog Code session) at Jon's direction to mirror the logs usage meters for Distributed Tracing. Skills invoked: /writing-tests, /polish (simplify + /code-review at high effort, fixes applied). An earlier revision aggregated `trace_spans_distributed` directly under the mistaken belief no `app_metrics2` emitter existed for traces; Jon pointed out the traces ingestion consumer, and the query was rewritten to read `app_metrics2` `app_source='traces'` like logs (which also removed a fragile dependency on the trace spans Kafka MV's header semantics and its guard test). Deviations from the initial plan: the `USAGE_TYPES` entries were dropped entirely after checking the billing repo does not yet accept `apm_tracing_mb_in_period` (this PR is tracking-only); user-facing copy says "Distributed tracing" per Jon's naming direction, while field names keep the billing-defined `apm_tracing_*` prefix.

---


posthog/tasks/usage_report.py
```diff
@@ -2922,6 +2975,9 @@ def _get_team_report(all_data: dict[str, Any], team: Team) -> UsageReportCounter
         android_logs_records_in_period=all_data["teams_with_android_logs_records_in_period"].get(team.id, 0),
         flutter_logs_records_in_period=all_data["teams_with_flutter_logs_records_in_period"].get(team.id, 0),
         ruby_logs_records_in_period=all_data["teams_with_ruby_logs_records_in_period"].get(team.id, 0),
+        apm_tracing_bytes_in_period=all_data["teams_with_apm_tracing_bytes_in_period"].get(team.id, 0),
+        apm_tracing_spans_in_period=all_data["teams_with_apm_tracing_spans_in_period"].get(team.id, 0),
+        apm_tracing_mb_in_period=int(all_data["teams_with_apm_tracing_bytes_in_period"].get(team.id, 0) // 1_000_000),
```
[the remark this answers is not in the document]
jonmcwest: This is intentional and mirrors the existing, live `logs_mb_in_period` meter exactly: bytes are floored to whole MB per team (`bytes // 1_000_000`) and then summed at the org level by the generic rollup. That same per-team-floor behavior is documented on `UsageReportCounters`, so tracing stays consistent with logs.

The precision that matters is preserved: `apm_tracing_bytes_in_period` is carried at the org level as raw, unfloored bytes (the generic rollup sums the byte fields), so the exact org byte total is always available. `apm_tracing_mb_in_period` is a floored convenience field, not the source of truth for the byte count.

Also, as noted in the PR description, this ships tracking-only — there is no `USAGE_TYPES` entry and billing does not consume `apm_tracing_mb_in_period` yet — so there is no billable meter being underreported here. When billing does adopt a tracing meter, it can bill off the exact `apm_tracing_bytes_in_period` total, exactly as the logs pattern already allows.

posthog/temporal/usage_report/queries.py
```diff
@@ -520,6 +521,16 @@ class QuerySpec:
             "90d": "teams_with_logs_retention_90d_bytes_in_period",
         },
     ),
+    # ---- ClickHouse: APM tracing ---------------------------------------------
+    QuerySpec(
+        name="apm_tracing_usage",
+        fn=get_teams_with_apm_tracing_usage_in_period,
+        output="multi",
+        multi_keys_mapping={
```
pawel-cebula, line 529: Small test coverage thought: the legacy test covers the counters, but Temporal currently exercises this multi-output spec only with empty rows, so a `bytes`/`spans` mapping swap would still pass. Could we seed non-zero `apm_tracing_usage` in one of the aggregate/parity tests and assert bytes, spans, and MB reach the org report?

pawel-cebula (approved):
Looks good, just one small test suggestion.

posthog/temporal/usage_report/queries.py
jonmcwest: Good catch. Added it to the Temporal aggregate activity test (`test_aggregate_writes_chunks_and_manifest`), which runs the real fan-out end-to-end. It now seeds non-zero `apm_tracing_usage` (`bytes: 2_500_000`, `spans: 42` for team A) and asserts all three counters reach the org report: `apm_tracing_bytes_in_period == 2_500_000`, `apm_tracing_spans_in_period == 42`, and `apm_tracing_mb_in_period == 2` (the floored bytes), plus zeros for the other org. The values are deliberately distinct, so a `bytes`<->`spans` mapping swap in the `QuerySpec` fails these assertions instead of passing silently. Commit fa4140b.

jonmcwest: Done. Non-zero `apm_tracing_usage` is now seeded and asserted in two places:

- The **aggregate-activity** test (`test_aggregate_activity.py`) seeds distinct bytes/spans (`2_500_000` bytes, `42` spans) and asserts `apm_tracing_bytes_in_period`, `apm_tracing_spans_in_period`, and the floored `apm_tracing_mb_in_period` (= 2) reach the org report — so a bytes↔spans mapping swap now fails.
- The **parity** test (`test_parity.py`) additionally covers the multi-team org rollup: org_a sums a 2 MB team and a sub-MB team that floors to 0, so its org MB is 2 (not `floor(org total)`). The single-team aggregate test can’t distinguish per-team flooring from org-total flooring; this pins the intentional per-team floor across both the Celery and Temporal paths.

PR #73486, opened 2026-07-24
feat(conversations): identify who failed to send us a support ticket

christiaan-ph:
## Problem

When someone fails to send us a support ticket we want to know who, so we can reach out to them. Today we can't. The `support ticket send failed` event fires from the widget endpoint via `report_team_action`, which captures with `team.uuid` as the distinct_id and attaches org and project only as groups, so nothing on the event names the person who just failed to reach us.

Raised from a support thread investigating a stream of these events.

## Changes

Put the submitter's `distinct_id` on the failed-send event, and a numeric `team_id` alongside it.

The endpoint serves two populations. Support tickets sent to PostHog itself come from PostHog users, and their `distinct_id` is a person in our own project, already the thing every in-app capture carries. Tickets sent through a customer's own widget embedded on their own site come from that customer's end users, who are not ours to identify or contact. Only the first is in scope here, so the submitter is gated on `CONVERSATIONS_INTERNAL_SUPPORT_TEAM_ID` and every other team keeps exactly the counter it had before, with no identifier attached.

**That setting is unset by default and has to be configured per deployment before this does anything.** `posthoganalytics.api_key` is assigned unconditionally in `posthog/apps.py`, so `report_team_action` reaches our own Cloud project from any instance that hasn't set `OPT_OUT_CAPTURE`, self-hosted included. Defaulting it to 2 the way `GROWTH_ENRICHMENT_INTERNAL_TEAM_ID` does would mean a self-hosted install treats its own second project as ours and ships that project's end users to us. Unset means no submitter is ever attached, so the wrong answer is silence rather than someone else's data.

`widget_session_id` is deliberately not included, for either population. It's the credential the widget ticket and message endpoints authorize on, and the common rejection path (a valid session with an empty or over-long message) carries a live one, so putting it on the event would pair a working credential with the `ticket_id` it unlocks. The submitter's `distinct_id` is what identifies the person to follow up with, and it does that without carrying an access token.

`organization_id` is not added either: `groups(team=team)` in `posthog/event_usage.py` already attaches it as a group key, as it does the project.

The submitter's `distinct_id` is length-bounded before it goes on the event. This endpoint is public and unauthenticated, and the failure path runs before validation has established anything about the value.

Automated, run locally against the dev stack (`hogli test products/conversations/backend/api/tests/test_widget.py`, 53 passed).

- `test_ticket_sent_to_posthog_identifies_its_submitter` is parameterized over both auth modes and asserts the event names the submitter and the team. The verified-identity case covers the `identity_distinct_id` fallback, which the widget_session case doesn't reach.
- `test_ticket_not_sent_to_posthog_identifies_nobody` is the guard on the scope, covering another team on both auth modes plus the unconfigured case. Verified it bites by deleting the team check, which fails it, so a change that widens this beyond our own tickets can't pass silently.
- Both assert no property contains the `widget_session_id`, so a future change that reintroduces it fails.
- `test_rejected_send_bounds_an_overlong_submitter_id` covers the length cap. Regression it catches: an unbounded attacker-supplied string inflating an event from a public endpoint.

Also verified `uv run mypy --cache-fine-grained .` repo-wide (clean, 17197 files), `hogli ci:preflight` (no failures), and `hogli build:openapi` (no diff, so no schema change).

No manual browser testing: the change is server-side telemetry on an existing failure path with no user-visible surface.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Originally authored by the PostHog Slack app from a support thread, then reworked twice: once after automated security review, and once after clarifying that the goal is tickets sent to PostHog rather than tickets sent through customers' widgets.

- An earlier revision put the raw `widget_session_id` and an unkeyed truncated SHA-256 of the `distinct_id` on the event, for every team. Three review bots flagged it. Each finding was checked against the code rather than taken at face value: `widget_session_id` is documented as the access control mechanism at the top of `widget.py` and enforced in all four widget views, and the revision's own test showed a valid session reaching the failure path.
- One bot argued the failure path is only reachable with garbage values because validation had already failed. That is wrong: a valid session plus an empty message fails validation while carrying a live credential, which is the common case.
- Two bots suggested keying the digest with `team.secret_api_token`. Not taken, and moot now. That is the customer's HMAC identity secret, so telemetry would break on rotation and an auth secret would be mixed into an analytics path.
- A revision in between retained rejected submissions in a team-scoped table so a customer's end user stayed in the customer's own data. Dropped once the scope was confirmed as our own tickets only: with the identity limited to our own users there is nothing that can't live on the event, and the table would have been written for exactly one team.
- The discriminator on this event is left as `channel_source` even though the four client-side emitters of the same event name use `channel`. Renaming it would change a property existing queries and alerts may filter on.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0AMXLYHZ8U/p1784890519705819?thread_ts=1784890519.705819&cid=C0AMXLYHZ8U)*


christiaan-ph:
Closing this as overtaken by the client-side path.

When this was opened, the only telemetry for a failed support submission was the widget endpoint's `support ticket send failed`, which captures with `team.uuid` and so names nobody. Since then the client-side event has grown into the better answer: it captures from posthog-js with the submitter's own distinct_id, and carries the surface, the failure reason and a preview of what they wrote, so a Slack alert on it can link straight to the person and their replay. #78928 gave it its own name, `support ticket send blocked`, so the two are now separable in queries. Anything queried before 2026-08-06 mixes them under the old name.

That leaves the server-side event covering only submissions the client capture never made: the widget on a page without our app's support form, capture blocked in the browser, or the tab closing first. For tickets sent to PostHog itself that is a handful a month, and this PR would only have identified them for one team behind a setting that has to be configured per deployment. Not enough to carry the extra surface area.

Not proposed as follow-up work, just noting where it landed: the internal Slack destination on `support ticket send failed` renders `{person.name}` and a replay link, neither of which can resolve for an event keyed on a team uuid. If that alert stays enabled alongside the one on `support ticket send blocked`, it will keep posting with those fields empty.

PR #73492, opened 2026-07-24
feat(auth): org setting to enforce verified email domains

a-lider:
## Problem

Enforce-SSO means "emails on this domain must use this provider", not "only these domains may be members".

`random@gmail.com` can be invited into an SSO-enforced org, sign up however they like, and become a full member. 

An earlier attempt (#73170) derived the membership restriction from the existing enforce-SSO config. Security review pushed back, it breaks the legitimate "SSO for employees + a couple of external contractors" setup and enforced orgs already have outside-domain members who'd be affected. So: a **separate, opt-in setting**.

## Changes

New org-level setting **`enforce_verified_domains`** (default `off`, nullable, similar to `enforce_2fa`). When `on`, an email outside the org's verified domains is not admitted to that organization — at login, at signup, at invite creation, and on every subsequent request. When off — everything behaves exactly as today.

**Backend**

- `Organization.enforce_verified_domains` + migration. `posthog_organization` is a hot table: the migration is a metadata-only nullable `ADD COLUMN` and is acknowledged in `hot_table_acknowledged_migrations.txt`
- One predicate everything shares: `OrganizationDomain.objects.is_email_blocked_by_domain_enforcement(email, org)`, wrapped by `posthog/helpers/verified_domain_enforcement.py` for the access paths
- Enforcement points, all returning code `verified_domain_required`:
  - **invite creation** — single, bulk, and delegation invites (`posthog/api/organization_invite.py`)
  - **invite acceptance** — social pipeline and password signup (`posthog/api/signup.py`), which also catches invites created before enablement
  - **login** — password, passkey, SSO, and email verification, so no session-minting path is left open
  - **every org-scoped API request** — `VerifiedDomainEnforcementPermission` in the shared permission path, checked against the URL-resolved organization, for every user-bound authenticator (session, personal API keys, OAuth, JWT)
  - **every session-authenticated request** — `SessionAuthentication`, next to the existing 2FA enforcement, as the app-level cutoff for live sessions

**Scoped to one organization, enforced at two layers.** Checking every org a member belongs to at login would let one org lock a contractor out of all the others, including their own, with no self-service way back — so login is refused only when no organization admits the member's email, and when their current one stops admitting them they're moved to one that does. But the current organization is a UI preference routing never validates, so the authoritative boundary is the permission layer: every org-scoped request is checked against the URL-resolved organization, which closes direct API access from a session parked in a permitted org and covers personal API keys, OAuth, and views declaring their own `authentication_classes`. A session-auth check against the current organization sits on top as the app-level cutoff, so turning the setting on takes effect on sessions that are already live. Both layers short-circuit on the org's own flag and reuse rows the request already loaded, so orgs without enforcement pay no extra query.

**Lockout guards.** Enabling is rejected when the acting admin's own email wouldn't be admitted (this also covers an empty allow-list), deleting a verified domain is rejected when it would leave the acting admin blocked (which covers deleting the last one), and disabling is never gated on the entitlement, so an org that downgrades can always let its members back in.

**Frontend**

- "Domain enforcement" section on Settings → Organization → Authentication domains & SSO, below the domains table. Disabled until the org has at least one verified domain and the admin's own domain is among them.
- A blocked member gets a clear message and a way out instead of a silently broken app.

domain-enforcement-final

> [!WARNING]
> With the setting **on**, existing members whose email is outside the verified domains lose access to that organization on their next request — that's the intended strictness, the setting is explicit opt-in. Members of other organizations keep those.

Known follow-ups, deliberately not in scope: decoupling organization access from account login is a real architecture gap the reviewers rightly named (the permission-layer check is a step in that direction, not the destination), and the legacy `TeamInviteSurrogate` join path is unchecked but unreachable without a Django-admin-set `Team.signup_token`.

Tests added/updated for each enforcement point, plus `ruff` and repo-wide `uv run mypy --cache-fine-grained .` clean.

- `posthog/api/test/test_authentication.py` - password login blocked/unblocked as the member's domain gets verified; a multi-org member lands in an org that admits them instead of being refused; a live session is cut off when enforcement is switched on.
- `posthog/api/test/test_webauthn.py` - passkey login is blocked and mints no session.
- `posthog/api/test/test_user.py` - email verification stops auto-login for a blocked member.
- `posthog/api/test/test_organization.py` - enabling is refused when it would block the acting admin.
- `posthog/api/test/test_authentication.py` - a cross-org member's session and their personal API key are both refused against the enforcing org's projects and accepted against a permitted org's.
- `posthog/api/test/test_signup.py`, `test_organization_invites.py`, `posthog/models/test/test_organization_domain.py` - SSO login and invite paths, as before.

> [!NOTE]
> The review round (passkey/verify_email gap, current-org scoping, per-request enforcement, lockout guards) was written without a local Postgres, so the Python suite has not been executed against it — CI is the first run. The manual smoke below covered the pre-review behavior (login/signup/invite blocking, toggle round-trip), not the per-request enforcement or the organization move.

Earlier live smoke against the local stack: with enforcement on — password login of an outside-domain member → 400 `verified_domain_required`; verified-domain member → success; accepting a pre-existing outside-domain invite with a password → 400; invite to gmail via the authenticated API → 400 with no invite row; invite to a verified domain → 201. Toggle round-trips verified in the settings UI, including the disabled state for non-admin members.

## Automatic notifications

- [ ] Publish to changelog?


posthog/api/authentication.py
```diff
@@ -328,6 +328,16 @@ def create(self, validated_data: dict[str, str]) -> Any:
                 code="not_verified",
             )
 
+        # Domain enforcement: an org that requires a verified email domain blocks logins for members
+        # whose email is outside its verified domains. Checked after authentication so the error is
+        # only shown for valid credentials.
+        for organization in user.organizations.all():
+            if OrganizationDomain.objects.is_email_blocked_by_domain_enforcement(user.email, organization):
```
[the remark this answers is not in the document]
ReeceJones: Bot is right here, but this is a general problem with the way we authenticate organizations right now. Ideally org auth is decoupled from user auth. So a user can log into their account. but accessing organizations they are a member of is governed by a separate process.

posthog/api/authentication.py
```diff
@@ -328,6 +328,16 @@ def create(self, validated_data: dict[str, str]) -> Any:
                 code="not_verified",
             )
 
+        # Domain enforcement: an org that requires a verified email domain blocks logins for members
+        # whose email is outside its verified domains. Checked after authentication so the error is
+        # only shown for valid credentials.
+        for organization in user.organizations.all():
```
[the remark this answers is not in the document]
ReeceJones: I think the bot is partially right here. Also be sure to check the webathun sign up registration viewset

posthog/api/signup.py
```diff
@@ -948,6 +969,22 @@ def social_create_user(
...
+    if invite_organization is not None:
+        organizations_to_check.append(invite_organization)
+    if enforcement_email:
+        for organization in organizations_to_check:
+            if OrganizationDomain.objects.is_email_blocked_by_domain_enforcement(enforcement_email, organization):
+                logger.warning(
+                    "social_create_user_blocked_domain_enforcement",
+                    organization=str(organization.id),
+                )
+                return redirect("/login?error_code=verified_domain_required")
```
ReeceJones: Is it possible to abstract this loop into the SQL for the `is_email_blocked_by_domain_enforcement` function by doing a join of some form? This could potentially issue a lot of DB requests since the query is inside a for loop

ReeceJones (approved):
left some comments, but otherwise looks good

posthog/api/authentication.py
```diff
@@ -328,6 +328,16 @@ def create(self, validated_data: dict[str, str]) -> Any:
...
+            if OrganizationDomain.objects.is_email_blocked_by_domain_enforcement(user.email, organization):
+                raise serializers.ValidationError(
+                    "Your organization requires a verified email domain to log in. Please contact your administrator.",
+                    code="verified_domain_required",
+                )
```
Piccirello: **High — passkey login bypasses this check entirely.**

This is one of several paths that mint a session, and it's the only interactive one that got the check. `WebAuthnLoginViewSet.complete` (`posthog/api/webauthn.py`, `permission_classes = [AllowAny]`) calls `login(request, verified_user, backend="posthog.auth.WebauthnBackend")` after axes lockout, `_check_sso_enforcement`, and `is_email_verified_for_login` — and nothing else. `is_email_blocked_by_domain_enforcement` has four call sites in this PR and none of them are there.

Scenario: `contractor@gmail.com` is a member of an org that verifies `acme.com` and has a passkey registered. Passkey registration is `IsAuthenticated`, so this happens while they can still authenticate — the common case for anyone who used the product before the toggle was flipped. Admin enables `enforce_verified_domains`. Password login → 400 `verified_domain_required`. Passkey login → 200, full session, unchanged access.

That makes the PR description's warning ("existing members whose email is outside the verified domains are blocked at their next login") false for that population. Worth noting `webauthn.py` already carries an SSO-enforcement parity check at both registration and login, so this reads as an omission rather than a scoping decision.

Suggest factoring this loop into a helper and calling it from `webauthn.py` before `login()`.

posthog/api/signup.py
Piccirello: **Medium–High (design) — the block's blast radius is every org the user belongs to, not just the enforcing one.**

`organizations_to_check` is the user's full org list, and the same shape exists in `LoginSerializer.create` (`user.organizations.all()`), where any blocking org raises and refuses the whole login.

Scenario: a consultant on `gmail.com` belongs to Org A (client), Org B (another client), and their own org. Org A enables the setting. The consultant is now locked out of PostHog entirely — Org B and their own org included — and can't log in to leave Org A. One tenant's config change produces denial in unrelated tenants, with no self-service remedy.

The safer shape is per-org: deny access *to the enforcing org* (exclude it from accessible orgs / block switching into it) rather than refusing the session. If a global block really is intended, the setting's copy should say so, because admins won't expect their toggle to affect other companies' access.

This is the one finding I'd want settled before merge rather than in follow-up, since it's a design choice rather than a patch.

posthog/api/organization.py
```diff
@@ -312,6 +313,15 @@ def validate_enforce_2fa(self, value: bool | None) -> bool | None:
                 )
         return value
 
+    def validate_enforce_verified_domains(self, value: bool | None) -> bool | None:
```
Piccirello, line 332: **Medium — nothing stops the enabling admin from locking themselves out.**

This validator checks entitlement only. `EnforceVerifiedDomains.tsx` disables the switch when the org has zero verified domains, but never checks the actor's own address.

Scenario: sole owner `founder@gmail.com` verifies `company.com`, flips the switch, and is blocked at their next login by the global check in `LoginSerializer`. No admin is left who can turn it off, so recovery needs PostHog support and direct DB access.

Cheapest fix in this review: reject the update when the acting user's own email would be blocked by the new value. A UI warning listing the members who'd lose access would help too, since the current copy gives no preview of the blast radius.

posthog/models/organization.py
```diff
@@ -235,6 +235,11 @@ class DefaultExperimentStatsMethod(models.TextChoices):
         help_text="When True, in-app callouts inviting members to enable AI training are shown.",
     )
     enforce_2fa = models.BooleanField(null=True, blank=True)
+    enforce_verified_domains = models.BooleanField(
+        null=True,
+        blank=True,
+        help_text="When True, logins, signups, and invites for this organization are restricted to email addresses on its verified domains.",
+    )
```
Piccirello, line 242: **Medium — enabling the setting doesn't invalidate authentication material that already exists.**

Flipping this flag revokes nothing. An outside-domain member with a live session keeps full access until it expires, and a personal API key issued before enablement keeps working indefinitely — PAK auth has no org-level gate. Same for OAuth tokens.

Related, lower priority: the partner deep-link logins in `ee/partners/stripe/api/provisioning/login.py` and `ee/api/agentic_provisioning/views.py` both mint sessions after only an `is_email_verified` check. Those bypass SSO enforcement today as well, so that part is pre-existing rather than introduced here.

`revoke_other_sessions` already exists in `posthog/api/user.py`, so revoking blocked members' sessions when the flag is enabled is a small addition — and it's what makes the setting mean what the help text says.

posthog/models/organization_domain.py
```diff
@@ -167,6 +167,22 @@ def get_sso_enforcement_for_email_address(
...
+        return self.verified_domains().filter(organization=organization, domain__iexact=domain).exists()
+
+    def is_email_blocked_by_domain_enforcement(self, email: str, organization: Organization) -> bool:
+        """
+        Whether a login or join for `email` into `organization` should be blocked: the org requires
+        a verified email domain, and `email`'s domain is not one of the org's verified domains.
+        """
+        if not organization.enforce_verified_domains:
+            return False
+        return not self.is_domain_verified_for_organization(email, organization)
```
Piccirello, line 191: **Low — one more session-minting path that doesn't call this predicate: `verify_email`.**

`UserViewSet.verify_email` in `posthog/api/user.py` is `AllowAny` and calls `login()` after token validation. It explicitly guards SSO enforcement ("don't hand a non-SSO session to an account whose domain enforces SSO") but not domain enforcement.

Reachability is genuinely bounded — `request_email_verification` refuses when `user.is_email_verified and not user.pending_email`, so this needs an unverified account or an in-flight email change. The realistic case is an outside-domain user who signed up via invite before enablement and never verified: they click the verification link and get a session.

Worth closing at the same time as the passkey path, since both are the same missing call.

posthog/api/signup.py
```diff
@@ -922,6 +933,16 @@ def process_social_domain_jit_provisioning_signup(
...
 
 
+def _resolve_invite_organization(invite_id: str) -> Optional[Organization]:
+    """Organization an invite grants access to, or None for legacy team-invite surrogates / missing invites."""
+    try:
+        # nosemgrep: idor-lookup-without-org (invite UUID from server session serves as auth token)
+        invite = OrganizationInvite.objects.select_related("organization").get(id=invite_id)
+    except (OrganizationInvite.DoesNotExist, ValidationError):
+        return None
+    return invite.organization
```
Piccirello, line 948: **Low — the surrogate path the docstring names is a real unchecked join.**

Returning `None` here means no org gets checked, and `process_social_invite_signup` then falls back to `TeamInviteSurrogate`, whose `validate()` unconditionally returns `True` and whose `use()` calls `user.join(organization=...)` with no email check at all. That bypasses both the domain rule and the invite/email match.

Precondition is narrow: it needs a `Team` row with a non-null `signup_token`, and nothing in current code sets one — only Django admin exposes the field. So this is legacy-data-only, not reachable on a fresh org.

Still worth closing rather than leaving implicit, given the docstring already identifies it as a hole. Either check the surrogate's organization the same way, or have the surrogate path refuse when the target org enforces.

posthog/api/organization.py
```diff
@@ -312,6 +313,15 @@ def validate_enforce_2fa(self, value: bool | None) -> bool | None:
...
+        if self.instance and self.instance.enforce_verified_domains != value:
+            if not self.instance.is_feature_available(AvailableFeature.AUTOMATIC_PROVISIONING):
```
Piccirello: **Low — this gates changes in both directions, so a downgraded org can't turn enforcement off.**

`self.instance.enforce_verified_domains != value` is true for `True → False` as well, so an org that loses `AUTOMATIC_PROVISIONING` keeps blocking logins with no way to disable it.

This mirrors `validate_enforce_2fa` exactly, so it's a consistent house pattern — but the consequence differs. 2FA enforcement forces enrollment; this one denies access outright, which turns a billing state change into a lockout.

Suggest gating only the enable direction: `if value and not self.instance.is_feature_available(...)`.

frontend/src/scenes/settings/organization/VerifiedDomains/EnforceVerifiedDomains.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const hasVerifiedDomains = verifiedDomains.some((domain) => domain.is_verified)
+
+    return (
+        <PayGateMini feature={AvailableFeature.AUTOMATIC_PROVISIONING}>
+            <p>
+                Only allow people with an email address on a verified domain to log in or join this organization.
+                Invites to other domains are blocked.
+            </p>
+            <LemonSwitch
+                label="Require a verified email domain to log in"
```
Piccirello: I find this wording slightly confusing because we're not really blocking log in, we're blocking access to this org
```suggestion
                label="Restrict membership to verified email domains"
```

frontend/src/scenes/settings/organization/VerifiedDomains/EnforceVerifiedDomains.tsx
```diff
@@ -0,0 +1,46 @@
...
+        scope: RestrictionScope.Organization,
+    })
+
+    const hasVerifiedDomains = verifiedDomains.some((domain) => domain.is_verified)
+
+    return (
+        <PayGateMini feature={AvailableFeature.AUTOMATIC_PROVISIONING}>
+            <p>
+                Only allow people with an email address on a verified domain to log in or join this organization.
+                Invites to other domains are blocked.
```
Piccirello:
```suggestion
                Only people with an email address on one of your verified domains can be members of this organization. Invites to other addresses are blocked, and existing members outside your verified domains lose access.
```

Piccirello (changes requested):
Added 7 findings from a security review and 2 wording changes

posthog/permissions.py
```diff
@@ -238,6 +239,79 @@ def has_permission(self, request, view) -> bool:
...
+        # Non-user principals (sharing links, project secret keys, internal API) aren't members
+        # and can't be domain-gated.
+        if not isinstance(user, User):
+            return True
+
+        if not OrganizationDomain.objects.is_email_blocked_by_domain_enforcement(user.email, organization):
+            return True
+
+        if is_impersonated_session(request):
+            return True
```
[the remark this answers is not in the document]
a-lider: Re: `is_impersonated_session(request)` vs `request._request` — this is a false positive; the current code is correct.

`is_impersonated_session` only does `hasattr(request, "session") and USER_SESSION_FLAG in request.session`. A DRF `Request` proxies unknown attribute access to the wrapped Django `HttpRequest` via `__getattr__`, so `request.session` already resolves to the real underlying session — passing the DRF `Request` behaves identically to passing `request._request`, and the impersonation check does not fail.

It's the established pattern in this file: sibling permission classes at `permissions.py:411` and `:448` pass the DRF `request` straight into `is_impersonated_session`, as do `posthog/api/oauth/views.py:155` and `ee/api/conversation.py:334`. If the DRF-request form broke impersonation detection, those would already be broken. The explicit `._request` form in `verified_domain_enforcement.py:30` and `two_factor_session.py:220` is just defensive for helpers that may receive a plain `HttpRequest`; it isn't required here. No change needed.

posthog/api/authentication.py
a-lider: Resolved. `WebAuthnLoginViewSet.complete` now calls `_check_domain_enforcement(verified_user)` (webauthn.py:358), via the shared `resolve_login_organization` predicate, before `login()` at line 372. Password, passkey, email-verification, and the social pipeline all route through that one predicate now.

posthog/api/signup.py
a-lider: Resolved. The all-orgs loop is gone — both this social path and `LoginSerializer` now use `resolve_login_organization`, which is scoped to the current org and, when the current org stops admitting the user, moves them to another org that does; login is refused only when *no* org admits them. The invite branch just below checks only the invite's own org. So one tenant enabling the setting no longer locks the user out of unrelated tenants.

posthog/api/organization.py
a-lider: Resolved. `validate_enforce_verified_domains` now rejects enabling when the acting user's own email isn't on a verified domain (`would_block_self`, organization.py:330-336), the frontend disables the toggle with the same reason, and a later commit also refuses deleting the domain that admits the acting admin while enforcement is on.

posthog/models/organization.py
a-lider: Addressed, via per-request denial rather than revocation: the `SessionAuthentication` hook (`enforce_verified_domain`, auth.py:180) cuts live sessions on the next request, and `VerifiedDomainEnforcementPermission` — now appended even to viewsets with custom `dangerously_get_permissions` chains and applied to root org/team/project objects (commit cb4f7394) — gates personal API keys and OAuth per-request against the resolved org. So blocked members lose access on their next request regardless of credential type. `revoke_other_sessions` wasn't used; flagging in case eager session revocation on enable is also wanted.

posthog/models/organization_domain.py
a-lider: Resolved. `UserViewSet.verify_email` now calls `resolve_login_organization(user)` before `login()` (user.py:1114) and returns `requires_login` instead of minting a session when no org admits the user — right next to the existing SSO-enforcement guard.

posthog/api/signup.py
a-lider: Still open, but mitigated. `TeamInviteSurrogate.use()` (signup.py:801) still joins with no email/domain check, and `_resolve_invite_organization` returns `None` for surrogates so the invite-org gate here is skipped. In practice a surrogate-joined blocked user can't use the enforcing org — its enforcement is applied per-request by the session hook + permission — and it stays reachable only via a `Team.signup_token`, which nothing outside Django admin sets. Not blocking, but the specific join isn't refused; a one-line guard in `use()`/`validate()` would close it explicitly.

posthog/api/organization.py
a-lider: Resolved. The validator now short-circuits on `not value` (organization.py:321), so only enabling is entitlement-gated and disabling is always allowed — the `if value and not is_feature_available(...)` shape suggested. A downgraded org can always turn enforcement back off.

frontend/src/scenes/settings/organization/VerifiedDomains/EnforceVerifiedDomains.tsx
a-lider: Applied — the switch label now reads "Restrict membership to verified email domains" (line 48).

frontend/src/scenes/settings/organization/VerifiedDomains/EnforceVerifiedDomains.tsx
a-lider: Addressed with a close reword (lines 44-45): "Only allow people with an email address on a verified domain into this organization. Invites to other domains are blocked, and existing members on other domains lose access." — membership framing instead of "log in," and it now notes existing outside-domain members lose access. Not verbatim, so flag if the original phrasing is preferred.

posthog/permissions.py
```diff
@@ -238,6 +239,81 @@ def has_permission(self, request, view) -> bool:
...
+        if organization is None:
+            return True
+        return self._admits(request, organization)
+
+    def has_object_permission(self, request: Request, view, object: Model) -> bool:
+        if isinstance(object, Organization):
+            return self._admits(request, object)
+        if isinstance(object, Team | Project):
+            return self._admits(request, object.organization)
+        return True
```
Piccirello, line 281: **Medium — a blocked admin can no longer turn the setting off, so there's no escape hatch.**

`PATCH /api/organizations/<id>/` reaches this method: `parent_query_kwargs` is empty on the root route so `has_permission` returns `True`, then `get_object()` → `safely_get_object` → `check_object_permissions` lands here, `isinstance(object, Organization)` matches, and `_admits` raises. So removing the paygate from the disable path doesn't help the one person who needs to use it. (Traced by reading the chain, not by running a repro.)

Two ways to get there without a second admin cleaning up:

- `UserViewSet` isn't a `TeamAndOrgViewSetMixin` view and `/api/users/@me/` is on the enforcement bypass whitelist, so an admin can change their own email to an outside domain, verify it, and wedge the organization permanently.
- The new deletion guard in `organization_domain.py` computes `still_admitted` against the *acting* admin's address only, so an admin on one verified domain can delete the domain that admits another admin.

2FA has exactly this escape hatch and this doesn't: `WHITELISTED_PATHS` includes `/api/users/@me/two_factor_disable/`. Cheapest equivalent here is to let admins through to organization settings, or to special-case a PATCH that sets `enforce_verified_domains` to `false`.

posthog/permissions.py
```diff
@@ -238,6 +239,81 @@ def has_permission(self, request, view) -> bool:
...
+        if not isinstance(request.user, User):
+            return True
+
+        # Root viewsets (organizations, projects, environments) carry no parent URL kwargs, and the
+        # mixin's `organization` falls back to the user's current organization there, which is not
+        # the request's target. Gate on the fetched object below instead. Views deriving their
+        # target from the current team (`param_derived_from_user_current_team`) are the exception:
+        # for those the current team is the target by construction.
+        if not view.parent_query_kwargs and not view.param_derived_from_user_current_team:
+            return True
```
Piccirello, line 269: **Low — the early return leaves a read window into the enforcing organization for members of two or more orgs.**

Root-viewset *list* actions never invoke `has_object_permission`, so `GET /api/organizations/` returns every organization the user belongs to, blocked one included, serialized with `OrganizationSerializer` — whose `teams` field is `TeamBasicSerializer`, which carries `api_token`, alongside `available_product_features` and `customer_id`. `GET /api/users/@me/` does the same for the current organization and is whitelisted besides, and `set_current_organization` on that same ungated endpoint lets a blocked member point back at the organization they were just moved off.

Not reachable when the enforcing org is their only one, since login is refused outright. And `api_token` is the client-side ingestion key rather than a private secret, so this is an information-disclosure gap in the boundary rather than a credential leak.

Worth noting the whitelist reuse carries different weight here than for 2FA: a 2FA-blocked user can complete setup and leave the whitelisted set, whereas a domain-blocked user can never satisfy the check, so the exemption is permanent rather than transitional. Refusing `set_current_organization` into a blocked organization would close the re-pointing half cheaply.

posthog/api/signup.py
```diff
@@ -922,6 +934,16 @@ def process_social_domain_jit_provisioning_signup(
...
```
Piccirello, line 948: **Low — the surrogate path this docstring names is still an unchecked join.**

Returning `None` here means no organization gets checked, and `process_social_invite_signup` then falls back to `TeamInviteSurrogate`, whose `validate()` is unconditionally `True` and whose `use()` calls `user.join()` with no email check at all — bypassing both the domain rule and the invite/email match.

Precondition is narrow and unchanged from my last pass: it needs a `Team` row with a non-null `signup_token`, and nothing in current code sets one — only Django admin exposes the field. So this is legacy-data-only, not reachable on a fresh organization.

Still worth closing rather than leaving implicit, since the docstring already identifies it as a hole. Either resolve the surrogate's organization the same way, or have that path refuse outright when the target organization enforces.

posthog/models/organization_domain.py
```diff
@@ -167,6 +170,34 @@ def get_sso_enforcement_for_email_address(
...
```
Piccirello, line 191: **Low — this gives the stale-verification TODO a lockout consequence it didn't have before.**

`verified_domains()` a few lines up carries `# TODO: Verification becomes stale on Cloud if not reverified after a certain period`. Until now that was cosmetic. With enforcement on, a domain aging out of verified state stops admitting every member on it at once — including all admins, which then runs into the no-escape-hatch problem on the organization PATCH.

Not asking for the staleness work in this PR. But it's worth an explicit decision on whether re-verification can ever clear `verified_at` while `enforce_verified_domains` is on, and refusing to if so, because the failure mode is a whole organization locked out by a DNS change or a background job.

Piccirello (approved):
Agent-driven security review, second pass on the reworked enforcement.

Approving. The rework closed everything blocking, and the shape it landed on is better than what I suggested: splitting "may you have a session" (`resolve_login_organization`) from "may you reach this tenant" (`VerifiedDomainEnforcementPermission` on the URL-resolved organization) is the right decomposition, and gating on the resolved org rather than `current_organization` is the part that makes it hold. Moving enforcement into `SessionAuthentication` also incidentally neutralizes the partner deep-link login paths, which mint a session but can't use it.

Resolved since my last pass, so the matching comments there are outdated:

- Passkey login now gated (`webauthn.py` `_check_domain_enforcement`).
- `verify_email` now gated.
- Login no longer denies access to a member's other organizations.
- Enabling is refused when it would block the acting admin, and the guard was extended to domain deletion.
- Live sessions and non-session authenticators are now covered.
- Disabling is no longer paygated.

Four things left, none blocking. The first is the one I'd act on before this ships widely, because it turns a misconfiguration into an unrecoverable one.

posthog/permissions.py
a-lider: Fixed in 5d4fad5dd7f + cc7cebebaff, mirroring the 2FA escape hatch as you suggested. The org-detail PATCH now passes both domain gates, the serializer rejects every field change from a blocked admin except turning the setting off, and the endpoint's admin-write permission still applies (so only admins of the enforcing org can use it). Testing this locally surfaced the second half of the problem: after session expiry the blocked admin couldn't log back in to reach the hatch. So login now refuses blocked *members* with a clear error, while blocked *admins* get a session that's gated to the whitelist + the hatch. The full recovery loop (expired session → gated login → disable → full access) is covered by `test_blocked_admin_can_log_in_gated_and_recover_via_the_escape_hatch`. Both wedge paths you traced (self-service email change, deleting the domain that admits another admin) are recoverable now.

posthog/permissions.py
a-lider: Closed the cheap half in c18cbade739: `set_current_organization` now refuses an organization that blocks the member, with the standard `verified_domain_required` error. The list-serialization half stays a follow-up — whether the blocked org should disappear from the org switcher entirely is a product call I'd rather not smuggle into this PR. Also relevant to your porch-permanence point: gated sessions are now admin-only (cc7cebebaff), so the population standing on the whitelisted porch shrank to the people who can actually act from it.

posthog/api/signup.py
a-lider: Closed in 6d6918ac39c: the surrogate branch now runs the domain gate itself before `use()` — a legacy signup token into an enforcing organization resolves like an invalid invite, while tokens on non-enforcing organizations keep working. `test_legacy_team_signup_token_respects_domain_enforcement` covers both directions.

posthog/models/organization_domain.py
a-lider: Decision recorded in 586ecf91480 as an invariant comment next to the TODO: re-verification must never clear `verified_at` while `enforce_verified_domains` is on — whatever staleness mechanism gets built has to suspend or flag the domain rather than silently un-verify it, since clearing evicts every member on the domain at once. Nothing in current code clears `verified_at`, so there's no runtime guard to add yet; the comment sits where the future implementer will collide with it. Worth noting the blast radius also shrank since your pass: with the escape hatch and admin gated login, even this failure mode is now self-recoverable rather than a support escalation.

posthog/permissions.py
```diff
@@ -243,6 +244,85 @@ def has_permission(self, request, view) -> bool:
...
+    def has_permission(self, request: Request, view) -> bool:
+        if not isinstance(request.user, User):
+            return True
+
+        # Root viewsets (organizations, projects, environments) carry no parent URL kwargs, and the
+        # mixin's `organization` falls back to the user's current organization there, which is not
+        # the request's target. Gate on the fetched object below instead. Views deriving their
+        # target from the current team (`param_derived_from_user_current_team`) are the exception:
+        # for those the current team is the target by construction.
+        if not view.parent_query_kwargs and not view.param_derived_from_user_current_team:
```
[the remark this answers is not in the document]
a-lider: Verified, but this isn't a data leak and not critical — it's the already-tracked follow-up from the earlier review thread. What's readable is the org's directory card only: name, settings flags, and project names with the ingestion `api_token` (the same public token embedded in the org's website HTML — it can't read anything). No product data, no member list, no writes; every request into the org still 403s. And the reader isn't an outsider — it's someone the org still lists as a member, just no longer admitted. That they can see the org exists while being unable to do anything there is acceptable for now. Also worth noting the suggested fix as stated would break the enforcement escape hatch: a blocked admin must keep seeing the org in `GET /api/organizations/` to reach the setting and turn it off. The proper fix is role-aware (hide for blocked members, keep for admins) and lands as a small focused follow-up rather than expanding this PR post-approval.

PR #73498, opened 2026-07-24
fix(conversations): show existing tickets to free plans

clr182:
## Problem

Free plan orgs can't see their support tickets in the side panel, even when they have them. The whole tickets UI sits behind the paid support gate, so the panel renders nothing but the "upgrade to email support" message.

They do end up with legitimate tickets though: billing questions, and PostHog AI bug reports (`TicketPrompt` files a real ticket with no plan check). Their nav badge even ticks up with unread replies, with no way to open the thread. Nothing on the backend gates reads or replies, this was purely the panel.

## Changes

The gate moves from ticket **visibility** to ticket **creation**.

- New `canCreateTicket` selector in `sidepanelTicketsLogic` holding the entitlement rule (paid/custom subscription, active boost/scale/enterprise trial, or org under 3 months old), plus a guard in `startTicketFromSupportForm` so a support CTA can't hand the composer to a plan without support. The billing target area stays exempt, matching current behavior.
- `TicketsList` renders "Create new ticket" only when entitled. The list, threads, reply box and recovery link are ungated.
- `SidePanelSupport` renders the section whenever you can create a ticket **or** already have one, titled "Your tickets" for free plans. Its local `canEmail` was a duplicate of the same rule and now derives from `canCreateTicket`, since that duplication is what let the two gates drift apart in the first place.

Free plans keep the existing community and upgrade messaging below the list.

I (christian) ran this in a local branch and confirmed that free plans can see tickets created in their side bar but the create ticket button is still hidden.

Automated, run by Claude:

- `products/conversations/frontend`: 11 suites, 110 tests
- sidepanel + Support suites: 9 suites, 119 tests
- `pnpm --filter=@posthog/frontend typescript:check` clean

One existing test in `sidepanelTicketsLogic.test.ts` was asserting a bug CTA opens the composer, and it only passed because the old logic never checked the plan. It now sets up a paid plan explicitly. The new parameterized case is the regression this PR guards: on a free plan an analytics CTA must stay on the list while a billing CTA still opens the composer, and both consume the support intent so `supportRouterLogic` doesn't replay it. No existing test covered the plan gate at all.

Manual, locally: created a ticket through the billing CTA, then aged the org past 3 months so it reads as free and confirmed the ticket stays visible and repliable with no create button.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None, no user facing docs cover this gating.

## 🤖 Agent context

Claude Code (Opus 4.8) wrote this after I pointed it at the free plan visibility problem. Skills invoked: `/writing-tests`.

The main decision was where the billing escape hatch lives. First pass folded `targetArea === 'billing'` into `canCreateTicket` itself, but that selector flips back to false the moment the support intent is consumed and `sendSupportRequest` resets, which would unmount the section out from under someone mid-compose. So `canCreateTicket` is now pure plan entitlement and the billing exemption is checked once at the transition into the composer, with section visibility also keyed off the tickets view to stay stable.

Also considered leaving `canEmail` alone in `SidePanelSupport` and only touching the tickets block, but two copies of the same entitlement rule is exactly how this bug happened, so they're one source of truth now.


clr182:
I don't think the vera comment is in scope of this PR, and existed prior to this PR.

abigailbramble (approved):
Looks good.

Issue #72591, opened 2026-07-21
Email tracking-pixel consent for workflow emails (CNIL compliance)


mayteio:
## Background

France's CNIL published a recommendation on email tracking pixels (mandatory since July 14, 2026): open-tracking pixels in marketing emails require consent (or at minimum an easy objection mechanism), while transactional emails are exempt from consent but should not carry unnecessary tracking. Italy has equivalent rules. Reference: https://www.cnil.fr/fr/recommandation-pixel-suivi-courriels

The industry has converged on a common shape for this: a workspace-level consent mode (off/opt-in/opt-out), a per-recipient consent attribute, a hosted consent page recipients can use without logging in, per-message tracking disable, and open-rate math that excludes untracked sends.

We have customer requests for this via support, tracked in this Slack thread: https://posthog.slack.com/archives/C06GG249PR6/p1784013338232699

## Current behavior (the gap)

The "Capture email engagement events" toggle (`products/workflows/backend/models/team_workflows_config.py`) does **not** disable tracking. It only gates whether internal metrics are mirrored out as customer-facing `$workflows_email_*` events (`email-tracking.service.ts` `trackMetric`, `email.service.ts` send-time capture). With the toggle off:

- The 1x1 pixel is still injected into the email HTML (`nodejs/src/cdp/services/messaging/email-tracking.service.ts` `addTrackingToEmail()`). In production this pixel records nothing (dev-only handler), but it is still present in the message body.
- The real production tracking comes from the hardcoded SES configuration set `posthog-messaging` (`nodejs/src/cdp/services/messaging/email.service.ts`), which has open/click tracking enabled at the ESP layer. SES injects its own pixel and rewrites links. This config set is provisioned out-of-band, not in this repo.
- Links are still rewritten (both by us and by SES), and the workflow Metrics tab still records opens/clicks via the SES webhook path.

So there is currently no way, at any level (project, workflow, message, or recipient), to actually prevent open tracking. Per-link click-tracking opt-out exists (`data-ph-no-track` / `clicktracking="off"`), but there is no equivalent for the open pixel.

Related: #44269 (opt-out of engagement tracking for deliverability) overlaps with this but is framed around deliverability rather than consent.

## Design: two independent controls, one send-time decision

[...]

PR #73508, opened 2026-07-24
feat(workflows): track opens and clicks toggle on email steps

mayteio:
## Problem

#73506 added a per-send email tracking gate keyed on a `tracking_enabled` config field on the `function_email` action, but there's no way to set it from the workflow editor yet.
This is the UI slice of the MVP in #72591: a per-email-step control so transactional (or any) emails can opt out of open/click tracking, as required by CNIL's tracking-pixel recommendation.

## Changes

- New "Track opens and link clicks" switch in the email step's detail panel (next to "Message category"), default on, writing `config.tracking_enabled` via the existing `setWorkflowAction` flow. The info tooltip explains that untracked sends won't record opens or clicks, while delivery, bounce, and unsubscribe events still flow.
- Mirrored `tracking_enabled` in the frontend zod schema for `function_email` (kept in sync with the nodejs schema from #73506).
- Documented the key in the `HogFlowActionSerializer` config `help_text` (flows into generated API/MCP schemas, regenerated in this PR) and in the building-workflows skill's graph schema reference.
- Clarified the "Engagement events" settings description: that toggle only controls whether engagement is captured as PostHog events, it does not disable tracking itself.

> [!NOTE]
> Depends on #73506 - without it the flag is stored but has no effect at send time, so this should merge after.

- Frontend typecheck passes for all touched files (pre-existing unrelated errors in my local checkout come from stale generated kea type artifacts, not this change). Lint/format clean; OpenAPI types regenerated with `hogli build:openapi`.
- No new tests: the switch is declarative UI writing one config key through the already-tested `setWorkflowAction` path, and the behavioral contract (what `tracking_enabled: false` does at send time) is covered by the jest cases added in #73506. A component test here would only re-assert prop wiring.
- I (the agent) did not run the app to click through the editor manually.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Settings copy and the building-workflows skill reference updated in this PR. posthog.com user docs (workflows email tracking page) live in the website repo and should follow once the feature ships.

## 🤖 Agent context

Authored with PostHog Code (Claude). Skills invoked: /improving-drf-endpoints, /writing-tests (gate concluded no new tests were warranted here).
Placement decision: the switch sits in `HogFlowEditorPanelBuildDetail` beside the message category selector rather than inside the email template configuration, since it's step-level delivery behavior (like category) rather than content. It's gated on `isEmailAction` because pixels/link rewriting are email-only.

---


mayteio:
Consolidated into #73506 - the step toggle now ships together with the consent system so the UI lands as one coherent story, with the toggle framed as the message-level override rather than the compliance control.

PR #73513, opened 2026-07-24
chore(data-warehouse): cross-reference self-driving inbox from source skill

Gilbert09:
## Problem

The `implementing-warehouse-sources` skill walks an engineer (or agent) through building a data warehouse import source, but it never mentions the Self-driving Inbox. So someone shipping a source that syncs a table of actionable records — issues, tickets, or conversations — has no signal that the source is also a candidate for self-driving, and no pointer to the separate `adding-inbox-sources` skill that wires it up. The two skills existed side by side with no cross-reference.

## Changes

`chore(data-warehouse)` — docs only, no production code.

Adds a "Self-driving Inbox candidacy" section to `.agents/skills/implementing-warehouse-sources/SKILL.md` that:

- Explains the signal for candidacy is the **table** synced (`issues` / `tickets` / `conversations`), not the vendor, and which source categories those tend to fall under.
- Calls out which sources are explicitly **not** inbox candidates (analytics, billing, ad-platform, CRM, raw databases).
- Makes clear that wiring a source into the inbox is separate, additive work covered by `/adding-inbox-sources`, only possible once the warehouse source (this skill's deliverable) exists, and that none of that plumbing belongs in `products/warehouse_sources/` code.

Also adds an optional line to the implementation checklist so the candidacy check surfaces at handoff without being mistaken for a mandatory build step.

Docs-only change to an agent skill. Ran `markdownlint-cli2 --fix` with the repo's `.config/.markdownlint-cli2.jsonc` against the edited file — passes clean with no changes. No automated test suite covers skill markdown content; `oxfmt` couldn't run in this local environment (requires a newer Node than is available here), but the change follows the file's existing semantic-line-break style and markdownlint is clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable — this is agent-facing skill documentation, not user-facing posthog.com docs.

## 🤖 Agent context

Tom asked to review the self-driving inbox source skill (`adding-inbox-sources`) and add a reciprocal reference from the source-implementation skill (`implementing-warehouse-sources`) covering whether a new source should be included in self-driving. I read both skills, then added a single cross-referencing section plus an optional checklist line to the implementing skill. Edits were made against a fresh `origin/master` worktree (the skill on master is ahead of the local checkout, so I diffed to edit the current content). No changes were made to the `adding-inbox-sources` skill itself, per the scope of the request.


estefaniarabadan (approved):

Gilbert09:
Heads-up on scope: besides the doc cross-reference, I pushed two small commits to get CI green.

- Fixed the markdown formatting in `SKILL.md` so `oxfmt --check` passes (a stray line-initial `+` was being parsed as a bullet list, which had broken the "three surfaces" sentence).
- Fixed two Semgrep bash-parser errors in `.github/scripts/verify-playwright-new-tests-and-snapshots.sh`. Semgrep's bash parser can't handle an arithmetic ternary `$(( ... ? ... : ... ))` nor multiplication combined with a comparison inside `(( a * b > c ))`; either one makes the whole Semgrep scan exit non-zero on a partial parse, which was failing the `semgrep` check on every open PR. Rewrote both as equivalent plain arithmetic (precompute + compare / explicit if-else). Verified locally with Semgrep 1.171.0 that the script now parses at 100%, and the `semgrep` check is green on the latest commit.

The second change is unrelated to the doc content but was the only thing standing between this PR and a green `semgrep` check.

PR #73518, opened 2026-07-24
feat(navigation): drop starred sidebar item, instrument files tree

fercgomes:
## Problem

Closes GROW-217

Two asks came out of a Slack thread (linked below):

1. The "Starred" item in the left navigation is being retired, so it should no longer show up in the sidebar.
2. The Files feature (the project tree) had almost no product analytics, so we couldn't tell how people actually use it.

## Changes

- Removed the `Shortcuts` / "Starred" entry from `panelTriggerItems` in `NavTabBrowse`, so it no longer renders in the browse tab. The underlying shortcuts logic and the "Add to starred" action on recent items are untouched.
- Added usage instrumentation to the Files tree, mostly in `projectTreeLogic` so it fires regardless of which UI path triggers the action, plus a couple of interaction-only events in `ProjectTree`. Every event carries the tree `root` so the Files tree (`project://`) can be isolated from the other trees that reuse the same component.

New events: `project tree folder created`, `project tree item renamed`, `project tree item deleted`, `project tree items moved` (bulk), `project tree item moved` (drag), `project tree searched`, `project tree sort changed`, `project tree multi-select toggled`. These sit alongside the existing `project tree item clicked` and `project tree folder toggled`.

Context: https://posthog.slack.com/archives/C0113360FFV/p1784832773192549

I (Claude) could not run the frontend toolchain in this environment (`node_modules` not installed, so typecheck / lint / typegen did not run). Changes are limited to adding `posthog.capture` calls and removing one array entry, following the existing capture patterns already in these files. No automated tests were added or run.

## 🤖 Agent context

Authored by Claude (PostHog Code). I located the Files feature (the ProjectTree / file-system panel) and the "Starred" nav trigger, then added the instrumentation in the kea logic rather than the component so events fire from any entry point. Care was taken to merge captures into existing `loadSearchResultsSuccess` and `moveCheckedItems` listeners instead of declaring duplicate listener keys, which would have silently shadowed the real handlers. I was unable to read the linked Slack thread (auth-gated, no Slack tool available) and worked from the task description.
---

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -7095,37 +7095,37 @@ snapshots:
     scenes-app-settings-user--settings-user-reminders-modal--light:
         hash: v1.k794b7964.fbcf3f0f69a5b43c28ea7537ce6f1aab3c21e70f9c2cbb6fdfef4546dadba86c.yYIPurSofuNP1kZXYzNBOa3YU5Zy7VLaSegvGrNZ8Mw
     scenes-app-sidepanels--side-panel-access-control--dark:
-        hash: v1.k794b7964.ab76b9589b4b7e1e1ca84725b693b53629f9309f9472966d360b87377f7a7f2c.zbWOHBcSUM1lyU4uwh85J9XpPlcKOJSDMvLJ33AQUAc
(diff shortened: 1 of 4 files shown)


adamleithp (approved):
🪓

PR #73523, opened 2026-07-24
feat(early-access-features): allow assigning features to a person or role

joethreepwood:
> [!NOTE]
> Stacked on #72993 (`posthog-code/eaf-created-by-column`) — this PR replaces #73466, which conflicted with #72993 because both added `created_by`. Here the `created_by` layer comes from #72993 and only the assignee layer is added on top. Merge #72993 first, then retarget this PR to `master`.

> [!NOTE]
> The posthog.com roadmap consumes the public `assignee` field added here: [posthog.com#18882](https://github.com/PostHog/posthog.com/pull/18882). That PR falls back to its hard-coded team map until this one merges and deploys, so there is no ordering constraint — but the roadmap only reflects in-app assignments once this ships.

## Problem

Early access features have no ownership concept beyond the `created_by` field added in #72993. There is still no way to hand a feature to a person or team in the app, so surfaces that want to show an owner (like the public roadmap on posthog.com, which currently renders these features as "Unassigned" via a hand-maintained override map) have nothing to read.

## Changes

Same substance as #73466, rebased onto #72993's branch:

- `EarlyAccessFeature` gains `assigned_user` and `assigned_role` (the same "person or role" assignee convention error tracking uses). New migration `0009` in `products/early_access_features/backend/migrations`. Unlike #73466, `assigned_user` is declared with `db_constraint=False`, matching how #72993 adds `created_by` — no lock on the hot `posthog_user` table.
- The authenticated API gains a read/write `assignee` field shaped `{"type": "user" | "role", "id": ...}` or `null`, validated against the organization's members and roles. On create the assignee defaults to the creator unless an explicit `assignee` (including `null`) is sent. `created_by` handling itself is #72993's and is unchanged here.
- The **public** `/api/early_access_features/` endpoint (project-token auth, consumed by posthog-js and posthog.com's roadmap) now serves the assignee as a display name only — `{"type": "user" | "role", "name": "..."}` — no user IDs or emails. The public view and the flag-list `features` prefetch join the assignee relations to avoid N+1s.

> [!WARNING]
> This makes assignee display names (a member's first/last name, or a role name) visible to anyone with the project token, for every project. That felt like the minimal exposure that still lets consumers attribute features to owners, but it deserves a deliberate review. Happy to gate it or strip user names if we'd rather not expose them.

- The early access feature scene gets an "Assignee" control in the scene panel, reusing the error tracking `AssigneeSelect` components (cross-product import, same as `products/dashboards` already does). Changes save immediately in view mode and with the form in edit mode.

Automated tests carried over from #73466 in `products/early_access_features/backend/test/test_early_access_feature.py`:

- create defaults the assignee to the creator (the regression: creating silently leaves features ownerless)
- create with an explicit role assignee, and with an explicit `null` assignee (stays unassigned rather than falling back to the creator)
- update to another user, to a role, and back to `null`
- rejects users and roles from other organizations, and malformed payloads (wrong type, missing id, non-dict, invalid UUID)
- the public endpoint serves assignee display names for user and role assignments without per-feature queries (`assertNumQueries(2)`)

I (Claude, running in PostHog Code) could not run the Django test suite or frontend typechecking in this sandbox — no database or node_modules — so CI is the first full run. The postgres query snapshots (`.ambr`) were hand-merged on top of #72993's versions and may need the snapshot-update automation to true them up. `ruff check`/`ruff format` pass on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude via PostHog Code, directed by Joe. This PR is a restack of #73466 (closed) onto #72993, resolving the conflict between them: both PRs added `created_by` to `EarlyAccessFeature` with clashing `0007` migrations. Decisions: kept #72993's `created_by` implementation (including its backfill and `db_constraint=False`) verbatim and layered only the assignee changes on top as migration `0009`; switched `assigned_user` to `db_constraint=False` to match and to satisfy the hot-table FK guidance; dropped #73466's unrelated generated-type drift in `products/experiments`. Skills invoked: /django-migrations, /improving-drf-endpoints.

---


products/early_access_features/frontend/EarlyAccessFeature.tsx
gustavohstrassburger, line 1: I think we should update the LemonTable to display the assignee instead of the created by.

gustavohstrassburger (approved):
Stamping to unblock, there are some review comments (mine and AI's) that I think should be addressed before merging though.

joethreepwood:
Pushed fixes for the CI failure and review feedback:

- **mypy** (`Python code quality`): the assignee update test now re-fetches the feature between assertions instead of `refresh_from_db()`, so mypy's attribute narrowing no longer marks later assertions unreachable.
- **@gustavohstrassburger's comment**: the features list table now shows an **Assignee** column (using error tracking's `AssigneeResolver`/`AssigneeDisplay`) instead of the created-by column.
- **veria-ai's public-exposure finding**: intentionally not changed here — this is the deliberate trade-off already called out in the PR description's warning (the posthog.com roadmap needs a display name to attribute features). If we'd rather not expose member names via the project-token endpoint, happy to strip user names or gate it, but that's a product call for a human to make.

The frontend typechecking and bundle size jobs on the previous run were canceled rather than failed; they should complete on this push.

joethreepwood:
@gustavohstrassburger Resolving conflicts, but if I follow you this should now be good to merge once that's done?

gustavohstrassburger:
> @gustavohstrassburger Resolving conflicts, but if I follow you this should now be good to merge once that's done?

yes, but maybe worth arranging with @haacked if it's better to merge yours on his and then only once to master, or if he rather merge his on master and then yours. but I'll leave this up to you, just giving a suggestion here.

PR #73541, opened 2026-07-24
feat(data-warehouse): add schema-scoped syncs tab

luke-belton:
## Problem

On the data warehouse schema detail page, the Configuration → Details section had a "View sync history" button. Clicking it navigated *away* from the schema you were looking at, back to the parent source page's Syncs tab (deep-linked with `?schema=<name>` to pre-filter). That context switch is jarring — you leave the schema to see its own sync history.

## Changes

Give the schema page its own **Syncs** tab, scoped to that schema only, and repoint the button at it.

The source-level `SyncsTab` already renders the jobs table, polls for live status, supports "load older", and can filter by schema name. Rather than duplicate that, I taught it an optional single-schema lock and mounted it on the schema scene:

- **`SyncsTab.tsx`** — new optional `lockedSchema?: string` prop. When set, the tab pre-filters to that one schema on mount and hides the schema multi-select (which would be a no-op). The existing `?schema=` deep-link path is untouched.
- **`schemaSceneLogic.ts`** — added `'syncs'` to `SCHEMA_SCENE_TABS`. Routing and the `dataWarehouseSourceSchema` URL helper are already generic over the tab list, so no routing changes were needed. Regenerated kea types.
- **`SchemaScene.tsx`** — renders `<SyncsTab id={cleanedSourceId} lockedSchema={schema.name} />` as a new tab after Configuration, gated on `shouldShowManagedSourceSyncsTab` (hidden for direct-query sources, matching the source page). Added a guard effect + `activeTab` fallback so navigating to `/syncs` on a direct source redirects to Configuration. The tab also carries a "View all syncs for source" link so there's an explicit way up to the unfiltered source-level history.
- **`ConfigurationTab.tsx`** — the "View sync history" button stays, but is now a plain link to the schema's own Syncs tab (`syncHistoryUrl` prop) rather than an imperative push to the source page. It's omitted for direct-query sources, which have no Syncs tab to land on.

Live status polling comes for free since the reused `SyncsTab` / `sourceSettingsLogic` already polls.

> [!NOTE]
> No backend change. The jobs endpoint (`GET .../external_data_sources/:id/jobs/`) already accepts a `schemas` filter, and the frontend API client already passes it.

The source page's SchemasTab status-tag click that deep-links to the source Syncs tab with `?schema=` is left as-is — it's on the source page, still valid.

Automated, run locally by me (Claude):
- `tsgo --noEmit` — no errors in the touched files
- `oxlint` on the changed files — clean (two pre-existing warnings unrelated to this change)
- `oxfmt --check` — clean
- `hogli ci:preflight` — clean
- kea typegen — up to date

I (Claude) did **not** run the app manually. Suggested manual checks for a reviewer:
- Managed source → schema shows a **Syncs** tab scoped to that one schema, no schema dropdown, expandable log rows, "Load older jobs" footer.
- "View sync history" in Details moves to that Syncs tab and the URL becomes `.../schemas/:schemaId/syncs`; back returns to Configuration.
- "View all syncs for source" lands on the source's Syncs tab showing *all* schemas, with the Schema filter empty.
- Trigger "Sync now" → Syncs tab flips Running → Completed without manual refresh (polling).
- Direct-query source: no Syncs tab, no "View sync history" button; visiting `.../schemas/:id/syncs` redirects to Configuration.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Luke directed this; I (Claude) implemented it. The ask was to give the schema page a Syncs tab local to the schema, keep the existing "View sync history" button as the shortcut into it, and offer a link back out to the source's full sync history.

Key decision: reuse the existing source-level `SyncsTab` via a `lockedSchema` prop rather than build a parallel schema-jobs table/logic. Kept it one source of truth for the jobs table and its polling. Confirmed the backend already supports schema-name filtering on the jobs endpoint, so this stayed frontend-only. Gated both the tab and the button on `shouldShowManagedSourceSyncsTab` to match how the source page hides syncs for direct-query sources.

No skills were required (no DRF/migration/API-type changes).

---

### DIFF
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -6451,6 +6451,7 @@ export interface ExternalDataSourceSchema extends SimpleExternalDataSourceSchema
 export interface ExternalDataSchemaSourceSummary {
     id: string
     source_type: ExternalDataSourceType
+    access_method?: ExternalDataSource['access_method']
     supports_column_selection?: boolean
     supports_row_filters?: boolean
     user_access_level: AccessControlLevel | null
--- a/products/data_warehouse/frontend/scenes/SchemaScene/ConfigurationTab.tsx
+++ b/products/data_warehouse/frontend/scenes/SchemaScene/ConfigurationTab.tsx
@@ -83,7 +83,6 @@ export interface ConfigurationTabProps {
     source: SchemaSceneSource | null
     section: SchemaConfigurationSection
     onConfigureSyncMethod: () => void
-    onViewSyncHistory: () => void
 }
 
 export function ConfigurationTab({
@@ -92,7 +91,6 @@ export function ConfigurationTab({
     source,
     section,
     onConfigureSyncMethod,
-    onViewSyncHistory,
 }: ConfigurationTabProps): JSX.Element {
     const logic = schemaSceneLogic({ sourceId, schemaId: schema.id })
     const { isProjectTime, refreshingSchemas, resyncingSchema, supportsRowFilters } = useValues(logic)
@@ -110,7 +108,6 @@ export function ConfigurationTab({
                     cancelSchema={cancelSchema}
                     updateSchema={updateSchema}
                     onConfigureSyncMethod={onConfigureSyncMethod}
-                    onViewSyncHistory={onViewSyncHistory}
                 />
             )
         case 'sync-method':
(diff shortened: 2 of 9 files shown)


Gilbert09 (approved):
The greptile comments may need addressing

PR #73555, opened 2026-07-24
fix(onboarding): drop --region from wizard subcommand commands

christiaan-ph:
## Problem

An EU customer hit `✖ Unknown argument: region` running the MCP analytics setup command the app handed them:

```
npx -y @posthog/wizard@latest mcp-analytics --region eu --project-id=22631
```

The wizard removed `--region` as a global flag in [PostHog/wizard#934](https://github.com/PostHog/wizard/pull/934) (merged 2026-07-21). It now resolves the cloud region from the OAuth login token (`posthog_region`) and only accepts `--region` on the three API-key / no-OAuth paths: base integration (`npx @posthog/wizard`), `audit`, and `provision`. Every other subcommand rejects it.

Our app still appends `--region eu` for EU Cloud users. `useWizardCommand` builds `base + subcommand + ' --region eu' + ' --project-id=<team>'`, which is exactly the failing command above. So the copy-paste commands on the MCP analytics, data warehouse, self-driving, workflows, and ingestion-warnings onboarding surfaces are all broken for EU users. `--project-id` is unaffected (still a valid global flag).

[#73263](https://github.com/PostHog/posthog/pull/73263) already did this fix for the error-tracking / source-maps surfaces, but its sweep missed the shared `useWizardCommand` hook and the inline data-warehouse hint. This PR covers those two.

## Changes

Drop the `--region` suffix from the two remaining surfaces that appended it. Region resolves from the OAuth login, so the flag is redundant on the paths that still accept it and breaks the ones that don't.

- `useWizardCommand.ts` — remove `regionPart` (and the now-unused `preflight.region` / `Region` reads). This fixes every subcommand caller at once (MCP analytics, data warehouse, self-driving, workflows, ingestion-warnings, product empty states).
- `WarehouseWizardHint.tsx` — remove the inline `' --region eu'` suffix.

Left the headless cloud-run wizard in `products/tasks/.../run_wizard.py` untouched: it runs non-interactively with an API key, where `--region` is still valid and required (same call #73263 made).

I'm an agent (Claude Code, Opus 4.8). I have not run the app manually. Automated checks I ran:

- `hogli ci:preflight --fix` — clean (nothing in the diff maps to a known CI failure class).
- `oxlint` on both changed files — clean.
- lint-staged formatter ran on commit — clean.
- Confirmed no test files reference `useWizardCommand` or `WarehouseWizardHint`, and the `run_wizard.py` region test still asserts the headless `--region` path (unchanged).

Suggested manual check before merge: on EU Cloud, open the MCP analytics empty state and the data warehouse new-source catalog, and confirm the copied commands no longer contain `--region eu` and run cleanly.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change. posthog.com already omits `--region` (the `<WizardCommand />` component reads region off the auth token). The `wizard-signup.mdx` doc keeps `--region` on purpose since account signup has no token to infer the region from.

## 🤖 Agent context

I (Christiaan) directed this from a customer report of the `Unknown argument: region` error. Claude Code traced it: confirmed via the wizard source that `--region` was dropped from `GLOBAL_OPTIONS` in PostHog/wizard#934 and now lives only on base integration / `audit` / `provision`, then found the app's `useWizardCommand` still emits it for EU. Swept every `--region` + wizard call site in the app; #73263 had already fixed the error-tracking ones and correctly left `run_wizard.py` alone, so this PR is scoped to the two shared surfaces that sweep missed. No skills were required.


gewenyu99 (approved):
Yes, region is no longer supported (or rather, never did anything)

PR #73563, opened 2026-07-24
fix(data-imports): stop _safe_expose_ch_error crashing on errors without .message

Gilbert09:
## Problem

A data warehouse sync surfaced this in error tracking:

> `AttributeError: 'EOFError' object has no attribute 'message'`

The real failure is a transient one: ClickHouse closed the socket mid-read during warehouse column introspection (`DataWarehouseTable.get_count` → `sync_execute`), and `clickhouse_driver` raised `EOFError: Unexpected EOF while reading bytes`. That error then reached `_safe_expose_ch_error`, which is supposed to translate a ClickHouse error into a friendly message. Instead it crashed on its own error handling.

`wrap_clickhouse_query_error` only attaches a `.message` to ClickHouse `ServerException`-derived errors. It returns everything else untouched. An `EOFError` is not a `ServerException`, so it came back with no `.message`, and the `ExtractErrors` substring loop (`if key in err.message`) blew up with `AttributeError`. The handler had a guard for exactly this shape, but only for `ClickHouseAtCapacity` - every other message-less error (the `EOFError` here, plus the timeout / memory-limit `APIException`s `wrap_clickhouse_query_error` can also return) hit the crash.

The net effect: a retryable connection blip was replaced by a defect-shaped `AttributeError` in error tracking, and the true error was hidden.

## Changes

Generalize the existing `ClickHouseAtCapacity` guard: after wrapping, if the error has no `.message`, re-raise it untouched.

```diff
-        # Capacity errors are transient — surface them so the caller can retry. Check this
-        # before the message matching below, since ClickHouseAtCapacity is an APIException
-        # and has no `.message` attribute.
-        if isinstance(err, ClickHouseAtCapacity):
-            raise err
+        # Only ClickHouse ServerException-derived errors carry a `.message`. Everything else —
+        # transient connection/read errors (e.g. an EOFError from a dropped ClickHouse socket) and
+        # already-translated APIExceptions like ClickHouseAtCapacity — has no `.message`, so re-raise
+        # it untouched. Masking these as a storage-bucket misconfiguration would both crash here on
+        # the missing attribute and hide a retryable error behind a misleading user-facing message.
+        if not hasattr(err, "message"):
+            raise err
```

Only `ServerException`-derived errors carry a `.message` and are meant to be matched against `ExtractErrors`. Anything else is either a transient error that should stay retryable, or an already-translated `APIException` that carries its own meaning. Both should propagate as-is, not be crushed into the generic "check your bucket URL" message. This keeps the transient `EOFError` retryable via Temporal instead of turning it into a defect.

This is not the ClickHouse-layer retry work (#72081, #70389) - those add retries on other read paths and don't touch this handler. Even with a retry in place, an exhausted retry still lands here, so this crash needed fixing on its own.

Added two parameterized cases to `TestSafeExposeChError` in `products/warehouse_sources/backend/tests/test_table.py`:

- `test_transient_errors_without_message_are_reraised_untouched` - feeds an `EOFError` and a `ConnectionResetError` (both message-less, non-`ServerException`) and asserts the exact original exception is re-raised, not an `AttributeError` and not the generic bucket message. This is the regression the reported issue is: no existing test covered a message-less error other than `ClickHouseAtCapacity`.

Ran the class (4 passed) and `ruff check` on both changed files (clean). These are pure unit tests - no DB, no ClickHouse.

## 🤖 Agent context

Authored by PostHog Code (Claude Opus 4.8) from an error-tracking issue on the data-warehouse import pipeline. Traced the two-frame exception chain (transient `EOFError` from the clickhouse driver → `AttributeError` in `_safe_expose_ch_error`) to the missing-`.message` assumption. Invoked `/writing-tests` to gate the test additions. Considered adding `EOFError` to a retryable/non-retryable list but rejected it - the error was already retryable; the only bug was the handler crashing on it. Kept the fix scoped to the handler.

---


estefaniarabadan (approved):

Gilbert09:
Rebased onto `master` to clear the merge conflict, reconciling this fix with the `_safe_expose_ch_error` changes that landed on master in the meantime (the `raw_message` capture + `CORRUPTED_PARQUET_METADATA_MESSAGE` mapping). The fix is now expressed as: re-raise any error without a `.message` (transient connection errors like `EOFError`, plus already-translated `APIException`s such as `ClickHouseAtCapacity`) untouched, rather than masking it as a storage-bucket misconfiguration. Tests pass locally and comments were updated so they no longer claim a crash that master's `raw_message` change already prevents.

**Status:** the branch is `MERGEABLE`, review is `APPROVED`, and every check this PR owns is green (Django tests, product tests, `semgrep-python`, lint).

The only red check is the top-level **`semgrep`** app, and it is **not caused by this PR** — it's a repo-wide CI-infra failure:

- The scan reports `Findings: 0 (0 blocking)`.
- It exits non-zero purely because semgrep's bash parser chokes on `.github/scripts/verify-playwright-new-tests-and-snapshots.sh:111` (a `(( ))` arithmetic expression) and the job runs with `--error`.
- That script lives on `master` and is untouched by this PR (which only changes two `products/warehouse_sources/backend/` files).
- The identical failure (same script, same line, 0 findings, exit code 3) is present on other unrelated open PRs, so it affects everything in flight, not just this branch.

I can't re-run or fix it from here — the `Semgrep` workflow is defined in a cross-repo/private location, and the parse error is deterministic. This needs a devex/infra fix (e.g. semgrep-ignore that script or drop `--error` on parse warnings) rather than a change on this PR. Everything else is ready to merge.

PR #73572, opened 2026-07-24
feat(logs): tri-state exclusions on the service facet

jonmcwest:
## Problem

Stacked on #73564. The Service facet was the last logs facet without the excluded state, and it needed one real backend change: the facet-counts query only gated severity filters out of their own facet, so a service exclusion would have zeroed out its own value's count.

## Changes

- Service values cycle unchecked → included → excluded → unchecked. Includes keep the dedicated `serviceNames` field; exclusions are an `is_not` `service_name` log filter in the filterGroup, compiled by the existing generic property path to `NOT IN` on the column. The trace list and sparkline share the same filter builder, so both scope correctly.
- Backend: added the own-facet strip for `service_name` log filters, mirroring severity. Note the strip is by key, so when faceting on service it also ignores a hand-authored `exact` service chip, matching severity's behavior and standard faceted-search semantics.
- The severity and service cycle paths collapsed into one `cycleColumnValue` helper in the listener (third occurrence of the shape, as flagged in the previous PR's review).
- The taxonomic picker doesn't offer `service_name` as a log property; rail-authored exclusion chips still render and edit fine in the chips bar. Adding it to the picker is a separate product decision.

Screenshots: none yet, one manual pass over the whole stack at the end.

Automated only. The backend gate was done test-first: the parameterized own-facet test failed on the service row before the gate and passed after.

- `test_log_facet_values.py`: the severity-exclusion pair from the previous PR became parameterized over (severity, service) — own-facet counts must ignore the exclusion, and other facets must honor it (proves the `NOT IN` translation on real ClickHouse data). The service row is the new coverage; it catches the missing gate.
- `facetRailLogic.test.ts`: service cycle across both stores, simultaneous mixed include/exclude, and that service and severity exclusions stay under their own keys (catches key-clobbering between the two column facets, which share the storage mechanism).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude Code (Claude Fable 5), directed by @jonmcwest. Skills invoked: `/writing-tests`, `/writing-kea-logics`, `/code-review` (combined two-axis pass on the commit; its one partial finding, a missing simultaneous mixed-state service test, was fixed in the follow-up commit).


jonmcwest:
### Branch brought up to date + CI status

Rebased this branch onto the current tip of the base (`jon/logs-facet-level-exclusion`) to clear the merge conflict. The base had been squash-rebased onto master, so I replayed only the two service-facet commits onto it (`git rebase --onto`) and dropped the four level-facet commits that are now part of the base. The PR's intent and diff are unchanged — 5 logs files (backend `logs_query_runner.py` + tests, and the `FacetRail` frontend logic/config). The branch is now **mergeable** with the base as an ancestor, and no base-only files leaked into the diff.

All functional CI is green (backend, frontend typecheck, Jest, lint, semgrep, bundle size, etc.).

The only remaining red is the **Visual Review** gate, and it's flagging a single snapshot that is unrelated to this PR:

- Story: `scenes-app-dashboards--show-across-viewports--narrow--light` — a **Dashboards** scene story. This PR only touches the logs facet code, which cannot affect how the Dashboards scene renders.
- The diff is a page-height/size mismatch (the lower dashboard tiles are absent from the capture) — the signature of a flaky async render of that heavy multi-viewport story, not a real UI change.
- The base branch's own latest Visual Review run passed with **0 changed** snapshots on the identical Dashboards code, confirming this isn't introduced here.

This check is non-blocking. Greening it requires finalizing the Visual Review run, which commits a new baseline — a human decision I've intentionally left to you rather than shipping an unrelated (and truncated-looking) Dashboards baseline through this logs PR. You can either accept/finalize that one snapshot in the Visual Review UI, or re-trigger a fresh Storybook run to let the flaky capture clear.

jonmcwest:
### Rebased onto the advanced base + CI green

The base (`jon/logs-facet-level-exclusion`) moved ahead by several commits and picked up a refactor of the level-facet cycling, which conflicted with this PR in `facetRailLogic.ts`. Rebased onto the current base tip and resolved the one conflict by reconciling both intents:

- **Base's improvement, kept:** the included → excluded transition writes both stores in a single `setFilters(..., false)` so the query, URL sync, and filter history all observe the final state (no half-applied cycle).
- **This PR's generalization, kept:** the cycling logic is factored into `cycleColumnValue`, now driven by the facet source's `exclusionKey`, so the Service facet gets the same tri-state (include → exclude → clear) as Level. The single-field transitions still go through the dedicated `setSeverityLevels` / `setServiceNames` setters, preserving their filter-change analytics and product-intent.

Verified locally: the 9 `facetRailLogic` Jest specs pass (including the new mixed service/severity exclusion cases), and `tsgo` reports no errors in the touched FacetRail files. The PR's diff is unchanged in scope — the same 5 logs files, with the base now a true ancestor and no base-only files leaked in.

CI is fully green on the latest commit (183 passing, 0 failing), and this fresh commit's Visual Review ran as a full `storybook` pass — the earlier stale `(partial)` status that lingered on the previous commit is no longer present.

The only remaining item is the required review approval, which is a human gate I can't satisfy.

jonmcwest:
### Merge conflict resolved; branch is mergeable again

The base moved on while this PR sat behind the level-facet work, and that work (`feat/fix(logs): tri-state exclusions on the level facet`) has since landed on `master`. This branch was carrying those same commits, which is what produced the conflict in `facetRailLogic.ts`.

Fix: replayed only this PR's own service-facet commits onto the current `master` (`git rebase --onto origin/master`), dropping the level-facet commits now present upstream. The typegen commit was detected as already-upstream and dropped automatically. `master` is now a true ancestor, the branch is `0` behind, and the diff is unchanged in intent — the same service-facet work, now expressed as a delta on top of the level facet:

- `products/logs/backend/logs_query_runner.py` — own-facet strip for `service_name` exclusions
- `products/logs/backend/test/test_log_facet_values.py`
- `products/logs/frontend/.../facetRailLogic.ts` — factor the include→exclude→clear cycle into a shared `cycleColumnValue`, wired for both Level and Service
- `products/logs/frontend/.../facetRailLogic.test.ts`
- `products/logs/frontend/.../facets.ts` — `exclusionKey: 'service_name'`

No base-only files leaked in: `FacetRail.tsx` and `facets.test.ts` dropped out of the diff because their changes were level-facet-only and are byte-identical to what's already on `master`.

### CI

All required product checks are green (backend, Django, frontend typecheck, Jest, lint, **Semgrep Checks Pass**, etc. — 182 passing). The one required check I re-ran was **Frontend typechecking**, which had been killed mid-run by a runner shutdown ("received a shutdown signal"), not a code error; it passed on re-run.

Two checks remain red, and both are inherited from `master` — neither is caused by this logs-only change:

1. **Visual regression tests pass** (and its `Complete Visual Review run` / `PostHog Visual Review` children) — the only changed snapshots are `endpoints-usagetrendschart--breakdown--light` and `--dark`. That's the **Endpoints** usage-trends chart, which this PR does not touch. The rendering changed on `master` itself in [#70835](https://github.com/PostHog/posthog/pull/70835) ("render usage trends chart via quill-charts directly"): the line interpolation is now smoother, so the current render no longer matches the committed `snapshots.yml` baseline. This is not flaky — it's a deterministic, structural change, and `master`'s own push-run flags the identical snapshots as `changed`. Every PR that rebases onto the current `master` will hit this until the baseline is regenerated. Greening it means blessing an Endpoints chart baseline that changed elsewhere; I've left that to the Endpoints owners rather than committing an unrelated baseline through a logs PR.

2. **semgrep** (the org-level `Semgrep` check, distinct from the required `Semgrep Checks Pass`, which passes) — a parser error on `.github/scripts/verify-playwright-new-tests-and-snapshots.sh:111` (a C-style `for (( ))` loop the pinned semgrep can't parse), reporting `0 findings`. It's failing identically on every currently-open PR, so it's a pre-existing, repo-wide infra issue and not part of the required merge gate.

The branch is `MERGEABLE`. The remaining gate is the required review approval.

jonmcwest:
### Status update

- ✅ Approval is in (1/1 required), and it won't be dismissed on push.
- ✅ Branch is `MERGEABLE` — no conflict. Master's "require up-to-date" policy is not strict, so the branch being a couple of commits behind a fast-moving `master` is not a merge blocker (and re-syncing would only re-run the full suite and re-land the same visual-review state below, so I've left it).
- ✅ All required product checks are green.

The **only** remaining required check that's red is **Visual regression tests pass**, and it's the Endpoints `usagetrendschart` baseline drift from [#70835](https://github.com/PostHog/posthog/pull/70835) described above — not something this logs PR changed or can legitimately bless. It needs the Endpoints owners to regenerate that snapshot baseline on `master` (or a maintainer to override the check for this PR). I've deliberately not approved an unrelated product's chart baseline through here, since the visual-review gate exists precisely to let that product's owners confirm the new rendering is intended rather than a regression.

PR #73573, opened 2026-07-24
feat(llm-gateway): add changelog bot to llm gateway routes

sarahxsanders:
## Problem

for the changelog bot agent, I need LLM gateway access for the inference calls (writing posts). it's minimal cost, so the cap is low

## Changes
- updates to gateway config files to add changelog bot route


services/llm-gateway/src/llm_gateway/products/config.py
richardsolomou, line 278: heyo can we make `changelog_bot` use exact model matching? the shared matcher uses `startswith`, so `openai/gpt-5.6-sol-pro` and `openai/gpt-5.6-terra-pro` both pass even though theyre distinct models in the cost registry. that defeats the model/cost pin, and the access test should reject both too.

richardsolomou (changes requested):

richardsolomou (approved):

PR #73576, opened 2026-07-24
fix(experiments): descriptive error when flag loses variants

rodrigoi:
## Problem

Opening the results of an experiment whose feature flag was converted from multivariate to a boolean or rollout flag broke: the exposure query returned HTTP 500 and the page gave no explanation.

- In that state `filters.multivariate` is explicitly `null`, so the runner's `dict.get("multivariate", {})` returned `None` and crashed with `AttributeError`.
- Without variants an experiment can't compute exposures or metrics at all, so coalescing `null` to `{}` would only trade the crash for an empty, endlessly refreshing results view. A typed error beats a silently empty chart.

## Changes

- Both experiment query runners now raise a descriptive 4xx `ValidationError` (code `feature_flag_variants_removed`) when the flag has no variants. A shared null-safe helper (`variants_from_flag_filters`) centralizes the extraction, and `FeatureFlag.variants` reuses the same logic.
- The results view shows an error banner for this state and stops the auto-refresh loop. A user-initiated reload refetches the experiment and its flag, then re-runs results, so the page recovers once the variants are restored (the exposure query embeds the flag snapshot, so a plain retry could not). The banner shows when the flag changed, read best-effort from the ungated per-flag activity endpoint.
- The feature flag serializer now blocks converting a flag away from multivariate while a non-deleted experiment references it, naming the blocking experiments. Only deleted experiments are exempt.
- The experiment precompute canary classifies this state as skipped instead of error, keeping its error gauge meaningful.

All four ReviewHog findings are addressed: manual reload recovery, the ungated activity endpoint, the canary skip classification, and dropping the suggestion to "detach" experiments (no such action exists).

- Jest (`experimentLogic.test.ts`, ran locally): detection sets the error state and records the change time; unrelated error codes are ignored; auto-refresh stops once variants are gone; a manual refresh clears the state and retries. The last case guards the reload-button recovery added for review.
- Backend (written, not run here: this sandbox has no Postgres/ClickHouse, so they run in CI): runner tests assert the 4xx and its code for `multivariate: null` and empty variants; serializer tests assert the conversion block and its exemptions; `test_flag_without_variants_is_skipped` guards the canary skip classification.
- Ran locally: full frontend typecheck, oxlint/oxfmt, ruff, repo-wide mypy, and `hogli ci:preflight`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None: the change replaces a crash with an in-product error state; no documented workflow changes.

## 🤖 Agent context

Authored with Claude Code via a PostHog Code task. Skills invoked: `/improving-drf-endpoints`, `/writing-tests`, `/writing-user-facing-copy`, `/writing-pr-descriptions`.

This revision rebased the branch onto current master (integrating master's new `excluded_variants` handling in both query runners) and folded in the four ReviewHog findings. The notable behavior decision: the variants-removed guard now only swallows `auto_refresh` triggers, so the error state is recoverable in-app instead of requiring a full page reload.

---

### DIFF
--- a/frontend/src/scenes/experiments/ExperimentView/FlagVariantsRemovedBanner.tsx
+++ b/frontend/src/scenes/experiments/ExperimentView/FlagVariantsRemovedBanner.tsx
@@ -1,0 +1,23 @@
+import { useValues } from 'kea'
+
+import { LemonBanner } from '@posthog/lemon-ui'
+
+import { dayjs } from 'lib/dayjs'
+
+import { experimentLogic } from '../experimentLogic'
+
+export function FlagVariantsRemovedBanner(): JSX.Element {
+    const { flagVariantsRemovedAt } = useValues(experimentLogic)
+
+    return (
+        <LemonBanner type="error">
+            <p className="font-semibold mb-1">This experiment's feature flag no longer has variants</p>
+            <p className="mb-0">
+                The flag was changed from a multivariate flag to a boolean or rollout flag
+                {flagVariantsRemovedAt ? ` on ${dayjs(flagVariantsRemovedAt).format('MMMM D, YYYY [at] h:mm A')}` : ''},
+                so exposures and metrics can't be computed. Restore the flag's variants, or delete this experiment, to
+                see results again.
+            </p>
+        </LemonBanner>
+    )
+}
--- a/frontend/src/scenes/experiments/experimentLogic.test.ts
+++ b/frontend/src/scenes/experiments/experimentLogic.test.ts
@@ -2119,4 +2119,66 @@ describe('experimentLogic', () => {
             })
         })
     })
+
+    describe('flag variants removed', () => {
+        const CHANGED_AT = '2024-06-01T10:30:00Z'
+
+        beforeEach(() => {
+            logic.actions.setExperiment(experiment)
+            useMocks({
+                get: {
+                    '/api/projects/:team/activity_log': {
+                        count: 1,
+                        next: null,
+                        previous: null,
+                        results: [
+                            {
+                                created_at: CHANGED_AT,
+                                detail: {
+                                    changes: [
+                                        {
+                                            field: 'filters',
+                                            before: {
+                                                multivariate: { variants: [{ key: 'control' }, { key: 'test' }] },
+                                            },
+                                            after: { multivariate: null },
+                                        },
+                                    ],
+                                },
+                            },
+                        ],
+                    },
+                },
+            })
+        })
+
+        it('flags the error and records the change time when exposures fail with the removed-variants code', async () => {
+            await expectLogic(logic, () => {
+                logic.actions.loadExposuresFailure('no variants', { code: 'feature_flag_variants_removed' } as any)
+            })
+                .toDispatchActions(['stopAutoRefreshInterval', 'setFlagVariantsRemoved'])
+                .toMatchValues({
+                    flagVariantsRemoved: true,
+                    flagVariantsRemovedAt: CHANGED_AT,
+                })
+        })
+
+        it('ignores exposure failures with an unrelated error code', async () => {
+            await expectLogic(logic, () => {
+                logic.actions.loadExposuresFailure('boom', { code: 'server_error' } as any)
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['setFlagVariantsRemoved'])
+                .toMatchValues({ flagVariantsRemoved: false })
+        })
+
+        it('stops re-running results once the flag has lost its variants', async () => {
+            logic.actions.setFlagVariantsRemoved(CHANGED_AT)
+            await expectLogic(logic, () => {
+                logic.actions.refreshExperimentResults(true, 'auto_refresh')
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['loadExposures', 'markRefreshStarted'])
+        })
+    })
 })
--- a/frontend/src/scenes/experiments/experimentLogic.test.ts
+++ b/frontend/src/scenes/experiments/experimentLogic.test.ts
@@ -2457,7 +2457,7 @@ describe('experimentLogic', () => {
             logic.actions.setExperiment(experiment)
             useMocks({
                 get: {
-                    '/api/projects/:team/activity_log': {
+                    '/api/projects/:team/feature_flags/:flag_id/activity': {
                         count: 1,
                         next: null,
                         previous: null,
@@ -2502,13 +2502,22 @@ describe('experimentLogic', () => {
                 .toMatchValues({ flagVariantsRemoved: false })
         })
 
-        it('stops re-running results once the flag has lost its variants', async () => {
+        it('stops auto-refreshing results once the flag has lost its variants', async () => {
             logic.actions.setFlagVariantsRemoved(CHANGED_AT)
             await expectLogic(logic, () => {
                 logic.actions.refreshExperimentResults(true, 'auto_refresh')
             })
                 .toFinishAllListeners()
                 .toNotHaveDispatchedActions(['loadExposures', 'markRefreshStarted'])
+        })
+
+        it('clears the error state and retries on a user-initiated refresh', async () => {
+            logic.actions.setFlagVariantsRemoved(CHANGED_AT)
+            await expectLogic(logic, () => {
+                logic.actions.refreshExperimentResults(true, 'manual')
+            })
+                .toDispatchActions(['clearFlagVariantsRemoved', 'markRefreshStarted'])
+                .toMatchValues({ flagVariantsRemoved: false })
         })
     })
 })
--- a/frontend/src/scenes/experiments/experimentLogic.test.ts
+++ b/frontend/src/scenes/experiments/experimentLogic.test.ts
@@ -2511,12 +2511,12 @@ describe('experimentLogic', () => {
                 .toNotHaveDispatchedActions(['loadExposures', 'markRefreshStarted'])
         })
 
-        it('clears the error state and retries on a user-initiated refresh', async () => {
+        it('reloads the experiment to recover on a user-initiated refresh', async () => {
             logic.actions.setFlagVariantsRemoved(CHANGED_AT)
             await expectLogic(logic, () => {
                 logic.actions.refreshExperimentResults(true, 'manual')
             })
-                .toDispatchActions(['clearFlagVariantsRemoved', 'markRefreshStarted'])
+                .toDispatchActions(['loadExperiment'])
                 .toMatchValues({ flagVariantsRemoved: false })
         })
     })
--- a/frontend/src/scenes/experiments/experimentLogic.test.ts
+++ b/frontend/src/scenes/experiments/experimentLogic.test.ts
@@ -2449,4 +2449,66 @@ describe('experimentLogic', () => {
             })
         })
     })
+
+    describe('flag variants removed', () => {
+        const CHANGED_AT = '2024-06-01T10:30:00Z'
+
+        beforeEach(() => {
+            logic.actions.setExperiment(experiment)
+            useMocks({
+                get: {
+                    '/api/projects/:team/activity_log': {
+                        count: 1,
+                        next: null,
+                        previous: null,
+                        results: [
+                            {
+                                created_at: CHANGED_AT,
+                                detail: {
+                                    changes: [
+                                        {
+                                            field: 'filters',
+                                            before: {
+                                                multivariate: { variants: [{ key: 'control' }, { key: 'test' }] },
+                                            },
+                                            after: { multivariate: null },
+                                        },
+                                    ],
+                                },
+                            },
+                        ],
+                    },
+                },
+            })
+        })
+
+        it('flags the error and records the change time when exposures fail with the removed-variants code', async () => {
+            await expectLogic(logic, () => {
+                logic.actions.loadExposuresFailure('no variants', { code: 'feature_flag_variants_removed' } as any)
+            })
+                .toDispatchActions(['stopAutoRefreshInterval', 'setFlagVariantsRemoved'])
+                .toMatchValues({
+                    flagVariantsRemoved: true,
+                    flagVariantsRemovedAt: CHANGED_AT,
+                })
+        })
+
+        it('ignores exposure failures with an unrelated error code', async () => {
+            await expectLogic(logic, () => {
+                logic.actions.loadExposuresFailure('boom', { code: 'server_error' } as any)
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['setFlagVariantsRemoved'])
+                .toMatchValues({ flagVariantsRemoved: false })
+        })
+
+        it('stops re-running results once the flag has lost its variants', async () => {
+            logic.actions.setFlagVariantsRemoved(CHANGED_AT)
+            await expectLogic(logic, () => {
+                logic.actions.refreshExperimentResults(true, 'auto_refresh')
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['loadExposures', 'markRefreshStarted'])
+        })
+    })
 })
--- a/frontend/src/scenes/experiments/experimentLogic.tsx
+++ b/frontend/src/scenes/experiments/experimentLogic.tsx
@@ -70,6 +70,7 @@ import {
 } from '~/queries/schema/schema-general'
 import { setLatestVersionsOnQuery } from '~/queries/utils'
 import {
+    ActivityScope,
     BreakdownAttributionType,
     BreakdownType,
     CohortType,
@@ -240,6 +241,42 @@ interface MetricLoadingSummary {
     successfulCount: number
     erroredCount: number
     cachedCount: number
+}
+
+// Mirrors FLAG_WITHOUT_VARIANTS_ERROR_CODE in
+// products/experiments/backend/hogql_queries/__init__.py — the 4xx code raised when the
+// experiment's flag was converted from multivariate to a boolean/rollout flag.
+const FLAG_VARIANTS_REMOVED_ERROR_CODE = 'feature_flag_variants_removed'
+
+function errorHasFlagVariantsRemovedCode(error: any): boolean {
+    return error?.code === FLAG_VARIANTS_REMOVED_ERROR_CODE
+}
+
+// Best-effort lookup of when the flag lost its variants, from the flag's activity log. Returns the
+// timestamp of the change where multivariate variants went from present to absent, or null if it
+// can't be found (activity unavailable, entry missing) — the error state degrades to no timestamp.
+async function findFlagVariantsRemovedAt(flagId?: number | null): Promise<string | null> {
+    if (!flagId) {
+        return null
+    }
+    try {
+        const response = await api.activity.list({ scope: ActivityScope.FEATURE_FLAG, item_id: flagId })
+        for (const item of response.results ?? []) {
+            for (const change of item.detail?.changes ?? []) {
+                if (change.field !== 'filters') {
+                    continue
+                }
+                const before = (change.before as any)?.multivariate?.variants ?? []
+                const after = (change.after as any)?.multivariate?.variants ?? []
+                if (before.length > 0 && after.length === 0) {
+                    return item.created_at ?? null
+                }
+            }
+        }
+    } catch {
+        // Best-effort — show the error without the timestamp if activity can't be read.
+    }
+    return null
 }
 
 function parseMetricErrorDetail(error: any): { detail: any; hasDiagnostics: boolean } {
@@ -537,6 +574,9 @@ export interface experimentLogicValues {
     exposuresLoading: boolean
     featureFlagValidationError: string | null
     firstPrimaryMetric: ExperimentFunnelsQuery | ExperimentMetric | ExperimentTrendsQuery | undefined
+    flagVariantsRemoved: boolean
+    flagVariantsRemovedAt: string | null
+    flagVariantsRemovedInfo: { changedAt: string | null } | null
     formMode: FormModes
     freezeExposureLoading: boolean
     getExperimentMetricType: (metric: ExperimentMetricUnion | undefined) => ExperimentMetricType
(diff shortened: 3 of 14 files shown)


frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2448,6 +2513,11 @@ export const experimentLogic = kea<experimentLogicType>([
             }
         },
         refreshExperimentResults: async ({ forceRefresh, triggeredBy, refreshIfStale }) => {
+            // The flag lost its variants — exposures and metrics can't be computed, so don't keep
+            // re-attempting (which would spin forever). The error state is shown instead.
+            if (values.flagVariantsRemovedInfo) {
+                return
+            }
```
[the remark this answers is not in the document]
rodrigoi: Valid. Fixed in b4c58a0: `refreshExperimentResults` now only short-circuits `auto_refresh` triggers. Any user-initiated refresh (manual, page load, config change) dispatches the new `clearFlagVariantsRemoved` action and retries for real, so the reload button recovers the results view once the flag's variants are restored; if they are still missing, `loadExposuresFailure` re-arms the error state. Covered by a new jest case ("clears the error state and retries on a user-initiated refresh").

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -242,6 +243,42 @@ interface MetricLoadingSummary {
...
+                if (before.length > 0 && after.length === 0) {
+                    return item.created_at ?? null
+                }
+            }
+        }
+    } catch {
+        // Best-effort — show the error without the timestamp if activity can't be read.
+    }
+    return null
+}
```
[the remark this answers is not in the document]
rodrigoi: Valid. Fixed in b4c58a0: `findFlagVariantsRemovedAt` now uses `api.activity.listLegacy({ scope: ActivityScope.FEATURE_FLAG, id })`, which routes to the ungated per-flag `/feature_flags/:id/activity` endpoint (the same path the flag History tab uses), so the timestamp lookup works for orgs without the audit logs add-on.

products/experiments/backend/hogql_queries/experiment_query_runner.py
```diff
@@ -241,6 +246,13 @@ def __init__(
         self.group_type_index = self.feature_flag.aggregation_group_type_index
         self.entity_key = get_entity_key(self.group_type_index)
 
+        # An experiment can't be scored without variants to split metrics across. This happens when
+        # the flag was converted from multivariate to a plain boolean/rollout flag (see
+        # FeatureFlag.variants — filters.multivariate is null or has an empty variants list). Raise a
+        # 4xx so the UI can explain it instead of failing later with an opaque error.
+        if not self.feature_flag.variants:
+            raise ValidationError(FLAG_WITHOUT_VARIANTS_ERROR_MESSAGE, code=FLAG_WITHOUT_VARIANTS_ERROR_CODE)
+
```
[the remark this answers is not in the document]
rodrigoi: Valid. Fixed in b4c58a0: `run_metric_canary_sync` now returns a skipped result when a run raises the `feature_flag_variants_removed` ValidationError, consistent with the other deterministic dead ends, so this state no longer pollutes the canary error gauge or burns retries. Added `test_flag_without_variants_is_skipped`, which exercises the real runner raise end to end.

products/feature_flags/backend/api/feature_flag.py
```diff
@@ -1155,6 +1157,35 @@ def _validate_archived_flags_are_disabled(self, attrs: dict) -> None:
...
+        experiments = list(self.instance.experiment_set.filter(deleted=False))
+        if not experiments:
+            return
+
+        experiment_names = ", ".join(f'"{exp.name}" (ID: {exp.id})' for exp in experiments)
+        raise serializers.ValidationError(
+            f"Cannot remove this flag's variants while it is linked to experiment(s): {experiment_names}. "
+            "Converting the flag to boolean or rollout would break those experiments' exposure and "
+            "metric results. Delete or detach the experiment(s) first."
+        )
```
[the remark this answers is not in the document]
rodrigoi: Valid. Fixed in b4c58a0: the message now ends with "Delete the experiment(s) first." and the "or detach" clause is gone.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2574,6 +2647,17 @@ export const experimentLogic = kea<experimentLogicType>([
...
+            // The flag lost its variants — exposures and metrics can't be computed, so the
+            // auto-refresh loop must not keep re-attempting (it would spin forever). A
+            // user-initiated refresh clears the error state and retries for real, so the "Reload
+            // results" button recovers once the flag's variants are restored; if they're still
+            // missing, loadExposuresFailure re-arms the state.
+            if (values.flagVariantsRemovedInfo) {
+                if (triggeredBy === 'auto_refresh') {
+                    return
+                }
+                actions.clearFlagVariantsRemoved()
```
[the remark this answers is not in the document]
rodrigoi: Valid. Fixed in 4112aed: a user-initiated refresh in this state now dispatches `loadExperiment()` instead of just clearing the error state. That refetches the experiment with its current flag, clears the state via the reducer, and `loadExperimentSuccess` re-runs results with the fresh flag, so recovery works without a full page reload. If the variants are still missing, `loadExposuresFailure` re-arms the banner. The jest case now asserts the reload path.

frontend/src/scenes/experiments/experimentLogic.tsx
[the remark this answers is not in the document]
rodrigoi: Update: the recovery mechanism changed in 4112aed after a follow-up review finding. The exposure query embeds the flag snapshot held in state, so clearing the state and retrying would re-send the stale variant-less flag. A user-initiated refresh now dispatches `loadExperiment()` (refetch flag, clear state, re-run results on success) instead.

frontend/src/scenes/experiments/experimentLogic.test.ts
```diff
@@ -2449,4 +2449,75 @@ describe('experimentLogic', () => {
...
+        it('stops auto-refreshing results once the flag has lost its variants', async () => {
+            logic.actions.setFlagVariantsRemoved(CHANGED_AT)
+            await expectLogic(logic, () => {
+                logic.actions.refreshExperimentResults(true, 'auto_refresh')
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['loadExposures', 'markRefreshStarted'])
+        })
+
+        it('reloads the experiment to recover on a user-initiated refresh', async () => {
```
haacked, line 2514: `blocking`: The test doesn't exercise the full recovery chain. `experimentLogic` is mounted unkeyed, so `experimentId` is `'new'` and `loadExperiment` short-circuits to the `NEW_EXPERIMENT` constant without fetching anything. That constant has no `start_date`, so `loadExperimentSuccess` never reaches `refreshExperimentResults`, and both assertions resolve before the success action. Mount a keyed logic on a launched experiment fixture as the other tests do, and extend the assertion to the full action chain through `loadExposures`.

products/experiments/backend/hogql_queries/__init__.py
```diff
@@ -2,6 +2,34 @@
...
+# explain the state instead of hitting a 500 or spinning forever.
+FLAG_WITHOUT_VARIANTS_ERROR_CODE = "feature_flag_variants_removed"
+FLAG_WITHOUT_VARIANTS_ERROR_MESSAGE = (
+    "This experiment's feature flag no longer defines any variants, so exposures and metrics "
+    "can't be computed. The flag was changed from a multivariate (experiment) flag to a boolean "
+    "or rollout flag. Restore the flag's variants, or delete the experiment, to resolve this."
+)
+
+
+def variants_from_flag_filters(filters: dict | None) -> list:
```
haacked, line 18: `suggestion`: `variants_from_flag_filters` has the same body as `variants_from_filters` in `products/feature_flags/backend/models/feature_flag.py`, and the two copies have already drifted in type hints and docstring wording, so the next person tightening the null handling is likely to fix only one. Nothing forces the duplication: other files in `products/experiments/backend` already import free functions from that model module.

Consider deleting this copy and importing `variants_from_filters` in `experiment_exposures_query_runner.py`, its only caller.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -3603,6 +3684,17 @@ export const experimentLogic = kea<experimentLogicType>([
...
             }
         },
+        loadExposuresFailure: async ({ errorObject }) => {
+            // The flag was converted from multivariate to boolean/rollout, so exposures and metrics
+            // can't be computed. Stop the auto-refresh loop — retrying can't succeed until the flag
+            // regains variants — and surface a descriptive state with the change time when we can find it.
+            if (values.flagVariantsRemovedInfo || !errorHasFlagVariantsRemovedCode(errorObject)) {
+                return
+            }
+            actions.stopAutoRefreshInterval()
```
haacked, line 3694: `suggestion`: Stopping the interval here only works temporarily. The `refreshExperimentResults` call whose exposure query just failed runs its `finally` block, which re-arms the interval for a launched experiment on a visible tab. Neither that block nor `resetAutoRefreshInterval` checks `flagVariantsRemovedInfo`, so every later tick fires `reportExperimentMetricsRefreshed` with `triggered_by: 'auto-refresh'` before the new guard returns early. The analytics stream records refreshes that never happened for as long as the tab stays open.

Consider gating the re-arm on the new state, e.g. `values.autoRefresh.enabled && !values.flagVariantsRemovedInfo` inside `resetAutoRefreshInterval`, so every re-arm path respects it.

products/experiments/backend/temporal/canary_logic.py
```diff
@@ -377,7 +379,13 @@ def _skipped(detail: str) -> CanaryMetricResult:
             query_id = f"experiment-canary-{canary_id}-{label}"
             try:
                 runs.append(_execute_canary_run(experiment, metric, mode, label, query_id))
-            except Exception:
+            except Exception as e:
+                # A flag converted away from multivariate is a deterministic dead end (the runner
+                # refuses in __init__ on every attempt), not a precompute regression — classify it
+                # as skipped so it doesn't pollute the canary's error gauge.
+                codes = e.get_codes() if isinstance(e, ValidationError) else None
+                if isinstance(codes, list) and FLAG_WITHOUT_VARIANTS_ERROR_CODE in codes:
```
haacked, line 387: `suggestion`: The canary classifies a flag that lost its variants as `skipped`, but the other two background paths treat it as a genuine failure. When a flag is converted away from multivariate while a recalculation or a scheduled timeseries backfill runs, `recalculation_logic.py` calls `capture_exception` and emits an `experiment metric error` event on every cycle. The Temporal activity in `posthog/temporal/experiments/activities.py` stores `FAILED` and re-raises without marking the error non-retryable, so the workflow burns all three retry attempts on a state that cannot succeed.

Consider extracting the classification check into a shared predicate and consulting it from both paths: skip the capture and the event in recalculation, and raise non-retryable in the activity.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2574,6 +2642,19 @@ export const experimentLogic = kea<experimentLogicType>([
...
         refreshExperimentResults: async ({ forceRefresh, triggeredBy, refreshIfStale }) => {
+            // The flag lost its variants — exposures and metrics can't be computed, so the
+            // auto-refresh loop must not keep re-attempting (it would spin forever). A
+            // user-initiated refresh reloads the experiment instead: the exposure query embeds
+            // the flag snapshot held in state (still the variant-less one), so a plain retry
+            // would fail even after the flag was fixed. Reloading refetches the flag, clears
+            // this state (reducer), and re-runs results on success; if variants are still
+            // missing, loadExposuresFailure re-arms the state.
+            if (values.flagVariantsRemovedInfo) {
+                if (triggeredBy !== 'auto_refresh') {
```
haacked, line 2653: `suggestion`: While the flag-variants-removed banner is showing, saving a metric, date, or exposure criteria blanks the whole page to a loading skeleton instead of updating in place. The guard treats every trigger except `'auto_refresh'` as user-initiated. A `'config_change'` refresh dispatches `loadExperiment()`, which sets `experimentLoading` and swaps the header and tabs for the skeleton that the separate `experimentUpdate` loader exists to avoid. Since callers like `updateExposureCriteria` don't await the `updateExperiment` PATCH before refreshing, the GET can land after it and momentarily show the just-saved changes disappearing from the form.

Consider narrowing the condition to `triggeredBy === 'manual'`, leaving `config_change` and `page_load` refreshes as they were.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -245,6 +246,44 @@ interface MetricLoadingSummary {
     cachedCount: number
 }
 
+// Mirrors FLAG_WITHOUT_VARIANTS_ERROR_CODE in
+// products/experiments/backend/hogql_queries/__init__.py — the 4xx code raised when the
+// experiment's flag was converted from multivariate to a boolean/rollout flag.
+const FLAG_VARIANTS_REMOVED_ERROR_CODE = 'feature_flag_variants_removed'
```
haacked, line 252: `suggestion`: The banner only appears after the exposure query round-trips and fails, so a user opening an experiment whose flag already lost its variants watches a spinner first. The flag's `filters` are already in the experiment payload through `MinimalFeatureFlagSerializer`. The other flag-state warnings in this file derive their state from `experiment.feature_flag.filters` with no query, and deriving this one the same way would remove the hand-mirrored error code, the failure listener, the activity-log lookup, and their typegen entries.

Consider making `flagVariantsRemoved` a plain selector over `experiment.feature_flag.filters`, with a `hasNoVariants(filters)` helper next to `hasMultipleVariantsActive`. The recovery guard still needs to stay, since `loadExperiment()` is what refetches the flag. The tradeoff is that the banner loses the best-effort change timestamp.

frontend/src/scenes/experiments/ExperimentView/FlagVariantsRemovedBanner.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export function FlagVariantsRemovedBanner(): JSX.Element {
+    const { flagVariantsRemovedAt } = useValues(experimentLogic)
+
+    return (
+        <LemonBanner type="error">
+            <p className="font-semibold mb-1">This experiment's feature flag no longer has variants</p>
+            <p className="mb-0">
+                The flag was changed from a multivariate flag to a boolean or rollout flag
+                {flagVariantsRemovedAt ? ` on ${dayjs(flagVariantsRemovedAt).format('MMMM D, YYYY [at] h:mm A')}` : ''},
```
haacked, line 17: `suggestion`: The banner tells the user to restore the flag's variants but never names the flag, so they have to go find it themselves. The five other flag-state banners in `ExperimentWarningBanners.tsx` all render the flag key as a link via `urls.featureFlag(experiment.feature_flag.id)`.

Consider rendering the flag key as the same link here.

products/feature_flags/backend/api/feature_flag.py
```diff
@@ -1161,6 +1163,35 @@ def _validate_archived_flags_are_disabled(self, attrs: dict) -> None:
...
+        _MISSING: Any = object()
+        new_filters = attrs.get("get_filters", _MISSING)
+        if new_filters is _MISSING:
+            return
+
+        # Only the multivariate -> boolean/rollout transition matters: had variants, now has none.
+        if not self.instance.variants or variants_from_filters(new_filters):
+            return
+
+        experiments = list(self.instance.experiment_set.filter(deleted=False))
```
haacked, line 1184: `suggestion`: Any non-deleted experiment now blocks converting a flag away from multivariate, including a draft, completed, or archived one. The message's only way forward is to delete the experiment. The flag-deletion guard a few hundred lines below draws the line at running experiments, and its comment says draft, stopped, and completed experiments may keep the flag. A user with an old archived experiment loses the ability to convert the flag unless they destroy the experiment history that guard was written to preserve.

Consider exempting archived experiments, e.g. `experiment_set.filter(deleted=False, archived=False)`, matching the non-destructive escape hatch the deletion guard already relies on.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2574,6 +2642,19 @@ export const experimentLogic = kea<experimentLogicType>([
...
         },
         refreshExperimentResults: async ({ forceRefresh, triggeredBy, refreshIfStale }) => {
+            // The flag lost its variants — exposures and metrics can't be computed, so the
+            // auto-refresh loop must not keep re-attempting (it would spin forever). A
+            // user-initiated refresh reloads the experiment instead: the exposure query embeds
+            // the flag snapshot held in state (still the variant-less one), so a plain retry
+            // would fail even after the flag was fixed. Reloading refetches the flag, clears
+            // this state (reducer), and re-runs results on success; if variants are still
+            // missing, loadExposuresFailure re-arms the state.
+            if (values.flagVariantsRemovedInfo) {
```
haacked, line 2652: `suggestion`: The recovery branch calls `actions.loadExperiment()` without the `triggeredBy` it was handed, unlike every other `loadExperiment` call. On success, `loadExperimentSuccess` re-dispatches `refreshExperimentResults(false, payload?.triggeredBy ?? 'manual', true)`, so a `config_change` that recovers the page is reported as a manual refresh. This skips the metrics recalculation a genuine config-change trigger would run.

Forward it as `actions.loadExperiment({ triggeredBy })`, or add a short comment if skipping recalculation here is deliberate.

frontend/src/scenes/experiments/experimentLogic.test.ts
```diff
@@ -2449,4 +2449,75 @@ describe('experimentLogic', () => {
...
+                    flagVariantsRemovedAt: CHANGED_AT,
+                })
+        })
+
+        it('ignores exposure failures with an unrelated error code', async () => {
+            await expectLogic(logic, () => {
+                logic.actions.loadExposuresFailure('boom', { code: 'server_error' } as any)
+            })
+                .toFinishAllListeners()
+                .toNotHaveDispatchedActions(['setFlagVariantsRemoved'])
```
haacked, line 2501: `suggestion`: This test only asserts that `setFlagVariantsRemoved` isn't dispatched for an unrelated error code. The same guard in `loadExposuresFailure` also protects `stopAutoRefreshInterval`. A regression that stopped auto-refresh for any exposure error would pass here while silently ending the refresh loop after a transient failure.

Extend the assertion to `.toNotHaveDispatchedActions(['stopAutoRefreshInterval', 'setFlagVariantsRemoved'])`.

products/feature_flags/backend/api/feature_flag.py
```diff
@@ -1161,6 +1163,35 @@ def _validate_archived_flags_are_disabled(self, attrs: dict) -> None:
...
+    def _validate_experiment_variants_preserved(self, attrs: dict) -> None:
+        """Block converting a flag from multivariate to boolean/rollout while a non-deleted
+        experiment references it. Every such experiment — draft, running, or ended — runs an
+        exposure query that needs the flag's variants, so dropping them breaks the results view.
+        Only deleted experiments are exempt."""
+        if self.instance is None:
+            return
+
+        # `get_filters` is the DictField's source; absent means filters aren't being updated.
+        _MISSING: Any = object()
```
haacked, line 1175: `nit`: The `_MISSING` sentinel only tests whether `get_filters` is present, since the resolved value is discarded when it's absent, and a plain membership check already says that. Perhaps use `if "get_filters" not in attrs: return` followed by `new_filters = attrs["get_filters"]`.

frontend/src/scenes/experiments/ExperimentView/FlagVariantsRemovedBanner.tsx
haacked, line 17: `nit`: This renders the change time as an unlabeled browser-local timestamp, while the same activity entry on the flag's History tab goes through `<TZLabel>`, which gives a hover popover with project, UTC, and local time. Perhaps use `<TZLabel time={flagVariantsRemovedAt} timestampStyle="absolute" />` here, with `absolute` since TZLabel otherwise renders relative text.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2141,6 +2199,16 @@ export const experimentLogic = kea<experimentLogicType>([
                 clearMetricsResults: () => [],
             },
         ],
+        // Set once the exposure/metric query reports the flag no longer has variants. Holds the
+        // best-effort timestamp of the change (or null). Reset when the experiment is reloaded,
+        // which is also how a user-initiated refresh recovers from this state.
+        flagVariantsRemovedInfo: [
+            null as { changedAt: string | null } | null,
```
haacked, line 2206: `nit`: `{ changedAt: string | null } | null` is spelled out four times here: the values interface, the reducer, and both selectors, where this file's convention elsewhere is a named interface such as `ExperimentWarning`. Perhaps declare `FlagVariantsRemovedInfo` and use it in the reducer and both selectors so typegen regenerates the rest.

frontend/src/scenes/experiments/experimentLogic.tsx
```diff
@@ -2574,6 +2642,19 @@ export const experimentLogic = kea<experimentLogicType>([
             }
         },
         refreshExperimentResults: async ({ forceRefresh, triggeredBy, refreshIfStale }) => {
+            // The flag lost its variants — exposures and metrics can't be computed, so the
```
haacked, line 2645: `nit`: The recovery mechanism is explained in full three times in this file: on the reducer, on this guard, and in the failure listener, so a later change to the recovery path has to keep three explanations in sync. Perhaps keep the full version here and cut the other two to one line each.

products/feature_flags/backend/api/test/test_feature_flag.py
```diff
@@ -4048,6 +4048,81 @@ def test_soft_delete_flag_allowed_with_non_running_experiment(self, _name, start
...
+                "multivariate": {
+                    "variants": [
+                        {"key": "control", "rollout_percentage": 50},
+                        {"key": "test", "rollout_percentage": 50},
+                    ]
+                },
+            },
+        )
+
+    @parameterized.expand(
```
haacked, line 4067: `nit`: In this parameterization `None` means "no `multivariate` key at all" while the string `"null"` means an explicit null, which reads backwards and forces the `if multivariate == "null"` branch inside the test. Perhaps pass the filters payload itself as the parameter instead.

haacked (commented):
LGTM! A couple non-blocking suggestions inline, plus one blocking test-coverage gap.

PR #73582, opened 2026-07-24
refactor: return frozen dataclasses from two swap-prone tuple helpers

arnohillen:
## Problem

Two internal helpers returned positional tuples whose elements share a type, so a reordered unpack swaps them silently with no type error:

- `_get_state_class` returned `(state_class, partial_state_class)` — two bare `type` values.
- `update_all_orgs_billing_quotas` returned `(quota_limited_orgs, quota_limiting_suspended_orgs, stats)` where the two org maps are both `dict[str, dict[str, int]]`. This one runs on a cron and drives billing quota limiting, so getting the two maps backwards would be a nasty bug.

Same idea as the sibling refactor in #73574, applied to the ee/hogai and ee/billing helpers.

## Changes

Each helper now returns a small frozen dataclass with named fields, and every call site reads by attribute. Frozen dataclasses are non-iterable on purpose, so any leftover positional unpack fails loudly instead of quietly swapping.

No behavior change. Backwards compatible.

| Function | Old return | New return |
| --- | --- | --- |
| `_get_state_class` | `(state_class, partial_state_class)` | `StateClasses(state_class, partial_state_class)` |
| `update_all_orgs_billing_quotas` | `(quota_limited_orgs, quota_limiting_suspended_orgs, stats)` | `BillingQuotaUpdateResult(quota_limited_orgs, quota_limiting_suspended_orgs, stats)` |

Call sites touched: the taxonomy agent and its nodes for `_get_state_class`, and the management command, the quota-limiting Temporal activity, and the quota-limiting test suite for `update_all_orgs_billing_quotas`.

> [!NOTE]
> The Temporal edit is in the activity body, not a `@workflow.defn`, so it does not affect workflow determinism or in-flight executions.

No manual testing (mechanical type-shape refactor). Automated checks I (Claude) actually ran:

- Repo-wide mypy the way CI runs it (`uv run mypy --cache-fine-grained .`): clean across the tree.
- The affected suites (`ee/billing/test/test_quota_limiting.py`, taxonomy agent + nodes tests): all green.
- `ruff check` and `ruff format`: clean, no unused-variable stragglers from the rewritten unpacks.

No new tests. Existing tests that unpacked these returns were mechanically updated to read the new attributes; the dataclasses' non-iterability is itself the regression guard against a stale positional unpack.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Arnaud directed this as part of a series of small, safe refactors turning swap-prone positional tuple returns into named frozen dataclasses. Tool: Claude Code (Opus 4.8).

Decisions:

- Frozen dataclass over `NamedTuple` on purpose: a `NamedTuple` stays iterable, so a stale `a, b = f(...)` would keep working and could still swap. A frozen dataclass makes that mistake fail loudly.
- For the ~33 quota-limiting call sites I kept the local variable names via attribute reads (`quota_update.quota_limited_orgs`) rather than rewriting every downstream usage, which keeps the diff mechanical and the test bodies unchanged.
- Verified every call site three ways (grep for remaining positional unpacks, repo-wide mypy, tests) rather than grep alone.


pawel-cebula (approved):
looks good, thanks for catching this!

PR #73586, opened 2026-07-24
feat(conversations): add clear filters button to tickets view

christiaan-ph:
## Problem

A support agent who has narrowed the tickets view has to clear each filter one at a time. The bar holds nine filters (status, priority, channel, SLA, AI triage, tags, assignee, date, search) and nothing resets them together.

## Changes

- The tickets filter bar gains a "Clear all filters" button after the assignee selector.
- The button calls the existing `resetFilters` action, which clears the filters and any active saved view.
- No new state. `resetFilters` already existed, surfaced only as a side action on the Saved Views button while a view was active.
- The button is a `secondary` LemonButton with no icon, so it matches the sibling filter controls.
- It carries no leading `IconX`. The siblings use `IconX` as a right-edge `sideAction` meaning "clear this one filter", so a leading copy of the same glyph would read as that narrower action.

> [!NOTE]
> `hasActiveFilters` gates the button, and that selector counts the date range as a filter. `dateFrom` defaults to `-7d`, so the button shows on a fresh load before the agent picks anything. Clicking it then widens the range to all time, because `resetFilters` falls back to a null date range when no saved view was applied.

CleanShot 2026-08-07 at 11 10 07@2x

I ran this branch on a local stack, with current master merged in, and drove the tickets scene headlessly with Playwright. No committed test was added.

What I confirmed in the browser:

- The button appears once a filter is active.
- Clicking it clears the filters, re-queries the list, and hides the button.
- The button's computed styling matches the sibling filter dropdowns.
- The filter bar fits on one row at 1700px.

What I did not do:

- No automated test. The change is presentational and reuses an action that already ships.
- `typescript:check` is not clean on this branch. It reports 179 errors, all in generated kea `*LogicType.ts` files across other products, none in conversations. Merging master left that typegen stale, so the errors predate this change.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Requested via a Slack thread. The first version came from the PostHog Slack app on Opus 4.8. It found that `resetFilters` already existed in `supportTicketsSceneLogic` and surfaced it as a `LemonButton` gated on the existing `hasActiveFilters` selector, rather than adding new state.

A later Claude Code session on Opus 5 merged current master in and brought the branch up on a local stack. It then revised the button after review. `tertiary` became `secondary` so the control renders as a button rather than plain text, the label became "Clear all filters", and the leading `IconX` came out. That session also found the date-default behavior recorded in the note above. Skills invoked: `writing-user-facing-copy`, `writing-pr-descriptions`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C075D3C5HST/p1784912112445459?thread_ts=1784912112.445459&cid=C075D3C5HST)*


christiaan-ph:
need to check how this looks visually...

veryayskiy (approved):

PR #73588, opened 2026-07-24
feat(data-warehouse): add deltalite streaming partition upsert crate

Gilbert09:
## Problem

The warehouse-sync incremental path writes to Delta tables through delta-rs SQL `MERGE`, which executes a DataFusion hash join. Under a bounded memory pool the join deadlocks instead of spilling ([delta-io/delta-rs#4614](https://github.com/delta-io/delta-rs/issues/4614), reproduced), and unbounded its memory scales with the scanned target, OOMing worker pods on large tables. A validated spike proved a purpose-built streaming partition upsert completes the same work with ~4x less memory that does not grow with table size, at matching write volume.

## Changes

Adds `rust/deltalite`, a library crate replacing only the merge *execution* – delta-rs stays the storage and protocol layer (log, checkpoints, Parquet writing, Add-action stats, conditional-put commits, conflict resolution). Core algorithm: build a PK hash set over the (small) source, stream the (large) target one row group at a time, drop replaced rows, write survivors then source rows, commit all partitions in one atomic commit carrying the `run_uuid`/`batch_index` idempotency tag.

Two crates, both workspace members:

- `rust/deltalite/core` – all the logic, pure Rust, holds all 39 Rust unit tests. Pins `deltalake = "=0.32.4"` without the `datafusion` feature, so the #4614 deadlock is impossible by construction.
- `rust/deltalite/python` – thin pyo3/maturin binding (abi3-py312, imported as `deltalite`), shaped exactly like `rust/hogql/parser` (cdylib+rlib, extension-module, macOS `dynamic_lookup` build.rs, no Rust tests – `extension-module` test harnesses cannot link on Linux, which is why the logic lives in the pyo3-free core crate).

**Stacked on #73619** (base branch `tom/hogql-parser-pyo3-028`), which migrates `hogql_parser_rs` from pyo3 0.22 to 0.28. Cargo allows one pyo3 version per workspace (`links = "python"`), and deltalite's pyarrow ingest (arrow 58's `pyarrow` feature, arrow major fixed by deltalake) requires 0.28 – so the parser moves up first, and this PR's binding crate joins the workspace cleanly.

Semantics preserved from the spike, each pinned by tests: SQL NULL semantics on primary keys (NULL never matches, always inserts), identity is `(PK, partition)` (a PK moving partition leaves the old row behind), duplicate source PKs rejected (stricter than MERGE, which silently double-inserts), exact content-based file selection (`prune_strategy=probe` reads only the PK column via projection pushdown – this closes the 32x write-amplification gap with random UUID keys), decimal buffer realignment on ingest (delta-rs#3884), `target_file_size` resolved as argument > `delta.targetFileSize` > delta-rs default, source rows written last within each partition (this ordering is what makes an update carry source values), and refusal of tables with deletion vectors or column mapping (including the legacy `minReaderVersion=2` form the feature-list check misses).

Hardening beyond the spike:

- **Process-global concurrency budgets** (`DELTALITE_PROCESS_MAX_PARALLEL_PARTITIONS` / `_FILES` / `_MAX_BUFFERED_BYTES`) enforced on top of the per-call knobs, so ~15 concurrent upsert threads in one Temporal worker cannot multiply the memory bound the spike measured per call.
- **Source-size guard**: a batch whose resident size + estimated PK set exceeds a configurable ceiling (default 2 GiB) fails with a typed `DeltaLiteSourceTooLargeError` at the front door instead of growing RSS until the pod dies.
- **Multipart upload** for large output files via a delegating `LogStore`/`ObjectStore` wrapper – `RecordBatchWriter` uploads each file as a single ~100 MB PUT and exposes no hook, but it writes through the table's object store, so the wrapper turns large plain-overwrite puts into multipart uploads while forwarding conditional puts (Delta log commits) untouched.
- Typed error hierarchy mapped onto a `DeltaLiteError` Python exception tree, `tracing` spans, and `metrics` counters/histograms with static label values only.
- A small knob-tuning helper (`python/deltalite_planner.py`): concurrent upserts + pod memory in, rough `max_parallel_partitions`/`max_parallel_files`/byte-budget suggestions and process-global env ceilings out. Deliberately a heuristic, not a fitted capacity model.

Not wired into `DeltaTableHelper` yet – the flag-gated seam is a follow-up, as are the spike's remaining pre-rollout conditions (real-S3 verification, conditional-put startup probe, `vacuum(full=True)` policy), documented in `rust/deltalite/README.md`.

> [!NOTE]
> `rust/Cargo.lock` grows by ~62 packages (the deltalake tree). Nothing outside `rust/deltalite` compiles any of them, but cold CI compile times for the workspace cache will grow accordingly, and `deltalite-core` itself is compile-heavy (weighted 180s in the test shard table).

All automated, all run locally:

- **159/159 pytest tests** (`rust/deltalite/python/tests`) against `deltalake==1.6.1` + `pyarrow==23.0.1` (the repo's pins), re-run against the final workspace-built wheel. The centrepiece is the differential parity harness: every scenario runs the same batch sequence through the real delta-rs `merge` (driven exactly as `DeltaTableHelper` drives it) and through `deltalite.upsert`, then asserts identical logical content and schema. Covers NULL PKs, composite PKs, partition moves, duplicates, schema evolution, decimal misalignment, crash/kill-before-commit + redelivery, interop (DuckDB `delta_scan`, compact/vacuum interleaving), the knob-tuning helper, plus new tests for the source guard and a forced-multipart parity run. The ported test set is function-for-function identical to the spike's 171 non-S3 tests (the 23 S3 tests need a live MinIO and were not ported).
- **39 Rust unit tests** in `deltalite-core` for the pure logic the spike only tested through Python: PK-set NULL semantics, row selection, stats-pruning decisions, target-file-size resolution, buffer realignment, source guard, multipart splitting and conditional-put passthrough.
- `cargo check --workspace --all-features` (498 crates, 0 errors), `cargo build -p deltalite-core --locked`, `cargo clippy --all-targets` on the three touched crates, `cargo fmt --check`, `cargo shear`, `ruff check`/`ruff format --check` – all clean.
- New `ci-deltalite-python.yml` workflow reproduces the wheel build + pytest run in CI, path-filtered to `rust/deltalite/**`.

## Automatic notifications

- [ ] Publish to changelog?


gantoine (approved):

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,1793 @@
...
+    let reader = ParquetObjectReader::new(store, path).with_file_size(f.size);
+    let builder = ParquetRecordBatchStreamBuilder::new(reader)
+        .await?
+        .with_batch_size(opts.read_batch_size);
+    let mut stream = builder.build()?;
+
+    let mut rows_updated = 0usize;
+    let mut rows_copied = 0usize;
+
+    while let Some(batch) = stream.try_next().await? {
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 5243e0e. Readers now reserve budget **before** `try_next()` decodes: an adaptive estimate (last batch's size, 4 MiB initially) is acquired up front, then topped up to the actual decoded size, with the total capped at each budget's capacity so an oversized batch still makes progress instead of deadlocking. The probe loop is budgeted the same way (256 KiB initial estimate for the narrow PK-column batches). Covered by a new unit test (`budget_permit_tops_up_and_caps_at_capacity`) plus the full parity suite rerun.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,1794 @@
...
+                return Err(Error::Generic(format!(
+                    "partition column '{col}' contains NULL at row {}; \
+                     Delta partition values must not be null",
+                    row_base + i
+                )));
+            }
+            let v = as_str.value(i);
+            if !indices.contains_key(v) {
+                order.push(v.to_string());
+                indices.insert(v.to_string(), vec![Vec::new(); n_batches]);
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 5243e0e. Planning is now sparse: selections are stored as (batch index, selection) pairs per contributing batch only, with contiguous rows of one batch merged into a single entry — so cost is O(total rows + contributing pairs) rather than O(distinct partitions x batches), and `PartitionSource.sel` no longer materialises a dense matrix either. Covered by new unit tests (`plan_stores_selections_sparsely_per_contributing_batch`, `plan_merges_contiguous_rows_of_one_batch_into_one_selection`).

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,1946 @@
...
+            held_global_kb: global_kb,
+        })
+    }
+
+    /// Grow `permit` so it covers `total_bytes` (used after a decode turns out larger
+    /// than the pre-decode estimate). Total held stays capped at each budget's
+    /// capacity, preserving the oversized-batch progress guarantee; a decode smaller
+    /// than the estimate keeps the estimate's permits until release, which only errs
+    /// conservative.
+    async fn top_up(&self, permit: &mut BudgetPermit, total_bytes: usize) -> Result<()> {
```
estefaniarabadan, line 430: (sent from claude-code): `top_up` acquires extra budget while holding the estimate's permits and the decoded batch; if all current holders block here at once (per-call or process-global budget), nothing can ever release and every upsert in the worker wedges permanently — release-and-reacquire on shortfall, or pre-reserve from Parquet row-group metadata.

rust/deltalite/core/src/schema.rs
```diff
@@ -0,0 +1,262 @@
...
+    // arrow-rs represents Decimal256 as `i256`, a `#[repr(C)]` pair of 128-bit lanes, so
+    // both decimal widths require 16-byte alignment.
+    let align = match data.data_type() {
+        DataType::Decimal128(_, _) | DataType::Decimal256(_, _) => {
+            Some(std::mem::align_of::<i128>())
+        }
+        _ => None,
+    };
+
+    let mut children = Vec::with_capacity(data.child_data().len());
```
estefaniarabadan, line 48: (sent from claude-code): `realign_array_data` detects repaired children by comparing the child's first buffer pointer, which never changes for Struct (no buffers) or List/Map intermediates, so a misaligned decimal at depth ≥ 2 (e.g. `map<string, decimal>`) is realigned then discarded and the delta-rs#3884 panic still fires — return `(ArrayData, changed)` from the recursion instead.

.github/workflows/ci-deltalite-python.yml
```diff
@@ -0,0 +1,82 @@
...
+# the compatibility check that replaces version-lockstep between the `deltalake`
+# Rust crate and the `deltalake` Python package (see rust/deltalite/README.md).
+#
+# ci-rust.yml covers the crates' build/lint/test; this workflow adds the thing
+# cargo cannot do -- build the actual wheel with maturin and run the Python
+# suite against the repo's pinned deltalake/pyarrow versions.
+
+on:
+    pull_request:
+        paths:
```
estefaniarabadan, line 15: (sent from claude-code): the `paths` filter omits `pyproject.toml`/`uv.lock`, so a `deltalake`/`pyarrow` pin bump never re-runs the parity suite this workflow exists to provide, and the hardcoded pins here can silently drift from the repo's.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,1946 @@
...
+            .sum::<usize>();
+    }
+    Ok(SourceFootprint {
+        source_bytes,
+        pk_set_bytes: estimate_pk_set_bytes(pk_bytes, rows),
+    })
+}
+
+/// Refuse tables whose features would make a blind file rewrite unsafe.
+fn ensure_supported_table(table: &DeltaTable) -> Result<()> {
```
estefaniarabadan, line 788: (sent from claude-code): `ensure_supported_table` refuses deletion vectors and column mapping but not `delta.enableChangeDataFeed` (blind rewrite emits no `_change_data`, corrupting the feed), `delta.appendOnly` (upsert tombstones files), or CHECK constraints/invariants (`RecordBatchWriter` doesn't enforce them) — refuse all three.

estefaniarabadan:
(sent from claude-code): the pyo3 0.22→0.28 migration of `rust/hogql/parser` (stacked base #73619) ships at 1.3.107 while production still pins 1.3.106, so any behavior change from the migration surfaces later, decoupled from this review — confirm the publish+pin is scheduled and the parser's parity gate covers the migrated paths.

estefaniarabadan (approved):

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,1946 @@
...
+        // upserts) would multiply resident memory. Reserve the estimate first, decode,
+        // then top the permit up if the batch came out larger. The permits ride with
+        // the data through filter and the writer queue, and are released by the
+        // writer only after write().
+        let mut permit = budgets.acquire_bytes(estimate_bytes).await?;
+        let Some(batch) = stream.try_next().await? else {
+            break;
+        };
+        let actual = batch.get_array_memory_size();
+        budgets.top_up(&mut permit, actual).await?;
```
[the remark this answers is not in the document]
Gilbert09: Fixed in 8bc271ec. \`top_up\` now **releases the estimate's permits before re-acquiring the grown amount** instead of acquiring the extra while still holding them. A reader therefore never waits on the byte budget while holding any of it, so the hold-and-wait cycle you described can't form — concurrent readers whose decoded batches exceed their estimate now serialize through the budget instead of wedging. Both budgets are released before either is re-acquired, so holding one semaphore while blocking on the other can't deadlock either. Covered by a new regression test (`concurrent_top_ups_beyond_estimate_do_not_wedge`): eight readers each reserve a small estimate then top up to a batch far larger than the whole budget, and all complete within a timeout.

rust/deltalite/core/src/upsert.rs
Gilbert09: Fixed in 8bc271ec. \`top_up\` now **releases the estimate's permits before re-acquiring the grown amount** instead of acquiring the extra while still holding them. A reader therefore never waits on the byte budget while holding any of it, so the hold-and-wait cycle you described can't form — concurrent readers whose decoded batches exceed their estimate now serialize through the budget instead of wedging. Both budgets are released before either is re-acquired, so holding one semaphore while blocking on the other can't deadlock either. Covered by a new regression test (`concurrent_top_ups_beyond_estimate_do_not_wedge`): eight readers each reserve a small estimate then top up to a batch far larger than the whole budget, and all complete within a timeout.

rust/deltalite/core/src/schema.rs
Gilbert09: Fixed in 8bc271ec. `realign_array_data` now threads a `changed` flag up from the recursion (via an inner `realign_inner` returning `(ArrayData, bool)`) instead of comparing the child's first-buffer pointer — which never moves for a Struct (no buffers) or a List/Map intermediate (its first buffer is the unchanged offsets). A repaired decimal nested two or more levels deep is now rebuilt into its parents rather than discarded. Covered by a new `struct<struct<decimal>>` regression test (`realigns_decimal_nested_two_levels_deep`) that fails the #3884 alignment path before the fix.

.github/workflows/ci-deltalite-python.yml
Gilbert09: Fixed in 8bc271ec. `pyproject.toml` and `uv.lock` are now in both the `pull_request` and `push` path filters, and the install step reads the `deltalake`/`pyarrow`/`duckdb` pins straight out of `pyproject.toml` (sparse-checked-out for this job) rather than hardcoding them. So a pin bump both re-runs this suite and is the single source the suite installs — the pins can no longer drift from what production installs.

rust/deltalite/core/src/upsert.rs
Gilbert09: Fixed in 8bc271ec. `ensure_supported_table` now also refuses `delta.enableChangeDataFeed=true` (a blind file rewrite emits no `_change_data` and would corrupt the feed), `delta.appendOnly=true` (upsert rewrites and tombstones files), and any `delta.constraints.<name>` CHECK constraint (the raw `RecordBatchWriter` does not enforce constraints/invariants, so a rewrite could persist violating rows). Each returns `Error::Unsupported` with a specific message.

rust/deltalite/core/src/upsert.rs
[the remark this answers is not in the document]
Gilbert09: The up-front reservation this raised was addressed in 5243e0e (readers reserve budget before `try_next()` decodes; the probe loop too). The follow-up risk that topping the reservation up could itself block while holding budget is now also fixed in 8bc271ec — `top_up` releases before re-acquiring, so it can't wedge. Resolving.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,2030 @@
...
+        let want = Self::kb(total_bytes);
+        let want_local = want.min(self.local_cap_kb as u64) as u32;
+        let want_global = want.min(self.limits.buffer_cap_kb() as u64) as u32;
+        if want_local <= permit.held_local_kb && want_global <= permit.held_global_kb {
+            return Ok(());
+        }
+        // Release both budgets before re-acquiring either, so no growth ever waits while
+        // holding budget (holding one semaphore while waiting on the other would deadlock
+        // just as surely as holding an estimate while waiting to grow it).
+        permit.local = None;
```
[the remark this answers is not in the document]
Gilbert09: Addressed in 26af51b. Two-layer defense now:

1. **The gap is closed at the source.** Both readers size the Parquet read batch from the widest row group's uncompressed bytes/row (`total_byte_size / num_rows` from row-group metadata) rather than a fixed row count, so a decoded batch stays near its pre-decode reservation regardless of row width. The wide/compressible-target case you describe now yields *smaller* batches (down to one row), not larger ones — a batch can no longer balloon past the reservation and sit unaccounted.

2. **The residual is bounded and deadlock-free.** When Arrow in-memory expansion still exceeds the reservation, `top_up` releases-and-reacquires (never holding budget while waiting), so it can't wedge. The unaccounted window is now one near-target batch per reader, and reader count is bounded by the process file limit — not unbounded exhaustion.

There is no pre-decode upper bound on `get_array_memory_size` that is both tight and safe (Arrow size can exceed uncompressed Parquet size), so eliminating the last byte of the window would mean serializing decoders and losing the streaming concurrency this path is built for. The sizing + release-and-reacquire combination keeps memory bounded without that cost. `probe_file` gets the same sizing. Covered by `batch_rows_bounds_decoded_size_by_bytes_per_row` and `concurrent_top_ups_beyond_estimate_do_not_wedge`.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,2030 @@
...
+            "table has delta.appendOnly=true; deltalite rewrites and removes files, which an \
+             append-only table forbids"
+                .into(),
+        ));
+    }
+
+    // CHECK constraints (`delta.constraints.<name>`) and column invariants are enforced by
+    // delta-rs's merge/write path, not by the raw `RecordBatchWriter` deltalite uses, so a
+    // rewrite could persist rows violating them. Refuse rather than write unchecked data.
+    if let Some(name) = config
```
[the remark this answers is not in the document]
Gilbert09: Partly addressed in 6be32cd, with a deliberate scope note.

`ensure_supported_table` now also refuses the writer features `invariants` and `checkConstraints` (writer protocol v7) — the current representation of column invariants and CHECK constraints — alongside the `delta.constraints.*` config guard for the legacy CHECK form.

The one residual is the *legacy* column-invariant form (writer protocol v2), stored per-field as `delta.invariants` in the schema field metadata rather than in table configuration or the feature list. Detecting it means reaching into the delta-kernel typed schema's per-field metadata, and there is no other reliable signal (writer-version alone can't distinguish it — v2 covers nearly every non-table-features table). deltalite writes to tables it manages (warehouse sync targets created by delta-rs/deltalite), which do not carry Spark-era legacy invariants, so I've scoped detection to the writer-feature + config representations rather than couple the guard to the kernel schema-metadata API. Happy to add the field-metadata scan as a follow-up if a real table surfaces it.

rust/Cargo.lock
```diff
@@ -11572,6 +12421,17 @@ dependencies = [
  "cfg-if",
 ]
 
+[[package]]
+name = "thrift"
+version = "0.17.0"
```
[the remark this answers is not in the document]
Gilbert09: This is a pre-existing transitive dependency of `parquet` (58.4.0 → `thrift` 0.17.0), not something this PR introduces — the `Cargo.lock` entry comes from the base branch, and `parquet` is used across the Rust workspace. Bumping `thrift`/`parquet` is a workspace-wide change that belongs in its own PR (it touches every crate that reads Parquet), not here.

Worth noting the crafted-metadata excessive-allocation vector is also bounded on the deltalite side by exactly the memory controls this PR hardens: the per-call and process-global byte budgets, and the new content-based batch sizing, cap resident decode memory regardless of what the file claims. I'll leave the dependency bump to a dedicated dependency-hygiene change so it can be reviewed and CI-run across the whole workspace.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,2108 @@
...
+                "source batch has {} rows, more than a single batch supports",
+                batch.num_rows()
+            )));
+        }
+        let idx = batch.schema().index_of(col).map_err(|_| {
+            Error::SchemaMismatch(format!(
+                "partition column '{col}' is missing from the incoming batch"
+            ))
+        })?;
+        let as_str = arrow_cast::cast(batch.column(idx), &DataType::Utf8)?;
```
danielcarletti, line 936: [should-fix] Partitions are identified by arrow_cast::cast(col, Utf8) here, and that same string is used at ~L1026 as PartitionValue::Equal(value) to list the partition's existing files. For any partition type whose arrow Utf8 cast diverges from delta-rs's stored partition-value string (timestamps, floats, some boolean encodings), the file listing matches nothing, the partition is treated as empty, and every source row is re-inserted as a silent duplicate. Not reachable today (production _ph_partition_key is always String; tests only use a string key), but the code accepts arbitrary partition types with no guard, so it's a latent silent-corruption trap. Fix: reject a non-Utf8 partition column in ensure_supported_table/planning, or add differential-parity coverage for int/date/timestamp partition columns before a non-string key is ever used.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,2108 @@
...
+            .iter()
+            .filter_map(|a| match a {
+                Action::Add(add) => add.partition_values.get(c).cloned().flatten(),
+                _ => None,
+            })
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .map(|v| format!("'{}'", v.replace('\'', "''")))
+            .collect();
+        format!("{c} IN ({})", vals.join(", "))
```
danielcarletti, line 711: [nit] For an empty source against a partitioned table, actions is empty so this predicate becomes the malformed string "<col> IN ()" (empty IN list) recorded in commitInfo. Informational only (never executed), but reads as broken. Fix: emit None for the predicate when there are no partition values to list.

rust/deltalite/core/src/upsert.rs
```diff
@@ -0,0 +1,2108 @@
...
+                adds.extend(writer.flush().await?);
+            }
+        }
+    }
+    adds.extend(writer.flush().await?);
+
+    // Every source row is written, but a source row that replaced an existing one is an
+    // update, not an insert -- the batch is deduped, so each match consumes exactly one
+    // source row. Counting the whole slice here double-counted matches as both.
+    let rows_inserted = source.rows.saturating_sub(rows_updated);
```
danielcarletti, line 1693: [nit] rows_inserted = source.rows - rows_updated assumes each match consumes exactly one source row, which holds for a deduped source against a PK-unique target. If the target already holds duplicate PKs within a partition (e.g. an external writer), one source row can match several target rows, so rows_updated exceeds the matched source-row count and rows_inserted saturates to 0 — a stats miscount only; the data outcome (deduping the target) is fine. Fix: cap the rows_updated used for this subtraction at the number of matched source rows, or document the assumption on the field.

danielcarletti (commented):
Reviewed against DC's checklist (design, blast radius, correctness, tests). No blockers — very high quality, with a differential parity harness against real delta-rs doing the heavy lifting on semantic equivalence. 3 findings inline: one latent silent-duplication trap worth guarding before this is wired into DeltaTableHelper, plus two minor stats/commitInfo nits. Note: the bundled pyo3 0.22→0.28 parser migration is the stacked base PR #73619's content and is the only change touching existing product code (all HogQL parsing) — make sure it's reviewed there.

PR #73597, opened 2026-07-24
feat(data-warehouse): link DAGs to their filtered data ops graph

sakce:
## Problem

From the Models scene's DAGs tab, there was no quick way to jump into the data ops Modeling graph filtered to a specific DAG - you'd have to open data ops and manually select the DAG from the toolbar.

## Changes

- Each DAG name in the DAGs tab is now a link to `/data-ops?tab=modeling&dag=<id>`, shown only when the Modeling tab is actually reachable (gated on the same `data-modeling-tab` feature flag that gates the tab itself).
- `urls.dataOps()` now takes an optional `dagId` second argument, added as a `?dag=` query param.
- `dataModelingLogic` reads `?dag=` on load and pre-selects that DAG, without touching the persisted selection when the param is absent.

No automated tests added - this is a small, additive UI wiring change (a link + a URL param) with no new business logic to regression-test; existing `dataWarehouseSceneLogic` tests still pass unaffected since `urls.dataOps(tab)` stays backward compatible.

I wasn't able to manually click through the flow in a running app in this environment.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with Claude Code, directed by Jovan Sakovic in a chat session: he asked for a link from the Models DAGs list to the corresponding data-ops scene, guarded the same way the data-ops scene itself is guarded, then asked to make it filter to the specific DAG clicked via a URL param.

Decisions: the DAG link is gated on `FEATURE_FLAGS.DATA_MODELING_TAB` (the flag for the destination Modeling tab) rather than the broader `DATA_WAREHOUSE_SCENE` flag, since that's what actually determines whether the destination is reachable. An initial "Data ops" header button (gated on `DATA_WAREHOUSE_SCENE`) was added then removed at Jovan's request, since only the per-DAG link was wanted.

### DIFF
--- a/frontend/src/scenes/data-warehouse/scene/dataModelingLogic.ts
+++ b/frontend/src/scenes/data-warehouse/scene/dataModelingLogic.ts
@@ -11,9 +11,11 @@ import {
 import { deepEqual as equal } from 'fast-equals'
 import { MakeLogicType, actions, afterMount, beforeUnmount, kea, listeners, path, reducers, selectors } from 'kea'
 import { loaders } from 'kea-loaders'
+import { urlToAction } from 'kea-router'
 import type { RefObject } from 'react'
 
 import api from 'lib/api'
+import { urls } from 'scenes/urls'
 
 import {
     DataModelingDAG,
@@ -1084,6 +1086,13 @@ export const dataModelingLogic = kea<dataModelingLogicType>([
             actions.setEdges([])
             actions.loadDataModelingNodes()
             actions.loadDataModelingEdges()
+        },
+    })),
+    urlToAction(({ actions, values }) => ({
+        [urls.dataOps()]: (_, searchParams) => {
+            if (typeof searchParams.dag === 'string' && searchParams.dag !== values.selectedDagId) {
+                actions.setSelectedDagId(searchParams.dag)
+            }
         },
     })),
     afterMount(({ actions }) => {
--- a/frontend/src/scenes/models/DagsTab.tsx
+++ b/frontend/src/scenes/models/DagsTab.tsx
@@ -29,7 +29,8 @@ export function DagsTab(): JSX.Element {
     const { dags, dagsLoading } = useValues(dagsLogic)
     const { loadDags, updateDag, deleteDag } = useActions(dagsLogic)
     const { featureFlags } = useValues(featureFlagLogic)
-    const dataOpsModelingEnabled = !!featureFlags[FEATURE_FLAGS.DATA_MODELING_TAB]
+    const dataOpsModelingEnabled =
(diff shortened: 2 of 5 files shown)


thiagosalvatore (approved):
@sakce approved, but qq: I thought that with your changes the DAG would be invisible for users?

sakce:
@thiagosalvatore this is purely for me while i work on things :joy_cat:

frontend/src/scenes/models/DagsTab.tsx
```diff
@@ -24,6 +28,8 @@ const DEFAULT_DAG_NAME = 'Default'
 export function DagsTab(): JSX.Element {
     const { dags, dagsLoading } = useValues(dagsLogic)
     const { loadDags, updateDag, deleteDag } = useActions(dagsLogic)
+    const { featureFlags } = useValues(featureFlagLogic)
+    const dataOpsModelingEnabled = !!featureFlags[FEATURE_FLAGS.DATA_MODELING_TAB]
```
thiagosalvatore: This only checks `DATA_MODELING_TAB`, but the destination still returns `NotFound` unless `DATA_WAREHOUSE_SCENE` is also enabled (`DataWarehouseScene.tsx:49`). Since flags can be overridden independently, this can expose links that always dead-end. Please gate the link on both flags, or make the Modeling destination genuinely independent of the scene flag.

frontend/src/scenes/data-warehouse/scene/dataModelingLogic.ts
```diff
@@ -1086,6 +1088,13 @@ export const dataModelingLogic = kea<dataModelingLogicType>([
             actions.loadDataModelingEdges()
         },
     })),
+    urlToAction(({ actions, values }) => ({
```
thiagosalvatore, line 1091: Could we add coverage for this URL-driven selection? This mutates persisted state and triggers both filtered loaders, but the existing `dataWarehouseSceneLogic` tests never mount this logic. At minimum, please cover `?dag=<id>` overriding the persisted selection and filtering both requests, plus an absent `dag` preserving the persisted value; a stale ID fallback case would also be valuable.

thiagosalvatore (changes requested):
The generated URL issue is fixed in the current head, but there are still two blocking gaps: the link's feature-flag gate does not match the destination's gate, and the new persisted URL-selection behavior is untested.

sakce:
@thiagosalvatore fair comments on both. unsure if this was an agent reviewing, but would appreciate a `requested changes` PR review to cases that should be really blocking. this is a low stakes change and having to do another round trip is worse than trusting that I will address the comments in the best way possible for the context the PR is in

thiagosalvatore (approved):

PR #73602, opened 2026-07-24
fix(conversations): surface email connect errors instead of a generic toast

HaynesPostHog:
## Problem

Connecting a support sending address showed "Failed to connect email" whatever went wrong. Every 400 from the connect endpoint carries a specific reason, and the listener discarded it. A blocked domain, a typo, and an unconfigured instance all looked identical.

The report behind this blamed the address's top-level domain. Nothing rejects a top-level domain. The frontend regex, Django's `EmailField`, the sending provider's domain regex, and the free and disposable domain blocklists all accept one. The real reason sat in the response the UI threw away.

## Changes

- The connect form now shows the reason the backend sent, and keeps "Failed to connect email. Please try again." for responses that carry none.
- `supportSettingsLogic` reads the view's `{error}` field and DRF's `{detail}` field, so validation failures and the view's own rejections both reach the toast.
- Error strings on connect, verify-domain and send-test drop the environment variable name and internal phrasing they exposed. The specifics stay in the structured logs.
- Verify-domain and send-test still discard their errors in the UI, so their reworded strings reach only the logs for now.
- A test fixture moves to a reserved domain. That bullet is the only mechanical one.

> [!NOTE]
> The shared-domain guard added in #93462 tells a user to send from a domain they own. This PR is what makes that message reach them.

No screenshot: the visible change is the text inside an existing error toast, and it was verified through the logic test rather than in a browser.

- `supportSettingsLogic.test.ts` asserts both response shapes reach `lemonToast.error`. No test covered the discarded-reason regression before.
- `test_email_endpoints.py` and `supportSettingsLogic.test.ts` both pass locally, with master merged in.
- Not checked: a real connect failure in a running browser, and the verify-domain and send-test strings, which no test asserts.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None. No documented workflow or API changes.

## 🤖 Agent context

The first session traced the connect path end to end and ruled out domain validation as the cause, which is why this PR surfaces the existing error instead of changing a validator.

A later session brought the branch up to date after it sat behind master. Master moved the connect flow's provider registration inside a `kind == SUPPORT` branch, so the reworded strings were re-applied at the new nesting. That session also replaced a test fixture that carried a real company's domain and display name. The committed fixture uses a reserved domain, and the assertion never depended on the value.

Skills invoked: `/writing-code-comments`, `/writing-pr-descriptions`.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08S2L0S5MZ/p1784916538414989?thread_ts=1784916538.414989&cid=C08S2L0S5MZ)*


phillram (approved):
LGTM

veryayskiy (approved):
🚢

PR #73618, opened 2026-07-24
feat(llm-gateway): add claude opus 5

charlesvien:
## Problem

Claude Opus 5 is out (https://www.anthropic.com/news/claude-opus-5) and the gateway doesn't serve it yet.

## Changes

Same as the Sonnet 5 rollout (#67138):

- gateway model allowlists + Bedrock inference profile mapping for `claude-opus-5`
- tasks catalog + task composer picker entries

Pricing comes from LiteLLM's cost map, which already has `claude-opus-5` ($5/$25 per MTok), so no override needed.

Extended the existing parameterized gateway tests (bedrock mapping, product allowlists) with `claude-opus-5` cases. Full llm-gateway suite passes locally (only 2 pre-existing stream-test failures that also fail on master).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A

## 🤖 Agent context

Written with Claude Code, mirroring #67138. Skills invoked: /writing-tests, /claude-api (model id + pricing reference). An earlier revision carried a `MODEL_COST_OVERRIDES` bridge entry for pricing; dropped after LiteLLM upstream landed `claude-opus-5` with matching numbers.


tatoalo (approved):

cvolzer3 (changes requested):
Can you update `BEDROCK_RUNTIME_COUNT_TOKENS_UNSUPPORTED` in `/Users/chrisvolzer/projects/posthog/services/llm-gateway/src/llm_gateway/bedrock.py` to include Opus 5?

Bedrock runtime doesnt support Opus 5, but Bedrock mantle does so adding it to that list will trigger the fallback

Context: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-5.html

cvolzer3 (approved):

PR #73625, opened 2026-07-24
feat: enforce warehouse-source access control when querying

a-lider:
> Stacked on #72440 (per-table access control)

## Problem

**Querying** a source's tables still ignored the source's access rules. Enforcement only works for the warehouse tables and views resource. So a member restricted from a source (for example a Zendesk) could still `SELECT * from zendesk.tickets`. 

## Changes

Query enforcement for warehouse sources, most-specific rule first. A rule on the table itself wins outright: a table granted inside a denied source stays queryable, a table denied inside an allowed source doesn't. Without table rules the more restrictive of the table's and the source's access applies, each resolved from its own object and resource-level rules — so denying all tables covers an allowed source, and denying one source covers tables allowed at the resource level.

- `_is_warehouse_table_denied` (the gate every HogQL build funnels through) resolves the level through `warehouse_table_effective_level` and denies below viewer. Userless builds stay fail-closed, org admins bypass.
- The query cache partitions on source access — warehouse reads declare `external_data_source` as a queried resource — so a restricted user is not served a cached result someone with access populated.

> [!NOTE]
> Views aren't part of the cascade: `DataWarehouseSavedQuery` has no `external_data_source` FK (it's authored SQL that can reference several sources or none), so there's no parent source to inherit from. Non-materialized views re-resolve to their underlying tables at execution, so a denied table still stops them; a materialized view is its own snapshot governed by its `warehouse_view` rule. The `warehouse_objects` resource still covers tables and views alike.

`hogli test posthog/hogql/printer/test/test_hogql_access_control.py` — `TestWarehouseSourceAccessControl` covers the precedence matrix: source denied covers its tables (and lands in the cache key) but not other sources', source viewer keeps them queryable, a table grant overrides a denied source, a table deny overrides an allowed source, source allowed + all-tables denied at resource level, all-tables allowed + source denied, resource-level source deny, and org-admin bypass. 

`hogli test posthog/hogql_queries/test/test_access_controlled_resources.py` — updated for `external_data_source` now being in scope wherever warehouse tables are read.

Plus the 6 schema-AC tests from #72440 (they share the helper)

Also manually in the SQL editor against a local source restricted for a non-admin member: querying an unruled table is denied, a table with its own grant still works, admin bypasses.

## Automatic notifications

- [ ] Publish to changelog? — No, gated behind the `hogql-warehouse-access-control` flag
- [ ] Alert Sales and Marketing teams? — No

## 🤖 Agent context

Main design decision (Alex pushed on this): modeling the source→table cascade. Rejected extending `RESOURCE_INHERITANCE_MAP` — it's resource-*type* inheritance and can't express object-*instance* parent links (this table's source is that source row), and there's no precedent for object-A deriving access from object-B's rules. A generic object-inheritance layer would touch every resolver in `UserAccessControl` (object resolution, queryset filtering, the blocked-ids set feeding cache keys) — too much blast radius on security-critical code for one FK pair. Instead the precedence is expressed once via `warehouse_table_effective_level` + `has_object_rules` and called at the three enforcement sites. If a second parent→child object pair shows up, that helper generalizes into an object-parent map with an FK resolver as its own refactor.

A first pass had this as min(source, table) and then as "only per-source object rules cascade" — both wrong, corrected to the precedence above after Alex worked through the cross-level cases.

### DIFF
--- a/posthog/hogql/database/database.py
+++ b/posthog/hogql/database/database.py
@@ -791,12 +791,25 @@ class Database(BaseModel):
         """
         Returns True if the user can't query this warehouse table.
         Userless context (no UserAccessControl) fails closed - every table is denied.
+        Table-level rules take precedence; a table without its own rules also requires viewer
+        access to its parent source, so restricting a source covers all its tables.
         """
         uac = self.user_access_control
-        if uac is not None and (
-            uac.is_organization_admin or uac.check_access_level_for_object(table, required_level="viewer")
-        ):
-            return False
+        if uac is not None:
+            if uac.is_organization_admin:
+                return False
+            table_allowed = uac.check_access_level_for_object(table, required_level="viewer")
+            source = table.external_data_source
+            # Only object rules on the source cascade to querying its tables; resource-level
+            # external_data_source rules keep governing the source APIs only.
+            source_allowed = (
+                source is None
+                or uac.has_object_rules("warehouse_table", str(table.pk))
+                or not uac.has_object_rules("external_data_source", str(source.pk))
+                or uac.check_access_level_for_object(source, required_level="viewer")
+            )
+            if table_allowed and source_allowed:
+                return False
 
         # Add table names to denied tables so the query raises "You don't have access" instead of "Unknown table"
         self._denied_tables.add(table.name)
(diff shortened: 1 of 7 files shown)


ReeceJones (approved):

a-lider:
Decided to make this more general and introduce a new pattern in access control: https://github.com/PostHog/posthog/pull/74201

PR #73634, opened 2026-07-24
chore(cohorts): add oracle modes to parity harness

matheus-vb:
## Problem

Verifying a behavioral cohort backfill canary meant hand rolling shell scripts on a toolbox pod each time: fold the shadow topic, recompute expected membership in ClickHouse, three way diff, then sort the missing persons by which backfill day owned them. Every future canary needs that same comparison, so it should be a command.

## Changes

Adds `--oracle recompute` to `compare_cohort_membership`. It recomputes membership from `events` with evaluator semantics (inclusive N+1 day window, `count >= 1 AND op(count)` floor, operator mapping, tree composition) and diffs that against the folded shadow topic. Over count is the hard gate. Under count is segmented by backfill day domain so the expected boundary day gap is separated from real seed or live path misses, with a decay curve predicting when it clears.

New modules under `products/cohorts/backend/parity/`: `tzdates.py` (team tz calendar math), `recompute.py` (pure screen, tree evaluation, classification), `oracle.py` (ClickHouse reads plus the Django run context). `--oracle old-pipeline` stays the default and its output is unchanged.

Leaf identity is the full Stage 1 `LeafStateKey` tuple rather than `conditionHash`, which digests only the event matcher. That tuple now lives in `models/leaf_shape.py` so the shape hash and the oracle share one field list.

> [!NOTE]
> The oracle only reproduces `performed_event` and `performed_event_multiple` leaves with a string event key, no property filters, and whole day windows. Anything else reports SKIP with a specific reason. The report never claims PASS for a diff it could not adjudicate.

Tested both oracle modes end to end against the local dev stack on a realtime cohort, where the recompute oracle exactly reproduced the folded membership. All three ClickHouse queries were executed against live ClickHouse across every operator, checking that the SQL predicate agrees with the Python comparator, that day bucketing lands inside the computed window, and that per person day sums match the member set counts. I also verified the persisted `behavioral_filters_shape_hash` is byte identical to master across 531k field combinations, since `leaf_shape.py` feeds it.

## Docs update

No.


products/cohorts/backend/parity/oracle.py
```diff
@@ -0,0 +1,308 @@
...
+  AND e.timestamp <= %(at)s
+  AND coalesce(e.inserted_at, e._timestamp) <= %(at)s
+  AND {resolved_person} IN %(person_ids)s
+GROUP BY resolved_person_id
+"""
+
+_SEGMENTATION_SQL = """
+SELECT {resolved_person} AS resolved_person_id,
+       toDate(e.timestamp, %(team_tz)s) AS day,
+       multiIf(e.timestamp >= %(grace_start)s, 'grace',
```
gustavohstrassburger, line 91: `suggestion`: `grace` wins outright here with no check that the grace window sits entirely after the boundary. `_validate_recompute_flags` only bounds `--grace-minutes` against a fixed `MAX_GRACE_MINUTES = 24 * 60` (compare_cohort_membership.py:453), not against `ctx.boundary_at`, so a boundary established a few hours before `--at` is already inside that window.

When `grace_start` reaches past the boundary, real seed-day matches get stamped `grace` instead of `seed`. `_classify_missing_person`'s first branch (recompute.py:491-493) returns `missing_grace` whenever `seed + boundary + unseeded + post < min_count`, and `missing_grace` is excluded from the FAIL gate (recompute.py:615). So a cohort whose entire missing set is a real seeder drop can report PASS.

Worth rejecting the overlap explicitly at the call site (compare_cohort_membership.py:678) rather than approximating it with a constant, something like:

```python
grace_start = at - grace
if grace_start < ctx.boundary_at:
    rows.append(skip_comparison(cohort_id=cid, name=name, reason="grace_window_predates_boundary", reconcile_runs=reconcile_runs))
    continue
```

products/cohorts/backend/parity/recompute.py
```diff
@@ -0,0 +1,696 @@
...
+    *,
+    window: frozenset[date],
+    ctx: RunContext,
+    min_count: int,
+) -> str:
+    """Assign a missing (oracle - fold) person to a day-domain class. Precedence grace ≻ seed ≻
+    boundary ≻ unseeded ≻ post: the highest-precedence domain whose cumulative count crosses the
+    membership threshold. The classes are exhaustive — the final fall-through implies ``post > 0``,
+    since the three preceding sums were all below the threshold the total clears."""
+    grace, seed, boundary, unseeded, post = _domain_counts(matches, window=window, ctx=ctx)
```
gustavohstrassburger, line 494: `suggestion`: when `matches` is empty (the day-counts read returned nothing for this person), all five domain counts are 0, so `seed + boundary + unseeded + post < min_count` is always true and this unconditionally returns `missing_grace` — which gates PASS, not FAIL.

That's the shape of a real failure, not just an edge case: `load_leaf_members` and `load_day_counts` are separate ClickHouse reads with different person-id chunking. A person present in one and absent from the other means the two reads disagree (a merge/override drift, a dropped chunk), and today that silently reads as lag noise.

`TestMissingSegmentation.test_missing_person_precedence` covers all five classes but always supplies at least one `DayMatch` — nothing exercises the empty case. Worth deciding explicitly whether an empty match list should stay `missing_grace` (and pin it with a comment + test) or become its own class that gates FAIL.

products/cohorts/backend/parity/recompute.py
```diff
@@ -0,0 +1,696 @@
...
+    #   (b) they entered within the grace window — the processor enters on the event and only leaves
+    #       on a sweep tick, so a just-ingested historical event legitimately shows as a member
+    #       (`event_path.rs` sets has_match on any match; only the sweep clears it).
+    entered_after = at - grace
+    eviction_pending: set[str] = set()
+    for person in false_set:
+        if fold_records[person].last_updated >= entered_after:
+            eviction_pending.add(person)
+            continue
+        if not extended_leaf_counts:
```
gustavohstrassburger, line 614: `suggestion`: worth naming the asymmetry here explicitly. When the over-count cap blows, every false member gets scored hard and the cohort FAILs on evidence the command chose not to gather (`extended_leaf_counts` stays empty here). The under-count side handles the mirror case oppositely: a blown cap sets `day_counts_loaded = False` and routes to SKIP instead of FAIL (compare_cohort_membership.py:668-669).

If that's intentional (over-count fails closed because over-inclusion is the hard gate; under-count skips because it can't be attributed), a line in `RECOMPUTE_CAVEATS` would save the next person from reading a cap-driven FAIL as a real over-inclusion bug.

gustavohstrassburger (approved):
Nice work, careful DST/window handling throughout. A couple of suggestions.

PR #73640, opened 2026-07-24
feat(growth): unified score lab page and result input snapshots

MattBro:
## Problem

Iterating a classifier prompt in the score lab (base PR https://github.com/PostHog/posthog/pull/72710) takes five admin surfaces: changelist, action dropdown, options form, results page, and a separate add form to persist a version. Comparing an edited prompt against archived data means committing a version first or dropping to the CLI. The tools this workflow competes with do the whole loop in one screen.

This PR is the bridge surface for the label owner while the score lab is admin-hosted; a staff-only in-product scene is the planned successor, at which point these admin views are deleted.

## Changes

**Unified Lab page** at `/admin/growth/enrichmentpromptconfig/lab/<label>/`:

- Version rail: every version of the label, active badge, click to load into the editor
- Editor: prompt textarea, model dropdown, input-field checkboxes, prefilled from the selected version, with a dirty-state line that flips to "editing (unsaved experiment)"
- Run: streams verdicts for the editor's current (possibly unsaved) state against recent archived orgs - fetch + ReadableStream appending rows as each LLM call completes. Persists nothing
- Save as new version: persists exactly the validated editor bytes as a new immutable version, with an auto-suggested version bump. Duplicate versions return a form error
- The streaming core is shared with the existing changelist dry-run action (refactored into one async-generator helper, built via `posthog.api.streaming.streaming_response`)

**Input snapshots on results** (migration 0007): `EnrichmentLabelResult.inputs` stores the rendered classifier inputs (extracted payload fields + signup domain) at compute time. The signup domain derives from the earliest current member and drifts as members leave; without the snapshot, results are not reproducible even under a fixed model. The unknown short-circuit snapshots the domain with empty fields.

**CSP nonces on inline template scripts**: the admin's Content-Security-Policy (script-src with per-request nonce) silently blocks bare inline `<script>` tags - discovered because the lab's Run button did nothing in a real browser while all tests passed. Both lab and options-page scripts now carry `nonce="{{ request.csp_nonce }}"`.

Automated: 32 tests (`test_enrichment_lab_ui.py` new, `test_enrichment_labels.py` extended) - lab page renders and prefills from the active version, run streams verdicts for an edited prompt and persists nothing (no results, no configs), save persists exact submitted bytes and rejects duplicates without a 500, changelist links the lab, batch stores the inputs snapshot and keeps it out of `output`, unknown verdicts snapshot the domain.

Live browser verification (local dev + local LLM gateway, via Playwright): edited the prompt in place, dirty-state line flipped, Run streamed 10 real verdicts row-by-row into the page, nothing persisted; the CSP issue was found and fixed through this pass. Migration safety checks (`analyze_migration_risk`, `makemigrations --check`) clean.

## Next steps

- Staff-only in-product scene (products/growth manifest + IsStaffUser DRF viewset, following the cohorts staff-tools pattern) replaces this admin surface; these views are deleted in that PR series
- Warehouse dashboards for stored results (config, no code)

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

**Autonomy:** Human-directed (agent-built)

Built by two parallel Claude sub-agents (lab UI vs data snapshotting, disjoint file ownership) in a stacked worktree, then integration-verified and live-browser-tested by the coordinating session. The CSP finding is the notable catch: Django test client cannot execute JS, so template-script regressions are invisible to the suite - the live pass exists for exactly that class of bug. An adversarial review agent ran over the final diff before this PR opened.


MattBro:
Superseded: going straight to the in-product staff scene instead of an interim admin lab. The surviving pieces (input snapshots, CSP fixes, review hardening) carry over into the replacement PR.

PR #73644, opened 2026-07-24
chore(devenv): single-source mprocs shell generation

haacked:
## Problem

The docker-compose proc's shell was hand-duplicated across three places (`bin/mprocs.yaml`, `bin/mprocs-e2e.yaml`, and the hogli-generated shell used by `hogli dev:setup`), and they'd drifted into three different behaviors:

- `bin/mprocs.yaml` ran `docker compose -f docker-compose.dev.yml up` with no project name and no profiles, so it silently dropped every service gated behind a profile (temporal, etcd, localstack, observability, replay) even though other procs in that same file depend on them.
- `bin/mprocs-e2e.yaml` ran without a `-p` flag, so its compose project name came from directory inference instead of being pinned.
- The hogli-generated shell (used when `hogli dev:setup` resolves intents) had no `POSTHOG_SANDBOX` branch at all, so a sandboxed environment driven through the intent resolver would try to run `docker compose` and fail instead of short-circuiting.

## Changes

- Extract `_build_docker_compose_shell` in `generator.py` as the single builder for the up/logs/sandbox-guard shell, and have all three shells call it.
- That consolidation changes real behavior, each worth a specific look in review:
  - `bin/mprocs.yaml` now runs with `-p posthog -f docker-compose.profiles.yml --profile dev_tools --profile etcd --profile localstack --profile observability --profile replay --profile temporal`, matching the service set its own procs actually depend on instead of a bare `up`.
  - `bin/mprocs-e2e.yaml` now pins `-p posthog` on all three `docker compose` calls (up, personhog up, logs).
  - The hogli-generated shell gets the `POSTHOG_SANDBOX` short-circuit for the first time, so sandboxed `hogli dev:setup` environments skip `docker compose` instead of erroring.
- Add `hogli dev:regenerate-mprocs` (with a `--check` flag) to sync the checked-in shell lines in `bin/mprocs.yaml` and `bin/mprocs-e2e.yaml` from the generator functions.
- Wire a new ci_preflight "mprocs" check that runs `dev:regenerate-mprocs --check` whenever `generator.py` or either mprocs file changes, with `--fix` calling the plain regenerate command.
- Rename `TrackedMprocsFile.get_path` (a callable) to a plain `path` field, and consolidate `_find_repo_file`/`get_default_e2e_mprocs_path` into one `find_repo_file` helper.

Two things came up repeatedly across `/simplify` passes on this branch and I deliberately left them alone:

1. `_generate_docker_compose_config` reads the existing shell out of `bin/mprocs.yaml` via the registry, then discards it and rebuilds a fresh string from `_build_docker_compose_shell`. Every other special-case method on `MprocsGenerator` (`_add_startup_message`, `_add_nodejs_capability_groups`, `_add_personhog_env`, `_add_uv_groups`, `_add_sandbox_wrapper`) transforms the existing string in place instead. Switching to transform-in-place would remove most of this PR's new code (the regex reader/writer, the regenerate command, the sync tests), but it's a redesign, not a cleanup.
2. `build_e2e_docker_compose_shell()` lives in `generator.py`, a module whose docstring says it builds config "from resolved environment" via hogli's intent system, a path E2E never takes (`bin/mprocs-e2e.yaml` runs directly via `phrocs --config bin/mprocs-e2e.yaml` from `bin/e2e-test-runner`). Whether E2E's shell-building logic belongs nearer `bin/mprocs-e2e.yaml` instead is an open question for later.

Added coverage in `test_devenv.py`:
- Each checked-in file's docker-compose shell equals its generator function's output, and its `ready_pattern` matches `_READY_MARKER`.
- Regenerating a drifted copy of each file rewrites only the docker-compose shell line.
- `dev:regenerate-mprocs --check` passes without writing when up to date, and fails without writing when stale.
- The static and e2e shells only emit the ready marker after `docker compose up` (and, for e2e, the personhog step) succeeds, and never invoke docker at all when `POSTHOG_SANDBOX=1`.

Ran the devenv test suite locally.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs to update; this only touches internal dev tooling.


tools/hogli-commands/hogli_commands/ci_preflight.py
```diff
@@ -138,6 +139,17 @@ class DiffCheck:
...
+        key="mprocs",
+        label="bin/mprocs*.yaml docker-compose shell out of sync with generator.py",
+        # From TRACKED_MPROCS_FILES so preflight can't drift on which files need a regen.
+        triggers=[
+            "tools/hogli-commands/hogli_commands/devenv/generator.py",
+            *(t.name for t in TRACKED_MPROCS_FILES),
+        ],
+        verify=["hogli", "dev:regenerate-mprocs", "--check"],
+        fix=["hogli", "dev:regenerate-mprocs"],
+    ),
```
gantoine, line 152: 👍🏼

gantoine (approved):

PR #73651, opened 2026-07-24
feat(query): replace http coalescing with cache-key single flight

andyzzhao:
## Problem

When N identical blocking queries arrive at once (a dashboard fan-out, a shared link going around, API clients polling), all N run on ClickHouse. The HTTP-level query coalescer (PRs 49295, 51049, 52511) catches some of this, but it keys on the request surface (team, path, normalized body), so the same computation reaching us through different endpoints or differently-shaped bodies never collapses, and two users with the same body but different resolved modifiers can wrongly share. The correct identity for "same computation" is the query `cache_key`, which only exists inside the query runner.

Concurrent failures are the expensive half: before the circuit breaker (PR 72777) has enough evidence to open, N concurrent doomed requests each burn a full execution-time budget hitting the same error.

## Changes

Stacked on PR 72777. Adds single-flight execution per `cache_key` to `posthog/query_cache/` (`single_flight.py`, surfaced as `QueryCache.flight()`), gated behind a `query-single-flight` flag (off by default), and removes the HTTP-layer coalescer it replaces (`query_coalescer.py`, its middleware registration, the viewset mixins on `QueryViewSet` and `InsightViewSet`, the `QUERY_COALESCING_MAX_WAIT_SECONDS` setting, and its tests; generated OpenAPI artifacts regenerated since the mixin's docstring leaked into endpoint descriptions).

```mermaid
flowchart LR
    A{{"blocking execution"}} --> L["acquire flight?"]
    L -- "leader" --> C["run on ClickHouse"]
    C -- "success" --> S["store_result publishes via the cache"]
    L -- "follower" --> W["poll until release or timeout"]
    W -- "leader's fresh entry" --> S2["serve it"]
    W -- "nothing fresh" --> B["breaker recheck"]
    B -- "open" --> E["remembered error"]
    B -- "closed" --> R["run it yourself"]
    class A phYellow;
    class S,S2,W phBlue;
    class C,E phRed;
    class L,B,R phGray;
```

The contract is deliberately minimal: the cache layer either hands you a result or tells you to run the query, and coalescing changes nothing about that shape. Errors are never transported.

- Success needs no publication mechanism of its own: the leader's cache write is the publication, and followers serve the fresh entry (custom-name patching applied). Followers refuse stale entries; a stale entry means the leader failed, and old data must not mask that.
- Leader failures are not shared. A follower that finds nothing fresh rechecks the circuit breaker (the leader's deterministic failure opens it during the wait, on the first failure for memory-limit and query-size kinds) and otherwise runs the query itself, surfacing whatever happens honestly. Not transporting errors is what killed PR 49295's version of this: preserving arbitrary, runtime-defined exception classes across Redis is brittle, and this contract makes the problem vanish instead of solving it.
- The accepted cost: the first concurrent burst of transient failures (timeouts) runs more than once, until the breaker's 3-strike threshold fills. Deterministic failures are stopped after one execution by the breaker recheck.
- Scope is the blocking execution path only; async dispatch already dedups via the running-queries map. The single-flight check runs after the circuit-breaker gate.
- Every storage error and every ambiguous outcome (leader crash, lock expiry, uncacheable result) fails open to independent execution; the worst case is today's behavior.
- `posthog_query_single_flight_total{action}` tracks activity.

> [!WARNING]
> Rollout sequencing: HTTP coalescing (gated by the `http-query-coalescing` flag) is deleted by this PR, and single flight ships off by default behind `query-single-flight`. Between merge and flag ramp there is no request coalescing at all. If `http-query-coalescing` is currently serving real traffic, enable `query-single-flight` as part of shipping this.

Behavior deltas versus the HTTP coalescer, all deliberate: followers now pay normal request processing (parse, cache key computation) before waiting, in exchange for collapsing on the correct identity; cross-user response sharing is gone, since `cache_key` includes per-user resolved modifiers that the HTTP body key ignored; and concurrent failures are no longer replayed to followers at all. v1 polls (250ms interval, 65s cap) instead of pub/sub.

Automated only. `posthog/query_cache/test/test_single_flight.py` covers the primitives (leader exclusivity, wait outcomes, storage errors failing open). `TestQuerySingleFlightRunner` in `test_query_runner.py` drives the real `run()` flow: leaders release on success and failure, followers serve the leader's fresh entry, refuse stale entries, run the query themselves after a transient leader failure, hit the breaker recheck after a deterministic one, fall back on timeout, and flag-off never touches the flight. Full `test_query_runner.py`, `posthog/query_cache/test`, and `posthog/api/test/test_query.py` suites pass and repo-wide mypy is clean. The HTTP coalescer's tests are deleted with it.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Flag-gated off; docs will come with a broad rollout.

## 🤖 Agent context

Authored by Claude (Claude Code session) directed by the PR assignee, as the third PR in the query-cache stack (73425 consolidation, 72777 circuit breaker). The design went through two rounds: an initial version transported leader failures to followers through a short-TTL envelope replayed via a generic carrier exception, sidestepping the runtime-defined exception reconstruction problem that killed PR 49295's runner-level coalescing. The driver then simplified the contract to "the cache either gives you a result or tells you to run the query", which removes error transport entirely and delegates repeated-failure suppression to the circuit breaker, at the accepted cost of duplicate executions in the first burst of concurrent transient failures. The driver also called for removing the HTTP-layer coalescer in this PR rather than as a follow-up. Other v1 choices: polling over pub/sub, blocking-path-only scope, fail-open on every ambiguous outcome.


andyzzhao:
Superseded by a fresh stack: PR 100666 adds runner-level single flight behind a flag and keeps the HTTP coalescer, and a second layer on top removes the HTTP coalescer once the flag is at 100%.

PR #73662, opened 2026-07-25
fix(data-imports): treat empty envelope body as zero rows for required selectors

Gilbert09:
## Problem

A data warehouse sync that uses a required `data_selector` fails with a `ValueError` when the source API returns an empty envelope body (`{}` or `[]`) that omits the list key.

Surfaced by error tracking in the shared REST source pipeline:

> `ValueError: Required data_selector 'datasets' matched nothing in the response (body keys: []). The API response shape may have changed.`

[Error tracking issue](https://us.posthog.com/project/2/error_tracking/019f9721-9486-7913-8a45-fccbea623458)

The failing frame is `_extract_response` in `sources/common/rest_source/rest_client.py`, shared by ~90 REST sources that set `data_selector_required=True`. Plenty of REST APIs drop the envelope key entirely when a collection is empty and just return `{}`. The `data_selector_required` guard was meant to fail loud on a genuine shape change (a wrong or renamed key), but it also caught this empty-collection case and turned a normal "no rows" sync into a hard failure.

## Changes

In `_extract_response`, when a required selector matches nothing, first check whether the body is an empty container. An empty `{}` or `[]` carries no rows and no alternative shape, so it is treated as a legit zero-row page and returns `[]`. A body that carries different keys is still a real shape change and still fails loud with the original error.

```diff
 if required and not matches:
+    if isinstance(body, dict | list) and not body:
+        return []
     keys = sorted(body.keys())[:20] if isinstance(body, dict) else type(body).__name__
     raise ValueError(...)
```

This is a general fix at the shared layer rather than a per-source patch, since the empty-envelope shape is common across REST sources, not specific to the source that surfaced it.

Added two parameterized cases to `TestRequiredDataSelector` in `test_required_selector.py` covering an empty object and an empty array body with a required selector. Both previously raised `ValueError`; they now return `[]`. The existing cases still guard the real failure modes: an absent key on a non-empty body (`{"unexpected": ...}`) still fails loud, and a present-but-empty list (`{"data": []}`) is still a valid zero-row page.

Ran the suite locally:

```
uv run pytest .../rest_source/tests/test_required_selector.py -q
8 passed
```

## 🤖 Agent context

I (Claude) triaged an error-tracking issue in the data-warehouse import pipeline. The exception fired in shared REST source code, so I fixed it there rather than in the source connector that happened to surface it. I considered making the empty case retryable, but an empty `{}` is a stable, correct response, so retrying would just burn the budget before failing anyway. Returning an empty page is the right behavior. I scoped the carve-out to the selector branch, leaving the no-selector "body must be a list" contract untouched, and kept the loud failure for genuine shape changes (a body with unexpected keys).

---


Gilbert09:
### CI status: green on this PR's scope; blocked only by a pre-existing master breakage

The branch is up to date with `master` (clean two-parent merge, no file leaks — the PR diff is still only the 8 data-warehouse REST-source files), has no unresolved review threads, and is `MERGEABLE`. All of this change's own tests pass in CI (the `Django test matrix` reports `django | success`), and there are **no** failing tests anywhere under `products/warehouse_sources` / `rest_source` / `data_imports`.

The three non-green checks all trace back to a single failure that is unrelated to this PR:

- **`Product tests (replay-vision, workflows, events schema legacy)` — FAILURE**
  Two tests in `products/workflows/backend/api/test/test_hog_flow.py` fail:
  `TestHogFlowSecretInputs::test_publish_promotes_draft_secret_to_live_without_wiping` and
  `::test_restore_reattaches_live_secret_not_historical`.
  They PATCH `{"actions": ...}` with the `x-posthog-client: mcp` header, which the MCP-update guard now rejects:
  > `actions/edges can't be replaced via workflows-update - a partial list silently drops every step it omits.`
- **`Django Tests Pass` — FAILURE**: the aggregate gate, red solely because of the two workflows tests above.
- **`Validate migrations` — CANCELLED**: collateral — Backend CI cancels downstream jobs once a sibling job fails.

This is a collision between two independently-merged workflows PRs (the MCP-update guard in #73166 and the secret-inputs tests in #72913); neither rebased against the other, so `master` itself is red on this job. Every one of the last 8 `master` commits fails the same `Product tests (replay-vision, workflows, events schema legacy)` job, so CI (which runs against the PR merged with `master`) inherits the failure regardless of this branch's contents.

This branch touches nothing in `products/workflows`, and the fix belongs there — it needs a workflows-team call on whether those secret-inputs tests should stage drafts through the id-addressed graph-patch endpoint instead of a whole-`actions` PATCH. I've left it alone rather than pull an unrelated product change into a data-imports PR. Once `master` is green on that job again, this PR's CI will go green with no further changes here.

Gilbert09:
### Update: CI is green

The master-wide `products/workflows` breakage flagged above was fixed upstream by #74000; I merged the latest `master` in and CI now passes. One further red on the re-run was an unrelated pre-existing flake — `posthog/hogql/printer/test/test_printer.py::TestPrinter::test_union_by_name_executes_on_clickhouse`, which asserts an exact row order on a `UNION ALL BY NAME` query that has no `ORDER BY` (ClickHouse doesn't guarantee row order without one). A re-run passed. This PR touches nothing in hogql, so I left that test alone rather than pull an unrelated file into a data-imports diff.

All checks are green, no unresolved review threads, and the branch is `CLEAN`/mergeable against `master`.

PR #73708, opened 2026-07-25
fix(cli): normalize Windows paths before spawning Node

### LINKED ISSUE #73389: posthog-cli api crashes on Windows with `EISDIR ... lstat 'C:'` (v0.9.1) — every api subcommand unusable
> **Note:** This issue was drafted by an AI coding agent (Claude) while helping a
> user set up the PostHog CLI for agent use. The diagnosis below is based on
> direct reproduction on the user's machine, but please take the root-cause
> analysis with a grain of salt and verify independently.

## Summary

On Windows, every `posthog-cli api …` subcommand crashes instantly with a Node
main-path resolution error, before any tool logic runs. This makes the entire
agent-facing `api` catalog unusable on Windows, and also breaks the
`@posthog/wizard cli add` flow, which calls `posthog-cli api agents-md install`
and `posthog-cli api skill install` during "set up Claude Code / Codex".

## Environment

- OS: Windows 11 (build 26200)
- `posthog-cli --version`: **0.9.1** (latest on npm at time of report)
- Installed via: `npm i -g @posthog/cli` (global; wrapper at `%APPDATA%\npm\posthog-cli`)
- Node: v24.16.0 (`C:\Program Files\nodejs\node.exe`)

## Repro

```
posthog-cli api tools
```

## Actual

```
node:fs:2745
    const out = binding.lstat(base, false, undefined, true /* throwIfNoEntry */);
                        ^
Error: EISDIR: illegal operation on a directory, lstat 'C:'
    at Object.realpathSync (node:fs:2745:25)
    at toRealPath (node:internal/modules/helpers:63:13)
    at Module._findPath (node:internal/modules/cjs/loader:803:22)
    at resolveMainPath (node:internal/modules/run_main:35:21)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:146:20)
    at node:internal/main/run_main_module:33:47 {
  errno: -4068, code: 'EISDIR', syscall: 'lstat', path: 'C:'
}
```

Exit code 1. Reproduces both in a plain PowerShell and in Git Bash. Non-`api`
subcommands (`--version`, `--help`, `login`, `sourcemap`, etc.) are unaffected —
the crash is specific to the `api` path, which is the only one that shells out to
`node <bundled .mjs>`.

## Expected

`posthog-cli api tools` prints the tool catalog.

## Root cause (narrowed down)

The `api` subcommand spawns `node` to run the bundled agent-cli script. The crash
is in Node's **main-path resolution** (`resolveMainPath` -> `_findPath` ->
`toRealPath` -> `realpathSync('C:')`), i.e. it happens *before the script runs* —
so the script content is not at fault. Node is being handed a main path that
resolves to a bare drive root (`C:`) instead of the full script path. This looks
like an argument-quoting / path-construction bug in the Windows `node` spawn
inside `posthog-cli.exe`.

**Proof the script itself is fine** — invoking it directly works every time:

```
node "C:\Users\<user>\AppData\Roaming\npm\node_modules\@posthog\cli\lib\posthog-api-cli.mjs" tools   # prints catalog
node "…\@posthog\cli\lib\posthog-api-cli.mjs" search insight                                         # works
```

(The same script is also cached at `~/.posthog/api-cli/<version>/posthog-api-cli.mjs`
and runs identically.)

## Impact

- `posthog-cli api …` is completely unusable on Windows.
- `@posthog/wizard cli add` fails at the Claude Code / Codex setup step because it
  invokes `posthog-cli api agents-md install` / `api skill install`.

## Workaround

Bypass the exe and run the bundled script directly via a small wrapper on PATH:

```bat
@echo off
node "%~dp0node_modules\@posthog\cli\lib\posthog-api-cli.mjs" %*
```

Then `posthog-api <args>` == `posthog-cli api <args>`.

## Suggested fix

Review how `posthog-cli.exe` builds the `node` invocation for the `api` subcommand
on Windows — the script path passed as Node's main argument is being truncated to
the drive root. Likely needs proper quoting / verbatim-arg handling when
constructing the child-process command line.

CreepyGnome:
## Problem

On Windows, the CLI canonicalizes the API bundle path to a verbatim `\\?\` path before passing it to Node. Recent Node releases can resolve that path as `C:` and exit with `EISDIR`.

Closes #73389

## Changes

Use `dunce::canonicalize` for paths passed to Node so Windows paths remain normalized without the verbatim prefix. Add a Windows assertion that prevents this regression.

- `cargo test embedded_bundle_is_materialized_into_a_versioned_install_dir -- --nocapture`
- `cargo fmt --all -- --check`
- Ran `posthog-cli api skill list` with Node 22.23.1, 24.18.0, and 26.5.0. Each returned 210 skills with exit code 0.
- The relevant tests passed. The full suite encountered a Windows newline mismatch in sourcemap::test_pair_inject, which this PR does not modify.
- Clippy was blocked by a `needless_borrows_for_generic_args` finding in unchanged code at `src/login.rs:195`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No documentation changes are needed. This corrects Windows path handling without changing CLI usage.

## 🤖 Agent context

I directed the investigation and fix. Codex identified the Windows verbatim path interaction, implemented path normalization, added regression coverage, and ran validation. The repository `writing-tests` skill and GitHub publication guidance were used.

Codex used PowerShell, Git, Cargo, and Node.js, plus the repository `writing-tests` skill and GitHub publication guidance. We reproduced the Windows verbatim path failure, chose `dunce::canonicalize` to avoid manual path rewriting, and added Windows regression coverage

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -2093,6 +2093,7 @@ dependencies = [
  "colored",
  "crossterm 0.28.1",
  "dirs",
+ "dunce",
  "globset",
  "inquire",
  "libc",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -24,6 +24,7 @@ path = "src/main.rs"
 [dependencies]
 clap = { version = "4.5.31", features = ["derive", "env"] }
 dirs = "6.0.0"
+dunce = "1.0.5"
 inquire = "0.7.5"
 chrono = "0.4.42"
 posthog-symbol-data = "0.5.0"
--- a/cli/src/api_proxy.rs
+++ b/cli/src/api_proxy.rs
@@ -108,7 +108,7 @@ impl ApiProxyError {
 }
 
 fn canonicalize_file(path: &Path) -> Option<PathBuf> {
-    let resolved = path.canonicalize().ok()?;
+    let resolved = dunce::canonicalize(path).ok()?;
     if resolved.is_file() {
         Some(resolved)
     } else {
@@ -152,8 +152,7 @@ fn materialize_embedded_script(
         .map_err(|source| materialize_failed(MaterializeStep::CreateDir, source))?;
     fs::write(&path, bundle)
         .map_err(|source| materialize_failed(MaterializeStep::Write, source))?;
-    let resolved = path
-        .canonicalize()
+    let resolved = dunce::canonicalize(&path)
         .map_err(|source| materialize_failed(MaterializeStep::Canonicalize, source))?;
     if !resolved.is_file() {
         return Err(materialize_failed(
@@ -212,11 +211,10 @@ fn find_script() -> Result<PathBuf, ApiProxyError> {
         }
         let path = PathBuf::from(path);
         let resolved =
-            path.canonicalize()
-                .map_err(|source| ApiProxyError::ConfiguredPathMissing {
-                    path: path.display().to_string(),
-                    source,
-                })?;
+            dunce::canonicalize(&path).map_err(|source| ApiProxyError::ConfiguredPathMissing {
+                path: path.display().to_string(),
+                source,
+            })?;
         if !resolved.is_file() {
             return Err(ApiProxyError::ConfiguredPathNotAFile {
                 path: resolved.display().to_string(),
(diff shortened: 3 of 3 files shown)


AlexlaGuardia:
This matches what I traced on #73389: the `\\?\` verbatim prefix out of `Path::canonicalize`, not argv quoting. Thanks for picking it up.

Worth flagging for reviewers that all three producers of the script path are covered here (`canonicalize_file`, `materialize_embedded_script`, and the `POSTHOG_API_CLI_PATH` branch of `find_script`). Any one left on `Path::canonicalize` keeps a live route to the same `EISDIR`.

`dunce::canonicalize` is also the safer choice over stripping the prefix by hand, since it keeps the verbatim form where it is actually needed (paths past the 260 char limit, UNC shares), so a long install dir does not regress the other way.

PR #73712, opened 2026-07-26
feat(data-modeling): let suspended models be resumed

sakce:
## Problem

When a data modeling node fails to materialize five times in a row, we suspend it: scheduled DAG runs skip it and everything downstream of it.
That part works.
Getting back out does not.

Suspension only ever cleared on a *successful materialization*, and a suspended node is skipped by every scheduled run.
So it can never succeed, so it can never clear.
Fixing the broken SQL does nothing. Waiting does nothing. The model just stops updating, quietly, forever.

The trap is easiest to see from the user's side: you get a suspended model, you fix the query that broke it, and the model still never runs again.

## Changes

Five ways back, where before there was only one that a suspended node could never reach.

| Trigger | Where |
| --- | --- |
| The saved query's SQL changes | `logic/saved_query_dag_sync.py` |
| Manual materialize, and saved query "Sync now" | `logic/node_materialization.py` |
| Run upstream/downstream, for the whole cone the request names | `views/node.py::run` |
| `POST .../data_modeling_nodes/{id}/resume/`, plus a `suspended` read field on the node | `views/node.py::resume` |
| `POST .../warehouse_saved_queries/{id}/resume/`, scoped `warehouse_view:write` | `views/saved_query.py::resume` |

Three things worth a reviewer's attention.

**Query changes are detected by fingerprint, not by a flag passed in from the edit path.**
`mark_node_suspended` records a hash of the SQL that got the node suspended, and the DAG sync clears only when the current hash differs.
That matters because `sync_saved_query_to_dag` has many internal callers (backfills, resync commands, managed viewset sync).
Clearing unconditionally there would mean every resync silently un-suspends everything.
Markers written before this PR carry no fingerprint, so they read as stale and clear on the first edit. That is deliberate: otherwise they are unclearable by this path.

**Resuming stamps a watermark, and the consecutive-failure counter ignores jobs from before it.**
Without that, the five failures that caused the suspension are still the five most recent jobs, so a resumed node re-suspends on its very next failure and an edit buys exactly one attempt.

**There are two resume endpoints on purpose, and the saved query one is the one to use.**
`NodeViewSet` is `scope_object = "INTERNAL"`, which means `_get_required_scopes` returns `None`, which means no personal API key or OAuth token can call it at all - the refusal is literally "This action does not support personal API key access". It also means `AccessControlPermission` skips it, so warehouse RBAC does not apply (that is why `lineage` re-applies the check by hand).
The saved query is what users, MCP tools and scripts actually hold, so the resume lives there too under `warehouse_view:write`, where scopes and RBAC both work.
It resumes every node backing the query, which matters because one query can back several nodes when it landed in duplicate DAGs.
That follows the existing shape of that viewset: `run` and `revert_materialization` already do node and DAG work through the data_modeling facade.

> [!NOTE]
> Behavior change on the run button. `POST .../data_modeling_nodes/{id}/run/` now resumes every node in the cone it was asked to run. Previously that request was a silent no-op for suspended nodes, since `ExecuteDAGWorkflow` skips them. Detection is unchanged: a node that keeps failing still suspends after five consecutive failures, it just gets a fresh window after a resume.

Suspension state moved to `products/data_modeling/backend/logic/node_suspension.py`, next to the paths that clear it.
`posthog/temporal/.../activities/utils.py` re-exports it through the facade, so the activities keep their existing imports and the JSON shape has one definition instead of two.

Automated only. I (Claude) ran these; I did not exercise any of it by hand in a browser.

- `products/data_modeling/backend/test/test_node_suspension.py` (new, 5 cases) - covers the regression this PR exists for: a suspended node whose query changes must clear, and one whose query is re-synced unchanged must **not** (that second case is the guard against a resync command mass-un-suspending everything). Also that resume stamps the watermark and is a no-op on a healthy node, so "Sync now" doesn't write properties on every dispatch.
- `posthog/temporal/tests/data_modeling/test_materialize_view_activities.py` - one new case: failures from before a resume must not re-suspend. Verified it is load-bearing by disabling the `since` bound and watching it fail. The existing suspension cases already cover the unbounded path.
- `products/data_modeling/backend/tests/api/test_node_api.py` - three new cases: read `suspended` then resume it away, run resumes the cone, materialize resumes the node.
- `products/data_warehouse/backend/tests/api/test_saved_query.py` - one new case, deliberately driven with a personal API key carrying `warehouse_view:write`, over a query backed by two nodes in two DAGs. Catches both the thing the node endpoint can't do (key access) and the multi-node fan-out.
- Suites green: 477 in `products/data_modeling/backend`, 87 in `test_saved_query.py`, 50 node API, 108 across the temporal materialize/execute-dag and DAG-sync suites.
- `uv run mypy --cache-fine-grained .` clean repo-wide, `ty` clean, ruff clean, `tach check --dependencies --interfaces` clean.
- `hogli build:openapi` regenerated six files, all of them downstream of these two serializers.

`ci:preflight` flags one openapi advisory I did not introduce: the local generator produces different `products/dashboards/frontend/generated/widget-*.json` than what is committed, on files this branch leaves at master's content. I left them alone rather than commit a regeneration of another product's generated output. Shout if you'd rather I fold it in.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover model suspension yet, so nothing to update here. Surfacing it in the UI is the follow-up, and that is where docs will want to land.

## 🤖 Agent context

Written by Claude Opus 5 in Claude Code, driven by me. Skills invoked: `/writing-tests` before adding any test (it is what pushed the API cases into existing test files instead of new ones, and killed a couple of near-duplicates), and `/improving-drf-endpoints` before touching either viewset (that one is why `suspended` gets a typed sub-serializer rather than a bare `DictField`, so the generated TypeScript is `NodeSuspensionApi` with real field docs instead of `Record<string, unknown>`).

The first design put the reset-on-edit call in the `data_warehouse` saved-query serializer, which already knows whether the query changed. Rejected: it leaves every other edit path uncovered. The fingerprint approach covers all of them. We also considered treating a fingerprint mismatch as "not suspended" at read time instead of actively clearing, which self-heals with no writes, but then stored state and enforced state disagree and any UI has to re-derive the same logic.

The saved-query endpoint started as a follow-up and got folded in after we looked at what `INTERNAL` actually costs: no key access and no RBAC, paid back by hand on exactly one action. Worth reviewing as a direction, not just a diff - the same argument applies to `run` and `materialize`, which are node-level today.

Left out on purpose: no per-engine argument on either resume (nothing calls for it yet), no activity log entry on the saved-query action (its siblings log, so say the word if you want one), and no `suspended` field on the saved-query serializer - the UI that needs it is the next PR. Separately, the ungated node writes deserve their own fix; I have not touched that here.

Tests only prove the mechanism. Models already suspended in production still need a resume once this deploys.


products/data_modeling/backend/logic/node_suspension.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def clear_node_suspension(node: Node, *, engine: str | None = None, by: str = "materialization") -> bool:
+    """Clears one engine, or all of them when engine is None.
+
+    Stamps a watermark because the failures that caused the suspension are still the most recent
+    jobs — without it the node re-suspends on its very next failure.
+    """
+    suspended = suspension_state(node)
+    engines = [str(engine)] if engine is not None else list(suspended)
```
thiagosalvatore, line 75: why the str(engine) here if engine is already typed as str?

thiagosalvatore (approved):

posthog/temporal/data_modeling/activities/utils.py
```diff
@@ -125,13 +114,17 @@ def maybe_suspend_node_for_engine(
...
     with transaction.atomic():
         node = Node.objects.select_for_update().get(id=node_id, team_id=team_id, dag_id=dag_id)
         if is_node_suspended(node, engine):
             return False
-        mark_node_suspended(node, engine=engine, reason=reason, job_id=job_id)
+        since = suspension_reset_at(node, engine)
+        if _count_leading_failures(saved_query_id, engine, since=since) < CONSECUTIVE_FAILURES_TO_SUSPEND:
+            return False
+        fingerprint = query_fingerprint(
+            DataWarehouseSavedQuery.objects.filter(id=saved_query_id).values_list("query", flat=True).first()
```
thiagosalvatore, line 125: The fingerprint is captured here from the *live* saved query at suspend time. If a query edit (A→B) lands in the narrow window where the 5th failure is suspending and this path wins the node row lock, it records fingerprint(**B**) — then `_clear_stale_fingerprints` sees B == B and won't clear on the edit, so fixing the SQL leaves the node stuck (the exact trap this PR fixes).

It's a tight race and self-heals via any further edit / `run` / `materialize` / `resume`, so not a blocker. I confirmed the mechanism with a probe (a marker carrying the current query's fingerprint survives a same-query sync). If you want to close it, fingerprinting the query the *failing job* actually ran — rather than the current saved query — would make the edit-clear reliable.

products/data_modeling/backend/presentation/views/node.py
```diff
@@ -213,9 +242,21 @@ def perform_destroy(self, instance: Node) -> None:
...
         instance.delete()
 
+    def _require_warehouse_access(self, *, level: AccessControlLevel, message: str) -> None:
+        """`scope_object = "INTERNAL"` makes AccessControlPermission skip this viewset entirely, so
+        warehouse RBAC has to be re-applied by hand (warehouse_view inherits warehouse_objects)."""
+        if not self.user_access_control.check_access_level_for_resource("warehouse_view", required_level=level):
+            raise PermissionDenied(message)
+
+    def retrieve(self, request, *args, **kwargs):
+        self._require_warehouse_access(level="viewer", message=_READ_DENIED)
```
thiagosalvatore, line 252: `retrieve` and `list` gain a `warehouse_view:viewer` gate here; on master neither was warehouse-gated (only team membership — just `lineage` had the check). The tightening is well-justified (the `suspended` field carries the raw materialization error), but it's a behavior change: a team member with project access but no warehouse access will now get a 403 on the data-modeling scene. Worth confirming that matches the UI's current access model so it doesn't surface as "I can't see my models anymore."

products/data_modeling/backend/presentation/views/node.py
```diff
@@ -280,6 +321,10 @@ def run(self, req: request.Request, *args, **kwargs) -> response.Response:
 
         node_ids.add(str(node.id))
 
+        # ExecuteDAGWorkflow skips suspended nodes, so without this the request is a silent no-op
+        # for exactly the nodes that need it most.
+        resume_nodes(Node.objects.filter(team_id=self.team_id, id__in=node_ids), by="manual_run")
```
thiagosalvatore, line 326: This clears suspension (a write) with no access check, while the dedicated `resume` action is gated at `editor` for the same effect ("resuming puts a model back on the schedule"). By that logic `run`/`materialize` arguably want the same gate. Not newly exploitable — these already triggered materializations without a check — and you flagged the ungated node writes as a separate follow-up, so fine to defer; noting the inconsistency.

products/data_warehouse/backend/presentation/views/saved_query.py
```diff
@@ -1131,6 +1135,20 @@ def run(self, request: request.Request, *args, **kwargs) -> response.Response:
...
+    @action(methods=["POST"], detail=True, required_scopes=["warehouse_view:write"])
+    def resume(self, request: request.Request, *args, **kwargs) -> response.Response:
+        """Resume materialization suspended after repeated failures.
+
+        Scheduled runs skip a suspended model and everything downstream of it, so it cannot succeed
+        its way back on its own.
+        """
+        from products.data_modeling.backend.facade.api import resume_saved_query
+
+        resumed = resume_saved_query(self.get_object())
```
thiagosalvatore, line 1148: Unlike `run` / `revert_materialization` / `materialize` on this viewset, `resume` doesn't `log_activity`. Resuming is a meaningful state change on a shared resource — worth an audit entry for parity. (You noted this in the description; flagging inline so it's tracked.)

products/data_modeling/backend/logic/node_suspension.py
```diff
@@ -0,0 +1,136 @@
...
+    node.properties = properties
+    return True
+
+
+def _persist_change(node: Node, change: Callable[[Node], bool]) -> bool:
+    """Apply a change to the node's suspension state as a locked read-modify-write.
+
+    `properties` is one JSON blob, so writing it from a caller's in-memory instance would drop
+    anything a materialization committed to the same row since that instance was read. Every other
+    writer of this field takes the row lock too (see the data-modeling Temporal activities).
```
thiagosalvatore, line 96: Small accuracy nit: `set_declared_target` in `node_frequency.py` writes `properties` without `select_for_update`, so "every other writer of this field takes the row lock too" isn't quite true. `_persist_change` still protects resume against prior-committed writes (the tested case); it's only an unlocked *concurrent* writer that could lose an update, and that race predates this PR. Worth softening the comment.

products/data_modeling/backend/logic/node_suspension.py
```diff
@@ -0,0 +1,136 @@
...
+        if locked is None or not change(locked):
+            return False
+        locked.save(update_fields=["properties"])
+    node.properties = locked.properties
+    return True
+
+
+def resume_nodes(nodes: Iterable[Node], *, by: str, engine: str | None = None) -> int:
+    """Returns how many of the nodes were actually suspended, not how many were passed in."""
+    return sum(
```
thiagosalvatore, line 109: `resume_nodes` opens one `atomic()` + `select_for_update` per node via `_persist_change`. The `run` endpoint passes the whole downstream cone through here, so healthy (non-suspended) nodes each still take a transaction + row lock as a no-op — N synchronous round-trips on the request path for large cones. Pre-filtering to actually-suspended nodes (e.g. a `properties__system__suspended` JSON filter) would skip most of them. Low priority since runs are manual/infrequent.

products/data_warehouse/backend/presentation/views/saved_query.py
sakce: good shout - will follow up

products/data_modeling/backend/presentation/views/node.py
sakce: yep, intentional. UI will follow

PR #73737, opened 2026-07-26
feat(cohorts): backfill completion protocol django foundation

matheus-vb:
## Problem

The behavioral cohorts backfill has shipped the seeder and reconcile snapshot pieces (B1 to B4), but nothing terminalizes a fully confirmed run. `last_backfill_events_at` never gets stamped, so `is_flag_compatible` never opens. This PR lays the inert Django foundation for the B5 completion protocol. It is the base of a three PR stack (this, then #73738, then #73739).

## Changes

- Migration `0006_cohort_backfill_completion` adds nullable completion columns the Rust seeder will write: `chunks_planned_at`, `reconcile_dispatched_at`, `reconcile_observed_at`, `marker_watch` (opaque JSON) on `CohortBackfillRun`, and `reconcile_completed_at` plus `reconcile_marker_bits` (bigint bitmap, maintained as an i64 by Rust) on `CohortBackfillRunCohort`. Django never interprets the bits or the watch blob.
- New finalizer in `backfill/finalize.py` plus a thin `finalize_cohort_backfill_runs` Celery task on an every 2 minute beat, gated by `BEHAVIORAL_BACKFILL_FINALIZER_ENABLED` (default off). It discovers fully observed reconciling runs, CASes each to `completed` or `superseded`, stamps events readiness per participation, and invalidates the behavioral cohort and flags caches per team after commit.
- Readiness fixes in `readiness.py`: refuse to stamp a superseded participation (blocks the A to B to A revert resurrection), and add `superseded_at` guards to the stamp CAS with a cohort rollback when a supersession races the stamp.
- `supersede_active_runs` reorders its writes (run rows before participations) to match the finalizer's lock order and avoid a deadlock on cohort scoped runs.

Everything is inert on merge: the columns are unread and the finalizer returns before touching the DB while gated off.

New `test_finalize.py` covers the outcome matrix (completed, completed plus superseded, all superseded, superseded trumps completed, shortfall held, missing outcome held, mixed held still stamps), unobserved runs skipped, second fire noop, disabled setting does not touch the DB, cache invalidation targets, and the cohort scoped supersede by divergence path. `test_readiness.py` is extended for superseded never resurrected and the stamp rollback race.


products/cohorts/backend/backfill/finalize.py
```diff
@@ -0,0 +1,230 @@
...
+            if terminal_status == CohortBackfillRunStatus.COMPLETED:
+                result.completed += 1
+                RUNS_FINALIZED_COUNTER.labels(status="completed").inc()
+            else:
+                result.superseded += 1
+                RUNS_FINALIZED_COUNTER.labels(status="superseded").inc()
+        else:
+            # We hold the run row's FOR UPDATE lock, so a missed CAS can only mean our own
+            # stamp_events_readiness call superseded a cohort-scoped run inside this transaction —
+            # the run is terminal (superseded + finished_at) and counted as such. This relies on a
```
gustavohstrassburger: `suggestion`: This relies on a cohort-scoped run having exactly one participation, and that's enforced by application code in `runs.py` (`create_backfill_run_for_cohort`), not by a DB constraint. `CohortBackfillRunCohort`'s only constraint is `cohort_bfrc_run_cohort_uq` on `(run, cohort)`, which allows any number of participations per run regardless of scope. If a second participation ever lands on a COHORT-scoped run, one could stamp while the terminal CAS returns 0 rows, and the run gets counted `superseded` even though a cohort was stamped and its cache invalidated. Worth pointing this comment at exactly where the invariant is created (`runs.py:100-119`), or adding a defensive check in `_finalize_one_run` that holds the run and logs loudly if `run.scope == COHORT and len(participations) != 1`, rather than silently miscounting.

products/cohorts/backend/backfill/test/test_finalize.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+
+    def test_flags_cache_task_name_matches_the_registered_task(self) -> None:
+        # The dispatch is by name (a static import would create a product-dependency cycle), so pin
+        # the string against the task's registered name to catch a rename or move.
+        module_path, task_name = FLAGS_CACHE_TASK.rsplit(".", 1)
+        task = getattr(importlib.import_module(module_path), task_name)
+        self.assertEqual(task.name, FLAGS_CACHE_TASK)
+
+    def test_cohort_scoped_run_superseded_by_stamp_divergence(self) -> None:
```
gustavohstrassburger, line 215: `blocking`: This builds the single-participation case by hand, so the invariant `finalize.py:217` depends on (a cohort-scoped run has exactly one participation) is exercised only in the shape that already holds it. Nothing tests two participations on a COHORT-scoped run, so the miscount that invariant guards against is unverified behavior rather than a known, asserted one. Either enforce the invariant with a test proving rejection, or add a test with two participations pinning the actual (miscounted) outcome so a regression here fails a test instead of drifting silently.

products/cohorts/backend/backfill/test/test_readiness.py
```diff
@@ -132,6 +132,54 @@ def test_already_stamped_readiness_is_not_overwritten(self) -> None:
...
+
+        # A supersession commits after the cohort stamp but before the participation CAS ratifies
+        # it. The stamp is this transaction's own write, so it has to be taken back.
+        with connection.execute_wrapper(supersede_after_the_cohort_stamp):
+            self.assertFalse(stamp_events_readiness(run, cohort.id))
+
+        cohort.refresh_from_db()
+        participation.refresh_from_db()
+        self.assertIsNone(cohort.last_backfill_events_at)
+        self.assertIsNone(participation.stamped_at)
```
gustavohstrassburger, line 181: `blocking`: This test asserts the cohort rollback and that `stamp_events_readiness` returns `False`, but never asserts the participation actually ended up `superseded`, which is the terminal state the finalizer's `else` branch reads to count the run. The test would pass unchanged if the rollback path stopped setting `superseded_at`. Worth adding:

```suggestion
        self.assertIsNone(participation.stamped_at)
        self.assertIsNotNone(participation.superseded_at)
```

products/cohorts/backend/backfill/finalize.py
```diff
@@ -0,0 +1,230 @@
...
+    # per team inside the loop, so the unscoped read is discovery only.
+    observed = list(
+        CohortBackfillRun.objects.unscoped()
+        .filter(
+            # B5 stamps only the events column, so never terminalize a person-properties run (B7).
+            backfill_kind=CohortBackfillKind.BEHAVIORAL,
+            status=CohortBackfillRunStatus.RECONCILING,
+            reconcile_observed_at__isnull=False,
+        )
+        .values_list("id", "team_id")
```
gustavohstrassburger: `suggestion`: The index in migration 0007 fixes the indexing half of ReviewHog's earlier finding on this scan, but there's still no `[:N]` slice on `.values_list("id", "team_id")`. Steady state is bounded by the unique constraints on `CohortBackfillRun`, but a run with any unresolved participation gets returned to `reconciling` and re-scanned every 2-minute pass indefinitely. If the Rust seeder stalls across many teams, the result set grows monotonically with no cap on the pass itself. A slice plus `order_by("reconcile_observed_at")` (free here, it's the index's second column) keeps each pass bounded and gives held runs a fair shot:

```python
.order_by("reconcile_observed_at")
.values_list("id", "team_id")[: settings.BEHAVIORAL_BACKFILL_FINALIZER_MAX_RUNS_PER_PASS]
```

products/cohorts/backend/backfill/readiness.py
```diff
@@ -51,19 +51,41 @@ def stamp_events_readiness(run: CohortBackfillRun, cohort_id: int) -> bool:
 
     This update intentionally bypasses signals. B5 must explicitly invalidate feature-flag and
     behavioral-cohort caches after a successful stamp.
+
+    Call inside a transaction: the cohort stamp and the participation CAS that ratifies it must
+    commit or roll back together.
     """
     participation = CohortBackfillRunCohort.objects.for_team(run.team_id).get(run_id=run.id, cohort_id=cohort_id)
```
gustavohstrassburger, line 58: `suggestion`: This re-fetches the participation row `_finalize_one_run` already has in hand a few lines earlier at `finalize.py:154`, inside a transaction holding a `FOR UPDATE` lock on the run row. It's not buying freshness either, the actual race protection is the `superseded_at__isnull=True` guard on the ratifying CAS below, not this read. Passing the row in instead of `cohort_id` drops a round trip that scales with the team's cohort count on every pass. Your call given the signature churn at other call sites.

products/cohorts/backend/backfill/finalize.py
```diff
@@ -0,0 +1,230 @@
...
+    CohortBackfillRunStatus,
+)
+from products.cohorts.backend.models.dependencies import invalidate_team_behavioral_cohort_cache
+
+logger = structlog.get_logger(__name__)
+
+# products.feature_flags already depends on products.cohorts, so a static import of its task here
+# would close a forbidden product-dependency cycle; dispatch by registered task name instead.
+# test_finalize.py pins this string against the task's registered name.
+FLAGS_CACHE_TASK = "products.feature_flags.backend.tasks.update_team_service_flags_cache"
```
gustavohstrassburger, line 41: `suggestion`: This pins the implicit registered name of `update_team_service_flags_cache`, but that task has no explicit `name=` on its `@shared_task` decorator, so the string is just the dotted module path, it changes the moment `tasks.py` moves into a package. The one other cross-product by-name dispatch in this repo (`customer_analytics.process_custom_property_sync`) pins the name on the task itself for exactly this reason (`hogli isolate`'s `pin_task_names` step exists because a module move silently renames a task's registration path and strands queued messages at deploy). Adding `name="products.feature_flags.backend.tasks.update_team_service_flags_cache"` to that decorator declares the contract on both ends, your `test_flags_cache_task_name_matches_the_registered_task` still catches a rename either way, but a pinned name makes it a non-event instead of something to reason through mid-deploy.

products/cohorts/backend/backfill/finalize.py
```diff
@@ -0,0 +1,230 @@
...
+    runs_scanned: int = 0
+    completed: int = 0
+    superseded: int = 0
+    held: int = 0
+    stamped_participations: int = 0
+    invalidated_teams: int = 0
+
+
+@dataclass
+class _RunOutcome:
```
gustavohstrassburger: `suggestion`: This dataclass declares four fields but the caller only ever reads `invalidate_team`. `stamped`, `superseded`, and `held` are read exclusively inside `_finalize_one_run` and are already duplicated in the `result: FinalizerPass` parameter mutated in place a few lines below. Returning a plain `bool` for "does this team need invalidating" instead of `_RunOutcome | None` drops the dataclass and the `outcome is not None and` guard at the call site, since `None` and `invalidate_team=False` are already treated identically there.

products/cohorts/backend/backfill/finalize.py
```diff
@@ -0,0 +1,230 @@
...
+    ["result"],  # labels: "stamped", "superseded"
+)
+
+RUNS_FINALIZED_COUNTER = Counter(
+    "posthog_cohort_backfill_runs_finalized_total",
+    "Backfill runs terminalized by the finalizer, by terminal status",
+    ["status"],  # labels: "completed", "superseded"
+)
+
+HELD_RUNS_GAUGE = Gauge(
```
gustavohstrassburger, line 57: `suggestion`: `held` counts two different things and this gauge publishes their sum, a legitimately-waiting shortfall and a crashed/rolled-back run. The help text describes only the first, so an alert on this gauge can't distinguish routine backpressure from a crash loop. Splitting into `held`/`errored`, or a labeled gauge with `["reason"]`, lets the alert target the one that actually needs paging.

gustavohstrassburger (approved):
Solid foundation PR, careful concurrency work. A few non-blocking suggestions inline.

PR #73738, opened 2026-07-26
feat(cohorts): seeder planning proof and auto reconcile dispatch

matheus-vb:
## Problem

With the completion columns in place (#73737), the Rust seeder still can't move a fully confirmed run out of `seeding`, and the reconcile dispatch receipt is print only. This PR adds the producer half of the completion protocol: prove planning finished, auto transition a confirmed run `seeding` to `reconciling`, produce the reconcile control tiles, and persist the dispatch record (produce HWMs, marker watch start positions, fence epoch). Stacked on #73737; all new behavior is dark by default.

## Changes

- New completion domain (`domain/completion.rs`, `partition.rs`, `pinned.rs`): parse don't validate newtypes that make illegal states unrepresentable. `PartitionBitmap` packs the 64 partition set into the signed BIGINT with an exact bit 63 round trip, `ReconcileHwms` refuses any non full partition set, `ProducedOffset` / `CommittedOffset` / `NextOffset` keep the read versus next offset off by one unrepresentable, `DispatchEpoch` is `#[must_use]` and mintable only from a stored timestamp, and `CompletionPhase::from_parts` classifies each run into `Observed`, `Reconciling`, `ReconcilingUndispatched`, `SeedingPlanned`, `SeedingUnplanned`, or `SeedingAnomalous`. `SeedPartition` moves out of the kafka module into the domain (kafka re-exports it).
- Store (`store/completion.rs`): `mark_chunks_planned` and `read_planning_stamp` record the planning proof; `cas_run_reconciling` freezes the ledger and transitions `seeding` to `reconciling` only when planning is proven and every chunk confirmed; `confirm_reconciling` re-claims an already reconciling run for self healing re-dispatch; `ReconcilingClaim::record` atomically writes the fence epoch per INV-3 and `revert` rolls back; `discover_completions` classifies each behavioral run's phase; `mark_run_observed_unreconcilable` hands an all superseded run to Django. Every observation write is fenced on `reconcile_dispatched_at` plus status, returning a typed fence loss on a miss.
- App: `CompletionDriver` is ticked from the orchestrator poll arm after `fill_claim_slots`. Dispatch runs in a spawned task with per replica in flight dedupe, so a run of cohorts times 64 acks can't blow the liveness deadline. The planning stamp is wired into `prepare.rs` and withheld when no pinned conditions survive an active participation (fail closed). A `Certified` / `Uncertified` split means only a completion proven dispatch can reach persistence; `validate_completion` now gates on the planning proof (`PlanningUnproven`) instead of an empty chunk ledger. The `reconcile_dispatch` CLI is upgraded so its default invocation is a mutation (claim, persist, print the fence); `--allow-incomplete` stays print only with a loud "NOT persisted" warning.
- Kafka producer: `capture_topic_offsets` reads the membership topic high watermarks before producing the tiles, so every marker of the dispatch lands at or above the recorded watch positions.
- Config: `seeder_reconcile_auto_dispatch_enabled` and `seeder_confirm_register_backfilled` (both default false) plus `cohort_membership_changed_topic`. Arming requires both flags and `COHORT_PARTITION_COUNT == 64`, otherwise a startup error. New producer side metrics added.

New `tests/pg_completion.rs` integration suite covers the planning stamp, the reconciling CAS single winner and its guards, the atomic dispatch record and fresh epoch, fenced writes rejected under a stale epoch, and phase discovery. Domain and app unit tests cover the bitmap round trip, the liveness boundary, `CompletionPhase` classification, and the auto dispatch policy gating. The test fixture is regenerated against migration `0006`. Run with `cargo test -p cohort-seeder` (the pg suite needs a local Postgres).

### DIFF
--- a/rust/cohort-seeder/src/app/completion.rs
+++ b/rust/cohort-seeder/src/app/completion.rs
@@ -218,10 +218,18 @@ impl CompletionDriver {
         }
         let context = self.context.clone();
         let guard = InFlightGuard::enter(Arc::clone(&self.in_flight), run_id);
-        tokio::spawn(async move {
+        let dispatch = tokio::spawn(async move {
             let _permit = permit;
             let _guard = guard;
             run_dispatch(&context, run_id, kind).await;
+        });
+        // `run_dispatch` reports its own failures, so a `JoinError` is a panic. Dropping the handle
+        // would make a deterministic one — a poison row, say — an invisible per-tick respawn loop.
+        tokio::spawn(async move {
+            if let Err(error) = dispatch.await {
+                counter!(RECONCILE_DISPATCHES, "outcome" => "panicked").increment(1);
+                warn!(error = %error, run_id = ?run_id, "reconcile dispatch task panicked");
+            }
         });
         true
     }
--- a/rust/cohort-seeder/src/app/completion.rs
+++ b/rust/cohort-seeder/src/app/completion.rs
@@ -18,6 +18,7 @@ use std::time::Duration;
 use common_types::cohort::TeamAllowlist;
 use metrics::{counter, gauge};
 use sqlx::PgPool;
+use tokio::sync::Semaphore;
 use tokio::task::JoinError;
 use tracing::warn;
 
@@ -30,12 +31,13 @@ use crate::domain::{
 };
 use crate::kafka::producer::{CaptureOffsetsError, SeedTileProducer};
 use crate::observability::metrics::{
-    RECONCILE_CAS_LOST, RECONCILE_DISPATCHES, RECONCILE_RECORD_INVALID, RUNS_RECONCILING,
+    RECONCILE_CAS_LOST, RECONCILE_DISPATCHES, RECONCILE_DISPATCHES_IN_FLIGHT,
+    RECONCILE_RECORD_INVALID, RUNS_RECONCILING,
 };
-use crate::store::chunks::PgChunkStore;
 use crate::store::completion::{
     cas_run_reconciling, confirm_reconciling, discover_completions,
-    mark_run_observed_unreconcilable, CompletionStoreError, ReconcilingClaim,
+    mark_run_observed_unreconcilable, runs_with_all_chunks_confirmed, CompletionStoreError,
+    ReconcilingClaim,
 };
 use crate::store::runs::ReconcileRunError;
 
@@ -112,9 +114,9 @@ struct DispatchContext {
 
 pub struct CompletionDriver {
     context: DispatchContext,
-    store: PgChunkStore,
     allowlist: TeamAllowlist,
     in_flight: Arc<Mutex<HashSet<RunId>>>,
+    dispatch_slots: Arc<Semaphore>,
 }
 
 impl CompletionDriver {
@@ -124,25 +126,27 @@ impl CompletionDriver {
         allowlist: TeamAllowlist,
         membership_topic: String,
         max_inflight: NonZeroUsize,
+        max_concurrent_dispatches: NonZeroUsize,
         register_backfill: RegisterBackfillConfirmation,
     ) -> Self {
         Self {
             context: DispatchContext {
+                pool,
                 producer,
                 membership_topic: Arc::from(membership_topic),
                 max_inflight,
                 register_backfill,
-                pool: pool.clone(),
             },
-            store: PgChunkStore::new(pool),
             allowlist,
             in_flight: Arc::new(Mutex::new(HashSet::new())),
+            dispatch_slots: Arc::new(Semaphore::new(max_concurrent_dispatches.get())),
         }
     }
 
     /// One driver pass: classify every discovered run and spawn dispatch tasks for the ones that need
     /// one. A discovery failure is logged and retried next tick; the dark path (driver absent) costs
-    /// zero queries.
+    /// zero queries. Every read is per-tick, never per-run — the pass blocks the orchestrator's
+    /// liveness heartbeat.
     pub async fn tick(&self) {
         let discovered = match discover_completions(&self.context.pool, &self.allowlist).await {
             Ok(discovered) => discovered,
@@ -153,6 +157,7 @@ impl CompletionDriver {
         };
 
         let mut reconciling = 0_u64;
+        let mut planned = Vec::new();
         for completion in discovered {
             match completion.phase {
                 CompletionPhase::Observed | CompletionPhase::Reconciling(_) => {
@@ -172,9 +177,7 @@ impl CompletionDriver {
                         );
                     }
                 }
-                CompletionPhase::SeedingPlanned => {
-                    self.maybe_dispatch_confirmed(completion.run_id).await;
-                }
+                CompletionPhase::SeedingPlanned => planned.push(completion.run_id),
                 CompletionPhase::SeedingUnplanned => {}
                 CompletionPhase::SeedingAnomalous => {
                     warn!(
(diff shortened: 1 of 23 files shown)


rust/cohort-seeder/src/store/completion.rs
```diff
@@ -0,0 +1,881 @@
...
+        watch: &MarkerWatch,
+    ) -> Result<DispatchEpoch, CompletionStoreError> {
+        let run_id = self.run_id;
+        let mut tx = pool.begin().await?;
+        let dispatched_at = sqlx::query_scalar::<_, DateTime<Utc>>(
+            r#"
+            UPDATE cohort_backfill_runs
+            SET reconcile_hwms = $2, marker_watch = $3,
+                reconcile_dispatched_at = now(), reconcile_observed_at = NULL, updated_at = now()
+            WHERE id = $1 AND status = 'reconciling'
```
gustavohstrassburger, line 236: `suggestion`: this `UPDATE` fences on `WHERE id = $1 AND status = 'reconciling'` only, no epoch check. Compare `revert` two methods up, which carries `AND reconcile_dispatched_at IS NOT DISTINCT FROM $2`, and every other completion write in this file (`persist_marker_watch`, `persist_marker_observations`, `mark_run_observed`), which all fence the same way. `record` is the one write that doesn't, despite `ReconcilingClaim` already carrying `dispatched_at_mint` for exactly this and using it in `revert`.

`confirm_reconciling` is documented as non-exclusive by design, so two replicas (or a replica racing an operator re-dispatch via the CLI) can both hold a claim for the same run and both call `.record()`. Whichever commits second silently overwrites the first's `reconcile_hwms`, `marker_watch`, and fence timestamp, and also zeroes every non-superseded participation's marker bits, completion, and error, including any the winner's observer already merged under the winner's epoch. No `CompletionFenceLost` is ever raised for this path, so the loser doesn't even know it lost.

```suggestion
        let dispatched_at = sqlx::query_scalar::<_, DateTime<Utc>>(
            r#"
            UPDATE cohort_backfill_runs
            SET reconcile_hwms = $2, marker_watch = $3,
                reconcile_dispatched_at = now(), reconcile_observed_at = NULL, updated_at = now()
            WHERE id = $1 AND status = 'reconciling'
              AND reconcile_dispatched_at IS NOT DISTINCT FROM $4
            RETURNING reconcile_dispatched_at
            "#,
        )
        .bind(run_id)
        .bind(Json(hwms))
        .bind(Json(watch))
        .bind(self.dispatched_at_mint)
```

The loser then gets the existing `CompletionFenceLost { operation: RecordDispatch }`, which `run_dispatch` already handles cleanly (counts `record_failed`, leaves the run for the next tick) — the right outcome for a replica that lost the race.

rust/cohort-seeder/src/app/reconcile_dispatch.rs
```diff
@@ -106,8 +179,9 @@ fn validate_completion(
                 remaining_chunks,
             });
         }
-        if total_chunks == 0 && status == RunStatus::Seeding {
-            return Err(PrepareReconcileDispatchError::EmptyChunkLedger(run_id));
+        let planned = status == RunStatus::Reconciling || planning_proven;
```
gustavohstrassburger: `suggestion`: the `status == RunStatus::Reconciling` disjunct treats "already reconciling" as proof of planning, but that's proof the ledger is frozen, not proof planning ran. `UndispatchedReason::NeverDispatched` documents exactly this case: a pre-protocol manually-reconciling run with NULL `chunks_planned_at`. `CompletionDriver::tick` classifies such a run `ReconcilingUndispatched` and spawns a re-dispatch, which reaches this check and gets certified purely on status, producing a full reconcile tile set for a run whose seeding domain was never planned. That defeats the `prepare.rs` withhold the moment anything else moves the run to `reconciling`.

```suggestion
        if !planning_proven {
            return Err(PrepareReconcileDispatchError::PlanningUnproven(run_id));
        }
```

If pre-protocol reconciling runs genuinely need adopting without a planning stamp, that reads more like a case for an explicit one-time operator action through `--allow-incomplete`, not a bypass the automatic driver also takes on every tick.

rust/cohort-seeder/src/store/completion.rs
```diff
@@ -0,0 +1,881 @@
...
+/// participation is never shortcut this way.
+pub async fn mark_run_observed_unreconcilable(
+    pool: &PgPool,
+    run_id: RunId,
+) -> Result<(), CompletionStoreError> {
+    let updated = sqlx::query_scalar::<_, RunId>(
+        r#"
+        UPDATE cohort_backfill_runs
+        SET reconcile_observed_at = now(), updated_at = now()
+        WHERE id = $1 AND status = 'reconciling'
```
gustavohstrassburger: `suggestion`: this is one of three writes in the file (along with `record` and `revert`) that skip the `backfill_kind = 'behavioral'` recheck every other completion query performs. The module doc asserts this filter is universal ("a `person_property` run id ... can never enter this protocol"), and this write is one of the exceptions.

The reachable path is the CLI, which now mutates by default: an operator-supplied run UUID goes through `load_reconcile_run`, which checks status and cross-team participation but not `backfill_kind`. A mistyped `person_property` run UUID (indistinguishable by eye from a behavioral one in the same table) that happens to already be `reconciling` reaches this write with no epoch fence at all, handing a wrong-kind run to Django's finalizer to terminalize as superseded.

```suggestion
        WHERE id = $1 AND backfill_kind = 'behavioral' AND status = 'reconciling'
```

Worth the same addition on `record` and `revert`, and maybe rejecting a wrong-kind UUID up front in `load_reconcile_run` so the CLI fails fast with a clear message instead of failing deep in the claim path.

rust/cohort-seeder/src/app/completion.rs
```diff
@@ -0,0 +1,517 @@
...
+    fn spawn_dispatch(&self, run_id: RunId, kind: DispatchKind) -> bool {
+        let Ok(permit) = Arc::clone(&self.dispatch_slots).try_acquire_owned() else {
+            return false;
+        };
+        if !lock_in_flight(&self.in_flight).insert(run_id) {
+            return false;
+        }
+        let context = self.context.clone();
+        let guard = InFlightGuard::enter(Arc::clone(&self.in_flight), run_id);
+        tokio::spawn(async move {
```
gustavohstrassburger: `suggestion`: this drops the `JoinHandle`, so nothing ever observes the task's outcome. `InFlightGuard`'s Drop keeps the in-flight set and semaphore permit correct across a panic, so bookkeeping self-heals, but a panicking dispatch produces no log line and no metric. It looks identical to a dispatch that simply hasn't started, so a deterministic panic (a malformed `marker_watch` row, say) becomes a silent hot loop: the run re-spawns and re-panics every tick with nothing in the metrics to show it. `app/execute.rs::record_task_result` already handles this for chunk tasks by matching on `Result<ChunkOutcome, JoinError>`.

```suggestion
        let handle = tokio::spawn(async move {
            let _permit = permit;
            let _guard = guard;
            run_dispatch(&context, run_id, kind).await;
        });
        tokio::spawn(async move {
            if let Err(error) = handle.await {
                counter!(RECONCILE_DISPATCHES, "outcome" => "panicked").increment(1);
                warn!(error = %error, run_id = ?run_id, "reconcile dispatch task panicked");
            }
        });
```

rust/cohort-seeder/src/bin/reconcile_dispatch.rs
```diff
@@ -24,7 +31,11 @@ const PARTITION_VERIFY_TIMEOUT: Duration = Duration::from_secs(10);
 #[derive(Debug, Parser)]
 #[command(
     name = "reconcile_dispatch",
-    about = "Dispatch partition-targeted reconcile snapshots for one behavioral cohort backfill run"
+    about = "Dispatch partition-targeted reconcile snapshots for one behavioral cohort backfill run",
```
gustavohstrassburger, line 34: `suggestion`: the bare invocation of this CLI used to be print-only; now it transitions the run, produces tiles, and persists the dispatch record, while `--allow-incomplete` is the print-only path. That inverts the usual convention where the destructive mode needs the extra flag. `--confirm-register-backfilled` being required does force a deliberate invocation, but it's an attestation about data lineage, not consent to mutate, and combined with the `backfill_kind` gap flagged elsewhere, a pasted wrong run UUID becomes a real state change instead of a harmless printout.

Worth considering an explicit `--dry-run` that runs preparation and prints the plan without claiming. Your call whether the existing required attestation is enough of a gate on its own.

gustavohstrassburger (approved):
Careful protocol work overall. One concurrency gap inline (record() isn't epoch-fenced like its sibling revert()), plus a couple more inline and one on the shared Kafka pacer that couldn't map to this diff's hunks. Stamped to unblock.

PR #73739, opened 2026-07-26
feat(cohorts): seeder reconcile marker watcher and observation pass

matheus-vb:
## Problem

#73738 dispatches reconcile and records the produce HWMs, but nothing observes the `reconcile_complete` markers or writes per cohort outcomes, so a run never reaches `reconcile_observed_at` and the finalizer never fires. This PR adds the consumer half of the completion protocol: watch the markers, check seed group liveness, settle each run, and record its outcomes. Stacked on #73738; dark by default behind an independent gate.

## Changes

- cohort-core: `ReconcileCompleteMarker` is hoisted verbatim from `cohort-stream-processor` into `cohort_core::seed::reconcile` so the processor producer and the seeder watcher agree on the exact bytes. The processor now re-exports it and the golden wire tests move with it.
- Domain (`domain/ledger.rs`, `completion.rs`): `MarkerLedger` is a pure per run fold; `observe` routes each marker to `Novel`, `Duplicate`, `ForeignRun`, `ForeignTeam`, or `UnknownCohort`, and `settle` consumes a non clonable `SettleProof` to return `AllComplete`, `Partial`, or `NoMarkers`. `WatchPositions` and `ObservationEnds` gain the merge and watermark helpers the watcher needs.
- Kafka: `SeedGroupOffsetReader` queries the seed processor group's committed offsets as the liveness signal, read only and mirroring the personhog leader warming pattern so it can't disturb that group's durable progress. `MarkerWatcher` is an assign only consumer with a throwaway group that watches all membership partitions, surfaces retention truncation as a typed error, and cheaply skips person keyed rows via a `':'` key probe before parsing.
- App: `observe_run` early exits when all participations are complete, otherwise checks liveness, captures the end watermarks, and settles with INV-1 hash attribution (a diverged or deleted cohort becomes a terminal partial supersede, a still matching one becomes a retryable shortfall); `reconcile_observed_at` is always the last write. `MarkerWatchTask` is a second global lifecycle component folding markers into per run ledgers and flushing bits and positions on interval, batch cap, or a cohort completing. `CompletionDriver` is split into independent dispatch and observe arms, each gated separately.
- Store: `persist_observation_ends` is rewritten to patch only the `ends` key via `jsonb_set`, so it can't clobber the positions the watch task flushes.
- Config: `seeder_reconcile_observer_enabled` (default false) plus `kafka_seed_consumer_group`, the offsets timeout, and the persist interval and batch cap. New observation metrics added, and the crate picks up `async-trait` for the observe and watch seams.

Domain unit tests cover the ledger fold routing and order independence and the settle verdict discrimination. `tests/pg_completion.rs` is extended for the observation writes (OR merge round trip, fenced writes). App tests use fakes for the committed offsets, topic offsets, and marker stream to exercise the early exit, liveness hold, ends captured once, zero markers, and the hash attribution split. No broker integration test; the kafka plumbing mirrors proven prior art and the seams carry the logic. Run with `cargo test -p cohort-seeder -p cohort-core`.


rust/cohort-seeder/src/app/watch.rs
```diff
@@ -0,0 +1,973 @@
...
+        for run_id in run_ids {
+            self.flush_run(run_id).await;
+            // One round trip per run, all of it before the runner loop's own heartbeat: a large
+            // watched set must slow flushes down, not spend the whole liveness budget.
+            if let Some(handle) = &self.heartbeat {
+                handle.report_healthy();
+            }
+        }
+        self.messages_since_flush = 0;
+        self.positions_advanced = false;
```
gustavohstrassburger: `suggestion`: this clears unconditionally, even when a run's persist failed transiently this cycle. Say a run's coverage advances to its captured end watermark (`positions_advanced` goes true), the next flush hits a transient DB error on a run whose bits are already clean (`has_dirty() == false`), and this line still fires. On an idle topic (the steady state once a reconcile drain finishes), nothing sets `positions_advanced` again, so `flush_run`'s guard (`!has_dirty() && !positions_advanced`) skips that run on every future flush. Its positions never reach the DB, `caught_up` never mints a proof, and the run sits in `MarkerLagging` until a process restart happens to re-seek it.

```suggestion
        self.positions_advanced = !all_persisted;
```

That needs `flush_run` to report whether the persist landed (return `bool`, `true` on success/fence-loss/no-op, `false` only on a transient error) and the loop above to fold those into `all_persisted` before this line. Worth a test alongside `flush_persists_dirty_bits_then_clears_them`: script `FakeFlush` to fail once, assert a second `flush()` with no intervening `ingest` still issues a persist call.

rust/cohort-seeder/src/kafka/markers.rs
```diff
@@ -0,0 +1,286 @@
...
+            next_offset,
+            marker,
+        })
+    }
+}
+
+/// Build the assignment TPL for `start`, rejecting any partition whose start fell out of retention.
+/// `start` iterates in partition order, so the truncation reported is the lowest offending partition
+/// rather than whichever watermark call happened to return first.
+fn build_assignment(
```
gustavohstrassburger, line 149: `suggestion`: this is the truncation detector and it has no test, even though this file is the one explicitly exempted from broker integration coverage. The comparison being tested is `next.get() < low` (a few lines down) — if that ever flips to `<=`, a start exactly at the low watermark (the normal case right after a retention sweep) gets misreported as truncated, and every run on that partition gets dropped and tombstoned until re-dispatch. It's a pure function over `(topic, WatchPositions, HashMap)`, so it's cheap to pin down directly:

```rust
#[test]
fn a_start_below_the_low_watermark_is_truncation_and_the_boundary_is_not() {
    let partition = MembershipPartition::new(0);
    let lows = HashMap::from([(partition, 20_i64)]);

    let mut below = WatchPositions::new();
    below.insert(partition, NextOffset::from_high_watermark(5));
    assert!(matches!(
        build_assignment("membership", &below, &lows),
        Err(WatchError::Truncated { partition: 0, requested: 5, low: 20 })
    ));

    let mut at_low = WatchPositions::new();
    at_low.insert(partition, NextOffset::from_high_watermark(20));
    let tpl = build_assignment("membership", &at_low, &lows).unwrap();
    assert_eq!(
        tpl.find_partition("membership", 0).map(|e| e.offset()),
        Some(Offset::Offset(20))
    );
}
```

rust/cohort-seeder/src/domain/completion.rs
```diff
@@ -312,6 +312,29 @@ impl WatchPositions {
...
+    /// Advance a partition's position to the later of its current and `offset`. The watcher reads a
+    /// single global consumer, so positions must never regress even when a re-assignment re-reads an
+    /// earlier offset; taking the max keeps the persisted resume state monotone.
+    ///
+    /// A partition this position set does not already name is ignored rather than inserted. Such a
+    /// partition appeared after the dispatch captured its start offsets, so the run never read it
+    /// from the beginning; inserting it would claim coverage of everything below `offset`. Leaving
+    /// it absent keeps [`ObservationEnds::caught_up`] fail-closed — the run holds until a
+    /// re-dispatch recaptures a start position for it.
+    pub fn advance(&mut self, partition: MembershipPartition, offset: NextOffset) {
```
gustavohstrassburger, line 331: `suggestion`: neither the monotonicity guard (`offset > *current`) nor the fail-closed guard (ignoring an untracked partition) has a direct test. Deleting either one leaves every existing test passing — a re-assignment reading an earlier offset would silently rewind persisted coverage, and inserting an unknown partition would let `caught_up` mint a proof against a range the run never actually read.

```rust
#[test]
fn advance_is_monotone_and_ignores_partitions_it_does_not_track() {
    let tracked = MembershipPartition::new(0);
    let untracked = MembershipPartition::new(1);
    let mut positions = WatchPositions::new();
    positions.insert(tracked, NextOffset::from_high_watermark(10));

    positions.advance(tracked, NextOffset::from_high_watermark(20));
    assert_eq!(positions.get(tracked), Some(NextOffset::from_high_watermark(20)));

    positions.advance(tracked, NextOffset::from_high_watermark(5));
    assert_eq!(
        positions.get(tracked),
        Some(NextOffset::from_high_watermark(20)),
        "advance must never regress a position"
    );

    positions.advance(untracked, NextOffset::from_high_watermark(99));
    assert_eq!(positions.get(untracked), None);
}
```

rust/cohort-seeder/src/store/completion.rs
```diff
@@ -726,6 +735,9 @@ pub async fn runs_with_all_chunks_confirmed(
 pub struct DiscoveredCompletion {
     pub run_id: RunId,
     pub team_id: TeamId,
+    /// When the run was created. The observer uses it to age the oldest not-yet-observed run for the
+    /// stalled-observation pager gauge.
+    pub created_at: DateTime<Utc>,
```
gustavohstrassburger: `suggestion`: this field is dead, and its own doc comment says the opposite of what the code does. Both `track_oldest` call sites in `app/completion.rs` explicitly age from the dispatch epoch or first-sighting time and name `created_at` as the thing they deliberately avoid ("Age from the phase, not `created_at`"). It costs a column on every discovery query for nothing, and the comment will mislead the next person into thinking it's the gauge's source. Worth dropping from here, `completion_columns!`, `CompletionRunRow`, and `classify_row` — or if it's groundwork for a follow-up, land it with that PR.

rust/cohort-seeder/src/app/watch.rs
```diff
@@ -0,0 +1,973 @@
...
+    async fn persist(
+        &self,
+        run_id: RunId,
+        epoch: DispatchEpoch,
+        bits: &[(CohortId, PartitionBitmap)],
+        positions: &WatchPositions,
+    ) -> Result<(), CompletionStoreError>;
+}
+
+/// The real flush: PR-B's epoch-fenced OR-merge of bits plus the watcher positions.
```
gustavohstrassburger: `blocking`: this and five other doc comments in the diff reference the originating PR by shorthand ("PR-B's fenced writes", "PR-C's dedicated task" in `app/observe.rs:13`, `domain/ledger.rs:3`, `observability/metrics.rs:44`). These stop meaning anything the moment the stack merges, and the substance is already in the same sentence — dropping the label loses nothing:

```suggestion
/// The real flush: the epoch-fenced OR-merge of bits plus the watcher positions.
```

rust/cohort-seeder/src/app/watch.rs
```diff
@@ -0,0 +1,973 @@
...
+        }
+        for run in self.runs.values_mut() {
+            run.positions.advance(item.partition, item.next_offset);
+        }
+        self.positions_advanced = true;
+        self.messages_since_flush += 1;
+        if let Some(marker) = item.marker {
+            let (fold, completed) = match self.runs.get_mut(&marker.run_id) {
+                Some(run) => (
+                    run.ledger.observe(&marker),
```
gustavohstrassburger, line 232: `suggestion`: the `None` arm just below reports `MarkerFold::ForeignRun` for "no watched run names this run id at all," which is a different thing than what `ForeignRun` is documented to mean ("this run's team, but a different run"). Since the watcher tails the whole membership topic, the unwatched case will dominate the `seeder_reconcile_markers_observed_total{fold}` metric by orders of magnitude, which makes the label useless for the distinction it exists to draw. A separate `Unwatched` variant would keep it meaningful.

gustavohstrassburger (approved):
Solid implementation of a genuinely hard completion protocol. An issue inline (a flush-failure edge case that can stall a run's position tracking on an idle topic), plus a couple of test-coverage gaps and some cleanup. Approved even with the issue, since this is not fully running in prod yet.

PR #73749, opened 2026-07-26
fix(mcp): classify exec dispatcher rejections as validation, not internal

mcoll-posthog:
## Problem

The single-exec `exec` dispatcher throws a plain `Error` for every agent-recoverable command failure — an unknown inner tool, an unknown command verb, malformed JSON input, a scope-gated tool, an unknown `learn` topic, or a usage error. Because none of those are a typed error class, they fall through `resolveToolErrorClassification` to the catch-all `internal` bucket. Three things follow from that:

- **They inflate the `internal` error rate.** Ops alerting keys on `mcp_tool_errors_total{error_type=~"...internal..."}`, so ordinary agent mistakes read as server faults — and a single agent retrying a bad command in a tight loop can masquerade as an outage.
- **They're undebuggable from analytics.** `resolveSafeErrorMessage` only captures messages from an allowlist of typed errors, so these arrive with a blank `$mcp_error_message`.
- **They spam error tracking.** `handleToolError` runs `captureException` on them, minting a fresh `exec`-fingerprinted error-tracking issue for each slip-up.

This surfaced while investigating an `internal` error signature on the `exec` dispatcher (~0 ms duration, empty message) that spanned many tenants and harnesses — see the [Slack thread](https://posthog.slack.com/archives/C0BKT3R76RK/p1785083202535329?thread_ts=1785083202.535329&cid=C0BKT3R76RK) for context.

## Changes

- Add `ExecCommandError` (in `lib/errors.ts`) carrying a value-free `reason` enum (`unknown_tool`, `unknown_command`, `invalid_json`, `missing_scope`, `deprecated_tool`, `usage`, `invalid_regex`, `unknown_learn_topic`).
- Throw it from the recoverable branches of the `exec` dispatcher instead of a plain `Error`. The agent-facing messages are unchanged.
- Classify `ExecCommandError` as `validation` (not `internal`), and stamp a value-free `$mcp_error_message` derived from `reason` — never the raw message, which can echo a caller tool name or a JSON-parser fragment.
- Short-circuit exception capture for it in `handleToolError`, like the other recoverable branches.

Genuine unexpected throws inside the dispatcher stay plain `Error`s and remain `internal`, so that bucket regains its meaning as "real server fault."

**Why:** an alert kept re-firing on the dispatcher's `internal` error rate; the errors turned out to be agent-side command mistakes miscategorized as server faults, with no captured message to debug from.

Added unit tests in `tests/hono/tool-executor-metrics.test.ts` asserting a dispatcher rejection (unknown tool, unknown command verb, malformed JSON) is classified as `validation` and carries a value-free `$mcp_error_message` — the regression being that these previously landed in the `internal` bucket with a blank message.

Ran locally:

- `test:hono tests/hono/tool-executor-metrics.test.ts` — 26 passed (incl. the 4 new/updated cases).
- `vitest run tests/unit/exec.test.ts` — 105 passed (dispatcher error messages unchanged; `ExecCommandError extends Error` so existing `.rejects.toThrow('message')` assertions still hold).

`pnpm --filter=@posthog/mcp run typecheck` fails in this environment only on pre-existing `@posthog/quill` / `@posthog/quill-charts` module-resolution errors under `src/ui-apps/` (workspace packages not built here); none of those files are touched by this PR.

## 🤖 Agent context

Investigated from Slack by the PostHog Slack app (Claude). The dispatcher `internal` signature was traced to plain-`Error` throws in `services/mcp/src/tools/exec.ts` being bucketed by `resolveToolErrorClassification` as `internal`, and to `resolveSafeErrorMessage` / `handleToolError` treating them as unexpected faults. Chose to add a dedicated typed error and reuse the existing `validation` bucket rather than introduce a new `$mcp_error_type` enum value, to avoid a taxonomy/dashboard migration. `$mcp_error_message` intentionally carries only the `reason` enum to stay value-free. No repo skills were required for these files.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BKT3R76RK/p1785083202535329?thread_ts=1785083202.535329&cid=C0BKT3R76RK)*

### DIFF
--- a/services/mcp/src/hono/tool-executor.ts
+++ b/services/mcp/src/hono/tool-executor.ts
@@ -331,12 +331,12 @@ export class ToolExecutor {
             return response
         } catch (error: unknown) {
             const metricTool = execToolName()
+            const classification = classifyToolError(error, metricTool)
             if (!execMetrics.innerToolName) {
                 // Match the inner-tool path, which labels rejected input `validation_error`.
-                const status = error instanceof ExecCommandError ? 'validation_error' : 'error'
+                const status = classification.errorType === 'validation' ? 'validation_error' : 'error'
                 toolCallsTotal.inc({ tool: 'exec', status })
             }
-            const classification = classifyToolError(error, metricTool)
 
             void trackToolCall(
                 metricTool,
--- a/services/mcp/src/hono/tool-executor.ts
+++ b/services/mcp/src/hono/tool-executor.ts
@@ -7,6 +7,7 @@ import {
     type ToolResultPayload,
 } from '@/lib/build-tool-result'
 import {
+    ExecCommandError,
     handleToolError,
     MissingOrganizationContextError,
     MissingProjectContextError,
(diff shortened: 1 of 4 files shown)


Twixes (approved):
I think we should merge, should help the agents!

PR #73750, opened 2026-07-26
feat(growth): score lab - versioned LLM classifiers with an in-product lab

MattBro:
## Problem

The signup-enrichment pipeline archives every provider company payload verbatim, but the "AI-pilled company?" classifier that consumes this kind of data still lives in a third-party tool, where prompt iterations are unversioned and results aren't reproducible against our own data. That tool is being wound down, so the capability needs a home in the pipeline.

This ports it as a "score lab": rails are code and PR-gated, brains (prompt, model, input selection, output shape) are versioned config rows the label owner edits in the product without a deploy. It's also the first component of the v-next ICP score.

> [!NOTE]
> Shadow means shadow: nothing is written to group properties, person properties, or events, and nothing runs on a schedule. Zero LLM calls happen until someone runs a dry run or the batch command. Scheduling the batch runner, and the alert that fires on absence of output, are a follow-on PR stacked on this one. Promotion to a group property comes later still.

### Data model (`products/growth/backend/models.py`, migrations 0005-0006)

- `EnrichmentPromptConfig` - a versioned classifier definition: name, version, prompt_text, model, input_fields (dotted paths into the archived payload), input_query (HogQL), output_fields (the output schema), is_active
- `EnrichmentLabelResult` - one verdict per (org, label, prompt version, fetch). Keying on the fetch means a re-enriched org recomputes under the same version instead of freezing. Each row stamps `prompt_hash` (a hash of the behavior-defining config fields), the served gateway model and token usage, and an `inputs` snapshot of exactly what the classifier saw, so a verdict stays reproducible even though the signup domain it derives from drifts as org membership changes
- Versions are server-assigned (`v{n}`, derived from the highest existing suffix under a row lock, never a row count, so a deleted version's string is never reused for a different prompt)
- Both models are excluded from the IDOR coverage check: they carry no `team_id` because they are instance-global and org-scoped rather than team data, and access is gated on `IsStaffUser`

### Classifier runner (`enrichment/labels.py`, `enrichment/lab.py`, `enrichment/input_query.py`)

- Calls the LLM gateway (`get_llm_client(product="growth")`) and parses exactly the keys the config's `output_fields` declares. `output_fields` is the sole output contract: a label name is a human name and is never used as a data key, so renaming a label cannot change what a classifier computes or invalidate stored verdicts
- Prompt inputs are either selected payload fields or the rows of a HogQL query frozen into the version. Queries are parsed with `parse_select` (never raw SQL), bounded by pushing the row limit into the AST, and executed through `execute_hogql_query` under `tags_context` with `LimitContext.POSTHOG_AI`
- **The batch runner supports the archived-payload path only.** It refuses to run a query-mode config rather than classifying it against an empty input dict. Wiring the query path through the batch runner is deliberate future work
- Email addresses are reduced to their domain at any depth before they reach the model or the stored snapshot, and `input_fields` is restricted to a curated allow-list served by `GET /input_fields/`, because every selected value reaches the LLM and is then stored indefinitely
- A missing or empty payload, or one where no configured field resolves, returns `unknown` without calling the LLM, so absent data is never a confident `false`
- Bounded tenacity retry with jitter, and JSON extracted via the shared `posthog.llm.semantic_enrichment.extract_json_object`, which already handles the gateway's fenced-response quirk

### In-product score lab (staff only)

Scene at `/score_lab/<label>`, registered from the growth manifest as an `instanceLevel` scene:

- Version rail with the active badge, and activate-with-confirmation
- Prompt editor (monaco), model select fed from the gateway's live model list, input mode toggle (payload fields or HogQL query), and an output-schema editor
- **Run** streams verdicts row by row into the results table as each LLM call completes, against the editor's current unsaved state. Persists nothing, so experiments are free
- **Save as new version** persists exactly the validated editor bytes under a server-assigned version

Backed by a staff-only DRF API (`/api/growth_score_lab/*`) following the cohorts staff-tools pattern: `scope_object="INTERNAL"`, `permission_classes=[IsAuthenticated, IsStaffUser]`, registered unscoped at the API root because prompt configs are instance-global rather than team data. `run` streams `application/x-ndjson` through an async generator and `posthog.api.streaming.streaming_response`, and carries its own `UserRateThrottle`: the global throttles are `PersonalApiKeyRateThrottle` subclasses that pass straight through for a session-authenticated request, so this endpoint would otherwise have no limit on an action that costs up to `MAX_SAMPLE_SIZE` completions per call.

Django admin keeps only a read-only results browser. Config authoring lives entirely in the scene.

### Commands

- `manage.py enrichment_label_dry_run --label ai_pilled --sample 50 [--prompt-file X] [--compare-version v]` - classifies a sample without persisting, for terminal iteration and larger samples than the scene's cap. Display columns are rendered from the config's `output_fields`, and a key the schema doesn't declare renders as `-` rather than as a default that reads like data
- `manage.py enrichment_label_batch --label ai_pilled [--limit N] [--workers K]` - the shadow run. Idempotent (skips computed (org, label, version, fetch)) and resumable, with an end-of-run summary, `capture_exception` on every LLM or parse failure, and a non-zero exit when any item failed. Coverage check:

```sql
SELECT count(*) FROM posthog_organizationenrichmentfetch f
LEFT JOIN posthog_enrichmentlabelresult r
  ON r.fetch_id = f.id AND r.label_name = 'ai_pilled' AND r.prompt_version = '<active version>'
WHERE r.id IS NULL;
```

### Seed

Migration 0006 seeds `ai-pilled-clay-v1` with the label owner's existing prompt captured verbatim, so the port is comparable against the run it replaces. Iterating a v2 on richer inputs is the label owner's job via the lab, not this PR's.

### Why not reuse `LLMPrompt`

`products/ai_observability/backend/models/llm_prompt.py` is conceptually similar but team-scoped (CASCADE FK to Team, per-team uniqueness) and customer-facing, and it versions only the prompt, not the model/inputs/output tuple this lab's provenance hashes. Parking internal instance-global config inside a customer product's tenant boundary would leak an internal surface into their data model and couple these provenance guards to their migrations.

**Automated** - 81 backend tests across `api/test_score_lab.py` and `tests/test_enrichment_labels.py`, plus 18 Jest tests:

- staff-only: every endpoint 403s for an authenticated non-staff user, including the streaming one and `rename`
- missing, empty, and not-found payloads return `unknown` and never call the LLM
- results are stamped with version, content hash, model, exact fetch, and the rendered input snapshot
- batch re-run creates no rows and makes zero LLM calls (broken idempotency would mean double spend)
- the batch runner refuses a query-mode config before spending anything
- a newer fetch and a version bump both recompute with old rows intact
- an unparseable LLM response is captured and counted without persisting or aborting the run
- a rename propagates to every sibling version and every stamped result, and a rename mid-run stamps the live label rather than the retired one
- renaming a label leaves both the admin verdict column and the dry-run comparison columns resolving, with the label name and the boolean output key deliberately different
- email local parts never reach the rendered prompt, at any nesting depth
- runs persist nothing (CLI and API), and the streamed run returns verdicts for an edited, unsaved prompt
- save rejects invalid label slugs; activate flips the active version atomically
- HogQL input queries reject syntax errors and non-SELECT statements, and execution failures surface as 400s rather than 500s
- custom output schemas reject bad keys, types, duplicates, and reserved names

**Live browser verification** (local dev + local LLM gateway, via Playwright): loaded the scene, ran the active version against real archived orgs and watched verdicts stream in, switched to HogQL mode and streamed a classified verdict from a query, and exercised a custom output schema end to end. This pass caught three bugs the test suite structurally cannot see: ClickHouse rejecting untagged HogQL queries, DRF returning 406 on the stream's Accept header, and a monaco mount race.

**Migrations**: `makemigrations --check` clean; `analyze_migration_risk` reports 1 Safe / 1 Needs Review / 0 Blocked, the review item being the single-row `RunPython` seed. 0005 creates two tables (org FK is `db_constraint=False`, off the hot-table lock path) and 0006 is that seed.

## Next steps

- **Stacked follow-on**: schedule the batch runner, and ship it with an alert that fires on absence of output rather than only on errors, since a component that never runs emits zero failures
- Warehouse dashboards over stored verdicts. The results table does not reach the internal warehouse on its own: that Postgres source is per-table opt-in with auto-sync of new schemas off, so `growth_enrichmentlabelresult` has to be enabled on it once this deploys, plus `growth_enrichmentpromptconfig` if dashboards join to the config instead of reading the denormalized `label_name`/`prompt_version` columns. Enabling a schema is config, with two gotchas carried over from the enrichment-fetch table's sync: `created_at` is `auto_now_add`, so the incremental sync needs a lookback window or rows committed after a sync captured a higher max are skipped permanently; and a missing JSON key stringifies non-empty in HogQL, so `toString(output.<key>) != ''` is not a presence test, which matters now that `output_fields` lets the key set differ per version
- First full shadow pass, with the label owner adjudicating disagreements into a golden set. There is no model for human ground truth yet, and "is v3 better than v2" is currently answered by reading two columns rather than by a recorded judgement
- Wire the HogQL input path through the batch runner, or drop it if the payload path proves sufficient
- Org evidence dossier: one assembled evidence document per org (archived payload, headcount trajectory, first-party product usage, GitHub) that classifiers read instead of a thin payload
- Promotion to a group property, gated on the golden set

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

**Autonomy:** Human-directed (agent-built)

Built by Claude Code from a written handoff spec, with implementation parallelized across sub-agents (runner, commands, API, scene frontend) against a fixed contract, then integrated, adversarially reviewed, and live-browser-verified by the coordinating session.

Two earlier iterations were removed rather than left to rot. The lab was first built inside Django admin (~322 lines) and replaced by the in-product scene; its admin surfaces were deleted. A version-immutability guard on `EnrichmentPromptConfig` (`FROZEN_FIELDS` plus overridden `save()`/`delete()` taking row locks, paired with a mid-run re-check in the batch runner) was removed once review established that no production path could reach it: the save endpoint only ever creates, activate's `save()` touches a field that was never frozen, rename goes through a queryset `update()`, and there is no delete endpoint and no config admin. The invariant it claimed to protect is enforced by the create-only endpoint and a unique constraint. `content_hash` stayed, because that is the part carrying real provenance.

The runner and API shapes are deliberately kept free of org assumptions (config version + input rows + verdict rows) so the classifier core could survive a future re-scoping.


rafaeelaudibert (approved):
This is a lot of code, I'll run stamphog, it'll be better than me.

My suggestion is you simply merge some of the migrations together if possible, that's gonna be easier to merge

MattBro:
> This is a lot of code, I'll run stamphog, it'll be better than me.
> 
> My suggestion is you simply merge some of the migrations together if possible, that's gonna be easier to merge

I'm going to break this up into smaller prs and close this one

MattBro:
Superseded by the sliced approach: classifier + verdict store merged in #74912, scheduling in #77786, and the in-product lab UI continues in #75754. Branch kept for reference.

PR #73751, opened 2026-07-26
chore(ci): deploy signals worker on canonical scout skill changes

andrewm4894:
## Problem

The signals scout coordinator and runner both live in the video-export Temporal worker, and the canonical `signals-scout-*` fleet ships inside that worker's image. The coordinator reads the fleet off local disk at runtime (`products/signals/backend/scout_harness/lazy_seed.py`).

That worker's deploy change-filter matched `^products/signals/backend/` but not `^products/signals/skills/`. So a commit that only adds or edits a scout SKILL.md built an image that nothing ever deployed.

Every step of the chain stays quiet, which is what makes this nasty:

```mermaid
flowchart TD
    A[skills-only commit on master] --> B[image builds and pushes to ECR]
    B --> C{video-export change filter}
    C -->|no match| D[no commit_state_update dispatch]
    D --> E[charts state file stays pinned]
    E --> F[Argo healthy, nothing to sync]
    F --> G[worker disk keeps the old fleet]
    G --> H[sync finds nothing to create, logs no error]
    class A phYellow;
    class C phRed;
    class B,D,E,F,G phGray;
    class H phRed;
```

After:

```mermaid
flowchart TD
    A[skills-only commit on master] --> B[image builds and pushes to ECR]
    B --> C{video-export change filter}
    C -->|matches skills path| D[commit_state_update dispatched]
    D --> E[charts state file moves to new sha]
    E --> F[Argo syncs the worker]
    F --> G[new fleet on disk, scout seeds next tick]
    class A phYellow;
    class C phBlue;
    class B,D,E,F phGray;
    class G phYellow;
```

Skills-only changes did eventually ship, by riding along on the next `products/signals/backend/` commit. That is why this looked intermittent rather than broken, and why a new scout can appear days late or, over a quiet weekend, not at all. Over the 9 days before this change, 5 of 13 commits touching `products/signals/skills/` would not have triggered a deploy.

I found it after a newly merged canonical scout still had no skill row in a project roughly 19 hours after merge, while the coordinator ticked 48 times with zero sync errors.

> [!NOTE]
> Widening a path filter is safe for PRs that have not rebased. It only changes which deploys get dispatched and adds no prerequisite an older branch could be missing.

## Changes

One path added to the video-export worker's change filter, plus a comment so the next person does not read it as redundant with the backend path and prune it.

```diff
-  grep -qE '...|^products/signals/backend/|^products/review_hog/backend/|...'
+  grep -qE '...|^products/signals/backend/|^products/signals/skills/|^products/review_hog/backend/|...'
```

I checked the neighbouring workers for the same class of bug and deliberately left them alone. Three looked affected at first and were not:

| Candidate | Why it is fine |
| --- | --- |
| `products/review_hog/skills/` | Same lazy-seed pattern, but seeding is only called from `products/review_hog/backend/api/*.py`, so it runs on web, which deploys every commit. |
| `products/warehouse_sources/skills/` | `ai_builder.py` loads the skill at runtime, but only via `draft_manifest_sync` from a DRF view. Also web. |
| Every `products/*/skills/` on tasks-agent | `LocalSkillsCache` walks all of them, but only in the local image-build path (`_prepare_local_modal_build_context`, and the docker sandbox builder). In production the bundle is baked in by `cd-sandbox-base-image.yml`, which already filters on `products/*/skills/**`. |

I could not test the deploy end to end. Triggering a production deploy is not something I can do from a PR, and the behaviour only shows on the next skills-only merge to master. What I did verify:

- Replayed the new filter against real commits, extracting the regex out of the workflow file rather than retyping it. The 5 skills-only commits that were silently skipped now match, the last commit that did deploy (`966941bc`) still matches, and an unrelated replay commit still does not.
- `hogli lint:workflows`: 6 of 6 checks pass across 105 workflows.
- `hogli ci:preflight --strict`: 0 failures.
- `hogli format:yaml` (oxfmt): no changes, so the edit is already canonical.
- The workflow still parses as YAML, and `bash -n` accepts the step's script.

No automated tests added. The filter is a shell regex inside a workflow step and there is no existing harness for asserting on these path filters. Building one for a single regex would be more surface area than the fix, and it would not have caught this: the bug was a missing path, not a malformed one.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Nothing user-facing changes here.

## 🤖 Agent context

Andy asked why a canonical scout he had merged was not showing up in his scout list, and I (Claude, via Claude Code) traced it. Session: https://claude.ai/code/session_011yWwbtAMb1Wm6YHsp6ZLak

Skills invoked: `/authoring-ci-workflows`.

The diagnosis went through a few wrong turns worth knowing about, since they shape how narrow this fix is. The first hypothesis was that canonical sync itself had broken, which the data killed: sync was writing normally across the fleet on the day in question, and an earlier new scout had reached 420 existing teams incrementally. The second was an Argo weekend freeze, which was also wrong, Argo has auto-sync with selfHeal on and was behaving correctly. The state file in PostHog/charts was simply never asked to move. Reading the image tags per service is what settled it: web was on master tip while the video-export worker sat on the commit immediately before the scout, with no pod restarts in between.

I originally scoped this as four filter edits across three workers and wrote it up that way. Verifying each caller before editing knocked three of them out, per the table above. Worth flagging for the reviewer, because "add the path everywhere skills are read" is the intuitive fix and it would have added real deploy churn to tasks-agent for no benefit.

The deeper fix is not this PR. The canonical fleet is data that the sync already versions per team, so shipping it inside a container image is what couples scout content to worker deploys in the first place. Seeding from a build artifact or the skills store would remove this whole class of bug. That is tracked separately, along with the observability gap that let this hide: a no-op sync logs nothing, so a healthy sync and a sync running against a stale fleet look identical from outside.

---
_Generated by [Claude Code](https://claude.ai/code/session_011yWwbtAMb1Wm6YHsp6ZLak)_

### DIFF
--- a/.github/workflows/container-images-cd.yml
+++ b/.github/workflows/container-images-cd.yml
@@ -475,7 +475,7 @@ jobs:
             - name: Check for changes that affect video export temporal worker
               id: check_changes_video_export_temporal_worker
               run: |
-                  if git diff --name-only HEAD^ HEAD | grep -qE '^posthog/temporal/exports_video|^posthog/tasks/exports/video_exporter.py|^products/signals/backend/|^products/review_hog/backend/|^products/business_knowledge/backend/|^products/warehouse_sources/backend/temporal/data_imports/workflow_activities/emit_signals|^products/signals/backend/emission/|^nodejs/src/scripts/record-replay|^ee/hogai/session_summaries|^posthog/temporal/common|^posthog/management/commands/start_temporal_worker.py$|^pyproject.toml$|^bin/temporal-django-worker$|^posthog/clickhouse/migrations/'; then
+                  if git diff --name-only HEAD^ HEAD | grep -qE '^posthog/temporal/exports_video|^posthog/tasks/exports/video_exporter.py|^products/signals/backend/|^products/signals/skills/|^products/review_hog/backend/|^products/business_knowledge/backend/|^products/warehouse_sources/backend/temporal/data_imports/workflow_activities/emit_signals|^products/signals/backend/emission/|^nodejs/src/scripts/record-replay|^ee/hogai/session_summaries|^posthog/temporal/common|^posthog/management/commands/start_temporal_worker.py$|^pyproject.toml$|^bin/temporal-django-worker$|^posthog/clickhouse/migrations/'; then
                       echo "changed=true" >> $GITHUB_OUTPUT
                   else
                       echo "changed=false" >> $GITHUB_OUTPUT


.github/workflows/container-images-cd.yml
```diff
@@ -474,8 +474,14 @@ jobs:
 
             - name: Check for changes that affect video export temporal worker
               id: check_changes_video_export_temporal_worker
+              # `products/signals/skills/` is deliberately here alongside the backend paths: the
+              # canonical `signals-scout-*` fleet ships inside this image and the scout coordinator
+              # reads it off local disk at runtime (scout_harness/lazy_seed.py). Without it a
+              # skills-only commit builds an image nothing ever deploys, so a new canonical scout
+              # silently never reaches existing teams — the sync finds nothing to create and logs no
+              # error. See PostHog/posthog#73685.
```
gantoine: we really don't need this comment

gantoine (approved):

.github/workflows/container-images-cd.yml
andrewm4894: Agreed, dropped it in `ba6f0fa`. You're right on convention too: it ended with a PR reference, and CLAUDE.md says change context belongs in the commit message and PR description, not the code. None of the other paths in that filter carry a comment either.

The diff is now exactly the one added path. Re-verified after the change that the filter still behaves: the previously-skipped skills-only commits match, `966941bc` still matches, unrelated commits still don't.

---

andrewm4894: yeah bots mad for excessive comments - sorry about that - actually have [this scout](https://us.posthog.com/project/2/inbox/scouts/signals-scout-excessive-comments) set up to follow up on excessive comments too (think needs a bit of tweaking though)

PR #73833, opened 2026-07-27
fix(hogql): resolve SQL insight date filters like other insights

thmsobrmlr:
## TL;DR

When you put a date filter like "This month" on a SQL insight, it filtered from "the 1st at whatever time it is right now" instead of "the 1st at midnight", and it never applied an upper bound at all. So SQL insights dropped rows from the first day and included rows from the future. Now they resolve date filters the same way every other insight does. This PR also adds a dismissable in-app banner (behind a feature flag, off by default) that tells users on affected SQL insights why their numbers changed.

## Problem

SQL insights resolve `{filters.dateRange.from}` / `{filters.dateRange.to}` / `{filters}` through their own date parsing in `ReplaceFilters` instead of `QueryDateRange`, with two defects:

- Relative `date_from` values (`mStart`, `-7d`, `wStart`) kept the wall-clock time of day, so e.g. "This month" started at `07-01T15:15:xx` instead of `07-01T00:00:00` — rows from the first day of the range before "now o'clock" were silently dropped, and the cutoff moved with every refresh. The parse also ignored the team's week start day.
- A missing `date_to` (true for every "…until now"-style preset) removed the upper-bound comparison entirely instead of bounding at now, so future-dated rows were included.

Every other insight type resolves the same strings through `QueryDateRange`, which truncates `date_from` and bounds a missing `date_to` at end of today — so a SQL tile and a trends tile on the same dashboard with the same date filter disagreed. Raised via [a support ticket](https://us.posthog.com/project/2/support/tickets/64760) where a warehouse-backed SQL funnel on a dashboard showed fewer conversions than the underlying table.

## Changes

`ReplaceFilters` now resolves date bounds with `QueryDateRange`'s day-interval semantics:

- Relative day-or-coarser `date_from` snaps to start of day, honoring the team's week start day and one pinned "now" per query.
- Relative and date-only `date_to` snap to end of day; a missing `date_to` resolves to end of today when a date filter is in use.
- `{filters}` compares a snapped end-of-day upper bound inclusively (`timestamp <= 23:59:59.999999`), so the whole day is covered; exact bounds (explicit datetimes, sub-day relatives, `explicitDate`) keep the strict `<`, preserving half-open windows like the logs count-ranges bucket ends, which are documented as exclusive drill-in bounds.
- Deliberate exceptions: explicit datetimes are used verbatim; sub-day rolling windows (`-1h`, `-30M`) keep exact bounds and stay open-ended; `explicitDate` disables the end-of-day snapping of `date_to` (like `QueryDateRange`, which consults it only for the upper bound); `date_from="all"` still means a fully unbounded range — no lower bound (not an earliest-event lookup, which wouldn't make sense for warehouse tables) and no upper bound either, since "All time" promises the whole table including future-dated warehouse rows.
- Logs and tracing resolve their exact rolling windows themselves: their runners now pass absolute datetimes from their existing `QueryDateRange(exact_timerange=True)` through the new `QueryDateRange.to_hogql_filters()` instead of raw relative strings, so the logs/traces list keeps matching its count and sparkline (which never went through `{filters}`) on day-level presets like `-7d`.
- `HogQLQueryRunner.apply_dashboard_filters` now carries the dashboard's `explicitDate` along with its date bounds (coerced to bool so the serialized value stays `false`), matching the "date range is one unit" doctrine in `apply_dashboard_filters.py`. The base `QueryRunner` still keeps the insight's flag when the dashboard's is `None`; reconciling that is a follow-up.
- The heatmaps test-accounts subquery passes an explicit next-day-midnight datetime; compared with the strict `<`, its window matches the inclusive days of the main heatmap query (a date-only value would snap to the end of the next day and widen the window).
- Adds a dismissable in-app banner (`SqlInsightDateFilterNotice`, behind the `sql-insight-date-filter-notice` flag, off by default) on affected SQL insight views, so users understand why their results shifted. Folded in from #73842; retire the flag and component a few weeks after the fix ships.

Behavioral note: any SQL insight using a relative or open-ended date range will show different (now correct) numbers on its next refresh, converging over ~a day as day-stale caches expire. Worth a changelog entry.

Automated only (no manual testing):

- `posthog/hogql/test/test_filters.py`: new parameterized tests pin each regression no existing test caught — relative `date_from` snapping to midnight (the reported bug), week-start handling, missing `date_to` resolving to end of today (previously asserted as unbounded), date-only `date_to` covering the whole day, sub-day windows staying rolling/open-ended for both bounds, `explicitDate` bypassing snapping, `date_from="all"` staying fully unbounded (no end-of-today cap), verbatim datetime upper bounds staying strict/exclusive while snapped ones are inclusive, and the skip-to-`true` contract surviving when no date filter is in use.
- `posthog/hogql/test/test_query.py` (snapshots updated for the end-of-day changes), `posthog/hogql_queries/test/test_hogql_query_runner.py` + `ee/hogai/chat_agent/sql/test/test_sql_mixins.py`, `products/web_analytics` heatmaps API + session attribution suites. The logs boundary tests keep their original exclusive-boundary assertions (exact bounds are compared strictly again), plus a new case pinning that a day-level relative preset (`-1d`) keeps its exact wall-clock window on the list query instead of snapping to midnight — the desync no existing test caught.
- `SqlInsightDateFilterNotice.test.ts`: parameterized cases pinning which date-range shapes the banner fires on, matching the backend exemptions — including `explicitDate` paired with a relative preset (where `date_from` still snaps) and bare relative `date_to` values.
- mypy clean on the changed files.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

The SQL docs on posthog.com describing `{filters.dateRange.*}` placeholders may want a note that open-ended ranges now bound at end of today.

## 🤖 Agent context

- Authored with PostHog Code (Claude Code) from a support-ticket investigation; skills invoked: `/writing-tests`, `/analyzing-insights-across-teams` (impact sizing).
- Considered reusing `QueryDateRange` directly inside `ReplaceFilters`, rejected: it requires an interval (SQL queries have none), resolves `date_from="all"` via an earliest-event ClickHouse lookup (wrong for warehouse tables), and would snap explicit datetimes to end of day. Instead the resolver mirrors its day-interval behavior with documented exceptions.
- The sub-day rolling exemption exists because ad-hoc SQL insights over the logs/traces tables use rolling live-view-style windows (`date_from: "-1h"`) through the same `{filters}` placeholder; without it this fix would widen "last hour of logs" to "since the start of the hour" and cap live tails at end of today. The logs and tracing products themselves now resolve their windows before the placeholder (see Changes), so day-level presets stay exact there too.
- Review follow-ups: scoped the inclusive upper bound to snapped end-of-day values only (georgemunyoro — the unconditional `<=` broke the logs count-ranges exclusive bucket contract); kept `date_from="all"` fully unbounded instead of capping at end of today (georgemunyoro); made logs/tracing pass resolved bounds so their list views stay in sync with count/sparkline (georgemunyoro); fixed the banner to fire on `explicitDate` + relative presets since `date_from` still snaps (georgemunyoro); coerced the dashboard `explicitDate` carry to bool to keep the response shape (georgemunyoro). Declined honoring `explicitDate` for `date_from` (greptile P1) since `QueryDateRange.date_from()` ignores the flag too, and diverging would re-introduce the wall-clock bug for dashboards that combine `explicitDate` with relative presets.

---


posthog/hogql/filters.py
```diff
@@ -52,6 +59,87 @@ def __init__(
...
+        """Lower bound of the date range, resolved the way QueryDateRange resolves it for insights:
+        relative values in day-or-coarser units ("mStart", "-7d") snap to the start of the day instead
+        of keeping the current time of day. Sub-day units ("-1h") stay exact rolling windows, explicit
+        datetimes are used verbatim, and "all" means no lower bound."""
+        parsed_date, delta_mapping = self._parse_date_from()
+        if parsed_date is None:
+            return None
+        if delta_mapping is None or SUB_DAY_DELTA_KEYS & delta_mapping.keys():
+            return parsed_date
+        return parsed_date.replace(hour=0, minute=0, second=0, microsecond=0)
```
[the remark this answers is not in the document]
thmsobrmlr: Not applied — this mirrors `QueryDateRange` on purpose. `QueryDateRange.date_from()` never consults `explicitDate`: relative day-or-coarser values always truncate to start of day (`always_truncate=...`); only `date_to()` checks the flag. Honoring `explicitDate` here would create a fresh divergence — a dashboard sending `explicitDate: true` with a relative preset would keep the wall-clock bug on SQL tiles while trends tiles on the same dashboard snap to midnight.

Explicit datetime lower bounds already pass through verbatim (they parse with `delta_mapping=None`), and the UI never pairs `explicitDate: true` with a relative `date_from` (`dateFilterLogic` resets the flag on every date selection).

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

posthog/hogql/filters.py
```diff
@@ -52,6 +59,87 @@ def __init__(
...
+                now=self._current_time(),
+                team_week_start_day=self.team.week_start_day,
+            )
+            if delta_mapping is not None and SUB_DAY_DELTA_KEYS & delta_mapping.keys():
+                return parsed_date
+        if parsed_date is None:
+            return None
+        if explicit_date:
+            return parsed_date
+        return parsed_date.replace(hour=23, minute=59, second=59, microsecond=999999)
```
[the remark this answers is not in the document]
thmsobrmlr: Fixed in 3b2633c — `{filters}` now compares the upper bound with `lessOrEquals`, matching how trends compares against `QueryDateRange`'s `date_to` (`timestamp <= {date_to}` in `trends_query_builder.py`). The heatmaps events subquery keeps its next-day-midnight bound; with `<=` it now matches the main heatmap predicate `timestamp <= {date_to} + interval 1 day` exactly.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

posthog/hogql/filters.py
georgemunyoro, line 92: This snapping desyncs the logs and traces **list** from its own count and sparkline.

`LogsQueryRunner` is the only one of the three logs runners that goes through `{filters}` — `filters=HogQLFilters(dateRange=self.query.dateRange)` at `logs_query_runner.py:669`, placeholder at `:395` — and it's the only precise timestamp bound in that query (the other predicate is day-granular `toStartOfDay(time_bucket)`). Meanwhile `count_query_runner.py` and `sparkline_query_runner.py` call `execute_hogql_query` **without** `filters=`, so their `{filters}` collapses to `true` and their real bounds come from `QueryDateRange(..., exact_timerange=True)` → `always_truncate=False`, i.e. deliberately *not* truncated.

Concretely: open Logs at 15:00 on the "Last 7 days" preset. The frontend passes the raw relative string through (`logsViewerFiltersLogic.ts` `utcDateRange` only converts dayjs-parseable values; `logsViewerFiltersLogic.test.ts:113` pins `date_from: '-7d'` unchanged). Post-change the list starts at 00:00 seven days ago while the sparkline and the header count — which is a sum over the sparkline buckets — start at 15:00 seven days ago. So up to 15h of logs sit at the bottom of the list that the chart above it says don't exist, and the count under-reports the list it labels.

`products/tracing/backend/logic.py` has the same shape (`exact_timerange=True` at `:381`, `filters=HogQLFilters(...)` at `:428`, placeholder at `:290`), though there the count runner does pass `filters=`, so for traces only the sparkline diverges.

The sub-day carve-out protects the `-1h` / `-24h` live tail, which is the case the PR body calls out, but it stops one unit short of the `-3d` / `-7d` presets these products also ship in `DEFAULT_DATE_RANGE_PICKER_OPTIONS`.

posthog/hogql/filters.py
```diff
@@ -52,6 +59,87 @@ def __init__(
...
+
+    def _open_ended_date_to(self) -> Optional[datetime]:
+        """A missing date_to means "up to now" when a date filter is in use — matching how insights
+        resolve the same range instead of leaving the query unbounded (which would include
+        future-dated rows). With no date filter at all, or with a sub-day rolling date_from
+        ("-1h"), the range stays open-ended."""
+        date_range = self.filters.dateRange if self.filters else None
+        if date_range is None or date_range.date_from is None:
+            return None
+        if date_range.date_from != "all":
```
georgemunyoro: `"all"` skips the sub-day check here and falls through to `_current_time()`, which `_resolve_date_to` then snaps to `23:59:59.999999` — so "All time" now caps at end of today.

On master a missing `date_to` set `compare_op_wrapper.skip = True`, so `{filters.dateRange.to}` collapsed to `true` and genuinely returned everything. And unlike the bare `{filters}` placeholder, the `dateRange.from` / `dateRange.to` placeholders have no table restriction, so this reaches warehouse tables — which is where future-dated rows actually live. A SQL insight like `SELECT * FROM renewals WHERE renewal_date <= {filters.dateRange.to}` on the "All time" preset now silently drops every renewal dated after today.

I don't think the `QueryDateRange` parity argument covers this one: there `"all"` resolves `date_from` via `get_earliest_timestamp()`, so the range is never open at the bottom, and "all" means "since the first event" rather than "unbounded". This resolver deliberately keeps `"all"` as *no lower bound* (documented in `_resolve_date_from`), which is the right call for warehouse tables — but it means the upper bound is the only thing left, and capping it changes what the preset promises.

The PR body says `date_from="all"` still means no lower bound and is silent on the upper one, and there's no test covering `"all"` in the new `test_filters.py`. If the cap is intended, worth a test plus a line in the description; if not, an early `if date_range.date_from == "all": return None` here.

posthog/hogql/filters.py
```diff
@@ -142,36 +230,26 @@ def visit_placeholder(self, node):
...
-                    parsed_date = relative_date_parse(dateTo, self.team.timezone_info)
+            date_to = self._resolve_date_to()
+            if date_to is not None:
                 exprs.append(
+                    # Inclusive, like insights compare against QueryDateRange's date_to — the resolved
+                    # end-of-day bound lands on 23:59:59.999999, so a strict `<` would drop rows at
+                    # exactly that microsecond
                     ast.CompareOperation(
-                        op=ast.CompareOperationOp.Lt,
+                        op=ast.CompareOperationOp.LtEq,
```
georgemunyoro: The inclusive bound is well justified for the snapped `23:59:59.999999` case, but it also applies to verbatim explicit datetimes — and that breaks the logs bucket contract.

`count_ranges_query_runner.py` emits half-open buckets (`bucket_end = bucket_start + interval`, and its own WHERE is `timestamp >= {date_from} AND timestamp < {date_to}`), serialized as absolute ISO like `2025-12-15T00:10:00Z`. The API says so explicitly in `products/logs/backend/presentation/views/api.py:493`: *"Bucket end as ISO 8601 timestamp. Exclusive upper bound. Pass back as `dateRange.date_to` to drill in."* Those values take the verbatim branch in `_resolve_date_to`, so with `LtEq` a log landing exactly on a bucket edge is now returned by both adjacent drill-downs, and a drill-down returns one more row than the bucket count that launched it.

The two `_boundary_query` edits in this PR are the same semantics change surfacing — `"2025-12-16 00:00:00Z"` → `"2025-12-15 23:59:59.999999Z"` was needed to keep `boundary-log-dec16-midnight-exact` out. Those assertions weren't wrong; "midnight-exact belongs to the next day" was the contract.

Narrow in practice: `count-ranges` isn't used by the logs frontend (it's an API/MCP path) and a collision needs a timestamp on an exact interval boundary — common for second-granularity buckets, rare for ms data in minute buckets. But the `help_text` is now inaccurate regardless of data. Either scope the `LtEq` to the snapped end-of-day case, or update the bucket contract and its docs in the same PR.

frontend/src/queries/nodes/DataVisualization/Components/SqlInsightDateFilterNotice.tsx
```diff
@@ -0,0 +1,56 @@
...
+    if (!source.query?.includes('{filters')) {
+        return false
+    }
+    const dateRange = source.filters?.dateRange
+    if (!dateRange || (dateRange.date_from == null && dateRange.date_to == null)) {
+        return false
+    }
+    const openEndedWithBoundedStart =
+        dateRange.date_to == null && dateRange.date_from != null && !SUB_DAY_RELATIVE_REGEX.test(dateRange.date_from)
+    if (dateRange.explicitDate) {
```
georgemunyoro, line 34: This branch assumes `explicitDate` disables the snapping, but it only disables half of it: `_resolve_date_from` never reads `explicit_date` (deliberately, per your greptile P1 note — and I agree with the reasoning). So `date_from` still snaps to midnight when `explicitDate` is set, while this returns `openEndedWithBoundedStart`, which is `false` whenever `date_to` is non-null. Net effect: the banner is hidden from a set of users whose numbers do change.

Reachable from the UI rather than just via the API — the "Exact time range" toggle renders unconditionally under `showExplicitDateToggle` (`DateFilter.tsx:375`) and `SqlInsightFilters.tsx:66-70` writes `filters.dateRange.explicitDate`, independent of which preset is selected. So: pick "Last month" on a SQL insight (`values: ['-1mStart', '-1mEnd']`) and flip the toggle → `{date_from: '-1mStart', date_to: '-1mEnd', explicitDate: true}`. `-1mStart` has `delta_mapping = {months: 1, day: 1}` — no sub-day key — so it snaps to `00:00:00` where master returned the 1st at wall-clock time. Results shift, no banner.

The `.test.ts` cases only pair `explicitDate` with an absolute `date_from` (`'2026-07-01'`), so this shape isn't covered. Not a merge blocker since the banner is flagged off, but worth closing before the flag goes live, since suppressing support tickets is the whole job here.

Related: the `date_to` arm only tests `DATE_ONLY_REGEX`, so a bare relative `date_to` like `-3d` (which the backend does now snap to end of day) isn't matched. I couldn't find a UI path that produces that shape — every preset with a `date_to` pairs it with a relative `date_from` the first arm already catches — so it's latent rather than a live gap, but a wider regex would be cheap insurance.

posthog/hogql_queries/hogql_query_runner.py
```diff
@@ -178,6 +178,8 @@ def apply_dashboard_filters(self, dashboard_filter: DashboardFilter):
                 self.query.filters.dateRange = DateRange()
             self.query.filters.dateRange.date_to = dashboard_filter.date_to
             self.query.filters.dateRange.date_from = dashboard_filter.date_from
+            # The date range is one unit — don't pair the dashboard's bounds with the insight's own explicitDate
+            self.query.filters.dateRange.explicitDate = dashboard_filter.explicitDate
```
georgemunyoro: This assigns unconditionally, where `query_runner.py:2486` guards with `if dashboard_filter.explicitDate is not None:`. Since `DashboardFilter.explicitDate` defaults to `None` and is only persisted when a dashboard's exact-time toggle is used, `None` is the common case — so the two paths diverge on almost every dashboard.

Worked example: dashboard filter `{date_from: '-7d', explicitDate: None}`. A Trends tile saved with `explicitDate=True` keeps it, and `QueryDateRange.date_to()` (line 105) skips the end-of-day snap, so its upper bound is now. The SQL tile next to it gets clobbered to `None` and snaps to end of today. Two tiles, one dashboard filter, two upper bounds — which is the divergence this PR is closing.

That said, your version matches the newer doctrine in `apply_dashboard_filters.py:143-149` (*"an override that sets either bound supplies both bounds and explicitDate"*), and pairing an insight's saved `explicitDate` with a dashboard's relative range is genuinely dubious — the flag was recorded for a date range that no longer exists. So I suspect this is intentional and correct, and `query_runner.py:2486` is the stale one. Mostly flagging that leaving both conventions in the tree guarantees someone rediscovers this; a reconciliation or a pointer comment would settle it.

Minor, separate: writing `None` instead of leaving the `DateRange.explicitDate` default flips the API-serialized value from `false` to `null` — that's the three `test_insight_model.py` assertion changes. Frontend readers use `?? false` so nothing breaks, and `tileLogic.tsx:90`'s `!== undefined` now stores `null` rather than skipping, which is cosmetic. Just isn't called out anywhere as a response-shape change.

posthog/hogql/filters.py
thmsobrmlr: Fixed in 24be1e5 — logs and tracing now resolve their windows before the placeholder. The runners that passed `filters=` with raw strings (logs list + services, traces list + count + the attribute/service-name helpers) now pass absolute datetimes from their existing `QueryDateRange(exact_timerange=True)` via a new `QueryDateRange.to_hogql_filters()`. Absolute datetimes take the verbatim branch, so nothing snaps, and with the strict `<` restored for exact bounds (see the `LtEq` thread) the list predicate is identical to the count runner's `timestamp < {date_to}`. The tracing aggregation runner already did exactly this inline; it now shares the helper.

Added `test_relative_date_from_keeps_exact_window`: `-1d` on the list query at a mid-day frozen now, asserting the midnight-exact fixtures stay out.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

posthog/hogql/filters.py
thmsobrmlr: Fixed in 24be1e5 — the cap wasn't intended. `_open_ended_date_to` now returns `None` for `date_from="all"`, so both placeholders collapse and `{filters}` adds no bounds at all. Pinned by `test_replace_filters_all_time_stays_unbounded` plus an `"all"` row in the skipped-placeholder cases, and called out in the PR body. Also dropped `"all"` from the banner's affected shapes, since nothing changes for those insights anymore.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

posthog/hogql/filters.py
thmsobrmlr: Fixed in 24be1e5 — `_resolve_date_to` now reports whether it snapped, and `{filters}` uses `<=` only for snapped end-of-day bounds. Verbatim datetimes, sub-day relatives, and `explicitDate` keep the strict `<`, so the count-ranges bucket contract and its help_text hold: midnight-exact belongs to the next day again, and the two `_boundary_query` edits are reverted. Heatmaps keeps its explicit next-day-midnight bound, now compared `< next-day 00:00`, which is the same window as `<= 23:59:59.999999` — its snapshots revert to master.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

frontend/src/queries/nodes/DataVisualization/Components/SqlInsightDateFilterNotice.tsx
thmsobrmlr: Fixed in 24be1e5 — `explicitDate` now suppresses only the `date_to` arm; a day-or-coarser relative `date_from` triggers the banner regardless, and the `date_to` arm also matches bare relative values like `-3d`. Added the "Last month" + exact-time-toggle case and a relative `date_to` case to the tests. One more shape moved while closing this: `date_from="all"` no longer triggers the banner, since the open-ended cap was removed for it (see the other thread) and those insights don't change.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

posthog/hogql_queries/hogql_query_runner.py
thmsobrmlr: Settled in 24be1e5, mostly your way — the clobber stays (it matches the `apply_dashboard_filters.py` doctrine, and a saved `explicitDate` paired with a dashboard's new range is stale by construction), but coerced to `bool(...)`, so the serialized value stays `false`, the three `test_insight_model.py` assertions revert to master, and the response-shape change disappears. Added a comment at the assignment pointing at the divergence from the base `QueryRunner`'s guarded version. Aligning `query_runner.py:2486` changes every insight type, so that's a follow-up PR rather than a rider here.

*Comment from PostHog Code, on behalf of @thmsobrmlr.*

georgemunyoro (approved):

PR #73842, opened 2026-07-27
feat(insights): notice on SQL insights affected by date filter fix

thmsobrmlr:
## TL;DR

Stacked on [#73833](https://github.com/PostHog/posthog/pull/73833). That PR changes what numbers SQL insights show for relative and open-ended date ranges. This one tells users why: a dismissable banner appears on exactly the SQL insights whose results can shift, behind a feature flag we flip on when the fix deploys.

## Problem

#73833 makes SQL insights resolve date filters like every other insight. Insights with a relative (`mStart`, `-7d`) or open-ended date range will show different, now correct, numbers on their next refresh. Users should see an explanation on the affected insights instead of wondering why the numbers moved.

## Changes

- New `SqlInsightDateFilterNotice`: an info `LemonBanner` (dismissable once per user via `dismissKey`) rendered in `DataVisualization`, so it appears on SQL insight views but not on dashboard tiles or other embedded contexts. The SQL editor's output pane renders its own visualization fork, so it is not covered.
- `isAffectedByDateFilterResolutionChange` mirrors the affected shapes from `posthog/hogql/filters.py`: query contains a `{filters` placeholder, and the date range is day-or-coarser relative, open-ended, or has a date-only upper bound. Sub-day rolling windows and verbatim datetimes are excluded, matching the backend exemptions.
- Gated behind the new `sql-insight-date-filter-notice` flag (off by default): enable when #73833 deploys, retire the flag and component after a few weeks.

Automated only: `SqlInsightDateFilterNotice.test.ts` covers the predicate with a case per affected/exempt shape — it pins the contract between this notice and the backend fix's exemptions (sub-day rolling windows, verbatim datetimes, explicitDate), which no other test encodes. The banner rendering itself is LemonBanner framework behavior and is not re-tested. Jest/typecheck could not run in this sandbox (no node_modules); relying on CI.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None; the behavior change itself is documented in #73833.

## 🤖 Agent context

- Authored with PostHog Code (Claude Code) as a follow-up to #73833; skill invoked: `/writing-tests`.
- Modeled on the existing `SamplingDeprecationNotice` pattern (conditional banner on matching insights). Considered a per-insight dismiss key and rejected it as nagging; one global dismissal reads once and stays away. Placed in `DataVisualization`'s non-embedded path rather than the editor-filters list because SQL insights don't render `InsightViz`.

---

### DIFF
--- a/frontend/src/lib/constants.tsx
+++ b/frontend/src/lib/constants.tsx
@@ -483,6 +483,7 @@ export const FEATURE_FLAGS = {
     SLACK_DWH: 'slack-dwh', // owner: @MarconLP #team-warehouse-sources
     SQL_EDITOR_QUERY_HISTORY: 'sql-editor-query-history', // owner: @annaszell #team-data-tools
     SQL_EDITOR_VIM_MODE: 'sql-editor-vim-mode', // owner: @arthurdedeus
+    SQL_INSIGHT_DATE_FILTER_NOTICE: 'sql-insight-date-filter-notice', // owner: #team-product-analytics, gates the notice on SQL insights affected by the date filter resolution fix
     SSE_DASHBOARDS: 'sse-dashboards', // owner: @aspicer #team-analytics-platform
     SUBSCRIPTION_AI_PROMPT: 'ai-subscriptions', // owner: #team-analytics-platform, gates AI prompt-based subscriptions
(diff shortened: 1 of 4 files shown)


thmsobrmlr:
Folded into #73833 — the notice now ships there (commit 9a7f0a9) and that PR is green. Closing this as redundant.

_Comment from PostHog Code._

PR #73853, opened 2026-07-27
feat(data-warehouse): add 24 tables to the Klaviyo source

Gilbert09:
## Problem

The Klaviyo source only covered eight tables. The coverage audit in #73746 (which this is stacked on) diffed our endpoint map against the [klaviyo/openapi](https://github.com/klaviyo/openapi) stable spec and found we expose a small slice of the 42 top-level collections Klaviyo publishes.

The big gaps: segments (Klaviyo users organise everything around them and we had none of it), flow step structure, and Klaviyo's own computed campaign and flow performance numbers. Without those you cannot break a flow down by step, cannot join a profile to a segment, and have to recompute engagement metrics that Klaviyo already calculates.

## Changes

24 new tables, all additive. No existing table, incremental field, primary key, partition setting, or sync path changed.

| Area | Tables |
| --- | --- |
| Segments | `segments`, `segment_profiles` |
| Flow structure | `flow_actions`, `flow_messages` |
| Reporting | `campaign_values_reports`, `flow_values_reports` |
| Content | `templates`, `forms`, `images`, `web_feeds` |
| Catalog | `catalog_items`, `catalog_variants`, `catalog_categories` |
| Commerce | `coupons`, `coupon_codes`, `reviews` |
| Organisation | `tags`, `tag_groups`, `custom_metrics` |
| Platform | `data_sources`, `object_types`, `push_tokens`, `webhooks`, `accounts` |

Three pieces of plumbing were needed:

**Generalised fan-out.** The list membership fan-out was hardcoded to `/lists` and a `list_id` column. It is now driven by a `KlaviyoFanOutConfig` on the endpoint (parent path, parent page size, parent id column), so `segment_profiles` reuses it verbatim. `list_profiles` keeps its exact URLs, row shape, resume bookmark, and `sort_mode`, which the existing tests still pin.

**Two-level fan-out.** Klaviyo only exposes flow messages under `/flow-actions/{id}/flow-messages`, and flow actions only under `/flows/{id}/flow-actions`, so `flow_messages` walks flows, then actions, then messages. Rows carry both `flow_action_id` and `flow_id`. It is off by default because of the cost.

**POST transport for reporting.** The values reports take their query in a JSON:API body rather than query params, so `_fetch_page` grew an optional body and switches to POST. Each grouping plus its statistics flattens into one row. Klaviyo requires a conversion metric on every report and has no account default to read, so there is a new optional `conversion_metric_id` source field. Left blank, it resolves the Placed Order metric that Klaviyo's own reporting defaults to, and falls back to the account's first metric. Every row records which metric was used.

Incremental support follows what the spec actually documents per endpoint. Where there is a server-side timestamp filter (`segments`, `templates`, `forms`, `images`, `web_feeds`, `flow_actions`, `flow_messages`, `segment_profiles`) the table is incremental. Where there is not (catalog, coupons, tags, push tokens, and friends) it is full refresh, rather than paging the whole collection and pretending. `reviews` only documents `greater-or-equal` on its `created` filter, so endpoint configs now carry an operator, defaulting to the existing `greater-than`.

Page sizes are pinned per endpoint from the spec because Klaviyo caps them individually and 400s anything larger.

Each new table has a canonical description with a docs URL and per-column descriptions taken from the spec.

> [!NOTE]
> New tables only auto-enable on sources that opted into auto-sync of new schemas. The expensive fan-outs (`segment_profiles`, `flow_messages`) and the existing `list_profiles` are default-off regardless.

I did not have a Klaviyo account to hit, so everything below is automated. Endpoint paths, filter operators, sort enums, page size caps, request bodies, and response shapes all come from the vendor spec at `openapi/stable.json`, pinned to the `2026-07-15` revision this source already targets.

- `hogli test products/warehouse_sources/backend/temporal/data_imports/sources/klaviyo/` - 99 passed. The 51 pre-existing tests still pass unchanged apart from one assertion that moved from `fan_out_over_lists` to the config object that replaced it.
- `hogli test products/warehouse_sources/backend/temporal/data_imports/sources/tests/` - 3778 passed (registry, categories, version pins, generated configs).
- `uv run mypy --cache-fine-grained .` - clean, 16930 files.
- `ruff check` and `ruff format` on the source directory, plus `hogli ci:preflight --fix` with 0 failures.

New tests and the regression each one catches:

- Generalised fan-out (4 tests): that `segment_profiles` uses the segment parent and `segment_id` column rather than falling back to the hardcoded list path; that non-membership fan-out keeps the resource fields and gains the parent id (dropping `flow_id` makes the table unjoinable); that the two-level walk formats the intermediate path with the grandparent id and carries both ancestors; that a flow deleted mid-enumeration is skipped rather than failing the sync.
- Values reports (5 tests): that groupings and statistics flatten into one row rather than staying nested under the declared primary key; that the body carries statistics, timeframe, and conversion metric, which Klaviyo 400s without; that a configured metric wins and skips the lookup call; that the fallback picks the first metric when Placed Order is absent; that an account with no metrics yields nothing instead of posting an invalid report.
- Request params (3 tests): that each endpoint's `page[size]` stays within its documented cap and the unsized endpoints send none, since either mistake fails the whole table; that reviews use the inclusive operator Klaviyo documents.
- Schemas (3 tests): that the expensive fan-outs stay opt-in; that lookback endpoints are merge-only, since append would materialise the deliberate re-pulls as duplicates; that every endpoint reaches the schema list.
- Extended the existing `sort_mode` case list to cover the new fan-outs.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

The posthog.com source doc renders its supported tables from `get_documented_tables`, which this picks up automatically. The connect form gained one optional field, so `<SourceParameters />` updates itself too.


estefaniarabadan (approved):

Gilbert09:
/trunk merge

Gilbert09:
/trunk merge

PR #73869, opened 2026-07-27
fix(workflows): make workflows list search server-side

benjackwhite:
## Problem

The workflows list search was purely client-side. The list logic fetched a single capped page (up to 200 flows) once on mount and then fuzzy-matched over that in-memory array. Any workflow past the first page was invisible to search, and the status and creator filters had the same blind spot. Reported in Slack: searching for a workflow that clearly exists returned no results.

## Changes

Moved filtering and pagination to the backend, following the same shape as the feature flags list.

Backend (`products/workflows/backend/api/hog_flow.py`):
- added a server-side `search` param that matches on workflow name and description (case-insensitive, spaces match `-`/`_` runs, same regex approach as feature flag search)
- added a `created_by` filter keyed on user uuid (the list UI's member picker uses uuid, not pk), and dropped the old pk-based `created_by` filterset field
- documented both params with `@extend_schema_view` so they flow into generated types and MCP
- dropped the inflated default page size that only existed to let the client load everything at once (`default_limit` 200 to 100, `max_limit` unchanged at 500)

Frontend:
- `getHogFlows` now forwards `search`/`status`/`created_by`/`limit`/`offset`
- `workflowsLogic` sends those params, drives a server-side `PaginationManual`, debounces reloads on filter changes, and syncs filters to the URL so a search is shareable
- the table renders the server page and its `count`-based pagination instead of a client-side slice

> [!WARNING]
> Breaking change for direct API consumers. `created_by` is now a user uuid, so `?created_by=<pk>` returns a 400 with `"Must be a valid user uuid"` instead of filtering. The previous generated types advertised `created_by?: number`, so any script or integration written against that shape breaks with no deprecation window. Nothing in the product does this, and the app's member picker already sends uuids.

### Follow-up fixes from local QA

Daniel took this branch for a manual end to end pass on a project with 259 workflows and pushed two commits on top. Three issues, all created or made materially worse by moving paging server-side:

**Unstable pagination.** `order_by("-updated_at")` has no tiebreaker, and LIMIT/OFFSET over a non-unique sort key is unstable. With 120 rows sharing one `updated_at`, walking every page returned 39 rows twice and never showed 54 of them at all. That is the same invisibility this PR set out to fix. Adding `-id` as a tiebreaker takes it to 0 and 0. The old code was not exposed because it fetched one page and sliced client-side.

**Empty page showed onboarding.** An out-of-range `?page=` (a stale link, or rows deleted since it was shared) rendered the "Create your first workflow" empty state on a project full of workflows, with no table, no search box and no pagination control to get back. Now the page clamps to the last real page after a load, and the product introduction is gated on being on page 1. The `page` URL param is also parsed defensively, which covers the Graphite review comment about `NaN` and negative values.

**Workflow names stopped resolving in ticket activity.** `WorkflowActivityLink` in `products/conversations/` resolved a workflow name by scanning `workflowsLogic`'s list page. That page went from 200 rows to 30, so on teams with more than 30 workflows the ticket activity feed fell back to a generic "A workflow" label. Worse, `workflowsLogic` is a singleton that stays mounted across navigation, so after searching on the workflows list the feed resolved names against that filtered page. It now fetches the single workflow by id through a new keyed `workflowNameLogic`, which also dedupes when several activity rows cite the same workflow.

Added backend tests in `test_hog_flow.py`: a parameterized matrix for `search` (name match, case-insensitivity, description-only match, space-to-separator, no match) plus a `created_by`-by-uuid test. They guard the regression this PR fixes, a workflow that exists but isn't on the first loaded page must still be findable now that the query runs in the DB.

Automated, all passing locally: `test_hog_flow.py` (191 tests, re-run after the ordering change since that was the riskiest edit), the MCP unit suite (2401 tests, covering the regenerated `workflows-list` schema), the `WorkflowActivityLink` jest tests rewired to the new by-id lookup, and the frontend typecheck.

Manual: Daniel did a full browser pass on a local project seeded to 259 workflows, with the search target at position 254 so it fell outside the old fetch. Reproduced the bug on master first (footer read "1-30 of 200 workflows", searching the target returned "No workflows matching filters"), then confirmed the fix and exercised search by name and description, the creator and status filters, pagination over a filtered set, filter changes resetting to page 1, deep links, the `workflows-list` MCP tool against a running local server, and the ticket activity feed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover the workflows list filters, so nothing to update.

## 🤖 Agent context

Requested by Ben White from a Slack thread after a report that workflows search returned nothing. Claude confirmed the root cause (client-side Fuse search over a single capped page, no `search` filter on the backend), then implemented the fix modeled on the feature flags list, which already does server-side search + pagination + URL sync end to end.

Skills invoked: `/improving-drf-endpoints`, `/adopting-generated-api-types`, `/writing-tests`, and `/posthog-local-e2e-qa` for the follow-up QA pass.

Decisions: moved status and creator filtering server-side too, not just search, because introducing server-side pagination would otherwise leave those filters operating over a single page. Kept the handwritten `api.hogFlows.getHogFlows` helper rather than switching to the generated `hogFlowsList` function, to stay consistent with the rest of the hog flows API surface in `api.ts`.

On the follow-up commits, Daniel drove the QA and Claude wrote the fixes. Two things worth flagging for the reviewer. The activity link fetches the workflow with `getHogFlow(id)` rather than the list endpoint filtered by `?id=`, because the web list serializer returns the full action and edge graph anyway, so `?id=` is no cheaper and retrieve gives a clean 404 for a deleted workflow. No regression test was added for the activity link fix, since the existing test renders with no list state and already fails if anyone reimplements it by scanning the list.

Both the missing tiebreaker and the missing page clamp look systemic rather than specific to this PR. Feature flags falls back to `order_by("-created_at")` with no tiebreaker and has no clamp either, so the same two one-liners probably belong there.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C06GG249PR6/p1785144586618789?thread_ts=1785144586.618789&cid=C06GG249PR6)*


mayteio (approved):
LGTM just check the bot comment! Cheers for this

dmarchuk (approved):

PR #73871, opened 2026-07-27
fix(insights): scale axis decimals to small values

sampennington:
## Problem

An insight tracking mean query latency in seconds (values roughly 0.002 to 0.012) rendered a y-axis where every tick read either `0.01` or `0`. Four identical labels stacked on top of each other, and the axis told you nothing.

The cause is a flat two-decimal default. Once a series sits below 0.1, two decimals can't separate one tick from the next, so `.nice()` produces perfectly good tick values and the formatter collapses them all into the same string. The same rounding hit tooltips, value labels, tables, and bold numbers for those insights.

## Changes

Numeric and percentage formatting now picks its precision from the value's magnitude when no explicit `decimalPlaces` is set: two decimals at or above 0.1, plus one more decimal for each additional leading zero below that, capped at ten so near-zero float residue doesn't render as a wall of digits. `0.012` renders as `0.012`, `0.002` as `0.002`, and anything at 0.1 or above is unchanged.

The rule is applied in three places so the whole insight agrees:

- `buildYTickFormatter` in `@posthog/quill-charts` (y-axis ticks)
- `formatAggregationAxisValue` in the app (tooltips, value labels, tables, bold numbers)
- `autoFormatYTick`, the chart's fallback formatter for charts with no format config, which scales precision to the axis domain

An explicit `decimalPlaces` still wins, `minDecimalPlaces` now floors the derived precision (which also closes a latent `RangeError` when it was set above two), and currency and duration formats are untouched.

Automated, all run locally by me (Claude). No manual browser run, so there is no screenshot — the assertions exercise the same formatters the axis, tooltip, and table call.

- `y-formatters.test.ts` gains parameterized cases asserting a small-valued tick set stays distinct across `numeric`, `percentage`, and `percentage_scaled`, including a negative tick (the `Math.abs` in the helper is load-bearing — without it `-0.002` renders as `-0`). This is the regression: before the change those ticks all formatted to `0.01` / `0`.
- It also covers a null `minDecimalPlaces`, which caught a real bug introduced while addressing review: the trends adapter assigns `minDecimalPlaces` straight from the filter, and a saved insight sends that field as `null`, not `undefined`. Leaning on a default parameter meant `Math.max(null, …)` coerced the floor to 0, so a y-axis tick at `1.2345` rendered as `1.2`. Fixed by validating the floor instead. Verified the case fails against the pre-fix helper.
- `scales.test.ts` gains cases for `autoFormatYTick` over sub-0.1 domains, covering the path changed there.
- `aggregationAxisFormats.test.ts` gains rows for small numeric and percentage values, plus one that pins the `* 100` on the `percentage_scaled` branch — without it a refactor could silently regress to over-precision.
- Full `packages/quill/packages/charts` suite plus the insights / product analytics / lib-utils suites: 159 files, 3535 tests, all passing. Lint, format, typecheck, and `ci:preflight` clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the y-axis `format` bullet in `packages/quill/packages/charts/AGENTS.md` to document the precision rule.

## 🤖 Agent context

Written by Claude (PostHog Code) from a report of duplicate y-axis labels on one dashboard tile. I pulled the insight config over MCP to confirm the axis was on the default `numeric` format with `decimalPlaces` unset, then reproduced the exact label sequence from the screenshot by running the old formatter over a 0 to 0.012 tick set. That ruled out tick generation and pinned it on formatting.

I considered making the formatter tick-aware, passing the whole tick array in so precision could be derived from the step. That is the fully general fix, but it means changing the `tickFormatter` signature across the charts public API for a case that value-magnitude precision already covers for `.nice()`-generated linear ticks. Went with the local rule instead. I also chose to fix the app-side formatter rather than only the axis, since leaving the tooltip at `0.01` while the axis says `0.012` is its own bug.

The PR was then put through Sam's `sp-ship` review loop (`sp-swarm` fan-out on opus + `sp-triage` on sonnet, both loaded from the PostHog skills store). That surfaced twelve findings; seven were applied — dropping the helper from the charts package's public exports, capping the digit ceiling, threading `minDecimalPlaces`, and three test gaps — and four were deferred as out of scope and left open on the PR: unifying the per-domain and per-value precision policies, the `duration` / `currency` formats that still collapse distinct ticks, a Python port of the formatter under `products/alerts` that will now disagree with the chart, and the near-duplicate switch shared with `buildYTickFormatter`. The null-coercion bug above came out of one of those applied fixes and was caught on verification.

Skills invoked: `/writing-tests`, `skills-store`, `sp-ship`, `sp-swarm`, `sp-triage`, `hog-charts-review`, `xp-reviewer`.

---


packages/quill/packages/charts/src/index.ts
```diff
@@ -173,7 +173,7 @@ export { createXAxisTickCallback } from './utils/dates'
 export type { TimeInterval } from './utils/dates'
 export { buildYTickFormatter } from './utils/y-formatters'
 export type { YAxisFormat, YFormatterConfig } from './utils/y-formatters'
-export { percentage } from './utils/format'
+export { percentage, significantDecimalPlaces } from './utils/format'
```
sampennington: 🤖 **sp-ship** · ♻️✨📊📐 `[reuse+quality+hog-charts+xp]` · **Blocker**

This puts `significantDecimalPlaces` on the published surface while `frontend/src/lib/utils/numbers.ts:71` carries a byte-identical copy — so one of the two is superfluous, and nothing pins them together (each has its own tests, so drift is silent). Pick one before merge: drop this export and keep the app copy, or delete the app copy and import from `@posthog/quill-charts` here. Note the `format.ts` no-PostHog-imports constraint only runs one way — nothing stops the app importing charts.

packages/quill/packages/charts/src/utils/format.ts
```diff
@@ -17,6 +17,17 @@ function validateFractionDigits(maximumFractionDigits: number, fallback: number)
...
 
+/** Fraction digits needed to keep two significant digits — {@link DEFAULT_DECIMAL_PLACES} for
+ *  anything at or above 0.1, and one more for every extra leading zero below that. A flat two
+ *  decimals collapses a small-valued axis into repeated labels: ticks over 0–0.012 all round to
+ *  "0.01" or "0". */
+export function significantDecimalPlaces(value: number, minimum: number = DEFAULT_DECIMAL_PLACES): number {
+    if (!isFinite(value) || value === 0) {
+        return minimum
+    }
+    return Math.min(Math.max(minimum, 1 - Math.floor(Math.log10(Math.abs(value)))), 100)
```
sampennington: 🤖 **sp-ship** · 💻✨ `[code+quality]` · **Suggestion / follow-up**

The `100` ceiling is Intl's hard limit, not a legibility limit, so float residue now renders as a very long label where it used to collapse to `0`. A trends formula like `A/B*100 - 100` that should be exactly zero often lands on `1.7763568394002505e-15`; this returns 16 and the value renders as `0.0000000000000018` in ticks, tooltips, legend rows and `InsightsTable` cells. Consider `Math.min(..., 10)` — still covers microsecond-scale seconds (1e-6 needs 7) while keeping noise at `0`. Also worth naming the two constants: the `1 -` silently encodes "two significant digits" (so it can drift from the doc comment above) and `100` is already hardcoded in `validateFractionDigits` a few lines up. Same line in the twin at `frontend/src/lib/utils/numbers.ts:75`.

packages/quill/packages/charts/src/utils/y-formatters.ts
```diff
@@ -45,7 +46,7 @@ function formatNanoseconds(value: number): string {
 export function buildYTickFormatter(config: YFormatterConfig): (value: number) => string {
     const { format, prefix, suffix, decimalPlaces, minDecimalPlaces, currency } = config
     return (value: number): string => {
-        let formatted = humanFriendlyNumber(value, decimalPlaces, minDecimalPlaces)
+        let formatted = humanFriendlyNumber(value, decimalPlaces ?? significantDecimalPlaces(value), minDecimalPlaces)
```
sampennington: 🤖 **sp-ship** · 💻📊📐♻️✨ `[code+hog-charts+xp+reuse+quality]` · **Suggestion / follow-up**

`minDecimalPlaces` isn't threaded into `significantDecimalPlaces`, so the helper's `minimum` parameter has no caller anywhere — and it's the fix for a live crash. `{ minDecimalPlaces: 3 }` with no `decimalPlaces` gives `humanFriendlyNumber(50, 2, 3)` → `RangeError: maximumFractionDigits value is out of range`, because `validateFractionDigits` sanitizes each bound independently so `min > max` reaches `toLocaleString`. Pre-existing, but this change makes it value-dependent (throws at ≥ 0.1, works below), so an axis can start crashing as its data grows. `significantDecimalPlaces(value, minDecimalPlaces)` closes it and gives the parameter a purpose. While you're here, `decimalPlaces ?? significantDecimalPlaces(...)` is restated three times with a different argument each time — a hoisted `const resolveDecimals = (v: number): number => decimalPlaces ?? significantDecimalPlaces(v, minDecimalPlaces)` states the rule once. Same shape at `frontend/src/scenes/insights/aggregationAxisFormat.ts:95`.

packages/quill/packages/charts/src/core/scales.ts
```diff
@@ -728,7 +729,12 @@ function padValueRange(
 
 export function autoFormatYTick(value: number, domainMax: number): string {
     if (domainMax < 2) {
-        return value.toFixed(2)
+        // Two decimals only resolve a domain down to ~0.1; below that every tick rounds to the same
+        // label, so scale the precision to the domain instead.
+        return value.toLocaleString('en-US', {
+            minimumFractionDigits: 2,
+            maximumFractionDigits: significantDecimalPlaces(domainMax),
```
sampennington, line 736: 🤖 **sp-ship** · 📊✨📐♻️⚡ `[hog-charts+quality+xp+reuse+efficiency]` · **Suggestion / follow-up**

Keying precision to `domainMax` only fixes the `domainMax < 2` branch, because a domain's max magnitude doesn't bound its tick *step* — the same collapse is still reachable on the two branches left alone. A log axis with no explicit `format` and ticks `[0.001, 0.01, 0.1, 1, 10, 100]` falls to `maximumFractionDigits: 0`, so the first three all render `0`. With `startAtZero: false` / `floatBaseline: true`, a 1.234–1.240 domain renders `1.23 | 1.24 | 1.24 | 1.24 | …` and a 100.01–100.05 domain renders every tick as `100`. So the package now answers "how much precision does this axis use?" two different ways — per domain here, per value in `buildYTickFormatter` — and which one an axis gets depends only on whether `format` is set, which isn't a distinction a reader would predict. Keying off each tick's own magnitude (or the tick step) would cover all three cases with one rule; the AGENTS.md bullet currently describes only the per-value one. Separately, this expression is exactly `humanFriendlyNumber(value, significantDecimalPlaces(domainMax), 2)` — the import is already in hand, output is identical in this branch, and it picks up `validateFractionDigits` clamping for free; hoisting the digit count into `autoFormatterFor` (which already closes over a fixed `domainMax`) would also stop rebuilding it per tick.

packages/quill/packages/charts/src/utils/y-formatters.ts
```diff
@@ -45,7 +46,7 @@ function formatNanoseconds(value: number): string {
...
         switch (format) {
             case 'duration':
                 formatted = humanFriendlyDuration(value)
```
sampennington, line 56: 🤖 **sp-ship** · 📊 `[hog-charts]` · **Suggestion / follow-up**

`duration` and `currency` still produce identical labels for distinct ticks — the same user-visible symptom being fixed for `numeric`/`percentage`. `duration` on a 0–2s axis: `[0, 0.5, 1, 1.5, 2]` → `0s | 500ms | 1s | 2s | 2s` (`humanFriendlyDuration` uses `toFixed(0)` for 1 ≤ d < 60). `currency` on this PR's own 0–0.012 domain: `$0.00 | $0.00 | $0.00 | $0.01 | $0.01` (`humanFriendlyCurrency` hardcodes 2 dp). `short` is fine. Related: `decimalPlaces` silently no-ops for `short`, `duration*` and `currency` — a config a caller can set and never see honored.

frontend/src/scenes/insights/aggregationAxisFormat.ts
```diff
@@ -84,7 +90,11 @@ export const formatAggregationAxisValue = (
     const aggregationAxisPostfix =
         (trendsFilter as TrendsFilter)?.aggregationAxisPostfix ??
         (trendsFilter as Partial<TrendsFilterType>)?.aggregation_axis_postfix
-    let formattedValue = humanFriendlyNumber(value, maxDecimalPlaces, minDecimalPlaces)
+    let formattedValue = humanFriendlyNumber(
+        value,
+        maxDecimalPlaces ?? significantDecimalPlaces(value),
```
sampennington: 🤖 **sp-ship** · 💻📊 `[code+hog-charts]` · **Suggestion / follow-up**

There's a third copy of this formatter that didn't get the fix: `products/alerts/backend/evaluation/formatting.py` documents itself as a backend port of `formatAggregationAxisValue` ("render a raw metric value as the user sees it on the chart"), and its `_human_friendly_number` / `_percentage` still clamp to `DEFAULT_DECIMAL_PLACES` when no explicit `decimal_places` is set. After this change an alert on a sub-0.1 insight reports the breach at `0` while the chart it links to shows `0.002` — the exact mismatch the port exists to prevent. Worth porting `significantDecimalPlaces` there in this PR, or filing the follow-up.

frontend/src/scenes/insights/aggregationAxisFormat.ts
```diff
@@ -84,7 +90,11 @@ export const formatAggregationAxisValue = (
     const aggregationAxisPostfix =
         (trendsFilter as TrendsFilter)?.aggregationAxisPostfix ??
         (trendsFilter as Partial<TrendsFilterType>)?.aggregation_axis_postfix
-    let formattedValue = humanFriendlyNumber(value, maxDecimalPlaces, minDecimalPlaces)
+    let formattedValue = humanFriendlyNumber(
```
sampennington, line 93: 🤖 **sp-ship** · ♻️📐⚡ `[reuse+xp+efficiency]` · **Suggestion / follow-up**

This switch and `buildYTickFormatter` (`y-formatters.ts:48`) are the same eight cases over the same eight-member union, `formatNanoseconds` is duplicated verbatim, and `AggregationAxisFormat` duplicates the exported `YAxisFormat` — this PR had to patch both in lockstep, which is the third strike. `buildYTickFormatter` is already public and `YFormatterConfig` covers `format`/`prefix`/`suffix`/`decimalPlaces`/`minDecimalPlaces`/`currency`, so this function could shrink to mapping `trendsFilter` (camelCase + legacy snake_case) into that config, keeping only the two app-specific bits: `captureException` on a bad currency code, and the currency prefix dedup. Also, while you're in this line: `formattedValue` is built eagerly with `humanFriendlyNumber` — an `Intl.NumberFormat` construction — then overwritten by every branch except `numeric`/`default`, and this function runs per table cell, per value label, per tooltip row and per map region. Folding it into the `numeric` case stops a duration- or currency-formatted insight paying a discarded formatter build per rendered value. Same pattern at `y-formatters.ts:49`.

sampennington (commented):
🤖 **sp-ship** · 🚫 Blockers

The precision fix is correct and the tests genuinely fail without it, but the diff leaves `significantDecimalPlaces` on the charts package's public surface *and* byte-identical in `lib/utils/numbers.ts`, with nothing tying the two together — pick one before merge.

Two things worth deciding rather than deferring. The `100`-digit ceiling means near-zero float residue (the `1.7e-15` a formula insight produces instead of exactly 0) now renders as sixteen decimals where it used to collapse to `0`. And the fix lands on two different precision policies: per-value in `buildYTickFormatter`, per-domain in `autoFormatYTick` — so log axes, floated baselines, and the `duration` / `currency` formats still collapse distinct ticks into identical labels. Reviewers also flagged that `products/alerts/backend/evaluation/formatting.py` is a documented port of the app formatter and will now disagree with the chart it links to.

packages/quill/packages/charts/src/utils/y-formatters.test.ts
```diff
@@ -36,5 +36,24 @@ describe('buildYTickFormatter', () => {
...
+        // An explicit setting wins over the small-value precision below
+        expect(fmt(0.012)).toBe('0.01')
+    })
+
+    // A two-decimal default collapsed every tick of a small-valued axis (e.g. latency in seconds)
+    // into the same "0.01" / "0" label.
+    const smallTickCases: { format: YAxisFormat; ticks: number[]; expected: string[] }[] = [
+        {
+            format: 'numeric',
+            ticks: [0, 0.002, 0.008, 0.01, 0.012],
```
sampennington: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

No added case exercises a negative small value, and that's the one branch of `significantDecimalPlaces` these formatter-level tests can actually observe. `Math.abs(value)` is load-bearing: drop it and `Math.log10(-0.002)` → `NaN` → `Math.max(2, NaN)` → `NaN` → `validateFractionDigits` silently falls back to 2, so `-0.002` renders as `-0`. Insight series can go negative, so that's this PR's own bug mirrored. Cheapest fix: extend this row to `ticks: [0, -0.002, 0.002, 0.008, 0.01, 0.012]` / `expected: ['0', '-0.002', '0.002', '0.008', '0.01', '0.012']`.

frontend/src/scenes/insights/aggregationAxisFormats.test.ts
```diff
@@ -84,6 +84,13 @@ describe('formatAggregationAxisValue', () => {
         { candidate: 0.8709423, filters: { decimalPlaces: 3 }, expected: '0.871' },
         { candidate: 0.8709423, filters: { decimal_places: 9 }, expected: '0.8709423' },
         { candidate: 0.8709423, filters: { decimal_places: -1 }, expected: '0.87' }, // Fall back to default for unsupported values
+        // Two decimals aren't enough for small values — 0.012 and 0.002 both used to render as "0.01" / "0"
+        { candidate: 0.012, filters: {}, expected: '0.012' },
+        { candidate: 0.002, filters: {}, expected: '0.002' },
+        { candidate: 0.00012345, filters: {}, expected: '0.00012' },
+        { candidate: 0.012, filters: { decimal_places: 2 }, expected: '0.01' }, // An explicit setting still wins
+        { candidate: 0.005, filters: { aggregation_axis_format: 'percentage' }, expected: '0.005%' },
+        { candidate: 0.00005, filters: { aggregation_axis_format: 'percentage_scaled' }, expected: '0.005%' },
```
sampennington: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

This case doesn't pin the `* 100` the fix deliberately added — `significantDecimalPlaces(value * 100)` and `significantDecimalPlaces(value)` both yield `'0.005%'` for `0.00005`, because the extra allowed digits are invisible once trailing zeros drop (same for the sibling row at `y-formatters.test.ts:52`). A refactor that collapses the two percentage branches onto a single `significantDecimalPlaces(value)` would pass every test while regressing labels to over-precision. Pick a value where the scaling shows: `{ candidate: 0.0123456, filters: { aggregation_axis_format: 'percentage_scaled' }, expected: '1.23%' }` renders `'1.235%'` without the `* 100`. The `percentage` branch's argument is already pinned by the existing `3.944 → '3.94%'` row.

packages/quill/packages/charts/src/core/scales.test.ts
```diff
@@ -689,6 +689,18 @@ describe('hog-charts scales', () => {
             expect(autoFormatYTick(value, domainMax)).toBe(expected)
         })
 
+        it.each([
+            { domainMax: 0.012, value: 0.012, expected: '0.012' },
+            { domainMax: 0.012, value: 0.002, expected: '0.002' },
+            { domainMax: 0.012, value: 0, expected: '0.00' },
```
sampennington: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

This row passes both before and after the change — `0` renders as `'0.00'` under the old `value.toFixed(2)` and under the new `toLocaleString` with `minimumFractionDigits: 2` at any max-digit count. It can only fail if the hardcoded `minimumFractionDigits: 2` goes away, which the pre-existing `domainMax: 0, value: 0.5 → '0.50'` row already covers. Drop it and keep the two that discriminate.

frontend/src/scenes/insights/aggregationAxisFormats.test.ts
```diff
@@ -84,6 +84,13 @@ describe('formatAggregationAxisValue', () => {
         { candidate: 0.8709423, filters: { decimalPlaces: 3 }, expected: '0.871' },
         { candidate: 0.8709423, filters: { decimal_places: 9 }, expected: '0.8709423' },
         { candidate: 0.8709423, filters: { decimal_places: -1 }, expected: '0.87' }, // Fall back to default for unsupported values
+        // Two decimals aren't enough for small values — 0.012 and 0.002 both used to render as "0.01" / "0"
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

Two added comments here restate the table rows they sit on. This one — `// Two decimals aren't enough for small values — 0.012 and 0.002 both used to render as "0.01" / "0"` — narrates the cases directly below, which already say it (`{ candidate: 0.012, expected: '0.012' }`), and it's PR context rather than a constraint a future reader needs. Line 91's `// An explicit setting still wins` restates its own row (`decimal_places: 2` → `'0.01'`). Suggest deleting both.

packages/quill/packages/charts/src/utils/y-formatters.test.ts
```diff
@@ -36,5 +36,24 @@ describe('buildYTickFormatter', () => {
     it('respects decimalPlaces for numeric format', () => {
         const fmt = buildYTickFormatter({ format: 'numeric', decimalPlaces: 2 })
         expect(fmt(1.2345)).toBe('1.23')
+        // An explicit setting wins over the small-value precision below
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

Two added comments here restate what's beside them. This one — `// An explicit setting wins over the small-value precision below` — restates the assertion beneath it, inside a test already named `respects decimalPlaces for numeric format`. Lines 43–44 (`// A two-decimal default collapsed every tick of a small-valued axis …`) are PR/bug context the `it.each` name (`keeps small $format ticks distinct`) plus the expected values already convey. Suggest deleting both.

packages/quill/packages/charts/src/index.ts
sampennington: 🤖 **sp-triage**

Fixed in `3783713` — dropped `significantDecimalPlaces` from the charts package barrel export (`export { percentage } from './utils/format'`). Every in-package caller (`scales.ts`, `y-formatters.ts`) already imports it directly from `./utils/format`, and nothing outside the package used the barrel export, so the app-side copy in `frontend/src/lib/utils/numbers.ts` stays as the one public entry point.

packages/quill/packages/charts/src/utils/format.ts
sampennington: 🤖 **sp-triage**

Fixed in `d3597d2` — clamped `significantDecimalPlaces` to a `MAX_SIGNIFICANT_DECIMAL_PLACES = 10` ceiling instead of Intl's `100`, in both `format.ts` and the `frontend/src/lib/utils/numbers.ts` twin. Still covers microsecond-scale seconds while keeping float residue from rendering a 16-decimal label. Left the constant-naming/hoist suggestions as-is for now — the ceiling was the correctness bug.

packages/quill/packages/charts/src/utils/y-formatters.ts
sampennington: 🤖 **sp-triage**

Fixed in `d3597d2` — threaded `minDecimalPlaces` into `significantDecimalPlaces(value, minDecimalPlaces)` at all three call sites here (and the twin in `aggregationAxisFormat.ts:95`), so `minimum` can no longer exceed the computed maximum and trip `RangeError` via `validateFractionDigits`. Left the "hoist a `resolveDecimals` helper" cleanup for a follow-up since it's a pure readability change, not part of the crash fix.

packages/quill/packages/charts/src/core/scales.ts
sampennington: 🤖 **sp-triage**

Deferring — this is a design decision about how axis precision should be computed (per-domain-max vs per-tick-magnitude vs per-tick-step), not a localized fix. Agreed the two current strategies (`autoFormatYTick`'s per-domain and `buildYTickFormatter`'s per-value) can still collapse log-scale and floated-baseline axes, and that unifying them would also let this branch reuse `humanFriendlyNumber(value, significantDecimalPlaces(domainMax), 2)` directly. Leaving open for a human to weigh in on the right unification.

packages/quill/packages/charts/src/utils/y-formatters.ts
sampennington: 🤖 **sp-triage**

Deferring — extending the same small-value precision fix to `duration` and `currency` formats is a real gap (confirmed the `humanFriendlyDuration`/`humanFriendlyCurrency` hardcoded precision), but it broadens this PR's scope beyond the `numeric`/`percentage` fix and touches format-specific rendering choices (e.g. how many duration units to show, currency subunit conventions) that deserve their own review rather than a bundled fix. Also flagging the related `decimalPlaces` no-op for `short`/`duration*`/`currency` as part of the same follow-up.

frontend/src/scenes/insights/aggregationAxisFormat.ts
sampennington: 🤖 **sp-triage**

Deferring — porting `significantDecimalPlaces` into `products/alerts/backend/evaluation/formatting.py` is real drift (confirmed the Python port still clamps to `DEFAULT_DECIMAL_PLACES`) and would produce the exact chart-vs-alert mismatch described, but it's a second-language change to a different product's code, outside this PR's single-file scope. Leaving open for a human to decide whether to fold it into this PR or file as a follow-up.

frontend/src/scenes/insights/aggregationAxisFormat.ts
sampennington: 🤖 **sp-triage**

Deferring — collapsing `formatAggregationAxisValue` onto the exported `buildYTickFormatter`/`YFormatterConfig` is a good de-duplication, but it's a cross-file refactor of a public function's shape (mapping `trendsFilter`'s camelCase/snake_case duality into `YFormatterConfig`, preserving the `captureException` and currency-prefix-dedup behavior) rather than a localized fix, so it needs a human judgement call on scope and risk. Also noted: the eager `humanFriendlyNumber` formatter build being discarded on every non-numeric branch, worth folding in together.

packages/quill/packages/charts/src/utils/y-formatters.test.ts
sampennington: 🤖 **sp-triage**

Fixed in `52fd69c` — extended the numeric small-tick case to include a negative value (`-0.002` -> `-0.002`), pinning `Math.abs` as load-bearing in `significantDecimalPlaces`.

frontend/src/scenes/insights/aggregationAxisFormats.test.ts
sampennington: 🤖 **sp-triage**

Fixed in `52fd69c` — swapped the `percentage_scaled` case to `{ candidate: 0.0123456, expected: "1.23%" }`, which discriminates the `* 100` scaling (the old `0.00005` case passed identically either way once trailing zeros dropped).

packages/quill/packages/charts/src/core/scales.test.ts
sampennington: 🤖 **sp-triage**

Fixed in `52fd69c` — dropped the `domainMax: 0.012, value: 0, expected: "0.00"` row; it passes under both the old and new formatting regardless of max-digit count, and the pre-existing `domainMax: 0, value: 0.5 -> "0.50"` row already covers the `minimumFractionDigits: 2` behavior it was meant to pin.

frontend/src/scenes/insights/aggregationAxisFormats.test.ts
sampennington: 🤖 **sp-triage**

Fixed in `52fd69c` — deleted both restating comments (the small-value narration above the test cases, and the "explicit setting still wins" note on its own row).

packages/quill/packages/charts/src/utils/y-formatters.test.ts
sampennington: 🤖 **sp-triage**

Fixed in `52fd69c` — deleted both restating comments (the "explicit setting wins" note above the decimalPlaces test, and the small-value narration above `smallTickCases`); the test names and expected values already convey both.

PR #73875, opened 2026-07-27
fix(charts): repaint the canvas after a lost 2D context is restored

sampennington:
## Problem

A user reported that the line on the Web vitals charts in web analytics disappears after switching between the INP/LCP/FCP/CLS tabs a few times. It renders on first load, then eventually stops. Reported on Firefox.

A screenshot of the broken state is the whole diagnosis: y-axis labels, x-axis labels and both goal lines render, correctly positioned, and **nothing** on the canvas does — no grid, no axis lines, no tick marks, no series line.

That split is meaningful. Overlays (axis labels, goal lines, value labels, legend, tooltip) are DOM positioned from the scales; the grid, axis lines, tick marks and series are canvas. A goal line only renders when its y-value lands inside the plot box, so the layout was healthy. The canvas specifically failed, and it stayed failed.

Two mechanisms can do that, and neither was handled:

1. **A lost 2D context.** The browser can discard a canvas's backing storage under memory pressure. The bitmap resets to transparent black and draws are silently ignored until `contextrestored`. Nothing listened for it, so once it happened the canvas never came back — no exception, no error boundary, just an empty plot under fully rendered chrome. Each metric tab switch fully remounts the insight (`<Query key={webVitalsTab}>`), so a few switches allocates and drops several canvas pairs.
2. **Redundant reallocation.** The sizing path assigned `canvas.width`/`height` on *every* ResizeObserver delivery with no check that the size changed. That resets the bitmap even when the value is identical, and the repaint only lands on the next animation frame — so each redundant delivery costs a blank painted frame. `observe()` always fires one initial observation on top of the synchronous first measure, so it happened on every mount.

## Changes

All in `@posthog/quill-charts` core. No consumer-facing API change.

- **Repaint a restored context.** `useChartCanvas` listens for `contextrestored` and calls `updateSize(forceRepaint)`, which publishes fresh state even though no value moved. `contextlost` is deliberately *not* handled: `preventDefault` on it tells the browser the author will restore the context, and it then never restores.
- **Idempotent canvas sizing.** New `canvas-size.ts` holds `syncCanvasSize`, which only touches `width`/`height` when the *rounded* backing size differs. Rounding matters: `canvas.width` is an integer, so comparing a fractional target would never match and every sync would still reallocate.
- **Only publish state on a real change**, so a no-op resize delivery no longer re-renders, rebuilds the scales, and cancels the pending frame. A reallocation that wiped a bitmap without moving a value still forces the repaint.
- **Resolve both 2D contexts before touching the backing store**, so a failed allocation can't leave two wiped bitmaps with nothing scheduled to repaint them.
- **Clamp the plot clip rect.** `useChartMargins` grows the left gutter from measured label widths with a floor but no ceiling against the container, so a narrow multi-axis chart can reserve more width than it has. That made the `withVerticalClip` rect negative, which canvas reads as a reversed rectangle off the right edge — discarding the entire series layer while the DOM axis labels still rendered.
- **`try`/`finally` around both draw callbacks**, so a throwing chart-type draw can't leak an unbalanced `ctx.save()` into every later frame.
- **Testing surface:** `chart.canvas` added to the shared `HogChart` accessor, so tests reach the static canvas through the accessor instead of re-querying the DOM.

> [!NOTE]
> This branch is not pixel-neutral. `syncCanvasSize` rounds the backing store where the old code let the WebIDL `unsigned long` coercion truncate, which shifts the derived drawing scale by a fraction of a percent when the CSS width is fractional at `dpr > 1`. That moves antialiased edges in the exporter's `device_scale_factor=2` path, which is why one visual baseline (`exporter-funnels--funnel-time-to-convert-insight`) was regenerated. Rounding is the more correct behaviour; flagging it so a future bisect doesn't read the baseline update as unrelated churn.

Automated only. I (Claude) could not run Firefox in this environment, so there is no manual verification of the original report.

- `useChartCanvas.test.tsx` — renders a `Chart`, dispatches `contextrestored`, asserts it repaints *and* that the backing store came back. Verified to fail without the listener.
- `canvas-size.test.ts` — a repeat sync must not write to `width`/`height` at all (asserted by shadowing the setters, not by trusting the return value); a real size or dpr change must reallocate; a sub-pixel change that rounds to the same backing store must still update the CSS size; `buildDimensions` pinned against literals, since it is the only thing clamping the plot box and the shared jsdom fixture now derives from it.
- `canvas-renderer.test.ts` — `withVerticalClip` degrades to unclipped when the reserved gutter exceeds the container (and on a non-finite width), and restores the context when `draw` throws. It had no coverage at all before.
- Full `@posthog/quill-charts`, `@posthog/quill-components` and `products/product_analytics` suites: 99 suites, 2245 tests, green. `tsc --noEmit` clean on the package apart from pre-existing missing-jest-types noise. oxlint and oxfmt clean.

Error tracking shows no draw-time canvas exceptions on web analytics pages over 30 days, which is consistent with context loss (it throws nothing).

Known gap: Safari never fires `contextrestored` (Firefox 125+ and Chrome 99+ do), so a context lost there still stays blank. `isContextLost()` polling would be the lever and isn't implemented.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs. Updated the charts `AGENTS.md` sizing section (the DOM-vs-canvas split is what made this diagnosable) and `docs/TESTING.md` for the new accessor field.

## 🤖 Agent context

Written by Claude (Claude Code). I couldn't reproduce the bug, so I traced the render pipeline instead — the DOM/CSS height chain from the web vitals tile down to the chart wrapper, then every path in the charts core that can leave the canvas blank while overlays keep rendering. The screenshot ruled out the alternatives: a sizing bug wouldn't place the goal lines correctly, and a mid-draw throw would leave the grid visible since it's drawn before the series layer.

This went through two review rounds via `/sp-ship`, and the second round changed the shape of the PR substantially. The first version also shipped an `isDrawableDimensions` gate that skipped draw frames whose plot box was degenerate. Two reviewers independently blocked it: it had no loop-level test (deleting it left the suite green), it had no re-arm path so a chart could latch permanently blank — the exact symptom being fixed — and the sizing doc justifying it described frame retention that the resize path already wipes. It also made two other new guards unreachable. Sam called the cut, and it's dropped along with a `clearAndPrepare` dpr fallback that only defended a sub-half-pixel container.

The `withVerticalClip` clamp was initially dropped with it, then re-landed: the gate was the *reason* it looked unreachable, and without the gate it defends a state a narrow multi-axis chart can actually reach. The second round also caught that the restore handler zeroed `canvas.width` to trigger the resize path, which destroyed the bitmap before confirming it could be repainted — under memory pressure, when `getContext` is most likely to fail, that reintroduced this bug through its own fix. Hence `updateSize(forceRepaint)`.

Also rejected along the way: a `surfaceGeneration` counter threaded through four files to signal the repaint (unnecessary — the existing wipe-forces-republish path already covered it); coalescing the static draw against a render storm (cancel-and-re-arm always yields a paint on the next frame, so it starves nothing, and the test I wrote for it passed against the old code too); and a reviewer claim that Firefox lacks `contextlost`/`contextrestored`, which would have undercut the whole fix — it shipped in Firefox 125.

Unrelated but worth someone's attention: `TypeError: error loading dynamically imported module .../TrendsLineChart-*.js`, ~48 occurrences in 30 days, all Firefox. `TrendsLineChart` is lazy-loaded behind `<Suspense>`, so that's a chunk-load failure, probably stale hashed chunks after deploy.

Skills invoked: `/writing-tests`, `/sp-ship` (which drove `sp-swarm` and `sp-triage`).

---

### DIFF
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -56,6 +56,10 @@ Charts fill their container and need a parent with real dimensions — a `0`-hei
 A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing — so a chart squeezed to nothing keeps its last drawn frame until the container is usable again, rather than going blank while its DOM axis labels stay put.
 Growing the container back repaints normally.
 
+Overlays (axis labels, axis titles, reference/goal lines, value labels, legend, tooltip) are DOM, positioned from the scales; the grid, axis lines, tick marks, and the series themselves are canvas.
+So "labels and goal lines render but the plot is empty" means the canvas specifically failed, not that the chart is mis-sized — the layout was fine or the overlays wouldn't have placed correctly.
+The two ways that happens are a bitmap discarded without a repaint (a resize wiping `canvas.width`, or a lost-and-restored 2D context) and a draw that painted nothing; both are handled in `useChartCanvas` / `useChartDraw`.
+
 ## Composition
 
 ```tsx
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -55,8 +55,10 @@ Charts fill their container and need a parent with real dimensions — a `0`-hei
 
 Overlays (axis labels, axis titles, reference/goal lines, value labels, legend, tooltip) are DOM, positioned from the scales; the grid, axis lines, tick marks, and the series themselves are canvas.
 That split is the first thing to check when a chart looks broken: if the labels and goal lines render in the right places but the plot area is empty, the layout is fine and the canvas alone failed.
-The bitmap is only ever discarded by `syncCanvasSize` reallocating it (a real resize) or by the browser losing the 2D context, and `useChartCanvas` repaints after both — a blank plot under correct overlays means one of those two paths didn't schedule its repaint.
+Check the cheap causes first: `theme.skipDraw` suppresses all canvas painting by design, and a `drawStatic` that throws leaves the frame it was clearing empty.
+Beyond those, the bitmap is discarded either by `syncCanvasSize` reallocating it (a real resize) or by the browser losing the 2D context, and `useChartCanvas` repaints after both — so a blank plot under correct overlays means one of those paths didn't schedule its repaint.
 Note the browser does not report a lost 2D context everywhere: Firefox 125+ and Chrome 99+ fire `contextrestored`, Safari doesn't fire it at all, so a context lost on Safari stays blank.
+The backing store is rounded, not truncated, to device pixels, so a fractional CSS width at `dpr > 1` can shift the drawing scale by a fraction of a percent versus the raw multiply.
 
 ## Composition
 
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -53,12 +53,10 @@ const series: Series[] = [
 
 Charts fill their container and need a parent with real dimensions — a `0`-height flex child renders nothing. Give the wrapper an explicit height (`h-64`, `flex-1` in a sized column). Sparkline alone takes `height`/`width` props.
 
-A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing. When the container itself doesn't resize (only `margins` shrinking the plot box) the canvas bitmap is untouched, so the chart keeps its last drawn frame rather than going blank while its DOM axis labels stay put. A container resize is different: `syncCanvasSize` reallocates and wipes the bitmap for the new rect before the draw loop's skip check ever runs, so a chart resized down to zero plot area shows an empty canvas, not its last frame, until it's resized back to something drawable.
-Growing the container back repaints normally.
-
 Overlays (axis labels, axis titles, reference/goal lines, value labels, legend, tooltip) are DOM, positioned from the scales; the grid, axis lines, tick marks, and the series themselves are canvas.
-So "labels and goal lines render but the plot is empty" means the canvas specifically failed or the container was resized below the axis margins — not that the chart is otherwise mis-sized.
-The two ways the canvas can fail while sized correctly are a bitmap discarded without a repaint (a lost-and-restored 2D context) and a draw that painted nothing; both are handled in `useChartCanvas` / `useChartDraw`.
+That split is the first thing to check when a chart looks broken: if the labels and goal lines render in the right places but the plot area is empty, the layout is fine and the canvas alone failed.
+The bitmap is only ever discarded by `syncCanvasSize` reallocating it (a real resize) or by the browser losing the 2D context, and `useChartCanvas` repaints after both — a blank plot under correct overlays means one of those two paths didn't schedule its repaint.
+Note the browser does not report a lost 2D context everywhere: Firefox 125+ and Chrome 99+ fire `contextrestored`, Safari doesn't fire it at all, so a context lost on Safari stays blank.
 
 ## Composition
 
@@ -102,7 +100,7 @@ The two ways the canvas can fail while sized correctly are a bitmap discarded wi
 
 Import test helpers from `@posthog/quill-charts/testing` (jsdom-only).
 
-- `getHogChart(scope?)` reads rendered overlays via stable `data-attr` hooks — `yTicks()`, `xTicks()`, `valueLabels()`, `referenceLines()`, axis titles, `seriesCount`, etc. Its `waitForTooltip()` returns a structured `TooltipSnapshot` but only when the chart was mounted via `renderHogChart` (it needs the captured tooltip context).
+- `getHogChart(scope?)` reads rendered overlays via stable `data-attr` hooks — `yTicks()`, `xTicks()`, `valueLabels()`, `referenceLines()`, axis titles, `seriesCount`, etc. Its `waitForTooltip()` returns a structured `TooltipSnapshot` but only when the chart was mounted via `renderHogChart` (it needs the captured tooltip context). `canvas` is the static (data) canvas, for asserting the backing store or dispatching a canvas-level event — use it instead of `container.querySelector('canvas')`.
 - For a chart rendered outside `renderHogChart` (e.g. a real product scene), read the tooltip from the DOM: `getHogChartTooltip()` / `waitForHogChartTooltip()` return the portal element; `createHogChartTooltip(el)` wraps it as `{ element, isPinned }`. If the chart renders the built-in `DefaultTooltip`, use `createDefaultTooltipAccessor(el)` for `label()`, `rows()`, `value(seriesLabel)`, `swatchColors()`, and `total()` — it reads `DefaultTooltip`'s `data-attr="hog-chart-tooltip-*"` hooks (a stable testing contract; keep the accessor and the component's attrs in sync).
 
 ## Maintenance
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -57,8 +57,7 @@ Overlays (axis labels, axis titles, reference/goal lines, value labels, legend, 
 That split is the first thing to check when a chart looks broken: if the labels and goal lines render in the right places but the plot area is empty, the layout is fine and the canvas alone failed.
 Check the cheap causes first: `theme.skipDraw` suppresses all canvas painting by design, and a `drawStatic` that throws leaves the frame it was clearing empty.
 Beyond those, the bitmap is discarded either by `syncCanvasSize` reallocating it (a real resize) or by the browser losing the 2D context, and `useChartCanvas` repaints after both — so a blank plot under correct overlays means one of those paths didn't schedule its repaint.
-Note the browser does not report a lost 2D context everywhere: Firefox 125+ and Chrome 99+ fire `contextrestored`, Safari doesn't fire it at all, so a context lost on Safari stays blank.
-The backing store is rounded, not truncated, to device pixels, so a fractional CSS width at `dpr > 1` can shift the drawing scale by a fraction of a percent versus the raw multiply.
+Safari is the exception: it never fires `contextrestored`, so a context lost there stays blank.
 
 ## Composition
 
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -53,6 +53,9 @@ const series: Series[] = [
 
 Charts fill their container and need a parent with real dimensions — a `0`-height flex child renders nothing. Give the wrapper an explicit height (`h-64`, `flex-1` in a sized column). Sparkline alone takes `height`/`width` props.
 
+A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing — so a chart squeezed to nothing keeps its last drawn frame until the container is usable again, rather than going blank while its DOM axis labels stay put.
+Growing the container back repaints normally.
+
 ## Composition
 
 ```tsx
--- a/packages/quill/packages/charts/AGENTS.md
+++ b/packages/quill/packages/charts/AGENTS.md
@@ -53,12 +53,12 @@ const series: Series[] = [
 
 Charts fill their container and need a parent with real dimensions — a `0`-height flex child renders nothing. Give the wrapper an explicit height (`h-64`, `flex-1` in a sized column). Sparkline alone takes `height`/`width` props.
 
-A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing — so a chart squeezed to nothing keeps its last drawn frame until the container is usable again, rather than going blank while its DOM axis labels stay put.
+A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing. When the container itself doesn't resize (only `margins` shrinking the plot box) the canvas bitmap is untouched, so the chart keeps its last drawn frame rather than going blank while its DOM axis labels stay put. A container resize is different: `syncCanvasSize` reallocates and wipes the bitmap for the new rect before the draw loop's skip check ever runs, so a chart resized down to zero plot area shows an empty canvas, not its last frame, until it's resized back to something drawable.
 Growing the container back repaints normally.
 
 Overlays (axis labels, axis titles, reference/goal lines, value labels, legend, tooltip) are DOM, positioned from the scales; the grid, axis lines, tick marks, and the series themselves are canvas.
-So "labels and goal lines render but the plot is empty" means the canvas specifically failed, not that the chart is mis-sized — the layout was fine or the overlays wouldn't have placed correctly.
-The two ways that happens are a bitmap discarded without a repaint (a resize wiping `canvas.width`, or a lost-and-restored 2D context) and a draw that painted nothing; both are handled in `useChartCanvas` / `useChartDraw`.
+So "labels and goal lines render but the plot is empty" means the canvas specifically failed or the container was resized below the axis margins — not that the chart is otherwise mis-sized.
+The two ways the canvas can fail while sized correctly are a bitmap discarded without a repaint (a lost-and-restored 2D context) and a draw that painted nothing; both are handled in `useChartCanvas` / `useChartDraw`.
 
 ## Composition
 
--- a/packages/quill/packages/charts/src/core/canvas-renderer.ts
+++ b/packages/quill/packages/charts/src/core/canvas-renderer.ts
@@ -922,9 +922,19 @@ export function withVerticalClip(
     // Matches drawAxes' snapping: the axis line's 1px column starts at round(plotLeft), so trimming
     // there leaves the stroke flush against the axis line.
     const left = clipLeft ? Math.round(dimensions.plotLeft) : 0
+    const clipWidth = dimensions.width - left
+    // `useChartMargins` grows the left margin from measured label widths with a floor but no ceiling
+    // against the container, so a narrow chart with stacked y-axis gutters can reserve more than it
+    // has. That makes the rect negative, which canvas reads as a reversed rectangle sitting entirely
+    // off the right edge — clipping to it would discard the whole series layer while the DOM axis
+    // labels still render. Draw unclipped rather than invisibly. `!(> 0)` also bails on a NaN width.
+    if (!(clipWidth > 0)) {
+        draw()
+        return
+    }
     ctx.save()
     ctx.beginPath()
-    ctx.rect(left, dimensions.plotTop - pad, dimensions.width - left, dimensions.plotHeight + pad * 2)
+    ctx.rect(left, dimensions.plotTop - pad, clipWidth, dimensions.plotHeight + pad * 2)
     ctx.clip()
     try {
         draw()
--- a/packages/quill/packages/charts/src/core/canvas-renderer.ts
+++ b/packages/quill/packages/charts/src/core/canvas-renderer.ts
@@ -922,9 +922,18 @@ export function withVerticalClip(
     // Matches drawAxes' snapping: the axis line's 1px column starts at round(plotLeft), so trimming
     // there leaves the stroke flush against the axis line.
     const left = clipLeft ? Math.round(dimensions.plotLeft) : 0
+    const clipWidth = dimensions.width - left
+    // A left margin wider than the container (very narrow charts, where the reserved y-tick column
+    // outgrows the available width) makes the rect negative. Canvas reads that as a reversed
+    // rectangle sitting entirely off-canvas, so clipping to it would discard the whole plot — draw
+    // unclipped rather than invisibly. Phrased as `!(> 0)` so a non-finite width bails too.
+    if (!(clipWidth > 0)) {
+        draw()
+        return
+    }
     ctx.save()
     ctx.beginPath()
-    ctx.rect(left, dimensions.plotTop - pad, dimensions.width - left, dimensions.plotHeight + pad * 2)
+    ctx.rect(left, dimensions.plotTop - pad, clipWidth, dimensions.plotHeight + pad * 2)
     ctx.clip()
     try {
         draw()
--- a/packages/quill/packages/charts/src/core/canvas-renderer.ts
+++ b/packages/quill/packages/charts/src/core/canvas-renderer.ts
@@ -922,18 +922,9 @@ export function withVerticalClip(
     // Matches drawAxes' snapping: the axis line's 1px column starts at round(plotLeft), so trimming
     // there leaves the stroke flush against the axis line.
     const left = clipLeft ? Math.round(dimensions.plotLeft) : 0
-    const clipWidth = dimensions.width - left
-    // A left margin wider than the container (very narrow charts, where the reserved y-tick column
-    // outgrows the available width) makes the rect negative. Canvas reads that as a reversed
-    // rectangle sitting entirely off-canvas, so clipping to it would discard the whole plot — draw
-    // unclipped rather than invisibly. Phrased as `!(> 0)` so a non-finite width bails too.
-    if (!(clipWidth > 0)) {
-        draw()
-        return
-    }
     ctx.save()
     ctx.beginPath()
-    ctx.rect(left, dimensions.plotTop - pad, clipWidth, dimensions.plotHeight + pad * 2)
+    ctx.rect(left, dimensions.plotTop - pad, dimensions.width - left, dimensions.plotHeight + pad * 2)
     ctx.clip()
     try {
         draw()
--- a/packages/quill/packages/charts/src/core/canvas-size.test.ts
+++ b/packages/quill/packages/charts/src/core/canvas-size.test.ts
@@ -1,6 +1,6 @@
-import { dimensions, mockRect } from '../../testing'
+import { dimensions, mockRect } from '../testing'
 import { buildDimensions, sameDimensions, syncCanvasSize } from './canvas-size'
-import { DEFAULT_MARGINS } from './useChartMargins'
+import { DEFAULT_MARGINS } from './hooks/useChartMargins'
 
 /** Records assignments to an integer canvas dimension, which jsdom would otherwise apply silently. */
 function trackWrites(canvas: HTMLCanvasElement, prop: 'width' | 'height', writes: string[]): void {
--- a/packages/quill/packages/charts/src/core/canvas-size.ts
+++ b/packages/quill/packages/charts/src/core/canvas-size.ts
@@ -1,4 +1,4 @@
-import type { ChartDimensions, ChartMargins } from '../types'
+import type { ChartDimensions, ChartMargins } from './types'
 
 /** The subset of `DOMRect` the canvas sizing needs. */
 export interface SizeRect {
--- a/packages/quill/packages/charts/src/core/hooks/canvas-size.ts
+++ b/packages/quill/packages/charts/src/core/hooks/canvas-size.ts
@@ -69,18 +69,3 @@ export function sameDimensions(a: ChartDimensions, b: ChartDimensions): boolean 
         a.plotHeight === b.plotHeight
     )
 }
-
-/** Whether a frame drawn against these dimensions can produce a picture. Canvas silently no-ops
- *  on non-finite coordinates, and a zero-or-negative plot box clips the series away — either way
- *  the draw loop would clear the canvas and paint nothing, so callers skip the frame and leave the
- *  last good one on screen instead. */
-export function isDrawableDimensions(dimensions: ChartDimensions): boolean {
-    return (
-        Number.isFinite(dimensions.width) &&
-        Number.isFinite(dimensions.height) &&
-        Number.isFinite(dimensions.plotLeft) &&
-        Number.isFinite(dimensions.plotTop) &&
-        dimensions.plotWidth > 0 &&
-        dimensions.plotHeight > 0
-    )
-}
--- a/packages/quill/packages/charts/src/core/hooks/canvas-size.ts
+++ b/packages/quill/packages/charts/src/core/hooks/canvas-size.ts
@@ -1,0 +1,86 @@
+import type { ChartDimensions, ChartMargins } from '../types'
+
+/** The subset of `DOMRect` the canvas sizing needs. */
+export interface SizeRect {
+    width: number
+    height: number
+}
+
+/** `canvas.width`/`height` are integers, so a fractional target has to be rounded before it's
+ *  compared or assigned — otherwise it never equals the stored value and every sync reads as a
+ *  change. */
+function backingSize(cssSize: number, dpr: number): number {
+    return Math.round(cssSize * dpr)
+}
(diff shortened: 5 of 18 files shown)


packages/quill/packages/charts/AGENTS.md
```diff
@@ -53,6 +53,13 @@ const series: Series[] = [
 
 Charts fill their container and need a parent with real dimensions — a `0`-height flex child renders nothing. Give the wrapper an explicit height (`h-64`, `flex-1` in a sized column). Sparkline alone takes `height`/`width` props.
 
+A container too small to hold the axis margins leaves the plot area at zero, and the draw loops skip such a frame instead of clearing the canvas and painting nothing — so a chart squeezed to nothing keeps its last drawn frame until the container is usable again, rather than going blank while its DOM axis labels stay put.
```
sampennington: 🤖 **sp-ship** · 💻📊📐 `[code+hog-charts+xp]` · **Blocker**

This documents behaviour the code doesn't deliver, and it's the stated justification for the skip in `useChartDraw`. On the path described — a container squeezed below the axis margins — the shrink arrives via `ResizeObserver`, `syncCanvasSize` reallocates the backing store and wipes the bitmap, and *only then* does the draw loop skip. There is no last frame left to keep. It holds only for a margins-only change at a constant rect.

That makes the next paragraph ("labels and goal lines render but the plot is empty means the canvas specifically failed, not that the chart is mis-sized") actively misleading: post-PR, a mis-sized chart is one of the ways to produce exactly that symptom. Either make the code match the doc (skip `syncCanvasSize` too when the incoming rect isn't drawable, leaving the old bitmap intact under the `overflow: hidden` wrapper) or narrow the guide to what actually happens.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
```diff
@@ -56,25 +57,31 @@ export function useChartDraw({
             cancelAnimationFrame(staticRafRef.current)
             staticRafRef.current = null
         }
-        if (!ctx || !dimensions || !scales || theme.skipDraw) {
+        // A frame that can't produce a picture is skipped rather than drawn: clearing first and
+        // painting nothing replaces the last good frame with a blank canvas, while the DOM axis
+        // overlays keep rendering — the chart then reads as "chrome but no data".
+        if (!ctx || !dimensions || !scales || theme.skipDraw || !isDrawableDimensions(dimensions)) {
```
sampennington: 🤖 **sp-ship** · 🧪📐 `[test+xp]` · **Blocker**

This gate has no re-arm path and no loop-level test, which is the riskiest possible shape of change for a bug whose symptom is *nothing painted*.

Untested: `isDrawableDimensions` is referenced only by `canvas-size.test.ts` and the two draw loops, and every chart test renders through `ensureJsdom`, which pins `getBoundingClientRect` to 800x400 so the predicate is always `true`. Delete `|| !isDrawableDimensions(dimensions)` from both loops and the whole suite still passes.

Latchable: once a publish lands with `plotWidth <= 0` both loops return early, and only another dimensions change will ever draw again. A narrow chart where the measured y-tick column plus `margins.right` exceeds the width gives `plotWidth = 0`; margins then stabilise, no resize follows, and the chart is latched blank with the DOM overlays still rendering — precisely the screenshot being debugged. Previously that frame drew grid and axis lines into a zero-width plot.

Add the test the docs promise ("growing the container back repaints normally"): `renderHook(useChartDraw)` with `{width: 40, height: 400, plotWidth: 0, ...}`, assert `drawStatic` *and* `ctx.clearRect` were not called, then re-render drawable and assert it paints.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+            series={SERIES}
+            labels={LABELS}
+            theme={THEME}
+            createScales={createScales}
+            drawStatic={drawStatic}
+            drawHover={() => false}
+        />
+    )
+    const canvas = container.querySelector('canvas')
+    if (!canvas) {
```
sampennington: 🤖 **sp-ship** · 🦔📊♻️ `[personal+hog-charts+reuse]` · **Blocker**

Two problems in one helper.

Branching logic in a test file: `const canvas = container.querySelector('canvas')` / `if (!canvas) { throw ... }`. No `if` in test bodies or test helpers — narrow with `!` or use a query that throws itself.

And `docs/TESTING.md:176` blocks this exact pattern: "Don't fall back to `container.querySelector('canvas')` for canvas presence. `renderHogChart` already throws when the canvas is missing." The reason it was forked is real — `HogChart` exposes no canvas element and `findCanvas` is private in `accessor.ts` — so the fix is to add it to the shared surface (a `canvas` getter on `HogChart`, or a `dispatchOnChartCanvas(chart, event)` helper), not to inline a one-off.

packages/quill/packages/charts/src/core/canvas-renderer.ts
```diff
@@ -922,9 +922,18 @@ export function withVerticalClip(
     // Matches drawAxes' snapping: the axis line's 1px column starts at round(plotLeft), so trimming
     // there leaves the stroke flush against the axis line.
     const left = clipLeft ? Math.round(dimensions.plotLeft) : 0
+    const clipWidth = dimensions.width - left
+    // A left margin wider than the container (very narrow charts, where the reserved y-tick column
+    // outgrows the available width) makes the rect negative. Canvas reads that as a reversed
+    // rectangle sitting entirely off-canvas, so clipping to it would discard the whole plot — draw
+    // unclipped rather than invisibly. Phrased as `!(> 0)` so a non-finite width bails too.
+    if (!(clipWidth > 0)) {
```
sampennington, line 931: 🤖 **sp-ship** · 📊📐🧪 `[hog-charts+xp+test]` · **Suggestion / follow-up**

This branch can't execute from its only call path. `drawLineSeriesLayer` runs inside the draw loops, which now gate on `isDrawableDimensions`, requiring `plotWidth > 0` i.e. `width > left + right`; reaching `round(plotLeft) > width` would need `margins.right < 0.5px`. So the comment asserts a state that can't occur, which will mislead the next reader, and it's untested either way (`grep withVerticalClip` finds only the definition and one call site).

Pick one place to own "the plot box is sane". Either drop this and let the gate own the invariant, or keep it and re-word to "defensive; unreachable while the draw loops gate on `isDrawableDimensions`".

packages/quill/packages/charts/src/core/hooks/clearCanvas.ts
```diff
@@ -9,7 +9,10 @@ import type { ChartDimensions } from '../types'
  *  headless export that resizes the viewport under device_scale_factor=2); reading it again here
  *  would scale the transform inconsistently with the canvas and magnify/clip the drawing. */
 export function clearAndPrepare(ctx: CanvasRenderingContext2D, dimensions: ChartDimensions): void {
-    const dpr = dimensions.width > 0 ? ctx.canvas.width / dimensions.width : window.devicePixelRatio || 1
+    // A sub-pixel wrapper rounds the backing store down to 0, which would derive a zero scale and
+    // collapse every coordinate onto a point (and stop `clearRect` from clearing anything).
+    const backingRatio = ctx.canvas.width / dimensions.width
+    const dpr = backingRatio > 0 && Number.isFinite(backingRatio) ? backingRatio : window.devicePixelRatio || 1
```
sampennington: 🤖 **sp-ship** · 🧪📐📊 `[test+xp+hog-charts]` · **Suggestion / follow-up**

Same unreachability, plus a coverage gap. The scenario in the comment (a sub-pixel wrapper rounding the backing store to 0) needs `dimensions.width < 0.5`, which `isDrawableDimensions` now rejects upstream, and the old `dimensions.width > 0` guard already covered `0/0 → NaN`.

Note the existing test `falls back to window.devicePixelRatio when dimensions have no width` uses `makeCtx(0)` with `dims(0)`, so it passed before this change too — the genuinely new `backingRatio === 0` branch (`makeCtx(0)` with `dims(0.4)`) is untested. Either revert this and keep the diff honest about what it changes, or add that case.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
```diff
@@ -68,24 +51,35 @@ export function useChartCanvas(options: UseChartCanvasOptions): UseChartCanvasRe
...
+            const next = buildDimensions(rect, marginsRef.current)
+            setCanvasState((prev) =>
+                prev &&
+                !staticWiped &&
+                !overlayWiped &&
+                prev.ctx === context &&
+                prev.overlayCtx === overlayContext &&
+                sameDimensions(prev.dimensions, next)
+                    ? prev
+                    : { ctx: context, overlayCtx: overlayContext, dimensions: next }
```
sampennington, line 86: 🤖 **sp-ship** · 📊 `[hog-charts]` · **Suggestion / follow-up**

The repaint after a wipe rests entirely on this literal allocating a *referentially* new `dimensions`, because the draw effects key on `[ctx, dimensions, scales, ...]`, not on the state object. In both the dpr-change and `contextrestored` cases every value is identical and `sameDimensions` is true — only the reference changed. Nothing here says so.

A future "avoid invalidating downstream memos" tidy-up that reuses `prev.dimensions` when the values are equal (the natural mirror of the bail-out three lines up) silently reinstates the blank canvas, and only the one `contextrestored` test would catch it. Make the mechanism explicit in the comment, or carry an explicit repaint token in `CanvasState`.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
```diff
@@ -95,13 +89,26 @@ export function useChartCanvas(options: UseChartCanvasOptions): UseChartCanvasRe
         })
         observer.observe(wrapper)
 
+        // A 2D context can be lost when the browser reclaims canvas memory. Its bitmap comes back
+        // blank on `contextrestored` and nothing repaints on its own, so the canvas stays empty
+        // while every DOM overlay (axis labels, goal lines, tooltip) keeps rendering against valid
+        // dimensions. Zeroing the backing size makes `updateSize` treat it as a real resize, which
+        // repaints. `contextlost` is deliberately not handled: preventing its default tells the
+        // browser we'll restore the context ourselves, and then it never restores.
+        const onContextRestored = (event: Event): void => {
```
sampennington: 🤖 **sp-ship** · ✨📐💻 `[quality+xp+code]` · **Suggestion / follow-up**

Signalling "repaint needed" by zeroing the backing width so `updateSize` misreads it as a resize is a hidden coupling to `canvas-size.ts` — nothing there hints that its `canvas.width !== width` compare is load-bearing for context restore. If someone later caches the last synced size instead of reading the DOM (a natural optimisation) this fix dies silently and its test goes green for the wrong reason. An explicit `updateSize({ forceRepaint: true })` states the intent and survives refactoring.

Minor while you're in there: `event.currentTarget as HTMLCanvasElement` asserts away `EventTarget | null` and depends on `currentTarget` still being populated; a per-canvas closure drops the cast and makes it obvious which canvas is zeroed.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
```diff
@@ -68,24 +51,35 @@ export function useChartCanvas(options: UseChartCanvasOptions): UseChartCanvasRe
...
             if (!context || !overlayContext) {
                 return
             }
 
-            setCanvasState({
-                ctx: context,
-                overlayCtx: overlayContext,
-                dimensions: buildDimensions(rect, marginsRef.current),
-            })
+            const rect = wrapper.getBoundingClientRect()
```
sampennington, line 64: 🤖 **sp-ship** · ⚡ `[efficiency]` · **Suggestion / follow-up**

The observer callback discards `entries` and re-measures with `getBoundingClientRect()`. Layout is clean inside the callback, so this is free for one chart — but `updateSize` then writes `canvas.style.width/height` on two canvases, dirtying layout for the next observer in the same delivery batch. On a dashboard with N charts a real resize frame costs ~N-1 forced layout flushes interleaved with the style writes, per frame, for the duration of the drag.

Read the size from the entry (`entry.borderBoxSize[0]`, falling back to `contentRect`) in the observer path and keep `getBoundingClientRect()` for the synchronous first measure and the restore path.

sampennington (commented):

Nine reviewers, eight returned (react errored mid-response and was dropped). Three blockers, all clustered on the same soft spot: the `isDrawableDimensions` gate. It has no loop-level test — delete it from both draw loops and the suite still passes — it can latch a chart blank with no re-arm path, and the `AGENTS.md` paragraph justifying it describes frame retention the resize path already wipes, so the guide now hands the next debugger the wrong heuristic. The third is a test helper that branches and forks the harness `docs/TESTING.md` explicitly blocks.

Consensus on the rest: `syncCanvasSize`, resolving the contexts before resizing, and the two `try`/`finally` additions are the load-bearing, well-evidenced fixes and read cleanly. The clip-rect clamp and the dpr fallback defend states the gate now makes unreachable — worth dropping rather than testing unreachable branches. Two things that didn't anchor to a line: `contextrestored` does land on the reported browser (Firefox 125+, Chrome 99+) but Safari has no support at all, so a lost context there is still permanently blank and `isContextLost()` polling is the only lever; and `Sparkline`'s `fill` mode with `LINE_MARGINS` reserves 12px vertically, so a sparkline in a flex slot at or below that now draws nothing where it previously drew a flat line — `MetricCard`'s `sparklineFill` is the surface to sanity-check.

packages/quill/packages/charts/src/core/canvas-size.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            ['a fractional css size', 507.328125, 311.5, 1],
+            ['a fractional css size at dpr 2', 507.328125, 311.5, 2],
+        ])('leaves the backing store alone on a repeat sync with %s', (_, width, height, dpr) => {
+            const canvas = document.createElement('canvas')
+
+            expect(syncCanvasSize(canvas, { width, height }, dpr)).toBe(true)
+            // Assigning canvas.width/height wipes the bitmap even when the value is identical, and
+            // the repaint only lands on the next animation frame — so a no-op sync must not touch
+            // it, or a container that keeps reporting resizes never gets to show a drawn frame.
+            expect(syncCanvasSize(canvas, { width, height }, dpr)).toBe(false)
```
sampennington, line 32: 🤖 **sp-ship** · 🧪📊 `[test+hog-charts]` · **Suggestion / follow-up**

The test is named "leaves the backing store alone" but asserts only the returned boolean, not the property the comment describes. A plausible tidy-up — `const resized = canvas.width !== width || ...; canvas.width = width; canvas.height = height; return resized` — returns identical booleans, re-introduces the unconditional wipe this PR exists to remove, and passes unchanged. Shadow the setter and assert nothing was written.

Also: these rows assert `width`/`style.width` but never `height`/`style.height`, and `syncCanvasSize` guards the two axes in independent `if` blocks — a regression reintroduced on the height branch would slip through, since the return is `true` if *either* axis changed.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
```diff
@@ -56,25 +57,31 @@ export function useChartDraw({
...
+                    scales,
+                    series,
+                    labels,
+                    hoverIndex: -1,
+                    hoverPosition: null,
+                    theme,
+                    hoverProgress: 1,
+                    resetHoverFade: () => 1,
+                })
+            } finally {
```
sampennington, line 78: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

The `try`/`finally` is untested, and it matters more than it looks: a `drawStatic` that throws after `clearAndPrepare`'s `ctx.save()` used to leak an unbalanced save, so every later frame inherited the wrong transform and clip — another route to a permanently blank canvas, i.e. the same bug class. Cheap to cover: render with a throwing `drawStatic`, assert `ctx.restore` ran, then re-render with a working one and assert the next frame draws (the throw propagates synchronously under `setupSyncRaf`, so wrap in `expect(...).toThrow()`).

packages/quill/packages/charts/src/core/hooks/useHoverAnimation.ts
```diff
@@ -70,7 +71,7 @@ export function useHoverAnimation({
             cancelAnimationFrame(hoverRafRef.current)
             hoverRafRef.current = null
         }
-        if (!overlayCtx || !dimensions || !scales || themeRef.current.skipDraw) {
+        if (!overlayCtx || !dimensions || !scales || themeRef.current.skipDraw || !isDrawableDimensions(dimensions)) {
```
sampennington: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion / follow-up**

This guard is now byte-identical to `useChartDraw.ts:63` (`!ctx || !dimensions || !scales || skipDraw || !isDrawableDimensions(dimensions)`). Since the PR already adds a shared module for this concern, a `canPaint({ ctx, dimensions, scales, skipDraw })` there would stop the two loops drifting.

packages/quill/packages/charts/src/core/canvas-size.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (canvas.style.width !== cssWidth) {
+        canvas.style.width = cssWidth
+    }
+    if (canvas.style.height !== cssHeight) {
+        canvas.style.height = cssHeight
+    }
+    return resized
+}
+
+export function buildDimensions(rect: SizeRect, margins: ChartMargins): ChartDimensions {
```
sampennington, line 51: 🤖 **sp-ship** · ♻️ `[reuse]` · **Suggestion / follow-up**

Now that `buildDimensions` is a plain non-hook export, the hand-written copy of the same computation in `src/testing/jsdom.ts:4-11` is redundant — it spells out `plotWidth: 800 - DEFAULT_MARGINS.left - DEFAULT_MARGINS.right` by hand, which is exactly `buildDimensions(mockRect, DEFAULT_MARGINS)`. `jsdom.ts` already imports `DEFAULT_MARGINS` and this module imports only types, so there's no cycle risk. Collapsing it removes the last place the margin math is written twice, which is the reason this module was extracted.

packages/quill/packages/charts/src/core/hooks/canvas-size.ts
```diff
@@ -0,0 +1,86 @@
...
+        a.width === b.width &&
+        a.height === b.height &&
+        a.plotLeft === b.plotLeft &&
+        a.plotTop === b.plotTop &&
+        a.plotWidth === b.plotWidth &&
+        a.plotHeight === b.plotHeight
+    )
+}
+
+/** Whether a frame drawn against these dimensions can produce a picture. Canvas silently no-ops
```
sampennington: 🤖 **sp-ship** · 🦔✨ `[personal+quality]` · **Suggestion / follow-up**

The "skipping beats clearing-and-painting-nothing" rationale is written out four times (here, `useChartDraw.ts:60`, `useChartCanvas.test.tsx:45`, `AGENTS.md:56`) and the "assigning `canvas.width` wipes the bitmap" rationale three times (lines 20-27 here, `canvas-size.test.ts:16`, `useChartCanvas.ts:92`). Each copy can drift, and finding 1 above is what that looks like in practice — the `AGENTS.md` copy is already wrong.

One home per fact (the helper's doc), with call sites naming the helper instead of re-arguing it. Same theme: `SizeRect`'s doc restates a two-field interface, and `backingSize` carries three lines of doc over a one-line body.

packages/quill/packages/charts/AGENTS.md
sampennington: 🤖 **sp-ship**

Fixed in `63f8169a` — corrected the doc. The "keeps its last frame" claim only holds for a margins-only change at a constant rect; a real container resize wipes the bitmap via `syncCanvasSize` before the draw loop's `isDrawableDimensions` gate runs, so a squeezed *container* shows an empty canvas, not the last frame. Narrowed the "canvas specifically failed" framing to account for that case too.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
sampennington: 🤖 **sp-ship**

Deferring — this is a judgement call about production behavior, not a mechanical fix. Adding a re-arm path (or removing the gate) changes what the chart does when it latches blank, and a loop-level regression test needs to pin down the intended recovery semantics first (does it need another dimensions change, or should it retry every frame?). That's a design decision for a human, not something to guess at in triage.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.test.tsx
sampennington: 🤖 **sp-ship**

Partially fixed in `63f8169a` — dropped the `if (!canvas) throw` branch, narrowed with `!` to match the no-if-in-test-helpers convention used elsewhere in this package (e.g. `Legend.test.tsx`, `Sparkline.test.tsx`).

Left the second half open: adding a `canvas` getter (or `dispatchOnChartCanvas`) to the shared `HogChart` testing surface changes a contract used across every chart test file, which is a wider design call than a triage-scoped fix should make unilaterally. Deferring that part to a human — the fork stays local to this file for now.

packages/quill/packages/charts/src/core/canvas-renderer.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — agreed the branch is currently unreachable while the draw loops gate on `isDrawableDimensions`, but picking which place owns the invariant (drop the defensive branch vs. re-word it) is a design call for the PR author, not a mechanical fix.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — making the reference-identity repaint mechanism explicit (a comment vs. an explicit repaint token on `CanvasState`) is a design decision, not a single-file mechanical fix.

packages/quill/packages/charts/src/core/hooks/clearCanvas.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — same unreachability question as the sibling thread on `canvas-renderer.ts:930`; deciding whether to revert the now-redundant guard or add the missing `backingRatio === 0` test case is a design/coverage call for the PR author.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — reading size from the `ResizeObserverEntry` instead of re-measuring via `getBoundingClientRect()` is a real optimization but changes the observer callback's data source, which needs verification against the entry's `borderBoxSize`/`contentRect` fallback behavior across browsers rather than a blind swap.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — swapping the implicit `canvas.width = 0` signal for an explicit `updateSize({ forceRepaint: true })` option changes a hook's public shape, which is a design call for the PR author rather than a mechanical fix.

packages/quill/packages/charts/src/core/canvas-size.test.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — agreed the test name promises more than it asserts, but choosing between shadowing the `canvas.width`/`height` setters and adding height-branch assertions is a test-design call the PR author should make.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — this is a real test gap (an untested `try`/`finally` guarding against unbalanced `ctx.save()`), but writing a throwing-`drawStatic` test with a synchronous-RAF assertion is a substantive test addition, not a mechanical single-file fix.

packages/quill/packages/charts/src/core/canvas-size.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — collapsing `src/testing/jsdom.ts`'s hand-written margin math into `buildDimensions(mockRect, DEFAULT_MARGINS)` is a reasonable dedup but touches the shared testing surface, which the PR author should own.

packages/quill/packages/charts/src/core/hooks/useHoverAnimation.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — extracting the duplicated guard into a shared `canPaint(...)` helper is a reasonable follow-up but touches both `useChartDraw.ts` and `useHoverAnimation.ts`, which is broader than a single localized edit.

packages/quill/packages/charts/src/core/hooks/canvas-size.ts
sampennington: 🤖 **sp-ship**

Out of scope for this triage pass — consolidating the repeated "skip vs. clear-and-paint-nothing" and "assigning canvas.width wipes the bitmap" rationale into one doc home spans four files; a broader documentation consolidation for the PR author to take on, not a single-file fix.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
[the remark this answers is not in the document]
sampennington: 🤖 **sp-ship**

Deferring — this is the same root issue as the sp-swarm blocker on `useChartDraw.ts:63` (the `isDrawableDimensions` gate's re-arm/latch behavior), which was deferred to a human as a production-behavior judgement call rather than auto-fixed. Resolving this one would require the same design decision, so leaving it open alongside that thread.

sampennington:
🤖 **sp-ship** · 💬 **Suggestions only**

Second review pass (code, test, hog-charts, personal — scoped, since the increment was mostly deletions plus test changes). No blockers this time, and the three from the first pass are closed by deletion rather than by testing unreachable code: `isDrawableDimensions` is gone repo-wide, and `withVerticalClip` / `clearAndPrepare` were verified byte-identical to master. The reviewers traced every degenerate `dimensions` to the end of the pipeline and confirmed the worst case is still a blank frame, never a throw or a latch, and mutation-tested the strengthened tests to confirm they actually bite.

Two findings were worth acting on and are fixed in 2075215. The restore handler was zeroing `canvas.width` before confirming it could repaint, so a failed `getContext` under memory pressure — exactly when a context is lost — would have left a 0-size backing store with nothing scheduled to repaint it, reintroducing this bug through its own fix; it now signals `updateSize(forceRepaint)` instead. And the `withVerticalClip` clamp turned out to be reachable *because* the gate was removed: `useChartMargins` reserves a left gutter with no ceiling against the container, so a narrow multi-axis chart can produce a negative clip rect that discards the whole series layer. Re-landed, with tests.

The rest were suggestions and are also applied: `buildDimensions` is pinned against literals (it's the only clamp left, and the shared jsdom fixture now derives from it), an assertion that couldn't fail is gone, comments that restated production docstrings are trimmed, the sizing guide now names `theme.skipDraw` and a throwing `drawStatic` ahead of context loss, and `chart.canvas` is documented in `docs/TESTING.md`. One reviewer claim I checked and rejected: that Firefox lacks `contextlost`/`contextrestored` — it shipped in Firefox 125, Chrome 99; Safari is the real gap and the guide now says so.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
sampennington: 🤖 **sp-ship**

Fixed in 2075215 (cut landed in eb1ca9a). The gate is deleted from both draw loops and from `canvas-size.ts` — `grep isDrawableDimensions` is empty repo-wide — and the sizing paragraph that justified it is gone with it. Both loops now always clear and repaint, so there is no early return left to latch on.

packages/quill/packages/charts/src/core/hooks/useChartCanvas.test.tsx
sampennington: 🤖 **sp-ship**

Fixed in eb1ca9a. `canvas` is now part of the shared `HogChart` accessor, the test goes through `renderHogChart` + `chart.canvas`, and the `if`/throw branch is gone. Documented in `docs/TESTING.md` in 2075215, including the caveat that absence assertions still need a raw query since the accessor throws when there is no canvas.

packages/quill/packages/charts/src/core/hooks/useChartDraw.ts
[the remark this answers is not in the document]
sampennington: 🤖 **sp-ship**

Addressed in eb1ca9a by removing the early return entirely rather than trying to make it safe. Your read was correct and was one of the reasons for dropping it: retaining stale pixels under freshly positioned DOM overlays is the same failure this PR is about. Both draw loops now clear and repaint on every dimensions change.

packages/quill/packages/charts/AGENTS.md
```diff
@@ -53,6 +53,13 @@ const series: Series[] = [
...
+That split is the first thing to check when a chart looks broken: if the labels and goal lines render in the right places but the plot area is empty, the layout is fine and the canvas alone failed.
+Check the cheap causes first: `theme.skipDraw` suppresses all canvas painting by design, and a `drawStatic` that throws leaves the frame it was clearing empty.
+Beyond those, the bitmap is discarded either by `syncCanvasSize` reallocating it (a real resize) or by the browser losing the 2D context, and `useChartCanvas` repaints after both — so a blank plot under correct overlays means one of those paths didn't schedule its repaint.
+Note the browser does not report a lost 2D context everywhere: Firefox 125+ and Chrome 99+ fire `contextrestored`, Safari doesn't fire it at all, so a context lost on Safari stays blank.
+The backing store is rounded, not truncated, to device pixels, so a fractional CSS width at `dpr > 1` can shift the drawing scale by a fraction of a percent versus the raw multiply.
```
gantoine: these last 2 lines necessary?

gantoine (approved):

packages/quill/packages/charts/AGENTS.md
sampennington: Mostly not, agreed — dropped both.

The device-pixel rounding line was pure trivia (a fraction-of-a-percent scale shift nobody will ever debug from), and the *why* for rounding is already on `backingSize` in `canvas-size.ts` where it belongs.

I kept the Safari fact but compressed it to one short sentence, because dropping it outright would leave the paragraph above actively wrong — it says `useChartCanvas` repaints after a lost context, and on Safari it never gets the chance. That is a distinct terminal outcome (permanently blank, no re-arm) rather than "a path forgot to schedule a repaint", so it is worth the one line:

> Safari is the exception: it never fires `contextrestored`, so a context lost there stays blank.

Also moved `canvas-size.ts` out of `core/hooks/` into `core/` while I was in here — it exports three pure functions, no hook, and it was the odd kebab-case file sitting among the `use*` ones.

PR #73883, opened 2026-07-27
fix(insights): show renamed series name in trends legend

sampennington:
## Problem

On a trends line chart, the new in-chart quill legend shows the raw event name instead of a renamed series name — so two series that filter the same event on different properties are indistinguishable.

The legend derives each series' label from the query *result*'s `action`, which honors `custom_name` but not a series `name`. `series_event()` returns the raw event, so `action.name` is the event and `action.custom_name` only reflects `series.custom_name`. A series renamed via the query editor or API (which sets `name`) never reaches the legend. The legacy legend rendered labels from the frontend query *definition* (which carries `name`), so this regressed once the quill in-chart legend became the default.

Raised in a Slack support thread (linked below).

## Changes

Surface a series' display override into the result `action.custom_name`: use `custom_name` when set, otherwise a `name` that differs from the raw event/action label (a `name` echoing the event is not a rename and is ignored). This mirrors the frontend label resolver and restores parity now that the quill legend reads from the query result. Applied in both the aggregate and time-series result branches of the trends query runner.

Added a parameterized regression test (`test_trends_series_custom_name`) covering: `custom_name` set, `name` set, `custom_name` winning over `name`, a `name` echoing the event (no rename → `None`), and no override. It guards the reported regression — a revert to reading only `series.custom_name` fails the `name` case.

The sandbox has no Postgres/ClickHouse, so I couldn't execute the DB-backed suite here; the cases collect correctly and the `series_custom_name` resolution logic was verified standalone against the real schema nodes.

## 🤖 Agent context

Diagnosed and authored by the PostHog Slack app. Traced the legend label path (`getTrendsSeriesDisplayLabel` → `getDisplayNameFromEntityFilter`) to the backend result builder, and found the `name`-vs-`custom_name` gap by inspecting the affected insight's series config. Chose a backend fix (one resolver used by both result branches) over re-plumbing series nodes through every chart, so legend and tooltip stay consistent across surfaces. Invoked the `/writing-tests` skill before adding the test.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1784901266426799?thread_ts=1784901266.426799&cid=C0ACRAMJUAG)*

### DIFF
--- a/posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
+++ b/posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
@@ -456,11 +456,9 @@ class TestTrendsQueryRunner(ClickhouseTestMixin, APIBaseTest):
 
     @parameterized.expand(
         [
-            # (series, expected action.custom_name) — the legend/tooltip render `custom_name` over the raw event.
             ("custom_name set via rename modal", EventsNode(event="$pageview", custom_name="Renamed"), "Renamed"),
             ("name set via query editor / API", EventsNode(event="$pageview", name="Renamed"), "Renamed"),
             ("custom_name wins over name", EventsNode(event="$pageview", name="A", custom_name="B"), "B"),
-            # A `name` echoing the event is not a rename — must not leak into custom_name.
             ("name echoing the event is not a rename", EventsNode(event="$pageview", name="$pageview"), None),
             ("no override", EventsNode(event="$pageview"), None),
         ]
--- a/posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
+++ b/posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
@@ -454,6 +454,31 @@ class TestTrendsQueryRunner(ClickhouseTestMixin, APIBaseTest):
 
         self.assertEqual("$pageview", response.results[0]["label"])
 
+    @parameterized.expand(
+        [
+            # (series, expected action.custom_name) — the legend/tooltip render `custom_name` over the raw event.
+            ("custom_name set via rename modal", EventsNode(event="$pageview", custom_name="Renamed"), "Renamed"),
+            ("name set via query editor / API", EventsNode(event="$pageview", name="Renamed"), "Renamed"),
+            ("custom_name wins over name", EventsNode(event="$pageview", name="A", custom_name="B"), "B"),
+            # A `name` echoing the event is not a rename — must not leak into custom_name.
+            ("name echoing the event is not a rename", EventsNode(event="$pageview", name="$pageview"), None),
+            ("no override", EventsNode(event="$pageview"), None),
+        ]
+    )
+    def test_trends_series_custom_name(self, _name, series, expected_custom_name):
+        self._create_test_events()
+
+        response = self._run_trends_query(
+            self.default_date_from,
+            self.default_date_to,
+            IntervalType.DAY,
+            [series],
+            None,
+            None,
+        )
+
+        self.assertEqual(expected_custom_name, response.results[0]["action"]["custom_name"])
+
     def test_trends_count(self):
         self._create_test_events()
 
--- a/posthog/hogql_queries/insights/trends/trends_query_runner.py
+++ b/posthog/hogql_queries/insights/trends/trends_query_runner.py
@@ -73,7 +73,7 @@ from posthog.hogql_queries.insights.utils.breakdowns import (
     has_breakdown_filter,
 )
 from posthog.hogql_queries.insights.utils.utils import get_response_hogql
-from posthog.hogql_queries.query_runner import AnalyticsQueryRunner
+from posthog.hogql_queries.query_runner import AnalyticsQueryRunner, resolve_series_custom_name
 from posthog.hogql_queries.utils.formula_ast import FormulaAST
 from posthog.hogql_queries.utils.query_compare_to_date_range import QueryCompareToDateRange
 from posthog.hogql_queries.utils.query_date_range import QueryDateRange
@@ -608,7 +608,7 @@ class TrendsQueryRunner(AnalyticsQueryRunner[TrendsQueryResponse]):
                         "type": "events",
                         "order": series.series_order,
                         "name": series_label or "All events",
-                        "custom_name": self.series_custom_name(series.series, series_label),
+                        "custom_name": resolve_series_custom_name(series.series, series_label),
                         "math": series.series.math,
                         "math_property": series.series.math_property,
                         "math_hogql": series.series.math_hogql,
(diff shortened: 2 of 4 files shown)


posthog/hogql_queries/insights/trends/trends_query_runner.py
```diff
@@ -842,6 +842,20 @@ def query_previous_date_range(self):
             exact_timerange=self.exact_timerange,
         )
 
+    def series_custom_name(
+        self, series: Union[EventsNode, ActionsNode, DataWarehouseNode, GroupNode], series_label: str | None
+    ) -> str | None:
+        # A series' display override. `custom_name` is set by the in-app "Rename graph series" modal;
+        # a `name` that differs from the event/action label is a rename applied via the query editor or
+        # API. Either should win over the raw event name in legends and tooltips.
+        custom_name = getattr(series, "custom_name", None)
```
sampennington: 🤖 **sp-ship** · 💻📐 `[code+xp]` · **Suggestion / follow-up**

All four node types in the `Union` (Events/Actions/DataWarehouse/Group) statically declare `custom_name` and `name` as `str | None`, so the `getattr(..., None)` defaults can never fire — and `getattr` collapses the return to `Any`, against CLAUDE.md's mypy `--strict` / no-`Any` rule. The sibling `series_event` right below uses direct attribute access. Consider `series.custom_name` / `series.name`.

posthog/hogql_queries/insights/trends/trends_query_runner.py
```diff
@@ -842,6 +842,20 @@ def query_previous_date_range(self):
             exact_timerange=self.exact_timerange,
         )
 
+    def series_custom_name(
```
sampennington: 🤖 **sp-ship** · ✨ `[quality]` · **Suggestion / follow-up**

`series_custom_name` names only the first branch, but the method also returns a differing `name`. It maps to the `action.custom_name` field it populates, so it's defensible — optional rename to e.g. `series_display_name` if you want the name to cover both cases. The comment already carries the intent.

posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
```diff
@@ -454,6 +454,31 @@ def test_trends_label(self):
...
+            # (series, expected action.custom_name) — the legend/tooltip render `custom_name` over the raw event.
+            ("custom_name set via rename modal", EventsNode(event="$pageview", custom_name="Renamed"), "Renamed"),
+            ("name set via query editor / API", EventsNode(event="$pageview", name="Renamed"), "Renamed"),
+            ("custom_name wins over name", EventsNode(event="$pageview", name="A", custom_name="B"), "B"),
+            # A `name` echoing the event is not a rename — must not leak into custom_name.
+            ("name echoing the event is not a rename", EventsNode(event="$pageview", name="$pageview"), None),
+            ("no override", EventsNode(event="$pageview"), None),
+        ]
+    )
+    def test_trends_series_custom_name(self, _name, series, expected_custom_name):
```
sampennington, line 466: 🤖 **sp-ship** · 💻🧪📐 `[code+test+xp]` · **Suggestion / follow-up**

The cases only cover `EventsNode`, where `series_label` is always the raw event. For `ActionsNode`/`DataWarehouseNode`, `series_label` is the resolved action name / table name — the more realistic input for the "name echoing the label is not a rename" no-op. Consider one `ActionsNode` row exercising the same branches. Non-blocking (all types share the `series_custom_name` path).

posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py
```diff
@@ -454,6 +454,31 @@ def test_trends_label(self):
 
         self.assertEqual("$pageview", response.results[0]["label"])
 
+    @parameterized.expand(
+        [
+            # (series, expected action.custom_name) — the legend/tooltip render `custom_name` over the raw event.
+            ("custom_name set via rename modal", EventsNode(event="$pageview", custom_name="Renamed"), "Renamed"),
+            ("name set via query editor / API", EventsNode(event="$pageview", name="Renamed"), "Renamed"),
+            ("custom_name wins over name", EventsNode(event="$pageview", name="A", custom_name="B"), "B"),
+            # A `name` echoing the event is not a rename — must not leak into custom_name.
```
sampennington: 🤖 **sp-ship** · 🦔 `[personal]` · **Suggestion / follow-up**

This comment narrates the test case directly below it, whose description string already says "name echoing the event is not a rename". Suggest deleting. (Same for the tuple-shape half of the comment on line 459 — the `why custom_name` clause is the useful part.)

sampennington (commented):

No blockers — the fix is correct and well-scoped: both result branches are updated, `series_label` is the right per-type label to compare against, and the frontend consumer (`getDisplayNameFromEntityFilter`) honors `custom_name` with the same precedence, so no downstream consumer regresses. The convergent notes are that the `getattr(..., None)` defensiveness is superfluous (every node type in the Union — including `GroupNode` — declares both fields, and the sibling `series_event` uses direct access), and the regression test only exercises `EventsNode` so the per-type `name != series_label` comparison isn't covered for actions/warehouse series.

sampennington:
🤖 **sp-ship** · triage

Actioned in 1c03275:
- 💻📐 `getattr(..., None)` → direct `series.custom_name` / `series.name` (all Union types, incl. `GroupNode`, declare both fields; mypy-clean and matches sibling `series_event`).
- 🦔 removed the two test comments that restated the cases below them.

Deferred (with reasoning):
- ✨ method name `series_custom_name` — kept as-is: it maps to the `action.custom_name` field it populates, and the docstring carries the "also a differing `name`" case.
- 🧪 `ActionsNode`/`DataWarehouseNode` test case — `series_custom_name` is a type-agnostic string comparison and the `EventsNode` cases already cover all its branches; the per-type label resolution lives in `series_event` (separately covered), so an extra node-type row would mostly re-test the same path. Verified manually that an `ActionsNode` rename resolves correctly.

posthog/hogql_queries/insights/trends/trends_query_runner.py
```diff
@@ -608,7 +608,7 @@ def get_value(name: str, val: Any):
                         "type": "events",
```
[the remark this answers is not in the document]
sampennington: 🤖 **sp-ship** · triage — good catch, confirmed and fixed in 7ccaf7c. `apply_series_custom_names` re-patched cached responses from `series.custom_name` only, overwriting the `name`-based rename on every cached read. Both paths now share `resolve_series_custom_name`; the cache normalizer honors `name` only for TrendsQuery (matching the fresh path), so stickiness/lifecycle stay in sync. Added `applies_name_based_rename_from_cache` + `name_echoing_the_event_is_not_a_rename` cases to TestApplySeriesCustomNames.

PR #73893, opened 2026-07-27
feat(conversations): apply saved ticket views server-side via ?view= param

luke-belton:
## Problem

Saved ticket views ("Ticket Views") store their filters as an opaque JSON blob that only the frontend knows how to interpret. Opening a view in the app works because `supportTicketsSceneLogic` unpacks the blob and re-encodes it into snake_case query params, but nothing outside the frontend can do that. In practice:

- An agent using the MCP tools can list saved views (`conversations-views-list` returns `short_id` + `name`) but has no way to get the tickets a view contains.
- The filter shape existed in three hand-maintained vocabularies (the camelCase saved-view blob, the snake_case list-endpoint params, and the URL-param encoding), with the camelCase-to-snake_case translation living only in frontend TypeScript, free to drift from the backend's filtering.

## Changes

`GET /conversations/tickets/?view=<short_id>` now applies a saved view's filters server-side, and the filter shape is a single backend-owned contract.

```mermaid
flowchart LR
    View[(TicketView.filters<br/>raw JSON)] --> Ser{{TicketViewFiltersSerializer<br/>canonical shape}}
    Params[flat query params<br/>status, tags, order_by, ...] --> Adapter[query_params_to_view_filters]
    Ser --> Apply{{apply_ticket_filters<br/>one filters-to-ORM implementation}}
    Adapter --> Apply
    Apply --> QS[filtered, ordered queryset]
    Ser -. hogli build:openapi .-> TS[generated TS + MCP schemas]
    class Ser,Apply phBlue;
    class Params,View phGray;
    class QS,TS phYellow;
```

- **New `products/conversations/backend/api/ticket_filters.py`**: `TicketViewFiltersSerializer` is the source of truth for the saved-view filter shape (all fields optional, `help_text` on everything, legacy single-value `assignee` normalized). `apply_ticket_filters` is the only place filter values become ORM predicates, and `query_params_to_view_filters` adapts the existing flat snake_case params into the same canonical shape.
- **`TicketViewSet.safely_get_queryset`** now resolves `?view=<short_id>` (team-scoped, 400 with a clear message on unknown or invalid views), overlays explicit query params on top of the view's values so a caller can apply a view and narrow further, and funnels everything through the single applier. The `me` assignee token still resolves per-caller, so a shared view scoped to "me" means each caller's own tickets.
- **`TicketView.filters` writes are validated** against the canonical shape via a `JSONField` subclass (`@extend_schema_field`), while storing and returning the raw dict so unknown keys and legacy blobs survive round-trips unchanged. The 10KB size guard stays.
- **Frontend** deletes the hand-written `TicketViewFilters` interface and derives it from the generated `TicketViewFiltersApi` (with `assignee` widened locally for the legacy single-value shape old rows can still return). No behavior change.
- **MCP**: the `conversations-tickets-list` tool now documents the `view` param and the views-list to tickets-list workflow; no new tool needed. Also documented the previously-missing `ai_triage_result` and `snoozed` params in the OpenAPI schema.

> [!NOTE]
> No breaking change. The flat query params keep working exactly as before (behavior-preserving refactor, verified by the existing 201-test suite), and existing saved-view blobs continue to validate and render unchanged.

Automated (all run locally by the agent):

- Full `products/conversations/backend/api/tests/` ticket + views suites: 223 passed, including all pre-existing filter tests against the refactored path (behavior-preservation guard).
- New `TestTicketViewParamFilter` (6 tests): `?view=` equals the equivalent flat params, explicit param overrides the view, unknown/other-team `short_id` returns 400, `me` resolves to the requesting user, and a legacy blob (`'all'` sentinels, single-value assignee, unknown keys) applies without error and respects sorting. Each catches a regression no existing test did, since the `view` param is new.
- New `TestTicketViewFiltersValidation` (no-DB `SimpleTestCase`, parameterized): the canonical shape accepts everything the frontend saves today (defaults, legacy assignee, unknown keys) and rejects invalid enum values and oversized payloads. Guards against tightening validation in a way that breaks existing saved views.
- New endpoint wiring guards: invalid filters rejected with 400 on create, unknown keys survive round-trips.
- Frontend: `supportTicketsSceneLogic.test.ts` (17 passed), `typescript:check` clean for conversations files.
- Repo-wide `uv run mypy --cache-fine-grained .` clean, `hogli ci:preflight --fix` green, `hogli build:openapi` clean after adding the `TicketPriorityEnum` name override.

Not done manually: I did not click through the app UI; the frontend change is type-only.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The `view` param is documented in the OpenAPI schema and MCP tool descriptions, which are the generated public surfaces for this API.

## 🤖 Agent context

Authored with PostHog Code (Claude). Skills invoked while producing this PR: `/improving-drf-endpoints`, `/writing-tests`, `/adopting-generated-api-types`, `/implementing-mcp-tools`, `/writing-code-comments`.

Decisions along the way: a first design added a backend-only camelCase-to-snake_case translator, rejected as fragile (two hand-mirrored copies of the same mapping, the exact failure mode session-recording playlists live with). Prior art review (insights `query`, `DashboardFilter`, error tracking's `PropertyGroupFilterValue`) pointed to "one validated shape + one applier" instead. The shape lives in a DRF serializer rather than the `frontend/src/queries/schema` pipeline because that pipeline is explicitly for frontend-authored query types, and a serializer gives the MCP tool schema for free. A full wire-format unification (dropping the flat params for a filter object) was considered and deliberately deferred: it would break existing API/MCP consumers and the flat params are the better ergonomics for agents anyway.

---

### DIFF
--- a/products/conversations/backend/api/ticket_filters.py
+++ b/products/conversations/backend/api/ticket_filters.py
@@ -91,7 +91,18 @@ def normalize_assignee_filter(value: Any) -> list[Any]:
     }
 )
 class TicketViewAssigneeFilterField(serializers.Field):
+    """Lenient by default so legacy stored blobs still apply (an unrecognized entry is
+    dropped, and the legacy 'all' sentinel means no filter). Writes pass
+    context={"strict_writes": True}: silently dropping a malformed entry there would
+    save a view that quietly matches more assignees than the caller asked for."""
+
     def to_internal_value(self, data: Any) -> list[Any]:
+        if self.context.get("strict_writes"):
+            entries = data if isinstance(data, list) else [data]
+            if any(not _is_assignee_entry(entry) and entry != "all" for entry in entries):
+                raise serializers.ValidationError(
+                    "Each assignee entry must be 'me', 'unassigned', or an object with type ('user' or 'role') and id."
+                )
         return normalize_assignee_filter(data)
 
     def to_representation(self, value: Any) -> Any:
--- a/products/conversations/backend/api/ticket_filters.py
+++ b/products/conversations/backend/api/ticket_filters.py
@@ -16,7 +16,7 @@ from collections.abc import Mapping
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any
 
-from django.db.models import CharField, Exists, OuterRef, Q, QuerySet
+from django.db.models import CharField, F, OrderBy, Q, QuerySet
 from django.db.models.functions import Cast
 from django.utils import timezone
 
@@ -47,8 +47,19 @@ AI_TRIAGE_FILTER_VALUES = [
     "in_progress",
 ]
 
-ALLOWED_ORDER_COLUMNS = {"updated_at", "sla_due_at", "snoozed_until", "created_at", "ticket_number"}
-DEFAULT_ORDER_BY = "-updated_at"
+ALLOWED_ORDER_COLUMNS = ("updated_at", "sla_due_at", "snoozed_until", "created_at", "ticket_number")
+
+VALID_STATUS_VALUES = frozenset(s.value for s in Status)
+VALID_PRIORITY_VALUES = frozenset(p.value for p in Priority)
+VALID_CHANNEL_VALUES = frozenset(c.value for c in Channel)
+
+# Named choice sets for ChoiceFields below, registered in ENUM_NAME_OVERRIDES
+# (posthog/settings/web.py) so drf-spectacular doesn't mint generic globals like
+# ChannelEnum/SlaEnum/OrderEnum in the shared OpenAPI namespace.
+TICKET_CHANNEL_FILTER_CHOICES = [*(c.value for c in Channel), "all"]
+TICKET_SLA_FILTER_CHOICES = [*SLA_FILTER_VALUES, "all"]
+TICKET_TAGS_MATCH_CHOICES = ["any", "all"]
+TICKET_SORT_ORDER_CHOICES = [1, -1]
 
 
 def _is_assignee_entry(value: Any) -> bool:
@@ -111,10 +122,10 @@ class TicketViewAssigneeFilterField(serializers.Field):
 
 class TicketViewSortingSerializer(serializers.Serializer):
     columnKey = serializers.CharField(
-        help_text="Ticket column to sort by (updated_at, sla_due_at, snoozed_until, created_at, ticket_number). "
+        help_text=f"Ticket column to sort by ({', '.join(ALLOWED_ORDER_COLUMNS)}). "
         "Unknown columns fall back to updated_at."
     )
-    order = serializers.ChoiceField(choices=[1, -1], help_text="1 for ascending, -1 for descending.")
+    order = serializers.ChoiceField(choices=TICKET_SORT_ORDER_CHOICES, help_text="1 for ascending, -1 for descending.")
 
 
 class TicketViewFiltersSerializer(serializers.Serializer):
@@ -132,12 +143,12 @@ class TicketViewFiltersSerializer(serializers.Serializer):
         help_text="Ticket priorities to include. Empty or omitted means all priorities.",
     )
     channel = serializers.ChoiceField(
-        choices=[*(c.value for c in Channel), "all"],
+        choices=TICKET_CHANNEL_FILTER_CHOICES,
         required=False,
         help_text="Channel the ticket originated from. 'all' disables the filter.",
     )
     sla = serializers.ChoiceField(
-        choices=[*SLA_FILTER_VALUES, "all"],
+        choices=TICKET_SLA_FILTER_CHOICES,
         required=False,
         help_text="SLA state: 'breached' is past due, 'at-risk' is due within the next hour, "
         "'on-track' has more than an hour remaining. 'all' disables the filter.",
@@ -159,7 +170,7 @@ class TicketViewFiltersSerializer(serializers.Serializer):
         help_text="Tag names to match, combined according to tagsMatch.",
     )
     tagsMatch = serializers.ChoiceField(
-        choices=["any", "all"],
+        choices=TICKET_TAGS_MATCH_CHOICES,
         required=False,
         help_text="'any' returns tickets with at least one of tags (OR); 'all' requires every tag (AND).",
     )
@@ -189,6 +200,21 @@ class TicketViewFiltersSerializer(serializers.Serializer):
         help_text="Free-text search. A numeric value matches a ticket number exactly; otherwise matches "
         "the customer's name or email, the email subject, or message content.",
     )
+
+
+def parse_stored_view_filters(stored: Any) -> dict[str, Any]:
+    """Leniently validate a stored TicketView.filters blob into the canonical shape.
+
+    Stored blobs predate write validation, so a legacy value must not make the view
+    unusable: offending keys are dropped and the rest still apply."""
+    if not isinstance(stored, dict):
+        return {}
+    serializer = TicketViewFiltersSerializer(data=stored)
+    if not serializer.is_valid():
+        cleaned = {key: value for key, value in stored.items() if key not in serializer.errors}
+        serializer = TicketViewFiltersSerializer(data=cleaned)
+        serializer.is_valid()
+    return dict(serializer.validated_data)
 
 
 def _decode_assignee_param(raw: str) -> list[Any]:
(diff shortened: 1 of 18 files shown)


products/conversations/backend/api/ticket_filters.py
```diff
@@ -0,0 +1,434 @@
...
+        },
+        "description": (
+            "Assignees to match (any of): 'unassigned', 'me' (resolved to the requesting user), "
+            "or an object with type ('user' or 'role') and id."
+        ),
+    }
+)
+class TicketViewAssigneeFilterField(serializers.Field):
+    def to_internal_value(self, data: Any) -> list[Any]:
+        return normalize_assignee_filter(data)
```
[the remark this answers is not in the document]
luke-belton: Addressed the Greptile finding in https://github.com/PostHog/posthog/commit/7cec2f9ef7a6418c98623b488b4fe3f61dea6d3c. Writes to a view's filters now reject malformed assignee entries with a 400 instead of silently dropping them, so a saved view can no longer quietly match more assignees than the caller asked for. The ?view= read path stays lenient on purpose: legacy stored blobs (the 'all' sentinel, unrecognized entries) still normalize away and apply, since 400ing an old view on read would break it retroactively. Covered by new parameterized serializer tests for both modes plus an endpoint wiring guard.

products/conversations/backend/api/tickets.py
```diff
@@ -494,120 +417,30 @@ def safely_get_queryset(self, queryset: QuerySet) -> QuerySet:
...
 
-        # ticket_number is unique per team (the queryset is already team-scoped), so it breaks
-        # ties deterministically. Without it, rows equal on the primary key — every no-SLA
-        # ticket shares NULL sla_due_at — have no stable order across the separate LIMIT/OFFSET
-        # page queries, so pages overlap or drop rows and the sort looks lost past page 1.
-        return queryset.order_by(primary, "-ticket_number")
+        user = cast("User", self.request.user) if self.request.user and self.request.user.is_authenticated else None
+        return apply_ticket_filters(queryset, filters, team=self.team, user=user)
+
+    def _get_view_filters(self, short_id: str) -> dict[str, Any]:
```
veryayskiy, line 440: _get_view_filters 400s the whole list request when a stored blob fails validation (:442). Those blobs were never validated before this PR, so a legacy value makes an existing view permanently unusable via ?view=. Be lenient here like TicketViewAssigneeFilterField is on read (ticket_filters.py:99-112) and drop the offending keys.

products/conversations/backend/api/ticket_filters.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+def query_params_to_view_filters(params: Mapping[str, str]) -> dict[str, Any]:
+    """Translate the flat snake_case ticket-list query params into the canonical
+    TicketViewFilters shape. Only keys whose param is present are set, so values coming
+    from a saved view survive unless explicitly overridden. Invalid values are dropped
+    rather than rejected, matching the endpoint's long-standing behavior."""
+    filters: dict[str, Any] = {}
+
+    status_param = params.get("status")
```
veryayskiy, line 299: status, priority, and assignee set their key even when every value parsed away, so ?view=X&status=bogus or ?view=X&assignee=user:abc drops the view's filter and widens the result set. Guard on the parsed list like aiTriageResult (:248-252) and tags (:258-263) already do.

products/conversations/backend/api/ticket_views.py
```diff
@@ -17,23 +19,35 @@
...
+    """Validates writes against the canonical filter shape but stores and returns the raw
+    dict, so unknown keys survive round-trips and legacy blobs render unchanged."""
+
+    def to_internal_value(self, data: Any) -> dict:
+        value = super().to_internal_value(data)
+        if not isinstance(value, dict):
+            raise serializers.ValidationError("Expected a JSON object.")
+        if len(json.dumps(value)) > MAX_FILTERS_SIZE_BYTES:
+            raise serializers.ValidationError("Filters payload is too large.")
+        filters_serializer = TicketViewFiltersSerializer(data=value, context={"strict_writes": True})
```
veryayskiy, line 33: strict write validation rejects saves the UI can produce: SupportTicketsScene.tsx:297-304 has no maxLength, currentFilters puts search in the blob (supportTicketsSceneLogic.ts:683), and search is capped at 200 in ticket_filters.py:185-191, so saving a view after pasting a long search returns 400 with a raw DRF error. Clamp search on write or cap the input.

products/conversations/backend/api/ticket_filters.py
```diff
@@ -0,0 +1,467 @@
...
+        child=serializers.ChoiceField(choices=Status.choices),
+        required=False,
+        help_text="Ticket statuses to include. Empty or omitted means all statuses.",
+    )
+    priority = serializers.ListField(
+        child=serializers.ChoiceField(choices=Priority.choices),
+        required=False,
+        help_text="Ticket priorities to include. Empty or omitted means all priorities.",
+    )
+    channel = serializers.ChoiceField(
```
veryayskiy, line 163: these inline ChoiceField choice sets mint generic globals in the shared OpenAPI/MCP namespace (ChannelEnum with an all sentinel the real Channel doesn't have, plus SlaEnum, TagsMatchEnum, OrderEnum). Lift each choice list to a module constant and register ENUM_NAME_OVERRIDES entries next to posthog/settings/web.py:571 so the next serializer with a channel or order field doesn't collide.

products/conversations/backend/api/tickets.py
luke-belton: Fixed in 26033e66e2a — `_get_view_filters` now drops just the offending keys from a legacy blob and applies the rest, so an old view stays usable via `?view=`. If somehow nothing survives cleaning it applies no view filters rather than 400ing. Covered by a test with a blob mixing one invalid key and one valid one.

products/conversations/backend/api/ticket_filters.py
luke-belton: Fixed in 26033e66e2a — status, priority, and assignee now only set their key when at least one value survives parsing, matching the aiTriageResult/tags guards, so `?view=X&status=bogus` keeps the view's saved filter instead of widening. Locked in with an API test plus a parameterized unit test over the three params.

products/conversations/backend/api/ticket_filters.py
luke-belton: Fixed in 26033e66e2a - lifted the four choice sets to module constants (`TICKET_CHANNEL_FILTER_CHOICES` etc.) and registered ticket-scoped names for them in ENUM_NAME_OVERRIDES next to the existing TicketStatusEnum/TicketPriorityEnum entries.

products/conversations/backend/api/ticket_views.py
luke-belton: Fixed in a842470b482 — two halves: the search input now has maxLength={200} matching the backend's MAX_SEARCH_LENGTH (which the list endpoint has always enforced by ignoring longer searches, so no capability is lost), and the `?view=` read path drops an overlong stored search instead of 400ing (falls out of the lenient-read fix on the other thread).

veryayskiy:
1. ticket_filters.py:545-550 — tagsAll is in the saved-view contract but the frontend can't apply it (supportTicketsSceneLogic.ts:495-510, SavedViewsModal.tsx:34-39), so a view created via API/MCP with tagsAll behaves differently through ?view= than in the app.
2. ticket_filters.py:579-593 — parse_stored_view_filters drops a whole key on any child error, so status: ["open", "bogus_legacy"] widens to all statuses instead of keeping open.
3. Confirm the dropped self._filter_queryset_by_access_level(queryset) call is intentional (harmless, since posthog/api/routing.py:337 applies it).

veryayskiy (approved):
left a comment. stamping to not hold

luke-belton:
All three addressed in 362de536910:

1. **tagsAll removed from the saved-view contract** — it's now param-only: `tags_all=` still composes with `tags=` on the list endpoint, but the serializer no longer accepts `tagsAll` on view writes, and a stored blob carrying one is ignored on `?view=` — matching what the app does. Locked in with a test that a stored `tagsAll` key doesn't filter.
2. **Per-entry salvage in parse_stored_view_filters** — an errored list field now keeps its individually-valid entries, so `status: ["open", "bogus_legacy"]` narrows to `open` instead of dropping the key and widening to all statuses. Whole-key dropping remains only for scalar fields and lists with no valid entries.
3. **Access-level call restored** — the dropped `_filter_queryset_by_access_level` was a rebase casualty, not intentional. You're right that it's redundant (routing.py applies it after `safely_get_queryset`, and the #70081 test suite passes either way), but I've restored it so this PR doesn't silently diverge from master.

268 backend tests pass locally. The CI bot's regenerated types on the previous push also confirm the enum overrides now apply (TicketChannelFilterEnum etc. replaced the generic names).

PR #73900, opened 2026-07-27
fix(slack_app): migrate user-link sign-in to Slack OpenID Connect

VojtechBartos:
## Problem

Slack Marketplace review blocks submission while the app requests the legacy Sign in with Slack user scopes (`identity.basic`, `identity.email`) - Slack retired that protocol. Our Slack user-link flow (binding a Slack identity to a PostHog user so @mentions resolve without email matching) was built on it.

## Changes

The flow only ever needed identity - the resolver matches links on `slack_user_id` + `slack_team_id` and the stored user token is never read after linking - so this is a protocol swap, not a rebuild:

```diff
- https://slack.com/oauth/v2/authorize?user_scope=identity.basic,identity.email&scope=
+ https://slack.com/openid/connect/authorize?scope=openid+email+profile&response_type=code

- POST oauth.v2.access  →  users.identity
+ POST openid.connect.token  →  openid.connect.userInfo
```

Both calls now go through `slack_sdk`. Views, state signing, CSRF/org guards, and link storage are unchanged - OIDC claims map 1:1 onto the existing `SlackIdentity` fields, and existing links keep working (same keys). Also updated the manifest snippet in the internal Slack local setup guide.

> [!NOTE]
> The Slack app config on api.slack.com needs its user scopes swapped to `openid`, `email`, `profile` before resubmitting (redirect URLs stay the same).

- Updated the authorize-URL test - catches a revert to the legacy endpoint/scope shape, which is what Marketplace review rejects
- New `TestExchangeCode` units (mocking `WebClient`) - catch a wrong OIDC claim key or an unwrapped `SlackApiError` 500-ing the callback; this path was previously only mocked at the view layer
- Both affected suites pass locally (39 tests); ruff and repo-wide mypy clean for changed files
- Manually tested on local

## 🤖 Agent context

Claude Code authored this from the Marketplace rejection message. Investigation confirmed the persisted user token is write-only, which made the swap safe without touching views or storage. Skills invoked: /writing-tests.

### DIFF
--- a/docs/internal/slack-local-setup-guide.md
+++ b/docs/internal/slack-local-setup-guide.md
@@ -118,8 +118,9 @@ oauth_config:
       - users:read
       - users:read.email
     user:
-      - identity.basic
-      - identity.email
+      - openid
+      - email
+      - profile
 settings:
   event_subscriptions:
     request_url: https://<you>-posthog.ngrok.dev/slack/event-callback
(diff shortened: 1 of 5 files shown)


tatoalo (approved):

Twixes (commented):
Didn't run, but knowing OAuth and a bit of OIDC, this seems all correct to me

PR #73902, opened 2026-07-27
fix(flags): match replay usage by JSONB containment instead of integer cast

jakesciotto:
## Problem

The feature flags `bulk_update_tags` endpoint returned an unhandled 500 whenever a team in the project stored a non-integer `id` in its `session_recording_linked_flag` JSON.

The flag queryset annotated `is_used_in_replay_settings` by casting `session_recording_linked_flag -> 'id'` to an integer. That is a jsonb-to-integer cast, so Postgres errors on any stored id that isn't a JSON number (even a numeric-looking string like `"123"`, or `null`) with `cannot cast jsonb string to type integer`. The annotation ran on every action, and because the EXISTS scopes by project, one team's bad row broke the endpoint for every sibling team in the project. The same data also 500s the flag list endpoint.

**Why:** raised from an error-tracking signal on this endpoint. The endpoints should tolerate a non-integer replay-linked-flag id rather than crash, and `validate_session_recording_linked_flag` only checks the key set (never the type of `id`), so such rows can still arrive.

## Changes

Compute the annotation with a cast-free JSONB containment check, `session_recording_linked_flag @> {"id": flag.id}`, applied on every action. This matches the semantics the serializer fallback (`FeatureFlagSerializer.get_is_used_in_replay_settings`) and the delete guard already use: a non-integer stored id simply yields `False` instead of erroring the query. `bulk_update_tags` itself never reports replay usage (it returns raw dicts without building the serializer), so on that path the annotation only needs to be harmless, which containment is.

This supersedes the first revision of this PR, which kept the cast but scoped it to the `list` action; as review pointed out, that left the list endpoint 500ing on the same data.

Extended the regression test (`test_bulk_update_tags_with_non_integer_replay_linked_flag_id`): a sibling team stores a non-integer `session_recording_linked_flag` id, and the test asserts `bulk_update_tags` returns 200 with the tag applied, and that the list endpoint returns 200 with `is_used_in_replay_settings=True` for a genuinely linked flag. No existing test verified the annotation's correctness (as opposed to query count) through the list endpoint, so a wrong OuterRef correlation would previously have shipped silently. Existing replay-settings tests still cover the serializer fallback. I (the agent) was not able to run the Python suite in this environment (no local toolchain); CI will run it.

## 🤖 Agent context

Authored by the PostHog Slack app at Jake's request from a Slack thread; revised after review to adopt the suggested containment approach, which also fixes the list endpoint and removes the action conditional. The query snapshot was hand-updated to the expected containment SQL; if the exact formatting differs, the snapshot CI job will amend it as it did for the first revision.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B8ZE81ZT7/p1785151173481609?thread_ts=1785151173.481609&cid=C0B8ZE81ZT7)*


products/feature_flags/backend/api/feature_flag.py
```diff
@@ -2879,23 +2879,24 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
-        # This checks if any team in the same project uses this flag for session recording
-        # Extract the 'id' key from the JSONB field and cast to integer for safe comparison
-        from django.db.models import IntegerField
-        from django.db.models.functions import Cast
-
-        queryset = queryset.annotate(
-            is_used_in_replay_settings_annotation=Exists(
-                Team.objects.filter(
-                    project_id=OuterRef("team__project_id"),
+        if self.action == "list":
```
haacked: `suggestion`: The flags list page still 500s on the same data this fixes for `bulk_update_tags`. The cast stays on `list`, and per the snapshot this PR deletes it compiles to `((session_recording_linked_flag -> 'id'))::integer`, a jsonb cast — `('{"id": "123"}'::jsonb -> 'id')::integer` raises `cannot cast jsonb string to type integer`, and so does `{"id": null}`. Because the EXISTS scopes by project, one team's row breaks `GET /api/projects/{id}/feature_flags/` for every sibling team in that project.

That behavior is unchanged from before your diff, so it isn't a merge blocker. But [the containment check the serializer already uses](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/products/feature_flags/backend/api/feature_flag.py#L1037-L1040), and [the delete guard](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/products/feature_flags/backend/api/feature_flag.py#L1891), need no cast, are safe on every action, and would make this diff smaller by removing the conditional entirely. Worth fixing at the read layer regardless of the write path, since [`validate_session_recording_linked_flag`](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/posthog/api/team.py#L983) only checks the key set and never the type of `id`, so more of these rows can still arrive.

```suggestion
        # Matches the containment check in FeatureFlagSerializer.get_is_used_in_replay_settings,
        # so the annotated and unannotated paths agree. Containment never casts, so a
        # non-integer id in the JSON yields False instead of erroring the query.
        queryset = queryset.annotate(
            is_used_in_replay_settings_annotation=Exists(
                Team.objects.filter(
                    project_id=OuterRef("team__project_id"),
                    session_recording_linked_flag__contains=JSONObject(id=OuterRef("id")),
                )
            )
        )

        if self.action == "list":
```

with `from django.db.models.functions import JSONObject` at module level. On Postgres this emits `session_recording_linked_flag @> JSONB_BUILD_OBJECT('id', "posthog_featureflag"."id")`, which matches the serializer's semantics exactly. Two lines on your new test show the gap:

```python
        list_response = self.client.get(f"/api/projects/{self.team.id}/feature_flags/")
        assert list_response.status_code == status.HTTP_200_OK, list_response.json()
```

products/feature_flags/backend/api/feature_flag.py
```diff
@@ -2879,23 +2879,24 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
+            # Annotate replay settings usage to avoid N+1 queries when serializing a whole page.
+            # The cast of session_recording_linked_flag->>'id' to integer errors on non-integer
```
haacked: `suggestion`: Two claims in this comment will mislead the next reader. The generated SQL is `-> 'id'`, a jsonb cast, not the `->>'id'` text cast written here — visible in the `.ambr` lines this PR deletes — and that's why even a numeric-looking `"123"` errors rather than only non-numeric text. And `bulk_update_tags` never builds a `FeatureFlagSerializer`: [it runs the queryset](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/posthog/api/tagged_item.py#L351) and [returns raw dicts from `apply_bulk_tag_changes`](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/posthog/api/tagged_item.py#L382), so there's no containment fallback on that path. It's safe because nothing asks for replay usage there at all. The PR description carries both inaccuracies too.

```suggestion
            # Annotate replay-settings usage to avoid N+1 queries across a whole page.
            # Casting session_recording_linked_flag -> 'id' to integer errors on any team in
            # the project whose stored id isn't a JSON number, so keep it off other actions.
            # Single-object responses get the answer from the serializer's cast-free
            # containment check, and bulk_update_tags never reports replay usage at all.
```

products/feature_flags/backend/api/test/test_feature_flag.py
```diff
@@ -8227,6 +8227,24 @@ def test_bulk_update_tags_works_with_personal_api_key(self):
...
+
+        response = self.client.post(
+            f"/api/projects/{self.team.id}/feature_flags/bulk_update_tags/",
+            {"ids": [flag.id], "action": "add", "tags": ["foo"]},
+            format="json",
+        )
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        body = response.json()
+        assert body["updated"] == [{"id": flag.id, "tags": ["foo"]}]
+        assert body["skipped"] == []
```
haacked, line 8317: `suggestion`: The current test doesn't verify that `is_used_in_replay_settings` returns `True` through the list endpoint, which is now the only path using the annotation. [`test_is_used_in_replay_settings_serializer_field`](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/products/feature_flags/backend/api/test/test_feature_flag.py#L3998) tests a single flag, exercising [the cast-free serializer path](https://github.com/posthog/posthog/blob/ca2777283594d492788f7bb17b3831b3a29ebdbd/products/feature_flags/backend/api/feature_flag.py#L1037) instead. A wrong `OuterRef` correlation in the annotation would pass that test but break every flag's replay-usage badge in the list UI. `test_getting_flags_is_not_nplus1` validates query count but not correctness.

Extend the test to verify both implementations return the correct value.

```python
        # The list action is the only path that uses the annotation
        response = self.client.get(f"/api/projects/{self.team.id}/feature_flags/")
        assert response.status_code == 200
        listed = next(f for f in response.json()["results"] if f["id"] == flag.id)
        assert listed["is_used_in_replay_settings"] is True
```

products/feature_flags/backend/api/test/test_feature_flag.py
```diff
@@ -8227,6 +8227,24 @@ def test_bulk_update_tags_works_with_personal_api_key(self):
         assert body["updated"] == [{"id": flag.id, "tags": ["foo"]}]
         assert body["skipped"] == []
 
+    def test_bulk_update_tags_with_non_integer_replay_linked_flag_id(self):
+        # A team with a non-integer session_recording_linked_flag id used to 500 the endpoint:
+        # get_queryset cast that JSON id to integer, and Postgres errors on a non-numeric value.
```
haacked: `nit`: This comment sounds like a changelog entry ("used to 500 the endpoint") that the commit message already covers; CLAUDE.md asks to keep it out of code. It also refers to `get_queryset` when the method is `safely_get_queryset`, breaking the reference for grepping.

```suggestion
        # A non-integer id breaks any queryset that casts session_recording_linked_flag's id
        # to integer, so this pins that bulk_update_tags never applies that cast.
```

haacked (approved):
Nice root-cause work. The one thing worth a look: the list endpoint still runs the same cast, so a project with a bad row keeps 500ing there. Details plus a smaller alternative inline.

Also, flag IDs should always be integers. So I'll follow-up with some PRs to address why we're writing non-integers into that field.

products/feature_flags/backend/api/feature_flag.py
jakesciotto: Taken as suggested in 98aeb1c: the annotation now uses the containment check via `JSONObject(id=OuterRef("id"))` on every action, and the conditional is gone. Also added your two list-endpoint lines to the test (plus a sibling team holding the bad row, so the list assertion exercises the cross-team EXISTS).

products/feature_flags/backend/api/feature_flag.py
jakesciotto: The conditional and this comment are gone with the containment approach, and I've corrected the PR description on both points: it now describes the `-> 'id'` jsonb cast (failing on any non-number, including `"123"` and `null`), and no longer claims bulk_update_tags falls back to the serializer.

products/feature_flags/backend/api/test/test_feature_flag.py
jakesciotto: Added in 98aeb1c: self.team now genuinely links the flag (integer id), the bad row moved to a sibling team in the project, and the test asserts `is_used_in_replay_settings is True` for the flag through the list endpoint, so a broken OuterRef correlation fails it.

products/feature_flags/backend/api/test/test_feature_flag.py
jakesciotto: Reworded in 98aeb1c without the changelog framing or the stale get_queryset reference; it now states what the test pins (containment, never an integer cast) since the cast is gone entirely.

PR #73909, opened 2026-07-27
feat(warehouse): add HubSpot leads table (default-disabled)

luke-belton:
## Problem

The HubSpot data warehouse source syncs 7 CRM objects (contacts, companies, deals, tickets, quotes, emails, meetings) but not Leads.
Customers on Sales Hub who work leads in HubSpot can't pull that object into PostHog.

This adds Leads as an 8th syncable table.
It has to start disabled for existing connections, and that's the crux of the change (more below).

## Changes

New `leads` endpoint on the HubSpot source: `/crm/v3/objects/leads`, incremental on `hs_lastmodifieddate`, partitioned on `hs_createdate`.
It follows the same shape as the existing objects, so most of the machinery (search-path routing, association backfill, primary key, the per-object custom-property picker) extends to it for free.

Two things make leads different from the other 7:

> [!WARNING]
> Leads needs the `crm.objects.leads.read` OAuth scope, and that scope is Sales Hub Pro+/Enterprise only.

- **Default-disabled.** Existing HubSpot connections were authorized before this scope existed, so their tokens don't carry it. If leads defaulted on, every existing customer's sync would start 403-ing. A new `should_sync_default` field on `HubspotEndpointConfig` (set `False` only for leads, mirroring the pattern Algolia and Airbrake already use) threads through `get_schemas`, so leads starts off in both the new-source wizard and `auto_enable_new_schemas` for existing sources. Nothing else needed changing.
- **Scope added as optional, not mandatory.** `crm.objects.leads.read` goes in HubSpot's `optional_scope`, next to the existing plan-gated behavioral-events scopes. Requesting it in the mandatory `scope` string would fail the whole OAuth authorization for any portal whose plan doesn't include leads.

A customer who wants leads reconnects HubSpot (picking up the optional scope where their plan allows), then enables the table. A customer who enables it without the scope hits the existing `403 ... reconnect your HubSpot account` error from `get_non_retryable_errors` - that path is unchanged.

Property handling needed no leads-specific wiring. The per-object custom-properties textarea is generated off `DEFAULT_PROPS.items()` and read via `getattr(config, f"{schema_name}_properties")`, so adding `leads` to `DEFAULT_PROPS` produces a `leads_properties` field automatically (regenerated into the source config). One nuance: lead standard properties are all `hs_`-prefixed, and auto-discovery only pulls non-`hs_` props, so `DEFAULT_LEAD_PROPS` lists them explicitly.

Files:
- `settings.py` - `leads` endpoint, `DEFAULT_LEAD_PROPS`, `should_sync_default` field, catalog wiring (`ENDPOINTS`, `OBJECT_TYPE_SINGULAR/PLURAL`, `DEFAULT_PROPS`)
- `source.py` - thread `should_sync_default` into `SourceSchema`
- `integration.py` - add `crm.objects.leads.read` to `optional_scope`
- `canonical_descriptions.py` - `leads` table + column descriptions
- `generated_configs/hubspot.py` - regenerated (`leads_properties`)

Automated, run by me (Claude):

- `hogli test .../hubspot/test/` - all 152 HubSpot tests pass, including a new assertion that leads is `should_sync_default=False` while the other 7 stay on. That test guards the core requirement: flipping leads on by default would 403 existing customers.
- A standalone consistency check that `ENDPOINTS`, `HUBSPOT_ENDPOINTS`, `DEFAULT_PROPS`, `OBJECT_TYPE_SINGULAR` and the canonical descriptions all agree on the endpoint set, and that leads' cursor + `hs_object_id` are present in its default props (the parametrized `TestSettingsShape` suite covers this too).
- ruff lint/format, `ty check`, and `hogli ci:preflight --strict` all clean.

I did **not** run a live end-to-end sync against a real HubSpot portal - that needs credentials and a Sales Hub Pro+ account with leads enabled. Worth a manual verification before/after merge: confirm leads shows up **unchecked** in the source wizard, and that enabling it on a scoped connection produces a `leads` table.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The user-facing source doc lives in the posthog.com repo and its table list auto-renders via `<SourceTables />`, so no doc change is needed here.
A short "leads is off by default, reconnect to enable" note there would be a good follow-up.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) - Luke directed this; I (Claude) did the implementation.

Started as a scoping question ("how much work to add a leads table, default-disabled for existing customers"). Exploration surfaced the two non-obvious constraints that shaped the design: the missing OAuth scope (there's no reconnect-prompt mechanism for HubSpot the way there is for Slack's `missing_scopes()`, so we rely on the existing reactive 403), and that the scope is plan-gated (hence `optional_scope`, not `scope` - requesting a plan-unavailable scope fails the whole authorization).

One thing I rejected: building a proactive scope-upgrade/reconnect prompt for existing customers. Luke chose to lean on the existing error path instead, so that's out of scope here.

Skills invoked: `/writing-tests` (gated the one test I added - it catches the default-disabled regression no existing test did).

Note for reviewers: `pnpm generate:source-configs` also rewrote ~15 unrelated generated configs (formatter line-wraps on some, genuine pre-existing drift on stripe/slack). I reverted all of those; this PR only touches `generated_configs/hubspot.py`. That drift is worth a separate look by the warehouse-sources owners.

---

### DIFF
--- a/posthog/models/integration.py
+++ b/posthog/models/integration.py
@@ -683,8 +683,10 @@ class OauthIntegration:
                 client_secret=settings.HUBSPOT_APP_CLIENT_SECRET,
                 scope="tickets crm.objects.contacts.write sales-email-read crm.objects.companies.read crm.objects.deals.read crm.objects.contacts.read crm.objects.quotes.read crm.objects.companies.write",
                 additional_authorize_params={
-                    # NOTE: these scopes are only available on certain hubspot plans and as such are optional
-                    "optional_scope": "analytics.behavioral_events.send behavioral_events.event_definitions.read_write"
+                    # NOTE: these scopes are only available on certain hubspot plans and as such are optional.
+                    # crm.objects.leads.read is Sales Hub Pro+/Enterprise only — requesting it as a
+                    # mandatory scope would fail the whole authorization for portals that lack it.
+                    "optional_scope": "analytics.behavioral_events.send behavioral_events.event_definitions.read_write crm.objects.leads.read"
                 },
                 id_path="hub_id",
                 name_path="hub_domain",
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs/hubspot.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/generated_configs/hubspot.py
@@ -14,6 +14,7 @@ class HubspotCustomPropertiesConfig(config.Config):
     quotes_properties: str | None = None
     emails_properties: str | None = None
     meetings_properties: str | None = None
+    leads_properties: str | None = None
 
 
 @config.config
--- a/products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/canonical_descriptions.py
+++ b/products/warehouse_sources/backend/temporal/data_imports/sources/hubspot/canonical_descriptions.py
@@ -139,7 +139,6 @@ CANONICAL_DESCRIPTIONS: CanonicalDescriptions = {
             "hs_lead_name": "The lead's name.",
             "hs_lead_type": "The type of lead (e.g. new business).",
             "hs_lead_label": "The current status label of the lead (e.g. warm, cold).",
-            "hs_lead_source": "The analytics source of the lead at creation, set automatically by HubSpot.",
             "hs_pipeline": "The pipeline the lead is in.",
             "hs_pipeline_stage": "The pipeline stage the lead is in (e.g. new, attempting, connected, qualified).",
             "hs_createdate": "Date the lead was created in HubSpot.",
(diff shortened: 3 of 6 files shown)


luke-belton:
Verified `DEFAULT_LEAD_PROPS` against HubSpot's docs (follow-up to a review question about provenance):

| Property | Verified by |
|---|---|
| `hs_lead_name`, `hs_lead_type`, `hs_lead_label` | Named verbatim in the [Leads API guide](https://developers.hubspot.com/docs/api-reference/crm-leads-v3/guide) |
| `hs_object_id` | Same guide (batch read section) |
| `hs_createdate`, `hs_lastmodifieddate`, `hs_pipeline`, `hs_pipeline_stage` | Real API response JSON in a [HubSpot community thread](https://community.hubspot.com/t5/APIs-Integrations/Leads-created-through-API-is-not-coming-in-UI-but-coming-in-API/m-p/1087005) |
| ~~`hs_lead_source`~~ | **Removed in bd08b07** - not a default lead property. The [KB's default lead properties](https://knowledge.hubspot.com/properties/hubspots-default-lead-properties) "Lead Source" is an inherited analytics property whose internal name isn't published, and no API response shows `hs_lead_source`. |

Also re-confirmed from the guide: endpoints (`/crm/v3/objects/leads` + search), the `crm.objects.leads.read` scope, and the Sales Hub Pro/Enterprise requirement.

The definitive per-portal check remains `GET /crm/v3/properties/leads` on a real Sales Hub Pro+ account, worth doing as part of the manual verification already flagged in the description.

Gilbert09 (approved):

PR #9618, opened 2022-05-03
feat(apps): frontend apps in the sidebar

mariusandra:
## Problem

[We want](https://github.com/PostHog/posthog/issues/9533) to ~plugins~ apps to add scenes in the frontend. This PR adds basic support for that, with inline editing.

## Changes

- This adds support for frontend plugins, created through the plugin source editor. (Support via github/npm coming next)
- Code is transpiled in the plugin server, using only one thread per cluster.
- The transpiled code is then loaded on the frontend through a `import` call.

## How did you test this code?

- Added several tests
- Tested locally many times

PR #73919, opened 2026-07-27
chore: remove the agent platform

### LINKED ISSUE #71123: Agent Builder should hand off directly to newly created agents
## Is your feature request related to a problem?

When someone creates and promotes an agent from Agent Builder in PostHog Code, there is no obvious way to leave the authoring conversation and start talking directly to the newly created agent.

Authoring and runtime use separate surfaces and identities. Agent Builder can create the application and its revisions through the PostHog MCP control plane, but creation alone does not attach the new agent to the current PostHog Code conversation. The user must currently understand triggers and then use an out-of-band route such as configuring the agent's MCP endpoint or calling its chat ingress with a credential.

This makes a basic expectation feel surprisingly complicated: “I created this agent, now let me try it.”

Owning team: @PostHog/team-agents

## Describe the solution you'd like

After Agent Builder successfully promotes an agent, PostHog Code should offer a clear **Chat with this agent** action.

That action should:

1. Open a new PostHog Code conversation running as the newly created agent.
2. Use the user's existing PostHog identity without asking them to create or paste a personal API key.
3. Preserve the agent's declared runtime authentication contract rather than bypassing its trigger configuration.
4. Explain when the agent has no compatible user-invocable trigger and offer to add one safely.
5. Keep Agent Builder available so the user can return to editing after testing the agent.

A sensible default for agents intended for interactive use may be a `chat` trigger with PostHog authentication. Whether an `mcp` trigger should also be added by default is a separate product decision.

## Describe alternatives you've considered

- Connect directly to the deployed agent's `/mcp` endpoint. This works for external MCP clients, but requires registering another server, providing credentials, and starting a new client session.
- Invoke the agent through the central PostHog MCP. [PR #69846](https://github.com/PostHog/posthog/pull/69846) adds value for build → promote → invoke → verify automation, but it makes Agent Builder relay the result rather than giving the user a direct conversation with the created agent.
- Call `/run`, `/send`, and `/listen` directly with a personal API key. This is useful for API consumers, but should not be the default PostHog Code experience.

## Additional context

The desired distinction is:

- **Agent Builder:** author and configure an agent.
- **PostHog MCP invocation tools:** automate verification and agent-to-agent orchestration.
- **Direct handoff:** let the person immediately use the agent they just created.

The authentication implementation should make this transition feel continuous while retaining the control-plane/runtime security boundary.

### LINKED ISSUE #71714: Agent platform: harden the untrusted-caller (authenticated/public) audience — follow-ups
Follow-ups deferred from #71696, which adds the posthog `authenticated` audience and the cross-agent team-wide memory read, and hardens the "reachable by untrusted callers" (authenticated audience OR public auth) envelope. These are accepted residual risks tracked here rather than blocking that PR.

- [ ] **Per-principal / per-app rate limit on invocation.** An untrusted-caller audience drops the tenant gate; spend is billed to the owning team. Needs a platform-wide invocation rate limit keyed on caller principal + app. (Own PR — platform-wide.)
- [ ] **Capability-flag denylist instead of hardcoded tool sets.** `AUTHENTICATED_FORBIDDEN_TOOLS` / `AUTHENTICATED_OWNER_CREDENTIAL_TOOLS` are hardcoded id lists. A new shared-state-read or owner-credential tool is not fail-closed by default. Add a `readsSharedState` / `usesOwnerCredential` capability flag on the native tool definitions and derive the envelope from that, so new tools are safe by construction.
- [ ] **`introspect` timeout + short cache.** The PostHog bearer introspection (`/api/users/@me/`) on every `posthog`-auth request has no timeout and no cache. Add both.
- [ ] **Re-check team-wide-share grant at read time.** The `memory_shared_team_wide` grant is snapshotted per session; a mid-session revoke does not cut in-flight readers (same-team, fail-closed). Re-check at read for immediate revocation.
- [ ] **Pairing guard for `table-append` on a cross-caller store.** Rows an untrusted-caller agent appends are caller-influenceable; downstream readers must treat them as untrusted input. Consider a guard/annotation so a consumer can't accidentally trust them.

Context: #71696.

benjackwhite:
## Problem

We're shelving the agent platform. This removes it from the monorepo.

Closes #71714
Closes #71123

## Changes

Deletes `products/agent_platform` — the Django app, its product-DB migrations, and the four Node services (agent-ingress, agent-runner, agent-janitor, agent-sandbox-host) — plus every reference into it: `INSTALLED_APPS`, `db_routing.yaml`, `posthog/settings/agents.py`, the `agents` / `agent_approvals` API scopes, the MCP tools and builder playbooks, dev-env procs and intents, three CI workflows, and four semgrep rules.

**On the migrations** (the open question): deleting them is right here, but it needs an explicit allowlist entry, which I've added.

The repo has a hard guard against deleting historical migrations (`hogli lint:migration-deletions`) and `safe-django-migrations.md` gives four reasons. Three don't apply to agent_platform: the tables only ever existed in the dedicated `agent_platform` product database, so nothing lingers in a shared cluster, no `django_migrations` orphans survive, and fresh databases can't diverge because the database stops existing. The fourth does apply regardless of which database the tables live in — the risk analyzer replays each open PR's tree, so any in-flight branch predating the deletion re-flags these files as phantom new migrations until it merges master.

That's what the allowlist is for, and a whole-product retirement is the case it names. Entry added at `.github/scripts/migration-deletion-allowlist.txt` with the reasoning inline. The phased `SeparateDatabaseAndState` + `DROP TABLE` dance the doc prescribes would be ceremony here — there's no shared table to orphan and the cluster is dropped wholesale in step 3.

## Merge order

This is the first of three PRs. Each step must deploy before the next merges.

1. **This PR.** Django stops connecting to the `agent_platform` database and stops calling ingress/janitor. Running pods keep serving off their pinned images.
2. **[PostHog/charts#13530](https://github.com/PostHog/charts/pull/13530)** — deletes the three ArgoCD apps and the Django DB credentials. Has to come second: it removes the `PRODUCT_DB_AGENT_PLATFORM_*` env and the `agent-platform` psql cluster that this PR's `db_routing.yaml` entry would otherwise still expect.
3. **[#9618](https://github.com/PostHog/posthog-cloud-infra/pull/9618) then [#9617](https://github.com/PostHog/posthog-cloud-infra/pull/9617)** — destroys the database, buckets, cache, certs and DNS. Point of no return.

Revertible up to step 3.

I'm an agent. No manual testing — everything below is a command I actually ran.

Passed: `ruff check` (all clean) and `ruff format --check`, `tach check`, `lint-imports` (1 contract kept), `pytest tools/hogli-commands/.../test_devenv.py` (85 passed — this covers the `bin/mprocs.yaml` ↔ `devenv/intent-map.yaml` capability invariant), the MCP unit tests for tool filtering / definitions / schema snapshots, MCP `oxlint` and `oxfmt --check`, `oxfmt --check` on every edited root and frontend file, and a parse check over all 18 edited YAML/JSON/TOML files. Settings-level assertions confirm `agent_platform` is absent from `INSTALLED_APPS` and `DATABASES` and that `get_scope_descriptions()` still builds.

Couldn't run here (no Postgres in the sandbox): `django.setup()`, the URL conf import, and the DB-backed portion of `posthog/test/test_scopes.py` (49 pure-logic tests passed, 37 errored on connection refused). MCP `typecheck` reports no agent- or playbook-related errors; its remaining errors are unresolved `@posthog/quill` and `react` imports caused by my partial `--filter` install, not by this change.

CI turned up two things I'd missed, both now fixed on this branch:

- An MCP test asserted `toHaveLength(24)` on the required-feature-flag list alongside the entries I'd edited. Being straight about how it slipped through: my local re-run used `vitest run -u` with path filters and reported "1 passed (1)" — it ran a single test, not the file, and I read that as green. The full suite (93 files, 2385 tests) now passes.
- `hogli lint:migration-deletions` blocked the migration deletions, resolved via the allowlist entry described above.

Two other fixes came out of it:

- **`fix(ci)`: the `dockerfiles` paths-filter matched deleted files**, so hadolint was handed the two removed agent_platform Dockerfiles and failed on paths that no longer exist. Scoped the filter to `added|modified`. This isn't specific to this branch — it would break any PR that deletes a Dockerfile. Strictly narrower matching, so unrebased PRs are unaffected.
- `services/mcp/src/api/generated.ts` regenerated itself as predicted: CI's auto-commit dropped all 1469 lines of dead `Agent*` Orval types.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Product docs lived under `products/agent_platform/docs` and go with it.

## 🤖 Agent context

Claude Code, driven by Ben. I mapped the footprint with parallel read-only explore agents across all three repos before touching anything, then applied this repo's edits directly and delegated the charts and infra repos to separate agents.

Two things worth flagging from the sweep:

- I kept `bin/frontend-exclude-filter` and its two CI steps. agent_platform was the only nested product workspace, so the mechanism is now a no-op, but it already handles the zero-match case by emitting `- '**'` (always true, never skips CI) and it re-arms automatically for the next product that adds one. Deleting it would have meant editing two workflows for no gain.
- `ai_gateway` survives in `devenv/intent-map.yaml` because a proc still declares that capability, but the `agents` mode was its only consumer, so nothing enables it by default now. I repointed its `requires` from `agent_runtime` to `core_infra` and left it there rather than guessing at a new home. If someone wants the local gateway reachable from a mode, that's a follow-up.

`services/agent-proxy`, `temporal-worker-tasks-agent`, the PR-approval bot and the VictoriaMetrics vmagent all match an "agent" grep but belong to other systems. I left them alone. The `agent-platform` strings left in `products/engineering_analytics` fixtures are historical GitHub PR records, not a dependency.

No skills were invoked — no migrations were authored, no DRF endpoints or tests changed beyond deletions.


feliperalmeida (approved):

gantoine (approved):
🫡

dmarticus (approved):
world-class diff

PR #73935, opened 2026-07-27
refactor(agentic-provisioning): restructure into drf class-based views

rafaeelaudibert:
## Problem

`ee/api/agentic_provisioning/` was one 2,689-line `views.py` of function views, wired route by route from `ee/urls.py` (26 `path()` entries). Everything lived in that file: auth resolution, team creation, token minting, region proxying, wizard kickoff, validation, throttling, and the response envelopes. Meanwhile #72799 and #73387 left `ee/partners/stripe/api/provisioning/` as the house pattern for a partner provisioning surface: DRF class-based views on an envelope-aware base, `RegionProxyMixin` on `dispatch`, serializers that raise spec errors, a self-contained `urls.py`. The two partner surfaces had drifted into different shapes, so a fix in one didn't obviously translate to the other.

The routes carried a second problem. Every partner endpoint was registered twice, under `/api/agentic/...` and a generic `/api/provisioning/...` alias, which doubled the route table and its test surface.

## Changes

Restructured the package to mirror the Stripe module. Nothing about the wire contract moves: error envelopes, check order, and rate-limit consumption order are unchanged, which is why validation still raises `ProvisioningError` rendered per view rather than DRF's default error shape.

```mermaid
flowchart TD
    P{{Partner}} --> U["ee/urls.py<br/>26 path() entries"]
    U --> V["views.py<br/>2,689 lines"]
    V --> DB[(PostHog)]
    class P phBlue;
    class U,V phYellow;
    class DB phRed;
```

```mermaid
flowchart TD
    P{{Partner}} --> U["agentic_provisioning/urls.py<br/>15 path() entries"]
    U --> V["views/<br/>base · account_requests · oauth_token<br/>resources · deep_links · github_grants · authorize"]
    V --> S["serializers · throttling · exceptions"]
    V --> H["accounts · teams · tokens<br/>credentials · regions · wizard"]
    H --> DB[(PostHog)]
    class P phBlue;
    class U,V phYellow;
    class S,H phGray;
    class DB phRed;
```

- **Views** become class-based on `ProvisioningAPIView` / `BearerResourceAPIView`, which own the envelope-aware `handle_exception`, the region proxy on `dispatch`, and the shared `validated_body` / `get_scoped_team` helpers.
- **Domain logic** moves out of the view layer into `accounts`, `teams`, `tokens`, `credentials`, `regions`, and `wizard`, so a view reads as request handling and nothing else.
- **Throttling** moves onto DRF's `get_throttles()` where the throttle key is known before the handler runs, with `enforce_*` helpers kept only where the key appears mid-flow. These are additive on purpose: `DEFAULT_THROTTLE_CLASSES` really does apply here (`get_cache_key` falls back to client IP for anonymous requests), so a plain `throttle_classes = [...]` would have silently dropped the per-IP protection.
- **Routes** live in the package's own `urls.py`, and `ee/urls.py` just includes it.

> [!WARNING]
> `/api/provisioning/*` is removed. Those paths now 404, same as any unknown API path. Partners keep using `/api/agentic/*`, which is the only prefix the public docs ever described.

I checked the self-capture telemetry before deleting the alias (Claude ran the queries, split on `$request_path` the same way #73387 did). Over 120 days the alias took roughly 40 requests, four orders of magnitude below `/api/agentic/*`, and not one was a working partner integration: invalid signatures, wrong API version, missing code, plus a handful of calls from our own tooling. Every real caller is on `/api/agentic/*`.

Automated only, all run locally by the agent. I did not exercise this against a real partner.

- `hogli test ee/api/agentic_provisioning/test/` → 306 passed. The suite was already rewritten around bearer/PKCE partner auth in #73387; this PR keeps it and only removes what the alias removal makes untestable.
- `uv run mypy --cache-fine-grained .` → clean, 16,928 files. `ruff check` / `ruff format` clean. `hogli ci:preflight --strict` → 0 failures.
- Route check: every one of the 15 URL names reverses to its expected path and resolves to its view, and the removed alias paths return 404 through the test client while `/api/agentic/provisioning/account_requests` still returns 401. Worth knowing if you assert on this later: the alias paths still `resolve()` successfully because they fall through to the global `^api.+` catch-all in `posthog/api/rest_router.py`, so a `Resolver404` assertion would be misleading. Status codes are the honest check.

Test changes are deletions only, no new tests. `test_url_aliases.py` existed to assert the alias routes resolved, so it goes with them, and `test_repositories_happy_path` in `test_github_grants.py` loses its second parameterized case. Route registration for all 11 partner endpoints is still covered through their real paths by `test_account_requests`, `test_oauth_token`, `test_resources`, `test_rotate_credentials`, `test_deep_links`, `test_github_grants`, and `test_wizard_resource_actions`.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No doc change needed for the alias removal, since `contents/docs/integrate/provisioning.mdx` only ever documented `/api/agentic/...` paths.

Separately, that doc is stale from #73387 and still needs a companion PR on posthog.com: it documents `service_id` as a request and response field (now ignored, and no longer echoed back) and the `API-Version: 0.1d` header (no longer enforced here). It also documents `/resources` as step 3 of onboarding without mentioning `project_id`, the one field that changes what the endpoint does, and omits `GET /resources/{id}` and `POST /resources/{id}/remove` entirely.

## 🤖 Agent context

Authored with Claude Code (Claude Opus 5), directed by @rafaeelaudibert. The restructure happened across a few sessions: the initial split, then a `/simplify` and review pass that pulled out the shared `validated_body` / `get_scoped_team` helpers, a single `resolve_bearer_access_token` primitive behind all three bearer paths, and `ConfidentialPartnerAuthentication` for the GitHub grant endpoints.

Deliberately left out: extracting a shared `provisioning_common` package with the Stripe module. That is the obvious next step, but it is a large cross-namespace change and the Stripe side is spec-frozen, so it wants its own PR. Also skipped moving PKCE checks into the serializer and deriving the throttle envelope from the view, since both would change observable wire order or response shapes.

On the alias: the route table used to generate both prefixes from one list so an endpoint could not ship under one and not the other. With a single prefix that indirection stopped paying for itself, so `urls.py` is now a flat list of explicit `path()` entries, matching how the Stripe module reads.


rafaeelaudibert:
* **#74364** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74351** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74089** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74089?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#73935** <a href="https://app.graphite.com/github/pr/PostHog/posthog/73935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/73935?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rafaeelaudibert:
Fixed the hex-security-app finding in 962324b2b5ee09345c0611248ae761b29d56fa4e.

The finding is real and pre-existing: master's `_resolve_team_for_existing_user` has the identical organization-only check, so this PR moved it verbatim rather than introducing it. `user_can_access_team` already existed for this exact purpose (its docstring describes this hole) and was wired into the resource paths, just never into team selection.

Four sites, not the two the bot flagged:

| Site | Before | After |
|---|---|---|
| `resolve_team_for_existing_user`, requested team | org membership only | + `user_can_access_team` |
| `resolve_team_for_existing_user`, auto-select | no access check at all | inaccessible teams filtered out |
| `agentic_authorize`, trusted-partner auto-approve | mints against `non_demo_teams[0]` unchecked | same filter applied before selection |
| `AuthorizeConfirmView` | org membership only | + `user_can_access_team` |

The two extra sites matter because neither needs attacker-supplied input: the auto-select and auto-approve paths both hand back the org's single non-demo team without asking whether the user can open it.

Filtering auto-select routes a user whose only teams are restricted into the existing create-a-new-project fallback instead of returning a project they cannot open. Wire responses are unchanged: the requested-team path still fails `team_resolution_failed`, and confirm still returns 403 `team_not_accessible`.

**Tests.** Three regression cases, added as parameterized cases on the existing "inaccessible team" tests rather than new near-duplicates, plus one new test for the auto-approve path that had no equivalent:

- `test_existing_user_with_inaccessible_team_id_returns_400[restricted_in_own_org]` - same-org team denied by project access controls, alongside the existing cross-org case
- `test_existing_user_auto_select_skips_restricted_team` - no `team_id` supplied, the org's only team restricted
- `test_confirm_rejects_inaccessible_team[restricted_in_own_org]`
- `test_trusted_partner_auto_redirect_skips_restricted_team`

All four fail with the fix reverted (verified by stashing only the source changes) and pass with it. `hogli test ee/api/agentic_provisioning/test/` is green at 313, and `uv run mypy --cache-fine-grained .` is clean.

One note for whoever reviews: `.../test/base.py` gained a shared `_restrict_team_access` helper. `test_scope_hydration.py` and `test_resources.py` each carry a local copy of that setup; I left those alone to keep this diff to the security fix, but they could collapse onto the shared helper later.

ee/api/agentic_provisioning/views/github_grants.py
```diff
@@ -0,0 +1,93 @@
...
+                "email": grant.email,
+                "expires_in": github_grants.GITHUB_GRANT_TTL_SECONDS,
+            }
+        )
+
+
+class GitHubGrantRepositoriesView(ConfidentialPartnerAPIView):
+    # Keyed per grant id, so polls against an unknown id can only exhaust that
+    # id's own budget.
+    partner_throttle_classes = [GrantPollThrottle]
```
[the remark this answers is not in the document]
rafaeelaudibert: Accurate, and fixed in d8fe3bef2a4.

You are right on the mechanics: DRF runs `check_throttles` in `initial()`, before the handler calls `load_grant()`, and the key was the grant id alone. So a confidential partner that learned another partner's grant id could spend the owner's poll budget on it with requests that all end in 404.

Went with your second suggestion (partner in the key) rather than moving the check after `load_grant()`, because keeping it a declarative `partner_throttle_classes` entry means it still gets the base view's `rate_limited` envelope and `Retry-After` for free. A grant belongs to exactly one partner, so scoping the key does not split any legitimate caller's quota.

Added `test_repositories_poll_budget_is_not_shared_across_partners`, which fails with a 429 on the old key.

Worth noting the practical severity matches your "low": `grant_id` is an opaque token only ever returned to the partner that created it, so the attacker needs a secret they should not have. Fixed anyway since it was two lines.

ee/api/agentic_provisioning/teams.py
```diff
@@ -0,0 +1,118 @@
...
+        return None
+
+    org_ids = [m.organization_id for m in memberships]
+
+    if requested_team_id is not None:
+        try:
+            team = Team.objects.get(id=requested_team_id, is_demo=False)
+        except Team.DoesNotExist:
+            return None
+        if team.organization_id not in org_ids:
```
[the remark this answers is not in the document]
rafaeelaudibert: Accurate, and fixed in 962324b2b5e.

The escalation path was real: `resolve_team_for_existing_user` checked `team.organization_id in org_ids` and nothing else, so a partner could name any team id in one of the user's orgs via `configuration.team_id` and get a code minted against a project that project-level access controls exclude the user from.

`user_can_access_team` already existed and was wired into the resource paths, just not into the selection sites. It is now applied at all four:

- `resolve_team_for_existing_user`, requested-team path (the one you flagged)
- `resolve_team_for_existing_user`, auto-select. Worth calling out that this one needed no attacker input at all: the org's single non-demo team was returned unchecked
- `agentic_authorize`, the trusted-partner auto-approve path, which minted against `non_demo_teams[0]`
- `AuthorizeConfirmView`, since the user picking the team in the consent screen does not imply access to it, exactly as you noted

Inaccessible teams are now filtered out of auto-select, so those users fall into the existing create-a-new-project path rather than being handed a project they cannot open. Wire responses are unchanged: the requested-team path still fails `team_resolution_failed`, confirm still returns 403 `team_not_accessible`.

Regression tests added in `test_authorize.py` and `test_account_requests.py` via a `_restrict_team_access` helper that makes the user an org member with an `access_level="none"` project ACL.

ee/api/agentic_provisioning/accounts.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+    if app is None:
+        return None
+    redirect_uris = app.redirect_uris.strip()
+    if not redirect_uris:
+        return None
+    return redirect_uris.split()[0]
+
+
+def mint_auth_code(
```
fercgomes, line 60: since we're doing refactoring for cleanliness sake...

this function is minting a code but also storing it, so maybe we can be more descriptive with its name?

ee/api/agentic_provisioning/accounts.py
```diff
@@ -0,0 +1,419 @@
...
+
+    capture_provisioning_event(
+        "account_request",
+        "new_user",
+        partner=partner,
+        region=region,
+        team_id=team.id,
+    )
+
+    # Emit the standard signup event so provisioned accounts flow into the shared
```
fercgomes, line 314: this comment block is useless

ee/api/agentic_provisioning/accounts.py
```diff
@@ -0,0 +1,419 @@
...
+        code_challenge_method=code_challenge_method,
+    )
+
+    response_body: dict[str, Any] = {"id": request_id, "type": "oauth", "oauth": {"code": code}}
+    if wizard_payload is not None:
+        response_body["wizard"] = wizard_payload
+    return response_body
+
+
+def process_wizard_block(*, partner: OAuthApplication, user: User, team: Team, wizard_config: dict) -> dict[str, Any]:
```
fercgomes, line 370: process_wizard_block is a bit confusing

ee/api/agentic_provisioning/credentials.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ProvisioningError(
+            "invalid_label_prefix",
+            f"label_prefix must be {PROVISIONED_PAT_LABEL_PREFIX_MAX_LENGTH} characters or fewer",
+        )
+
+    # Reject Unicode control (Cc), format (Cf), and line/paragraph separators (Zl/Zp).
+    # Cf is the important one - it includes bidi overrides (U+202A-U+202E) and
+    # isolates (U+2066-U+2069), which a partner could use to re-order surrounding
+    # text in the user's settings page (Trojan Source class). Cc covers C0 + DEL.
+    if any(unicodedata.category(c) in {"Cc", "Cf", "Zl", "Zp"} for c in stripped):
```
fercgomes, line 51: isn't it better to use a battle-tested lib to sanitize this kind of user input, instead of always reimplementing from scratch?

ee/api/agentic_provisioning/throttling.py
```diff
@@ -0,0 +1,271 @@
...
+    return _fixed_window_count(cache_key, PARTNER_RATE_LIMIT_WINDOW_SECONDS), limit
+
+
+def _capture_partner_rate_limited(partner: OAuthApplication, endpoint: str, limit: int, count: int) -> None:
+    capture_provisioning_event(
+        PARTNER_RATE_LIMIT_EVENT_NAMES[endpoint], "rate_limited", partner=partner, limit=limit, count=count
+    )
+
+
+def enforce_partner_rate_limit(
```
fercgomes, line 109: why not make these class methods?

fercgomes (approved):
this is awesome, it already looks much more organized and clean

ee/api/agentic_provisioning/views/github_grants.py
```diff
@@ -0,0 +1,93 @@
...
+from ee.api.agentic_provisioning import github_grants
+from ee.api.agentic_provisioning.analytics import capture_provisioning_event
+from ee.api.agentic_provisioning.exceptions import ProvisioningError
+from ee.api.agentic_provisioning.serializers import GitHubGrantCreateSerializer
+from ee.api.agentic_provisioning.throttling import GitHubGrantsThrottle, GrantPollThrottle
+from ee.api.agentic_provisioning.views.base import ConfidentialPartnerAPIView
+
+
+class GitHubGrantsCreateView(ConfidentialPartnerAPIView):
+    partner_throttle_classes = [GitHubGrantsThrottle]
```
MattBro: Declaring `GitHubGrantsThrottle` here runs it in DRF's `initial()`, before the `provisioning_can_create_accounts` check in `post()`. Master ran the capability check first (`views.py:653`) and the quota second (`views.py:657`).

So a partner with the flag off used to get 403 forever and now gets 403 ten times then 429, with its real hourly grant budget spent on requests master rejected for free. `AccountRequestsView` keeps the old order with an inline `enforce_partner_rate_limit` after the capability check.

ee/api/agentic_provisioning/tokens.py
```diff
@@ -0,0 +1,204 @@
...
+            for rt in orphan_refresh:
+                rt.scoped_teams = [t for t in (rt.scoped_teams or []) if t != team_id]
+                rt.save(update_fields=["scoped_teams"])
+
+
+def get_available_teams_for_user(user: User) -> list[dict[str, object]]:
+    """Return the user's non-demo teams for inclusion in the token exchange response."""
+    org_ids = list(user.organization_memberships.values_list("organization_id", flat=True))
+    teams = Team.objects.filter(organization_id__in=org_ids, is_demo=False).select_related("organization")
+    return [
```
MattBro, line 201: This is `ee/partners/stripe/api/provisioning/core.py:664` minus its `if user_can_access_team(user, team)` clause, so team and org names from restricted teams in advanced-permissions orgs get listed in the token response. The helper is already imported in this file at line 23.

Pre-existing on master, but converging on the Stripe shape is the point of this PR and this is the one behavioral difference between the two. It also feeds a bug in posthog.com#18359, which uses `available_teams[0]` as a team id.

MattBro (approved):
Approving. Two comments above, both small. The throttle ordering one is worth doing before merge since it contradicts the "check order and rate-limit consumption order are unchanged" claim in the description.

ee/api/agentic_provisioning/credentials.py
rafaeelaudibert: This is just a reimplementation of what we had before. I agree it could be improved

ee/api/agentic_provisioning/accounts.py
rafaeelaudibert: Just moved it around, improving the code is something we can do later

ee/api/agentic_provisioning/tokens.py
rafaeelaudibert: Nah, the point of this is PR is to NOT converge on Stripe's shape. The idea is that they should not be the same. I'll work on the fix, though.

rafaeelaudibert:
### Merge activity

* **Jul 28, 11:17 PM UTC**: [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/73935) couldn't add this PR to the merge queue because it failed for an unknown reason (GitHub is reporting that this PR is not mergeable, despite passing required status checks defined by your branch protection rules. Please check your rulesets for additional blocking criteria. Graphite Merge Queue does not currently support rulesets. Please contact Graphite support for further assistance.).

PR #73946, opened 2026-07-27
feat(playground): add access control

ReeceJones:
## Problem

LLM playground does not have any access control, meaning any user of a project with the playground setup and tokenmax and blow up the company credit card.

## Changes

Adds all of the plumbing for access control for the llm playground:
- New `llm_playground` scope. Open to change this, but it felt more descriptive than just `playground`
- One-off access control implementation in the playground viewset since it is stateless and the existing mixin only works for stateful (model-backed) viewsets
- Updated the frontend code to gate actions using on the user's permissions
- Set access control icon for the llm playground scope in `ScopeIcon.tsx`
- Register the playground scene for access control
- A bunch of other random stuff needed to make it worky

New tests + manual clicking

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None – this adds an access control resource, which the access control docs describe generically rather than enumerating.

## 🤖 Agent context

I directed this; Claude (PostHog Code, Opus) wrote it. Skills invoked: a personal `rj-retroactively-adding-access-control` skill that carries the checklist for this kind of retrofit, plus the repo's `/django-migrations` and `/writing-tests`.

Three decisions worth flagging for review, since each had a reasonable alternative:

- **Dedicated resource vs. reusing `llm_analytics`.** I chose a dedicated one so admins can deny the playground (it spends money) while still allowing trace viewing. The cheaper option was an inheritance-map entry, which would have given the proxy a scope object but no separate settings row.
- **One-off permission class vs. moving the proxy under the project router.** Moving it would have let `TeamAndOrgViewSetMixin` handle enforcement automatically and would be more consistent with every other access-controlled viewset, but it changes the API path and needs frontend changes. I kept the URL stable and wired the check by hand, with a docstring explaining why it is hand-rolled.
- **Scope named `llm_playground`, not `playground`.** Renamed in a follow-up commit after review feedback, to match the sibling AI observability scopes (`llm_analytics`, `llm_prompt`, `llm_provider_key`, `llm_skill`) and the `llm_playground` key the product tree already uses for this scene's icon. It also sits with the other `llm_*` entries in every ordered registry rather than under `p`.
- **No level cap.** I tried restricting to none/editor, then reverted it (see the revert commit). Access control validation is index-based over `ordered_access_levels`, so `minimum`/`maximum` can only trim the ends of the ladder; omitting viewer from the middle meant threading a per-resource level list through the settings payload and two frontend call sites just to stop the dropdown offering a level the API would reject. Not worth it for this, so the standard four levels stand and `manager` behaves like `editor`.

One mechanical note for review: three access control logic files carry hand-inlined `APIScopeObject` string-literal unions, so `'llm_playground'` had to be added at 22 separate sites. I scripted that, restricting it to unions that already carried `'toolbar'` so stale unions were left alone, and the full typecheck confirms none were missed.

---


richardsolomou (approved):
Could we also assert the access-denied toast in the keyboard-submit test? That would cover both blocking the request and explaining why nothing happened.

PR #73951, opened 2026-07-27
fix(replay-vision): show ineligible reasons on scan views

willwearing:
## Problem

Replay Vision's on-demand scan table and observation detail status pill render `ObservationStatusTag` without the observation's `error_reason`. The tooltip only mounts when it has content, so ineligible statuses have no hover explanation in those views.

## Changes

- Preserve the observation reason in `scannerRunTabLogic` row state and pass it to the scan table's status tag.
- Pass the existing reason to the observation detail status tag.
- Extend the existing Kea logic test to cover preservation of an ineligible reason.

Screenshot omitted because this restores content in an existing tooltip without introducing new UI.

- `pnpm --filter=@posthog/frontend jest --runTestsByPath ../products/replay_vision/frontend/replay_scanners/scannerRunTabLogic.test.ts --runInBand`
- `pnpm exec oxlint --quiet` on the four changed files
- `pnpm exec oxfmt --check` on the four changed files
- `flox activate -- bash -c "./bin/hogli ci:preflight --fix"`

The extended Kea test catches the shipped regression where an ineligible observation reaches the run tab but loses its reason before rendering.

`pnpm --filter=@posthog/frontend typescript:check` could not complete in the fresh worktree because workspace packages including Quill and HogVM were unbuilt.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs update. This is a wiring fix for existing tooltip behavior.

## 🤖 Agent context

Codex authored the patch using the repository's `/writing-tests` skill and the GitHub publish workflow. I extended the existing Kea test because the reason is lost in logic before render, then fixed both affected status-tag call sites. Human review is required.


arnohillen:
Reviewed this and dug into it locally. The diagnosis is right: `ObservationStatusTag` already accepted `errorReason`, and LemonUI's `Tooltip` returns its children unwrapped when the title is falsy, so those two call sites rendered a pill with no hover affordance at all. Two of the four call sites already passed the reason, so this makes them consistent.

I pushed one commit rather than leaving review comments, because the fix was a line away from being permanent.

**`errorReason` is now a required prop.** Optional meant forgetting it degraded the UI silently, which is exactly how this shipped. Required moves the check to the compiler. The one genuinely reason-less call site (the synthetic pending pill) passes `errorReason={null}` explicitly.

I checked that the guard really catches the original bug by re-introducing the exact omission on the detail page:

```
ReplayObservation.tsx(459,30): error TS2741: Property 'errorReason' is missing in type
'{ status: ObservationStatusEnumApi; }' but required in type
'{ status: ObservationStatusEnumApi; errorReason: string | null; }'
```

One error, nothing else in the tree, so this regression cannot come back quietly.

**Restored the coverage the fixture change dropped.** `s2` had been converted from `succeeded` to `ineligible`, which left that test with no success case. I added `s3` for the ineligible reason and kept `s2` as `succeeded`, so one assertion now covers running, succeeded, and ineligible.

**Simplified the row type** from `ReplayObservationApi['error_reason'] | null` (which derives a non-nullable `string` and then widens it) to plain `string | null`.

## Before and after

Hovering the Ineligible pill on a scanner's on-demand tab, local dev:

### Before

ineligible-tooltip-before

### After

ineligible-tooltip-after

## Verified locally

- `scannerRunTabLogic.test.ts` 3/3, re-run after the master merge and after the pre-commit formatter
- `typescript:check` clean repo-wide. Build `@posthog/quill` and `@posthog/hogvm` first (`pnpm --filter '@posthog/quill...' --filter '@posthog/hogvm...' build`), that unbuilt-workspace noise is what drowned your local run in ~195 unrelated errors
- Confirmed the observations list endpoint returns `error_reason` in list responses, so the run tab genuinely has the data to pass through

## Two things worth fixing in the description

1. The kea test catches the logic-layer loss, not the render omission. The detail page half was a pure JSX omission on data already in hand, so the required prop is what covers it now.
2. Screenshots are above, so the "omitted because no new UI" note can go.

Worth knowing for context: on the detail page the reason is already rendered in the body when it parses, so that tooltip is redundant in the common case. It does help on format drift, where the body block is suppressed and previously nothing was shown at all.

willwearing:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists + generalists), paul-reviewer, xp-reviewer

## Verdict: ✅ APPROVE

All 3 independent reviewers found the change focused and correct. The patch carries `error_reason` through the scanner row model, requires it at every status-tag call site, and the focused Jest suite passes all 3 tests.

### Key findings

- No actionable findings.

### Convergence

- All 3 reviewers independently found no correctness, security, coupling, naming, or test coverage concerns.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Correctly preserves and displays the reason with no regression found. |
| 👤 paul | Clean type-level enforcement and complete call-site coverage. |
| 📐 xp | Focused change, realistic regression test, no unnecessary abstraction. |

arnohillen (approved):

PR #73961, opened 2026-07-27
fix(ci): repair weekly flaky report

rnegron:
## Problem

- The weekly flaky report sends unsupported Slack table cells.
- It filters pytest after a mixed-runner API limit, so Jest rows can hide qualifying backend tests.
- Enrichment was not repository-scoped.
- **The report was blind to Trunk's quarantine**, which is the review question on this PR. Our `quarantined` signal only covers `.test_quarantine.json`, where the pytest adapter xfails the test and the span records `xfailed`. Trunk instead masks the job verdict and leaves a hard failure in the junit, so its quarantines reach the spans as plain failures.

> [!NOTE]
> One Trunk-quarantined test has 16 failures in 7 days, so it would currently headline the report, told to "park it", while CI already ignores it.

## Changes

- Build Slack links with supported `rich_text` cells and surface validation details.
- Add a `runner` API filter before ranking and limiting; request `runner=pytest` from the report.
- Preserve proved flakes, suppress unproved master bursts, and scope enrichment by repository.
- Read Trunk's quarantine state, then park or label those tests.

| Decision | Choice | Why |
| --- | --- | --- |
| Where the join lives | The report script, not the `flaky_tests` endpoint | The endpoint reads `trace_spans` on the LOGS cluster while `trunkio.*` is warehouse, so it is not one query. SPEC §3 also bans hardcoded warehouse table names, and there is no Trunk source resolver. |
| Label vs drop | Both, split by variable | Masking on means CI no longer fails on the test, so it leaves the ranked 10 and is listed below the table. Masking off means Trunk marked it but CI still fails, so it stays ranked with a `(Trunk flagged)` label. |

Dropping them outright was rejected: a long-lived auto-quarantine sitting on real breakage is what nobody notices.

### Control and fail-open

No new toggle. `TRUNK_UPLOAD_ENABLED` gates whether the query runs at all; the second only picks park vs label.

```diff
+ TRUNK_UPLOAD_ENABLED: ${{ vars.TRUNK_UPLOAD_ENABLED }}
+ TRUNK_QUARANTINE_ENABLED: ${{ vars.TRUNK_QUARANTINE_ENABLED }}
+ TRUNK_QUARANTINE_TABLE: ${{ vars.TRUNK_QUARANTINE_TABLE }}
```

Variables unset, table missing, query error, and zero rows all degrade to the previous report.

<details>
<summary>A pytest node id is not a Trunk test id</summary>

Trunk keys a test by `file` + `classname` + `name`, putting the class inside `classname` rather than between file and name. The node id is rebuilt by trimming the file's module prefix off `classname`. Against live data, 6 of 8 quarantined pytest tests match exactly; the 2 misses are product suites reported product-relative, which the existing path-variant logic already covers.

`trunkio.quarantinedtests` has no repository column, so the query cannot be repo-scoped. It does not need to be: a row only annotates a selector the repo-scoped endpoint already returned.

</details>

- 9 Node tests and 29 flaky-tests API tests. The 5 new Node cases cover: degrading when uploads are off / the table is missing / it returns nothing; a product-relative selector matching a repo-relative Trunk path; park-vs-label switching on the masking variable; and a quarantined test that would otherwise be buried in a collapsed cluster row.
- Ran the script end to end with `DRY_RUN=1` in all three variable states against a local stub seeded with real warehouse rows. With the variables unset it issues no Trunk query at all.
- OpenAPI/MCP generation, repo-wide mypy, workflow lint, actionlint, and strict preflight.

> [!WARNING]
> No API key was available in the session, so the live endpoint path is stubbed rather than exercised. The SQL that rebuilds node ids was validated by running it against the real warehouse, not by a unit test.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the Engineering analytics README with the runner filter behavior.

## 🤖 Agent context

- Codex for the original fixes, then Claude for the Trunk work, with the CI workflow, tests, copy, and code comment skills.
- Confirmed the trunkio schema and the pytest-vs-jest discriminator against live data before designing the join, rather than assuming a node id matches Trunk's test id. It does not.
- A review pass caught two defects, fixed here: a test asserted the parked note was the last Slack block, which broke once Actions set `GITHUB_WORKFLOW_REF` and appended one after it; and parking ran after cluster collapsing, so a quarantined test could hide inside a cluster row whose bare file path can never match a Trunk node id. Parking now runs first.

### DIFF
--- a/.github/scripts/weekly-flaky-report.mjs
+++ b/.github/scripts/weekly-flaky-report.mjs
@@ -137,9 +137,7 @@ function isMasterBurst(item) {
 }
 
 function selectReportCandidates(items) {
-    return collapseClusters(
-        items.filter((item) => item.runner === REPORT_RUNNER && !isMasterBurst(item)).slice(0, CANDIDATE_POOL)
-    )
+    return items.filter((item) => item.runner === REPORT_RUNNER && !isMasterBurst(item)).slice(0, CANDIDATE_POOL)
 }
 
 // Product suites run from their product dir, so a selector path may be repo- or
@@ -277,10 +275,9 @@ async function enrich(items, runHogql = hogql) {
 // rows would have to carry an identical node id to collide.
 const TRUNK_QUARANTINED_QUERY = `
     SELECT concat(file, '::', if(cls = '', '', concat(cls, '::')), name) AS nodeid,
-        quarantined_at,
-        quarantine_setting
+        quarantined_at
     FROM (
-        SELECT file, name, quarantined_at, quarantine_setting,
+        SELECT file, name, quarantined_at,
             replaceAll(substring(file, 1, length(file) - 3), '/', '.') AS module,
             if(startsWith(classname, concat(module, '.')),
                replaceAll(substring(classname, length(module) + 2, length(classname)), '.', '::'),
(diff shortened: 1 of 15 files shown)


gantoine (approved):
does this report need context of trunk flaky tests/quarantining in any way or does it run before that?

PR #61547, opened 2026-06-04
feat(signals): configurable base branch for inbox auto-PRs

tatoalo:
## Summary

configurable base branch for inbox auto-PRs (PostHog/code#2470). Stores per-repository base branch overrides and applies them when auto-starting implementation tasks, so inbox PRs can target a chosen branch instead of the repo default if needed

configurable_base

PR #73967, opened 2026-07-27
feat(signals): add base branch overrides to inbox settings

christiaan-ph:
## Problem

Self-driving already supports per-repo base branches for the PRs it opens. `SignalTeamConfig.autostart_base_branches` maps `org/repo` to a branch, and `auto_start` passes it through as the task branch, so the sandbox checks that branch out and the PR targets it instead of the repo default.

PostHog Desktop exposes this under Settings > Self-driving. The web app never did. The field appears only in `types.ts` and a Storybook mock, and it's read-only in Django admin, so from the web app the only way to set it was calling the config API by hand. Teams on a develop-then-master or QA-then-main flow reasonably conclude self-driving can't target their integration branch, when it can. This came up via a customer support ticket asking exactly that.

## Changes

Adds a "Base branches" control to the PR generation card in inbox settings, shown when PR generation is on, mirroring the Desktop app's version of the same setting.

- Existing overrides render as rows with a live branch picker and a remove button, so a branch can be changed in place. Each row resolves its GitHub integration by matching the repo's org prefix against the integration account name. Rows with no matching integration (nothing connected, or an org the team no longer has) degrade to plain text but stay removable, so stale entries can't get stuck and can never list branches from the wrong integration.
- New overrides are added with the existing `GitHubRepositoryCombobox` and `GitHubBranchCombobox`, the same pickers the task repo selector uses. Free text was the wrong call here: a mistyped repo key silently never matches, and a mistyped branch fails much later inside the sandbox at checkout, so neither mistake surfaces anywhere near where it's made.
- An organization switcher appears only when more than one GitHub integration is connected.
- With no GitHub integration connected, the section points at the connect step directly above it.

Both the add, update, and remove paths persist the full map, since the API replaces `autostart_base_branches` wholesale rather than merging. The repo key is lowercased and the branch trimmed client-side to match what the serializer stores, so the optimistic value equals what comes back.

This supersedes #73469, which added the same capability with two free-text inputs.

> [!NOTE]
> No backend change. The field, its validation, and the auto-start behavior already shipped in #61547.

**Not manually tested.** Claude wrote this code and I reviewed it, but my local environment has no GitHub integration configured, so neither of us has rendered this against a real integration. It needs a click-through from someone with GitHub connected before merging. Specifically worth checking:

- [ ] The picker row and override rows fit the narrow setup rail (`AgentSetupColumn`) without awkward wrapping
- [ ] Adding an override round-trips and survives a reload
- [ ] Changing a row's branch persists and other repos' entries are untouched
- [ ] The org switcher appears with two or more GitHub integrations and the repo list follows it
- [ ] With GitHub disconnected, existing rows still render and can be removed

Automated, actually run:

- `jest --testPathPattern signalTeamConfigLogic`, 8 cases, all passing. The file is new; no test covered this logic before.
- `pnpm --filter=@posthog/frontend typescript:check`. No errors in the changed files; the repo-wide run has pre-existing failures on an unrelated tree, identical with and without this branch.
- `hogli ci:preflight --fix`, clean.

What each test group catches:

- **Add preserves the map** and lowercases the key, trims the branch, clears the draft. Guards the data-loss shape: the API replaces this field, so an add that writes only the new entry silently drops every other repo's override.
- **Update changes only the target repo's branch.** Same wholesale-replace hazard from the edit path.
- **Re-selecting the stored branch doesn't hit the API.** The combobox fires `onValueChange` on a same-item re-select, and without the guard every re-open of a row's picker could write the map.
- **Remove deletes only the target.**
- **Incomplete draft doesn't reach the API** (three parameterized cases: no slash, empty half, blank branch). Guards against writing a key the backend would 400 on.
- **Drafted branch is dropped when the repo changes.** Without this, picking repo A, choosing its branch, then switching to repo B would point B at a branch that only exists in A, which fails at sandbox checkout.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed. The field is already documented in `products/signals/ARCHITECTURE.md`.

## 🤖 Agent context

Claude Code (Opus 5) wrote this after I asked it to verify a support question about whether the inbox agent's PR base branch is configurable. Tracing `autostart_base_branches` from the serializer through `auto_start.py` into the sandbox checkout showed the capability was complete and only the web UI was missing.

Skills invoked: `/adopting-generated-api-types`, `/writing-tests`, `/writing-user-facing-copy`.

- **Kept the manual `api.signalTeamConfig` client** instead of generated functions. `SignalTeamConfigViewSet` marks both actions `@extend_schema(exclude=True)`, so there's no generated function or type to adopt. Making the endpoint schema-visible is a bigger change than this PR should carry.
- **Pickers over free text**, as above. This is the main difference from #73469.
- **Per-row integration resolution is a heuristic.** Desktop resolves the owning integration per repo via its aggregated repo list; the web comboboxes are per-integration with server-side search, so there's no cheap equivalent. Matching the repo's `org/` prefix against `display_name` works because the backend sets a GitHub integration's display name to the installation's account login (`integration.py`, `account.name` from `account.login`). Rows with no match render as plain text rather than guessing an integration, since a wrong integration's branch list would let a user persist a branch the repo doesn't have (Greptile's one review finding, fixed in add51d5). Flagging in case a reviewer knows a firmer lookup.
- **The add picker doesn't hide already-configured repos** (Desktop's does). Re-picking a configured repo updates its entry in place rather than duplicating, so the outcome is equivalent; hiding would mean adding an exclude prop to the shared combobox.
- `GitHubBranchCombobox` prefills the repo's default branch when nothing is selected, so a freshly picked repo shows its default before you change it. Committing that value is a no-op against current behavior.


tatoalo (approved):

Twixes:
Pushed some UI tweaks:

Screenshot 2026-07-30 at 16 30 48@2x

Twixes (approved):

PR #73971, opened 2026-07-27
feat(customer-analytics): point account MCP tools at custom property definitions

### LINKED ISSUE #73924: Improve Customer analytics custom properties MCP tools discoverability
The agents currently struggle a bit to find the correct custom properties. We never ask questions about the accounts, so we need to nudge them to take a look at the custom property definitions to understand better what data they have available and use that data instead of running random queries.

AlexlaGuardia:
Closes #73924.

## Problem

An agent asked about accounts reaches for `accounts-list` (or `accounts-retrieve`) first. Both descriptions enumerate the built-in typed properties (`csm`, `account_executive`, `stripe_customer_id`, `hubspot_deal_id`, and so on) and never mention that team-defined custom properties exist.

So the agent reasonably concludes the built-ins are the whole schema, and when the answer isn't in them it starts guessing at property names and running queries to find out. That's the behaviour described in the issue.

The information it needs is already written and good. `custom-property-definitions-list` documents the definition shape and even spells out the HogQL path for bulk reads against `system.accounts`. The cross-references to it exist too, but only on `accounts-custom-property-values-create` and `accounts-custom-property-values-list` — tools an agent only discovers *after* it already knows custom properties are a thing.

That's the gap: it isn't missing documentation, it's a missing forward pointer at the one place every agent starts.

## Change

Adds the pointer to the two entry-point tools:

- **`accounts-list`** — states the returned `properties` are built-ins only, and to call `custom-property-definitions-list` first to see what a team actually tracks, rather than assuming the built-ins are the whole schema or guessing at names.
- **`accounts-retrieve`** — same, then routes on to `accounts-custom-property-values-list` for that account's values.

Description text only, in the existing house style. No schema, scope, annotation, or handler changes.

## Notes

- Edits the YAML source, not the generated TypeScript.
- The tool-level description text lives only in `products/customer_analytics/mcp/tools.yaml` — nothing generated or snapshotted carries it, so this needs no codegen run and produces no snapshot churn. (Confirmed by grepping the description text across `services/` and `products/`: single hit, this file. The customer-analytics tools also have no entries under `services/mcp/tests/unit/__snapshots__/tool-schemas/`.)
- YAML parses clean and line wrapping matches the surrounding file.

Happy to extend the same pointer to other entry points, or to word it differently if you'd rather nudge toward the HogQL path than the per-account value tools.


lricoy:
Hey @AlexlaGuardia thank you for the PR! Are you using this yourself?

AlexlaGuardia:
Hey @lricoy, no. I'm not a Customer Analytics user, so I'll be straight about where this came from. Arthur's issue #73924 came across my feed and the failure mode was readable from `tools.yaml` alone: `accounts-list` and `accounts-retrieve` both enumerate the built-in properties and then stop, so an agent reading the catalog has no reason to believe custom properties exist at all. It guesses names or falls back to a query instead of calling `custom-property-definitions-list`.

That made it a catalog problem rather than a docs one, which is why the diff is description text and nothing else.

If you'd rather that nudge live somewhere central than on two tool descriptions, say so and I'll close this.

products/customer_analytics/mcp/tools.yaml
```diff
@@ -355,7 +359,10 @@ tools:
         description: >
             Fetch a single Customer Analytics account by id. Returns the full account including typed `properties` (csm,
             account_executive, account_owner assignments plus stripe_customer_id, hubspot_deal_id, billing_id, sfdc_id,
-            zendesk_id external identifiers) and `tags`. To link a user to this account rather than the bare list,
+            zendesk_id external identifiers) and `tags`. These are the built-in fields only; a team's own typed
+            attributes are custom properties and are not returned here. Call `custom-property-definitions-list` to see
+            what the team tracks, then `accounts-custom-property-values-list` for this account's values. To link a user
```
arthurdedeus, line 363: this is pretty spot on

arthurdedeus (approved):
both additions look solid! thanks for contributing

PR #73987, opened 2026-07-27
chore(onboarding): remove the quickstart landing page and its flag

fercgomes:
## Problem

The `quickstart-homepage` experiment ([#385438](https://us.posthog.com/project/2/experiments/385438)) has been ended as **lost**. A catalog of products as the post-onboarding landing page made the UI more overwhelming rather than less: people sign up with one specific task in mind, and a product catalog pulls them away from it.

So the whole exploration comes out and the post-onboarding destination returns to the AI home.

## Changes

- Deletes the `/quickstart` scene and its logic, plus the `QUICKSTART_HOMEPAGE` flag constant, the sidebar entry, the Configure homepage option, and the flag-gated redirect after onboarding.
- **Keeps** the completion fixes that shipped alongside the experiment: both onboarding-completion signals are still persisted together before navigating, and setup tasks are still ticked only after that PATCH commits.
- `sceneLogic` now ignores a persisted homepage whose scene no longer ships. Without this, anyone who had picked Quickstart as their homepage would have `/` redirect into a dead route on every visit, with no way out except reconfiguring it. It also guards the next scene removal.

Not manually tested in a running app — no dev stack in this environment. Automated checks actually run:

- `jest src/scenes/onboarding src/layout/navigation src/scenes/sceneLogic.test.*` — 461 passing. The onboarding suite is the meaningful one: it asserts the completion redirect now falls back to `/` for every product without its own destination, and that both completion signals are still persisted.
- `oxlint` and `oxfmt --check` clean; `hogli ci:preflight --strict` reports 0 failures.
- `tsgo --noEmit` reported no errors in any changed file. The full pass could not run to completion here (it exhausts memory in this sandbox once the quill workspace packages are built), so CI is the real gate for repo-wide types.

Tests removed rather than added: the two helpers extracted purely to hold the Quickstart check (`shouldShowNoEventsProjectNotice`, `shouldHideWizardSyncCard`) are inlined back, so their unit tests go with them.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No published docs describe the post-onboarding destination, so nothing to update.

## 🤖 Agent context

Built with PostHog Code (Claude) directed by Fernando. Because the checkout is a shallow clone, the revert was reconstructed from the GitHub API rather than `git revert`: the introducing PR's diff was pulled file by file and separated from the follow-ups that touched the same files, so the Cmd+K nav work and the SDK-onboarding additions stayed intact.

Two judgement calls worth a reviewer's attention:

- The original PR bundled a genuine fix in with the experiment — writing `completed_snippet_onboarding` alongside `has_completed_onboarding_for` so a completed user isn't bounced back into onboarding after a refresh. That is kept, along with the later ordering fix for the racing team PATCHes. Only the flag branch and the homepage persistence were pulled out.
- Deleting the scene would have stranded users whose homepage pointed at it, so the `sceneLogic` guard was added. That is the one line here that isn't a straight revert. A backend cleanup of those stored settings would also work, but the frontend guard is cheaper and generalizes.

Where a helper existed only to carry the Quickstart branch it was inlined back; where a helper stands on its own without mentioning Quickstart (`getHomepageMode`) it was kept rather than churned back into a nested ternary.

Skills invoked: `/writing-tests` and `/writing-code-comments` rules applied via repo instructions; no scene, migration, or DRF skills were relevant to a deletion-only change.

The production flag `quickstart-homepage` still exists at 100% control and is safe to delete once this deploys.

---


rafaeelaudibert (approved):
Do we need to do anything with the folk who ended up with "quickstart" as their homepage in the DB? Is that even something that could've happened?

fercgomes:
afaik, no. everything is front-end only

fercgomes:
> Do we need to do anything with the folk who ended up with "quickstart" as their homepage in the DB? Is that even something that could've happened?

It could have, and it isn't purely front-end: the homepage choice is persisted server side in `UserHomeSettings.homepage` (a JSONField in `posthog/models/user_home_settings.py`), written through `api/user_home_settings/@me/` and bootstrapped back into `APP_CONTEXT` in `posthog/utils.py`. First-time onboarders in the test arm got Quickstart written there automatically by `setQuickstartAsDefaultHomepageOnce`, and that arm wasn't empty, so some of those rows do exist.

It's handled in the read path rather than with a migration. `getBootstrappedHomepage()` in `sceneLogic` now drops any persisted homepage whose `sceneId` is no longer in `sceneConfigurations`, so those rows go inert and the user falls back to the project default. That guard is load-bearing rather than defensive: without it, `/` would redirect into a dead route on every visit, and combined with the `/quickstart` compat redirect added for the bookmark case the two would bounce off each other indefinitely. There's a test pinning it (`ignores a bootstrapped homepage whose scene no longer ships`), verified to fail when the guard is removed.

The residue is that the stale JSON stays on the row until the user next picks a homepage in the Configure homepage modal. Same for anyone who pinned Quickstart as a sidebar tab (`UserHomeSettings.tabs`), except the compat redirect now sends those clicks home instead of to a 404. I went with the read-path guard over a data migration because it also covers the next scene removal, but happy to add a migration that nulls the affected `homepage` values if you'd rather the rows were actually clean.

PR #73991, opened 2026-07-27
feat(tasks): mark wizard cloud runs in task telemetry

gewenyu99:
Adds `is_wizard_run` to task activity telemetry, derived from the protected `wizard_config` state key, so cloud setup-wizard runs are distinguishable from other `origin_product=onboarding` tasks (which callers can set themselves).

---


gewenyu99:
I don't think this is it

PR #73993, opened 2026-07-27
chore(query): relocate query-adjacent modules from posthog/caching

andyzzhao:
## Problem

With the query cache consolidated into `posthog/query_cache/` (PR 73425), the `posthog/caching/` folder was left holding a confusing mix: insight API bridge code, the insight cache warming tasks, and the staleness policy all look query-cache-related but are not part of the cache, while the folder also holds genuinely unrelated Redis plumbing and per-feature caches. 

## Changes

Four relocations with import updates and no behavior change:

- `QUERY_CACHE_ALIAS` moves into `posthog/query_cache/backend.py`. It is the query cache's own Redis alias name, and this severs the package's last import from `posthog/caching/`.
- The insight API bridge (`calculate_results.py`, `insight_result.py`, `insights_api.py`) moves to `products/product_analytics/backend/`. It is insight code that owed nothing to caching except history.
- `warming.py` moves to `products/product_analytics/backend/` too, taking the warming-only `largest_teams()` helper from `utils.py` with it. Celery references updated: the module list in `settings/celery.py`, `tasks/scheduled.py`, the logger config, and the task-registration snapshot.
- The staleness policy (`is_stale`, `cache_target_age`, `ThresholdMode`) moves to `posthog/hogql_queries/staleness.py`, next to its main consumer. It reads `Team` and interval specs, so it is request-layer policy, mirroring how exception-to-kind classification lives app-side of the cache package. `posthog/caching/utils.py` is deleted.

What deliberately stays in `posthog/caching/`: the zstd compressor and the Redis cluster connection factory (settings reference them by dotted path, and the compressor serves every cache alias) plus the five per-feature Redis caches. After the moves the folder is coherently "Redis cache plumbing and per-feature caches".

> [!NOTE]
> Moving `warming.py` renames its celery tasks to the new module path. Queued warm tasks under the old names are dropped during the deploy window; warming is a best-effort hourly job, so the worst case is some insights staying stale one extra cycle, once.

Automated only. All 311 tests across the moved and consuming suites collect cleanly (which imports every touched module through Django), the DB-free suites pass, ruff and repo-wide mypy are clean. The DB-backed suites (warming, calculate_results, task-registration snapshot, query runner) run in CI; the local dev database was unavailable in this session, so they were not run locally.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal refactor; no user-facing behavior change.

## 🤖 Agent context

Authored by Claude (Claude Code session) directed by the PR assignee, as the cleanup follow-up to the query cache consolidation stack (PRs 73425, 72777, 73651). Decisions made with the driver: keep the folder name (contents become honest once the query-adjacent tenants leave, so no rename churn), leave plumbing and feature caches in place (settings dotted-path references and settings-load-time imports make moving them risk without payoff), accept the one-deploy celery task-name blip for warming rather than carrying legacy task-name aliases forever, and place staleness in the query layer rather than the cache package to keep the package free of Team and interval knowledge.


posthog/hogql_queries/staleness.py
```diff
@@ -95,3 +77,16 @@ def is_stale(
...
+# From https://github.com/PostHog/posthog-cloud-infra/blob/master/ansible/config/clickhouse-users.xml#L11
+# Keep in sync with the above! And note that this doesn't hold for async queries (which are flagged as of Feb 2023)
+CLICKHOUSE_MAX_EXECUTION_TIME = timedelta(seconds=180)
+
+# Default minimum wait time for refreshing an insight
+BASE_MINIMUM_INSIGHT_REFRESH_INTERVAL = timedelta(minutes=15)
+# Wait time for short-term insights
+REDUCED_MINIMUM_INSIGHT_REFRESH_INTERVAL = timedelta(minutes=3)
+# Wait time for "real-time" insights
+REAL_TIME_INSIGHT_REFRESH_INTERVAL = timedelta(minutes=1)
```
MattPua, line 92: what defines a short time insight?

MattPua (approved):

PR #74012, opened 2026-07-27
fix(warehouse): coerce Intercom owner_id and waiting_since to string

kyleswank:
## Problem

The Intercom data warehouse sync fails on two tables with PyArrow schema-merge errors:

- `contacts`: `ArrowTypeError: Unable to merge: Field owner_id has incompatible types: int64 vs string`
- `conversation_parts`: `ArrowTypeError: Unable to merge: Field waiting_since has incompatible types: string vs int64`

Intercom returns `owner_id` (a contact's admin identifier) and `waiting_since` (a Unix epoch value on conversation parts) inconsistently: an integer in some records, a string in others. The pipeline builds one Arrow table per page and infers each column's type from that page's data, so pages disagree. When they get concatenated before the Delta write, PyArrow can't unify `int64` with `string` and the whole sync fails.

## Changes

Normalize both fields to string at the source, before any row reaches Arrow, so every batch infers one stable type. String is the safe target: casting int to string never fails, and it matches the codebase's existing "ambiguous field to string" convention (UUID and IP handling in the pipeline utils). `None` is left as-is so the column stays nullable.

The two endpoints reach Arrow by different paths, so the coercion lands in two places, both driven by one new declarative config field `coerce_string_fields` on `IntercomEndpointConfig`:

- `contacts` flows through the REST framework, so it's coerced via a `data_map` transform (the framework's per-item hook, same pattern the guru/webflow/campayn sources use).
- `conversation_parts` is a hand-written substream that bypasses the REST framework, so it's coerced inline in `_substream_items`. This also centralizes coercion for any future substream field.

> [!NOTE]
> This clears the failure for every sync going forward because each sync now produces internally consistent types. For a table whose Delta schema was already established with these columns as `int64` by an earlier partial sync, incoming string values are cast back to the stored `int64` (numeric IDs cast cleanly), so the column keeps its original type rather than flipping to string. To actually store the column as string, reset and re-sync that schema. No action is needed just to clear the error.

Automated tests I (Claude) actually ran, all green (`hogli test .../intercom/tests/test_intercom.py`, 107 passed), plus repo-wide `uv run mypy --cache-fine-grained .` (clean) and `hogli ci:preflight --strict` (0 failures). I did not run a live Intercom sync.

New tests in `test_intercom.py` (`TestCoerceStringFields`), each catching a regression no existing test did:

- Parameterized `_coerce_fields_to_str` cases: guards the core coercion (int to string, string unchanged, `None` preserved, missing field is a no-op, only configured fields touched). This is the direct guard for the int64-vs-string merge failure.
- `contacts` REST-path wiring: if `get_resource` stops attaching the `data_map`, `owner_id` silently stops being coerced. This asserts the data_map is present and stringifies `owner_id`.
- `conversation_parts` substream-path wiring: coercion lives in `_substream_items` (bypasses the framework data_map). Feeds parts with mixed int/string/None `waiting_since` and asserts they all normalize to string/None.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude (Opus 4.8) via PostHog Code, directed by Kyle. Skills invoked: `/writing-tests` (gated the new tests). Explored the Intercom source and the data-imports Arrow/merge path first to confirm the root cause was per-batch type inference on vendor-inconsistent fields, not the Delta write.

Considered fixing at the merge boundary (a general cast step before concat) but rejected it as broader and riskier than normalizing at the source. Chose string over integer for the target type after confirming int coercion can throw on a genuinely non-numeric value while string coercion never fails. Made `coerce_string_fields` a declarative config field rather than hardcoding the two field names, so future type-flip fields are a one-line addition.

---


products/warehouse_sources/backend/temporal/data_imports/sources/intercom/intercom.py
```diff
@@ -195,6 +195,17 @@ def _build_paginator(cfg: IntercomEndpointConfig) -> BasePaginator:
     return SinglePagePaginator()
 
 
+def _coerce_fields_to_str(item: dict[str, Any], fields: list[str]) -> dict[str, Any]:
```
danielcarletti: Can we move this to a helper? Its probably already implemented somewhere but we can definitely reuse it

danielcarletti (approved):

products/warehouse_sources/backend/temporal/data_imports/sources/intercom/intercom.py
kyleswank: Good call. There wasn't an existing shared helper (Linear and Elasticsearch each hand-rolled their own float-coercion variants), so I pulled this out into a reusable `coerce_fields_to_str` in `common/row_transforms.py` and wired Intercom to import it. Any future source hitting the same JSON-type-flip problem can reuse it now. Done in a0f01e9.

PR #74014, opened 2026-07-27
fix(aio): don't fail eval emission when team is over billing quota

Radu-Raicea:
## Problem

`emit_evaluation_event_activity` (and the trace- and tagger-level emit activities) spiked in errors and drove the `LLMAEvalsHighErrorRate` alert. Root cause: when an evaluation fires for a team that's over its ingestion quota, `capture_internal` returns a `402 billing_limit_exceeded` whole-request failure. `raise_for_status()` raises `CaptureInternalError`, the activity re-raises, Temporal retries it to the max (the quota can't recover in the retry window), and the whole workflow fails — counted as a system error.

That's a customer billing condition, not a system fault: ingestion would drop the team's events anyway, so there's nothing to recover by retrying or alerting.

## Changes

- Classify billing-limit failures through `CaptureInternalError.is_billing_limit_exceeded`, then swallow them in the generation, trace, and tagger emit activities so the workflow returns cleanly instead of retrying.
- For the eval emit paths, record a distinct `dropped_billing_limited` emission outcome, kept out of the `failed` bucket the error-rate alert watches.
- Any other capture error still fails and retries exactly as before.

Added parameterized activity tests (`402` → swallowed, non-402 → still raises) for all three emit activities: `test_run_evaluation.py`, `test_run_trace_evaluation.py`, and `test_run_tagger.py`. Each guards the exact regression here — a billing-limited team's emit exhausting retries and failing the workflow / tripping the alert — which no existing test covered.

The focused billing-limit activity tests pass locally for all three emit paths (six cases total). Ruff lint and format checks pass on all changed files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Investigated from a workflow-failure stack trace shared in the originating Slack thread. Traced `CaptureInternalError` back through `capture.py:raise_for_status` (which carries `status_code`) to the 402 quota case, then found all three emit activities share the same raise-and-retry shape. Invoked `/writing-tests` before adding coverage and scoped tests to the one regression each catches. Extended the fix to the tagger emit path since it had the identical latent bug even though only the eval path fired the alert.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0A006STKHR/p1785054604085999?thread_ts=1785054604.085999&cid=C0A006STKHR)*


posthog/temporal/ai_observability/evaluation_workflow_activities.py
```diff
@@ -325,6 +326,15 @@ def _emit() -> None:
     try:
         await database_sync_to_async(_emit, thread_sensitive=False)()
         increment_emit_event_outcome("success")
+    except CaptureInternalError as e:
+        if e.status_code == HTTPStatus.PAYMENT_REQUIRED:
+            # Team is over its ingestion quota, so ingestion would drop this event anyway.
+            # That's a billing condition, not a system fault — don't retry it or count it as an error.
+            increment_emit_event_outcome("dropped_over_quota")
```
carlos-marchal-ph: "quota" already means something else in this package: `quota_error` in `evaluation_errors.py:77` and `increment_user_errors("quota_error")` in `evaluation_llm_judge.py:354` both mean the provider API key is out of credit.

Could we call it `dropped_billing_limited` or `dropped_ingestion_limited`? Cheap now, annoying once someone has built a panel on the label.

posthog/temporal/ai_observability/evaluation_workflow_activities.py
```diff
@@ -325,6 +326,15 @@ def _emit() -> None:
...
+            logger.info("Skipping eval event emission; team over billing quota", team_id=event_data["team_id"])
+            return
+        increment_emit_event_outcome("failed")
+        raise
```
carlos-marchal-ph, line 334: There are two `increment_emit_event_outcome("failed")` sites now, so whoever adds the next status-code branch has to remember to re-increment. Wouldn't inverting the check read better, given only 402 is special?

```suggestion
    except CaptureInternalError as e:
        if e.status_code != HTTPStatus.PAYMENT_REQUIRED:
            increment_emit_event_outcome("failed")
            raise
        # Team is over its ingestion quota, so ingestion would drop this event anyway.
        # That's a billing condition, not a system fault — don't retry it or count it as an error.
        increment_emit_event_outcome("dropped_over_quota")
        logger.info("Skipping eval event emission; team over billing quota", team_id=event_data["team_id"])
        return
```

posthog/temporal/ai_observability/test_run_evaluation.py
```diff
@@ -375,6 +376,51 @@ async def test_emit_evaluation_event_activity(self, setup_data):
...
                 assert props["$ai_evaluation_type"] == "online"
 
+    @pytest.mark.asyncio
+    @pytest.mark.django_db(transaction=True)
+    @pytest.mark.parametrize(
+        "status_code,should_raise",
+        # A 402 means the team is over its ingestion quota, which is a billing condition we swallow
+        # so the workflow succeeds instead of exhausting retries on a request that can't recover.
+        [(402, False), (500, True)],
+    )
```
carlos-marchal-ph, line 387: Six of the eight parametrize blocks in this file use `pytest.param(..., id=...)`, so failures read as `[over_quota]` rather than `[402-False]`. Could we match them and fold the comment into the ids?

```suggestion
    @pytest.mark.parametrize(
        "status_code,should_raise",
        [
            pytest.param(402, False, id="over_quota_is_swallowed"),
            pytest.param(500, True, id="server_error_still_raises"),
        ],
    )
```

posthog/temporal/ai_observability/metrics.py
```diff
@@ -131,10 +131,12 @@ def increment_tokens(token_type: str, count: int) -> None:
...
 
 def increment_emit_event_outcome(outcome: str) -> None:
-    """Track $ai_evaluation event emission outcomes (success/failed).
+    """Track $ai_evaluation event emission outcomes (success/failed/dropped_over_quota).
 
     Distinguishes Activity 4 failures from other workflow failures so we can
-    measure and alert on dropped eval events specifically.
+    measure and alert on dropped eval events specifically. `dropped_over_quota`
+    is an expected billing condition, not a system failure, so it's kept out of
+    the `failed` bucket the error-rate alert watches.
```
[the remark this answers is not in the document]
carlos-marchal-ph: I think it's right

carlos-marchal-ph (approved):
The same five lines now appear in three files, along with the same comment and the same `HTTPStatus` import. The fact being encoded, that a whole-request 402 from `capture_internal` means the team is over its ingestion quota, is a property of `capture_internal` rather than of AI observability.

Would an `is_over_quota` property on `CaptureInternalError`, or a subclass of that error, be worth it, so the semantics live next to the code that produces them and the other products get it for free?

PR #68725, opened 2026-07-06
chore(tests): deflake backend test offenders

rnegron:
## Problem

Backend CI has several high-signal flaky tests that fail on the first attempt and pass on a pytest rerun.

This PR now keeps only test-local determinism fixes.
Product-adjacent changes and assertion-weakening changes were removed from scope so the PR stays in DevEx territory.

## Changes

Kept fixes:

- **embeddings TTL**: seed document embeddings inside the ClickHouse table TTL window.
- **temporal query_log**: run `SYSTEM FLUSH LOGS` before polling `system.query_log`.
- **sessions**: use two distinct session IDs for property-value fixtures so merged raw sessions do not collapse `utm_source`.
- **oauth**: freeze the clock for JWT `iat` bounds checks.
- **hog-functions**: assert reload calls without relying on unordered database result order.
- **flush-guard**: test the guard against a dedicated probe table instead of Django's full flush.
- **trends**: assert tied cohort breakdown rows by cohort instead of response position.

Removed from this PR after review:

- `products/user_interviews/backend/facade/__init__.py` import-side-effect cleanup.
- Warehouse query-folder clocking changes.
- Cohort membership set assertion.
- Group property CRUD `+1s` timestamp changes.
- Snowflake wrong-passphrase message assertion weakening.

## How did you test this code?

Ran the narrowed affected test set:

```bash
hogli test ee/api/test/test_admin_oauth2_middleware.py posthog/api/test/test_session.py posthog/hogql_queries/insights/trends/test/test_trends_query_runner.py posthog/hogql_queries/test/test_document_embeddings_query_runner.py posthog/management/commands/test/test_resave_hog_functions.py posthog/temporal/tests/test_clickhouse.py posthog/test/test_flush_lock_guard.py
```

Result: `299 passed in 595.00s`.

Also ran:

```bash
hogli ci:preflight --strict
```

Result: exited 0 with ruff and format checks passing.
It reported an advisory that the branch is behind `master` and OpenAPI drift exists on the branch state; there were no preflight failures for the changed test files.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

No docs update.
This is a test-only cleanup.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

[...]

PR #74020, opened 2026-07-27
fix(ci): key the schema cache on product db routing

rnegron:
## Problem

Backend CI restores a cached Postgres schema instead of replaying every migration. That key never accounted for which products live on their own database, so a job could restore a dump that disagrees with the code and hard-fail:

| Routing differs by | Then `migrate` |
| --- | --- |
| label added | replays from 0001, `CREATE TABLE` on an existing table |
| label removed | skips those migrations, the table is never created |

A cache miss falls back to a full migrate, but a successful restore of the wrong dump does not, so the job dies with no obvious cause.

<details>
<summary>Why the dump and the code can disagree</summary>

The dump is built under `TEST=1`, where `ProductDBRouter` routes the app labels in `products/db_routing.yaml` to separate databases. Their tables never reach the dump, but `django_migrations` still records those migrations as applied. Consumers compensate by deleting exactly those rows so `migrate` replays them, reading the app list from their own checked-out copy of the config rather than the copy the dump was built from.

</details>

## Changes

Routing drifts from two sides, so two mechanisms.

| Drift source | Fix | Sites |
| --- | --- | --- |
| master edits routing | config and both router modules join the hashed key inputs, save and every restore | 9 |
| a PR edits routing | merge base and HEAD differ, so drop the migration-set key and the per-SHA fallback | 5 |
| both | `SCHEMA_CACHE_EPOCH` v1 to v2, so no v1 entry survives the new formula | 5 |

`allow_migrate` decides which tables get created, so the modules implementing it are hashed alongside the config. The per-SHA fallback is itself a merge-base-routing dump, so hashing alone would not cover the PR case.

```diff
-MIG_HASH=$(printf '%s\n%s'     "$MIG_FILES" "$PG_IMAGE"               | sha256sum | cut -c1-40)
+MIG_HASH=$(printf '%s\n%s\n%s' "$MIG_FILES" "$PG_IMAGE" "$DB_ROUTING" | sha256sum | cut -c1-40)
```

> [!WARNING]
> Expect one-time misses after merge: nothing exists under `posthog-schema-mig-v2-*` until a master push saves the first entries, and the per-SHA key carries no epoch so it absorbs most of them. Later routing changes de-cache open PRs until they rebase past them. These three files have changed in 5 commits ever.

> [!NOTE]
> `posthog/settings/data_stores.py` also shapes routing, since it registers the router and synthesizes the `TEST` writer URLs. Deliberately not hashed: its routing block has only ever moved in commits that already touch a hashed file, while the file overall changes ~14 times per 90 days. The durable fix is a self-describing dump that records its own routed app set, retiring every routing input here.

No new tests. The behavior is entirely workflow YAML.

| Check | Result |
| --- | --- |
| `hogli lint:workflows` | 7/7 green across 124 workflows |
| `hogli ci:preflight --strict` | 0 failures |
| 9 key blocks pinned to one commit | 1 distinct hash, save and restore agree |
| Per-site refs | routing, migrations and image all read that site's own ref |
| Drop or swap `DB_ROUTING` | hash moves both ways, so the input is load-bearing |
| Mutate the router module blob | hash moves, so the implementation is load-bearing too |
| Guard replayed against real commits | changed routing skips, unchanged proceeds, absent at base skips |

Replayed against every commit that has ever changed routing:

| Commit | Change | Caught |
| --- | --- | --- |
| [deflake backend test offenders](https://github.com/PostHog/posthog/pull/68725) | routing block plus `data_stores.py` | yes |
| [gate deploys on unconfigured routes](https://github.com/PostHog/posthog/pull/70338) and [its revert](https://github.com/PostHog/posthog/pull/70398) | routing block plus `data_stores.py` | yes |
| [stamphog PR review product](https://github.com/PostHog/posthog/pull/70407) | route added | yes |
| [remove the agent platform](https://github.com/PostHog/posthog/pull/73919) | route removed | yes |

This PR leaves `products/db_routing.yaml` untouched, so its own CI only exercises the unchanged-routing path.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed. No user-facing surface.

## 🤖 Agent context

Claude Code (Opus). Skills: `/authoring-ci-workflows`, `/writing-code-comments`, `/simplify`.

- First pass hashed the merge-base copy only, so a PR editing routing still matched the old key.
- Chose blanking both keys over rehashing at HEAD: `ci-backend` gates its restore on the per-SHA key, and one rule across 5 sites stays in parity more easily.
- Rejected folding routing into the `MIG_FILES` grep, since `git ls-tree` always lists it and the `[ -z "$MIG_FILES" ]` guard would go dead.
- Parity across the 9 copies is hand-maintained; a follow-up extracts a composite action.


gantoine (approved):

rnegron:
will merge tomorrow once the merge queue is a bit cleaner

PR #74025, opened 2026-07-27
fix(personhog): snapshot freeze quorum and bound handoff age

z0br0wn:
## Problem

A handoff's freeze quorum is evaluated against the live router registry, so the requirement can grow after the handoff is created. A router that registers mid-freeze becomes a required acker for a Freezing event it can never receive — its watch is anchored at its own bootstrap revision, so nothing replays it — leaving bootstrap catch-up as its only path to acking. If that misses, the quorum is unsatisfiable for as long as that router lives, and nothing removes the handoff: cleanup_stale_handoffs only deletes handoffs whose new owner is gone, and an in-flight handoff pins its partition against re-planning. The result is a handoff stuck in Freezing indefinitely, with its partition unroutable, until a human restarts a pod.

Rolling deploys make this likely rather than exotic: they introduce new routers precisely while the resulting rebalance's handoffs are in flight.

## Changes

Freeze quorum is snapshotted at handoff creation (HandoffState::freeze_quorum) and intersected with the live registry at evaluation, so the requirement is monotonic — a departed router drops out, and a late joiner is never added. Excluding a late joiner is safe because the freeze quorum is an availability mechanism, not the safety boundary: a correctly bootstrapping joiner opens stashes for in-flight handoffs before its routing table loads and never forwards to the old owner, and even one that missed the handoff cannot lose an acked write — pre-fence forwards land in the changelog ahead of the warm HWM, and post-fence forwards are rejected by the broker's transactional-producer epoch and never acked. The full argument lives on the field's doc comment. Records without a snapshot (written by older coordinators) keep the all-live-routers rule — the only safe reading of an empty snapshot.

A total-age deadline backstops causes not yet found (COORDINATOR_HANDOFF_DEADLINE_SECS, default 120s, orders of magnitude above healthy handoffs). The reconcile tick cancels an over-age handoff using the same CAS-guarded atomic handoff-plus-acks delete cleanup uses, counts it under handoffs_cancelled_total{reason="phase_deadline"} so the deadline can't quietly absorb a bug, and wakes the planning loop via a Notify so a fresh plan — new handoff id, new quorum snapshot — retries immediately. The wake matters: a deletion fires no pod event, so without it no re-plan would run until the next unrelated pod change. Waking the one planning loop rather than planning inline keeps a single planner task. Force-advancing past a missing ack would be exactly the split-brain the phase exists to prevent, so cancel-and-retry is the only safe response. Deliberately no backoff: with the known cause fixed, repeated cancellation of one partition means something new, and the counter makes that visible.

Participants already handle deletion-without-Complete (routers drain stashes back to the current assignment; pods feed deletions into desired-state convergence), so cancellation reuses existing protocol machinery rather than introducing a new event shape.

- Three new unit tests on freeze_quorum_met pin the semantics, each catching a distinct regression: a late joiner blocking the quorum (the bug), a snapshot member being dropped while still registered (would advance past a router still forwarding), and an empty snapshot being read as "nobody must ack" (would advance with no freeze at all).
- New integration test a_handoff_that_can_never_reach_quorum_is_cancelled: injects a Freezing handoff whose required router never acks, with a live new owner so the existing cleanup path stays out of it, and asserts the wedged handoff does not outlive its deadline — no existing test covered any removal path for a live-new-owner handoff.
- Full suites green: 32 lib, 52 integration, 20 protocol-hardening. The stateright model checker passes (12 properties, ~6 min); with static router membership the model exercises the fallback path, which is identical to the previous rule — machine-checking the snapshot-divergence case needs dynamic router membership in the model and is a scoped follow-up.
- Existing tests that deliberately park handoffs mid-phase get an effectively-disabled deadline so the deadline can't delete the state under test; the cancellation test passes a short one explicitly.

## 🤖 Agent context

Root-caused from a stuck handoff observed via the new handoffs_in_flight{phase} metric plus router logs showing a router that started seconds after the stuck handoffs were created and only ever processed later events. An adversarial review before commit traced the safety argument to its actual foundation (drain + broker epoch fencing + fence-then-HWM ordering, machine-verified in personhog-stateright, rather than router behavior) and caught that cancellation originally had no re-plan trigger — recovery would have waited for the next unrelated pod event; fixed with the Notify wake into the single planning loop. A per-phase deadline was considered and rejected: started_at is the record's only timestamp, so per-phase budgets would silently shrink for whichever phase ran last; one total-age budget is the honest statement of intent. Session run in Claude Code; /writing-tests invoked for the test additions.

### DIFF
--- a/rust/personhog-coordination/src/coordinator.rs
+++ b/rust/personhog-coordination/src/coordinator.rs
@@ -809,11 +809,11 @@ impl Coordinator {
     /// next plan is free to try again, with a fresh handoff id and a
     /// fresh quorum snapshot.
     ///
-    /// Deliberately not backed off: with the quorum snapshotted at
-    /// creation the known cause of unsatisfiable quorums is gone, so
-    /// repeated cancellation of the same partition means something new,
-    /// and `handoffs_cancelled_total{reason="phase_deadline"}` is meant
-    /// to make that visible rather than absorbed.
+    /// Deliberately not backed off: the snapshot rule keeps freeze
+    /// quorums satisfiable, so repeated cancellation of one partition
+    /// means a new cause, and
+    /// `handoffs_cancelled_total{reason="phase_deadline"}` exists to
+    /// make that visible rather than absorbed.
     ///
     /// Returns whether anything was cancelled, so the caller can wake
     /// the planning loop — a deletion fires no pod event, and without
--- a/rust/personhog-coordination/src/coordinator.rs
+++ b/rust/personhog-coordination/src/coordinator.rs
@@ -4,6 +4,7 @@ use std::time::Duration;
 
 use etcd_client::{EventType, WatchStream};
 use metrics::{counter, gauge, histogram};
+use tokio::sync::Notify;
 use tokio_util::sync::CancellationToken;
 
 use assignment_coordination::store::parse_watch_value;
@@ -35,6 +36,29 @@ pub struct CoordinatorConfig {
     /// indefinitely, and doubles as defense-in-depth for anything else
     /// that slips through the event-driven paths.
     pub reconcile_interval: Duration,
+    /// How long a handoff may sit in a pre-terminal phase before the
+    /// coordinator cancels it and lets the next plan try again.
+    ///
+    /// This is the backstop for causes we have not found: a participant
+    /// that never acks leaves a handoff that no other path removes —
+    /// cleanup only deletes handoffs whose new owner is gone, and an
+    /// in-flight handoff pins its partition so no re-plan can touch it.
+    /// Cancelling is the only safe response; force-advancing past a
+    /// missing freeze ack is exactly the split-brain the quorum exists
+    /// to prevent.
+    ///
+    /// Measured against the handoff's total age rather than time in its
+    /// current phase: `started_at` is the only timestamp the record
+    /// carries, and a per-phase budget derived from it would silently
+    /// shrink for whichever phase happened to run last. One end-to-end
+    /// budget is also the honest statement of intent — a handoff should
+    /// finish, and how it divides its time between freezing, draining,
+    /// and warming is not something to police.
+    ///
+    /// Generous by design: healthy handoffs complete in a few seconds,
+    /// so this sits orders of magnitude above them. Too tight a bound
+    /// would cancel the handoffs that are merely slow.
+    pub handoff_deadline: Duration,
 }
 
 impl Default for CoordinatorConfig {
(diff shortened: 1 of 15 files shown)


rust/personhog-coordination/src/protocol.rs
```diff
@@ -135,9 +139,33 @@ pub fn freeze_quorum_met(
...
-        .all(|r| acked.contains(r.router_name.as_str()))
+        .map(|r| r.router_name.as_str())
+        // An absent snapshot is a pre-upgrade record: fall back to
+        // requiring every live router, which is what it was written
+        // under.
+        .filter(move |name| {
+            handoff.freeze_quorum.is_empty()
+                || handoff.freeze_quorum.iter().any(|member| member == name)
+        })
 }
```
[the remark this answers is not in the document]
z0br0wn: fixed, worth noting after this lands and there's no previous empty state, the optional bit can be removed / simplified

rust/personhog-coordination/src/types.rs
```diff
@@ -122,6 +122,45 @@ pub struct HandoffState {
     /// past a drain or warm that never happened.
     #[serde(default)]
     pub handoff_id: String,
+    /// The routers that must ack this handoff's freeze, snapshotted when
+    /// the coordinator created it.
+    ///
+    /// Evaluating the quorum against the *live* router set instead lets
```
nickbest-ph: nit: i've noticed that fable seems to like to anchor long-winded comments on the behaviour that is changing in its edit, rather than just elaborating on/specing out what the new behaviour is.

sometimes this leaks implementation details of code/behaviours that don't exist anymore so the comments become stale almost immediately on merge

doesn't need to gate this PR, but i think this comment could be shorter, outlining the expected behaviour of this snapshot and what it guards against without fully going into the details of the "live" router set we are moving away from

rust/personhog-coordination/src/coordinator.rs
```diff
@@ -732,6 +799,91 @@ impl Coordinator {
...
+    /// change. `handoffs` is the caller's already-listed snapshot; each
+    /// candidate is re-read under mod_revision before deletion, so the
+    /// snapshot's staleness only ever skips a cancel, never misdirects
+    /// one.
+    async fn cancel_expired_handoffs(
+        store: &PersonhogStore,
+        handoffs: &[HandoffState],
+        handoff_deadline: Duration,
+    ) -> Result<bool> {
+        let now = util::now_seconds();
```
nickbest-ph, line 840: so, this action makes clock skew between machines a bit sensitive/relevant to the overall system, because the cancellation potentially uses timestamps originating from two different machines to compute whether or not something is expired.

if clock skew is substantial we could be cancelling hand offs inadvertently quickly.

maybe a comment about that just in case we get bit by this somehow?

nickbest-ph (approved):

PR #74026, opened 2026-07-27
feat(conversations): quick actions for support replies

HaynesPostHog:
## Problem

Support agents repeat the same replies and the same follow-up steps (set status, bump priority, tag) all day.
This ships the first half of a feature we started as "macros" and are calling **Quick actions**: reusable saved replies that can also apply ticket actions on insert.

This is split out of [#72792](https://github.com/PostHog/posthog/pull/72792), which a reviewer asked to break up because it was ~3k lines.
This PR is the saved-response half and is useful on its own; a stacked follow-up adds the second kind (running a workflow against the ticket).

## Changes

Adds **Quick actions** to the Support composer. An agent inserts one by typing `/` in the message box or clicking the quick action button.

- A quick action inserts a saved reply and can apply ticket actions (status, priority, tags, assignee) at the same time.
- Replies support `{{variables}}` (customer name, ticket number, agent name) filled from the ticket client-side.
- Quick actions are team-shared or personal to their creator, managed under Support, Settings, Quick actions.

Backend adds a team-scoped `QuickAction` model (fail-closed `TeamScopedRootMixin`) and a DRF viewset with visibility scoping plus an allow-listed ticket-actions serializer.
Frontend adds the `/` slash-command picker, the composer toolbar button, template-variable substitution, and the settings management UI.

Automated, all green locally. I (Claude, via PostHog Code) ran these; @HaynesPostHog directed the work.

- **Backend** `test_quick_actions.py` (7 tests): create sets team and creator, visibility scoping, the actions serializer strips unknown keys, a non-creator cannot demote a shared quick action to personal, an API-set assignee survives a partial update, content cap, and the "must do something" guard.
- **Frontend jest** (12 tests): `templateVariables` substitution, and `applyQuickAction` covering rich vs plain-text doc building, the empty-doc guard, and reply-insert vs actions-only dispatch.
- Repo-wide `mypy` clean (17010 files), `pnpm typescript:check` clean for the changed files, `ruff`, and `hogli ci:preflight --strict` with 0 failures.

I could not capture live screenshots in this session: no browser driver was available and the shared dev stack was held by another worktree. The response flow is covered by the tests above; @HaynesPostHog is doing a visual pass of the settings UI and the `/` composer picker before merge.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs exist for this composer feature yet; nothing to update in this PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). @HaynesPostHog directed the work; I (Claude, via PostHog Code) implemented it.

This is the first of a two-PR split of [#72792](https://github.com/PostHog/posthog/pull/72792). This PR is the saved-response half. A stacked follow-up ([#74031](https://github.com/PostHog/posthog/pull/74031)) adds workflow-trigger quick actions plus the workflows facade, and depends on the [CDP `manual_invocations` route](https://github.com/PostHog/posthog/pull/73058), which is already its own PR.

Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`, `/writing-user-facing-copy`, `/writing-code-comments`.

Two things worth a reviewer's eye. The `QuickAction` migration is `0055` because the original branch's `0053` collided with master's SLA-snooze migrations, so I regenerated it against the current leaf. I also carried over and fixed a review finding from #72792: `{{customer.name}}` now resolves from the ticket's embedded person first (present synchronously and not nulled when the separate person lookup fails) instead of the async person loader, matching the message-display resolution in the same logic.

---


products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,221 @@
...
+    def safely_get_queryset(self, queryset: QuerySet[QuickAction]) -> QuerySet[QuickAction]:
+        # `for_team` resolves child environments to the canonical (parent) team id, matching the
+        # rewrite `RootTeamMixin.save()` performs on write. Filtering by the raw `self.team_id`
+        # would miss quick actions created in a child environment (stored under the parent).
+        queryset = QuickAction.objects.for_team(self.team_id).select_related("created_by")
+        # Team quick actions are visible to everyone; personal ones only to their creator.
+        return queryset.filter(
+            Q(visibility=QuickActionVisibility.TEAM)
+            | Q(visibility=QuickActionVisibility.PERSONAL, created_by=self.request.user)
+        )
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — confirmed. `safely_get_queryset` scopes via `for_team()` (canonical parent) while `TeamAndOrgViewSetMixin.get_queryset()` re-ANDs the raw URL `team_id`, so any child/project environment gets `team_id = parent AND team_id = child` and matches nothing. The root-team fixtures are why the tests stay green. Fixing with `_should_skip_parents_filter -> True`, mirroring `MetricRuleViewSet`.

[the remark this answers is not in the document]
HaynesPostHog: **valid** — the override rebuilds the queryset and drops the class-level `.order_by("-created_at")`, and the model `Meta` sets no default ordering, so pagination is non-deterministic. Adding `.order_by("-created_at")` to the override, as `ticket_views.py` already does.

products/conversations/frontend/components/QuickActions/quickActionsLogic.ts
```diff
@@ -0,0 +1,340 @@
...
+            }
+        },
+        loadQuickActionsFailure: () => {
+            lemonToast.error('Failed to load quick actions')
+        },
+    })),
+
+    afterMount(({ actions }) => {
+        actions.loadQuickActions()
+    }),
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — on a cold/direct load the loader can fire before `currentTeamId` resolves, sending `/api/projects/null/...` with no auto-retry. Guarding the initial load on `currentTeamId`, consistent with the sibling logics cited.

products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,221 @@
...
+        return getattr(self.instance, field, None)
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        instance = self.instance
+
+        # A quick action must do something: insert a reply or apply ticket actions.
+        has_reply = bool(self._effective(attrs, "content") or self._effective(attrs, "rich_content"))
+        has_actions = bool(self._effective(attrs, "actions"))
+        if not (has_reply or has_actions):
+            raise serializers.ValidationError("A quick action needs a reply or a ticket action.")
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — `validate()` measures "has actions" on the raw incoming `actions` while `update()` merges the previous assignee back, so an assignee-only quick action fails to re-save from the UI once the reply is cleared. Computing the merged/effective actions once and reusing it in both places.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
```diff
@@ -914,6 +950,35 @@ export const supportTicketSceneLogic = kea<supportTicketSceneLogicType>([
...
+                } else {
+                    actions.setAssignee(null)
+                }
+                changed = true
+            }
+            // Persist the field changes; text insertion into the composer is handled separately.
+            if (changed) {
+                actions.updateTicket()
+            }
+        },
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — `applyTicketActions -> updateTicket()` PATCHes every locally-edited field, not just the ones the quick action set, bypassing the unsaved-changes guard the send-and-set-status flow already uses. Scoping the write to the quick action's own fields.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
```diff
@@ -0,0 +1,280 @@
...
+        parts.push(`status → ${quickAction.actions.status}`)
+    }
+    if (quickAction.actions?.priority) {
+        parts.push(`priority → ${quickAction.actions.priority}`)
+    }
+    if (quickAction.actions?.tags?.length) {
+        parts.push(`${quickAction.actions.tags.length} tag${quickAction.actions.tags.length === 1 ? '' : 's'}`)
+    }
+    return parts.length ? parts.join(', ') : '—'
+}
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — the summary omits `actions.assignee`, so an assignee-carrying quick action under-reports what it does. Narrow (assignee is API-only today, no Settings control creates it), so tracking as a follow-up rather than in this fix pass.

products/conversations/frontend/components/QuickActions/QuickActionPicker.tsx
```diff
@@ -0,0 +1,160 @@
...
+                {filtered.length === 0 && (
+                    <div className="text-secondary p-2 text-center">
+                        {quickActionsLoading
+                            ? 'Loading quick actions...'
+                            : loadFailed
+                              ? "Couldn't load quick actions. Try again."
+                              : query
+                                ? 'No matching quick actions'
+                                : 'No quick actions yet'}
+                    </div>
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — the copy names a retry the UI can't perform (a toast fires, but the picker branch has no button and the logic stays mounted). Minor/recoverable; follow-up to either wire a retry button or soften the copy.

products/conversations/frontend/components/QuickActions/quickActionsLogic.ts
```diff
@@ -0,0 +1,340 @@
...
+        },
+        isModalOpen: [
+            false,
+            {
+                openCreateModal: () => true,
+                openEditModal: () => true,
+                closeModal: () => false,
+                quickActionSaved: () => false,
+            },
+        ],
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — `quickActionSaved` closes the modal unconditionally, so a stale save for A can close a modal since reopened for B via escape/backdrop. Reachable but narrow (low-frequency settings surface, recoverable). Follow-up.

products/conversations/frontend/components/QuickActions/quickActionsLogic.ts
```diff
@@ -0,0 +1,340 @@
...
+                        const response = await conversationsQuickActionsList(teamId, { limit: pageSize, offset })
+                        all.push(...response.results)
+                        if (!response.next || response.results.length === 0) {
+                            break
+                        }
+                        offset += pageSize
+                    }
+                    return all
+                },
+            },
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — no `breakpoint()` between paged awaits, so a superseded multi-page fetch could clobber newer state. As the analysis notes, the logic is a singleton loaded only from `afterMount`, so the realistic trigger is narrow and self-correcting. Follow-up.

products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,233 @@
...
+        # `safely_get_queryset` already scopes by the canonical team via `for_team`. The default
+        # parent-lookup filter would AND the raw URL team id back on top, which for a child
+        # environment can never match a quick action stored under the parent team.
+        return True
+
+    def safely_get_queryset(self, queryset: QuerySet[QuickAction]) -> QuerySet[QuickAction]:
+        # `for_team` resolves child environments to the canonical (parent) team id, matching the
+        # rewrite `RootTeamMixin.save()` performs on write. Filtering by the raw `self.team_id`
+        # would miss quick actions created in a child environment (stored under the parent).
+        queryset = QuickAction.objects.for_team(self.team_id).select_related("created_by")
```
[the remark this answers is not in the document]
HaynesPostHog: Thanks — the mechanics are accurate (data is canonicalized to the parent via `for_team`, authorization is checked against the URL team), but this is the intended project-scoping rather than a bug, so I'm not changing it here. Reasoning:

- `QuickAction` uses `TeamScopedRootMixin` deliberately so quick actions are **project-shared** across every environment. The earlier must-fix (`_should_skip_parents_filter`) existed precisely to make that sharing work — a quick action created in one environment must be usable from its siblings.
- This isn't cross-tenant. `TeamAndOrgViewSetMixin` still verifies the requester's membership/scope against the URL team within their org before `safely_get_queryset` runs, so you can't reach another project's or org's quick actions — the only boundary crossed is between environments of the *same* project, which is the design.
- The `for_team(self.team_id)` + parent-canonicalization pattern is shared by ~10 other project-scoped viewsets (`metric_rules_api`, `parser_recipes`, `tracing`, `data_catalog`, `message_suppression`, …). Adding a stricter parent-only permission here would diverge from that convention and could actually block a legitimate environment member from the shared config.

If we do want quick actions to respect per-environment isolation, that's a platform-level access-control decision (affecting all those models), not something to change in this PR — flagging for @mayteio / @veryayskiy to weigh in.

mayteio (commented):
Went through the whole diff. The security shape looks right: fail-closed manager, visibility scoping with a test that would actually catch a regression, the actions serializer stripping unknown keys, and size caps on the JSON fields. Nothing that looked like an IDOR or an injection path.

Comments inline. The two I would want resolved before merge are the slash trigger swallowing arrow keys and unknown template tokens being deleted silently, since both land in the customer-facing composer. The rest are questions or small cleanups.

One process note: for a change this size, a short video or a few screenshots of the settings UI and the `/` picker would make review much easier. Most of what I flagged is behavior you can only really judge by driving it, and there is currently nothing in the PR to check the visual and interaction side against.

products/conversations/frontend/scenes/ticket/SupportTicketScene.tsx
```diff
@@ -235,6 +237,9 @@ export function SupportTicketScene({ ticketId }: { ticketId: string }): JSX.Elem
                         feedbackByMessageId={feedbackByMessageId}
                         showAiReplyFeedback={aiSuggestionsEnabled}
                         onSubmitAiReplyFeedback={submitAiReplyFeedback}
+                        enableQuickActions
```
mayteio, line 258: Suggestion rather than a request: should the composer half go out behind a flag?

The settings section is inert on its own, so it is just this prop. This adds a keystroke intercept to the live customer reply composer, and without a flag the only rollback is reverting the PR.

products/conversations/frontend/components/QuickActions/QuickActionPicker.tsx
```diff
@@ -0,0 +1,160 @@
...
+            }
+        },
+        [onSelect]
+    )
+
+    const onKeyDown = useCallback(
+        (event: KeyboardEvent): boolean => {
+            const count = filtered.length
+            const keyMappings = {
+                ArrowUp: (): boolean => {
```
mayteio, line 68: These return `true` (handled) even when `filtered.length === 0`, so TipTap preventDefaults the keypress and the cursor does not move.

With `startOfLine: true` and `allowSpaces: true`, the suggestion stays active for the whole line once a line starts with `/`. So an agent who begins a line with a path (`/api/projects/@current/insights/`, `/usr/local/bin/...`) gets a "No matching quick actions" popover over their compose box and loses up/down arrow navigation for the rest of that reply, with nothing on screen explaining why.

Did this come up when you were testing the slash trigger? Curious whether it reproduces as readily as it looks like it should from the code.

```ts
ArrowUp: (): boolean => {
    if (count === 0) {
        return false
    }
    setSelectedIndex((i) => (i - 1 + count) % count)
    return true
},
```

Separately, worth verifying: `Escape` calls `renderer.destroy()`, but the suggestion stays active in ProseMirror state, so the next keystroke hits `renderer.updateProps()` on a destroyed renderer.

products/conversations/frontend/components/Editor/templateVariables.ts
```diff
@@ -0,0 +1,60 @@
...
+const VARIABLE_PATTERN = /\{\{\s*([\w.]+)\s*\}\}/g
+
+/**
+ * Replace {{variable}} tokens in a string with their resolved values.
+ * Unknown or unset tokens resolve to an empty string so no raw {{...}} leaks to the customer.
+ */
+export function applyTemplateVariables(text: string, values: TemplateVariableValues): string {
+    return text.replace(VARIABLE_PATTERN, (_match, token: string) => {
+        const value = values[token as TemplateVariableToken]
+        return value ?? ''
```
mayteio: `value ?? ''` erases any `{{...}}` the variable list does not know about. A saved reply explaining a Liquid or Jinja template loses `{{ user.name }}` on insert, and the agent has no signal it happened before sending to the customer.

Given how often support replies are about template syntax, leaving unknown tokens untouched seems safer:

```ts
return value ?? _match
```

posthog/settings/web.py
```diff
@@ -614,6 +614,12 @@ def static_varies_origin(headers, path, url):
         "UserInterviewSearchDocumentTypeEnum": "products.user_interviews.backend.facade.enums.SEARCH_DOCUMENT_TYPES",
         "BatchExportRunStatusEnum": "products.batch_exports.backend.models.batch_export.BatchExportRun.Status",
         "HeatmapType": "products.web_analytics.backend.models.heatmap_saved.SavedHeatmap.Type",
+        # Conversations ticket + quick-action priority share one choice set; pin its name so the
+        # quick-action `priority` field doesn't collide with other products' `priority` enums.
+        "SupportPriorityEnum": "products.conversations.backend.models.constants.Priority",
+        # Keep the existing private/shared visibility enum name stable so the conversations
+        # quick-action `visibility` field (team/personal) doesn't rename it and churn unrelated types.
+        "VisibilityEnum": "posthog.models.column_configuration.ColumnConfiguration.Visibility",
```
mayteio: The comment says this keeps the existing name stable, but the effect is the reverse: `ColumnConfigurationVisibilityEnumApi` becomes `VisibilityEnumApi` in `products/product_analytics/frontend/generated/api.schemas.ts`. Nothing hand-written imports it so nothing breaks, but it hands the generic `VisibilityEnum` name to column configurations, so the next product that adds a `visibility` field collides with it instead.

Pinning to `ColumnConfigurationVisibilityEnum` avoids the collision and drops the cross-product churn from this diff.

products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,233 @@
...
+        if not (has_reply or has_actions):
+            raise serializers.ValidationError("A quick action needs a reply or a ticket action.")
+
+        # Only the creator may turn a shared team quick action personal — otherwise a teammate's
+        # edit would make it vanish for everyone else (and the editor), with no way to reach it again.
+        if (
+            instance is not None
+            and attrs.get("visibility") == QuickActionVisibility.PERSONAL
+            and instance.visibility == QuickActionVisibility.TEAM
+            and instance.created_by_id != self.context["request"].user.id
```
mayteio, line 163: The comment justifies this as stopping a shared quick action from vanishing for everyone, but a non-creator can already delete it outright, or select all and clear the body. The guard blocks one of three routes to the same outcome.

The one asymmetry is that deleting is visible whereas demoting makes it silently disappear from every picker while the row still exists, with no way for the creator to reach it. If that is the concern it is worth saying so. Otherwise dropping the guard and its test removes the branch, the `_merged_actions` interaction, and the inconsistency.

products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,233 @@
...
+        required=False,
+        allow_null=True,
+        help_text="Set the ticket priority when the quick action is applied.",
+    )
+    tags = serializers.ListField(
+        child=serializers.CharField(),
+        required=False,
+        help_text="Replace the ticket's tags with this list when the quick action is applied.",
+    )
+    assignee = QuickActionAssigneeSerializer(
```
mayteio, line 52: Nothing in this PR writes `assignee`. Carrying it means `_merged_actions()`, the ordering constraint it puts on `validate()`, and two regression tests all exist to protect a field with no producer.

Is it needed before the assignee control lands? Dropping it now and adding it back with the UI would take out roughly 70 lines, and `actions` stays a straight full replace.

products/conversations/backend/api/quick_actions.py
```diff
@@ -0,0 +1,233 @@
...
+    serializer_class = QuickActionSerializer
+    lookup_field = "short_id"
+
+    def _should_skip_parents_filter(self) -> bool:
+        # `safely_get_queryset` already scopes by the canonical team via `for_team`. The default
+        # parent-lookup filter would AND the raw URL team id back on top, which for a child
+        # environment can never match a quick action stored under the parent team.
+        return True
+
+    def safely_get_queryset(self, queryset: QuerySet[QuickAction]) -> QuerySet[QuickAction]:
```
mayteio, line 201: This is the only conversations viewset using `_should_skip_parents_filter` plus `for_team`. I checked and it is safe, `get_queryset` runs `safely_get_queryset` before the parents filter, and the child-environment test covers it.

Question rather than a change request: does `TicketViewSet` have the same child-environment gap and nobody has hit it, or is there something about quick actions specifically that needs the parent resolution? Asking because if it is the former, tickets are silently broken for multi-environment projects.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
```diff
@@ -914,6 +954,36 @@ export const supportTicketSceneLogic = kea<supportTicketSceneLogicType>([
...
+                fields.push('priority')
+            }
+            if (ticketActions.tags) {
+                actions.setTags(ticketActions.tags)
+                fields.push('tags')
+            }
+            if (ticketActions.assignee !== undefined) {
+                const assignee = ticketActions.assignee
+                if (assignee && assignee.id != null) {
+                    const type = assignee.type === 'role' ? 'role' : 'user'
```
mayteio, line 1037: `assignee.id` is a `CharField` server-side, so a non-numeric value yields `NaN` here, serializes to `null`, and silently clears the ticket's assignee instead of failing. Worth guarding:

```ts
const id = type === 'user' ? Number(assignee.id) : assignee.id
if (type === 'user' && Number.isNaN(id)) {
    return
}
```

products/conversations/frontend/scenes/ticket/SupportTicketScene.tsx
HaynesPostHog: Fair call, and it's yours to make since it's a rollout decision. Two things that lower the risk in the meantime: the intercept only arms on a `/` at the *start* of a line (not mid-text), and I fixed the arrow-key swallow below so a non-matching `/…` line no longer traps the caret. If you'd still like it behind a flag before it reaches the live composer, say the word and I'll gate the `SupportEditor` prop on one.

products/conversations/frontend/components/QuickActions/QuickActionPicker.tsx
HaynesPostHog: $Fixed both, thanks — this was a real bug.\n\n- ArrowUp/ArrowDown now return \`false\` when \`filtered.length === 0\`, so a line that starts with \`/\` but has no matches (a path, etc.) lets the keypress through and the caret moves normally.\n- The renderer is now nullable with an idempotent \`close()\`; Escape/exit/click-away all null it out, so a later \`onUpdate\`/\`onKeyDown\` from ProseMirror no-ops instead of hitting a destroyed renderer.\n\nAnd yes, it reproduced exactly as you read it — a \`/usr/local/...\` line trapped arrow nav.

products/conversations/frontend/components/Editor/templateVariables.ts
HaynesPostHog: $Fixed. I kept the blank-on-unset behavior for *known* tokens (so a literal \`{{customer.name}}\` never ships to a customer when we just have no value), but unknown tokens are now left untouched — a reply explaining Liquid/Jinja keeps its \`{{ user.name }}\` example. Regression test updated to cover both.

posthog/settings/web.py
HaynesPostHog: $Fixed. Pinned \`ColumnConfigurationVisibilityEnum\` to its existing name instead, so QuickAction takes \`QuickActionVisibilityEnum\` and product_analytics reverts to \`ColumnConfigurationVisibilityEnumApi\` — the cross-product churn is gone from the diff, and the generic \`VisibilityEnum\` name is no longer handed out to collide later. Good catch that I had it backwards.

products/conversations/backend/api/quick_actions.py
HaynesPostHog: $You read it right — the concern is exactly the asymmetry: delete is visible, but flipping team→personal silently removes the row from everyone else's picker while it still exists and is unreachable. Kept the guard and rewrote the comment to say that explicitly rather than implying it's protecting the content.

products/conversations/backend/api/quick_actions.py
HaynesPostHog: $It's not dead — \`assignee\` is a live, writable field today via the API/MCP (the serializer accepts it and \`applyQuickAction\` applies it in the composer), just with no Settings control yet. The merge-back exists so a UI partial-update (which never sends \`assignee\`) doesn't wipe one set via the API. So dropping it would drop a supported capability, not just speculative UI plumbing. Happy to revisit if we decide assignee shouldn't be API-settable before the control lands.

products/conversations/backend/api/quick_actions.py
HaynesPostHog: $Checked — no latent gap. \`TicketViewSet\` doesn't use \`for_team\`; \`Ticket\` isn't a \`TeamScopedRootMixin\` model, so tickets are genuinely environment-scoped and the raw \`team_id\` filter is correct for them. Quick actions are the opposite: \`TeamScopedRootMixin\` canonicalizes them to the parent so they're project-shared, which is *why* this viewset needs the parent-resolution + skip-parents-filter. Different scoping by design, not tickets being silently broken.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
HaynesPostHog: $Fixed. A non-numeric user id now skips the assignee action (with a toast) instead of serializing \`NaN\`→\`null\` and silently clearing the assignee. I skip just that action rather than early-returning, so the rest of the quick action still applies.

abigailbramble (commented):
few questions:
- do we want to do anything more complete about assignees only being able to be set via API?
- I'm assuming we're planning to roll this roll out as part of a feature flag please so we can test it / play with it before rolling it out to customers? make sure it's gated in the settings /definitions of quick actions and in the composer.
- do we want the quick actions available in the outbound ticket composer?

PR #74031, opened 2026-07-27
feat(conversations): run a workflow from a quick action

HaynesPostHog:
## Problem

Support agents often want to kick off an automation (escalate, notify, enrich data) while handling a ticket, not just send a reply.
This adds the second kind of **Quick action**: one that runs a workflow against the current ticket.

Stacked on [#74026](https://github.com/PostHog/posthog/pull/74026) (the saved-response half). Both are a split of [#72792](https://github.com/PostHog/posthog/pull/72792), which a reviewer asked to break up.

> [!IMPORTANT]
> Review [#74026](https://github.com/PostHog/posthog/pull/74026) first. This PR is stacked on it, so its diff here is the workflow-trigger delta only.

## Changes

A quick action can now carry an optional workflow that runs against the ticket when the action is used, alongside its reply and ticket actions.

- `workflow_id` on the `QuickAction` model, added by migration `0056`.
- A `run` endpoint that RBAC-checks the runner's access to the workflow, then enqueues a full-graph run against a synthesized ticket event.
- `products/workflows/backend/facade` as the single entry point conversations calls: `invoke_hog_flow_now`, `workflow_is_runnable`, `user_can_run_workflow`.
- A `create_hog_flow_manual_invocation` plugin-server helper.
- Composer and settings wiring: a workflow picker in settings, a `Workflow` tag in the picker, and running the workflow only after any ticket-field PATCH settles so it sees the updated fields.
- Re-expose the manual workflow trigger type, so you can build a workflow that only runs when triggered by hand.

> [!NOTE]
> The nodejs `manual_invocations` route this calls lands in [#73058](https://github.com/PostHog/posthog/pull/73058). Until that deploys, the run endpoint returns a graceful 502 instead of erroring. Flagging the facade and run path for CDP and workflows review.

Automated, all green locally. I (Claude, via PostHog Code) ran these; @HaynesPostHog directed the work.

- **Backend** `test_quick_actions.py` adds 5 workflow tests: a workflow must be runnable to attach, attaching requires workflow access, reply and workflow combine on one action, the run endpoint invokes the workflow with ticket-scoped globals and rejects non-workflow actions, and the runner's access (not the creator's) is enforced on run.
- **Frontend jest** `applyQuickAction` covers workflow-only, reply-only, and both-at-once dispatch.
- Repo-wide `mypy` clean, `pnpm typescript:check` clean for the changed files, `ruff`, and `hogli ci:preflight --strict` with 0 failures.

The live run path depends on #73058, so I did not exercise an end-to-end workflow run locally. It is covered by the backend run-endpoint test (the facade is called with the right globals) and the nodejs route test in #73058.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs exist for this composer feature yet; nothing to update in this PR.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). @HaynesPostHog directed the work; I (Claude, via PostHog Code) implemented it.

Second of a two-PR split of [#72792](https://github.com/PostHog/posthog/pull/72792), stacked on [#74026](https://github.com/PostHog/posthog/pull/74026). This PR declares the `products.conversations` to `products.workflows` dependency in `tach.toml` because the run path calls the workflows facade.

Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`, `/writing-code-comments`.

One decision worth a reviewer's eye: quick actions can run any active workflow, not just manual-trigger ones, which is why the run path needs the dedicated CDP route in #73058 rather than reusing the public webhook path. `workflow_id` is a soft reference with no cross-product FK; the API validates that it resolves to an active workflow for the team.

---


products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
```diff
@@ -975,10 +981,45 @@ export const supportTicketSceneLogic = kea<supportTicketSceneLogicType>([
...
             }
             // Persist the field changes; text insertion into the composer is handled separately.
+            // Publish a settle promise so a workflow run from the same quick action waits for the
+            // PATCH to commit before the backend reads the ticket (otherwise it sees stale fields).
             if (changed) {
+                cache.ticketUpdateSettled = new Promise<void>((resolve) => {
+                    cache.resolveTicketUpdate = resolve
+                })
                 actions.updateTicket()
             }
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — confirmed. `cache.resolveTicketUpdate`/`ticketUpdateSettled` are shared slots with no per-invocation identity guard (unlike `ticketUpdateRequest === request` one line up), so chaining two quick actions lets an earlier PATCH's `finally` resolve a later action's settle-promise and fire the workflow against stale ticket state. Capturing the resolver synchronously at the start of the `updateTicket` listener, as suggested.

products/conversations/backend/api/quick_actions.py
```diff
@@ -219,3 +299,44 @@ def perform_update(self, serializer: serializers.BaseSerializer) -> None:
...
+        except HogFlowNotRunnableError as e:
+            raise serializers.ValidationError({"workflow_id": str(e)})
+        except RequestException:
+            # The workflow service (CDP) was unreachable — a transient upstream failure, not a bad
+            # request. Surface a clean 502 rather than an unhandled 500.
+            logger.exception("quick_action_run_workflow_service_unreachable", workflow_id=str(quick_action.workflow_id))
+            return Response(
+                {"detail": "Couldn't reach the workflow service. Try again shortly."},
+                status=status.HTTP_502_BAD_GATEWAY,
+            )
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — good catch, and it defeats the interim behavior documented in the PR body. `internal_requests` doesn't `raise_for_status()`, so a live CDP without the route returns 404 (no `RequestException`), the facade turns any `not response.ok` into `HogFlowNotRunnableError`, and `run()` maps that to a 400 `{"workflow_id": ...}` instead of the intended 502. Splitting the facade so HTTP-layer failures (4xx/5xx) raise a distinct service error routed to the 502 branch, keeping `HogFlowNotRunnableError` for the `workflow_is_runnable` pre-check only.

products/conversations/backend/api/quick_actions.py
```diff
@@ -219,3 +299,44 @@ def perform_update(self, serializer: serializers.BaseSerializer) -> None:
     def perform_destroy(self, instance: QuickAction) -> None:
         self._track("conversations quick action deleted", instance)
         super().perform_destroy(instance)
+
+    @extend_schema(
+        request=QuickActionRunRequestSerializer,
+        responses={202: OpenApiResponse(description="Workflow run enqueued.")},
+    )
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — the hand-built 502 `Response` isn't discoverable by drf-spectacular, so it's missing from the generated OpenAPI/MCP/typed-client contract. Declaring it in `responses` with a small error serializer, following the `zendesk_import.py` precedent. Fixed alongside the 404->502 handling below.

products/conversations/backend/api/quick_actions.py
```diff
@@ -129,11 +159,25 @@ def _effective(self, attrs: dict[str, Any], field: str) -> Any:
...
         has_actions = bool(self._effective(attrs, "actions"))
-        if not (has_reply or has_actions):
-            raise serializers.ValidationError("A quick action needs a reply or a ticket action.")
+        workflow_id = self._effective(attrs, "workflow_id")
+        if not (has_reply or has_actions or workflow_id):
+            raise serializers.ValidationError("A quick action needs a reply, a ticket action, or a workflow to run.")
+
+        # If it runs a workflow, that workflow must be active for the team.
+        if workflow_id and not workflow_is_runnable(self.context["team_id"], workflow_id):
+            raise serializers.ValidationError({"workflow_id": "That workflow does not exist or is not active."})
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — the active-workflow check runs on the `_effective` value, so once a workflow is attached and later archived, unrelated edits (rename, visibility, ticket actions) get rejected. Scoping it to `attrs.get("workflow_id")` so it only re-validates when the field is actually being set/changed, matching the RBAC check just below it.

products/conversations/backend/api/quick_actions.py
```diff
@@ -129,11 +159,25 @@ def _effective(self, attrs: dict[str, Any], field: str) -> Any:
...
+
+        # RBAC: attaching a workflow requires access to that workflow — otherwise a ticket-scoped
+        # user could wire up (and later run) privileged automations by UUID. Only checked when the
+        # reference is being set/changed, so unrelated edits by teammates without workflow access
+        # don't get blocked. The run endpoint re-checks the actual runner at execution time.
+        if attrs.get("workflow_id") and not user_can_run_workflow(
+            self.context["request"].user, self.context["get_team"](), attrs["workflow_id"]
+        ):
+            raise serializers.ValidationError({"workflow_id": "You don't have access to that workflow."})
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — real correctness bug and the multi-environment twin of the parent-team queryset issue on #74026: QuickAction is project-scoped (canonicalized to the parent) but the workflow lookups use the raw environment `team_id`, so a workflow-backed quick action created in env A fails to run from sibling env B. It fails closed (globally-unique UUID, no cross-tenant leak), so should_fix is the right level. Canonicalizing the team_id used for validate-time and run-time workflow lookups so the reference resolves the same way from any environment in the project. Thanks for the nuance that the surfaced error is actually the 403 path.

products/workflows/frontend/Workflows/hogflows/steps/StepTrigger.tsx
```diff
@@ -226,16 +226,12 @@ export function StepTriggerConfiguration({ node }: { node: Node<TriggerAction> }
...
-                          label: 'Manual',
-                          description: 'Trigger your workflow manually... with a button!',
-                          value: 'manual',
-                          icon: <IconButton />,
-                      },
-                  ]
-                : []),
+            {
+                label: 'Manual',
+                description: 'Run only when triggered by hand — from the button here or a support quick action',
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — these are new strings this PR added, and em dashes are called out in CLAUDE.md's copy rules. Rewriting both without them.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
```diff
@@ -103,8 +125,8 @@ export function QuickActionsSection(): JSX.Element {
         <div className="flex flex-col gap-3">
             <p>
                 Save things you do often, then trigger them in a conversation by typing <code>/</code> in the message
-                box or using the quick action button. A quick action can insert a saved reply and set the ticket's
-                status, priority, or tags.
+                box or using the quick action button. A quick action can insert a saved reply, set the ticket's status,
+                priority, or tags, and run one of your workflows — any combination.
```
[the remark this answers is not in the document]
HaynesPostHog: **valid** — fixing the new lines (129 intro, 300 tooltip). Fair point that the file already has pre-existing `Optional —` tooltips; I'll clean those up too in this pass so the file stays internally consistent rather than leaving a mix.

products/conversations/backend/api/quick_actions.py
```diff
@@ -183,7 +268,9 @@ class QuickActionViewSet(
     # Reuse the "ticket" scope: quick actions are a support-agent tool and shouldn't grant access
     # beyond what ticket access already implies.
     scope_object = "ticket"
-    scope_object_write_actions = ["create", "update", "partial_update", "patch", "destroy"]
+    # Custom @action methods fall through to no required scope unless listed here, which 403s every
+    # personal-API-key / MCP caller. Keep the DRF defaults and add `run`.
+    scope_object_write_actions = ["create", "update", "partial_update", "patch", "destroy", "run"]
```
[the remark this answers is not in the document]
HaynesPostHog: **Valid — fixed** in 85467bf. You're right that deriving `run`'s authorization from `scope_object` alone let a `ticket:write` personal API key trigger workflow execution (and its stored-secret actions) without `hog_flow:write`.

Added explicit method scopes to the action:

```python
@action(detail=True, methods=["post"], required_scopes=["ticket:write", "hog_flow:write"])
def run(self, ...):
```

`APIScopePermission` requires *every* listed scope (it loops and 403s on the first missing one), so a key now needs both `ticket:write` and `hog_flow:write` — on top of the existing per-workflow `user_can_run_workflow` RBAC check in the body. `run` stays in `scope_object_write_actions` so session/access-control still treats it as a ticket-write action. Added a regression test asserting the declared scopes so this can't silently regress.

mayteio (commented):
Reviewed on top of #74026. The parts I want to call out as solid first: RBAC is checked both at attach time and again at run time against the runner rather than the creator, the `run` action requires `hog_flow:write` on top of `ticket:write`, and the migration is a nullable UUID add so it takes no lock.

Two things I would want resolved before this merges. The first is `INTERNAL_API_SECRET` (comment on `plugin_server_api.py`), which is a documented repo rule rather than a preference. The second is the manual trigger being re-exposed, which is a workflows product decision being made here as a side effect. The rest are questions and cleanups.

Same note as on the base PR: a short video or a few screenshots would help a lot, particularly the settings workflow picker and what an agent sees between clicking a workflow quick action and it completing.

posthog/plugins/plugin_server_api.py
```diff
@@ -111,6 +111,20 @@ def create_hog_flow_scheduled_invocation(
     )
 
 
+def create_hog_flow_manual_invocation(team_id: int, hog_flow_id: str, payload: dict) -> requests.Response:
```
mayteio, line 134: This adds a new `INTERNAL_API_SECRET` caller (via `get_internal_api_headers()`), and #73058 adds a new endpoint protected by the same secret. `.agents/security.md` is explicit about this:

> Don't extend `INTERNAL_API_SECRET` to new service-to-service calls. It's a single fleet-wide shared secret already trusted by many services across Django, Node, and Rust, one leak reaches all of them, so it must not grow new callers or protected endpoints (it's being actively retired edge-by-edge; adding one moves it backwards).

The documented alternatives, in order: a scoped JWT with claims pinning the token to team and operation (canonical example is `RECORDING_API_JWT_SECRET`, minted in `posthog/session_recordings/recordings/recording_api_jwt.py` and verified in `nodejs/src/session-replay/recording-api/auth.ts`), or failing that a dedicated `<PURPOSE>_API_SECRET` scoped to this one caller and callee.

I realise `create_hog_flow_scheduled_invocation` directly above does the same thing, so this looked like the local pattern. But that is the growth the rule exists to stop, and nothing in CI catches it. Worth settling with whoever owns the retirement before this lands, since it also fixes the shape of the new route in #73058.

products/workflows/frontend/Workflows/hogflows/steps/StepTrigger.tsx
```diff
@@ -226,16 +226,12 @@ export function StepTriggerConfiguration({ node }: { node: Node<TriggerAction> }
...
-                      {
-                          label: 'Manual',
-                          description: 'Trigger your workflow manually... with a button!',
-                          value: 'manual',
-                          icon: <IconButton />,
-                      },
-                  ]
-                : []),
+            {
+                label: 'Manual',
```
mayteio: This makes Manual available as a trigger type on every workflow. It was deliberately hidden in `60ccc795624` ("chore(workflows): batch workflow tweaks pre soft launch", #50072), which reduced the trigger surface ahead of soft launch at the same time as adding the batch blast-radius limit.

Worth noting the machinery was never broken (`_FUNCTION_TRIGGER_CONFIG_TYPES` in `hog_flow.py` includes `manual`, and the trigger button works), so this reads as a launch-surface decision rather than a correctness one. It also was not written down, unlike the schedule trigger immediately below which carries a comment explaining why it is hidden from new workflows.

Could we get an explicit call from the workflows side before this lands, rather than un-hiding it as a side effect of a conversations feature? Cc @HaynesPostHog for routing.

Separately, and independent of that decision: the new copy below points at "a support conversation... (Support, settings, Quick actions)". That renders in the workflow editor for teams that do not have conversations enabled at all. Either gate it on `conversations_enabled` or keep the copy generic.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
```diff
@@ -9,6 +9,8 @@ import { LemonDialog } from 'lib/lemon-ui/LemonDialog'
 import { LemonField } from 'lib/lemon-ui/LemonField'
 import { userLogic } from 'scenes/userLogic'
 
+import { workflowsLogic } from 'products/workflows/frontend/Workflows/workflowsLogic'
```
mayteio, line 12: This reaches directly into workflows' internal kea logic, which sidesteps the facade the PR introduces. `products/workflows/backend/facade` describes itself as "the single, stable entry point callers should use... rather than reaching into the model", and the `tach.toml` entry declares the dependency, but tach only covers Python so nothing enforces the equivalent on the frontend.

Net effect is that a refactor of `workflowsLogic`'s exposed values silently breaks the support settings page. If the facade is the contract, it is worth having a frontend equivalent (even just a thin exported selector/hook from workflows) rather than importing the logic module itself.

products/workflows/backend/facade/api.py
```diff
@@ -0,0 +1,70 @@
...
+    the same access as operating the workflow — not just access to the surface triggering it.
+    With no explicit access controls on the workflow, every team member passes (editor default).
+    """
+    hog_flow = HogFlow.objects.filter(team_id=team.id, id=workflow_id).first()
+    if hog_flow is None:
+        return False
+    return UserAccessControl(user=user, team=team).check_access_level_for_object(hog_flow, "editor")
+
+
+def invoke_hog_flow_now(team_id: int, workflow_id: str | UUID, globals: dict[str, Any]) -> None:
```
mayteio, line 68: The docstring says `HogFlowServiceError` is raised "if the plugin-server rejects the run or is unreachable at the HTTP layer", but only the first half is true. `response.ok` catches a completed non-2xx round trip; a `Timeout` or `ConnectionError` propagates out of here as a raw `requests` exception.

The tell is in the caller: `quick_actions.py` has to `from requests.exceptions import RequestException` and catch it alongside `HogFlowServiceError`. A facade whose callers need to know it is HTTP underneath is not buying much. Wrapping the call here and re-raising as `HogFlowServiceError` makes the docstring true and lets callers catch one thing.

products/conversations/backend/api/quick_actions.py
```diff
@@ -171,6 +219,43 @@ def create(self, validated_data: dict[str, Any]) -> QuickAction:
...
+    except Exception:
+        logger.exception("quick_action_run_group_resolution_failed", ticket_id=str(ticket.id))
+    return {
+        "event": {
+            "event": "$conversation_quick_action_triggered",
+            "properties": properties,
+            "distinct_id": ticket.distinct_id or ticket.channel_source or "unknown",
+            "timestamp": timezone.now().isoformat(),
+        },
+        "groups": properties.get("$groups"),
```
mayteio, line 295: The globals here are `{event, groups}` with no `person`, while the facade docstring advertises `{event, person?, groups?}`.

Is that deliberate? Asking because any workflow that filters on person properties will silently never match when triggered from a quick action, which is not obvious from either end and would be a confusing thing to debug.

products/conversations/backend/api/quick_actions.py
```diff
@@ -171,6 +219,43 @@ def create(self, validated_data: dict[str, Any]) -> QuickAction:
...
 
+class QuickActionRunRequestSerializer(serializers.Serializer):
+    ticket_id = serializers.UUIDField(help_text="Ticket to run the workflow against.")
+
+
+class QuickActionRunErrorSerializer(serializers.Serializer):
+    detail = serializers.CharField(help_text="Human-readable explanation of why the run failed.")
+
+
+def _build_ticket_event_globals(ticket: Ticket) -> dict:
```
mayteio: This is a hand-copy of the composition in `events.py`. `capture_message_received` (events.py:552-572) runs the same helpers in the same order and has the same org-groups fast path (`_groups_from_org_id` when `ticket.organization_id` is set, `_resolve_org_groups` otherwise), near enough verbatim.

The comment says this mirrors the `$conversation_*` events so workflow filters see the ticket, but it has already drifted in two ways before merging:

- No `_get_actor_properties`. Every real ticket event records who performed it. A quick action run has an obvious actor, the agent who clicked it, and a workflow cannot branch on it.
- No `_get_sla_properties`, which `capture_message_sent` includes. A workflow filtering on SLA state will not see them.

Rather than importing five private helpers, could `events.py` export one public builder that both this and `capture_message_received` call? That removes the private coupling, makes the two payloads impossible to drift apart, and is probably a smaller diff than what is here.

products/conversations/frontend/components/QuickActions/QuickActionPicker.tsx
```diff
@@ -123,6 +123,7 @@ export const QuickActionPicker = forwardRef<QuickActionPickerRef, QuickActionPic
                         active={index === selectedIndex}
                         onClick={() => execute(quickAction)}
                         tooltip={quickAction.description || undefined}
+                        sideIcon={quickAction.workflow_id ? <LemonTag type="completion">Workflow</LemonTag> : undefined}
```
mayteio: Related to the parent-team resolution on the base PR: quick actions resolve to the canonical parent team, so one list is shared by every environment in a project, but workflows are environment-scoped and `run` checks `workflow_is_runnable(self.team_id, ...)` against the requesting environment.

So a quick action created in one environment shows up in another environment's picker with this Workflow tag on it, and only fails once an agent clicks it. The run endpoint returns a clear message, which is good, but the picker is still advertising something that environment cannot do.

Could the list endpoint resolve `workflow_id` against the current environment and return a `workflow_runnable` flag, so this entry can be rendered with a `disabledReason` instead of failing on click?

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
```diff
@@ -980,10 +986,45 @@ export const supportTicketSceneLogic = kea<supportTicketSceneLogicType>([
...
+            // Publish a settle promise so a workflow run from the same quick action waits for the
+            // PATCH to commit before the backend reads the ticket (otherwise it sees stale fields).
             if (fields.length > 0) {
+                cache.ticketUpdateSettled = new Promise<void>((resolve) => {
+                    cache.resolveTicketUpdate = resolve
+                })
                 actions.updateTicket(fields)
             }
         },
+        runWorkflowQuickAction: async ({ quickAction }) => {
```
mayteio, line 1069: Running a workflow is not idempotent and there is no in-flight state surfaced anywhere. The agent clicks, and nothing changes on screen until the round trip returns, which can be the full 5s CDP timeout. `cache.runningWorkflowQuickAction` correctly blocks a second click, but silently: no toast, no disabled state, so a second click just disappears.

CLAUDE.md asks that anything triggering a network request disables its trigger and shows a loading state for the duration. Putting the in-flight flag in a reducer rather than `cache` would let the picker entry show it and would also let the second click explain itself.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
```diff
@@ -980,10 +986,45 @@ export const supportTicketSceneLogic = kea<supportTicketSceneLogicType>([
             }
             // Persist only the fields this quick action set, so any unrelated unsaved ticket edits
             // aren't saved as a side effect. Text insertion into the composer is handled separately.
+            // Publish a settle promise so a workflow run from the same quick action waits for the
+            // PATCH to commit before the backend reads the ticket (otherwise it sees stale fields).
             if (fields.length > 0) {
+                cache.ticketUpdateSettled = new Promise<void>((resolve) => {
```
mayteio, line 1063: This is a fair amount of hand-rolled coordination for one click path: a settle promise in `cache`, its resolver in a second `cache` slot, and an ownership transfer at the top of `updateTicket` to stop an earlier update releasing a later run. `cache.ticketUpdateSettled` is also never cleared, so it stays latched at whatever the last update left behind.

Since both halves are triggered by the same quick action, would it be simpler for one listener to own the sequence, awaiting the PATCH and then issuing the run? That deletes the promise handshake and the ownership guard entirely.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
```diff
@@ -60,11 +66,27 @@ export function QuickActionsSection(): JSX.Element {
...
         deleteQuickAction,
     } = useActions(quickActionsLogic)
 
     const { user } = useValues(userLogic)
+    const { workflows, workflowsLoading } = useValues(workflowsLogic)
+    const { loadWorkflows } = useActions(workflowsLogic)
+
+    // workflowsLogic doesn't load on mount, so kick the fetch to populate the workflow picker.
+    useEffect(() => {
+        loadWorkflows()
```
mayteio: Data loading in a `useEffect` rather than the logic. There is a kea logic right here, and CLAUDE.md asks for business logic to live in it and to avoid hooks. An `afterMount` on `quickActionsLogic`, or a listener triggered by `openCreateModal` / `openEditModal`, would also mean the workflow list is only fetched when the modal is actually opened rather than on every render of the settings page.

posthog/plugins/plugin_server_api.py
HaynesPostHog: $Fixed. Ben confirmed the scoped-JWT direction, so this now mints a dedicated per-(team, workflow) JWT (\`WORKFLOWS_MANUAL_INVOCATION_JWT_SECRET\`) with a 2-minute expiry, mirroring \`_mint_reschedule_parked_jwt\` — no new \`INTERNAL_API_SECRET\` caller. Empty in prod until provisioned, so it fails closed. The verify side is in #73058 (a separate per-purpose key, not reused from reschedule).

products/workflows/frontend/Workflows/hogflows/steps/StepTrigger.tsx
HaynesPostHog: $Reverted — Ben's call was to re-hide it. Manual is now shown only when the current trigger is already \`manual\` (mirroring the schedule pattern), so it's hidden from new workflows again and this PR no longer makes a workflows launch-surface decision as a side effect. Quick actions still run any active workflow via the run endpoint regardless of trigger type, so nothing here depends on it being selectable. Also made the manual-config copy generic (dropped the Support/conversations reference) so it doesn't render conversations-specific text for teams without the product.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
HaynesPostHog: $Valid — there's no frontend equivalent of the facade today, so this import still reaches into \`workflowsLogic\`. I made the fetch lazy (only when the editor modal opens) but the cross-import remains. Creating a stable frontend export (a thin selector/hook from workflows) is a workflows-product architecture decision, so I'd rather not invent one unilaterally here — flagging for the workflows side. Happy to wire conversations onto it the moment it exists.

products/workflows/backend/facade/api.py
HaynesPostHog: $Fixed. \`invoke_hog_flow_now\` now wraps connection-level \`RequestException\` (timeout/refused/DNS) into \`HogFlowServiceError\` too, so callers catch exactly two facade exceptions and the raw \`requests\` error never escapes. Dropped the \`from requests.exceptions import RequestException\` from the caller in \`quick_actions.py\` accordingly — the docstring is now true.

products/conversations/backend/api/quick_actions.py
HaynesPostHog: $Deliberate, and now documented. The run path avoids the person resolver on purpose (it's the expensive lookup the comment guards against), so I fixed the facade docstring to stop advertising \`person\` and added a note that person-property filters won't match from a quick-action run. If we want them to, that's a product tradeoff (pay the person lookup on the synchronous run path) — flagging for a call rather than silently eating the cost. I did add actor + SLA globals, which are cheap and were the other half of the drift.

products/conversations/backend/api/quick_actions.py
HaynesPostHog: $Partly addressed. The drift you named was omission, not reimplementation — the builder already calls the same per-group helpers, it was just missing \`_get_actor_properties\` and \`_get_sla_properties\`. Added both, so each property group has a single source of truth and a workflow can branch on who ran it + SLA state. A fully-unified public builder is the right end state but has to reconcile the received-path's org-persist side effect (it writes the resolved org back to the ticket + does person processing, which a manual run must NOT do), so I left that as a follow-up rather than reshaping the hot capture path in this PR. Person stays out by design (above).

products/conversations/frontend/components/QuickActions/QuickActionPicker.tsx
HaynesPostHog: $Done exactly as suggested. The serializer now exposes \`workflow_runnable\`, resolved against the current environment in a single batched query (\`active_workflow_ids\`, memoized per response — no N+1), and the picker renders a workflow quick action with a \`disabledReason\` when its workflow isn't active here instead of letting the click fail.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
HaynesPostHog: $Addressed with the observable half. The run now fires an immediate \`Running "…"\` toast on click and an explicit "still running" toast if you click again while it's in flight, so the second click no longer vanishes silently. I didn't move the flag into a reducer to disable a trigger because the trigger (the \`/\` picker / toolbar item) closes on select — there's no persistent element left on screen to bind a loading state to. Open to a different surface if you have one in mind.

products/conversations/frontend/scenes/ticket/supportTicketSceneLogic.ts
HaynesPostHog: $Kept the mechanism but tightened it. The settle handshake was the subject of the must-fix race earlier in review (an earlier PATCH releasing a later run), which now has an invocation-identity guard and test coverage — so I was reluctant to swap it for an untested rewrite. I did clear the latch you flagged (\`ticketUpdateSettled\` is now dropped after the run consumes it). Collapsing it into one listener means either duplicating the PATCH body or extracting it out of the shared \`updateTicket\` (which the Save button and send-and-set-status also use), so I'd rather do that as its own change than fold it in here. Tracking it as a follow-up.

products/conversations/frontend/scenes/settings/QuickActionsSection.tsx
HaynesPostHog: $Made it lazy — the fetch now only fires when an editor modal actually opens, not on every settings-page render. I kept it in the component rather than \`quickActionsLogic\` because that logic is also mounted by the composer/\`/\` picker, and moving the workflow fetch into it would couple every composer mount to \`workflowsLogic\`. The lazy gate addresses the per-render fetch, which was the real cost.

PR #74034, opened 2026-07-27
fix(onboarding): reconnect wizard sync SSE streams after fatal closes

fercgomes:
## Problem

A native EventSource only auto-reconnects after a cleanly dropped connection. Any non-200 response on a (re)connect leaves it in readyState CLOSED permanently, and the SSE helper in posthog/api/streaming.py documents that consumers must schedule their own reconnect from onerror (its admission control returns 503 with a Retry-After that EventSource ignores). Both wizard sync stream consumers, taskRunStreamLogic and wizardSessionStreamLogic, handled the CLOSED case by storing an error string that says "call connect() to retry" while nothing ever called connect again, so one fatal close silenced the stream for the rest of the pageload.

## Changes

- Add `sseReconnectDelayMs` to `lib/wizard-sync/pollLoop.ts`: exponential backoff from 2s to a 30s cap over consecutive attempts, with the existing poll jitter so clients dropped together do not reconnect in lockstep.
- `taskRunStreamLogic`: when onerror fires with readyState CLOSED, schedule a `connect()` retry through a keyed disposable timer. The attempt counter lives on the logic cache and resets on a successful open.
- `wizardSessionStreamLogic`: same treatment for the session stream's SSE branch. The retry goes through `connect()`, which re-samples the sync-mode flag, so a mid-outage flip to polling takes effect on the next attempt.
- A scheduled retry marks itself on the cache so it continues the connect cycle that failed instead of starting a new one. The session logic's per-cycle transport telemetry would otherwise capture one "wizard sync transport error" per backoff step, which is one every 30s per tab for as long as a stream stays closed, and that metric is the SSE versus polling comparison.
- Both logics dispose the pending reconnect timer at the top of `connect` (a manual or scheduled connect supersedes it) and in `disconnect`, so an unmounted or explicitly disconnected stream never reconnects. The timers ride the kea-disposables lifecycle, which also pauses them while the tab is hidden.
- Add `lib/wizard-sync/eventSource.mock.ts`, a driveable EventSource stand-in for tests, since jsdom ships none.

> [!NOTE]
> A server that accepts the connection and then drops it fatally in a loop is retried every ~2s at first. The counter only resets on a successful open, so sustained failures settle at the 30s cap. Non-CLOSED transport errors get no scheduled reconnect: the browser is already retrying those with the Last-Event-ID cursor.

No manual testing. I am an agent and could not exercise this in a browser, so everything below is automated and was run locally:

- `jest src/scenes/onboarding/shared/wizard-sync/taskRunStreamLogic.test.ts` (48 tests)
- `jest products/wizard/frontend/wizardSessionStreamLogic.test.ts` (7 tests)

What the new tests catch that nothing did before:

- `sseReconnectDelayMs` cases: a base, exponent or cap regression that turns the retry into a hot loop.
- `taskRunStreamLogic transport`: a fatal close that never reconnects (the bug this PR fixes), a reconnect fired while the browser is already retrying natively, and a pending retry surviving `disconnect` or unmount.
- `wizardSessionStreamLogic SSE mode`: the same reconnect path, plus the telemetry guard, which fails as soon as a scheduled retry restarts the reporting cycle.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs impact. This is transport behavior behind an existing surface.

## 🤖 Agent context

I (actually Claude Code, Opus 5) wrote this. No repo skills were invoked.

Decisions worth knowing: the reconnect deliberately goes through the existing `connect` action rather than a private helper, so it re-runs every guard connect already owns (project resolved, run still the active one, sync mode re-sampled). The per-cycle telemetry was the one thing that could not ride along with that, so scheduled retries flag themselves on the cache and leave the counters alone. The per-run TaskRun stream stays SSE only and is deliberately not put behind the sync-mode flag: its REST detail carries no progress steps, so a polled fallback would render an empty pipeline.

## Validating after merge

The change is client-side, so the server's own stream counters are where the reconnects show up.

- `posthog_tasks_task_run_stream_connections_opened_total{origin_product="onboarding"}` should rise against `posthog_tasks_task_run_stream_connections_closed_total{origin_product="onboarding",outcome="stream_error"}` and `{outcome="unavailable"}`. Before this, one fatal close ended a client's stream for the rest of the pageload, so opens per run were effectively capped at one per surface. After it, every non-`client_disconnect` close should be followed by another open from the same client.
- The rate is the guard against a hot loop. Backoff runs 2s to a 30s cap and only resets on a successful open, so a client failing continuously settles near 2 opens/minute. Sustained `opened_total` above that per concurrent run means the backoff or the counter reset regressed.
- `posthog_tasks_task_run_stream_resume_gap_total{origin_product="onboarding"}` will rise from close to zero, because reconnects that never happened before now arrive with a `Last-Event-ID`. That is the expected direction. What would be a regression is the gap count tracking `opened_total` one for one, which would mean reconnects are consistently landing past the trim threshold and clients are losing events rather than resuming.
- `wizard sync transport error` (project 2) must stay at most one per `workflow_id` per session. More than one is the per-cycle telemetry guard failing, which turns the SSE-versus-polling comparison into a count of backoff steps. `wizard sync transport ready` with `transport=sse` should rise, since a stream that reconnects can still report ready.
- `posthog_tasks_task_run_stream_connection_duration_seconds{outcome="completed"}` should not shorten. Shorter completed connections alongside more opens would mean reconnects are displacing healthy streams instead of replacing dead ones.

Fixes #74316

### DIFF
--- a/frontend/src/lib/wizard-sync/eventSource.mock.ts
+++ b/frontend/src/lib/wizard-sync/eventSource.mock.ts
@@ -1,0 +1,100 @@
+/**
+ * A driveable EventSource stand-in for tests. jsdom ships no EventSource at all, and the wizard sync
+ * transports need one a test can step frame by frame: open, data frames, named events (`stream-end`,
+ * the server's `error` payload) and browser transport failures, which differ only by readyState.
+ *
+ * Handlers registered both ways are invoked, since the transports use `onmessage`/`onerror`
+ * assignment for data and `addEventListener` for named events.
+ */
+export class MockEventSource {
+    static readonly CONNECTING = 0
+    static readonly OPEN = 1
+    static readonly CLOSED = 2
+
+    /** Every source opened since the last `reset()`, in order — the count is how many connects happened. */
+    static instances: MockEventSource[] = []
+
+    readyState: number = MockEventSource.CONNECTING
+    onopen: ((event: Event) => void) | null = null
+    onmessage: ((event: MessageEvent) => void) | null = null
+    onerror: ((event: Event) => void) | null = null
+
+    private listeners = new Map<string, ((event: Event) => void)[]>()
+
+    constructor(readonly url: string) {
+        MockEventSource.instances.push(this)
+    }
+
+    static reset(): void {
+        MockEventSource.instances = []
+    }
+
+    static last(): MockEventSource {
+        const source = MockEventSource.instances[MockEventSource.instances.length - 1]
+        if (!source) {
+            throw new Error('No EventSource was opened')
+        }
+        return source
+    }
+
+    addEventListener(name: string, handler: (event: Event) => void): void {
+        this.listeners.set(name, [...(this.listeners.get(name) ?? []), handler])
+    }
+
+    removeEventListener(name: string, handler: (event: Event) => void): void {
+        this.listeners.set(
+            name,
+            (this.listeners.get(name) ?? []).filter((h) => h !== handler)
+        )
+    }
+
+    close(): void {
+        this.readyState = MockEventSource.CLOSED
+    }
+
+    emitOpen(): void {
+        this.readyState = MockEventSource.OPEN
+        const event = new Event('open')
+        this.onopen?.(event)
+        this.dispatch('open', event)
+    }
+
+    /** An unnamed data frame, the shape both transports read run/session state from. */
+    emitMessage(data: string): void {
+        const event = new MessageEvent('message', { data })
+        this.onmessage?.(event)
+        this.dispatch('message', event)
+    }
+
+    /** A named event. With `data` it is a MessageEvent, which is what tells a server-sent
+     * `error` apart from a transport failure. */
+    emitNamed(name: string, data?: string): void {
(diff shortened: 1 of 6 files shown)


fercgomes:
* **#74037** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74037?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74036** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74036?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74035** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74034** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74034?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74034?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #74035, opened 2026-07-27
fix(onboarding): recover the run card when the stream delivers no state

fercgomes:
## Problem

The task-run SSE endpoint emits a NAMED `error` event (a MessageEvent carrying `{"error": "Stream not available"}`) once it gives up waiting for a run's Redis stream key, then closes cleanly. `taskRunStreamLogic` never listened for that event, and a clean close is exactly what a native EventSource answers by reconnecting, so the client re-entered the same server-side wait forever while `cloudProgress` held the surfaces on the `idle` phase. `idle` renders a spinner with no recovery controls, and its only other exit was a state event that, by construction, was never coming. `taskRunDetailToStreamState` already existed to project the REST run snapshot onto the stream's state shape but had no caller.

## Changes

- `taskRunStreamLogic` listens for `error` with `addEventListener` and branches on `event.data`: a payload means the server has no stream to serve, no payload means a browser transport failure (the existing CLOSED and backoff handling, now factored into `scheduleReconnect`).
- On a server error payload the logic drops the EventSource (so the browser cannot reconnect into the same wait), fetches the run once through `tasksRunsRetrieve`, and feeds it through `taskRunDetailToStreamState`: a terminal status settles the stream via `streamCompleted`, a live run schedules a backed-off reconnect, and a permanent fetch error (401/403/404/410, per `isPermanentPollError`) marks the stream dead with `runStalled`.
- New `NO_STATE_STALL_MS` (4 minutes, longer than the server's own wait for a missing stream key) fixes an absolute deadline the first time the stream connects with nothing to show. A scheduled retry keeps that deadline instead of restarting it, so a transport looping faster than the window cannot push it back forever, and a connect made after the deadline passed reports the stall straight away rather than clearing it. The first state event disarms the whole thing.
- `cloudProgress` maps "no run state plus stalled" to the `error` phase with a `Setup lost contact` title, which is the phase that already carries the retry CTAs and the dismiss control.
- `parseStreamErrorMessage` parses the named event's payload, returning null for anything that is not a `{error: string}` body.

> [!NOTE]
> `isStalled` is now set by two independent conditions. The queued-stall path is unchanged. The no-state path can only fire while `taskRunState` is null, and `cloudProgress` checks that case first. A terminal state arriving through the fallback does not fire the funnel's completion event: `cloudRunCompletionReport` requires an observed non-terminal to terminal transition, and the backend already reports runs that finish unobserved.

No manual testing. I am an agent and could not exercise this in a browser, so everything below is automated and was run locally:

- `jest src/scenes/onboarding/shared/wizard-sync` (7 suites, 174 tests)

What the new tests catch that nothing did before:

- `parseStreamErrorMessage` cases: a transport failure being mistaken for the server's named event, which would fire the REST fallback on every reconnect.
- run-detail fallback cases: a server error payload that does not settle a finished run, does not retry a live one, or retries a run that is gone for good.
- `reports a stall when the stream delivers no state at all`: the eternal spinner this PR removes.
- `keeps the no-state stall through a scheduled reconnect`: a reconnect clearing `isStalled` with nothing left to re-arm, which puts the spinner back with no error CTAs and no dismiss.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs impact. Recovery behavior inside an existing surface.

## 🤖 Agent context

I (actually Claude Code, Opus 5) wrote this. No repo skills were invoked.

Decisions worth knowing: the named `error` event and the browser's transport failures both dispatch under the name `error`, so the branch is on `typeof event.data === 'string'` rather than on two separate handlers, and the old `onerror` assignment had to go for `addEventListener` to see both. The no-state timer went through two designs before this one. Armed once per logic lifetime it could never be re-armed after firing, and re-armed per connect it could be pushed back forever by a retry loop that is faster than the window. An absolute deadline, refreshed only by a fresh connect attempt, is what satisfies both. The run-detail fallback resolves outside kea, so `scheduleReconnect` returns early when the disposables manager is already gone.

## Validating after merge

This mostly turns an invisible dead end into a visible one. It does not change how many runs finish, so the server-side run counters are the control, not the signal.

- `posthog_tasks_task_run_stream_connections_closed_total{origin_product="onboarding",outcome="unavailable"}` is the exact server event this PR now handles. Its ratio to `posthog_tasks_task_run_stream_connections_opened_total{origin_product="onboarding"}` should fall: each `unavailable` close used to be answered by an immediate native reconnect into the same wait, and is now answered by one REST fetch plus at most one backed-off reconnect.
- `wizard sync run dismissed` (project 2) with `mode="cloud"` and `phase="error"` should appear for a class of runs that could not produce it before, because the surface held `phase="idle"` with no dismiss control. Direction: up from roughly zero.
- The falsifier for that same event is `elapsed_seconds`. `NO_STATE_STALL_MS` is 4 minutes, so dismissals at `phase="error"` with `elapsed_seconds` under 240 mean the no-state deadline is firing on healthy runs and the recovery CTAs are being offered while the stream is still working.
- Control: `posthog_tasks_runs_in_status{origin_product="onboarding"}` and `posthog_tasks_oldest_open_run_age_seconds{origin_product="onboarding"}` should be flat. Nothing here cancels or completes a run, it only lets the client stop waiting on one, so movement there points at something other than this change.
- Server-side `task_run_started` with no `task_run_completed`, `task_run_failed` or `task_run_cancelled` on the same `properties.run_id` is the population this PR is for. That population should not shrink; what should change is that its clients reach a terminal surface instead of a spinner.

Fixes #74316

### DIFF
--- a/frontend/snapshots.yml
+++ b/frontend/snapshots.yml
@@ -5573,7 +5573,7 @@ snapshots:
     scenes-app-dashboards--show-across-viewports--narrow--dark:
         hash: v1.k794b7964.e919c4955d7113170ab00269c5f628c318afe14f26dc8aafb046b172e17f362e.YDXD84rrb2KR1I36pHMxEsGZ0sN0os_cBBdNIf89Pu8
     scenes-app-dashboards--show-across-viewports--narrow--light:
-        hash: v1.k794b7964.2a88f06991ca4c13ad27ee62db8eaa037c542301e28c24934eedc1bbb78cb338.3Ne0cm6_2xQiGE1wGcRzmw0QBd3QrEYkELz3PCg9J48
+        hash: v1.k794b7964.a46d840722e971910236d6fcd8a8f69e0b1db5ee9d472d62055b02d6569f7a99.kHEdVPNUm5DRKSlMFS12SLKf1TyChQJZKKuZ2PKvtfA
     scenes-app-dashboards--show-across-viewports--superwide--dark:
         hash: v1.k794b7964.9948b52cf842206c1584057a6a16b8caed7a32d95f103dbea07089a02df75a53.AuXGqOvw-P-XuzOL2Zl-BkBPb1-Q3R1949f5-IkTjZA
     scenes-app-dashboards--show-across-viewports--superwide--light:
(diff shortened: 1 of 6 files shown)


fercgomes:
* **#74037** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74037?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74036** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74036?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74035** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74035?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74034** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74034?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

PR #74038, opened 2026-07-27
fix(tasks): stop infra log lines resetting the run inactivity timer

fercgomes:
## Problem

A cloud run must stop when its agent goes silent. The workflow enforces this with an inactivity timer. Today the timer never expires, because the run keeps resetting it with its own infra output.

The loop works like this:

1. `append_task_run_log` sends a heartbeat with `agent_active=True` for every batch of entries the sandbox posts.
2. The workflow runs a credential refresh loop. The first interval is 20 minutes.
3. Each refresh makes the sandbox write an infra log line.
4. That line arrives through the same append path. It sends a heartbeat.
5. The heartbeat restarts the inactivity window. The default window is 30 minutes.

20 minutes is shorter than 30 minutes. Therefore a run whose agent stopped completely resets its own timer forever. The run never reaches the timeout. It stays `in_progress`.

The workflow signal handler already ignores `agent_active=False`. The fix therefore belongs at the point where the facade decides the value.

Fixes #74316

## Changes

- Add `_entries_show_agent_activity` to `facade/api.py`. The function classifies one batch of log entries.
- Change `append_task_run_log`. It now sends the classified value instead of a constant `True`.
- Keep the S3 write unchanged. The code appends every entry in both cases.

The classifier applies these rules:

| Batch content | Result | Reason |
|---|---|---|
| An ACP batch with a `session/*` method | `True` | The agent sends messages, tool calls, or permission requests. |
| An ACP batch with only `_posthog/console`, `_posthog/error`, or a method that is not a string | `False` | These are infra frames. They must not extend the window. |
| A batch with no ACP frame, for example plain `{type, message}` | `True` | These callers have no `session/*` frame. They keep the old behavior. |
| An empty batch | `False` | There is no evidence of work. |

> [!NOTE]
> `agent_active` is a one-way switch. `TaskRun.heartbeat_workflow` returns immediately on `False`, and the workflow signal handler ignores `False` as well. A `False` value therefore cannot shorten a window. It can only fail to extend one.

Two stronger heartbeat paths do not change:

- The event ingest path keeps state. A `session/update` sets a latch in Redis. After that, every event sends a heartbeat, at most one per 30 seconds. A long tool turn that writes only console output stays alive.
- The agent proxy posts `kind: "heartbeat"` with `agent_active=True`. That path does not depend on ACP traffic.

This PR narrows the third and weakest path. That path holds no state between requests.

## Risk

An agent runtime that never sends a `session/*` notification would stop extending the window. It would time out on the idle grace instead. The relay path and the agent proxy path both send ACP session notifications for real agent output. The change should therefore affect only runs that produce no agent output. That result is the intended behavior.

Automated tests only. Nobody ran a manual test of this path.

- `pytest products/tasks/backend/tests/test_facade.py -k AppendLogAgentActivity`. 10 tests pass locally.
- The updated test in `test_api.py` covers the endpoint wiring.

New tests and the regression that each one catches:

- Parameterized cases for `_entries_show_agent_activity`. They catch the original defect: an infra-only batch must not report agent activity. They also catch the opposite defect: a batch with no ACP frame must keep the old behavior.
- One wiring test. It asserts that `append_task_run_log` passes the classified value into `heartbeat_workflow`. Without it, a correct classifier could still be ignored at the call site.

## Validating after merge

- `posthog_tasks_oldest_open_run_age_seconds{status="in_progress", origin_product="onboarding"}` is the primary signal. Before this change the age could grow without a bound. After this change the age must stop above the effective inactivity timeout and stay in a bounded band. An age that keeps growing means that the classifier still counts an infra line as agent activity.
- `posthog_tasks_runs_in_status{status="in_progress", origin_product="onboarding", run_environment="cloud"}` must fall. Silent runs now reach the timeout. A flat or rising value after one week of full rollout is the regression signal.
- The opposite failure is over-classification. A live agent could time out too early. Watch `task_run_completed` for a rise in short `duration_seconds` on onboarding runs. Also watch Loki with `namespace="temporal-worker-tasks-agent"` and `origin_product="onboarding"` for inactivity terminations on runs that still produced output.
- Log completeness must not move. The code appends every entry in both cases.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior changes. No docs update is needed.

## 🤖 Agent context

I wrote this change with Claude Code.

The first version judged every batch and returned `False` for any batch without a `session/*` frame. A review comment showed the problem: callers that post only generic `{type, message}` entries have no `session/*` frame, and those runs would start to time out while they still work. The second version therefore checks the shape of the batch first. It judges only a batch that carries at least one dict `notification`. Everything else keeps the old heartbeat.

I also rejected a gate on the protected wizard marker. The leak is not specific to the wizard. The credential refresh loop runs for every run that holds GitHub credentials. A wizard gate would leave the same loop in place for code runs, and those runs live longer.

The written parts of this description follow ASD-STE100, a controlled-English standard: one meaning per word, active voice, simple tenses, and short sentences.


fercgomes:
* **#74041** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74041?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74040** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74040?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74039** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74039?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74038** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74038?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74038?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

products/tasks/backend/facade/api.py
```diff
@@ -2192,15 +2192,38 @@ def set_task_run_output(
     return _task_run_detail_to_dto(run)
 
 
+def _entries_show_agent_activity(entries: list[dict]) -> bool:
```
tatoalo, line 2418: a single tool turn can emit one `session/update` followed by separate `_posthog/console` batches which will be treated as inactive even though the tool is still producing output. The stream-ingestion path retains active state until turn completion. Can this path use the same stateful behavior or an explicit heartbeat independent of each log batch?

products/tasks/backend/facade/api.py
```diff
@@ -2192,15 +2192,38 @@ def set_task_run_output(
...
+    The heartbeat only reports the agent as active when the entries show real agent
+    activity; infra entries are still appended but heartbeat with ``agent_active=False``,
+    which the workflow's signal handler ignores.
+    """
     run = _get_visible_run(run_id, task_id, team_id)
     if run is None:
         return None
     run.append_log(entries)
-    run.heartbeat_workflow(agent_active=True)
+    run.heartbeat_workflow(agent_active=_entries_show_agent_activity(entries))
```
tatoalo, line 2457: if we change this for all runs, we can have generic `{type, message}` entries without ACP notificatios sothose runs previously remained active but will now time out while still working which seems not ideal 🤔  

could we gate this classification on the protected wizard marker and preserve `agent_active=True` for other runs?

products/tasks/backend/facade/api.py
fercgomes: `agent_active=False` is not an idle marker, which I think is the crux here. `TaskRun.heartbeat_workflow` returns before connecting and `ProcessTaskWorkflow.heartbeat` ignores it, so a console-only batch simply fails to extend the timer. It never shortens it, and it cannot move a run toward timeout on its own.

Mid-turn extension still comes from the stateful path you are describing. Every cloud run starts either the SSE relay (when `sandbox_event_ingest_enabled` is off) or the sequenced ingest endpoint. Both latch until `end_of_turn` and heartbeat every 30s while events arrive, including console output from a tool that is still running. So the stateful behaviour is present, it just does not live in the facade classifier.

Putting the latch in the facade would need turn state, and the only store for it is the ingest Redis flag, which relay-path runs never write. That would heartbeat less often than today, not more.

The case genuinely not covered is a dead event channel, where no ACP frames arrive at all. There, falling back to the idle window is the behaviour we want.

products/tasks/backend/facade/api.py
fercgomes: Good catch on the generic entries, fixed, but without the wizard gate.

`_entries_show_agent_activity` now decides whether the batch speaks ACP at all before it judges it. It tracks whether any entry carried a dict `notification`, returns True on the first `session/*` method, and otherwise returns `bool(entries) and not saw_acp_frame`. So a batch of plain `{type, message}` entries keeps the old heartbeat and those runs stay active exactly as before, while a sandbox batch that does carry ACP frames stays fail-closed: `_posthog/console`, `_posthog/error` and malformed methods do not extend the inactivity window. Empty batches remain False.

On gating this on the protected wizard marker: the leak is not wizard-specific. The credential refresh loop runs for every run that has GitHub credentials, on a 20 minute cadence under a 30 minute inactivity window, so each refresh lands inside the window and rearms it. Gating on wizard would leave exactly the same self-sustaining heartbeat in place for code runs, which are the longer-lived ones. Keying on the ACP shape fixes both classes of run and keeps the classifier honest about what it is actually looking at. Anything that still slips through is bounded by the wall clock cap in #74039.

Tests updated to match: `no_notification` and `malformed_notification` now expect True, `non_string_method` and `empty_entries` stay False, and I added a case where a plain line rides alongside a `_posthog/console` frame and correctly returns False. The append_log API test is back on the original generic entry and still asserts `agent_active=True`, which now covers the path you flagged.

tatoalo: a real long-running turn of a task can time out despite continuing to produce output. Does the ACP protocol guarantee recurring session/* notifications within the inactivity window?

fercgomes: No, ACP does not guarantee that, and we should not rely on it. The window is not extended by ACP notification cadence though, so a console-only tool turn still heartbeats.

The extension comes from a latch in Redis, not from each batch being judged on its own. In `_heartbeat_workflow_if_needed` (event_ingest.py:372) the first `session/update` sets `agent_active=True` (`_is_session_update`, event_ingest.py:431-435). After that the latch is read rather than recomputed, so **any** subsequent event heartbeats while it is on, including `_posthog/console` frames from a long-running tool, throttled to one per `HEARTBEAT_THROTTLE_SECONDS = 30` (event_ingest.py:41). It only clears on `is_turn_complete`. So the guarantee we need is not "recurring `session/*`", it is "the turn produces some output and eventually ends", which is the case you are describing.

There is a second, independent source that does not depend on ACP traffic at all: the agent proxy posts `kind: "heartbeat"` with `agent_active=True`, and that calls `heartbeat_workflow` directly (agent_proxy_callback.py:112-120). A turn that went genuinely silent at the ACP layer is still covered by that.

Against a 30 minute default window (`INACTIVITY_TIMEOUT_DEFAULT_SECONDS`, 1 hour for user-driven runs) a 30 second throttle leaves a wide margin.

What this PR changes is the third and weakest path, `append_task_run_log` (api.py:2203). That one is stateless per request by construction, and it was the one extending the window off infra frames alone, which is the leak: credential refresh runs every 20 minutes for any run with GitHub credentials, so its console output alone kept rearming a 30 minute window forever. Narrowing it is the point.

The honest residual: a run whose only signal is the log-append endpoint, with no ingest events and no proxy heartbeat, that emits only infra ACP frames, will now fall back to the idle window. That is the fail-closed case we want, since by then we have no evidence of a live agent. If you would rather that case stayed active, the fix is a latch with real turn state rather than widening the classifier, and the only store for it is the ingest Redis flag, which relay-path runs never write.

MattBro (approved):
Checked the latch claim independently and it holds. `_heartbeat_workflow_if_needed` sets `agent_active` on the first `session/update` and reads it rather than recomputing for every later event, so console-only output from a still-running tool keeps heartbeating until the turn ends. tatoalo's slow-turn case is covered.

The premise checks out too: `CREDENTIAL_REFRESH_INITIAL_DELAY` is 20m against a 30m `INACTIVITY_TIMEOUT_DEFAULT_SECONDS`, so every refresh lands inside the window and rearms it. Seen in prod on a stuck onboarding run: `sandbox_credentials_refreshed` every 20 minutes for 6 hours after the agent had finished.

PR #74040, opened 2026-07-27
feat(tasks): sweep stale in-progress cloud runs and bound workflow lifetime

fercgomes:
## Why

Both existing TaskRun janitors sweep `QUEUED` only, so they cover runs that never started and nothing else. Once a run reaches `IN_PROGRESS` its `process-task` workflow owns the terminal write; if that workflow is gone (worker lost, history unusable, execution closed out of band) the row stays `IN_PROGRESS` with no writer left, and every check that looks for a live run keeps finding it.

The `process-task` start calls also pass no `execution_timeout`, so a workflow execution has no server-side ceiling at all. The in-workflow wall-clock cap only helps a workflow that is still executing its own code.

## What

- Add `facade.get_stale_in_progress_cloud_task_run_ids(older_than, limit)`: cloud runs in `IN_PROGRESS` whose `updated_at` is older than the window. Local runs are excluded, since a desktop-driven run reports on its own cadence and silence there is not evidence of a dead owner. Interactive runs are excluded too: they are exempt from the wall-clock cap, and the live path writes S3 and Temporal rather than the run row, so a quiet `updated_at` there says nothing about whether the workflow is alive.
- Add the `kill_stale_in_progress_task_runs` celery task, registered hourly at minute 20 (offset from the QUEUED sweep at minute 0). Its window is `TASKS_MAX_RUN_DURATION_SECONDS` plus a one-hour grace, read from the setting so the two bounds cannot drift apart. It terminalizes through `claim_and_fail_stale_run` with `error_type="stale_in_progress_cleanup"`, carries swept/error Prometheus counters, and emits a `sweep_done` log matching the QUEUED sweeps.
- Set `execution_timeout=PROCESS_TASK_EXECUTION_TIMEOUT` (12h) on all four `process-task` `start_workflow` calls in `temporal/client.py`.

## Notes for reviewers

- `claim_and_fail_stale_run` is a compare-and-set over `QUEUED`/`IN_PROGRESS`, so a workflow that wakes up and terminalizes between selection and claim wins and the sweep skips the row. Ordering by `updated_at` with a 500-row batch keeps a backlog draining oldest-first across runs.
- `execution_timeout` bounds the whole execution chain, `continue_as_new` continuations included. 12h is deliberately far above the in-workflow cap so it never preempts the proper terminalization path, but an interactive session (exempt from the in-workflow cap) held open past 12h will now be closed by the server.
- The sweep window keys off the setting, so lowering `TASKS_MAX_RUN_DURATION_SECONDS` for a local repro also tightens the sweep. Covered by `test_window_tracks_the_configured_run_duration_cap`. When the setting is 0 or negative the cap is disabled, so the window falls back to `PROCESS_TASK_EXECUTION_TIMEOUT`, the only remaining bound; without that floor a disabled cap would reap every run older than the grace hour.
- The interactive exclusion is written as `Q(state__mode__isnull=True) | ~Q(state__mode="interactive")` rather than a plain `exclude`. An absent `mode` key means background, and `exclude` on a JSON key compares against NULL, which would drop those rows and silently disable the sweep for the majority of runs. `test_sweeps_runs_without_an_interactive_mode` pins that.
- Test plan: `pytest posthog/tasks/test/test_kill_stale_in_progress_task_runs.py` (17 passed), `pytest products/tasks/backend/tests/test_facade.py -k stale` (5 passed). The registration snapshot is updated.

## Validating after merge

- The sweep is running and finding work when `posthog_task_run_stale_in_progress_swept_total` (new in this PR) is non-zero. Expected shape is a burst on the first few hourly runs as the existing backlog drains, then a low steady rate. Staying at zero forever means the sweep is selecting nothing and the janitor is inert; a rate that never decays means rows are being created stuck as fast as they are reaped, which points upstream rather than here.
- `posthog_task_run_stale_in_progress_errors_total` (new in this PR) should stay at zero. Any sustained non-zero value means `mark_failed` is raising after the claim committed, leaving terminal rows without `completed_at`.
- `posthog_tasks_oldest_open_run_age_seconds{status="in_progress", run_environment="cloud"}` is the outcome metric: it should fall to and stay under `TASKS_MAX_RUN_DURATION_SECONDS` plus the one-hour grace for capped runs, and under the Temporal execution timeout plus the same grace for interactive ones. A value above either bound means the corresponding tier of the query is not matching.
- `posthog_tasks_runs_in_status{status="in_progress", run_environment="cloud"}` should stop growing monotonically and start tracking live concurrency instead.
- Regression to watch: healthy long runs being reaped mid-flight. Those surface as `task_run_failed` (server event) with `properties.error_type = "stale_in_progress_cleanup"` on runs whose age is under the cap, and in the `kill_stale_in_progress_task_runs.sweep_done` log in Loki, `namespace="temporal-worker-tasks-agent"`, where a `saturated=true` line means the 500-row batch filled and the window is likely too tight.

Fixes #74316


fercgomes:
* **#74041** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74041?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74040** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74040?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74039** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74039?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74038** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74038?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

products/tasks/backend/temporal/constants.py
```diff
@@ -82,6 +82,12 @@ def resolve_max_run_duration() -> timedelta | None:
     return timedelta(seconds=seconds) if seconds > 0 else None
 
 
+# Server-side ceiling on a whole `process-task` execution chain (continue-as-new included), set on
+# every start_workflow call. Deliberately far above `resolve_max_run_duration()`: the in-workflow cap
+# is the one that terminalizes the run properly, and this only stops an execution that lost the
+# ability to do so from living forever in Temporal.
+PROCESS_TASK_EXECUTION_TIMEOUT = timedelta(hours=12)
```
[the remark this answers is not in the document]
fercgomes: Both halves were real. Fixed.

**Ceiling can no longer be outrun by the cap.** `PROCESS_TASK_EXECUTION_TIMEOUT` is now `resolve_process_task_execution_timeout()`, returning `max(12h floor, TASKS_MAX_RUN_DURATION_SECONDS + 1h headroom)`, and `client.py` calls it at each `start_workflow`. A cap configured at or above 12h pushes the ceiling out with it instead of being preempted by it, so the in-workflow cap stays the bound that actually terminalizes the run.

**Interactive runs are no longer excluded from the sweep, only held to a wider window.** `get_stale_in_progress_cloud_task_run_ids` now takes two windows: capped runs at cap + grace, interactive runs at the execution timeout + the same grace. The wider window is sound because the run row is written by the workflow itself when it takes the run IN_PROGRESS, so the workflow started no later than the last write to the row; a row quiet for longer than the ceiling therefore belongs to an execution Temporal has already closed. Local runs stay excluded at every window, since they have no Temporal execution behind them at all.

Covered by `test_sweeps_interactive_runs_past_the_execution_timeout`, `test_leaves_interactive_runs_inside_the_execution_timeout_alone`, `test_leaves_local_interactive_runs_alone_at_any_age`, `test_uncapped_window_tracks_a_cap_raised_past_the_execution_timeout_floor`, and `test_execution_timeout_cutoff_stays_above_a_cap_raised_past_the_floor`.

posthog/tasks/tasks.py
```diff
@@ -267,6 +277,58 @@ def _sweep_each(
...
+    BATCH_SIZE = 500
+    GRACE = datetime.timedelta(hours=1)
+    STALE_AFTER = tasks_facade.stale_in_progress_cutoff(GRACE)
+    REASON = "Run was stuck in IN_PROGRESS with no live workflow and was killed by the cleanup job."
+
+    # Janitor sweep is intentionally cross-team — it runs without a team context.
+    stale_ids = tasks_facade.get_stale_in_progress_cloud_task_run_ids(STALE_AFTER, BATCH_SIZE)
+    swept = errors = 0
+    for run_id in stale_ids:
+        try:
```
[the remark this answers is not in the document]
fercgomes: Not changing this. `claim_and_fail_stale_run` is unchanged by this PR (byte-identical on master) and is the same helper the existing QUEUED sweeps already call; this PR only adds a caller.

The end state described is also not a stuck row. The compare-and-set has already committed `FAILED`, which is terminal, so the run has left `IN_PROGRESS` and stops counting as running, which is the entire job of the sweep. What is missing is `completed_at` and the error message: cosmetic fields on an already-terminal row. The `mark_failed` exception is not silent either, it increments `posthog_task_run_stale_in_progress_errors_total` and is sent to error tracking, so the case is observable.

Re-selecting `FAILED` rows to backfill those fields would mean a sweep that repeatedly walks terminal rows on every run for no change in run lifecycle behavior.

products/tasks/backend/temporal/client.py
```diff
@@ -244,6 +245,7 @@ async def execute_task_processing_workflow_async(
             id_reuse_policy=WorkflowIDReusePolicy.ALLOW_DUPLICATE_FAILED_ONLY,
             task_queue=settings.TASKS_TASK_QUEUE,
             retry_policy=RetryPolicy(maximum_attempts=3),
+            execution_timeout=resolve_process_task_execution_timeout(),
```
tatoalo: this timeout terminates unrelated background and interactive runs, can we gate against wizard eligibility/deadline into these start paths and leave existing timeout behavior unchanged for other runs?

tatoalo (commented):
we do already have janitors for orphaned QUEUED runs, this handles IN_PROGRESS  which is nice as well.
Can we make that relationship explicit and scope this reconciler to Wizard runs? The new shared execution timeout creates stale rows for every cloud-run producer, and this global age-based sweep then terminate them without confirming that their workflow has stopped.

posthog/tasks/tasks.py
```diff
@@ -267,6 +277,63 @@ def _sweep_each(
...
+    untouched for longer than that cap plus a grace hour provably has no workflow behind it. Keying
+    on the setting (rather than a second hardcoded window) keeps the two bounds from drifting apart:
+    raising the cap automatically pushes this sweep out with it.
+
+    Interactive sessions are uncapped and stay quiet for as long as their human does, so that window
+    says nothing about them. They get the wider one: the Temporal execution timeout, which is the
+    only bound they have and which the server enforces whether or not the workflow can still write.
+    A session closed at that ceiling would otherwise sit IN_PROGRESS with no writer left at all.
+
+    ``claim_and_fail_stale_run`` is a compare-and-set on ``QUEUED``/``IN_PROGRESS``, so a workflow
```
tatoalo: status transition and the remaining terminal fields are separate writes. If mark_failed or a later side effect raises, the row is already FAILED and will never be selected by this sweep again

products/tasks/backend/temporal/client.py
fercgomes: The ceiling is derived from the cap rather than fixed: `resolve_process_task_execution_timeout()` returns `max(12h, cap + 1h)`, so a capped run is always terminalized by its own in-workflow cap first and never reaches the Temporal ceiling. The ceiling only exists as a backstop for runs the cap does not apply to.

Interactive runs are the only ones that can actually reach it, and their sandbox is created with a 6h Modal timeout (`SANDBOX_TTL_SECONDS`), so 12h is already twice the life of the sandbox the session depends on. A resume starts a fresh execution with a fresh clock, so a long-lived interactive session is not truncated by this either.

Gating it to wizard runs would also break the sweep this PR adds: the interactive tier's window is defined by the ceiling, so without one we would be failing rows whose executions have no bound at all, which is the situation the sweep exists to detect.

posthog/tasks/tasks.py
fercgomes: Same point greptile raised above, and that reply carries most of the reasoning. Two refinements specific to your framing.

`mark_failed` writes status, `error_message` and `completed_at` in a single save, so the only window that leaves a partial row is a DB error between the compare-and-set claim and that save. The side effects after it (stream publish, analytics, push) all run against an already complete row, so they cannot produce the half-written state.

A partial row is still terminal, which is what the sweep is for: it is no longer occupying a slot or blocking the user. What it would be missing is `error_message` and `completed_at`. That case is counted by `posthog_task_run_stale_in_progress_errors_total`, reported to error tracking, and pinned by `test_one_failure_does_not_block_the_sweep`, so it is visible rather than silent.

If you would rather it be airtight than merely observable, the fix is folding both fields into the compare-and-set inside `claim_and_fail_stale_run`. That is a shared helper this PR does not otherwise touch, so I left it alone, but I am happy to do it here if you prefer.

MattBro (approved):

fercgomes:

PR #74044, opened 2026-07-27
fix(tasks): retry wizard run cancellation from PR-close webhooks

fercgomes:
## Why

`_cancel_wizard_run_on_close` ran `cancel_task_run` inline from the GitHub PR webhook, inside a bare `except Exception` that logged and returned so the handler could stay 2xx. That combination has no recovery path: `cancel_task_run` does a synchronous Temporal round-trip plus DB writes, GitHub does not redeliver a delivery we answered 2xx, and nothing else watches for a closed setup PR. Any transient failure dropped the cancellation permanently, leaving the run non-terminal with its sandbox alive until the TTL expired.

The swallow is also wider than the failure it was there for: `cancel_task_run` does not raise on an unreachable Temporal, it returns the `unavailable` outcome, which the handler discarded entirely.

## What

- New `cancel_wizard_run_on_pr_close_task` celery task in `facade/tasks.py`. It calls `cancel_task_run` with `source="pr_closed"` and retries on both failure shapes: an exception, and the `unavailable` outcome that `cancel_task_run` returns instead of raising. Bounded at `CANCEL_WIZARD_RUN_MAX_RETRIES = 5` with the countdown doubling from 30s, so the last attempt lands about 15 minutes out and then the task fails for good.
- `_cancel_wizard_run_on_close` now enqueues that task from `transaction.on_commit` and keeps its existing wizard/cloud/non-terminal pre-checks. Only the enqueue is wrapped in `except Exception`, so the handler still cannot 5xx on GitHub, but the failure it swallows is now a broker hiccup rather than the whole cancellation.
- The reason string moves to a module constant `PR_CLOSED_CANCEL_REASON`; `webhooks.py` no longer imports `cancel_task_run`.
- The enqueue itself still cannot raise (the handler has already answered 2xx and GitHub does not redeliver on its own), so its failure increments `posthog_tasks_cancel_enqueue_failed_total{kind="pr_closed"}` instead of being logged and forgotten.
- Tests: the two PR-close tests now assert on the enqueue instead of the inline call, plus new coverage that the task retries on `unavailable` and on a raised error, and does not retry on `accepted` or `not_found`.

## Notes for reviewers

Retries make the cancellation at-least-once. `cancel_task_run` short-circuits on `run.is_terminal` and the underlying PATCH is `only_if_non_terminal`, so a second delivery on an already-cancelled run is a no-op; the visible effect of a duplicate is at most a repeated `task_run_cancel_requested` analytics event.

The webhook response is unchanged: still 200 on every path, including when the broker is unreachable.

Test plan: `pytest products/tasks/backend/tests/test_webhooks.py` (76 tests) passes locally.

## Validating after merge

- `posthog_tasks_cancel_enqueue_failed_total{kind="pr_closed"}` stays at zero. It is the only remaining way a PR-close cancellation can be dropped, so a nonzero rate is the signal that the broker, not Temporal, is now the weak link.
- `task_run_cancel_requested` with `cancel_source = "pr_closed"` should rise slightly relative to the pre-merge baseline, since deliveries that previously failed on a transient error now land on a retry. It should not fall.
- Join on `properties.run_id`: every `task_run_cancel_requested` with `cancel_source = "pr_closed"` gets a `task_run_cancelled_terminal` with the same `cancel_source`. The unmatched share should be zero once the 15 minute retry budget elapses; a persistent gap means the retries are exhausting.
- In Loki, `namespace="temporal-worker-tasks-agent"` and the celery workers: `github_pr_webhook_wizard_cancel_enqueue_failed` should be absent, and repeated `Cancellation for task run ... could not be delivered` means `cancel_task_run` kept returning `unavailable` across all five attempts.
- The webhook response is unchanged at 200 on every path, so GitHub delivery success rates for `pull_request` are a flat control: a drop there would mean the handler started failing rather than deferring.

Fixes #74316


fercgomes:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/PostHog/posthog/74044?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#74044** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74044?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74043** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74043?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74042** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74042?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

products/tasks/backend/webhooks.py
```diff
@@ -301,21 +306,17 @@ def _cancel_wizard_run_on_close(task_run: TaskRun) -> None:
...
-            cancel_task_run(
-                task_run.id,
-                task_run.task_id,
-                task_run.team_id,
-                reason="Setup pull request was closed",
-                source="pr_closed",
+            cancel_wizard_run_on_pr_close_task.apply_async(
+                args=[str(task_run.id), str(task_run.task_id), task_run.team_id, PR_CLOSED_CANCEL_REASON],
             )
         except Exception:
```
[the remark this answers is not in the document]
fercgomes: The enqueue runs in `transaction.on_commit`, after the handler has already answered 2xx, and GitHub does not automatically redeliver a delivery it received a 2xx for. Raising there loses the cancellation exactly the same way and adds a 500 on top, so the swallow stays.

What changed is that the swallowed window is now a broker publish rather than the whole Temporal round trip, and the loss is counted: `posthog_tasks_cancel_enqueue_failed_total{kind="pr_closed"}`, asserted in `test_pr_closed_cancel_enqueue_failure_keeps_webhook_successful`. Making the enqueue itself durable needs an outbox row plus a sweep, which is a separate change.

products/tasks/backend/facade/tasks.py
```diff
@@ -11,6 +11,40 @@
...
+
+    countdown = CANCEL_WIZARD_RUN_RETRY_DELAY_SECONDS * (2**self.request.retries)
+    try:
+        outcome, _ = cancel_task_run(run_id, task_id, team_id, reason=reason, source="pr_closed")
+    except Exception as error:
+        raise self.retry(exc=error, countdown=countdown)
+
+    # cancel_task_run reports an undeliverable cancellation as an outcome instead of raising,
+    # so the retry has to be driven off the return value as well as off exceptions.
+    if outcome == "unavailable":
```
MattBro, line 42: When this keeps coming back `unavailable` you give up after 6 tries, but `cancel_task_run` returns early on that path before `_schedule_cancel_verification` runs, so nothing was ever armed to catch the run. Temporal down for ~15min and an IN_PROGRESS run ends up with nothing watching it (the stale sweeper only takes QUEUED, the IN_PROGRESS one is in #74040).

Could we arm `verify_task_run_cancelled_task` on exhaustion instead of just raising? `force_terminalize_cancelled_run` doesn't need the run to be CANCELLED, so it still works even though the signal never landed.

products/tasks/backend/tests/test_webhooks.py
```diff
@@ -365,14 +366,57 @@ def test_pr_closed_cancel_failure_keeps_webhook_successful(self, _mock_capture,
...
+        # cancel_task_run reports an undeliverable cancellation as an outcome rather than
+        # raising, so the retry has to be driven off the return value.
+        run_id, task_id = str(self.task_run.id), str(self.task.id)
+
+        with (
+            patch(
+                "products.tasks.backend.facade.cancellation.cancel_task_run",
+                return_value=(outcome, None),
+            ),
+            patch.object(cancel_wizard_run_on_pr_close_task, "retry", side_effect=Retry()) as mock_retry,
```
MattBro, line 395: Both of these stub `retry`, so the countdown is only ever asserted at `retries == 0` and `max_retries=5` never gets hit. Drop the `2**` and these still pass.

Unstubbing won't help (calling the task directly sets `called_directly=True` and celery short-circuits before reading either), but `push_request(retries=N, called_directly=False, is_eager=True)` then `task.run(...)` does. There's an example in `test_ai_observability_usage_report.py`.

MattBro (approved):

fercgomes:

PR #74045, opened 2026-07-27
fix(tasks): fail the run when the initial prompt cannot be delivered

fercgomes:
## Why

`forward_pending_user_message` delivers the initial prompt to the sandbox agent once its server is ready. On a non-retryable delivery result (success False, turn_in_flight False, retryable False) it currently clears the pending message from run state, logs, and returns normally, so the workflow keeps executing with an agent that never received its prompt. The run then idles until the inactivity timeout and finishes with no visible error even though nothing was ever asked of the agent. A delivery failure that cannot be retried is terminal for the run and should surface as one.

## What

- On the non-retryable path, raise a non-retryable `ApplicationError` carrying the delivery error text instead of clearing the pending message and returning
- Keep the pending message in state on that path, matching what the retryable path already does, so the failed run retains what it tried to deliver
- Keep the Slack delivery-failure relay and the delivery-failed metric on that path
- turn_in_flight and retryable semantics are unchanged
- Update the activity unit tests for the new behavior

## Notes for reviewers

- No workflow change is needed: both `process-task` and `execute-sandbox` call this activity with `maximum_attempts=1` and no try/except, so the raise flows into their existing failure handlers, which persist status `failed` with the cause message
- The error text goes through `user_facing_agent_error`, so the transient content-block RPC error keeps its friendlier wording
- Tests run on this branch: `pytest products/tasks/backend/temporal/process_task/tests/test_forward_pending_message.py`, 17 passed. Nothing else was run, and no manual sandbox run was exercised

## Validating after merge

- `posthog_tasks_task_run_failed_total{temporal_activity_type="forward_pending_user_message"}` goes from flat zero to non-zero (both regions). Every increment is a run that previously kept executing with an agent that never received its prompt, so a small non-zero rate is the change working, not a regression. The companion labels on those series are `error_type="ActivityError"`, `cause_error_type="ApplicationError"`
- That counter should track `posthog_tasks_followup_delivery_failed_total` (both `retryable` values) close to 1:1 for the initial-prompt path, since the same two branches emit both. A sustained gap in either direction means a delivery failure is taking a path that does not raise
- Regression signal: `posthog_tasks_runs_terminal_1h{status="failed"}` stepping up by more than the delivery-failure rate above, or `posthog_tasks_task_run_failed_total{temporal_activity_type="forward_pending_user_message"}` climbing while the follow-up delivery counter stays flat. Either says the raise is firing on deliveries that actually landed
- Loki, `namespace="temporal-worker-tasks-agent"`, `origin_product="onboarding"`: `forward_pending_message_non_retryable_failure` and `forward_pending_message_retryable_failure` lines should now be followed by the workflow failure handler for the same run_id rather than by the run continuing to the inactivity timeout
- Product analytics (project 2): server `task_run_failed` events whose `properties.error_message` contains `forward pending message failed`, joinable on `properties.run_id`. Before this change those runs surfaced as `task_run_completed` after idling, so a matching drop in `task_run_completed` for runs with no agent output is the other half of the same move

Fixes #74316


fercgomes:
* **#74046** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74046?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74045** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74045?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74045?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tatoalo (commented):
the new exception changes all of our initial prompt from continuing after a rejected initial prompt to failing the run altogether, can we gate this on wizard only? retrial in other cases are a good fallback we need

fercgomes:

PR #74046, opened 2026-07-27
fix(tasks): re-forward prompts delivered while a turn is in flight

fercgomes:
## Why

`forward_pending_user_message` treats a turn-in-flight result as delivered and
clears the pending message from run state. That result comes from a single
source: `send_agent_command` catching `requests.ReadTimeout`, meaning the
request body was written but no response came back. It is evidence that the
request reached the sandbox, not that the agent server ingested the prompt, so
clearing the state discards the only copy of a message whose delivery is
unknown. A run in that position holds a live sandbox with an agent that may
never have been asked anything, and the state it would need to retry is gone.

`send_user_message` already carries `message_id` as an idempotency key and the
agent server answers a redelivery of an accepted id with `result.duplicate`
instead of starting a second turn. That makes re-forwarding the same id a safe
probe: an acknowledged response settles the question either way, and only the
undelivered case produces a new turn.

## What

- On turn-in-flight the activity keeps `pending_user_message` and its
  `pending_user_message_id` in state, sets a `pending_user_message_delivery_unconfirmed`
  marker, and returns `FORWARD_OUTCOME_TURN_IN_FLIGHT`; an acknowledged
  delivery returns `FORWARD_OUTCOME_DELIVERED` and clears the marker with the
  rest of the pending-message keys
- A `duplicate` ack no longer enqueues a Slack reply relay: the turn it refers
  to belongs to an earlier copy and carries no reply for this call
- Both workflows start a background `_confirm_pending_user_message_delivery`
  loop on that outcome: up to `PENDING_MESSAGE_CONFIRM_MAX_ATTEMPTS` (5)
  re-forwards spaced `PENDING_MESSAGE_CONFIRM_DELAY` (120s) apart, each skipped
  once the run is no longer active, stopping as soon as an attempt returns
  anything other than turn-in-flight
- Agent activity outranks the forward outcome. Both workflows now keep a
  monotonic `_agent_activity_count`, bumped by `heartbeat(agent_active=True)`,
  and the loop snapshots it on entry: an increment stops the loop before the
  next re-forward and before the exhaustion failure. A turn in flight is the
  ordinary result for a first turn longer than the activity's 90s synchronous
  timeout, so an unacknowledged forward alone must never terminalize a run
  whose agent is visibly streaming
- That counter is workflow-wide, so activity only confirms the prompt while the
  prompt is the only thing that was sent. Both workflows also keep a monotonic
  `_followup_dispatch_count`, bumped where a follow-up is handed to
  `send_followup_to_sandbox`, and the loop snapshots it alongside the activity
  count: a follow-up dispatched during the window drives the agent on its own,
  so its heartbeats no longer stop the re-forwards. They do still bar the
  exhaustion failure, which never fires while the agent is streaming
- Exhausting the attempts with no agent activity, or a re-forward raising,
  terminalizes the run `failed` with `error_type=pending_message_delivery_failed`
  and an error_message naming the cause, matching the parent branch's failure
  contract
- `pending_user_message_delivery_unconfirmed` joins the pending-message key
  group cleared by `create_resume_snapshot` and `TaskRun.prepare_for_cloud_handoff`
- Unit tests for the new activity outcomes, and for the confirmation loop and
  its `run()` wiring in both workflows

## Notes for reviewers

- Temporal versioning: the loop's timer and repeat activity are new workflow
  commands, gated behind `workflow.patched(_PATCH_ID_CONFIRM_PENDING_MESSAGE_DELIVERY)`
  in each workflow (distinct ids, since the two record separate histories).
  The gate sits behind an outcome check so pre-rollout histories, whose
  recorded activity result decodes to `None`, never reach `patched` at all
- The activity's declared return type is `str | None` on purpose:
  `value_to_type(str, None)` raises, so a bare `str` annotation would break
  replay of any history that recorded the old `null` result. Verified against
  the installed `temporalio.converter`
- The loop terminalizes by setting `_completion_status`/`_completion_error` and
  `_task_completed` rather than raising, mirroring the existing follow-up
  delivery failure path. Raising from an unawaited background task would sit
  unobserved until the inactivity timeout
- Worst case for a healthy run: the prompt was queued and consumed, the agent is
  working, and the first `agent_active` heartbeat inside the 120s window ends the
  loop with no extra request at all. If the agent is silent for that window the
  loop spends one re-forward, which the agent server drops as a duplicate
- The exhaustion failure is therefore reachable only when five consecutive
  re-forwards go unacknowledged and no heartbeat reports the agent streaming for
  the whole ~10 minute window, which is a run with nothing running in it
- Known gap left as is: a `continue_as_new` during the confirmation window
  cancels the loop, and the continuation skips forwarding because
  `resumed_sandbox` is set. History length cannot reach the continue-as-new
  threshold within the 10-minute window on a fresh run
- Tests run on this branch: `pytest
  products/tasks/backend/temporal/process_task/tests/test_forward_pending_message.py
  products/tasks/backend/temporal/process_task/tests/test_workflow.py
  products/tasks/backend/temporal/execute_sandbox/tests/test_execute_sandbox_workflow.py`,
  165 passed, 4 skipped, plus one failure that also fails on master
  (`test_get_sandbox_for_repository_falls_back_to_fresh_sandbox_when_resume_injection_fails`,
  an unrelated image_source assertion). No sandbox run was exercised manually

## Validating after merge

- Loki, `namespace="temporal-worker-tasks-agent"`: `forward_pending_message_delivery_unconfirmed` is the entry rate into the new loop (one line per turn-in-flight forward). Expect it non-zero and roughly the share of first turns that outlive the activity's 90s timeout. Zero everywhere means the patch gate is not being reached and nothing below can be read
- Of the runs that enter it, `pending_user_message_delivery_confirmed` plus `pending_user_message_delivery_confirmed_by_agent_activity` (and the `execute_sandbox_`-prefixed twins) should account for nearly all of them, with `pending_user_message_delivery_still_unconfirmed` appearing at most a few times per run before one of the two lands
- The terminalization goes through the completion path rather than a raise, so it does not reach `posthog_tasks_task_run_failed_total`; the signal is the product analytics (project 2) server event `task_run_failed` with `properties.error_type = "pending_message_delivery_failed"`, joinable on `properties.run_id`. Expect near zero. Sustained non-zero means the loop is failing runs whose agent was working, and the attempt budget or the 120s spacing is too tight
- `pending_user_message_delivery_unconfirmed_agent_active` marks the case where a concurrently dispatched follow-up could account for the agent activity: the prompt stays unconfirmed and the run is left alone. Expect it rare. A rate comparable to the confirmed lines would mean follow-ups routinely land inside the confirmation window and the loop is mostly declining to conclude anything
- Regression signals: `posthog_tasks_runs_terminal_1h{status="failed"}` stepping up after rollout beyond the `pending_message_delivery_failed` event count, and `posthog_tasks_followup_delivery_failed_total` rising, which would mean the re-forwards are being treated as new deliveries instead of dropped as duplicates on the same `message_id`
- Expected in the other direction: `posthog_tasks_oldest_open_run_age_seconds{status="in_progress"}` loses its undelivered-prompt tail, since those runs now terminalize inside the ~10 minute confirmation window instead of idling to the inactivity timeout

Fixes #74316


fercgomes:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/PostHog/posthog/74046?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#74046** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74045** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74045?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

products/tasks/backend/temporal/process_task/workflow.py
```diff
@@ -1442,14 +1471,94 @@ async def _await_agent_server_ready(self, sandbox_output: GetSandboxForRepositor
...
+            f"the agent never acknowledged it across {PENDING_MESSAGE_CONFIRM_MAX_ATTEMPTS} re-deliveries"
+        )
+
+    def _agent_is_working_on_the_prompt(self, agent_activity_before: int, attempt: int) -> bool:
+        """True once a heartbeat has reported the agent streaming since the loop began.
+
+        Nothing else can be producing agent output at this point: the initial
+        prompt is the only thing sent to a fresh sandbox, so an increment means
+        it was ingested regardless of what the forward's ack did.
+        """
```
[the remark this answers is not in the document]
fercgomes: Valid, fixed. Both workflows now keep a monotonic `_followup_dispatch_count`, bumped where a follow-up is handed to `send_followup_to_sandbox`, and the confirmation loop snapshots it next to `_agent_activity_count`. Activity only confirms the prompt while the follow-up count has not moved, so a concurrently dispatched follow-up no longer stops the re-forwards; they continue and stay idempotent on the same `message_id`.

Activity that a follow-up can explain still bars the exhaustion failure, since a streaming agent is not a run to terminalize. New tests: `test_confirm_pending_message_ignores_activity_a_concurrent_followup_can_explain` and `test_ignores_activity_a_concurrent_followup_can_explain`.

fercgomes:
Addressed the P1 on the shared activity counter. Both workflows now keep a monotonic `_followup_dispatch_count`, bumped where a follow-up is handed to `send_followup_to_sandbox`, and the confirmation loop snapshots it next to `_agent_activity_count`. Agent activity only confirms the prompt while the follow-up count has not moved, so a concurrently dispatched follow-up no longer stops the re-forwards; they continue on the same `message_id`, which the agent server drops as a duplicate.

Activity a follow-up could explain still bars the exhaustion failure, logged as `pending_user_message_delivery_unconfirmed_agent_active`, since a streaming agent is not a run to terminalize. Also fixed the `test_create_resume_snapshot` remove_keys assertion that the new state key broke in CI.

products/tasks/backend/temporal/execute_sandbox/workflow.py
```diff
@@ -1223,14 +1256,118 @@ async def _start_agent_server(self, sandbox_output: GetSandboxForRepositoryOutpu
...
-        await workflow.execute_activity(
+    async def _forward_pending_user_message(self) -> str | None:
+        return await workflow.execute_activity(
             forward_pending_user_message,
             self.context.run_id,
             start_to_close_timeout=timedelta(seconds=PENDING_MESSAGE_FORWARD_TIMEOUT_SECONDS),
             retry_policy=RetryPolicy(maximum_attempts=1),
         )
 
+    async def _confirm_pending_user_message_delivery(self) -> None:
```
tatoalo, line 1267: we keep delivered and in-flight ID sets in process memory. If we have restarts after accepting the first prompt but before this workflow observes the response, this retry executes the prompt again and can duplicate agent side effects. which creates a lot of regressions

tatoalo (changes requested):
do you have evidence of actual dropped prompts? we can persist that the message_id was accepted, then retry safely only when acceptance is unknown?

products/tasks/backend/temporal/execute_sandbox/workflow.py
fercgomes: The sets are Temporal workflow state, not process memory, and they are not on this path.

`_acked_ids` (workflow.py:324) and `_in_flight_followup_ack_ids` (workflow.py:331) are instance attributes of `ExecuteSandboxWorkflow`, which is `@temporalio.workflow.defn` (workflow.py:258). Temporal rebuilds them by replaying history on whichever worker picks the workflow up, and this workflow never calls `continue_as_new`, so there is no boundary that drops them. They also only dedupe parent-to-child signals (`_is_duplicate_signal`, workflow.py:964); nothing on the pending prompt path reads or writes them, and this PR does not touch either set.

The prompt path's idempotency key is `pending_user_message_id`, written to Postgres under a row lock before the send (forward_pending_message.py:103-110, models.py:1630-1636) and generated only when absent. Every re-forward reuses it (forward_pending_message.py:146). A turn in flight keeps the message and its id in state (forward_pending_message.py:211-220); an ack removes them (forward_pending_message.py:228-237). So the restart you describe replays with the same id: the agent server drops a redelivery it already accepted and returns `duplicate: true`, which we read at forward_pending_message.py:286-290 and settle on. A workflow that restarts after a clean delivery finds no pending message at all and returns DELIVERED without sending (forward_pending_message.py:113-114).

This is the same mechanism `send_followup_to_sandbox` already relies on (send_followup_to_sandbox.py:128-132 and 289-291, agent_command.py:290-294), and the execute_sandbox follow-up path carries it too via `message_id or str(workflow.uuid4())` (workflow.py:1577). Covered by `test_re_forward_reuses_message_id_and_settles_state_on_duplicate_ack` and `test_connection_error_retries_with_longer_timeout` (test_forward_pending_message.py:192 and :242).

The one thing that would break idempotency is the agent server losing its accepted-id table, which means its process died. No path restarts it inside a run, and if it goes we mark `_sandbox_gone` and the confirm loop returns immediately (workflow.py:1293-1294).

fercgomes:

PR #74049, opened 2026-07-27
feat(tasks): tag task_run_started with origin_product

fercgomes:
## Why

`task_run_failed` carries `origin_product` from the workflow context, but `task_run_started` (fired
right after the IN_PROGRESS transition) does not. The two events therefore cannot be joined on the
same dimension, so any per-origin start/failure pairing has to fall back on a separate lookup of the
task row. The value is already resolved on `TaskProcessingContext` at workflow start, so the started
event can carry it directly.

## What

- add `origin_product` to the `task_run_started` property bag in `ProcessTaskWorkflow._provision_and_start_agent`, sourced from `self.context.origin_product` exactly as the failure path does
- add an `origin_product` keyword to the `_build_context` test helper
- unit test asserting both `task_run_started` and `task_run_failed` report the same origin

## Notes for reviewers

- No Temporal patch gate. Per `.claude/rules/temporal-workflow-versioning.md`, activity input
  dataclasses are safe to edit; this only adds a key to the `properties` dict inside the existing
  `TrackWorkflowEventInput`, and adds no command to the workflow's sequence, so replaying histories
  see an unchanged command list.
- `origin_product` is `str | None` on the context, so runs without an origin send an explicit null
  rather than omitting the key, matching `task_run_failed`.
- Test plan: pytest products/tasks/backend/temporal/process_task/tests/test_workflow.py -k "origin_product or provisioning_fails" (2 passed locally).

## Validating after merge

- This changes no run behavior; it only makes an existing dimension observable at start time. What
  proves it works is the property's presence, not a shift in any run outcome.
- Signal: the server event `task_run_started` (project 2) carries `properties.origin_product`. Within
  one deploy the share of `task_run_started` events emitted by `ProcessTaskWorkflow` that carry the
  key should go from 0% to 100%, with the value `onboarding` for wizard cloud runs and null for runs
  whose task has no origin.
- The pairing this exists for: for a given `properties.run_id`, `task_run_started` and
  `task_run_failed` must report the same `origin_product`. Any `run_id` where the two disagree is a
  defect in how the context is resolved, not in the events.
- Known gap, so the overall share stays under 100%: `execute_sandbox/workflow.py` emits its own
  `task_run_started` without this key. Scope the check to runs that also produce a `task_run_failed`
  or `task_run_completed` from `ProcessTaskWorkflow` before reading the number as short.
- Regression: `origin_product` null on `task_run_started` while the same run's `task_run_failed`
  reports a non-null value. Nothing else about the run changes, so there is no other failure mode to
  watch for.

Fixes #74316

### DIFF
--- a/products/tasks/backend/temporal/process_task/tests/test_workflow.py
+++ b/products/tasks/backend/temporal/process_task/tests/test_workflow.py
@@ -77,6 +77,7 @@ def _build_context(
     environment: str | None = None,
     use_modal_vm_sandbox: bool = False,
     custom_image_name: str | None = None,
+    origin_product: str | None = None,
 ) -> TaskProcessingContext:
     return TaskProcessingContext(
         task_id="task-id",
@@ -87,6 +88,7 @@ def _build_context(
         github_integration_id=github_integration_id,
         repository=repository,
         distinct_id="distinct-id",
+        origin_product=origin_product,
         environment=environment,
         create_pr=True,
         state=state or {},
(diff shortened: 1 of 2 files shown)


fercgomes:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/PostHog/posthog/74049?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#74050** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74050?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74049** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74049?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74049?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74048** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

tatoalo (commented):
workflow.py also emits task_run_started, and its context already carries origin_product, right now attribution coverage depends on which workflow implementation runs the task. can we have the property to both producers and cover both paths?

fercgomes:

PR #74050, opened 2026-07-27
feat(wizard): validate repository detectability before creating a cloud run

fercgomes:
## Why

`create_wizard_cloud_run` accepts any `owner/repo` string that passes the endpoint's two-segment
format check. Two of the run's preconditions are only tested inside the sandbox, minutes after
kickoff: whether the team's GitHub installation can reach the repository at all (the clone fails with
"remote: Repository not found"), and whether the repository holds any project manifest the wizard's
framework detection keys off (the wizard exits with "Could not auto-detect your framework for this
project"). Both are answerable from two GitHub reads at kickoff, where the answer can still be turned
into an error message the caller can act on.

## What

- new `products/tasks/backend/logic/wizard_preflight.py`: reads `GET /repos/{owner}/{repo}` and, when
  that succeeds, the recursive git tree of the ref the sandbox will check out, and reports an access
  tri-state plus a tri-state `framework_detectable`. Never raises
- new `resolve_team_github_integration` in `products/tasks/backend/models.py`, extracted from
  `Task._build_task` and shared with the pre-flight, so the check and the run it predicts cannot
  drift onto different credentials. Public sandbox repos, which `_build_task` exempts from needing an
  integration at all, skip the pre-flight rather than being read through one
- `WIZARD_PROJECT_MANIFESTS` mirrors `PROJECT_MANIFESTS` in the wizard's
  `src/lib/detection/agentic.ts`, matched on the basename of every blob in the tree rather than only
  at the root, because the wizard globs the whole checkout for manifests
- new facade entry point `validate_wizard_cloud_run_target`, raising
  `WizardRepositoryInaccessibleError` / `WizardFrameworkUndetectableError` on a confirmed negative.
  `create_wizard_cloud_run` calls it, so no caller can start a run the pre-flight would reject
- `POST /api/wizard/cloud_run` calls it directly, above the daily attempt reservation, and maps each
  rejection to a 400 carrying a stable `code` (`repository_inaccessible`, `framework_undetectable`).
  The kickoff counter reports them under their own outcome label instead of folding them into
  `invalid`
- the partner provisioning endpoint maps the same two errors to 400s under the same codes, instead of
  letting them fall through to its generic 500 handler
- `WIZARD_CLOUD_RUN_DAILY_PREFLIGHT_CAP = 40`, an atomic per-user daily reservation taken immediately
  above the pre-flight and never released. The attempt cap deliberately does not charge a rejection,
  so without this a caller looping on rejected repositories would make unmetered authenticated GitHub
  reads. Sized above the attempt cap so it only binds on that shape
- `_release_cloud_run_attempt` gives the attempt back when the facade's repeated pre-flight rejects a
  target the endpoint's first read accepted. `validate_wizard_cloud_run_target` is the first statement
  of `create_wizard_cloud_run`, so a rejection there provably created nothing and booted no sandbox
- facade, endpoint, and partner-endpoint tests

## Notes for reviewers

- The blocking rule is the risk surface, so it is deliberately narrow: only a 404 on the repository
  read blocks, and only a tree read that returned 200 with `truncated: false` and zero manifest
  matches blocks. Rate limits, timeouts, non-200s, malformed payloads, truncated trees, an
  unresolvable ref, a public sandbox repo, and a team with no usable GitHub integration all create
  the run exactly as today. The parameterized facade test enumerates those cases; a change that
  starts blocking on any of them fails it.
- Adding a manifest to `WIZARD_PROJECT_MANIFESTS` is always safe; omitting one the wizard supports is
  what would reject a healthy repository, hence the pointer to the wizard file it mirrors.
- The attempt reservation is the hard ceiling on sandbox boots, so the target check sits above it: a
  repository typo boots no sandbox and now costs none of the day's budget. `create_wizard_cloud_run`
  re-checks after the reservation, which costs two more GitHub reads on the accepted path only, next
  to a sandbox that runs for up to 45 minutes. That second check can still reject, which is what the
  release path exists for, so the attempt counter stays a count of sandbox boots and nothing else.
- The two caps answer different questions and are deliberately separate: the attempt cap bounds
  sandbox boots and must not be spent by a request that boots none, the pre-flight cap bounds the
  GitHub reads a request makes before it can reach creation. Both are `cache.add` + `cache.incr`, so
  neither can be raced the way the read-then-create DB throttles can.
- Two GitHub reads on the rejected kickoff path, both with a 5s timeout, through the existing egress
  transport and the installation's shared budget.
- The repository and ref are interpolated into request paths, so both are containment-checked
  (`_is_safe_path`) before any read; the endpoint serializer only validates the two-segment shape.
- Test plan: `pytest products/tasks/backend/tests/test_facade.py products/tasks/backend/tests/test_models.py`
  (217 passed), `pytest posthog/api/wizard/test/test_http.py` (34 passed),
  `pytest ee/api/agentic_provisioning/test/test_wizard_resource_actions.py` (16 passed).

## Validating after merge

- The two new labels on `posthog_wizard_cloud_run_requests_total` are the primary signal:
  `{outcome="repository_inaccessible"}` and `{outcome="framework_undetectable"}` go from absent to
  non-zero, and `{outcome="created"}` falls by about their sum. Both regions.
- The point of the change is that those rejections used to be sandbox boots that failed minutes
  later, so `posthog_tasks_runs_in_status{status="failed",origin_product="onboarding"}` should fall
  by a comparable amount over the same window. If `created` drops but failed onboarding runs do not,
  the pre-flight is rejecting repositories that would have worked.
- The regression this can cause is exactly that false rejection. Sustained ratio of the two new
  outcomes to `created` above roughly 10%, or client `onboarding cloud run queued` no longer being
  followed by `onboarding cloud run completed` for repositories that used to work, is the line to
  investigate. `WIZARD_PROJECT_MANIFESTS` drifting behind the wizard's `PROJECT_MANIFESTS` is the
  most likely cause.
- `{outcome="throttled"}` should not rise. The pre-flight cap is 40/user/day against an attempt cap
  of 15, so throttling should keep coming from the attempt cap and the DRF run-counting throttles. A
  rise here means the pre-flight cap is binding on ordinary use and is sized wrong.
- Kickoff latency now includes two GitHub reads at a 5s timeout each. p95 of `POST
  /api/wizard/cloud_run` should move by well under a second; a jump into the 5s to 10s range means
  the reads are timing out against the installation's shared budget rather than answering.

Fixes #74316

### DIFF
--- a/posthog/api/wizard/http.py
+++ b/posthog/api/wizard/http.py
@@ -58,14 +58,50 @@ OPENAI_SUPPORTED_MODELS = {"o4-mini", "gpt-5-mini", "gpt-5-nano", "gpt-5"}
 # Absolute ceiling on sandbox boots per user per day, reserved atomically right before run
 # creation. The DB-counted throttles above the view are read-then-create and can be raced by
 # parallel requests; this cache.incr cannot, so it is the hard bound a start-cancel or crash
-# loop lands on. Only requests that reach creation consume it.
+# loop lands on. Only requests that boot a sandbox consume it: a reservation whose run creation
+# then rejects is released again, so it stays a count of sandbox boots and nothing else.
 WIZARD_CLOUD_RUN_DAILY_ATTEMPT_CAP = 15
+
+# Ceiling on target pre-flights per user per day. The pre-flight spends GitHub reads from the
+# installation's shared budget before anything is created, and it deliberately costs no sandbox
+# boot, so the attempt cap above does not bound it: a caller looping on repositories that fail the
+# pre-flight would otherwise make unmetered authenticated GitHub requests. Sized well above the
+# attempt cap so it only ever binds on that looping shape, never on ordinary retrying.
+WIZARD_CLOUD_RUN_DAILY_PREFLIGHT_CAP = 40
 
 WIZARD_CLOUD_RUN_REQUESTS_TOTAL = Counter(
     "posthog_wizard_cloud_run_requests_total",
-    "Cloud-run wizard kickoff requests, by outcome (created/unavailable/invalid/permission_denied/throttled)",
+    "Cloud-run wizard kickoff requests, by outcome (created/unavailable/invalid/permission_denied/throttled/"
+    "repository_inaccessible/framework_undetectable)",
     labelnames=["outcome"],
 )
+
+# Pre-flight rejections are reported under their own outcome rather than folded into "invalid", so a
+# pre-flight that starts rejecting healthy repositories is visible without reading request logs.
+CLOUD_RUN_PREFLIGHT_OUTCOMES = frozenset({"repository_inaccessible", "framework_undetectable"})
+
+
+def _cloud_run_validation_outcome(error: exceptions.ValidationError) -> str:
+    codes = error.get_codes()
+    if isinstance(codes, list) and len(codes) == 1 and codes[0] in CLOUD_RUN_PREFLIGHT_OUTCOMES:
+        return cast(str, codes[0])
+    return "invalid"
+
+
+def _preflight_validation_error(error: Exception) -> exceptions.ValidationError:
+    """Turn a facade pre-flight rejection into a 400 carrying its own stable code.
+
+    Clients act on the code, not on the prose: each of these rejections has a specific remedy (fix
+    the repository name, grant the GitHub app access, point the wizard at another repository) that
+    a generic input-validation code would not tell them apart from.
+    """
+    code = (
+        "repository_inaccessible"
+        if isinstance(error, tasks_facade.WizardRepositoryInaccessibleError)
+        else "framework_undetectable"
+    )
+    return exceptions.ValidationError(str(error), code=code)
+
 
 # Supported Gemini models
 GEMINI_SUPPORTED_MODELS = {
@@ -466,8 +502,8 @@ class SetupWizardViewSet(viewsets.ViewSet):
         except exceptions.PermissionDenied:
             WIZARD_CLOUD_RUN_REQUESTS_TOTAL.labels(outcome="permission_denied").inc()
             raise
-        except exceptions.ValidationError:
-            WIZARD_CLOUD_RUN_REQUESTS_TOTAL.labels(outcome="invalid").inc()
+        except exceptions.ValidationError as e:
+            WIZARD_CLOUD_RUN_REQUESTS_TOTAL.labels(outcome=_cloud_run_validation_outcome(e)).inc()
             raise
         except exceptions.Throttled:
             # The atomic attempt reservation inside _cloud_run raises after check_throttles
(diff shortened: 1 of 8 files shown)


fercgomes:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/PostHog/posthog/74050?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#74050** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74049** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74048** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

posthog/api/wizard/http.py
```diff
@@ -524,6 +566,10 @@ def _cloud_run(self, request: Request) -> Response:
                 repository=repository,
                 branch=branch,
             )
+        except (tasks_facade.WizardRepositoryInaccessibleError, tasks_facade.WizardFrameworkUndetectableError) as e:
```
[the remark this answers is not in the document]
fercgomes: Correct, fixed. `_release_cloud_run_attempt` now gives the reservation back on that path. `create_wizard_cloud_run` calls `validate_wizard_cloud_run_target` as its first statement, before it creates any row, so a rejection there provably booted no sandbox and the reservation can be released rather than charged. New test `test_creation_side_preflight_rejection_gives_the_attempt_back` pins it with the cap at 1: the request whose second pre-flight rejects is followed by one that still succeeds.

posthog/api/wizard/http.py
```diff
@@ -515,6 +546,17 @@ def _cloud_run(self, request: Request) -> Response:
         if project.id not in visible_project_ids:
             raise exceptions.PermissionDenied("You don't have access to this project.")
 
+        # Above the reservation on purpose: a rejection here starts no sandbox, so it must not
+        # spend one of the day's sandbox boots.
+        try:
+            tasks_facade.validate_wizard_cloud_run_target(
```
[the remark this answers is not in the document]
fercgomes: Valid, fixed. The pre-flight now carries its own atomic daily reservation, `WIZARD_CLOUD_RUN_DAILY_PREFLIGHT_CAP = 40`, taken by `_reserve_cloud_run_preflight` immediately above the call and never released, so every request that reaches a GitHub read is metered whether or not it goes on to create a run. It is separate from the attempt cap on purpose: that one bounds sandbox boots and must not be spent by a rejection, this one bounds the reads. Sized above the attempt cap so it only binds on the looping shape you describe. New test `test_preflight_reads_are_capped_even_though_they_spend_no_attempt` drives the cap to 2 and asserts the third rejected request gets a 429 without reaching the pre-flight.

products/tasks/backend/logic/wizard_preflight.py
```diff
@@ -0,0 +1,205 @@
...
+# reads below at a different GitHub endpoint.
+_SAFE_REPOSITORY_RE = re.compile(r"^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$")
+# Same containment rule for the ref, which is also interpolated into a path, with ``/`` allowed for
+# namespaced branches ("feature/x").
+_SAFE_REF_RE = re.compile(r"^[A-Za-z0-9._/-]+$")
+
+# Basenames of every project manifest the wizard identifies a framework by. Mirrors PROJECT_MANIFESTS
+# in the wizard's src/lib/detection/agentic.ts; keep the two in sync, and when in doubt add the entry
+# here, since a manifest missing from this set is what turns a healthy repository into a rejection.
+WIZARD_PROJECT_MANIFESTS = frozenset(
```
MattBro, line 43: This mirrors the wizard's `PROJECT_MANIFESTS`, but the "couldn't auto-detect your framework" abort actually comes from `detectFramework()`, which walks `FRAMEWORK_REGISTRY`. There's no Go/Rust/Elixir/Dart/JVM target in that registry, so a repo with just a `go.mod` passes here and still dies in the sandbox. Worth keying off the registry instead, otherwise the rejection count won't line up with the failures you're trying to cut.

posthog/api/wizard/http.py
```diff
@@ -515,15 +598,36 @@ def _cloud_run(self, request: Request) -> Response:
...
                 repository=repository,
                 branch=branch,
             )
+        except (tasks_facade.WizardRepositoryInaccessibleError, tasks_facade.WizardFrameworkUndetectableError) as e:
+            # The facade re-runs the pre-flight, so GitHub answering differently between the two
+            # reads lands here rather than as a 500. It runs before the facade creates anything, so
+            # this request booted no sandbox and the attempt it reserved goes back.
+            self._release_cloud_run_attempt(user_id)
+            raise _preflight_validation_error(e)
         except ValueError as e:
```
MattBro, line 631: This branch doesn't release the attempt, unlike the one right above it. `_build_task` raises here before creating anything (team has no GitHub integration), so the user burns one of their 15 on a run that never booted, which the new comment on the cap says can't happen. Reachable from the UI too, since the frontend only checks `kind === 'github'` and doesn't filter the broken-installation states.

products/tasks/backend/models.py
```diff
@@ -60,6 +60,22 @@ def resolve_schema(schema: type[BaseModel] | dict) -> dict:
     return schema.model_json_schema()
 
 
+def resolve_team_github_integration(team: Team) -> Integration | None:
```
MattBro, line 63: We already have a `resolve_team_github_integration` in `logic/repo_selection/agent.py` with different rules (takes `team_id`, has user/org fallbacks), and it's exported through the facade. Two with the same name is going to bite someone, rename this one?

MattBro (approved):

fercgomes:

PR #74057, opened 2026-07-27
fix(replay-vision): state what each config version changed

arnohillen:
## Problem

The scanner configuration tab's config-versions history only rendered the prompt plus a handful of type-specific config fields (tags, scale, length, allow_inconclusive).
Versions bump on any change to `scanner_type`, `scanner_config`, `query`, `sampling_rate`, `sampling_mode`, `provider`, `model` or `emits_signals`, so anything outside that small set produced a new version that looked identical to the one before it.

I hit this on a real scanner where v4 only turned on "Multiple tags per session".
The history rendered v4 as a copy of v3 and, worse, the header said "prompt unchanged from v3" — which reads as "nothing changed" when something clearly had.

Two version-tracked fields were also missing from the run snapshots the history is rebuilt from, so a filter- or sampling-only bump was not just unshown, it was unrecoverable.

## Changes

Every version entry now states what it changed since the previous version, field by field, and lists the full version-tracked config it ran with.

- **Change summary per version.** "Changed since vN" with each differing field as `before → after`. A prompt change shows "edited" plus a button that opens the side-by-side diff. It compares against the previous version *present* in the history, not `version - 1`, since versions that scanned nothing leave gaps — the label names the version so it can't mislead.
- **All eight version-tracked fields in the readout**, including the recording filters as pills, with changed fields emphasized.
- **Snapshots now record `query`, `sampling_rate` and `sampling_mode`** (`ScannerSnapshot`), so a filter- or sampling-only bump is explainable at all. These never reach the LLM.
- **Old versions say "Not recorded", never "unchanged".** Observations written before this lack those three fields, so pre-existing versions read `Not recorded` and the change block adds a line naming what couldn't be compared. That's the one case the page can't be specific about, and it says so rather than implying nothing changed.

Each version's prompt renders as a labeled Prompt row in the same field grid, in the UI font. It used to be unlabeled monospace, which read as a code block next to the sans-serif fields.

Each entry leads with its change summary and keeps the full config behind a toggle, so a history of several versions stays scannable instead of repeating a near-identical table per version.

Also dropped `promptUnchangedSince` (the old "prompt unchanged from vN" hint), which is redundant and actively misleading once the change list is explicit, and moved the prompt diff into a shared `PromptDiff.tsx` so the quality tab and the version history use one implementation.

### Before

v2 only turned on "Multiple tags per session". The entry says "prompt unchanged from v1" and the flag appears nowhere.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/PostHog/pr-assets/2e17263d4c08404a5ba0c67cff5a0c3c36893ba5/2026/07/a6d88cae-5b3a-425a-b34f-6eceb5d97ed7.png">
  Config version entry before the change
</picture>

### After

Same version, same data. Each entry leads with what changed.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/PostHog/pr-assets/ecaf2d92d1af3e85ac44f3bca4f7ea155a143f08/2026/07/9736133f-c1c6-41da-9b8b-3a62d918c5ed.png">
  Config version entry after the change
</picture>

The full config sits behind a toggle, since the summary above already names what moved:

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/PostHog/pr-assets/a883f941cd0ca3b1dbb5a08c3cc5a6c08b2c6a42/2026/07/bf1c73e1-f368-4ab6-9bd4-e1bb48f54b70.png">
  Config version entry with the full config expanded
</picture>

> [!NOTE]
> No migration and no backfill. Existing observations keep their snapshots and their versions render with "Not recorded" for the three new fields. Versions scanned after this deploy are fully explained.

Automated, all run locally and passing:

- **`versionChanges.test.ts` (new).** Guards the bug this PR fixes: a config key absent on one side must read as `off → on`, not as no change. Existing tests couldn't catch it because the diff logic didn't exist. Cases cover flag added / removed / flipped, each top-level field, sampling and model label formatting, the prompt carrying raw text so the diff button works, query changed, query unrecorded on either side, an all-legacy-fields-unrecorded version pair, identical versions, and comparing across a version gap.
- **`test_temporal.py`.** Extended the existing snapshot assertion with the three new fields. Catches a regression where `_build_scanner_snapshot` drops them and every future version silently loses the ability to explain a sampling or filter bump.
- **`test_observation_labels_api.py`.** Updated the marker assertion for the new keys (null on legacy snapshots), and added a case where v1 to v2 changes *only* sampling. Nothing existing covered a version pair distinguishable solely by a non-config field.

Suites: 264 Jest tests across 24 suites, 196 backend tests, `typescript:check` clean, repo-wide `mypy --cache-fine-grained` clean, oxlint/oxfmt/ruff clean.

Manual: I drove the local dev app and seeded scanner versions that differ by exactly one field to check each path renders — a flag-only bump (the screenshots above), a sampling-only bump, a model-only bump, a prompt-only bump, the oldest-version entry, and the live "no scans yet" entry. Seeded data was reverted afterwards.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover the config-versions card, so nothing to update.

## 🤖 Agent context

Claude Code (Opus 5) wrote this after I pointed it at a scanner whose v4 looked identical to v3 and asked it to find the actual change and make the overview cover everything that can change. It queried the scanner's observation snapshots through the PostHog MCP to identify the real diff (`multi_label` off to on) before touching code.

Skills invoked: `/writing-tests` (the new Jest suite and the two extended backend tests were gated through it, which is why the test list names a specific regression per group rather than mirroring the implementation).

The main decision was where version history should come from. Scanner changes are not activity-logged, so the only record is the per-observation run snapshot, which is why the fix extends `ScannerSnapshot` rather than reading an audit trail. I deliberately did not backfill: the honest "Not recorded" state costs less than inventing history for old rows. The other call was to distinguish "not recorded" from "unset" — top-level snapshot fields only compare when both versions recorded them, while `scanner_config` keys always compare, since that dict is snapshotted whole and an absent key genuinely means the value is off. Getting that backwards is exactly what would either hide a real change or invent a fake one.


fasyy612 (approved):
Nice 👌

PR #74065, opened 2026-07-27
chore(ci): route merges through the trunk merge queue

gantoine:
## Problem

We're turning on Trunk's merge queue. Two gaps to close before that's usable.

First, `gh pr merge` and the GitHub merge button become dead ends. Agents don't know that. Left alone they'll try to merge directly, hit the branch ruleset, and either give up or start improvising. And "merge it" is really two jobs: enqueue, then babysit the PR until Trunk lands it or kicks it out.

Second, Trunk can't place a PR into a queue lane without an impacted-targets upload for its head SHA. Without one, PRs never enter the queue at all. This bit posthog/code, where fork PRs silently sat outside the queue and had to be merged by hand.

## Changes

**Impacted-targets upload.** `.github/workflows/trunk-impacted-targets.yml` POSTs to Trunk's `setImpactedTargets` on every PR event. It's a port of the workflow already running in posthog/code, including the fork-auth handling: fork `pull_request` runs get no repo secrets, so they authenticate with `x-forked-workflow-run-id` and Trunk verifies the run id against the head SHA. Internal PRs use `x-api-token`.

It reports `impactedTargets: "ALL"` rather than a computed list. That's always correct - it can never under-report and let the queue merge conflicting PRs in parallel. It buys no parallelism, it just gets every PR into the queue. Narrowing it to real target names later is a follow-up, and `ci-paths-filter.yml` already computes something close to what that would need.

Two deviations from the posthog/code version, both because this repo's conventions differ:

- It's a `workflow_call` reusable workflow folded into `pr-housekeeping.yml`, not its own dispatch. That parent exists precisely to keep small always-fire PR workflows from each burning a run against the 500-runs/10s dispatch cap. The parent's concurrency group is already keyed on PR number, which also happens to be what the posthog/code version worked around by hand (fork PRs frequently share a `head_ref`, so keying on branch name lets one PR cancel another's upload).
- The step no-ops with a warning annotation when `TRUNK_API_TOKEN` is absent on an internal PR, so this can land before the secret does.

**The `merging-prs` skill** covers the whole merge loop: preflight the PR, enqueue with a `/trunk merge` comment, confirm Trunk picked it up, watch until it lands, handle the kick-out case by reading the Trunk bot's comment. A failed impacted-targets upload is the first thing it checks when a PR doesn't appear in the queue.

**AGENTS.md** gets the rule itself, since not every agent will load the skill, plus a line in the stacked-PRs section - restacking force-pushes, and that silently drops a queued branch.

> [!WARNING]
> Also worth confirming: the AGENTS.md line names the check run `Trunk Merge Queue (master)` exactly. The skill matches on a `Trunk Merge Queue` prefix so it tolerates any suffix, but the AGENTS.md line is the thing to re-check against the real queue config.

`bin/hogli lint:workflows` and `actionlint` (1.7.12, the pinned CI version, with `SHELLCHECK_OPTS=--severity=error`) both pass on the two changed workflows. `hogli ci:preflight --strict` passed on push.

I did not exercise the upload or the merge flow end to end - the queue isn't wired up on this repo yet, and I'm not going to POST to Trunk's API from a local shell to find out. The payload shape and auth headers are copied from the workflow currently running in posthog/code, not written from the docs.

No tests added. There's nothing here a test would catch that the two linters don't.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable - this is agent-facing repo documentation and CI wiring, not user docs.

## 🤖 Agent context

I (actually Claude, via Claude Code) started from a draft skill written against a `main`-default repo, then ported the impacted-targets workflow off posthog/code once it came up that the queue needs it.

Skills invoked: `/writing-skills` before writing the skill, `/authoring-ci-workflows` before the workflow. The latter is why this is a reusable workflow under `pr-housekeeping.yml` rather than a standalone dispatch, and why the fork handling is shaped the way it is.

- I considered gating the whole thing on a repository variable, the way `/gating-production-deploys` gates deploys on `CD_DEPLOY_ENABLED`. Rejected it: that skill notes gating `pull_request` builds on a variable breaks contributor CI, which suggests variables don't reach fork PR runs. Since forks are the case this workflow most needs to serve, a vars gate would defeat the point. Hence the token-presence check instead, which only covers internal PRs, and the merge-ordering warning above for the fork half.
- The draft skill had an `allowed-tools` list scoped to a handful of `gh` invocations. I dropped it - it would have blocked the skill's own "fix the failure, push, re-enqueue" path, and most skills in `.agents/skills` don't restrict tools. The `gh pr merge` prohibition stays as a hard rule in the prose, and the ruleset backs it up regardless.
- The draft's polling loop was a foreground `sleep 60`. Claude Code blocks foreground sleep, so it would have stalled. Rewrote it as a state-transition script for the `Monitor` tool or background Bash, printing only on change and exiting on MERGED or CLOSED.
- On skill placement: `/writing-skills` turned out to be aimed at product skills under `products/*/skills/`, which get published to PostHog/skills for customer-facing agents. That's a different thing from the repo-engineering skills in `.agents/skills` this PR adds. Naming and description conventions still applied, but the tooling it describes (`hogli lint:skills`, `hogli build:skills`) doesn't cover `.agents/skills`. Might be worth splitting that guidance.


.github/workflows/trunk-impacted-targets.yml
```diff
@@ -0,0 +1,89 @@
...
+            IS_FORK: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
+            RUN_ID: ${{ github.run_id }}
+            REPO_OWNER: ${{ github.repository_owner }}
+            REPO_NAME: ${{ github.event.repository.name }}
+            PR_NUMBER: ${{ github.event.pull_request.number }}
+            # Head SHA (not the synthetic merge commit): it matches the workflow
+            # run's head_sha, which is what Trunk checks on a fork upload.
+            PR_SHA: ${{ github.event.pull_request.head.sha }}
+            TARGET_BRANCH: ${{ github.event.pull_request.base.ref }}
+            TRUNK_API_TOKEN: ${{ secrets.TRUNK_API_TOKEN }}
```
Piccirello, line 47: Does Trunk support authenticating via OIDC? So that we can get rid of this long lived credential.

Piccirello (approved):

.github/workflows/trunk-impacted-targets.yml
gantoine: no only  `x-api-token` and `x-forked-workflow-run-id` headers, but i can ask in the slack channel.

PR #69100, opened 2026-07-07
feat(mcp-analytics): structured intent themes in the activity tab

lucasheriques:
## Problem

The activity tab's "what agents are trying to do" card is a paragraph of prose. Real agent intents are all worded differently, so there is structure the text cannot show: which themes dominate, how big each is, which tools they involve. Here is the current card on a project with real traffic, which is a fair illustration of the problem. Ten lines that say "agents are analyzing feature flags and investigating user behavior", which is close to unactionable:

> Agents are extensively analyzing and validating feature flags, especially participant and scope-related flags, across multiple PostHog projects to ensure correct configuration and deployment. They are investigating user behavior, event schemas, and telemetry data to support product analytics, adoption, revenue reporting, and to diagnose issues such as interface unresponsiveness, conversion tracking failures, and LLM-related errors.

## Changes

The endpoint now returns a one-sentence summary plus up to five semantic themes, and the tab renders them as rows with a share bar, a description, a verbatim example intent, and tool chips.

Here it is rendered against 100 seeded intents with a real `gemini-3.5-flash-lite` call. Same question, answered with structure:

themes-card

<sub>Local stack, synthetic seeded data.</sub>

**The LLM does not report any figure.** It only names each group and lists the numbers of the intents belonging to it. `resolve_themes` maps those numbers back to the corpus and derives the count, the tools, and the example. Out-of-range numbers and ones an earlier theme already claimed get dropped, so counts sum to at most the number of intents analyzed and the share bar is a real share of the total. This mirrors what `products/replay_vision/backend/feedback_themes.py` already does with `comment_numbers`. The alternative, asking the model for `intent_count` and `example_intent` directly, produces figures nothing can verify.

Digest generation moves to `gemini-3.5-flash-lite`. Clustering short free text is the cheap tier's strength and the endpoint calls it inline, so latency is user-visible. Same choice `feedback_themes` made. Session summaries stay on OpenAI, so `INTENT_MODEL` splits into `SESSION_INTENT_MODEL` and `DIGEST_THEMES_MODEL`.

> [!WARNING]
> The digest now needs `GEMINI_API_KEY`, not `OPENAI_API_KEY`. Without it the endpoint 503s and the card falls back to the verbatim intent list, same as before.

[...]

PR #74074, opened 2026-07-27
feat(mcp-analytics): tool-centric intent analytics with per-call attribution

DanielVisca:
## Problem

Intent clustering answers "what are agents trying to do?" but not the reverse question a tool owner asks: when agents have an intent my tool serves, do they find it, or does another tool take the traffic? Three gaps blocked that:

- Attribution was session-level: the session's first `$mcp_intent` was smeared across every call, so tools used for a later intent were credited to the wrong goal.
- `$mcp_tool_description` capture silently died when the hosted server's analytics moved to the hono dispatcher (the category stamp survived because it was added hono-side later, from the same catalog). The Descriptions table has rendered empty since, and description-to-intent fit had nothing to read.
- `$mcp_listed_tool_names` (the advertised catalog on `$mcp_tools_list`) was captured but had no consumer, so "advertised but never called" was never measured.

**Why:** tool owners need to see discoverability, not just error rates. High description fit with low capture means agents should find the tool but don't; that's a description rewrite, and the overlap data shows what a rewrite might steal traffic from.

## Changes

**Capture (services/mcp):** restore `$mcp_tool_description` stamping via a throw-safe `getToolDescription` helper (clipped to 512 chars since catalogued descriptions run to ~13 KB), on both the hono dispatch path and the CLI path (which stamped neither description nor category). The `@posthog/mcp` SDK's `instrument()` wrapper already stamps descriptions for customer servers, so no cross-repo change.

**Pipeline (intent_clustering.py + Temporal activity):** per-call corpus with last-observation-carried-forward for intentless calls; journeys/switches derived from the same fetch (one less ClickHouse scan); new pure functions for the tool pivot (capture/rank/top competitor/contested score), discovery rate against advertisement-known sessions (null below a 5-session floor), description fit (second embedding kind sharing the content-addressed cache under a distinct prefix), and capped tool overlaps. Snapshot blob is versioned (v2); every cap and coverage limit is reported in `computed_with`, never silently truncated. The Temporal workflow body is unchanged (still one activity), so no determinism gating is needed.

**API:** additive DTO/serializer extensions with null-safe backward compat; pre-v2 blobs render with an empty pivot. OpenAPI/MCP types regenerated.

**Frontend:** the clustering tab gains a "By tool" view (pivot table, per-tool intent detail with capture bars and competitor callouts, discovery-vs-fit scatter with median quadrant lines, overlap table), a tool-name filter, and a sample-coverage line; the tool detail page gains an "Intents served" section fed by the same snapshot.

**Docs:** `exploring-mcp-intent-clusters` and `improving-mcp-tools` skills, the shared `models-mcp.md` reference (including the `$mcp_listed_tool_names` array-extraction gotcha), and the intent-cluster MCP tool descriptions.

> [!NOTE]
> Description fit and the scatter stay empty until roughly a week of description-stamped traffic accumulates in the lookback window after deploy; the UI states this rather than showing zeros. Discovery rates are only measurable in sessions whose tools-list catalog was observed (about half of tool-call sessions, qualitatively), and exec-wrapper-mode sessions advertise only the wrapper; the coverage line and null semantics make both limits visible.

## Walkthrough with data (screenshots)

Captured end-to-end on a locally running stack checked out at this PR: ~56 agent sessions / 130 tool calls of realistic MCP traffic were seeded into ClickHouse (five intent themes: data questions, feature-flag checks, error triage, how-to questions, report building — with error-then-switch sequences, LOCF-style intentless follow-up calls, tools-list catalogs, and stamped descriptions), and the snapshot was computed by running the shipped pipeline code in the activity's exact order against the live ClickHouse — including the new SQL-boundary clips. The only substitution was the embedding-worker HTTP call (no provider locally), replaced with deterministic topic vectors; every other line is the code in this PR. The UI below reads that snapshot through the real API.

### Use case 1: "What are agents trying to do, and does it work?"

The clustering tab's **By intent** view (default). The coverage banner states what the sample represents — 56/56 sessions, 43% of calls carried an intent, advertisement data for 82% of sessions — so percentages read as sample statistics. The scorecards and heatmap surface the contested data-questions cluster at a 17.5% error rate:

03-clustering-by-intent

Clicking a cluster shows its routing breakdown (execute-sql 66.7% vs query-trends at a 42.9% error rate), sample intents, and the journey sankey with the biggest leak called out — six sessions drain into Error via query-trends:

04-cluster-detail

### Use case 2: a tool owner asks "is my tool discoverable?"

Toggle to the new **By tool** view. Every tool gets calls, errors, intents served, a contested-routing badge, discovery rate (called-when-advertised), and description fit. `docs-search` stands out: 5.6% discovery, 2 calls:

05-by-tool-pivot

Selecting it shows the failure precisely: for "how do I filter out internal test users" its description fit is **0.98** — yet it captures only **9.1%** of the intent while `execute-sql` takes 90.9% (with a 0.06 fit of its own). Agents should find this tool and don't — a naming/positioning problem, not a capability gap:

06-tool-intent-detail-docs-search

The discovery-vs-fit scatter puts every tool on that map at once (lower right = fits but rarely picked = rewrite candidates). Validating this chart against real data caught — and this PR now fixes — quadrant copy that was flipped relative to the axes:

11-discovery-scatter

### Use case 3: "which tools compete, and which just work together?"

`execute-sql`'s panel shows the strongest confusion evidence: **query-trends → execute-sql ×6 after errors** for the same intent — agents fall back to SQL when trends queries fail:

07-tool-intent-detail-execute-sql

The overlap table separates substitution from cooperation: `error-tracking-list + session-replay-list` and `dashboard-create + insight-create` are "Used together" (workflows), while `execute-sql + query-trends` is "One or the other" (competition, 14 contested calls):

12-tool-overlaps

### Use case 4: the tool's own page tells the same story

The new **Intents served** section on the tool detail page reads the same snapshot, so numbers agree across surfaces — docs-search's 5.6% discovery badge and per-intent capture appear here too:

09-tool-page-intents-served

### Use case 5: narrowing to one tool family

The tool-name filter works in both views — typing "flag" narrows the pivot to the two flag tools:

08-tool-search-filter

### Blast radius: neighboring surfaces on the same data

The dashboard (including its Intent clusters tile), tool quality, and sessions tabs all render the same seeded traffic through their existing query runners — no regressions from the pivot/attribution changes. Jest (163) and backend (81) suites pass on the branch:

01-dashboard

02-tool-quality

10-sessions

TDD throughout: tests written red first, then implemented.

- `test_intent_clustering.py`: 26 new/updated pure-function cases. The headline regression: per-intent tool counts must contain only same-intent calls (v1's session smear would fail it). Also LOCF session isolation, switch vs self-retry discrimination, journey outcomes scoped to the cluster's own calls, discovery-rate floor semantics, cap accounting, description-fit cosine against hand-built vectors, prefix-separated cache keys, and v1-blob backward compat through the DTO layer. 78 pass locally.
- `test_presentation.py`: extended the stored-snapshot test to pin v1-blob compat through the HTTP layer, plus a v2-sections round-trip asserting nulls survive as nulls. These need the full test DB setup, which this sandbox can't run (no Docker/sqlx); CI runs them. I did verify the same DTO-to-serializer round trip locally with a standalone script against both blob shapes.
- ClickHouse-backed corpus-query tests (sampling, effective-tool coalesce, advertised-catalog union, window stats, latest-description argMax, the LIMIT-100 truncation regression) follow the existing seeded-event pattern; CI validates them. The advertised-tools SQL shape was validated against production data before implementation.
- services/mcp: extended the existing shape-pinning tests for both capture paths (15 pass); typecheck clean in touched files.
- Frontend: `weightedMeanFit` unit tests (5 pass), full `tsgo --noEmit` clean, oxlint clean.
- `hogli ci:preflight --fix` run; repo-wide mypy run as CI does.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Skills and the shared MCP data-model reference are updated in this PR (see Changes).

## 🤖 Agent context

Authored by Claude (PostHog Code session) directed by Daniel Visca (assignee). Skills invoked: /writing-tests, /writing-code-comments, /improving-drf-endpoints, /writing-user-facing-copy. Design decisions of note: per-call LOCF attribution was chosen over excluding intentless calls (strictly narrower than v1's whole-session smear, with the imputed share reported); discovery denominators only count advertisement-known sessions rather than treating a missing tools-list event as "nothing advertised"; quadrant lines on the scatter sit at data medians instead of tuned constants; and the snapshot caps all report dropped counts. The capture regression window and root cause were verified against production data and commit history before writing code.

---


DanielVisca:
CI status notes from the agent run:

- **Django tests Core shard 8/20** fails on `test_team_project_parity.py::test_viewset_actions_are_a_superset_of_team_viewset` (`default_evaluation_contexts` / `default_release_conditions`). This same test [fails on master itself](https://github.com/PostHog/posthog/actions/runs/30313482398/job/90134418247), so it is pre-existing master breakage unrelated to this PR; not fixing it here. A merge from master once it's fixed will clear the shard.
- **Frontend typechecking** failed on the typegen drift step (a hand-ordered kea values interface); fixed in 86acb32.
- Greptile's P1 (stale advertised description for `execute-sql`) was valid and is fixed in 86acb32: `trackToolCall` now stamps the served, per-request formatted description on both dispatch paths, with wiring tests for each.
- Also bounded the description-embedding fan-out to the top corpus tools by call volume (flagged by an unlisted review bot; verified independently: without the cap, unique tool names per call would each trigger an embedding request).
- **Visual review: 2 changed** is expected: this PR adds the By tool view and the Intents served section; needs human approval.

DanielVisca:
Overnight status, and one thing that needs your call.

**Pushed three review fixes** (`9ec92c578`):
- The advertised-catalog union clipped tool-name *length* but nothing else, so both the array inside each `$mcp_tools_list` event and the number of those events per session were unbounded, and `groupArray` + `arrayFlatten` materialised all of it. Each event's array is now sliced and each session's events capped before the union.
- `compute_tool_pivot` rebuilt the advertised-session list per tool by scanning every session — O(tools × sessions) over an uncapped, sender-controlled tool set. It now uses a one-pass inverted index, matching the `sessions_calling` index already built nine lines above.
- The cap comment claimed *every* cap reports its dropped count in `computed_with`, and the skill repeated that as a completeness check. Only `dropped_tools` and `dropped_overlap_pairs` exist; the per-tool and per-cluster caps truncate silently. Both now say so.

Verified: ruff, mypy, and 38 non-ClickHouse tests pass locally. The new bound has a ClickHouse test, so CI covers it end to end.

**Left alone deliberately: this PR now conflicts with master**, and I did not resolve it. #75075 landed at 02:35 and rewrote the same clustering files, so six files conflict: `constants.py`, `intent_clustering.py`, `test_intent_clustering.py`, `MCPAnalyticsClustering.tsx`, `mcpClusteringLogic.ts`, `mcpClusteringLogic.test.ts`.

The reason I stopped rather than resolving it: #74080 and #74081 are *also* rewriting `mcpClusteringLogic.ts` and `MCPAnalyticsClustering.tsx`, and this PR is branched off master rather than stacked with them. Whichever lands first forces the other to re-resolve the same UI by hand, so resolving it now mostly guarantees doing it twice — and the resolution needs a product decision I can't make from the code (this PR keeps the tool-by-cluster heatmap that #74081 deletes).

Suggested order: land #74080 → #74081 (both green and non-conflicting now), then rebase this one onto the result, at which point the heatmap question resolves itself because #74081 will have already removed it.

Also worth flagging: `stamphog` REFUSED on size and the `deps_toolchain` deny, so **nothing has reviewed these 3400 lines** — this needs a human approval regardless of CI.

gesh (commented):
Amazing job, thanks a lot for your help!! 🙌 

I'll do a code review later, just something product/ux related - what do you think of reorganizing the UI a bit? Seems like there is a lot of data, stats, searching, sorting, and we could make it easier to digest. 

Dropping a quick mockup I made (very rough, the boxes should be tiles - same as what we have on the dashboard 😄) , I'd love to hear your thoughts on this. Not a blocker for this PR, I can pull it off next week. :)

Image

DanielVisca:
Rebased onto master (the conflicts came from #75075's snapshot/heatmap caps and #69100 landing in the same files) and went through every open review thread. Status of each:

**Fixed in [923a1e6](https://github.com/PostHog/posthog/commit/923a1e6cf9749f07f5f782228afb25f5d60134e8):**
- Advertised-catalog memory bound (hex-security): the per-event and per-session-event caps could still multiply across sampled sessions into gigabytes of worker-bound strings. The distinct union is now also capped per session (`MAX_ADVERTISED_TOOLS_PER_SESSION`), with the absent-vs-bounded catalog distinction preserved.
- Unbounded tool-description aggregation (veria): the descriptions query now filters to the bounded corpus tool set before `GROUP BY`, so sender-invented names never build `argMax` state; the activity passes `top_corpus_tools` up front and the Python-side filter stays as defense for other callers.
- Heatmap empty state now distinguishes an active tool filter from a snapshot with no tool calls.
- The tool page's intents panel now surfaces a failed clustering run (mirroring the clustering tab) instead of reading as never-computed.
- `self_retries` gets the same one-line gloss as `switches` in the skill.

**Fixed earlier in [9ec92c5](https://github.com/PostHog/posthog/commit/9ec92c5781618fd9ac33f1646196d42928c3798e)** (threads still open, content addressed): advertised-catalog event/array caps, the `compute_tool_pivot` inverted index, and the SKILL.md/comment over-claim about cap visibility.

**Considered, deliberately unchanged:**
- Scorecards staying whole-snapshot while the heatmap filters: #75075 gave the scorecards an explicit whole-snapshot framing ("Scorecards cover the top X of Y clusters by call volume"), so filtering them would fight that caption; the review bot itself called the global-cards-over-filtered-table pattern defensible.
- `database_sync_to_async` vs the pool variant: this file uses the default helper throughout, the activity runs at max concurrency 4 on a daily schedule, and runtime is dominated by the embedding step — happy to switch the file wholesale in a follow-up if wanted.
- The discovery scatter ignoring the tool search: intentional — its quadrant lines are medians over the plotted population, so filtering would silently move the reference lines a tool is judged against. The heatmap's render cap does now compose with the search filter (filter first, then cap).

Merge conflicts were resolved by keeping both sides' intent: the cluster-count cap from #75075 now truncates before the tool pivot is computed, so `tools[].clusters` only ever references clusters that exist in the persisted snapshot, and `n_clusters` keeps the pre-truncation count.

products/mcp_analytics/frontend/clustering/MCPAnalyticsClustering.tsx
```diff
@@ -166,28 +151,70 @@ function Scorecards(): JSX.Element {
...
+        </div>
+    )
+}
+
 function Heatmap(): JSX.Element {
-    const { sortedClusters, toolColumns, selectedClusterId } = useValues(mcpClusteringLogic)
+    const { filteredClusters, toolColumns, selectedClusterId } = useValues(mcpClusteringLogic)
     const { selectCluster } = useActions(mcpClusteringLogic)
 
     if (toolColumns.length === 0) {
```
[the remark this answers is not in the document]
DanielVisca: Fixed. The empty state now branches on the active filter: `MCPAnalyticsClustering.tsx:233-235` renders `No clusters call a tool matching "<query>". Clear the search to see all clusters.` when `toolSearch` is set, and keeps the no-data copy only when there genuinely are no tool calls.

products/mcp_analytics/frontend/clustering/ToolDetailIntentsSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+                </>
+            ) : (
+                <>
+                    Intent clustering shows which goals agents pursue when they call this tool.{' '}
+                    <LinkPrimitive to={urls.mcpAnalyticsIntentClustering()} className="text-accent">
+                        Compute clusters
+                    </LinkPrimitive>{' '}
+                    to see them here.
+                </>
+            )}
```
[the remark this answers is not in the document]
DanielVisca: Fixed. `ToolDetailIntentsSection.tsx:25` now checks `snapshot.status === 'error'` and surfaces the failure with `error_message`, mirroring the clustering tab, so a failed run no longer reads as never-computed.

products/mcp_analytics/skills/exploring-mcp-intent-clusters/SKILL.md
```diff
@@ -35,10 +43,13 @@ posthog:mcp-analytics-intent-clusters-retrieve
...
-has a `label`, `intent_count`, `call_count`, `error_count`, `error_rate_pct`,
-`routing_entropy`, a `tool_distribution` (which tools that goal routes to, with
-per-tool error rates), and `sample_intents`.
+embedding model, clustering parameters, and sample-coverage percentages), a
+`clusters` array, a `tools` array (the tool pivot), and `tool_overlaps`. Each
+cluster has a `label`, `intent_count`, `call_count`, `error_count`,
+`error_rate_pct`, `routing_entropy`, a `tool_distribution` (which tools that
+goal routes to, with per-tool error rates), `sample_intents`, plus `switches`
+(errored call immediately followed by a different tool for the same intent —
+the strongest "agents mix these tools up" evidence) and `self_retries`.
```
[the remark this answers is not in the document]
DanielVisca: Fixed. `self_retries` now carries the same gloss as its sibling in the snapshot-shape paragraph (`SKILL.md:53-54`): "errored call immediately retried with the same tool, a sign the tool's error messages aren't helping agents self-correct".

products/mcp_analytics/skills/exploring-mcp-intent-clusters/SKILL.md
```diff
@@ -48,6 +59,35 @@ points at a class of agent goals the tools serve badly.
...
+`tool_overlaps` lists pairs competing for the same intents; use
+`sessions_with_both` vs `sessions_with_either` to separate workflows (used
+together) from confusion (one or the other).
+
+Read coverage before quoting numbers: `computed_with.sampled_sessions` /
+`session_coverage_pct` say how much of the window the corpus represents, and
+`advertisement_coverage_pct` bounds what discovery rates can see. Only sessions
+with an observed tools-list catalog enter discovery denominators, and sessions
+in exec-wrapper mode advertise only the wrapper, so per-tool discovery is
+measured on full-catalog sessions.
```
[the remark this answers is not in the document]
DanielVisca: Fixed, and you were right that the invariant was false for 4 of 6 caps. The over-claiming comment is gone: `intent_clustering.py:67-69` now states that only `dropped_tools` and `dropped_overlap_pairs` report what they dropped and that the per-tool and per-cluster caps truncate silently. The skill says the same at `SKILL.md:93-97`, including how to read a capped list ("at least that many", not "exactly").

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +737,231 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
...
+            sessions_calling[call.tool].add(session_id)
+
+    tools: list[dict[str, Any]] = []
+    for tool, cluster_entries in per_tool_clusters.items():
+        call_count = call_totals[tool]
+        weighted_entropy = sum(entry["calls"] * entry["cluster_entropy"] for entry in cluster_entries)
+        advertised_sessions = [
+            session_id for session_id, advertised in advertised_by_session.items() if tool in advertised
+        ]
+        called_when_advertised = sum(1 for session_id in advertised_sessions if session_id in sessions_calling[tool])
```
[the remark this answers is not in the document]
DanielVisca: Fixed. `compute_tool_pivot` now builds a one-pass inverted index (`intent_clustering.py:906-909`, `advertised_sessions_by_tool`) and looks each tool up in it at `:915`, mirroring the adjacent `sessions_calling` index. The per-tool rescan of every advertised session is gone, so the O(tools x sessions) path no longer exists.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -124,257 +194,391 @@ class IntentRecord:
...
+_TOOL_DESCRIPTIONS_SQL = """
+SELECT
+    left({tool_expr}, {max_tool_len}) AS tool,
+    argMax(left({description_expr}, {max_desc_len}), timestamp) AS description
+FROM events
+WHERE event = {event}
+    AND timestamp >= now() - INTERVAL {lookback_days} DAY
+    AND notEmpty({tool_expr_where})
+    AND notEmpty({description_expr_where})
+GROUP BY tool
```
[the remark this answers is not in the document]
DanielVisca: Fixed. The descriptions query now filters to the bounded corpus tool set before `GROUP BY` (`intent_clustering.py:280`: `AND left({tool_expr_where}, {max_tool_len}) IN {tools}`), so `argMax` state is only built for tools already in the corpus rather than for every sender-invented name. The activity passes `top_corpus_tools` in; the Python-side filter stays as defense for other callers.

DanielVisca:
Went through CI and every open review thread. Pushed two commits; CI now has **zero failing checks**, and the only thing blocking merge is review approval.

## What was actually broken

Eight red contexts, three root causes, and only one belonged to this PR:

- **Frontend typechecking (this PR's fault).** The rebase resolution took master's copy of `mcpClusteringLogic.test.ts` and re-added only its import line, so the fixtures never picked up the fields this PR makes required: `switches` / `self_retries` on `MCPIntentClusterApi`, `tools` / `tool_overlaps` on the snapshot, and the thirteen coverage fields on `MCPIntentClusterSnapshotMetaApi`. Fixed in 6d65c01. `Frontend Tests Pass` was the aggregator failing off the same cause; Jest itself was green.
- **Repo checks (master-side).** `hogli product:lint --all` was failing on `customer_analytics` and `error_tracking` turbo.json input narrowing, neither of which this branch touches. Confirmed by running it both ways: master printed `All 68 products passed`, this branch printed `2 product(s) failed`. Merging master in cleared it. That failure was also cancelling all of Backend CI, which is why `Django Tests Pass` was red with every Django job marked `cancelled`.
- **Three flakes.** A Hog-function-template seeding race in MCP integration, two `sql-editor-*-postgres` warehouse specs, and an instant `Install uv` setup failure (which took `Sandbox Base Image Pass` with it). All green on master, none touching this PR's paths, all cleared on re-run.

## Two checks that had never actually run

`Validate OpenAPI types` and the repo-wide mypy step were both cancelled by the repo-checks cascade, so they were unverified rather than passing. Worth noting `ci:preflight` cannot catch the OpenAPI one on its own: that check is `verify=None`, so it only advises you to regenerate whenever a serializer changes, without ever comparing. Ran the real thing against a live DB, and there was no drift. Both are green on the current head now.

## Review threads

18 threads, 6 were already resolved, 12 open. Checked each against the tree rather than the status comments, and 6 of the 12 were fixed in 9ec92c5 / 923a1e6 with the thread just left open. Replied on each with the verified line and resolved them.

**Three are genuinely open, all low severity:**

1. The advertised-catalog union bounds its *output*, not its aggregation state. `arraySlice(arrayDistinct(arrayMap(..., arrayFlatten(groupArray(listed)))), 1, N)` at `intent_clustering.py:229-233` materializes the full union before slicing, so `MAX_ADVERTISED_TOOLS_PER_SESSION` does not limit what ClickHouse builds. `groupUniqArray(N)` over `arrayJoin`ed names would. This one landed about four minutes after the last push, so it had not been looked at.
2. `MAX_CORPUS_SESSIONS` applies after `GROUP BY session_id ORDER BY cityHash64(...)` (`intent_clustering.py:198-200`), so the cap does not bound the grouping either.
3. `ToolOverlapTable` still ignores the tool-name filter (`toolOverlaps` is `(s) => [s.snapshot]`, `mcpClusteringLogic.ts:326`). Only the scatter half of that finding was addressed, and the scatter half is defensible since its quadrant lines are medians over the plotted population.

The first two are the same shape as the bounds already fixed on this branch, so the pattern is established and these look like two spots the sweep missed. Left them for a decision rather than reworking reviewed SQL unprompted.

## Still needs a human

`stamphog` REFUSED on size and a `deps_toolchain` deny, so nothing has reviewed these lines holistically, only per-line bots. Green CI does not substitute for that.

adamleithp (approved):
no context

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +796,242 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
...
+        for rank, entry in enumerate(distribution, start=1):
+            tool = entry["tool"]
+            competitor = next(
+                ({"tool": other["tool"], "pct": other["pct"]} for other in distribution if other["tool"] != tool),
+                None,
+            )
+            per_tool_clusters[tool].append(
+                {
+                    "cluster_id": cluster["id"],
+                    "label": cluster["label"],
```
gesh: 🔴 **Critical — snapshot payload.** Every pivot entry copies the cluster's `label`, which is the medoid intent text clipped at `MAX_INTENT_TEXT_LENGTH` (1000 chars), plus `cluster_call_count` and `cluster_entropy` — all per-cluster constants.

At the caps this file sets, that's 300 tools × 20 clusters × ~1 KB ≈ **6 MB of duplicated text** in one JSONB blob. `MAX_SNAPSHOT_CLUSTERS = 100` exists specifically because "hundreds of near-singleton clusters make a multi-MB payload the UI then has to render in one go" — this reintroduces that on the tool axis.

`cluster_id` is already in the entry and the frontend has `clusters[]` from the same logic (`ToolIntentDetail` already reads it for switches), so the label/count/entropy can be joined client-side.

products/mcp_analytics/frontend/MCPAnalyticsToolDetail.tsx
```diff
@@ -612,6 +613,13 @@ function MCPAnalyticsToolDetailContent({ toolName }: { toolName: string }): JSX.
 
             <LemonDivider />
 
+            <div className="flex flex-col gap-3 px-4 pb-4">
+                <SectionHeader title="Intents served" subtitle="From the latest intent cluster snapshot" />
+                <ToolDetailIntentsSection toolName={toolName} />
```
gesh, line 630: 🔴 **Critical — this pulls the whole snapshot onto the tool detail page.** `ToolDetailIntentsSection` calls `useValues(mcpClusteringLogic)`, and that logic's `afterMount` fires `loadSnapshot()`. So every tool page view downloads the full cluster snapshot (100 clusters + up to 300 tool pivots, see the payload comment on `intent_clustering.py`) to render one tool's section — and `loadSnapshotSuccess` will start the poll interval if a run happens to be computing.

Options: a tool-scoped read on the API, or lazy-mount the section (render on expand / intersection) so the fetch isn't on the critical path.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +796,242 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
...
+                    "cluster_call_count": cluster["call_count"],
+                    "cluster_entropy": cluster["routing_entropy"],
+                    "description_fit": description_fit.get(tool, {}).get(cluster["id"]),
+                    "top_competitor": competitor,
+                }
+            )
+            call_totals[tool] += entry["count"]
+            error_totals[tool] += entry["errors"]
+
+    sessions_calling: dict[str, set[str]] = defaultdict(set)
```
gesh, line 914: **Inconsistent session population.** `sessions_calling` is built from every call, including those with `intent_text is None`, while `compute_tool_overlaps` builds `session_tools` with `if call.intent_text is not None`.

Consequence: `session_count` and the `called_when_advertised` numerator count sessions whose only calls to the tool were unattributed, but `call_count` — documented as "Intent-attributed calls" — does not. A tool can show a healthy discovery rate driven entirely by calls that never entered any cluster.

Worth picking one definition (attributed-only would match `call_count`) or documenting the split in the serializer help text.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -641,24 +1097,82 @@ def build_snapshot(
     n_clusters_total = len(clusters)
     del clusters[MAX_SNAPSHOT_CLUSTERS:]
 
+    fit = compute_description_fit(embeddings, labels, description_embeddings or {})
+    if calls_by_session is not None:
+        tools, dropped_tools = compute_tool_pivot(
```
gesh, line 1128: **The pivot runs on the truncated cluster list.** `del clusters[MAX_SNAPSHOT_CLUSTERS:]` happens a few lines above, so per-tool `call_count` / `error_count` silently exclude calls belonging to clusters that didn't survive the cap, and `dropped_tools` doesn't cover that loss.

This file is otherwise careful that "every cap and coverage limit is reported in `computed_with`, never silently truncated" — either compute the pivot before the `del` (keeping the entry list itself capped), or add a `pivot_excludes_dropped_clusters` / dropped-call count to `computed_with`.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -124,257 +200,408 @@ class IntentRecord:
...
+    toString(properties.$mcp_is_error) IN ('true', '1') AS is_error
 FROM events
 WHERE event = {event}
     AND $session_id IN {session_ids}
-    AND properties.$mcp_tool_name IS NOT NULL
-    AND properties.$mcp_tool_name != ''
     AND timestamp >= now() - INTERVAL {lookback_days} DAY
-GROUP BY session_id, tool_name
+    AND notEmpty({tool_expr_where})
+ORDER BY session_id, timestamp
```
gesh, line 225: **Row-cap truncation biases discovery rates downward.** `ORDER BY session_id` + `LIMIT MAX_QUERY_ROWS` means that when the cap is hit, whole sessions at the tail of the id ordering contribute zero calls. Those same session ids are still passed to `fetch_advertised_tools`, so they stay in `advertised_sessions` denominators with nothing in the numerator — discovery rates skew low exactly when the truncation warning fires.

Either drop sessions absent from `calls_by_session` before the advertised query, or lower `MAX_CORPUS_SESSIONS` for the run when the cap trips.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +796,242 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
     return indices[best_local]
 
 
+def top_corpus_tools(records: list[IntentRecord], max_tools: int = MAX_TOOLS_IN_SNAPSHOT) -> set[str]:
```
gesh, line 810: **Different rank basis from the pivot.** This ranks tools over *all* records, but `compute_tool_pivot` caps its 300 by volume over the *surviving* clusters. With more than `MAX_TOOLS_IN_SNAPSHOT` tools the two sets can diverge, so a tool that made the pivot may never have had its description fetched — it renders as "no description captured" with a null fit for no reason the user can see.

Ranking both on the same population fixes it.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -10,49 +10,74 @@
...
 from posthog.api.embedding_worker import EmbeddingResponse, async_generate_embedding
 from posthog.clickhouse.query_tagging import Feature, Product, tags_context
 from posthog.models.team.team import Team
 from posthog.sync import database_sync_to_async
 
-from products.mcp_analytics.backend.constants import MAX_SNAPSHOT_CLUSTERS, MCP_TOOL_CALL_EVENT
-from products.mcp_analytics.backend.models import MCPIntentEmbeddingCache, MCPSession
+from products.mcp_analytics.backend.constants import MAX_SNAPSHOT_CLUSTERS, MCP_TOOL_CALL_EVENT, MCP_TOOLS_LIST_EVENT
+from products.mcp_analytics.backend.hogql_queries.base import EFFECTIVE_DESCRIPTION_SQL, EFFECTIVE_TOOL_SQL
+from products.mcp_analytics.backend.models import MCPIntentEmbeddingCache
```
gesh, line 52: ❓ **`MCPSession` dropped from the corpus.** v1's `fetch_intent_corpus` overlaid the on-demand LLM summaries from `posthog_mcp_session` (they won per session, and pulled in summarised sessions the event sample missed). The per-call corpus removes that path entirely, but the model and the generate-intent endpoint are still live in the product.

If that's deliberate, worth saying so in the PR description and the skill docs — a user who generates a session summary will reasonably expect it to influence clustering.

Related: the `NO_INTENT_RECORDED_FALLBACK` check in `build_call_corpus` is now near-dead (that placeholder was written by the summariser, never by an event `$mcp_intent`), and its comment still refers to "the summariser's 'nothing here' placeholder".

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -124,257 +200,408 @@ class IntentRecord:
...
+LIMIT {max_rows}
+"""
+
+# Whole-window totals for the coverage metrics in computed_with. One cheap
+# aggregate row; deliberately unsampled so the UI can honestly relate the
+# sampled corpus to the traffic it represents.
+_WINDOW_STATS_SQL = """
+SELECT
+    count() AS total_calls,
+    countIf(coalesce(toString(properties.$mcp_intent), '') != '') AS calls_with_intent,
```
gesh, line 269: **Denominator doesn't match the corpus population.** `total_calls` here counts every `$mcp_tool_call` with a non-empty `$session_id`, with no `notEmpty(<effective tool>)` filter — but `_SESSION_CALLS_SQL` requires a tool name. `intent_coverage_pct` is then presented in the coverage banner as if the two were the same set. Adding the same `notEmpty` predicate keeps them comparable.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +796,242 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
...
-    ``journeys_by_cluster`` is the output of ``aggregate_journeys_per_cluster``
-    keyed by cluster id. Optional; clusters without an entry get a null journey.
+    All keyword arguments are optional so degraded runs still snapshot: without
+    ``calls_by_session`` there are no journeys, switches, pivot, or overlaps;
+    without descriptions there is no fit. Missing coverage inputs serialize as
+    null meta fields, never as fabricated zeros.
     """
     if len(records) == 0:
-        return _empty_snapshot(distance_threshold, n_intents=0)
+        return empty_snapshot(distance_threshold, n_intents=0)
```
gesh: **Known coverage is thrown away on the empty path.** This returns before `window_stats` / `corpus_stats` are applied, so a run that saw plenty of traffic but couldn't attribute any of it reports every coverage field as null — indistinguishable from "no data at all". Passing the stats through here would let the UI say "0% of calls carried an intent", which is the actionable message.

products/mcp_analytics/frontend/clustering/ToolPivotTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+                                    {tool.error_count > 0 ? (
+                                        <span className="tabular-nums text-danger">
+                                            {humanFriendlyNumber(tool.error_count)}
+                                        </span>
+                                    ) : (
+                                        <span className="text-muted">0</span>
+                                    )}
+                                </TableCell>
+                                <TableCell align="right">
+                                    <span className="tabular-nums">{tool.clusters.length}</span>
```
gesh: **"Intents served" shows a truncated count.** `tool.clusters` is capped server-side at `MAX_CLUSTERS_PER_TOOL = 20`, so a tool serving 40 clusters displays `20` with no indication. Either send a real `n_clusters_served` on the pivot, or rename the column to something like "Top intents" and cap-annotate at 20.

products/mcp_analytics/frontend/clustering/DiscoveryScatter.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    This chart needs tools with both a captured description and enough advertised sessions to measure
+                    discovery. Descriptions accumulate as new calls are captured; check back after a few days of
+                    traffic.
+                </p>
+            </div>
+        )
+    }
+
+    const fits = scatterPoints.map((p) => p.fit)
+    const fitDomain: [number, number] = [Math.min(...fits, 0), Math.max(...fits, 0.01)]
```
gesh: **x-domain anchored at 0 flattens the chart.** Cosine similarities for one server's tool descriptions against its own intent centroids tend to sit in a narrow band (e.g. 0.55–0.85). Forcing the domain to start at 0 pushes every bubble against the right edge and puts the median line right through the pile, which defeats the quadrant read the copy above promises.

Padding around the observed min/max (with a floor on the span so a single point doesn't explode the scale) would keep the quadrants legible.

services/mcp/src/tools/toolDefinitions.ts
```diff
@@ -128,6 +128,23 @@ export function getToolCategory(toolName: string): string | undefined {
...
+export const MAX_CAPTURED_DESCRIPTION_LENGTH = 512
+
+/**
+ * The description a tool advertises to agents, clipped for analytics capture, or
+ * undefined for tools without a catalogued definition (e.g. the `exec` wrapper).
+ * Like {@link getToolCategory} this never throws, so it is safe to call from the
+ * analytics hot path where a missing definition must not break the request.
+ */
+export function getToolDescription(toolName: string): string | undefined {
+    return getToolDefinitions()[toolName]?.description.slice(0, MAX_CAPTURED_DESCRIPTION_LENGTH)
```
gesh: Nit: the optional chain stops at the definition lookup, so `?.description.slice(...)` still throws if a definition ever lacks `description`. The zod schema makes it required today, so this is purely defensive — but the JSDoc right above promises "never throws" on the analytics hot path, and `?.description?.slice(...)` is free.

Also worth noting while here: `getToolDefinitions()` rebuilds the merged `{ ...local, ...generated }` record on every call, and the hot path now invokes it twice per tool call (`getToolCategory` + `getToolDescription`). Memoising the merge would be a cheap win.

products/mcp_analytics/backend/intent_clustering.py
```diff
@@ -564,27 +796,242 @@ def _medoid_index(embeddings: np.ndarray, indices: list[int]) -> int:
...
+    population and cap the tool pivot uses. Everything description-related
+    (the ClickHouse aggregation, the embedding fan-out) is bounded to this set."""
+    totals: Counter[str] = Counter()
+    for record in records:
+        for tool, count in record.tool_counts.items():
+            totals[tool] += count
+    return {tool for tool, _ in sorted(totals.items(), key=lambda item: (-item[1], item[0]))[:max_tools]}
+
+
+def select_tools_for_description_fit(
```
gesh: Nit: this has no production caller — the activity bounds the fan-out by passing `top_corpus_tools(...)` into `fetch_tool_descriptions`, and the only references are in `test_intent_clustering.py`. Either wire it in as the actual guard or drop it; a tested-but-unreachable helper tends to rot.

products/mcp_analytics/backend/logic.py
```diff
@@ -756,9 +758,19 @@ def get_intent_cluster_snapshot(team: Team) -> contracts.IntentClusterSnapshot:
         last_computed_by_email=snapshot.last_computed_by.email if snapshot.last_computed_by else "",
         clusters=[_to_cluster_dto(item) for item in clusters_raw if isinstance(item, dict)],
         computed_with=_to_meta_dto(meta_raw) if isinstance(meta_raw, dict) else None,
+        tools=[_to_tool_pivot_dto(item) for item in tools_raw if isinstance(item, dict)],
+        tool_overlaps=[_to_overlap_dto(item) for item in overlaps_raw if isinstance(item, dict)],
     )
 
 
+def _opt_float(value: Any) -> float | None:
```
gesh, line 807: Nit: `isinstance(True, int)` is `True`, so a boolean in the blob coerces to `1.0` / `1` instead of being rejected. Only reachable via a hand-edited or future-shaped blob, but `isinstance(value, bool)` guard is one line.

gesh (commented):
Really nice work — the per-call LOCF attribution and per-cluster subsequence scoping are a genuine correctness fix over the v1 session smear, the SQL-boundary clipping is careful about sender-controlled dimensions, and the null-vs-zero discipline holds all the way from pipeline to UI. Notes below.

**🔴 Critical — worth fixing before merge**
1. Pivot entries duplicate the cluster `label` (up to 1 KB of intent text) per tool×cluster. Worst case ~6 MB in a JSONB blob whose 100-cluster cap exists precisely to avoid this.
2. The tool detail page now mounts `mcpClusteringLogic`, so every tool page view downloads that whole blob (and may start polling). Together with (1) that's a real regression on a page that only needs one row.

**❓ Needs a decision**
3. The `MCPSession.intent` summary overlay is gone from the corpus while the model and generate-intent endpoint remain. Intended? It isn't in the PR description.

**Correctness (non-blocking, but they make numbers disagree)**
4. `sessions_calling` counts unattributed calls; overlaps exclude them — `session_count` / `discovery_rate_pct` and `call_count` measure different populations.
5. The pivot is computed after clusters are truncated, so per-tool counts silently miss dropped clusters (unreported in `computed_with`).
6. Row-cap truncation leaves dropped sessions in discovery denominators.
7. `top_corpus_tools` and the pivot rank over different populations → null descriptions past 300 tools.
8. `intent_coverage_pct` denominator lacks the `notEmpty(tool)` filter the corpus has.
9. Empty-corpus path returns all-null coverage even when window stats are known.

**UI + nits:** capped "Intents served" count, scatter x-domain anchored at 0, `?.description.slice()`, catalog re-merge per call, dead `select_tools_for_description_fit`, bool-accepting `_opt_*`.

PR #74077, opened 2026-07-27
feat(replay-vision): deliver digests to webhooks

ksvat:
> Part of a 3-PR stack adding webhook delivery to Replay Vision:
> 1. **#74077 (this PR)** – backend delivery to webhooks
> 2. #74078 – API accepts webhook targets
> 3. #74079 – frontend picker

## Problem

Replay Vision digests and alerts can only be delivered to Slack. A user asked to send them somewhere else via a webhook, so they can route findings into their own systems or tools we don't natively integrate with.

Delivery already runs through a per-action `internal_destination` HogFunction driven by a private internal event, so the plumbing to add a second destination type is already there. This is the first of three PRs and does the backend half.

## Changes

- `delivery.py` now dispatches on the delivery target's `type`. A `webhook` target provisions the `template-webhook` HTTP destination (URL, `POST`, versioned header) instead of `template-slack`. Shared scaffolding for both types lives in `_base_destination_payload`.
- The webhook body is a structured JSON envelope (`{id, type, timestamp, data}`) following the same convention the alerts and logs webhook destinations use, including the `X-PostHog-Webhook-Version: 1` header. `type` routes on the event kind so a consumer can tell a digest from an alert firing from an alert recovering.
- The emitted `$replay_vision_action_ready` event now carries the structured fields the webhook body references (`event_kind`, `report_markdown`, `run_url`, action/scanner names, `observation_count`), so consumers get clean fields rather than Slack-formatted text.

No migration: `delivery_config` was already a generic list, and the emit change is activity-only (no workflow command change, replay-safe). The API serializer still gates which target types it accepts – that widens in #74078.

> [!NOTE]
> This PR generalizes provisioning but the API won't accept a `webhook` target until #74078 lands. On its own it's exercised through `provision_delivery` directly.

Automated only (I'm Claude, an agent – I did not manually click through the UI). Ran the affected backend suites locally, all green:

- `test_vision_action_delivery.py` – new cases: webhook target provisions `template-webhook` with the URL, versioned header, and a body referencing the event props; mixed slack+webhook targets create both; reconcile switches slack→webhook (archives the stale one).
- `test_vision_action_engine.py` – the emit event now carries the structured props; all three `event_kind` values (digest / alert_fired / alert_recovered) are covered via a parameterized case.

Each new test catches a regression the Slack-only tests didn't: a broken webhook dispatch (wrong template, missing header, body not referencing event props) or a dropped structured field would silently ship an empty/malformed payload.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

I (Claude, via Claude Code) wrote this under @ksvat's direction. The design deliberately copies the pattern from `products/alerts/backend/destination_configs.py` (Slack/Discord/Webhook/Teams destinations) rather than importing it, since `replay_vision` doesn't depend on `products.alerts` in tach and coupling them for a small pattern isn't worth it. Scoped v1 to Slack + Webhook; Discord/Teams would be a few more lines in the same shape if wanted later.

Invoked `/writing-tests` and `/writing-code-comments` while shaping this.

### DIFF
--- a/products/replay_vision/backend/api/delivery.py
+++ b/products/replay_vision/backend/api/delivery.py
@@ -20,6 +20,11 @@ logger = structlog.get_logger(__name__)
 EVENT_NAME = "$replay_vision_action_ready"
 _INTERNAL_DESTINATION = "internal_destination"
 _SLACK_TEMPLATE = "template-slack"
+_WEBHOOK_TEMPLATE = "template-webhook"
+
+# Marks our webhook payloads so a consumer can pin the schema and we can evolve it without breaking them.
+# Mirrors the alerts/logs destination convention (products/alerts/backend/destination_configs.py).
+_WEBHOOK_HEADERS = {"Content-Type": "application/json", "X-PostHog-Webhook-Version": "1"}
 
 
 def _managed_destinations(action: VisionAction, team: Team) -> QuerySet[HogFunction]:
@@ -42,11 +47,13 @@ def _channel_id(value: str) -> str:
     return value.split("|", 1)[0].strip()
 
 
-def _slack_destination_payload(action: VisionAction, target: dict[str, Any]) -> dict[str, Any]:
+def _base_destination_payload(action: VisionAction, template_id: str) -> dict[str, Any]:
+    """The internal_destination scaffold every delivery type shares: the trigger filter that binds the
+    destination to this action, plus its enabled state. Callers fill in `inputs` per template."""
     return {
         "type": _INTERNAL_DESTINATION,
         "enabled": action.enabled,
-        "template_id": _SLACK_TEMPLATE,
+        "template_id": template_id,
         "name": f"Replay Vision · {action.name}",
         "filters": {
             "events": [{"id": EVENT_NAME, "type": "events"}],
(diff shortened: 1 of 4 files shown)


ksvat:
* **#74079** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74079?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74078** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74078?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#74077** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74077?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ksvat:
### Merge activity

* **Jul 28, 4:09 AM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog/74077).

PR #74085, opened 2026-07-28
chore(personhog): protocol deploy resilience

z0br0wn:
## Problem

 A rolling deploy terminates leader pods while handoffs are in flight. In dev this wedged the whole cluster: a cancellation told a router to drain its stash back to a pod that had just died, that drain can never converge (nothing can be delivered while new arrivals and expiry-retries keep the queue alive), and it ran inline in the router's only coordination loop. The loop froze, the router stopped acking freezes while its healthy lease kept it in every freeze quorum, and every handoff cycled through the 120s deadline forever.
  
The scoped fix (drain to a dead owner) kept growing better versions of the same guess, so we stepped back and audited the whole protocol by information flow. The structural finding: cancellation was communicated as a bare key deletion, forcing every observer to reconstruct facts the coordinator already owned (is the owner alive? is a successor coming?). Reconstructions are where this protocol's bugs have lived. Two more findings fell out of the same audit: the handoff deadline measured total age and would livelock any partition whose changelog replay legitimately outlives it, and the model checker, while validating the redesign, found a pre-existing acked-write-loss bug unrelated to cancellation.

## Changes

**Cancellation before**

  ```mermaid
  flowchart LR
      D[handoff deleted] --> G{router guesses disposal}
      G --> X[dead owner: drain never converges<br/>loop blocked, cluster wedges]
      class D phYellow;
      class G phGray;
      class X phRed;
  ```

**Cancellation after**

  ```mermaid
  flowchart LR
      P[planner classifies doomed handoff] --> R{one CAS replacement}
      R -- placement re-planned --> S[successor Freezing<br/>stashes ride in, no gap]
      R -- placement stands --> A[reaffirm Complete to live owner<br/>stashes drain home]
      R -- nothing placeable --> F[guarded delete, fail closed]
      N[deletion events are inert:<br/>only post-Complete cleanup]
      class P,S,A phBlue;
      class R phGray;
      class F,N phYellow;
  ```

  - **Containment and supervision** (first commits): stash drains run off the watch loop in per-partition lanes, serialized within a partition, cancellable at batch boundaries, with different-target supersession and same-target absorption. A stall watchdog fails the router's run when the watch loop stops making progress, so registration follows participation instead of just process liveness. Deadline cancellations name the routers whose freeze acks never arrived (`freeze_ack_missing_total{router}`), turning a wedge diagnosis into reading a label. Leader channels gain a connect timeout, since dials to an unassigned pod IP black-hole at TCP connect where the request timeout never starts.
  - **Cancellation by replacement**: a handoff record is never deleted while non-terminal. The planner classifies doomed handoffs (dead new owner, or past the current phase's deadline) and atomically CAS-replaces each with whatever resolves its stashes, clearing predecessor acks in the same transaction. The router's Delete arm is inert. The reaffirm carries `old_owner: None` on purpose: naming the owner on both sides would make the pod derive Released and drop the partition instead of resuming it, a landmine the model surfaced before any code ran.
  - **Phase-aware deadline**: time-in-phase via `phase_entered_at_ms`, not total age. Freezing and Draining wait only on acknowledgements and keep the 120s budget; Warming replays a changelog that scales with the partition and gets `COORDINATOR_WARMING_DEADLINE_SECS` (default 1800s, 0 disables). A total-age deadline can never complete a partition whose replay outlives it: cancel, replan, warm from zero, repeat.
  - **Level-triggered reconciliation everywhere**: the router gains a truth path as an arm of its own watch loop, re-deriving stashes, freeze acks, the table (from the assignment authority), drain disposal, and dial addresses from a fresh snapshot every `ROUTER_RECONCILE_SECS`; the pod re-converges every involved partition, including locally-held state the durable world no longer mentions; the coordinator's tick wakes the planner unconditionally, which is exactly what the stateright model always assumed. A missed event, a silently dead watch stream, or an out-of-protocol raw deletion now costs at most one reconcile interval, never correctness.
  - **The independent safety fix** (`fix(personhog): rebuild partition warms across handoff eras`): the pod's Acquiring convergence skipped warming whenever any warm existed, so a warm from an earlier ownership era could satisfy a new handoff's warming with a stale cutoff, making acked writes beyond it invisible to the new owner forever. Found by the checker in a two-partition crash-plus-rejoin space, reproduced with cancellation disabled, so it predates everything here. Warms now carry provenance and only the current handoff's warm counts; anything else is released and rebuilt.

  > [!NOTE]
  > Documented honestly rather than fixed here: the double-zombie residual is the one known acked-write-loss path in the shipped protocol (a lease-expired router feeding a lease-expired pod, both windows bounded to about a heartbeat tick by self-fencing). The model reproduces the interleaving and validates the fix, broker transactional epoch fencing, which is designed but not yet implemented. It is queued behind the stash rework.

Model-first, then automated suites; no manual testing beyond the earlier dev-cluster diagnosis that motivated the work.

  - The stateright model was extended before production code was written: replacement actions with a shared disposition, warm provenance, three reachability probes, and four new exhaustive configs. The full suite (21 checks) passes in release; the double-zombie residual still reproduces exactly as documented, and `rejoin_two_partitions_without_cancellation` permanently pins the stale-warm counterexample, failing in seconds against the old pod convergence.
  - The three delete-driven integration tests were rewritten to exercise the real path with no synthetic deletion: the coordinator detects a dead new owner, replaces by reaffirm, and the router drains home, at Freezing, Draining, and Warming.
  - New regressions no prior test caught: a raw deletion is inert and the successor resolves the stash; a reaffirm drains the stash back to the owner; a blocked drain cannot stall freeze acks for other partitions; the reconcile pass heals an out-of-protocol deletion and re-asserts lost freeze acks; a stalled watch loop trips the watchdog and deregisters; per-phase deadlines spare a long warm while cancelling a wedged freeze; lane supersession never overlaps drains and same-target requests never restart a healthy one.
  - Full `cargo test` for `personhog-coordination`, `personhog-router`, `personhog-leader`, plus clippy and fmt, all green. The k3s suite runs in CI (resource-bound locally).

## 🤖 Agent context

  Claude Code (Fable 5) authored the changes under my direction, starting from a live diagnosis of the dev wedge. The `/writing-tests` skill shaped test selection. The PR evolved deliberately: the first version fixed the drain-to-dead-owner trigger with a router-side registration check, a reviewer comment ("failed lookup revives dead owner") plus my own pushback exposed that as a reconstruction of coordinator-owned facts, and the redesign replaced the entire decision rather than hardening the guess — the criticized branch no longer exists. The replacement design was modeled in stateright before implementation; the checker rejected the first modeling run and produced the stale-warm counterexample, which turned out to be a pre-existing production bug and became this PR's standalone safety fix. A put-back-on-supersession refinement was implemented and then deliberately reverted: it belongs to the follow-up stash rework that addresses the whole client-visible retry class at once, rather than one producer of it here.


rust/personhog-coordination/src/routing_table.rs
```diff
@@ -418,12 +628,116 @@ impl RoutingTable {
...
+            table
+                .write()
+                .await
+                .insert(assignment.partition, assignment.owner.clone());
+            lanes.request(
+                Arc::clone(handler),
+                router_name.to_string(),
+                assignment.partition,
+                assignment.owner,
+            );
```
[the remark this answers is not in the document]
z0br0wn: this is true - the errors will bounce back to the client to retry until the stash is pointed to the new owner. what's needed here is a bit of a reworking of the stash mechanics as a whole to close this gap completely - and it's one already in progress that will land as a follow up to this.

rust/personhog-coordination/src/routing_table.rs
```diff
@@ -506,11 +820,107 @@ impl RoutingTable {
                     new_owner = %handoff.new_owner,
                     "updated routing table and draining stash to new owner"
                 );
-                handler
-                    .drain_stash(handoff.partition, &handoff.new_owner)
-                    .await?;
+                lanes.request(
```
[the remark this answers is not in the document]
z0br0wn: this is intended, but predicated on the leader being able to fence appropriately. a bigger rework of the stash will be in a follow up pr

rust/personhog-coordination/src/routing_table.rs
```diff
@@ -205,8 +402,10 @@ impl RoutingTable {
             }
         };
 
-        // Abort and await all remaining tasks for clean shutdown
+        // Abort and await all remaining tasks for clean shutdown, and stop
+        // any in-flight drains at their next batch boundary.
         tasks.shutdown().await;
+        lanes_cancel.cancel();
```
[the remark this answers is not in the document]
z0br0wn: addressed in the follow up pr as that introduces the pause mechanics https://github.com/PostHog/posthog/pull/74325

rust/personhog-coordination/src/routing_table.rs
```diff
@@ -357,13 +556,51 @@ impl RoutingTable {
...
+                _ = reconcile_tick.tick() => {
+                    Self::reconcile_pass(
+                        &store,
+                        &table,
+                        &addresses,
+                        &handler,
+                        &lanes,
+                        &router_name,
+                    )
+                    .await?;
```
nickbest-ph: i think `reconcile_pass` failures will propagate up and eventually trigger `signal_failure`, shutting the pod down.

if that's right, with every leader and router pod ticking every 5s, brief etcd blips are going to cause quite a bit of damage across the fleet. if etcd isn't available for 5s we kill every pod in the whole fleet simultaneously.

we could add some metric and warn on failure here and then try to tolerate staleness instead?

rust/personhog-coordination/src/pod.rs
```diff
@@ -596,9 +668,21 @@ impl PodHandle {
...
+        // purpose: convergence stays serialized with event handling.
+        let mut reconcile_tick = tokio::time::interval_at(
+            tokio::time::Instant::now() + self.config.reconcile_interval,
+            self.config.reconcile_interval,
+        );
         loop {
             tokio::select! {
                 _ = cancel.cancelled() => return Ok(()),
+                _ = reconcile_tick.tick() => {
+                    self.reconcile_all().await?;
```
nickbest-ph: same coupling of data plane availability to control plane health as in the router exists here

consider whether or not we need this to be fatal

rust/personhog-coordination/README.md
```diff
@@ -66,7 +66,11 @@ Several failure modes are handled explicitly:
...
+- **Participant stall watchdog**: A router's registration is lease-backed, and the lease keepalive is its own task — so a router whose handoff watch loop has stalled would stay registered (counted in freeze quorums) while never acking. The watch loop stamps a progress marker from a ticker arm inside its own select; a watchdog task fails the run when the stamp goes stale (`ROUTER_PARTICIPANT_STALL_SECS`, default 60 s), deregistering the router on the way down so quorums stop counting it and the process supervisor restarts a healthy participant. Pods deliberately get no such watchdog: their converge loop legitimately blocks for warm-duration, and a pod that stops acking is routed *around* by the phase deadline and replacement instead of being restarted mid-warm.
```
nickbest-ph, line 71: nit: README `ROUTER_PARTICIPANT_STALL_SECS` value doesn't match config.rs value `PARTICIPANT_STALL_SECS`

nickbest-ph (changes requested):
do a check quick to see if single request unavailabilities on the etcd ticks are fatal to router/leader pods and consider tolerating staleness for a certain amount of ticks instead if they are

rust/personhog-coordination/src/routing_table.rs
z0br0wn: this is already happening in dev on current master (the fleet got wiped out on a blip today) - this definitely amplifies it, but regardless once this and my stash prs land my immediate follow up was already to introduce some etcd resilience. for the time being i'll make this non fatal, and still plan on addressing the overall etcd fragility after

nickbest-ph (approved):

rust/personhog-coordination/src/routing_table.rs
z0br0wn: updated!

PR #74093, opened 2026-07-28
fix(data-warehouse): give actionable copy when file upload column detection fails

Gilbert09:
## Problem

In the data-warehouse new-source onboarding flow, users uploading a self-managed file (CSV / JSON / Parquet) sometimes hit a wall when column detection fails. Replay Vision watched real onboarding sessions where users retried the same upload many times, cycling through the same failure without ever understanding what to fix.

The message they saw was `Could not read columns from the uploaded file: <raw ClickHouse error>`. The appended engine error is opaque and doesn't tell a user what to actually change about their file.

## Changes

Swap the raw-error dump for plain, format-specific guidance that names the exact shape the reader expects:

- **CSV** is read as `CSVWithNames`, so it needs a header row and a consistent column count.
- **JSON** is read as `JSONEachRow`, so it must be newline-delimited (one object per line), not a single array.
- **Parquet** gets a generic "make sure it's a valid Parquet file".

The underlying exception is no longer shown to the user but is now sent to error tracking via `capture_exception`, so we keep the same debuggability we had from the interpolated error string.

The hint copy lives next to `FILE_FORMAT_TO_TABLE_FORMAT` in `file_uploads.py`, where the header-row / newline-delimited read assumptions are already documented, so the guidance and the reason for it stay in one place. It's re-exported through the warehouse-sources facade and used by the create-from-upload endpoint.

- Confirmed `FILE_FORMAT_READ_HINTS` covers every entry in `SUPPORTED_FILE_FORMATS`.
- Ran `ruff check` and `ruff format --check` over the three touched files.
- No existing test asserted on the old message string, so nothing regressed there.
- I (Claude) did not exercise the endpoint end-to-end against a live upload.

## Docs update

No docs change: this is an in-product error message only.

## 🤖 Agent context

Found via a Replay Vision pass over real data-warehouse onboarding sessions: CSV/file uploads were a recurring friction point, with users retrying the same failing upload repeatedly against an unhelpful raw-engine error. This is the self-contained copy fix from that pass.

I invoked `/writing-user-facing-copy` for the message wording and kept `/improving-drf-endpoints` in mind, though the change is a pure error-string swap with no request/response schema impact. I deliberately scoped this to copy only, and did not touch upload validation, supported formats, or the read-format mapping.

---


Gilbert09:
Heads-up for whoever merges this: the one red required check (`Django tests – Core … (8/20)`) is failing on a test unrelated to this diff.

`posthog/api/test/test_team_project_parity.py::test_viewset_actions_are_a_superset_of_team_viewset` fails with:

> Unexpected project-only actions: `['default_evaluation_contexts', 'default_release_conditions']`. If intentional, add them to `PROJECT_ONLY_ACTIONS`.

Those are project-viewset actions from recent master changes (migrations `0975`–`0977` and #73294), and `PROJECT_ONLY_ACTIONS = {"change_organization"}` on `origin/master` doesn't yet include them, so the test is red on master itself. This PR only touches three data-warehouse upload files and adds no viewset actions, so it isn't the cause. It'll go green once master's parity allowlist is fixed and this branch picks it up. The other cancelled/failed shards earlier were self-hosted-runner timeouts, since re-run.

PR #74108, opened 2026-07-28
feat(signals): add self-validation follow-up runs to scout harness

andrewm4894:
## Problem

Scouts surface findings, teams ship fixes, and then nobody checks whether the fix actually worked. The canonical `signals-scout-inbox-validation` scout closes part of that loop by re-measuring recently resolved inbox reports, but it may not be enabled on a team, and its watched surface is limited to resolved reports: signal-channel findings and recommendation-shaped work never become a resolved report it could re-measure.

Custom scouts already have a self-improvement channel (evidence-backed `improve:` scratchpad suggestions). This PR gives every scout the analogous self-validation channel: follow up on your own past work and confirm it was actually fixed or improved.

## Changes

All in the scout harness prompt - no runner, model, or config changes. The validation cadence is deliberately the scout's own judgment, not a harness schedule.

**Prompt (`scout_harness/prompt.py`)**

- A shared "Follow up on your own past work" section on both channels and both origins. Scouts record `followup:<skill-name>:<entity>` scratchpad entries for work whose fix is measurable later. Each entry carries what was surfaced, the exact probe that confirms the fix, the baseline number, and a validate-after date (deploy + soak time).
- The scout decides, run by run, whether to spend the run validating that queue. The section carries the decision criteria (due entries accumulated, a while since the queue was last worked, a dismissal note saying a fix shipped) but no trigger - there is no fixed schedule and no harness-side cadence. A validation pass is self-reported in the close-out summary and in the touched queue entries, which is also how future runs know when the queue was last worked.
- Verdict rules per due entry: fix held is quiet memory maintenance (update or forget the entry, no "it worked" output); fix didn't hold is a real finding, re-surfaced through the scout's normal channel with a clause matched per-tool to the same fail-closed rule as the existing channel sections (a signal scout is pointed at `scout-emit-signal`, an emit-only report scout is never pointed at `scout-edit-report`, and so on).
- The section defers resolved-report re-measurement to the canonical inbox-validation scout when the `scout_fleet` roster shows it actively running, so the two loops stay complementary.
- Queue entries are treated as team-shared data, not trusted memory: the scout is told to check `created_by_skill` and re-derive, rather than execute, a probe from an entry its own runs didn't write (raised by a security review comment - the scratchpad is writable by any scout on the team).

- Extended the two prompt-tail tests to assert the shared follow-up section rides in both tails (they're assembled by separate code paths, so either can lose it independently) and that it states the scout-owned cadence.
- New parameterized prompt test over all four channel variants asserting the section's re-surface clause names a path the scout actually holds. The wrong-tool half of the fail-closed rule is already policed by the existing emit-only/edit-only tests, which assert the unheld tool appears nowhere in the whole prompt - the section now falls under those assertions automatically.

I could not run the Python suite in this remote container (the repo pins Python 3.13.13 and only 3.13.12 is available, so the venv won't sync and `hogli` / `ci:preflight` can't run) - the tests run in CI. I verified all four section variants render correctly (right tool named, forbidden tools absent, template placeholders fully resolved) with a standalone script exec'ing the Django-free parts of `prompt.py`, and ruff check/format pass.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Harness behavior docs updated in `products/signals/backend/scout_harness/AGENTS.md` in this PR. No user-facing docs affected.

## 🤖 Agent context

Authored by Claude Code from a request to add a self-validation feature for all scouts, mirroring how custom scouts do self-improvement reports: keep a follow-up list in the scratchpad and periodically dedicate a run to it, since the canonical inbox-validation scout may not be enabled. Skills invoked: `/writing-tests` and `/improving-drf-endpoints`.

Design history reviewers should know:
- The first revision had a harness-driven deterministic cadence (every Nth run per team/skill lane became a validation run, stamped `run_focus: "self_validation"` in the run metadata and on lifecycle events). Andy redirected: the scout should own the decision. This revision removes the runner/limits/serializer changes entirely and moves the cadence into the scout's judgment, which also mooted two Greptile P1s against the cadence helper and fixed a mypy failure in a since-deleted test.
- Trade-off accepted with that redirect: no deterministic `run_focus` stamp for analytics. Validation passes are self-reported in the close-out summary instead. A follow-up idea (Andy's): let the scout set run metadata itself for observability - the natural seam is extending `SignalScoutRunSummary` (the structured close-out the harness already parses and persists) with optional self-reported fields the harness copies into the run row's `metadata` at finalize.


andrewm4894:
Re the two review findings on `1ddbedc2`:

- **Cold follow-up stranding**: applied in `f42f24c5` - the handoff to the inbox-validation scout now excludes reports resolved before that scout's ~14 day intake window, so a due follow-up on an older resolved report stays with the scout that recorded it.
- **Treat queue state as untrusted**: declining this one at the prompt layer. The scratchpad is deliberately co-writable across the fleet (steering and cross-scout memory depend on it), and a compromised sibling that can rewrite an entry's state header can just as easily `forget` the entry outright - no amount of in-prompt re-verification of scheduling metadata defends against a writer with full mutation rights, it only burns every run's budget re-deriving state that is usually fine. The prompt already treats entry *content* (probes) as untrusted and re-derives from live sources. The real fix for state integrity is structural - a server-side write gate reserving `followup:<skill>:*` keys to the owning skill on the scratchpad surface - which is beyond this prompt-only PR; flagging it as a possible follow-up alongside the planned scout-set run metadata work.

---

PR #74117, opened 2026-07-28
perf(hogvm): cut transformation execution ~35% in prod, ~2x on geoip

jose-sequeira:
## Problem

PostHog ingestion runs user-defined Hog transformations on every event, and the highest-volume one is the geoip template. On the Rust HogVM (the executor behind `CDP_HOG_RUST_VM_EXECUTION_ENABLED`) it ran meaningfully slower than the legacy native-JS plugin it replaces. Profiling showed the cost was the interpreter's per-step allocation churn, not the mmdb lookup. This PR is a series of allocation- and dispatch-focused optimizations to the VM crate and its napi addon.

## Changes

Six independent optimizations, each measured in isolation on a canonical geoip harness (interleaved A/B, kept only when it beat a 2% gate). They compound to roughly a 2x speedup on the geoip workload.

| Change | What it does |
| --- | --- |
| Box large `HogLiteral` variants | Boxes the `Object` / `Callable` / `Closure` payloads so `HogValue` shrinks from 120 to 32 bytes. Every stack push/pop/clone and heap emplacement moves a quarter of the bytes it used to. |
| `HogStr` string payload | Gives `HogLiteral::String` a two-arm payload: `Owned(String)` for computed strings, `Shared(Arc<str>)` for constants sliced out of the pre-decoded token stream. String-constant pushes become a refcount bump instead of a malloc+copy. Equality is content-based across arms. |
| ahash object maps | Hashes object maps with `ahash` instead of SipHash. Still a keyed, DoS-resistant hash (object keys come from user programs), several times faster on the short keys the property-write loop hammers. |
| In-place object emplacement | `walk_emplacing` rebuilt every native-function object result via `collect` (fresh map, re-hash, re-insert). It now walks children in place through `values_mut()`, keeping the original allocation. The geoip record has ~10 nested sub-objects per event. |
| Cached chunk token slice | The per-opcode fetch went through a symbol match plus, for module code, a two-`String` `Symbol` hash and table lookup. The VM now caches the current chunk's decoded token slice and the fetch is a bounds-checked index. |
| jemalloc in the addon | Sets jemalloc as the addon's global allocator; the interpreter's small-allocation churn was the dominant profile cost under glibc malloc. Built with `disable_initial_exec_tls` rather than via `common-alloc`: node dlopens the cdylib late, and jemalloc's default initial-exec TLS fails the load with "cannot allocate memory in static TLS block". This was the single largest win. |

The napi addon also gains a program registry (`registerProgram` / `executeRegisteredSync`) that validates and token-decodes a hogFunction's bytecode once and executes by handle, skipping the per-event JS→Rust marshal + copy + decode.

Semantics are unchanged throughout - the TS reference VM defines behavior, and the crate's parity suites are the gate.

### Measured in production

The branch image ran on the EU main analytics ingestion lane. Old and new pods overlapped during the rollout, so the comparison below is the same traffic in the same window (`hogvm_rust_execution_duration_ms`, all transformations mixed):

| | before | after | change |
| --- | --- | --- | --- |
| p50 | 0.34 ms | 0.23 ms | −34% |
| p90 | 0.76 ms | 0.48 ms | −36% |
| mean | 0.39 ms | 0.29 ms | −26% |

The settled fleet holds the same numbers. Execution outcomes stayed flat: ~0.003% errors on both images, zero fallbacks, and the addon loads cleanly with jemalloc.

> [!NOTE]
> Two rollout notes for whoever ships this. The addon now sets jemalloc as the Rust global allocator inside the plugin-server process (already the standard across PostHog Rust services, but the addon's memory footprint will shift, so RSS is worth watching). And wiring the program registry into `nodejs/src/cdp/hog-transformations/rust-vm-executor.ts` is still a follow-up, not part of this PR.

Automated only - no manual testing was done. On every optimization the following ran, and stayed green:

- `cargo test -p hogvm` (77 tests - stl / vm / parity suites)
- `cargo test --features noop` in the addon workspace (19 tests)
- `cargo check -p cymbal -p cohort-core` (the two crates that depend on this one; public API stayed additive)
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo shear` in both workspaces

Performance was measured with an interleaved A/B harness (candidate vs the prior HEAD binary, back to back, several rounds, comparing per-round ratios rather than absolute numbers since the runner hardware drifted between sessions). Each committed change cleared a 2% improvement gate; changes that didn't were reverted. The benchmarking scaffolding used to produce those numbers has been removed from the branch - this PR is only the shipping VM adaptation.

Production validation: the "Measured in production" table comes from the `hogvm_rust_execution_duration_ms` and `hogvm_rust_execution_total` series in prod-EU Grafana, with the running image confirmed against this PR's head commit.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

José directed this work; I (Claude) did the implementation, measurement, and cleanup. It ran as an iterative optimization loop: each iteration profiled the geoip path, formed one hypothesis, implemented the smallest change for it, gated on the full correctness suite, then measured an interleaved same-machine A/B and kept the change only if it beat 2%.

A few things were tried and rejected, worth surfacing so they aren't re-attempted:

- Making all VM strings `Arc<str>` (rather than the two-arm `HogStr`) was a ~9% regression on x86/glibc - the atomic refcount traffic on the VM's constant string clones cost more than glibc's small-alloc fast path saved, and it added copies on the owned-string paths. The `HogStr` split exists specifically to get the constant-push win without that downside.
- Rewriting `ok_or_else(|| VmError::...)` to eager `ok_or(...)` (a clippy suggestion) was a ~3% regression on the token-fetch hot path, since it constructs the large `VmError` enum on every successful fetch. The lint is suppressed crate-wide with a comment explaining why.
- A standalone dealloc-free `CallGlobal` symbol probe measured ~1%, under the gate, so it was reverted.

No repo skills applied to this Rust crate. The final commit strips the benchmarking harness (perf docs, fixtures, the profile binary, and a dev bench script) so the diff is just the VM changes plus three dependency additions (ahash, common-alloc, lockfiles). A later session validated the prod-EU rollout through the Grafana MCP, comparing old and new pods side by side during the rollout; `/writing-pr-descriptions` shaped this body.

https://claude.ai/code/session_01SrCAzhbacGX1cdb3Kcx1bi

---
_Generated by [Claude Code](https://claude.ai/code/session_01SrCAzhbacGX1cdb3Kcx1bi)_


rust/common/hogvm/src/program.rs
pl, line 1: nit: Asked the bot to review tests for gaps, found this, which sound like they could be useful to cover:

What is covered

- Token::Int under both opcodes. Every parity program and the numeric coercion suite push integer constants. large_literals.rs pushes 500 of them under INTEGER.
- Token::Float under the FLOAT opcode. numeric_coercion.rs uses fractional values like 2.5 and whole-number floats like 1.0. The whole-number case matters because json!(1.0) serializes as the float variant, so it exercises the branch where as_i64 returns None and as_f64 picks it up. The equality assertions against integer 1 confirm the value survives as a float that compares equal.
- Token::Str, Token::Null, Token::Bool are hit constantly by the parity fixtures.

What has no coverage

- Token::from_json itself. There are no unit tests in program.rs, so the decode function is only tested indirectly through execution.
- INTEGER opcode with a non-integer operand. The comment in vm.rs says the SQL-AST compiler emits INTEGER with a boolean operand, and that is the stated reason the arm matches every token variant. No test does this. The bytecode examples fixture has zero occurrences of the INTEGER opcode followed by true or false, and that fixture is only parsed as JSON in ops.rs anyway, never executed. Same for INTEGER with a float operand.
- Integers above i64::MAX. Nothing in the tests or fixtures contains a 19-digit number. This is the one place the decode differs from master's serde path, since master's next::<usize>() would accept a u64 that the new next_usize rejects. It is a practically unreachable divergence but it is untested.
- Token::Json for numbers. Unreachable without arbitrary_precision, so no test is possible or needed.
- FLOAT opcode with an integer operand. The fixture has one such line but it is never executed. next_f64 accepts Token::Int, matching master's serde behavior, but nothing asserts it.

rust/common/hogvm/src/vm.rs
```diff
@@ -835,26 +843,82 @@ impl<'a> HogVM<'a> {
...
     }
 
-    fn next<T>(&mut self) -> Result<T, VmError>
-    where
-        T: DeserializeOwned + Any,
-    {
-        let next = self.context.get_bytecode(self.ip, &self.current_symbol)?;
+    // Change the chunk being executed, keeping `current_tokens` in lockstep with
+    // `current_symbol`. Every assignment of `current_symbol` must go through here.
+    fn set_chunk(&mut self, symbol: Option<Symbol>) -> Result<(), VmError> {
```
pl, line 848: nit: `set_chunk` is now load-bearing for instruction fetching (the cached slice is what `next_token` indexes), but only two of its five call sites are exercised. `calling_hog_stl.hog` and `icontains.hog` cover call into a Hog-written STL fn and return. The catch-frame switch (`set_chunk(frame.catch_symbol)` in the `Throw` arm) and the resume switch (`set_chunk(chunk_to_symbol(&active.chunk))`) are reached by nothing in this crate, and neither cymbal nor cohort-core tests ever call a Hog-written STL function.

Two tests would close that:

1. A parity fixture that throws inside a lambda invoked by `arrayMap`, catches in the root program, and then calls a root function so a wrong chunk after unwind fails loudly:

```
let caught := false;
try {
  arrayMap(x -> throw Error('boom'), [1, 2]);
} catch (e) {
  caught := e.message = 'boom';
}
fun after() { return true; }
return caught and after();
```

2. A test in `tests/async_resume.rs` (via `run_injecting`) that suspends from inside an STL frame, so the snapshot carries a non-root chunk on the call stack:

```
return arrayMap(x -> asyncFetch(x) + 1, [1, 2])
```

Every resume test today snapshots with only root frames.

_— written by Claude, posted on Paweł's behalf_

rust/common/hogvm/src/context.rs
```diff
@@ -255,7 +254,34 @@ impl ExecutionContext {
...
+                .symbol_table
+                .get(symbol)
+                .ok_or_else(|| VmError::UnknownSymbol(symbol.to_string()))?
+                .body_tokens()),
+            None => Ok(self.program.body_tokens()),
+        }
+    }
+
+    /// The pre-decoded form of `get_bytecode` — the interpreter's per-step fetch path.
+    pub fn get_token(&self, ip: usize, symbol: &Option<Symbol>) -> Result<&Token, VmError> {
```
pl: nit: `get_token` has no callers. The VM only uses `chunk_tokens`, and `debug_step` still goes through `get_bytecode`. Looks like an intermediate step before the slice cache landed; can it be removed?

_— written by Claude, posted on Paweł's behalf_

pl (approved):
Left a couple of minor things with the bot, looks good to ship.

rust/common/hogvm/src/context.rs
jose-sequeira: Removed in 9b0ce77ca40, along with the underlying Program/ExportedFunction get_token methods it was the only caller of.

rust/common/hogvm/src/vm.rs
jose-sequeira: Added both in 35278e08f8f: a chunk_switch_unwind test program that throws from a lambda under an stl/arrayMap frame and catches in root (throw is a statement in Hog, so the lambda calls a small boom() fn), and stl_frame_suspend_round_trips in async_resume.rs, which suspends inside the lambda while the stl/arrayMap frame is on the call stack, so the snapshot round-trips a non-root chunk and the resume returns into module code.

rust/common/hogvm/src/program.rs
jose-sequeira: Covered the reachable gaps in 35278e08f8f (tests/token_decode.rs): INTEGER with a bool operand, INTEGER with a float operand, an integer above i64::MAX (decodes as float), and FLOAT with an integer operand. Left Token::Json numbers out since they are unreachable without arbitrary_precision.

PR #74120, opened 2026-07-28
perf(hogvm): register transformation programs once instead of per event

jose-sequeira:
> [!NOTE]
> Stacked on #74117 — base is `jose-sequeira/rust-vm-geoip-loop`, so this diff is only the registry wiring. Merge #74117 first and this retargets to master.

## Problem

#74117 made the Rust HogVM interpreter roughly 2x faster on the geoip transformation, but the executor still hands the addon a hog function's entire bytecode array on **every event**. That array gets marshalled across the napi boundary, copied, validated, and token-decoded per invocation. For a transformation running on every ingested event that is the same few hundred tokens decoded over and over.

The addon already had `registerProgram` / `executeRegisteredSync` to fix this, but nothing called them, they were never declared in the addon's type definitions, and wiring them up as-is would have leaked.

## Changes

**Executor now registers each program once and executes by handle.** The per-event cost drops to the globals crossing.

Three things had to change to make that safe:

**The addon registry leaked.** It was an append-only `Vec` where the handle is the index, so every re-registration of an edited function permanently stranded a slot holding a decoded token stream. It's now a slot map with a free list plus a `releaseProgram` binding, so freed slots get reused. Releasing an unknown or already-released handle is a no-op, so a double release can't push the same handle onto the free list twice and alias one slot between two live registrations.

**A cache keyed on function id alone would serve stale bytecode.** The executor keys on `id` + `updated_at`, so an edited transformation re-registers instead of silently running its pre-edit program.

**The cache is LRU, bounded at 500 entries**, using the `lru-cache` already depended on in this package. Eviction order matters above the cap: dropping by registration order would evict the function that runs on every event just because it was registered first, then re-register and re-evict it in a loop. `dispose` is the single owner of releasing handles — it fires on both eviction and the explicit delete on version change — so no caller releases directly and a handle can't be released twice.

One deliberate degradation: a hog function with no `updated_at` has no version to key on, and serving it a cached handle could run outdated bytecode after an edit. Those execute through the old unregistered `executeSync` path instead. Correct beats fast.

A `hogvm_rust_program_registrations_total{reason}` counter distinguishes first-time registrations from re-registrations, so a function thrashing its cache entry is visible.

**Rust — ran locally, all green.** 22 addon tests (19 existing + 3 new), plus fmt and clippy. The three new tests each pin a specific regression in the registry rewrite:

- released slots are reused, and the reused slot holds the *new* program (catches the leak, and catches a reuse bug that would run the released program)
- executing a released handle errors rather than running whatever now occupies the slot
- a double release doesn't hand the same slot to two live registrations

They share the process-global registry, so they take a mutex — cargo runs tests in parallel threads and they assert on which slot a registration lands in.

**TypeScript — CI is the verifier.** I could not install `nodejs/node_modules` in my sandbox (`pnpm install` 403s fetching the uWebSockets.js tarball), so neither jest nor tsc ran locally. The first push failed the build on exactly that gap — the addon's hand-maintained `index.d.ts` never declared the registry functions, so the test's mock didn't typecheck. That is fixed here; treat any further red as unverified-by-me rather than a surprise.

The 5 jest cases cover: handle reuse across events (the point of the change), re-registration plus release when `updated_at` changes (the stale-bytecode regression), LRU keeping a hot function resident while an idle one is evicted (the eviction-order regression), the release-at-cap bound, and the no-version fallback to `executeSync`.

No manual testing — I have not run this against a real ingestion process.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

José asked for this as a follow-up to #74117, where I had flagged the registry wiring as out of scope. The reason it was out of scope is the leak: the registry as written could not be wired in safely, so this PR is the fix plus the wiring rather than just the hookup. The LRU came from José pushing back on my first cut, which evicted in insertion order.

Design notes worth a reviewer's attention. I considered keying the registry by string on the Rust side (`${id}:${updated_at}`) to sidestep index recycling entirely, and rejected it — passing a string per event partly re-introduces the marshalling this change exists to remove, and the slot map with a no-op double release closes the aliasing hazard. On the cache side I delete-then-set rather than overwrite, so the behavior doesn't depend on whether `lru-cache` fires `dispose` for an in-place `set` (a semantic I couldn't verify locally without the dependency installed). The 500-entry cap is a judgement call, not a measured number; worth a second opinion if you think a process can see more distinct transformations than that.

The benchmark harness that would have quantified the win was removed at the end of #74117, so I have no in-repo measurement for this change. The argument here is structural (decode once instead of N times), not measured.

https://claude.ai/code/session_01SrCAzhbacGX1cdb3Kcx1bi


jose-sequeira:
Noting the red `semgrep-general` / `Semgrep Checks Pass` here, because it is not from this change and I am deliberately not fixing it.

All 7 blocking findings are `unsafe-add-mask-workflow-command` in GitHub Actions YAML:

```
.github/workflows/hogbox-preview-env.yml:482, :767
.github/workflows/pr-closed.yml:100
.github/workflows/hogbox-preview-cleanup.yml:67
```

This PR touches no workflow files — the diff is `rust/common/hogvm/node/**` and `nodejs/src/cdp/hog-transformations/**`. Those exact lines are present on `master` right now, so the findings predate the branch.

Why it shows up here but not on the base PR (#74117, where semgrep is green): the scan ran full-repo (`Targets scanned: 1524`) rather than diff-scoped. The likely cause is this PR's base being a branch rather than `master`, so the baseline diff doesn't resolve and semgrep falls back to scanning everything, surfacing pre-existing findings a diff-scan would filter out. I'd expect it to go green on its own once #74117 merges and this retargets to `master`.

Fixing the workflow findings from a HogVM perf PR would be unrelated scope, and workflow edits hit every in-flight PR immediately, so that belongs in its own change.

---

jose-sequeira:
CI is red on `Node.js parallel Tests 1/1`, and the failure does not come from this PR.

Failing test, in `tests/ingestion/ingestion-e2e.test.ts:3619`:

```
● Event Pipeline E2E tests (fold+groupBatchCreates=true) › we do not alias users if distinct id changes but we are already identified

    expect(received).toBe(expected) // Object.is equality

    Expected: 2
    Received: 3

      3617 |                 const persons = await fetchPostgresPersons(infra.postgres, team.id)
      3618 |                 // Still 2 separate persons — merge was rejected because both are identified
    > 3619 |                 expect(persons.length).toBe(2)
```

That is a `waitForExpect` on a count of Postgres person rows in the identify/merge path. This PR touches only the HogVM program registry — `rust-vm-executor.ts`, `rust-vm.ts`, `index.d.ts`, `lib.rs`, and the executor's own test file — and nothing in ingestion or person merging. Trunk already has a [flaky-test record](https://app.trunk.io/posthog-inc/flaky-tests/test/b20b2aad-a4bb-503b-866c-6a364b02c8a1?repo=PostHog%2Fposthog) for it; it just was not quarantined on this run, so the `Fail on test failure` step failed the job. The `Test with Jest` step itself passed, at 11337/11338.

Everything else on `f2ebe8e` is green: all three serial shards, Node.js Build, Node.js Code quality, the Rust ingestion consumer e2e, Rust tests, and Semgrep. The PR's own suite passes locally, 22/22 in `rust-vm-executor.test.ts`.

I have re-run the failed jobs and will keep watching until the PR is green.

---

jose-sequeira:
#74117 merged, so this PR auto-retargeted to `master` and is now conflicted. I tried to resolve it and stopped, because the conflict is a real design collision rather than a mechanical one. Not pushing a guess.

**The mechanical part is easy.** This PR's history still carries #74117's pre-merge commits, which the merge queue squashed into `c851fc01`. The resolution is to replay only the four registry commits onto current `master`. Three of them (`register transformation programs once`, `declare the registry bindings and evict by LRU`, `fall back when the addon has no registry bindings`) cherry-pick clean.

**The fourth one conflicts, and picking either side loses something.** `perf(hogvm): route batched executions through registered programs` conflicts with `rust-vm-batch-scheduler.ts` in `master`, over how the scheduler keys its queues:

- `master` keys by a **sha256 hash of the bytecode content**, cached per array instance in a `WeakMap`. Its comment is explicit about why: *"Every team has its own HogFunction row, so one template (GeoIP on thousands of teams) arrives as thousands of distinct arrays with identical content; keying by instance would keep each team in its own batch."*
- The batched-execution commit makes the scheduler generic over `P` and keys by **object reference**, passing the `hogFunction` object itself: `this.scheduler.execute(invocation.hogFunction, invocation.state.globals)`.

One queue per `hogFunction` row means one queue per team. Applied on top of `master`, that shrinks GeoIP batches from "every team at this instant" to "one team at this instant" — on the single highest-volume transformation in ingestion, which is the workload both PRs exist to speed up. Nothing fails; only `hogvm_rust_batch_flush_size` would show it.

Worth noting the ordering: the content-hash keying is new, added while #74117 was being rewritten before merge. The batched-execution commit predates it, so it reverts that keying without having seen it, rather than overriding it deliberately.

**Suggested resolution, for a second opinion before I implement it:** key the registry by bytecode content hash instead of by `id` + `updated_at`. Teams sharing a template then share one handle, which keeps `master`'s cross-team batching and amortizes the token-decode across all those teams instead of repeating it per team — strictly better than what this PR currently does. It also drops the `updated_at` versioning entirely, since the content hash *is* the version, which removes the deliberate degradation this PR documents (a hog function with no `updated_at` falling back to `executeSync`).

That is a design change, not a conflict resolution, so I would rather have agreement than land it inside a merge fix. Tell me which way to go and I will do the rebase and push.

---

PR #74124, opened 2026-07-28
feat(personhog): add lifecycle saga tables and is_deleted columns

jose-sequeira:
## Problem

Person deletion in personhog is moving from hard deletes to a durable saga. That needs two schema pieces on the persons database:

1. State tables for lifecycle operations (delete now, merge later): one row per operation with step, lease, frozen request, and recorded outcome, plus one row per person the operation touches. Living on the same primary as the person rows lets each Postgres step commit its work and its step advance in one transaction.
2. Soft-delete tombstones on person and distinct id rows. Hard-deleting a person row means a later re-creation of the same key restarts at version 0 and loses to its ClickHouse tombstone forever. Keeping the row (`is_deleted = true`, version bumped) turns re-creation into a revival upsert that floors the version above the tombstone.

This PR is schema only — nothing writes tombstones or saga state yet. Tombstone-aware reads and revival land in the stacked follow-up.

## Changes

- New `lifecycle_op` and `lifecycle_op_person` tables with sweep, GC, and mark indexes (`rust/persons_migrations/20260727000001`). The partial unique mark index is the claim primitive: at most one live op per person, a unique violation is the conflict.
- `is_deleted BOOLEAN NOT NULL DEFAULT false` on `posthog_person`, `posthog_persondistinctid`, and `personhog_person_tmp` (`20260727000002`). ADD COLUMN with a constant default is metadata-only on PG 11+; the remaining ACCESS EXCLUSIVE lock wait is bounded by `SET LOCAL lock_timeout = '2s'`, and the idempotent file retries on the next run if it times out.
- Django `Person` / `PersonDistinctId` get the matching state-only field. Both models are `managed = False`, so there is no Django migration.
- nodejs person write paths now return explicit columns instead of `RETURNING *`, so the new personhog-only column does not leak into ingestion's `InternalPerson` objects (ingestion stays unaware of it).
- Deliberately no partial indexes: the `(team_id, distinct_id)` unique index stays total so a re-created distinct id conflicts with its tombstone and can take the revival path.

- Migrations applied to the local persons database (idempotent re-run verified).
- `npx jest src/common/persons/repositories/postgres-person-repository.test.ts` on the stacked branch: 108 passed. Without the explicit-columns change, two existing tests fail on this schema (deep-equality between `RETURNING *` results and explicit-column fetches) — that is the regression the change guards.
- Rust personhog storage tests pass on the stacked branch (13, including the new revival scenarios).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Internal architecture change, no user-facing docs.

## 🤖 Agent context

Authored with Claude Code, directed by @jose-sequeira. Skills invoked: /django-migrations and /clickhouse-migrations conventions were reviewed in the earlier session that authored the SQL; /writing-tests gated the test additions on the stacked branch. Decisions: schema deviates from the internal design doc where the real tables do (`team_id INTEGER`, `created_at` kept for ops visibility); a shield index was dropped because the tombstone row itself is the durable shield. The nodejs explicit-columns change rides in this PR (not the follow-up) because the columns alone break nodejs CI deep-equality tests the moment they merge.


rust/persons_migrations/20260727000002_add_is_deleted_to_person_tables.sql
```diff
@@ -0,0 +1,36 @@
...
+-- — no table rewrite, no backfill, no NOT NULL validation scan, regardless of
+-- table size. The remaining hazard is the brief ACCESS EXCLUSIVE lock: on a
+-- hot table the danger is queueing for it behind a long-running query while
+-- everything else queues behind us. lock_timeout bounds that wait; on timeout
+-- the per-file transaction aborts, nothing is recorded, and the next
+-- migration run retries this idempotent file.
+
+SET LOCAL lock_timeout = '2s';
+
+ALTER TABLE posthog_person
```
nickbest-ph, line 26: i think we should consider a separate "bridge" table that stores the death version linked to the id, rather than soft deletion tombstones sticking around in these tables and won't require having to add the is_deleted flag to every query that exists.

this also lets us set a new id on for the distinct id version of this table that can be a hash of the id, obscuring the PII that is the distinct id.

nickbest-ph (approved):
lgtm

PR #74138, opened 2026-07-28
chore(insights): make refreshed quill chart styles the default

sampennington:
## Problem

`quill-chart-style-refresh` is at 100% rollout with no property filters, so the refreshed chart styling is already what everyone sees. The flag is just a branch nobody is on the false side of.

## Changes

Deleted the flag and the `useChartStyleRefreshEnabled` hook. `useChartTheme` and `useChartConfig` now apply unconditionally what used to be the flag-on defaults: monotone curves, axis lines, tick marks, faint dashed grid, crosshair, rounded bar corners. Caller overrides still win.

The "Line style" option in the insight display menu was gated on the same flag, so that gate is gone too.

Renamed `REFRESHED_CONFIG_DEFAULTS` to `CHART_CONFIG_DEFAULTS` and `refreshedThemeOverrides` to `chartThemeDefaults`, since "refreshed" no longer distinguishes anything.

> [!NOTE]
> No behavior change for users — the flag was fully rolled out. Storybook and Jest run with flags off, though, so the visual baselines were captured on the pre-rollout styling. The Visual Review diffs are the snapshots catching up to what production already renders, not a new look.

`InsightDisplayConfig.test.tsx` caught the one real consequence: "Line style" is now an always-present menu section for trends line/area, retention, and stickiness. Updated those four `it.each` expectation rows. No new tests — a flag removal with nothing left on the disabled path doesn't have a regression to name that the existing suite misses.

I (actually Claude) could not run jest, lint, or typecheck locally — `node_modules` isn't installed in this environment — so those are verified by CI rather than by me. Grepped the repo to confirm no reference to the flag survives.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed.

## 🤖 Agent context

Claude via PostHog Code. I asked it to find the chart style flag in `useChartTheme` / `useChartConfig` and delete it. Skills invoked: `sp-ship` (and its `sp-swarm` / `sp-triage` sub-skills) from the PostHog skills store, plus `/writing-tests` before touching the test file.

The swarm review raised one finding as a merge blocker: `chartThemeDefaults` hardcodes rgba values that shadow the `--color-graph-*` design tokens. That's a real problem, but it's pre-existing at 100% rollout, not something this PR introduces — see the open thread for the reasoning and why the token cleanup wants its own PR (both tokens are shared with the two remaining Chart.js graphs, and `Sparkline`/`Metric` bypass this hook entirely). Left open as a record, not a gate.

---

### DIFF
--- a/frontend/src/lib/charts/hooks.ts
+++ b/frontend/src/lib/charts/hooks.ts
@@ -9,7 +9,7 @@ import { themeLogic } from 'lib/logic/themeLogic'
 
 import { buildTheme } from './utils/theme'
 
-const REFRESHED_CONFIG_DEFAULTS = {
+const CHART_CONFIG_DEFAULTS = {
     curve: 'monotone',
     showAxisLines: true,
     showTickMarks: true,
@@ -18,7 +18,7 @@ const REFRESHED_CONFIG_DEFAULTS = {
     barCornerRadius: 4,
 } as const
 
-function refreshedThemeOverrides(isDarkModeOn: boolean): Partial<ChartTheme> {
+function chartThemeDefaults(isDarkModeOn: boolean): Partial<ChartTheme> {
     return {
         gridColor: isDarkModeOn ? 'rgba(255,255,255,0.06)' : 'rgba(0,0,0,0.06)',
         gridDashPattern: [3, 3],
@@ -28,23 +28,13 @@ function refreshedThemeOverrides(isDarkModeOn: boolean): Partial<ChartTheme> {
     }
 }
 
-export function useChartStyleRefreshEnabled(): boolean {
-    const { featureFlags } = useValues(featureFlagLogic)
-    return !!featureFlags[FEATURE_FLAGS.QUILL_CHART_STYLE_REFRESH]
-}
-
 /** Theme for app quill charts. `buildTheme()` reads CSS variables from the DOM, so the memo keys on
- *  `isDarkModeOn` to re-read them when the app theme flips. Behind `QUILL_CHART_STYLE_REFRESH` it
- *  also applies the refreshed chart colors (faint dashed grid, muted axis lines); caller `overrides`
- *  win over both. Pass a stable (memoized or module-level) `overrides` object — a fresh object every
- *  render defeats the memo. */
+ *  `isDarkModeOn` to re-read them when the app theme flips. It also applies the app chart colors
+ *  (faint dashed grid, muted axis lines); caller `overrides` win over those. Pass a stable (memoized
+ *  or module-level) `overrides` object — a fresh object every render defeats the memo. */
 export function useChartTheme(overrides?: Partial<ChartTheme>): ChartTheme {
     const { isDarkModeOn } = useValues(themeLogic)
-    const refreshEnabled = useChartStyleRefreshEnabled()
-    return useMemo(
-        () => buildTheme({ ...(refreshEnabled ? refreshedThemeOverrides(isDarkModeOn) : {}), ...overrides }),
-        [isDarkModeOn, refreshEnabled, overrides]
-    )
+    return useMemo(() => buildTheme({ ...chartThemeDefaults(isDarkModeOn), ...overrides }), [isDarkModeOn, overrides])
 }
 
 /** The single rollout gate for chart drag-to-zoom, applied inside `useDateRangeZoom` so every
@@ -83,20 +73,18 @@ export function useDateRangeZoom(
 }
 
 /** Drop-in replacement for the `useMemo` that builds a chart's config object. On top of memoizing,
- *  it applies app-level rendering defaults — currently the refreshed style (monotone curve, axis
- *  lines, tick marks, crosshair, grid) behind `QUILL_CHART_STYLE_REFRESH`. Keys the config sets
- *  explicitly (non-undefined) always win over the defaults. */
+ *  it applies app-level rendering defaults (monotone curve, axis lines, tick marks, crosshair,
+ *  grid). Keys the config sets explicitly (non-undefined) always win over the defaults. */
 export function useChartConfig<T extends object>(factory: () => T, deps: DependencyList): T
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined {
-    const refreshEnabled = useChartStyleRefreshEnabled()
     // eslint-disable-next-line react-hooks/exhaustive-deps
     const config = useMemo(factory, deps)
     return useMemo(() => {
-        if (!refreshEnabled || !config) {
+        if (!config) {
             return config
         }
         const defined = Object.fromEntries(Object.entries(config).filter(([, value]) => value !== undefined))
-        return { ...REFRESHED_CONFIG_DEFAULTS, ...defined } as T
-    }, [refreshEnabled, config])
+        return { ...CHART_CONFIG_DEFAULTS, ...defined } as T
+    }, [config])
 }
--- a/frontend/src/lib/charts/hooks.ts
+++ b/frontend/src/lib/charts/hooks.ts
@@ -29,9 +29,8 @@ function chartThemeDefaults(isDarkModeOn: boolean): Partial<ChartTheme> {
 }
 
 /** Theme for app quill charts. `buildTheme()` reads CSS variables from the DOM, so the memo keys on
- *  `isDarkModeOn` to re-read them when the app theme flips. It also applies the app chart colors
- *  (faint dashed grid, muted axis lines); caller `overrides` win over those. Pass a stable (memoized
- *  or module-level) `overrides` object — a fresh object every render defeats the memo. */
+ *  `isDarkModeOn` to re-read them when the app theme flips. Pass a stable (memoized or module-level)
+ *  `overrides` object — a fresh object every render defeats the memo. */
 export function useChartTheme(overrides?: Partial<ChartTheme>): ChartTheme {
     const { isDarkModeOn } = useValues(themeLogic)
     return useMemo(() => buildTheme({ ...chartThemeDefaults(isDarkModeOn), ...overrides }), [isDarkModeOn, overrides])
@@ -72,19 +71,18 @@ export function useDateRangeZoom(
     return enabled && dates?.length && onZoom ? handler : undefined
 }
 
-/** Drop-in replacement for the `useMemo` that builds a chart's config object. On top of memoizing,
- *  it applies app-level rendering defaults (monotone curve, axis lines, tick marks, crosshair,
- *  grid). Keys the config sets explicitly (non-undefined) always win over the defaults. */
+/** Builds a chart's config object, memoized on `deps`, applying `CHART_CONFIG_DEFAULTS` for any
+ *  key the factory leaves undefined. Keys the factory sets explicitly always win over the defaults. */
 export function useChartConfig<T extends object>(factory: () => T, deps: DependencyList): T
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined {
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    const config = useMemo(factory, deps)
     return useMemo(() => {
+        const config = factory()
         if (!config) {
             return config
         }
         const defined = Object.fromEntries(Object.entries(config).filter(([, value]) => value !== undefined))
         return { ...CHART_CONFIG_DEFAULTS, ...defined } as T
-    }, [config])
+    }, deps)
 }
--- a/frontend/src/queries/nodes/InsightViz/insightDisplayOptions.tsx
+++ b/frontend/src/queries/nodes/InsightViz/insightDisplayOptions.tsx
@@ -63,7 +63,6 @@ export function useInsightDisplayOptions(): { items: LemonMenuItems; count: numb
     } = useValues(trendsDataLogic(insightProps))
     const { featureFlags } = useValues(featureFlagLogic)
     const hideWeekendsEnabled = !!featureFlags[FEATURE_FLAGS.PRODUCT_ANALYTICS_HIDE_WEEKENDS]
-    const styleRefreshEnabled = !!featureFlags[FEATURE_FLAGS.QUILL_CHART_STYLE_REFRESH]
 
     // The slope graph shows the first vs last interval, so it drops the options that need the points
     // between them (smoothing, multiple axes, alert/annotation overlays, statistical analysis).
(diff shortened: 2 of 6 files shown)


frontend/src/lib/charts/hooks.ts
```diff
@@ -18,7 +18,7 @@ const REFRESHED_CONFIG_DEFAULTS = {
     barCornerRadius: 4,
 } as const
 
-function refreshedThemeOverrides(isDarkModeOn: boolean): Partial<ChartTheme> {
+function chartThemeDefaults(isDarkModeOn: boolean): Partial<ChartTheme> {
     return {
         gridColor: isDarkModeOn ? 'rgba(255,255,255,0.06)' : 'rgba(0,0,0,0.06)',
```
sampennington, line 23: 🤖 **sp-ship** · ♻️✨📊📐 `[reuse+quality+hog-charts+xp]` · **Blocker**

These rgba literals are spread *after* `buildTheme`'s base, so they now permanently discard `gridColor: graphColors.axisLine` and `crosshairColor: graphColors.crosshair` for every app quill chart. `--color-graph-axis-line` and `--color-graph-crosshair` are already light/dark-aware in `base.scss` and already read by `getGraphColors()`, so the concrete failure is: a designer edits the token and nothing changes anywhere. It also re-derives dark mode from `themeLogic.isDarkModeOn` when the CSS variable already encodes it, which is why quill's own `useChartTheme` watches DOM theme mutations rather than trusting a boolean.

Behind a flag this was a reasonable experiment shape. As the permanent default it wants to move down: put the refreshed values on the existing graph tokens (adding one for `axisLineColor`, which has none yet), let `buildTheme` carry them in `base`, and drop this function. That also picks up `Sparkline.tsx` and `Metric.tsx`, which call quill's `useChartTheme` directly and so never see any of this.

frontend/src/lib/charts/hooks.ts
```diff
@@ -83,20 +73,18 @@ export function useDateRangeZoom(
...
- *  explicitly (non-undefined) always win over the defaults. */
+ *  it applies app-level rendering defaults (monotone curve, axis lines, tick marks, crosshair,
+ *  grid). Keys the config sets explicitly (non-undefined) always win over the defaults. */
 export function useChartConfig<T extends object>(factory: () => T, deps: DependencyList): T
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined
 export function useChartConfig<T extends object>(factory: () => T | undefined, deps: DependencyList): T | undefined {
-    const refreshEnabled = useChartStyleRefreshEnabled()
     // eslint-disable-next-line react-hooks/exhaustive-deps
     const config = useMemo(factory, deps)
     return useMemo(() => {
```
sampennington, line 80: 🤖 **sp-ship** · ⚛️✨📐 `[react+quality+xp]` · **Suggestion / follow-up**

`rerender-split-combined-hooks`. The split was load-bearing while the flag existed: `refreshEnabled` could flip independently of `deps`, so the merge needed its own memo. Now `config`'s identity changes exactly when `deps` change, so this second memo is keyed on a value keyed on the same thing.

```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
    return useMemo(() => {
        const config = factory()
        if (!config) {
            return config
        }
        const defined = Object.fromEntries(Object.entries(config).filter(([, value]) => value !== undefined))
        return { ...CHART_CONFIG_DEFAULTS, ...defined } as T
    }, deps)
```

frontend/src/lib/charts/hooks.ts
```diff
@@ -9,7 +9,7 @@ import { themeLogic } from 'lib/logic/themeLogic'
 
 import { buildTheme } from './utils/theme'
 
-const REFRESHED_CONFIG_DEFAULTS = {
+const CHART_CONFIG_DEFAULTS = {
```
sampennington, line 12: 🤖 **sp-ship** · 📊♻️ `[hog-charts+reuse]` · **Suggestion / follow-up**

Chart-type-specific keys in a type-agnostic hook, and the `as T` cast below suppresses the excess-property check that would catch it: `barCornerRadius` is declared only on the bar/funnel/combo configs but now lands on `BoxPlotConfig` and every line config, and `curve` lands on every bar config. Harmless today since chart types destructure known keys, but a future library-side name collision would land silently on every consumer. Splitting into line / bar / shared-axis sets, or typing this `Partial<ChartConfig>` with narrow overloads, keeps the contract honest.

Also worth doing in the same pass: five `mcp_analytics` call sites (`ActivityChart`, `ToolUsageChart`, `ToolErrorRateChart`, `ToolQualityCharts`, `MCPAnalyticsToolDetail`) still spell out these exact literals inside their `useChartConfig` factories. Meaningful as flag-era opt-ins, redundant now — and they'd silently miss any future change here.

frontend/src/lib/charts/hooks.ts
```diff
@@ -83,20 +73,18 @@ export function useDateRangeZoom(
...
     const config = useMemo(factory, deps)
     return useMemo(() => {
-        if (!refreshEnabled || !config) {
+        if (!config) {
             return config
         }
         const defined = Object.fromEntries(Object.entries(config).filter(([, value]) => value !== undefined))
-        return { ...REFRESHED_CONFIG_DEFAULTS, ...defined } as T
-    }, [refreshEnabled, config])
+        return { ...CHART_CONFIG_DEFAULTS, ...defined } as T
```
sampennington, line 86: 🤖 **sp-ship** · 📊🧪 `[hog-charts+test]` · **Suggestion / follow-up**

This merge is now the sole styling path for every chart in the app and nothing imports `lib/charts/hooks` from a test. There's a nameable regression it would let through: `chartStyleCurve` returns `undefined` when no line style is picked and `'linear'` when one is, and `chartStyleAdapter.test.ts` asserts only the adapter half. The other half of that contract lives here — drop the `value !== undefined` filter or flip the spread order and the "Line style → straight" toggle silently stops working while every existing chart test still passes.

A `renderHook` test covering defaults-fill-when-absent, explicit-value-wins, explicit-`undefined`-falls-through, and `undefined` config passthrough is cheap and deterministic.

frontend/src/lib/charts/hooks.ts
```diff
@@ -28,23 +28,13 @@ function refreshedThemeOverrides(isDarkModeOn: boolean): Partial<ChartTheme> {
...
-    const { featureFlags } = useValues(featureFlagLogic)
-    return !!featureFlags[FEATURE_FLAGS.QUILL_CHART_STYLE_REFRESH]
-}
-
 /** Theme for app quill charts. `buildTheme()` reads CSS variables from the DOM, so the memo keys on
- *  `isDarkModeOn` to re-read them when the app theme flips. Behind `QUILL_CHART_STYLE_REFRESH` it
- *  also applies the refreshed chart colors (faint dashed grid, muted axis lines); caller `overrides`
- *  win over both. Pass a stable (memoized or module-level) `overrides` object — a fresh object every
- *  render defeats the memo. */
+ *  `isDarkModeOn` to re-read them when the app theme flips. It also applies the app chart colors
```
sampennington: 🤖 **sp-ship** · 🦔✨📐 `[personal+quality+xp]` · **Suggestion / follow-up**

The "also" is dangling now that there's nothing to contrast with, and the sentence narrates the spread on line 37, which already reads defaults-then-overrides. The colors listed are a second copy of `chartThemeDefaults` too, and "colors" undersells it (`gridDashPattern` / `crosshairDashPattern` aren't colors). Suggest cutting that sentence and keeping the two ideas a reader can't recover from the code: why the memo keys on `isDarkModeOn`, and the stable-`overrides` requirement.

frontend/src/lib/charts/hooks.ts
```diff
@@ -83,20 +73,18 @@ export function useDateRangeZoom(
 }
 
 /** Drop-in replacement for the `useMemo` that builds a chart's config object. On top of memoizing,
- *  it applies app-level rendering defaults — currently the refreshed style (monotone curve, axis
- *  lines, tick marks, crosshair, grid) behind `QUILL_CHART_STYLE_REFRESH`. Keys the config sets
- *  explicitly (non-undefined) always win over the defaults. */
+ *  it applies app-level rendering defaults (monotone curve, axis lines, tick marks, crosshair,
```
sampennington: 🤖 **sp-ship** · 🦔✨📐 `[personal+quality+xp]` · **Suggestion / follow-up**

The parenthetical restates `CHART_CONFIG_DEFAULTS` from the top of this same file and already drifted — it omits `barCornerRadius`. Point at the constant instead so it can't drift again. "Drop-in replacement for the `useMemo`" also only parses for someone who remembers there used to be a bare `useMemo` here; post-flag this is just how you build a chart config. Keep the non-obvious part: only non-undefined keys override the defaults.

frontend/src/queries/nodes/InsightViz/insightDisplayOptions.tsx
```diff
@@ -100,12 +99,10 @@ export function useInsightDisplayOptions(): { items: LemonMenuItems; count: numb
...
     // Only the quill line charts (trends/stickiness line and area, retention and funnel-trends
-    // graphs) draw curves, and only the style-refresh flag curves lines by default — without it
-    // there's no curvature to straighten. Retention and funnel trends default to their line graph
-    // when display is unset.
+    // graphs) draw curves, so they're the only ones with curvature to straighten. Retention and
+    // funnel trends default to their line graph when display is unset.
     const isLineChartInsight = isLineDisplay || ((isRetention || isTrendsFunnel) && !display)
-    const showLineStyleConfig =
-        (isTrends || isStickiness || isRetention || isTrendsFunnel) && isLineChartInsight && styleRefreshEnabled
+    const showLineStyleConfig = (isTrends || isStickiness || isRetention || isTrendsFunnel) && isLineChartInsight
```
sampennington, line 105: 🤖 **sp-ship** · 🧪 `[test]` · **Suggestion / follow-up**

This goes from flag-gated to always-on, which makes the "Line style" menu item and its contribution to the modified-options `count` badge visible for every trends/stickiness/retention/funnel-trends line insight. `useInsightDisplayOptions` has no test coverage at all, so `isLineChartInsight` — including the "retention and funnel trends default to a line graph when `display` is unset" branch — is now load-bearing for everyone and unverified. A small test asserting the item appears for line/area trends and for retention/funnel-trends with `display` unset, and not for bar/pie/calendar-heatmap, locks in the new default cheaply.

sampennington (commented):
🤖 **sp-ship** · 🚫 Blockers

The flag removal itself is complete and correct — nothing in the repo still references `QUILL_CHART_STYLE_REFRESH`, the surviving `featureFlagLogic` imports are still earning their place, and `showLineStyleConfig` correctly kept its insight-type guard when it lost the flag. The one blocker is that `chartThemeDefaults` now permanently shadows the `--color-graph-axis-line` / `--color-graph-crosshair` design tokens that `buildTheme()` reads: acceptable while flag-scoped, but as the default it kills those tokens and encodes dark mode twice (CSS vars in `buildTheme`, a kea boolean here).

Two things that didn't anchor to a line and are worth a look. `frontend/src/scenes/insights/EditorFilters/LineStylePicker.tsx:28` still says the default is smooth "under the style-refresh flag" — the last stale mention of the flag anywhere. And the config defaults now reach surfaces that never opted into them: `RatingBarChart` sets only `hideYAxis`, so it picks up value grid lines, a crosshair, and tick marks, while `MultipleChoiceBarChart` opts out of grid and axis lines but still inherits orphaned tick marks. Meanwhile the two funnel bar charts bypass `useChartConfig` entirely and keep square corners, so "every app chart" isn't quite true. Worth a screenshot pass, and note every chart Storybook baseline was captured with the old styling.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Fixed in `095effa` — collapsed the two `useMemo` calls into one keyed on `deps`, applying your suggested diff verbatim.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Fixed in `095effa` — cut the dangling "also" sentence and the redundant colors recap, kept the memo-key and stable-overrides rationale.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Fixed in `095effa` — points at `CHART_CONFIG_DEFAULTS` instead of recapping it, dropped the flag-era "drop-in replacement" framing, kept the non-obvious override rule.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Agree this is a real bug — confirmed `chartThemeDefaults` values pass through `buildTheme(...)` as `overrides`, permanently shadowing `graphColors.axisLine`/`graphColors.crosshair`, and confirmed `Sparkline.tsx`/`Metric.tsx` bypass this wrapper entirely via `@posthog/quill-charts`'s own `useChartTheme`. Deferring rather than auto-fixing: the right fix (new `--color-graph-*` token(s) in `base.scss`, wiring through `buildTheme`, touching the two direct-`useChartTheme` callers) is a multi-file design-token change that goes well beyond this PR's scope of removing a flag. Leaving open for a human decision on scope — should be addressed as a fast follow-up.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Out of scope for this PR — this is a type-safety refactor (splitting `CHART_CONFIG_DEFAULTS` into narrower per-chart-type sets) plus deduplicating 5 unrelated `mcp_analytics` call sites, not a flag-removal change. Resolving here; worth a follow-up.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-triage**

Good catch on the coverage gap, but adding a new test suite is a substantive addition beyond triaging this flag-removal PR. Resolving here; worth a follow-up `renderHook` test for `useChartConfig` covering the defaults-fill/explicit-value/explicit-undefined cases you outlined.

frontend/src/queries/nodes/InsightViz/insightDisplayOptions.tsx
sampennington: 🤖 **sp-triage**

Agree `useInsightDisplayOptions` deserves test coverage now that this branch is always-on, but writing that suite is a substantive addition beyond this flag-removal PR. Resolving here; worth a follow-up covering the line/area/retention/funnel-trends-default cases you outlined.

frontend/src/lib/charts/hooks.ts
sampennington: 🤖 **sp-ship**

Downgrading this from Blocker. The finding is real, but it is not introduced by this PR.

`quill-chart-style-refresh` was at 100% rollout with no property filters before this change, so every user was already on the flag-on path. `chartThemeDefaults` has been shadowing `--color-graph-axis-line` and `--color-graph-crosshair` in production for the whole rollout — deleting the flag removes an `if` that nobody was on the false side of. Behavior is unchanged.

So the token cleanup is worth doing, but it is its own change with its own blast radius: both tokens are also read by the two remaining Chart.js graphs (`DataVisualization/Charts/LineGraph.tsx` and `BillingLineGraph.tsx`), and the refreshed values are visibly fainter, so retuning them restyles those too. There is also no token for `axisLineColor` yet. That belongs in a follow-up alongside pointing `Sparkline.tsx` and `Metric.tsx` at this hook, not bolted onto a flag deletion.

Leaving the thread open as the record; not treating it as a merge gate.

PR #74142, opened 2026-07-28
feat(conversations): show plan response time in new ticket placeholder

christiaan-ph:
## Problem

Customers opening a support ticket from the sidebar get no signal on when to expect a reply, and chat-style UIs set an expectation of near-immediate answers. The composer placeholder is a cheap place to set the right expectation up front, using the response time for the customer's actual plan.

Context: https://posthog.slack.com/archives/C075D3C5HST/p1785200408298969

## Changes

- The new ticket composer placeholder now reads "Describe what you need help with and our support engineers aim to get back to you within 48 hours." (or whatever response time the customer's plan carries). "Aim to" rather than "will", because these are target response times, not contractual SLAs.
- The plan resolution is the exact logic the "Support is open Monday - Friday" response times table already uses. The table's inline `getCurrentPlan` and `getResponseTimeFeature` helpers are extracted into `lib/components/Support/supportResponseTime.ts` and both surfaces point at it, so the placeholder and the table can't drift apart. #74173 builds the /ticket confirmation on the same module and is stacked on this branch, so this PR merges first.
- `getSupportResponseTime` only ever returns a time billing actually publishes for the plan (via the `support_response_time` feature note), or the fixed pay-as-you-go "72 hours". Free plans, active trials, unloaded billing, missing notes, and notes that don't read as a duration all resolve to null, and the placeholder falls back to the existing generic copy "...will get back to you." An earlier revision invented "1 business day" in those cases; that fallback is gone so we never promise a time nobody signed up to.
- The response times table itself is unchanged in behavior: it keeps its existing "1 business day" display default for rows billing publishes no note for, exactly as on master.

**Automated:** `frontend/src/lib/components/Support/supportResponseTime.test.ts`, a parameterized suite (15 cases) over the full plan-to-response-time mapping: plans with and without billing notes, active and expired trials, unloaded billing, non-duration notes, and the known enterprise org path. This logic previously lived untested inside the table component. Existing `sidepanelTicketsLogic.test.ts` suite passes (8 tests). Lint and typecheck pass for all touched files.

**Manual (local stack, stubbed `/api/billing` responses):** an initial round verified every plan scenario end to end: Boost/Scale/Enterprise notes rendered as "within 48/24/8 hours", pay-as-you-go as "within 72 hours", and free/unresolvable plans fell back to the generic copy, with the response times table unchanged throughout. Two changes landed after that round and have been verified by the unit suite but not yet re-checked in the browser: the "aim to" wording, and active trials / missing notes now showing the generic placeholder instead of "1 business day". A fresh pass over those scenarios is pending before merge.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed, no documented workflow changes.

## 🤖 Agent context

Written with Claude Code. Skills invoked: `/writing-user-facing-copy`, `/writing-tests`. The extraction-and-reuse approach was chosen over a second implementation because the support response times table already resolves the current plan. After review feedback the wording moved to "aim to get back to you" and the invented "1 business day" fallback was removed: the helper now returns null whenever billing publishes no duration, and callers leave the promise out. The helper module was also consolidated with the near-identical one from #74173 into a single `supportResponseTime.ts` that both PRs share; #74173 is stacked on this branch. Manual testing of the original revision was done pair-style with stubbed billing scenarios; the post-review changes still need a browser pass.


abigailbramble:
Can we have 'aim to get back to you within' to reflect the target response times, not contractual SLAs

christiaan-ph:
@abigailbramble good catch, updated in b52ab9f. The placeholder now reads:

> Describe what you need help with and our support engineers aim to get back to you within 48 hours.

The duration itself still comes from whatever billing publishes for the org's plan, so it always matches the response times table in the help panel rather than being hardcoded here. When the plan has no published time (free plans, active trials), no time is mentioned at all:

> Describe what you need help with and our support engineers will get back to you.

Same wording change applied to the `/ticket` confirmation message in #74173.

Also fixed the failing frontend typecheck on this PR, which was stale generated kea types for `sidepanelTicketsLogic` rather than anything in the copy.

christiaan-ph:
Follow-up to the above: 7d59840 also removes the invented "1 business day" fallback. Active trials and plans billing publishes no duration for now get the generic placeholder with no time promised, matching the behavior #74173 already had. That PR is now stacked on this branch and shares the same helper, so the two surfaces can't disagree.

abigailbramble (approved):
LGTM

PR #74145, opened 2026-07-28
feat(warehouse_sources): support Intercom API version 2.16

Gilbert09:
## Problem

Intercom's latest stable REST API version is 2.16. The data-warehouse Intercom source topped out at 2.15, so new connections couldn't be created on the current version. No deprecation has been announced for 2.13 or 2.15, so this is a plain update, not a sunset.

## Changes

- Add `2.16` to `supported_versions` and flip `default_version` to `2.16`, so new sources are stamped with it. Existing sources keep their pin (2.13 or 2.15) and are byte-for-byte unaffected.
- No request-layer changes needed. The source already dispatches on the resolved pin by threading it into the `Intercom-Version` header via `intercom_source` and `validate_credentials`, so adding a label is a declaration change.
- 2.16's differences from 2.15 for the endpoints this source reads are response-shape only (`owner_id` returns as string, `articles` swaps `parent_id`/`parent_type` for `parent_ids`, ticket assignee ids become integers). The source uses an auto-inferred schema, so these are absorbed without a data or schema transform. No per-version module is warranted.
- No migration to repin existing rows: this is not a deprecation, and existing pins stay put.

> [!NOTE]
> The default flip only affects newly created sources and any row with a NULL pin. Rows pinned to 2.13/2.15 continue to sync on their pinned version.

Automated only (I'm an agent, no manual sync run):

- Extended `test_intercom_source.py` so version resolution, the new default, and the dispatched request path cover 2.16 alongside 2.13/2.15 (default-is-latest, `source_for_pipeline` resolution, and the `validate_credentials` probe pin).
- The existing `TestVersionDispatch` tests parameterize over `supported_versions`, so they now assert the `Intercom-Version: 2.16` header reaches the wire on both the session (scroll/substream) and framework REST paths.
- Ran the Intercom suite (132 passed) and the registry invariant suite `test_source_versions.py` (1255 passed), which enforces `default_version == supported_versions[-1]`, default ∈ supported, and deprecated ⊆ supported.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code. Invoked the `/warehouse-source-new-version` skill and followed its version-declaration + dispatch conventions.

Decisions along the way:
- Ran the skill's "does the version need to exist at all?" gate against the changelog. 2.16 qualifies on two grounds: the version is a request input (the `Intercom-Version` header the source already sends), and response shapes genuinely diverge for endpoints we read. Because the schema is auto-inferred, those divergences need no per-version branching, so the correct shape is declaration-only.
- Checked open PRs (`intercom`, `2.16`, and the maintainer's own list via `--author Gilbert09`) for a duplicate. The nearest neighbor, #74012 (coerce Intercom `owner_id`/`waiting_since` to string), is an unrelated Arrow type-merge fix, not a version bump. No duplicate exists.
- Left the stale `INTERCOM_API_VERSION = "2.13"` module constant alone: `IntercomSource.validate_credentials` always passes `self.resolve_api_version(...)`, so that constant only backs direct standalone calls and touching it is outside this bump's scope.

---


.agents/skills/warehouse-source-new-version/SKILL.md
```diff
@@ -106,3 +106,6 @@ After you finish a version-update or deprecation PR using this skill, **append w
...
 - Vendor OpenAPI is often embedded in the docs pages themselves (`/reference/get_v3-<resource>.md`, indexed by `llms.txt`) — parse it to confirm paths, filter params, response envelope, and id types for all endpoints at once instead of trusting a prose migration guide.
 - The registry invariant now enforces `default_version == supported_versions[-1]` (test_source_versions.py): declare versions oldest→newest and flip the default in the same PR — a held-back default fails CI, and `[-1]` is a trustworthy "latest" accessor.
+- Intercom 2.16: pure declaration bump. The source already threads the resolved pin into the `Intercom-Version` header (`intercom_source`/`validate_credentials`), and 2.16's changelog diffs vs 2.15 for the endpoints it reads are response-shape only (`owner_id`→string, articles `parent_id`/`parent_type`→`parent_ids`, ticket assignee ids→int), absorbed by the auto-inferred schema — no per-version module, no repin migration (not a deprecation).
```
danielcarletti: remove these additions to the SKILL, they are not necessary at all

danielcarletti (approved):

PR #74163, opened 2026-07-28
feat(warehouse_sources): deprecate Zonka Feedback v1 in favor of v2.1

Gilbert09:
## Problem

Zonka is deprecating its older API generation, which its docs market as "API 2.0" (`docs.api2.zonkafeedback.com`), in favor of the current v2.1 API (`apidocs.zonkafeedback.com`). We want customers steered onto the current version before the vendor retires the old one, without breaking any source still pinned to a supported version.

The task framed this as "deprecate v2.0 and migrate v2.0 to v2.1", but our source never adopted the vendor's "2.0" label. It supports two labels: the legacy `UNVERSIONED_API_VERSION` placeholder `v1`, and `v2.1` (already the default). There is no per-version request dispatch, so a `v1` pin already hits the current v2.1 host (`apis.zonkafeedback.com`). So the label to deprecate is the one the source actually stores, `v1`, not a `v2.0` that exists nowhere in the code. This is the same collapsed shape as the recent Brex and Calendly deprecations.

## Changes

- Flag `v1` deprecated via `deprecated_versions = (VersionDeprecation(version=ZONKA_API_VERSION_V1, sunset_at=None),)`. No sunset date is published, so `sunset_at=None`. `v2.1` stays the default and `v1` stays supported, so existing pins keep syncing. The generic in-product deprecation warning lights up from this metadata with no source-specific UI.
- Migration `0105_repin_zonka_feedback_api_version` repins source-level `ExternalDataSource.api_version` from `v1` to `v2.1`. It filters on the exact old value (idempotent, re-run matches nothing), only touches source-level pins (NULL already resolves to the new default), and leaves user-managed schema-level `ExternalDataSchema.api_version` overrides untouched. Reverse is a no-op because a repinned row is indistinguishable from a natively-created v2.1 one.

> [!NOTE]
> Because there is no per-version dispatch, `v1` and `v2.1` resolve to byte-identical requests today, so the repin needs no data or schema transform. It is a pure pin move. **Written, not run** - a human reviews and executes the migration.

I (Claude) ran the source unit tests and the registry invariant suite:

```
pytest products/warehouse_sources/backend/temporal/data_imports/sources/zonka_feedback/tests/test_zonka_feedback_source.py \
       products/warehouse_sources/backend/temporal/data_imports/sources/tests/test_source_versions.py
# 1281 passed
```

Added `test_v1_is_deprecated_without_sunset_and_v2_1_is_not` to the source tests. It catches a regression where the `deprecated_versions` line is dropped, points at the wrong label, or gains a stray sunset date - none of the existing tests assert deprecation metadata. `ruff check` and `ruff format` pass on the touched files.

I could not run `makemigrations --check` or `sqlmigrate` in this environment (no database reachable), but both got past migration-graph loading and failed only at the Postgres connect step, so the migration is structurally valid and linear off `0104_easybill_source`. The migration mirrors the sibling repin migrations (`0081` Klaviyo, `0092` Lightspeed Retail, and the open Brex/Calendly `0105`s).

> [!WARNING]
> Migration number collision: the open Brex (#74144) and Calendly (#74148) PRs also add a `0105_*` migration off `0104_easybill_source`. Whichever merges first wins `0105`; the rest (including this one) need renumbering before merge. `ci:preflight` flags this.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change - the deprecation surfaces through the existing generic version-deprecation warning.

## 🤖 Agent context

Authored by Claude (Claude Code). Skills invoked: `/warehouse-source-new-version` (version-declaration and deprecation conventions), `/django-migrations` (migration safety), `/writing-tests` (test gate).

Key decision: the ticket said "deprecate v2.0", but the codebase has no `v2.0` label - it has `v1` (the pre-versioning `UNVERSIONED_API_VERSION` placeholder) and `v2.1`. Per the versioning framework's rule that labels are opaque and copied exactly, I deprecated the real stored label `v1` rather than inventing a `v2.0` label that no row carries (a `v2.0` repin filter would match zero rows). I confirmed v2.1 was already implemented and default, that there is no per-version dispatch (so the repin is behavior-neutral), and that no open PR already covers Zonka before opening this one. I appended the label-reconciliation learning to the skill.

---


danielcarletti (approved):

danielcarletti:

PR #74172, opened 2026-07-28
feat(warehouse_sources): support Google Ads API version v25

Gilbert09:
## Problem

Google is retiring the **v24** Google Ads API on its rolling ~12-month schedule, and **v25** is now the latest major. The data-warehouse Google Ads source defaulted to v24, so new sources would keep landing on a version headed for sunset. We need to support v25, move new sources onto it, and warn customers still pinned to v24 — without disturbing any source pinned to a supported version (v23/v24).

Vendor references: [v25 announcement](https://ads-developers.googleblog.com/2026/07/announcing-v25-of-google-ads-api.html) · [release notes](https://developers.google.com/google-ads/api/docs/release-notes) · [sunset dates](https://developers.google.com/google-ads/api/docs/sunset-dates)

## Changes

- **New version v25.** `supported_versions` is now `("v23", "v24", "v25")` and `default_version` flips to `v25`, so new sources are stamped v25. Existing pins are honored verbatim and keep syncing on their own version.
- **SDK bump.** The API version is bundled inside the `google-ads` package — v25 first ships in `google-ads==31.2.0` (was 30.1.0). That release bundles v21–v25, so the source's v23/v24 proto imports survive. Nothing else in the repo imports `google.ads.googleads.v*`. 31.2.0 is newer than uv's rolling `exclude-newer` cutoff, so `google-ads` is added to `exclude-newer-package` to override it (same escape hatch `django`/`modal` use).
- **v25 proto modules** registered in `_PROTO_MODULES_BY_VERSION` so v25 responses decode against v25 protos.
- **v24 deprecated.** Added to `deprecated_versions`, which lights up the generic in-product warning — no source-specific UI. Google hasn't published a firm sunset date yet (projected ~May 2027, 12 months after v24's April 2026 release), so `sunset_at=None` until it does.
- **Repin migration** (`0105_repin_google_ads_api_version`) moves source-level `ExternalDataSource.api_version` from `v24` → `v25`. Idempotent (exact-value filter), reverse is a no-op, and it leaves schema-level `ExternalDataSchema.api_version` overrides untouched (customer-pinned by design). v25 reads the same resources/fields this source already reads — its only breaking change removes the lifecycle-goal resources, which this source never touches — so **no data/schema transform is needed**. **Written, not run** — a human reviews and executes it.

Out of scope: v23 is also on the vendor's deprecation path (earlier projected sunset) but this change is scoped to the v24 deprecation triggered by v25's release; the CDP Google Ads destination (#74075) and `integration.py`'s REST v21 helper are separate surfaces and untouched.

Agent-run automated tests only (no manual/live sync — there are no stored credentials or live-sync harness; verification against vendor docs is the source of truth per the source's conventions):

- `test_google_ads_source.py` + `test_source_versions.py` — **1416 passed**. Extended the existing version-dispatch/discovery/type-URL/validate-credentials parameterizations to include `v25` and flipped the default expectations to `v25`; added a focused deprecation-metadata case (v24 deprecated, v25/default not). These catch a v25 pin failing to reach the client/discovery layer and a missing v25 proto-module registration — regressions no prior case covered.
- `ruff check` / `ruff format` clean on touched files; `mypy` clean on changed source + migration; `hogli ci:preflight` reports 0 failures (uv.lock in sync).

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (PostHog Code). Skills invoked: `/warehouse-source-new-version` (version-add + deprecation conventions), `/django-migrations` (repin migration), `/writing-tests` (test gate).

Key decisions: (1) Confirmed the version genuinely needs adding — the vendor is retiring v24 (our default) and v25 is a real request input (SDK routing + proto-module selection), so this isn't a rename-only relabel. (2) The SDK-bundled-version constraint made the `google-ads` bump the crux; verified 31.2.0 keeps v23/v24 before bumping. (3) Chose `sunset_at=None` over a fabricated-precise date because Google hasn't published one — matching the `lightspeed_retail` precedent. (4) Left `get_non_retryable_errors` alone: v24 stays fully functional until sunset, and there's no confirmed version-rejection signature to add yet. Appended these learnings to the skill for the next bump.

---


Gilbert09:
Heads-up on the migration number: this PR adds `warehouse_sources/0105_repin_google_ads_api_version`, and two sibling PRs currently also sit on `0105` (#74148 Calendly, #74144 Brex). `0105` is correct relative to master (leaf `0104`); whichever merges first keeps it and the rest just need `python manage.py rebase_migration warehouse_sources` before merge. Didn't pre-renumber since none has merged yet — that would only move the collision.

danielcarletti (approved):

PR #74176, opened 2026-07-28
feat(warehouse_sources): support Zendesk Sunshine API version v2

Gilbert09:
## Problem

The **Zendesk Sunshine** data warehouse source reads the legacy Sunshine custom objects API (`/api/sunshine/`, version `v1`). Zendesk is retiring that API: no new legacy objects can be created since 2026-01-15, and it is removed on **2026-06-30**. Sources pinned to `v1` will stop working at that point, and the current custom objects data lives under a different, standard API (`/api/v2/custom_objects`).

## Changes

- **New version `v2`.** `supported_versions = (v1, v2)`, `default_version = v2` — new sources use the current custom objects API. The request layer dispatches on the resolved pin: base path (`/api/sunshine/` vs `/api/v2/`), the endpoint catalog, and each resource's response envelope (`data` vs `custom_objects` / `custom_object_records` / `custom_object_fields`) are selected per version. The shared `links.next` cursor paginator and http_basic (`email/token`) auth are reused, so the v2 branch stays small.
  - `v1` (legacy) is unchanged: same table set (object types/records, relationship types/records, policies, limits) and the same incremental `objects/query` path.
  - `v2` exposes `custom_objects`, `custom_object_records`, and `custom_object_fields`, synced **full-refresh** (sorted by `updated_at`). The v2 records list endpoint has no server-side `updated_at` filter, only `sort` + cursor, so no incremental path is offered.
  - `get_schemas`, `validate_credentials`, and `source_for_pipeline` consume the resolved pin, so a pinned source discovers and syncs under its own version. An unsupported pin raises a non-retryable error rather than looping the ~6h discovery cadence.
- **Deprecation.** `v1` is added to `deprecated_versions` with `sunset_at = 2026-06-30`. The generic in-product deprecation banner and API fields light up from this metadata; no source-specific UI.
- **Migration `0105`.** Backfills source-level `api_version` `NULL -> v1` for Zendesk Sunshine sources, so the `default_version` flip to `v2` cannot silently move unpinned rows onto the divergent v2 table set. Idempotent (filters `NULL`), reverse is a no-op, and schema-level `api_version` overrides are left untouched.
- Updated the source caption, docs page, and canonical table descriptions to cover both versions.

### Customer migration (`v1 -> v2`): manual, not scripted

The `v1` and `v2` table sets genuinely differ — legacy relationships and permission policies have no v2 equivalent, and records reshape from a flat `attributes` field to nested `custom_object_fields`. Repinning an existing source from `v1` to `v2` is therefore a **resync under new tables**, not an in-place `api_version` flip: the next discovery cycle would orphan the v1 tables and start the v2 ones fresh, and the record columns change shape. That transform is lossy/ambiguous, so per the `warehouse-source-new-version` skill it is **not** scripted. The manual path (repin the source to `v2` from its settings, then let it resync under the new tables) is documented in the source docs and surfaced by the deprecation banner; a human runs it per source.

Automated (run by the agent with `uv run pytest`):

- `products/warehouse_sources/backend/temporal/data_imports/sources/zendesk_sunshine/tests/` — extended for both versions. New/changed cases catch: `get_base_url` and the endpoint catalog resolving per version (and raising on an unsupported pin); `get_schemas` returning the correct table set per pin and **defaulting to v2** for a NULL pin (a version-blind `get_schemas` would silently discover the wrong tables for every unpinned legacy source); `validate_credentials` probing the version-appropriate endpoint; the resolved pin threading into the request layer; the deprecation metadata (v1 deprecated with the 2026-06-30 sunset, v2 not deprecated); and a framework-integration test driving the real REST layer against a mocked v2 HTTP surface (v2 paths, `custom_object_records` envelope, `sort=updated_at`, `per_page=100`).
- `products/warehouse_sources/backend/temporal/data_imports/sources/tests/test_source_versions.py` — the registry invariants (default in supported, default is newest, deprecated ⊆ supported, default not deprecated) pass. 1330 tests green.
- `uv run mypy --cache-fine-grained` on the touched source files: no issues.
- The migration is a pure `RunPython` data migration (no schema ops); `makemigrations --check` could not run here (no DB in the sandbox), so a reviewer should confirm no state drift when applying.

## Docs update

Updated `products/warehouse_sources/backend/temporal/data_imports/sources/zendesk_sunshine/docs/zendesk-sunshine.md` in this PR.

## 🤖 Agent context

Authored with Claude Code. Skills invoked: `warehouse-source-new-version` (version-declaration, dispatch, pinning semantics, deprecation, and the migration decision), `django-migrations`, `writing-user-facing-copy` (caption, docs, error strings), `writing-tests`, `writing-code-comments`.

Key decisions:

- **v2 needed to exist** (not just a renamed label): the base path, endpoint paths, response envelopes, and table set all diverge from v1, and v1 is being retired — both gate conditions from the skill.
- **v2 shipped full-refresh, not incremental.** The vendor docs confirm the v2 records list endpoint supports `sort` + cursor pagination but has no server-side `updated_at` filter; rather than fabricate an unverified incremental-export path from partially-readable docs, v2 records sync full-refresh and this is documented.
- **The repin is not scripted.** Because the table sets and record shapes diverge, an automatic `api_version` UPDATE would be lossy; the shipped migration only backs out the NULL-pin cohort, and the manual repin path is documented. New learnings were appended to the `warehouse-source-new-version` skill.

---


danielcarletti (approved):

danielcarletti:

PR #74180, opened 2026-07-28
feat(warehouse_sources): support ShipStation API v2 and deprecate v1

Gilbert09:
## Problem

ShipStation runs two live vendor APIs. v1 is the original `ssapi.shipstation.com` OpenAPI (HTTP Basic with an API key + secret, US Pacific DateTimes). v2 is the newer ShipEngine-based API at `api.shipstation.com/v2` (a single `API-Key` header, ISO 8601 UTC, and an entirely different resource model). The vendor has deprecated v1 and will remove it at a future, not-yet-announced date. The source only implemented v1, so new connections were being created on the deprecated API and there was no path onto v2.

## Changes

- **Declare v2 and deprecate v1.** `supported_versions = (v1, v2)`, `default_version = v2` (new sources start on the current API), and `deprecated_versions` pins v1 with `sunset_at=None` (no date announced). The generic in-product deprecation warning and the public `/api/public_source_configs/` fields light up from this metadata — no source-specific UI.
- **Per-version transport dialect.** A frozen `ShipStationDialect` carries the axes that differ: host, auth scheme (Basic key+secret vs single `API-Key` header), page-size/sort param spellings, ascending-sort value, and date-filter format (Pacific vs ISO 8601 UTC). v1's request path is byte-for-byte unchanged. v2 requests pin to the base host and refuse cross-host redirects, since `requests` would replay a custom header (not just `Authorization`) across a redirect.
- **Per-version resource catalog.** v1 and v2 expose different tables (v1 orders/customers/products/…; v2 shipments/labels/carriers/batches/manifests/pickups/tags), so `get_schemas` returns the catalog for the resolved pin and the sync path builds from it. The v2 catalog (paths, primary keys, response envelopes, date filters, pagination) is grounded in the ShipEngine OpenAPI spec.
- **Version-aware credentials.** `api_secret` is now optional at the form level (v2 users have no secret); `validate_credentials` probes the resolved version's host and enforces the credentials that version needs (v1 requires the key + secret pair).
- **Non-retryable auth errors** extended to the v2 host.

### Migration

Repinning an existing v1 source to v2 is **not safely scriptable**: v2 needs a single API key the customer must generate in ShipStation's v2 API settings (it can't be derived from the v1 key + secret), and v2 exposes a different table set. An `api_version` UPDATE alone would point rows at credentials they don't have. So the customer repin is a **documented manual path** (generate a v2 API key, re-enter credentials, re-select tables), not a migration.

Migration `0105` is the safe part: it pins any ShipStation rows with a NULL `api_version` back to `v1`, so flipping the default to v2 can't silently move an unpinned row onto v2. It's idempotent (isnull guard), has a no-op reverse (same reasoning as the `0075` backfill), and is **written but not run** — a human reviews and executes it.

Automated only (the agent can't exercise the live vendor API; the v2 catalog is grounded in the ShipEngine OpenAPI spec):

- `sources/shipstation/tests/` and `sources/tests/test_source_versions.py` — all pass (`1332 passed`), plus `ruff` and `mypy` clean on the touched files.

- `test_request_targets_version_host_and_pagination` — a v2 pin syncing against v1's host/paths/params (every table 404s).
- `test_build_auth_scheme_per_version` / `test_v2_uses_snake_case_filter_and_sort_spellings` / `test_format_values_v2_uses_iso_utc` — v1 auth/param/date spellings leaking into a v2 request.
- `test_get_schemas_returns_the_pinned_versions_catalog` — discovery under the wrong version (tables disappear/duplicate on reconciliation).
- `test_declares_both_versions_defaulting_to_v2_with_v1_deprecated` — a dropped deprecation or held-back default silently keeping new sources on v1.
- `validate_credentials` version cases — probing the wrong host, or accepting a v1 connection with no secret.

## Docs update

No posthog.com doc change: the version pin and deprecation are surfaced by existing generic API/UI fields.

## 🤖 Agent context

Authored by PostHog Code. Skills invoked: `/warehouse-source-new-version` (version declaration + dispatch + migration conventions), `/django-migrations`, `/writing-tests`, `/writing-user-facing-copy`.

Key decisions: v2 is a full resource-API replacement, not a host/auth reskin, so it needed its own endpoint catalog and version-aware `get_schemas` rather than routing the v1 catalog through a v2 host. Because v2 credentials can't be derived from v1's and the table sets diverge, the v1→v2 repin is documented as manual instead of scripted; the only migration is the NULL-pin safety backstop for the default flip. Built on the current `rest_source`-based transport (paginator handles paging), reusing the shared `PageNumberPaginator(total_path="pages")` for both versions since both envelopes carry `page`/`pages`.

Supersedes the earlier attempt in #71679, which wired v2 as a transport-only reskin over the v1 catalog (breaking new-source syncs) and did not deprecate v1 or add a migration. That PR can be closed in favor of this one.

---


danielcarletti (approved):

danielcarletti:

PR #74181, opened 2026-07-28
feat(llm-gateway): add Kimi K3 on Modal

tatoalo:
## Problem

PostHog Code needs a token-priced open model that runs through the Claude adapter.

**Why:** Kimi K3 provides another shared-infrastructure model without requiring new provider credentials.

## Changes

- route `moonshotai/kimi-k3` to its Modal endpoint using the existing proxy credentials
- advertise it to the Claude adapter and record its input, output, and cache-read token costs

### DIFF
--- a/services/llm-gateway/src/llm_gateway/api/anthropic.py
+++ b/services/llm-gateway/src/llm_gateway/api/anthropic.py
@@ -46,6 +46,8 @@ from llm_gateway.metrics.prometheus import (
     REQUEST_COUNT,
     REQUEST_LATENCY,
 )
+from llm_gateway.modal import is_modal_served_model
+from llm_gateway.modal_routing import send_modal_anthropic_messages
 from llm_gateway.models.anthropic import GATEWAY_ONLY_FIELDS, AnthropicCountTokensRequest, AnthropicMessagesRequest
 from llm_gateway.products.config import validate_product
 from llm_gateway.request_context import (
@@ -535,6 +537,9 @@ async def _handle_anthropic_messages(
     if provider == "cloudflare" or is_cloudflare_model(body.model):
         return await send_glm_anthropic_messages(data, user, body.stream or False, product)
 
+    if is_modal_served_model(body.model):
+        return await send_modal_anthropic_messages(data, user, body.stream or False, product)
+
     if provider == "bedrock":
         return await _send_bedrock_messages(data, user, request, body.stream or False, product)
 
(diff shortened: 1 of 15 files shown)


services/llm-gateway/src/llm_gateway/api/anthropic.py
```diff
@@ -535,6 +537,9 @@ async def _handle_anthropic_messages(
     if provider == "cloudflare" or is_cloudflare_model(body.model):
         return await send_glm_anthropic_messages(data, user, body.stream or False, product)
 
+    if is_modal_served_model(body.model):
```
richardsolomou, line 540: kimi messages reach Modal here but `/v1/messages/count_tokens` still sends the same model to anthropic, where it fails as an unknown model. can we give it a local or modal-compatible counting path too?

services/llm-gateway/src/llm_gateway/modal_routing.py
```diff
@@ -0,0 +1,69 @@
...
+    ProviderConfig,
+    handle_llm_request,
+)
+from llm_gateway.auth.models import AuthenticatedUser
+from llm_gateway.config import get_settings
+from llm_gateway.modal import (
+    ensure_modal_model_allowed,
+    ensure_modal_model_configured,
+    make_modal_anthropic_call,
+    make_modal_completion_call,
```
richardsolomou, line 19: couldnt this share `glm_routing._send_via_modal`? both paths validate the model, resolve the Modal credentials, copy the request, then call `handle_llm_request`, so fixes could easily land in only one.

services/llm-gateway/src/llm_gateway/rate_limiting/model_cost_overrides.py
```diff
@@ -11,6 +11,28 @@
...
+        "mode": "chat",
+        "max_input_tokens": 262_144,
+        "max_output_tokens": 32_768,
+        "input_cost_per_token": 3e-06,
+        "output_cost_per_token": 1.5e-05,
+        "cache_read_input_token_cost": 3e-07,
+        "supports_vision": False,
+        "supports_prompt_caching": True,
+    },
+    "openai/moonshotai/kimi-k3": {
```
richardsolomou: could we define the Kimi cost metadata once and reuse it for both model IDs? theyre the same pricing contract, so a future update could easily change one and miss the other.

richardsolomou (changes requested):
could we add behavior coverage for Kimi’s Messages, count-tokens, chat-completions, and model-discovery paths? the current helper tests dont hit the new routing or registry branches.

richardsolomou (approved):

PR #74183, opened 2026-07-28
fix(aio): stop negative AI costs when cache reads exceed input tokens

christiaan-ph:
## Problem

When `$ai_generation` events arrive without explicit cost properties, ingestion computes `$ai_input_cost_usd` from token counts and a matched pricing entry. Cache tokens are reported two ways by providers: inclusive, where cache reads are a subset of `$ai_input_tokens` (OpenAI style), and exclusive, where they are separate pools (Anthropic style). `resolveCacheReportingExclusive` hard-codes the inclusive assumption for every non-Anthropic provider, so an event from an exclusive-reporting integration gets billed `prompt_rate * (input_tokens - cache_read_tokens)`. As soon as the cached prefix outgrows the new input tokens, that expression goes negative and the event is stored with a negative `$ai_input_cost_usd` and `$ai_total_cost_usd`, which then pollutes trace totals, dashboards, and spend rollups.

This is a real reported failure, not a hypothetical: a user's cache-heavy `xiaomi/mimo-v2.5` conversations carried the full cached prefix in `$ai_cache_read_input_tokens` with only the incremental tokens in `$ai_input_tokens`, producing visibly negative trace costs. The inclusive formula reproduces their stored negative totals exactly, and ingestion already counts these outcomes in `llma_ai_cost_outcome_total{outcome="negative"}` without preventing them.

## Changes

The Anthropic-via-Vercel path already has the right tool for this: a consistency check that falls back to exclusive accounting when the token counts are provably not inclusive (see the analysis that closed #69770, which validated the guard direction empirically). This PR generalizes that check to every provider instead of leaving it as an Anthropic special case.

Resolution order in `resolveCacheReportingExclusive` is now three explicit layers:

1. An explicit `$ai_cache_reporting_exclusive` boolean still wins, in both directions.
2. Provider priors are unchanged: Anthropic-family resolves exclusive (except the Vercel gateway inclusive case), everything else defaults to inclusive.
3. New universal fallback: an undeclared event reporting `input_tokens < cache_read_tokens + cache_write_tokens` cannot be inclusive, because inclusive counting means cache tokens are a subset of the input total. Such events resolve to exclusive, so the full input pool bills at the prompt rate, cache reads bill at the cache rate, and the input cost can no longer go negative.

Events that satisfy the inclusive invariant are billed exactly as before. The resolved flag written back onto events also feeds the cache-hit-rate analytics, which previously computed hit rates above 100% for these integrations.

> [!NOTE]
> An exclusive-reporting integration whose small requests happen to fit `cache_read <= input` still resolves inclusive and is slightly underbilled (never negative). Token counts alone cannot distinguish those events; full correctness there still requires the documented explicit flag.

Fallback flips are now observable via a new counter, `aio_ai_cost_cache_exclusive_fallback_total{prior}`, instead of only surfacing after the fact as negative totals.

Automated tests only, run locally by the agent:

- `hogli test nodejs/src/ingestion/pipelines/ai/costs/input-costs.test.ts` (67 passed) and `hogli test nodejs/src/ingestion/pipelines/ai/process-ai-event.test.ts` (126 passed).
- New resolver cases each guard a distinct regression no existing test caught: removing or narrowing the universal fallback (negative costs return), making the fallback over-eager (normal inclusive OpenAI events would flip and overbill), and breaking resolution precedence (the fallback must not override an explicit `false`).
- Updated the existing "cache read tokens exceeding input tokens" cost test, which previously asserted a total of 0. That assertion only held by coincidence: the test model has no cache read rate, and the 0.5 default multiplier made the cache cost cancel the negative text pool exactly. It now asserts the exclusive-accounting result and the resolved flag on the event.

I did not run a live ingestion pipeline; the change is confined to pure functions covered by the tests above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The in-repo reference (`products/ai_observability/skills/exploring-llm-costs/references/cache-accounting.md`) already describes the behavior as "ingestion auto-detects and writes the resolved value", which this change makes true for all providers, so no doc change is needed here. The property table on posthog.com (Generations docs) describes the default as provider-based only and could gain a sentence about the consistency fallback as a follow-up.

## 🤖 Agent context

- Authored by Claude Code (Claude Fable 5) in a support-investigation session driven by @christiaan-ph, starting from a user report of negative AI costs.
- Skills invoked: `/writing-tests`, `/writing-code-comments`.
- Alternatives considered and rejected: clamping negative totals to zero (hides the signal and stores a different wrong number), hardcoding per-provider reporting styles (contradicts the documented auto-detect contract and breaks when SDKs change), fixing in the SDKs (an SDK cannot know how an arbitrary OpenAI-compatible endpoint counts cache tokens, and it would not cover OTel or manual capture), and stateful per-integration detection (cross-event state in a stateless pipeline step, disproportionate to the problem).
- Verification: the reporter's stored negative trace total was reproduced to the third decimal from their token counts and the matched pricing entry under the inclusive formula, confirming the mechanism before writing the fix.


richardsolomou (approved):
Could we cover the fallback counter and both `prior` labels? The accounting behavior is well tested, but the new observability could silently regress while these tests remain green.

christiaan-ph:
Added coverage in 4dd712afec9: both fallback flips now assert their `prior` label on the real counter, plus two no-increment guards (a well-formed inclusive event, and an explicit flag with provably-not-inclusive tokens, so the counter can't silently move above the explicit-flag check). Mutation-checked that swapping the two labels fails exactly the new label cases.

PR #74189, opened 2026-07-28
feat(experiments): move hypothesis next to results, cap content width

jurajmajerik:
## Problem

The hypothesis is part of reading an experiment's results, but it sat up in the config area at the top, squeezed into a narrow column. And the experiment content ran edge to edge, so on wide screens the text and tables stretched further than is comfortable to read.

## Changes

Moved the hypothesis into its own box on the Metrics tab, right above Exposures, so it sits next to the results you read it against. Also capped the width of the content under the tabs (the tab bar itself stays full width) so everything stays readable on wide monitors.

| Before | After |
| --- | --- |
| before | after |

Ran it locally and compared before and after on the same experiment (screenshots above). No tests added - it's a layout change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs affected.

## 🤖 Agent context

Juraj directed this; I (Claude) implemented it, iterating on the layout against the live local app and verifying with screenshots.


rubychilds:
@jurajmajerik took a look at this. Agree right now the current space and design is limiting. But my concern with this change, is that the text would become super wide. if we break the space into columns, I think a max with for this, would be about 1/2 of a wide space, and maybe 2/3 on a normal desktop screen. I haven't looked deeply into how others solve this, particularly with wide tables - I know notion limits width, unless you change the setting to wide (maybe that is what we need longer term - layout options for the page)

jurajmajerik:
As discussed capped the tab contents:

image

image

rubychilds (approved):

PR #74196, opened 2026-07-28
fix(data-warehouse): query synced ClickHouse source tables via S3

xljones:
## Problem

A synced (warehouse) ClickHouse data-warehouse source can't be queried in the SQL editor.
Any HogQL that references its tables fails with `Direct queries require selecting a connection.`, even though a synced source has no connection to select.
Hit in prod-EU (project 69647) querying `clickhouse_raffle`.

## Changes

`DataWarehouseTable.hogql_definition()` rendered every ClickHouse-typed source's tables as a `DirectSQLTable` because the ClickHouse branch gated only on `direct_engine` (which keys off `source_type`), unlike the postgres/mysql/snowflake/redshift branches that also require `access_method == direct`.
So a synced ClickHouse source's tables were forced down the direct-query path and rejected.
Added the `is_direct_query` guard so a synced ClickHouse source falls through to the normal S3-backed table; pure-direct ClickHouse sources and the dual-mode virtual-table path are unaffected.

Added a parameterized `SimpleTestCase` (`TestHogqlDefinitionDirectGating`): a `warehouse` ClickHouse source resolves to `HogQLDataWarehouseTable`, a `direct` one to `DirectClickHouseTable`.
It catches a revert of the access-method guard; no existing test covered the engine-vs-access_method gating.
I (Claude) ran it locally, 2 passed. No manual UI testing was done by the agent.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Diagnosed with Xander from a prod-EU SQL editor error. Confirmed the source config via Metabase (ClickHouse, `access_method=warehouse`, `direct_query_enabled=false`) and the failing request via a HAR (no `connectionId` sent).
Skills invoked: `/writing-tests`, `/writing-code-comments`, `/writing-user-facing-copy`.
Two sibling issues surfaced in the same session are being split into their own PRs: the connection selector mislabels ClickHouse sources as "Postgres", and dual-mode ClickHouse execution 500s instead of returning a clean error.

---

### DIFF
--- a/products/warehouse_sources/backend/models/table.py
+++ b/products/warehouse_sources/backend/models/table.py
@@ -750,8 +750,17 @@ class DataWarehouseTable(CreatedMetaFields, UpdatedMetaFields, UUIDTModel, Delet
                 connection_metadata=self.external_data_source.connection_metadata,
             )
 
-        # Engine-keyed (no is_direct_clickhouse) to satisfy the source-agnostic guard.
-        if self.external_data_source and self.external_data_source.direct_engine == "clickhouse":
+        # Requires is_direct_query (access_method == direct) so a synced ClickHouse source falls
+        # through to the S3-backed table below, the same way is_direct_postgres/mysql/etc. gate their
+        # branches. Without it every ClickHouse-typed source (synced included) becomes a direct table,
+        # which forces the direct-query path and rejects synced-table queries that have no connection.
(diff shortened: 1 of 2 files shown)


xljones:
Superseded by #74193, which merged the identical `is_direct_query` guard (with a test that also covers ClickHouseCloud) before this could merge. Closing as a duplicate rather than re-landing the same code.

PR #74199, opened 2026-07-28
fix(data-warehouse): label ClickHouse connections in the SQL editor selector

xljones:
## Problem

In the SQL editor's connection selector, a ClickHouse data-warehouse source shows as "Postgres" with the Postgres icon.
The engine label/icon map had no `clickhouse` entry, so `getConnectionEngine` fell through to its `'postgres'` default.

## Changes

Added `clickhouse` to the `ConnectionEngine` union, the `ENGINE_LABELS` and `ENGINE_ICONS` maps, and the engine detection in `getConnectionEngine`, using the existing `public/services/clickhouse.svg` asset.
A ClickHouse source now renders as "ClickHouse" (or "ClickHouse · synced" for a warehouse source).

No new automated test: the maps are `Record<ConnectionEngine, ...>`, so a missing engine entry is a TypeScript compile error rather than a runtime bug, which is a stronger guard than a unit test would be.
The agent did not capture a screenshot (no running app in this session).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs affected.

## 🤖 Agent context

Split out from a prod-EU SQL editor investigation with Xander. This is the cosmetic sibling of the backend fix in #74196 (synced ClickHouse tables were also wrongly treated as direct).
Skills invoked: `/writing-user-facing-copy`, `/writing-code-comments`.

---

### DIFF
--- a/frontend/src/scenes/data-warehouse/editor/connectionSelectorLogic.ts
+++ b/frontend/src/scenes/data-warehouse/editor/connectionSelectorLogic.ts
@@ -5,6 +5,7 @@ import { ApiConfig } from 'lib/api'
 import { urls } from 'scenes/urls'
 
 import IconPostHog from 'public/posthog-icon.svg'
+import IconClickHouse from 'public/services/clickhouse.svg'
 import IconDuckDB from 'public/services/duckdb.svg'
 import IconMySQL from 'public/services/mysql.png'
 import IconPostgres from 'public/services/postgres.png'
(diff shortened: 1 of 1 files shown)


sakce:
@Gilbert09 has #74019

xljones:
Closing in favour of https://github.com/PostHog/posthog/pull/74019

PR #74211, opened 2026-07-28
feat(signals): let a scout opt into extra sandbox domains

andrewm4894:
## Problem

Scout sandboxes run on a restricted network: the default trusted allowlist (version control, package registries, container registries, PostHog's own APIs) and nothing else. That is a reasonable default for a scout, which is close to the worst-case workload for prompt injection — it ingests attacker-influenceable text by design (Slack messages, warehouse rows, survey responses, PR bodies), runs unattended on a schedule, and has no human reviewing a run before it writes to an inbox.

But it leaves no way for the occasional scout that genuinely has to read a source outside PostHog to do so, and the fleet has been routing around it. `authoring-scouts` documents a "code ∩ the outside world" pattern that says to check a provider's published deprecation schedule "fetched from the web", which nothing could actually execute.

## Changes

An optional `sandbox_allowed_domains` list on `SignalScoutConfig`, the per-`(team, skill)` config row.

- **The default is unchanged.** Empty (the default, and what every existing row gets) leaves the scout on the shared `TRUSTED` environment exactly as before. There is no config value that grants unrestricted egress.
- **Non-empty** moves that one scout to its own `CUSTOM` environment carrying the default trusted hosts **plus** the declared ones (`include_default_domains=True`), so opting in is strictly additive and can never take away a host the scout could already reach.
- **The per-scout environment is load-bearing, not cosmetic.** `get_or_create_signals_sandbox_env` reasserts policy on every run, so two scouts with different allowlists sharing one environment row would rewrite each other's policy run by run. Widened scouts get `SIGNALS_SCOUT_<skill_name>`; naming it after the skill also means a 403 in a sandbox log points at the scout that owns it.

This follows the existing least-privilege pattern rather than inventing one: `SIGNALS_REPO_DISCOVERY` already runs `CUSTOM` with `GITHUB_ONLY_DOMAINS`, and `CONVERSATIONS_SUPPORT_REPLY` runs `CUSTOM` with an empty list. The difference is that those lists are module constants picked by an engineer in a reviewed PR, and this one is set through the config API — so the shape is validated rather than trusted.

Validation accepts exact hostnames only, capped at 20, deduped and normalized (lowercased, trailing dot and whitespace stripped). Rejected: wildcards, schemes, ports, paths, bare TLDs, and IP literals. Wildcards are rejected rather than range-checked because `*.example.com` and `*.co.uk` are the same syntax with very different scope, telling them apart needs the public suffix list, and the two enforcement paths (agentsh, Modal's `outbound_domain_allowlist`) do not treat wildcards identically. This can be loosened later; it can't be tightened once scouts depend on it.

Changes are activity-logged for free — `SignalScoutConfig` is already on `ModelActivityMixin` because its fields drive spend, and this one moves a security boundary.

Docs updated in the same PR: the harness `AGENTS.md`, and the two stale network claims in `authoring-scouts` (the "External systems" row and the "code ∩ the outside world" pattern that promised a web fetch nothing could perform).

Automated tests only. I did not exercise a real sandbox — nothing here was verified against a live agentsh or Modal run, so the claim that a declared domain is actually reachable end to end is **untested**. A reviewer with sandbox access should confirm that before anyone relies on it, on both enforcement paths, since they differ (`provision_sandbox.py:584`).

Ran locally: `products/signals/backend/test/test_scout_harness.py` and `test_scout_harness_api.py` (255 passed), repo-wide `uv run mypy --cache-fine-grained .` (clean, what CI runs), and `hogli ci:preflight --fix` (0 failures). Migration checked with `sqlmigrate`: a single `ADD COLUMN ... DEFAULT '{}' NOT NULL`, no stray `DROP DEFAULT`, on a product table rather than a hot one.

Tests added, and the regression each catches:

- `TestSandboxEnvResolution` (2 cases) — that an empty list still resolves to the shared `TRUSTED` environment with no extra kwargs, and that a non-empty list resolves to a differently-named `CUSTOM` environment. The first is the one that matters: it locks the unchanged default so a later edit to this function can't silently alter every scout's network policy. The second asserts the name differs from the shared one, which is the invariant that stops one scout's widening from rewriting another's policy.
- `TestSandboxAllowedDomainsValidation` (20 parameterized cases) — that wildcards, schemes, ports, paths, bare TLDs, IP literals, and over-cap lists are rejected, and that accepted input is normalized and deduped. Guards against the validator being loosened so that `*` or a URL reaches a sandbox allowlist. No DB (`SimpleTestCase`), since field validation runs entirely in memory.
- Two endpoint cases on the config `PATCH` — round-trip and a 400 on an invalid domain. These are the wiring guard: the `SimpleTestCase` matrix proves the validation logic but not that the viewset invokes this serializer.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Covered in this PR (harness `AGENTS.md` + `authoring-scouts` references). No external docs site changes.

## 🤖 Agent context

Written with Claude Code (Opus 5) at Andy's direction, coming out of a design discussion on the scouts backlog about what network access scouts should get by default.

The framing moved twice during that discussion. It started as "scouts are missing a reference-domain tier, widen `DEFAULT_TRUSTED_DOMAINS`", which was wrong: that constant is shared with every task sandbox, so editing it widens far more than scouts. It then became "default scouts tighter and add an opt-in", which Andy explicitly scoped back down — the default stays exactly as it is, and this is purely an optional field. Two things surfaced while reading the code that shaped the result: the scout env is shared by name with the report-research and custom-agent workloads, and its policy is reasserted per run, which is why a widened scout must get its own environment row rather than different kwargs on the shared one.

The rejected alternative worth naming is putting the declared domains in the scout's skill body. Scouts are shared skills, so an author's list would then widen egress on every team that enables the scout, with no reviewer — the skills store is a database write, not a PR. Keeping the grant on the per-team config row means a widening is always scoped to one team and always attributable. A future refinement could let a skill *request* domains that a config then grants, which would give the UI a "this scout is asking for X" affordance; that isn't in this PR.

Skills invoked: `/django-migrations`, `/improving-drf-endpoints`, `/writing-tests`, `/writing-code-comments`, plus `/phs scouts-work` for the backlog context (issue 18).

Known gap, repeated from above because it decides how much this is worth: I could not establish whether the sandbox runtime exposes a web-fetch tool at all. There is no `WebFetch`/`WebSearch` wiring anywhere in `scout_harness/` or `tasks/backend/`, so a widened scout may only be able to reach these hosts via `curl`. If the runtime instead exposes a provider-side *search* tool, that executes off-sandbox and this allowlist doesn't gate it. Worth confirming from a real run's agent logs before promoting this field to users.


andrewm4894:
Moving this to draft while the layer underneath it settles.

The config half (the \`sandbox_allowed_domains\` field on \`SignalScoutConfig\`, its validation, and the docs) should hold regardless. The enforcement half is the part to revisit: it resolves a \`SandboxEnvironment\` and leans on the agentsh / Modal allowlist, and that mechanism is in flux for scout runs. Enforcement is also not uniform across runtimes today (see the `vm_runtime` caveat at `products/tasks/backend/temporal/process_task/activities/provision_sandbox.py:584`), so the right shape may not be a per-scout environment at all.

Worth pairing with the separate question of a web-search tool for the harness. A domain allowlist on its own only gets a scout as far as `curl`; there is no fetch tool in `scout_harness/` today, so the two are halves of the same capability.

Context and the wider discussion: https://posthog.slack.com/archives/C09G8Q32R6F/p1785241568540189

PR #74213, opened 2026-07-28
docs(replay): point at where the in-flight count is set

robbie-c:
## Problem

`scrub-client.ts` described the jitter as keeping "a pod's **eight** in-flight images" from re-posting in lockstep. That count is `SESSION_RECORDING_ML_IMAGE_SCRUB_SCRUB_CONCURRENCY`, set per deployment and tracking the sidecar's worker count — [PostHog/charts#13595](https://github.com/PostHog/charts/pull/13595) has since moved it to 14, so the comment was already wrong.

## Changes

Replace the number with a pointer to where it is actually set. A reader who needs the current value can follow it; a reader who only needs the reasoning does not have to trust a stale figure.

This matches how the neighbouring code already handles deployment-owned values — `ml-mirror/config.ts` and `produce-collected-images-step.ts` both cite the chart file they depend on rather than restating its contents.

Comment-only, no behaviour change. Caught in review of the charts PR that invalidated it.


arnohillen (approved):
Robot review:

Good trade. A number in a comment goes stale, a pointer to where it's set doesn't.

The failing Django, Frontend and VR jobs can't be from this diff, it's comment-only.

PR #74219, opened 2026-07-28
fix(review-hog): serve code review at /code-review not /code_review

andrewm4894:
## Problem

`/code_review` was the odd one out. Nearly every route in the app is kebab-case (`data-management`, `error-tracking`, `feature-flags`…), and the underscore stuck out enough to get flagged. Raised in a Slack thread — the maintainer was fine with the rename.

## Changes

Rename the Code review scene route from `/code_review` to `/code-review` (manifest + regenerated `products.tsx`).

Add a redirect from the old `/code_review` path that forwards query params. The `?review=` / `?reviews_scope=` params are a permanent public contract baked into GitHub PR status comments, so the redirect carries them across and those links keep opening the right report. This mirrors the existing `/feature-flags` → `/feature_flags` redirect precedent (though in the opposite direction).

Also updated the backend deep-link builder, the MCP tool `url_prefix` and its generated output, and the architecture doc. The `DECISIONS.md` log keeps the historical `/code_review` mentions since it records decisions as they were made.

> [!NOTE]
> Old `/code_review` links (including those already sitting in GitHub PR comments) redirect to `/code-review` with their `?review=` intact, so nothing breaks.

Added a redirect test in `sceneLogic.test.tsx` asserting `/code_review?review=r-9` lands on `/code-review` with the `review` param preserved — catches the regression where the permanent PR-comment deep links break or silently drop the param. Updated the backend `test_status_comment.py` assertions to the new path.

I (Claude, via the PostHog Slack app) could not run the JS/Python suites in this environment — no `node_modules`/toolchain. `products.tsx` and the generated MCP file were hand-edited to match their generators; CI preflight regenerates and verifies consistency.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Directed by Andy Maguire from a Slack thread that pointed out the underscore inconsistency. I checked `urls.ts` first and confirmed hyphens are the dominant convention (~9% of paths use underscores, mostly legacy `/instance/*` admin routes), then made the change.

Key decision: keep a redirect rather than a clean rename, because the `?review=` deep link is documented in-code as a permanent public contract embedded in GitHub PR comments — a bare rename would 404 those. The redirect forwards all incoming search params via `combineUrl` so both `review` and `reviews_scope` survive. Left `iconType: 'code_review'` and the signals `code_review` artefact type untouched — those are enum/icon identifiers, not URLs.

No repo skills were invocable in this environment; applied the test value-gate manually.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1785242874571769?thread_ts=1785242874.571769&cid=C09SK2PAGKF)*

### DIFF
--- a/frontend/src/products.tsx
+++ b/frontend/src/products.tsx
@@ -191,7 +191,7 @@ export const productRoutes: Record<string, [string, string]> = {
     '/replay-vision/:id/triggers': ['ReplayVisionScannerEditor', 'replayVisionScannerTriggers'],
     '/replay-vision/:id/self-driving': ['ReplayVisionScannerEditor', 'replayVisionScannerSelfDriving'],
     '/replay-vision/:id': ['ReplayVisionScanner', 'replayVision'],
-    '/code_review': ['CodeReview', 'codeReview'],
+    '/code-review': ['CodeReview', 'codeReview'],
     '/session-summaries': ['SessionGroupSummariesTable', 'sessionGroupSummariesTable'],
     '/session-summaries/:sessionGroupId': ['SessionGroupSummary', 'sessionGroupSummary'],
     '/skills': ['Skills', 'skills'],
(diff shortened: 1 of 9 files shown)


sortafreel (approved):

tatoalo (approved):
mkay

PR #74221, opened 2026-07-28
fix: enforce prop access control across all supported tables

ReeceJones:
## Problem

Some tables were excluded from property access control because there was not a central registry listing all tables for which it should be enforced.

## Changes

Centralize the list of tables for which property access control should be enabled.

Tests

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context


posthog/hogql/restricted_properties.py
```diff
@@ -34,13 +50,26 @@ def restricted_property_keys_for_table_type(table_type: ast.Type, context: HogQL
...
+    # they carry person/group properties — match them before the EventsTable branch either way.
     if isinstance(table, EventsPersonSubTable):
         prop_def_type = PropertyDefinition.Type.PERSON
+    elif isinstance(table, EventsGroupSubTable):
+        prop_def_type = PropertyDefinition.Type.GROUP
     elif isinstance(table, EventsTable):
         prop_def_type = PropertyDefinition.Type.EVENT
     elif isinstance(table, (PersonsTable, RawPersonsTable)):
         prop_def_type = PropertyDefinition.Type.PERSON
+    elif isinstance(table, (GroupsTable, RawGroupsTable)):
```
a-lider: +1 to the hex comment.
`system.groups` exposes the same blob as a `PostgresTable` with no `access_scope`, and the postgres printer has no JSONDropKeys equivalent, so a restricted group property is still readable via `SELECT group_properties FROM system.groups`.

posthog/hogql/restricted_properties.py
```diff
@@ -34,13 +50,26 @@ def restricted_property_keys_for_table_type(table_type: ast.Type, context: HogQL
...
+        # Group property definitions are per group type index, but `context.restricted_properties` carries only
+        # (name, type) — so a restricted name is declined for every group type. Over-restricting one group type's
+        # property on another is the fail-closed direction; the alternative leaks the restricted value.
+        prop_def_type = PropertyDefinition.Type.GROUP
     else:
+        # PropertyDefinition.Type.SESSION is deliberately absent: the sessions tables expose each session property as
+        # its own column rather than a JSON blob, so there is nothing for this function's callers to scrub. Restricting
+        # a session property therefore has no query-time effect yet — enforcing it needs field-level denial, not a
+        # blob-key drop.
         return set()
```
a-lider, line 76: nit: `upsert_property_access_control` doesn't validate the property type, so an admin can still create a rule on a session property that silently does nothing. Not for this PR, but might be worth fixing.

a-lider (approved):
The hex system.groups comment looks like a gap worth resolving before merge. Otherwise looks good.

ReeceJones:
I forgor about this PR. Will fix it up and ship

posthog/hogql/restricted_properties.py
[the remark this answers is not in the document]
ReeceJones: Fixed in 85877af408abc7dbbf177d5229828c1069a7f7f4. The Postgres-backed system.groups table now uses group property restrictions, with coverage for direct key and whole-blob reads.

ReeceJones: Fixed in 85877af408abc7dbbf177d5229828c1069a7f7f4. system.groups now receives the same restricted-key scrubbing, with regression tests for direct and whole-blob reads.

posthog/hogql/restricted_properties.py
ReeceJones: Agreed. Session properties need separate field-level enforcement because they are exposed as columns rather than through a JSON blob. That is outside this PR’s blob-scrubbing scope.

georgemunyoro (approved):

PR #74228, opened 2026-07-28
refactor(capture): outputs plan, routing goldens, pure route(), output registry

pl:
## Problem

> [!WARNING]
> This PR now also carries the overflow and rate-limit parity fixes from #78014, folded in
> at my request. Two of those commits change production wire behavior. See "Overflow and
> rate-limit parity" below.

`rust/capture`'s produce path mixes routing policy, serialization, and transport in one place, which makes automatic failover, cluster migrations without code changes, and payload format evolution hard and risky. #72691 is the combined draft of the outputs refactor that fixes this; per its plan doc, each step graduates to its own standalone PR.

This is the first graduation: the working contract (plan doc) plus the groundwork steps — the parity instrument and the first two mechanical extractions.

## Changes

Four commits, each green on its own:

1. **Plan doc** (`rust/capture/OUTPUTS_REFACTOR_PLAN.md`): the architecture, per-step goals, parity proofs, and progress tracker for the whole refactor, including the reconciliation notes for the AI-events topic routing and Import capture mode that landed on master while the draft was in flight.
2. **Routing golden oracle consolidation**: every routing case in the kafka sink tests now pins the full fingerprint of a produced record — topic, partition-key presence, person-processing/heatmap headers, content-encoding, the DLQ header set, and which reroute counter fires. The standalone DLQ header tests fold into their routing cases. This suite is the oracle every later refactor step proves wire parity against, unmodified.
3. **Pure `route()`**: the routing decision (target, key policy, header/metric effect) becomes a pure function over `ProcessedEventMetadata` plus the AI overflow valve, replacing the inline topic/key match and `route_with_overflow`. The sink resolves the decision; outcomes are byte-identical (pinned by the goldens).
4. **`OutputRegistry` + startup completeness check**: the output→topic wiring moves into one registry; `KafkaSink::new` refuses to boot when any registered output resolves to an empty topic, so a misconfig fails at startup instead of at first produce. The opt-in AI topics stay outside the completeness check (setup already validates them against the routing mode). Drops the unused traces topic field.

### Overflow and rate-limit parity (folded in from #78014)

The v0 and v1 pipelines disagreed on how the overflow limiter and the global rate limiter combine. Each path had tests asserting the opposite of the other.

| input | v0 before | v1 before | now |
| --- | --- | --- | --- |
| burst rate-limited, locality off | key dropped, person processing on | key dropped, person processing **off** | see the two fixes below |
| globally limited plus burst, locality on | key **kept** | key dropped | key dropped on both |

5. **v1 stops using a wire header as a control channel**: the v1 sink read its partition-key decision out of `headers.force_disable_person_processing`, so asking for a partition spread also told downstream to skip person processing. `spread_partitions` now carries the partitioning intent and `Event::ordering()` carries the guarantee. `OrderingGuarantee` moves to `crate::ordering` so both paths name one type.
6. **Person processing off drops the ordering guarantee on both paths**: legacy's burst limiter overwrote the rate limiter's stamp, so a key the rate limiter declared too hot went back to hashing onto one partition.
7. **Keyless records never reach a person-writing lane**: the analytics main and overflow consumers write persons keyed on distinct id, so a keyless record with person processing on hands the same person rows to several workers. `Destination::writes_persons` scopes the spread to lanes without person writes.
8. **`ForceLimited` implies person processing off on its own**: `person_processing_disabled` derives the skip from the flag or the reason, so a stamping site that forgets the flag cannot produce a keyless person-on record.
9. **Cross-path parity suite** (`rust/capture/src/overflow_parity.rs`): drives both pipelines over the overflow and rate-limit matrix and pins lane, partition-key presence, and the person-processing header.

> [!WARNING]
> Two production wire changes, both narrow and both deliberate.
>
> **prod-US** (`OVERFLOW_PRESERVE_PARTITION_LOCALITY` on): a key over the rate-limit threshold that also bursts the overflow limiter moves from keyed to round-robin on the overflow topic. Watch overflow partition skew.
>
> **prod-EU** (locality off): a burst-limited key whose person processing stays on moves from keyless to keyed, so it concentrates on one overflow partition instead of spreading. That trade avoids contended person writes.

No wire behavior changes: routing goldens and all existing endpoint/integration tests pass unmodified (except the test-harness consolidation itself in commit 2).

- `cargo test -p capture` — full suite (lib + all integration binaries) green after each commit.
- `cargo clippy -p capture --all-targets -- -D warnings` and `cargo fmt` clean.
- New tests: the consolidated oracle now catches header/counter regressions per routing case that only spot-check tests caught before (e.g. DLQ header set on every DLQ-routed lane, reroute counters firing exactly once); `route_tests` pin the pure decision including the AI overflow valve gating; `create_sink_refuses_boot_on_missing_output_topic` pins the boot-time completeness check per capture mode; the registry tests pin AI-topic fallback behavior.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

The plan doc in this PR is the refactor's own documentation; no user-facing docs are affected.

## 🤖 Agent context

Authored with Claude Code, graduating steps 0–3 of the combined draft PR #72691 onto current master. The original branch commits predate the `$ai_*` topic routing (#68719) and Import mode (#71800) on master, so the oracle consolidation, `route()`, and the registry were re-derived to cover the AI lanes and the overflow valve rather than cherry-picked. Decision of note: the AI outputs are excluded from the registry's boot completeness check because they are opt-in (setup validates them against `CAPTURE_ANALYTICS_AI_EVENTS_MODE`), and `route()` takes the valve as an explicit input to stay pure.


rust/capture/src/sinks/kafka.rs
```diff
@@ -236,50 +195,388 @@ impl<P: KafkaProducer> Clone for KafkaSinkBase<P> {
...
+            // Precedence: force_overflow (restrictions) -> overflow_reason
+            // (pipeline-stamped) -> default main-topic routing.
+            if metadata.force_overflow {
+                Route {
+                    target: Outputs::AnalyticsOverflow,
+                    ordering: person_ordering(metadata.skip_person_processing),
+                }
+            } else {
+                match &metadata.overflow_reason {
+                    Some(OverflowReason::ForceLimited) => Route {
```
eli-r-ph, line 256: Dropping the sink's redundant `set_force_disable_person_processing(true)` here is a no-op today — I checked both production stamping sites (`events/overflow_stamping.rs:109`, `events/analytics.rs:497`) and each sets the flag before stamping the reason. But the failure mode if that ever drifts is hot-key events reaching person processing with a null partition key, which is exactly what overflow exists to prevent, and it fails silently.

Worth making it structurally impossible rather than documented: carry the flag inside `OverflowReason::ForceLimited`, or `debug_assert!(metadata.skip_person_processing)` in this arm.

rust/capture/src/sinks/registry.rs
```diff
@@ -0,0 +1,296 @@
...
+                &self.main
+            }
+        }
+    }
+
+    /// Startup completeness check: every registered output must resolve to a
+    /// non-empty topic, so a misconfigured or newly-added-but-unwired output
+    /// fails fast at boot instead of at first produce. `Custom` is excluded (it carries its own topic per event), as
+    /// are the opt-in AI outputs (validated by setup against the routing mode).
+    pub fn check_complete(&self) -> anyhow::Result<()> {
```
eli-r-ph, line 189: `SessionReplayMain` resolves to `main`, and all 9 registered outputs are demanded regardless of `CaptureMode`, so a recordings or import pod must have the analytics/overflow/historical/heatmaps/warnings/error-tracking topics non-blank even though it can never produce to them.

Every one of those has a non-empty envconfig default, so this is safe as shipped. The risk is arming `CAPTURE_OUTPUTS_COMPLETENESS_CHECK_ENABLED` against a deployment that explicitly sets one to `""` — that's a fleet crashloop, not a degradation. Suggest landing the mode-scoping (plan Step 9) before arming it anywhere, and auditing the chart env for blank topic values first.

rust/capture/src/sinks/registry.rs
```diff
@@ -0,0 +1,296 @@
...
+    AiOverflow,
+    /// Admin-configured custom topic borrowed from `redirect_to_topic`. Resolved
+    /// inline by the sink; never registered (it carries its own topic).
+    Custom(&'a str),
+}
+
+impl Outputs<'_> {
+    /// Every registered always-required output. `check_complete` walks this so
+    /// a newly added output is caught at boot rather than at first produce.
+    const REGISTERED: [Outputs<'static>; 9] = [
```
eli-r-ph: The module doc says adding an output is a one-place change caught at boot, but only `topic_for`'s match is compiler-exhaustive — `REGISTERED` is a hand-maintained array. A new variant compiles fine and silently skips the completeness check.

An exhaustive `fn is_required(&self) -> bool` plus a test asserting the array length equals the required-variant count would make the guarantee real.

rust/capture/src/sinks/kafka.rs
```diff
@@ -236,50 +195,388 @@ impl<P: KafkaProducer> Clone for KafkaSinkBase<P> {
...
+            // skip_person_processing (v1 only nulls keys for
+            // Main/Overflow-shaped destinations). AI events never reroute
+            // historical.
+            if ai_events_overflow_armed && metadata.force_overflow {
+                Route {
+                    target: Outputs::AiOverflow,
+                    ordering: person_ordering(metadata.skip_person_processing),
+                }
+            } else if ai_events_overflow_armed {
+                match &metadata.overflow_reason {
```
eli-r-ph, line 303: This four-arm match is a verbatim copy of the `AnalyticsMain` one above (line 279). The deleted `route_with_overflow` carried a docstring saying it existed specifically so the two lanes could not diverge — that guarantee is now two copies a future edit can drift apart.

A small shared helper taking the (overflow, main) target pair would restore it without giving up the purity of `route()`.

rust/capture/src/sinks/kafka.rs
```diff
@@ -408,7 +705,16 @@ impl KafkaSink {
             info!("connected to Kafka brokers");
         };
 
-        let topics = Arc::new(KafkaTopicConfig::from(&config));
+        // Refuse to boot on incomplete output wiring: a blank topic fails
+        // here, at startup, instead of at first produce. Config-only — the
+        // broker is never probed.
+        let registry = OutputRegistry::from(&config);
+        if config.outputs_completeness_check_enabled {
```
eli-r-ph: The config-only check runs after the `fetch_metadata(..., Timeout::After(10s))` ping above, so a blank topic costs a full broker connect attempt before the refusal — fail-fast in intent, fail-after-connect in practice. Moving `check_complete()` to the top of `KafkaSink::new`, before the producer is built, makes it instant.

rust/capture/src/sinks/kafka.rs
```diff
@@ -1327,56 +1553,182 @@ mod tests {
...
+                    match value {
+                        metrics_util::debugging::DebugValue::Counter(v) => Some(*v),
+                        _ => None,
+                    }
+                })
+            };
+            let dlq_count = count("capture_events_rerouted_dlq");
+            let custom_count = count("capture_events_rerouted_custom_topic");
+            match expected.rerouted {
+                Rerouted::None => {
```
eli-r-ph, line 1782: `Rerouted::None` asserts the metric is *absent*, which passes both when the counter didn't fire and when the thread-local recorder never saw it. Since the recorder only works while `prepare_record` runs inline on the test thread (per the comment at line 1596), and plan Steps 6/12 explicitly move prep into a hoisted/spawned path, the negative cases go quietly vacuous exactly when this oracle is supposed to be catching drift.

Snapshotting the counter before and after and asserting a delta of 0 would fail loudly instead.

rust/capture/src/sinks/kafka.rs
```diff
@@ -408,7 +705,16 @@ impl KafkaSink {
...
 
-        let topics = Arc::new(KafkaTopicConfig::from(&config));
+        // Refuse to boot on incomplete output wiring: a blank topic fails
+        // here, at startup, instead of at first produce. Config-only — the
+        // broker is never probed.
+        let registry = OutputRegistry::from(&config);
+        if config.outputs_completeness_check_enabled {
+            registry.check_complete()?;
+        } else {
+            warn!("outputs completeness check disabled; a blank output topic will fail at first produce instead of at boot");
```
eli-r-ph: With the flag off (the shipping default), every capture pod logs this at startup with nothing the reader can act on. `info!` — or dropping it — avoids training people to ignore capture boot warnings.

rust/capture/OUTPUTS_REFACTOR_PLAN.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Stage A — oracle and groundwork
+
+#### Step 1 · Consolidate the routing golden oracle
+
+The `assert_routing` suite pins topic + partition key + every stamped header + the rerouted counters for all pipelines and lanes. It is the parity instrument for every later step.
+
+#### Step 2 · Extract routing into a pure `route()`
+
+`route(&ProcessedEventMetadata) -> Route { target, key_policy, effect }`, consulted by the sink. This is the function Step 5 hoists out of the sink into lane resolution.
```
eli-r-ph: The doc is already stale against the code it ships with: the shipped shape is `Route { target, ordering }` plus an `ai_events_overflow_armed` argument, not `Route { target, key_policy, effect }`. Same for Step 3 (line 71), which doesn't mention the config gate the last commit added, and Steps 5 / the repartitioning note (lines 89, 91, 211), which reference `KeyPolicy`, `RouteEffect`, and `KeyPolicy::Null` — none of which exist.

Since this is the working contract for 17 more steps, worth a reconciliation pass. Also worth deciding now who deletes it at Step 20 — a 274-line plan at a crate root will outlive its usefulness.

rust/capture/OUTPUTS_REFACTOR_PLAN.md
```diff
@@ -0,0 +1,274 @@
...
+- **Output** = the addressed destination: targets + selection policy + serializer. **OutputTable** = the `(Pipeline, Lane)` → output mapping with the per-mode completeness check. The Step-3 `OutputRegistry` (output → topic) is absorbed into the table's targets.
+- **Sink** = backend mechanism. Kafka, S3, print, noop. Nothing else is a sink; a thing that picks between sinks is an output policy.
+- **Serializer** = format × envelope. Formats and envelopes compose; neither knows about sinks or outputs.
+
+### Invariants
+
+- Metric names and labels stay stable: `capture_events_rerouted_*`, `capture_primary_sink_health`, `capture_fallback_sink_failovers_total`, `capture_split_sink_selected`, `capture_failover_breaker_*`, `capture_event_batch_size`. Renames are out of scope for every step.
+- Wire parity is proven by the Step-1 goldens and existing endpoint/integration tests passing **unmodified**, except where a step explicitly says otherwise.
+- Never mix a mechanical move with a behavior change in one commit.
```
eli-r-ph, line 52: The `route decides output and ordering guarantee only` commit breaks this invariant: it restructures `Route` and removes the ForceLimited header set in one go. The commit message is honest about it, but splitting the header removal into its own commit would make the one wire-semantics change in this PR revertable on its own.

rust/capture/src/sinks/registry.rs
```diff
@@ -0,0 +1,296 @@
...
+/// Which configured output a routing decision selects, named **pipeline +
+/// lane** — the vocabulary the refactor converges on (typed per-pipeline
+/// lanes; see the plan doc). The sink resolves each output to a concrete
+/// topic string against the [`OutputRegistry`]; distinct outputs may share a
+/// topic (analytics main and session-replay main both resolve the
+/// deployment's main topic today). Mirrors v1's `Destination` split — the
+/// convergence target when the v1 stack folds onto this registry (see the
+/// plan doc).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Outputs<'a> {
```
eli-r-ph: Naming and shape nits, batched — these propagate through 17 more steps so cheaper to settle now:

- `Outputs` is plural for a type that names exactly one output; `Output` reads better everywhere it appears.
- `Custom(&'a str)` forces a lifetime parameter onto the whole enum and an awkward `Outputs<'static>` in `REGISTERED`. A `Cow<'a, str>`, or keeping the custom topic outside the enum, would keep the registry API lifetime-free.
- `sinks/kafka.rs:816`: `let route = route(...)` shadows the function name.
- `route()` can return `PerSession` for metadata with no `session_id`, i.e. a "pure" decision the sink then can't realize (`MissingSessionId` surfaces later at key resolution). Folding the reject into the decision would make `Route` always realizable.

eli-r-ph (approved):
A couple of things worth updating (a few important ones) but nothing blocking; approving to unblock merge after an update pass 👍 

I also found a parity bug between legacy and v1 wrt partition key handling between the old OverflowLimiter and the new global rate limiter. [This PR](https://github.com/PostHog/posthog/pull/78014) (stacked on yours since it touches some of the same code) addresses it, getting the behavior into parity for legacy and v1.

You can fold it into your PR here or I can rebase after you land yours.

rust/capture/src/sinks/registry.rs
pl: Yeah, 100%

rust/capture/src/sinks/kafka.rs
pl: Fixed structurally in `7d24030be37` on #78014 (stacked on this PR), where the person-writing-lane rule made this the last representable keyless-person-on path: `ProcessedEventMetadata::person_processing_disabled()` derives the skip from the flag **or** a `ForceLimited` reason, and both the sink's header derivation and `route()`'s `person_ordering` sites consult it. A stamping site that forgets the flag now produces header + keyless together (fail-safe) instead of keyless with person processing silently on. Stamping sites still set the flag for pipeline-level readers, and the goldens now pin the implication: reason-only ⇒ header present.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/kafka.rs
pl: Moved in `15f064637e5`: the check now runs at the top of `KafkaSink::new`, before the producer is built — refusal is instant.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/kafka.rs
pl: Downgraded to `info!` in `15f064637e5`.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/registry.rs
pl: All four taken, in three commits:

- `e85cfe08ddf`: `Outputs` → `Output`.
- `985105029b4`: `Custom(String)` — custom redirects are a rare admin path, so one clone per redirected event buys a lifetime-free registry API (`Cow` would have kept the parameter).
- `7f693bcf101`: `route()` returns `Result<Route, CaptureError>` — a replay event with no session_id is rejected in the decision, so every returned `Route` is realizable; same error, one layer earlier. The shadowing binding became `decision` in the same commit.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/registry.rs
pl: Recorded the arming precondition in the plan's Step 9 (`a94922675ba`): the flag stays off everywhere until mode-scoping lands and the chart env is audited for explicitly-blank topics.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/registry.rs
pl: Done in `0c6d1f5e5a9`: exhaustive `is_required()` (cfg(test) — its only caller is the pin) forces every new variant to declare its side at compile time, and `registered_is_exactly_the_required_outputs` asserts membership-iff-required over an all-variants list. That list is the one hand-maintained enumeration left, so the guarantee is belt-and-braces rather than airtight — a macro generating both from one list felt like more cleverness than scaffolding that Step 7/9 rebuilds deserves.

_— written by Claude, posted on Paweł's behalf_

rust/capture/OUTPUTS_REFACTOR_PLAN.md
pl: Reconciled in `36023785329`: Steps 2/5 now state the shipped shape (`Result<Route { target, ordering }, CaptureError>` with the valve as an explicit argument; `OrderingGuarantee` as the decision data), Step 3 mentions the config gate, the repartitioning note uses `OrderingGuarantee::None`, and deletion is decided — the 20c commit retires the doc, with anything load-bearing graduating to module docs or DESIGN.md first.

_— written by Claude, posted on Paweł's behalf_

rust/capture/OUTPUTS_REFACTOR_PLAN.md
pl: Acknowledged — the invariant was breached here, and it stands for the remaining steps. Leaving the history as-is though:

- The split's payoff was independent revertability, but `7d24030be37` (#78014) has since re-decided that exact semantic at the read site (`ForceLimited` ⇒ header, derived) — a split-out commit would describe a state that no longer exists, and its revert would conflict with the derivation.
- The behavior change was a production no-op (both stamping sites set the flag), pinned by goldens either way, so the blast radius the invariant protects against was minimal here.
- A rewrite force-restacks this branch plus everything stacked on it, and orphans the existing review anchors.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/kafka.rs
pl: Fixed in `94104eb1cc9`, with one deviation from the snapshot-delta suggestion: a detached thread-local recorder yields an empty snapshot before *and* after, so delta-0 is equally vacuous. Instead `assert_routing` now asserts the recorder observed `capture_kafka_produce_bytes_total` — which enqueue emits unconditionally on the same inline path as the reroute counters — in every case. A recorder that goes deaf (prep hoisted to spawned threads) fails all goldens loudly at that line, so the absence assertions can never pass vacuously.

_— written by Claude, posted on Paweł's behalf_

rust/capture/src/sinks/kafka.rs
```diff
@@ -236,50 +196,452 @@ impl<P: KafkaProducer> Clone for KafkaSinkBase<P> {
...
+                    // directions. A burst keeps its key while person processing
+                    // is on — the overflow consumer updates persons keyed on
+                    // distinct id, so spreading one distinct id across
+                    // partitions turns a hot key into contended person-row
+                    // updates — which makes the locality preference irrelevant
+                    // on this lane. And a key whose person processing is
+                    // already off (the global rate limiter stamps its verdict
+                    // before the overflow limiter overwrites the reason) must
+                    // not get its partition back.
+                    Some(OverflowReason::RateLimited { .. }) => Route {
```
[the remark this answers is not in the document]
eli-r-ph: this would undo Pawel's changes to the overflow vs. GRL person processing stamp behavior. Skipping it

PR #74229, opened 2026-07-28
feat(aio): add Eve observability support

Radu-Raicea:
## Problem

Eve emits Vercel AI SDK OpenTelemetry spans, but PostHog currently identifies them as Vercel spans and does not map `eve.session.id` into AI session grouping. An Eve root turn that only carries `eve.*` attributes can also be filtered before normalization.

Eve creates `ai.eve.turn` inside a Vercel Workflow `step.execute` span. The PostHog exporter filters the Workflow span, so the Eve turn otherwise retains a parent ID that PostHog never receives. Eve also places the session ID on the turn root, while generation and tool children do not carry `$ai_session_id`.

## Changes

- Accept Eve root spans through the existing Vercel AI provider path.
- Normalize `ai.eve.turn` as the logical trace root when its filtered Workflow parent is absent.
- Identify Eve root spans with `$ai_framework = "eve"` and map `eve.session.id` to `$ai_session_id`.
- Resolve session trace IDs before aggregation so generations, tools, tokens, costs, and latency remain in the session when only the root carries the session ID.
- Add Eve to the AI observability integration picker with setup guidance for `PostHogTraceExporter`.
- Install the OpenTelemetry HTTP exporter required by the `@posthog/ai/otel` optional peer dependency.
- Reuse the existing Vercel AI normalizer, PostHog OTLP endpoint, and Vercel logo.

### Before

```mermaid
flowchart LR
    Eve["Eve OTel spans"] --> Filter["AI OTLP provider filter"]
    Filter --> Result["Root turn may be dropped<br/>or retain a missing parent<br/>child events omitted from session"]
    classDef phBlue fill:#1d4aff,stroke:#1d4aff,color:#fff
    classDef phRed fill:#f54e00,stroke:#f54e00,color:#fff
    classDef phYellow fill:#f9bd2b,stroke:#f9bd2b,color:#000
    classDef phGray fill:#e5e7eb,stroke:#c7ccd1,color:#000
    class Eve phYellow
    class Filter phBlue
    class Result phRed
```

### After

```mermaid
flowchart LR
    Eve["Eve OTel spans"] --> Filter["Vercel and Eve provider filter"]
    Filter --> Normalize["Logical Eve trace root"]
    Normalize --> Session["Trace-aware session aggregation"]
    Session --> Result["Complete Eve session trace"]
    classDef phBlue fill:#1d4aff,stroke:#1d4aff,color:#fff
    classDef phRed fill:#f54e00,stroke:#f54e00,color:#fff
    classDef phYellow fill:#f9bd2b,stroke:#f9bd2b,color:#000
    classDef phGray fill:#e5e7eb,stroke:#c7ccd1,color:#000
    class Eve,Result phYellow
    class Filter,Normalize,Session phBlue
```

- Node middleware suite: 51 tests passed, including a realistic Eve turn with a filtered Workflow parent and `functionId`.
- Session query suite: 8 tests passed against the dedicated AI events table and events fallback. The regression fixtures put the session ID only on the root and verify child generations, tools, tokens, cost, and latency.
- Rust OTLP provider suite: 9 tests passed.
- Frontend formatting and full TypeScript check passed.
- Repository-wide mypy passed across 17,018 source files.
- `hogli ci:preflight --fix` completed with no failures.
- I did not run manual browser testing.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Added in-product Eve onboarding based on the current Vercel Eve instrumentation documentation. The setup installs the optional OpenTelemetry HTTP exporter required by `@posthog/ai/otel` and links to the PostHog Eve installation page.

## 🤖 Agent context

Codex implemented and validated this change under human direction. Tools used: PostHog docs search, upstream source inspection, web documentation lookup, and GitHub CLI. The local Codex task does not expose a shareable session URL.

Skills invoked: `/llma-git-workflow`, `/writing-tests`, `/writing-user-facing-copy`, `/writing-code-comments`, and `/posthog-efficient-validation`.

The implementation was checked against the current Eve turn-span source, Vercel Workflow parent-span behavior, `@vercel/otel` 2.1.3, and the published `@posthog/ai` 8.4.1 exporter contract.


frontend/src/scenes/onboarding/legacy/sdks/allSDKs.tsx
```diff
@@ -225,6 +225,14 @@ export const ALL_SDKS: SDK[] = [
         image: 'https://res.cloudinary.com/dmukukwp6/image/upload/vercel_373fa70879.svg',
         docsLink: 'https://posthog.com/docs/ai-observability/installation/vercel-ai',
     },
+    {
+        name: 'Eve',
+        key: SDKKey.EVE,
+        tags: [SDKTag.FRAMEWORK],
+        searchTerms: ['Vercel Eve'],
+        image: 'https://res.cloudinary.com/dmukukwp6/image/upload/vercel_373fa70879.svg',
+        docsLink: 'https://vercel.com/docs/eve/observability#export-ai-sdk-spans-with-opentelemetry',
```
carlos-marchal-ph: Every other SDK in this list links to a posthog.com/docs page. Are we okay with having providers where the extended docs are fully external? I have nothing against it but just double checking this is on purpose.

frontend/src/scenes/onboarding/legacy/sdks/allSDKs.tsx
```diff
@@ -225,6 +225,14 @@ export const ALL_SDKS: SDK[] = [
         image: 'https://res.cloudinary.com/dmukukwp6/image/upload/vercel_373fa70879.svg',
         docsLink: 'https://posthog.com/docs/ai-observability/installation/vercel-ai',
     },
+    {
+        name: 'Eve',
+        key: SDKKey.EVE,
+        tags: [SDKTag.FRAMEWORK],
+        searchTerms: ['Vercel Eve'],
```
carlos-marchal-ph: This is a single two-word phrase where the other entries list discrete terms. The card is already called Eve, so would `['Vercel', 'agent']` match more of what people actually type?

nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.ts
```diff
@@ -284,8 +303,15 @@ function process(event: PluginEvent, next: () => void): void {
     delete props['ai.response.finishReason']
     delete props['gen_ai.response.finish_reasons']
 
+    if (eveSpan) {
+        const eveSessionId = props['eve.session.id']
+        if (props['$ai_session_id'] === undefined && isNonEmptyString(eveSessionId)) {
+            props['$ai_session_id'] = eveSessionId
+        }
+    }
```
carlos-marchal-ph, line 310: The eve.* keys are added to MARKER_KEYS but never to STRIP_KEYS, so eve.session.id ends up duplicated (kept raw and copied here into $ai_session_id) and eve.version / eve.environment ride along as raw props on every event.

posthog/hogql_queries/ai/session_query_runner.py
```diff
@@ -239,7 +246,12 @@ def _build_query(self) -> ast.SelectQuery:
             WHERE event IN (
                 '$ai_span', '$ai_generation', '$ai_embedding', '$ai_metric', '$ai_feedback', '$ai_trace'
             )
-              AND {filter_conditions}
+              AND {trace_filter_conditions}
+              AND trace_id IN (
+                  SELECT trace_id
+                  FROM posthog.ai_events AS ai_events
+                  WHERE {session_filter_conditions}
+              )
```
carlos-marchal-ph, line 254: This changes session membership for every framework, not just Eve. A session now pulls the full trace for any trace with at least one event carrying the session id, rather than only the events that carry it themselves.

TBH sounds like the better solution, just a bit weird to add this in a "add a new provider" PR.

carlos-marchal-ph (approved):

bradlocking:
Thanks folks! Really appreciate the work getting this in

bradlocking:
Hey @Radu-Raicea, a quick follow up on the eve integration. I had trouble getting spans to show up in PostHog, and the fix was different from your docs, so I wanted to share it.

My setup: eve `0.24.6` running through withEve inside next dev on my machine (`@posthog/ai 7.16.4`, `@vercel/otel`, `@opentelemetry/api 1.9.1`).

The problem: I set it up the way the docs say, `registerOTel` with `PostHogTraceExporter` in the eve setup:

```tsx
registerOTel({
  serviceName: agentName,
  traceExporter: new PostHogTraceExporter({ apiKey, host }),
})
```

> Note: `PostHogTraceExporter` expects `apiKey`, but the [docs](https://posthog.com/docs/ai-observability/installation/eve#add-eve-instrumentation) document it as `projectToken`.

Nothing showed up in PostHog. When I looked closer: the exporter was built and setup ran, but the exporter's `export()` was never called, so no spans ever left the app. A test span I made right after `registerOTel` never went out either. And `trace.getTracerProvider()` came back empty (a plain `ProxyTracerProvider` with nothing behind it). So the provider `registerOTel` set up was not the one eve was sending its spans to. My guess is there are two copies of `@opentelemetry/api` in the dev build, and `registerOTel` set up its provider on one copy while eve makes its spans on the other.

What fixed it: setting up the provider by hand, on the same `@opentelemetry/api` the agent imports, instead of using `registerOTel`:

```tsx
import { trace } from '@opentelemetry/api'
import { BasicTracerProvider, BatchSpanProcessor } from '@opentelemetry/sdk-trace-base'

const provider = new BasicTracerProvider({
  spanProcessors: [new BatchSpanProcessor(new PostHogTraceExporter({ apiKey, host }))],
});

trace.setGlobalTracerProvider(provider)
```

After that everything came through, and the send to PostHog came back with no errors.

Two questions:
1. Should `registerOTel` work when eve runs through `withEve` inside a Next dev server, or is setting up the provider by hand the better way for this setup?
2. If `registerOTel` is meant to work here, any idea why it would land on a different copy of `@opentelemetry/api` than the one eve uses?

Happy to share more or put together a repro.

Radu-Raicea:
Thanks, this was really helpful. I looked into both questions.

1. `registerOTel` should work in this setup, and I would keep it as the recommended approach. Eve’s own 0.24.6 documentation uses it from `defineInstrumentation({ setup })`. The manual `BasicTracerProvider` setup is a useful workaround, but it does not configure the async context manager and propagators that `registerOTel` installs. Without those, spans may lose their parent-child relationships and `getActiveSpan()` may not work correctly.

2. Your multiple-copy explanation is plausible, although I cannot confirm it without a reproduction. In local development, `withEve` starts a separate `eve dev` process rather than running Eve inside the Next.js server process. Inside that Eve process, however, Eve vendors its own `@opentelemetry/api`, while `@vercel/otel` and the application may load other instances. These should normally share the process-global provider, but initialization order or a cached no-op provider could leave Eve attached to a different provider instance. The fact that registering through the API imported by the agent fixed exporting is consistent with that.

Eve also has an [open change to load authored instrumentation earlier](https://github.com/vercel/eve/pull/1302), so this appears more likely to be an Eve bundling or initialization issue than a PostHog exporter or ingestion issue.

You are also right about `apiKey`: `@posthog/ai@7.16.4` expects `apiKey`. `projectToken` was added in 7.19.6 and is the option used in 8.x. We have updated our documentation to make that version boundary clear.

bradlocking:
> You are also right about `apiKey`: `@posthog/ai@7.16.4` expects `apiKey`. `projectToken` was added in 7.19.6 and is the option used in 8.x. We have updated our documentation to make that version boundary clear.

Thanks for this, I saw in https://github.com/PostHog/posthog/pull/74329.

Shall I wait for this to land to test with `registerOTel` again?

Radu-Raicea:
@bradlocking yes, it's in the merge queue so should be deployed soon!

PR #74234, opened 2026-07-28
fix(ingestion): treat ai meta-events as ai event types

pl:
## Problem

The AI ingestion lane DLQs the internal AI meta-events with `event_not_in_allowlist`: `$ai_tag` (LLM taggers), `$ai_generation_summary` and `$ai_trace_summary` (trace summarization), and `$ai_evaluation_report` (eval reports).
They are emitted server-side via `capture_internal`, share the `$ai_` namespace, and carry AI telemetry (trace IDs, model, token counts), but `AI_EVENT_TYPES` only listed the seven core types, so the lane's allow step rejects them.

## Changes

Adds the four meta-events to `AI_EVENT_TYPES`. Effects across the code paths that gate on the set:

- AI lane allow step: meta-events pass instead of DLQing (the fix).
- `processAiEvent` enrichment: they get trace-property normalization only; cost and model-param extraction gates on the exact core names and is unaffected.
- `splitAiEventsStep`: they now double-write to `ai_events` alongside `events`.
- Analytics joined pipeline: they route down the `ai` branch, which does full person processing; the branch's "AI events are never merge events" assumption holds for them.

The Python query-routing list (`AI_EVENT_NAMES` in `ai_table_resolver.py`) is deliberately **not** expanded: `ai_events` has no meta-event rows from before this change, and routing their queries to the dedicated table would make the resolver's retention probe misread the missing history as expired data. Their queries stay on the shared `events` table. Both lists now carry comments explaining the divergence.

Extended the parameterized `process-ai-event-step` case with the four meta-events — catches the regression where they are removed from the set and start DLQing again. The existing non-AI DLQ cases (`$pageview`, `$$heatmap`, ...) still pass unchanged.
Ran the AI pipeline, analytics pipeline, and common step suites; the only failures are pre-existing environment-dependent tests (blob-store integration needs object storage, personless step needs a dedicated test database) that fail identically without this change.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code. Skills invoked: /writing-tests, /writing-code-comments.
Decisions: an earlier approach made the AI lane's allow step prefix-based (`$ai_*`) with non-canonical events passing through enrichment untouched (#74232); review showed `processAiEventStep` still DLQs anything outside `AI_EVENT_TYPES`, and expanding the canonical set was chosen as the cleaner fix. The Python mirror was left unexpanded on purpose to keep meta-event queries on the events table until the dedicated table has history.


nodejs/src/ingestion/common/subpipelines/ai-event-types.ts
```diff
@@ -1,11 +1,22 @@
...
     '$ai_embedding',
     '$ai_evaluation',
     '$ai_span',
     '$ai_trace',
     '$ai_metric',
     '$ai_feedback',
+    // Internal meta-events emitted about AI telemetry (taggers, summarization, eval
+    // reports). Cost and model-param enrichment gates on the exact core names above,
+    // so these only get trace-property normalization and the ai_events double-write.
+    '$ai_tag',
```
Radu-Raicea, line 18: This is not a meta event though, it's similar to `$ai_evaluation` events, and I want to get rid of it, but we have a few customers using it.

I want to merge it into `$ai_evaluation` events, once those support what `$ai_tag` need (non-boolean outputs - categorical)

Radu-Raicea (approved):

PR #74238, opened 2026-07-28
fix(alerts): sign the slack snooze target to prevent forged alert ids

MattBro:
## Problem

A project member can make a Slack snooze button act on an alert that the message does not name.

[feat(alerts): add snooze button to slack alert notifications](https://github.com/PostHog/posthog/pull/72698) added the snooze dropdown. That dropdown holds the target alert id in hog-function block config, and any project member can edit that config. Nothing tied the id to the notification. An author could therefore put a different alert's id behind option text that looks correct, such as "For 1 day". A user who has access to that other alert then snoozes it without knowing.

The existing access checks confirm that the user who clicks has access to the alert. They do not confirm that the id belongs to the alert the message describes.

This defect came from a [review comment](https://github.com/PostHog/posthog/pull/72698#discussion_r3660795412) on that PR.

## Changes

The snooze action now carries a server-side signature, and the author cannot produce that signature.

- `sign_insight_alert_id` and `verify_insight_alert_id` are new in the alerts facade (`products/alerts/backend/facade/api.py`). Each one uses an HMAC-SHA256 over the alert id, keyed on `settings.SECRET_KEY`.
- `trigger_alert_hog_functions` (`posthog/tasks/alerts/utils.py`) now sends `alert_id_sig` next to `alert_id` in the `$insight_alert_firing` event. The server computes the signature when it sends the notification, so a hog-function author cannot produce a valid signature for an arbitrary alert id.
- The snooze dropdown option value is now `{alert_id}|{alert_id_sig}|{token}`, in `sub-templates.ts` and in the matching `cdp_functions.yaml`. The handler in `products/slack_app/backend/api.py` verifies the signature while it parses the value, and rejects any value whose signature does not match the id.
- The date and time modal path does not change. The server writes its `private_metadata` only after it verifies the interaction that selected the option.

One weakness stays, and this PR documents it instead of fixing it. A per-alert signature does not change between messages. It stops an author from naming an arbitrary alert. It does not stop a replay. An author can copy an id and signature pair out of a legitimate message for an alert they already saw. The author can then send that pair again. A per-notification nonce with server-side storage would stop that. The finding is low severity. It requires a project member. The action is reversible, and the audit log records it. The nonce therefore costs more than the risk.

I ran `hogli test` locally over `products/slack_app/backend/tests/test_posthog_code_interactivity.py` and the new `products/alerts/backend/test/test_snooze_signature.py`. 67 tests passed.

- The new `test_snooze_signature.py` asserts four properties. A signature verifies the id it was made for. A signature made for one id does not verify a different id, which is the property the fix depends on. A changed value fails verification. The firing event carries an `alert_id_sig` that verifies. These tests fail if the emit drops the signature, or if the verify accepts more than it should.
- I updated the interactivity tests to use signed values. I added `test_snooze_rejects_bad_signature`, which runs over a forged signature and over an unsigned two-part value. Neither value snoozes the alert, even when the alert is real and the user has access. This test fails if the verification step is removed.
- `tach check` and `mypy` are clean on the changed files.

I did not click the button in Slack on this branch. The parent PR was verified by hand end to end. This PR changes the value format and adds a verification step, and the tests above cover both.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Built with Claude Code ([session](https://claude.ai/code/session_01Q2wDEGbVefb6oJJSkjAsnq)) as a follow-up to [feat(alerts): add snooze button to slack alert notifications](https://github.com/PostHog/posthog/pull/72698), in answer to the review comment linked above.

Skills invoked: `/writing-tests`, `/pr-ready`, `/unambiguous-agent-text`.

I chose a static HMAC binding instead of a signed per-notification nonce. The nonce is stronger, because it also stops a replay, but it needs server-side state for each message. The finding is low severity, it requires a project member, the action is reversible, and the audit log records it. The HMAC binding matches that level of risk, and the section above states the remaining weakness for a reviewer to judge.


mayteio:
@MattBro did you still want to get this in? If you get it ready for review & reviewhog it happy to stamp 😄

PR #74239, opened 2026-07-28
feat(ingestion): track parse timing and message sizes in tophog

pl:
## Problem

TopHog gives per-tenant top-N visibility into the ingestion pipelines, but nothing before the transform step is tracked. Events that fail at parse (malformed payloads that go to the DLQ) are invisible in TopHog, and there is no per-sender view of raw message sizes at all. When a single token sends oversized or malformed events, identifying it today means sampling the DLQ topic by hand.

## Changes

- The common ingestion pipeline builder now **requires a `topHog: TopHogRegistry`** in its config, and `parseMessage()` always records parse metrics through it. Every common-builder pipeline (AI, analytics joined, error tracking, heatmaps, client warnings) gets these for free:
  - `messages_by_token` (count)
  - `parse_time_ms_by_token` (timer)
  - `message_size_by_token` (sum of raw Kafka message bytes)
  - `max_message_size_by_token` (max)
  - `message_size_by_token_per_partition` (sum)
- Metrics are keyed by token (the team is not resolved yet at parse time) and recorded from the raw message on the input side, so events that DLQ at parse still show up. This is the earliest per-sender visibility the pipeline has. Naming follows the session replay convention.
- The builder surfaces take metric factories through a `topHog` key in their options, so pipelines never touch `createTopHogWrapper`: `.pipe(step, { retry, topHog: [...] })`, `.parseMessage({ topHog: [...] })` (appended to the defaults), `.resolveTeam({ topHog: [...] })` (replaces the old `resolveTeam({ wrap })`). The AI pipeline's transform/emit metrics moved into pipe options accordingly.
- `CommonIngestionConsumerScope` now installs the topHog registry itself via a new `TopHogComponent` (same lifecycle-component pattern as the other scope services), in its own scope level directly above the consumer: the started registry is in the container when the pipeline factory runs, and the flush loop starts before and stops after the consumer. Consumers neither build nor pass it — `container.topHog` is just there. TopHog stays deliberately per-consumer rather than in the shared services scope: an instance bakes in the pipeline/lane labels its rows are stamped with and drains through that consumer's outputs, and one server process runs several consumers off the same shared scope.
- **Heatmaps and client warnings get TopHog bootstrapped**: the `tophog` output registered in their output registries and the registry threaded through their pipeline configs. No deployment config changes needed: the output resolves from code defaults (`clickhouse_tophog` via the `INGESTION_DOWNSTREAM` producer, which their deployments already configure through the shared ingestion values), and both apps already set `INGESTION_PIPELINE`/`INGESTION_LANE`.

> [!NOTE]
> Error tracking's pre-existing `resolved_teams` counter is removed: it recorded once per event whose token resolved to a team, which the name didn't convey, and the new `messages_by_token` parse default plus error tracking's existing `emitted_events` (per team, at emit) cover inbound and outbound counts with clearer semantics. Its two pipeline tests were repointed at `messages_by_token`.

- Shared test fakes in `tests/helpers/tophog.ts` (`createNoopTopHog`, `createRecordingTopHog` with per-metric capture). Builder tests assert `resolveTeam` metrics record ok/miss outcomes and that parse timing/sizes record for every parsed message; `parse-kafka-message.test.ts` asserts exact per-metric keys and values on both the DLQ path (the visibility property this PR exists for) and the success path, including the `unknown`-token fallback and that no unexpected metrics record.
- Ran the affected jest suites (common builder, parse step, AI, analytics, error tracking, heatmaps, client warnings — pipelines, consumers, and framework extensions): 1224 passed; the only failure is the S3 blob-store integration test that also fails on clean master without local object storage.
- `pnpm typescript:check` shows only pre-existing errors from unbuilt local packages (identical on master).
- No manual testing against a running stack.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, internal observability metrics only.

## 🤖 Agent context

Authored with Claude Code in an ops session about AI-pipeline runbooks, where we noticed events that DLQ at parse are invisible in TopHog. Iterated per review: first a generic `wrap` option on `parseMessage()`; then TopHog required in the builder config with metric-factory lists; then the lists moved under a `topHog` options key on `pipe`/`parseMessage`/`resolveTeam` so pipelines don't build wrappers; then the per-consumer wiring settled as `TopHogComponent` (colocated with the `TopHog` class, like other lifecycle components) installed by `CommonIngestionConsumerScope` in a scope level above the consumer — a flat single-scope variant was tried and reverted since handing the registry out pre-start bends the scopes lifecycle model and loses start/stop ordering. Putting TopHog in the shared services scope was evaluated and rejected: instances bake in per-pipeline identity and outputs. Metrics are keyed by token rather than team_id since the parse step runs pre-team-resolution.


eli-r-ph (approved):
Nice change, LGTM!

PR #74247, opened 2026-07-28
fix(tasks): route CI follow-ups to a GitHub comment instead of Slack

Twixes:
## Problem

Follow-up messages on PRs opened via the Slack app are noisy. After the "Pull request opened :rocket:" card, the agent keeps posting CI/review follow-ups into the thread, which reads as a wall of verbose updates the requester didn't ask for. The card itself is useful; the trailing chatter is what people want gone.

## Changes

Once a run has already opened a PR, `relay_task_run_message` now posts later agent messages as a comment on that PR instead of relaying them to the Slack thread. Slack keeps just the "PR opened" card; the CI/review follow-ups land on the PR where they belong.

It falls back to the normal Slack relay when the task has no GitHub integration or the GitHub API rejects the comment, so nothing is silently dropped. Direct replies to a user's own follow-up (a different code path) still go to Slack, since that's an active conversation.

Added two endpoint tests in `products/tasks/backend/tests/test_api.py`:
- a run with an open PR relays as a GitHub comment and does not enqueue a Slack relay
- when the GitHub comment is rejected, it falls back to the Slack relay

These lock in the new routing decision and the fallback. I (Claude, via the PostHog Slack app) could not run them in this environment because it has no database or Docker available, so I verified the changed files compile and pass ruff. They need a run in CI or a local stack.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

N/A

## 🤖 Agent context

Directed from a Slack thread: the ask was to stop CI follow-ups from cluttering Slack threads and instead post a GitHub comment when changes to the PR were needed, kept as simple as possible.

Authored by Claude (PostHog Slack app). I traced the agent→Slack relay path to the single chokepoint (`relay_task_run_message` in the tasks facade) and added the PR-comment routing there, reusing the existing `GitHubIntegration.comment_on_pull_request_from_url` helper. I checked that the other relay callers in `forward_pending_message.py` are direct replies to user follow-ups and left them on Slack on purpose. Invoked the `/writing-tests` skill before adding tests. Could not execute the DB-backed tests here (no DB/Docker); compile + ruff pass.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1785248561288099?thread_ts=1785248561.288099&cid=C09SK2PAGKF)*


tatoalo (approved):
yes

PR #74250, opened 2026-07-28
fix(cdp): match microsoft ads conversions on email or phone without a click id

luke-belton:
## Problem

A customer testing the new Microsoft Ads destination noticed the template bails out of any event that has no `msclkid`, and asked whether that's right given the API can also match on email.

They're correct. In [Microsoft's UET Conversions API schema](https://learn.microsoft.com/en-us/advertising/guides/uet-conversion-api-integration?view=bingads-13), `msclkid` is not a required field, and the docs explicitly recommend sending hashed email (`em`) and phone (`ph`) for enhanced conversions so measurement survives third-party cookie loss. Hard-gating on the click ID drops every conversion from a user who arrived without one, or whose click ID we never captured, even when we have a perfectly good identifier to send.

## Changes

The click ID is now optional. The template builds `userData` from whatever it has and only skips the event when none of `msclkid`, `email`, or `phone` are available, which is the point where Microsoft genuinely can't attribute anything. This matches how the Snapchat destination already handles its identifier set.

```diff
-if (empty(inputs.microsoftClickId)) {
-    print('Empty `microsoftClickId`. Skipping...')
-    return
-}
-
-let userData := {
-    'msclkid': inputs.microsoftClickId
+let userData := {}
+if (not empty(inputs.microsoftClickId)) {
+    userData.msclkid := inputs.microsoftClickId
 }
```

The `microsoftClickId` input drops `required: true` and its description now says it gives the best attribution rather than that it's mandatory.

Review also turned up that the `phone` input defaulted to an empty string while `email` defaulted to `{person.properties.email}`, so the phone match path was unreachable without hand-wiring the field. It now defaults to `{person.properties.phone}`, matching the snapchat and tiktok destinations.

> [!NOTE]
> Behavior change for anyone already running this destination: events that were previously skipped will now send if the person has an email or phone. That's the intent, but it means conversion volume in Microsoft Advertising will go up after this ships.

Automated only, I didn't exercise this against a live UET tag.

Claude updated `microsoft.template.test.ts` and ran it (5 passed). Two cases changed:

- A parameterized case replaces the old `skips when microsoftClickId is missing`, asserting we still fire when only the click id, only the email, or only the phone is present. That's the direct regression guard for this fix, and each identifier gets its own branch rather than email standing in for all three.
- `skips when no identifier is available` covers the remaining skip path, so a future refactor can't quietly drop the guard entirely and start posting unattributable events.

I also checked our hashing against the worked examples in Microsoft's CAPI docs: both the email and phone hashes match, including keeping the leading `+` on E.164 numbers, so that needed no change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed, the destination's user-facing copy is the input description which is updated here.

## 🤖 Agent context

Claude Code (Opus) did the change. I pointed it at the customer's question and it verified the claim against Microsoft's published CAPI schema before touching anything, rather than taking the report at face value. Worth confirming that reading in review.

It considered widening the payload further while in here, `externalId`, `clientIpAddress`, and `clientUserAgent` are all supported and would help match rates, but left them out to keep this scoped to the reported bug. Skills invoked: `/writing-user-facing-copy` for the input description, `/writing-tests` for the test changes.

---

### DIFF
--- a/nodejs/src/cdp/templates/_destinations/microsoft_ads/microsoft.template.test.ts
+++ b/nodejs/src/cdp/templates/_destinations/microsoft_ads/microsoft.template.test.ts
@@ -82,16 +82,27 @@ describe('microsoft template', () => {
             `"Error from capi.uet.microsoft.com (status 401): {'error': {'code': 'Unauthorized', 'message': 'You are not authorized to access this resource.'}}"`
         )
     })
-    it('skips when microsoftClickId is missing', async () => {
+    it('sends the conversion without a click id when email is present', async () => {
         const response = await tester.invokeMapping(
             'Conversion',
             { tagId: '12345678', apiToken: 'api-token', eventName: 'purchase' },
             createAdDestinationPayload({ person: { properties: { msclkid: null } } })
         )
+        expect(response.error).toBeUndefined()
+        expect(JSON.parse(response.invocation.queueParameters!.body as string).data[0].userData).toEqual({
+            em: '3d4eee8538a4bbbe2ef7912f90ee494c1280f74dd7fd81232e58deb9cb9997e3',
+        })
+    })
+    it('skips when no identifier is available', async () => {
+        const response = await tester.invokeMapping(
+            'Conversion',
+            { tagId: '12345678', apiToken: 'api-token', eventName: 'purchase' },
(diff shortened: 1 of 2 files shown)


mayteio (approved):
LGTM, please check the bot comments before merging 😄

PR #74254, opened 2026-07-28
chore: move frontend and django admin guidance to path-scoped rules

gantoine:
## Problem

`AGENTS.md` (and its `CLAUDE.md` symlink) loads into every agent session, in full, no matter what the session is about. It is around 26k characters, so roughly 6.5k tokens of context before anyone types anything.

Some of that guidance is not universal. Four bullets only matter when you are actually touching frontend files or a Django `admin.py`. Every backend-only or infra-only session pays for them anyway.

## Changes

Moved four bullets out of `AGENTS.md` into `.claude/rules/`, which the repo already uses for path-scoped guidance (13 rule files today, all with `paths:` frontmatter).

| Moved | New file | Loads on |
| --- | --- | --- |
| Frontend AGENTS.md pointer, quill design system, quill vs LemonUI | `.claude/rules/frontend-quill.md` | `frontend/src/**`, `products/*/frontend/**`, `packages/quill/**` |
| Django admin `ForeignKey` widget config | `.claude/rules/django-admin-widgets.md` | `**/admin.py`, `**/admin/**` |

`AGENTS.md` goes from 26,122 to 23,973 characters, so about 537 tokens off every session. Content is unchanged, only where it lives.

> [!NOTE]
> Nothing safety-critical moved. Every "never do X" prohibition stays always-loaded, along with the double-submit guard, the code style bullets, and the mandatory skill invocation list. A lazily-loaded prohibition is a prohibition you cannot rely on.

No automated tests here, this is agent guidance content. I verified the mechanical parts: the new rule files match the frontmatter shape of the existing `.claude/rules/*.md` files, relative links resolve from `.claude/rules/` (they needed `../../` prefixes), and `hogli ci:preflight --strict` passes on the push, including the AGENTS.md symlink check and markdown formatting.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not applicable, no user-facing behavior.

## 🤖 Agent context

This came out of a broader Claude Code setup health check I ran on my machine. Most of the findings were local to my own config (unused skills, stale memory files, an overly broad permission rule) and are not in this PR. This is the one piece that touches the repo.

The check first looked for content in `AGENTS.md` a session could just derive by reading the codebase. There was almost none, the file is dense with real gotchas and failure contracts, so I did not propose a trim. The second pass looked for content that is real but not universal, which is what this PR acts on.

Claude picked a larger migration set initially, including the MCP tool bullets and the Temporal payload size limit. I dropped those. The MCP bullets are pointers people need to know exist before they go looking, and the Temporal 2 MiB limit is the kind of thing you want in context before you write the activity, not after you have already opened the file.

Tools: Claude Code (Opus 5). No repo skills were invoked for the content itself.


rnegron (commented):
hmm I dunno... I'd rather keep the agent-agnostic logic (ie cleanly supporting codex as well) rather than moving these short paragraphs into `.claude` rules? im not sure I see the friction?

gantoine:
> im not sure I see the friction?

this was a recommendation from `claude /doctor`, if it doesn't make sense we can close it.
_i'm no slave to the machine yo_

PR #74259, opened 2026-07-28
fix(batch-exports): make missing Postgres connection inputs non-retryable

rossgray:
## Problem

A `postgres-export` batch export whose backing integration is missing credentials (e.g. `user` is `null`) hit `raise ValueError("Missing required inputs")` in `PostgresInsertInputs.credentials()`. A bare `ValueError` isn't in the Postgres destination's non-retryable list, so the activity kept retrying until it breached the 1-hour SLA and paged on-call. Retrying can never recover a misconfigured/absent integration, so it should fail fast.

## Changes

- Add a dedicated `PostgreSQLMissingRequiredInputsError` and raise it (instead of `ValueError`) from `credentials()` and `authority()` when required connection inputs are missing.
- Register it in `NON_RETRYABLE_ERROR_TYPES` so the export fails fast instead of retrying to an SLA breach.

Using a distinct exception type rather than reusing `ValueError` avoids masking unrelated `ValueError`s that might arise elsewhere in the activity.

Added a parameterized unit test (`test_missing_connection_inputs_raise_non_retryable_error`) covering missing `user`/`password` (via `credentials()`) and missing `host`/`port` (via `authority()`). It asserts the new exception is raised and that its class name is registered in `NON_RETRYABLE_ERROR_TYPES` — this guards against a future rename/refactor silently making these errors retryable again (the exact SLA-breach regression this PR fixes, since the retry classifier matches on class name string).

The local sandbox has no Postgres/ClickHouse, so I couldn't run the DB-backed suite here; the new test is pure logic. `ruff check`/`format` and `py_compile` pass.

## 🤖 Agent context

Authored by the PostHog Slack app from a Slack thread where @rossgray  diagnosed a batch-export SLA-breach alert down to `user` being `null` and the resulting bare `ValueError`, and asked for a new non-retryable exception type. Invoked the `/writing-tests` skill to gate the added test. The choice of a distinct exception type (over reusing `ValueError`) follows Ross's suggestion to avoid masking other errors.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C085MQ3P96X/p1785247496332049?thread_ts=1785247496.332049&cid=C085MQ3P96X)*

### DIFF
--- a/products/batch_exports/backend/temporal/destinations/postgres_batch_export.py
+++ b/products/batch_exports/backend/temporal/destinations/postgres_batch_export.py
@@ -160,10 +160,13 @@ class PostgreSQLTransactionError(Exception):
 
 
 class PostgreSQLMissingRequiredInputsError(Exception):
-    """Raised when required connection inputs (credentials or host/port) are missing.
+    """Raised when the export is missing required connection inputs (credentials or host/port).
 
     This usually means the backing integration is misconfigured or absent, so retrying won't recover it.
     """
+
+    def __init__(self, err_msg: str):
+        super().__init__(f"The export is missing required connection inputs: {err_msg}")
 
 
 class _PostgreSQLClientInputsProtocol(typing.Protocol):
@@ -193,7 +196,7 @@ class PostgresInsertInputs(BatchExportInsertInputs):
         password = self.password
 
         if user is None or password is None:
-            raise PostgreSQLMissingRequiredInputsError("Missing required credentials (user and/or password)")
+            raise ValueError("Missing required inputs")
 
         return Credentials(user, password)
 
@@ -202,7 +205,7 @@ class PostgresInsertInputs(BatchExportInsertInputs):
         port = self.port
 
         if host is None or port is None:
-            raise PostgreSQLMissingRequiredInputsError("Missing required connection details (host and/or port)")
+            raise ValueError("Missing required inputs")
 
         return Authority(host, port)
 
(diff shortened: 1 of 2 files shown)


products/batch_exports/backend/temporal/destinations/postgres_batch_export.py
```diff
@@ -183,7 +193,7 @@ def credentials(self) -> Credentials:
         password = self.password
 
         if user is None or password is None:
-            raise ValueError("Missing required inputs")
+            raise PostgreSQLMissingRequiredInputsError("Missing required credentials (user and/or password)")
```
tomasfarias: Just a thought: I think I would rather we raise this error in the batch export, and keep the generic `ValueError` here. The inputs themselves have no context of the batch export function, so it makes sense that they react more generic. Whereas the batch export function is the one that controls which error is ultimately raised.

products/batch_exports/backend/temporal/destinations/postgres_batch_export.py
```diff
@@ -933,7 +946,12 @@ async def insert_into_postgres_activity_from_stage(inputs: PostgresInsertInputs)
         )[:63]
 
         client_inputs = await _get_postgresql_integration(inputs) or inputs
-        pg_client = PostgreSQLClient.from_inputs(client_inputs, database=inputs.database)
+        try:
+            pg_client = PostgreSQLClient.from_inputs(client_inputs, database=inputs.database)
+        except ValueError as err:
+            # Missing credentials/host usually means a misconfigured or absent integration, which retrying
+            # can't recover. Surface it as a non-retryable error so the export fails fast instead of breaching SLA.
+            raise PostgreSQLMissingRequiredInputsError(str(err)) from err
```
tomasfarias, line 954: Thanks, I prefer this. Maybe in the future we need better abstractions around retryable/non-retryable, but this is great.

tomasfarias (approved):

products/batch_exports/backend/temporal/destinations/postgres_batch_export.py
rossgray: yep!

PR #74264, opened 2026-07-28
feat(aio): Add config to prompt versions

jurajmajerik:
## Problem

Prompt management stores only the prompt text. Model parameters (model, temperature, tools, response schemas) stay in application code, so tuning them still requires a deploy. This is also the main blocker for teams migrating from competitors, where a prompt carries a `config` object next to the text.

## Changes

A prompt version now carries an optional `config`: a schemaless JSON object stored on the version row next to the prompt content, versioned with it, and returned as-is by the fetch endpoint. This PR is the backend. SDK exposure (`prompt.config` in posthog-python and @posthog/ai) and the UI come in follow-up PRs.

- Nullable `config` JSONField on `LLMPrompt` plus migration (plain `ADD COLUMN`)
- Accepted on create and publish. Must be a JSON object or null, same 1 MB cap as the prompt payload
- Config-only publish: `config` plus `base_version` with no `prompt` or `edits` publishes a new version with the content unchanged, so tuning a parameter doesn't require resending the whole prompt
- Returned by the public fetch endpoint with `content=full`, omitted for `preview` and `none`, included in the HyperCache payloads
- MCP create/update prompt tools accept `config`
- `prompt_has_config` on `$llm_prompt_fetched`; the activity log records that config changed, never its contents

Design choices, briefly:

- **Schemaless object, not structured fields.** Whatever shape the app needs (sampling params, tool definitions, response schemas) fits without us maintaining a provider schema, and competitor configs migrate as-is. The top level must be an object so we can recognize well-known keys like `model` later without a breaking change.
- **Carry-forward on publish.** A publish that omits the `config` key keeps the previous version's config; explicit `config: null` clears it. Without this, every text-only publish (including the `edits` find/replace path that agents use) would silently drop the config.
- **No cache migration.** Cache entries written before the field existed just read as `config: null`, which is correct: config can only become non-null through a post-deploy write, and those invalidate through the existing signals.

New tests in `test_llm_prompt.py`: a `TestLLMPromptConfigAPI` class covering publish semantics (config carried forward through both full-prompt and edits publishes, explicit null clears it, config-only publish keeps content unchanged), content modes omitting config, duplicate copying it, and a pre-deploy cache entry without the key reading as null instead of erroring. Validation (non-object, oversize) is a no-DB `SimpleTestCase` matrix on the publish serializer plus endpoint wiring guards for create and publish.

Full prompt suites (168 tests) green locally, repo-wide mypy clean, `ci:preflight` green. Not tested: SDK consumption, since the SDKs don't read `config` yet (follow-up PRs).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Docs follow together with SDK support in the follow-up PRs, so we don't document a field the SDKs can't read yet.

## 🤖 Agent context

I directed Claude Code through research, plan, and implementation. Skills invoked: /django-migrations, /improving-drf-endpoints, /writing-tests, /implementing-mcp-tools. Two decisions emerged during implementation rather than in the plan: allowing config-only publishes (the original plan required `prompt` or `edits` on every publish), and adding `config` to the MCP tools' `include_params` allowlists, which would otherwise have silently excluded it from agent tooling.


posthog/api/llm_prompt_serializers.py
```diff
@@ -224,6 +236,16 @@ class LLMPromptPublishSerializer(serializers.Serializer):
             "Mutually exclusive with prompt."
         ),
     )
+    config = serializers.JSONField(
```
Radu-Raicea: `config` is object-or-null at runtime, but this `JSONField` generates `unknown` and an untyped MCP property, so clients accept strings and arrays the API rejects. Could we use an object-annotated field across create, publish, and response serializers?

Radu-Raicea (approved):

posthog/api/llm_prompt_serializers.py
jurajmajerik: Done in 3aa811f. Added a LLMPromptConfigField (a JSONField annotated with {"type": "object", "nullable": true}) and used it in all three places: the create serializer, the publish serializer, and the public fetch response. The generated TypeScript is now { [key: string]: unknown } | null, the Zod schema is a nullish record, and the MCP tool schemas show object-or-null too (snapshots updated). Clients now reject strings and arrays before the request is sent, matching what the API enforces.

PR #74277, opened 2026-07-28
chore(managed-warehouse): extract ducklake into its own product

EDsCODE:
## Problem

Managed warehouse and DuckLake infrastructure lived in the core Django app, even though several products consume it. That left the code unowned, exposed ORM models and internal modules across product boundaries, and forced unrelated managed warehouse changes through the full Django test suite.

The initial extraction moved the code into `products/managed_warehouse`, but the product was not safely isolated yet. External callers still received Django models or imported presentation and state modules directly, so enabling selective CI would have hidden real downstream coupling.

### Extract managed warehouse into its own product

- Moved the DuckLake, Duckgres, Temporal, model, migration, and test surfaces into `products/managed_warehouse/backend`.
- Kept physical table names pinned to `posthog_duckgres*`. The state-only migrations change Django ownership without moving data.
- Preserved Temporal workflow and activity identities for in-flight executions.
- Changed Django admin URL names from the `posthog` app label to `managed_warehouse`.

### Seal the product boundary

- Replaced external `DuckgresServer`, `DuckgresSinkSchemaState`, `CPTeam`, `cp_teams`, and `team_state` access with frozen contracts and capability functions.
- Preserved sink compare-and-swap updates, concurrency budgets, failure tracking, timestamps, batching, and backfill semantics behind the facade.
- Moved control-plane reads into the managed warehouse product while preserving the 60-second cache and existing failure postures.
- Moved SQL connection lifecycle, presentation, Django admin registration, admin templates, and their behavior tests into the product.
- Removed `facade/models.py`. External code now uses the managed warehouse facade, presentation, or routes rather than backend internals.
- Added `backend:contract-check` and narrowed Turbo inputs so implementation-only changes run product tests without cascading through unrelated products. Template changes remain inputs to the product test task.
- Updated the remaining core, Dagster, Temporal, data warehouse, endpoints, and warehouse sources callers to use the new capabilities.

### Boundary before

```mermaid
flowchart LR
    A[Core and sibling products] --> B[Managed warehouse ORM models]
    A --> C[Presentation and state internals]
    B --> D[(Managed warehouse state)]
    C --> D
    class A phYellow;
    class B,C phRed;
    class D phGray;
```

### Boundary after

```mermaid
flowchart LR
    A[Core and sibling products] --> B[Contracts and capability facade]
    B --> C[Managed warehouse implementation]
    C --> D[(Managed warehouse state)]
    class A phYellow;
    class B phBlue;
    class C phRed;
    class D phGray;
```

> [!NOTE]
> The product boundary is sealed and isolated CI is enabled. Managed warehouse still has explicit dependencies on the public facades of data warehouse, data modeling, and warehouse sources because it coordinates those products.

I am an agent. I did not manually exercise Duckgres, admin screens, or Temporal workflows.

Local checks run:

- `tach check --dependencies --interfaces`
- `lint-imports`
- `hogli product:lint managed_warehouse`
- `pnpm --dir products/managed_warehouse backend:lint`
- Ruff and formatting checks across the changed Python files
- 248 focused control-plane and Dagster tests
- Django admin registration and all six relocated admin templates loaded successfully

Database-backed tests and OpenAPI generation could not run locally because PostgreSQL was unavailable. Repo-wide mypy was also OOM-killed on the development VM. The latest CI run passes Python code quality, Django tests, Dagster tests, frontend tests, and all affected product test shards.

The Visual Review workflow detected three snapshot differences and still needs human review. This PR does not intentionally change the frontend UI.

Regression coverage added or moved with the product covers:

- frozen contract mapping without returning managed warehouse ORM objects
- sink state transitions and trusted organization concurrency lookup
- control-plane cache, fallback, and failure behavior
- SQL connection creation, reconciliation, password rotation, and deprovision cleanup
- managed warehouse presentation and Django admin behavior
- stored-bucket fallback and earliest-event-date synchronization through the facade

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the Temporal and DuckLake verification documentation to point at the product-owned workflow and configuration paths. There is no intended user-facing product behavior change.

## 🤖 Agent context

Authored with Claude Code and Shelley at Eric's direction. Repo skills invoked during the work include `/isolating-product-facade-contracts`, `/django-migrations`, `/writing-tests`, `/writing-user-facing-copy`, and `/triaging-visual-review-runs`.

The main design choice was to finish the isolation wave in this PR rather than leave the extracted product with ORM exports and an unsound CI skip. Capability facades preserve the existing transaction and control-plane semantics while keeping Django models private. Presentation and admin ownership moved into the product so core no longer needs model access solely for routing or registration.


sakce (approved):

fuziontech (changes requested):
Automated review agent on behalf of @fuziontech.

Blocking findings:

- `posthog/admin/test_duckgres_server_admin.py:65`: `_grant_admin_permissions()` still filters the `posthog` content type. After this move Django creates these permissions under `managed_warehouse`, so the helper grants nothing and permission-gated admin tests/actions raise `PermissionDenied`. Update the app label (and the stale admin test URLs while here).
- `products/managed_warehouse/package.json:5`: the new `backend:lint` script explicitly uses the root `pyproject.toml`, bypassing the exemptions added in `products/ruff.toml`. Running this script reports 53 `PLC0415` failures, so the product's declared lint task is broken. Point it at the product Ruff config or resolve the violations.

fuziontech (approved):
Automated review agent on behalf of @fuziontech.

The extraction and state-only model move look safe to ship. Migration validation, product lint, workflow lint, Python quality, repo checks, and migration checks are green.

Non-blocking follow-ups:
- Update the Duckgres admin test permission helper to grant `managed_warehouse` permissions rather than `posthog` permissions.
- Point the product `backend:lint` script at `products/ruff.toml`; its current root config ignores the product-specific legacy exemptions.

Focused database-backed tests could not run locally because the review VM has no PostgreSQL service, but the corresponding CI suites are running.

rnegron (commented):
@EDsCODE looks like some conflicts must be solved first, regardless here is a claude-assited devex review:

## Production deploy triggers quietly widened

- `container-images-cd.yml:981`: the old pattern was `^posthog/temporal/ducklake|^posthog/ducklake`, and the temporal tests lived at `posthog/temporal/tests/ducklake/`, so they never matched. The new `^products/managed_warehouse/backend` matches `backend/tests/**`. Test-only edits now redeploy the ducklake temporal worker. Add `grep -v '.../backend/tests/'`.
- The same substitution widened `dataModeling`. Master listed `posthog/ducklake/**` but never `posthog/temporal/ducklake/**`; one path now covers both. Deliberate?
- `ci-dagster.yml:126` keeps `posthog/ducklake/**`. Right call for unrebased PRs, but with no comment and no removal date it lives forever.
- `ci-backend.yml:673` still says warehouse_sources is the only product running its own temporal suite.

## `.test_durations`: drop the change, regenerate

It resurrects `test_duckgres_server_team_admin.py::…`, deleted by [drop DuckgresServerTeam](https://github.com/PostHog/posthog/pull/73876), and breaks sort order (master has 0 out-of-order transitions, this has 2). Let the generator own the file:

```bash
gh workflow run ci-backend-update-test-timing.yml --ref master
```

Dispatch on `master` sets `should_commit=true` and opens a PR with real measured durations. Run it after a master `ci-backend` run that covers the product. Staleness fallback handles the gap until then.

## Description doesn't match the migration

The body says the RunPython "repoints" content types. It copies them and retains both labels for rolling-deploy safety, which is the better design, and the docstring says so. Anyone reviewing off the description evaluates a deploy-unsafe migration instead.

## Copied fixture diverged

`tests/temporal/conftest.py`'s `aorganization` dropped `is_ai_data_processing_approved=True` versus `posthog/temporal/tests/conftest.py`. Probably inert here, but unintended.

Gilbert09 (approved):

EDsCODE:
Addressed, thanks for the detailed review:

- Excluded `products/managed_warehouse/backend/tests/**` from the DuckLake worker deploy trigger.
- Preserved the previous `dataModeling` deploy scope by excluding the moved Temporal implementation and Temporal tests.
- Added an August 14, 2026 removal date to the legacy Dagster path.
- Updated the stale backend selector comment to include `managed_warehouse`.
- Dropped the `.test_durations` changes so the generator remains authoritative.
- Corrected the PR description: the migration copies content types, permissions, and grants while retaining both labels for rolling deploys.
- Restored `is_ai_data_processing_approved=True` in the copied organization fixture.

Validation: workflow lint and Ruff pass. The focused Temporal file ran 9 tests successfully; 3 database-backed cases could not set up because Postgres is unavailable in this environment.

jghoman (approved):

PR #74301, opened 2026-07-28
chore(retention): flag data churn in compare recheck [do not merge]

thmsobrmlr:
## TL;DR (eli5)

When we run the same retention insight twice with two different query implementations and the numbers differ, there are two possible reasons: the queries really disagree (a bug), or the data changed while we were looking (people got merged, late events arrived). Until now the tool couldn't tell these apart. Now it checks: if the numbers are exactly the same on a second look, it's a bug. If the numbers moved, it's the data.

Stacked on #64819 (the commands only exist on that branch), same [do not merge] status.

## Problem

The prod sweep flagged two insights as MISMATCH (team 15506, both monthly first-time retention with different start/return events):

```
[3046/10000] MISMATCH Y6DH6tPi (team 15506) — 2 stable cell diff(s), rows legacy=13 dwh=13
[3226/10000] MISMATCH 00a7uYNz (team 15506) — 1 stable cell diff(s), rows legacy=4 dwh=4
```

Local investigation found no parity gap: all four query shapes (legacy two-arm, legacy single-scan, DWH single-scan, DWH union) return identical results on synthetic data for both insight shapes, including under `SAMPLE 0.25` and with person-override (PoE v2) data. The remaining suspect is live data churn: person merges and late ingest move historical monthly buckets between the two sequential variant runs.

The recheck couldn't discriminate. `intersect_stable_mismatch` matched recheck cells by key only, so a cell that differed in both passes with *different values* (the churn signature) still counted as a stable mismatch. A deterministic query bug reproduces *identical* values.

## Changes

- `CellDiff` gains `values_stable`: `True` = second pass returned the exact same legacy/dwh values (deterministic difference), `False` = the cell differed again but values moved (churn), `None` = no recheck ran.
- `compare_retention_correctness` mismatch lines now end with `2/2 value-identical (deterministic)` or `all values moved between passes (churn, not deterministic)`.
- The full report's cell diff table gains a `Values` column (`identical` / `moved`), only when a recheck ran.

Rerunning the sweep (or just `--insight-id 1798742 --insight-id 1818123`) with this change answers churn-vs-bug directly.

I'm an agent (Claude Code). Automated checks only:

- Extended `test_reproduced_mismatch_kept` to assert the kept cell is marked `values_stable=True` (catches the annotation being dropped), and added `test_moved_values_kept_but_marked_unstable` (catches moved-value cells being marked deterministic). No existing test covered value-level recheck semantics.
- Full unit suites for both commands: 76 passed.
- `ruff check` and `ruff format` clean.

Not run end-to-end against prod ClickHouse; the two flagged insights should be re-checked on a prod pod with this change.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change, internal validation tooling.

## 🤖 Agent context

Built with PostHog Code (Claude), directed by the assignee, while investigating the two sweep mismatches above. Skills invoked: /writing-tests.

Decisions:

- Ruled out a deterministic query gap locally first: generated all four builder shapes for the two prod insight configs and ran an empirical parity probe (400 synthetic actors, multi-distinct-id persons, months scattered around the window). Also force-ran the seven tests excluded from the legacy-vs-variant comparison harness; all pass, so the exclusion list predates the parity fixes and is stale (possible separate cleanup on master).
- Kept moved-value cells in the mismatch verdict rather than dropping them: churn plus a real bug can coexist, so the annotation reports rather than filters.
- Originally pushed to #64819 directly; moved to this stacked PR so the base branch stays at the state its sweep ran from.

---

### DIFF
--- a/posthog/management/commands/compare_retention_correctness.py
+++ b/posthog/management/commands/compare_retention_correctness.py
@@ -296,6 +296,14 @@ def _check_one(insight: Insight, url: str, freeze: bool, recheck: bool) -> Row:
                 f"{len(diff.cell_diffs)} {stable}cell diff(s), "
                 f"rows legacy={diff.row_count_legacy} dwh={diff.row_count_dwh}"
             )
+            if rechecked and diff.cell_diffs:
+                # Value-identical cells reproduce exactly (deterministic difference); moved values
+                # mean the data changed under the queries in both passes (merge/late-ingest churn).
+                identical = sum(1 for c in diff.cell_diffs if c.values_stable)
+                if identical:
+                    detail += f", {identical}/{len(diff.cell_diffs)} value-identical (deterministic)"
+                else:
+                    detail += ", all values moved between passes (churn, not deterministic)"
             if recheck and not rechecked:
                 detail += " (recheck errored — stability unverified)"
         return Row(insight.id, insight.short_id, insight.team_id, url, diff.status, detail)
--- a/posthog/management/commands/compare_retention_legacy_vs_dwh.py
+++ b/posthog/management/commands/compare_retention_legacy_vs_dwh.py
@@ -111,6 +111,11 @@ class CellDiff:
     dwh: float
     abs_diff: float
     rel_diff: Optional[float]  # None when legacy == 0
+    # Recheck verdict: True = the second pass returned the exact same legacy/dwh values
+    # (a deterministic difference), False = the cell differed again but with moved values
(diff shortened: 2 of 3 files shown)


thmsobrmlr:
Note: this shows as merged because its commit was fast-forwarded onto the base branch (the #64819 tooling branch) — the change now lives directly in #64819. Nothing landed on master. Comment from PostHog Code on behalf of @thmsobrmlr.

PR #74317, opened 2026-07-28
feat(error-tracking): sentry migration import pipeline

danielcarletti:
## Problem

Teams on Sentry who want to switch to PostHog error tracking lose their issue history: there is no way to bring existing Sentry issues, events, statuses, and stack traces into PostHog. We already sync Sentry data into the warehouse (issues, issue_events tables), but nothing bridges it into the error tracking product.

Sentry is also just the first of these migrations - the same shape (warehouse source as staging, transform, ingest, reconcile statuses) applies to Rollbar, Bugsnag, and friends - so the durable surfaces here are deliberately source-generic.

## Changes

Backend for a "Migrate from X" flow that turns synced warehouse data into real error tracking issues, with Sentry as the first source:

- `ErrorTrackingMigration` model (source_type + config JSON, fail-closed team scoping, `db_constraint=False` FKs so the migration takes no locks on hot tables) tracking a migration end to end
- a source-agnostic `error-tracking-migration` Temporal workflow: wait for the source's warehouse schemas to finish syncing, page + transform + emit through `capture_batch_internal(historical_migration=True)`, wait for ingestion to settle, then reconcile issue statuses through `issue_mutations.bulk_update_issues`
- everything source-specific lives behind a `MigrationSourceAdapter` protocol (`temporal/source_migration/base.py`): required schemas, page/count/status queries, row-to-`$exception` transform, status mapping, fingerprint scheme. `sentry.py` is the only adapter; adding a source is one adapter module plus a `SourceType` value - no per-source branches in shared code
- the Sentry adapter: deterministic `sentry:{org}:{issue_id}` fingerprints (one PostHog issue per Sentry issue), already-symbolicated frames become `platform: "custom"` frames with `resolved: true` so cymbal skips re-symbolication, `$issue_name`/`$issue_description` seed issues from Sentry's title/culprit, deterministic uuid5 event ids make re-runs idempotent via ClickHouse dedupe, oldest-first emission so `first_seen` lands correctly, `resolved`/`ignored` map to resolved/suppressed
- generic `error_tracking/migrations` API: create (with `source_type`)/list/retrieve/cancel/attach_code_migration, `@validated_request` throughout, registered on `routers.projects`; regenerated OpenAPI types

Quota-dropped events are counted and surfaced, not fatal. Historical timestamps fall outside billing windows (usage reporting counts `$exception` by event timestamp).

Automated only, all run locally against current master:

- 18 Sentry adapter unit tests: field mapping against a realistic Sentry `full=true` payload, exception chain reversal, frame context splitting, frame cap, distinct id fallback chain, status mapping, deterministic ids, anchor event conditions
- 8 API tests: create starts the workflow, adapter config validation (missing org_slug rejected), source-type mismatch rejected, single-active-migration guard, cancel transitions and terminal-state guard, attach_code_migration, team scoping on list/retrieve
- `sqlmigrate` inspected: the migration is a plain `CREATE TABLE` + indexes, no hot-table locks
- repo-wide mypy fully green; `tach check` clean; `hogli ci:preflight --strict` clean

Not done: an end-to-end run against a live Sentry org. The dlt-normalized warehouse column names (`date_created`, `entries`) are tolerated in both camel and snake case by the transform, but the first live import should confirm them.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Docs will land with the user-facing scene PR (stacked on this one).

## 🤖 Agent context

Authored by Claude (Claude Code) with me directing. Skills invoked: /django-migrations, /improving-drf-endpoints, /writing-tests. Design choices worth knowing at review time: the transformation is pure Python in a Temporal activity rather than a materialized view (testability; building cymbal's nested `$exception_list` in HogQL is impractical); emission goes through `capture_batch_internal` rather than direct Kafka so quota, routing, and dedupe stay in one blessed path; frames use cymbal's `CustomFrame` escape hatch because Sentry API frames are already symbolicated. The PR originally shipped Sentry-named surfaces end to end; after review discussion we generalized the durable ones (model, API route, workflow/adapter seam) before merge since they are expensive to rename later, and kept the scene Sentry-branded since each source will want its own UI anyway. The `source_type` enum name collision (`SourceTypeBb2Enum`) is resolved via ENUM_NAME_OVERRIDES per the drf-spectacular workflow.


danielcarletti:
Pushed a follow-up commit generalizing the durable surfaces so this serves as the template for future migrations (Rollbar, Bugsnag, ...): model is now `ErrorTrackingMigration` with `source_type` + config JSON, the API moved to `error_tracking/migrations`, and the Temporal package is a source-agnostic workflow with a per-source `MigrationSourceAdapter` protocol - Sentry is the only adapter. Adding a source is one adapter module plus a SourceType value. PR body updated to match. All suites re-run green; repo-wide mypy fully green.

PR #74319, opened 2026-07-28
feat(data-warehouse): sync full sentry event bodies in fanout children

danielcarletti:
## Problem

The Sentry warehouse source's `issue_events` and `project_events` schemas call Sentry's event-list endpoints without `full=true`, so synced rows carry only the simplified event shape - no `entries`, meaning no stack traces. The upcoming Sentry -> error tracking migration needs full event bodies, and there was no way for a fanout child endpoint to pass extra query params at all.

## Changes

- `DependentEndpointConfig` gains an opt-in `child_params: dict[str, Any]` merged into fanout child request params (the resolve param always wins on collision). Sources that don't set it are byte-identical, and the existing `child_params_extra` call-site override keeps final say.
- Sentry's `issue_events` and `project_events` endpoints set `child_params={"full": "true"}`, so synced rows include the complete event body (stacktrace `entries`).

Automated, run locally against current master:

- new fanout test proving the merge lands, the resolve param can't be clobbered, and parent params are untouched
- existing backwards-compat test strengthened to exact-dict equality on parent and child params - the regression guard that configs without `child_params` produce identical requests to before
- parameterized Sentry test asserting both event endpoints carry `full=true` - catches anyone dropping the setting
- full `common/rest_source` + Sentry suites: 73 passed

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change beyond richer synced columns; source docs render tables from code.

## 🤖 Agent context

Authored by Claude (Claude Code) with me directing, as part of the Sentry migration series. Skills invoked: /implementing-warehouse-sources, /writing-tests. The `params` key rejection in `child_endpoint_extra` was intentionally left in place; `child_params` is the sanctioned path.


products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/fanout.py
```diff
@@ -111,6 +112,8 @@ def build_dependent_resource(
     }
     if page_size_param is not None:
         child_params[page_size_param] = child_config.page_size
+    # The resolve param binds child requests to their parent row — config must not clobber it.
+    child_params.update({k: v for k, v in fanout.child_params.items() if k != fanout.resolve_param})
```
estefaniarabadan: collision is dropped silently, maybe is worth raising a ValueError

products/warehouse_sources/backend/temporal/data_imports/sources/sentry/settings.py
```diff
@@ -128,6 +128,8 @@ class SentryEndpointConfig:
             resolve_field="slug",
             include_from_parent=["id", "slug"],
             parent_field_renames={"id": "project_id", "slug": "project_slug"},
+            # full=true makes Sentry return complete event bodies (incl. stacktrace entries).
+            child_params={"full": "true"},
```
estefaniarabadan, line 195: this affects also existing tables, right? could this cause payload growth on the next syncs?

estefaniarabadan (approved):

products/warehouse_sources/backend/temporal/data_imports/sources/common/rest_source/fanout.py
danielcarletti: Good call — done in d472a05. `build_dependent_resource` now raises a `ValueError` when `child_params` contains the fanout `resolve_param`, instead of silently dropping it, so the misconfiguration surfaces at build time. Added a test (`test_build_dependent_resource_rejects_child_params_clobbering_resolve`) covering it.

products/warehouse_sources/backend/temporal/data_imports/sources/sentry/settings.py
danielcarletti: Yes, it applies to the existing `project_events` and `issue_events` tables, but the growth is bounded to events synced *after* this ships, not a retroactive re-fetch. Both endpoints sync incrementally on `dateCreated` (merge/upsert on the event primary key), so already-synced rows are never re-requested and keep their current shape — only newly ingested events carry the full body (stacktrace entries etc.). The delta table format absorbs the extra columns via schema evolution, so no manual migration is needed. Larger per-row payloads on new events is exactly the intent of this change, so this is expected rather than a regression.

PR #74329, opened 2026-07-28
fix(aio): support Eve runtime context attributes

Radu-Raicea:
## Problem

Eve uses AI SDK 7 `runtimeContext`, which OpenTelemetry exports as `ai.settings.context.*`. PostHog only understood the legacy `ai.telemetry.metadata.*` namespace, so Eve child spans could lose framework and session attribution, and `posthog_distinct_id` did not identify the user.

Follow-up to #74229.

## Changes

- Read PostHog AI context from both namespaces while preserving legacy metadata precedence.
- Normalize namespaced Eve attributes for framework and session grouping.
- Resolve `posthog_distinct_id` per span in OTLP capture.
- Update Eve onboarding to prefer the session initiator, fall back to the current caller, and identify both the Eve turn and its AI SDK child spans.
- Link Eve's runtime-context contract, document the `apiKey` to `projectToken` version boundary, and use per-span export so Vercel Workflow does not depend on a background batch timer.

- Automated Jest coverage catches namespaced Eve spans losing framework, session, custom properties, groups, or user identity while retaining the legacy path.
- Automated Rust coverage catches runtime-context identity falling back to a resource or anonymous ID.
- TypeScript checks, Rust formatting, and CI preflight passed.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Updated the Eve onboarding example with user identification, anonymous-capture guidance, the upstream runtime-context reference, the exporter option version boundary, and per-span export for Workflow runtimes.

## 🤖 Agent context

Codex implemented and validated this follow-up with `/writing-tests`, `/writing-user-facing-copy`, `/writing-code-comments`, `/posthog-efficient-validation`, `/create-pr`, `/llma-git-workflow`, and `/github-efficient-operations`. The implementation keeps the existing metadata namespace authoritative when both forms are present and uses the verified Eve `initiator` and `current` fields rather than introducing a `creator` field.

### DIFF
--- a/docs/onboarding/ai-observability/eve.tsx
+++ b/docs/onboarding/ai-observability/eve.tsx
@@ -14,16 +14,22 @@ export const getEveSteps = (ctx: OnboardingComponentsContext): StepDefinition[] 
             content: (
                 <>
                     <Markdown>
-                        Install the PostHog AI exporter, OpenTelemetry HTTP exporter, and Vercel's OpenTelemetry
-                        package.
+                        Install PostHog AI, its OpenTelemetry peer dependencies, and Vercel's OpenTelemetry package.
                     </Markdown>
 
                     <CodeBlock
                         language="bash"
                         code={dedent`
-                            npm install @posthog/ai @opentelemetry/api @opentelemetry/exporter-trace-otlp-http @vercel/otel
+                            npm install @posthog/ai @opentelemetry/api @opentelemetry/exporter-trace-otlp-http @opentelemetry/sdk-trace-base @vercel/otel
                         `}
                     />
+
+                    <Blockquote>
+                        <Markdown>
+                            **Version note:** This example uses `projectToken`, which is available in `@posthog/ai`
+                            7.19.6 and later. Earlier 7.x versions use `apiKey`.
+                        </Markdown>
+                    </Blockquote>
                 </>
             ),
         },
@@ -51,14 +57,16 @@ export const getEveSteps = (ctx: OnboardingComponentsContext): StepDefinition[] 
                 <>
                     <Markdown>
                         Create `agent/instrumentation.ts`. Eve discovers this file and starts the exporter when your
-                        agent server starts. The optional `events` handler identifies spans using the user who started
-                        the session, falling back to the caller for the current turn.
+                        agent server starts. The optional `events` handler uses [Eve runtime
+                        context](https://eve.dev/docs/guides/instrumentation#runtime-context) to identify spans with the
+                        user who started the session, falling back to the caller for the current turn.
                     </Markdown>
 
                     <CodeBlock
                         language="typescript"
                         code={dedent`
                             import { trace } from '@opentelemetry/api'
+                            import { SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base'
                             import { PostHogTraceExporter } from '@posthog/ai/otel'
                             import { registerOTel } from '@vercel/otel'
                             import { defineInstrumentation } from 'eve/instrumentation'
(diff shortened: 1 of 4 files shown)


docs/onboarding/ai-observability/eve.tsx
```diff
@@ -70,6 +71,20 @@ export const getEveSteps = (ctx: OnboardingComponentsContext): StepDefinition[]
                                     host: process.env.POSTHOG_HOST,
                                   }),
                                 }),
+                              events: {
+                                'step.started'(input) {
```
carlos-marchal-ph, line 89: This step is badged required and the hook adds 14 lines of auth plumbing to it. Every sibling doc marks the distinct id optional inline (`posthog_distinct_id: 'user_123', // optional`).

docs/onboarding/ai-observability/eve.tsx
```diff
@@ -82,6 +97,13 @@ export const getEveSteps = (ctx: OnboardingComponentsContext): StepDefinition[]
                         </Markdown>
                     </CalloutBox>
 
+                    <Markdown>
+                        Eve exposes the user who started the session as `session.auth.initiator`. If that value is not
```
carlos-marchal-ph: This paragraph duplicates the "How this works" callout directly above it, and the sentence about the active span attribute versus the runtime context is plumbing the reader can't act on. Every other AI observability doc puts the one useful line before the snippet and then closes with a standard blockquote:

> Note: If you want to capture LLM events anonymously, omit the `posthog_distinct_id` metadata field. See our docs on [anonymous vs identified events](https://posthog.com/docs/data/anonymous-vs-identified-events) to learn more.

Could we match that shape? 37 other docs use it, and the link is worth having here.

nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.ts
```diff
@@ -62,6 +70,50 @@ function isPromptVersion(value: unknown): value is string | number {
...
+    }
+}
+
+function promotePosthogContext(props: Record<string, unknown>): void {
+    for (const prefix of AI_CONTEXT_PREFIXES) {
+        promotePosthogCustomMetadata(props, prefix)
+    }
+}
+
+function stripProcessedContext(props: Record<string, unknown>): void {
```
carlos-marchal-ph, line 95: Unrecognised runtime-context keys survive into the event. `ai.telemetry.metadata.org_id` gets dropped but `ai.settings.context.org_id` lands verbatim as an `ai.`-prefixed property, and so does any `$ai_*` key that isn't in `STRIPPED_RUNTIME_CONTEXT_KEYS`. I ran the middleware to confirm rather than reading it off.

This isn't a corner case. `runtimeContext` is AI SDK 7's general-purpose per-call bag, handed to tool callbacks, so it holds application data by design. `getRuntimeContextAttributes` exports every key with no allowlist and recurses into nested objects, so `runtimeContext: { tenantId, user: { id, email } }` arrives as three separate attributes and all three land on the event. On the legacy integration it's spread with no opt-in gate at all. Beyond the documented contract, that's an uncontrolled path for customer app data into event properties and unbounded property-name growth.

nodejs/src/ingestion/pipelines/ai/otel/middleware/vercel-ai.test.ts
```diff
@@ -13,6 +13,14 @@ jest.mock('~/ingestion/pipelines/ai/otel/attribute-mapping', () => ({
 }))
 
 const mockedMapOtelAttributes = jest.mocked(mapOtelAttributes)
+const AI_CONTEXT_PREFIXES = ['ai.telemetry.metadata.', 'ai.settings.context.']
+const PROMOTED_CONTEXT_CASES: [string, string, string | number][] = AI_CONTEXT_PREFIXES.flatMap((prefix) => [
```
carlos-marchal-ph, line 17: The two `$ai_prompt_version` rows render identical test names under `'promotes %s%s to event properties'`

carlos-marchal-ph (approved):

docs/onboarding/ai-observability/eve.tsx
```diff
@@ -14,16 +14,22 @@ export const getEveSteps = (ctx: OnboardingComponentsContext): StepDefinition[]
...
+                            npm install @posthog/ai @opentelemetry/api @opentelemetry/exporter-trace-otlp-http @opentelemetry/sdk-trace-base @vercel/otel
                         `}
                     />
+
+                    <Blockquote>
+                        <Markdown>
+                            **Version note:** This example uses `projectToken`, which is available in `@posthog/ai`
+                            7.19.6 and later. Earlier 7.x versions use `apiKey`.
+                        </Markdown>
+                    </Blockquote>
```
bradlocking, line 32: 👍

PR #74331, opened 2026-07-28
feat(experiments): add tags support with table filtering and bulk updates

rubychilds:
## Problem

Experiments have no tags, unlike feature flags. There's no way to organize or filter experiments by team, initiative, or theme: no tags on the experiment detail page, no tags column or tag filter on the experiments table, no row multi-select, and no bulk tagging. This PR brings experiments to parity with the feature flags table and detail page.

## Changes

**Backend**

- `TaggedItem` gains a nullable `experiment` FK, wired into the existing generic tagging system. Shipped as the established safe 3-migration pattern (same as the ticket/account/endpoint FKs before it): `SeparateDatabaseAndState` with a `NOT VALID` check-constraint swap, `CREATE INDEX CONCURRENTLY` builds, then `UNIQUE USING INDEX`. No data migration, no table rewrite, no blocking scans on `posthog_taggeditem`.
- `ExperimentSerializer` / `ExperimentBasicSerializer` expose `tags` on detail and list responses. Create and update pop tags around the facade layer's strict field allowlist.
- The list API accepts `tags` / `excluded_tags` filters, and the viewset adopts `TaggedItemViewSetMixin` for a `bulk_update_tags` endpoint (add / remove / set, per-object editor ACL, PAT access via `experiment:write`) plus a `matching_ids` endpoint for select-all-across-pages.
- MCP: `experiment-create` / `experiment-update` accept tags, get/list/update responses include them, and a new `experiments-bulk-update-tags-create` tool mirrors the flags one.

**Frontend**

- Experiments table: tags column, "Tags" / "Exclude tags" filter dropdowns (URL-synced), row multi-select with an "Update tags" bulk action and a select-all-matching banner.
- Experiment detail page: a new "Advanced options" card on the Settings tab with inline tag editing, laid out to match the feature flag detail page (Advanced options card on the left, the statistics settings grouped in a card on the right, Notifications below Statistics).
- Saving tags refreshes `tagsModel`, so newly created tags immediately appear in the table's tag filter without a reload.

All built from existing shared pieces: `TaggedItemSerializerMixin` / `TaggedItemViewSetMixin`, `ObjectTags`, `TagSelect`, `BulkUpdateTagsButton`, and `LemonTable`'s `bulkSelection`.

Bulk tagging on the table:

bulk tagging

Screenshot 2026-07-30 at 11 13 36 AM

Settings tab with the Advanced options card:

settings tab

> [!NOTE]
> `WebExperiment` is a proxy model over the same table, so tags apply to web experiments too. Bulk delete and bulk copy-to-project (which flags also have in their bulk bar) are left as follow-ups.

Manual tests:
- adding a tag, and removing it. Instant save on adding a tag.
- viewing tag as a column in a table, and being able to filter on that tag.

Automated tests, all run locally and green:

- **Backend** (`TestExperimentTags`, 10 cases in `test_presentation_api.py`): create with tags persists and normalizes them (guards the facade `expected_fields` allowlist, where forgetting to pop tags breaks every tagged create); partial update replaces tags while an update omitting tags leaves them untouched; list includes tags and `?tags=` / `?excluded_tags=` filter correctly (parameterized); `bulk_update_tags` mutates plus reports missing IDs as skipped; PAT with `experiment:write` works (guards the `scope_object_write_actions` opt-in, a config-only regression no session test would catch); `matching_ids` applies list filters. Full experiments presentation suite (276 tests) passes.
- **Frontend** (`experimentLogic.test.ts`, 2 new cases): saving tags dispatches `tagsModel.loadTags` so the tag filter offers new tags without a reload (the exact staleness bug found during manual testing), and an update without tags doesn't. Full suite: 91 passed.
- `makemigrations --check` clean, migrations applied and reversed locally, `hogli ci:preflight --fix` reports 0 failures, frontend typecheck green.

Also verified live against the local dev stack with a scripted browser (Playwright): created tagged experiments, filtered the table by tag and by excluded tag, bulk-added tags to multi-selected rows, edited tags on the Settings tab, and confirmed persistence through the API. Malformed filter values (`?tags=not-json`) are ignored gracefully and empty-tags bulk removes return a clean 400, matching flags behavior.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored with Claude Code (PostHog Code session). Skills invoked while producing this PR: /django-migrations, /improving-drf-endpoints, /implementing-mcp-tools, /writing-tests, /writing-user-facing-copy, /verify, /run-posthog.

Decisions along the way: the `TaggedItem` migrations were hand-written on the endpoint-FK precedent rather than autogenerated, since `makemigrations` would emit a blocking CHECK validation and non-concurrent index builds. Tags editing initially landed in the scene-panel sidebar, then moved to an "Advanced options" card on the Settings tab to match the feature flag detail page after review feedback; the Settings tab was restructured into the flags-style two-column card layout in the same pass. Bulk delete and copy-to-project for experiments were deliberately left out to keep the PR scoped to tags.

---


products/experiments/backend/experiment_service.py
```diff
@@ -119,6 +120,16 @@
...
 ExperimentCreationMode = Literal["new", "duplicate", "copy_to_project"]
 
+
+def _parse_tag_names(value: Any) -> list[str]:
+    """Parse a tags query param that arrives as a list or a JSON-encoded string."""
+    try:
+        tags = value if isinstance(value, list) else json.loads(value) if isinstance(value, str) else []
+    except (json.JSONDecodeError, TypeError):
+        return []
+    return [tag for tag in tags if isinstance(tag, str)]
```
haacked, line 160: `blocking`: A tags filter carrying a JSON scalar, like `?tags=5`, `?tags=true`, or `?tags=null`, returns a 500 instead of being ignored. The `try` above covers only the decode, so the scalar parses fine and this comprehension then raises `TypeError: 'int' object is not iterable`, uncaught. Same for `excluded_tags`, and on `matching_ids` since it parses through here too. Quieter variants are wrong rather than fatal: `?tags="growth"` filters on the six characters of the string, and `?tags={"a":1}` filters on the dict's keys.

The flags version keeps its queryset call [inside the try](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/products/feature_flags/backend/api/feature_flag.py#L3745), which is why the same input degrades to an unfiltered list there. Worth a `tags=123` case in `test_list_filters_by_tags` so the ignore-and-return-everything path is actually exercised.

```suggestion
    if not isinstance(tags, list):
        return []
    return [tag for tag in tags if isinstance(tag, str)]
```

products/experiments/backend/presentation/views.py
```diff
@@ -400,6 +426,50 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
+        )
+
+        # Only IDs are needed; drop the class queryset's joins/prefetches (only("id") would
+        # otherwise conflict with select_related).
+        queryset = queryset.select_related(None).prefetch_related(None)
+
+        if not self.user_access_control:
+            editable_ids = list(queryset.values_list("id", flat=True))
+        else:
+            experiments = list(queryset.only("id"))
```
haacked: `blocking`: "Select all matching" costs two extra queries per experiment. `only("id")` leaves `created_by_id` deferred, and the creator check inside `get_user_access_level` reads the [`created_by` object](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/posthog/rbac/user_access_control.py#L1332), so each row triggers a deferred-field refresh plus a fetch of the user. The button only renders above 100 experiments and the action is unpaginated, so the floor is roughly 200 extra queries per click, growing linearly from there. Every org pays it, including ones with access controls off, because the precheck that would short-circuit them runs after both queries have already fired.

`bulk_object_access_levels` resolves the same levels from one preload and takes `(pk, created_by_id)` pairs, so no instances need hydrating. It also collapses the `if not self.user_access_control` branch above, which never fires, since that property always returns an instance and the class defines no `__bool__`:

```python
        rows = list(queryset.values_list("id", "created_by_id"))
        levels = self.user_access_control.bulk_object_access_levels(
            "experiment", [(str(pk), created_by_id) for pk, created_by_id in rows]
        )
        editable_ids = [
            pk
            for pk, _ in rows
            if (user_access_level := levels.get(str(pk)))
            and access_level_satisfied_for_resource("experiment", user_access_level, "editor")
        ]
```

Flags' `matching_ids` has the same pattern, so the same fix applies there.

frontend/src/scenes/experiments/ExperimentView/SettingsTab.tsx
```diff
@@ -52,101 +68,155 @@ export function SettingsTab(): JSX.Element {
...
+            <div className="flex gap-4 flex-wrap items-start">
+                <div className="flex-1 min-w-64 flex flex-col gap-4">
+                    <div className="rounded border p-4 bg-bg-light flex flex-col gap-4">
+                        <div className="font-semibold">Advanced options</div>
+                        <div className="flex flex-col gap-2">
+                            <label className="text-sm font-medium">Tags</label>
+                            {canEditExperiment ? (
+                                <ObjectTags
+                                    tags={experiment.tags ?? []}
+                                    onChange={(tags) => updateExperimentSettings({ tags })}
```
haacked: `blocking`: Adding or removing a tag re-runs every metric query on the experiment. `updateExperimentSettings` persists and then calls [`refreshExperimentResults(true, 'config_change')`](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/frontend/src/scenes/experiments/experimentLogic.tsx#L2694), which clears the primary and secondary metric results and re-queries them with the cache bypassed. On the Results tab every metric card flips back to loading after a tag edit, and the ClickHouse work is redone from scratch. The action's other three callers change stats method, CUPED, and conversion windows, which genuinely change how metrics compute; tags don't.

`updateExperiment` still reaches the `updateExperimentSuccess` listener that reloads `tagsModel`, so the tag-refresh behavior your new tests cover is preserved:

```suggestion
                                    onChange={(tags) => updateExperiment({ tags, update_feature_flag_params: false })}
```

That needs `updateExperiment` added to the `useActions(experimentLogic)` destructure at line 32.

products/experiments/backend/experiment_service.py
```diff
@@ -3700,6 +3711,19 @@ def filter_experiments_queryset(
                 # narrow the queryset by primary key to preserve ordering and pagination.
                 queryset = queryset.filter(pk__in=self._experiments_matching_event(queryset, event))
 
+            tags = _parse_tag_names(query_params.get("tags"))
+            if tags:
+                queryset = queryset.filter(tagged_items__tag__name__in=tags).distinct()
```
haacked: `suggestion`: Filtering by tag makes Postgres dedupe the whole matched set before pagination. The join on `tagged_items` plus `.distinct()` runs `SELECT DISTINCT` across every column of the [six tables the list queryset joins](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/products/experiments/backend/presentation/views.py#L373), including jsonb ones like `team.test_account_filters` and `feature_flag.filters`. Flags get away with the same `.distinct()` because their queryset uses only `prefetch_related`, no joins.

The ID-subquery form that the `excluded_tags` block just below already uses avoids both the row multiplication and the dedupe:

```suggestion
                # Match by ID subquery so the tagged_items join can't multiply rows, which
                # would otherwise need a DISTINCT over every select_related column.
                experiments_with_tags = Experiment.objects.filter(
                    team__project_id=self.team.project_id, tagged_items__tag__name__in=tags
                ).values("pk")
                queryset = queryset.filter(pk__in=experiments_with_tags)
```

products/experiments/backend/presentation/views.py
```diff
@@ -400,6 +426,50 @@ def safely_get_queryset(self, queryset) -> QuerySet:
             request_data=getattr(request, "data", None),
         )
 
+    @extend_schema(
```
haacked, line 477: `suggestion`: The generated client for this endpoint only accepts `tags`, so callers can't reach the rest of its filtering. The docstring says `matching_ids` takes the same query params as `list` and the body does hand them all to `filter_experiments_queryset`, but `parameters=` declares one. That's why `ExperimentsMatchingIdsRetrieveParams` comes out with a lone `tags` field, and why [Experiments.tsx:651](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/frontend/src/scenes/experiments/Experiments.tsx#L651) hand-builds the URL and casts the response instead of calling the function this PR just generated.

Lifting the list's parameter list into a module-level constant and pointing both `list` and `matching_ids` at it would fix the schema in one place. After a regen the frontend can drop the template-string URL and the cast.

products/experiments/backend/test/test_presentation_api.py
```diff
@@ -8846,3 +8846,107 @@ def test_api_schema_exposes_every_runtime_field(self) -> None:
...
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json()["tags"] == ["one"]
+
+        response = self.client.patch(
+            f"/api/projects/{self.team.id}/experiments/{experiment['id']}/",
+            {"tags": ["two", "three"]},
+            format="json",
+        )
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert sorted(response.json()["tags"]) == ["three", "two"]
```
haacked, line 8992: `suggestion`: Nothing covers clearing tags with `tags: []`, which is what the Settings tab sends when you remove the last one. The omitted-tags case is covered right here, and both behaviors hang off the same `pop("tags", None)` in `update`: `None` preserves, `[]` clears. The shared `test_empty_tags_clears_all_tags` doesn't reach this path, since it goes through the mixin's own `update`, which reads `initial_data` rather than `validated_data`.

A GET after the replace assertion is worth adding too. The PATCH response is rendered from `prefetched_tags`, which holds exactly what was just written, so it would still read `["three", "two"]` even if the old row were never deleted.

posthog/migrations/1293_taggeditem_experiment.py
```diff
@@ -0,0 +1,284 @@
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
```
haacked: `suggestion`: Eight of these app pins block unrelated rollbacks for no reason. The only new relation this migration introduces is the FK to `experiments.Experiment`; the other FK columns already exist in model state from earlier migrations that carry their own dependencies. As written, rolling back `conversations` or `product_analytics` would have to unapply 1277-1279 first.

The endpoint precedent does the same constraint and unique_together swap over the same columns with only its FK target plus `posthog` ([1179_taggeditem_endpoint.py:6](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/posthog/migrations/1179_taggeditem_endpoint.py#L6)), and by the time it landed, dashboards, actions, event_definitions, conversations and customer_analytics had already moved into their own apps. Trimming to `experiments` and `posthog` looks safe; worth confirming a fresh `migrate` still applies cleanly.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -551,6 +605,75 @@ const ExperimentsTable = ({
                             page: 1,
                         })
                     }
+                    bulkSelection={{
+                        getKey: (experiment: Experiment): number =>
+                            typeof experiment.id === 'number' ? experiment.id : -1,
+                        isRowSelectable: (experiment: Experiment) =>
```
haacked: `suggestion`: Changing a filter leaves the previous selection intact, so the bulk bar can tag experiments that are no longer on screen. `useBulkSelection` keeps its keys across the reload, the bar still reads "N experiments selected", and Update tags posts the stale IDs. After a "Select all N matching" click it also keeps claiming "All N matching experiments selected", because `matchingExperimentIds` only resets on a successful bulk update (line 669).

Clearing both when the filter identity changes fixes it. `FeatureFlags.tsx` has the same gap, so whether to fix both here is your call.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -551,6 +605,75 @@ const ExperimentsTable = ({
...
+                                            onClick={async () => {
+                                                setMatchingExperimentIdsLoading(true)
+                                                try {
+                                                    const { limit, offset, ...restFilters } = paramsFromFilters
+                                                    const response = (await api.get(
+                                                        `api/projects/${currentProjectId}/experiments/matching_ids/?${toParams(restFilters)}`
+                                                    )) as { ids: number[]; total: number }
+                                                    setMatchingExperimentIds(response.ids)
+                                                    ctx.setSelectedKeys(response.ids)
+                                                } finally {
```
haacked: `suggestion`: If the `matching_ids` request fails, clicking "Select all N matching experiments" does nothing and explains nothing. The `try`/`finally` has no `catch`, so the spinner stops, the selection is unchanged, and the rejection surfaces only as an unhandled promise in the console. A user would click again and conclude the button is broken.

```suggestion
                                                } catch {
                                                    lemonToast.error("Couldn't select all matching experiments")
                                                } finally {
```

products/experiments/backend/presentation/views.py
```diff
@@ -340,9 +362,13 @@ class EnterpriseExperimentsViewSet(
     ForbidDestroyModel,
     TeamAndOrgViewSetMixin,
     AccessControlViewSetMixin,
+    TaggedItemViewSetMixin,
```
haacked, line 395: `suggestion`: Bulk tagging pulls the full experiment payload for every selected row. The mixin builds its queryset from `self.get_queryset()`, and `safely_get_queryset` only narrows for the list action, so up to 500 experiments come back joined across team, feature_flag, created_by, exposure_cohort and holdout with every metrics and filters jsonb column loaded, plus the saved-metric prefetch. The mutation itself reads only `id`, `team_id`, `name` and the tags.

A `bulk_update_tags` branch in `safely_get_queryset` that keeps `created_by` for the editor check and drops the rest would cut it down.

products/experiments/backend/test/test_presentation_api.py
```diff
@@ -8846,3 +8846,107 @@ def test_api_schema_exposes_every_runtime_field(self) -> None:
...
+        )
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json()["updated"] == [{"id": experiment["id"], "tags": ["via-pat"]}]
+
+    def test_matching_ids_applies_list_filters(self):
+        tagged = self._create_experiment("Tagged", "tags-matching-flag-a", tags=["growth"])
+        self._create_experiment("Untagged", "tags-matching-flag-b")
+        response = self.client.get(f'/api/projects/{self.team.id}/experiments/matching_ids/?tags=["growth"]')
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json() == {"ids": [tagged["id"]], "total": 1}
```
haacked, line 9069: `suggestion`: The editor filter in `matching_ids` never runs in this test. Both experiments are created by `self.user` with default access, so the check at [views.py:464](https://github.com/PostHog/posthog/blob/fb25e7990f9e908ec6109691a90369186336d4d3/products/experiments/backend/presentation/views.py#L464) passes everything and the endpoint's "only IDs the user has permission to edit" goes unverified. A regression there feeds view-only IDs into the select-all banner, and the user picks "select all N matching" then gets a silent partial result when `bulk_update_tags` skips them.

`test_excludes_private_experiments` in `test_session_context_api.py:603` has the setup to copy: `_enable_access_controls()` plus an `AccessControl` row scoped to the experiment's pk.

frontend/src/scenes/experiments/ExperimentView/SettingsTab.tsx
```diff
@@ -52,101 +68,155 @@ export function SettingsTab(): JSX.Element {
...
-                        type="internal_destination"
-                        subTemplateIds={['experiment-significant']}
-                        forceFilterGroups={[
-                            {
-                                events: [{ id: '$experiment_metric_significant', type: 'events' }],
-                                properties: [
+
+                <div className="flex-[2] min-w-80 flex flex-col gap-4">
+                    <div className="rounded border p-4 bg-bg-light flex flex-col gap-4">
+                        <div className="font-semibold">Statistics</div>
```
haacked: `suggestion`: The Settings tab now has no headings, so there's nothing for a screen reader user to navigate between. Statistics and Notifications went from `<h2>` to `<div className="font-semibold">`, and CUPED, Baseline variant and Conversion windows went from `<h2>` to `<label>`. A `<label>` with no `htmlFor` and no control inside it names nothing, so those five section titles are now carried by font weight alone.

```suggestion
                        <h2 className="font-semibold">Statistics</h2>
```

products/experiments/backend/presentation/serializers.py
```diff
@@ -756,6 +765,9 @@ def create(self, validated_data: dict, *args: Any, **kwargs: Any) -> Experiment:
 
         # Pop fields not needed for DTO but needed for validation
         validated_data.pop("update_feature_flag_params", None)
+        # Tags live in the shared TaggedItem table, not on the experiment row, so they must not
```
haacked: `nit`: This says the pop keeps `tags` out of the facade DTO, but `to_facade_dto` builds from `self.validated_data` and names every field explicitly, so tags never reach it either way. What the pop actually protects is the `expected_fields` check below.

products/experiments/backend/presentation/views.py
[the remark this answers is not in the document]
rubychilds: Fixed in 85cea5725e5 — `bulk_update_tags` now returns the same skipped reason ("Not found or no edit access") for both missing and inaccessible objects, so responses no longer distinguish restricted IDs from nonexistent ones. Applied in the shared `TaggedItemViewSetMixin`, so feature flags, dashboards, insights, and event definitions get the same fix.

products/experiments/backend/presentation/views.py
```diff
@@ -351,9 +373,13 @@ class EnterpriseExperimentsViewSet(
...
```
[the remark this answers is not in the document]
rubychilds: Fixed in 85cea5725e5 — the request serializer now caps `tags` at 100 per request and enforces the model's 255-character tag name limit, alongside the existing 500-ID cap. Guarded by a new test in `test_tagged_item.py` mirroring the existing ids-cap test.

haacked (commented):
Three blocking issues inline; the rest are suggestions and a nit.

products/experiments/backend/experiment_service.py
rubychilds: Fixed in db8d13ec5c2 — `_parse_tag_names` now ignores anything that doesn't decode to a list, so `?tags=5`, `?tags=true`, `?tags=null`, `?tags=\"growth\"`, and `?tags={...}` all fall through to no filtering instead of 500ing or misfiltering. Applies to `excluded_tags` and `matching_ids` too since they share the helper.

products/experiments/backend/presentation/views.py
rubychilds: Fixed in db8d13ec5c2 — the queryset now does `select_related(\"created_by\").only(\"id\", \"created_by\")`, so the creator check reads from the single joined query instead of two deferred-refresh queries per row.

frontend/src/scenes/experiments/ExperimentView/SettingsTab.tsx
rubychilds: Fixed in db8d13ec5c2 — the tags editor now calls `updateExperiment({ tags })` directly, so a tag edit persists without triggering the config-change results refresh. `updateExperimentSettings` stays for the settings that genuinely change metric computation.

products/experiments/backend/experiment_service.py
rubychilds: Fixed in db8d13ec5c2 — the tags filter now uses a `pk__in` subquery (same shape as `excluded_tags`) instead of joining `tagged_items` and running `SELECT DISTINCT` over the six-table list queryset.

products/experiments/backend/presentation/views.py
rubychilds: Fixed in db8d13ec5c2 — the list parameters moved to a shared `EXPERIMENT_LIST_FILTER_PARAMETERS` constant that both `list` and `matching_ids` declare, so `ExperimentsMatchingIdsRetrieveParams` now carries the full filter set.

products/experiments/backend/test/test_presentation_api.py
rubychilds: Fixed in db8d13ec5c2 — the update test now PATCHes `tags: []` after the replace and asserts both the response and a follow-up GET come back empty, covering the clear path through `ExperimentSerializer.update`'s `validated_data` pop.

posthog/migrations/1293_taggeditem_experiment.py
rubychilds: Fixed in db8d13ec5c2 — trimmed to the FK target (`experiments`) plus the previous `posthog` migration, matching the endpoint precedent.

frontend/src/scenes/experiments/Experiments.tsx
rubychilds: Fixed in db8d13ec5c2 — added a `clearSelectionKey` option to the shared `bulkSelection` config (LemonTable/useBulkSelection) and keyed it on the serialized filters (page excluded, so cross-page selection still works). `matchingExperimentIds` resets on the same signal, so the \"All N matching selected\" claim can't go stale either.

frontend/src/scenes/experiments/Experiments.tsx
rubychilds: Fixed in db8d13ec5c2 — added the catch with a `lemonToast.error` so a failed request tells the user to try again instead of silently doing nothing.

products/experiments/backend/presentation/views.py
rubychilds: Fixed in db8d13ec5c2 — `safely_get_queryset` now has a `bulk_update_tags` branch that drops the detail joins/prefetches (keeping `created_by` for the access check) and defers the heavy JSON columns.

products/experiments/backend/test/test_presentation_api.py
rubychilds: Fixed in db8d13ec5c2 — added `test_matching_ids_excludes_view_only_experiments`: an experiment created by another user with an object-level `viewer` access control is excluded from the response while the requester's own experiment is returned.

frontend/src/scenes/experiments/ExperimentView/SettingsTab.tsx
rubychilds: Fixed in db8d13ec5c2 — card titles are `h2` again and the sub-section titles (Tags, Method, CUPED, Baseline variant, Conversion windows) are `h3`, so the tab has a navigable heading hierarchy and no control-less `label`s.

products/experiments/backend/presentation/serializers.py
rubychilds: Fixed in db8d13ec5c2 — the comment now says what the pop actually protects: the `expected_fields` check.

products/experiments/backend/presentation/serializers.py
```diff
@@ -229,6 +232,11 @@ class ExperimentBaseSerializer(UserAccessControlSerializerMixin, serializers.Mod
             "them, such as duplicate and copy-to-project."
         ),
     )
+    tags = serializers.ListField(
```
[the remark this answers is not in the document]
rubychilds: Fixed in db8d13ec5c2 — the experiment `tags` write field is capped at 100 items with 255 characters each, mirroring the bulk endpoint's caps.

jurajmajerik:
I like the 2 column layout! However:
- There's large empty space on the left; I think the majority of content should be on the left, with some empty space on the right?
- Calling tags "Advanced options" seems a bit off to me

Alternatively, I think tagging could live in the `Info.tsx` section, if you find a way to squeeze it in.

image

jurajmajerik:
The filtering selects have inconsistent sizes. I think the bigger size is actually the one to use here.

image

rubychilds:
@jurajmajerik filter dropdown sizes have been fixed. Reason for the layout is to align it with feature flags, and where tags are placed there. Agree it's not ideal honestly, but would prefer to make a wider layout fix of both of these pages in another PR, than conflating this one. 

This is feature flags and where there tags are. I want to try and ensure more consistency in our UX, so user isn't working out how to add tags differently in every product. 

Screenshot 2026-08-11 at 1 14 18 PM

> I like the 2 column layout! However:
> 
> There's large empty space on the left; I think the majority of content should be on the left, with some empty space on the right?
> Calling tags "Advanced options" seems a bit off to me
> Alternatively, I think tagging could live in the Info.tsx section, if you find a way to squeeze it in.

frontend/src/scenes/experiments/ExperimentView/SettingsTab.tsx
```diff
@@ -52,101 +68,157 @@ export function SettingsTab(): JSX.Element {
...
+                    <div className="rounded border p-4 bg-bg-light flex flex-col gap-4">
+                        <h2 className="font-semibold text-base m-0">Advanced options</h2>
+                        <div className="flex flex-col gap-2">
+                            <h3 className="text-sm font-medium m-0">Tags</h3>
+                            {canEditExperiment ? (
+                                <ObjectTags
+                                    tags={experiment.tags ?? []}
+                                    // Not updateExperimentSettings: tags don't affect metric
+                                    // computation, so don't trigger its results refresh.
+                                    onChange={(tags) => updateExperiment({ tags })}
```
haacked: `blocking`: Picking two tags in a row here silently deletes the first one. Every change saves immediately, and `ObjectTags` passes `experiment.tags` straight through as [the input's current value](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/lib/components/ObjectTags/ObjectTags.tsx#L81), which `LemonInputSelect` [re-reads to build the next selection](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/lib/lemon-ui/LemonInputSelect/LemonInputSelect.tsx#L453). Nothing updates `experiment.tags` until [the PATCH response lands](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/scenes/experiments/experimentLogic.tsx#L3618). So the second pick computes from the list as it was before the first pick, and sends a payload holding only the second tag. `set_tags_on_object` then deletes every tag that isn't in it.

The window is real and nothing closes it. The dropdown stays open after a selection in multiple mode, so two clicks a few hundred milliseconds apart is the ordinary way to use this. `saving` only drives a loading bar, not `disabled`, so the input stays clickable for the whole round trip. The version guard can't catch it either, because [tags are popped out](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/experiments/backend/presentation/serializers.py#L874) before the service sees the payload, which leaves `update_data` empty and no conflict to detect.

Every other editable `ObjectTags` caller keeps a local buffer so the input's value updates synchronously. [`SceneTags.tsx:28-41`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/lib/components/Scenes/SceneTags.tsx#L28-L41) is the closest model.

```tsx
const [draftTags, setDraftTags] = useState<string[] | null>(null)
// …
<ObjectTags
    tags={draftTags ?? experiment.tags ?? []}
    onChange={(tags) => {
        setDraftTags(tags)
        updateExperiment({ tags })
    }}
    saving={experimentUpdateLoading}
    tagsAvailable={allExistingTags.filter((tag: string) => !experiment.tags?.includes(tag))}
    actionButtonSize="medium"
    data-attr="experiment-tags"
/>
```

products/experiments/backend/presentation/views.py
```diff
@@ -418,6 +458,45 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
+            self.get_queryset(),
+            action="list",
+            query_params=request.query_params,
+        )
+
+        # Drop the class queryset's joins/prefetches; the access check below reads only the
+        # experiment ID and its creator, so load exactly those (a bare only("id") would leave
+        # created_by deferred and cost two refresh queries per row).
+        queryset = (
+            queryset.select_related(None).prefetch_related(None).select_related("created_by").only("id", "created_by")
```
haacked: `suggestion`: This loads every column of the joined user row, including the password hash and five JSON config columns, once per matching experiment on an endpoint with no pagination. Naming the bare foreign key in `only("id", "created_by")` puts it in the select mask with an empty sub-mask, and Django reads an empty mask as "no restriction", so the join is unrestricted. I confirmed this against Django 5.2.14: the emitted SQL is `SELECT experiment.id, experiment.created_by_id, user.id, user.email, user.password, …`. The per-row refresh queries really are gone, so the comment above is right about that; it's the "load exactly those" part that isn't what happens.

The `preload_object_access_controls` call on line 491 also performs unnecessary work. It fills `self._cache`, but `_get_access_controls` [returns from a fresh scan of `_cached_access_controls`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/posthog/rbac/user_access_control.py#L657) before it ever consults that cache. So the team's access-control rows are scanned once per experiment and the preload's result is discarded.

[`bulk_object_access_levels`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/posthog/rbac/user_access_control.py#L1413) collapses both. It takes `(pk, created_by_id)` pairs, so it feeds straight from `values_list` with no join and no model instances, and it groups the rows once instead of per object. [`posthog/api/search.py:236`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/posthog/api/search.py#L236) already uses it, and `experiment` isn't in `RESOURCE_FALLBACK_MAP`, so the semantics match the per-object path you have now. If you'd rather keep this shape, `only("id", "created_by__id")` alone fixes the wide join and still costs zero extra queries, since the creator check only compares primary keys.

```python
        if not self.user_access_control:
            editable_ids = list(queryset.values_list("id", flat=True))
        else:
            # values_list ignores the class queryset's joins and prefetches, so each row is two
            # integers and no model instance is built.
            rows = list(queryset.values_list("id", "created_by_id"))
            access_levels = self.user_access_control.bulk_object_access_levels(
                "experiment", [(str(experiment_id), created_by_id) for experiment_id, created_by_id in rows]
            )
            editable_ids = [
                experiment_id
                for experiment_id, _ in rows
                if (user_access_level := access_levels.get(str(experiment_id)))
                and access_level_satisfied_for_resource("experiment", user_access_level, "editor")
            ]
```

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -206,13 +239,35 @@ const ExperimentsTable = ({
...
     const hasMultipleProjects = (currentOrganization?.projects?.length ?? 0) > 1
 
+    const [matchingExperimentIds, setMatchingExperimentIds] = useState<readonly number[] | null>(null)
+    const [matchingExperimentIdsLoading, setMatchingExperimentIdsLoading] = useState(false)
+    // State rather than a ref so mounting the slot re-renders and the bar's portal finds it.
+    const [bulkSelectionBarContainer, setBulkSelectionBarContainer] = useState<HTMLDivElement | null>(null)
+
+    // Changing a filter changes which experiments the bulk bar refers to, so drop the selection and
+    // any cached "all matching" IDs. Page is excluded: selections deliberately span pages.
+    const { page: _page, ...filtersWithoutPage } = filters
```
haacked: `suggestion`: Sorting the table throws away the selection. `order` is part of this identity string, and [the sort handler](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/scenes/experiments/Experiments.tsx#L621) writes into it, so clicking a column header clears the selection and nulls `matchingExperimentIds`. Select twenty experiments, click a header to check something, and the bulk bar disappears. It's worst right after "Select all N matching", where one header click discards a request that just fetched every matching ID.

The comment above says the clear exists because a filter change alters which experiments the bar refers to. Sorting doesn't do that: the matched set and the count are identical, only the row order moves.

```suggestion
    const { page: _page, order: _order, ...selectionScopeFilters } = filters
    const filterIdentity = JSON.stringify(selectionScopeFilters)
```

frontend/src/lib/lemon-ui/LemonTable/useBulkSelection.ts
```diff
@@ -42,6 +42,9 @@ export interface BulkSelectionConfig<T extends Record<string, any>, K extends Bu
      *  it inline with an existing toolbar row. While the element is null (not mounted yet), the bar
      *  falls back to its default position above the table. */
     barPortalTarget?: HTMLElement | null
+    /** Clear the selection whenever this value's identity changes (e.g. a serialized filter state),
+     *  so the bar can't act on rows that a filter change removed from view. */
+    clearSelectionKey?: unknown
```
haacked: `suggestion`: A caller who passes an object here hangs the tab, and the type doesn't stop them. `unknown` accepts `clearSelectionKey={filters}`, and an object literal gets a fresh identity every render, so the effect at line 90 re-runs every render and calls `setSelectedKeysState([])`, whose new array triggers the next render. The doc comment describes the value as a serialized filter state, but passing the filters object itself is the obvious wrong turn, and eight tables reach this hook.

A functional update removes the loop outright, since React bails out when the setter returns the same reference. It also drops a wasted commit on the experiments table today: `filterIdentity` changes on every keystroke in the search box, and each change re-renders all 100 rows to replace an already-empty selection with a different empty array. Narrowing the type on top of that makes the contract something the compiler checks. The same declaration is repeated at line 70.

```suggestion
    clearSelectionKey?: string | number | boolean | null
```

and at line 95:

```ts
        setSelectedKeysState((current) => (current.length === 0 ? current : []))
```

products/experiments/backend/experiment_service.py
```diff
@@ -131,6 +132,22 @@ class CleanupRepositoryTarget(TypedDict):
...
+    Anything that doesn't decode to a list is ignored rather than an error: a scalar like
+    ``?tags=5`` or ``?tags="growth"`` decodes fine but isn't a tag list.
+    """
+    try:
+        tags = value if isinstance(value, list) else json.loads(value) if isinstance(value, str) else []
+    except (json.JSONDecodeError, TypeError):
+        return []
+    if not isinstance(tags, list):
+        return []
+    return [tag for tag in tags if isinstance(tag, str)]
```
haacked: `suggestion`: A caller who filters on `Growth` gets nothing back, and one who excludes `Growth` gets back everything they asked to exclude. Tag names are lowercased on write by `tagify`, which [your own test](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/experiments/backend/test/test_presentation_api.py#L9045) asserts, but this helper returns the query value raw and `tagged_items__tag__name__in` is case-sensitive. The UI is safe because it feeds tag names from `tagsModel`, which already holds normalized values. The new callers aren't: the OpenAPI `tags` and `excluded_tags` params and the MCP `experiment-list` filter, where an agent turns "experiments tagged Growth" into `["Growth"]`.

One caveat on scope. Every tag filter in the repo behaves this way, including [the flags reference](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/feature_flags/backend/api/feature_flag.py#L3758) that this helper was forked from, so fixing it only here makes experiments diverge from flags. If you'd rather not carry that, the alternative is moving this helper into `posthog/api/tagged_item.py` and calling it from both sides, which also hands flags the malformed-input hardening you added.

```suggestion
    return [normalized for tag in tags if isinstance(tag, str) and (normalized := tagify(tag))]
```

Needs `from posthog.models.tag import tagify`. A `'tags=["Growth"]'` case on the existing parameterized list would cover it; every case there uses an already-lowercase name.

posthog/api/tagged_item.py
```diff
@@ -184,6 +184,11 @@ def update(self, instance, validated_data):
 
 
 BULK_UPDATE_TAGS_MAX_IDS = 500
+BULK_UPDATE_TAGS_MAX_TAGS = 100
+TAG_NAME_MAX_LENGTH = 255  # Mirrors Tag.name's max_length
```
haacked: `suggestion`: This constant reads like a rule for the whole module, but only the bulk serializer applies it, so a 300-character tag still returns a 500 on every single-object write. `TaggedItemSerializerMixin.tags` at line 154 is a bare `ListField` with no child validator, and that's the field dashboards, insights, actions, tickets, and endpoints inherit on create and update. The tag reaches `Tag.objects.get_or_create`, which doesn't run validators, and Postgres rejects it against `varchar(255)`. Experiments is the only resource that escapes this, because [presentation/serializers.py:235](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/experiments/backend/presentation/serializers.py#L235) redeclares the field with the cap.

Putting the length cap on the mixin's field covers every resource and lets that override drop the length half. Worth keeping the 100-item count cap off this path though: the length cap only turns a 500 into a 400, while a count cap could reject requests existing callers make today.

The constants block would need to move above the class at line 149.

```python
    tags = serializers.ListField(required=False, child=serializers.CharField(max_length=TAG_NAME_MAX_LENGTH))
```

posthog/api/tagged_item.py
```diff
@@ -198,8 +203,9 @@ class BulkUpdateTagsRequestSerializer(serializers.Serializer):
         help_text="'add' merges with existing tags, 'remove' deletes specific tags, 'set' replaces all tags.",
     )
     tags = serializers.ListField(
```
haacked: `suggestion`: This caps the request body but not the work, so a maxed-out request still runs for minutes and holds a web worker the whole time. The cost is `ids` times `tags`, not either one: `apply_bulk_tag_changes` calls `set_tags_on_object` per object, and that runs a `get_or_create` for the tag and another for the link, per tag. At 500 objects and 100 tags that's 50,000 links and at least 100,000 round trips even when every link already exists. New links cost roughly ten queries each on top, because `TaggedItem.save()` calls `full_clean()` and that revalidates both foreign keys plus three constraints inside a savepoint.

The per-object write path isn't yours and every existing mixin user already pays it. What's new is that the cap sitting right here reads like a bound on the work, and it isn't. Bounding the product instead would make it one:

```python
        if len(attrs["ids"]) * len(attrs["tags"]) > BULK_UPDATE_TAGS_MAX_LINKS:
            raise serializers.ValidationError(
                {"tags": f"At most {BULK_UPDATE_TAGS_MAX_LINKS} tag assignments per request (ids x tags)."}
            )
```

with `BULK_UPDATE_TAGS_MAX_LINKS = 5000` alongside the other constants. That still allows 500 objects with 10 tags, or 50 with 100, which covers anything the UI can produce.

posthog/api/tagged_item.py
```diff
@@ -371,13 +377,13 @@ def bulk_update_tags(self, request, **kwargs):
             if user_access_level and access_level_satisfied_for_resource(scope_object, user_access_level, "editor"):
                 editable_objects.append(obj)
             else:
-                errors.append({"id": obj.id, "reason": "Permission denied"})
+                errors.append({"id": obj.id, "reason": BULK_UPDATE_TAGS_SKIPPED_REASON})
```
haacked: `suggestion`: The unified reason doesn't close the leak, because the shape of `skipped` still separates the two cases. This branch appends one entry per object the queryset returned, while line 386 appends one per requested ID, and `ids` is never deduplicated. So `{"ids": [7, 7], "action": "add", "tags": ["x"]}` comes back with one skipped entry when object 7 exists and you can't edit it, and two when it doesn't exist. Ordering leaks the same bit without the duplicate trick, since every found-but-inaccessible entry is appended before any missing one: post `[9999999, 42]` and the position of 42 in the response tells you whether it exists.

Worth calibrating before you spend much on this. The same bit is already reachable from a single GET, which returns 403 for a restricted object and 404 for a missing one, and nothing here crosses a team boundary. So it's not new attacker capability, it's that the comment on line 189 promises a property the code doesn't have, in shared code nine other products inherit. Building the list once from the request IDs makes the promise true.

```python
        editable_ids = {obj.id for obj in editable_objects}
        errors: list[dict[str, Any]] = [
            {"id": obj_id, "reason": BULK_UPDATE_TAGS_SKIPPED_REASON}
            for obj_id in dict.fromkeys(validated_ids)
            if obj_id not in editable_ids
        ]
```

That replaces both append sites and drops the `found_ids` set. `dict.fromkeys` preserves request order and works for the UUID variant too. The two existing assertions in `test_tagged_item.py` use single non-duplicated IDs, so they still pass.

products/experiments/backend/test/test_presentation_api.py
```diff
@@ -9036,6 +9036,138 @@ def test_api_schema_exposes_every_runtime_field(self) -> None:
...
+        response = self.client.post(
+            f"/api/projects/{self.team.id}/experiments/bulk_update_tags/",
+            {"ids": [experiment["id"]], "action": "set", "tags": ["via-pat"]},
+            format="json",
+            headers={"authorization": f"Bearer {personal_api_key}"},
+        )
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json()["updated"] == [{"id": experiment["id"], "tags": ["via-pat"]}]
+
+    def test_matching_ids_applies_list_filters(self):
```
haacked: `suggestion`: Nothing here would notice `matching_ids` going back to two queries per experiment. The fix rests entirely on `only()` and `select_related()` lining up, so dropping the `select_related`, narrowing `only()`, or reading one more field inside the access loop leaves the endpoint correct and just makes it slow again. Both `matching_ids` tests assert response contents only, and the endpoint is unpaginated, so the regression scales with the project.

[`test_listing_experiments_with_related_objects_is_not_nplus1`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/experiments/backend/test/test_presentation_api.py#L469) is the pattern to copy: one experiment versus five, `CaptureQueriesContext` around each call, and assert the count doesn't grow with the row count.

products/experiments/backend/test/test_presentation_api.py
```diff
@@ -9036,6 +9036,138 @@ def test_api_schema_exposes_every_runtime_field(self) -> None:
...
+            f"/api/projects/{self.team.id}/experiments/{experiment['id']}/",
+            {"tags": []},
+            format="json",
+        )
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json()["tags"] == []
+        detail = self.client.get(f"/api/projects/{self.team.id}/experiments/{experiment['id']}/").json()
+        assert detail["tags"] == []
+
+    @parameterized.expand(
```
haacked: `suggestion`: The `?tags=5` crash you just fixed can come straight back without failing anything. Every case in this list passes a well-formed array, so the `isinstance(tags, list)` guard in `_parse_tag_names` is never exercised. Delete that guard and `?tags=5` raises a `TypeError` again while `?tags="growth"` silently filters on the string's six characters, and the suite stays green.

It's four rows on the list you already have.

```python
            ("tags_and_excluded", 'tags=["shared"]&excluded_tags=["growth"]', {"B"}),
            ("tags_scalar_number_ignored", "tags=5", {"A", "B", "C"}),
            ("tags_scalar_string_ignored", 'tags="growth"', {"A", "B", "C"}),
            ("tags_object_ignored", 'tags={"a":1}', {"A", "B", "C"}),
            ("excluded_tags_scalar_ignored", "excluded_tags=5", {"A", "B", "C"}),
```

frontend/src/lib/lemon-ui/LemonTable/useBulkSelection.ts
```diff
@@ -75,12 +79,23 @@ export function useBulkSelection<T extends Record<string, any>, K extends BulkSe
...
     isRowSelectable,
     initialSelectedKeys,
+    clearSelectionKey,
 }: UseBulkSelectionParams<T, K>): UseBulkSelectionResult<T, K> {
     const [selectedKeys, setSelectedKeysState] = useState<K[]>(() =>
         initialSelectedKeys ? Array.from(new Set(initialSelectedKeys)) : []
     )
     const previouslyCheckedIndexRef = useRef<number | null>(null)
 
+    const isFirstClearKeyRender = useRef(true)
```
haacked: `suggestion`: This mount guard has no test, and dropping it breaks every table that seeds `initialSelectedKeys` by wiping the selection on first paint, silently. The clearing behavior it guards has no test either. `useBulkSelection.test.ts` has 17 cases covering every other branch, including `ignores changes to initialSelectedKeys after mount`, which already has the `rerender` shape to mirror.

Two cases cover both directions: the key changing clears, and the key holding steady across a rerender preserves. The second one matters as much as the first, since a clear that fires when the key is unchanged would drop the selection on every pagination click, which is what excluding `page` from `filterIdentity` exists to prevent.

```ts
it('clears the selection when clearSelectionKey changes', () => {
    const { result, rerender } = renderHook(
        ({ key }) => useBulkSelection({ pageRecords: PAGE, getKey, clearSelectionKey: key }),
        { initialProps: { key: 'a' } }
    )
    act(() => result.current.toggleRow(2, 1))
    expect(result.current.selectedKeys).toEqual([2])
    rerender({ key: 'b' })
    expect(result.current.selectedKeys).toEqual([])
})

it('keeps initialSelectedKeys on mount when clearSelectionKey is set', () => {
    const { result } = renderHook(() =>
        useBulkSelection({ pageRecords: PAGE, getKey, initialSelectedKeys: [1, 2], clearSelectionKey: 'a' })
    )
    expect(result.current.selectedKeys).toEqual([1, 2])
})
```

products/experiments/backend/test/test_presentation_api.py
```diff
@@ -9036,6 +9036,138 @@ def test_api_schema_exposes_every_runtime_field(self) -> None:
...
+        assert response.json()["updated"] == [{"id": experiment["id"], "tags": ["via-pat"]}]
+
+    def test_matching_ids_applies_list_filters(self):
+        tagged = self._create_experiment("Tagged", "tags-matching-flag-a", tags=["growth"])
+        self._create_experiment("Untagged", "tags-matching-flag-b")
+        response = self.client.get(f'/api/projects/{self.team.id}/experiments/matching_ids/?tags=["growth"]')
+        assert response.status_code == status.HTTP_200_OK, response.json()
+        assert response.json() == {"ids": [tagged["id"]], "total": 1}
+
+    def test_matching_ids_excludes_view_only_experiments(self):
```
haacked: `suggestion`: The editor check in `bulk_update_tags` has no test anywhere in the repo, so if it broke or inverted, a viewer could retag objects across all ten resources sharing the mixin and nothing would fail. I checked the flags, event-definition, and experiment bulk-tag suites. Both skipped-reason assertions in `test_tagged_item.py` reach the missing-ID branch instead: one uses a nonexistent ID, the other another team's dashboard, which the team-scoped queryset drops before the access check runs.

That file can't host the test, because it's an `APIBaseTest` and `access_controls_supported` is false unlicensed, so the precheck short-circuits. This test right here is licensed and already builds the exact fixture, another user's experiment plus an object-level `viewer` row, so adding a test right next to it is cheap: assert the view-only experiment comes back in `skipped` with its tags unchanged.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -553,6 +622,81 @@ const ExperimentsTable = ({
...
+                                            loading={matchingExperimentIdsLoading}
+                                            onClick={async () => {
+                                                setMatchingExperimentIdsLoading(true)
+                                                try {
+                                                    const { limit, offset, ...restFilters } = paramsFromFilters
+                                                    const response = (await api.get(
+                                                        `api/projects/${currentProjectId}/experiments/matching_ids/?${toParams(restFilters)}`
+                                                    )) as { ids: number[]; total: number }
+                                                    setMatchingExperimentIds(response.ids)
+                                                    ctx.setSelectedKeys(response.ids)
```
haacked: `question`: What should happen when a project has more than 500 experiments and someone clicks "Select all N matching"? `matching_ids` is unpaginated and uncapped, so it returns every editable ID, this line puts all of them in the selection, and [`BulkUpdateTagsButton.tsx:54`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/frontend/src/lib/components/BulkActions/BulkUpdateTagsButton.tsx#L54) posts the lot. The shared serializer caps `ids` at 500, so the request 400s and the user gets the generic "Failed to update tags" toast with nothing about a limit. The two features are sized against different numbers: the banner appears once the count passes one page of 100.

Flags has the same limitation today, so this isn't something you introduced. Capping `matching_ids` at 500, chunking the bulk request, or just naming the limit in the toast would each close it, if you want to close it here rather than in a follow-up.

posthog/api/tagged_item.py
```diff
@@ -244,7 +250,7 @@ class BulkUpdateTagsUUIDItemSerializer(serializers.Serializer):
 
 class BulkUpdateTagsUUIDErrorSerializer(serializers.Serializer):
     id = serializers.UUIDField(help_text="UUID of the object that was skipped.")
-    reason = serializers.CharField(help_text="Why the object was skipped, e.g. 'Not found'.")
+    reason = serializers.CharField(help_text="Why the object was skipped, e.g. 'Not found or no edit access'.")
```
haacked: `suggestion`: This help text now describes a response that event definitions doesn't send. The string flowed into the generated clients, including [products/event_definitions/frontend/generated/api.schemas.ts:217](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/products/event_definitions/frontend/generated/api.schemas.ts#L217) and [services/mcp/src/api/generated.ts:12882](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/services/mcp/src/api/generated.ts#L12882), but [`posthog/api/event_definition.py:494`](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/posthog/api/event_definition.py#L494) overrides `bulk_update_tags` and still returns the literal `"Not found"`. That override is deliberate, since it has no per-object editor check, so the schema is the side that's wrong. [test_event_definition.py:541](https://github.com/PostHog/posthog/blob/b92ecc30958198110f545b4bb403b820ffc8a211/posthog/api/test/test_event_definition.py#L541) pins the old string and still passes, which is how it stayed hidden.

Either give this serializer's `reason` its own help text describing what the UUID path actually returns, or have the override use the shared constant. The second only works if returning "or no edit access" from an endpoint that never checks edit access reads acceptably to you.

posthog/api/tagged_item.py
```diff
@@ -198,8 +203,9 @@ class BulkUpdateTagsRequestSerializer(serializers.Serializer):
         help_text="'add' merges with existing tags, 'remove' deletes specific tags, 'set' replaces all tags.",
     )
     tags = serializers.ListField(
-        child=serializers.CharField(),
-        help_text="Tag names to add, remove, or set.",
+        child=serializers.CharField(max_length=TAG_NAME_MAX_LENGTH),
+        max_length=BULK_UPDATE_TAGS_MAX_TAGS,
+        help_text="Tag names to add, remove, or set (up to 100 per request, 255 characters each).",
```
haacked: `nit`: These two numbers are spelled out two lines below the constants that enforce them, and this help text is codegen input baked into 16 generated schema and zod files. Bumping a cap would ship a description that contradicts the validator.

```suggestion
        help_text=(
            f"Tag names to add, remove, or set (up to {BULK_UPDATE_TAGS_MAX_TAGS} per request, "
            f"{TAG_NAME_MAX_LENGTH} characters each)."
        ),
```

products/experiments/backend/presentation/views.py
```diff
@@ -418,6 +458,45 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
+        )
+
+        if not self.user_access_control:
```
haacked: `suggestion`: This branch can't run, so `matching_ids` reads as though it has an unfiltered path for orgs without access control. It doesn't. `user_access_control` is a [`cached_property` that always returns a `UserAccessControl`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/api/routing.py#L639), and that class defines no `__bool__` or `__len__`, so the negation is always false.

CI's patch-coverage report says the same thing from the other direction: line 487 is the only uncovered changed line in this file.

Unindenting the `else` body drops the false signal:

```python
        experiments = list(queryset)
        self.user_access_control.preload_object_access_controls(cast(list, experiments))
        editable_ids = [
            experiment.id
            for experiment in experiments
            if (user_access_level := self.user_access_control.get_user_access_level(experiment))
            and access_level_satisfied_for_resource("experiment", user_access_level, "editor")
        ]
```

Feature flags carries the same dead branch, so whether to fix it there too is your call.

products/experiments/backend/presentation/views.py
```diff
@@ -418,6 +458,45 @@ def safely_get_queryset(self, queryset) -> QuerySet:
...
+        parameters=EXPERIMENT_LIST_FILTER_PARAMETERS,
+    )
+    @action(methods=["GET"], detail=False, pagination_class=None, required_scopes=["experiment:read"])
+    def matching_ids(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        """
+        Get IDs of all experiments matching the current list filters.
+        Accepts the same query params as the list endpoint and returns only
+        IDs of experiments the user has permission to edit.
+        """
+        service = ExperimentService(team=self.team, user=request.user)
```
haacked: `suggestion`: The list filters run twice on every `matching_ids` request. `self.get_queryset()` on the next line routes through [`TeamAndOrgViewSetMixin.get_queryset`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/api/routing.py#L324), which calls `safely_get_queryset`, which already hands the queryset to `filter_experiments_queryset` with `action="matching_ids"`. This call then runs it a second time with `action="list"`.

The `search`, `order`, and `deleted` handling sits outside the `if action == "list"` block, so `?search=foo` ends up carrying the same `name ILIKE` predicate twice. Results match today because Django overwrites a duplicate annotation alias rather than raising, which I confirmed against the pinned Django 5.2.17.

What it costs is legibility. Someone reading `matching_ids` has no way to tell which filters already ran on the queryset they were handed, and any step added to `filter_experiments_queryset` later that isn't safe to run twice will change this endpoint without touching it.

Teaching the service that `matching_ids` is a list-shaped action keeps the filtering in one place:

```python
# experiment_service.py
-        if action == "list":
+        if action in ("list", "matching_ids"):
```

This block then becomes `queryset = self.get_queryset()`, which also drops the `action="list"` string that currently has to misreport the action to get the right filters.

posthog/models/tagged_item.py
```diff
@@ -107,6 +108,13 @@ class TaggedItem(ModelActivityMixin, UUIDTModel):
...
+    experiment = models.ForeignKey(
+        "experiments.Experiment",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="tagged_items",
+    )
 
     class Meta:
         unique_together = ("tag", *RELATED_OBJECTS)
```
haacked: `suggestion`: This constraint can't reject a duplicate at either layer, so migrations 1294 and 1295 build and promote a twelve-column index that never fires.

Postgres treats a row as distinct from every other row when any indexed column is NULL, and every `posthog_taggeditem` row has ten NULL foreign keys by design. Django skips the check too: `save()` calls `full_clean()`, and `_perform_unique_checks` abandons a `unique_together` check as soon as one of its fields is None. I confirmed both against the pinned Django 5.2.17.

The repo already documents why. [`build_partial_uniqueness_constraint`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/models/utils.py#L486) says "All permutations of null columns must be explicit as Postgres ignores uniqueness across null columns", which is exactly what the eleven partial `unique_<field>_tagged_item` constraints are for. `unique_experiment_tagged_item` in 1294 is the one doing real work here.

Dropping it rather than widening it would shrink this PR and leave the table one fewer index to maintain on every insert: delete this line, the second `AlterUniqueTogether` in 1293, the third `RunSQL` in 1294, and 1295 entirely. Nothing outside `posthog/migrations/` references the constraint by name.

Worth saying that this diverges from the endpoint precedent in 1179 and touches schema every `TaggedItem` consumer sits on, so it's a bigger call than the rest of the PR and reasonable to leave for later.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -553,6 +622,81 @@ const ExperimentsTable = ({
...
+                                                } catch {
+                                                    lemonToast.error(
+                                                        "Couldn't select all matching experiments. Please try again."
+                                                    )
+                                                } finally {
+                                                    setMatchingExperimentIdsLoading(false)
+                                                }
+                                            }}
+                                        >
+                                            Select all {count} matching experiments
```
haacked: `blocking`: In a project with more than 500 experiments, this builds a selection the server refuses, and the only thing the user sees is "Failed to update tags".

[`matching_ids`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/products/experiments/backend/presentation/views.py#L498) returns every matching ID with no cap and no pagination, that whole array becomes the selection, and [`BulkUpdateTagsForm` posts all of them in one request](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/frontend/src/lib/components/BulkActions/BulkUpdateTagsForm.tsx#L58). [The `ids` field caps at 500](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/api/tagged_item.py#L198), so the request comes back 400 and nothing is tagged.

The threshold shape is what makes it awkward. The button only appears once 100 rows are selected, so 100 to 500 works fine, and anything above 500 fails wholesale with no partial result and no hint about the cap.

This is the answer to the question I left here last round. Batching inside `submit` fixes it, and since the form is shared it closes the same gap for feature flags:

```typescript
const BULK_UPDATE_TAGS_MAX_IDS = 500
const ids = Array.from(selectedIds)
const updated: BulkUpdateTagsResult['updated'] = []
const skipped: BulkUpdateTagsResult['skipped'] = []
for (let i = 0; i < ids.length; i += BULK_UPDATE_TAGS_MAX_IDS) {
    const batch = (await api.create(`api/projects/${currentProjectId}/${resource}/bulk_update_tags/`, {
        ids: ids.slice(i, i + BULK_UPDATE_TAGS_MAX_IDS),
        action: tagAction,
        tags: selectedTags,
    })) as BulkUpdateTagsResult
    updated.push(...batch.updated)
    skipped.push(...batch.skipped)
}
```

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -553,6 +622,81 @@ const ExperimentsTable = ({
...
+                                        AccessControlResourceType.Experiment,
+                                        experiment.user_access_level,
+                                        AccessControlLevel.Editor
+                                    )
+                                  ? true
+                                  : { disabledReason: "You don't have permission to edit this experiment." },
+                        rowAriaLabel: (experiment: Experiment) => `Select experiment ${experiment.name}`,
+                        headerAriaLabel: 'Select all experiments on this page',
+                        noun: ['experiment', 'experiments'],
+                        renderActions: (ctx) => {
```
haacked: `suggestion`: The select-all-matching flow now exists twice, and the two copies have already drifted apart during this PR's own review.

This block and [the flags equivalent](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/frontend/src/scenes/feature-flags/FeatureFlags.tsx#L612) share the same two `useState` hooks, the same `isAllMatchingSelected` comparison, the same banner condition, and the same fetch body. The real differences are the endpoint path, the per-page constant, and the noun.

Two fixes from this review landed here only. The toast on a failed `matching_ids` request is at line 676 and flags still has a bare `try`/`finally`, and `clearSelectionKey` is passed here but nowhere in the flags config, so filter changes still leave a stale selection there.

Two instances is below the bar where I'd normally push for an abstraction, so this is your call. The thing that tips it is that the shared part is the piece you least want drifting, and the PR description lists bulk delete and copy-to-project as follow-ups that would deepen the parallel. A hook next to `useBulkSelection` owning the fetch, the loading state, the error toast, and the reset would let both scenes keep only their banner copy and buttons.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -553,6 +622,81 @@ const ExperimentsTable = ({
...
+                                ctx.selectedCount >= EXPERIMENTS_PER_PAGE &&
+                                count > ctx.selectedCount
+                            return (
+                                <>
+                                    {isAllMatchingSelected && (
+                                        <span className="text-muted text-sm">
+                                            All {ctx.selectedCount} matching experiments selected
+                                        </span>
+                                    )}
+                                    {showSelectAllMatchingBanner && (
```
haacked: `suggestion`: Activating "Select all N matching experiments" with the keyboard drops focus to the top of the page, and a screen reader user gets no confirmation that anything happened.

The button renders only while `showSelectAllMatchingBanner` is true. A successful fetch flips `isAllMatchingSelected`, that condition goes false, and the focused button leaves the DOM. Browsers move focus to `<body>` when that happens, so the next Tab restarts from the beginning of the document. The replacement text appears in a fresh element, so nothing announces it either.

Always rendering the status text as a live region fixes the announcement, since the region is in the DOM before its content changes:

```diff
-                                    {isAllMatchingSelected && (
-                                        <span className="text-muted text-sm">
-                                            All {ctx.selectedCount} matching experiments selected
-                                        </span>
-                                    )}
+                                    <span className="text-muted text-sm" role="status">
+                                        {isAllMatchingSelected
+                                            ? `All ${ctx.selectedCount} matching experiments selected`
+                                            : ''}
+                                    </span>
```

For the focus, moving it to the bar's "Clear" button after `ctx.setSelectedKeys(response.ids)` keeps the keyboard user where they were working. Flags has the identical structure if you want them to stay in step.

posthog/api/tagged_item.py
```diff
@@ -198,8 +203,9 @@ class BulkUpdateTagsRequestSerializer(serializers.Serializer):
         help_text="'add' merges with existing tags, 'remove' deletes specific tags, 'set' replaces all tags.",
     )
     tags = serializers.ListField(
-        child=serializers.CharField(),
-        help_text="Tag names to add, remove, or set.",
+        child=serializers.CharField(max_length=TAG_NAME_MAX_LENGTH),
```
haacked: `suggestion`: The 255-character cap on individual tag names has no test, so dropping it turns a clean 400 back into a 500 with the suite still green. Only the 100-tag cap on the line below is covered.

Without the `max_length`, a 256-character tag reaches `Tag.objects.get_or_create`, where `Tag.name` is `varchar(255)`, and Postgres raises. [`test_too_many_tags_returns_400`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/api/test/test_tagged_item.py#L410) is the same shape: the payload becomes `{"ids": [dashboard.id], "action": "set", "tags": ["x" * 256]}` with the same `attr == "tags"` assertion.

products/experiments/backend/presentation/serializers.py
```diff
@@ -229,6 +232,12 @@ class ExperimentBaseSerializer(UserAccessControlSerializerMixin, serializers.Mod
...
```
haacked: `suggestion`: These caps have no test, so they can be dropped from experiment create and update without anything failing. [`test_too_many_tags_returns_400`](https://github.com/PostHog/posthog/blob/7d24cb1add886ae072e1b7b697715c9b774bc36a/posthog/api/test/test_tagged_item.py#L410) covers the bulk request serializer, which is a different field on a different path.

A create carrying 101 tags and expecting a 400 mirrors it, and `TestExperimentTags._create_experiment` already has the setup.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -553,6 +622,81 @@ const ExperimentsTable = ({
...
+                                            type="secondary"
+                                            size="small"
+                                            loading={matchingExperimentIdsLoading}
+                                            onClick={async () => {
+                                                setMatchingExperimentIdsLoading(true)
+                                                try {
+                                                    const { limit, offset, ...restFilters } = paramsFromFilters
+                                                    const response = (await api.get(
+                                                        `api/projects/${currentProjectId}/experiments/matching_ids/?${toParams(restFilters)}`
+                                                    )) as { ids: number[]; total: number }
```
haacked: `nit`: This PR generates `ExperimentMatchingIdsResponseApi` with exactly these two fields, so importing it from `products/experiments/frontend/generated/api.schemas` drops the hand-written shape.

products/experiments/backend/experiment_service.py
```diff
@@ -131,6 +132,22 @@ class CleanupRepositoryTarget(TypedDict):
...
+
+def _parse_tag_names(value: Any) -> list[str]:
+    """Parse a tags query param that arrives as a list or a JSON-encoded string.
+
+    Anything that doesn't decode to a list is ignored rather than an error: a scalar like
+    ``?tags=5`` or ``?tags="growth"`` decodes fine but isn't a tag list.
+    """
+    try:
+        tags = value if isinstance(value, list) else json.loads(value) if isinstance(value, str) else []
+    except (json.JSONDecodeError, TypeError):
```
haacked: `nit`: `TypeError` can't be raised here, since `json.loads` is only reached when `value` has already passed `isinstance(value, str)` and a `str` argument gives `JSONDecodeError` for bad input. Leaving it in sends the next reader looking for a call that could raise it.

jurajmajerik (approved):

products/experiments/frontend/scenes/ExperimentsScene.tsx
```diff
@@ -543,6 +603,95 @@ const ExperimentsTable = ({
...
+                                                    setMatchingExperimentIds(response.ids)
+                                                    ctx.setSelectedKeys(response.ids)
+                                                } catch {
+                                                    lemonToast.error(
+                                                        "Couldn't select all matching experiments. Please try again."
+                                                    )
+                                                } finally {
+                                                    setMatchingExperimentIdsLoading(false)
+                                                }
+                                            }}
```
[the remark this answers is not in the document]
rubychilds: Fixed in a5efb63e95e — the click handler now captures the filter identity when the request starts and drops the response if filters changed while it was in flight, so a stale matching_ids result can no longer repopulate the selection the filter-change reset just cleared.

PR #74338, opened 2026-07-28
feat(conversations): open support panel on ticket created via /ticket

christiaan-ph:
## Problem

When someone files a support ticket through PostHog AI's `/ticket` command, we tell them the ticket exists and then leave them to find it. The success toast offered a "View" button, but that is easy to miss and disappears on its own. Nothing takes you to the thread you just created.

## Changes

Creating a ticket via `/ticket` now lands you in that ticket's thread in the support side panel. The success toast keeps its message and drops the now-redundant "View" button on this path.

The interesting part is the sequencing. Callers opt in with `view_created_ticket` and navigate themselves once their own success handling is finished, using the ticket recorded on `lastSubmittedConversationsTicket`. The submit listener deliberately does not open the panel, and there is a comment in the code saying so, because opening it there writes the `#panel=support` hash mid-submit. `supportRouterLogic` answers that hash by reopening the support form, which resets `lastSubmittedTicketId`. The caller then reads `null`, concludes the submit failed, and leaves its modal open with no confirmation message in the chat.

The `/feedback` handover is untouched: it still shows the toast with a "View" button and does not move you anywhere.

I tested the flow on my local stack with the conversations widget enabled: sent `/ticket` in PostHog AI, described an issue, submitted, and confirmed the support panel opens on the new ticket's thread, the toast has no "View" button, and the chat still gets its confirmation message. I also checked that `/feedback` still behaves as before.

Claude wrote the code and ran the automated checks:

- One new test in `supportLogic.test.ts`. It catches the failure above: with `supportRouterLogic` mounted, a submit that opts into viewing must leave `lastSubmittedTicketId` readable and must not navigate on its own. It fails against the first version of this change and passes now. No existing test covered it, because none of them mounted the router logic alongside a submit.
- Existing suites for regressions: `supportLogic.test.ts` (18 passing) and `sidepanelTicketsLogic.test.ts` (8 passing). The panel side of this already had coverage in `sidepanelTicketsLogic.test.ts`, which asserts the specific ticket thread opens when the toast "View" path fires, so I did not duplicate it.
- Lint, formatting, and `tsgo --noEmit` are clean for the changed files.
- Headless Playwright passes over the real flow in three states: a fresh tab, a tab with the support panel already open, and cancelling the modal instead of submitting. The cancel run confirms the panel stays closed when you back out.

There is no component render test for the three lines of wiring in `TicketPrompt`. That path is covered by the manual run above.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No docs change needed. This is in-product navigation behavior with no new setting or API.

## 🤖 Agent context

I drove this with Claude Code (Claude Fable 5). Skills invoked: `/writing-tests` before adding the test, and `/run-posthog` to bring up the dev stack for manual verification.

The first version put the panel-open inside `submitSupportTicket`, which looked tidier since the toast decision lives there too. It broke in a way unit tests did not catch: the modal stayed open and no confirmation appeared in the chat, while the panel did open on the right ticket. Claude reproduced it with a test that mounts `supportRouterLogic`, which is what turned the URL hash write into an `openSupportForm` dispatch and wiped the submitted ticket id. Moving the navigation to the caller matches the timing of the toast "View" button, which has always worked precisely because it fires after the submit settles.

We also considered hanging the navigation off `LemonModal`'s `onAfterClose` so the panel opens only after the modal's 250ms close animation finishes. Rejected: once the confirmation message is appended, `getTicketPromptData` stops rendering `TicketPrompt`, so the component can unmount before that callback fires and the navigation would be silently dropped. A guaranteed navigation is worth more than removing a sub-300ms crossfade.

Worth a reviewer's eye: `openSupportForm` resetting `lastSubmittedTicketId` means any code that opens the support panel during a submit corrupts success detection for both `TicketPrompt` and `FeedbackPrompt`, which both poll that value after awaiting. The new test guards this path. A sturdier design would have `submitSupportTicket` hand its result back to callers instead, which felt like too much for this PR.


christiaan-ph:
Note to self: this PR overlaps with #74173 in `TicketPrompt.tsx` (both change the `useValues(supportLogic)` destructure and `handleTicketCreated`) and in `supportLogic.ts` (both add to connect and selectors). Whichever lands second needs a rebase, and the conflict resolution in `handleTicketCreated` must keep both behaviors: append the response-time confirmation message and open the support panel. Don't let one silently drop the other.

PR #74348, opened 2026-07-28
chore(ci): parameterize flaky report runners

rnegron:
## Problem

- The weekly flaky report hardcoded pytest across request construction and candidate selection, blocking runner extensions.

## Changes

- Pass an explicit runner through URLs and candidate filtering.
- Keep the configured runner list pytest-only.
- Fetch each configured runner into an isolated candidate pool.

## How did you test this code?

- Node test: 10 passing cases guard server-side runner filtering, explicit selection, and isolated pools, run under both a bare shell and a CI-like environment.
- Rebased onto the merged Trunk work: `selectReportCandidates` no longer collapses clusters (parking has to run first), and the Trunk lookup takes the runner.
- Oxfmt check passed.
- `hogli ci:preflight --fix` completed without failures.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

- No documentation change because observable behavior remains pytest-only.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

- Codex used `/wt`, `/writing-tests`, `/writing-code-comments`, `/hogli`, `/running-ci-preflight`, `/code-review`, and `/simplify`.
- The refactor is limited to seams required by the Jest follow-up.

PR #74349, opened 2026-07-28
feat(ci): add Jest to flaky report

rnegron:
## Problem

- The weekly Slack report omits Jest failures and cannot preserve separate runner limits.
- Trunk parking, added in [repair weekly flaky report](https://github.com/PostHog/posthog/pull/73961), is pytest-shaped. Trunk stores `parent` as the runner for pytest but as the **file path** for Jest, so a Jest lookup matched zero rows. Adding Jest without fixing that would rank Trunk-parked Jest tests as actionable, with a "park it" call to action, for a test CI already ignores.

> [!NOTE]
> Not hypothetical: `activityLogLogic.person.test.tsx` is Trunk-quarantined **and** failing in spans right now, so it would have landed in the first Jest report.

## Changes

- Fetch and rank pytest and Jest separately, with up to 10 rows each.
- Add runner labels and JavaScript-family ownership.
- Keep Logs rescue counts and links pytest-only (the failures view has no Jest rows).
- Extend the Trunk lookup to Jest, and collapse both quarantine systems into one status.

```mermaid
flowchart LR
    A{{Weekly report}} --> P[pytest request] --> PT[pytest top 10]
    A --> J[Jest request] --> JT[Jest top 10]
    A --> T[Trunk quarantine<br/>per runner] --> PK[(parked)]
    T --> PT
    T --> JT
    PT --> S[Slack table]
    JT --> S
    PK --> S
    class A phYellow;
    class P,J,T phBlue;
    class PT,JT phYellow;
    class PK phGray;
    class S phRed;
```

### Two simplifications this made possible

**One park status instead of two.** The quarantine file and Trunk both answer "is this failure already suppressed", but were handled separately, and disagreed: a quarantine-file test kept its top-10 slot while a Trunk one lost it. `testStatusFor` now answers once. The only status a row can still carry is `(Trunk flagged)`, which means Trunk marked it but CI still goes red, so it is still real work.

**Path matching is general, not a prefix list.** `selectorVariants` hardcoded a `products/<x>/` strip. Jest needs the opposite direction, since Trunk reports package-relative (`src/…`) while the report is repo-relative (`frontend/src/…`). It now generates every path suffix, which handles both directions in less code and needs no edit when a package is added.

- 14 Node tests, run under both a bare shell and a CI-like environment. New cases: a Jest selector from the package root matching a Trunk row, both quarantine systems parking while a masking-off Trunk flag stays ranked, and a parked test that would otherwise hide inside a collapsed cluster.
- Dry run with both runners against real Trunk rows pulled from the warehouse: Trunk queried once per runner, and the Jest test parked despite the `frontend/` prefix mismatch.
- Oxfmt, workflow lint, and `hogli ci:preflight --fix` clean.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

- Updated the engineering analytics README with runner limits and enrichment scope.

## 🤖 Agent context

- Codex wrote the original runner split. Claude rebased it onto the merged Trunk work and resolved the conflict, using the CI workflow, tests, copy, and code comment skills.
- Confirmed the Jest selector shape against live spans before designing the join. Jest selectors are test-grained (`file::full title`), which is what makes per-test parking possible; a file-grained selector would have forced all-or-nothing parking per file and a different design.
- Known approximation: Trunk tracks FOSS and EE variants separately, the report does not, so a test parked only in EE is treated as parked for both.


rnegron:
* **#74349** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74349?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74348** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74348?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

PR #74369, opened 2026-07-28
feat(web-analytics): add a dedicated runner for bot analytics tables

lricoy:
## Problem

The web analytics bot analytics tab issues its two tables — **Crawlers** and **Most crawled paths** — as raw `HogQLQuery` DataTables. In `system.query_log` they show up as anonymous `hogql_query` blobs: no `query_type`, no product/breakdown tagging, invisible to the per-strategy observability every other web analytics query has. Web analytics convention is a dedicated runner per query family.

## Changes

- **New `WebBotsTableQuery` query kind** (`breakdownBy: Crawler | Path`, optional `limit`) in the query schema, with codegen through `schema.json` / `posthog/schema.py`.
- **`WebBotsTableQueryRunner`** (`products/web_analytics/backend/hogql_queries/web_bots.py`): produces the same SQL as the previous inline HogQL — bot gate (`$virt_is_bot` + `$virt_bot_name`), the bot event set (`$pageview`/`$screen`/`$http_log`), date range and property/test-account filters via the standard `WebAnalyticsQueryRunner` helpers — tagged `query_type="web_bots_query"`, paginated with a default limit of 50.
- **Registration everywhere the exhaustiveness checks demand it**: runner dispatch, the web-analytics product branch in `query_tagging` (mypy's `assert_never` caught it), the runner query union, DataTable query features, type guard, insight metadata.
- **Frontend swap**: `botAnalyticsLogic.ts`'s two DataTable sources go from inline `HogQLQuery` SQL to the typed kind. The trends tiles on the tab already used `TrendsQuery` and are untouched.

Behavior is intended to be identical for users; the change is typing + observability.

- New CH-backed `test_web_bots.py` (3 tests + query snapshots): a Googlebot-UA event vs a human-UA event asserting only bot traffic appears with the right first column for both breakdowns — catches the two realistic regressions (dropping the bot gate → human traffic in bot tables; inverted breakdown dispatch), plus a date-range exclusion case. CH-backed because the regression lives in the virtual-property → HogQL-function → ClickHouse chain.
- Repo-wide `mypy --cache-fine-grained .` clean (its `assert_never` exhaustiveness caught a missing tagging branch — fixed).

## 🤖 Agent context

I directed this; Claude (Claude Code) implemented it after a query_log audit showed the bot tables as one of the two remaining runner-less web analytics query families. Design followed `WebExternalClicksTableQueryRunner` as the closest template; one query kind with a breakdown enum (mirroring `WebStatsTableQuery`) was chosen over two kinds. The other runner-less family — the live metrics dashboard's minute-bucket HogQL — is deliberately left as a follow-up. Skills invoked: `/writing-tests`.


arthurdedeus (approved):
nice and clean 🧹

PR #74373, opened 2026-07-28
feat(wizard): attribute wizard runs to the user who started them

rafaeelaudibert:
## Problem

When you open a project the setup wizard has run in, the FAB in the corner shows the run's progress. It always labels that run "On your machine", even when a teammate is the one who ran the wizard. On a shared org the first thing a brand-new member sees on login is someone else's run presented as their own, which is confusing.

The root cause: `WizardSession` had no user attribution at all. It's keyed only by team + workflow/skill, and `/latest` returns whatever the newest run for the *project* is. So every member's browser renders the same run, with no way to say whose it is.

## Changes

Attribute a run to the person who started it, end to end.

- Backend: `WizardSession` gains a `created_by` FK via the `CreatedMetaFields` mixin. It's overridden to `db_constraint=False` (posthog_user is a hot table, so the plain constrained FK would be blocked by `HotTableAlterPolicy` and take a lock on it) and `related_name="+"` to match this app's no-reverse-relation convention. It's populated from the authenticated user on create only, via `create_defaults`, so a later push for the same run never reattributes it.
- The initiator flows through the facade DTO, the serializer, and the generated frontend types.
- Frontend: the FAB's mode chip and dialog now read "On <name>'s machine" for a teammate's local run, and keep "On your machine" for the viewer's own (matched on email). Falls back to the initiator's email when they have no name, and to "On your machine" for older runs with no attribution.

> [!NOTE]
> Cloud runs still read "Cloud run" - the "whose machine" framing only applies to local runs. The completed-state detail subtitle ("finished its work on your machine") is left as-is; it contrasts local vs cloud rather than claiming ownership.

I couldn't attach a screenshot from this environment, but the `WizardSyncCard` story `LocalRunningStartedByTeammate` (and the `AllStates` canvas) render the new chip for visual review.

Automated tests only - I (Claude, via the PostHog Slack app) could not run the suite, codegen, or lint in this environment (no local Postgres and no `node_modules`), so CI is the real check. Tests I added:

- `test_logic.py::test_created_by_is_set_on_create_and_not_overwritten_on_update` - locks in the create-only semantics: a second push for the same session_id carrying a different user must not reattribute the run. Guards against someone moving `created_by_id` into `defaults`.
- `test_presentation.py::test_create_session_attributes_run_to_authenticated_user` - wiring guard that the viewset actually stamps `request.user` and the serializer exposes it.
- `installationProgressLogic.test.ts` - the initiator mapping (prefer first name, fall back to email, null when absent) and that the finished-run snapshot carries it so the handoff surface keeps the attribution after the live stream gates off.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Raised in a Slack thread: a new member logging into a shared org saw a teammate's wizard run labelled "On your machine", and the ask was to name the initiator instead. I (the PostHog Slack app, running on Claude) traced it to `WizardSession` having zero user attribution and implemented the fix across the stack.

Skills invoked while shaping this: `/django-migrations` (the migration is hand-written because there's no DB in this environment - it mirrors what `makemigrations` would emit for a nullable FK, with the swappable-user dependency), `/improving-drf-endpoints` (help_text on the new serializer field), `/writing-user-facing-copy` (the "On <name>'s machine" copy), and `/writing-tests` (each test above names the regression it catches).

A note for the reviewer: the generated `api.schemas.ts` change was written by hand to match Orval's output for a nullable nested dataclass (identical shape to the existing `UserBasicInfoApi` pattern), since I couldn't run `hogli build:openapi` here. Please re-run codegen to confirm no drift - I expect none beyond possibly component ordering.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1785272205733679?thread_ts=1785272205.733679&cid=C09GTQY5RLZ)*


products/wizard/backend/logic/sessions.py
```diff
@@ -35,29 +41,44 @@ def upsert_session(params: UpsertWizardSessionInput) -> tuple[WizardSessionDTO,
...
+        # A session belongs to the user who created it. A later push from a
+        # different user would overwrite the run data while `created_by` stays
+        # the original owner, so reject it. Legacy rows with a null
+        # `created_by_id` predate attribution and stay updatable by anyone.
+        if (
+            previous_session is not None
+            and previous_session.created_by_id is not None
+            and previous_session.created_by_id != params.created_by_id
+        ):
+            raise WizardSessionOwnershipError("This wizard session belongs to a different user and can't be updated.")
```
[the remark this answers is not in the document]
fercgomes: \+1

products/wizard/backend/logic/sessions.py
```diff
@@ -35,29 +41,44 @@ def upsert_session(params: UpsertWizardSessionInput) -> tuple[WizardSessionDTO,
             .first()
         )
         previous_run_phase = previous_session.run_phase if previous_session else None
+
+        # A session belongs to the user who created it. A later push from a
```
fercgomes, line 45: not sure we want this... but i guess its fine for now

products/wizard/backend/logic/sessions.py
```diff
@@ -79,12 +100,12 @@ def _enqueue_event_definition_sync(team_id: int, session_id: str) -> None:
...
 
 def get_session(team_id: int, session_id: str) -> WizardSessionDTO | None:
-    instance = WizardSession.objects.filter(team_id=team_id, session_id=session_id).first()
+    instance = WizardSession.objects.select_related("created_by").filter(team_id=team_id, session_id=session_id).first()
     return _to_dto(instance) if instance else None
 
 
 def get_latest_session(team_id: int, workflow_id: str, skill_id: str | None = None) -> WizardSessionDTO | None:
-    qs = WizardSession.objects.filter(team_id=team_id, workflow_id=workflow_id)
+    qs = WizardSession.objects.select_related("created_by").filter(team_id=team_id, workflow_id=workflow_id)
```
fercgomes, line 109: this is a bit weird...

we're adding the possibility of seeing a run created by someone else in the team, with the right attribution

but this line here makes it so you only return sessions you've created, so you'll never see sessions created by anyone else

fercgomes (dismissed):
i like the idea, but the get_latest_session created_by filter renders the entire feature useless

products/wizard/backend/logic/sessions.py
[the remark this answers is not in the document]
rafaeelaudibert: The amount of effort to make this happen is huge, I don't think it's worth the cost of adding a lock here, basically:

Someone has to create a session run with the same session ID at the same time to make this happen, it's so unlikely I don't think we need

fercgomes (approved):

rafaeelaudibert:
* **#74373** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74373?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74373?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

products/wizard/backend/logic/sessions.py
rafaeelaudibert: This is not filtering to the current user, this is simply selecting the user without an extra join

PR #74380, opened 2026-07-28
fix(aio): match Bedrock pricing by inference profile

### LINKED ISSUE #74366: US Bedrock inference-profile events use Global token rates
## Bug Description

I noticed that PostHog prices a US Bedrock inference-profile event using the Global-equivalent provider entry, even though the existing cost catalog already contains the correct US entry.

For this input:

```ts
findCostFromModel('us.anthropic.claude-sonnet-5', {
    $ai_provider: 'bedrock',
})
```

PostHog currently selects:

```text
amazon-bedrock-claude-on-aws
```

It should select:

```text
amazon-bedrock-us-east-1
```

The raw model ID contains the `us.` inference-profile prefix, but [`findCostFromModel`](https://github.com/PostHog/posthog/blob/97295889b4dfa4399c9f4fe2d877bde32156375d/nodejs/src/ingestion/pipelines/ai/costs/cost-model-matching.ts#L41-L60) resolves it to the canonical model before provider selection. The provider resolver then sees only the generic `bedrock` provider and [takes the first partial Amazon Bedrock match](https://github.com/PostHog/posthog/blob/97295889b4dfa4399c9f4fe2d877bde32156375d/nodejs/src/ingestion/pipelines/ai/costs/provider-matching.ts#L137-L175).

The catalog already has separate [Global and US provider rows](https://github.com/PostHog/posthog/blob/97295889b4dfa4399c9f4fe2d877bde32156375d/nodejs/src/ingestion/pipelines/ai/costs/providers/llm-costs.json#L1020-L1048). At the current Claude Sonnet 5 promotional rates, the difference is:

| Token category | Current selection | US Geo rate |
|---|---:|---:|
| Input | $2.00 / 1M | $2.20 / 1M |
| Output | $10.00 / 1M | $11.00 / 1M |
| Cache read | $0.20 / 1M | $0.22 / 1M |
| 5-minute cache write | $2.50 / 1M | $2.75 / 1M |
| 1-hour cache write | $4.00 / 1M | $4.40 / 1M |

The one-hour rate is [derived from the selected input price](https://github.com/PostHog/posthog/blob/97295889b4dfa4399c9f4fe2d877bde32156375d/nodejs/src/ingestion/pipelines/ai/costs/input-costs.ts#L187-L207), so the wrong provider selection affects that bucket as well.

For these token categories, selecting Global-equivalent rates understates the correct US charge by 9.09%.

## How to reproduce

1. Check out PostHog at `97295889b4dfa4399c9f4fe2d877bde32156375d`.
2. Run:

   ```ts
   const result = findCostFromModel('us.anthropic.claude-sonnet-5', {
       $ai_provider: 'bedrock',
   })

   console.log(result?.cost.provider)
   ```

3. Observe:

   ```text
   amazon-bedrock-claude-on-aws
   ```

4. Compare it with the `amazon-bedrock-us-east-1` entry already present for the same model.

## Additional context

AWS documents distinct [`us.anthropic...` and `global.anthropic...` inference-profile IDs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-5.html). Its [cross-Region inference documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html) describes Global inference as approximately 10% cheaper than Geographic inference, and the exact token-category rates are listed on the [Bedrock pricing page](https://aws.amazon.com/bedrock/pricing/). Anthropic also documents the input, output, cache-read, 5-minute-write, and 1-hour-write categories in its [pricing reference](https://platform.claude.com/docs/en/about-claude/pricing), while noting that Bedrock maintains its own regional pricing.

I think the matching rule should be:

- An explicit regional provider remains authoritative.
- With a generic Bedrock provider, a recognized inference-profile prefix can select the corresponding provider row when one exists.
- `global.` selects the Global row.
- Unprefixed models and prefixes without a corresponding row retain the current fallback.
- The matcher uses prices already present in the provider catalog rather than hardcoding model names or rates.

This is separate from #71306 and #71323. Those changes preserve and price Anthropic’s cache TTL breakdown after a provider price has been selected; this issue is about selecting the correct regional provider price in the first place.

## Debug info

```shell
Source-level reproduction against PostHog master at 97295889b4dfa4399c9f4fe2d877bde32156375d.
Synthetic model and provider values only; no project or customer data is required.
```

gouveags:
## Problem

Bedrock events can use a generic `$ai_provider` while the raw model contains routing information, such as `us.anthropic...` or a full inference-profile ARN.

After model lookup found the catalog row, provider resolution only used the generic Bedrock provider. It could therefore fall back to the first matching catalog entry instead of the existing regional entry. This can price a US inference profile using Global-equivalent rates.

Closes #74366

This is separate from #71323. That PR corrected Anthropic cache-write pricing after a provider price is selected. This PR corrects the provider selection itself.

## Changes

- Inspect the raw model when the provider is a generic Bedrock alias.
- Use the raw inference-profile prefix to select an exact catalog provider, or the only regional provider sharing that prefix.
- Use the AWS region from a full inference-profile ARN when the catalog has an exact regional provider.
- Keep an explicit provider authoritative.
- Preserve the current fallback for unknown prefixes, missing catalog entries, unprefixed models, and non-Bedrock providers.
- Apply the same resolution to both cost lookup paths.

The matcher does not hardcode model names or prices. It only selects among provider rows already present in the catalog.

> [!NOTE]
> A profile prefix changes selection only when the matched catalog row contains an exact provider key or one unambiguous regional match. Otherwise, the existing fallback remains unchanged.

- Ran the matcher and provider Jest suites: 2 suites and 131 tests passed.
- Added regressions for raw US and Global profile IDs, full inference-profile ARNs, explicit-provider precedence, missing regional rows, provider aliases, and non-Bedrock providers.
- Ran six runtime probes against the real catalog: raw US and Global IDs, an exact ARN region, a missing regional row, an explicit provider override, and a non-Bedrock provider.
- ESLint and Prettier passed for both changed files.
- `git diff --check` passed.

I did not run the full PostHog stack or the full monorepo test suite.

## Automatic notifications

- [x] Publish to changelog?

## Docs update

No docs update. This corrects provider selection for an existing event shape; it adds no fields, configuration, or workflow.

## 🤖 Agent context

I directed the work and used Codex sub-agents and GitHub CLI for research, regression-first tests, implementation, validation, and review. Skills invoked: `development-workflow`, `create-pr`, and the repository's `/writing-tests` skill. No public agent session link is available.

The implementation keeps the existing pricing catalog as the source of truth. It avoids model-specific pricing rules, preserves explicit-provider precedence, and retains the existing fallback when the routing information cannot identify a catalog provider safely.


gouveags:
@richardsolomou I found this while following the Bedrock pricing path from #71323. The fix keeps the catalog as the source of truth and only uses the raw profile ID when it maps safely to an existing provider row. Would you mind taking a look when you have time?

gouveags:
@greptileai review

gouveags:
@rafaeelaudibert The PostHog × Tractian meetup in São Paulo was a major catalyst for this contribution. It was great meeting you there, and everyone at MOVEdot is genuinely excited to start contributing back to PostHog. I hope this is the first of many!

rafaeelaudibert:
@gouveags loved meeting y'all there! We'd love to get your contributions in.

I can help routing requests to the right teams if GitHub doesn't do it automatically :)

richardsolomou (approved):

gouveags:
Can you guys help with the merge? @richardsolomou or @rafaeelaudibert

PR #74421, opened 2026-07-28
feat(replay): add per-person sessions tab to the player sidebar

ksvat:
## Problem

When you're watching a recording and want to see another session from the same user, today you have to leave the player, go to the recordings list, and filter by that person. That's a lot of clicks for a common move. This came out of session replay feedback asking for an easy in-player way to browse a user's other sessions.

## Changes

Adds a **Sessions** tab to the session replay player sidebar (next to Overview / Inspector / Network waterfall / Linked issues). It lists the current person's recordings, newest first, with timestamp and duration. Click any row to jump straight to that recording without leaving the player. The current recording is marked "Playing now", and there's a "Load older recordings" button that walks the cursor for people with long histories.

The tab reuses the existing sidebar tab machinery, so it gets URL sync (`?tab=sessions`) for free, and it only appears once we know who the person is.

It's gated behind the `replay-player-person-sessions-tab` feature flag (client-side, currently 0% rollout) so we can dogfood it in prod first.

> [!NOTE]
> No screenshot yet. The change is behind a flag at 0% rollout and I couldn't exercise it against real data locally (few recordings locally), so I haven't captured the UI in the running app. Happy to add one once it's enabled on a project with real multi-session users.

I (Claude) verified this at the type and unit level, not in a running browser:

- `pnpm --filter=@posthog/frontend typescript:check` passes clean.
- Added one Jest test for `mergeRecordingPage`, the one piece of non-trivial logic. It guards against "Load older" replacing the accumulated list instead of appending to it, which would silently drop earlier recordings. Selectors (`recordings`, `hasMore`) are trivial passthroughs and are covered as the observable outcome rather than as separate tests.
- Lint and format pass.

I did **not** manually test the tab in the app. Getting a person with multiple recordings requires prod data, which is the whole reason for the flag.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs needed while it's flag-gated and off.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Kim drove this, I (Claude) wrote the code.

The direction shifted mid-design. We first sketched prev/next chevrons in the player header, then Kim pivoted to a tab-with-a-list instead, which fit the existing sidebar tab system (`SessionRecordingSidebarTab` + `playerSidebarLogic`) much better and made pagination natural rather than an edge case. I anchored the sibling set on the person (via `person_uuid`), not on whatever filtered list you arrived from, since that's what the feedback actually asked for and it works even on deep-linked recordings.

Notable implementation decisions:
- The list logic keys on the person so hopping between their recordings reuses the cached list instead of refetching each jump.
- The fetch fires on both `loadRecordingMetaSuccess` and `afterMount`, because the person can resolve before or after the tab is opened.
- I extracted `mergeRecordingPage` as a pure function specifically so the load-more merge could be unit-tested cheaply, rather than standing up the whole player logic tree in a test.

Skills invoked while producing this: `/writing-tests` (gated the single test that earns its place), `/writing-user-facing-copy` (tab label and empty-state copy). The feature flag was created through the PostHog MCP.


arnohillen:
robot review:

1. The query never sets `date_from`, so the backend default of `-3d` applies. The tab only lists recordings from the last 3 days, "Load older recordings" stops there, and a current recording older than 3 days doesn't appear in the list. The person page playlist sets `date_from: '-30d'` ([sessionRecordingsPlaylistLogic.ts:221](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/session-recordings/playlist/sessionRecordingsPlaylistLogic.ts#L221)).
2. The logic key is `${playerKey}-${sessionRecordingId}`, but the comment and PR description say the list is keyed on the person and cached across jumps. As written it refetches on every jump. The open graphite thread flags the same thing.
3. The `snapshots.yml` change updates the baseline for `scenes-app-inbox-scene--self-driving-paused--dark`, which this PR doesn't touch.

Smaller ones: "This is the only recording for this person" also renders when the request failed or returned nothing. The flag owner comment says `@kim` but the author handle is `ksvat`. `distinct_ids` is passed on the GET query string alongside `person_uuid`, which can produce very long URLs for merged persons.

ksvat:
Thanks all — good catches. Addressed in `7b5e59f`:

**The logic key vs. the comment** (Graphite, stamphog, arno #2): real discrepancy. The comment and PR text claimed per-person caching, but the key is `sessionRecordingId`, so it refetches on each jump. `sessionPerson` isn't available synchronously when `key()` runs, so keying on the person has a timing hazard. I took the safe route: kept per-recording keying and corrected the comment to match reality. A jump is one cheap paged request, so the refetch is fine for now.

**`date_from` defaulting to -3d** (arno #1): fixed — now `-30d`, matching `sessionRecordingsPlaylistLogic`. The 3-day default would have hidden older sessions and dropped a current recording older than 3 days from its own list.

**Empty vs. error state** (arno smaller): the "only recording for this person" copy now only shows after a successful load. Failures get a distinct message plus a "Try again" button.

**`distinct_ids` on the query string** (arno smaller): capped at <100 like the playlist. `person_uuid` resolves the person server-side, so dropping distinct_ids for heavily merged persons is safe and avoids the huge URL.

**Flag owner handle** (arno smaller): fixed to `@ksvat`.

**`snapshots.yml` for the inbox scene** (arno #3): not from my code — that's the CI `chore(visual): update storybook baselines` commit auto-pushed to the branch (`4a678c2`). Left as-is.

arnohillen (commented):
Robot review:

Flag-gated, and the tab content only mounts when selected, so this adds no query to normal recording views. Four small things:

- Duplicate import from `~/queries/schema/schema-general` on two consecutive lines.
- `import type { PersonType } from '../../../../types'` sits two lines below a `~/types` import.
- `loadError` doesn't cover `loadMoreRecordingsFailure`, so a failed "Load older" does nothing visible.
- `DATE_FROM = '-30d'` means the tab can omit the recording you're currently watching, if that recording is older than 30 days.

The graphite comment about the logic key is stale: the key comment now documents the per-recording tradeoff explicitly rather than claiming to cache per person.

ksvat:
Good call on the retention window — addressed in `5a47927`:

**Window scoped to retention, not `-30d`** (your #4 + Kim's suggestion): the date window is now derived from the current recording's `retention_period_days` (from `sessionPlayerMetaData`), with a 30-day fallback when it's unknown and a one-day buffer to avoid edge-of-window rounding. Since a recording can't outlive its retention period, the current recording is always inside its own list now, and the window matches how far back sessions can actually exist per team rather than an arbitrary constant. Added a small parameterized test for the fallback and buffer.

**Failed "Load older" was silent** (#3): `loadMoreRecordingsFailure` now sets a distinct `loadMoreError`, shown as a message next to the button with the button switching to "Try again". Kept it separate from the initial-load error so a failed page-2 doesn't blow away the already-loaded list.

**Duplicate schema-general import** (#1) and **`PersonType` import placement** (#2): merged into one import and moved `PersonType` up with `~/types`.

PR #74432, opened 2026-07-28
fix(api): tighten URL host validation

Piccirello:
## Problem

The `app_urls` allowlist check and the redirect that follows it disagreed about which host a URL points at.

`unparsed_hostname_in_allowed_url_list` percent-decoded the URL before parsing it, so it read the host out of a string no client ever resolves. An `appUrl` whose authority terminates early once decoded could inherit an allowlist decision made for a completely different host.

Two endpoints in `posthog/api/user.py` echo that URL straight back as a 302 with credentials in the fragment: `toolbar_oauth_callback` (authorization code) and `redirect_to_site` (project api_token, user email, distinct id). `normalize_and_validate_app_url` made the split worse by running its scheme and loopback checks on the raw URL while delegating the allowlist decision to the decoded parse, so only the more permissive of the two parses was ever enforced.

## Changes

**Root cause, `posthog/api/utils.py`.** The blanket `unquote()` is gone. `canonicalize_encoded_url` decodes only when the raw string parses to no host at all, which is the case [#23504](https://github.com/PostHog/posthog/pull/23504) added it for: some browsers encode a whole redirect target, and `https%3A%2F%2Fpermitted.example` has no authority to validate. A URL that already parses to a host is judged exactly as written, so an encoded terminator can no longer smuggle in a second one. Both sinks then redirect to what `canonicalize_encoded_url` returned rather than to the caller's raw string.

**Userinfo, `posthog/api/utils.py`.** `strip_url_userinfo` rebuilds a URL so its authority holds only host and port, and both toolbar sinks run the validated URL through it before it becomes a redirect target. That is what makes the validated host and the emitted host provably equal on those two paths: the ambiguous part of the authority is not in the string the browser receives.

**Two rules instead of one, `posthog/security/url_validation.py`.** URLs we fetch ourselves and URLs we hand back to someone else need different strictness, so they now get different functions.

| | rule | callers |
| --- | --- | --- |
| `has_authority_bypass_chars` | `\` and `%5c` | the SSRF validator in this module, nothing else |
| `has_ambiguous_authority` | the above, plus `%2f` `%3f` `%23` `%40` anywhere in the authority | every sink that returns a URL to a client |

Percent-encoded characters are ordinary in credentials. An email username encodes `@` as `%40`, and a password may encode `/`, `?`, or `#`. Applying the strict rule to outbound fetches would reject routine basic auth on integrations and warehouse sources, so the fetch path keeps the lenient rule. A URL we hand to someone else has no business carrying credentials in the first place, which is what makes the strict rule safe there.

> [!NOTE]
> Three sinks get the strict rule that only had the backslash rule on master: the OAuth `redirect_uri` validator, `absolute_uri`, and the conversations widget restore URL. All three deliver the authority to a client intact and none of them strip userinfo the way the toolbar path does. The OAuth loopback branch is the sharpest case: `urlparse` reads `http://evil.example%2F@localhost:50470/callback` as `localhost`, so the port-flexibility reconstruction matched a registered `http://localhost/callback` and approved it.

Deliberately untouched: `hostname_in_allowed_url_list`, which a sibling PR is changing for an unrelated ReDoS issue.

**Merge with master.** One commit moves code and changes none of it: `has_ambiguous_authority`, the new `test_utils.py` class, and one import. Master added `strip_userinfo` and new allowlist tests at the positions this branch used, and git conflicts when both sides insert at one position. The moves are what let the merge apply cleanly.

Automated only. I (Claude) did not exercise this in a browser or against a running stack.

Every group below failed before the corresponding fix and passes after.

- `posthog/api/test/test_utils.py::TestAppUrlCanonicalization` (9 cases, new). Direct unit tests for the two URL helpers, which previously had only indirect coverage. The `canonicalize_encoded_url` cases pin that a URL with an authority is never decoded, which is the property that keeps the bypass closed, and that an unparseable URL comes back untouched instead of raising. The `strip_url_userinfo` cases pin that the split happens on the last `@`, matching how `urlparse` resolves the host: splitting on the first would emit a different host than the allowlist approved. On a `SimpleTestCase`, since the helpers are pure string functions and do not need the DB the rest of the file sets up.
- `posthog/security/test/test_url_validation.py::test_has_ambiguous_authority` (13 cases, new) and 3 cases added to `test_has_authority_bypass_chars`. The pair pins the split itself: basic auth with `%40` in the username, or `/` or `#` encoded in the password, passes the lenient rule and fails the strict one. Two cases assert an unparseable authority fails closed rather than raising, covering the branch CI reported as the only uncovered lines in the previous revision.
- `posthog/api/oauth/test_views.py::test_encoded_authority_terminator_in_localhost_redirect_rejected` (4 cases, new). Asserted against `OAuthValidator.validate_redirect_uri` rather than the `/authorize` response, because an unbuilt frontend also produces a non-200 there, so the existing `assertNotEqual(200)` style would have passed with the guard removed.
- `products/conversations/backend/api/tests/test_restore.py::test_restore_request_rejects_ambiguous_authority` (4 cases added, renamed from the backslash-only version) and `posthog/test/test_utils.py::test_absolute_uri_rejects_ambiguous_authority` (3 cases added, same rename). Each covers a sink that emits the authority as supplied.
- `posthog/security/test/test_url_validation.py::test_is_url_allowed_permits_percent_encoded_credentials` (4 cases, new) and one case added to `posthog/test/test_utils.py::test_format_absolute_url`. These guard the over-block direction, which the unit tests above do not reach. The first fails if the fetch path is switched to the strict rule, breaking every integration and warehouse source whose connection URL carries basic auth. The second fails if the authority extraction ever falls back to the whole string when there is no netloc, which would make `absolute_uri` reject relative paths with a reserved character encoded in the query, and it builds most of our emailed links. I verified both by mutating the source and confirming each fails, since a guard that cannot fail is not a guard.
- `posthog/api/test/test_toolbar_oauth_primitives.py::test_rejects_unparseable_app_url` (2 cases, new) and `posthog/api/test/test_user.py::test_redirect_to_site_rejects_an_unparseable_app_url` (2 cases, new). `https://[::1` and an NFKC-unstable host both raise `ValueError` out of `urlparse`. Reaching either sink with one of those returned a 500 rather than the typed 400 and 403. No existing test passed an app_url that fails to parse at all.
- Existing groups from earlier commits still pass: `test_unparsed_hostname_in_allowed_url_list`, `test_redirect_to_site_rejects_encoded_authority_terminator`, `test_redirect_to_site_accepts_a_fully_encoded_app_url`, and `test_rejects_ambiguous_authority`.

`ruff check` and `ruff format` are clean, and a repo-wide `mypy --cache-fine-grained .` came back clean.

Commands run:

```
PGPORT=5432 PYTEST_XDIST_WORKER=gw0 pytest \
  posthog/api/test/test_utils.py \
  posthog/security/test/test_url_validation.py \
  posthog/test/test_utils.py \
  posthog/api/test/test_toolbar_oauth_primitives.py \
  products/conversations/backend/api/tests/test_restore.py \
  posthog/api/test/test_user.py \
  posthog/api/oauth/test_views.py \
  posthog/api/test/test_toolbar_oauth_authorize.py \
  posthog/api/test/test_toolbar_hash_routing_repro.py
```

<details>
<summary>Pre-existing and environmental failures in my local run</summary>

11 failures from an unbuilt frontend (`TemplateDoesNotExist: index.html`), confirmed identical on a clean tree with `git stash`:

- `test_user.py::TestLoginViews::test_redirect_to_preflight_when_no_users`
- `posthog/api/oauth/test_views.py` (10, including both `TestLocalhostLoopbackRedirectUri` acceptance cases, which sit next to a guard this PR changes; verified the same pass/fail split stashed and unstashed)

Plus several ClickHouse `Table is in readonly mode` errors at teardown from a wedged local replica, also confirmed identical on a clean tree and unrelated to a change that only parses strings.

</details>

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing behavior change to document. Authorized URLs still work the way the settings page describes them.

## 🤖 Agent context

Claude Code (Opus) wrote this from a verified finding I handed it, working in an isolated worktree. Skills invoked: `/writing-tests` and `/writing-code-comments`.

The design took three passes and the intermediate states are worth knowing about, since two of them are wrong in instructive ways.

The first revision put the encoded-terminator rule inside `has_authority_bypass_chars` and scoped it to the authority. That over-blocked legitimate credentials on the SSRF path, which is what @MattBro caught.

The second revision split the rule into a second function and had each client-facing caller invoke both. That looked tidy and was a footgun: I wired the allowlist path and missed the OAuth `redirect_uri` validator, `absolute_uri`, and both conversations sinks, silently loosening three of them relative to the previous revision. The failing tests for the OAuth and conversations sinks in this PR are the ones that caught it.

The third and current revision folds both rules into `has_ambiguous_authority` so there is one call, not two to remember. `has_authority_bypass_chars` stays as the lenient primitive and is now called only from inside `url_validation.py`. That inventory is the invariant worth preserving: if a new caller outside that module reaches for `has_authority_bypass_chars`, it is almost certainly the wrong one.

Worth knowing while reviewing: the terminator rule is defense in depth on the toolbar path, where `strip_url_userinfo` already makes the validated host and the emitted host the same string. It is load-bearing on the other three sinks, which emit the authority as supplied. The docstrings say which is which so nobody reads a guarantee into the wrong one.

Thanks to @MattBro, whose review caught the missing `ValueError` guard and the basic-auth over-block. His third point, that three of the five call sites only parameterized backslash cases, is answered by the split rather than by adding cases everywhere: the lenient rule genuinely is backslash-only, and every sink on the strict rule now has terminator cases. His fourth point, that the description described only the first commit, is why this body was rewritten.


Piccirello:
### Code review

Found 1 issue:

1. The stated justification for inverting the `url encoded needle` test case does not hold. The body argues a singly-decoded needle can only look encoded if it was double-encoded, "which no real client does" — but the `unquote()` call being removed, and that exact test case, were both added by #23504 (`0fa085b8a4`) in response to a reported case of exactly that input shape reaching this check. Worth taking the tradeoff knowingly, or noting it explicitly, rather than on the premise that the input does not occur. If backward compatibility matters here, one option is to decode as a fallback only when the raw parse yields no host, re-running the authority checks against the decoded variant.

https://github.com/PostHog/posthog/blob/ac78a88bf8a57c926aa57c411c689fd20fd22b13/posthog/api/utils.py#L499-L506

https://github.com/PostHog/posthog/blob/ac78a88bf8a57c926aa57c411c689fd20fd22b13/posthog/api/test/test_utils.py#L319-L324

posthog/api/utils.py
```diff
@@ -496,12 +496,45 @@ def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs):
...
+    Some browsers encode a whole redirect target, turning ``https://permitted.example`` into
+    ``https%3A%2F%2Fpermitted.example`` — a string that parses to no host at all.
+
+    Only a string with no host is decoded, and only once. Decoding a URL that already has an
+    authority is what lets ``https://permitted.example%2F@evil.example/`` be approved as
+    ``permitted.example`` while a browser resolves ``evil.example``, so that case is left alone.
+    Callers must redirect to what this returns, not to what they passed in — approving the decoded
+    form and emitting the raw one would reintroduce the same split.
+    """
+    if urlparse(url).hostname:
```
MattBro: `urlparse` is unguarded here and now runs before the `try/except ValueError` that used to catch this, so `https://[::1` and NFKC-unstable hosts like `https://exa℀mple.com/` return a 500 where master returned a clean 400. `has_authority_bypass_chars` already wraps its own `urlparse` in `except ValueError: return True`; this needs the same, returning `url` unchanged so the block still happens.

posthog/security/url_validation.py
```diff
@@ -151,21 +154,32 @@ def _is_private_ip_literal(host: str) -> bool:
...
     if "\\" in url:
         return True
     if "%5c" in url.lower():
         return True
-    return False
+    try:
+        authority = urlparse.urlparse(url).netloc.lower()
+    except ValueError:
+        return True
+    return any(terminator in authority for terminator in ENCODED_AUTHORITY_TERMINATORS)
```
MattBro, line 182: Scanning the whole authority also rejects `https://user%40example.com:pass@host.example.com/path`, which is ordinary basic auth that every parser resolves to `host.example.com`. Nothing breaks today, but `clickhouse/source.py:128` ships that shape as a placeholder and this fires through `is_url_allowed`. Narrowing to `netloc.rpartition("@")[2]` fixes it, at the cost of `good.example%2F@evil.test` no longer being blocked here and relying on the allowlist rejecting `evil.test` instead.

MattBro (commented):
Two changes worth making, one verified regression and one behavior narrowing, both inline. I have the guard fix and a regression test ready to push here unless you would rather.

The description covers the first commit only. It says the unquote was dropped and that a case asserts False; the second commit reintroduces the decode conditionally and that test asserts True at HEAD, so anyone reviewing from the body gets the wrong threat model.

Smaller: neither new helper has a direct unit test, and three of the five `has_authority_bypass_chars` call sites parameterize only the backslash cases.

posthog/api/utils.py
Piccirello: Confirmed and fixed in f3ba731. Reproduced both shapes you named, against master and against 6223adc:

| input | master | 6223adc |
| --- | --- | --- |
| `https://[::1` | `ToolbarOAuthError` 400 | unhandled `ValueError: Invalid IPv6 URL` |
| `https://exa℀mple.com/` | `ToolbarOAuthError` 400 | unhandled `ValueError: netloc contains invalid characters under NFKC normalization` |

`canonicalize_encoded_url` now returns the URL untouched on `ValueError`, so the caller's own `urlparse` raises inside its `try` and produces the 400, exactly as you described. That also makes `unparsed_hostname_in_allowed_url_list` return `False` rather than raise, which is better than master, where it raised too.

One thing the guard alone did not fix. `redirect_to_site` has its own unguarded `urlparse` in the rejection branch, which predates this PR and 500s on master for the same inputs. Once the guard makes the allowlist return `False` instead of raising, control newly *reaches* that line, so the 500 moved rather than went away. Both are guarded now, with regression tests at each sink: `test_rejects_unparseable_app_url` and `test_redirect_to_site_rejects_an_unparseable_app_url`.

posthog/security/url_validation.py
Piccirello: Confirmed and fixed, though not in the shape you suggested.

The over-block is real: `has_authority_bypass_chars("https://user%40corp.example:pass@example.com/path")` returns `False` on master and `True` at 6223adc.

On the supporting citation, for the record rather than to argue the finding: `clickhouse/source.py` never routes through `is_url_allowed` (it does not import `url_validation` at all, nor does `clickhouse_cloud`), and the placeholder is `https://user:password@play.clickhouse.com:8443/default`, which carries no `%40` and so would not have been blocked even if it did. The finding stands on its own.

I did not take `netloc.rpartition("@")[2]`, for two reasons.

`%40` is not special. A password containing `/`, `?`, or `#` has to encode it too, so `https://user:p%2Fss@host/` is equally legitimate and was equally blocked. A carve-out keyed on which characters, or on which part of the authority, leaves the same trap for whoever hits it next.

And scoping to the host would have left a check in the shared primitive that catches nothing. With the decode gone and `strip_url_userinfo` applied, the allowlist already sees the host the browser resolves, so `good.example%2F@evil.test` is rejected by the allowlist regardless.

So the rule is split by what happens to the URL afterward instead. `has_authority_bypass_chars` keeps the lenient rule and is now called only from the SSRF validator inside `url_validation.py`. A new `has_ambiguous_authority` carries the strict rule for every sink that hands a URL back to a client. Credentials are ordinary in the first and have no business in the second, which is what makes each rule safe where it sits.

Pulling on this is what found the actual bugs. Three sinks deliver the authority to a client intact, do not strip userinfo, and had only the backslash rule on master: the OAuth `redirect_uri` validator, `absolute_uri`, and the conversations widget restore URL. The OAuth one is the sharpest, because `urlparse` reads `http://evil.example%2F@localhost:50470/callback` as `localhost`, so the RFC 8252 port-flexibility branch reconstructs `http://localhost/callback`, matches the registered URI, and approves it. I have a failing-first test for that one asserting against `OAuthValidator.validate_redirect_uri` directly, since the existing `assertNotEqual(200)` style passes spuriously when the frontend is unbuilt.

All three take the strict rule now. As with the toolbar path, this is a guard against a consumer that percent-decodes before splitting the authority rather than a live divergence in anything we ship, and the docstrings say so.

Piccirello:
All four points actioned, thanks. Please hold off on pushing the guard fix: it is superseded, and the second change landed in a different shape than your suggestion, so the two would conflict.

**The `ValueError` guard.** Confirmed exactly as you described and fixed. Reproduction table inline. The guard on its own left a second 500 in `redirect_to_site`, whose rejection branch has an unguarded `urlparse` of its own, so that one is guarded too.

**The `%40` narrowing.** Confirmed and fixed, but not via `rpartition`. Reasoning inline. Short version: `%40` is not special, since a password encoding `/`, `?`, or `#` is equally legitimate and was equally blocked, so any carve-out keyed on characters or on the host portion leaves the same trap. The rule is now split by what happens to the URL afterward rather than by where in the authority it sits. `has_authority_bypass_chars` keeps the lenient rule and is reachable only from the SSRF validator; a new `has_ambiguous_authority` carries the strict rule for sinks that hand a URL back to a client.

**The description.** Rewritten. You were right that it described the first commit only, and worse, the `NOTE` block was describing code that no longer existed.

**The test coverage.** Answered by the split rather than by adding cases everywhere. The lenient rule genuinely is backslash-only now, so backslash-only tests are correct at its call sites. Every sink on the strict rule has terminator cases, and both new helpers have direct unit tests. I also added the over-block direction, which nothing covered: that the fetch path keeps allowing basic auth, and that `absolute_uri` keeps allowing relative paths with a reserved character encoded in the query. Both mutation-verified, since a guard that cannot fail is not a guard.

Your second point is what surfaced the real bugs. Three sinks that hand a URL back to a client had only the backslash rule on master, and the OAuth `redirect_uri` loopback branch approved an authority that `urlparse` and a decoding client read differently. Worth a fresh look at that one in particular.

Two things still open and not in this PR. The branch is a long way behind master and conflicts in `posthog/security/url_validation.py`, the file this restructures, so the merge wants doing deliberately rather than as a drive-by. And nothing stops a future caller outside `url_validation.py` from reaching for `has_authority_bypass_chars` instead of the strict one, which is exactly the mistake I made mid-way through this. A semgrep rule would fit the convention ladder better than a test; I left it out to avoid widening a security PR.

feliperalmeida (approved):

PR #74434, opened 2026-07-28
fix(activity-log): validate detail_filters field paths

Piccirello:
## Problem

The advanced activity logs endpoints accept a `detail_filters` map where the caller supplies the field paths.
Those paths were used to build ORM lookup names with only a check for `__`, and the segment names themselves were never validated.
Django does not treat every segment of a path as a JSON key, so a path can end up selecting the ORM operation rather than just the data to compare.
The existing allowlist covers the `operation` field of the filter but not the field path, so it never sees this.

Separately, `_generate_indexed_paths` expands each `[]` marker in a path into five index permutations, recursively.
That is fine at the depth the UI produces, but the number of generated paths and `Q` objects grows as `5 ** markers`, and nothing bounded the marker count.
That expansion happens in Python before any SQL is emitted, so query-side limits do not apply to it.

## Changes

Added `validate_detail_filters` in `posthog/api/advanced_activity_logs/filters.py` and made it the single gate for `detail_filters`.
It runs at the top of `_apply_detail_filters`, and `AdvancedActivityLogFiltersSerializer` delegates to it too, so bad input is rejected at the API boundary instead of at export-render time.
It rejects:

- field paths whose segments collide with a Django lookup or transform name.
  The set is computed at import from `JSONField().get_lookups()` and `KeyTransform.get_lookups()` rather than hardcoded, so it stays right across Django upgrades.
- field paths containing `__`, empty segments, or non-string/oversized keys.
- field paths where `5 ** path.count("[]")` exceeds 125, which is the depth the endpoint already supports (`changes[].after[].field.subarray[].value`).
- filter configs that are not objects. Previously those raised `AttributeError` and returned a 500.
- more than 100 filter entries.

The array-index fan-out constant moved to a module-level `_MAX_ARRAY_INDICES_TO_CHECK` so the cap and the generator cannot drift apart.

> [!WARNING]
> Behavior change: invalid `detail_filters` now return a 400 instead of being silently dropped.
> Before this, a filter with `__` in the path or an unsupported operation was skipped, and the caller got an unfiltered result set with no signal that their filter did nothing.
> The UI only ever sends paths discovered by `available_filters` with the three supported operations, so it is unaffected.

Also corrected three `nosemgrep: orm-field-injection` justifications in this file.
They claimed the field path was validated, which was not true of the segment names.
They now name the validator and state what it actually enforces.
And fixed the `orm-field-injection` rule message itself, which told readers that anchoring a path on a JSONField mitigates the risk.
It does not, for the reason above, and that advice is roughly how this got shipped.

Automated only. I did not exercise this in a browser or against a real dataset.

```
PGPORT=5433 pytest posthog/api/advanced_activity_logs/ -q
PGPORT=5433 pytest posthog/api/test/test_activity_log.py -q
```

Both green (39 and 23 tests).

New tests, and the regression each catches:

- `TestDetailFilterValidation.test_rejects_unsafe_detail_filters` (parameterized, 6 cases) fails if a caller-supplied path segment can become the ORM operation, if array nesting can exceed the generated-path cap, or if a non-object filter config 500s.
  No existing test covered any of these.
  It is a `SimpleTestCase` because rejection happens before a query is built, so it needs no database.
- `test_serializer_rejects_unsafe_detail_filter_paths` is the wiring guard for the serializer path, so a refactor that drops the `validate_detail_filters` delegation is caught.

Replaced `test_rejects_dunder_field_paths` and `test_rejects_invalid_operations`.
They asserted the old silent-skip behavior, which this PR intentionally changes, and their cases are now rows in the parameterized test above.
That also drops two DB-backed tests in favor of no-DB ones.

Before applying the fix, I confirmed each new case failed for the right reason (`ValidationError not raised`, not an import or fixture error), and inspected the compiled SQL to confirm the old behavior was what the finding described rather than something incidental.

Two gaps a reviewer should know about.
I did not exercise the HTTP endpoint, so the hop from `is_valid(raise_exception=True)` through exceptions-hog to a 400 response is inferred from how every other serializer here behaves, not observed.
And the fan-out bound lives in `validate_detail_filters`, not in `_generate_indexed_paths`, so the generator is still unbounded if called directly.
`_apply_detail_filters` is its only production caller, so the boundary holds, but it is at the validator rather than at the generator.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover `detail_filters` key syntax, so nothing to update.
The serializer `help_text` already documents the allowed operations.

## 🤖 Agent context

I had Claude Code (Opus) do this one, working from two findings I handed it.
Skills invoked: `/writing-tests` and `/writing-code-comments`.

Worth knowing for review: the first fix it tried was structural rather than validation-based, appending an explicit terminal lookup to every generated path so the caller's last segment could never land in the lookup position.
A probe against the real ORM killed that.
Lookups registered on `KeyTransform` shadow key names at intermediate positions too, so the suffixed path raises `FieldError` instead.
That would have traded one bad outcome for a 500, so it was dropped entirely rather than kept as defense in depth.
Validation is the only mechanism here, which is also why the `nosemgrep` justifications now point at the validator by name.

Related, and not obvious from the diff: rejecting reserved segment names in every position rather than only the last also closes a latent 500.
An intermediate segment named after a `KeyTransform` lookup raises `FieldError` today.

The task description suggested capping array markers at one or two.
I went with three (125 generated paths), because `test_deeply_nested_array_fields` pins three as supported behavior and I did not want to silently drop a feature inside a hardening change.
Expressed as a path-count cap rather than a marker count, so it survives a change to how many array indices we check.


Piccirello:
### Code review

Found 1 issue:

1. The per-segment reserved-lookup check in `_validate_detail_filter` can be bypassed when a segment ends or begins with an underscore next to a dot, so a Django lookup name can still reach the query as an isolated path component.

`_validate_detail_filter` splits `field_path` on `.` (after stripping `[]` markers) and rejects any segment that exactly matches a name in `_DJANGO_LOOKUP_NAMES`. The path is later joined into Django's lookup syntax with `field_path.replace('.', '__')` (and equivalently in `_build_indexed_path`). Because that join runs after validation, an underscore at the edge of a segment can combine with the inserted `__` to shift where Django's separator falls once joined — so a path can pass the split-on-`.` check while still resolving, after the join, to a lookup name as its own isolated component in the final Django path. That's the same gap the updated `nosemgrep: orm-field-injection` justifications on lines 168, 382, and 385 assert is closed ("no segment named after a Django lookup").

https://github.com/PostHog/posthog/blob/072764f0ccda6e8c080d94d48c4fb6fe9011f403/posthog/api/advanced_activity_logs/filters.py#L72-L80

The join that creates the mismatch:

https://github.com/PostHog/posthog/blob/072764f0ccda6e8c080d94d48c4fb6fe9011f403/posthog/api/advanced_activity_logs/filters.py#L147-L151

posthog/api/advanced_activity_logs/filters.py
```diff
@@ -1,12 +1,95 @@
...
+_DJANGO_LOOKUP_NAMES = frozenset(JSONField().get_lookups()) | frozenset(KeyTransform.get_lookups())
+
+_MAX_DETAIL_FILTERS = 100
+_MAX_FIELD_PATH_LENGTH = 500
+# Array filters only look at the first _MAX_ARRAY_INDICES_TO_CHECK elements of each array. Each
+# `[]` marker in a field path multiplies the number of generated index permutations by that number,
+# and every permutation costs a path string plus a Q object, so capping the product keeps a deeply
+# nested path from exhausting the worker's memory before any SQL runs.
+_MAX_ARRAY_INDICES_TO_CHECK = 5
+_MAX_INDEXED_PATHS = 125
```
ReeceJones, line 28: nit: add a comment here saying that this is 5^3 and why that is significant. A bit of a magic number right now.

ReeceJones (approved):

posthog/api/advanced_activity_logs/filters.py
Piccirello: Good catch, done in 8efb87ae898. Folded it into the comment above: 125 is 5^3, so a path can nest three arrays and a fourth is rejected.

Kept it as a literal rather than `_MAX_ARRAY_INDICES_TO_CHECK ** 3` on purpose, since the cap is on generated paths rather than on depth. If we ever raise the per-array index count, the literal narrows the allowed depth automatically, whereas deriving it from the base would pin depth at three and let the fan-out grow with it (20^3 = 8000 Q objects per filter, times the 100 filter limit).

posthog/api/advanced_activity_logs/filters.py
```diff
@@ -1,12 +1,96 @@
...
+
+    # Check the components Django will actually see, not the dot-separated segments. A segment
+    # ending in `_` next to one starting with `_` joins into a run of underscores that shifts the
+    # `__` boundary, so a lookup name can surface as its own component without ever having been a
+    # segment -- and reach the database as a lookup.
+    for component in stripped.replace(".", "__").split("__"):
+        if not component:
+            raise serializers.ValidationError(
+                f"Detail filter field path '{field_path}' has an ambiguous underscore boundary."
+            )
```
a-lider, line 88: Wouldn't this block a legit filter on a detail key that's named something like `date` / `day`?

a-lider (approved):

posthog/api/advanced_activity_logs/filters.py
Piccirello: No, `date` and `day` are safe. The set comes from `JSONField().get_lookups()` and `KeyTransform.get_lookups()`, which under our settings is exactly 21 names:

```
contained_by contains endswith exact gt gte has_any_keys has_key has_keys
icontains iendswith iexact in iregex isnull istartswith lt lte range regex startswith
```

`date`/`day`/`year`/`hour` are transforms on `DateField`. They're never registered on `JSONField` or `KeyTransform`, so they pass validation and compile to a plain key transform:

| path | SQL |
| --- | --- |
| `detail__date` | `detail -> date` |
| `detail__day` | `detail -> day` |
| `detail__contains` | `detail @> 'x'` |
| `detail__regex` | `detail::text ~ '...'` |

The 21 that are blocked could plausibly be real detail keys (`contains`, `range`). But none of them were filterable before this change either: on master `contains` silently returns containment results instead of key results, and `range` raises `IndexError`, so a 500. The validator turns a silently wrong answer into a 400.

One genuine behavior change worth naming: master `continue`s past a rejected path, dropping that one filter and returning results for the rest. This rejects the whole request. Fail-closed is right here, but a saved filter set containing such a path now hard-fails instead of quietly degrading.

Added a test in 76598f5 pinning the accept side (`date`, `day`, `changes[].after.date`, `context.trigger_name`) so that widening the set to cover Django transforms rather than lookups fails loudly instead of silently rejecting ordinary keys.

Also worth flagging separately: `field_discovery.py` advertises whatever keys it finds in real detail JSON, unfiltered, so an org with a `context.range` key would be offered a filter this now rejects. It returned wrong results on master, so it isn't a regression, but filtering discovery to match the validator would be a reasonable follow-up.

PR #74436, opened 2026-07-28
fix(oauth): gate introspection on confidential clients

Piccirello:
## Problem

Anyone who knows a PostHog OAuth token's value can read its scopes, teams, and organizations, without holding any credential. Two gaps on public endpoints, both reachable without authentication.

1. **A secretless caller passes the introspection client-credentials gate.** `/oauth/introspect/` accepts client credentials as an alternative to a bearer token with the `introspection` scope. oauth2_provider's `_authenticate_basic_auth` and `_authenticate_request_body` both return True for a public client presenting `grant_type=urn:ietf:params:oauth:grant-type:device_code`. Both return before they reach `_check_secret`, so master's blank-secret guard `verify_client_secret` never runs on that path. The caller needs a public application's `client_id`, which is public by definition, and no secret.
2. **`/oauth/token/` returns a 500 for some well-formed JSON bodies.** `OAuthTokenView.post` parses the body then calls `.items()`, catching only decode errors. A body that parses to a list, `null`, a string, or a number raises `AttributeError`. A nested object as a parameter value raises `KeyError` further down. The endpoint is public, csrf-exempt, and unauthenticated, so each case returns a 500 rather than the RFC 6749 `invalid_request` 400.

Self-introspection already gives an access token that data about itself. The delta is refresh tokens, plus any token value the caller holds but is not presenting as its own bearer.

> [!NOTE]
> This PR started larger and has been narrowed twice as its other fixes landed elsewhere. https://github.com/PostHog/posthog/pull/74913 supersedes the CIMD verification-token binding it originally added, and is merged. Its blank-`client_secret` guard shipped on master as `verify_client_secret` in `posthog/api/oauth/client_auth.py`, so this PR uses master's.

## Changes

**Introspection.** Only a confidential client now passes the endpoint's client-credentials path. `ConfidentialClientOnlyOAuthValidator` re-checks `client_type` after the library's `authenticate_client` returns True.

This is a separate validator subclass rather than a change to `OAuthValidator`. The token endpoint legitimately tolerates a public client sending a secret it then ignores, because PKCE binds that grant, and an existing test asserts it.

**Token endpoint.** A JSON body that is not an object is rejected with an `invalid_request` 400. Values are coerced to strings before they reach `request.POST`. `null` is dropped as absent. An object or array value is rejected rather than left to fail as a string operation downstream.

**Stripe dev-app command.** `generate_stripe_app_tokens` created its application with `client_secret=""`. That row is confidential and authenticates with a secret, so under master's guard it can no longer authenticate at all. It now generates a real secret and prints it once, matching how Django admin issues one.

No migration. Nothing user-facing changes.

Automated only. I (Claude) did not exercise any of this by hand, and I did not test against a running instance.

I ran both groups against master with the fix reverted. Every case failed there and passed with it.

- `test_introspection_rejects_a_public_client_claiming_the_device_code_grant` (2 cases). Catches a regression where `validator_class` is dropped from the view, or the confidential check leaves the validator, and a secretless public client authenticates again. The two cases are the two transports, because `_authenticate_basic_auth` and `_authenticate_request_body` each carry their own copy of the short-circuit. Both returned 200 with a full introspection body before the change.
- `test_token_endpoint_rejects_json_payloads_that_are_not_string_objects` (5 cases). Catches a regression where a non-object JSON body, or a non-string parameter value, 500s the public token endpoint. It asserts the `invalid_request` error code rather than a non-500, so an incidental 400 does not pass for a spec-compliant one. All five 500'd before the change.

This PR previously tested the introspection gate with three blank-secret cases. Those now pass on master without the fix, because `verify_client_secret` covers them, so I replaced them rather than keeping them.

Ten tests in `posthog/api/oauth/test_views.py` fail locally with `TemplateDoesNotExist: index.html`, because `frontend/dist` is not built in that worktree. They fail the same way on a clean master checkout, which I verified by checking the two touched files out from master and rerunning.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

None needed.

## 🤖 Agent context

Claude (Claude Code, Opus 5) wrote the fixes and the tests. Skills invoked: `/writing-tests`, `/writing-code-comments`, `/writing-pr-descriptions`.

The original decision worth recording was where to put the empty-secret fix. Patching only the creation sites would have gone green on a creation-path test while leaving every existing row exploitable, so the check belongs in the validator, where it covers both auth paths and every row regardless of client type. Master reached the same conclusion independently, so this PR defers to it.

A later session found that the blank-secret rationale no longer justified the introspection change on its own, and traced the still-uncovered path to the library's device_code branch. That is why the test group changed shape.

Master came in through GitHub's update-branch API rather than a local merge. The branch was more than 1000 commits behind, and the signature ruleset rejects a push that newly reaches that much history.

Left alone: `OAuthIntrospectTokenView._is_self_introspection` parses the JSON body the same loose way the token endpoint did. It runs before any token validation, so an unauthenticated request with a bearer header and a JSON array body also 500s. Same shape as the token endpoint bug, and worth its own PR rather than widening this one.


MattBro:
@greptileai review

The verification-token binding this flagged has been removed from this PR. It is replaced by binding the URL at issuance time in https://github.com/PostHog/posthog/pull/74913, which closes the race rather than narrowing it. What is left here is client-credential hardening only.

MattBro (approved):
Approving. Merged master in to clear the 145-commit lag and re-ran CI, green.

Checked the caller binding for a fail-open and didn't find one: Basic auth goes through `client_credentials_from_basic_auth`, form bodies through `request.POST`, and credentials in a JSON body can't authenticate in the first place. The `resource_owner` assumption the whole thing rests on holds too, `dispatch` only sets it on the bearer fallback.

One question rather than a blocker: the binding only covers client-credentials callers, and `introspection` is in `ALWAYS_ALLOWED_SCOPES` so it skips an app's scope ceiling. A consented bearer token can still introspect any token in the deployment, which `test_introspecting_different_token_succeeds_with_introspection_scope` pins as intended. Pre-existing and not this PR's job, but worth a line somewhere on whether that door stays open.

Title still says "gate introspection on confidential clients" when the bigger change is now the caller binding, and the description doesn't mention the RecursionError work.

PR #74437, opened 2026-07-28
fix(api): require staff for instance debug endpoints

Piccirello:
## Problem

Two internal debugging endpoints were gated on a deployment-shape check rather than on the caller's privilege. The check assumed a self-hosted instance has one tenant. It doesn't - a self-hosted instance routinely hosts several organizations and teams, so being logged in there says nothing about whether you should be reading instance-wide diagnostics.

Both endpoints return executed query text: `/api/debug_ch_queries/` returns ClickHouse queries with the literals they ran with, and `/api/instance_status/queries` returns currently running and recent queries across the whole instance plus Postgres activity. That belongs to operators, not to every account on the box.

The `insight_id` / `experiment_id` lookups on the debug endpoint also had no team predicate - they matched on the ClickHouse `log_comment` field alone.

## Changes

- `DebugCHQueries.list` now gates on `is_staff or impersonated`, with `DEBUG` kept for local development. The deployment-shape disjunct is gone, and the now-unused `is_cloud` import with it.
- The filtered branch of `queries()` / `stats()` / `hourly_stats()` binds the requesting user's team into the WHERE clause (`JSONExtractInt(log_comment, 'team_id') = %(team_id)s`), so a filtered read stays inside one tenant regardless of who is calling. A caller with no current team is refused rather than run unscoped. The unfiltered branch was already user-scoped (`user_id` in the query comment) and is unchanged.
- `InstanceStatusViewSet.queries` moves to `[IsAuthenticated, IsStaffUser]` on the action. `SingleTenancyOrAdmin` is left alone: `instance_status` is its only caller, and the other two actions on the viewset (`list`, `navigation`) legitimately serve every user - `navigation` in particular feeds the app shell.
- `InsightPanelToggles` and `InsightSceneMenuBar` - the two places the insight scene renders the "Debug panel" toggle - used a broader condition than the backend now allows (`!preflight?.cloud`, i.e. "show it to everyone off Cloud"). Fixed to mirror the backend gate: staff, impersonated, or local `DEBUG` (via `preflight.is_debug`).

I checked the team predicate against a live ClickHouse rather than only through the tests: `JSONExtractInt` coerces both the numeric and quoted forms, `team_id` and `insight_id` are fields on the same `QueryTags` model so they land in the same `log_comment` object, and the combined clause parses and runs.

> [!NOTE]
> Two behavior changes worth knowing about.
>
> 1. The staff gate on the debug endpoint narrows a legitimate workflow: reading another team's query history now needs impersonation or a project switch.
> 2. Self-hosted operators who enable the `debug_queries` instance preference (or run with `DEBUG`) surface the CH debug modal to every user. Non-staff users who open it will now get a 403 where they previously got results. Cloud users have always seen that 403, so this is the existing Cloud behavior extended everywhere.
>
> No UI change is needed for `instance_status/queries`: its only consumer is the Internal metrics tab, which the frontend already renders behind `user.is_staff`. The insight "Debug panel" toggle did need one - a self-hosted, non-staff user viewing their own saved insight could still see and enable it, then hit an unexplained empty panel on the 403. That's fixed here by tightening the toggle's condition to match.

The sibling `list` action on `InstanceStatusViewSet` still reports Postgres / Redis / ClickHouse versions, object storage state, and deployment metadata to any authenticated user. Left as is here to keep this diff tight; worth a separate look.

Automated only - I (Claude) did not exercise these endpoints or the frontend toggle by hand.

Four tests, written before their respective fix and confirmed failing first, each for the reason it's meant to catch rather than on a fixture or import error:

| Test | Regression it catches |
| --- | --- |
| `test_non_staff_denied_off_cloud` | Reinstating any deployment-shape escape hatch in the `DebugCHQueries.list` guard. |
| `test_filtered_queries_are_scoped_to_the_requesting_users_team` | Dropping the bound `team_id` predicate from the filtered branch. Asserts every `sync_execute` call binds `%(team_id)s` to the requester's team - `system.query_log` can't be seeded, so asserting the bound parameter is the cheapest way to catch the removal. |
| `test_filtered_queries_denied_when_requester_has_no_current_team` | Dropping the `team is None` refusal on the filtered branch, which would otherwise fall through to an unscoped read across every team on the instance. |
| `test_queries_requires_staff_off_cloud` (parameterized non-staff / staff) | Reverting `instance_status/queries` to a permission that turns permissive off Cloud, and, via the staff case, over-tightening it so operators lose the page. |

Both endpoint tests pin `CLOUD_DEPLOYMENT` explicitly rather than inheriting it from how the suite is invoked, and mock ClickHouse at the boundary so they stay deterministic and cheap.

Two existing tests needed updating alongside the fix:

- `test_instance_status_routes` asserted a 200 on `queries` as a non-staff user. It now sets `is_staff` before that call; `list` and `navigation` stay non-staff.
- `test_denied` patched `posthog.api.debug_ch_queries.is_cloud`, a symbol the module no longer imports. That scaffolding is removed; its three real assertions (DEBUG on, DEBUG off, staff) are preserved, and the staff case is now asserted with `DEBUG` pinned off, which is strictly stronger than before.

```
pytest posthog/api/test/test_debug_ch_queries.py posthog/api/test/test_instance_status.py
19 passed
```

No existing Jest/Playwright test covers the "Debug panel" toggle's visibility, so the frontend fix has no automated regression test; I verified it with oxlint (clean) and confirmed the pre-existing `typescript:check` failures on this branch are unrelated (stale package links in `mcp_analytics`, `posthog_ai`, `error_tracking`, etc., not touching either changed file).

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs cover these internal endpoints.

## 🤖 Agent context

Written by Claude Code (Opus) in a worktree, directed by me. Skills invoked: `/writing-tests` (the value gate, and the push to parameterize the instance-status case, mock ClickHouse rather than lean on a live cluster, and judge whether the `team is None` branch was worth a dedicated test rather than coverage-chasing), `/improving-drf-endpoints` (viewset permissions; no serializer or schema surface changed, so no `hogli build:openapi` run - permission classes don't reach the OpenAPI output), and `superpowers:receiving-code-review` for triaging the review feedback below.

A few decisions along the way, from the original diff: I considered changing `SingleTenancyOrAdmin` itself, since `instance_status` is its only caller, but rejected it: that would also lock `navigation` behind staff, and the app shell calls it for every user. Action-level `permission_classes` replace the viewset's list rather than merging with it, so the action declares `[IsAuthenticated, IsStaffUser]` explicitly - `IsStaffUser` alone would turn an anonymous request into a 403 instead of a 401. For team scoping I put the predicate inside `_log_comment_filter`, the one place all three filtered reads go through, so a future caller can't forget it, and threaded `team_id` as a parameter instead of passing `request` down into the stats helpers.

Follow-up addressing review feedback: an automated code review on this PR flagged that the insight-scene "Debug panel" toggle wasn't updated alongside the backend tightening, so a self-hosted non-staff user could still see and click it, then get a silent empty panel from the 403. I verified the claim by reading both toggle sites and confirmed the other debug entry points (help menu, account menu, keyboard shortcut) already gate on `superpowersEnabled`, which never had the stale disjunct - only these two did. Fixed by swapping `!preflight?.cloud` for `preflight?.is_debug`, which mirrors the backend's `DEBUG` disjunct exactly (both trace to `settings.DEBUG`), rather than dropping the disjunct entirely - the latter would have hidden the toggle from non-staff local developers that the backend still serves. Also closed the one coverage gap the CI bot flagged (the `team is None` refusal branch) with a dedicated test, since it guards a real tenant-isolation invariant rather than being coverage-chasing. Deliberately did not add error-state handling for a 403 response in `DebugCHQueriesImpl.tsx` (no consumer of `loadDebugResponseFailure` exists today) - out of scope for this fix, since the corrected gate means an authorized viewer shouldn't hit that path.


Piccirello:
### Code review

Found 1 issue:

1. The frontend gate for the insight "Debug panel" toggle wasn't updated to match the tightened backend check, so it still offers an action the API now rejects for some users.

`DebugCHQueries.list` now requires `is_staff` unconditionally:

https://github.com/PostHog/posthog/blob/05bace4a8ad2f59bd76df74f5b70383c17e8b331/posthog/api/debug_ch_queries.py#L366-L370

But the frontend condition that decides whether to show the "Debug panel" toggle for a saved insight still uses the old logic (`is_staff || is_impersonated || !preflight.cloud`), unchanged by this PR:

https://github.com/PostHog/posthog/blob/05bace4a8ad2f59bd76df74f5b70383c17e8b331/frontend/src/scenes/insights/SidePanel/InsightPanelToggles.tsx#L28-L32

https://github.com/PostHog/posthog/blob/05bace4a8ad2f59bd76df74f5b70383c17e8b331/frontend/src/scenes/insights/SidePanel/InsightSceneMenuBar.tsx#L124-L128

On a self-hosted instance, a non-staff, non-impersonating user viewing their own saved insight will still see the toggle. Turning it on now gets a 403 from `/api/debug_ch_queries/`, and `DebugCHQueries` has no error state for that case, so the panel just renders empty with no explanation. The other entry points to this same debug view (help menu, account menu, the global keyboard shortcut) are already gated behind `is_staff`/`superpowersEnabled`, so this looks like the one spot the frontend wasn't updated alongside the backend fix.

Piccirello:
Addressed the toggle gap flagged above in 01ed6d07: `InsightPanelToggles` and `InsightSceneMenuBar` both used `!preflight?.cloud` to show the "Debug panel" toggle, which is exactly the deployment-shape check this PR removed from the backend. Swapped it for `preflight?.is_debug`, mirroring the backend's `DEBUG` disjunct (both trace to `settings.DEBUG`), so a self-hosted non-staff user no longer sees a toggle the API will now 403 on. Confirmed the other debug entry points (help menu, account menu, keyboard shortcut) already gate on `superpowersEnabled`, which never had this disjunct, so these were the only two spots.

Also added a test for the `team is None` refusal on the filtered debug-query branch, closing the one coverage gap CI flagged.

PR #74438, opened 2026-07-28
fix(api): bound authorized-URL wildcard matching

Piccirello:
## Problem

Authorized URL entries (`app_urls`, and the conversations `widget_domains`) support `*` wildcards. Matching compiled a regex per request, per entry, by escaping the stored pattern and turning each `*` into an unanchored greedy capture group. Several of those groups separated by literals, evaluated against a long hostname that cannot satisfy the trailing literal, is the shape that makes Python's backtracking engine explore an enormous number of ways to split the input. One request could hold a worker far longer than any request should, and the cost multiplied by the number of configured entries.

The hostname side is attacker-supplied on paths that accept a URL or read the `Origin` / `Referer` header. The pattern side is team configuration, which had no validation at all beyond dropping empty strings.

## Changes

Two independent layers, either of which fixes the problem on its own.

**Matching is now linear.** `hostname_in_allowed_url_list` calls a new `_matches_wildcard_host` instead of building a regex. It splits the pattern on `*`, anchors the prefix and suffix, then walks the interior literals left to right with `str.find`. That decides this grammar exactly in a single pass, so cost stays linear in hostname length no matter how many wildcards the entry carries.

> [!NOTE]
> Matching semantics are unchanged. `*` still spans any run of characters including dots, so `*.example.com` still matches `a.b.example.com` exactly as it did before. The obvious alternative — tightening `*` to a single label with `[^.]*` — would have been a smaller diff but a silent behavior change for anyone relying on multi-label wildcards, so it was rejected. The parameterized test table pins the existing semantics against the new implementation.

**Pathological entries are rejected at write time.** `validate_authorized_url_wildcards` caps an entry at 5 wildcards in the host and is wired into `validate_app_urls` and `validate_conversations_settings` on both the team and project serializers, which are separate code paths that had drifted. Genuine entries carry one wildcard, occasionally two. Wildcards in the path are not counted, since only the host reaches the matcher. An entry `urlparse` cannot parse, such as `http://[`, is rejected too, because every later allowlist check re-parses the stored entry and raises there instead. `hostname_in_allowed_url_list` skips such an entry as well, so a team that stored one before this change keeps working.

I (actually Claude) wrote the tests first and confirmed they failed against master before the fix.

- `test_hostname_in_allowed_url_list_wildcards` (8 parameterized cases, `test_utils.py`) pins the matching semantics the regex had: wildcards spanning several labels, a bare `*` entry, several wildcards in one entry, a wildcard inside a label, and the negative cases where a leading or trailing literal must still match. This is the regression that a naive `[^.]*` fix would have introduced, and none of it was covered before.
- `test_validate_authorized_url_wildcards` (7 parameterized cases, `test_utils.py`) covers the cap boundary either side, an entry list where only one member is over, wildcards in the path not counting, and the two unparseable-entry shapes.
- `test_rejects_url_with_too_many_wildcards` (`test_team.py`) runs both TeamSerializer and ProjectSerializer directly, so the validator is proven wired into each rather than merely present.
- `test_hostname_in_allowed_url_list_skips_unparseable_entry` (`test_utils.py`) puts an unparseable entry ahead of a valid one, so it fails if the matcher aborts the loop instead of skipping the entry.

No timing assertion and no catastrophic-backtracking reproduction is committed — both are flaky in CI. I did confirm the finding reproduces locally before fixing it, using a throwaway timing probe against the old regex, and that probe is not part of the diff.

Automated tests run: `posthog/api/test/test_utils.py` and `posthog/api/test/test_team.py`. I did not exercise this in a browser.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed. The wildcard cap is generous enough that no documented configuration hits it.

## 🤖 Agent context

Written by Claude, working from a security review finding. Skills invoked: `writing-tests`.

The decision worth reviewing is the choice to preserve matching semantics rather than tighten `*` to a single label. Tightening would arguably be the better security posture, since a bare `*` entry currently matches any host at all. But that is a product decision about existing customer configuration, not a hardening patch, and shipping it inside a fix for a different problem would have made the behavior change hard to spot. Worth a separate discussion.

The write-time cap is belt and braces. With linear matching the stored pattern can no longer cause a blowup, so the validator exists to keep obviously-bogus configuration out of the database rather than to prevent the original problem.

One coordination note for reviewers: a sibling PR touches `unparsed_hostname_in_allowed_url_list` in the same file. The two do not overlap — this one only changes `hostname_in_allowed_url_list` and adds a new function below it — so they merge cleanly in either order.


Piccirello:
### Code review

Found 1 issue:

1. Non-string `widget_domains` entries crash validation with a 500 instead of a 400

`validate_authorized_url_wildcards` only catches `ValueError` around `parse_domain(url)` (which calls `urlparse(url).hostname`). `conversations_settings` is a raw `JSONField` with no item-level schema, so a request like `{"conversations_settings": {"widget_domains": [123]}}` reaches this function with a non-string entry. `urlparse()` raises `AttributeError` for non-string input (e.g. `'int' object has no attribute 'decode'`), not `ValueError`, so the exception isn't caught and propagates out of `is_valid()` as an unhandled server error. `app_urls` doesn't have this gap since it's `ArrayField(CharField(...))`, so DRF coerces entries to strings before `validate_app_urls` runs — `conversations_settings.widget_domains` has no equivalent coercion.

https://github.com/PostHog/posthog/blob/ca2771b3de9eb3d127f1ff36ae17c65503546438/posthog/api/utils.py#L568-L586

posthog/api/test/test_utils.py
```diff
@@ -338,3 +341,38 @@ def test_unparsed_hostname_in_allowed_url_list(
...
+
+    @parameterized.expand(
+        [
+            ("no wildcards", ["https://example.com"], False),
+            ("wildcard count at the cap", ["https://*.*.*.*.*.example.com"], False),
+            ("wildcard count over the cap", ["https://*.*.*.*.*.*.example.com"], True),
+            ("only one entry is over the cap", ["https://example.com", "https://*.*.*.*.*.*.com"], True),
+            ("wildcards in the path are not counted", ["https://example.com/*/*/*/*/*/*/*/*"], False),
+            ("entry with no parseable host", ["*.*.*.*.*.*.*.example.com"], False),
+            ("entry that cannot be parsed at all", ["http://["], False),
```
ReeceJones: I think there is a test case missing here of what if there are multiple wildcards in part of the same domain level. For example:  
  
`https://foo*bar.example.com`

posthog/api/utils.py
```diff
@@ -535,6 +563,31 @@ def parse_domain(url: Any) -> Optional[str]:
     return urlparse(url).hostname
 
 
+MAX_WILDCARDS_PER_AUTHORIZED_URL = 5
```
ReeceJones, line 571: Can you double check that there are no existing app urls in prod that are over this limit?

ReeceJones (approved):

posthog/api/test/test_utils.py
Piccirello: Added in d1a81417e66, as two rows on the existing table:

```python
("wildcard inside a label may match nothing", ["https://foo*bar.example.com"], "foobar.example.com", True),
("literals between wildcards must appear in order", ["https://a*b*c.example.com"], "acb.example.com", False),
```

Your example is the first one, pinned at the zero-width case so it also guards against a future rewrite that makes `*` require at least one character.

The second one is the gap I would have missed without your prompt: every case in the table passed even with the middle-literal scan deleted entirely, because the only multi-wildcard entry was `*.*.example.com`, whose middle literal is `.` and whose expected result is `True` either way. `a*b*c.example.com` not matching `acb.example.com` is the case that actually fails if that loop goes away.

For what it is worth, I also ran the new matcher against the regex it replaces over an exhaustive 1364 patterns x 364 hosts, plus targeted greedy-scan traps like `*foo*foo.com` vs `xfoofoo.com`. Zero behavioral differences, so these are regression guards rather than bug fixes.

posthog/api/utils.py
Piccirello: Checked both prod regions. Nothing is close: the highest wildcard count in any single entry is 2, against a cap of 5. That holds for `app_urls`, `widget_domains`, and `recording_domains`, and it stays true even when counting `*` anywhere in the raw string rather than just in the host, which is a strict superset of what the validator counts. So no grandfathering needed and no reason to move the cap.

That also makes the docstring claim above ("genuine entries carry one wildcard, occasionally two") empirical rather than assumed.

While confirming the blast radius I traced the callers, in case it comes up: the matcher is reached with `app_urls` (`urls.py`, `user.py`, `oauth/toolbar_service.py`) and with `widget_domains` (via `on_permitted_recording_domain`). `recording_domains` looks like an omission here but is not, since it never reaches the matcher server side, it ships to the SDK through `remote_config.py` for client-side enforcement. So the two validated fields are exactly the two that matter.

posthog/api/utils.py
```diff
@@ -535,6 +563,31 @@ def parse_domain(url: Any) -> Optional[str]:
...
+    Genuine entries carry one wildcard, occasionally two. Entries far beyond that only ever
+    make matching more expensive, so they are rejected at write time.
+    """
+    for url in urls:
+        if not isinstance(url, str):
+            # widget_domains arrives on a raw JSONField, so entries are not string-coerced for us.
+            raise ValidationError("Each URL must be a string.")
+        try:
+            host = parse_domain(url)
+        except ValueError:
```
a-lider, line 586: nit: `hostname_in_allowed_url_list` calls `parse_domain` on every stored entry with no try/except, so an entry like `http://[` accepted here raises `ValueError: Invalid IPv6 URL` and every allowlist check against that team's list 500s. I'd reject on `ValueError` instead of skipping - only the hostname-is-None case produces no match pattern.

posthog/api/team.py
```diff
@@ -1211,7 +1211,9 @@ def validate_access_control(self, value) -> None:
     def validate_app_urls(self, value: list[str | None] | None) -> list[str] | None:
         if value is None:
             return value
-        return [url for url in value if url]
+        urls = [url for url in value if url]
+        validate_authorized_url_wildcards(urls)
```
a-lider, line 1397: nit: `recording_domains` flow through the same wildcard matcher (`on_permitted_recording_domain` -> `hostname_in_allowed_url_list`) but don't get this cap. The linear matcher makes that safe DoS-wise, so this isn't blocking, but if the write-time cap is worth having for `app_urls`. Maybe it's worth adding the cap in `validate_recording_domains` too.

posthog/api/utils.py
Piccirello: Good catch, taken in 1a7ea5701e5. The comment on that branch was also wrong: it claimed an unparseable entry "never yields a match pattern", but it doesn't reach the point of yielding anything, it raises out of `parse_domain`.

Two changes rather than one:

- `validate_authorized_url_wildcards` now raises on `ValueError` instead of skipping, as you suggested.
- `hostname_in_allowed_url_list` skips an entry it cannot parse. Rejecting at write time does nothing for a team that already has one stored, and I would rather that team's allowlist checks keep working than 500 until someone edits the field.

`test_hostname_in_allowed_url_list_skips_unparseable_entry` puts the bad entry ahead of a good one, so it fails if the matcher ever aborts the loop instead of skipping the entry.

Related, and out of scope here: the same `ValueError` is reachable from the request side, not just from stored config. `/api/user/redirect_to_site/?appUrl=http://[` reaches `unparsed_hostname_in_allowed_url_list`, which calls `urlparse(unquote(hostname))` on the raw query param, and `has_authority_bypass_chars` only screens for `\` and `%5c`. That is pre-existing on master and it is in the function the sibling PR is editing, so I have left it alone here rather than create a conflict. Flagging it there.

posthog/api/team.py
Piccirello: `recording_domains` doesn't reach the matcher. `on_permitted_recording_domain` has one caller, `validate_origin` in `products/conversations/backend/api/serializers.py`, and it passes `conversations_settings.widget_domains`. The function name is historical.

`recording_domains` only leaves Postgres through `remote_config.py`, which ships it to the SDK for client-side enforcement. Nothing server side ever parses it as a match pattern, so there is no matcher cost to bound.

Adding the cap would mean touching `validate_recording_domains` on both TeamSerializer and ProjectSerializer for a field with no server-side matcher path, so I'd do that separately rather than widen this diff.

PR #74439, opened 2026-07-28
fix(file-system): validate path and meta inputs

Piccirello:
## Problem

`FileSystemSerializer` accepted `path` and `meta` straight from the request body with no field validation at all.
Two things fall out of that.

`meta` is a writable `JSONField`, so a write can store `null`, a list, a number, or a string where every reader expects an object.
The list endpoint then calls `.get("created_by")` on whatever is stored, for every row on the page.
One bad row is enough to break the file tree for everyone in the project, not just whoever wrote it, and the row stays broken until someone deletes it.
The default list ordering is `Lower("path")` with a page size of 100, so a shallow row sorting early lands on page 1 of both the plain list and the `depth=1` sidebar load.
`publish_canvas` and `_canvas_share_url` read the same field and have the same problem.

`path` had no length or segment-count bound.
`FileSystemSerializer.create` walks the segments and does one existence check plus one insert per segment, so a single modest request body turns into a very large number of sequential queries and inserts on one worker.
We do not enable `ATOMIC_REQUESTS`, so each insert autocommits: even if the request dies partway, the folder rows already written stay in the project's tree, and the request is repeatable.
The `move` and `link` actions read `new_path` off `request.data` and hand it to the same loop without going through a serializer at all.

## Changes

Field validation on `FileSystemSerializer`, plus a read-side guard for rows that are already bad.

- `validate_meta` rejects anything that is not an object, and caps the serialized value at 1 MB.
- `validate_path` bounds a path at 4000 characters and 50 segments, via a new module-level `validate_file_system_path` helper.
- `move` and `link` call that same helper on `new_path` before `_assure_parent_folders`, so the actions that skip the serializer get the same bound.
- `_created_by_users` now skips a row whose `meta` is not a dict instead of raising. Write-time validation only stops new bad rows; this is what stops a row already in a live database from breaking unrelated listings.

On the bounds. Paths the product builds itself top out at three segments (`Unfiled/Insights/<name>`), so 50 segments leaves a lot of headroom for hand-made trees while making the folder-creation loop cheap. 4000 characters comfortably covers 50 segments of long object names even after `escape_path` inflates them. The `meta` ceiling is deliberately loose: it is a free-form client blob, and 1 MB is still a large narrowing against the 20 MB `DATA_UPLOAD_MAX_MEMORY_SIZE` ceiling.

Both validators are `validate_<field>` methods rather than redeclared fields, so the OpenAPI schema and the generated frontend types are unchanged.

> [!NOTE]
> Server-generated paths do not go through the serializer. `create_or_update_file` and `save_unfiled_files` build paths from object names and are untouched, which is why the bound lives at the API boundary and not in `split_path` or on the model.

Automated only. I did not exercise this in a running app.

Tests were written before the fix and each one was confirmed to fail on the vulnerability rather than on a fixture or import error. Pre-fix, the over-limit path POST returned 201 and created 51 rows, and the poisoned-row list request returned 500 from an `AttributeError`.

Two new test classes in `posthog/api/file_system/test/test_file_system.py`:

- `TestFileSystemSerializerInputValidation` (`SimpleTestCase`, no DB) holds the parameterized bounds matrix: `meta` as null, list, number, string, and over-size; `path` over the length and segment bounds; plus in-bounds control cases so the bounds can't silently swallow valid input. Catches a regression where a bound is removed or loosened.
- `TestFileSystemInputValidationAPI` (`APIBaseTest`) holds the three cases that genuinely need the DB. An over-limit create returns 400 and writes zero rows, which doubles as the wiring guard that the viewset actually runs this serializer. `move` and `link` with an over-limit `new_path` return 400 and leave the tree untouched, covering the two paths that never see a serializer. And a row created directly with a non-object `meta` no longer 500s the list endpoint, which is the "already poisoned in production" case that write-time validation cannot reach.

I put the `meta` read guard in `_created_by_users` rather than at each call site, so the one fix covers both `list` and `_list_recents`, and skipped a second test for the recents path as redundant.

```text
posthog/api/file_system/test/ + posthog/api/test/test_file_system.py + posthog/models/file_system/
255 passed
```

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

No user-facing docs change. The new 400s are input bounds far outside normal use, and no documented workflow changes.

## 🤖 Agent context

I directed this; Claude did the implementation. Skills invoked: `/writing-tests`, `/improving-drf-endpoints`, `/writing-user-facing-copy`.

A few decisions worth surfacing for review. We considered declaring `path = CharField(max_length=...)` and a typed `meta` field instead of validator methods, and rejected it: redeclaring fields changes the OpenAPI schema and drifts the committed generated types, which is a much bigger diff for no security benefit. We also considered pushing the path bound down into `split_path` or a model `clean()`, and rejected that because it would break unfiled-file saving and folder repair, which legitimately construct paths server-side.

Two things deliberately left out of scope. `bulk_create` for the parent folders was suggested but is pointless once the segment count is bounded, and it needs uniqueness-constraint reasoning this change doesn't have. And `CanvasPublishSerializer.name` is a third unbounded write that reaches `path` through the leaf rename; it's adjacent but not one of the two issues here.

One thing I could not verify: the desktop client is not in this repo, so I could not confirm nothing round-trips a full `meta` (canvas code plus version history) back through the generic endpoint. Canvas publishing itself writes `meta` directly on the model and skips the validator, so it is unaffected, and I searched the frontend and found no such round-trip. The 1 MB ceiling is generous for that reason. If a desktop client does round-trip a large `meta`, that is the thing to watch.


Piccirello:
### Code review

Found 2 issues:

1. `undo_delete`'s restore path skips the new path bound (`create`/`move`/`link` all validate first)

`_restore_file_system_path` passes the request-supplied `path` straight into `_assure_parent_folders`'s per-segment loop, without going through `validate_file_system_path` the way `create` (via the serializer), `move`, and `link` now do. `UndoDeleteItemSerializer.path` is a plain `CharField` with no length or segment cap, and `FileSystem.path` itself is an unbounded `TextField`, so this is the one remaining path into that loop with no cap at all.

https://github.com/PostHog/posthog/blob/7f36ae3dd181691f5c2d24608c23fe2221a544be/posthog/api/file_system/file_system.py#L982-L997

https://github.com/PostHog/posthog/blob/7f36ae3dd181691f5c2d24608c23fe2221a544be/posthog/api/file_system/file_system.py#L200-L205

2. `publish_canvas` / `_canvas_share_url` still assume `meta` is a dict

`_created_by_users` was changed in this PR to skip a row whose `meta` isn't a dict instead of crashing, but `publish_canvas`'s `meta = dict(dashboard.meta or {})` and `_canvas_share_url`'s `(dashboard.meta or {}).get("channelId")` read the same field on an existing row without that guard. A non-dict `meta` (list, string, or number) raises in both places, so a row in the same bad shape this PR defends against elsewhere will still 500 here.

https://github.com/PostHog/posthog/blob/7f36ae3dd181691f5c2d24608c23fe2221a544be/posthog/api/file_system/file_system.py#L1220-L1224

https://github.com/PostHog/posthog/blob/7f36ae3dd181691f5c2d24608c23fe2221a544be/posthog/api/file_system/file_system.py#L1327-L1331

mariusandra (approved):

PR #74444, opened 2026-07-28
feat(replay-vision): instrument user-facing events for GA

fivestarspicy:
## Problem

Before Replay Vision goes GA we need the user-facing surfaces instrumented so we can calibrate activation and quality. An audit of the code found the instrumentation was thin and partly non-functional:

- Only scanner *created* actually lands. The *enabled*/*disabled* events never fire (scanners are born enabled, and those events only fire on a post-creation transition), and the event names used spaces instead of the `replay_vision_` snake_case convention.
- The core quality signal (the thumbs up/down rating on an observation) fired **no** event on either surface.
- On-demand scans, bulk scans, scanner edits/deletes, affected-cohort creation, and quota exhaustion were entirely uninstrumented.

Because the MCP tools are thin wrappers over these same DRF endpoints, putting the events server-side covers both the UI and MCP in one path. Raised in a Slack thread (linked below).

## Changes

All events fire server-side via the existing `report_user_action` pattern, passing `request=` so each event carries the standard request-derived analytics props: the platform `source` vocabulary (`web` for the app UI, `mcp`, `api`, ...), `$session_id`, and `was_impersonated` (so support sessions can be excluded from calibration funnels).

- Renamed to `replay_vision_` snake_case: `replay_vision_scanner_created` (carries `enabled`), plus `replay_vision_scanner_enabled` / `replay_vision_scanner_disabled` on genuine post-creation toggles only.
- New scanner events: `replay_vision_scanner_edited` and `replay_vision_scanner_deleted`. The edited event fires only when submitted values actually changed and carries `edited_fields`. A save that changes nothing fires nothing, which matters because the UI PATCHes the entire form on every save.
- New scan events: `replay_vision_scan_requested_on_demand`, `replay_vision_bulk_scan_started`, and `replay_vision_quota_exhausted`. The quota event fires when an on-demand scan is blocked with a 402, or when a bulk batch actually left sessions unscanned for quota (judged by the per-session outcomes, not the headroom label, so partial batches that never hit the cap don't report exhaustion).
- New observation events: `replay_vision_observation_rated` (the priority calibration signal, carrying `is_correct` and `has_feedback`) and `replay_vision_scanner_viewed` / `replay_vision_observation_viewed` to support a later created → viewed → rated funnel. `observation_viewed` fires only for terminal observations: the observation scene polls the retrieve endpoint every few seconds while a scan is in flight, so in-flight fetches are poll ticks, not views.
- `replay_vision_affected_cohort_created` on the "Save as cohort" action.

The activation event itself is intentionally **not** defined here. That's a later call once these events have data. Pure client-side prompt-suggestion events (generate/apply/dismiss) are deferred to a follow-up since they have no server endpoint to hang off.

All 211 tests in the two touched test files pass locally after merging current master (the `bulk_observe` area had been reworked on master), and `ruff` and `mypy` are clean on the product. New and updated tests, each guarding a specific regression:

- `test_rating_reports_calibration_event` asserts at the `posthoganalytics.capture` boundary that the rating event carries `is_correct`, `has_feedback`, and `source` (session auth resolves to `web`). Catches both a dropped payload and a dropped `request=`.
- `test_full_body_save_reports_only_actually_changed_fields` (parameterized rename/no-op) locks in that a whole-form PATCH reports only real changes and a no-op save fires nothing.
- `test_retrieve_reports_viewed_only_for_terminal_observations` (parameterized pending/succeeded) guards the poll-tick gate on `observation_viewed`.
- `test_skips_for_quota_when_the_credit_budget_is_the_tighter_limit` now also asserts the `bulk_scan_started` + `quota_exhausted` pair, and the new `test_quota_bound_batch_that_fits_does_not_report_exhaustion` proves a quota-bound batch with an already-running session does not report exhaustion.
- `test_quota_blocked_observe_reports_exhaustion` covers the on-demand 402 path (`trigger: on_demand`).
- Existing lifecycle assertions updated for the renamed events.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Originally authored by the PostHog Slack app from a Slack thread, after a read-only audit of Replay Vision event instrumentation verified against the live event taxonomy. A follow-up Claude Code review session then merged master in, replaced a product-local source helper with `report_user_action`'s built-in `request=` analytics props (standard `source` vocabulary instead of a product-specific one), tightened the quota/viewed/edited event semantics described above, and added the regression tests.

Decisions worth noting: `source` uses the platform-wide values from `get_event_source` (`web`/`mcp`/`api`/...) rather than a product-local `ui` label, so cross-product dashboards filter one vocabulary. `replay_vision_scanner_edited` is separated from the enable/disable toggle by diffing the written fields against the instance, so a pure toggle doesn't double-report and a no-op save stays silent. No serializer fields changed, so no OpenAPI/MCP schema regeneration was needed.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B7TE4R8GH/p1784767180759699)*


arnohillen (approved):
Reviewed the instrumentation logic, event naming, privacy of event properties, and test coverage. Looks good.

PR #74450, opened 2026-07-28
feat(self-driving): Allow keeping report charts the same

Twixes:
## Problem

The `edit_report` tool now (as of  #73733) allows settings `charts` on a report. The problem: no way to keep the charts just the same as they were without repeating. `charts: []` was the only allowed empty value, and it was _resetting_ the charts.

I meant to land this with it but the PR merged first, so here it is on its own.

## Changes

`charts` on `edit_report` now has three states instead of two:
- field omitted or `null` - charts left as they are (new!)
- `[]` - charts cleared (existing)
- `[...]` - charts replaced with that set (existing)

A few unit & API tests to prevent regressions here.


andrewm4894 (approved):
sweet yep defo better to have this option too

PR #74098, opened 2026-07-28
fix(mcp-analytics): probe setup from event definitions

rafaeelaudibert:
## Problem

MCP Analytics ran an events-table query during app startup to decide whether setup was complete. The setup screen also showed only a listening state, which made it hard to tell whether MCP events were arriving with the expected names.

## Changes

- Preload setup status with one exact, batched event-definitions lookup. `$mcp_initialize` means the server connected, while `$mcp_tool_call` confirms usable MCP Analytics data.
- Keep polling for fresh events while the MCP Analytics page is open.
- Show the five latest MCP events beside the setup status. The stream is filtered to MCP Analytics event types and displays only event names and receipt times.
- Document the generated API namespace-import convention in the Kea logic skill.

## How did you test this code?

- 30 focused frontend tests pass, covering setup-state mapping, existing-event detection, polling transitions, MCP stream filtering, and safe event rendering.
- Added an event-definitions API test for repeated and comma-separated exact names. Local execution requires Postgres, which was not running in this checkout.
- OpenAPI generation, Ruff, frontend formatting, skill lint, and `git diff --check` pass.
- I could not run the full frontend TypeScript check because this checkout does not have the required workspace packages built.
- I did not capture a screenshot because the development stack was not running.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

## Docs update

No user documentation update is needed. The setup steps are unchanged.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Codex implemented this change using `/writing-user-facing-copy`, `/writing-tests`, `/adopting-generated-api-types`, `/improving-drf-endpoints`, `/using-kea-disposables`, `/writing-code-comments`, `/writing-skills`, and `/writing-kea-logics`.

The implementation moved the startup probe from property definitions to exact event definitions, then batched all registered setup probes into one generated API call. It reuses the shared livestream connection with an MCP-specific server filter and a defensive UI filter.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #74452, opened 2026-07-28
feat: add setup empty states for flags, experiments, and more products

rafaeelaudibert:
## Problem

Feature flags and Experiments are creation-first products, but until now their first-run experience was the deprecated `ProductIntroduction` banner sitting above an empty table. MCP analytics already ships the shared `ProductEmptyState` setup screen (pitch on the left, animated example-data preview on the right); this PR brings Feature flags and Experiments onto the same platform.

Stacked on #74098.

## Changes

Both products adopt the shared empty-state gate via `emptyState` on their `SceneExport`. Unlike MCP analytics there is no wizard and no waiting-for-data state: detection is a plain entity count (does the project have any flags/experiments?), and the single primary button links to the creation page (`/feature_flags/new`, `/experiments/new`).

- **Detection logics** (`featureFlagsSetupLogic`, `experimentsSetupLogic`): one `limit: 1` list call through the generated API clients, mapped to `needs-setup` / `has-data`, failing open to `unknown` so the gate never strands on its spinner.
- **Interactive previews**, CSS-only per the platform rules (hidden checkbox/radios driving `:checked ~` styles, no timers, `prefers-reduced-motion` and static-snapshot variants):
  - Feature flags: a mini flag list wired to a mini app. Flipping the `one-click-checkout` flag re-skins the checkout UI below it (new express row, new CTA) and the rollout jumps 0% → 100%.
  - Experiments: a running experiment's results card (animated conversion bars, leading badge, win probability) wired to the same mini app. Selecting a variant shows what users in it see.
- **Scene cleanup**: removed the deprecated `ProductIntroduction` call sites from both scenes (the `ApprovalsPromoBanner` gating now keys off `shouldShowEmptyState` directly).
- **Stories**: one per product in `ProductEmptyState.stories.tsx` via `productEmptyStateStory`, rendering the exact shipped configs with the count queries mocked empty.

> [!NOTE]
> Removing `ProductIntroduction` also removes the `MCPUseCaseCard` it rendered via `mcpSurfaceKey` (`feature_flags.create` / `experiments.create`). The empty-state platform has no equivalent surface yet; happy to wire one up as a follow-up if we want to keep that entry point.

### Eight more products (second commit)

The same platform now also gates **Early access features, Prompts, Skills, Endpoints, Product tours, Support, Web scripts, User research, and Links** (no screenshots for these - same layout as above, each with its own interactive example-data preview):

- Detection is an entity count via each product's generated list client (`limit: 1`), except Support, which uses the full three-state ladder: `conversations_enabled` off → needs-setup, on with no tickets → waiting-for-data (polling until the first ticket lands).
- The platform config grew a `PrimaryAction` component slot for create actions that need hooks: User research's button opens PostHog AI with the same `create_user_interview_topic` tool as the scene's "New topic" button, and a poll clears the gate once the agent creates a topic.
- Product tours' primary action creates a starter tour via the generated API and routes to the editor (`/product_tours/new` can't save, and the scene's creation modal can't render behind the gate).
- Visual review deliberately does **not** adopt an empty state: `/visual_review` now forwards to settings when no repo is connected (it already forwarded to the repo's runs page otherwise).
- Removed the remaining `ProductIntroduction` call sites the gate replaces (links, prompts, endpoints, web scripts' `site_app` mapping entry) and Support's bespoke first-run banner.
- Defined the missing product color tokens (endpoints - previously referenced but undefined - product tours, visual review, and dark variants for llm-prompts, links, user-interviews, data-pipeline) and wired the light/dark pairs through the manifests' `iconColor` tuples and the project tree.
- Flag-gated products (endpoints, product tours, links, user research) set `featureFlag` on their config so the gate only engages where the product is available.
- The links preview's podcast "Play episode" button reuses the inbox onboarding `playMeep()` easter egg.

### Feature flags

| Flag off | Flag flipped on |
| --- | --- |
| feature-flags-empty-state | feature-flags-empty-state-flag-on |

### Experiments

| Control selected | Test variant selected |
| --- | --- |
| experiments-empty-state | experiments-empty-state-test-variant |

<details>
<summary>Dark mode</summary>

flags-dark
experiments-dark

</details>

- New jest tests for every detection logic (11 suites, 56 tests): parameterized count/ladder → status mapping plus the failure path. They guard the connect wiring into `productSetupStatusLogic` - if that breaks, the scene gate strands on its spinner or shows the wrong surface, and no existing test catches it.
- Ran the existing `experimentsLogic` suite (40 passing) after removing the scene's `ProductIntroduction`.
- `pnpm --filter=@posthog/frontend typescript:check` clean; lint/format via `pnpm --filter=@posthog/frontend fix`.
- Screenshots above were taken from the new storybook stories (light + dark), including the interactive states, so the interactions are verified in a real render. I did not run the app end to end against a project with zero flags/experiments.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Authored by Claude Code (Fable 5) in an interactive session directed by Rafael; the second commit's nine adoptions were built by three parallel subagents from per-product fact sheets, then integrated and verified centrally. Skills invoked: `building-product-empty-states` (platform contract, detection rules, preview constraints), `writing-user-facing-copy`, `writing-tests`. Decisions: detection uses generated `featureFlagsList`/`experimentsList` with `limit: 1` rather than a HogQL event query, since these products are entity-based (per the platform skill, they skip the boot-time `setupProbe`, which is event-based); no feature flag gating the rollout, matching the direction to ship these as full adoptions; primary action targets `/feature_flags/new` rather than the templates picker, per the session's direction to link the creation pages.


rafaeelaudibert:
* **#74452** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog/74452?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#74098** <a href="https://app.graphite.com/github/pr/PostHog/posthog/74098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

fercgomes (approved):

products/feature_flags/frontend/emptyState/featureFlagsEmptyState.tsx
```diff
@@ -0,0 +1,35 @@
...
+        icon: <IconToggle />,
+        accentColor: 'var(--color-product-feature-flags-light)',
+        hedgehog: HedgehogTrafficController,
+        text: {
+            'needs-setup': {
+                headline: 'Ship code without shipping the feature',
+                lead: 'Wrap a change in a feature flag, roll it out to 1% of users, and watch what happens: the session replays, events, and exceptions from the people who got it. Turn it off the moment something looks wrong, no redeploy needed. Flags also power experiments, early access programs, kill switches, and remote config.',
+            },
+        },
+        primaryAction: { label: 'Create your first feature flag', to: urls.featureFlag('new') },
```
haacked: `blocking`: This is what turns `quickcreate.spec.ts` red, and it takes all 8 tests in the file with it. That spec's shared `beforeEach` creates a workspace with `no_demo_data: true`, so the project has zero flags, the gate renders the empty state instead of the scene, and [the `[data-attr="new-feature-flag"]` click](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/surveys/frontend/e2e/quickcreate.spec.ts#L118) has nothing to match. The describe is `mode: 'serial'`, so a failure in setup fails everything after it.

Adding a `data-attr` passthrough isn't enough on its own: the scene's button goes to `urls.featureFlagTemplates()` and the spec's next line clicks `blank-feature-flag-template`, while this config links straight to `/feature_flags/new` and skips the picker. Route the spec around the gated scene instead:

```ts
await page.goto(urls.featureFlagTemplates())
await page.evaluate(() => {
    window.posthog?.featureFlags?.override({})
})
await page.locator('[data-attr="blank-feature-flag-template"]').click()
```

products/experiments/frontend/emptyState/experimentsEmptyState.tsx
```diff
@@ -0,0 +1,35 @@
...
+        accentColor: 'var(--color-product-experiments-light)',
+        hedgehog: HedgehogExperiment,
+        text: {
+            'needs-setup': {
+                headline: 'Test changes on real users before you commit',
+                lead: 'Split traffic between variants and measure the impact on the metrics you care about. Choose Bayesian or frequentist statistics, keep a holdout to track long-term impact, and reuse shared metrics across experiments. Run no-code experiments on your site, or test anything a flag can gate, LLM prompts included.',
+            },
+        },
+        primaryAction: { label: 'Create your first experiment', to: urls.experiment('new') },
+        skippable: false,
```
haacked, line 39: `blocking`: A project that archived all of its experiments loses the whole scene and can't get back. The list endpoint [excludes archived rows unless asked for them](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/experiments/backend/presentation/views.py#L272), so the count returns 0 and the gate swaps in the setup screen. Unarchiving is done from that list, and `skippable: false` removes the only way through, so the recovery path is knowing an experiment's URL by heart.

Feature flags has the same hole via `exclude_archived_unless_requested`, and there's no param value that counts both states. Zero-entity projects also lose in-scene tabs with no route of their own: Holdout groups (created before your first experiment, not after) and Settings here, and the Projects tab on flags, which shows other projects' flags and has content even when this one has none. That's exactly what `skippable`'s own docstring carves out.

```diff
         primaryAction: { label: 'Create your first experiment', to: urls.experiment('new') },
-        skippable: false,
         docsUrl: 'https://posthog.com/docs/experiments',
```

Same removal in `featureFlagsEmptyState.tsx:30`. Links, product tours and skills can keep it: their scenes really are just a list, and the tours count includes archived rows.

products/early_access_features/frontend/emptyState/earlyAccessFeaturesEmptyState.tsx
```diff
@@ -0,0 +1,36 @@
...
+        accentColor: 'var(--color-product-early-access-features-light)',
+        accentColorDark: 'var(--color-product-early-access-features-dark)',
+        hedgehog: HedgehogRocket,
+        text: {
+            'needs-setup': {
+                headline: 'Run beta programs without building your own',
+                lead: 'Let users opt in and out of features at any stage, from early concepts to general availability. Run beta programs, collect waitlists for upcoming features, and give users control of their product experience, all without custom infrastructure. Each feature links to a flag, and a user opting in overrides its release conditions.',
+            },
+        },
+        primaryAction: { label: 'Create your first early access feature', to: urls.earlyAccessFeature('new') },
```
haacked: `blocking`: This is the `Early access feature new and list` failure CI is reporting. [The spec clicks a link named "New feature"](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/playwright/e2e/early-access-management.spec.ts#L24) against a workspace with zero features, but the gate has replaced the scene and this button reads "Create your first early access feature" with no `data-attr`. Both pre-creation clicks hit it, since the first feature is cancelled, and `skippable: false` means there's no path back to the scene's own button.

Matching either label keeps the spec working in both states:

```ts
const newFeatureLink = page.getByRole('link', {
    name: /New feature|Create your first early access feature/,
})
```

products/experiments/frontend/emptyState/experimentsEmptyState.tsx
```diff
@@ -0,0 +1,35 @@
...
+        icon: <IconFlask />,
+        accentColor: 'var(--color-product-experiments-light)',
+        hedgehog: HedgehogExperiment,
+        text: {
+            'needs-setup': {
+                headline: 'Test changes on real users before you commit',
+                lead: 'Split traffic between variants and measure the impact on the metrics you care about. Choose Bayesian or frequentist statistics, keep a holdout to track long-term impact, and reuse shared metrics across experiments. Run no-code experiments on your site, or test anything a flag can gate, LLM prompts included.',
+            },
+        },
+        primaryAction: { label: 'Create your first experiment', to: urls.experiment('new') },
```
haacked: `blocking`: A viewer-role user now gets an enabled "Create your first experiment", follows it, fills the form, and only finds out on save. This is a regression rather than an omission: the `ProductIntroduction` this replaced was wrapped in `AccessControlAction` at `Editor` and [forwarded `disabledReason` onto its create button](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/frontend/src/lib/components/ProductIntroduction/ProductIntroduction.tsx#L227), and [the scene's own button still gates](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/frontend/src/scenes/experiments/Experiments.tsx#L555). With `skippable: false` that button is the only thing on the screen. Grepping all eleven `emptyState/` directories for `AccessControl` returns nothing, so the same gap is in flags, links, skills and product tours.

`ProductEmptyStatePrimaryAction` can't express a permission, which is why this happened. The `PrimaryAction` slot already can:

```tsx
function CreateFirstExperimentButton(): JSX.Element {
    return (
        <AccessControlAction
            resourceType={AccessControlResourceType.Experiment}
            minAccessLevel={AccessControlLevel.Editor}
        >
            <LemonButton type="primary" to={urls.experiment('new')} className="self-start">
                Create your first experiment
            </LemonButton>
        </AccessControlAction>
    )
}
```

then `PrimaryAction: CreateFirstExperimentButton`. Worth deciding once for all eleven, since growing the config with `accessControl` and `dataAttr` would also hand the Playwright specs back their selectors.

products/cdp/frontend/emptyState/webScriptsSetupLogic.ts
```diff
@@ -0,0 +1,91 @@
...
+            teamLogic,
+            ['currentProjectId'],
+        ],
+    })),
+    loaders(({ values }) => ({
+        scriptCount: [
+            null as number | null,
+            {
+                loadScriptCount: async (_: void, breakpoint): Promise<number> => {
+                    const response = await hogFunctionsList(String(values.currentProjectId), {
```
haacked: `blocking`: A project whose web scripts are all legacy plugin site apps gets told to create its first one, and loses access to the scripts running on its site today. This counts only `site_app` hog functions, but the scene it gates also lists plugin-backed site apps loaded from `pipeline_frontend_apps_configs` and passed in as [`manualFunctions`](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/frontend/src/scenes/data-pipelines/DataPipelinesHogFunctions.tsx#L93). Those projects count zero, and `skippable: false` means there's no way to the list or the History tab. The loader that fetches them still calls the migration "super temporary", so these projects are expected to exist.

Count both sources:

```ts
const [hogFunctions, legacySiteApps] = await Promise.all([
    hogFunctionsList(String(values.currentProjectId), { type: ['site_app'], limit: 1 }),
    api.get<{ count: number }>(
        `api/projects/${values.currentProjectId}/pipeline_frontend_apps_configs?limit=1`
    ),
])
breakpoint()
return hogFunctions.count + legacySiteApps.count
```

If you'd rather not add a request, dropping `skippable: false` at least leaves an escape hatch.

products/links/frontend/emptyState/LinkPreview.tsx
```diff
@@ -0,0 +1,176 @@
...
+                        <span className="LinkPreview__on-a">Spring sale</span>
+                        <span className="LinkPreview__on-b">Episode 42</span>
+                    </span>
+                    <span className="LinkPreview__swap LinkPreview__page-sub">
+                        <span className="LinkPreview__on-a">20% off sitewide, this week only</span>
+                        <span className="LinkPreview__on-b">How small teams ship faster</span>
+                    </span>
+                    <span className="LinkPreview__swap LinkPreview__page-cta">
+                        <span className="LinkPreview__on-a">Shop the sale</span>
+                        <button type="button" className="LinkPreview__on-b LinkPreview__play" onClick={playMeep}>
```
haacked, line 121: `blocking`: A keyboard user tabbing through this empty state lands on an invisible button and gets a toast and an audio clip on Enter. The button carries `LinkPreview__on-b`, which is `opacity: 0` until the podcast radio is checked, and [the only thing gating it is `pointer-events`](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/links/frontend/emptyState/LinkPreview.scss#L236), which suppresses mouse and touch but leaves the button in the tab order. Its `:focus-visible` outline paints on an `opacity: 0` element, so the focus ring disappears for that stop too.

Hiding the swap variants with `visibility` as well as opacity fixes both. A `visibility: hidden` grid item still occupies its cell, so there's no layout shift, and it leaves the tab order and the accessibility tree:

```scss
    &__on-a,
    &__on-b {
        visibility: hidden;
        opacity: 0;
        transition: opacity 250ms ease, visibility 0s linear 250ms;
    }
```

with the mirror image on the `:checked` rule at line 318. The `pointer-events` pair on `__play` can then go.

products/conversations/frontend/emptyState/supportSetupLogic.ts
```diff
@@ -0,0 +1,128 @@
...
+        ],
+    })),
+    listeners(({ actions, values, cache }) => ({
+        checkSetup: () => {
+            // No team loaded yet: leave the status at `loading` and let the
+            // `currentTeam` subscription re-run this once it arrives.
+            if (!values.currentTeam) {
+                return
+            }
+            if (!values.currentTeam.conversations_enabled) {
```
haacked: `blocking`: A team that turns support off loses its ticket history, and the settings copy promises the opposite. This short-circuits to `needs-setup` on `conversations_enabled` without ever counting tickets, so `/support/tickets` renders the setup screen instead of the inbox. [The toggle says "Existing tickets stay accessible"](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/conversations/frontend/settings/GeneralSection.tsx#L75), and the tickets are still there, they're just unreachable from the only list view.

Count first, and let the toggle decide only between the two empty outcomes:

```diff
         checkSetup: () => {
             if (!values.currentTeam) {
                 return
             }
-            if (!values.currentTeam.conversations_enabled) {
-                actions.setDetectedStatus('needs-setup')
-                return
-            }
             actions.loadTicketCount()
         },
         loadTicketCountSuccess: () => {
             if (values.ticketCount && values.ticketCount > 0) {
                 actions.setDetectedStatus('has-data')
                 cache.disposables.dispose('poll')
+            } else if (!values.currentTeam?.conversations_enabled) {
+                actions.setDetectedStatus('needs-setup')
+                cache.disposables.dispose('poll')
             } else {
                 actions.setDetectedStatus('waiting-for-data')
             }
         },
```

This flips the assertion at `supportSetupLogic.test.ts:29`, which currently asserts the query is skipped when conversations are disabled.

products/ai_observability/frontend/emptyState/LLMPromptPreview.scss
```diff
@@ -0,0 +1,384 @@
...
+        margin-left: auto;
+    }
+
+    &__chip {
+        justify-self: end;
+        padding: 0 0.375rem;
+        font-family: var(--font-mono);
+        font-size: 0.625rem;
+        font-weight: 700;
+        color: var(--prompt-preview-accent);
```
haacked: `blocking`: In light mode the "production" chip is close to invisible. It's painted in `--color-product-llm-prompts-light` (`rgb(196 181 253)`) on the white card, which is about 1.8:1 against the 4.5:1 AA floor, and the same token colors the code keywords, the fetched version, and "serving v4". A low-vision reader can't see which version is live, which is the whole point of the widget.

Skills is worse at 1.85:1 for its command line and radio dots, flags is 2.70:1 for the rollout label and the `: true` console value, and web scripts is 2.9:1. All of them clear AA in dark mode, so it's only the light theme. For comparison the MCP reference accent lands at 4.5:1.

These tokens are sidebar icon tints, not text colors. Either darken the light token (`rgb(124 58 237)` clears 5.7:1 and keeps the family), or keep the pastel for fills and strokes and derive a text-safe variant:

```scss
--prompt-preview-accent-text: color-mix(in oklab, var(--prompt-preview-accent) 55%, var(--color-text-primary));
```

Worth checking the previews whose accent only touches fills and borders, since those are fine as they are.

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
```diff
@@ -0,0 +1,101 @@
...
+            teamLogic,
+            ['currentProjectId'],
+        ],
+    })),
+    loaders(({ values }) => ({
+        topicCount: [
+            null as number | null,
+            {
+                loadTopicCount: async (_: void, breakpoint): Promise<number> => {
+                    const response = await userInterviewTopicsList(String(values.currentProjectId), { limit: 1 })
```
haacked: `blocking`: A project with interview responses but no topics loses the whole scene. Interviews outlive topics: [`UserInterview.topic` is `SET_NULL`](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/user_interviews/backend/models.py#L44), so deleting a topic through the shipped `user_interview_topics_destroy` MCP operation orphans its responses, and audio uploads set `topic` read-only so they never have one at all. Those responses stay searchable through the scene's search box, which is the only UI for them, and `skippable: false` makes the swap permanent.

Count both:

```ts
const [topics, interviews] = await Promise.all([
    userInterviewTopicsList(String(values.currentProjectId), { limit: 1 }),
    userInterviewsList(String(values.currentProjectId), { limit: 1 }),
])
breakpoint()
return topics.count + interviews.count
```

Or drop `skippable: false` if you'd rather keep the probe as is.

products/endpoints/frontend/emptyState/endpointsEmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+        accentColorDark: 'var(--color-product-endpoints-dark)',
+        hedgehog: HedgehogCodeBubble,
+        text: {
+            'needs-setup': {
+                headline: 'Turn a SQL query into an API endpoint',
+                lead: 'Save a query as a named endpoint and call it from any application over the API. Update the query without changing your integration, and monitor every execution: bytes read, CPU usage, and duration. Materialize an endpoint to precompute its results and serve them fast.',
+            },
+        },
+        primaryAction: { label: 'Create your first endpoint', to: urls.sqlEditor({ source: 'endpoint' }) },
+        skippable: false,
```
haacked, line 31: `blocking`: `skippable: false` takes away two things a zero-endpoint project can do today. The Usage tab is on this same scene, so a project that soft-deleted its endpoints can no longer review historical cost and usage. And [the "Insight-based endpoint" path](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/products/endpoints/frontend/newEndpointMenu.tsx#L33) is only reachable from the scene's New button, while this config's single action goes to the SQL editor, so the most likely first-time path is the one that disappears.

```diff
         primaryAction: { label: 'Create your first endpoint', to: urls.sqlEditor({ source: 'endpoint' }) },
-        skippable: false,
         docsUrl: 'https://posthog.com/docs/endpoints',
```

frontend/snapshots.yml
```diff
@@ -5990,18 +6038,6 @@ snapshots:
...
-        hash: v1.k794b7964.2674e373d1160fbbfdb9a19a9c58784d5844a82b63fc71d32a614036faecbe9d.q9GyBN3PQJc2iUmR8pSQyvrNDZEaNSoEN-3I0cUNDYk
-    scenes-app-features--new-feature-flag--dark:
-        hash: v1.k794b7964.63525b4477e45f45dd5f6d3899ea40fbcddbe0740984e77ce85fdc30fda6889e.1ThZLV0EP8iSDYLCz1SRoPAAhA-vxBD8nY8Y-jL-2UA
-    scenes-app-features--new-feature-flag--light:
-        hash: v1.k794b7964.ce770bba480ba157f69520a090453521d4f07516be3aad472c7129d06c2737e6.XaEt155WvjsXhFH7EvO6bb6xGgeu0DeDz-B8u8L9Ymg
-    scenes-app-features--not-found-early-access--dark:
-        hash: v1.k794b7964.5906be874698c680dc921921cd28097b9567d94edccd183b4bec1bd20ff8ef5a.wh3QTvGjeN9sTOwhCa3ywx6DkOzDvEf2uX7cimZ6EoQ
-    scenes-app-features--not-found-early-access--light:
-        hash: v1.k794b7964.7c4c60522de8d6fd84b4334d53ed790c4094d4023de4d5defc07b613281bec68.zLhf-ChR2B4M8Gd-vTYGILy2agvAsBni1SFLC81icb0
     scenes-app-health-tables--data-modeling-default--dark:
```
haacked, line 6347: `blocking`: Three stories were deleted here rather than re-baselined, and two of them have nothing to do with this change. `EarlyAccessFeatures.stories.tsx` went in its own commit, and `NewFeatureFlag` and `NotFoundEarlyAccess` point at `/early_access_features/:id`, which routes to a different scene with no `emptyState` at all. The creation form and the 404 path lose their visual coverage. Early access is also the only adopted product whose scene stories were removed; links, skills and prompts all kept theirs and just took new hashes.

The list story broke because it mocks the legacy hyphenated `early-access-feature` path while the detection logic calls the generated underscored one, so it fell through to the default empty handler and detected `needs-setup`. Restoring the file and adding the path the logic actually uses puts it back on the `has-data` branch:

```ts
'/api/projects/:team_id/early_access_feature/': {
    count: 2,
    results: EARLY_ACCESS_FEATURE_RESULT as any[],
    next: null,
    previous: null,
},
```

CI is already reporting this product's Playwright spec failing, so dropping the visual coverage alongside it leaves the scene with less protection than before.

frontend/src/lib/components/ProductEmptyState/ProductEmptyStateGate.tsx
```diff
@@ -49,7 +49,9 @@ function ProductEmptyStateGateInner({ emptyState, children }: ProductEmptyStateG
     const { unskipEmptyState } = useActions(setupLogic)
     const { featureFlags } = useValues(featureFlagLogic)
 
-    if (skipped) {
+    // A lingering local skip is ignored for non-skippable products (the button may have
+    // been shown before the product opted out of skipping).
+    if (skipped && config.skippable !== false) {
```
haacked: `suggestion`: This guard doesn't do what its comment says, and it removes the only way back. Opting out of the skip branch here doesn't help, because `showEmptyState` at line 86 comes from a selector that is still `!skipped && (...)`. With `skipped` true that's false too, so line 93 renders the bare scene: an empty list, no setup screen, and no "Set up X" reminder banner. With `skippable: false` there's no skip button left to clear the stored flag either. That's strictly worse than before the guard existed, since the skip branch at least rendered the reminder.

Nobody can reach this today, since no product has shipped skippable and then opted out. It fires the first time one does, which is the case the comment is written for. Derive the decision once instead of mixing the local guard with the skip-aware selector:

```diff
-    if (skipped && config.skippable !== false) {
+    const skipHonored = skipped && config.skippable !== false
+    if (skipHonored) {
@@
-    if (showEmptyState) {
+    if (!skipHonored && (status === 'needs-setup' || status === 'waiting-for-data')) {
```

`showEmptyState` can then come out of the `useValues` destructure, since the gate is its only consumer.

products/visual_review/frontend/scenes/VisualReviewIndexScene.tsx
```diff
@@ -2,55 +2,43 @@ import { useValues } from 'kea'
...
 
     useEffect(() => {
-        if (!reposLoading && repos.length >= 1) {
+        if (reposLoading) {
+            return
+        }
+        if (repos.length >= 1) {
             router.actions.replace(urls.visualReviewRepoRuns(repos[0].id))
+        } else {
+            router.actions.replace(urls.visualReviewSettings())
```
haacked: `suggestion`: Any failure loading the repo list now sends the user to Settings as though they had no repos. `loadReposFailure` leaves `repos` at `[]` while `reposLoading` drops to false, so this branch can't tell "no repos connected" from "the request failed". A team with three repos connected lands on the connect-a-repo page with no error, no retry, and no way back, since `replace` drops `/visual_review` from history too. The inline copy this replaced at least said why you were there.

Forwarding from the loader's success action fixes it and matches how the rest of the codebase routes (`productToursLogic.ts:479`, `surveyLogic.tsx:2021`, `featureFlagLogic.ts:3474` all route from a `load*Success` listener). It also puts the forward back in the logic, where this PR deleted it from:

```ts
listeners(({ values }) => ({
    loadReposSuccess: () => {
        router.actions.replace(
            values.repos.length >= 1
                ? urls.visualReviewRepoRuns(values.repos[0].id)
                : urls.visualReviewSettings()
        )
    },
})),
```

The scene then drops the effect and can show a real failure state instead of a skeleton that never resolves.

products/product_tours/frontend/emptyState/productToursSetupLogic.ts
```diff
@@ -0,0 +1,140 @@
...
+            actions.setDetectedStatus(tourCount !== null && tourCount > 0 ? 'has-data' : 'needs-setup')
+        },
+        loadTourCountFailure: () => {
+            // Never strand the gate on its spinner: if nothing has answered yet, fail
+            // open to the real scene. Don't downgrade an existing answer on a blip.
+            if (values.setupStatus === 'loading') {
+                actions.setDetectedStatus('unknown')
+            }
+        },
+        createFirstTour: async () => {
```
haacked: `suggestion`: Tours created here won't show up in the `product tour created` event stream, because this is a second copy of a creation path that already exists. [`productToursLogic.createTour`](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/frontend/src/scenes/product-tours/productToursLogic.ts#L515) makes the same POST with the same `createDefaultTourContent()`, registers the same product intent and pushes to the same editor URL, and it also calls `reportProductTourCreated` and reloads the list. So the funnel measuring first-tour creation misses exactly the cohort this screen exists to convert.

On the premise: routing to the editor after a POST is right, it's what the scene's modal already does. But "the creation modal can't render behind the gate" describes where the modal is mounted, not a constraint; `NewProductTourModal` is fully props-driven, so a `PrimaryAction` component could render its own instance if you want the name prompt back. This file already imports that module for `createDefaultTourContent`, so dispatching the existing action adds no import edge:

```diff
     connect(() => ({
         actions: [
             productSetupStatusLogic({ productKey: ProductKey.PRODUCT_TOURS }), ['setDetectedStatus'],
+            productToursLogic, ['createTour'],
         ],
     })),
```

That also drops the local `creatingFirstTour` flag, so put the in-flight guard on `createTour` instead. The scene's modal submit needs it just as much and doesn't have it today.

products/links/frontend/emptyState/LinkPreview.scss
```diff
@@ -0,0 +1,381 @@
...
+    &__swap {
+        display: grid;
+    }
+
+    &__swap > * {
+        grid-area: 1 / 1;
+        min-width: 0;
+    }
+
+    &__on-a,
```
haacked, line 39: `suggestion`: A screen reader reads both halves of every crossfade at once, because `opacity: 0` hides an element visually but leaves it in the accessibility tree. The flags preview announces "0% 100%", ": false : true", "3.1% 4.6%" and both checkout CTAs together; the endpoints hero toggle gets the accessible name "Inline, Materialized, checkbox, not checked" so a screen reader user can't tell which state it's in; support's counter reads "3 tickets 4 tickets". The demo's whole payoff is that exactly one state is live, and that's the thing assistive tech is told is untrue.

The `visibility` change proposed on the play-button comment fixes this at the same time, and it's the same edit in `FeatureFlagPreview.scss:37`, `EndpointPreview.scss:37`, `SupportPreview.scss:37`, `LLMSkillPreview.scss:22`, `EarlyAccessFeaturePreview.scss:37` and `ProductTourPreview.scss`. Worth folding into `SKILL.md` next to the existing `:focus-visible` rule, since every preview inherits the pattern.

One place it isn't needed: `UserInterviewPreview` already gives its hidden radios accessible names through real `<label htmlFor>` associations.

products/links/frontend/emptyState/LinkPreview.scss
```diff
@@ -0,0 +1,381 @@
+.LinkPreview {
+    --link-preview-accent: var(--empty-state-accent, var(--color-primary-500));
```
haacked, line 4: `suggestion`: `accentColorDark` never reaches any preview. This binds to `--empty-state-accent` only, and `--empty-state-accent-dark` is read by nothing except two Tailwind utilities on the shell, so in dark mode the preview interior paints with the light token while the eyebrow icon and pulse dot right above it use the dark one. Links renders `rgb(143 7 255)` inside the card and `rgb(178 92 255)` in the header, visibly different shades. Support is starker: red ticket rows and Send button under an amber icon.

It also costs contrast where it matters. The selected-row border and radio dot are the only indication of which link is picked, and the light token lands at 2.69:1 on the dark card against the dark token's 4.34:1, so a dark-mode user can't tell the preview responded to their click.

One line on the platform side covers all thirteen previews including the pre-existing MCP one, instead of eleven per-preview overrides. On the preview column in `ProductEmptyState.tsx`:

```diff
-    className="hidden min-w-0 flex-col justify-center gap-3 p-10 md:flex rounded-md border border-primary"
+    className="hidden min-w-0 flex-col justify-center gap-3 p-10 md:flex rounded-md border border-primary dark:[--empty-state-accent:var(--empty-state-accent-dark)]"
```

Every preview keeps reading a single variable.

products/feature_flags/frontend/emptyState/FeatureFlagPreview.scss
```diff
@@ -0,0 +1,460 @@
+.FlagPreview {
```
haacked, line 3: `suggestion`: These stylesheets don't ship in the scene's lazy chunk, they ship in the app's single boot-time stylesheet, so every PostHog user downloads and parses them on cold load even though only a project with zero entities renders one. [The build resolves exactly one CSS output for the entry](https://github.com/PostHog/posthog/blob/4b57f7106f3940c494d488673f5e536a3d49596b/frontend/build.mjs#L184) and emits it as a render-blocking stylesheet, and esbuild has no CSS code splitting, so SCSS reachable through a dynamic scene import lands there. The eleven new previews add about 106 KiB of raw SCSS; the whole app was 676 KiB across 243 files before this PR.

Worth knowing because it changes the answer on the CI bundle warning: lazy-loading the preview components moves the JS half (a few KiB per scene chunk) and none of this.

About half of each file is the same boilerplate under a different BEM prefix. The chrome trio is byte-identical in 9 files, the sparkline group in 8, the swap grid in 9, the reduced-motion and static guards in all 12. Lifting those into one partial under `lib/components/ProductEmptyState/` and having each preview root carry both classes takes most of the weight out of the boot stylesheet and leaves each product's own artwork alone. They all already read `--empty-state-accent`, so there's no per-product variable plumbing.

products/conversations/frontend/emptyState/supportEmptyState.tsx
```diff
@@ -0,0 +1,39 @@
...
+            'needs-setup': {
+                headline: 'Answer support tickets with full product context',
+                lead: 'Collect tickets from an in-app chat widget, email, or Slack into one inbox. Every ticket shows the person behind it, with their events, session replays, and past tickets alongside the conversation. Set SLAs, assign owners, and trigger workflows on ticket events.',
+            },
+            'waiting-for-data': {
+                headline: 'Support is on. Waiting for your first ticket',
+                lead: 'Tickets from your chat widget, email, and any other connected channels will show up here as they arrive. Send a test message through the widget to see one land, or finish connecting a channel in settings.',
+            },
+        },
+        primaryAction: { label: 'Enable support', to: urls.supportSettings() },
```
haacked: `suggestion`: In `waiting-for-data` this screen contradicts itself: the headline says "Support is on. Waiting for your first ticket" while the button under it says "Enable support" and links to settings. That branch is only reachable when `conversations_enabled` is already true, so it's telling a team that just turned support on to go turn it on. `previewLabel: 'Your inbox, once enabled'` reads the same way.

`primaryAction` isn't mode-aware the way `text` is, and support is the only three-state adopter, so mode-neutral copy is the cheap fix:

```diff
-        primaryAction: { label: 'Enable support', to: urls.supportSettings() },
+        primaryAction: { label: 'Open support settings', to: urls.supportSettings() },
         docsUrl: 'https://posthog.com/docs/support',
-        previewLabel: 'Your inbox, once enabled',
+        previewLabel: 'Your inbox, once tickets arrive',
```

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
```diff
@@ -0,0 +1,101 @@
...
+                    breakpoint()
+                    return response.count
+                },
+            },
+        ],
+    })),
+    listeners(({ actions, cache, values }) => ({
+        loadTopicCountSuccess: () => {
```
haacked, line 92: `suggestion`: A failed first check hands the user the real scene and then takes it away twenty seconds later. `loadTopicCountFailure` correctly fails open to `unknown` so the gate renders the live topics list, but the poll keeps running behind it and this success handler pushes its answer unconditionally. The next successful poll on a project with zero topics calls `setDetectedStatus('needs-setup')` and swaps the mounted scene for the full-page setup screen, losing anything typed into the search box, with no way back because `skippable: false`.

All eleven logics carry the never-downgrade guard, but only in the failure listener. Only the two that re-check while mounted can violate it from the success path, and `supportSetupLogic.ts:100` has the same shape.

```diff
         loadTopicCountSuccess: () => {
             const hasData = (values.topicCount ?? 0) > 0
             if (hasData) {
                 cache.disposables.dispose('poll')
             }
-            actions.setDetectedStatus(hasData ? 'has-data' : 'needs-setup')
+            // Detection already failed open and the real scene is on screen; only a
+            // positive answer should move the gate from there.
+            if (hasData || values.setupStatus !== 'unknown') {
+                actions.setDetectedStatus(hasData ? 'has-data' : 'needs-setup')
+            }
         },
```

The durable version is to make `setDetectedStatus` refuse to move a settled status back to `needs-setup` within a session, so the next product that polls gets it for free.

products/feature_flags/frontend/emptyState/featureFlagsSetupLogic.ts
```diff
@@ -0,0 +1,88 @@
...
+        },
+        loadFlagCountFailure: () => {
+            // Never strand the gate on its spinner: if nothing has answered yet, fail
+            // open to the real scene. Don't downgrade an existing answer on a blip.
+            if (values.setupStatus === 'loading') {
+                actions.setDetectedStatus('unknown')
+            }
+        },
+    })),
+    afterMount(({ actions }) => {
```
haacked, line 95: `suggestion`: The gate's comment says statuses are preloaded at boot so the spinner "rarely renders in practice", but that's only true for products with a `setupProbe`, and `grep setupProbe products/*/manifest.tsx` returns only `mcp_analytics`. For all eleven adoptions the spinner is the normal path: the header paints, then the filters and table skeleton are replaced by a full-scene `SpinnerOverlay` for one round trip on every entry, including every trip back from a detail page. `productSetupStatusLogic` unmounts with the scene, so the answer is discarded and the next visit repeats it.

This probe isn't free either. `safely_get_queryset` materializes `Survey.get_internal_flag_ids` into a set and applies two `NOT IN` exclusions to both the count and the page query plus five prefetches, so it's roughly 8 to 10 SQL statements on one of the busier scenes in the app, and it duplicates a count the scene's own list call already returns.

Remembering only the positive answer removes the spinner for projects that have data without ever flashing the empty state on a stale read, since a stale `has-data` can only render the real scene, which is already the fail-open default. That needs the platform to keep the status instance alive the way `productSetupPreloadLogic` does for probed products, so it's a chunk-1 change rather than a local one. Either way the gate's comment should stop claiming a preload these products don't get.

products/conversations/frontend/emptyState/supportSetupLogic.ts
```diff
@@ -0,0 +1,128 @@
...
+        },
+        loadTicketCountFailure: () => {
+            // Never strand the gate on its spinner: if nothing has answered yet, fail
+            // open to the real scene. Don't downgrade an existing answer on a poll blip.
+            if (values.setupStatus === 'loading') {
+                actions.setDetectedStatus('unknown')
+            }
+        },
+    })),
+    subscriptions(({ actions }) => ({
```
haacked, line 119: `suggestion`: Every mount of the support scene fires the ticket-count request twice. kea-subscriptions invokes a subscription once at mount whenever the value isn't `undefined`, and `currentTeam` defaults to `null`, so this runs `checkSetup` before the explicit call in `afterMount` runs it again. Both reach `loadTicketCount` for a team with conversations enabled, and the loader's `breakpoint()` just discards the first response.

```diff
     subscriptions(({ actions }) => ({
-        currentTeam: () => actions.checkSetup(),
+        // kea-subscriptions fires once at mount; afterMount already ran checkSetup.
+        currentTeam: (_, previousTeam) => {
+            if (previousTeam !== undefined) {
+                actions.checkSetup()
+            }
+        },
     })),
```

This still fires on the `null` to team transition, so it keeps covering the case where the team arrives after mount.

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
```diff
@@ -0,0 +1,101 @@
...
+            if (values.setupStatus === 'loading') {
+                actions.setDetectedStatus('unknown')
+            }
+        },
+    })),
+    afterMount(({ actions, cache }) => {
+        actions.loadTopicCount()
+        // Topics are created by PostHog AI in the side panel while the gate stays on
+        // screen, so poll until one lands rather than waiting for a remount.
+        cache.disposables.add(() => {
```
haacked, line 113: `suggestion`: This poll has no stop condition other than a topic appearing, so someone who opens PostHog AI, closes the panel without creating anything and leaves the tab open keeps issuing a request every 20 seconds for as long as the scene is open. That's roughly 1,440 over a workday-length session, every one answering zero. It also keeps retrying after the endpoint fails: `loadTopicCountFailure` publishes `unknown` and the gate falls through to the real scene, but nothing disposes the poll, so a 500 becomes an indefinite retry loop underneath a scene that's already loading topics on its own.

Support doesn't have either problem, because its timer calls `checkSetup()`, which returns before any network call when the product is off. Starting the poll from the action it exists for gets the same property, using the `findMounted()` pattern product tours already uses. Capping it at a fixed number of ticks and letting a remount re-arm it is the cheaper version.

The interval is also inlined as `20_000` here, while both other pollers name it `POLL_INTERVAL_MS`, so anyone retuning the cadence by grep will miss this one.

frontend/src/scenes/experiments/Experiments.tsx
```diff
@@ -203,7 +198,7 @@ const ExperimentsTable = ({
     openSurveyModal: (experiment: Experiment) => void
     openCopyToProjectModal: (experiment: Experiment) => void
 }): JSX.Element => {
-    const { currentProjectId, experiments, experimentsLoading, tab, shouldShowEmptyState, filters, count, pagination } =
+    const { currentProjectId, experiments, experimentsLoading, filters, count, pagination } =
```
haacked, line 204: `nit`: Dropping `shouldShowEmptyState` from this destructure leaves `experimentsLogic`'s selector with no production caller, so a repo-wide grep now finds it only in its own definition and in a green test that guards nothing. The next person touching experiments filtering will read that test as saying the empty-state decision still lives there, when it moved to `experimentsSetupLogic`. `llmPromptsLogic` in this same PR deletes exactly this pair, so the cleanup shape is already set. `featureFlagsLogic.shouldShowEmptyState` is still live, so this is experiments only.

products/user_interviews/frontend/emptyState/NewTopicButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+    })
+
+    return (
+        <LemonButton
+            type="primary"
+            icon={<IconSparkles />}
+            className="self-start"
+            data-attr="user-research-empty-state-new-topic"
+            onClick={() => openMax?.()}
+            disabledReason={openMax ? undefined : 'PostHog AI is unavailable here'}
```
haacked, line 31: `nit`: This `disabledReason` can never fire, so on an instance without PostHog AI the button stays enabled and clicking it opens a side panel showing `MaxNotConfigured`. `useMaxTool` returns a null `openMax` only when `active` is false, and this call site leaves `active` at its default `true`. Passing `active: isMaxAvailable` from `maxGlobalLogic` would make the guard mean what it reads like. The same inert guard already ships on the scene's button, so this isn't new, but it matters a little more here because `skippable: false` and there's no `docsUrl`, making this the only control on the screen.

.agents/skills/building-product-empty-states/SKILL.md
```diff
@@ -49,17 +49,24 @@ Statuses: `loading` (not yet known - the gate holds a spinner, never flashes the
...
-- **Wizard vs primary action**: SDK-installed products set `wizard: { slug }` (the slug must exist in `@posthog/wizard`); creation-first products (flags, surveys) set `primaryAction` instead. Self-hosted degrades automatically: no cloud → the terminal hides and the manual path is promoted.
+- **Wizard vs primary action**: SDK-installed products set `wizard: { slug }` (the slug must exist in `@posthog/wizard`); creation-first products (flags, surveys) set `primaryAction` instead. Self-hosted degrades automatically: no cloud → the terminal hides and the manual path is promoted. If the create action needs hooks (e.g. it opens PostHog AI via `useMaxTool`, like user research's "New topic"), provide a `PrimaryAction` component instead of `primaryAction` - it renders in the same slot and takes precedence. If the create action needs hooks (e.g. it opens PostHog AI via `useMaxTool`, like user research's "New topic"), provide a `PrimaryAction` component instead of `primaryAction` - it renders in the same slot and takes precedence.
```
haacked: `nit`: The `PrimaryAction` sentence is pasted twice in a row in this bullet. Delete the second copy.

haacked (commented):
Nice platform work, and the detection logics are consistent. Main thing to look at before merge: `skippable: false` plus a count that's narrower than what the scene renders locks people out of scenes they have data in, in about five products. The two red Playwright specs are that same mechanism, not flakes.

products/feature_flags/frontend/emptyState/featureFlagsEmptyState.tsx
rafaeelaudibert: Fixed by 34a1a5e5b66. Agreed on the cause, took a different fix: the empty state now opens the template selector like the scene's New button, and carries `new-feature-flag`. Spec passes unchanged.

products/experiments/frontend/emptyState/experimentsEmptyState.tsx
rafaeelaudibert: Fixed by f36fbc9eb57. Agreed on the archived lockout. Kept `skippable: false` and made detection sum a live and an archived count instead, so an all-archived project keeps its scene.

products/early_access_features/frontend/emptyState/earlyAccessFeaturesEmptyState.tsx
rafaeelaudibert: Fixed by 34a1a5e5b66. Agreed. Used the shared `create-feature` attr both surfaces now carry rather than matching two labels, so one selector works whichever renders. CI green.

products/experiments/frontend/emptyState/experimentsEmptyState.tsx
rafaeelaudibert: Fixed by b40fadcbe50. Agreed, and took the config route: `primaryAction.accessControl` plus `dataAttr`, applied to the six products whose scene button gates today. Test covers viewer vs editor.

products/cdp/frontend/emptyState/webScriptsSetupLogic.ts
rafaeelaudibert: Fixed by 83452ef00a7. Agreed, added the request. Put it on `api.pipelineFrontendAppsConfigs.list` rather than a hardcoded URL, and the count now sums both sources.

products/links/frontend/emptyState/LinkPreview.tsx
rafaeelaudibert: Fixed by 2fe6ac90b7c. Agreed. Took the `visibility` fix into shared mixins and applied it to all ten crossfading previews, then dropped the redundant `pointer-events` pair.

products/conversations/frontend/emptyState/supportSetupLogic.ts
rafaeelaudibert: Fixed by 269bca420af. Agreed, this was a real hole. Count always runs now; the toggle only picks between the two empty outcomes. Flipped the test that asserted the old skip.

products/ai_observability/frontend/emptyState/LLMPromptPreview.scss
rafaeelaudibert: Fixed by 6946dc21677. Agreed. Derived a text-safe variant at 60% toward `--color-text-primary`, applied to all 29 `color:` sites. Note prompts and skills were already fixed by the token consolidation on master.

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
rafaeelaudibert: Fixed by 83452ef00a7. Agreed, counted both. Orphaned responses and audio uploads are exactly the cases that would have lost the scene.

products/endpoints/frontend/emptyState/endpointsEmptyState.tsx
rafaeelaudibert: Fixed by 0bed0580b4c. Agreed the insight route disappeared. Kept the gate and reproduced the scene's New button instead, dropdown and picker modal included, so both routes work.

frontend/snapshots.yml
rafaeelaudibert: Fixed by ead4300bee3. Agreed. Restored the file and all three snapshot entries; master had already corrected the mock path to the underscored one. Verified the gate reaches `has-data`.

frontend/src/lib/components/ProductEmptyState/ProductEmptyStateGate.tsx
rafaeelaudibert: Fixed by 5df95269061. Agreed, and took your derivation: one `skipHonored`, and the now-unused `showEmptyState` selector is gone rather than left dead.

products/visual_review/frontend/scenes/VisualReviewIndexScene.tsx
rafaeelaudibert: Fixed by c4d917ae111. Agreed. Forward moved into the logic's success listener, and the scene now shows a real error with a retry instead of a skeleton that never resolves.

products/product_tours/frontend/emptyState/productToursSetupLogic.ts
rafaeelaudibert: Fixed by c4d917ae111. Agreed on the missing event. Dispatches the shared `createTour` now, and the in-flight guard sits on that action, so the scene's modal gets it too.

products/links/frontend/emptyState/LinkPreview.scss
rafaeelaudibert: Fixed by 2fe6ac90b7c. Agreed. Same mixins as the play-button fix, applied across ten previews. Experiments already swapped `display`, so it was never affected. Rule documented in SKILL.md.

products/links/frontend/emptyState/LinkPreview.scss
rafaeelaudibert: Fixed by 2fe6ac90b7c. Agreed, and used your one-liner on the preview column, so all thirteen previews read a single variable.

products/feature_flags/frontend/emptyState/FeatureFlagPreview.scss
rafaeelaudibert: Partly fixed by 0bb822f1dc9. Measured your bundle claim: 74 KB of the 1.23 MB boot stylesheet. Only ~7 KB was identical though, not half, so the rest needs CSS splitting.

products/conversations/frontend/emptyState/supportEmptyState.tsx
rafaeelaudibert: Fixed by c4d917ae111. Agreed, the screen contradicted itself. Took your mode-neutral copy for both the button and the preview caption.

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
rafaeelaudibert: Fixed by 5df95269061. Agreed, and went with the durable version you suggested: `setDetectedStatus` refuses to walk a settled `unknown` or `has-data` back, so both pollers get it.

products/feature_flags/frontend/emptyState/featureFlagsSetupLogic.ts
rafaeelaudibert: Fixed by 5df95269061. Agreed the comment was wrong and corrected it. Left the status caching out: it is a real change but belongs in its own PR.

products/conversations/frontend/emptyState/supportSetupLogic.ts
rafaeelaudibert: Fixed by 5df95269061. Agreed, took your guard verbatim. Support is the only adopter with a subscription, so nothing else was double-fetching.

products/user_interviews/frontend/emptyState/userInterviewsSetupLogic.ts
rafaeelaudibert: Fixed by 5df95269061. Agreed. Poll now stops on failure, caps its ticks, and names the interval `POLL_INTERVAL_MS` like the other two.

frontend/src/scenes/experiments/Experiments.tsx
rafaeelaudibert: Fixed by b40fadcbe50. Agreed. Deleted the selector, its type entries, the vacuous test, and the orphaned `objectsEqual` import. Flags' copy is still live and untouched.

products/user_interviews/frontend/emptyState/NewTopicButton.tsx
rafaeelaudibert: Fixed by c4d917ae111. Agreed. Passing `active: isMaxAvailable` now. Also fixed the scene's button, since leaving one inert and one working would be worse.

.agents/skills/building-product-empty-states/SKILL.md
rafaeelaudibert: Fixed by b40fadcbe50. Agreed, duplicate sentence deleted.

PR #74467, opened 2026-07-29
fix(web-analytics): harden per-shape warming logs

lricoy:
## Problem

Follow-up to #74375, which merged before its review-fix commit landed (the branch auto-deleted on merge and the fix re-created it orphaned). Three review findings are therefore live in prod unhardened:

1. The per-shape log extracts `dateRange` non-defensively — a malformed shape raises *outside* the worker's failure boundary and aborts the whole op instead of skipping the shape.
2. The ~150k-line-per-pass INFO burst goes to **stderr** (root console handler) — stream-based collectors can relabel it and it risks backpressure.
3. The slow-shape WARNING timed only `runner.run`, missing shapes delayed by construction/cache lookups.

## Changes

- Defensive log call (isinstance checks + try/except — observability can never abort the pass).
- Route `products.web_analytics.dags.cache_warming` INFO to stdout (`console_stdout_info` + `console_stderr_warning`), same pattern as the hypercache verifier.
- Single per-shape event that escalates to WARNING on the **full** wall-clock, with the replayed range and cold-state stashed from the inner pass (`replay_date_from`, `was_cold`).

- `hogli test products/web_analytics/dags/tests/test_cache_warming.py` — 59 pass. Observability-only; no outcome changes.

## 🤖 Agent context

Claude (Claude Code) authored the fixes in response to review-bot findings on #74375; the merge raced the fix push, so this lands them separately.

### DIFF
--- a/posthog/settings/logs.py
+++ b/posthog/settings/logs.py
@@ -164,6 +164,13 @@ LOGGING: dict[str, Any] = {
             "handlers": ["console_stdout_info", "console_stderr_warning"],
             "propagate": False,
         },
+        # The web analytics warmer logs one line per shape per pass (up to the
+        # 400k selection cap) — keep the INFO burst off stderr for the same reason.
+        "products.web_analytics.dags.cache_warming": {
+            "level": "INFO",
+            "handlers": ["console_stdout_info", "console_stderr_warning"],
+            "propagate": False,
+        },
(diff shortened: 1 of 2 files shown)


lricoy:
Folding into #74471 — its branch already carries this commit, so retargeting that PR to master lands both in one merge instead of two serial CI cycles.

PR #74472, opened 2026-07-29
fix(sdk-health): show the status tooltip when the badge is clicked

slshults:
## Problem

In SDK Health, users click the coloured status badge ("Outdated", "Current", "Recent") expecting something to happen, and nothing does. Two separate things were in the way:

1. PostHog's `Tooltip` is built on Base UI, whose hover interaction is configured `mouseOnly: true`. On a touch device the tooltip is simply unreachable, so a tap gives you nothing at all.
2. Base UI's tooltip trigger defaults to `closeOnClick`, so on desktop, clicking a badge you were already hovering *dismissed* the tooltip you were reading.

This was reported as [Make status tag clickable or remove cursor-help affordance]. An earlier PR (#58957) tried to fix it by moving the version `LemonMenu` to cover the whole cell, so clicking the badge opened the Activity page / SQL query menu. That was the wrong behaviour: all the badge needs to do on click is show the same tooltip it shows on hover. That PR was closed for inactivity.

## Changes

`frontend/src/lib/lemon-ui/Tooltip/Tooltip.tsx` gains an opt-in `openOnClick` prop:

- passes `closeOnClick={false}` to the Base UI trigger, so a click no longer dismisses
- opens via the component's existing `handleOpenChange`, rather than force-controlling `visible`

That second point matters. Driving `visible` from the call site would make the tooltip fully controlled, which swallows Base UI's own `useDismiss`, and the tooltip becomes undismissable by Escape or by pressing outside. Routing through the existing open state keeps every dismissal path working. `openOnClick` defaults to `false`, so no existing call site changes behaviour (`closeOnClick`'s Base UI default is already `true`).

`SdkHealthComponents.tsx` extracts the three repeated badge branches into one `SdkVersionStatusTag` that uses `openOnClick`, and adds `tabIndex={0}` so keyboard users can reach the status reason too (Base UI's trigger opens the tooltip on keyboard focus).

Clicking only opens, it does not toggle closed. That is deliberate: making click toggle would reintroduce the original complaint, where clicking a badge you are hovering snatches the tooltip away. Dismissal is Escape, a press outside, or hovering away.

I am an agent (PostHog Code). Unit tests plus real-browser verification.

**Jest, 9 new tests:** click opens; click does nothing when the prop is off; hover still opens both with and without the prop; Escape dismisses; press-outside dismisses; the badge is focusable. I mutation-checked each one, and deleted one test that passed against a deliberately broken implementation rather than leave false confidence. jsdom cannot trigger Base UI's pointerdown dismiss at all, so the "click while hovering does not dismiss" half is untestable in jest, and there is a comment in the test file saying so.

**Playwright against Storybook**, which is where that untestable half got covered:

| check | result |
| --- | --- |
| desktop: click opens the tooltip | pass |
| desktop: Escape dismisses | pass |
| desktop: hover still opens | pass |
| desktop: click while already hovering keeps it open, no blink | pass |
| desktop: press outside dismisses | pass |
| keyboard: Tab lands on the badge and reveals the reason | pass |
| touch (Pixel 7 emulation): tap opens the tooltip | pass |
| touch: tap elsewhere dismisses | pass |

The touch case is the one that was completely broken before, so it is the one worth re-checking by hand on a real device.

Also ran: `oxlint` clean, `tsgo --noEmit` clean, and the full `frontend/src/lib/lemon-ui` suite (197 tests) green, since this touches a shared component.

**Known limitation, not introduced here:** Base UI's tooltip does not wire `aria-describedby` between trigger and popup, so the reason text is not in the accessibility tree for screen readers on any PostHog tooltip. Making the badge focusable is an improvement, not a complete fix. Worth a separate look at the shared `Tooltip` if we care.

Edit: I'm a human, and I tested in the UI on my local. Works as we hope, and other tooltips still work as before. 

## Publish to changelog?

no

## Docs update

No docs change required.

---

### DIFF
--- a/frontend/src/lib/lemon-ui/Tooltip/Tooltip.test.tsx
+++ b/frontend/src/lib/lemon-ui/Tooltip/Tooltip.test.tsx
@@ -1,0 +1,83 @@
+import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import { Tooltip } from './Tooltip'
+
+const TITLE = 'Released 2 years ago. Upgrade recommended.'
+
+function renderTooltip(props: { openOnClick?: boolean } = {}): void {
+    render(
+        <div>
+            <Tooltip title={TITLE} delayMs={0} {...props}>
+                <span>Outdated</span>
+            </Tooltip>
+            <button>Elsewhere</button>
+        </div>
+    )
+}
+
+describe('Tooltip', () => {
+    afterEach(() => {
+        cleanup()
+    })
+
+    it('opens on hover', async () => {
+        renderTooltip()
+
+        fireEvent.pointerEnter(screen.getByText('Outdated'), { pointerType: 'mouse' })
+        fireEvent.mouseEnter(screen.getByText('Outdated'))
+
+        expect(await screen.findByText(TITLE)).toBeTruthy()
+    })
+
+    it('does not open on click by default', async () => {
+        renderTooltip()
+
+        fireEvent.click(screen.getByText('Outdated'))
+
+        await waitFor(() => expect(screen.queryByText(TITLE)).toBeNull())
+    })
+
+    it('opens on click when openOnClick is set', async () => {
+        renderTooltip({ openOnClick: true })
+
+        fireEvent.click(screen.getByText('Outdated'))
+
+        expect(await screen.findByText(TITLE)).toBeTruthy()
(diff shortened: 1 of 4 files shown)


rafaeelaudibert (approved):
Makes sense, did you test this locally just to make sure you didn't break other tooltips?

PR #74478, opened 2026-07-29
chore(personhog): etcd resilience 

z0br0wn:
## Problem

Every etcd interaction in the router's and pod's run loops treats failure as fatal: one broken watch stream, one failed write, one lost lease, and the process dies. Because an etcd blip severs every participant's connections at once, this couples data plane availability to control plane health fleet-wide, a few seconds of etcd unavailability restarts every router and every leader pod simultaneously, and every leader cache is dumped and rebuilt for an event that changed nothing about ownership. The reconcile ticks make this deterministic rather than connection-lucky, since they actively probe etcd every 5 seconds.

There is also no test anywhere that exercises coordination-store failure, so this entire failure class was invisible until it happened.

## Changes

**Run supervisors.** The router's and pod's `run()` are each split into a supervisor and a `run_once` attempt. A failed attempt tears down cleanly (tasks joined, drain lanes cancelled and joined, lease revoked or left to lapse by TTL), then the supervisor backs off exponentially (500 ms base, 15 s cap) and rebuilds through the same bootstrap that recovers a restarted process. The data plane keeps serving through the gap: routers from their last-known table with parked stashes intact, pods from their caches. A consecutive-failure budget (default 10, configurable, healthy attempts reset the count) keeps a genuinely wedged participant on the process-restart path, and `personhog_coordination_run_restarts_total{component}` counts every rebuild.

Serving while disconnected is safe for a structural reason, now documented in the code: ownership cannot move while etcd is unreachable (the coordinator cannot advance handoffs), and once etcd recovers, any handoff created before the participant re-registers excludes it from the freeze quorum, so the old owner fences before a new owner warms and stale forwards bounce into the existing drain machinery rather than landing.

**Lease loss still self-fences, now locally.** On a lost lease the pod releases every held partition through the handler before any retry (dropping cache and serving authority, purely locally, callable with no lease), then rejoins as a fresh participant and re-acquires ownership only through the normal warm path. If the local fence itself fails, a poisoned flag refuses in-place retry entirely and the process restart clears everything. The zombie-prevention property is unchanged; only the blast radius of enforcing it shrank.

  ```mermaid
  flowchart LR
      A[etcd blip] --> B[watch/write fails] --> C[process dies] --> D[fleet restarts, caches dumped]
      class C,D phRed;
      class A,B phGray;
  ```
  
  ```mermaid
  flowchart LR
      A[etcd blip] --> B[attempt fails, torn down] --> S[backoff, rebuild in place]
      S -->|recovered| R[re-register, converge, catch up]
      S -->|budget spent| P[process restart]
      B -.-> D[data plane keeps serving, caches intact]
      class S,R phBlue;
      class D phYellow;
      class A,B,P phGray;
  ```

**The `etcd_bounce` chaos scenario.** The bed's chaos executor gains a dedicated scenario that abruptly kills one etcd member of the healthy three-member cluster: quorum and state  survive, but every client stream pinned to that member breaks at once, which is exactly the failure this PR makes survivable. etcd deliberately lives outside the general kill-target  pool (a separate `ChaosConfig` slot), so the pair and staggered scenarios can never draw it; only the bounce, behind an all-members-ready quorum guard, may touch it. Enabled by  `CHAOS_ETCD_NAMESPACE`, excluded from the draw when unset. A companion charts change grants the harness the namespace RBAC and sets the env for dev. One ripple: the new CLI field  tipped clippy's large-variant lint on `Command`, fixed by boxing `Traffic` the way `Gate` already was.

  > [!WARNING]
  > Behavior change: coordination failures and lease loss no longer restart the process by themselves. The stall watchdog now fails an *attempt* (rebuild in place); process restart happens only past the retry budget. Tests that asserted death-on-failure were rewritten to the new contract, see below.

All automated. Suites green: coordination 42 lib, 52 etcd-backed integration, 30 protocol hardening; router 48 lib, 22 proxy, 15 stash integration; harness 17.
  
New and rewritten coverage, by the regression each catches:

  - `a_failed_coordination_attempt_rebuilds_in_place` (new): the supervisor itself. Handler failures kill an attempt, re-bootstraps fail while the fault persists, and recovery re-registers *and acks a handoff that arrived during the outage*, so catch-up is asserted, not assumed.
  - `pod_self_fences_locally_and_rejoins_after_lease_loss` (rewritten): warms a partition, revokes the lease out from under the pod, then asserts the full sequence: `Released` before any rejoin (the safety property, previously only implied by process death, now directly observed), re-registration, and a second warm as convergence re-acquires. Catches any regression that lets a pod retain partitions across lease loss.
  - `router_rejoins_with_a_fresh_lease_after_revocation` (rewritten): transient deregistration then rejoin, catching a supervisor that either never gives up its dead registration or never comes back.
  - `a_stalled_watch_loop_trips_the_watchdog_and_deregisters` (updated): pinned to the fatal path via a retry budget of 1, so the watchdog's escalation contract stays regression-tested even though its default behavior is now rebuild.
  - The chaos scenario's draw gating is covered by the existing exclusion test extended to the new scenario; the scenario's end-to-end behavior is deliberately validated in the bed after deploy (its acceptance signature: `run_restarts_total` spikes with zero container restarts, zero warms, zero violations), since etcd-member kills are not reproducible in the unit environment.

## 🤖 Agent context

Built with Claude Code; the `/writing-tests` skill was invoked for the test work. Decisions a reviewer should know about: an earlier, smaller plan (making the `StashHandler` trait infallible) was dropped because the supervisor subsumes its benefit and the fallible trait is the test-injection point the new integration tests rely on. The chaos scenario was deliberately built as a dedicated, quorum-guarded scenario rather than a fourth entry in the random kill pool, so compound-kill scenarios can never take etcd below quorum. The consecutive-failure budget resets after healthy attempts specifically so it bounds crash loops rather than accumulating over weeks of occasional blips. The safety argument for serving while disconnected (ownership frozen while etcd is down, freeze-quorum exclusion after recovery) was worked out during review discussion and is documented in the supervisor doc comments, not just here.


rust/personhog-coordination/src/pod.rs
```diff
@@ -228,118 +243,322 @@ impl PodHandle {
...
+            let lease_id = match self.begin_session().await {
+                Ok(lease_id) => lease_id,
+                Err(e) => {
+                    if cancel.is_cancelled() {
+                        return Ok(());
+                    }
+                    if !self.note_run_failure(&mut consecutive_failures, session_started, &e) {
+                        return Err(e);
+                    }
+                    if self.run_backoff(&cancel, consecutive_failures).await {
```
nickbest-ph, line 329: Does running backoffs with sleeps open us up to a large split brain window?

If connection between a leader pod and etcd is down and we end up with a 15 second nap, this pod's owning lease could expire, etcd could come back up, the coordinator could trigger the dead pod handoff scenario because the lease expired, then as routers shift over to the new pod, the fleet of routers will simultaneously write to old pod and new pod? old pod is backing off and won't find out that it doesn't own the lease until it wakes up?

the pod termination before used to act as a hard fence here

z0br0wn: good call out - fixing

rust/personhog-coordination/src/util.rs
```diff
@@ -33,15 +33,34 @@ pub async fn run_lease_keepalive(
...
+                // stream, so an unbounded await here would leave the
+                // component serving while its lease expires server-side.
+                // A round that cannot confirm renewal within one interval
+                // is therefore treated as lease loss — with the interval
+                // at a third of the TTL, detection lands at worst two
+                // thirds of the way in, leaving the final third for the
+                // self-fence.
+                let round = async {
+                    keeper.keep_alive().await?;
+                    match stream.message().await? {
```
nickbest-ph: any connection blip with the etcd member we are currently connected to via the stream results in telling the pod the lease is lost, and forcing it to re-warm everything. i think that etcd pod restarts might be fairly routine?

the lease is available/replicated to other etcd pods, should we attempt to retry making a stream against another pod in the case we know that we aren't anywhere near the TTL of the lease? as i think the only way the lease will be revoked in this scenario is if the TTL expires

rust/personhog-coordination/src/pod.rs
z0br0wn: Good catch — the kernel of this is real and is now fixed on the branch. Two corrections to the failure model first, then the fix.

Routers never write to the old and new owner simultaneously: the dead-owner handoff still runs the freeze phase, every registered router must freeze-ack (and stash) before the coordinator advances, and routers only flip to the new owner at Complete. Getting a dual write requires a router that also wedged and missed the freeze — that's the pre-existing double-zombie residual that the epoch fencing roadmap item addresses, and it's unchanged by this PR.

The hard fence also still exists — lease loss triggers a local self-fence that fences and drains in-flight writes before releasing anything (process termination used to drop in-flights mid-produce, so this is actually stronger), with a poison-to-process-restart fallback if the fence can't complete.

What you found is deferred detection feeding that fence: run_backoff was awaited outside the select that races the heartbeat, so a lease lost mid-nap wasn't observed until the nap ended (up to the 15s cap). Auditing every await the session makes while holding a live lease for the same property turned up a second site — the graceful drain (pre-existing on master), which could keep draining leaseless up to drain_timeout. Both now race the heartbeat handle, so lease-loss detection is bounded by the keepalive round alone regardless of supervisor state, and both sites have regression tests that fail on the old code (a lease revoked mid-nap or mid-drain must self-fence within seconds, not when the await ends).

rust/personhog-coordination/src/util.rs
nickbest-ph: this is a bit in tension with my other comment around the backoffs/sleeps, but fencing here buys no safety  if there is TTL runway remaining because the coordinator can't have reassigned anything. and the fencing is at the cost of availability

z0br0wn: updated!

rust/personhog-coordination/src/pod.rs
```diff
@@ -228,118 +243,322 @@ impl PodHandle {
...
+    }
+
+    /// Record one supervisor-level failure (a bootstrap, an attempt, or
+    /// a lost lease). Returns `false` when the consecutive budget is
+    /// exhausted; a run that was healthy for a while resets the count
+    /// first, so the budget bounds crash loops rather than lifetime
+    /// failures.
+    fn note_run_failure(&self, consecutive: &mut u32, started: Instant, err: &Error) -> bool {
+        const HEALTHY_ATTEMPT: Duration = Duration::from_secs(60);
+        if started.elapsed() >= HEALTHY_ATTEMPT {
```
nickbest-ph: this consecutive failure counter resets at a threshold (60s) that is the same as two of the slow failure detectors:
- a watch loop runs a reconcile pass every 5s and tolerates 12 consecutive failures (60s)
- participant_stall_threshold on the router watchdog is 60s

doesn't seem right

maybe a better reset condition is on progress rather than elapsed time?

z0br0wn: updated, and made some other fixes upon closer review

nickbest-ph (approved):

PR #74502, opened 2026-07-29
feat(personhog): warm partitions in parallel during handoffs

z0br0wn:
## Problem

The pod's coordination watch loop converged partitions one at a time: every watch event and every reconcile pass awaited each partition's convergence inline. Convergence to Acquiring includes warming the partition's cache from Kafka, so when a deploy moved several partitions onto the same surviving pod, each warm queued behind the last. The visible cost is in handoff latency: warming-phase duration scales with queue depth × single-warm time rather than single-warm time, and since routers stash writes for a partition until its handoff completes, stash waits inherit that same tail. In our dev chaos bed the warming-phase p99 sat near the stash-expiry ceiling during deploys while the actual per-partition warm work was an order of magnitude smaller.

## Changes

- Single-flight per partition, concurrent across partitions. watch_handoff_loop now drives convergence futures through a FuturesUnordered polled as a select arm. An in_flight set guarantees at most one convergence per partition (the no-interleaving invariant apply relies on is unchanged), and a pending map coalesces signals that arrive mid-convergence — converge reads fresh durable state, so one re-run absorbs any number of coalesced signals.
- One dispatch path for events, startup, and the reconcile tick. reconcile_all became involved_partitions (derives the involved set and watch-anchor revision only); startup seeds the loop and the tick re-dispatches through the same lanes. Cold-restart recovery therefore warms in parallel too.
- Bounded warm concurrency. New PodConfig::warm_concurrency (default 4) backs a semaphore acquired only at the two warm sites in apply. Drains, releases, and acks never queue behind warm pressure — fence latency stays prompt regardless of how many warms are running.
- Error dispositions preserved. Each dispatch carries a Trigger: event-driven (and startup-seed) convergence failures remain fatal to the run; reconcile-triggered failures remain under the reconcile failure budget. Budget accounting is judged per tick window — any reconcile-convergence failure in a window counts one against the budget, an all-success window resets it — which keeps the budget at the pass granularity it was sized for while ensuring a single persistently-failing partition still exhausts it.
- Cancellation semantics unchanged by construction. Convergence futures live inside the loop future; dropping the loop drops them, exactly as dropping the old inline .await did. No spawned tasks, no join bookkeeping.

Three new tests in protocol_hardening.rs, backed by a new GatedWarmHandler whose warms park on per-partition gates and record concurrency high-water marks:

  - concurrent_inbound_handoffs_warm_in_parallel — with one partition's warm parked, another partition's handoff must still warm and ack; fails by construction under serialized convergence (the regression no existing test caught).
  - convergences_for_one_partition_never_interleave — a new handoff era arriving while that partition's warm is parked must coalesce behind it, then re-warm; asserts same-partition concurrency never exceeded 1 (pins the single-flight invariant if the in_flight bookkeeping is ever removed).
  - concurrent_warms_are_bounded_by_warm_concurrency — bound 2, four parked inbound handoffs, exactly two warms enter the handler (pins the semaphore).

All existing coordination, leader, and router suites pass unchanged, and the e2e harness gates pass locally — including kill + scale-up, which produces exactly the simultaneous-inbound-handoff burst this PR parallelizes.

## 🤖 Agent context

Claude Code session. Design was reviewed before implementation: the key decisions were keeping convergence futures inside the loop (cancel-by-drop parity, no task lifecycle) rather than spawned workers; a persistent-worker-per-partition design was rejected because the reconcile tick would respawn finished workers every pass; and per-converge budget counting was rejected in favor of per-tick-window judgment because one bad tick across 8 partitions would otherwise burn 8 of the 12-failure budget that today costs one. /writing-tests was invoked for the test additions.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -8465,6 +8465,7 @@ dependencies = [
  "assignment-coordination",
  "async-trait",
  "etcd-client",
+ "futures",
  "k8s-awareness",
  "k8s-openapi",
  "kube",
--- a/rust/personhog-coordination/Cargo.toml
+++ b/rust/personhog-coordination/Cargo.toml
@@ -8,6 +8,7 @@ publish = false
 assignment-coordination = { path = "../common/assignment-coordination" }
 async-trait = { workspace = true }
 etcd-client = "0.18"
+futures = { workspace = true }
 k8s-awareness = { path = "../common/k8s-awareness" }
 metrics = { workspace = true }
 serde = { workspace = true }
--- a/rust/personhog-coordination/src/pod.rs
+++ b/rust/personhog-coordination/src/pod.rs
@@ -4,8 +4,10 @@ use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
 use etcd_client::{EventType, WatchStream};
+use futures::future::BoxFuture;
+use futures::stream::{FuturesUnordered, StreamExt};
 use metrics::{counter, gauge, histogram};
-use tokio::sync::{Mutex, Notify};
+use tokio::sync::{Mutex, Notify, Semaphore, SemaphorePermit};
 use tokio_util::sync::CancellationToken;
 
 use assignment_coordination::store::parse_watch_value;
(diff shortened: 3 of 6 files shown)


eli-r-ph (approved):
makes sense, changes LGTM

PR #74519, opened 2026-07-29
fix(conversations): strip commonmark escapes from outbound plain text

xljones:
## Problem

The rich editor → Slack conversion was over-escaping the message text fallback, so Slack messages showed stray backslashes before ordinary punctuation (`\!`, `\.`, `\-`, `\#`, `\(`, etc.). The same escapes reach customers by email: the `text/plain` part of outbound support replies, and the message previews that notification templates render as plain text.

## Changes

Rich content is serialized to markdown via `_escape_markdown`, which backslash-escapes every CommonMark special char. Slack mrkdwn, plain-text email parts, and notification previews don't read CommonMark, so the escapes leaked through verbatim.

- Slack: un-escape at the end of `content_to_slack_mrkdwn`, after the mrkdwn rewrites run. Code and Slack control chars stay protected.
- Email: added `rich_content_to_email_payload`, the counterpart of `rich_content_to_slack_payload`. It returns the `text/plain` part without the escapes plus the `text/html` alternative, and messages without rich content now render through the same HTML shell instead of one bare escaped paragraph.
- Previews: the `message_content` event property and the denormalized ticket message preview are read as plain text downstream, so they drop the escapes too. Existing rows and events keep what they were written with.

Escapes are stripped through one shared helper that leaves code spans and code blocks literal. Markdown syntax itself (`**bold**`, `[label](url)`) is untouched in the plain-text part — only the escapes go.

Added email payload cases to `test_formatting.py` (text part loses the escapes, code keeps its backslashes, the fallback path is clean in both parts and still HTML-escapes), plus one preview case each in `test_signals.py` and `test_events.py` for the two surfaces notification templates read. Locally: `test_formatting.py`, `test_signals.py`, `test_events.py`, `test_email_endpoints.py`, `test_tickets.py`, `test_mailgun.py`, `test_slack_image_sync.py` all pass, and repo-wide `mypy` is clean.

## 🤖 Agent context

Investigated from a Slack support thread, then extended after the same effect showed up in an outbound email. Confirmed in project data that the escapes are present on the great majority of recent outbound message events, which is what the plain-text notification template renders. Root-caused the double-transform (escape-then-render without un-escape) in `products/conversations/backend/formatting.py`. Chose to un-escape at the destination boundaries rather than skip escaping upfront, which the intermediate markdown step still relies on for Slack blocks, GitHub comments, and round-tripping.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C08S2L0S5MZ/p1785258020763879?thread_ts=1785258020.763879&cid=C08S2L0S5MZ)*

### DIFF
--- a/products/conversations/backend/formatting.py
+++ b/products/conversations/backend/formatting.py
@@ -28,6 +28,9 @@ _RE_SINGLE_NEWLINE = re.compile(r"(?<!\n)\n(?!\n)")
 _RE_MD_ESCAPE = re.compile(r"([\\`*_{}\[\]()#+\-.!|])")
 _RE_MD_UNESCAPE = re.compile(r"\\([\\`*_{}\[\]()#+\-.!|])")
 _RE_ALT_ESCAPE = re.compile(r"([\\\]])")
+# Fenced blocks (```...```) or simple inline spans (`code`). Their contents are literal
+# and must skip mrkdwn rewriting and backslash unescaping.
+_RE_CODE_SEGMENT = re.compile(r"```[\s\S]*?```|`[^`\n]*`")
 _RE_SLACK_EMOJI = re.compile(r":([a-z0-9_+\-]+):")
 
 
@@ -159,12 +162,29 @@ def strip_slack_user_mentions(text: str) -> str:
     return _RE_SLACK_USER_MENTION.sub("", text)
 
 
+def _escape_slack_control_chars(text: str) -> str:
+    """Neutralize Slack's control characters so user text can't inject mentions/links/broadcasts."""
+    return text.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
+
+
 def content_to_slack_mrkdwn(content: str) -> str:
     """Convert markdown comment content to Slack mrkdwn text."""
     if not content:
         return ""
 
-    text = content
+    # Pull code out first: its contents are literal, so they must not be rewritten as
+    # mrkdwn, have escapes stripped, or have their `<`/`>`/`&` treated as user injection.
+    code_segments: list[str] = []
+
+    def capture_code(match: re.Match) -> str:
+        code_segments.append(match.group(0))
+        return f"\x00CODE{len(code_segments) - 1}\x00"
+
+    text = _RE_CODE_SEGMENT.sub(capture_code, content)
+
+    # Escape control chars in user text before the conversions below emit their own trusted
+    # `<...>` link tokens — otherwise a literal `<!channel>` would become an active broadcast.
+    text = _escape_slack_control_chars(text)
 
     text = _RE_MD_IMAGE.sub(r"<\2|\1>", text)
 
@@ -186,6 +206,9 @@ def content_to_slack_mrkdwn(content: str) -> str:
     # backslash escapes left over from serialization — Slack mrkdwn doesn't use them and
     # would otherwise show them literally (e.g. "world\!").
     text = _RE_MD_UNESCAPE.sub(r"\1", text)
+
+    for index, value in enumerate(code_segments):
+        text = text.replace(f"\x00CODE{index}\x00", _escape_slack_control_chars(value))
 
     def resolve_mention(match: re.Match) -> str:
         uuid_str = match.group(1)
--- a/products/conversations/backend/formatting.py
+++ b/products/conversations/backend/formatting.py
@@ -26,6 +26,7 @@ _RE_MD_LINK = re.compile(r"\[([^\]]+)\]\(([^)]+)\)")
 _RE_MD_MENTION = re.compile(r"@member:([a-f0-9-]+)")
 _RE_SINGLE_NEWLINE = re.compile(r"(?<!\n)\n(?!\n)")
 _RE_MD_ESCAPE = re.compile(r"([\\`*_{}\[\]()#+\-.!|])")
+_RE_MD_UNESCAPE = re.compile(r"\\([\\`*_{}\[\]()#+\-.!|])")
 _RE_ALT_ESCAPE = re.compile(r"([\\\]])")
 _RE_SLACK_EMOJI = re.compile(r":([a-z0-9_+\-]+):")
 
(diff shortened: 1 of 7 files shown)


products/conversations/backend/formatting.py
```diff
@@ -181,6 +182,11 @@ def capture_bold_italic(match: re.Match) -> str:
     for index, value in enumerate(bold_italic_matches):
         text = text.replace(f"\x00BI{index}\x00", f"*_{value}_*")
 
+    # Markdown syntax has been rewritten to Slack mrkdwn above, so drop the CommonMark
+    # backslash escapes left over from serialization — Slack mrkdwn doesn't use them and
+    # would otherwise show them literally (e.g. "world\!").
+    text = _RE_MD_UNESCAPE.sub(r"\1", text)
```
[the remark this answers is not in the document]
xljones: Good catch — fixed in 0543ad0. Code spans/fences are now stashed out before any mrkdwn rewriting or backslash unescaping and restored verbatim, so a code block like `C:\Projects\*` keeps its backslashes. Added `test_outbound_code_block_preserves_literal_backslashes` to lock it in.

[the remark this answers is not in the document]
xljones: Confirmed and fixed in 0543ad0. The unescape had re-armed Slack control syntax that the old CommonMark escaping happened to defang. `content_to_slack_mrkdwn` now HTML-escapes `<`/`>`/`&` in user text before the link/image conversions emit their own trusted `<...>` tokens (the same approach as `escape_slack_mrkdwn` in the signals module), so a literal `<!channel>` becomes `&lt;!channel&gt;` while converter-generated links stay live. Added `test_outbound_text_fallback_neutralizes_slack_control_chars` covering `<!channel>`, `<@U…>`, and a fake `<url|label>`.

products/conversations/backend/formatting.py
```diff
@@ -181,6 +202,14 @@ def capture_bold_italic(match: re.Match) -> str:
...
```
veryayskiy, line 208: unescape runs after the rewrites, but _RE_MD_ITALIC matches across backslashes, so \*not bold\* comes out as _not bold_ — literal asterisks render as italics. Stash \X sequences like you stash code, restore them unescaped at the end.

products/conversations/backend/formatting.py
```diff
@@ -26,7 +26,11 @@
 _RE_MD_MENTION = re.compile(r"@member:([a-f0-9-]+)")
 _RE_SINGLE_NEWLINE = re.compile(r"(?<!\n)\n(?!\n)")
 _RE_MD_ESCAPE = re.compile(r"([\\`*_{}\[\]()#+\-.!|])")
+_RE_MD_UNESCAPE = re.compile(r"\\([\\`*_{}\[\]()#+\-.!|])")
 _RE_ALT_ESCAPE = re.compile(r"([\\\]])")
+# Fenced blocks (```...```) or simple inline spans (`code`). Their contents are literal
+# and must skip mrkdwn rewriting and backslash unescaping.
+_RE_CODE_SEGMENT = re.compile(r"```[\s\S]*?```|`[^`\n]*`")
```
veryayskiy, line 33: Escaped backticks aren't accounted for, so in a \ b <@U1> *x*`` the escaped backtick pairs with the real code opener — the real code span loses protection (x→x) and the `` still leaks. Same fix as above.

veryayskiy:
in description documents only the unescape pass and "3 parameterized cases"; the diff also adds &/</> entity escaping (formatting.py:165-167, applied at formatting.py:187 and 211) to every outbound Slack message including the tasks.py:393 path, plus 7 test cases.

xljones:
Note to self, we should consider escaping slack `:emoji:` too.

products/conversations/backend/formatting.py
xljones: Confirmed — reproduced it: `*not bold*` came out as `_not bold_`, because `_RE_MD_ITALIC` matched across the backslash and the unescape then removed the leftover slash.

Not fixed here, because master already fixes it. This module moved to `posthog/comment/formatting.py` and was rewritten while this PR sat stale, and the rewrite takes exactly the shape you asked for: escapes are masked before the syntax conversions and restored after. Bold is stashed the same way, which also fixes a separate bold-renders-as-italic bug this branch had.

Closing this PR as superseded. Thanks for the catch — the same reasoning carries into #96134, which fixes what master still gets wrong.

products/conversations/backend/formatting.py
xljones: Confirmed, and this one still bites on master. Reproduced on this branch: `` a ` b <@U1> `x` `` left the backslash in place, dropped the code span's protection, and leaked a stray backtick.

Master masks escaped characters but has no code handling at all, so nothing mis-pairs there yet — the defect reappears the moment code is masked. #96134 does mask code, and orders the alternation exactly as you suggested: the escape branch is tried first, so an escaped backtick can never open a span. That PR also fixes the related half you and Greptile both pointed at — a code fence currently loses the author's own backslashes (`C:\Projects\*` reaches Slack as `C:\Projects*`).

xljones:
Closing as superseded.

`products/conversations/backend/formatting.py` no longer exists. The module moved to `posthog/comment/formatting.py` and was rewritten there while this PR was stale, which is also why the branch conflicts — the file moved, rather than a line clashing.

Master already fixes every Slack-side defect raised on this PR: the CommonMark escapes in the text fallback, the literal asterisks reading as italics, the escaped backtick capturing a code span, and the control-character injection. Rebasing would put a weaker implementation back on top of code master already handles.

What master still gets wrong is in [#96134](https://github.com/PostHog/posthog/pull/96134):

- a code block loses the author's own backslashes in the Slack fallback (`C:\Projects\*` arrives as `C:\Projects*`) — the defect Greptile flagged on this PR, still live
- `message_content` and `last_message_text` still store escaped text, and both are read as plain text

The plain-text email body, which prompted the later half of this PR, is fixed at the serializer in [#94745](https://github.com/PostHog/posthog/pull/94745).

PR #74535, opened 2026-07-29
feat(data-warehouse): authenticate managed warehouse sources with the org root credential

EDsCODE:
## Problem

The auto-provisioned managed warehouse `ExternalDataSource` authenticates as a per-project reader whose schema allowlist must be configured per team through a control-plane handshake. As a result the warehouse sidebar and SQL editor only show the calling team's own namespaces.

**Why:** the managed warehouse is an org-level resource; the connection should see every schema in the org's duckgres warehouse by default so everything shows up without per-team configuration (access controls are deliberately deferred to a follow-up).

## Changes

All in `products/data_warehouse/backend/`. Duckgres needs no changes (root bypasses `AllowedSchemas`).

- **`ensure_managed_warehouse_direct_source`** builds `job_inputs` from the server's root credentials (`DuckgresServer.username`/.`password`), sets `connection_metadata.credential_kind = "org_root"` and `direct_query_enabled = True` immediately. The `configure_project_reader` handshake, its second transaction, and its `RuntimeError` retry path are gone.
- **`project_reader_namespaces`** is deleted; its call site in `reconcile_managed_warehouse_tables` now uses an `internal_schemas()` denylist (`pg_catalog`, `information_schema`, `pg_toast`, `system`). This is sidebar hygiene, not permissioning, and stays one function so the later per-schema visibility control has a single place to plug into.
- **`_ensure_managed_source_locked`'s catalog-purge guard** accepts `org_root` alongside `project_reader`; without that it would fire on every sweep against migrated sources, deleting and rediscovering the catalog in a loop. A `project_reader` source keeps its already-bounded catalog and only swaps credential; metadata drops the obsolete `reader_configured` flag.
- **`_membership_table_suffix`** is deleted with the handshake. Its `ValueError` refusal for legacy-shared teams goes away, so those grandfathered teams gain a working connection.
- **Root password rotation**`update_managed_warehouse_root_password` now fans the rotation out to every managed source of the org in the same transaction. `_source_config` snapshots the password into each team's `job_inputs`, so updating only the server row would leave every source failing silently after a rotation.

> [!WARNING]
> Known gap (deliberate): any member of any team in an organization gets read and write access to every other team's warehouse tables via the SQL editor; root carries no `ReadOnly` policy. Accepted as a temporary posture. The follow-up is per-schema visibility controls or an org-scoped read-only access mode in duckgres, which would restore the write boundary without reintroducing per-team configuration.

Row growth: each team's source now registers the whole org catalog, so `DataWarehouseTable`/`ExternalDataSchema` rows scale with teams times total tables instead of total tables. Accepted for now; revisit if a large org is onboarded.

I could not run pytest in this environment (no Postgres/ClickHouse/Redis services available); verification was compile checks plus ruff lint/format on all touched files. Behavior covered by updated unit tests in `products/data_warehouse/backend/tests/test_managed_warehouse_connection.py`:

- root credentials land in `job_inputs` with `credential_kind="org_root"` and `direct_query_enabled` set immediately (catches a regression to reader-style deferred enablement)
- the denylist filters only engine-internal schemas, and a cross-team table is registered and queryable through HogQL on a team's source (catches accidental reintroduction of the per-team allowlist)
- migrating a `project_reader` source swaps the credential without purging its catalog, while a pre-reader (unknown-kind) source still purges (catches the purge-branch regression that would delete/rediscover the catalog every sweep)
- password rotation propagates to every managed source in the org in one transaction, skips tombstoned sources, and a revived source picks up the rotated password (catches the silent-stale-password failure)
- legacy-shared and no-CP-membership teams now get connections (catches reintroduction of the `_membership_table_suffix` refusal)

Removed the `configure_project_reader`/`project_reader_namespaces` adapter tests with the functions themselves.

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

`skip-inkeep-docs` (internal backend credential change, no user-facing docs impact).

## 🤖 Agent context

Agent: Claude (PostHog Code session). No repo-provided or public skills were invoked. The change follows an approved design doc (managed warehouse connection on the org root credential, 2026-07-29) that was supplied as the implementation spec; the four code changes plus the rotation fan-out fix map 1:1 onto its decision table.

Notable interpretation calls I made: the catalog-purge guard now exempts `project_reader`-kind sources too (their catalogs were discovered under the reader boundary, so only the credential needs swapping; purging them on every migration sweep would churn the catalog for every existing org). Deleted duckgres presentation-adapter tests were removed with the functions they covered rather than rewritten, since nothing else calls those endpoints. Full pytest run against Postgres/ClickHouse is still needed before undrafting.

---

### DIFF
--- a/products/data_warehouse/backend/managed_warehouse_connection.py
+++ b/products/data_warehouse/backend/managed_warehouse_connection.py
@@ -1,26 +1,27 @@
-"""Expose a managed Duckgres warehouse in the SQL editor as a read-only Postgres connection.
+"""Expose a managed Duckgres warehouse in the SQL editor as a Postgres connection.
 
 Each member team gets a Postgres ``ExternalDataSource`` pointed at the organization's
-``DuckgresServer``. Duckgres issues a distinct database login for each project and enforces
-read-only access to the project's schemas, so both HogQL and raw SQL stay inside that boundary.
-Setup happens in two steps:
+``DuckgresServer``, authenticated with the server's org root credential. Duckgres root
+sees every schema in the warehouse, so the connection discovers and exposes the whole
+org catalog without per-team namespace configuration — no control-plane handshake is
+involved in setting it up. Setup happens in two steps:
 
-1. ``ensure_managed_warehouse_direct_source`` creates the initially empty source row when a team
+1. ``ensure_managed_warehouse_direct_source`` creates the source row when a team
    joins, so the connection appears immediately.
-2. ``reconcile_managed_warehouse_tables`` runs once the warehouse is ready and records every
-   table in the project's event/person, data-import, team, and modeled-data namespaces.
+2. ``reconcile_managed_warehouse_tables`` runs once the warehouse is ready and records
+   every non-internal schema/table the root credential can see.
 
-This bypasses the user-facing create endpoint because the managed host is internal infrastructure
-and is not reachable for live schema validation during provisioning.
+This bypasses the user-facing create endpoint because the managed host is internal
+infrastructure and is not reachable for live schema validation during provisioning.
 
-Lifecycle is org-level only: ``soft_delete_managed_warehouse_sources`` handles deprovisioning of
-the whole warehouse. There is no per-team offboarding flow yet — a single team leaving keeps its
-connection until the org deprovisions (revisit if per-team removal becomes a product flow).
+Lifecycle is org-level only: ``soft_delete_managed_warehouse_sources`` handles
+deprovisioning of the whole warehouse. There is no per-team offboarding flow yet —
+a single team leaving keeps its connection until the org deprovisions (revisit if
+per-team removal becomes a product flow).
 """
 
 from __future__ import annotations
 
-import secrets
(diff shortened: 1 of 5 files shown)


benben (approved):

jghoman (approved):

EDsCODE:
Bot review disposition:

- **Cross-project access via root credential (veria-ai / greptile):** deliberate and documented in the PR description's "Known gap" — an approved posture change from the design this PR implements, with a follow-up planned (per-schema visibility or an org-scoped read-only mode in duckgres). Not addressed here.
- **Legacy-shared teams unscheduled in the periodic sweep (greptile):** valid — fixed in bfaf418c. The sweep's cp_table_suffix skip predates root-backed sources; it now schedules every backfill-enabled row, and the test asserts the legacy row is scheduled.
- **Migrated project_reader catalogs stay team-bounded until a sweep (greptile):** expected convergence behavior — the periodic reconcile widens the catalog on its next run; not a defect.

PR #74539, opened 2026-07-29
fix(experiments): emit metric error events from the client

andehen:
## Problem

We can only classify a fraction of the experiment metric failures users actually hit. Looking at one week of a high-usage project, the `experiment results refresh completed` events counted roughly 890 errored metrics while only 105 `experiment metric error` events existed for the same window. That is an 8x gap, and it lands right where we need the data most: this project's manual-refresh p90 is over two minutes against 6-8s for everyone else, with a visible error spike on one afternoon.

Every fix on the table for that (raising the client metric concurrency limit, reshaping funnel queries, widening the recalculation rollout) is a bet on *why* metrics fail. Raising client concurrency against an already-saturated ClickHouse budget would make things worse, not better. So the telemetry gap has to close first.

The cause is architectural, not a missing call. `_emit_runner_terminal_error_event` only fires when an exception escapes `ExperimentQueryRunner._calculate()`, which is the single method wrapped by `@experiment_error_handler`. On the async path most user-visible failures never get there:

| Failure class | Emitted before | Why not |
| --- | --- | --- |
| Exception inside `_calculate()` | yes | n/a |
| `ConcurrencyLimitExceeded` from the four rate limiters | no | raised in `_call_with_rate_limits`, outside `calculate()` |
| `ConcurrencyLimitExceeded` from `@limit_concurrency` on the Celery task | no | raised before the runner exists |
| Celery retries exhausted on `ClickHouseAtCapacity` | no | terminal state written by `_process_query_task_failure`, a bare `on_failure` hook with no runner |
| Client-side poll timeout after ~10min | no | synthetic frontend error, the server never saw a terminal failure |
| Actors / PersonModal drilldown | no | `to_actors_query()` is undecorated |

There is skew the other way too: intermediate `ClickHouseAtCapacity` retries each pass through `_calculate()` and emit, so today's `rate_limited` counts mix non-terminal attempts with real failures.

On the async path the backend cannot know when a metric has terminally failed from the user's point of view. Only the client can.

## Changes

The frontend catch block becomes the single source of truth for failure volume, tagged `mechanism: "client"`. One emit site covers all six rows above, including the three the backend structurally cannot see.

- New `reportExperimentMetricErrored` in `eventUsageLogic`, emitting `experiment metric error` with the same property names the backend uses so the two emitters stay joinable by `query_id`.
- `classifyExperimentMetricError` mirrors the backend's `classify_experiment_query_error` taxonomy. It lives next to the listener rather than at the call sites, because both experiment logics already duplicate `parseMetricErrorDetail` and a third copy was not worth it.
- `experiment_error_handler` now falls back to the taxonomy value as the DRF error code. That code already flows to the client through `query_status.error_code`, so the client trusts it over any inference from the HTTP status.

> [!NOTE]
> Counting semantics change, and this is the point of the PR. Count `mechanism = "client"` for how many failures users saw; that number should reconcile with `sum(errored_count)` on `experiment results refresh completed`. Join to `mechanism = "direct"` for the typed cause. Do not sum `direct` for volume, it double-counts capacity retries. That skew predates this PR; the fix is to stop counting that emitter, not to change it.

Deliberately out of scope: reworking `_process_query_task_failure` or the rate limiters to carry a runner. That is generic query infrastructure with a much larger blast radius, and the client event already covers those cases.

No UI changes, so no screenshots.

Automated, all run locally by the agent:

- New `classifyExperimentMetricError.test.ts`, 7 parameterized cases, pass.
- `src/scenes/experiments` plus `eventUsageLogic.test.ts`: 35 suites, 730 tests, pass.
- `products/experiments/backend/hogql_queries/test/test_error_handling.py`: 23 pass.
- `ruff check` and `ruff format --check` clean on the changed Python file.
- `typescript:check` reports no errors in any file this PR touches. The full run does fail on pre-existing `@posthog/quill-primitives` resolution errors under `products/*`, unrelated to this diff and present before it.

The new test earns its place: the client taxonomy has to stay in lockstep with the backend enum, and there is nothing that would catch a desync. Drop the error-code passthrough branch and every backend-classified failure silently reports `server_error`, which quietly poisons the exact analysis this PR exists to enable. It is a pure function, so the test is a cheap `it.each`.

I did not manually exercise a failing metric in a running app. The end-to-end check is the reconciliation query below, which only becomes meaningful after deploy.

<details>
<summary>Post-deploy verification query</summary>

```sql
SELECT
  countIf(event = 'experiment metric error' AND properties.mechanism = 'client') AS client_error_events,
  sumIf(toFloat(properties.errored_count), event = 'experiment results refresh completed') AS counted_errors
FROM events
WHERE timestamp >= now() - INTERVAL 1 DAY
```

These two should now track each other. Once they do, the `error_type` breakdown settles whether the slow project's error spike was concurrency exhaustion, and only then is it worth touching `METRIC_QUERY_CONCURRENCY_LIMIT` or the funnel queries.
</details>

## Automatic notifications

- [ ] Publish to changelog?

## Docs update

Not needed, this is internal telemetry with no user-facing surface.

## 🤖 Agent context

I asked Claude (PostHog Code, Opus) to analyze how experiment loading has felt for one customer over the last week using the `experiment ...` events, and to look for root causes. It pulled the data through the PostHog MCP and came back with a clear structural story (that project runs ~15.7 metrics per experiment against 4.2 for everyone else, into a fixed client concurrency of 10, with funnels carrying the latency tail) but flagged that the error-rate half of the analysis was unusable because most failures are never classified. I told it to fix that first rather than act on the slower-query hypotheses.

The interesting part was the diagnosis. The obvious read is "the frontend forgot to emit an event", and there is even a comment in `experimentLogic.tsx` asserting the backend owns it. Tracing it showed the comment encodes an assumption that does not hold on the async path, and that the missing classes are precisely the ones the backend cannot observe. That reframed the fix from "add a call" to "move ownership of terminality to the client and keep the backend emit as enrichment".

Rejected along the way: making the backend emit from `_process_query_task_failure` and from the rate limiters. Both would work, both touch shared query infrastructure used well beyond experiments, and neither covers the client-side poll timeout. The client emit covers all of them in one place.

Skills invoked: `/writing-tests` (gating the one new test) and `/writing-code-comments`.

---


andehen:
Both findings are valid, thanks. Fixed in 639276f180f.

Verified the mechanism before fixing: `ApiError` sets `this.detail = data?.detail` and passes the same `data.detail` as the `super(message)` argument, so on the diagnostics branch the fallback was provably re-emitting the JSON it had just parsed. `experimentMetricErrorMessage` now names the reasons the blob actually sets (`no-exposures, no-test-variant`) rather than storing the dump, and returns null when nothing is set.

The review under-called the impact, though. That blob only ever comes from the no-results validators, and those mean the experiment has no usable data yet, not that the metric config is bad. Both raise sites are a 400, so they were landing in `validation_error` — which means `error_type` would have mixed "your metric is broken" together with "this experiment just started", in the one dimension this telemetry exists to produce. The backend taxonomy already has `insufficient_data` for exactly this case. So the classifier now routes both the `no-results` code and the bare-`invalid`-with-diagnostics variant there, keyed off `hasDiagnostics` so it covers both raise sites in each legacy runner.

Test is up to 13 cases covering both branches.

